PR #3, opened 2015-04-17
Add command 'ba' for breaking on memory access

thebarbershopper:
Mimic windbg's `ba` command to break on memory access.

`ba` contains two types:
* break on read/watch - `ba r deadbeef`
* break on watch - `ba w deadbeef`

### DIFF
--- a/pwndbg/commands/windbg.py
+++ b/pwndbg/commands/windbg.py
@@ -185,6 +185,22 @@ def bp(where):
 def bp(where):
     gdb.execute('break *%#x' % int(where))
 
+@pwndbg.commands.Command
+@pwndbg.commands.OnlyWhenRunning
+def ba(kind='r', where=None):
+    '''Break on memory access to address'''
+    commands = {'r': 'awatch',
+                'w': 'watch'}
+    if len(kind) > 1:
+        kind = kind[0]
+
+    command = commands.get(kind, None)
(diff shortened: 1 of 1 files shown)


pwndbg/commands/windbg.py
```diff
@@ -185,6 +185,22 @@ def bp(where):
...
+                'w': 'watch'}
+    if len(kind) > 1:
+        kind = kind[0]
+
+    command = commands.get(kind, None)
+    if not command:
+        print("Type of access breakpoint not recognized: %s" % kind)
+        return
+
+    gdb.execute('%s *%#x' % (command, int(where, 16)))
```
zachriggle: `gdb.Breakpoint(...)` is better than `gdb.execute` when possible

pwndbg/commands/windbg.py
```diff
@@ -185,6 +185,22 @@ def bp(where):
 def bp(where):
     gdb.execute('break *%#x' % int(where))
 
+@pwndbg.commands.Command
```
zachriggle: Since you're using a `Command` instead of `ParsedCommand`, you'll have to pass the additional arguments to the parser yourself.

See `pwndbg.commands.fix` for a convenient routine to do this.

zachriggle:
Ping :)

PR #20, opened 2016-01-14
Add better features to search and searchmem function by resolving some issues.

P4N74:
While searching for `123456` in the the memory using either of `search` or `searchmem` as:
```
pwndbg> search 123456
pwndbg> search '123456'
```
it would convert the 123456 to integer and search for the corresponding hex values in memory. If the user want to search `123456` that was inserted into character array, it was not efficient to use the `search` or `searchmem` with their existing way of working.

Hence, I made this patch to solve this issue by letting the user to input what way he wants to search `123456` as. Hence I added a support for character search in such case which can be fulfilled as:
```
pwndbg> search/c 123456
pwndbg> search/c '123456'
```

Also the support to search directly for hex characters in memory is added in case user wants to search for any arbitrary hex value such as `\xf0\xf1\xf2\xf3` in memory which might have been injected into character array.
```
pwndbg> search/xc "f0f1f2f3"
pwndbg> search/xc "\xf0\xf1\xf2\xf3"
pwndbg> search/xc "\\xf0\\xf1\\xf2\\xf3"
```
The use of `\\xf0` is done because by default when the argument was send to function, gdb used to take `\` as escaping character and instead send `xf0xf1xf2xf3` for input `\xf0\xf1\xf2\xf3`. Hence if user is cautious and tries to send `\\xf0\\xf1\\xf2\\xf3` so that function gets parameter value as `\xf0\xf1\xf2\xf3`, it would be handled correctly.

### DIFF
--- a/pwndbg/commands/search.py
+++ b/pwndbg/commands/search.py
@@ -13,17 +13,31 @@ import pwndbg.vmmap
 
 @pwndbg.commands.Command
 @pwndbg.commands.OnlyWhenRunning
-def search(value):
+def search(searchtype, value=None):
     """
     Search memory for the specified value, provided
     either as a pointer-width integer, or a string.
 
     > search 0xdeadbeef
     > search "/bin/sh"
+    
+    To search 1234 in a character string instead of integer
+    > search/c 1234
+    
+    To search for characters using hex values in string
+    > search/xc f0f1f2f3
+    > search/xc \xf0\xf1\xf2\xf3
+    > search/xc \\xf0\\xf1\\xf2\\xf3
     """
+    
+    if value:
+        searchtype = searchtype[1:]
+    else:
+        value, searchtype = searchtype, value
+    
     hits = set()
 
-    for address in pwndbg.search.search(value):
+    for address in pwndbg.search.search(value, searchtype):
         if not address:
             continue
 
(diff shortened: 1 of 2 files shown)


zachriggle:
I like the idea, but I'd rather see it split into separate commands.  The method `pwndbg.search.search` should probably be modified to only take a `byte` array rather than trying to be intelligent about the search type.  Then the `pwndbg.commands.search.XXX` can perform the necessary translation.

P4N74:
Just to clarify, you mean that the `pwndbg.search.search` should only take argument as `byte` array that it will search for and `pwndbg.commands.search.XXX` would do the necessary translation. So it would be something like: `pwndbg.commands.search.search` calls `pwndbg.commands.search.XXX` with `searchtype` and `searchvalue` and then `pwndbg.commands.search.XXX` should do the translation and call `pwndbg.search.search` with that translated byte array. Am I right?

zachriggle:
Something along the lines.  Ultimately, the issue is that `pwndbg.search.search` attempts to be smart about identifying the type of data it is given.  This was a quick hack, a better implementation would just have it search for a byte sequence.

In order to translate user input (string, address, byte, word, dword, etc) into those byte arrays, new functions should be added to `pwndbg.commands.search`, for example `pwndbg.commands.search.searchd` or `pwndbg.commands.search.searchb`.  This would give us the `searchb` and `searchd` commands which would e.g. search for bytes or DWORDs.

P4N74:
I have update the `pwndbg.commands.search` and added `pwndbg.commands.search.searchb` and `pwndbg.commands.search.searchd` functions. While doing this I de-duplicated some code of `pwndbg.commands.search`.

PR #21, opened 2016-01-16
Correct the conversion of mem2chunk

P4N74:


P4N74:
I don't know if this file is needed or not since

P4N74:
According to [EGLIBC malloc implementation](http://www.eglibc.org/cgi-bin/viewvc.cgi/trunk/libc/malloc/malloc.c?view=markup,) mem2chunk is defined as `#define mem2chunk(mem) ((mchunkptr)((char*)(mem) - 2*SIZE_SZ))`

zachriggle:
Hah, thanks.  If you can't tell it's just a small stub, but I'll accept the fix :)

PR #22, opened 2016-01-17
Improved auxv entry parsing

saelo:
The previous code could crash if an entry was malformed or even execute arbitrary python code:

If e.g. the platform string (on the stack) was overwritten with something like " print('pwnd') " (note the whitespace), then gdb would produce the following line:
`15   AT_PLATFORM          String identifying platform    0x7fffffffe8a9 " print('pwnd') "...`
 and then parser would `eval("print('pwnd')")` :)

### DIFF
--- a/pwndbg/auxv.py
+++ b/pwndbg/auxv.py
@@ -1,4 +1,5 @@
 import os
+import re
 import sys
 
 import gdb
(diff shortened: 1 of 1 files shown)


zachriggle:
Hurray `eval`!

PR #23, opened 2016-03-14
show flags register in `context_regs`

br0ns:
Only X86 supported as of now.

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -56,7 +56,7 @@ def get_regs(*regs):
     result = []
 
     if not regs:
-        regs = pwndbg.regs.gpr + (pwndbg.regs.frame, pwndbg.regs.current.stack, pwndbg.regs.current.pc)
+        regs = pwndbg.regs.gpr + (pwndbg.regs.frame, pwndbg.regs.current.stack, pwndbg.regs.current.pc) + pwndbg.regs.flags
 
     changed = pwndbg.regs.changed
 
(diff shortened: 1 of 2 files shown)


zachriggle:
You can actually get away with just using the internal support built into GDB for the current architecture, rather than parsing everything out.  Again, only for x86 at the moment.

I need to figure out a way to set up some configuration for pwndbg, because the contents of `eflags` isn't something I (personally) want to see most of the time.

```
$ gdb /bin/sh
...
pwndbg> start
Temporary breakpoint 1 at 0x555555558e32
....
pwndbg> pi
>>> str(gdb.parse_and_eval('$eflags'))
'[ IF ]'
```

Gimme a few days to have a chance to figure this out, then I'll merge this in :)

br0ns:
Yes, I knew about GDB's support, but I wanted the flexibility of doing it
myself :).  Anyway, I got somewhat familiar with your code on the way, so
that's a plus.

I like having the flags shown all the time.  Time for some configuration
options?

BTW, I found a bug: try "set print elements 0".

--
br0ns

anthraxx:
@br0ns did you have any progress doing this? Else I would consider investing some time here

br0ns:
@anthraxx I'm back to PEDA for the moment, so that's a no.

zachriggle:
I've recently added support for [configuration points](https://github.com/pwndbg/pwndbg/blob/master/pwndbg/config.py), which was necessary for this feature.

anthraxx:
So you want this configurable, I see :-) default value enabled is ok, or you prefer other way around? Then I would work on a   PR

zachriggle:
Fixed with b0a2d5b38b8d79df5ed82aef35ff89bc0cfff9e9

zachriggle:
This is controlled by the (default off) flag `set-flags`.

Append to your `.gdbinit`: `set show-flags on`

PR #34, opened 2016-05-27
Fix .gdbinit modification line

kernel-sanders:
Your install instructions say to cd into the top level pwndbg directory and run ./setup.sh. This causes your last popd on line 27 to put you back there, and then line 32 adds an extra /pwndbg/ to the path of gdbinit.py so gdb can't find it on launch.


zachriggle:
Ah yep, this is an artifact of when I created the setup script.  Whoops!  Thanks :)

PR #36, opened 2016-05-30
Basic implementation of heap commands

gsingh93:
Implemented commands for printing heap chunks and fast bins.

### DIFF
--- a/pwndbg/color.py
+++ b/pwndbg/color.py
@@ -55,7 +55,7 @@ def get(address, text = None):
     if text is None and isinstance(address, (long, int)) and address > 255:
         text = hex(int(address))
     if text is None:
-        text = int(address)
+        text = str(int(address))
 
     if color == NORMAL:
         return text
@@ -71,4 +71,3 @@ def legend():
         UNDERLINE + 'RWX' + NORMAL,
         'RODATA'
     ))
-
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -11,13 +11,15 @@ PREV_INUSE = 1
 IS_MMAPED = 2
 NON_MAIN_ARENA = 4
 
+def value_from_type(type_name, addr):
+    gdb_type = pwndbg.typeinfo.load(type_name)
+    return gdb.Value(addr).cast(gdb_type.pointer()).dereference()
+
 def get_main_arena(addr=None):
     if addr == None:
         main_arena = gdb.lookup_symbol('main_arena')[0].value()
     else:
-        if isinstance(addr, int):
-            addr = hex(addr)
-        main_arena = gdb.parse_and_eval('(struct malloc_state)*' + addr)
+        main_arena = value_from_type('struct malloc_state', addr)
 
     if main_arena == None:
         print(red('Symbol \'main_arena\' not found. Try installing libc ' \
@@ -41,6 +43,9 @@ def get_heap_bounds():
 @pwndbg.commands.ParsedCommand
 @pwndbg.commands.OnlyWhenRunning
 def heap(addr=None):
+    """
+    Prints out all chunks in the main_arena, or the arena specified by `addr`.
+    """
     main_arena = get_main_arena(addr)
     if main_arena == None:
         return
@@ -71,6 +76,9 @@ def heap(addr=None):
 @pwndbg.commands.ParsedCommand
 @pwndbg.commands.OnlyWhenRunning
 def arena(addr=None):
+    """
+    Prints out the main arena or the arena at the specified by address.
+    """
     main_arena = get_main_arena(addr)
     if main_arena == None:
         return
@@ -80,6 +88,10 @@ def arena(addr=None):
 @pwndbg.commands.ParsedCommand
 @pwndbg.commands.OnlyWhenRunning
 def bins(addr=None):
+    """
+    Prints out the contents of the fastbins of the main arena or the arena
+    at the specified address.
+    """
     main_arena = get_main_arena(addr)
     if main_arena == None:
         return
@@ -87,22 +99,26 @@ def bins(addr=None):
     fastbins = main_arena['fastbinsY']
     bins = main_arena['bins']
 
-    size_t_size = gdb.lookup_type('size_t').sizeof
+    size_t_size = pwndbg.typeinfo.load('size_t').sizeof
     num_fastbins = int(fastbins.type.sizeof / fastbins.type.target().sizeof)
     num_bins = int(bins.type.sizeof / bins.type.target().sizeof)
     fd_field_offset = 2 * size_t_size
 
-    print(underline(yellow('Fastbins')))
+    print(underline(yellow('fastbins')))
     for i in range(num_fastbins):
         size = 2 * size_t_size * (i + 1)
         chain = pwndbg.chain.format(int(fastbins[i]), offset=fd_field_offset)
-        print(bold(size) + ': ' + chain)
+        print((bold(size) + ': ').ljust(13) + chain)
 
     # TODO: Print other bins
 
 @pwndbg.commands.ParsedCommand
 @pwndbg.commands.OnlyWhenRunning
 def top_chunk(addr=None):
+    """
+    Prints out the address of the top chunk of the main arena, or of the arena
+    at the specified address.
+    """
     main_arena = get_main_arena(addr)
     if main_arena == None:
         heap_start, heap_end = get_heap_bounds()
@@ -115,7 +131,7 @@ def top_chunk(addr=None):
         last_addr = None
         addr = heap_start
         while addr < heap_end:
-            chunk = gdb.parse_and_eval('(struct malloc_chunk)*' + hex(addr))
+            chunk = value_from_type('struct malloc_chunk', addr)
             size = int(chunk['size'])
 
             # Clear the bottom 3 bits
@@ -130,10 +146,13 @@ def top_chunk(addr=None):
 @pwndbg.commands.ParsedCommand
 @pwndbg.commands.OnlyWhenRunning
 def malloc_chunk(addr):
+    """
+    Prints out the malloc_chunk at the specified address.
+    """
     if not isinstance(addr, int):
         addr = int(addr)
 
-    chunk = gdb.parse_and_eval('(struct malloc_chunk)*' + hex(addr))
+    chunk = value_from_type('struct malloc_chunk', addr)
     size = int(chunk['size'])
     prev_inuse = (size & PREV_INUSE) == 1
     is_mmaped = (size & IS_MMAPED) == 1
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -53,8 +53,8 @@ def heap(addr=None):
     top = main_arena['top']
     last_remainder = main_arena['last_remainder']
 
-    print(bold('Top Chunk: ') + pwndbg.color.get(int(top)))
-    print(bold('Last Remainder: ') + pwndbg.color.get(int(last_remainder)))
+    print(bold('Top Chunk: ') + pwndbg.color.get(top))
+    print(bold('Last Remainder: ') + pwndbg.color.get(last_remainder))
     print()
 
     # Print out all chunks on the heap
@@ -96,7 +96,7 @@ def bins(addr=None):
     for i in range(num_fastbins):
         size = 2 * size_t_size * (i + 1)
         chain = pwndbg.chain.format(int(fastbins[i]), offset=fd_field_offset)
-        print(bold(str(size)) + ': ' + chain)
+        print(bold(size) + ': ' + chain)
 
     # TODO: Print other bins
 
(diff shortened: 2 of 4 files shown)


pwndbg/commands/heap.py
```diff
@@ -1,46 +1,127 @@
-#!/usr/bin/env python
```
zachriggle: Can you put the shebang and utf line back? It avoids Python2/3 issues in weird cases

pwndbg/commands/heap.py
```diff
@@ -1,46 +1,127 @@
...
-    gdb.execute('call brk(%i)' % n)
+from pwndbg.color import bold, yellow, red, underline
+
+PREV_INUSE = 1
+IS_MMAPED = 2
+NON_MAIN_ARENA = 4
+
+def get_main_arena(addr=None):
+    if addr == None:
+        main_arena = gdb.lookup_symbol('main_arena')[0].value()
```
zachriggle: I'd recommend using pwndbg.symbol.get if you just want the pointer

pwndbg/commands/heap.py
```diff
@@ -1,46 +1,127 @@
...
+    else:
+        if isinstance(addr, (long, int)):
+            addr = hex(addr)
+        main_arena = gdb.parse_and_eval('(struct malloc_state)*' + addr)
```
zachriggle: I'd recommend using lookup_type and then creating the object directly if it exists

pwndbg/commands/heap.py
```diff
@@ -1,46 +1,127 @@
...
-               help='Print more information')
-p.add_argument('--free', action='store_true',
-               help='Only show free slots')
-p.add_argument('address', type=int, default=0,
-               help='Heap allocation to display')
+        # Clear the bottom 3 bits
+        size &= ~7
+        addr += size
 
+@pwndbg.commands.Command
```
zachriggle: You'll want this to be a ParsesCommand type, unless you want the raw sting argument (eg "$rax")

pwndbg/commands/heap.py
```diff
@@ -1,46 +1,127 @@
...
+    main_arena = get_main_arena(addr)
+    if main_arena == None:
         return
 
+    fastbins = main_arena['fastbinsY']
+    bins = main_arena['bins']
+
+    num_fastbins = int(fastbins.type.sizeof / fastbins.type.target().sizeof)
+    num_bins = int(bins.type.sizeof / bins.type.target().sizeof)
+    fd_field_offset = 16
```
zachriggle: Can this be represented as a quanitity of longs or something similar?

pwndbg/commands/heap.py
```diff
@@ -1,46 +1,127 @@
...
+
+    print(underline(yellow('Fastbins')))
+    for i in range(num_fastbins):
```
zachriggle: Instead, or in addition to, fastbin index, can we also include the size?

pwndbg/commands/heap.py
```diff
@@ -1,46 +1,127 @@
...
+        chain = pwndbg.chain.format(int(fastbins[i]), offset=fd_field_offset)
+        print(bold(str(i)) + ': ' + chain)
+
+    # TODO: Print other bins
+
+@pwndbg.commands.Command
+@pwndbg.commands.OnlyWhenRunning
+def top_chunk(addr=None):
+    main_arena = get_main_arena(addr)
```
zachriggle: Is the top chunk always the beginning / end of the brk area? If so, we should be able to retrieve it directly via vmmap

pwndbg/commands/heap.py
gsingh93: Oops. Will do.

pwndbg/commands/heap.py
gsingh93: Yea, two pointers, good point.

pwndbg/commands/heap.py
gsingh93: It should be the last chunk in the sbrk area, but you don't know the size, so I don't know how you'd find it (I need to check is prev_size is set for the top chunk).

pwndbg/commands/heap.py
zachriggle: IIRC each bin corresponds to a particular allocation size (8/16/24/36 etc) in bytes

pwndbg/commands/heap.py
zachriggle: Can't one just walk each allocation linearly from the first (linear) allocation in the heap?

pwndbg/commands/heap.py
```diff
@@ -1,46 +1,127 @@
...
+    if main_arena == None:
+        return
+
+    top = main_arena['top']
+    print(top)
+
+@pwndbg.commands.Command
+@pwndbg.commands.OnlyWhenRunning
+def malloc_chunk(addr):
+    if isinstance(addr, (long, int)):
```
zachriggle: long is not Python3-compatible and should be avoided.  I think I added some bullshit hack to make it work, so it might not matter

pwndbg/commands/heap.py
```diff
@@ -1,46 +1,127 @@
...
+        addr = hex(addr)
+
+    gdb.lookup_type('struct malloc_chunk')
+    chunk = gdb.parse_and_eval('(struct malloc_chunk)*' + addr)
+    size = int(chunk['size'])
+    prev_inuse = (size & PREV_INUSE) == 1
+    is_mmaped = (size & IS_MMAPED) == 1
+    non_main_arena = (size & NON_MAIN_ARENA) == 1
+
+    header = bold(addr)
```
zachriggle: If recommend using the standard colorized for addresses, instead of or in addition to bold. This will make pointers which are outside of the heap more apparent.  pwndbg.color.get

zachriggle: I'd, colorizer

pwndbg/commands/heap.py
gsingh93: Oh yea, I can add that.

pwndbg/commands/heap.py
gsingh93: Yup, I just didn't want to do that. I'll do a fast return from the arena if we have symbols and walk all the chunks if we don't.

pwndbg/commands/heap.py
gsingh93: I want the struct Value, not the pointer.

pwndbg/color.py
```diff
@@ -53,7 +53,7 @@ def get(address, text = None):
     if text is None and isinstance(address, (long, int)) and address > 255:
         text = hex(int(address))
     if text is None:
-        text = int(address)
+        text = str(int(address))
```
gsingh93: I think not using `str` here was a bug.

The `address > 255` in the first if statement seems somewhat arbitrary. In my case, the `last_remainder` chunk starts of as `0` instead of `0x0` because of this.

pwndbg/commands/heap.py
```diff
@@ -1,46 +1,152 @@
...
 import pwndbg.commands
 
-@pwndbg.commands.OnlyWhenRunning
+from pwndbg.color import bold, yellow, red, underline
+
+PREV_INUSE = 1
+IS_MMAPED = 2
+NON_MAIN_ARENA = 4
+
+def get_main_arena(addr=None):
```
gsingh93: Should I be memoizing here?

pwndbg/commands/heap.py
```diff
@@ -1,46 +1,152 @@
...
+        main_arena = gdb.parse_and_eval('(struct malloc_state)*' + addr)
+
+    if main_arena == None:
+        print(red('Symbol \'main_arena\' not found. Try installing libc ' \
+                  'debugging symbols or specifying the main arena address ' \
+                  'and try again'))
+
+    return main_arena
+
+def get_heap_bounds():
```
gsingh93: Should I be memoizing here?

pwndbg/commands/heap.py
zachriggle: Commands shouldn't be memoized, library routines should.

pwndbg/commands/heap.py
zachriggle: Commands shouldn't be memoized, library routines should.

pwndbg/commands/heap.py
```diff
@@ -80,41 +87,59 @@ def bins(addr=None):
...
-
-    top = main_arena['top']
-    print(top)
+        heap_start, heap_end = get_heap_bounds()
+        if heap_start == None:
+            print(red('Could not find the heap'))
+            return
+
+        # If we don't know where the main_arena struct is, just iterate
+        # through all the heap objects until we hit the last one
```
zachriggle: I realized last night that this will only work in an uncorrupted heap.  Still useful to have if symbol data is missing, better than nothing.

pwndbg/commands/heap.py
```diff
@@ -80,41 +87,59 @@ def bins(addr=None):
...
+    else:
+        print(pwndbg.color.get(int(main_arena['top'])))
 
-@pwndbg.commands.Command
+@pwndbg.commands.ParsedCommand
 @pwndbg.commands.OnlyWhenRunning
 def malloc_chunk(addr):
-    if isinstance(addr, (long, int)):
-        addr = hex(addr)
+    if not isinstance(addr, int):
```
zachriggle: Not that it matters, but the `if` statement isn't necessary.  int(int(int(int(7)))) == 7.

gsingh93: From what I remember debugging last night, I think doing something like `malloc_chunk <addr>` makes `addr` a gdb.Value, which needs to be converted to an int to work with either hex, or pwndbg.color.get, or both.

zachriggle:
Revisiting structure creation;

Assuming that you can do this:

```
(gdb) p *(struct malloc_chunk*) 0xaddr
```

You can do the following:

```py
struct_malloc_chunk = pwndbg.typeinfo.load('struct malloc_chunk')
struct_malloc_chunk_pointer = struct_malloc_chunk.pointer()
addr = gdb.Value(pwndbg.regs.pc)
chunk = addr.cast(struct_malloc_chunk_pointer)
fd = chunk['fd']
print(hex(int(fd) & pwndbg.arch.ptrmask))
```

pwndbg/commands/heap.py
```diff
@@ -80,41 +87,59 @@ def bins(addr=None):
     fastbins = main_arena['fastbinsY']
     bins = main_arena['bins']
 
+    size_t_size = gdb.lookup_type('size_t').sizeof
```
zachriggle: `pwndbg.typeinfo.size_t.sizeof`

zachriggle: Actually, they're not `size_t`, they're pvoid.  This should use `pwndbg.typeinfo.pvoid.sizeof`

zachriggle: And in before "they're the same", see x32 ABI with 4-byte pointers and 8-byte size_t 😛

gsingh93: Where do you see a pointer being used? The two fields we're skipping are `prev_size` and `size`, which are both `INTERNAL_SIZE_T` which is just `size_t`.

zachriggle: Ah, I see that we're skipping over the `prevsize` and `size` field (size_t sized), I thought we were calculating an index into the `fastbinsY` array (pointer-sized).

pwndbg/commands/heap.py
zachriggle: I think that was true when you were calling `hex`, but now you're just calling `int`.  The code now is `if not isinstance(foo, int): foo = int(foo)`; the `if` is a noop.  In any case, it doesn't matter :)

pwndbg/commands/heap.py
```diff
@@ -27,31 +26,39 @@ def get_main_arena(addr=None):
...
     if heap_base == None:
         print(red('Could not find the heap'))
         return
 
     top = main_arena['top']
     last_remainder = main_arena['last_remainder']
 
-    print(bold('Top Chunk: ') + str(top))
-    print(bold('Last Remainder: ') + str(last_remainder))
+    print(bold('Top Chunk: ') + pwndbg.color.get(int(top)))
```
zachriggle: If if helps any, I just committed 0b86043 which auto-casts to `int`

pwndbg/commands/heap.py
```diff
@@ -80,41 +87,59 @@ def bins(addr=None):
...
     num_bins = int(bins.type.sizeof / bins.type.target().sizeof)
-    fd_field_offset = 16
+    fd_field_offset = 2 * size_t_size
 
     print(underline(yellow('Fastbins')))
     for i in range(num_fastbins):
+        size = 2 * size_t_size * (i + 1)
         chain = pwndbg.chain.format(int(fastbins[i]), offset=fd_field_offset)
-        print(bold(str(i)) + ': ' + chain)
+        print(bold(str(size)) + ': ' + chain)
```
zachriggle: I just added 2962dc9 which should save you all the casts to `str`

pwndbg/color.py
zachriggle: Agree on `str`, but I'd prefer to print decimal for small numbers.

Perhaps we can add a configuration point for this.  See `gdb.parameter` for more info.

gsingh93:
Updated. Because the dynamic type parsing is broken, this is still broken when you don't have symbols. I think that's fine for now, since installing debugging symbols is trivial (at least on Ubuntu).

zachriggle:
It turns out even with dynamic type generation, this doesn't work, since no
headers define these types :-(

zachriggle:
Though we could do the super ghetto stuff I used to do. Which would
*basically* just be to include a copy of dlmalloc headers, and use those.

PR #45, opened 2016-06-04
Added checksec command

gsingh93:
Don't merge until the GOT/PLT PR is merged, as that provides the `pwndbg.file.get_file` command.

We first try to call the binjitsu checksec command (which is named `checksec`), and if it doesn't exist (indicated by an exception being thrown), we attempt to call `checksec.sh`. Using an exception isn't the cleanest way to check for the existence of a file, so if you want me to do a proper check, let me know.


zachriggle:
Cherry-picked to save you the need to rebase

PR #48, opened 2016-06-06
remove re-introduced legacy import

anthraxx:
Seems the following commit: https://github.com/pwndbg/pwndbg/commit/1c6466d328a6fd53c39ec642f28aff1804aaaaa1#diff-6b7b09ae29c3f2feca1b2057a8b32da2 reintroduced the non existing elfheader command import, this break rinning pwndbg


zachriggle:
Fixed OOB with 3b512200717bd22878f918d413e2b96bc80b68b2 while we migrate to Travis CI which should prevent this kind of thing from happening 💖

PR #49, opened 2016-06-09
highlighting the current IP or line of the code/source section

anthraxx:
This helps visually distinguish the currently executed instruction and the according source line (if any is available).

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -141,6 +141,8 @@ def context_code():
 
     return banner + result
 
+pwndbg.config.Parameter('highlight-source', True, 'whether to highlight the closest source line')
+
 def context_source():
     try:
         symtab = gdb.selected_frame().find_sal().symtab
@@ -163,8 +165,16 @@ def context_source():
         if not source or closest_line <= 1:
             return []
 
+        # highlight the current code line
+        source_lines = source.splitlines()
+        if pwndbg.config.highlight_source:
+            for i in range(len(source_lines)):
+                if source_lines[i].startswith('%s\t' % closest_line):
+                    source_lines[i] = pwndbg.color.highlight(source_lines[i])
+                    break
+
         banner = [pwndbg.color.blue(pwndbg.ui.banner("code"))]
-        banner.extend(source.splitlines())
+        banner.extend(source_lines)
         return banner
     except:
         pass
--- a/pwndbg/commands/nearpc.py
+++ b/pwndbg/commands/nearpc.py
@@ -21,6 +21,8 @@ import pwndbg.ui
 import pwndbg.vmmap
 
 
+pwndbg.config.Parameter('highlight-pc', True, 'whether to highlight the current instruction')
+
 @pwndbg.commands.ParsedCommand
 @pwndbg.commands.OnlyWhenRunning
 def nearpc(pc=None, lines=None, to_string=False, emulate=False):
@@ -98,6 +100,10 @@ def nearpc(pc=None, lines=None, to_string=False, emulate=False):
         #     result.append('%s %s' % (line, lineno_to_src[line].strip()))
 
         line   = ' '.join((prefix, "%#x" % i.address, s or '', asm))
+
+        # Highlight the current line if the config is enabled
+        if pwndbg.config.highlight_pc and i.address == pc:
+            line = pwndbg.color.highlight(line)
 
         # If there was a branch before this instruction which was not
         # contiguous, put in some ellipses.
--- a/pwndbg/disasm/color.py
+++ b/pwndbg/disasm/color.py
@@ -51,7 +51,6 @@ def instruction(ins):
             # XXX: not sure when this ever happens
             asm += '<-- file a pwndbg bug for this'
         else:
-            colored_addr = pwndbg.color.get(ins.symbol_addr)
             asm = asm.replace(hex(ins.symbol_addr), ins.symbol)
             asm = '%-36s <%s>' % (asm, pwndbg.color.get(ins.symbol_addr))
 
--- a/pwndbg/color.py
+++ b/pwndbg/color.py
@@ -2,8 +2,8 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import functools
-
 import six
+import re
 
 import gdb
 import pwndbg.config
@@ -19,6 +19,7 @@ YELLOW         = "\x1b[33m"
 BLUE           = "\x1b[34m"
 PURPLE         = "\x1b[35m"
 CYAN           = "\x1b[36m"
+WHITE          = "\x1b[37m"
 GREY = GRAY    = "\x1b[90m"
 BOLD           = "\x1b[1m"
 UNDERLINE      = "\x1b[4m"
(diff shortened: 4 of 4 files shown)


anthraxx:
Feel free to give any feedback, I'm happy to make adjustments based on feedback. Hope you like this as it makes it much more convenient to distinguish the current line in the source (which was like impossible before) and also highlights the current instruction that the IP is pointing to. The arrow prefix is too decent to be the only highlight (in my opinion).
If you dislike having this yourself, I can also add some configuration options as the config module seems quote near! :+1: 

PS: A question that I'm interested in as I think i will contribute more and more in the future: How conservative or going nuts do you want the configuration system to be, should as much as possible be configurable or as least as possible? (was not sure as there currently seem to be 4 config options).

zachriggle:
Can you add a `pwndbg.config.Parameter` to control this behavior?

pwndbg/commands/nearpc.py
```diff
@@ -85,7 +85,7 @@ def nearpc(pc=None, lines=None, to_string=False, emulate=False):
 
     # Print out each instruction
     for i,s in zip(instructions, symbols):
-        asm    = pwndbg.disasm.color.instruction(i)
+        asm    = pwndbg.disasm.color.instruction(i, highlighted=i.address == pc)
```
zachriggle: `pwndbg.disasm.color.instruction()` also has access to `pc`, so there's no need for the conditional here

anthraxx:
@zachriggle sure, anything you prefer?
Otherwise I would go for: 
```
'highlight-pc', True, 'whether to highlight the current instruction'
'highlight-source', True, 'whether to highlight the closest source line'
```

pwndbg/disasm/color.py
```diff
@@ -11,8 +11,11 @@
 capstone.CS_GRP_JUMP
 ))
 
-def instruction(ins):
-    asm = u'%-06s %s' % (ins.mnemonic, ins.op_str)
+def instruction(ins, highlighted=False):
+    def maybe_highlight(s):
+        return s if not highlighted else pwndbg.color.bold(pwndbg.color.green(s))
```
zachriggle: This would be much simpler as:

```py
def highlight(line):
    if pwndbg.config.highlight_pc:
        return pwndbg.color.bold(pwndbg.color.green(line))
    return line
```

Also please pull the function to the outer scope.

pwndbg/commands/nearpc.py
anthraxx: Ok, will adjust that also, I need to get familiar with the codebase :smile_cat:

pwndbg/disasm/color.py
```diff
@@ -25,7 +28,7 @@ def instruction(ins):
 
         # If it's a constant expression, color it directly in the asm.
         if const:
-            asm = asm.replace(hex(ins.target), sym or target)
+            asm = asm.replace(hex(ins.target), maybe_highlight(sym or target))
```
zachriggle: I don't see a reason not to highlight the entire line (address included).

How about just:

```py
asm = asm.replace(hex(ins.target), sym or target)

if pwndbg.regs.pc == ins.address:
    asm = highlight(asm)
````

**Comment by zachriggle** [comment]
`highlight-pc` is fine :)

File: pwndbg/disasm/color.py (hunk 3 above)
Comment (anthraxx) [inline pwndbg/disasm/color.py]: the thing is that some part that adds colors by itself or make something bold also resets the color codes no normal so i will need to make that check on multiple locations, thats why i tried to just pass a bool flag to the function.
I removed that bool now, we can review the new patch once I tested it a bit, shouldn't take too long

**Comment by anthraxx** [comment]
patch adjusted but i needed to move the highlight function out and around both, target and sym (in case its not None) separately as both reset the colors to normal.
Tested it under quite some circumstances and looks good.

File: pwndbg/commands/context.py [hunk 4]
```diff
@@ -155,8 +157,16 @@ def context_source():
         if not source or source.startswith('1\t'):
             return []
 
+        # highlight the current code line
+        source_lines = source.splitlines()
+        if pwndbg.config.highlight_source:
+            for i in range(len(source_lines)):
```
zachriggle: This should probably use symbol-table lookups to find the current line.

There's an example here:
https://sourceware.org/gdb/onlinedocs/gdb/Line-Tables-In-Python.html

```py
     symtab = gdb.selected_frame().find_sal().symtab
     if symtab is not None:
         for line in symtab.linetable():
             print "Line: "+str(line.line)+" Address: "+hex(line.pc)
```

zachriggle: Nevermind, the code above already does this.

It appears there's not a way to get the actual line *contents* out via standard API calls :(

anthraxx: too bad :see_no_evil: so seems like the best option to do right now

zachriggle:
Hmm, yeah, nested colors end up causing issues.

This is probably something that we could detect and handle in a smarter way, but I don't want to bother with the code to do it 😛 

In any case, either way sounds fine 😄

anthraxx:
I thought a bit about this and I think it would make sense of I add functionality to the color class that does the following: when colorizing (f.e. blue()) it would make sense to replace the existing color reset codes of the passed text with the new one. This would have multiple advantages, f.e. an already colorized text, let's say its green, is part of a longer new string that should be colored blue. This would preserve the green part but everything else will be blue. Additionally a new method called reset() or normal() could just kill everything color related. 

For the currently highlighted instruction I would go with not resetting it so that red constants f.e. Will still be preserved (they look neat and usefully distinguish additional info) 

Will implement this if no objections follow

zachriggle:
This sounds cool.  I just did some of the heavy lifting for you in #63.  I'd recommend rebasing your changes against that, and adding a configurable point there.

anthraxx:
Yay this is really super awesome. I'm such a big fan of colors and adjustment configs :D and big thx to you, it's a lot of fun to contribute here as you are very open and really fast and active. Expect continuous help by me :+1

anthraxx:
adjusted the patch, it looks much cleaner and simpler now.
Added support to the color module to have nested colorization while preserving attributes like bold or underlined. also there is a strip() that can be used to eliminate existing colors/attributes if needed.

example:
```
print(pwndbg.color.red('|red|' + pwndbg.color.bold('|start-bold|' + pwndbg.color.blue("|blue|") + "|end-bold|") + '|red|'))
```

pwndbg/commands/nearpc.py
```diff
@@ -97,6 +99,10 @@ def nearpc(pc=None, lines=None, to_string=False, emulate=False):
 
         line   = ' '.join((prefix, "%#x" % i.address, s or '', asm))
 
+        # Highlight the current line if the config is enabled
```
zachriggle: This should probably be moved into pwndbg.color.instruction.  This ensures the coloring is applied when invoked internally, instead of just the command.

We should also be sure to only colorize the address and instruction, not ancillary data appended to the right (since this breaks color schemes; bold(blue(foo)) is not necessarily blue).

zachriggle: Hmm, though then we aren't highlighting the prefix and address.

anthraxx: I could switch it back but felt like being usefull when later we may get syntax highlighting for asm output, therefor it felt more appropriate to do it here. The example of the /x command pull request looks neat

anthraxx: Well it just highlights stuff in green when it does not have explicit color.  All constants etc will preserve it's color, did you try running the patch?

zachriggle: The screenshots that I posted were with your most recent suggested changes.

If you'd like to modify your PR to highlight as shown in that screenshot (i.e. prefix, address, and symbol name), that's fine with me.

anthraxx: Was that a typo or where did you post screenshots, I can't see any :cry: 

However the currently updated PR now looks like this:

highlight-source
highlight-with-color
highlight-with-colored-data

It also demonstrates that if at any point in the future the ASM output gets own coloring that those colors will be preserved and in such case only the prefix, address and symbol name will be highlighted. Right now it does color everything that does not have an explicit color.
Does this look OK to you? :smile_cat:

anthraxx:
hope you like the current version, I have already prepared another feature relying on this. Once this hits the master, I will open another feature pull request with more color definitions and config options for nearly full color theme freedom (all your chosen style and colors is preserved by default) :tada:

PR #50, opened 2016-06-09
Only register memory_changed and register_changed events if they exist (fixes GDB 7.7)

gsingh93:
An alternative solution would be to just check if the gdb version is 7.7 or less.


zachriggle:
Fixed with d632abe0c79f25ce0beefd56bf81167f3e2c6cee

PR #53, opened 2016-06-10
fix checksec regression under certain circumstances and version

anthraxx:
This resolves that a failing checksec command may output something when
falling back to the next version that will be tried. Additionally this
introduces a last option to try to call checksec before failing as some
distributions (like Arch Linix) package checksec under /usr/bin/checksec
and expect --file as command line option.
The returned output is decoded to ensure converting the bytes() object
to a proper string representation needed for python3, under python2 this
should simply be a no-op.
The line ending of the print command is stripped as the check_output
already returns a newline when not failing.

### DIFF
--- a/pwndbg/commands/checksec.py
+++ b/pwndbg/commands/checksec.py
@@ -1,6 +1,5 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-import gdb
 import pwndbg.commands
 
 import subprocess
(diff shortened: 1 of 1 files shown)


zachriggle:
Handled this a little bit cleaner with d7833ebe7b0597b437dda9c54619e544e6cb4e2e

zachriggle:
Also see binjitsu/binjitsu@e18eef7fae328ec0b44073a5fb5ab6f7dd86febe which adds a `--file` option

PR #55, opened 2016-06-10
Added .travis.yml

gsingh93:
Added .travis.yml to run a basic "does this load without crashing" test on Ubuntu 14.04/GDB 7.7

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,9 @@
+sudo: required
+dist: trusty
+language: python
+install:
+  - sudo apt-get -y install gdb
+  - lsb_release -a
+  - pip install -r requirements.txt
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -0,0 +1,9 @@
+sudo: required
+dist: trusty
+language: python
+install:
+  - sudo apt-get -y install gdb
```
zachriggle: We can probably just add this to setup.sh

gsingh93: What if you don't want the `gdb` from apt-get?

zachriggle: I suppose, but any custom version ought to be installed at not-/usr/bin/gdb (e.g. .../local/gdb)

PR #59, opened 2016-06-10
Enable caching on Travis

zachriggle:


zachriggle:
Looks like everything works, and the build time is down to ~1 minute.  We were also double-CI'ing for any branches pushed directly to this repo, that is now disabled.

PR #66, opened 2016-06-10
Fix docstrings, hide Defcon commands

zachriggle:


zachriggle:
Moving to `dev` branch

PR #70, opened 2016-06-12
Add color config of rodata and add cyan() colorize function

cebrusfs:
I make the color of rodata is able to config, and default is normal.

Also, I add missing colorize function `cyan()` so that the other can set the color to cyan() via `pwndbg config`.


zachriggle:
IIRC, The "rodata" color is used for everything not otherwise enumerated,
including constants (i.e. the value 7).

cebrusfs:
I think it does not include the some constant. It is only used for the address that can be found in vmmap but not match the other type.

I'm not sure that any other type of page will be misrecognized as RODATA. But the original code do that, I just make its color is configurable.

Complelte source for the reference:
```python
    if page is None:                 color = normal
    elif '[stack' in page.objfile:   color = stack
    elif '[heap'  in page.objfile:   color = heap
    elif page.execute:               color = code
    elif page.rw:                    color = data
    else:                            color = rodata

PR #73, opened 2016-06-12
Handle ImportError exception of pwndbg.emu.emulator in disasm

cebrusfs:
Since you handle the same exception when importing `pwndbg` (in `pwndbg/__init__.py`), I think it should be handle in `pwndbg/disasm/__init__.py`. Thus, people without installing `unicorn` still can use pwndbg in gdb and the emulator feature will turn off automatically.
Or maybe pwndbg can print some warning when emulator option is on but failure to import `emu`.

### DIFF
--- a/pwndbg/disasm/__init__.py
+++ b/pwndbg/disasm/__init__.py
@@ -15,7 +15,10 @@ import pwndbg.memory
 import pwndbg.symbol
 import pwndbg.memoize
 import pwndbg.jump
-import pwndbg.emu.emulator
+try:
+    import pwndbg.emu.emulator
+except:
+    pwndbg.emu = None
 
 import capstone
 from capstone import *
@@ -143,7 +146,7 @@ def near(address, instructions=1, emulate=False):
     emu = None
 
     # If we hit the current instruction, we can do emulation going forward from there.
-    if address == pc and emulate:
+    if address == pc and pwndbg.emu and emulate:
         emu = pwndbg.emu.emulator.Emulator()
 
         # For whatever reason, the first instruction is emulated twice.


zachriggle:
Building without Unicorn available isn't optional anymore, only whether to use it or not.  I don't remember if it was @gsingh93 or @thebarbershopper who wanted it to be optional.

Given that it's performed automatically via `setup.sh` and is a submodule to the `pwndbg` repo, I think it's OK to move it to be required.

Thoughts?

gsingh93:
I wanted it to be optional, but mainly because 1) things were still broken/experimental and 2) there wasn't a `setup.sh` script that did everything for you. I'm fine with it being required now.

zachriggle:
Sounds good to me. @cebrusfs, are you super attached to not needing Unicorn installed, or were you just making things more consistent (and thanks either way!)

cebrusfs:
I just don't want to execute `sudo` on my machine, lol. I tried to install capstone via `apt-get` (since it is requirement for disasm). However, I did not find a good way to install unicorn via `apt-get`. I just tried to use pwndbg without it.
Without this issue, I didn't seem any error during the normal usage (maybe there is, but i didn't trigger). It seems that unicorn do less thing than I think right now.

I may try to install unicorn manually in short future, and see what will be different.

zachriggle:
There is no supported usage model for Unicorn which does not require `sudo` (e.g. `sudo apt-get`), simply because of default linker paths and a lack of a nice package on PyPI.

However, you can set `LD_LIBRARY_PATH` in your RC files to point to whichever user-declared directories if you prefer.  For example, `pip install --user capstone` works nicely without `sudo`.  Unfortunately, there's no target for Unicorn yet (or Keystone).  I'd welcome a modification to `setup.sh` which performs these changes in lieu of this pull request.

As Unicorn grows more mature and stable (cc: @aquynh) I don't intend to keep it separable from pwndbg.  Most of the current `pwndbg.disasm.arch` was written only because Unicorn wasn't yet stable enough (or didn't exist, I'm not sure which).

While this particular pull request is a straightforward work-around, as time progresses Pwndbg and Unicorn (among other @aquynh projects) will become increasingly intertwined.  Rather than structure Pwndbg to work in a crippled state, I'd recommend working toward `sudo`-less capability with @aquynh (or an option for `setup.sh` to modify `LD_LIBRARY_PATH`).

anthraxx:
Sorry, I don't want to hijack this thread as it does not totally belong here, but just wanted to raise a side-note to keep in mind:
Even if capstone and unicorn are submodules aiding the default ```git clone ... && ./setup.sh``` type of usage it would be nice to make sure it always uses stuff that is compatible with the latest releases of capsone, unicorn, keystone. This would allow distributions (f.e. Arch Linux) to use the distribution based packages instead of forcing to build and install them separately from here. This would also aid a future where you tag a release of pwndbg that can be shipped in a distribution by referencing the distribution packages of unicorn etc :smile_cat: 
PS: I do that right now and I did not see any problems so far

zachriggle:
Merged with ded18e02ee992d0d3b14e2922883258ea55720e8

PR #74, opened 2016-06-12
Show "x" when the condition instruction will not be taken

cebrusfs:
As my understanding, `None` is mean "unconditional" and `False` is mean "it is conditional, but will not be taken". So I change the default value of `ins.condition` from `False` to `None` and set the `condition` attribute to `None` for all instructions whose address does not matched PC.

Then we can use `ins.condition` to know whether this condition instruction will be taken or not.

It is useful to show something when a condition instruction will not be taken because there is some instruction like `cmove` which is not a branch instruction. We can't know this instruction will be done or not by the origin UI.

screen shot 2016-06-12 at 11 29 06 pm

### DIFF
--- a/pwndbg/disasm/arch.py
+++ b/pwndbg/disasm/arch.py
@@ -72,7 +72,7 @@ class DisassemblyAssistant(object):
         instruction.condition = c
 
     def condition(self, instruction):
-        return False
+        return None
 
     def enhance_next(self, instruction):
         """
--- a/pwndbg/disasm/arch.py
+++ b/pwndbg/disasm/arch.py
@@ -244,7 +244,7 @@ class DisassemblyAssistant(object):
 
         for i, op in enumerate(ins.operands):
             rv.append('   operands[%i] = %s' % (i, ops.get(op.type, op.type)))
-            rv.append('       access   = %s' % (access.get(op.access, op.access)))
+            #rv.append('       access   = %s' % (access.get(op.access, op.access)))
 
             if op.int is not None:
                 rv.append('            int = %#x' % (op.int))
--- a/pwndbg/disasm/color.py
+++ b/pwndbg/disasm/color.py
@@ -57,11 +57,11 @@ def instruction(ins):
         asm = asm.replace(ins.mnemonic, pwndbg.color.bold(ins.mnemonic))
 
     # If we know the conditional is taken, mark it as green.
-    if ins.condition is True:
-        asm = pwndbg.color.green(u'✔ ') + asm
-    elif ins.condition is False:
-        asm = pwndbg.color.red(u'✘ ') + asm
-    else: # ins.codition is None
+    if ins.codition is None:
         asm = '  ' + asm
+    elif ins.condition:
+        asm = pwndbg.color.green(u'✔ ') + asm
+    else:
+        asm = pwndbg.color.red(u'✘ ') + asm
 
     return asm
--- a/pwndbg/disasm/color.py
+++ b/pwndbg/disasm/color.py
@@ -57,9 +57,11 @@ def instruction(ins):
         asm = asm.replace(ins.mnemonic, pwndbg.color.bold(ins.mnemonic))
 
     # If we know the conditional is taken, mark it as green.
-    if ins.condition:
+    if ins.condition is True:
         asm = pwndbg.color.green(u'✔ ') + asm
-    else:
+    elif ins.condition is False:
+        asm = pwndbg.color.red(u'✘ ') + asm
+    else: # ins.codition is None
         asm = '  ' + asm
 
     return asm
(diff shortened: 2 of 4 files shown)


pwndbg/disasm/color.py
```diff
@@ -57,9 +57,11 @@ def instruction(ins):
         asm = asm.replace(ins.mnemonic, pwndbg.color.bold(ins.mnemonic))
 
     # If we know the conditional is taken, mark it as green.
-    if ins.condition:
+    if ins.condition is True:
         asm = pwndbg.color.green(u'✔ ') + asm
-    else:
+    elif ins.condition is False:
+        asm = pwndbg.color.red(u'✘ ') + asm
+    else: # ins.codition is None
```
zachriggle: I'd prefer this check to be explicit for `is None` and then just truthiness (`if ins.condition` vs `... is True`)

zachriggle:
Overall I like it.  However, almost everything under the current `pwndbg.disasm.arch` module is going to be gutted in the near future, since Unicorn handles everything for us.

I'm in favor of merging this in the meantime, but I'd advise against spending much time on that module since it's going away (or being massively overhauled).

FWIW, you are correct on `True` (definitely taken) and `False` (definitely not taken), but `None` simply means that no data is added.  This means that either the jump is unconditional and does not benefit from extra annotation, or that we cannot reason about it.

zachriggle:
While there's a thread on open on this general functionality, what's the view on the current "X" vs "V" using Unicode symbols and color, versus just graying out the "not taken" case?

I think that our current color infrastructure is not robust enough to handle this (just by virtue of the way ANSI codes work), but it's worth exploring as an idea.

anthraxx:
@zachriggle I personally really love that unicode symbol, also i would really love that additional "X" symbol. As those "V" and "X" symbols are on the left side of the mnemonic its really easy to distinguish the result while still keeping the instruction easily readable by not graying it out. (just personal point of view).

cebrusfs:
Since I added the syntax highlight feature on my own version of pwndbg, it will not easy to gray it out on current color infrastructure.
In my version of UI, every part of asm have color. I can't find a good place to put the information of current branch. That's why I keep using "V" and "X". (The screenshot of the first post is a from my version of UI)

BTW, in my design, I want to make the width of every line as short as possible, because something I need to use gdb on my laptop or somewhere with a small screen.

zachriggle:
@cebrusfs: I would gladly accept your changes to make things less wide 😄 

ASM syntax highlighting is difficult.  If you could make a stand-alone library which adds color information to Capstone disassembly, I'd love to use it in Pwndbg.

pwndbg/disasm/color.py
zachriggle: Waiting for this change to accept the PR

cebrusfs:
@zachriggle: I've changed the `if`. Is it what you want?

For reducing the width, I remove some `<>` around the symbol. However, in the most cases, the bottleneck is the output of telescope. I may try to make it much tider.

For syntax highlighting, I just use `pygments` module to do that, it will parse a string and return a string with color. If you are interested what I did, you can see commits with "syntax highlighting" in [my fork](https://github.com/cebrusfs/217gdb/commits/master). Please ignore some hacking because its purpose is just for my personal using.

pwndbg/disasm/arch.py
```diff
@@ -244,7 +244,7 @@ def dump(self, instruction):
 
         for i, op in enumerate(ins.operands):
             rv.append('   operands[%i] = %s' % (i, ops.get(op.type, op.type)))
-            rv.append('       access   = %s' % (access.get(op.access, op.access)))
+            #rv.append('       access   = %s' % (access.get(op.access, op.access)))
```
zachriggle: Is there a reason this was commented out?

cebrusfs: I got `AttributeError` and I couldn't find where it is created, so I removed it to make dump be usable.
What's this attribute do? It seems that it's not capstone attribute.

zachriggle:
I have reverted this in 1f6d65f80b64d46528b07293a0bfd282a53146ee.

This breaks the situation where there is a conditional jump which we cannot reason about (i.e. it is not at `$PC`).  Since we cannot reason about it, it is marked with `None`, which makes it treated as an unconditional jump -- i.e., it is always taken.

This breaks disassembly by prematurely folding a jump that may not be taken.

cebrusfs:
Umm.... I didn't consider that case, but this feature is useful. I've not read the source of jump folding.
 Do you have any idea to fix it?

zachriggle:
If you are using Unicorn this feature doesn't matter since it only shows instructions which are executed.

Conditional instructions are a weird part in this since they are executed but may not do anything.

Anyhow, the current triplet of (yes, no, None) is insufficient since None also means unconditional.  Marking unconditional as Yes means every JMP gets checked.  We need a fourth state which requires some other changes.

I expect that best way to break it apart would be to mark an instruction as conditional and conditonal-None means "don't know" vs unconditional-None means followed.

PR #81, opened 2016-06-16
fixed condition label expression leading to always be None

anthraxx:
Show "x" when the condition instruction will not be taken (#74) in commit 6c04e5368e4d48472e10b5fbbb8cbf2a305fe187 introduced a regression because of a typo leading to always show nothing on conditional instructions.


zachriggle:
Can you submit a pull-request with:

- A binary
- A coredump of that binary (or multiple coredumps for different situatiosn)

That we can use to test this scenario?

In the meantime I'll accept the PR since it's a regression, but I'd like to avoid this in the future.

PR #83, opened 2016-06-16
ropper command added

sashs:
Hi,

I've implemented a command to use ropper in pwndbg. It is optional. If ropper is not installed you will see a message. 
ATM I've added just a few commands here, but it is possible to add all commands provided by ropper easily.

### DIFF
--- a/pwndbg/__init__.py
+++ b/pwndbg/__init__.py
@@ -32,6 +32,7 @@ import pwndbg.commands.peda
 import pwndbg.commands.procinfo
 import pwndbg.commands.reload
 import pwndbg.commands.rop
+import pwndbg.commands.ropper
 import pwndbg.commands.search
 import pwndbg.commands.segments
 import pwndbg.commands.shell
--- a/pwndbg/commands/ropper.py
+++ b/pwndbg/commands/ropper.py
@@ -18,7 +18,13 @@ parser = argparse.ArgumentParser(description="Gadget search, ropchaining and mor
 parser.add_argument('--gadgets',
                     help='Prints all gadgets found in the binary', action='store_true')
 parser.add_argument('--search', type=str,
-                    help='String to grep the output for')
+                    help='String to search for gadgets. ? - any character; %  - any string. Example: ropper --search "mov [e?x]"')
+parser.add_argument('--jmp', type=str,
+                    help='Looks for jmp reg or equivalent instructions. Example: ropper --jmp esp,eax', metavar='<reg[,<reg>]>')
+parser.add_argument('--opcode', type=str,
+                    help='Looks for the specified opcode. Valid formats examples: ffe4, ffe?, ff??. Example: ropper --opcode ffe4', metavar='<reg[,<reg>]>')
+parser.add_argument('--instructions', type=str,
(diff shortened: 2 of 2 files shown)


zachriggle:
Can you add a coredump and binary to the `tests/` directory, and add a test or two for this?

I apologize for making you a guinea pig for testing, but it has to happen somewhere.  This is a large-ish code addition, and I'm entirely unfamiliar with `ropper`.

sashs:
No problem. I will tests there. :)

anthraxx:
just as a side note, if you change existing commits of a PR, github will notice this when you force-push them to your fork. This keeps the commits inside the PR more useful _after_ the commits have been accepted (f.e. a "refactored" commit separately just clutters history if its within the same PR.).
PS: More useful commit messages that label its about the ropper command would also be great "refactored" or "changed help string" could be anything :smile:

zachriggle:
As far as I can tell, the only functionality that this provides over just shelling out to `ropper` command-line is to auto-dump a corefile, or provide the executable filename.

I'm willing to accept a pull request for `ropper` functionality which is limited to that.  Most of this functionality belongs in `ropper`, not pwndbg.

See how [the current `rop` command](https://github.com/pwndbg/pwndbg/blob/master/pwndbg/commands/rop.py) is implemented for an example of how to do that.

sashs:
No problem. 
I thought it is a good idea to provide more functionality, therefore I add those commands. But I understand your point of view. :)
So I suggest to add commands for:
- print all gadgets
- search gadgets
- settings (because of badbytes, etc)
- ropchain generation using the executable or the executable and the used libraries (similar to mona.py)

Is that ok for you?

sashs:
The reason why I added it in this way is the gadgets have to be collected just once and not everytime when you want to search a gadget. If you have a large binary it takes much time to gather the gadgets and if this has to be done everytime when you search for gadgets this is very time consuming.

zachriggle:
I recommend caching the gadgets like Pwntools / Binjitsu does.

```
$ time python -c 'from pwn import *; r=ROP("/lib/x86_64-linux-gnu/libc.so.6")'
[*] '/lib/x86_64-linux-gnu/libc.so.6'
    Arch:     amd64-64-little
    RELRO:    Partial RELRO
    Stack:    Canary found
    NX:       NX enabled
    PIE:      PIE enabled
[*] Loading gadgets for '/lib/x86_64-linux-gnu/libc.so.6'
python -c 'from pwn import *; r=ROP("/lib/x86_64-linux-gnu/libc.so.6")'  10.53s user 0.07s system 99% cpu 10.644 total
$ time python -c 'from pwn import *; r=ROP("/lib/x86_64-linux-gnu/libc.so.6")'
[*] '/lib/x86_64-linux-gnu/libc.so.6'
    Arch:     amd64-64-little
    RELRO:    Partial RELRO
    Stack:    Canary found
    NX:       NX enabled
    PIE:      PIE enabled
[*] Loaded cached gadgets for '/lib/x86_64-linux-gnu/libc.so.6'
python -c 'from pwn import *; r=ROP("/lib/x86_64-linux-gnu/libc.so.6")'  0.32s user 0.02s system 90% cpu 0.375 total
```

sashs:
Oh, I've never looked at it. This is cool. 
Also I forgot it is still possible to start the console for ropper or ropgadget. Then it is not needed to load the gadgets again. 

I will change the implementation today.

sashs:
I changed the implemenation.

zachriggle:
Thanks, @sashs!

PR #84, opened 2016-06-16
more colors and fully customizable and themeable styles

anthraxx:
this (more or less) allows fully customizing the style.


pwndbg/hexdump.py
```diff
@@ -12,40 +12,44 @@
...
+    for c in bytearray(b'\xff\x7f\x80'):
+        color_scheme[c] = pwndbg.color.hexdump_special("%02x" % c)
+        printable[c] = pwndbg.color.hexdump_special('.')
 
+    color_scheme[-1] = '  '
+    printable[-1] = ' '
 
 def hexdump(data, address = 0, width = 16, skip = True):
+    global color_scheme, printable
+    if not color_scheme or not printable:
```
zachriggle: I'd prefer to just call `load_color_scheme()` once on load.

Separately, since we do not modify `color_scheme` or `printable`, there's no need to declare them as `global`.

pwndbg/hexdump.py
```diff
@@ -12,40 +12,44 @@
...
-
-for c in bytearray((string.ascii_letters + string.digits + string.punctuation).encode('utf-8', 'ignore')):
-    color_scheme[c] = pwndbg.color.bold("%02x" % c)
-
-for c in bytearray(b'\x00\xff'):
-    color_scheme[c] = pwndbg.color.red("%02x" % c)
-
-for c in bytearray(b'\xff\x7f\x80'):
-    color_scheme[c] = pwndbg.color.yellow("%02x" % c)
+def load_color_scheme():
```
zachriggle: I don't see a useful need to support changing color schemes at-runtime.

pwndbg/config.py
```diff
@@ -62,10 +62,13 @@ def __init__(self, name, default, docstring):
                                         gdb.COMMAND_SUPPORT,
                                         getParam(default))
         self.value = default
+        self.trigger = trigger
 
         setattr(module, self.name, self)
 
     def get_set_string(self):
+        if self.trigger:
```
zachriggle: I understand the need for "setting changed, trigger function" and this may be useful in the future.

However, Pwndbg currently makes extensive use of annotations, which make it easy to see when a given function is invoked.

I'd much prefer annotation style to this style.  To change to annotation style, one simply needs to implement `__call__` to return a function wrapper, and that should work.

An example that's relevant may be:

```py
@pwndbg.color.hexdump_special
def reload_hexdump_colors():
    ...
```

Or something like that.  This would allow easy registration of *mutliple* functions for a given `Parameter`, and allow easy grepping and see when a given thing is triggered.

zachriggle: tl;dr I like the idea, but we can extend it to a general case (N triggers vs 1 trigger) and also make it easier to register triggers

pwndbg/hexdump.py
zachriggle: On further evaluation, the general idea of runtime setting changes *is* useful, so I'm not opposed for this to be the first use-case (and it is an easily-observable one).

pwndbg/commands/context.py
```diff
@@ -204,7 +200,7 @@ def context_backtrace(frame_count=10, with_banner=True):
     result = []
 
     if with_banner:
-        result.append(pwndbg.color.blue(pwndbg.ui.banner("backtrace")))
+        result.append(pwndbg.color.banner(pwndbg.ui.banner("backtrace")))
```
zachriggle: Does it make sense for `pwndbg.ui.banner` to automatically invoke the `color` property as well?  It would shorten this, and make it so that all banners are consistent.

pwndbg/color.py
```diff
@@ -80,6 +114,57 @@ def rwx(x):
 def highlight(x):
     return generateColorFunction(pwndbg.config.color_highlight)(x)
 
+def register(x):
```
zachriggle: Given the vast quantity of registers and the sub-heirarchies, (`register_xxx` and `nearpc_xxx`) it seems like this doesn't really scale particularly well.

In particular, this ends up with us needing to invoke functions like `pwndbg.color.nearpc_symbol`.

Consider instead in a future revision that we might do something like:

```py
import pwndbg.color.nearpc as N`

print(N.symbol('foobar'))
```

Having an actual object hierarchy rather than an underscore-separated hierarchy might let us get away with vaguely cleaner code.

In the meantime it's not important since we have a manageable number of color settings, but it's something worth considering.

zachriggle:
I added a lot of line-comments, so hopefully you can check those out and provide input.

In the future, I think it will be useful to separate out a `pwndbg.config.Parameter` into some specialized class for colors, given how many of them there are.  Ultimately, I don't want to litter the output of the `config` command with color options.  Instead, we can have a separate `theme` or `color` command or whatever that displays those.

pwndbg/hexdump.py
anthraxx: well we do modify  that with the hexdump color commands, therefor this is needed :smile:

pwndbg/config.py
anthraxx: make sense, good idea :+1:

pwndbg/commands/context.py
anthraxx: sure why not, just don't wanted to refactor too much of how you did things before. will do

pwndbg/color.py
anthraxx: makes sense, more structure is always good :smile:

anthraxx:
also makes sens, do you want me to change that separated command within this PR, or simply a new one soon?

zachriggle:
If you could do it in this PR, that'd be great :)

pwndbg/hexdump.py
zachriggle: `color_scheme` only needs to be declared as `global` if it is directly modified within the scope of this function.  I don't see that anywhere.

anthraxx: good catch, will adjust it

anthraxx:
sure, i will adjust that also :smile:

anthraxx:
i have adjusted all changes except the separation of color configs to theme command (will do this now) but you could already give some feedback. I have separated all colorization to separated modules and implemented all other feedback.

anthraxx:
work finished, waiting for feedback... some screens from the default and my local theme:

**default:**
default

**mine:**
mine

zachriggle:
I apologize for the delay, I'm currently at a conference and haven't had time to give this any attention.

I haven't forgotten, just very busy this week! <3

anthraxx:
@zachriggle sure, absolutely no problem... I hope you enjoy the stay and the conference!
I'm finished so far and I think everything can be adjusted now :stuck_out_tongue_closed_eyes: 
Guess pwndbg will be the first fully themable gdb extension :smirk_cat:

pwndbg/color/disasm.py
```diff
@@ -1,27 +1,33 @@
...
 capstone_branch_groups = set((
-capstone.CS_GRP_CALL,
-capstone.CS_GRP_JUMP
+    capstone.CS_GRP_CALL,
+    capstone.CS_GRP_JUMP
 ))
 
+config_branch = theme.Parameter('disasm-branch-color', 'bold', 'color for disasm (branch/call instruction)')
+
+def branch(x):
```
zachriggle: Can this be moved into the regular `pwndbg.color` module?

anthraxx: is already moved in a later commit :smile_cat:

anthraxx: oh mis,interpated the diff. this was moved to own module ``pwndbg/color/disasm.py`` and i think that is a good place for it. later if the soon following PR will add syntax highlighting I'm planning to provide disasm color override values in this file (``pwndbg/color/disasm.py``) to have own special colors for things like call, jmp, ret

anthraxx:
rebased all changes on top of current master and fixed several conflicts. would be nice if we can get this further ahead :cat:

zachriggle:
The current version still has the issue where the entire current line is bolded, rather than just the address / instruction.  This happens to be more apparent on the color scheme I have, which overloads "bold blue" as something different. Thoughts?

screen shot 2016-06-30 at 1 15 22 am

screen shot 2016-06-30 at 1 15 27 am

It also looks like we're coloring the backtrace information in `context backtrace`.  This isn't particularly helpful, as the code will ~= always be executable, but the user may have set a different color for executable code than the current default of `red` (which happens to coincide).  I think perhaps this should be a default-off toggle-able that uses the default legend colors.  What do you think?

screen shot 2016-06-30 at 1 17 10 am

The `theme` command should probably color the configuration options rather than simply stating them.  Thoughts?

screen shot 2016-06-30 at 1 18 53 am

The current `hexdump` address and offsets use default colors which deviate from the standard color palette.  Originally, they were just plain, but I'd be down for coloring them according to the standard color legend as a default.  The current value in the Pull Request is `green`.  Thoughts?  Separately, it looks like we're adding color now to the ascii on the side.  Was this intentional?  For the offsets themselves (rather than the addresses), I'd like to see consistent color defaults between `telescope` and `hexdump`.

screen shot 2016-06-30 at 1 20 29 am

For the `nearpc` and `context code` commands, the default-colored prefix overloads a default legend color of `red`.  Since this effectively inherits from the `highlight-pc` setting (because `set highlight-pc on` and `set nearpc-prefix-color normal` will still have a bolded cursor), I think that this should then default to `normal` and users can add a separate color if they like.  Thoughts?

screen shot 2016-06-30 at 1 35 02 am

screen shot 2016-06-30 at 1 38 12 am

anthraxx:
I try to use some numbers so we can more easily reference certain topics:

1) Didn't notice this as an issue, the implementation was way easier like this. I will change it to only highlight the address part and the instruction part.

2) Hm I see your point, will deactivate any default colors there. However I would still be really happy to have more freedom about that backtrace and be able to colorize the address and symbol also in a different color. I think if this is an default-off anyway it shouldn't be much of a problem to assume anybody who wants to have any colors there to set it to the same value as the code (if he/she likes it that way). Is that fine? A more complex solution could be to be able to reference other theme values f.e. via theme prefix like value: ```theme.memory-code-color```?

3) Cool idea with theme command colorizing values, will see how to achieve that in a easy way

4) Sure, will change hexdump address default value to none. The colorization of the ascii part was intentional as I find it incredibly handy (stolen from radare2 :smile_cat: ). As it looks like you dislike it, I will add a boolean toggle command with default value ```off```

5) I'm bit confused... do you mean by "```overloads a default legend color of red```" that the prefix cursor uses the same color as the legend color and therefor it should be stripped and just use ```normal``` or did i misunderstand this? If so, please explain me what exactly should have a ```normal``` default value :smile: 

6) Of cause, will make those characters the default ones

zachriggle:
1-4 yep yep yep yep

For (5) I mean that the default in the PR actually `red,bold`.  The `bold` should be tied to `highlight-pc` and the choice of red doesn't work with the rest of the color palette.  Red is only used anywhere else to color executable addresses or symbols (and in hexdump for NULL/whitespace bytes).  Using it for a glyph color displayed in `context` diverges from that and detracts from being able to find executable addresses at-a-glance by looking for red.

anthraxx:
5) Ah, i see! will switch the default value to not have any color property.

zachriggle:

anthraxx:
I have rebased all changes ontop of the current master's HEAD, so it should be easy to apply.
I (think) that all relevant things got adjusted.

However I decided for point 4) to not use colors for now as (to be honest) I start loosing the overview for this feature and I would appreciate if we can get this into the master before adding more features/additions.
I think right now all important parts should be like you wanted (f.e. I don't see any legend color overwrites with the default settings).

zachriggle:
Looks good to me!  Merging after much delay <3

anthraxx:
:tada: :balloon: cool, picking up new tasks then :yum: thanks a lot!

PR #86, opened 2016-06-22
Update disasm/__init__.py

cebrusfs:
Don't import `pwndbg.emu.emulator` twice. This is already handled by the some code below.


zachriggle:
Thanks, @cebrusfs!

PR #93, opened 2016-06-27
Add pwndbg commands list filtering

disconnect3d:
Hello,

I have added a possibility to filter the list of commands provided with `pwndbg`, simply by passing an argument to it:

```
pwndbg> pwndbg entry
entry                Set a breakpoint at the first instruction executed in
entry_point          GDBINIT compatibility alias to print the entry point.
pwndbg> 
```

Currently, this is done by simple lowercase string comparison. Imho we don't really need regexes here.

### DIFF
--- a/pwndbg/__init__.py
+++ b/pwndbg/__init__.py
@@ -130,7 +130,7 @@ handle SIGSEGV stop   print nopass
 for line in pre_commands.strip().splitlines():
     gdb.execute(line)
 
-msg = "Loaded %i commands.  Type pwndbg for a list." % len(pwndbg.commands._Command.commands)
+msg = "Loaded %i commands.  Type pwndbg [filter] for a list." % len(pwndbg.commands._Command.commands)
 print(pwndbg.color.red(msg))
 
 cur = (gdb.selected_inferior(), gdb.selected_thread())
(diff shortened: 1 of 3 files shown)


zachriggle:
I would prefer that all new commands be of the type `ArgparsedCommand`.  Would you mind modifying your pull request to use this?

Thanks!

disconnect3d:
I've changed the commands in misc.py to use `ArgparsedCommands`.

* Should I also split them into different files?
* Is the `type=int` proper for `errno` and `distance` commands arguments? Those arguments were `gdb.Value` before this change. Passing `type=gdb.Value` to `parser.add_argument` doesn't work:

  ```
  pwndbg> distance 2 0x33
    File "/home/dc/pwndbg/pwndbg/commands/__init__.py", line 54, in __call__
    File "/home/dc/pwndbg/pwndbg/commands/__init__.py", line 159, in _ArgparsedCommand
      return function(**vars(args))
    File "/home/dc/pwndbg/pwndbg/commands/misc.py", line 71, in distance
      a = int(a) & _arch.ptrmask
  gdb.error: Cannot convert value to long.
  ```

zachriggle:
Would you mind rebasing against `master` and removing the changes to the `distance` command?

Thanks!

disconnect3d:
Done.

Currently, as the 2nd commit states, only the `pwndbg` and `errno` commands use argparse in `misc.py` file.

By the way, if you think that we need tests for those commands, feel free to create a task for it and assign me to it.

I am glad to help in my free time ;).

zachriggle:

PR #102, opened 2016-07-07
fixed potential wrong config param type resolution

anthraxx:
This fixes an issue that can potentially happen with unordered dicts
in python3. A boolean typed config parameter can be handles as
integer dependin on the order of iteration in getParam(value).

The reason for this is because True/False are both, bool and int
at the same time, however its not the other way around.

```
checking type <class 'bool'>
matched True against <class 'bool'>

checking type <class 'str'>
checking type <class 'int'>
matched True against <class 'int'>

In [1]: isinstance(True, bool)
Out[1]: True

In [3]: isinstance(True, int)
Out[3]: True

In [4]: isinstance(1, bool)
Out[4]: False
```

### DIFF
--- a/pwndbg/config.py
+++ b/pwndbg/config.py
@@ -21,12 +21,17 @@ from __future__ import unicode_literals
 
 import sys
 import types
+import collections
 
 import six
 
 import gdb
 
-TYPES = {}
+TYPES = collections.OrderedDict()
+
+# The value is a plain boolean.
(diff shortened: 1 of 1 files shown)


zachriggle:

anthraxx:
@zachriggle awesome, have a nice stay and a lot of fun and relaxation! see ya soon mate :sunglasses:

PR #106, opened 2016-09-08
Remove the sudos in setup.py

Faalentijn:
sudo resets the environment variables causing the UNICORN_QEMU_FLAGS to be worthless and the installation to fail on non-Python 2 systems.

They were also not needed since we were already running in sudo.


zachriggle:
Could you confirm that 7bdec11e6d7bcfa79f7bd238646dca31080b1ad1 resolves this?

Faalentijn:
It resolves the problem for me.

PR #124, opened 2016-10-12
Avoid infinite loop on 0-sized chunks in heap command

andigena:
During heap traversal, we'll enter an infinite loop if a 0-sized chunk is found (happens to me frequently while developing exploits), printing the same chunk forever.


zachriggle:
Thanks for your contribution during #Hacktoberfest!

Issue #115, opened 2016-09-23
the context did not print out LR ( link register) while debugging ARM binary


bruce30262:
<img width="317" src="https://cloud.githubusercontent.com/assets/4478350/18794874/4ee57eda-81f4-11e6-9e25-df1126e3b616.png">  
  
As you can see in the picture, pwndbg did not print out the link register ( r14 ) while debugging the ARM binary . I thought it was supposed to print out the same content as the `i r` command ?

zachriggle:
It's not intended to print out the same contents as anything in particular.

The link register is not normally needed in the register context, and is generally more applicable to the stack context or something like that.

I'm not sure that adding what is effectively "the return address register" to the print-out is useful in the general case, but it'd be fine if a `config` point were added for it.

bruce30262:
@zachriggle agree. From my experience sometimes the LR register are used to store some local variables ( like the [binary](https://github.com/ctfs/write-ups-2016/blob/master/mma-ctf-2nd-2016/pwn/candy-store-300/candystore) I use to demonstrate the issue), that's why I would like to inspect the register context during the debug session.  
Anyway I see @dqi has already done the modification (thanks man!), guess I'll just add the configuration in my `.gdbinit`

zachriggle:
I think this is done via #127

PR #127, opened 2016-10-14
add option to show retaddr register to config

dqi:
Adds an option (show-retaddr-reg) to config to show return address registers like 'lr' on ARM. For #115 

```
show-retaddr-reg       False          whether to show return address register
```

Example:

```
*R0   0xf6fff23c —▸ 0xf66b6454 ◂— 0xe1a00005
*R1   0xf67937c4 —▸ 0x12000 ◂— 0x0
 R2   0x0
*R3   0xf6fff23c —▸ 0xf66b6454 ◂— 0xe1a00005
*R4   0x87ed (__libc_csu_init+1) ◂— 0x743f8e9
 R5   0x0
*R6   0x856d (_start+1) ◂— lsls   r0, r6, #3
*R7   0xf6fff238 ◂— 0x0
 R8   0x0
 R9   0x0
*R10  0xf67fdfb4 —▸ 0x2feec ◂— 0x0
 R11  0x0
*R12  0xb0
*SP   0xf6fff238 ◂— 0x0
*LR   0x87c9 (main+113) ◂— lsls   r6, r0, #1 /* u'F' */
*PC   0x86b2 (checkpassword+10) ◂— blx    #0x84c4

```

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -71,11 +71,14 @@ def regs(*regs):
     print('\n'.join(get_regs(*regs)))
 
 pwndbg.config.Parameter('show-flags', False, 'whether to show flags registers')
+pwndbg.config.Parameter('show-retaddr-reg', False, 'whether to show return address register')
 
 def get_regs(*regs):
     result = []
 
-    if not regs:
+    if not regs and pwndbg.config.show_retaddr_reg:
+        regs = pwndbg.regs.gpr + (pwndbg.regs.frame, pwndbg.regs.current.stack) + pwndbg.regs.retaddr + (pwndbg.regs.current.pc,)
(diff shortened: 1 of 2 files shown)


zachriggle:
I get the changes in `context.py` but why does `pwndbg.regs.common` need to change?

dqi:
It is to get the changed marker displayed for the retaddr register:

```
  @pwndbg.events.cont
  def update_last():
      M = sys.modules[__name__]
      M.last = {k:M[k] for k in M.common}
```

Adds to the 'last' dictionary which is used to see which registers changed, maybe better to add it here?

zachriggle:
Hmm, yeah that's probably useful.  If we can do it conditionally on the `config` setting I'm down 😎

Issue #126, opened 2016-10-14
python 3 error in inthook.py


Isaac0616:
When I invoke commands related to python subprocess, I get the following errors.

```python
pwndbg> checksec
bufsize must be an integer
'checksec': Prints out the binary security settings. Attempts to call the binjitsu
    checksec first, and then falls back to checksec.sh.
pwndbg> ls
Traceback (most recent call last):
  File "/root/pwndbg/pwndbg/commands/__init__.py", line 57, in __call__
    return self.function(*args, **kwargs)
  File "/root/pwndbg/pwndbg/commands/shell.py", line 82, in handler
    os.execvp(cmd, (cmd,) + a)
  File "/usr/lib/python3.5/os.py", line 615, in execvp
    _execvpe(file, args)
  File "/usr/lib/python3.5/os.py", line 643, in _execvpe
    path_list = get_exec_path(env)
  File "/usr/lib/python3.5/os.py", line 681, in get_exec_path
    warnings.simplefilter("ignore", BytesWarning)
  File "/usr/lib/python3.5/warnings.py", line 75, in simplefilter
    "lineno must be an int >= 0"
AssertionError: lineno must be an int >= 0
```

The first error is cause by the following check.
```python
if not isinstance(bufsize, int):
    raise TypeError("bufsize must be an integer")
```

After some debugging, I found that this problem is due to `inthook.py`. 
```python
pwndbg> python print(int)
<class 'pwndbg.inthook.xint'>
```

In `inthook.py`, it use the syntax
```
__metaclass__ = IsAnInt
```
to define metaclass, but this is a python 2 only syntax according to [this page](http://python-future.org/compatible_idioms.html#metaclasses).

I can fix it by using the `future` package like this:
```git
--- a/pwndbg/inthook.py
+++ b/pwndbg/inthook.py
@@ -9,6 +9,7 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
+from future.utils import with_metaclass

 import sys

@@ -28,8 +29,7 @@ class IsAnInt(type):
     def __instancecheck__(self, other):
         return isinstance(other, _int)

-class xint(builtins.int):
-    __metaclass__ = IsAnInt
+class xint(with_metaclass(IsAnInt, builtins.int)):
     def __new__(cls, value, *a, **kw):
         if isinstance(value, gdb.Value):
             if pwndbg.typeinfo.is_pointer(value):
```

[...]

PR #128, opened 2016-10-15
make metaclass syntax in inthook.py compatible with Python 3

Isaac0616:
As I mentioned in #126, I use the future package to make metaclass syntax compatible with both Python 2 and 3.

### DIFF
--- a/pwndbg/inthook.py
+++ b/pwndbg/inthook.py
@@ -13,6 +13,7 @@ from __future__ import unicode_literals
 import sys
 
 import gdb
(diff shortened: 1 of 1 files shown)


Isaac0616:
Checks seems to fail at `isort` of `pwndbg/ui.py`, which I didn't modify.

zachriggle:
It's failing on inthook.py

Isaac0616:
Oh, sorry for overlooking the error messages. I will try to fix it.

PR #129, opened 2016-10-19
workaround for python2 unicode printing

### LINKED ISSUE #117: Theme is broken on Python2
```
$ gdb
Loaded 94 commands.  Type pwndbg [filter] for a list.
pwndbg> theme
  File "/home/riggle/pwndbg/pwndbg/commands/__init__.py", line 54, in __call__
  File "/home/riggle/pwndbg/pwndbg/commands/theme.py", line 22, in theme
    longest_value = max(map(len, [extend_value_with_default(repr(v.value), repr(v.default)) for v in values]))
UnicodeEncodeError: 'ascii' codec can't encode characters
```

anthraxx:
repr() does not work well when printing unicode sequences (at least in python2)
therefor we add a small switch case in the theme command to properly print
the values depending on their types.

Remove u'' string prefxies as they are superseded by unicode_literals

Fixes #117

### DIFF
--- a/pwndbg/__init__.py
+++ b/pwndbg/__init__.py
@@ -8,6 +8,7 @@ from __future__ import unicode_literals
 import sys
 
 import gdb
+import six
 
 import pwndbg.android
 import pwndbg.arch
@@ -72,6 +73,12 @@ try:
 except:
     pass
 
+# this is an unconventional workaround to
+# support unicode printing for python2
+# https://github.com/pwndbg/pwndbg/issues/117
+if six.PY2:
+    reload(sys)
+    sys.setdefaultencoding('utf-8')
 
 
 __all__ = [
--- a/pwndbg/color/disasm.py
+++ b/pwndbg/color/disasm.py
@@ -29,7 +29,7 @@ def branch(x):
     return generateColorFunction(config.disasm_branch_color)(x)
 
 def instruction(ins):
-    asm = u'%-06s %s' % (ins.mnemonic, ins.op_str)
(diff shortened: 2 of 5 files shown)


zachriggle:
(to Python, not to you)

Are you fucking serious?  `repr` doesn't do unicode right? *shakes head*

If this works for you, I'm down to merge.  I'll wait for @disconnect3d to confirm

pwndbg/__init__.py
```diff
@@ -72,6 +73,12 @@
 except:
     pass
 
+# this is an unconventional workaround to
+# support unicode printing for python2
+# https://github.com/pwndbg/pwndbg/issues/117
+if six.PY2:
+    reload(sys)
+    sys.setdefaultencoding('utf-8')
```
zachriggle: I think this will cause more issues than it solves when trying to render colors, no?

zachriggle: Hmm, I was assuming the escape character wasn't valid UTF-8.  It looks like `\x1b` is fine with UTF-8.

anthraxx: I see your concern, but yes that works well :smile:

PS: python2 should honestly die! :see_no_evil:

zachriggle: Python3 has its own warts.  `latin-1` solves all ills 😛

anthraxx: okay okay, fair enough :stuck_out_tongue_closed_eyes:

disconnect3d:
As you can see on the [asciicinema record](https://asciinema.org/a/dla9krlyn8hd56imwbw3ovy34) it works fine. Good job! :)

PR #134, opened 2016-11-16
adding context-clear-screen feature with config value

anthraxx:
If enabled, this clears the whole screen before printing the context.
This way stepping through the program always starts from the top-left
corner and remains static in its poisition.

Default value: off


zachriggle:
Thanks for taking care of this!

glen-mac:
This is a great addition - thanks!

Issue #131, opened 2016-10-30
Same error as issue #117 with python3 (Ubuntu 16.04, gdb 7.11.1)


PetyrBaelish:
Hi guys, 
I have a problem that seems similar to closed issue #117:

GNU gdb (Ubuntu 7.11.1-0ubuntu1~16.04) 7.11.1
...
...
Traceback (most recent call last):
  File "/home/u/tools/pwndbg/gdbinit.py", line 15, in <module>
    import pwndbg # isort:skip
  File "/home/u/tools/pwndbg/pwndbg/**init**.py", line 18, in <module>
    import pwndbg.commands
  File "/home/u/tools/pwndbg/pwndbg/commands/**init**.py", line 15, in <module>
    import pwndbg.chain
  File "/home/u/tools/pwndbg/pwndbg/chain.py", line 43, in <module>
    config_arrow_left  = theme.Parameter('chain-arrow-left', '◂—', 'left arrow of chain formatting')
  File "/home/u/tools/pwndbg/pwndbg/color/theme.py", line 17, in **init**
    'theme')
  File "/home/u/tools/pwndbg/pwndbg/config.py", line 83, in **init**
    self.value = default
UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-1: ordinal not in range(128)
...
...

> (gdb) python print(sys.version)
> 3.5.2 (default, Sep 10 2016, 08:21:44) 
> [GCC 5.4.0 20160609]

disconnect3d:
@PetyrBaelish Can you provide an example so we will be able to reproduce?
Also, which ersion of pwndbg are you using (git commit sha would be awesome)?

PetyrBaelish:
On 30.10.2016 09:39, Disconnect3d
      wrote:
    
    
      @PetyrBaelish
        Can you provide an example so we will be able to reproduce?
        Also, which ersion of pwndbg are you using (git commit sha would
        be awesome)?
      —
        You are receiving this because you were mentioned.
        Reply to this email directly, view
          it on GitHub, or mute
          the thread.
      
        
          
          
        
        
      
[...]

PR #135, opened 2016-11-16
Add check for UTF-8 encoding (#131)

disconnect3d:
Also moved workaround from #117 to common place - gdbinit.py


zachriggle:
Thanks! ❤

Issue #137, opened 2016-11-23
Generate configuration file


zachriggle:
Pwndbg has grown to have sufficiently many configuration points that perhaps it's a good idea to have a command to generate a configuration file.

The config file should be valid GDB syntax, and sets the **current** values, while showing the default values.

This would expose all of the options to new and current users in an easy-to-manage fashion.

anthraxx:
neat feature :smiley: :heart:

disconnect3d:
Where should one put the generated `configfile`?

Either I do something wrong or something is bugged.

Suppose I have a config:
```
pwndbg> config
Name                       Value (Def) Documentation
----------------------------------------------------
auto-save-search           False       automatically pass --save to "search" command
```

The `configfile` command says:
```
pwndbg> configfile
# auto-save-search: automatically pass --save to "search" command
# default: False
py pwndbg.config.auto_save_search=False
```

So I put the last three lines in my `.gdbinit` just after `source pwndbg/gdbinit.py`.

Then I restart gdb and the `configfile` command doesn't show up the `auto_save_search` configuration.

zachriggle:
In fairness, I never actually tested the config files, I just assumed they'd work as expected.

Reopened for investigation <3

disconnect3d:
@zachriggle: Fixed in 1f56e5e7a64019dea793271c6953f4ff99393059 - the problem was that we changed the `pwndbg.config.attribute` so it was no more a `pwndbg.config.Parameter` instance and so checks like `isinstance(v, pwndbg.config.Parameter)` failed.

The proposed solution fixes this with generating settings such as:
```
py pwndbg.config.ida_rpc_host.value = '192.168.0.123'
```
Instead of:
```
py pwndbg.config.ida_rpc_host='192.168.0.123'
```

(The `.value` has been added).

I am also filtering the values and generating only the changed ones. If there are no changed values additional information is shown. See the screen below.
![pwndbg_configfile](https://cloud.githubusercontent.com/assets/10009354/20868285/5b06bbd8-ba58-11e6-802e-20a5e4d75043.png)


I have also added some additional information to the `config` and `theme` commands:
![pwndbg_config](https://cloud.githubusercontent.com/assets/10009354/20868273/eb88a2d0-ba57-11e6-9104-11ae9f953460.png)


[...]

Issue #132, opened 2016-11-05
Better IDA Pro integration


disconnect3d:
Currently it is not possible to integrate Ida Pro rpc server hosted on Windows host machine with pwndbg on a Linux vm guest without some hacky changes (see below).

The problem is there because of:
* pwndbg by default connects to localhost: https://github.com/pwndbg/pwndbg/blob/master/pwndbg/ida.py#L49
* the ida server script tries to save database into a path that is invalid on Windows - https://github.com/pwndbg/pwndbg/blob/master/ida_script.py#L19 - here is a proof: https://gyazo.com/651e3fda29ecac06fb5734554c537961
* this datetime path is also problematic in here - https://github.com/pwndbg/pwndbg/blob/master/pwndbg/commands/ida.py#L94 - we probably need a possibility to provide path to the database on both machines OR send the database file between them?

I have managed to make such integration working by making such changes:
```
diff --git a/pwndbg/commands/ida.py b/pwndbg/commands/ida.py
index b2c095d..cd6f69c 100644
--- a/pwndbg/commands/ida.py
+++ b/pwndbg/commands/ida.py
@@ -88,15 +88,20 @@ def save_ida():
         os.mkdir(backups)
 
     basename, ext = os.path.splitext(basename)
-    basename += '-%s' % datetime.datetime.now().isoformat()
-    basename += ext
+    #basename += '-%s' % datetime.datetime.now().isoformat()
+    #basename += ext + '.pwndbg'
+    print("BASENAME ='%s'" % basename)
 
     # Windows doesn't like colons in paths
-    basename = basename.replace(':','_')
+    #basename = basename.replace(':','_')
 
-    full_path = os.path.join(backups, basename)
+    #full_path = os.path.join(backups, basename)
+    full_path = basename  + ext
 
     pwndbg.ida.SaveBase(full_path)
+    full_path = full_path.replace("D:", "/d")
+    full_path = full_path.replace("\\", "/")
+    print("FULLPATH = '%s'" % full_path)
 
     data = open(full_path, 'rb').read()
 
diff --git a/pwndbg/ida.py b/pwndbg/ida.py
index 82311f7..528cb74 100644
--- a/pwndbg/ida.py
+++ b/pwndbg/ida.py
@@ -43,7 +43,8 @@ xmlrpclib.Marshaller.dispatch[type(0)] = lambda _, v, w: w("<value><i8>%d</i8></
 
 def setPort(port):
     global _ida
[...]

PR #139, opened 2016-12-04
Configurable IDA Pro RPC endpoints

disconnect3d:
<strike>**NOTE: This is still WIP (work in progress). I have added it here just to be able to discuss some changes if it is required.**</strike>

NOTE 2: It is good to review this PR commit by commit instead of reviewing all changes at once. The main reason is that it contains commits that does refactoring and imports optimization (fired from PyCharm IDE, which is more or less valid with pep8 style guide).

Questions:
1. I have added some verbosity if the Ida Pro xmlrpc connection failed or not in [pwndbg/ida.py:55-59](https://github.com/pwndbg/pwndbg/compare/master...disconnect3d:ida_pro_rpc_config#diff-ba15fea58a532436f4e9caa040736272R55). However this seems to spam at gdb startup:
```
dc@dc:/d/pwndbg$ gdb /bin/ls --quiet
Pwndbg failed to connect to Ida Pro xmlrpc: http://127.0.0.1:8888
Pwndbg failed to connect to Ida Pro xmlrpc: http://127.0.0.1:8888
Pwndbg failed to connect to Ida Pro xmlrpc: http://127.0.0.1:8888
Loaded 96 commands.  Type pwndbg [filter] for a list.
pwndbg> 
```
Maybe we should add a command `ida_connect` or `connect_ida` - with host/port argument and/or setting to automatically connect on startup. What do you think?

2) Is firing reformat/imports optimization fine? We might also start using a code linter (e.g. pylint) on the CI and lint the files that has been changed by particular PR. On the other side it might be a little bit overkill or an obstacle for people who just want to fix/add something and don't really care about purity/clean code.


disconnect3d:
About 1) and `Pwndbg failed to connect to Ida Pro xmlrpc: http://127.0.0.1:8888`. 

Because this is printed/invoked during pwndbg initialization (`source pwndbg/gdbinit.py`) if I put the configuration after the initialization, the pwndbg tries to connect to the default `ida-rpc-host`.

I believe this might be problematic with other settings values as well. Maybe it would be good to store the settings differently or set them in another fashion like:
```
py pwndbg_config = {'key': 'value'}
```

And then use the `pwndbg_config` variable in pwndbg to check if user has changed the setting.

zachriggle:
I think it may also be useful to add a command e.g. `ida` which dumps `ida_script.py` to some temporary file, with the host / port stuff filled in from the config options.

It can also emit a simple command to copy/paste into the IDA interpreter, e.g. `execfile("/path/to/idascript", globals())` which should make cut-paste easier.

Also the unit tests look to be failing due to import order.

zachriggle:
Regarding the failure message, I see no reason for it to print a message on failure.  99% of the time I'm not using Pwndbg with IDA, so the error message will just be useless noise.  A positive confirmation on success should be sufficient.

anthraxx:
I don't understand this user_config.py user_theme.py files, its very bad to edit files tracked by git. also its very bad to edit them at all, as pwndbg itself may not be writable to the user. Think about a scenario where pwndbg is packages (like in Arch Linux :yum:) and you are not root.
We should consider a different approach then this

disconnect3d:
@anthraxx Oh, that's a good point.

I have described the problems with generating gdb commands by `configfile`/`themefile` in #137.

So far, the only options that comes to my mind are:
1. Set cross-global variables in `.gdbinit` [using this](http://stackoverflow.com/questions/142545/python-how-to-make-a-cross-module-variable) - so that we could read them during Parameter instances creation.
2. Generate a different file - like `.pwndbg` in user home, but that might have some other edge cases as well...

anthraxx:
maybe I'm missing something obvious, but shouldn't a Parameter extend from the gdb.Parameter and just be a runtime changeable value? The Parameter instance can safely be preserved/referenced and its actual value changed on demand.
Just use the gdb syntax to set config values:

```
set highlight-color bold,green
set highlight-pc yes
```

this works perfectly, no matter when and where i change the values, they are reflected by the references.

so intead of
```
source ~/pwndbg/gdbinit.py
py pwndbg.config.ida_rpc_host.value = '192.168.0.123'
```
just have something like
```
source ~/pwndbg/gdbinit.py
set ida_rpc_host 192.168.0.123
```

disconnect3d:
@anthraxx The problem is that some of the values are used during the initialization (`source ~/pwndbg/gdbinit.py`). So we have to fetch the values changed by the user when creating `pwndbg.config.Parameter` instances.

For now I know that it occurs with the variables I have added for ida-rpc, as there is an attempt to connect to ida xmlrpc server.

We can of course skip that as there would be reattempt to establish connection to ida xmlrpc server when stepping through the code and the proper `ida-rpc-host` value would be there, set after the initialization. However, this issue might come back in the future.

anthraxx:
Well i still think its a very limited set of things that may ever need to actually access config values during initialization phase, surely the current implementation of IDA is one of them but most config and none of the themes will ever need to.

Maybe it would be a cleaner approach to disable it by default and use our config trigger system to initialize it when the config value is set to activate it, like here:
https://github.com/pwndbg/pwndbg/blob/b1892b27741a478cd5361061b0b4dda9ef46d02e/pwndbg/hexdump.py#L31

So IDA would be deactivated while pwndbg is initializing and once the user config values are loaded and ```set ida_rpc_host 192.168.0.123``` is being used (or maybe even a boolean like ```set ida_rpc_enabled yes```) it would call the initialization function via a trigger.

For me it feels like a cleaner and more KISS solution to the very concrete problem we are facing here.

zachriggle:
It's probably best to skip the IDA RPC setup at the beginning, or at least have it fail silently on the first connection.

We also support on-config-change hooks, so we can attempt to re-connect to IDA as each of the settings change.  There's no cost for the connection attempt against localhost, since it'll fail instantly.  We can also add an explicit `ida` or `ida-connect` (name doesn't matter) which has a verbose failure message.

The reason for using `py pwndbg.config.foo = bar` instead of `set foo bar` was just to avoid needing a translation layer between the two (`yes` vs `True`, and unicode characters vs. UTF-8 byte strings).  I have nothing against changing the syntax to native GDB syntax.

zachriggle:
An alternative / additional idea for reporting IDA's current status might be to add a section to `context` which just reports the path to the current IDB.  If IDA isn't connected, it's not displayed at all.  If IDA is connected, it can display the IDB path.

disconnect3d:
I rebased the branch and made the commits more atomic.

Some summary of the changes:
1. Added the Python value -> native gdb syntax translation (as a property) and changed the `configfile`/`themefile` to generate gdb commands that you put after pwndbg initialization in `.gdbinit` - see 9c6314c4ebd961b037a4cf986a92cd9716fbc4ba).
The translation might require some more work. I am not sure/aware of all the possible types yet (it seems to be a good candidate for a testcase).
Also there is one thing that I would just accept for now - the `set` command in `.gdbinit` prints a message that it has set a particular parameter... this might be repressed by [setting logging to /dev/null](http://stackoverflow.com/questions/11957417/gdb-do-not-print-output-of-gdbinit-commands). However, that would require to generate the logging commands along with configfile/themefile. Maybe there is a better way?

2. Added possibility to set Ida Pro xmlrpc host/port. Currently, the auto connect is done as a trigger (so when pwndbg is initialized, there is still an attempt to connect to localhost).
Currently there is only a message on a successful connection.
Also it seems that we should set some kind of timeout as if the host parameter has been set and the server is not running you have to wait for ages till it timeouts.

3. Fixed the bug with wrong idb path generation on Windows in ida_script.py (due to ADS NTFS; more [here](https://github.com/pwndbg/pwndbg/issues/132#issuecomment-264697778)) - see 627a0517aa8cccd813e41698d9d6450d0b34322c.

3. Extended FEATURES.md with information about Ida Pro xmlrpc connection settings and config/theme/configfile/themefile commands. See [it live](https://github.com/disconnect3d/pwndbg/blob/ida_pro_rpc_config/FEATURES.md#configuration-customization).

I am waiting for comments/review.

If the general approach is okay, maybe we can go with merging those changes and then revisiting the `set variable` printing and ida connection timeout problems?

pwndbg/commands/config.py
```diff
@@ -42,24 +45,37 @@ def config():
...
     """Generates a configuration file for the current Pwndbg theme options"""
     configfile_print_scope('theme')
 
+
 def configfile_print_scope(scope):
-    values = [v for k, v in pwndbg.config.__dict__.items()
-              if isinstance(v, pwndbg.config.Parameter) and v.scope == scope]
-    longest_optname = max(map(len, [v.optname for v in values]))
-    longest_value = max(map(len, [extend_value_with_default(repr(v.value), repr(v.default)) for v in values]))
+    filter_cond = lambda v: \
```
zachriggle: Separately, I agree with the style guide in PEP8 that suggests breaking *before* the operator.

```
if foo \
and bar:
```

versus:

```
if foo and \
bar:
```

(per https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator)

pwndbg/ida.py
```diff
@@ -159,41 +183,46 @@ def Anterior(addr):
...
 def UpdateBreakpoints():
     # XXX: Remove breakpoints from IDA when the user removes them.
     current = set(eval(b.location.lstrip('*')) for b in _breakpoints)
-    want    = set(GetBreakpoints())
+    want = set(GetBreakpoints())
 
     # print(want)
 
-    for addr in current-want:
+    for addr in current - want:
```
zachriggle: It looks like this was fed through a PEP8 formatter which doesn't follow the rules suggested in #141.  Your editor should be configured to ignore E226/E227.

pwndbg/ida.py
```diff
@@ -109,42 +126,49 @@ def base():
...
+    rv = _ida.GetFuncOffset(addr)
     return rv
 
+
 @withIDA
 @takes_address
 @pwndbg.memoize.reset_on_objfile
 def GetType(addr):
-    rv =  _ida.GetType(addr)
+    rv = _ida.GetType(addr)
```
zachriggle: E221

pwndbg/ida.py
```diff
@@ -109,42 +126,49 @@ def base():
...
 def Name(addr):
     return _ida.Name(addr)
 
+
 @withIDA
 @takes_address
 @pwndbg.memoize.reset_on_objfile
 def GetFuncOffset(addr):
-    rv =  _ida.GetFuncOffset(addr)
+    rv = _ida.GetFuncOffset(addr)
```
zachriggle: E221

pwndbg/ida.py
```diff
@@ -159,41 +183,46 @@ def Anterior(addr):
...
+
 
 @pwndbg.events.cont
 @pwndbg.events.stop
 @withIDA
 def UpdateBreakpoints():
     # XXX: Remove breakpoints from IDA when the user removes them.
     current = set(eval(b.location.lstrip('*')) for b in _breakpoints)
-    want    = set(GetBreakpoints())
+    want = set(GetBreakpoints())
```
zachriggle: E226/E227 is OK

pwndbg/config.py
```diff
@@ -64,50 +65,66 @@ def __call__(self, function):
...
 
+class Parameter(gdb.Parameter):
     def __init__(self, name, default, docstring, scope='config'):
         self.docstring = docstring.strip()
-        self.optname   = name
-        self.name      = name.replace('-','_')
-        self.default   = default
-        self.set_doc   = 'Set ' + docstring
-        self.show_doc  = docstring + ':'
+        self.optname = name
```
zachriggle: E221 is OK

pwndbg/ida.py
zachriggle: I'm adding `setup.cfg` to the project root, which contains the `pep8.py` options.

Since `pep8.py` picks these up automagically, so should PyCharm.

pwndbg/ida.py
zachriggle: Note: Not asking you to go back and un-fix, just pointing it out <3

pwndbg/ida.py
zachriggle: Note: Not asking you to go back and un-fix, just pointing it out <3

pwndbg/ida.py
zachriggle: Note: Not asking you to go back and un-fix, just pointing it out <3

pwndbg/ida.py
zachriggle: Note: Not asking you to go back and un-fix, just pointing it out <3

pwndbg/commands/ida.py
```diff
@@ -110,20 +112,23 @@ def save_ida():
...
         super(ida, self).__init__('ida')
+
     def invoke(self, name):
-        name   = name.string()
+        name = name.string()
         result = pwndbg.ida.LocByName(name)
 
-        if 0xffffe000 <= result <= 0xffffffff \
-        or 0xffffffffffffe000 <= result <= 0xffffffffffffffff:
+        if 0xffffe000 <= result <= 0xffffffff or 0xffffffffffffe000 <= result <= 0xffffffffffffffff:
```
zachriggle: It looks like we hit E122 the way it is.  From grepping for other E122 errors, it looks like it's not something we really want anyway, since it complains profusely about `dict` and `list` ordering, e.g.: https://github.com/pwndbg/pwndbg/blob/master/pwndbg/__init__.py#L76

zachriggle: Meh, probably just easiest to break this into two statements for readability.  Don't care enough to figure out how to make PEP8 happy and the code also be readable.

disconnect3d:
About https://github.com/pwndbg/pwndbg/pull/139#discussion_r92273623

Yeah. We should probably keep to one convention everywhere (e.g. in `config` or `theme` output).

I have added the `native_default` and fixed the ugly `filter_cond` (not sure if that's what you meant).

Anyway, to be honest I don't really like the gdb convention as I am not familiar with it.

As far as I understand a trigger is invoked whenever `set <var> = value` is launched.
Maybe we should change the triggers so that they would fire whenever the `value` attribute has changed (and so make it a property instead of pure field).

With that we probably could do:
`py pwndbg.config.parameter.value = python_value`

in the `.gdbinit` after the pwndbg initialization is done and that would be much easier - that would let us forget about `native_value`, `native_default`, translating the values between Python and gdb syntax (and other issues connected to this like unicode).

@zachriggle and @anthraxx What do you think about changing the triggers to fire them when `.value` is changed and so change `configfile`/`themefile` to generate python invocations instead of gdb's `set ...`?

zachriggle:
Re: firing config-change events in the `.value` setter, I'm down for that.  I believe my original assumption was that GDB would do that for us, but that's clearly false.

anthraxx:
donno... i really like the gdb syntax to setup the config values, they are much more human intractable/readable then using inline python... and at the end such config file is faced by humans.

I understand the silliness wit unicode, but it kind of feels inconsistent and ugly/unreadable to generate such configs compared to human usable format

zachriggle:
Another issue to consider with using GDB syntax is that its `input-radix` can truly mess everything up.  By using Python, we can avoid that particular issue.  If we stick with GDB, all integers will have to be appropriately prefixed (i.e. `0d64` vs `0x40`; `40` can mean either `0d40` or `0x40` depending on `input-radix`).

anthraxx:
I'm on a summit right now so hope you are patient enough to wait a bit... but I played around a bit by extending it slightly... and i don't think it will be hard to properly support it.

the cost benefit seems reasonable from a dev side (especially when we have great revewiws) compared to the end-user facing a hex string instead of beautiful unicode chars.

will add my tiny changes later today :smile:

zachriggle:
FWIW, there's nothing stopping us from using the pretty Unicode characters in Python, either 👅

anthraxx:
so here are my slightly roundup proposals, one fixes a bug with the type():

https://github.com/anthraxx/pwndbg/commit/e58071ce4c57d5b824b6cc319dd868771f026536
https://github.com/anthraxx/pwndbg/commit/e158afc5ec0f56acee1e93c3979895c6bc89fd9b
https://github.com/anthraxx/pwndbg/commit/70919fd932a63e3a0d022bb62d3aa5c8e7859be4

zachriggle:
I don't really see the benefit of an explicit function instead of read-only properties.  The first two, I like.

The last (https://github.com/anthraxx/pwndbg/commit/70919fd932a63e3a0d022bb62d3aa5c8e7859be4) shouldn't be included.  IMHO the code is more readable without that change.

anthraxx:
Sure, that's more like style then actual stuff. Just felt inappropriate to have additional properties doing a very specific job, no strong opinions on that one. Nuke it :P

zachriggle:
@anthraxx just stick it in a pull requeset <3

anthraxx:
Can do, in like aprox 5h.. Out for some party right now :D

zachriggle:
No rush.  Have an extra pint for me!


*Zach Riggle*

anthraxx:
did have two pints extra just for you :beers: 
anyway, creating a pull request will need to be based on these changes, therefor its better to wait until they hit master.... or you apply them by pulling them in from my remote :yum:

disconnect3d:
> you apply them by pulling them in from my remote

Will do that right now

Issue #68, opened 2016-06-10
heap command displays unnecessary data


zachriggle:
Given a simple binary below, breaking at function `B` produces the output below.  

The flags must be incorrect, as they are all marked `PREV_INUSE`.  The first entry has no `prev`, so that must be incorrect.

All of the entries display `fd` and `bk` and `fd_nextsize` and `bk_nextsize` even if the fields don't make sense to display in these instances.

We should not display fields that are unnecessary.  This means diverging from usage of GDB-provided symbols and type info, and declaring our own, similar to how it's done in GEF (#30), libheap (#5) and gdb-heap (#4).

```
pwndbg> break B
...
pwndbg> run
...
pwndbg> heap
Top Chunk: 0x6020a0
Last Remainder: 0

0x602000 PREV_INUSE
{
  prev_size = 0x0,
  size = 0x31,
  fd = 0xaaaaaaaaaaaaaaaa,
  bk = 0xaaaaaaaaaaaaaaaa,
  fd_nextsize = 0xaaaaaaaaaaaaaaaa,
  bk_nextsize = 0xaaaaaaaaaaaaaaaa
}
0x602030 PREV_INUSE
{
  prev_size = 0x700000000000004,
  size = 0x21,
  fd = 0xaaaaaaaaaaaaaaaa,
  bk = 0xaaaaaaaaaaaaaaaa,
  fd_nextsize = 0x700000000000002,
  bk_nextsize = 0x51
}
0x602050 PREV_INUSE
{
  prev_size = 0x700000000000002,
  size = 0x51,
  fd = 0xaaaaaaaaaaaaaaaa,
  bk = 0xaaaaaaaaaaaaaaaa,
  fd_nextsize = 0xaaaaaaaaaaaaaaaa,
  bk_nextsize = 0xaaaaaaaaaaaaaaaa
}
0x6020a0 PREV_INUSE
{
  prev_size = 0x70000000000000e,
  size = 0x20f61,
  fd = 0x0,
  bk = 0x0,
  fd_nextsize = 0x0,
  bk_nextsize = 0x0
}
```

Example binary:

```c
#include <malloc.h>

void A() {}
void B() {}
void C() {}
void D() {}
void E() {}
void F() {}

int main() {
    A();

    void* a = malloc(32);
    void* b = malloc(16);
    void* c = malloc(64);

    B();

    free(b);

    C();

    free(a);

    D();

    void* d = malloc(48);

    E();

    free(d);
    free(c);

[...]

PR #140, opened 2016-12-08
Added heap operations

blendin:
### Completed
- Implemented command **unsortedbin** to print the unsorted bin.
- Implemented command **smallbins** to print small bins.
- Implemented command **largebins** to print large bins.
- Implemented command **fastbins** to print the fast bins.
- Completed the **bin** command to print all bins.
- Changed heap, arena, and top_chunk commands to use the current heap implementation.
- Made bin commands default to only show bins in use (linked lists with more than one node).
- Added an extra parameter to **pwndbg.chain.get** that specifies a value that when reached stops the chain from continuing. 
- Augmented `pwndbg.chain.format` to accept lists (result from pwndbg.chain.get) as an argument.
- Use `malloc_par` struct (**mp_**) from glibc to determine _sbrk_base_ so heap base is more reliable. This fixes the bug in some ubuntu versions where a large _.bss_ can cause **[heap]** in _/proc/$(pid)/maps_ to be different from the actual start of the ptmalloc heap.
- Added a ptmalloc constants file to the constants/ directory, the spaces tables were brute forced using `largebin_index_32` and `largebin_index_64` functions from glibc's malloc.c

### Future work
- Add function to ptmalloc's Heap class that prints out a full representation of the heap.
- Implement other heap interfaces such as dlmalloc, tcmalloc, jemalloc.
- Trace calls to malloc, free, and malloc_consolidate.
- Automatically detect broken links in bins.
- Add flags for verbose and arena_addr to heap commands.


pwndbg/heap/dlmalloc.py
```diff
@@ -10,12 +10,7 @@
...
-malloc_state = None
-mallinfo     = None
+class Heap(pwndbg.heap.heap.Heap):
 
-@pwndbg.events.new_objfile
-def update():
-  malloc_chunk = gdb.lookup_type('struct malloc_chunk')
-  malloc_state = gdb.lookup_type('struct malloc_state')
-  mallinfo = gdb.lookup_type('struct mallinfo')
+    def __init__(self):
```
disconnect3d: I understand that this is still work in progress so jfyi - you can write an empty class using `pass`:

```python
In [1]: class A(object):
   ...:     pass
   ...: 

In [2]: a=A()

In [3]: 

```

pwndbg/heap/ptmalloc.py
```diff
@@ -5,17 +5,217 @@
...
+
+        # Symbols and types
+        self._malloc_chunk  = None
+        self._malloc_state  = None
+        self._mallinfo      = None
+        self._malloc_par    = None
+
+    @property
+    def main_arena(self):
+        from pwndbg.color import bold
```
disconnect3d: Is there any reason to keep the imports in here? If not, we should probably move them upper.

blendin:
@disconnect3d great suggestions! Will make these changes

pwndbg/heap/ptmalloc.py
```diff
@@ -5,17 +5,217 @@
...
+        from pwndbg.color import red
+
+        if not self._main_arena:
+            main_arena_symbol = gdb.lookup_symbol('main_arena')[0]
```
zachriggle: We should not use `gdb.lookup_symbol`, you should use `pwndbg.symbol.address()`.

pwndbg/constants/ptmalloc.py
```diff
@@ -0,0 +1,116 @@
...
+    are spaces 2 * SIZE_SZ bytes apart
+
+    Large bins are approximately logorithmaclly spaces with a bit of slop
+    to the actually numbers between spacing for the sake of speed.
+
+"""
+
+spaces_table_32 = [
+                   None, # there is no index 0
+                   8, 8, 8, 8, 8, 8, 8, 8,
```
zachriggle: I'd greatly prefer something like:

```py
spaces_table_32 = [ None ] + [ 8 ] * 64 + [64] * 32 + ...
```

It's much more readable.

Separately, rather than having two duplicate, hard-coded tables, it's probably better to have a function which derives them from the current architecture.  For example...

```
def get():
    return [ None ] + [ 2 * pwndbg.arch.ptrsize ] * 64 + ...
```

Though it appears there are small variances in what comes after the section of `[64]`, so perhaps the second half of this suggestion isn't useful.

pwndbg/commands/heap.py
```diff
@@ -166,23 +122,112 @@ def malloc_chunk(addr):
...
     chunk = value_from_type('struct malloc_chunk', addr)
     size = int(chunk['size'])
-    prev_inuse = (size & PREV_INUSE) == 1
-    is_mmaped = (size & IS_MMAPED) == 1
-    non_main_arena = (size & NON_MAIN_ARENA) == 1
+
+    prev_inuse, is_mmapped, non_main_arena = main_heap.chunk_flags(size)
 
     header = M.get(addr)
     if prev_inuse:
```
gsingh93: I don't think the `prev_inuse` is used for fastbins. Should we not display this if the size is in the fastbin range? #68 references this.

pwndbg/heap/__init__.py
```diff
@@ -5,19 +5,26 @@
...
-      current = pwndbg.heap.dlmalloc.Heap()
+    if pwndbg.symbol.address('ptmalloc_init'):
+        current_heap = pwndbg.heap.ptmalloc.Heap()
+
+    else:
+        # Default to ptmalloc heap for now until
+        # there are more implementations
+        current_heap = pwndbg.heap.ptmalloc.Heap()
+
+def get_heap():
```
zachriggle: I prefer the `current` global which is updated via the event function, to a `getter` function which returns the global.

pwndbg/heap/ptmalloc.py
```diff
@@ -5,17 +5,217 @@
...
+from collections import OrderedDict
+
 import gdb
 
 import pwndbg.events
 import pwndbg.typeinfo
+from pwndbg.constants.ptmalloc import *
+
+
+class Heap(pwndbg.heap.heap.Heap):
```
zachriggle: There are currently very few `import *` in the code base, which I'd prefer to keep to a minimum.  Long package names don't bother me so much `pwndbg.heap.heap.Heap`, however I might recommend renaming to e.g. `pwndbg.heap.base.BaseHeap`.

pwndbg/heap/ptmalloc.py
```diff
@@ -5,17 +5,217 @@
...
+        return self._malloc_par
+
+    def chunk_flags(self, size):
+        return ( size & PREV_INUSE, size & IS_MMAPPED, size & NON_MAIN_ARENA )
+
+    def get_arena(self, arena_addr=None):
+        if arena_addr is None:
+            return self.main_arena
+
+        return gdb.Value(arena_addr).cast(self.malloc_state.pointer()).dereference()
```
zachriggle: `pwndbg.memory.po(type, addr)` makes this much simpler: `return pwndbg.memory.poi(self.malloc_state, arena_addr)`

pwndbg/heap/ptmalloc.py
```diff
@@ -5,17 +5,217 @@
...
+        return (None, None)
+
+    def fastbins(self, arena_addr=None):
+        arena        = self.get_arena(arena_addr)
+
+        if arena == None:
+            return
+
+        fastbinsY    = arena['fastbinsY']
+        fd_offset    = self.malloc_chunk.keys().index('fd') * SIZE_SZ
```
zachriggle: See my recommendations about SIZE_SZ vs `pwndbg.arch`.

pwndbg/heap/ptmalloc.py
```diff
@@ -5,17 +5,217 @@
...
+            chain = self.bin_at(index, arena_addr=arena_addr)
+
+            if chain == None:
+                return
+
+            result[size] = chain
+
+        return result
+
+    def format_bin(self, bins, verbose=False):
```
zachriggle: IMHO any display-related code should be in another module.  I don't want to mix display code with logic code, at least where it can be avoided.  I'd suggest `pwndbg.commands.heap` or similar, perhaps a submodule if you like.

pwndbg/heap/ptmalloc.py
zachriggle: I've just added code to `master` which uses `lookup_symbol` in `pwndbg.symbol.address`

pwndbg/heap/ptmalloc.py
```diff
@@ -53,38 +52,70 @@ def mp(self):
...
+                      + [ pow(2, 9) ]                    * 16 \
+                      + [ pow(2, 12) ]                   * 8  \
+                      + [ pow(2, 15) ]                   * 4  \
+                      + [ pow(2, 18) ]                   * 2  \
+                      + [ pow(2, 21) ]                   * 1
+
+        # There is no index 0
+        spaces_table = [ None ] + spaces_table
+
+        # Fix up the slop in bin spacing ( part of libc - they made
```
disconnect3d: Hmm... do you have some reference for this? Like url, line of particular file or commit hash?

Mayyyybe we can add such info to this comment. However if it is an url it might be broken later...

blendin: Here is the line I'm referring to
https://github.com/andigena/glibc-2.23-0ubuntu3/blob/master/malloc/malloc.c#L1461

pwndbg/commands/heap.py
blendin: heap_fastbins

I implemented a check in the`malloc_chunk` command. If it is a fastbin then display that instead of PREV_INUSE. To figure out if the previous heap chunk is in use when a fastbin is after it, I believe we can go up the heap until we see the next chunk that isn't a fastbin.

blendin:
Some screenshots as well
unsortedbins
bins

pwndbg/heap/ptmalloc.py
disconnect3d: @blendin Nice. This program could be used in tests.

blendin: Here is the line in glibc specifying bin[0] doesn't exist as well
https://github.com/andigena/glibc-2.23-0ubuntu3/blob/master/malloc/malloc.c#L1402

disconnect3d:
@blendin Can you add those screenshots and a short description to `FEATURES.md`?

pwndbg/heap/ptmalloc.py
disconnect3d: We could add those links to commit descriptions - so that when someone do `git blame` it will be there for faster reference.

Maybe also to comments in the code - however I am not sure about it.

@zachriggle What do you think?

zachriggle:
I'd prefer if the bin sizes were printed in hex, since that's how we tend to print things (e.g. `telescope` / `stack` view).

gsingh93:
@blendin Like we discussed on Slack, could you make the `<- 0x0` value of the chain not point to the top of the chunk but point to the `fwd` pointer instead?

pwndbg/heap/ptmalloc.py
zachriggle: I don't believe the bin design has changed at all since 2000.  It uses the same bin design as the original `dlmalloc`.

It looks like `dlmalloc` 2.6.6 was the first version to have exactly that macro (`bin_at`) with that description:
http://g.oswego.edu/pub/misc/malloc-2.6.6.c

Since Ubuntu is our target, I'd rather link directly to their release, as far as "linking to documentation" goes.  For example:

https://bazaar.launchpad.net/~ubuntu-branches/ubuntu/trusty/eglibc/trusty-security/view/head:/malloc/malloc.c#L1356

pwndbg/commands/heap.py
zachriggle: Also, if there's any way to save a bunch of lines in the above screenshot by moving the opening brace onto the previous line, that'd be greeeeeeeaaaaaat </Lumberg>.

pwndbg/chain.py
```diff
@@ -44,8 +48,14 @@ def get(address, limit=LIMIT, offset=0):
 config_arrow_right = theme.Parameter('chain-arrow-right', '—▸', 'right arrow of chain formatting')
 config_contiguous  = theme.Parameter('chain-contiguous-marker', '...', 'contiguous marker of chain formatting')
 
-def format(value, limit=LIMIT, code=True, offset=0):
-    chain = get(value, limit, offset)
+def format(value, limit=LIMIT, code=True, offset=0, hard_stop=None, hard_end=0):
```
blendin: Will do

pwndbg/heap/ptmalloc.py
zachriggle: Per Slack discussion, a in-source comment / link is sufficient.

pwndbg/commands/heap.py
disconnect3d: Or even omitting the braces, do we really need em? :p

pwndbg/chain.py
blendin: Added a docstring to the format function [here](https://github.com/blendin/pwndbg/blob/ptmalloc/pwndbg/chain.py#L60). Modeled after the docstring for the enhance function.

pwndbg/constants/ptmalloc.py
```diff
@@ -0,0 +1,30 @@
...
+import pwndbg.arch
+
+# Heap flags
+PREV_INUSE      = 1
+IS_MMAPPED      = 2
+NON_MAIN_ARENA  = 4
+SIZE_BITS       = ( PREV_INUSE | IS_MMAPPED | NON_MAIN_ARENA )
+
+# Constants and macros taken from glibc malloc.c
+MIN_CHUNK_SIZE      = pwndbg.arch.ptrsize * 4
```
zachriggle: `pwndbg.arch.ptrsize` is not a global constant, this will change based on the current binary (e.g. `set arch`).  This unfortunately has to be a function, or needs an additional `event` to update the global.  See how `pwndbg.arch` handles this.

zachriggle: Same with` _ALLIGNMENT` (which is also typo'd) and `SMALLBIN_CORRECTION`

pwndbg/heap/ptmalloc.py
```diff
@@ -5,17 +5,244 @@
...
+        if mp_addr is not None:
+            self._mp = pwndbg.memory.poi(self.malloc_par, mp_addr)
+
+        return self._mp
+
+
+    @property
+    def malloc_chunk(self):
+        if not self._malloc_chunk:
+            self._malloc_chunk = pwndbg.typeinfo.load('struct malloc_chunk')
```
zachriggle: This can't realistically be cached, e.g. if I change the binary I'm debugging without killing GDB, this will be wrong.  I'd recommend just always returning the `typeinfo.load` object, and perhaps using `memoize.reset_on_objfile` to hold the value as cached ~= most of the time.

zachriggle: Same with all of the `@property` things above and below, just add `@pwndbg.memoize.reset_on_objfile` underneath `@property` and remove the underscore-prefixed caching / lazy initialization.

zachriggle:
Just waiting on the `memoize` updates (see review), and fixing the merge conflicts.

zachriggle:
All good? Let me know if this is ready to merge

anthraxx:
@zachriggle i didn't had the time to functionally test it, but from pure reading the code, +1 from me

blendin:
It is good on my 14.04 dev box with and without libc debugging symbols. I would wait until someone else tests it out.

PR #142, opened 2016-12-09
Fix windbg dX commands with explicit count

joeleong:
Tiny fix. Thanks for the tool!

```
pwndbg> dd 0 0
'gdb.Value' object cannot be interpreted as an integer
'dd': Starting at the specified address, dump N dwords
    (default 16).
```

### DIFF
--- a/pwndbg/commands/windbg.py
+++ b/pwndbg/commands/windbg.py
@@ -78,6 +78,7 @@ def dX(size, address, count, to_string=False):
     values = []
     address = int(address) & pwndbg.arch.ptrmask
     type   = get_type(size)
+    count = int(count)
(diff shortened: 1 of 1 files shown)


pwndbg/commands/windbg.py
```diff
@@ -78,7 +78,7 @@ def dX(size, address, count, to_string=False):
     values = []
     address = int(address) & pwndbg.arch.ptrmask
     type   = get_type(size)
-    for i in range(count):
+    for i in range(int(count)):
```
zachriggle: We use `count` directly below.  It's probably best to do `count = int(count)` above, rather than inline here.

PR #146, opened 2017-01-03
Fix exception when 'print elements' is unlimited. (Fixes #145)

### LINKED ISSUE #145: "show print elements 0" causes exceptions on stop
```
pwndbg> show print elements
Limit on string chars or array elements to print is unlimited.
  File "/home/david/.pwndbg/pwndbg/events.py", line 111, in caller
    func()
  File "/home/david/.pwndbg/pwndbg/strings.py", line 34, in update_length
    length  = int(message)
  File "/home/david/.pwndbg/pwndbg/inthook.py", line 44, in __new__
ValueError: invalid literal for int() with base 10: 'unlimited'
Python Exception <class 'ValueError'> invalid literal for int() with base 10: 'unlimited': 
```

Matir:
This now honors unlimited print elements by returning the full string in that case.

### DIFF
--- a/pwndbg/strings.py
+++ b/pwndbg/strings.py
@@ -31,13 +31,10 @@ def update_length():
     message = gdb.execute('show print elements', from_tty=False, to_string=True)
     message = message.split()[-1]
     message = message.strip('.')
-    try:
-        length  = int(message)
-    except ValueError:
-        if message == 'unlimited':
-            length = 0
-        else:
-            raise
+    if message == 'unlimited':
+        length = 0
+    else:
+        length = int(message)
 
 def get(address, maxlen = None):
     if maxlen is None:
--- a/pwndbg/strings.py
+++ b/pwndbg/strings.py
@@ -34,7 +34,7 @@ def update_length():
     try:
         length  = int(message)
     except ValueError:
-        if message == 'unlimited:
+        if message == 'unlimited':
             length = 0
         else:
             raise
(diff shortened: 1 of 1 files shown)


pwndbg/strings.py
```diff
@@ -31,7 +31,13 @@ def update_length():
     message = gdb.execute('show print elements', from_tty=False, to_string=True)
     message = message.split()[-1]
     message = message.strip('.')
-    length  = int(message)
+    try:
+        length  = int(message)
+    except ValueError:
+        if message == 'unlimited:
```
anthraxx: ```SyntaxError: EOL while scanning string literal```
I have the feeling this code wasn't tested much :stuck_out_tongue_closed_eyes: 

it would make more sense to first check for 'unlimited' before trying to make an integer out of it. I don't think the try-except is needed at all

pwndbg/strings.py
```diff
@@ -31,7 +31,13 @@ def update_length():
     message = gdb.execute('show print elements', from_tty=False, to_string=True)
     message = message.split()[-1]
     message = message.strip('.')
-    length  = int(message)
+    try:
+        length  = int(message)
+    except ValueError:
+        if message == 'unlimited':
```
anthraxx: it would make more sense to first check for 'unlimited' before trying to make an integer out of it. I don't think the try-except is needed at all

Matir: That's a reasonable approach, I'll update.

PR #153, opened 2017-01-18
Update __init__.py

diouziou:
Hi

here's a fix for the prompt. This should fix the problems one gets when entering some commands of various length in pwndbg, then navigating through these commands via up and down arrow keys.

This could also all be done with `set extended-prompt` but i have stuck to the coding style

Let me at the same time say thx for the good work that is pwndbg!


zachriggle:
Taking this on blind faith, thanks for the contribution!

diouziou:
You're welcome :)

PR #157, opened 2017-02-02
wrong dissassembler mode choice

g05u:
This will cause to choose a bad dissassembler so the output (context code) is not shown


zachriggle:
Thanks! I'll merge this back into `master` later

PR #164, opened 2017-02-14
Added commands for xorring regions of memory

blendin:
xor: xor a region of memory with a key
    memfrob: frobnicate a memory area (https://linux.die.net/man/3/memfrob)


zachriggle:
I like how I never actually finished the XOR command apparently 😂

PR #166, opened 2017-02-20
fix NoneType access if emulator not available through import

anthraxx:
This fixes a regression when the emulator support is disabled via
setting it to None when the import fails. The access happens when
checking if a particular architecture actually supports emulation.

Ensure that emulate has the correct value depending on the availability
of the import and deactivate it otherwise. The following core can safely
rely on the boolean state of the emulate variable.


zachriggle:
I'll figure out why CI is broken and merge this, then add a CI test that doesn't have Unicorn so we can catch this in the future (maybe)

PR #170, opened 2017-02-28
Add RPC-based tests

zachriggle:
Allow testing Pwndbg itself by embedding an XMLRPC server inside the GDB instance, which exposes both the `gdb` and `pwndbg` modules, as well as `eval`.

Since the default XMLRPC behavior is terrible and doesn't know how to marshall any of the following:

- `bytearray` object
- `bytes` object
- `str` or `unicode` with embedded `\x00`

We hack both the Marshaller and Unmarshaller to support a new datatype.

### DIFF
--- a/pwndbg/__init__.py
+++ b/pwndbg/__init__.py
@@ -63,6 +63,7 @@ import pwndbg.net
 import pwndbg.proc
 import pwndbg.prompt
 import pwndbg.regs
+import pwndbg.server
 import pwndbg.stack
 import pwndbg.stdio
 import pwndbg.typeinfo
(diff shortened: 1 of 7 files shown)


disconnect3d:
@zachriggle What is status of this task?
Do you need help with that?

zachriggle:
This is stalled on hacking some things into the XMLRPC layer.

Currently, there's no mechanism that you can just invoke an RPC method which returns a Python3 `bytes` object, because the XMLRPC spec does not know about it.

I added some stuff to hack in support, but it's not complete yet.

zachriggle:

PR #176, opened 2017-03-10
Use "pip" versions of Capstone and Unicorn

zachriggle:
This removes the Git submodules and explicit dependencies on a individual build of Capstone and Unicorn.  This should also greatly simplify the installation process.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -1,6 +1,15 @@
 #!/bin/bash
 set -ex
 
+PYTHON=''
+INSTALLFLAGS=''
+
+if [ "$1" == "--user" ] || (uname | grep -i Darwin &>/dev/null); then
+    INSTALLFLAGS="--user"
+else
+    PYTHON="sudo "
+fi
+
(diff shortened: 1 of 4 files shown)


setup.sh
```diff
@@ -1,6 +1,15 @@
 #!/bin/bash
 set -ex
 
+PYTHON=''
+INSTALLFLAGS=''
+
+if [ "$1" == "--user" ] || (uname | grep -i Darwin &>/dev/null); then
```
disconnect3d:
```bash
(uname | grep -i Darwin &>/dev/null)
```
Is this for checking whether we are on a Mac/OSX?

If so, maybe move it to a function:
```bash
is_osx() {
  uname | grep -i Darwin &>/dev/null
}
```

zachriggle:
Verified that this works now on a new Mac laptop, and works for system install on Ubuntu, and works via `bash setup.sh --user` on Ubuntu as well.

PR #184, opened 2017-03-19
add config for amount of stack, code context and nearpc lines

anthraxx:
Adding options so all commands and context lines can be configured independently. For more human friendly behavior of the lines value, we internally divide it by 2 as it functions as a value for before+after current pc.

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -148,15 +148,16 @@ def get_regs(*regs):
 pwndbg.config.Parameter('emulate', True, '''
 Unicorn emulation of code near the current instruction
 ''')
+code_lines = pwndbg.config.Parameter('context-code-lines', 10, 'number of additional lines to print in the code context')
 
 def context_code():
     banner = [pwndbg.ui.banner("code")]
     emulate = bool(pwndbg.config.emulate)
-    result = pwndbg.commands.nearpc.nearpc(to_string=True, emulate=emulate)
+    result = pwndbg.commands.nearpc.nearpc(to_string=True, emulate=emulate, lines=int(code_lines) // 2)
 
     # If we didn't disassemble backward, try to make sure
     # that the amount of screen space taken is roughly constant.
-    while len(result) < 11:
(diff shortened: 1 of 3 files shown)


pwndbg/commands/context.py
```diff
@@ -148,15 +148,16 @@ def get_regs(*regs):
 pwndbg.config.Parameter('emulate', True, '''
 Unicorn emulation of code near the current instruction
 ''')
+code_lines = pwndbg.config.Parameter('context-code-lines', 10, 'number of additional lines to print in the code context')
 
 def context_code():
     banner = [pwndbg.ui.banner("code")]
     emulate = bool(pwndbg.config.emulate)
-    result = pwndbg.commands.nearpc.nearpc(to_string=True, emulate=emulate)
+    result = pwndbg.commands.nearpc.nearpc(to_string=True, emulate=emulate, lines=int(int(code_lines) / 2))
```
zachriggle: You should be able to replace the second `int` call with integer division `//`

pwndbg/commands/nearpc.py
```diff
@@ -53,7 +54,7 @@ def nearpc(pc=None, lines=None, to_string=False, emulate=False):
         pc = pwndbg.regs.pc
 
     if lines is None:
-        lines = 5
+        lines = int(int(nearpc_lines) / 2)
```
zachriggle: `int(...) // 2` vs `int(int(...) / 2)`

pwndbg/commands/context.py
zachriggle: It may be desirable to extend the config class to have __div__ and __idiv__ and so-on methods to avoid the need to explicitly cast these

anthraxx: sounds cool, will do so in a separate commit

pwndbg/commands/nearpc.py
anthraxx: thats more readable, gonna change commit

anthraxx:
added the arithmetic operators for more convenient usage :smiley_cat:

PR #185, opened 2017-03-19
command robustness improvements and graceful errors

anthraxx:
- more robust and graceful error messages for malformed input
- fix word search by using uniform search type parameter
- graceful error messages on invalid memory regions and invalid count parameters
- only allow retaddr and canary command in running mode
- next commands: check if proc is alive while continuing to avoid exceptions

### DIFF
--- a/pwndbg/commands/hexdump.py
+++ b/pwndbg/commands/hexdump.py
@@ -6,6 +6,7 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import argparse
+import gdb
 
 import pwndbg.arch
 import pwndbg.commands
@@ -32,13 +33,17 @@ parser.add_argument('count', nargs='?', default=pwndbg.config.hexdump_bytes,
 def hexdump(address=None, count=pwndbg.config.hexdump_bytes):
     address = int(address)
     address &= pwndbg.arch.ptrmask
-    count   = int(count)
+    count   = max(int(count), 0)
     width   = int(pwndbg.config.hexdump_width)
 
     if count > address > 0x10000:
         count -= address
 
-    data = pwndbg.memory.read(address, count, partial=True)
+    try:
+        data = pwndbg.memory.read(address, count, partial=True)
+    except gdb.MemoryError as e:
+        print(e)
+        return
 
     for line in pwndbg.hexdump.hexdump(data, address=address, width=width):
         print(line)
--- a/pwndbg/commands/hexdump.py
+++ b/pwndbg/commands/hexdump.py
@@ -41,7 +41,7 @@ def hexdump(address=None, count=pwndbg.config.hexdump_bytes):
 
     try:
         data = pwndbg.memory.read(address, count, partial=True)
-    except gdb.MemoryError as e:
+    except gdb.error as e:
         print(e)
         return
 
--- a/pwndbg/commands/next.py
+++ b/pwndbg/commands/next.py
@@ -53,7 +53,7 @@ def next_syscall(*args):
     """
     Breaks at the next syscall.
     """
-    while not pwndbg.next.break_next_interrupt() and pwndbg.next.break_next_branch():
+    while pwndbg.proc.alive and not pwndbg.next.break_next_interrupt() and pwndbg.next.break_next_branch():
         continue
     pwndbg.commands.context.context()
 
--- a/pwndbg/commands/search.py
+++ b/pwndbg/commands/search.py
@@ -50,11 +50,11 @@ auto_save = pwndbg.config.Parameter('auto-save-search', False,
 parser = argparse.ArgumentParser(description='''
 Search memory for byte sequences, strings, pointers, and integer values
 ''')
-parser.add_argument('-t', '--type', choices=['byte','short','dword','qword','pointer','string','bytes'],
+parser.add_argument('-t', '--type', choices=['byte','short','word','dword','qword','pointer','string','bytes'],
                     help='Size of search target', default='bytes', type=str)
 parser.add_argument('-1', '--byte', dest='type', action='store_const', const='byte',
                     help='Search for a 1-byte integer')
-parser.add_argument('-2', '--word', dest='type', action='store_const', const='word',
+parser.add_argument('-2', '--word', '--short', dest='type', action='store_const', const='word',
                     help='Search for a 2-byte integer')
 parser.add_argument('-4', '--dword', dest='type', action='store_const', const='dword',
                     help='Search for a 4-byte integer')
@@ -107,6 +107,7 @@ def search(type, hex, string, executable, writable, value, mapping, save, next):
         }[pwndbg.arch.endian] + {
             'byte': 'B',
             'short': 'H',
+            'word': 'H',
             'dword': 'L',
             'qword': 'Q'
         }[type]
--- a/pwndbg/commands/search.py
+++ b/pwndbg/commands/search.py
@@ -6,6 +6,7 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import argparse
+import binascii
 import codecs
 import os
 import struct
@@ -95,7 +96,11 @@ def search(type, hex, string, executable, writable, value, mapping, save, next):
         save = bool(pwndbg.config.auto_save_search)
 
     if hex:
-        value = codecs.decode(value, 'hex')
+        try:
+            value = codecs.decode(value, 'hex')
+        except binascii.Error as e:
+            print('invalid input for type hex: {}'.format(e))
+            return
 
     # Convert to an integer if needed, and pack to bytes
     if type not in ('string', 'bytes'):
@@ -112,7 +117,12 @@ def search(type, hex, string, executable, writable, value, mapping, save, next):
             'qword': 'Q'
         }[type]
 
-        value = struct.pack(fmt, value)
+        try:
+            value = struct.pack(fmt, value)
+        except struct.error as e:
+            print('invalid input for type {}: {}'.format(type, e))
+            return
+
 
     # Null-terminate strings
     elif type == 'string':
--- a/pwndbg/commands/xor.py
+++ b/pwndbg/commands/xor.py
@@ -16,12 +16,12 @@ def xor_memory(address, key, count):
     Helper function for xorring memory in gdb
     """
     mem = pwndbg.memory.read(address, count, partial=True)
-    
+
     for index, byte in enumerate(mem):
         key_index = index % len(key)
         mem[index] = byte ^ ord(key[key_index])
 
-    return mem 
+    return mem
 
 @pwndbg.commands.Command
 @pwndbg.commands.OnlyWhenRunning
@@ -31,10 +31,17 @@ def xor(address, key, count):
     XOR ``count`` bytes at ``address`` with the key ``key``.
     '''
     if not isinstance(address, six.integer_types):
-        address = int(address, 16)
+        try:
+            address = int(address, 0)
+        except ValueError:
+            print('Invalid address %s' % address)
+            return
 
-    xorred_memory = xor_memory(address, key, count)
-    pwndbg.memory.write(address, xorred_memory)
+    try:
+        xorred_memory = xor_memory(address, key, count)
+        pwndbg.memory.write(address, xorred_memory)
+    except gdb.MemoryError as e:
+        print(e)
 
 @pwndbg.commands.Command
 @pwndbg.commands.OnlyWhenRunning
@@ -44,7 +51,14 @@ def memfrob(address, count):
     Run the memfrob command on a region of memory
     '''
     if not isinstance(address, six.integer_types):
-        address = int(address, 16)
+        try:
+            address = int(address, 0)
+        except ValueError:
+            print('Invalid address %s' % address)
+            return
 
-    xorred_memory = xor_memory(address, '*', count)
-    pwndbg.memory.write(address, xorred_memory)
+    try:
+        xorred_memory = xor_memory(address, '*', count)
+        pwndbg.memory.write(address, xorred_memory)
+    except gdb.MemoryError as e:
+        print(e)
--- a/pwndbg/commands/xor.py
+++ b/pwndbg/commands/xor.py
@@ -40,7 +40,7 @@ def xor(address, key, count):
     try:
         xorred_memory = xor_memory(address, key, count)
         pwndbg.memory.write(address, xorred_memory)
-    except gdb.MemoryError as e:
+    except gdb.error as e:
         print(e)
 
 @pwndbg.commands.Command
@@ -60,5 +60,5 @@ def memfrob(address, count):
     try:
         xorred_memory = xor_memory(address, '*', count)
         pwndbg.memory.write(address, xorred_memory)
-    except gdb.MemoryError as e:
+    except gdb.error as e:
         print(e)
--- a/pwndbg/commands/misc.py
+++ b/pwndbg/commands/misc.py
@@ -77,6 +77,7 @@ def distance(a, b):
     print("%#x->%#x is %#x bytes (%#x words)" % (a, b, distance, distance // _arch.ptrsize))
 
 @_pwndbg.commands.Command
+@_pwndbg.commands.OnlyWhenRunning
 def canary():
     """Print out the current stack canary"""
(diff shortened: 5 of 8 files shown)


pwndbg/commands/next.py
```diff
@@ -53,7 +53,7 @@ def next_syscall(*args):
     """
     Breaks at the next syscall.
     """
-    while not pwndbg.next.break_next_interrupt() and pwndbg.next.break_next_branch():
+    while pwndbg.proc.alive and not pwndbg.next.break_next_interrupt() and pwndbg.next.break_next_branch():
```
zachriggle: Is the concern that the process dies while in this loop?

anthraxx: no, for me i got a big ugly stacktrace yelling at me that i can't continue because the process has exited already. so it was "continuing until the next syscall" but in the meanwhile the program gracefully terminated, ultimately throwing an exception at me

pwndbg/commands/xor.py
```diff
@@ -31,10 +31,17 @@ def xor(address, key, count):
     XOR ``count`` bytes at ``address`` with the key ``key``.
     '''
     if not isinstance(address, six.integer_types):
-        address = int(address, 16)
+        try:
+            address = int(address, 16)
```
disconnect3d:
```
pwndbg> xor 0x7fffffffd820 0x10 0x2
  File "/home/dc/installed/pwndbg/pwndbg/commands/__init__.py", line 57, in __call__
  File "/home/dc/installed/pwndbg/pwndbg/commands/__init__.py", line 115, in _OnlyWhenRunning
    return function(*a, **kw)
  File "/home/dc/installed/pwndbg/pwndbg/commands/xor.py", line 41, in xor
    xorred_memory = xor_memory(address, key, count)
  File "/home/dc/installed/pwndbg/pwndbg/commands/xor.py", line 18, in xor_memory
    mem = pwndbg.memory.read(address, count, partial=True)
  File "/home/dc/installed/pwndbg/pwndbg/memory.py", line 41, in read
    count = max(int(count), 0)
  File "/home/dc/installed/pwndbg/pwndbg/inthook.py", line 45, in __new__
ValueError: invalid literal for int() with base 10: '0x2'
```

Maybe add:
```
count = max(int(count, 0), 0)
```
From `int` docstring:
> Base 0 means to interpret the base from the string as an integer literal.

pwndbg/commands/xor.py
```diff
@@ -44,7 +51,14 @@ def memfrob(address, count):
     Run the memfrob command on a region of memory
     '''
     if not isinstance(address, six.integer_types):
-        address = int(address, 16)
+        try:
+            address = int(address, 16)
```
disconnect3d: Same logic as in `xor` command. Isn't there any way to just forward it?


I have tried some stuff like:
```python
memfrob = functools.partial(xor, key=42)
```
even wrapping it with:
```python
pwndbg.commands.Command(pwndbg.commands.OnlyWhenRunning(...))
```

didn't really help...

And this:
```python
@pwndbg.commands.Command
def memfrob(address, count):
    '''memfrob(address, count)

    Run the memfrob command on a region of memory
    '''
    return xor(address, 42, count)
```

As this happens:
```
pwndbg> memfrob 0x7fffffffd838 10
object of type 'int' has no len()
'xor': xor(address, key, count)

    XOR ``count`` bytes at ``address`` with the key ``key``.
```

Btw. it might be good to have a `debug` flag and if it is turned on, such things would explode instead of just printing out an exception.

anthraxx: thats what logging and log levels should be for

pwndbg/commands/xor.py
zachriggle: We should probably not use `16` here, but rather `0` so that the value is auto-detected

anthraxx: makes sense, will do

zachriggle: Actually, how are we even getting strings here?  O_O

pwndbg/commands/xor.py
```diff
@@ -31,10 +31,17 @@ def xor(address, key, count):
...
 
-    xorred_memory = xor_memory(address, key, count)
-    pwndbg.memory.write(address, xorred_memory)
+    try:
+        xorred_memory = xor_memory(address, key, count)
+        pwndbg.memory.write(address, xorred_memory)
+    except gdb.MemoryError as e:
+        print(e)
 
 @pwndbg.commands.Command
```
zachriggle: The only reason we have to deal with `int`'ing a thing is because this is a `Command` instead of a `ParsedCommand`.  If you change it, we should be getting `gdb.Value` objects, which are much nicer.

pwndbg/disasm/__init__.py
```diff
@@ -100,11 +100,13 @@ def get_one_instruction(address):
         pwndbg.disasm.arch.DisassemblyAssistant.enhance(ins)
         return ins
 
-def one(address=None):
+
+
+def one(address='unset_argument'):
+    if address == 'unset_argument':
```
zachriggle: Wot?

anthraxx: well you want to distinguish between non set arugment and non argument. the whole codebase is build like it passes None and that can't be set to PC. however this whole function checks for none dont make any sense at all if ir returns None when being None :P 
so a) we fix _all_ million of calls to this func to never pass None
b) we dont allow to call it without argument
c) we have a special value to distiguis between none arguments and calls without argument

zachriggle: `None` implies "not set", any other value is explicit

anthraxx: i think you still miss the point, i fully understand that... but thats not how you wrote the whole other calls to one that pass an argument that is none. if you then set the address to $pc everything blows up ;-)

zachriggle: Do you mean the current value of PC, i.e. `pwndbg.regs.pc`, or the literal string `"$pc"`?  The former works fine, the latter should never work.

```
>>> pwndbg.disasm.one(pwndbg.regs.pc)
<capstone.CsInsn object at 0x7f62f89f64e0>
```

pwndbg/commands/xor.py
anthraxx: if i do that, no string literal can be passed anymore:

```
pwndbg> xor 0xffe78e9c ABC 100 
No symbol "ABC" in current context.
xor() takes exactly 3 arguments (2 given)
'xor': xor(address, key, count)

    XOR ``count`` bytes at ``address`` with the key ``key``.
```

pwndbg/commands/telescope.py
```diff
@@ -39,7 +39,7 @@ def telescope(address=None, count=telescope_lines, to_string=False):
     ($sp by default)
     """
     address = int(address if address else pwndbg.regs.sp) & pwndbg.arch.ptrmask
-    count   = int(count) & pwndbg.arch.ptrmask
+    count   = max(int(count), 0) & pwndbg.arch.ptrmask
```
zachriggle: This isn't actually what we want, and breaks the behavior of `telescope A B` to dump everything from `[A..B]`

pwndbg/commands/xor.py
zachriggle: I'm not sure what "ABC" would have been in any other case.  How *should* that be handled?

pwndbg/disasm/__init__.py
anthraxx: if cause pwndbg.regs.pc, that waht the second condition is supposed to do but it got taked to absurdum by returning before the assignment in case address is None.
If you fix it and set address to pwndbg.regs.pc once the argument is None, the whole disassembly will go nuts.

zachriggle: I can't parse the above comment.  Could you try re-writing / elaborating?  If you include a specific test case, I might understand better what the issue is.

pwndbg/commands/xor.py
anthraxx: its just a string. lets say i want to xor it with "FOOBAR" its quite handy to xor with a ascii xor key instead of only bein able to pass bytes via 0x42424242

pwndbg/disasm/__init__.py
anthraxx: can you come to the chat? :D

anthraxx:
```
def one(address=None):
    if address is None or not pwndbg.memory.peek(address):
        return None
    if address is None:
        address = pwndbg.regs.pc
```

that doesn't make any sense as "address = pwndbg.regs.pc" will never be assigned. therefor if you invoke one() without an argument it always returns None. however try removing the early exit yourself and step throug a program, it will screw up all the disassembly as other calls to one(var) having passed var=None because of some reasons this will scrwe up everything.

pwndbg/commands/xor.py
zachriggle: Then you have to pass in `"FOOBAR"`, not `FOOBAR`.

pwndbg/disasm/__init__.py
zachriggle: Where is `None` being explicitly passed in?  That sounds like a bug elsewhere that we're covering up with this statement.

anthraxx: no clue there is tons of code involved that does random stuff, its hard to say when its None while its not supposed to be like that. i can just tell calling one() without argument is broken :)
can we fix this in a own PR and i remove this from here? i don't like fixing the universe in one PR it looses the overview

pwndbg/regs.py
```diff
@@ -280,7 +280,7 @@ def __getattr__(self, attr):
 
             value = int(value)
             return value & pwndbg.arch.ptrmask
-        except (ValueError, gdb.error):
+        except (ValueError, AttributeError, gdb.error) as e:
```
zachriggle: Why are we catching `AttributeError` here? When is this actually thrown? This sounds like another issue we're hiding

pwndbg/commands/telescope.py
anthraxx: it never worked like that, try it in the current version. B behaves like a count right now and that is preserved

pwndbg/commands/xor.py
anthraxx: well i'm not that stupid
```
pwndbg> xor 0xffe78e9c "ABC" 100 
No symbol "ABC" in current context.
xor() takes exactly 3 arguments (2 given)
'xor': xor(address, key, count)

    XOR ``count`` bytes at ``address`` with the key ``key``.
```

pwndbg/regs.py
anthraxx: whenever some code accesses the regs.pc while its not yet available, i remove it for not and we handle it later. too much stuff

pwndbg/disasm/__init__.py
zachriggle: Let's revert the changes to `disasm.py`, then

pwndbg/commands/hexdump.py
```diff
@@ -32,13 +33,17 @@
 def hexdump(address=None, count=pwndbg.config.hexdump_bytes):
     address = int(address)
     address &= pwndbg.arch.ptrmask
-    count   = int(count)
+    count   = max(int(count), 0)
```
zachriggle: Does this break functionality of `hexdump A B` to hexdump `[A..B]` given that addresses may be interpreted as negative?

```
pwndbg> hexdump $rsp $rsp+10
+0000 0x7fffffffdc10  01 00 00 00  00 00 00 00  27 df ff ff  ff 7f 00 00  │....│....│'...│....│
+0010 0x7fffffffdc20
```

pwndbg/commands/search.py
```diff
@@ -50,7 +51,7 @@ def print_search_hit(address):
 parser = argparse.ArgumentParser(description='''
 Search memory for byte sequences, strings, pointers, and integer values
 ''')
-parser.add_argument('-t', '--type', choices=['byte','short','dword','qword','pointer','string','bytes'],
+parser.add_argument('-t', '--type', choices=['byte','word','dword','qword','pointer','string','bytes'],
```
zachriggle: Please put `short` back? Thanks!  `word` is ambiguous, which is why `short` was explicitly chosen.

pwndbg/commands/xor.py
zachriggle: Looks like we'd need QuietSloppyParsedCommand :P (or just to use `argparse`)

pwndbg/commands/telescope.py
anthraxx: i got the idea what its supposed to do, removing the block and i will rework this part in a new PR

pwndbg/regs.py
zachriggle: It does sound like we need some `@OnlyWhileRunning`, then.

pwndbg/commands/search.py
anthraxx: so reword the --word to --short? or can we support both values?

pwndbg/commands/hexdump.py
anthraxx: it still works like that as the address won't be negative even when the "binary representation" of a signed integer type would be a negative number, it just ansures. it will just subtract the value later.
the only behavior that triggers this was passing -10 as a second argument, which breaks this command.
so all fine :smile:

anthraxx:
I removed all "critical" parts (will do in separated PRs). I guess the except gdb.MemoryError is good enough for now, it hardens the call and I will rework the xor and memfrob commands later.
Also the hexdump functionality is still preserved so that should all be fine

pwndbg/commands/hexdump.py
```diff
@@ -32,13 +33,17 @@
...
     width   = int(pwndbg.config.hexdump_width)
 
     if count > address > 0x10000:
         count -= address
 
-    data = pwndbg.memory.read(address, count, partial=True)
+    try:
+        data = pwndbg.memory.read(address, count, partial=True)
+    except gdb.MemoryError as e:
```
zachriggle: We should catch `gdb.error` instead of `gdb.MemoryError`

pwndbg/commands/search.py
```diff
@@ -50,11 +51,11 @@ def print_search_hit(address):
 parser = argparse.ArgumentParser(description='''
 Search memory for byte sequences, strings, pointers, and integer values
 ''')
-parser.add_argument('-t', '--type', choices=['byte','short','dword','qword','pointer','string','bytes'],
+parser.add_argument('-t', '--type', choices=['byte','short','word','dword','qword','pointer','string','bytes'],
                     help='Size of search target', default='bytes', type=str)
 parser.add_argument('-1', '--byte', dest='type', action='store_const', const='byte',
                     help='Search for a 1-byte integer')
-parser.add_argument('-2', '--word', dest='type', action='store_const', const='word',
+parser.add_argument('-2', '--word', '--short', dest='type', action='store_const', const='word',
```
zachriggle: Does this actually work this way, with arbitrary arguments? Neat!

pwndbg/commands/xor.py
```diff
@@ -31,10 +31,17 @@ def xor(address, key, count):
...
+        except ValueError:
+            print('Invalid address %s' % address)
+            return
 
-    xorred_memory = xor_memory(address, key, count)
-    pwndbg.memory.write(address, xorred_memory)
+    try:
+        xorred_memory = xor_memory(address, key, count)
+        pwndbg.memory.write(address, xorred_memory)
+    except gdb.MemoryError as e:
```
zachriggle: gdb.error

zachriggle:
Looks good! Two small changes from `gdb.MemoryError` ==> `gdb.error`.

pwndbg/commands/search.py
anthraxx: neither did I... but a second add_argument felt dirty and so i tried it our and tested it... turns out it can do such things :smirk_cat:

pwndbg/commands/hexdump.py
anthraxx: ok changing

pwndbg/commands/xor.py
anthraxx: ok changing

anthraxx:
replaced the catch blocks with gdb.error (just a side note, the xor command parsing comes soon as own PR)

PR #186, opened 2017-03-20
Add README info about AUR packages on Arch Linux

disconnect3d:


zachriggle:
Rather than textual instructions, is there any way we can just invoke the correct `pacman` command in `setup.sh`?  It looks like the dependencies are:

```
python-capstone
python-future
python-psutil
python-ptrace
python-pycparser
python-pyelftools
python-six
python-unicorn (python-unicorn-git)
checksec (optional) – checksec command support
ropgadget (ropgadget-git) (optional) – ropgadget command support
ropper (ropper-git) (optional) – ropper command support
```

disconnect3d:
The AUR (Arch User Repository) packages can be installed either by some aur wrapper (e.g. pacaur, yaourt) or manually by:
```
git clone https://aur.archlinux.org/pwndbg-git.git
cd pwndbg-git && makepkg -i
cd .. && rm -rf pwndbg-git
```

This will install all the dependencies along with the package itself (in /usr/share/pwndbg/).

I am not sure whether we need those in setup.sh as this can be launched without cloning our repo at all.

zachriggle:
@anthraxx What's your take on the best way to do this? You own the AUR package

anthraxx:
Well i think that actually this are two different areas/fields. In my opinion I see the setup.sh more as either a configurator/setup tool if you check out the git repo and want to setup everything that is needed to work on this project.... or of cause to only use it as a user if there is no ready-to-install package provided by the distribution of you choice.
So speaking about Arch Linux itself, if you are "just" an end-user who wants to use it, I recommend to use the ready package to do so, this also doesn't require you to build capstone/unicorn.
Having such dependencies per-distribution explicitly in the seutp.sh may be handy for some cases, but that will also require to maintain them for all distributions handled there... i sense some people coming around and adding blocks for fedora and stuff... those then need to always be kept in sync if new dependencies are added or one distribution decides to rename it.

In my general opinion its more robust and less work for devs if a package maintainer itself is responsible to reflect the dependencies of something to his/her distribution and provide proper packages with dependencies. On the other hand i totally see why you may want to have them in f.e. for debian where no easy way to install it via package is available.

Your call, but these are my thoughts and opinions you asked for.

PS: I will soonish move the pwndbg (non-git) release version from the AUR into the official repositories making it installable for all Arch users without the need of an AUR wrapper :yum:

PR #187, opened 2017-03-21
telescope: fix printing to many values if count is an address

anthraxx:
This fixes an issue that too many values are printed when using
the second argument B as an address as it was simply subtracted
but in the context of telescope the count means ptrsize values
as its later (correctly) multiplied with ptrsize.
We also use ceil instead of floor division as its a better user
experience to not silently discard a value if the dest is unaligned
like with $sp $sp+1

On top do not allow count to be negative when B is used as actual
count integer rather then an address. When B is an address the
integer value is converted to an unsigned integer, therefor address
that, in binary, would represent an negative signed integer wont
be clamped by the max(count, 0).

### DIFF
--- a/pwndbg/commands/telescope.py
+++ b/pwndbg/commands/telescope.py
@@ -12,6 +12,7 @@ from __future__ import unicode_literals
 
 import argparse
 import collections
+import math
 
 import pwndbg.arch
 import pwndbg.chain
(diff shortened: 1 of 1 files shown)


pwndbg/commands/telescope.py
```diff
@@ -39,24 +40,27 @@ def telescope(address=None, count=telescope_lines, to_string=False):
     ($sp by default)
     """
     address = int(address if address else pwndbg.regs.sp) & pwndbg.arch.ptrmask
-    count   = int(count) & pwndbg.arch.ptrmask
+    count   = max(int(count), 0) & pwndbg.arch.ptrmask
```
zachriggle: If `count` is zero, do we do anything at all?  Might be better to bail with an error than use `max(..., 0)` since it's a no-op if `count == 0`.

anthraxx: sure, sounds fair. will make it print something and return... but will need that check also for address a bit later as one could pass `telescope $sp $sp` :smile:

zachriggle: Perhaps just do `max(..., 1)`?  Then we'll always print something

anthraxx: hmmm nifty idea, so simple and so effective... :+1: i love this :smile_cat:

anthraxx:
updated, this ensures that `telescope $sp $sp` and `telescope $sp 0` at least print one value

PR #188, opened 2017-03-21
adding simple radare2 command with auto-seek to $pc when running

anthraxx:
It's quite handy to drop into radare2 in the middle of a deep debugging
session to fire up the visual mode and examine the current location in
gdb using the ascii graph view (or something else) in radare2.


pwndbg/commands/radare2.py
```diff
@@ -0,0 +1,37 @@
...
+
+parser = argparse.ArgumentParser(description=".",
+                                 epilog="Example: radare2 -- -S -AA")
+parser.add_argument('--no-seek', action='store_true',
+                    help='Do not seek to current pc')
+parser.add_argument('arguments', nargs='*', type=str,
+                    help='Arguments to pass to radare2')
+
+@pwndbg.commands.ArgparsedCommand(parser)
+def radare2(arguments, no_seek=False):
```
zachriggle: Can we name the command `r2` instead? Easier to type :P

anthraxx: of cause, why not... i though about the same... but decided to use radare2 as r2 is a symlink to it.... but i like r2 also more.

anthraxx:
renamed command (and epilog reference) to r2

PR #189, opened 2017-03-22
fix disasm.one() to be called without an address

anthraxx:
This brings back the functionality to call disasm.one() without
and target address. As a default value the current regs.pc is
selected.
Fix the disasm.near() call to not pass None to disasm.one() when
the backward cache misses, otherwise it wrongly falls back to
the regs.pc value (which near() is not supposed to do in its
context).


zachriggle:
Thanks for hunting this down! <3

PR #190, opened 2017-03-22
early exit when require a valid file to be selected

anthraxx:


zachriggle:
I like this, but I'd prefer to have a bit more info and stick closer to GDB's default error messages.  Can we replace our message with whatever is printed when eg "info file" is done without a file selected?

anthraxx:
sure, can see i just tried to stick to what was used in multiple places already

anthraxx:
Hm it just silently does nothing (which actually i don't like too much :yum:)

zachriggle:
Ha, that's funny. Can we at least add a "See 'help file' for more
information." at the end?

anthraxx:
sure that sounds appropriate. as i don't want to duplicate that all over the place, i will create a error string that we can reference globally

zachriggle:
Good plan! Now add i18n! (Just kidding)

zachriggle:
Merging this for now, we can add the string error thing later

PR #192, opened 2017-03-27
nextsymbol command. halts at next call matching a given regex. 

ysf:


ysf:
will recommit with working tests

zachriggle:
Thanks for this!

I think I'd rather just change the `nextcall` command itself to take an optional argument, rather than introducing a new command.

pwndbg/next.py
```diff
@@ -88,6 +89,16 @@ def break_next_call(address=None):
...
             return ins
 
+def break_call_regex(regex=''):
+    while pwndbg.proc.alive:
+        ins = break_next_call()
+
+        if not ins:
+            break
+
+        if re.search(regex, ins.symbol):
```
zachriggle: I think `re.match` would be better, so that if I want to stop on `printf`, it doesn't accidentally stop on e.g. `sprintf`

zachriggle:
Please resume your work on this Pull Request rather than opening a new one.

pwndbg/next.py
ysf: I prefered search because I'd favorise writing ^printf instead of .*@plt or alike. If you needto match from the beginning a ^ is quickly typed. Still prefer match?

ysf:
Sorry for the inconvienience, am new to this. Can I somehow "attach" a new pull request here?

pwndbg/next.py
zachriggle: Hmm, are there actually any symbols which contain `@plt` in them?

setup.sh
```diff
@@ -3,7 +3,7 @@ set -ex
 
 if uname | grep -i Linux &>/dev/null; then
     sudo apt-get update || true
-    sudo apt-get -y install python-dev python3-dev python-pip python3-pip libglib2.0-dev libc6-dbg
+    sudo apt-get -y install git gdb python-dev python3-dev python-pip python3-pip libglib2.0-dev libc6-dbg
```
zachriggle: This is an unnecessary change for this Pull Request.

zachriggle: This is an unnecessary change for this Pull Request.

pwndbg/commands/next.py
```diff
@@ -36,6 +36,13 @@ def nextcall(*args):
 
 @pwndbg.commands.Command
 @pwndbg.commands.OnlyWhenRunning
+def nextsymbol(*args):
```
zachriggle: Instead of defining a new command, modify the existing `nextcall` command.

Thanks!

ysf: Ok, gladly do that. How should I differentiate if an address or regex parameter was given?

zachriggle: The easiest way is to just not use regex at all, and let GDB do symbol lookups for you, by making this command a `ParsedCommand`.

pwndbg/next.py
zachriggle: In either case, it seems like this case can be easily handled at runtime.  Assuming that the user tells us that they want to stop at the next call to `printf`, ideally it'd work like:

```
pwndbg> nextcall printf
```

Note the lack of regex.  We can use regex internally to match e.g. `(_*)(printf)(_*)(@plt)`.

zachriggle:

PR #194, opened 2017-03-29
pwndbg.ida Comment() broken

cloudburst:
@takes_address already performs l2() so doing it twice breaks this command

It might be worth adding logic for GetCommentEx(ea, 1) in the future, which checks for [a repeating comment](http://reverseengineering.stackexchange.com/questions/11024/getting-user-comments-with-idapython-api-user-cmts)


zachriggle:
Thanks for the contribution!

PR #195, opened 2017-03-29
Add IDA comments to context display

cloudburst:
This queries IDA for any comments or repeatable comments and adds them to the end of the assembly.

I also removed the GetCommentEx(addr) call.  I'm not sure if this is legacy idapython or what but it isn't needed on my setup.  And added the [repeatable comment](http://reverseengineering.stackexchange.com/questions/11024/getting-user-comments-with-idapython-api-user-cmts) function.

This adds a lot of RPC calls to each context() call on top of all of the IDA queries already for Anterior().  Next I want to hide all of these behind a config (similar to the existing unicorn emulate flag) to speed up context display when connected to IDA.

### DIFF
--- a/pwndbg/ida.py
+++ b/pwndbg/ida.py
@@ -130,7 +130,13 @@ def base():
 @withIDA
 @takes_address
 def Comment(addr):
-    return _ida.GetCommentEx(addr, 0) or _ida.GetCommentEx(addr)
+    return _ida.GetCommentEx(addr, 0)
+
+
+@withIDA
+@takes_address
+def RepeatableComment(addr):
+    return _ida.GetCommentEx(addr, 1)
 
 
 @withIDA
--- a/pwndbg/ida.py
+++ b/pwndbg/ida.py
@@ -326,7 +326,8 @@ def has_cached_cfunc(addr):
 @takes_address
 @pwndbg.memoize.reset_on_stop
 def decompile(addr):
-    return _ida.decompile(addr)
+    return None
+    # return _ida.decompile(addr)
 
 
 @withIDA
--- a/pwndbg/ida.py
+++ b/pwndbg/ida.py
@@ -97,19 +97,13 @@ def available():
 
 
 def l2r(addr):
-    exe = pwndbg.elf.exe()
-    if not exe:
-        raise Exception("Can't find EXE base")
-    result = (addr - int(exe.address) + base()) & pwndbg.arch.ptrmask
-    return result
+    vaddr = pwndbg.vmmap.find(pwndbg.regs.pc).vaddr
+    return (addr - vaddr + base()) & pwndbg.arch.ptrmask
 
 
 def r2l(addr):
-    exe = pwndbg.elf.exe()
-    if not exe:
-        raise Exception("Can't find EXE base")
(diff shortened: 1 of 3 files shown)


zachriggle:
I believe that the other `GetCommentEx` call that you removed is necessary to extract Hexrays decompiler output when copied to the disassembly via the `/` key.

cloudburst:
Ah ok, I don't have Hexrays unfortunately.  In fact calling `ida.decompile()` segfaults my ida and I had to comment it out of the pwndbg ida RPC server.  Took a while to track that down.

cloudburst:
Based on that, I'm not really sure if this patch works or not if Hexrays is installed, but I have no way to test that to see what the output would look like.

zachriggle:
Interesting! Could you open a separate bug for the `ida.decompile()` segfault?

cloudburst:
Sorry, what I meant specifically was `idaapi.decompile()` segfaults IDA.  So It's not a problem with pwndbg although pwndbg was calling ida.decompile as part of context/nearpc so that's what I had to comment out.

I'm one revision back on IDA Pro so it may be patched idk.

zachriggle:
Right, I understood what you meant.  I'd prefer to avoid segfaulting Pwndbg users' IDA instances if possible :P

pwndbg/ida.py
```diff
@@ -97,19 +97,13 @@ def available():
...
-    exe = pwndbg.elf.exe()
-    if not exe:
-        raise Exception("Can't find EXE base")
-    result = (addr - int(exe.address) + base()) & pwndbg.arch.ptrmask
-    return result
+    vaddr = pwndbg.vmmap.find(pwndbg.regs.pc).vaddr
+    return (addr - vaddr + base()) & pwndbg.arch.ptrmask
 
 
 def r2l(addr):
```
disconnect3d: Since we are touching it, can we add docstring or rename  `r2l` and `l2r` so that the purpose of those functions would be more clear?
It rebases the addreses, isn't it? But which one rebases to gdb and which one to ida?

zachriggle:
The `L` and `R` stand for `local` and `remote` respectively, where `gdb` is "local" and IDA is "remote".  They are indeed to convert addresses between the real addresses that GDB sees, and whatever addresses IDA is using.  We could rebase the IDB, but it's expensive and leads to really weird errors, so we just adjust addresses on-the-fly.

pwndbg/ida.py
zachriggle: The `L` and `R` stand for `local` and `remote` respectively, where `gdb` is "local" and IDA is "remote".  They are indeed to convert addresses between the real addresses that GDB sees, and whatever addresses IDA is using.  We could rebase the IDB, but it's expensive and leads to really weird errors, so we just adjust addresses on-the-fly.

zachriggle:
@cloudburst I think that you're accidentally adding additional commits, since you made this pull request against `master`.

cloudburst:
Yeah I am, sorry idk why the pull was from master ;/

Should I open a new pull, I don't see a way to remove those commits from this UI.

cloudburst:
Actually it's probably better filed as an Issue since it needs testing on how a comment is handled versus decompiler output so I'll do that.

Issue #180, opened 2017-03-15
Infinite loops - better code display


disconnect3d:
For such code:
```c
int main() { for(;;); }
```

We display it's assembly code as:
```
[─────────────────────────────────────────────────CODE──────────────────────────────────────────────────]
 ► 0x4004aa <main+4>    jmp    main+4                        <0x4004aa>
    ↓
 ► 0x4004aa <main+4>    jmp    main+4                        <0x4004aa>
    ↓
 ► 0x4004aa <main+4>    jmp    main+4                        <0x4004aa>
    ↓
 ► 0x4004aa <main+4>    jmp    main+4                        <0x4004aa>
    ↓
 ► 0x4004aa <main+4>    jmp    main+4                        <0x4004aa>
    ↓
 ► 0x4004aa <main+4>    jmp    main+4                        <0x4004aa>
    ↓
 ► 0x4004aa <main+4>    jmp    main+4                        <0x4004aa>
    ↓
 ► 0x4004aa <main+4>    jmp    main+4                        <0x4004aa>
    ↓
 ► 0x4004aa <main+4>    jmp    main+4                        <0x4004aa>
    ↓
 ► 0x4004aa <main+4>    jmp    main+4                        <0x4004aa>
    ↓
 ► 0x4004aa <main+4>    jmp    main+4                        <0x4004aa>
```

Which could be reduced to just one instruction...

I have added a print in line 81 in [pwndbg/commands/nearpc.py](https://github.com/pwndbg/pwndbg/blob/master/pwndbg/commands/nearpc.py#L80-L82) :
```python
instructions = pwndbg.disasm.near(pc, lines, emulate=emulate)
print(pc, instructions)
if pwndbg.memory.peek(pc) and not instructions:
```

Which shows that we actually know that those instructions are all the same:
```
[...]

PR #200, opened 2017-04-03
Better code display for 1 instruction inf loop #180

disconnect3d:
Removes all duplicate instructions from the tail end of the disassebled instructions list.

### DIFF
--- a/pwndbg/disasm/__init__.py
+++ b/pwndbg/disasm/__init__.py
@@ -226,4 +226,21 @@ def near(address, instructions=1, emulate=False):
         if insn:
             insns.append(insn)
 
+    # Removing all duplicate instructions from the end
+    # makes display more clear when we are in an 1 instruction infinite loop
+    if insns:
+        last_insn = insns[-1]
+        last_idx = len(insns) - 1
+
+        for i, insn in enumerate(reversed(insns)):
+            if i == 0:
+                continue
+
+            if insn == last_insn:
+                idx = last_idx - i
+                insns.pop(idx)
+            else:
+                break
+
     return insns
+


pwndbg/commands/nearpc.py
```diff
@@ -80,7 +80,23 @@ def nearpc(pc=None, lines=None, to_string=False, emulate=False):
     #             pc_to_linenos[line.pc].append(line.line)
     instructions = pwndbg.disasm.near(pc, lines, emulate=emulate)
```
zachriggle: I would much rather this logic be implemented in `pwndbg.disasm.near`, rather than in the command itself.

zachriggle:
I was playing around with this a bit, and have one suggestion and one problem.

Suggestion is that I think this is a bit more concise:

```py
    # Remove infinite loops repeating themselves
    while insns and len(insns) > 1 and insns[-1] == insns[-2]:
        del insns[-1]
```

The problem with doing the unrolling this way is that we're thinking of infinite loops, but the forward-emulation for ROP gadgets will break for a RET sled under this model.  It'll just look like a single RET, and then nothing more.

I think a fair way to do this would be to let the instruction repeat *once* (i.e. the last two instructions are the same), so that we can at least see the "next" instruction / gadget.

disconnect3d:
Do you mean something like this by a RET sled?:
```asm
[bits 32]

global _start
_start:

call test1
ret
test1:
  call test2
  ret
test2:
  call test3
  ret
test3:
  call test4
  ret
test4:
  call test5
  ret
test5:
    ret
```

Launching:
```
nasm -f elf test.asm && ld -m elf_i386 test.o -o test && gdb -q ./test
```

This produces the same output with and without your changes.
Also the output is... pretty weird for me, see https://asciinema.org/a/01iwnxnmxxd3jn2kjhfdb5jiw

zachriggle:
No, in your case all of the `ret` instructions are at different addresses.

The reason that the output is weird is because of the "previous instruction" tracking.  At `test4`, we call `test5`.  `test5` is just a `ret`, and the Unicorn emulation lets us know that the next instruction is the `ret` inside `test4`, and so on.  The **forward** instructions are all correct.  The **backward** instructions get a bit confused since we're going `test5 ret` --> `test4 ret`, so we save `test5` as the instruction which "preceded" `test4 ret`.  So when we unfold backward, things get weird.

The "ret sled" I'm referring to would be something like this (haven't tested it, but you should get the idea)

```
## "Terminate" our ROP stack with an CC
    call A
    int3
A:

## Push the "ret" address onto the stack a whole bunch
    mov ecx, 0x100
THELOOP:
    loop push_retaddr
    ret

## Push the "ret" address via "call"
push_ret_sled_address:
    call THELOOP

## Actual "ret sled" consists of this address on the stack a whole bunch,
## which Unicorn will emulate through so that we know the "next" instruciton
## is still this one
    ret
```

zachriggle:
screen shot 2017-04-05 at 2 48 43 pm

disconnect3d:
Closed - we've added an alternate implementation: https://github.com/pwndbg/pwndbg/pull/205.

Issue #156, opened 2017-01-31
Exception with ELF MIPSII


g05u:
Hey, when I set a breakpoint in an ELF MIPSII using remote debugging, pwndbg throws an exception (binary is attached):

[bender.gz](https://github.com/pwndbg/pwndbg/files/742558/bender.gz)

pwndbg> file bender_safe 
Reading symbols from bender_safe...(no debugging symbols found)...done.
pwndbg> b validate 
Breakpoint 1 at 0x401c84
pwndbg> c
Continuing.
Traceback (most recent call last):
  File "/home/user/pwndbg/pwndbg/events.py", line 111, in caller
    func()
  File "/home/user/pwndbg/pwndbg/vmmap.py", line 107, in explore_registers
    find(pwndbg.regs[regname])
...
gdb.MemoryError: Cannot access memory at address 0x34400000
Python Exception <class 'gdb.MemoryError'> Cannot access memory at address 0x34400000: 
Traceback (most recent call last):

disconnect3d:
This is a remote debugging session, right?

You probably have to set proper architecture, e.g.:
```
set arch mips
```
you must have the gdb installed with this architecture, or you can try gdb-multiarch.

However I am not sure the pwndbg works with MIPS.

g05u:
When I have a remotely debuged process and I try to read the local symbols (file path/to/bin) pwndbg doesn't manage it successfully and an exception is thrown.

zachriggle:
The issue is that the internal ELF structures declared via `ctypes` parse the data from the ELF files in-memory in Little Endian format, despite the actual contents being Big Endian.

There's some hackery we can probably do with metaclasses or __getattribute__ hooks.

zachriggle:
I've added some code to the `dev` branch which prevents all of these error messages.

However, it looks like Capstone's disassembler is broken for this architecture.

zachriggle:
Filed aquynh/capstone#849 and added a work-around to `dev`

zachriggle:
Looks like most things work.  The branch detection via Unicorn is busted, but that's it.
<img width="1039" alt="screen shot 2017-01-31 at 9 58 40 pm" src="https://cloud.githubusercontent.com/assets/111640/22493831/7983bfae-e800-11e6-917e-665bde7b3523.png">

zachriggle:
Merged to `master`

g05u:
Thanks, this issue is fixed

disconnect3d:
[...]

PR #202, opened 2017-04-04
Fix Ehdr and Phdr endian switching

disconnect3d:
Also fixes regression - issue #156.
Before this changes we were creating ElfXX_Ehdr and ElfXX_Phdr structures once during initialization using `pwndbg.ctypes.Structure` as base class (which was `ctypes.LittleEndianStructure` - as the default `pwndbg.arch.endian` is 'little').
We were thinking that the proper Ehdr/Phdr is set based on architecture (`pwndbg.arch.ptrsize`) and endian (as we updated the `pwndbg.ctypes.Structure` attribute to either `ctypes.LittleEndianStructure` or `ctypes.BigEndianStructure`).
However, changing the `pwndbg.ctypes.Structure` didn't really refresh the ELfXX_Ehdr and ElfXX_Phdr structures, so they still inherited from the old `pwndbg.ctypes.Structure` (which was `ctypes.LittleEndianStructure`).

Because of above, when we loaded a binary which was big endian, we failed on parsing its Ehdr and/or Phdr structures (as we just thought we updated Ehdr/Phdr to proper endian instead of really doing it; and so we interpreted Ehdr/Phdr fields as little endian).

This commit fixes the issue by removing the `pwndbg.ctypes` totally and by creating Ehdr and Phdr classes for each architecture (`pwndbg.arch.ptrsize` - 4 or 8) and endian (`pwndbg.arch.endian` - 'big' or 'little').
To make the code not so much bloated when it is possible the classes are generated.


zachriggle:
This is still sitting unmerged since I'm not a fan of the way this was done.  The code worked a while ago with the design that I was happy with, then I broke it for some reason.  I need to revisit that code and see why / what was broken.

The `pwndbg.ctypes` module should provide transparent endianness switching on-the-fly, and it's also important for some future work that I'm doing with being able to define structures manually in Python via `ctypes`, and then emit them as valid C structures, which are then loaded via `pwndbg.typeinfo`.  The design requires using a sub-class of `ctypes.Structure` so that we can add various methods / properties to it, and handle endianness-switching (i.e. struct reloading) on-the-fly (basically, `sys.modules[self.__module__].__dict__[self.__class__.__name__] = foobar`) or with some kind of crazy `__getattr__` and `__setattribute__` magic).

disconnect3d:
I am not really a big fan of reloading modules - it adds unnecessary complexity as noone expects that and overhead (that's not really the case I guess).

---

Also, there are two resources that might be helpful in here (or just cool to look through), that comes from a talk from a friend of mine:
* slides - http://slides.com/khorne/binarki#/
* similar code - https://gist.github.com/mrzechonek/b843be2d1b48236e40a07193272309c0

Unfortunately [the talk](https://www.youtube.com/watch?v=0TfZtitN_l4) is in polish.

zachriggle:

PR #206, opened 2017-04-04
add check for GDB being available in PATH

mauvehed:
If you run setup.sh and GDB is not available in your path the script simply fails later when it attempts to call GDB.


zachriggle:
Closing in favor of #207

PR #207, opened 2017-04-04
Enhance setup.sh

zachriggle:
A few enhancements to setup.sh

- Install `gdb` via `apt-get`
- Ensure that `gdb` is in `$PATH` before attempting to invoke it
- Ensure that the Python has access to `pip` via `ensurepip`
- Upgrade `pip` itself before installing dependencies
- Install dependencies into the correct `site-packages`
  - This code was previously removed, but is actually necessary when GDB does not use the system Python

Supersedes #206

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -28,7 +28,9 @@ PYTHON="${PYTHON}${PYVER}"
 SITE_PACKAGES=$(gdb -batch -q --nx -ex 'pi import site; print(site.getsitepackages()[0])')
 
 # Make sure that pip is available
-sudo ${PYTHON} -m ensurepip --upgrade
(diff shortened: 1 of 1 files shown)


disconnect3d:
Hmm.. probably we shouldn't fire `apt-get install python-pip` if pip already exists or if python from virtualenv is used.

ysf:
git should be installed via apt-get too if setup.sh uses it.

zachriggle:
Re: @disconnect3d about `python-pip`, the system `pip` may be old, we are upgrading it.  We use the system `pip` to install to GDB's Python site-packages.  There's no conflict here.

Re: @ysf Please see the diff

PR #208, opened 2017-04-05
Add ability to load individual symbols

cloudburst:
```c
$ cat kardashian.c
#include "kardashian.h"
struct sample foo;

$ cat kardashian.h
struct sample {
  int i;
  struct sample *less;
  struct sample *more;
};
```

```
pwndbg> ptype foo
No symbol "foo" in current context.
pwndbg> python pwndbg.typeinfo.compile("/home/kim/kardashian.c")
pwndbg> ptype foo
type = struct sample {
    int i;
    sample *less;
    sample *more;
}
```

### DIFF
--- a/pwndbg/typeinfo.py
+++ b/pwndbg/typeinfo.py
@@ -124,6 +124,8 @@ def load(name):
 
     with open(filename, 'w+') as f:
         f.write(source)
+        f.flush()
+        os.fsync(f.fileno())
 
     compile(filename)
 
--- a/pwndbg/typeinfo.py
+++ b/pwndbg/typeinfo.py
@@ -124,7 +124,6 @@ def load(name):
 
     with open(filename, 'w+') as f:
         f.write(source)
-        f.flush()
 
     compile(filename)
 
--- a/pwndbg/typeinfo.py
+++ b/pwndbg/typeinfo.py
@@ -136,12 +136,20 @@ def compile(filename=None, address=0):
         print("Specify a filename to compile.")
         return
 
-    objectname = filename.split(".")[0] + ".o"
+    objectname = os.path.splitext(filename)[0] + ".o"
 
     if not os.path.exists(objectname):
         gcc     = pwndbg.gcc.which()
         gcc    += ['-w', '-c', '-g', filename, '-o', objectname]
         subprocess.check_output(' '.join(gcc), shell=True)
 
+    add_symbol_file(objectname, address)
+
+def add_symbol_file(filename=None, address=0):
+    """Read additional symbol table information from the object file filename"""
+    if filename is None:
+        print("Specify a symbol file to add.")
+        return
+
     with pwndbg.events.Pause():
-        gdb.execute('add-symbol-file %s %s' % (objectname, address), from_tty=False, to_string=True)
+        gdb.execute('add-symbol-file %s %s' % (filename, address), from_tty=False, to_string=True)
--- a/pwndbg/typeinfo.py
+++ b/pwndbg/typeinfo.py
@@ -87,6 +87,7 @@ blacklist = ['regexp.h', 'xf86drm.h', 'libxl_json.h', 'xf86drmMode.h',
 'libunwind.h','libmjollnir-objects.h','libunwind-coredump.h','libunwind-dynamic.h']
 
 def load(name):
+    """Load symbol by name from headers in standard system include directory"""
     try:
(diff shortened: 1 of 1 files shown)


pwndbg/typeinfo.py
```diff
@@ -119,18 +120,28 @@ def load(name):
...
+
+    return gdb.lookup_type(name)
+
+def compile(filename=None, address=0):
+    """Compile and extract symbols from specified file"""
+    if filename is None:
+        print("Specify a filename to compile.")
+        return
+
+    objectname = filename.split(".")[0] + ".o"
```
zachriggle: Use `splitext` instead, thanks!

pwndbg/typeinfo.py
```diff
@@ -119,18 +120,28 @@ def load(name):
...
         gcc     = pwndbg.gcc.which()
-        gcc    += ['-w','-c','-g',filename + '.cc','-o',filename + '.o']
+        gcc    += ['-w', '-c', '-g', filename, '-o', objectname]
         subprocess.check_output(' '.join(gcc), shell=True)
 
     with pwndbg.events.Pause():
-        gdb.execute('add-symbol-file %s.o 0' % filename, from_tty=False, to_string=True)
-
-    return gdb.lookup_type(name)
+        gdb.execute('add-symbol-file %s %s' % (objectname, address), from_tty=False, to_string=True)
```
zachriggle: While you're adding functionality for this, would you mind breaking this bit out into a new function?

cloudburst:
Good idea, updated.

pwndbg/typeinfo.py
```diff
@@ -119,18 +120,36 @@ def load(name):
...
-    filename = '%s/%s_%s' % (tempdir, arch, '-'.join(name.split()))
+    filename = '%s/%s_%s.cc' % (tempdir, arch, '-'.join(name.split()))
 
-    if not os.path.exists(filename + '.o'):
-        with open(filename + '.cc', 'w+') as f:
-            f.write(source)
-            f.flush()
+    with open(filename, 'w+') as f:
+        f.write(source)
+        f.flush()
```
disconnect3d: Is there any reason to `flush` here? The "with statement" should do that for you (as well as closing the file).

cloudburst: No idea, I didn't write that :P

cloudburst: Just tested and it works fine with `flush()` removed but I don't know if there is some case for it being there.

zachriggle: Looks like for 2.7, `file.flush` invokes the native routine `file_flush` which invokes `fflush`.

`file.close` invokes the native routine `file_close` --> `close_the_file` --> `PyFileObject->f_close` --> `close(2)`.  There is no call to any flush routine.

`file.__exit__` just invokes the `close` method on the `file` object.

According to StackOverflow, `file.close()` implies `file.flush()`, but I don't see any code for it in the `cpython` source.

Separately, `flush()` does not guarantee that the data is actually on disk.  That requires calling `fsync()` or using `O_SYNC`.

So it appears that the call to `f.flush()` is both unnecessary *and* incorrect.

zachriggle:
We still need to call `fsync` to actually ensure the data hits the disk.  However, the file descriptor is closed outside of the `with` block, and we can't guarantee the buffers are flushed in the `with` block without `.flush()`.

An alternate option is to provide a `fileobj` instead of a `filename`, and run `gcc` in stdin mode` (e.g. `echo 'int main(){ puts("Hello!"); }' | gcc -xc - ; ./a.out`).

This would avoid the need for temporary files and `flush` / `fsync` altogether

cloudburst:
Do you have a preference of the two?

zachriggle:
The smallest change is to put back the `f.flush()` and add `os.fsync(f.fileno())` to ensure the data hits the disk

cloudburst:
Tests OK on my setup with 4c01d7d

disconnect3d:
Btw the `open` in Python 2 and 3 has a 3rd optional argument which specifies buffering.
However I am not really sure if `unbuffered` (3rd argument with value 0) really means "write the data to the disk immediately", so I am not sure if that helps anyhow...

The `open` also differs between Python 2 and 3. In Python 3 you can pass more arguments to it and it also deodes the string you read from it:
* https://docs.python.org/2/library/functions.html#open
* https://docs.python.org/3/library/functions.html#open

PR #210, opened 2017-04-05
Enable customizable context display and ordering

zachriggle:


pwndbg/commands/context.py
```diff
@@ -49,19 +51,19 @@ def context(*args):
...
-    if 'c' in args: result.extend(context_code())
-    if 'c' in args: result.extend(context_source())
-    if 'a' in args: result.extend(context_args())
-    if 's' in args: result.extend(context_stack())
-    if 'b' in args: result.extend(context_backtrace())
+    for arg in args:
+        if 'r' == arg: result.extend(context_regs())
+        if 'd' == arg: result.extend(context_code())
+        if 's' == arg: result.extend(context_source())
+        if 's' == arg: result.extend(context_stack())
```
disconnect3d: The `'s'` is doubled. If that is intentional, go with `result.extend((context_source(), context_stack())`.

I would also go with `if` + `elif`s instead of multiple `if`. If the arg is always one of `r,d,s,b` we can also make a dict mapping here.

zachriggle: Fixed!

pwndbg/commands/context.py
```diff
@@ -49,19 +61,17 @@ def context(*args):
     Accepts subcommands 'reg', 'code', 'stack', 'backtrace', and 'args'.
     """
     if len(args) == 0:
```
disconnect3d: Maybe just `if not args:`

pwndbg/commands/context.py
```diff
@@ -49,19 +61,17 @@ def context(*args):
...
-        args = ['reg','code','stack','backtrace','args']
+        args = str(config_context_sections).split()
 
     args = [a[0] for a in args]
```
disconnect3d: Since `args` is used just once, maybe `args = (a[0] for a in args)`.

pwndbg/commands/context.py
```diff
@@ -194,7 +195,7 @@ def context_source():
                     source_lines[i] = C.highlight(source_lines[i])
                     break
 
-        banner = [pwndbg.ui.banner("code")]
+        banner = [pwndbg.ui.banner("source")]
```
```python
banner = [pwndbg.ui.banner("source")] + source_lines
```

Or just `return ...`.

pwndbg/commands/context.py
zachriggle: I'm not sure it really matters to use a `list` versus a `genexpr` for a small number of elements

Issue #213, opened 2017-04-08
Missing typeinfos


disconnect3d:
Some of the binaries might not contain type information about `unsigned long`, `unsigned short` and so call to `gdb.lookup_type(...)` crashes.
I am attaching an example from one of the CTFs.

```
[dc@dc:ransomware_re]$ gdb ./ransomware -q
Loaded 106 commands.  Type pwndbg [filter] for a list.
Reading symbols from ./ransomware...(no debugging symbols found)...done.
pwndbg> run
Starting program: /home/dc/asis/ransomware_re/ransomware 
Exception during func=pwndbg.typeinfo.update ()
Traceback (most recent call last):
  File "/home/dc/installed/pwndbg/pwndbg/events.py", line 122, in caller
    func()
  File "/home/dc/installed/pwndbg/pwndbg/typeinfo.py", line 40, in update
    module.ulong  = gdb.lookup_type('unsigned long')
gdb.error: No type named unsigned long.
Exception during func=pwndbg.events._start_newobjfile (<gdb.NewObjFileEvent object at 0x7f2be28f6468>,)
Traceback (most recent call last):
  File "/home/dc/installed/pwndbg/pwndbg/events.py", line 122, in caller
    func()
  File "/home/dc/installed/pwndbg/pwndbg/events.py", line 178, in _start_newobjfile
    gdb.events.start.on_new_objfile()
  File "/home/dc/installed/pwndbg/pwndbg/events.py", line 60, in on_new_objfile
    function()
  File "/home/dc/installed/pwndbg/pwndbg/events.py", line 128, in caller
    raise e
  File "/home/dc/installed/pwndbg/pwndbg/events.py", line 122, in caller
    func()
  File "/home/dc/installed/pwndbg/pwndbg/typeinfo.py", line 40, in update
    module.ulong  = gdb.lookup_type('unsigned long')
gdb.error: No type named unsigned long.
Python Exception <class 'gdb.error'> No type named unsigned long.: 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".
[Inferior 1 (process 26086) exited with code 0377]
pwndbg> 
```

In above example we can actually fetch `size_t` (as we do later `module.size_t = module.ulong` anyway, and `size_t` works with `gdb.lookup_type`). However there are other types that are missing, e.g. `unsigned short`.

[...]

PR #214, opened 2017-04-08
Added alernate typeinfo lookup names for D lang (#213)

disconnect3d:
Names fetched from: https://dlang.org/ctod.html#types

### DIFF
--- a/pwndbg/typeinfo.py
+++ b/pwndbg/typeinfo.py
@@ -24,6 +24,7 @@ import pwndbg.memoize
 
 module = sys.modules[__name__]
 
+
 def is_pointer(value):
     type = value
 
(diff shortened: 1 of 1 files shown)


pwndbg/typeinfo.py
```diff
@@ -36,22 +36,31 @@ def is_pointer(value):
...
+    module.uint8  = module.uchar
+    module.uint16 = module.ushort
+    module.uint32 = module.uint
+    module.uint64 = lookup_first_type(('unsigned long long', 'ulong'))
 
     module.int8   = gdb.lookup_type('char')
     module.int16  = gdb.lookup_type('short')
     module.int32  = gdb.lookup_type('int')
-    module.int64  = gdb.lookup_type('long long')
+    module.int64  = lookup_first_type(('long long', 'long'))
```
zachriggle: This is strictly wrong for 32-bit binaries

pwndbg/typeinfo.py
```diff
@@ -36,22 +36,31 @@ def is_pointer(value):
...
+    module.uint   = lookup_first_type(('unsigned int', 'uint'))
     module.void   = gdb.lookup_type('void')
-    module.uint8  = gdb.lookup_type('unsigned char')
-    module.uint16 = gdb.lookup_type('unsigned short')
-    module.uint32 = gdb.lookup_type('unsigned int')
-    module.uint64 = gdb.lookup_type('unsigned long long')
+    module.uint8  = module.uchar
+    module.uint16 = module.ushort
+    module.uint32 = module.uint
+    module.uint64 = lookup_first_type(('unsigned long long', 'ulong'))
```
zachriggle: This is strictly wrong for 32-bit binaries

pwndbg/typeinfo.py
```diff
@@ -36,22 +36,31 @@ def is_pointer(value):
 @pwndbg.events.start
 @pwndbg.events.stop
 def update():
+
+    def lookup_first_type(types):
```
zachriggle: I'd prefer pushing this up to the global scope as `lookup_type` which just takes `*args` so we can comma-separate (rather than passing a tuple) and instead of raising a custom exception, raise the most recent exception `except:` -> `except e` and `raise Exception` -> `raise e`.

pwndbg/typeinfo.py
zachriggle: Nevermind, after looking at the `ctod` docs, I see that D defines these differently.

pwndbg/typeinfo.py
zachriggle: Nevermind, after looking at the `ctod` docs, I see that D defines these differently.

zachriggle:
It might be possible to just never reload certain types.  For example, the size of most of those types which are changed by this are actually identical for 32- and 64-bit binaries.  We don't need to reload any of these types except `long` and `ulong`.

It also seems that the cause for this is that GDB uses `lookup_symbol_in_language` which is language-specific (thus the problems with D binaries).  Is there any way for us to determine the binary language from within GDB?

disconnect3d:
@zachriggle Yes we can, just found that:
https://ftp.gnu.org/old-gnu/Manuals/gdb/html_chapter/gdb_11.html#SEC86
Here is it in action:
https://asciinema.org/a/6n2o6qliyfhiklttsy3kis5fu

However we would have to check whether that works all times (e.g. when symbols are stripped and so on).

disconnect3d:
@zachriggle 
There are also other languages.
```
pwndbg> set language 
Requires an argument. Valid arguments are auto, local, unknown, ada, c, c++, asm, minimal, d, fortran, objective-c, go, java, modula-2, opencl, pascal, rust.
```

However I am not sure whether they are problematic at all... I have tried to compile some of the Rust examples from [Rust By Example](http://rustbyexample.com/) and they were all reported as C by `show language` .

Anyway, what do you think about having a dict mapping between languages and their type names? Like this:
```python
languages_types = {
    'c': {
        'char': 'char',
        'ulong': 'unsigned long',
       // and so on..
    }
    'd': { ... }
}
```

zachriggle:
I think we should just be able to force the language to C and then put it back, right?

zachriggle:
In any case, I'd prefer to avoid a mapping, since it really shouldn't matter in what language the binary was written.  Ideally, we can just stick to the lowest-common-denominator (C).

disconnect3d:
@zachriggle Do you mean wrapping the type lookups with:
```python
gdb.execute('set language c')
// lookup and assign types
gdb.execute('set language auto')
```

When I did that and launched the binary compiled in D, the `entry` command crashed on `gdb.lookup_type`:
```
gdb.error: No type named unsigned long.
Python Exception <class 'gdb.error'> No type named unsigned long.: 
```
After that exception the `show language` shown `C` that was set by the script. It seems the language change takes place after all of our Python code is executed :|.

PR #216, opened 2017-04-11
Add withHexrays wrapper around decompile()

### LINKED ISSUE #196: IDA Pro Integration - decompile() segfault
Calling `idaapi.decompile()` causes my IDA `6.9.151221 (December 21, 2015)` to segfault (often corrupting the database) when Hex-Rays is not installed.

There needs to be some sanity checking in the [context](https://github.com/pwndbg/pwndbg/blob/master/pwndbg/commands/context.py#L209) command before calling `pwndbg.ida.decompile()`

I'm not sure if this is fixed in the latest IDA `6.95.160808 (August 08, 2016)`

zachriggle:
Fixes #196

### DIFF
--- a/pwndbg/ida.py
+++ b/pwndbg/ida.py
@@ -74,6 +74,12 @@ class withIDA(object):
             return self.fn(*args, **kwargs)
         return None
 
+def withHexrays(f):
+    @withIDA
+    @functools.wraps(f)
+    def wrapper(*a, **kw):
+        if _ida.init_hexrays_plugin():
+            return f(*a, **kw)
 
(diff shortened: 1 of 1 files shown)


pwndbg/ida.py
```diff
@@ -323,6 +329,7 @@ def has_cached_cfunc(addr):
 
 
 @withIDA
```
cloudburst: Is this line redundant since the Hexrays decorator also implies the same?

cloudburst:
This patch fixes the issue on my setup, thanks!

PR #221, opened 2017-04-17
Use inthook only in gdb with Python 2

disconnect3d:
Because of the inthook one can't use OR operation with Python 3 enums:

```
pwndbg> py import re; re.DOTALL | re.MULTILINE
  File "/usr/lib/python3.6/enum.py", line 794, in __or__
    return cls._missing_(value)
  File "/usr/lib/python3.6/enum.py", line 760, in _missing_
  File "/usr/lib/python3.6/enum.py", line 786, in _create_pseudo_member_
    pseudo_member._name_ = None
Error while executing Python code.
```

### DIFF
--- a/pwndbg/inthook.py
+++ b/pwndbg/inthook.py
@@ -18,37 +18,35 @@ from future.utils import with_metaclass
 
 import pwndbg.typeinfo
 
-if sys.version_info < (3,0):
+if sys.version_info < (3, 0):
     import __builtin__ as builtins
-else:
-    import builtins
 
-_int = builtins.int
+    _int = builtins.int
 
-# We need this class to get isinstance(7, xint) to return True
-class IsAnInt(type):
-    def __instancecheck__(self, other):
-        return isinstance(other, _int)
+    # We need this class to get isinstance(7, xint) to return True
+    class IsAnInt(type):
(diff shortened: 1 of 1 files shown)


disconnect3d:
Just as an information, the problem doesn't occur in gdb with Python 2: https://asciinema.org/a/cl5pjup8o9lpt8w1ue496q4mm

disconnect3d:
This bug is also a blocker for asm syntax highlighting ( https://github.com/anthraxx/pwndbg/tree/feature/syntax-highlight ) developed by @anthraxx .

zachriggle:
The inthook is necessary for Python3 as well

disconnect3d:
@zachriggle Is it? Quick look didn't show any bugs.

Anyway, I have found a solution for this. I will add it in another PR in a moment.

disconnect3d:
Another solution added in PR https://github.com/pwndbg/pwndbg/pull/222.

zachriggle:
@disconnect3d We use the `inthook` to force all invocations `int(gdb.Value object)` to be unsigned longs.  This is not the default behavior.

disconnect3d:
Solution will be merged in PR https://github.com/pwndbg/pwndbg/pull/222.

PR #222, opened 2017-04-17
Fix inthook enums OR operation in Python 3

disconnect3d:
The bug can be reproduced using such code:
```
pwndbg> py import re; re.DOTALL | re.MULTILINE
  File "/usr/lib/python3.6/enum.py", line 794, in __or__
    return cls._missing_(value)
  File "/usr/lib/python3.6/enum.py", line 760, in _missing_
  File "/usr/lib/python3.6/enum.py", line 786, in _create_pseudo_member_
    pseudo_member._name_ = None
Error while executing Python code.
```

The fix seems to work with Python 2 as well:
http://asciinema.org/a/czopcwobjwt2dk9vzif0zk49u

### DIFF
--- a/pwndbg/inthook.py
+++ b/pwndbg/inthook.py
@@ -18,17 +18,19 @@ from future.utils import with_metaclass
 
 import pwndbg.typeinfo
 
-if sys.version_info < (3,0):
+if sys.version_info < (3, 0):
     import __builtin__ as builtins
 else:
     import builtins
 
 _int = builtins.int
 
+
 # We need this class to get isinstance(7, xint) to return True
 class IsAnInt(type):
     def __instancecheck__(self, other):
         return isinstance(other, _int)
+
 
 class xint(with_metaclass(IsAnInt, builtins.int)):
     def __new__(cls, value, *a, **kw):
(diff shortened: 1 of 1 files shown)


pwndbg/inthook.py
```diff
@@ -12,26 +12,32 @@
 
 import os
 import sys
+import enum
```
zachriggle: isort is going to complain about this

zachriggle:
LGTM, just need to run `isort`

pwndbg/inthook.py
```diff
@@ -12,26 +12,32 @@
...
+
 # We need this class to get isinstance(7, xint) to return True
 class IsAnInt(type):
     def __instancecheck__(self, other):
         return isinstance(other, _int)
 
+
 class xint(with_metaclass(IsAnInt, builtins.int)):
     def __new__(cls, value, *a, **kw):
+        if isinstance(cls, enum.EnumMeta):
```
zachriggle: It might actually be better to have this at the end of the two `if` statements, and change them to `if` / `elif` so that we catch all unknown types.

disconnect3d: Fixed :+1:

disconnect3d:
@zachriggle Fixed imports, finally can be merged ;).
My `isort` returned different import ordering for `future` module... (probably py2 vs py3 issue...)

anthraxx:
awesome, then i can continue with my stuff :heart:

PR #226, opened 2017-04-18
Update arch.py

r1git:
Because of import unicode_literals, fmt is unicode and breaks struct.(un)pack on some configs.
Have to enforce str(fmt) before struct.(un)pack.

### DIFF
--- a/pwndbg/arch.py
+++ b/pwndbg/arch.py
@@ -68,10 +68,10 @@ def update():
         m.qemu = m.current
 
 def pack(integer):
-    return struct.pack(fmt, integer & ptrmask)
+    return struct.pack(str(fmt), integer & ptrmask)
 
 def unpack(data):
-    return struct.unpack(fmt, data)[0]
+    return struct.unpack(str(fmt), data)[0]
 
 def signed(integer):
     return unpack(pack(integer), signed=True)


disconnect3d:
Can you provide an example of configuration on which the break occurs?

r1git:
$ python
Python 2.7.5+ (default, Feb 27 2014, 19:37:08)

$ gdb -v
GNU gdb (GDB) 7.12.1

$ pip list | grep 'pip\|pycparser\|psutil\|python-ptrace\|pyelftools\|isort\|six\|future\|ROPgadget\|unicorn'
future (0.16.0)
futures (3.0.5)
isort (4.2.5)
pip (9.0.1)
psutil (5.2.2)
pycparser (2.17)
pyelftools (0.24)
python-ptrace (0.8.1)
six (1.10.0)
unicorn (1.0.0)

The error:
$ gdb -q /bin/ls
Loaded 106 commands.  Type pwndbg [filter] for a list.
pwndbg> entry
Temporary breakpoint 1 at 0x404888

Temporary breakpoint 1, 0x0000000000404888 in ?? ()
Struct() argument 1 must be string, not unicode

    Accepts subcommands 'reg', 'code', 'stack', 'backtrace', and 'args'.

zachriggle:
We should not be calling `str()` on every invocation of `pack` and `unpack`.  Any saved format strings should be cast to `str` once when they are instantiated.

Separately, I cannot reproduce this, even when using `unicode` strings specifically:

```
>>> struct.pack(u'L', 3)
'\x03\x00\x00\x00\x00\x00\x00\x00'
```

disconnect3d:
Can't reproduce either - tried on:
* machine with GNU gdb (Debian 7.7.1+dfsg-5) 7.7.1 + Python 2.7.9 (default, Jun 29 2016, 13:11:10) [GCC 4.9.2]
* machine with GNU gdb (GDB) 7.12 + Python 3.6.0 (default, Jan 16 2017, 12:12:55) [GCC 6.3.1 20170109]

Python version printed as:
```
pwndbg> py import sys; print(sys.version)
```

r1git:
@zachriggle Agreed, but when I tried to fix it at definition level with `fmt = str('=I')` instead, it strangely did not fix the problem.
@disconnect3d Ok, will try to do more test on the faulty machine to narrow it down.

r1git:
@zachriggle This is what I get on the faulty system:
```
$ python
Python 2.7.5+ (default, Feb 27 2014, 19:37:08) 
[GCC 4.8.1] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import struct
>>> struct.pack(u'L', 3)
TypeError: Struct() argument 1 must be string, not unicode
```
So it's definitely coming from that issue. Not linked to pwndbg.
Sorry for wasting your time.

PR #227, opened 2017-04-18
Add version command

disconnect3d:
This adds pwndbg's git commit id to pwndbg banner.
image

This might be handy for future bug reports as there were cases when someone had out-of-date pwndbg version and/or didn't report pwndbg version.

### DIFF
--- a/pwndbg/prompt.py
+++ b/pwndbg/prompt.py
@@ -5,9 +5,6 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-import os
-import subprocess
-
 import gdb
 
 import pwndbg.events
@@ -15,25 +12,8 @@ import pwndbg.memoize
 import pwndbg.stdio
 
 
-def show_version():
-    try:
-        git_path = os.path.join(os.path.dirname(os.path.dirname(pwndbg.__file__)), '.git')
-        commit_id = subprocess.check_output(['git', '--git-dir', git_path, 'rev-parse', 'HEAD'])
-        commit_id = commit_id[:8].decode('utf-8')
-
-        version_msg = 'Pwndbg version: %s' % commit_id
-        print(pwndbg.color.light_yellow(version_msg))
-    except:
-        pass
-
-
-def show_hint():
-    hint_msg = 'Loaded %i commands. Type pwndbg [filter] for a list.' % len(pwndbg.commands._Command.commands)
-    print(pwndbg.color.red(hint_msg))
-
-
-show_version()
-show_hint()
+hint_msg = 'Loaded %i commands. Type pwndbg [filter] for a list.' % len(pwndbg.commands._Command.commands)
+print(pwndbg.color.red(hint_msg))
 cur = (gdb.selected_inferior(), gdb.selected_thread())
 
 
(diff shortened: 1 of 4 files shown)


disconnect3d:
@anthraxx please look over this as I am not sure if we have `.git` when we install pwndbg through AUR on Arch Linux.

anthraxx:
No, there is definitivly no git checkout when installing as package, its a pure export as is

pwndbg/prompt.py
```diff
@@ -5,17 +5,28 @@
...
 import gdb
 
 import pwndbg.events
 import pwndbg.memoize
 import pwndbg.stdio
 
-msg = "Loaded %i commands.  Type pwndbg [filter] for a list." % len(pwndbg.commands._Command.commands)
-print(pwndbg.color.red(msg))
+git_path = os.path.join(os.path.dirname(os.path.dirname(pwndbg.__file__)), '.git')
+commit_id = subprocess.check_output(['git', '--git-dir', git_path, 'rev-parse', 'HEAD'])
```
zachriggle: This will throw an exception if Git is unavailable or the command fails

disconnect3d:
@zachriggle Added a workaround of doing nothing when git is not available... Also rebased to master.

pwndbg/prompt.py
```diff
@@ -5,17 +5,38 @@
...
 import pwndbg.stdio
 
-msg = "Loaded %i commands.  Type pwndbg [filter] for a list." % len(pwndbg.commands._Command.commands)
-print(pwndbg.color.red(msg))
 
+def show_version():
+    try:
+        git_path = os.path.join(os.path.dirname(os.path.dirname(pwndbg.__file__)), '.git')
+        commit_id = subprocess.check_output(['git', '--git-dir', git_path, 'rev-parse', 'HEAD'])
+        commit_id = commit_id[:8].decode('utf-8')
```
zachriggle: We should use `--short` instead of truncating to `8` characters, since 4 bytes of SHA is not guaranteed to be unique

pwndbg/prompt.py
```diff
@@ -5,17 +5,38 @@
...
 import gdb
 
 import pwndbg.events
 import pwndbg.memoize
 import pwndbg.stdio
 
-msg = "Loaded %i commands.  Type pwndbg [filter] for a list." % len(pwndbg.commands._Command.commands)
-print(pwndbg.color.red(msg))
 
+def show_version():
```
zachriggle: Can we move this into a `command` named `version`?

pwndbg/prompt.py
disconnect3d: I am truncating it to 8 characters (however it is 4 bytes of SHA I believe). Anyway `--short` returns shorter strings (7 characters here):
```bash
$ git rev-parse --short HEAD
e9443fd
```

pwndbg/prompt.py
disconnect3d: Yes. Should we launch this command on startup then?

What about situation when git is not there? Should I detect it and output "Can't fetch git commit id to display version" or try to fetch version first and don't load the command at all if git is not there?

zachriggle: We should likely do something similar to Pwntools and have a `version.py` which has the release version: https://github.com/Gallopsled/pwntools/blob/dev/pwnlib/version.py

disconnect3d:
This is how it looks right now:

image

zachriggle:
Love love love it <3

The "version number" should be the date in `YYYY.MM.bugfix` (i.e., 2016.04)

PR #233, opened 2017-04-19
Fixed GDB remote get command

reu-res:


zachriggle:
I expect that this simple quoting won't guarantee that the issue is fixed.

Can you provide a test-case where this failed?  I expect we have a need for a `sh_string` equivalent which generates GDB-compatible escaped strings.

reu-res:
Before patch
image

After patch
image

zachriggle:
I definitely understand that this fixes your specific issue.  However, while we're taking the time to fix this, I'd prefer to fix it once, in a reusable manner, and permanently.  Do we know how unprintable characters and embedded quotes work on the GDB command-line?

zachriggle:

PR #239, opened 2017-04-21
Fix Ehdr and Phdr endian switching

disconnect3d:
Also fixes regression - issue #156.
Before this changes we were creating ElfXX_Ehdr and ElfXX_Phdr structures once during initialization using `pwndbg.ctypes.Structure` as base class (which was `ctypes.LittleEndianStructure` - as the default `pwndbg.arch.endian` is 'little').
We were thinking that the proper Ehdr/Phdr is set based on architecture (`pwndbg.arch.ptrsize`) and endian (as we updated the `pwndbg.ctypes.Structure` attribute to either `ctypes.LittleEndianStructure` or `ctypes.BigEndianStructure`).
However, changing the `pwndbg.ctypes.Structure` didn't really refresh the ELfXX_Ehdr and ElfXX_Phdr structures, so they still inherited from the old `pwndbg.ctypes.Structure` (which was `ctypes.LittleEndianStructure`).

Because of above, when we loaded a binary which was big endian, we failed on parsing its Ehdr and/or Phdr structures (as we just thought we updated Ehdr/Phdr to proper endian instead of really doing it; and so we interpreted Ehdr/Phdr fields as little endian).

This commit fixes the issue by removing the `pwndbg.ctypes` totally and by creating Ehdr and Phdr classes for each architecture (`pwndbg.arch.ptrsize` - 4 or 8) and endian (`pwndbg.arch.endian` - 'big' or 'little').
To make the code not so much bloated when it is possible the classes are generated.


disconnect3d:
@zachriggle Want me to change it to module reloading and recreating the structures every X occurs (where X is appropriate event, haven't looked at previous implementation/source code yet)? :P

zachriggle:
For my "vision" of `pwndbg.ctypes` to work, we'd need to do dynamic reloading of classes when the architecture changes, which requires some extra work.  In the meantime, sticking as close to the existing code as possible is preferred.  I expect the issue is simply that the types aren't being reloaded correctly / re-instantiated :(

disconnect3d:
Closing as this should target `stable` anyway. Also found a solution with module reloading, PR inc.

zachriggle:
You can just change the target branch lol 😂

disconnect3d:
Lol, right! Didn't see that combobox earlier :P

Anyway, the other PR has different implementation ^^

zachriggle:
👍

Issue #232, opened 2017-04-19
Search command bugs and deficiencies


disconnect3d:
There are some weird cases about `search` command [all checked just on gdb with Python 3]:

1) "How can I use that?"
```
pwndbg> help search

Search memory for byte sequences, strings, pointers, and integer values

pwndbg> pwndbg search
ropper               ROP gadget search with ropper.
search               Search memory for byte sequences, strings, pointers, and integer values
pwndbg> search
usage: search [-h] [-t {byte,short,word,dword,qword,pointer,string,bytes}]
              [-1] [-2] [-4] [-8] [-p] [-x] [-s] [-e] [-w] [--save]
              [--no-save] [-n]
              value [mapping]
search: error: the following arguments are required: value
```
Probably `help search` and/or `pwndbg search` should show the usage as well?

2) The `-t <type>` is optional but I can't really specify `-t string`:
```
pwndbg> search -t string "aaa"
must be str, not bytes
'search': Search memory for byte sequences, strings, pointers, and integer values
```
It doesn't even work with `u'...'` or `unicode('...')`.

3) This probably should have better output:
```
pwndbg> search -t byte "a"
int() argument must be a string, a bytes-like object or a number, not 'NoneType'
'search': Search memory for byte sequences, strings, pointers, and integer values
```

4) I can do a bytes search which works with `u'...'` strings, not sure if that is intended:
```
pwndbg> search -t bytes u"/bin/ls"
warning: Unable to access 16000 bytes of target memory at 0x7ffff7a13507, halting search.
warning: Unable to access 16007 bytes of target memory at 0x7ffff7bda000, halting search.
pwndbg> search -t bytes "/bin/ls"
warning: Unable to access 16000 bytes of target memory at 0x7ffff7a13506, halting search.
warning: Unable to access 16006 bytes of target memory at 0x7ffff7bda000, halting search.
[stack]         0x7fffffffdc2f 0x736c2f6e69622f /* '/bin/ls' */
[stack]         0x7fffffffeff0 0x736c2f6e69622f /* '/bin/ls' */
```

5) The ugly `warning: Unable to access 16000 bytes of target memory at 0x7ffff7a13507, halting search.` - however this is actually a problem with gdb's `find` command and there is no good solution to disable this warning for now.
I will probably make a feature request to gdb to add a parameter/setting to disable warnings (e.g. `set warnings off`).

zachriggle:
To address them in order:

1. Maybe.  That'd be nice :)
[...]

PR #242, opened 2017-04-23
Fix issue when searching for strings

zachriggle:
Fixes part of #232

### DIFF
--- a/pwndbg/commands/search.py
+++ b/pwndbg/commands/search.py
@@ -129,6 +129,16 @@ def search(type, hex, string, executable, writable, value, mapping, save, next):
         value = value.encode()
         value += b'\x00'
 
+    # Find the mappings that we're looking for
+    mappings = pwndbg.vmmap.get()
+
+    if mapping:
+        mappings = [m for m in mappings if mapping in m.objfile]
+
+    if not mappings:
+        print(M.red("Could not find mapping %r" % mapping))
+        return
+
     # Prep the saved set if necessary
     global saved
     if save:
--- a/pwndbg/commands/search.py
+++ b/pwndbg/commands/search.py
@@ -126,6 +126,7 @@ def search(type, hex, string, executable, writable, value, mapping, save, next):
 
     # Null-terminate strings
     elif type == 'string':
+        value = value.encode()
         value += b'\x00'
 
     # Prep the saved set if necessary
--- a/pwndbg/commands/search.py
+++ b/pwndbg/commands/search.py
@@ -73,7 +73,7 @@ parser.add_argument('-w', '--writable', action='store_true',
                     help='Search writable segments only')
 parser.add_argument('value', type=str,
                     help='Value to search for')
-parser.add_argument('mapping', type=str, nargs='?', default=None,
+parser.add_argument('mapping_name', type=str, nargs='?', default=None,
                     help='Mapping to search [e.g. libc]')
 parser.add_argument('--save', action='store_true', default=None,
                     help='Save results for --resume.  Default comes from config %r' % auto_save.name)
@@ -84,7 +84,7 @@ parser.add_argument('-n', '--next', action='store_true',
 
 @pwndbg.commands.ArgparsedCommand(parser)
 @pwndbg.commands.OnlyWhenRunning
-def search(type, hex, string, executable, writable, value, mapping, save, next):
+def search(type, hex, string, executable, writable, value, mapping_name, save, next):
     # Adjust pointer sizes to the local architecture
     if type == 'pointer':
         type = {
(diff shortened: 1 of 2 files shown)


disconnect3d:
Are you sure this is working?

```
pwndbg> version
Gdb: GNU gdb (GDB) 7.12
Python: 3.6.0 (default, Jan 16 2017, 12:12:55)  [GCC 6.3.1 20170109]
Pwndbg: 1.0.0 build: 79c0e842

pwndbg> set context-sections 
Set which context sections are displayed by default (also controls order) to ''
pwndbg> b main
Breakpoint 1 at 0x419fd0

pwndbg> r somestringargv
Starting program: /usr/bin/bash somestringargv

Breakpoint 1, 0x0000000000419fd0 in main ()
Breakpoint main

pwndbg> search
usage: search [-h] [-t {byte,short,word,dword,qword,pointer,string,bytes}] [-1] [-2] [-4] [-8] [-p] [-x] [-s] [-e] [-w] [--save] [--no-save] [-n] value [mapping]
search: error: the following arguments are required: value

pwndbg> search somestring
search() got an unexpected keyword argument 'mapping'
'search': Search memory for byte sequences, strings, pointers, and integer values

pwndbg> search "somestring"
search() got an unexpected keyword argument 'mapping'
'search': Search memory for byte sequences, strings, pointers, and integer values

pwndbg> search -t string "somestring"
search() got an unexpected keyword argument 'mapping'
'search': Search memory for byte sequences, strings, pointers, and integer values

pwndbg> search -t string somestring
search() got an unexpected keyword argument 'mapping'
'search': Search memory for byte sequences, strings, pointers, and integer values

pwndbg> search -t bytes somestring
search() got an unexpected keyword argument 'mapping'
'search': Search memory for byte sequences, strings, pointers, and integer values

pwndbg> search -t bytes "somestring"
search() got an unexpected keyword argument 'mapping'
'search': Search memory for byte sequences, strings, pointers, and integer values
```

pwndbg/search.py
```diff
@@ -17,15 +17,28 @@
...
     value = searchfor
     size  = None
 
     i = gdb.selected_inferior()
 
-    maps = pwndbg.vmmap.get()
+    maps = mappings or pwndbg.vmmap.get()
     hits = []
 
     if end and start:
```
disconnect3d, line 44: Probably would be better to change below `maps = [...]` into `maps = (...)` or maybe moving the ifs into `for` (however the `start` and `end` in `for` override the outer ones so this would have to be renamed I guess).

However both changes are not really super needed. In worst case scenario we have ~400 items in the list (tried on `google-chrome`).

pwndbg/commands/search.py
```diff
@@ -126,8 +126,19 @@ def search(type, hex, string, executable, writable, value, mapping, save, next):
...
     # Null-terminate strings
     elif type == 'string':
+        value = value.encode()
         value += b'\x00'
 
+    # Find the mappings that we're looking for
+    mappings = pwndbg.vmmap.get()
+
+    if mapping:
+        mappings = [m for m in mappings if mapping in m.objfile]
```
disconnect3d: `mapping` and `mappings` seems confusing to me...

Also below there is `pwndbg.search.search` called with `mapping=mapping` keyword argument and you have changed that command to get `mappings` keyword argument (NOTE: additional `s`). This is the reason the code is failing atm.

disconnect3d (changes requested):

pwndbg/commands/search.py
disconnect3d: It should be just `mappings=mappings` in 149 line.

pwndbg/commands/search.py
```diff
@@ -126,8 +126,19 @@ def search(type, hex, string, executable, writable, value, mapping, save, next):
...
+
+    if not mappings:
+        print(M.red("Could not find mapping %r" % mapping))
```
disconnect3d: We've got above `import pwndbg.color.memory as M`

disconnect3d: Also if we want to show 'errors' in red, we should probably do that with all of them

pwndbg/commands/search.py
zachriggle: `mapping` is a string, e.g. `"libc.so"`.

`mappings` is a list of `pwndbg.memory.Page` objects which describe memory mappings and their regions.

zachriggle:
@disconnect3d can you approve the review so this can merge to `stable`?

disconnect3d (approved):

disconnect3d:
@zachriggle sure

EDIT: Weird as it still says `review required`...

zachriggle:
Huh.

PR #246, opened 2017-04-24
Fix Python 2 inthook (and so entry/start commands #234)

### LINKED ISSUE #234: Python2 Broken
```
pwndbg> start
Temporary breakpoint 1 at 0x41eab0
Struct() argument 1 must be string, not unicode

```

disconnect3d:
This fixes #234, at least the bug reported by me.

Tested and works on:
```
Pwndbg: 1.0.0 build: 52abfa95
```

And (here also tested the synchronization with IDA Pro and it works):
```
Gdb: GNU gdb (GDB) 7.12
Python: 3.6.0 (default, Jan 16 2017, 12:12:55)  [GCC 6.3.1 20170109]
Pwndbg: 1.0.0 build: 52abfa95
```

### DIFF
--- a/pwndbg/inthook.py
+++ b/pwndbg/inthook.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 """
-This hook is necessary for compatibility with Python2.7 versions of GDB
+This hook is necessary for compatibility with Python 2.7 versions of GDB
 since they cannot directly cast to integer a gdb.Value object that is
 not already an integer type.
 """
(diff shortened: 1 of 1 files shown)


zachriggle:
Since this is a bug-fix and not an improvement, it should target the oldest branch which is affected (in this case, it's likely `stable`)

disconnect3d:
Retargeted. However, this should probably wait until we understand #234 better.

PS: Should I also make a seperate PR for dev branch? Or we will merge/rebase(I guess not as this may break some development) stable to dev? Or just cherry-pick that commit to dev(?)

pwndbg/inthook.py
```diff
@@ -17,15 +17,19 @@
...
-if sys.version_info < (3, 0):
+if pwndbg.compat.python2:
     import __builtin__ as builtins
 else:
     import builtins
+    long = int
 
 _int = builtins.int
 
+types_default_int_new = six.string_types + (long, )
```
zachriggle, line 31: This should explode on Python3

zachriggle:
Fixed with #250

PR #247, opened 2017-04-24
WIP: Make pdb great again

disconnect3d:
Enchances debugging. The screen is worth more than ~~ words:

image

WIP: Checking this on Python 2...


disconnect3d:
I am closing this as there is better solution to debugging done in #251.

There also has been similar work in #248.

PR #248, opened 2017-04-24
Use pwndbg.stdio for all pdb function calls

zachriggle:
This is an example alternative to #247 which prevents stdio from being hooked by GDB when interacting with PDB

### DIFF
--- a/pwndbg/pdb.py
+++ b/pwndbg/pdb.py
@@ -1,0 +1,30 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+This module serves as utility for the developers to get pdb debugging session right, whenever standard:
+    import pdb; pdb.set_trace()
+Can't do its job properly (that happens often as gdb does some magic with stdout/stderr).
+
+Thanks to Dmoreno: http://stackoverflow.com/questions/17074177/how-to-debug-python-cli-that-takes-stdin
+
+Usage:
+    import pwndbg.pdb
+
+Importing the module will print out pretty message on what to do next.
+"""
+import functools
+import pdb
+import os
+
+import pwndbg.stdio
+
+for name, value in list(pdb.__dict__.items()):
+	if not callable(value):
+		continue
+
+	@functools.wraps(value)
+	def wrapper(*a, **kw):
+		with pwndbg.stdio.stdio:
+			return value(*a, **kw)
+
+	setattr(pdb, name, value)
--- a/pwndbg/stdio.py
+++ b/pwndbg/stdio.py
@@ -20,16 +20,7 @@ import pwndbg.compat
 
 
 def get(fd, mode):
-    if pwndbg.compat.python3:
-        file = io.open(fd, mode=mode, buffering=0)
-
-        kw ={}
-        if pwndbg.compat.python3:
-            kw['write_through']=True
-
-        return io.TextIOWrapper(file, **kw)
-    else:
-        return codecs.open(fd, mode, 'utf-8')
+    return io.open(fd, mode + 'b', buffering=0)
 
 
 class Stdio(object):
@@ -37,11 +28,10 @@ class Stdio(object):
 
     def __enter__(self, *a, **kw):
         self.queue.append((sys.stdin, sys.stdout, sys.stderr))
-
         if pwndbg.compat.python3 or True:
-            sys.stdin  = get('/dev/stdin', 'rb')
-            sys.stdout = get('/dev/stdout', 'wb')
-            sys.stderr = get('/dev/stderr', 'wb')
+            sys.stdin  = get('/dev/stdin', 'r')
+            sys.stdout = get('/dev/stdout', 'w')
+            sys.stderr = get('/dev/stderr', 'w')
 
     def __exit__(self, *a, **kw):
         sys.stdin, sys.stdout, sys.stderr = self.queue.pop()
--- a/pwndbg/__init__.py
+++ b/pwndbg/__init__.py
@@ -60,6 +60,7 @@ import pwndbg.heap
 import pwndbg.inthook
 import pwndbg.memory
 import pwndbg.net
+import pwndbg.pdb
 import pwndbg.proc
 import pwndbg.prompt
 import pwndbg.regs


disconnect3d:
That works much better then the version proposed by mine PR (#247) :heart:.

pwndbg/pdb.py
```diff
@@ -0,0 +1,30 @@
...
+
+Importing the module will print out pretty message on what to do next.
+"""
+import functools
+import pdb
+import os
+
+import pwndbg.stdio
+
+for name, value in list(pdb.__dict__.items()):
```
disconnect3d, line 21: Isn't patching `pdb.set_trace` enough?

pwndbg/pdb.py
```diff
@@ -0,0 +1,30 @@
...
+
+Thanks to Dmoreno: http://stackoverflow.com/questions/17074177/how-to-debug-python-cli-that-takes-stdin
+
+Usage:
+    import pwndbg.pdb
+
+Importing the module will print out pretty message on what to do next.
+"""
+import functools
+import pdb
```
disconnect3d, line 16: If the proposed solution works with `ipdb` patch it as well (if it is available).

Maybe something like:
```python
def wrap_with_stdio(func):
    # wrapping code
```

Or `module` instead of `func`.

pwndbg/stdio.py
```diff
@@ -20,28 +20,18 @@
...
+    return io.open(fd, mode + 'b', buffering=0)
 
 
 class Stdio(object):
     queue = []
 
     def __enter__(self, *a, **kw):
         self.queue.append((sys.stdin, sys.stdout, sys.stderr))
-
         if pwndbg.compat.python3 or True:
```
disconnect3d, line 31: This `if` is redundant because of `or True`.

zachriggle: Yes, it's some left-over stuff from troubleshooting this code.

tl;dr is that having a non-buffering file descriptor which transparently takes any of `str`, `unicode` (Python2), and `bytes` (Python3) is annoying.  I think I found a least-common denominator that works, though.

pwndbg/pdb.py
zachriggle: Probably would be nice.  The issue is that currently we already wrap all invoked `Command` objects (see `commands/__init__.py`) so anything you reach interactively *should* already be wrapped.  Anything you reach non-interactively (i.e. from the `pi` interpreter) does not need to be wrapped.  Clearly, things don't work as they're supposed to.

disconnect3d: Hm.. the description from stdio.py says:
```
Provides functionality to circumvent GDB's hooks on sys.stdin and sys.stdout
which prevent output from appearing on-screen inside of certain event handlers.
```

Can we use it to prevent warnings in gdb's `find` command that we use through `gdb.execute`?

disconnect3d: > Clearly, things don't work as they're supposed to.
 
Does it mean that current implementation has problems when `pdb` is invoked in already wrapped stuff?

pwndbg/pdb.py
```diff
@@ -0,0 +1,30 @@
...
+# -*- coding: utf-8 -*-
+"""
+This module serves as utility for the developers to get pdb debugging session right, whenever standard:
+    import pdb; pdb.set_trace()
+Can't do its job properly (that happens often as gdb does some magic with stdout/stderr).
+
+Thanks to Dmoreno: http://stackoverflow.com/questions/17074177/how-to-debug-python-cli-that-takes-stdin
+
+Usage:
+    import pwndbg.pdb
```
disconnect3d, line 11: Also the docstring here has to be changed.

Maybe something like:
```
"""
This module wraps pdb and eventually ipdb (if available) modules with pwndbg.stdio 
This fixes problems with potential stdout/stdin/stderr buffering done by gdb
"""
```

disconnect3d (changes requested):
The documentation is there in docstrings, but wouldn't it be good to have some list of "good to look through" stuff (maybe CONTRIBUTING, however this might be too overwhelming...)?
That would reduce bus factor :D.

pwndbg/pdb.py
zachriggle: Probably

zachriggle: > Can we use it to prevent warnings

No, the warnings aren't emitted from Python.

> Does it mean that current implementation has problems

Probably

disconnect3d: Seems to be working in Python 2 (not sure if I am checking right):
```
$ git ll
* 25d0dff - (HEAD, pdb-stdio) Fix Python 2 inthook (and so entry/start commands #234) (19 seconds ago) <disconnect3d>
* aa77e5f - (zach/pdb-stdio) Use stdio wrapper for pdb (63 minutes ago) <Zach Riggle>
* 247d242 - (origin/dev, dev) fixed GDB remote get command (#241) (2 days ago) <Roman>
```

Pwndbg:
```
$ gdb /bin/ls -q

pwndbg> version
Pwndbg: 1.0.0 build: 25d0dffd

pwndbg> start
> /home/dc/pwndbg/pwndbg/commands/start.py(51)start()
-> if not address:

(Pdb) list 40
 35  	    generally 'main', 'init', or the entry point.
 36  	    """
 37  	    run = 'run ' + ' '.join(a)
 38  	
 39  	    symbols = ["main",
 40  	                "_main",
 41  	                "start",
 42  	                "_start",
 43  	                "init",
 44  	                "_init"]
 45  	

(Pdb) 
 46  	    for symbol in symbols:
 47  	        address = pwndbg.symbol.address(symbol)
 48  	        import pdb
 49  	        pdb.set_trace()
 50  	
 51  ->	        if not address:
 52  	            continue
 53  	
 54  	        b = gdb.Breakpoint(symbol, temporary=True)
 55  	        gdb.execute(run, from_tty=False, to_string=True)
 56  	        return

(Pdb) up
> /home/dc/pwndbg/pwndbg/commands/__init__.py(56)__call__()
-> return self.function(*args, **kwargs)

(Pdb) down
> /home/dc/pwndbg/pwndbg/commands/start.py(51)start()
-> if not address:

(Pdb) p symbol
u'main'

(Pdb) p address
None

(Pdb) 
```

disconnect3d: For Python 3 it makes... any command unusable:
https://asciinema.org/a/0a4h2fe7yfq5d18djtoav3ezr

I have tried to remove binary mode from `pwndbg.stdio` and that's what I got, huh:
```
$ gdb
pwndbg> version
  File "/home/dc/installed/pwndbg/pwndbg/commands/__init__.py", line 55, in __call__
    with pwndbg.stdio.stdio:
  File "/home/dc/installed/pwndbg/pwndbg/stdio.py", line 34, in __enter__
    sys.stdin  = get('/dev/stdin', 'r')
  File "/home/dc/installed/pwndbg/pwndbg/stdio.py", line 25, in get
    return io.open(fd, mode, buffering=0)
ValueError: can't have unbuffered text I/O
```

pwndbg/stdio.py
```diff
@@ -20,28 +20,18 @@
...
-            sys.stdin  = get('/dev/stdin', 'rb')
-            sys.stdout = get('/dev/stdout', 'wb')
-            sys.stderr = get('/dev/stderr', 'wb')
+            sys.stdin  = get('/dev/stdin', 'r')
+            sys.stdout = get('/dev/stdout', 'w')
+            sys.stderr = get('/dev/stderr', 'w')
 
     def __exit__(self, *a, **kw):
         sys.stdin, sys.stdout, sys.stderr = self.queue.pop()
```
disconnect3d, line 37: Also there is redundant:
```
if False:
```
below

disconnect3d:
This can be closed as there is a better solution in #251.

Issue #249, opened 2017-04-27
context throw an exception when controlling $pc which from illegal memory


0xddaa:
### Description

I forged the FILE structure and tried to control program flow.
However, `context` will throw an exception if $pc is from illegal memory.
```
pwndbg> x/3i $pc
=> 0x7ffff7708193 <_IO_flush_all_lockp+371>:    call   QWORD PTR [rax+0x18]
   0x7ffff7708196 <_IO_flush_all_lockp+374>:    cmp    eax,0xffffffff
   0x7ffff7708199 <_IO_flush_all_lockp+377>:    mov    eax,0xffffffff
```
The traceback is as below:

```
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7708193 in _IO_flush_all_lockp (do_lock=do_lock@entry=0) at genops.c:786
786     in genops.c
Traceback (most recent call last):
  File "/home/dada/tools/pwndbg/pwndbg/commands/__init__.py", line 56, in __call__
    return self.function(*args, **kwargs)
  File "/home/dada/tools/pwndbg/pwndbg/commands/__init__.py", line 130, in _OnlyWhenRunning
    return function(*a, **kw)
  File "/home/dada/tools/pwndbg/pwndbg/commands/context.py", line 64, in context
    result.extend(func())
  File "/home/dada/tools/pwndbg/pwndbg/commands/context.py", line 77, in context_regs
    result.extend(get_regs())
  File "/home/dada/tools/pwndbg/pwndbg/commands/context.py", line 120, in get_regs
    desc = pwndbg.chain.format(value)
  File "/home/dada/tools/pwndbg/pwndbg/chain.py", line 97, in format
    enhanced = pwndbg.enhance.enhance(chain[-2] + offset, code=code)
  File "/home/dada/tools/pwndbg/pwndbg/enhance.py", line 100, in enhance
    instr = pwndbg.disasm.one(value)
  File "/home/dada/tools/pwndbg/pwndbg/disasm/__init__.py", line 108, in one
    for insn in get(address, 1):
  File "/home/dada/tools/pwndbg/pwndbg/disasm/__init__.py", line 128, in get
    i = get_one_instruction(address)
  File "/home/dada/tools/pwndbg/pwndbg/memoize.py", line 47, in __call__
    value = self.func(*args, **kwargs)
  File "/home/dada/tools/pwndbg/pwndbg/disasm/__init__.py", line 100, in get_one_instruction
[...]

PR #251, opened 2017-04-27
Simplify command exception debugging and make stdio work correctly

zachriggle:
Replaces #247 and #248 and is overall much simpler.

I did not previously know about `sys.__stdio__` and friends -- this is a much more portable solution.

Introduces two new configuration points:

1. `set exception-verbose on` will cause full stack-traces to be printed when an exception is raised inside of a `Command` function
2. `set exception-debugger` will do the above, and open an interactive debugger

Future work might use `sys.excepthook` instead, and it would be *much* more robust.  However, I couldn't find a clean way to not invoke `pdb` each type I mis-typed a command like:

```
pwndbg> pi
>>> foo
NameError: name 'foo' is not defined
```

...without what looked like hacky code.  For now, this is a 90% solution that we need anyway (since `sys.excepthook` only looks at uncaught exceptions -- we discard some exceptions which wouldn't make it there, and would still need this `Command`-specific code).


pwndbg/commands/__init__.py
```diff
@@ -20,7 +21,16 @@
 import pwndbg.symbol
 import pwndbg.ui
 
-debug = True
+
+verbose = pwndbg.config.Parameter('exception-verbose', False, 'whether to print a full stacktracefor exceptions raised in Pwndbg commands')
```
disconnect3d: Missing space in `stacktracefor`.

pwndbg/commands/__init__.py
```diff
@@ -20,7 +21,16 @@
...
+debug = pwndbg.config.Parameter('exception-debugger', False, 'whether to debug exceptions raised in Pwndbg commands')
```
disconnect3d: <3 <3

disconnect3d:
What's the difference between `sys.__stdout__` and `sys.stdout`?

For me it looks like the same, Python 2:
```python
In [3]: print sys.__stdout__
<open file '<stdout>', mode 'w' at 0x7f1b2c4c4150>

In [4]: print sys.stdout
<open file '<stdout>', mode 'w' at 0x7f1b2c4c4150>
```

Python 3:
```python
In [2]: sys.__stdout__
Out[2]: <_io.TextIOWrapper name='<stdout>' mode='w' encoding='UTF-8'>

In [3]: sys.stdout
Out[3]: <_io.TextIOWrapper name='<stdout>' mode='w' encoding='UTF-8'>
```

zachriggle:
https://docs.python.org/2/library/sys.html#sys.__stdin__

disconnect3d:
Oh and the difference is there in gdb, my bad I haven't checked inside gdb...:
```python
pwndbg> py import sys
pwndbg> py print(sys.stdout)
<gdb.GdbOutputFile object at 0x7f0f1fd4e630>
pwndbg> py print(sys.__stdout__)
<_io.TextIOWrapper name='<stdout>' mode='w' encoding='UTF-8'>
```

pwndbg/commands/__init__.py
```diff
@@ -20,7 +21,15 @@
 import pwndbg.symbol
 import pwndbg.ui
 
-debug = True
+verbose = pwndbg.config.Parameter('exception-verbose', False, 'whether to print a full stacktracefor exceptions raised in Pwndbg commands')
```
disconnect3d: Missing space: `stacktracefor`

zachriggle:
Yeah, and it doesn't look like it's always a thing (two copies of GDB built with Python2 it's the same object, on my Ubuntu 14 box with default GDB and Python3, I get what you have in 
https://github.com/pwndbg/pwndbg/pull/251#issuecomment-297853165)

disconnect3d (approved):
Good job! Works for me in both Py2 and Py3 :+1: 
Just add the missing space in help string before merging :P

disconnect3d:
Hmm I have just tested https://github.com/pwndbg/pwndbg/issues/249 with this on:
```
pwndbg> version
Pwndbg: 1.0.0 build: de2328e8
```

And here is the result - with no verbose/debug the command crashes with no info:
```
$ gdb ./test -q
Reading symbols from ./test...(no debugging symbols found)...done.
pwndbg> entry
Temporary breakpoint 1 at 0x400080

Temporary breakpoint 1, 0x0000000000400080 in _start ()
pwndbg> ni
0x0000000000400085 in _start ()
pwndbg> 
```

With exception-debugger we get pdb, which is cool:
```
pwndbg> set exception-debugger on
Set whether to debug exceptions raised in Pwndbg commands to True
pwndbg> entry
Temporary breakpoint 2 at 0x400080

Temporary breakpoint 2, 0x0000000000400080 in _start ()
  File "/home/dc/pwndbg/pwndbg/commands/__init__.py", line 63, in __call__
  File "/home/dc/pwndbg/pwndbg/commands/__init__.py", line 138, in _OnlyWhenRunning
    return function(*a, **kw)
  File "/home/dc/pwndbg/pwndbg/commands/nearpc.py", line 81, in nearpc
    instructions = pwndbg.disasm.near(pc, lines, emulate=emulate)
  File "/home/dc/pwndbg/pwndbg/disasm/__init__.py", line 213, in near
    target_candidate, size_candidate = emu.single_step()
  File "/home/dc/pwndbg/pwndbg/emu/emulator.py", line 407, in single_step
    insn = pwndbg.disasm.one(pc)
  File "/home/dc/pwndbg/pwndbg/disasm/__init__.py", line 108, in one
    for insn in get(address, 1):
  File "/home/dc/pwndbg/pwndbg/disasm/__init__.py", line 128, in get
    i = get_one_instruction(address)
  File "/home/dc/pwndbg/pwndbg/memoize.py", line 47, in __call__
  File "/home/dc/pwndbg/pwndbg/disasm/__init__.py", line 100, in get_one_instruction
    pwndbg.disasm.arch.DisassemblyAssistant.enhance(ins)
  File "/home/dc/pwndbg/pwndbg/disasm/arch.py", line 55, in enhance
    enhancer.enhance_next(instruction)
  File "/home/dc/pwndbg/pwndbg/disasm/arch.py", line 111, in enhance_next
    instruction.target = self.next(instruction, call=True)
  File "/home/dc/pwndbg/pwndbg/disasm/x86.py", line 112, in next
    return super(DisassemblyAssistant, self).next(instruction, call)
  File "/home/dc/pwndbg/pwndbg/disasm/arch.py", line 147, in next
    addr = int(pwndbg.memory.poi(pwndbg.typeinfo.ppvoid, addr))
  File "/home/dc/pwndbg/pwndbg/memory.py", line 264, in poi
    return gdb.Value(addr).cast(type.pointer()).dereference()
error: evaluation of this expression requires the program to have a function "malloc".

> /home/dc/pwndbg/pwndbg/memory.py(264)poi()
-> return gdb.Value(addr).cast(type.pointer()).dereference()
(Pdb) 
```

And below - weird message when disabling exception-debugger (the context stuff) and I guess two exceptions(?):
```
pwndbg> set exception-debugger off
Set whether to debug exceptions raised in Pwndbg commands to False
object of type 'NoneType' has no len()

Set whether to print a full stacktracefor exceptions raised in Pwndbg commands to True
pwndbg> entry
Temporary breakpoint 3 at 0x400080

Temporary breakpoint 3, 0x0000000000400080 in _start ()
  File "/home/dc/pwndbg/pwndbg/commands/__init__.py", line 63, in __call__
  File "/home/dc/pwndbg/pwndbg/commands/__init__.py", line 138, in _OnlyWhenRunning
    return function(*a, **kw)
  File "/home/dc/pwndbg/pwndbg/commands/nearpc.py", line 81, in nearpc
    instructions = pwndbg.disasm.near(pc, lines, emulate=emulate)
  File "/home/dc/pwndbg/pwndbg/disasm/__init__.py", line 213, in near
    target_candidate, size_candidate = emu.single_step()
  File "/home/dc/pwndbg/pwndbg/emu/emulator.py", line 407, in single_step
    insn = pwndbg.disasm.one(pc)
  File "/home/dc/pwndbg/pwndbg/disasm/__init__.py", line 108, in one
    for insn in get(address, 1):
  File "/home/dc/pwndbg/pwndbg/disasm/__init__.py", line 128, in get
    i = get_one_instruction(address)
  File "/home/dc/pwndbg/pwndbg/memoize.py", line 47, in __call__
  File "/home/dc/pwndbg/pwndbg/disasm/__init__.py", line 100, in get_one_instruction
    pwndbg.disasm.arch.DisassemblyAssistant.enhance(ins)
  File "/home/dc/pwndbg/pwndbg/disasm/arch.py", line 55, in enhance
    enhancer.enhance_next(instruction)
  File "/home/dc/pwndbg/pwndbg/disasm/arch.py", line 111, in enhance_next
    instruction.target = self.next(instruction, call=True)
  File "/home/dc/pwndbg/pwndbg/disasm/x86.py", line 112, in next
    return super(DisassemblyAssistant, self).next(instruction, call)
  File "/home/dc/pwndbg/pwndbg/disasm/arch.py", line 147, in next
    addr = int(pwndbg.memory.poi(pwndbg.typeinfo.ppvoid, addr))
  File "/home/dc/pwndbg/pwndbg/memory.py", line 264, in poi
    return gdb.Value(addr).cast(type.pointer()).dereference()
error: evaluation of this expression requires the program to have a function "malloc".

object of type 'NoneType' has no len()

  File "/home/dc/pwndbg/pwndbg/commands/__init__.py", line 63, in __call__
  File "/home/dc/pwndbg/pwndbg/commands/__init__.py", line 138, in _OnlyWhenRunning
    return function(*a, **kw)
  File "/home/dc/pwndbg/pwndbg/commands/context.py", line 64, in context
    result.extend(func())
  File "/home/dc/pwndbg/pwndbg/commands/context.py", line 160, in context_disasm
    while len(result) < code_lines + 1:
TypeError: object of type 'NoneType' has no len()
```

I guess we should rather report that something bad happened and either inform the user that he can redo the thing with verbose and report the issue.

zachriggle:
The `malloc` thing is happening regardless of this particular PR.  I finally tried to make debugging work correctly in order to hunt that issue down :P

disconnect3d:
@zachriggle Oki. What about the fact that this doesn't display an information to the user that something bad happened when command crashes?

zachriggle:
@disconnect3d Modified the code to handle that case (just prints the exception) when both settings are off.  (Also re-organized the code a bit)

disconnect3d:
@zachriggle This is how it looks now:
```
$ gdb ./test 
Reading symbols from ./test...(no debugging symbols found)...done.
pwndbg> start
Temporary breakpoint 1 at 0x400080

Temporary breakpoint 1, 0x0000000000400080 in _start ()
Python Exception <class 'NameError'> name 'e' is not defined: 
pwndbg> 
```

zachriggle:
@disconnect3d What binary / what version of Python?

pwndbg/exception.py
```diff
@@ -0,0 +1,27 @@
...
+
+verbose = pwndbg.config.Parameter('exception-verbose', False, 'whether to print a full stacktracefor exceptions raised in Pwndbg commands')
+debug = pwndbg.config.Parameter('exception-debugger', False, 'whether to debug exceptions raised in Pwndbg commands')
+
+def handle():
+    # Display the error
+    if debug or verbose:
+        print(traceback.format_exc())
+    else:
+        print(e)
```
disconnect3d: `e` doesn't exist here :P

@zachriggle sry, deleted my previous post to write comment here... (below is what hapens when something fails - here exactly case from #249)
```
pwndbg> start
Temporary breakpoint 1 at 0x400080

Temporary breakpoint 1, 0x0000000000400080 in _start ()
Python Exception <class 'NameError'> name 'e' is not defined: 
```

disconnect3d (changes requested):

pwndbg/exception.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import pdb
```
disconnect3d: Would be nice to have here a try to import ipdb ( https://pypi.python.org/pypi/ipdb ):
```python
try:
    import ipdb as pdb
except ImportError:
    import pdb
```

pwndbg/exception.py
zachriggle: Oh, duh.

disconnect3d: We can probably grab the exception using inspect module (docs: [py2](https://docs.python.org/2/library/inspect.html), [py3](https://docs.python.org/3/library/inspect.html)).

pwndbg/exception.py
```diff
@@ -0,0 +1,33 @@
...
+
+verbose = pwndbg.config.Parameter('exception-verbose', False, 'whether to print a full stacktracefor exceptions raised in Pwndbg commands')
+debug = pwndbg.config.Parameter('exception-debugger', False, 'whether to debug exceptions raised in Pwndbg commands')
+
+def handle():
+    # Display the error
+    if debug or verbose:
+        print(traceback.format_exc())
+    else:
+        print(sys.exc_info()[1])
```
disconnect3d: Just for the weird malloc case it says:
```
evaluation of this expression requires the program to have a function "malloc".
```

Might be good to do:
```python
print("{}: {}".format(*sys.exc_info())
```
This would produce:
```
<class 'gdb.error'>: evaluation of this expression requires the program to have a function "malloc".
```

However... in cases like #249 this is the output (with the `.format` stuff):
```
<class 'gdb.error'>: evaluation of this expression requires the program to have a function "malloc".

<class 'TypeError'>: object of type 'NoneType' has no len()
```

---

In the end I think that it would be better to do one of those:

1. Just inform the user...
```python
print("Pwndbg error. If you want to see full stracktrace, fire `set exception-verbose on` and redo the thing you have done. Report the issue on https://github.com/pwndbg/pwndbg/issues")
```

2. Save the exception in a global state and let the user see it by either `py print(pwndbg.exception.last)` or `pwndbg-error` command.

3. Save the exception and let the user generate issue/bug report through a command.

disconnect3d:
Would be nice to merge it finally :F

zachriggle:
Rebased against current `dev`

zachriggle:
@disconnect3d @anthraxx can I get a quick sanity check of this branch before merging?  It seems to work for me.

disconnect3d:
@zachriggle sure

disconnect3d:
@zachriggle Works fine for me on py2 & py3.

One thing I am little worried about - we hook either `pdb.set_trace` or `ipdb.set_trace`.
If one fires: `import pdb; pdb.set_trace()` having `ipdb` installed, the stdio problems might occur.

We should hook both :P

zachriggle:
The `ipdb` and `pdb` thing can be worked out in a different PR.  Just wanted to land this one so that debugging worked :)

PR #253, opened 2017-04-28
Add Docker containers for Pwndbg testing

zachriggle:


disconnect3d:
Bump @zachriggle could you find some time and deliver this, please? :PP It would be nice to run tests on both py2 and py3 :f.

disconnect3d:
Closing as there is no work in here.
I think we can make one docker image with both Py2+Py3 + gdb built from sources and launch tests on both.

PR #254, opened 2017-04-28
Add RPC-based tests

zachriggle:
(Note: Duplicate of #170 with the PR coming from my fork)

Allow testing Pwndbg itself by embedding an XMLRPC server inside the GDB instance, which exposes both the `gdb` and `pwndbg` modules, as well as `eval`.

Since the default XMLRPC behavior is terrible and doesn't know how to marshall any of the following:

- `bytearray` object
- `bytes` object
- `str` or `unicode` with embedded `\x00`

We hack both the Marshaller and Unmarshaller to support a new datatype.


disconnect3d:
Do I understand correctly that the idea behind this is to be able to launch unit tests over easily testable functions (the ones that doesn't need running program under the hood)? As I guess we won't want to mock everything around (`pwndbg.proc.*`, getting memory and stuff).

Do we really need it so complicated? I mean, we probably could have something like:
* launch gdb
* send a `pi import pwndbg.tests; pwndbg.tests.run()`
* gather results

At least for the functions that doesn't need a process under the hood (well we could mock stuff but it might be painful/hard?).

We could also start with having some *end to end tests* that would be produced manually, launched with predefined environment (envvars, config, theme) and just checked for _this exact output for given list of commands_.
E2E tests could also hook some interesting functions - the ones that instead of printing return specific values (e.g. list of `Page` objects etc).

zachriggle:
Won’t work. GDB exits with code zero even if Python throws an exception.
Hacking around this is more work than just exposing the Pwndbg module over
RPC to allow “import pwndbg” outside of GDBpython.

disconnect3d:
@zachriggle still how are we going to test commands that use gdb api? By mocking it?
This won't us check easily whether pwndbg works fine with particular gdb version.

zachriggle:
Nope, by actually running GDB and exposing it (also) over XMLRPC.

disconnect3d:
Guess this can be closed due to https://github.com/pwndbg/pwndbg/pull/375 ?

Issue #235, opened 2017-04-19
Do not eat stack traces


zachriggle:
We should not eat all exceptions anywhere.

Any time a generic exception is raised, the full backtrace should be printed.

This may impact usability, but greatly increases the ability to fix bugs.

We should **additionally** print a user-friendly error message, but all `except Exception`-style handlers should always end with `raise`.

disconnect3d:
This issue is connected to https://github.com/pwndbg/pwndbg/issues/217

disconnect3d:
Closing as this is more or less covered by https://github.com/pwndbg/pwndbg/pull/300 or to be more specific:
https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/exception.py#L42-L68

Issue #217, opened 2017-04-11
Top-level Exception Handlers


zachriggle:
It seems that we could do well to register a top-level exception handler (sys.excepthook) and check for common `gdb` exceptions.

From the exception handler, we can walk the Python call stack and see which `Command` object was invoked (if any) and the lowest `pwndbg` call-frame to extract the offending internal function's name.

This would allow us to *generally* not have to worry about having pretty messages for the user doing something silly.

A specific example is the `hexdump` command here: https://github.com/pwndbg/pwndbg/blob/433bf23177c60590bacfc37bbed3d244e7872cde/pwndbg/commands/hexdump.py#L44

```py
    try:
        data = pwndbg.memory.read(address, count, partial=True)
    except gdb.error as e:
        print(e)
        return
```

zachriggle:
The top-level exception handler might save us in cases where we also don't expect exceptions (yet).  A convenient thing we could do is to log the full stack-trace to a file, and point the user to the Issue tracker here on Github.

For functions where we *expect* the exception may be raised (and know that we only want to display an error message), we might be able to use a decorator or a context manager.  For example:

```py
@PrintGdbError
@Command
def foo(...):
    ...
```

```py
def bar(...):
    with GdbErrorMessageHandler:
        gdb.DoSomeStuff(...)
```

Where `GdbErrorMessageHandler` (terrible name) will raise some other `Exception`-derived object which the top-level `sys.excepthook` will silently discard.

disconnect3d:
Possible extensions of above idea:
* have a setting for developers like `set debug-postmortem` which make the excetion handler calling `import pdb; pdb.post_mortem(exception_obj)` when something fails
* ask user for automatic error report generation and upload (not sure if that's possible on github) - that would contain: stacktrace, user's gdb, python, pwndbg version, ~/.gdbinit content, analysed binary and commands history.

disconnect3d:
Closing as this is more or less covered by https://github.com/pwndbg/pwndbg/pull/300 or to be more specific:
https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/exception.py#L42-L68

PR #256, opened 2017-04-30
added command 'got' to display status of the GOT table

degrigis:
Hey there, I've finally found the time to contribute to the project.

In order to start to become familiar with the codebase I've ported and enhanced the "got" command that I've implemented in peda last year. ( https://github.com/longld/peda/pull/79 )

It works without problems with x86 and x64 architecture with and without PIE. I've not tested yet other archs but as long as the format of the binary is an ELF it shouldn't create problems.
Let me know if there are any code style that must be corrected or any other problems :-)

Screenshot:
screen_got

Signed-off-by: degrigis <degrigis@gmail.com>


pwndbg/commands/got.py
```diff
@@ -0,0 +1,94 @@
...
+                                print("[%s] %s -> %s" % (address, light_yellow(name), pwndbg.chain.format(got_address)))
+                        else:
+                            for (address, info, rtype, value, name, _ , _ ) in f_line:
+                                if ispie:
+                                    address = hex(bin_text_base + int(address, 16))
+                                    got_address = struct.unpack("<Q",pwndbg.memory.read(int(address, 16), 8))[0]
+                                    print("[%s] %s -> %s" % (address, light_yellow(name),pwndbg.chain.format(got_address)))
+                                else:
+                                    got_address = struct.unpack("<Q", pwndbg.memory.read(int(address, 16), 8))[0]
+                                    print("[%s] %s -> %s" % (address,light_yellow(name),pwndbg.chain.format(got_address)))
```
disconnect3d: This print has three duplicates. It would be better to have it once - probably in the same scope (or rather indent) as `if pwndbg.arch.ptrsize == 4:`.

pwndbg/commands/got.py
```diff
@@ -0,0 +1,94 @@
...
+                                got_address = struct.unpack("<I", pwndbg.memory.read(int(address, 16), 4))[0]
+                                print("[%s] %s -> %s" % (address, light_yellow(name), pwndbg.chain.format(got_address)))
+                        else:
+                            for (address, info, rtype, value, name, _ , _ ) in f_line:
+                                if ispie:
+                                    address = hex(bin_text_base + int(address, 16))
+                                    got_address = struct.unpack("<Q",pwndbg.memory.read(int(address, 16), 8))[0]
+                                    print("[%s] %s -> %s" % (address, light_yellow(name),pwndbg.chain.format(got_address)))
+                                else:
+                                    got_address = struct.unpack("<Q", pwndbg.memory.read(int(address, 16), 8))[0]
```
disconnect3d: This should be moved to a variable: `int(address, 16)`

This particular line is also duplicated, and can be moved to the same scope the `if ispie` is in.

disconnect3d:
Hey,

Thanks for the contribution!

Have you checked the `pltgot` and `plt` commands? It seems your command does pretty the same or similar but with a better output.

I have got some issues to have it working on both Py2 and Py3 versions. See below.

Python 2 with hello world binary:
```gdb
$ gdb ./a.out -q
Loaded 107 commands. Type pwndbg [filter] for a list.
pwndbg> got

GOT protection: No RELRO | GOT functions: 3
 
  File "/home/dc/pwndbg/pwndbg/commands/__init__.py", line 56, in __call__
  File "/home/dc/pwndbg/pwndbg/commands/__init__.py", line 119, in _OnlyWithFile
    return function(*a, **kw)
  File "/home/dc/pwndbg/pwndbg/commands/got.py", line 79, in got
    for (address, info, rtype, value, name) in f_line:
ValueError: too many values to unpack

pwndbg> version
Pwndbg: 1.0.0 build: 982bbba9
```

Python 3 with /bin/ls binary:
```gdb
$ gdb /bin/ls
Loaded 107 commands. Type pwndbg [filter] for a list.
pwndbg> got

GOT protection: Partial RELRO | GOT functions: 108
 
  File "/home/dc/installed/pwndbg/pwndbg/commands/__init__.py", line 56, in __call__
  File "/home/dc/installed/pwndbg/pwndbg/commands/__init__.py", line 119, in _OnlyWithFile
    return function(*a, **kw)
  File "/home/dc/installed/pwndbg/pwndbg/commands/got.py", line 79, in got
    for (address, info, rtype, value, name) in f_line:
ValueError: too many values to unpack (expected 5)

pwndbg> version
Gdb: GNU gdb (GDB) 7.12.1
Pwndbg: 1.0.0 build: 982bbba9
```

degrigis:
Hi, from my understanding `plt` and `pltgot` merely display symbols found in the .got.plt and .plt sections. This command search only for JUMP relocation entries and tries to dereference their address in order to understand if a symbol to a libc function has been resolved yet or not. I find it useful when I try to hijack GOT table during exploitation of ctf challenges. Do you think it's unuseful? 

I'm investigating the errors since I can't reproduce them on my machine :-)

degrigis:
Hey @disconnect3d, I've done some tests with Py2 and Py3 but I couldn't replicate your issues.
Are you playing with binaries on an x86 machine or another arch? 

```
pwndbg> got

GOT protection: Full RELRO | GOT functions: 18
 
[0x555555756f38] free@GLIBC_2.2.5 -> 0x7ffff7a91940 (free) ◂— push   r13
[0x555555756f40] puts@GLIBC_2.2.5 -> 0x7ffff7a7d690 (puts) ◂— push   r12
[0x555555756f48] fread@GLIBC_2.2.5 -> 0x7ffff7a7c1a0 (fread) ◂— push   r13
[0x555555756f50] __stack_chk_fail@GLIBC_2.4 -> 0x7ffff7b26500 (__stack_chk_fail) ◂— lea    rdi, qword ptr [rip + 0x7637d]
[0x555555756f58] setbuf@GLIBC_2.2.5 -> 0x7ffff7a846b0 (setbuf) ◂— mov    edx, 0x2000
[0x555555756f60] malloc_usable_size@GLIBC_2.2.5 -> 0x7ffff7a924c0 (malloc_usable_size) ◂— push   r13
[0x555555756f68] __libc_start_main@GLIBC_2.2.5 -> 0x7ffff7a2e740 (__libc_start_main) ◂— push   r14
[0x555555756f70] fgets@GLIBC_2.2.5 -> 0x7ffff7a7bad0 (fgets) ◂— test   esi, esi
[0x555555756f78] strcmp@GLIBC_2.2.5 -> 0x7ffff7aac9c0 (__strcmp_sse2_unaligned) ◂— mov    eax, edi
[0x555555756f80] __gmon_start__ -> 0x0
[0x555555756f88] malloc@GLIBC_2.2.5 -> 0x7ffff7a91580 (malloc) ◂— push   rbp
[0x555555756f90] _IO_getc@GLIBC_2.2.5 -> 0x7ffff7a84030 (getc) ◂— push   rbx
[0x555555756f98] realloc@GLIBC_2.2.5 -> 0x7ffff7a91b10 (realloc) ◂— push   r15
[0x555555756fa0] __printf_chk@GLIBC_2.3.4 -> 0x7ffff7b23cf0 (__printf_chk) ◂— push   r12
[0x555555756fa8] strtoul@GLIBC_2.2.5 -> 0x7ffff7a493f0 (strtouq) ◂— mov    rax, qword ptr [rip + 0x3879e1]
[0x555555756fb0] exit@GLIBC_2.2.5 -> 0x7ffff7a48030 (exit) ◂— lea    rsi, qword ptr [rip + 0x3895c1]
[0x555555756fb8] fwrite@GLIBC_2.2.5 -> 0x7ffff7a7c6e0 (fwrite) ◂— push   r14
[0x555555756fc0] __cxa_finalize@GLIBC_2.2.5 -> 0x7ffff7a482d0 (__cxa_finalize) ◂— push   r15
pwndbg> version
Gdb: GNU gdb (GDB) 7.12
Python: 2.7.12 (default, Nov 19 2016, 06:48:10)  [GCC 5.4.0 20160609]
Pwndbg: 1.0.0 build: 351d479f
```


```
pwndbg> got

GOT protection: Full RELRO | GOT functions: 18
 
[0x555555756f38] free@GLIBC_2.2.5 -> 0x7ffff7a91940 (free) ◂— push   r13
[0x555555756f40] puts@GLIBC_2.2.5 -> 0x7ffff7a7d690 (puts) ◂— push   r12
[0x555555756f48] fread@GLIBC_2.2.5 -> 0x7ffff7a7c1a0 (fread) ◂— push   r13
[0x555555756f50] __stack_chk_fail@GLIBC_2.4 -> 0x7ffff7b26500 (__stack_chk_fail) ◂— lea    rdi, [rip + 0x7637d]
[0x555555756f58] setbuf@GLIBC_2.2.5 -> 0x7ffff7a846b0 (setbuf) ◂— mov    edx, 0x2000
[0x555555756f60] malloc_usable_size@GLIBC_2.2.5 -> 0x7ffff7a924c0 (malloc_usable_size) ◂— push   r13
[0x555555756f68] __libc_start_main@GLIBC_2.2.5 -> 0x7ffff7a2e740 (__libc_start_main) ◂— push   r14
[0x555555756f70] fgets@GLIBC_2.2.5 -> 0x7ffff7a7bad0 (fgets) ◂— test   esi, esi
[0x555555756f78] strcmp@GLIBC_2.2.5 -> 0x7ffff7aac9c0 (__strcmp_sse2_unaligned) ◂— mov    eax, edi
[0x555555756f80] __gmon_start__ -> 0x0
[0x555555756f88] malloc@GLIBC_2.2.5 -> 0x7ffff7a91580 (malloc) ◂— push   rbp
[0x555555756f90] _IO_getc@GLIBC_2.2.5 -> 0x7ffff7a84030 (getc) ◂— push   rbx
[0x555555756f98] realloc@GLIBC_2.2.5 -> 0x7ffff7a91b10 (realloc) ◂— push   r15
[0x555555756fa0] __printf_chk@GLIBC_2.3.4 -> 0x7ffff7b23cf0 (__printf_chk) ◂— push   r12
[0x555555756fa8] strtoul@GLIBC_2.2.5 -> 0x7ffff7a493f0 (strtouq) ◂— mov    rax, qword ptr [rip + 0x3879e1]
[0x555555756fb0] exit@GLIBC_2.2.5 -> 0x7ffff7a48030 (exit) ◂— lea    rsi, [rip + 0x3895c1]
[0x555555756fb8] fwrite@GLIBC_2.2.5 -> 0x7ffff7a7c6e0 (fwrite) ◂— push   r14
[0x555555756fc0] __cxa_finalize@GLIBC_2.2.5 -> 0x7ffff7a482d0 (__cxa_finalize) ◂— push   r15
pwndbg> version
Gdb: GNU gdb (Ubuntu 7.11.1-0ubuntu1~16.04) 7.11.1
Python: 3.5.2 (default, Nov 17 2016, 17:05:23)  [GCC 5.4.0 20160609]
Pwndbg: 1.0.0 build: 351d479f
```

anthraxx:
I get exactly the same exceptions as @disconnect3d does

degrigis:
Hi @anthraxx, can you please paste here the content of `f_line` before the
` for (address, info, rtype, value, name) in f_line:` ?

anthraxx:
```
f_line: [['000000619018', '000100000007', 'R_X86_64_JUMP_SLO', '0000000000000000', '__ctype_toupper_loc@GLIBC_2.3', '+', '0'], ['000000619020', '000200000007', 'R_X86_64_JUMP_SLO', '0000000000000000', '__uflow@GLIBC_2.2.5', '+', '0'], ['000000619028', '000300000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'getenv@GLIBC_2.2.5', '+', '0'], ['000000619030', '000400000007', 'R_X86_64_JUMP_SLO', '0000000000000000', '__snprintf_chk@GLIBC_2.3.4', '+', '0'], ['000000619038', '000500000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'free@GLIBC_2.2.5', '+', '0'], ['000000619040', '000600000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'localtime@GLIBC_2.2.5', '+', '0'], ['000000619048', '000700000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'abort@GLIBC_2.2.5', '+', '0'], ['000000619050', '000800000007', 'R_X86_64_JUMP_SLO', '0000000000000000', '__errno_location@GLIBC_2.2.5', '+', '0'], ['000000619058', '000900000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'strncpy@GLIBC_2.2.5', '+', '0'], ['000000619060', '000a00000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'strncmp@GLIBC_2.2.5', '+', '0'], ['000000619068', '004e00000007', 'R_X86_64_JUMP_SLO', '00000000004018d0', 'localtime_r@GLIBC_2.2.5', '+', '0'], ['000000619070', '000b00000007', 'R_X86_64_JUMP_SLO', '0000000000000000', '_exit@GLIBC_2.2.5', '+', '0'], ['000000619078', '000c00000007', 'R_X86_64_JUMP_SLO', '0000000000000000', '__fpending@GLIBC_2.2.5', '+', '0'], ['000000619080', '000d00000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'clock_gettime@GLIBC_2.17', '+', '0'], ['000000619088', '000e00000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'setenv@GLIBC_2.2.5', '+', '0'], ['000000619090', '000f00000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'textdomain@GLIBC_2.2.5', '+', '0'], ['000000619098', '001000000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'fclose@GLIBC_2.2.5', '+', '0'], ['0000006190a0', '001100000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'bindtextdomain@GLIBC_2.2.5', '+', '0'], ['0000006190a8', '001200000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'dcgettext@GLIBC_2.2.5', '+', '0'], ['0000006190b0', '001300000007', 'R_X86_64_JUMP_SLO', '0000000000000000', '__ctype_get_mb_cur_max@GLIBC_2.2.5', '+', '0'], ['0000006190b8', '001400000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'strlen@GLIBC_2.2.5', '+', '0'], ['0000006190c0', '001500000007', 'R_X86_64_JUMP_SLO', '0000000000000000', '__stack_chk_fail@GLIBC_2.4', '+', '0'], ['0000006190c8', '001600000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'getopt_long@GLIBC_2.2.5', '+', '0'], ['0000006190d0', '001700000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'mbrtowc@GLIBC_2.2.5', '+', '0'], ['0000006190d8', '001800000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'strchr@GLIBC_2.2.5', '+', '0'], ['0000006190e0', '001900000007', 'R_X86_64_JUMP_SLO', '0000000000000000', '__overflow@GLIBC_2.2.5', '+', '0'], ['0000006190e8', '001a00000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'strrchr@GLIBC_2.2.5', '+', '0'], ['0000006190f0', '001b00000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'gmtime_r@GLIBC_2.2.5', '+', '0'], ['0000006190f8', '001c00000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'lseek@GLIBC_2.2.5', '+', '0'], ['000000619100', '001d00000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'gettimeofday@GLIBC_2.2.5', '+', '0'], ['000000619108', '001e00000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'fputs@GLIBC_2.2.5', '+', '0'], ['000000619110', '001f00000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'memset@GLIBC_2.2.5', '+', '0'], ['000000619118', '002000000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'fscanf@GLIBC_2.2.5', '+', '0'], ['000000619120', '002100000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'strncat@GLIBC_2.2.5', '+', '0'], ['000000619128', '002200000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'close@GLIBC_2.2.5', '+', '0'], ['000000619130', '002300000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'fputc@GLIBC_2.2.5', '+', '0'], ['000000619138', '002500000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'memcmp@GLIBC_2.2.5', '+', '0'], ['000000619140', '002600000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'fputs_unlocked@GLIBC_2.2.5', '+', '0'], ['000000619148', '002700000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'calloc@GLIBC_2.2.5', '+', '0'], ['000000619150', '002800000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'putenv@GLIBC_2.2.5', '+', '0'], ['000000619158', '002900000007', 'R_X86_64_JUMP_SLO', '0000000000000000', '__getdelim@GLIBC_2.2.5', '+', '0'], ['000000619160', '002a00000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'strcmp@GLIBC_2.2.5', '+', '0'], ['000000619168', '002c00000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'memcpy@GLIBC_2.14', '+', '0'], ['000000619170', '002d00000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'tzset@GLIBC_2.2.5', '+', '0'], ['000000619178', '002e00000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'time@GLIBC_2.2.5', '+', '0'], ['000000619180', '002f00000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'fileno@GLIBC_2.2.5', '+', '0'], ['000000619188', '003000000007', 'R_X86_64_JUMP_SLO', '0000000000000000', '__stpcpy_chk@GLIBC_2.3.4', '+', '0'], ['000000619190', '003100000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'settimeofday@GLIBC_2.2.5', '+', '0'], ['000000619198', '003200000007', 'R_X86_64_JUMP_SLO', '0000000000000000', '__xstat@GLIBC_2.2.5', '+', '0'], ['0000006191a0', '003300000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'malloc@GLIBC_2.2.5', '+', '0'], ['0000006191a8', '003400000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'fflush@GLIBC_2.2.5', '+', '0'], ['0000006191b0', '003500000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'nl_langinfo@GLIBC_2.2.5', '+', '0'], ['0000006191b8', '003600000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'ungetc@GLIBC_2.2.5', '+', '0'], ['0000006191c0', '003700000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'clock_settime@GLIBC_2.17', '+', '0'], ['0000006191c8', '003800000007', 'R_X86_64_JUMP_SLO', '0000000000000000', '__freading@GLIBC_2.2.5', '+', '0'], ['0000006191d0', '003900000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'realloc@GLIBC_2.2.5', '+', '0'], ['0000006191d8', '003a00000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'fdopen@GLIBC_2.2.5', '+', '0'], ['0000006191e0', '003b00000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'setlocale@GLIBC_2.2.5', '+', '0'], ['0000006191e8', '003c00000007', 'R_X86_64_JUMP_SLO', '0000000000000000', '__printf_chk@GLIBC_2.3.4', '+', '0'], ['0000006191f0', '003d00000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'timegm@GLIBC_2.2.5', '+', '0'], ['0000006191f8', '003e00000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'strftime@GLIBC_2.2.5', '+', '0'], ['000000619200', '003f00000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'error@GLIBC_2.2.5', '+', '0'], ['000000619208', '004000000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'open@GLIBC_2.2.5', '+', '0'], ['000000619210', '004100000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'fseeko@GLIBC_2.2.5', '+', '0'], ['000000619218', '004200000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'fopen@GLIBC_2.2.5', '+', '0'], ['000000619220', '004300000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'gmtime@GLIBC_2.2.5', '+', '0'], ['000000619228', '004400000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'unsetenv@GLIBC_2.2.5', '+', '0'], ['000000619230', '004500000007', 'R_X86_64_JUMP_SLO', '0000000000000000', '__cxa_atexit@GLIBC_2.2.5', '+', '0'], ['000000619238', '004600000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'exit@GLIBC_2.2.5', '+', '0'], ['000000619240', '004700000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'fwrite@GLIBC_2.2.5', '+', '0'], ['000000619248', '004800000007', 'R_X86_64_JUMP_SLO', '0000000000000000', '__fprintf_chk@GLIBC_2.3.4', '+', '0'], ['000000619250', '004900000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'mbsinit@GLIBC_2.2.5', '+', '0'], ['000000619258', '004a00000007', 'R_X86_64_JUMP_SLO', '0000000000000000', 'iswprint@GLIBC_2.2.5', '+', '0'], ['000000619260', '004b00000007', 'R_X86_64_JUMP_SLO', '0000000000000000', '__ctype_tolower_loc@GLIBC_2.3', '+', '0'], ['000000619268', '004c00000007', 'R_X86_64_JUMP_SLO', '0000000000000000', '__ctype_b_loc@GLIBC_2.3', '+', '0'], ['000000619270', '004d00000007', 'R_X86_64_JUMP_SLO', '0000000000000000', '__sprintf_chk@GLIBC_2.3.4', '+', '0']]
  File "/home/anthraxx/Projects/external/pwndbg/pwndbg/commands/__init__.py", line 56, in __call__
  File "/home/anthraxx/Projects/external/pwndbg/pwndbg/commands/__init__.py", line 119, in _OnlyWithFile
    return function(*a, **kw)
  File "/home/anthraxx/Projects/external/pwndbg/pwndbg/commands/got.py", line 80, in got
    for (address, info, rtype, value, name) in f_line:
ValueError: too many values to unpack (expected 5)
```

degrigis:
@anthraxx mmmm, seems that you are executing the code for 32bit compiled binaries nonetheless you are debugging a 64bit binary. Is ` if pwndbg.arch.ptrsize == 4` the correct way to check if the inferior is 32/64 bit right?

I have on a 64bit machine the value 4 when I debug 32bit binaries and 8 when running 64bit binaries. ( as expected obviously :-) )

anthraxx:
It indeed works when starting the application rather then invoking right after loading the file. this happens as pwndbg/arch.py only loads the ptrsize on update() which is invoked only on application start.
@zachriggle If possible we may want to set certain things when a file is loaded rather then just when started

degrigis:
@anthraxx  ah that's why! Out of curiosity: I don't understand how do you launch the debugged binary in order to replicate the issue?
Regarding the issue: is there a built-in way to check for this avoiding that problem? 

In PEDA that was accomplished using:

`        out = self.execute_redirect('maintenance info sections ?').splitlines()
`

anthraxx:
@degrigis you could use the @pwndbg.commands.OnlyWhenRunning decorator if there is no way to make this work without

degrigis:
@anthraxx  done, let me know! :-)

pwndbg/commands/got.py
```diff
@@ -0,0 +1,94 @@
...
+    local_path = pwndbg.file.get_file(pwndbg.proc.exe)
+
+    for program in ['checksec', 'checksec.sh']:
+        program = pwndbg.which.which(program)
+        if program:
+            argv = [program, "--file" , local_path]
+            out = subprocess.check_output(argv).decode('utf-8')
+            break
+    else:
+        print('Could not find checksec or checksec.sh in $PATH.')
```
anthraxx: instead of duplicating logic, we should make use of the checksec.py code and keep one uniform way to work with checksec logic

pwndbg/commands/got.py
```diff
@@ -0,0 +1,94 @@
...
+
+    relro_status = "No RELRO"
+    if "Full RELRO" in out:
+        relro_status = "Full RELRO"
+    if "Partial RELRO" in out:
+        relro_status = "Partial RELRO"
+
+    program = pwndbg.which.which("file")
+
+    if program:
```
anthraxx: i believe it would improve readability and reduce complexity if we inverse the logic and early-exit instead of nested if blocks to go through all requirements

pwndbg/commands/got.py
```diff
@@ -0,0 +1,94 @@
...
+        argv = [program, local_path]
+        out = subprocess.check_output(argv).decode('utf-8')
+
+        if "statically" in out:
```
anthraxx: i believe it would improve readability and reduce complexity if we inverse the logic and early-exit instead of nested if blocks to go through all requirements

pwndbg/commands/got.py
```diff
@@ -0,0 +1,94 @@
...
+            print('File is statically linked.')
+        else:
+            program = pwndbg.which.which("readelf")
+            if program:
```
anthraxx: i believe it would improve readability and reduce complexity if we inverse the logic and early-exit instead of nested if blocks to go through all requirements

pwndbg/commands/got.py
```diff
@@ -0,0 +1,94 @@
...
+                argv = [program, "-r", local_path]
+                out = subprocess.check_output(argv).decode('utf-8')
+
+                if "Error" in out:
```
anthraxx: i believe it would improve readability and reduce complexity if we inverse the logic and early-exit instead of nested if blocks to go through all requirements

pwndbg/commands/got.py
```diff
@@ -0,0 +1,94 @@
...
+                    print("Error during relocation info")
+                else:
+                    f_line = ""
+                    for line in out.splitlines():
+                        if "JUMP" not in line:
+                            continue
+                        else:
+                            f_line = f_line + line + "\n"
+                    if f_line == "":
```
anthraxx: just a nitpick / bikeshed but it feels like `if not f_line` is more readable then equality check of empty string

anthraxx (changes requested):

pwndbg/commands/got.py
degrigis: yes, definitely a good idea ( I had some problem to re-use the function, but probably I made something wrong, I'm going to fix it right now ). I will make the function checksec return an empty string if something went wrong and later check this value inside my function.

pwndbg/commands/got.py
degrigis: fixed in latest commit 635d92

pwndbg/commands/got.py
degrigis: fixed in latest commit 635d92

pwndbg/commands/got.py
degrigis: yup ;-)

pwndbg/commands/got.py
degrigis: well, you are right, give me 10 minutes!

pwndbg/commands/got.py
degrigis: Moreover, why not using the dictionary idea of PEDA in order to return info about the checksec output? I liked that idea.
f.i. checksec['RELRO']=2
     checksec['PIE'] = 1
     checksec['NX'] = 0
      ...

This can avoid the awful strings check inside my code keeping it inside the checksec.py module.

anthraxx: hm yeah we could separate and split out the logic in checksec.py so we have a command that prints out the shell output and a function that returns parsed values out of it

degrigis: Ok, I'm going to push also this in a while.

degrigis:
@anthraxx @disconnect3d  this should be ok, I've added also the support function inside checksec.py. I don't completely like it since now we strictly depends from the checksec.sh output, but as long as it doesn't change too often this shouldn't be a big problem, what do you think?

pwndbg/commands/checksec.py
```diff
@@ -27,3 +27,49 @@ def checksec(file=None):
...
+def _checksec():
+    local_path = pwndbg.file.get_file(pwndbg.proc.exe)
+    result = {}
+
+    for program in ['checksec', 'checksec.sh']:
+        program = pwndbg.which.which(program)
+        if program:
+            cs_out = subprocess.check_output([program, '--file', local_path]).decode('utf-8')
+
+            if "Full RELRO" in cs_out:
```
zachriggle: This code doesn't belong in the command, it belongs in a library routine.

pwndbg/commands/got.py
```diff
@@ -0,0 +1,96 @@
...
+                got_address = struct.unpack("<I", pwndbg.memory.read(int(address, 16), 4))[0]
+                print("[%s] %s -> %s" % (address, light_yellow(name), pwndbg.chain.format(got_address)))
+        else:
+            for (address, info, rtype, value, name, _, _) in f_line:
+                addressval = int(address,16)
+
+                if ispie: # if PIE, address is only the offset from the binary base address
+                    addressval = bin_text_base + addressval
+
+                got_address = struct.unpack("<Q",pwndbg.memory.read(addressval, 8))[0]
```
zachriggle: Use the other pwndbg.memory functions, not `struct.unpack`

pwndbg/commands/got.py
```diff
@@ -0,0 +1,96 @@
...
+    if not f_line:
+        print("NO JUMP_SLOT entries available in the GOT")
+        return
+    else:
+        f_line = [' '.join(x.split()).split(" ") for x in f_line.split("\n")][:-1]
+        print("\nGOT protection: %s | GOT functions: %d\n " %(green(relro_status), len(f_line)))
+
+        if pwndbg.arch.ptrsize == 4:
+            for (address, info, rtype, value, name) in f_line:
+                got_address = struct.unpack("<I", pwndbg.memory.read(int(address, 16), 4))[0]
```
zachriggle: Use the other pwndbg.memory functions, not `struct.unpack`

pwndbg/commands/got.py
```diff
@@ -0,0 +1,96 @@
...
+import subprocess
+import struct
+
+import pwndbg.commands
+import pwndbg.commands.checksec
+import pwndbg.enhance
+import pwndbg.chain
+import pwndbg.which
+
+from pwndbg.color import *
```
zachriggle: The general convention in other parts of `pwndbg` is to import color as a module with a short name (e.g. `as C`), rather than `import *`

pwndbg/commands/got.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+@pwndbg.commands.Command
+@pwndbg.commands.OnlyWhenRunning
+@pwndbg.commands.OnlyWithFile
+def got():
+    '''
```
zachriggle, line 23: Almost all of this logic belongs in a library routine, which is invoked by the command routine.

zachriggle (changes requested):

pwndbg/commands/got.py
degrigis: ok great, changed with pwndbg.memory.pvoid

pwndbg/commands/got.py
degrigis: ok great, changed with pwndbg.memory.pvoid

pwndbg/commands/got.py
degrigis: ok, I've imported only the two colors I need.

pwndbg/commands/checksec.py
degrigis: I move the code inside a module checksec.py in /pwndbg/pwndbg ok?

pwndbg/commands/got.py
degrigis: @zachriggle , not sure how to fix this. Are you suggesting to retrieve the got table from a library routine and then deref the pointer inside the command?

zachriggle: I mean that this code should be refactored into different routines.  The `Command` routines should ideally be simple and just display data which is procured via other routines.  For example, fetching the relocations could be its own routine, as could the `file` output, etc.

degrigis:
I've reorganized the code following the guidelines of @zachriggle. Waiting for your comments :-)

pwndbg/checksec.py
```diff
@@ -0,0 +1,60 @@
...
+import sys
+import subprocess
+
+import pwndbg.file
+
+def checksec():
+    local_path = pwndbg.file.get_file(pwndbg.proc.exe)
+    result = {}
+
+    for program in ['checksec', 'checksec.sh']:
```
disconnect3d: It would be better to use `map` and `filter`:
```python
programs = ('checksec', 'checksec.sh')
for program in filter(None, map(pwndbg.which.which, programs)):
```

Then you won't need `if program:` later on.

pwndbg/checksec.py
```diff
@@ -0,0 +1,60 @@
...
+        program = pwndbg.which.which(program)
+        if program:
+            cs_out = subprocess.check_output([program, '--file', local_path]).decode('utf-8')
+
+            if "Full RELRO" in cs_out:
+                result['RELRO'] = 2
```
disconnect3d: We should probably use something like enum here (e.g. `from enum import Enum`).

This would however require us to add `enum34` into `requirements.txt` (as enum is a package for Python 2). I am not sure if we want to do that. @zachriggle what do you think?

Then it might look something like:
```python
class Relro(Enum):
    NO = 0
    PARTIAL = 1
    FULL = 2
    
    @staticmethod
    def from_checksec_str(string):
        # ...
```

pwndbg/checksec.py
```diff
@@ -0,0 +1,60 @@
...
+            if "Partial RELRO" in cs_out:
+                result['RELRO'] = 1
+            if "No RELRO" in cs_out:
+                result['RELRO'] = 0
+
+            if "Canary found" in cs_out:
+                result['CANARY'] = 1
```
disconnect3d: I would go with `True`/`False` in this and below parameters.

pwndbg/commands/checksec.py
```diff
@@ -26,4 +26,4 @@ def checksec(file=None):
         if program:
```
disconnect3d: This command should probably just call `pwndbg.checksec.checksec(file)`.

For the error message - it might be good to raise an exception in `checksec(..)` and just catch it here.

pwndbg/commands/got.py
```diff
@@ -0,0 +1,76 @@
...
+    if not file_out:
+        return 'Could not find file command in $PATH.'
+    if 'statically' in file_out:
+        return 'Binary is statically linked.'
+
+    jmpslots = pwndbg.elf.getjmpslots(local_path)
+    if not jmpslots:
+        return 'Could not find readelf command in $PATH.'
+    if 'Error' in jmpslots:
+        return 'Error during retrieving of relocation info.'
```
disconnect3d: If there is an error, you are returning `Error` string, so it would be better to check if it exactly matches... (as there might be function called `Error`).

Anyway, the error situation should probably raise an exception (which might be caught here, idk).

pwndbg/commands/got.py
```diff
@@ -0,0 +1,76 @@
...
+        bin_text_base = pwndbg.memory.page_align(pwndbg.elf.entry())
+
+    relro_status = "No RELRO"
+    if cs_out['RELRO'] == 2:
+        relro_status = "Full RELRO"
+    if cs_out['RELRO'] == 1:
+        relro_status = "Partial RELRO"
+
+    f_line = [' '.join(x.split()).split(" ") for x in jmpslots.split("\n")][:-1]
+    print("\nGOT protection: %s | GOT functions: %d\n " %(green(relro_status), len(f_line)))
```
disconnect3d: Missing space after `%`.

Also, some strings are `""` and some are `''`, would be nice to stay with one or another :P.

pwndbg/commands/got.py
```diff
@@ -0,0 +1,76 @@
...
+    if pwndbg.arch.ptrsize == 4:
+        for (address, info, rtype, value, name) in f_line:
+            addressval = int(address, 16)
+            got_address = pwndbg.memory.pvoid(addressval)
+            print("[%s] %s -> %s" % (address, light_yellow(name), pwndbg.chain.format(got_address)))
+    else:
+        for (address, info, rtype, value, name, _, _) in f_line:
+            addressval = int(address,16)
+
+            if ispie: # if PIE, address is only the offset from the binary base address
```
disconnect3d: Not sure if we really need `ispie` variable if it is used just once.

pwndbg/elf.py
```diff
@@ -285,3 +285,27 @@ def map_inner(ei_class, ehdr, objfile):
...
+    """
+    Use the readelf command with the parameter -r
+    in order to retrieve the jump slots inside the 
+    binary specified in path.
+    """
+    program = pwndbg.which.which("readelf")
+    if program:
+        argv = [program, "-r", path]
+        readelf_out = subprocess.check_output(argv).decode('utf-8')
+        if "Error" in readelf_out:
```
disconnect3d: It would be better to check for return code instead of a string content. Imagine having a function/relocation called `"Error"` ;).

E.g.:
```bash
[dc@dc:~]$ readelf -r a
readelf: Error: 'a': No such file
[dc@dc:~]$ echo $?
1
[dc@dc:~]$ readelf -r /bin/ls 1>/dev/null
[dc@dc:~]$ echo $?
0
```

pwndbg/elf.py
```diff
@@ -285,3 +285,27 @@ def map_inner(ei_class, ehdr, objfile):
...
+            return "Error"
+        else:
+            jmpslots = ""
+            for line in readelf_out.splitlines():
+                if "JUMP" not in line:
+                    continue
+                else:
+                    jmpslots = jmpslots + line + "\n"
+            return jmpslots
```
disconnect3d: I would go with:
```python
return '\n'.join(line for line in readelf_out.splitlines() if 'JUMP' in line) + '\n'
```

However this is also not ideal - it is not resistant to a function called `JUMP`.  We should rather check for particular column here.

disconnect3d (changes requested):

disconnect3d:
About handling errors/exceptions - there is also a discussion about it in https://github.com/pwndbg/pwndbg/issues/235 and https://github.com/pwndbg/pwndbg/issues/217.

pwndbg/elf.py
zachriggle: This also requires the `readelf` implementation to be able to handle the target architecture.  Pwntools has a way to handle this, but it requires you to have lots of `readelf` installed!

pwndbg/commands/got.py
zachriggle: Meh, I have no preference on this ^^^^

pwndbg/commands/got.py
zachriggle: Meh, I have no preference on this ^^^^

pwndbg/elf.py
zachriggle: `check_output` will raise an exception if the program exits with a nonzero status

pwndbg/elf.py
zachriggle: :P 
```py
relocations = filter(lambda l: 'JUMP' in l, readelf_out.splitlines() # or the list comprehension, whichever
return '\n'.join(relocations)
```

pwndbg/checksec.py
degrigis: @disconnect3d ,my code has been taken directly from the module commands/checksec.py. I don't hate particularly the `if program`, but if this is your preferred coding style let's modify it in both positions. ( pwndbg/checksec.py and commands/checksec.py )

pwndbg/checksec.py
degrigis: Waiting for @zachriggle feedback here :-)

pwndbg/elf.py
```diff
@@ -296,16 +296,10 @@ def getjmpslots(path):
...
-            for line in readelf_out.splitlines():
-                if "JUMP" not in line:
-                    continue
-                else:
-                    jmpslots = jmpslots + line + "\n"
-            return jmpslots
+        try:
+            readelf_out = subprocess.check_output(argv).decode('utf-8')
+        except: raise OSError("Error during readelf execution.")
+        relocations = filter(lambda l: "JUMP" in l, readelf_out.splitlines())
```
disconnect3d:
```
[dc@dc:~]$ readelf -r /bin/ls

Relocation section '.rela.dyn' at offset 0x16e0 contains 8 entries:
  Offset          Info           Type           Sym. Value    Sym. Name + Addend
00000061eff0  003300000006 R_X86_64_GLOB_DAT 0000000000000000 __libc_start_main@GLIBC_2.2.5 + 0
00000061eff8  003e00000006 R_X86_64_GLOB_DAT 0000000000000000 __gmon_start__ + 0
00000061f5e0  006c00000005 R_X86_64_COPY     000000000061f5e0 __progname@GLIBC_2.2.5 + 0
00000061f5e8  007800000005 R_X86_64_COPY     000000000061f5e8 stdout@GLIBC_2.2.5 + 0
00000061f5f0  007a00000005 R_X86_64_COPY     000000000061f5f0 optind@GLIBC_2.2.5 + 0
00000061f5f8  007500000005 R_X86_64_COPY     000000000061f5f8 optarg@GLIBC_2.2.5 + 0
00000061f600  006f00000005 R_X86_64_COPY     000000000061f600 __progname_full@GLIBC_2.2.5 + 0
00000061f620  007200000005 R_X86_64_COPY     000000000061f620 stderr@GLIBC_2.2.5 + 0

Relocation section '.rela.plt' at offset 0x17a0 contains 108 entries:
  Offset          Info           Type           Sym. Value    Sym. Name + Addend
00000061f018  000100000007 R_X86_64_JUMP_SLO 0000000000000000 __ctype_toupper_loc@GLIBC_2.3 + 0
00000061f020  000200000007 R_X86_64_JUMP_SLO 0000000000000000 __uflow@GLIBC_2.2.5 + 0
00000061f028  000300000007 R_X86_64_JUMP_SLO 0000000000000000 getenv@GLIBC_2.2.5 + 0
00000061f030  000400000007 R_X86_64_JUMP_SLO 0000000000000000 cap_to_text + 0
00000061f038  000500000007 R_X86_64_JUMP_SLO 0000000000000000 sigprocmask@GLIBC_2.2.5 + 0
00000061f040  000600000007 R_X86_64_JUMP_SLO 0000000000000000 __snprintf_chk@GLIBC_2.3.4 + 0
00000061f048  000700000007 R_X86_64_JUMP_SLO 0000000000000000 raise@GLIBC_2.2.5 + 0
00000061f050  006e00000007 R_X86_64_JUMP_SLO 0000000000402260 free@GLIBC_2.2.5 + 0
00000061f058  000800000007 R_X86_64_JUMP_SLO 0000000000000000 abort@GLIBC_2.2.5 + 0
00000061f060  000900000007 R_X86_64_JUMP_SLO 0000000000000000 __errno_location@GLIBC_2.2.5 + 0
00000061f068  000a00000007 R_X86_64_JUMP_SLO 0000000000000000 strncmp@GLIBC_2.2.5 + 0
00000061f070  008100000007 R_X86_64_JUMP_SLO 00000000004022a0 localtime_r@GLIBC_2.2.5 + 0
00000061f078  000b00000007 R_X86_64_JUMP_SLO 0000000000000000 _exit@GLIBC_2.2.5 + 0
00000061f080  000c00000007 R_X86_64_JUMP_SLO 0000000000000000 strcpy@GLIBC_2.2.5 + 0
(...)
```
This should check for `JUMP` in `Type` column instead of all line.

pwndbg/checksec.py
```diff
@@ -8,53 +8,52 @@
...
+                else:
+                    result['RPATH'] = True
+
+                if "No RUNPATH" in cs_out:
+                    result['RUNPATH'] = False
+                else:
+                    result['RUNPATH'] = True
+                return result
+
+            except:
```
disconnect3d: This try/except handles too much code and too broad exception - imagine if someone modifies e.g. line 30 and makes some stupid mistake - he/she would waste time looking for that :p.

Just do:
```python
try:
    cs_out = subprocess.check_output([program, '--file', path]).decode('utf-8')
except (OSError, subprocess.CalledProcessError):
    ...
```

(like here https://github.com/pwndbg/pwndbg/blob/stable/pwndbg/version.py#L24-L27 - however now I think this tries too much code too, huh)

pwndbg/checksec.py
disconnect3d: Imho it is hard to read code with tons of ifs, so would be nice to reduce them if possible.

(also it is hard to read code which has a big if block and then else with one line, it is better to do the reverse - `if not program: error` and then continue the code without new indent/code block)

pwndbg/elf.py
degrigis: uhm, sorry, forgotten!

pwndbg/commands/got.py
```diff
@@ -0,0 +1,64 @@
...
+    jmpslots = pwndbg.elf.getjmpslots(local_path)
+    if not len(jmpslots):
+        return "NO JUMP_SLOT entries available in the GOT"
+
+    if cs_out['PIE']:
+        bin_text_base = pwndbg.memory.page_align(pwndbg.elf.entry())
+
+    relro_status = "No RELRO"
+    if cs_out['RELRO'] == 2:
+        relro_status = "Full RELRO"
```
zachriggle: It looks like we're converting between `"Full RELRO"` to `2` back to `"Full RELRO"`.  Since the output of `checksec` may change, perhaps it's just best to avoid this conversion and capture the output?

pwndbg/file.py
```diff
@@ -134,3 +135,18 @@ def readlink(path):
     data = data[:-len(ending)]
 
     return data
+
+def file(path):
```
zachriggle: Having this module named `pwndbg.file` might be misleading.  Since this and `pwndbg.checksec` are both wrappers around command-line utilities, perhaps they might be under e.g. `pwndbg.wrappers.file` and `pwndbg.wrappers.checksec` rather than second-level modules.

degrigis: mmm yes, I like the idea! 

@zachriggle  @anthraxx  @disconnect3d  I was thinking about this: are we planning to have enough tool wrappers to justify a dedicated module? wrappers can be dangerous since ( as the checksec issue ) we strictly depends on their results.

pwndbg/commands/got.py
degrigis: Do you think that it is not worth having a structured output from checksec results? Without that  in the commands code we will have to write `if "PIE enabled" in cs_out['PIE']` instead of `if cs_out['PIE']`. That's not a big problem, but the dependency from checksec output still remains... 

I agree that is not so good that we depends from checksec output, a solution could be to implement directly thechecksec's techniques instead of using it as external tool, but again, is it worth it?

pwndbg/file.py
zachriggle: It's not so much a "dedicated module" it's just slightly more organization.

pwndbg/commands/got.py
zachriggle: For the current purposes, I'd rather avoid the transformation.  The numerical value `2` doesn't tell me *anything* without having to look at the source of this function.  At least with the string I know what the value may be.  If the output changes, we have to update the code regardless -- just potentially in an additional place.

pwndbg/file.py
disconnect3d: Agree with @zachriggle.

To be honest I would even go with moving all `gdb` related stuff into its own module.
So it would be easier to change gdb into other debugger (e.g. lldb - heard it has much better python api) in the future, if needed.

pwndbg/commands/got.py
degrigis: Ok, fair enough, will fix this as soon as possible.

degrigis:
Ok @zachriggle  I've added the first version of the wrapper module and refactored a bit the code.

pwndbg/wrappers.py
```diff
@@ -0,0 +1,94 @@
...
+from __future__ import unicode_literals
+
+import subprocess
+
+import pwndbg.file
+
+'''
+ Wrapper to checksec utility.
+ Returns a dictionary for each protection supported containing the
+ output of the utility.
```
disconnect3d: Put those as function docstring pls

pwndbg/wrappers.py
```diff
@@ -0,0 +1,95 @@
...
+    else:
+        raise OSError("Could not find file command in $PATH.")
+
+def readelf(path,argv):
+    """
+     Utility: /usr/bin/readelf
+    """
+    program = pwndbg.which.which("readelf")
+    if program:
+        argv = [program, argv, path]
```
disconnect3d: Probably should be `cmd = [program] + argv.split() + [path]`

disconnect3d: `file` and `readelf` share most of the logic; might be good to add a function called `launch_program` or `call_shell_command`

degrigis: Not sure about the last comment @disconnect3d. What are you suggesting here?

One more thing: do you think that the functions in the module wrappers should remain functions or it's better to organize them as classes? The latter methods could be useful to provides different support functions per utility. ( f.i. the checksec class could contains different methods to return input in different way: the dictionary, the raw string etc... )

disconnect3d: Something like:
```python
def call_program(progname, args=None):
    program = pwndbg.which.which(progname)

    if not program:
        raise OSError('Could not find %s command in $PATH.' % progname)

    cmd = (progname,)
    
    if args:
        cmd += tuple(args)  # in case someone pass list

    try:
        return subprocess.check_output(cmd).decode()
    except Exception as e:
        raise OSError('Error during execution of %s command: %s' % (progname, e))

readelf = functools.partial(call_program, 'readelf')
file = functools.partial(call_program, 'file')
```

disconnect3d: About classes - do we really need that at this moment? I would keep it simple stupid :P

A little bit offtopic, might be cool to watch: [Stop Writing Classes, Jack Diederich [Python US 2012]](https://www.youtube.com/watch?v=o9pEzgHorH0)

degrigis: Ah! Now got it! Interesting video, I'm going to invest next 30 minutes there. Thanks!

pwndbg/wrappers.py
```diff
@@ -0,0 +1,38 @@
...
+import pwndbg.which
+
+
+def call_program(progname, *args):
+    program = pwndbg.which.which(progname)
+
+    if not program:
+        raise OSError('Could not find %s command in $PATH.' % progname)
+
+    cmd = [progname]
```
disconnect3d: or just `(progname,) + args` or  `[progname] + list(args)` (however we don't really need list here)

pwndbg/commands/got.py
```diff
@@ -0,0 +1,77 @@
...
+            address_val = int(address,16)
+
+            if "PIE enabled" in cs_out: # if PIE, address is only the offset from the binary base address
+                address_val = bin_text_base + address_val
+
+            got_address = pwndbg.memory.pvoid(address_val)
+            print("[%s] %s -> %s" % (hex(address_val), light_yellow(name), pwndbg.chain.format(got_address)))
+
+
+def _extract_Jumps(l):
```
disconnect3d: I don't like the uppercase 'J' :P

pwndbg/commands/got.py
```diff
@@ -0,0 +1,77 @@
...
+        for line in jmpslots.splitlines():
+            address, info, rtype, value, name = line.split()
+            got_address = pwndbg.memory.pvoid(int(address, 16))
+            print("[%s] %s -> %s" % (address, light_yellow(name), pwndbg.chain.format(got_address)))
+    else:
+        for line in jmpslots.splitlines():
+            address, info, rtype, value, name, _, _ = line.split()
+            address_val = int(address,16)
+
+            if "PIE enabled" in cs_out: # if PIE, address is only the offset from the binary base address
```
disconnect3d: What about PIE when `pwndbg.arch.ptrsize=4`?

pwndbg/commands/got.py
```diff
@@ -0,0 +1,77 @@
...
+    local_path = pwndbg.file.get_file(pwndbg.proc.exe)
+    cs_out = pwndbg.wrappers.checksec("--file", local_path)
+
+    file_out = pwndbg.wrappers.file(local_path)
+    if "statically" in file_out:
+        return "Binary is statically linked."
+
+    readelf_out = pwndbg.wrappers.readelf("-r", local_path)
+
+    jmpslots = '\n'.join(filter(lambda l: _extract_Jumps(l),
```
disconnect3d: `_extract_Jumps` -> `_extract_jumps`

degrigis: omg, sorry, it's late here :laughing:

pwndbg/commands/got.py
```diff
@@ -0,0 +1,77 @@
...
+
+    relro_status = "No RELRO"
+    if "Full RELRO" in cs_out:
+        relro_status = "Full RELRO"
+    elif "Partial RELRO" in cs_out:
+        relro_status = "Partial RELRO"
+
+    print("\nGOT protection: %s | GOT functions: %d\n " % (green(relro_status), len(jmpslots.splitlines())))
+
+    if pwndbg.arch.ptrsize == 4:
```
disconnect3d: It seems for me that both branches differs only in `line.split()` and maybe PIE (if x86-32 doesn't do anything with PIE).

Maybe:
```python
for line in jmpslots.splitlines():
    address, info, rtype, value, name = line.split()[:5]  # x86-64 has two columns redundant for us 
```

pwndbg/commands/got.py
degrigis: I'm not sure that PIE is ever enabled in a 32 bit binaries ( at least I've never seen it but please correct me if I am wrong ). Anyway your proposed modification resolve this issue by design removing the `if`

pwndbg/commands/got.py
disconnect3d: @degrigis This command function should more or less look like this:
```python
def got():
    '''
    Shows the state of the Global Offset Table
    '''
    sec = pwndbg.wrappers.checksec()  # i guess it can get_file on its own
    # not sure if relocs is proper name for jmpslots
    relocs = pwndbg.readelf.relocs()

    print("\nGOT protection: %s | GOT functions: %d\n" % (green(relro_status), len(relocs))

    for (address, name, got_addr) in relocs:
        print('[0x%x] %s -> %s' % (address_val, light_yellow(name), pwndbg.chain.format(got_address)))
```

...or we might have some kind of `pwndbg.got.got()` which would return GOT related data...

Basically it is good to split business logic with the view - this gives us possibility of code-reuse and makes stuff easier to read.

Also less code == less bugs :D

disconnect3d: Some nice commands, just for reference/as an examples:
* https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/commands/vmmap.py
* https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/commands/auxv.py
* https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/commands/misc.py

degrigis: Ok, but we have to agree about the organization of the code. In your hint now readelf is a separate module and not included in wrappers. 

How about this organization: 
`pwndbg/pwndbg/wrappers/readelf.py`, then a new wrapper will be `pwndbg/pwndbg/wrappers/checksec.py` etc...

@disconnect3d  @zachriggle  @anthraxx  feedbacks? :smiley:

disconnect3d: I've moved readelf to seperate module as it is kinda big tool that lets you do different things based on input.

In comparison checksec has like two(?) modes - one to check all the running apps and second to check particular file.

disconnect3d: Anyway, seperate files/submodules for wrappers are fine.

degrigis: Got it. And what do you think about my proposal?

PR #261, opened 2017-05-04
typeinfo: find useful system header files by finding the maximum fixed point of compilable header files

MaskRay:
The trial and errors process is quite slow, it would be better to cache the list of header files somewhere. Do you where is appropriate for storing the list?

Tested on Arch Linux x86_64, extra/gdb-7.12.1-3 (Python 3)

Hi Zach!

### DIFF
--- a/pwndbg/__init__.py
+++ b/pwndbg/__init__.py
@@ -65,6 +65,7 @@ import pwndbg.prompt
 import pwndbg.regs
 import pwndbg.stack
 import pwndbg.stdio
+import pwndbg.storage
 import pwndbg.typeinfo
 import pwndbg.version
 import pwndbg.vmmap
--- a/pwndbg/storage.py
+++ b/pwndbg/storage.py
@@ -1,0 +1,130 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import functools
+import os
+import pickle
+import re
+import sys
+
+import six
+
+import pwndbg.compat
+
+def get_cache_dir():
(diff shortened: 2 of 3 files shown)


zachriggle:
OMG ❤️

zachriggle:
There's not really a persistence mechanism for Pwndbg yet.  We haven't needed inter-session storage so far.

MaskRay:
Do you have a plan to add the persistence layer? I'll change the commit accordingly.

zachriggle:
Unfortunately, there's none currently.  I'm open to suggestions :)

disconnect3d:
@MaskRay Are you going to work on that too?

Here are some options for persistence directory:
* `~/.cache/pwndbg`
* `~/.local/share/pwndbg`
* `~/.config/pwndbg`
* `~/.pwndbg`

Probably would be good to check for an env var first, e.g.:
```python
persistence_path = os.environ.get('PWNDBG_PERSISTENCE_DIR', default_persistence_dir)
```

It might be good to handle it through `memoize`, e.g. adding a decorator called `forever` or `persistent`.
Another option could be an explicit `save`/`load`.

MaskRay:
Refined and rebased on https://github.com/pwndbg/pwndbg/pull/268

disconnect3d:
Can we finally make it into pwndbg? If we are not really having a proper persistance mechanism, maybe we should skip it for this PR?
@zachriggle @MaskRay

disconnect3d:
Hmm played a bit with this and here are some thoughts:
- this wouldn't be usable without cache
- maybe it would be better to add include paths and to have all includes and remove those that are still missing? (collisions might be problematic)
- maybe we should specify some 'standard include paths' for different OS/architectures?

ps: this needs to be rebased

disconnect3d:
Closing as there is no reply :(

PR #263, opened 2017-05-05
Fix git fatal msg display when pwndbg is installed w/o git

disconnect3d:
My bad. I haven't checked how pwndbg behaves without git when I added version command.

Before this commit:
```
$ gdb
fatal: Not a git repository: '/home/dc/installed/pwndbg/.git'
pwndbg> 
```

After this commit:
```
$ gdb
pwndbg> 
```

### DIFF
--- a/pwndbg/version.py
+++ b/pwndbg/version.py
@@ -15,12 +15,15 @@ def build_id():
     """
     try:
         git_path = os.path.join(os.path.dirname(os.path.dirname(__file__)), '.git')
-        cmd = ['git', '--git-dir', git_path, 'rev-parse', 'HEAD']
+        cmd = ['git', '--git-dir', git_path, 'rev-parse', '--short', 'HEAD']
 
         commit_id = subprocess.check_output(cmd, stderr=subprocess.STDOUT)
 
-        return 'build: %s' % commit_id[:8].decode('utf-8')
-    except:
+        return 'build: %s' % commit_id.decode('utf-8').strip('\n')
+
+    except (OSError, subprocess.CalledProcessError):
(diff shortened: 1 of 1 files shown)


pwndbg/version.py
```diff
@@ -15,10 +15,11 @@ def build_id():
     """
     try:
         git_path = os.path.join(os.path.dirname(os.path.dirname(__file__)), '.git')
-        commit_id = subprocess.check_output(['git', '--git-dir', git_path, 'rev-parse', 'HEAD'])
-        commit_id = commit_id[:8].decode('utf-8')
+        cmd = ['git', '--git-dir', git_path, 'rev-parse', 'HEAD']
```
zachriggle: Use `--short` instead of truncating to 8 characters

pwndbg/version.py
```diff
@@ -15,10 +15,11 @@ def build_id():
...
 
-        return 'build: %s' % commit_id
+        commit_id = subprocess.check_output(cmd, stderr=subprocess.STDOUT)
+
+        return 'build: %s' % commit_id[:8].decode('utf-8')
     except:
```
zachriggle: Catch the specific type of Exception raised when the command fails, so we don't silently mask other issues

pwndbg/version.py
zachriggle: Use `--short` instead of truncating to 8 characters

zachriggle (changes requested):

disconnect3d:
@zachriggle Fixed.

zachriggle (approved):

PR #266, opened 2017-05-08
Ehdr phdr commands

disconnect3d:
1. Adds `ehdr` and `phdr` commands and `__str__` for `ElfXX_Ehdr` and `ElfXX_Phdr`

Those might be handy for debugging issues with endianness (https://github.com/pwndbg/pwndbg/issues/156 or https://github.com/pwndbg/pwndbg/pull/239) or for debugging new commands that would have to use ehdr/phdr.

image

If the constant is not found:
image

2. Adds `Elf32_Dyn` and `Elf64_Dyn` structures and `elftypes.print_struct` function (which might be extended in the future to pretty print other ctypes fields).

This will be needed in the future for a command to dump process into Elf (e.g. `dumpelf`).

3. Refactors elftypes a bit

### DIFF
--- a/pwndbg/commands/elf.py
+++ b/pwndbg/commands/elf.py
@@ -37,6 +37,25 @@ def elfheader():
 
 
 @pwndbg.commands.Command
+@pwndbg.commands.OnlyWhenRunning
+def ehdr():
+    """
+    Prints ELF header structure.
+    """
+    print(pwndbg.elf.exe())
+
+
+@pwndbg.commands.Command
+@pwndbg.commands.OnlyWhenRunning
+def phdr():
+    """
+    Prints ELF Program Headers structures.
+    """
+    for phdr in pwndbg.elf.iter_phdrs(pwndbg.elf.exe()):
+        print(phdr)
+
+
+@pwndbg.commands.Command
 @pwndbg.commands.OnlyWithFile
 def gotplt():
     """
--- a/pwndbg/elftypes.py
+++ b/pwndbg/elftypes.py
@@ -35,11 +35,7 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import ctypes
-import sys
 
-import six
-
-import pwndbg.arch
 import pwndbg.ctypes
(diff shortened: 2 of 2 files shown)


zachriggle:
Can you make these be pretty-printers for the specified types?

https://sourceware.org/gdb/onlinedocs/gdb/Writing-a-Pretty_002dPrinter.html

We can then manually load the type information if needed via compiling a C file.

The commands can just be a shortcut to invoke the pretty-printer on the EXE's EHDR and PDHRs

zachriggle:
Also: Might be worthwhile converting your ~= generic code into generic code to generate a pretty-printer instead

disconnect3d:
> We can then manually load the type information if needed via compiling a C file.

Do you mean making a file like:
```c
#include <elf.h>
int main() {
    Elf32_Ehdr ehdr;
}
```

And obtaining the structure info from that somehow?

Can we generate/get ctypes structure from that? 

This is needed for a process dump command. Being able to point to any buffer offset and cast it to any type is pretty cool (and ctypes lets us do it - see https://github.com/disconnect3d/pwndbg/blob/dumpproc/pwndbg/dumpelf.py and https://github.com/disconnect3d/pwndbg/blob/dumpproc/pwndbg/ctypes.py#L24-L31).
This is till work in progress, a bit inspired by https://github.com/whatsbcn/skpd/

zachriggle:
> Can we generate/get ctypes structure from that?

That's one of the things that's on a back-burner, and part of the inspiration for `pwndbg.ctypes` existing.  Eventually I'd like to not use GDB's type and symbol system, so that we can inject our own stuff.

disconnect3d:
Another possibility would be using `cffi` instead of `ctypes`: http://eli.thegreenplace.net/2013/03/09/python-ffi-with-ctypes-and-cffi/

That would let us to parse header files and just grab structures from them (but might have its own quirks, no idea yet).

disconnect3d:
Closing as it is not that useful and can be inspected statically as well.

PR #267, opened 2017-05-08
Fix ehdr/phdr endianness not being switched properly

### LINKED ISSUE #156: Exception with ELF MIPSII

disconnect3d:
Also fixes #156.

An explanation of the issue can be found in #202 and #239, as well as discussion/thoughts.

Here it is in action:
https://asciinema.org/a/35li7yv0qleb7xdj57svoelzm

### DIFF
--- a/pwndbg/arch.py
+++ b/pwndbg/arch.py
@@ -5,7 +5,6 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-import collections
 import struct
 import sys
 
@@ -25,12 +24,14 @@ ptrsize = pwndbg.typeinfo.ptrsize
 fmt     = '=I'
 native_endian = str(sys.byteorder)
 
+
 def fix_arch(arch):
     arches = ['x86-64', 'i386', 'mips', 'powerpc', 'sparc', 'arm', 'aarch64', arch]
     return next(a for a in arches if a in arch)
 
+
 @pwndbg.events.start
-@pwndbg.events.stop
+@pwndbg.events.new_objfile
 def update():
     m = sys.modules[__name__]
 
@@ -67,14 +68,18 @@ def update():
     else:
         m.qemu = m.current
 
+
 def pack(integer):
     return struct.pack(fmt, integer & ptrmask)
+
 
 def unpack(data):
     return struct.unpack(fmt, data)[0]
 
+
 def signed(integer):
     return unpack(pack(integer), signed=True)
+
 
 def unsigned(integer):
     return unpack(pack(integer))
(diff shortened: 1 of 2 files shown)


pwndbg/arch.py
```diff
@@ -30,6 +30,7 @@ def fix_arch(arch):
     return next(a for a in arches if a in arch)
 
 @pwndbg.events.start
+@pwndbg.events.new_objfile
```
disconnect3d: We weren't reloading `pwndbg.arch.endian` when loading new object file (e.g. when connecting to remote target).

disconnect3d: Hm.. isn't `@pwndbg.events.stop` redudant here?

pwndbg/elf.py
```diff
@@ -30,16 +26,21 @@
 import pwndbg.proc
 import pwndbg.stack
 
+from imp import reload
```
disconnect3d: Not sure if that is the proper way for cross Python compatibility. [An online cheatsheet](http://python-future.org/compatible_idioms.html#reload) suggests it.

However, in our own reload command, we use [reload from either `builtins` or from `imp`](https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/commands/reload.py#L20-L23). Is there any reason to do so?

zachriggle:
Import reload from six.moves

disconnect3d:
I've imported `reload_module` from `six.moves` and removed `@pwndbg.events.stop` from `pwndbg.arch.update`.

disconnect3d:
@zachriggle any comments or merge? :P

zachriggle:
Ping me in 12 hours. I had planned to fix this myself but clearly I'm
slacking and should just merge this.

disconnect3d:
@zachriggle ping

zachriggle (approved):

Issue #264, opened 2017-05-06
Bare metal debugging


RageYL:
Hi,

When debugging on a bare metal mcu (arm, cortex-m4 processor) pwndbg 'hang' immediately after the connection with the remote gdbserver has been established

1) AUXV are not present: pwndbg tries to find those and will end up searching the whole memory (walk_stack2)
2) $xpsr is used instead of $cpsr. it uses a different bit to indicate thumb mode ($ps is not available on this arch)
xpsr: http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0413d/ch02s02s02.html
3) find_upper/lower_boundary is slow (each memory dereference is slow on remote): in my case max_pages is reached, it takes 30 seconds to display registers...

idea for 1) 2) 3):
the input binary contains lots of information that could be very useful:
- immediately available (do not look for them in the memory, eg: entry point)
- disable unneeded operations (no auxv with a cortex-m4, readelf -A)

idea 2)
it would be nice to display context without crashing even if a register is not found

disconnect3d:
@RageYL Can you look if we can determine somehow whether we are debugging bare metal or not?

Maybe there is some kind of `info ...` command that gives us such information after connecting to remote target?

RageYL:
I already (very briefly) looked and didn't find any command, that's why i suggested parsing the binary as i could get the information with readelf. 
But i'll take a more serious look and keep you updated ;) (should be done by the end of the week)

disconnect3d:
@RageYL Can you check `show osabi` after connecting to remote target?
Someone mentioned this on #gdb IRC channel on freenode.

RageYL:
@disconnect3d show osabi displays: The current OS ABI is "auto" (currently "none")

disconnect3d:
So maybe this is how we can determine debugging bare metal:
```python
is_bare_metal = '(currently "none")' in gdb.execute('show osabi', to_string=True)
```

For me (without debugging bare metal) it is:
```
$ gdb
Loaded 106 commands. Type pwndbg [filter] for a list.
pwndbg> show osabi
The current OS ABI is "auto" (currently "GNU/Linux").
The default OS ABI is "GNU/Linux".
```

This is also returned when debugging a MIPS binary remotely (launched as `qemu-mips -g 1234`).

RageYL:
[...]

PR #268, opened 2017-05-09
Add pwndbg.storage and four caching API file_cache_{,with_signature_}for_{text,pickle}

MaskRay:
Add pwndbg.storage and four caching API file_cache_{,with_signature_}for_{text,pickle}

Consider an imaginary file pwndbg/foo.py

```python
@pwndbg.storage.file_cache_for_text('{0}.txt')
def ft(x):
    print(x)
    return x
```

```
@pwndbg.storage.file_cache_for_text('c/{0}-{1}.txt')
def fx(x, y): return x
# fx('a', 'b') will save to ~/.cache/pwndbg/foo/ft/c/a-b.txt
```

The first call of `ft('a')` will save a file in `~/.cache/pwndbg/foo/ft/a.txt` .
A later call of `ft('a')` will load the text from the cache.

```python
@pwndbg.storage.file_cache_with_signature_for_text('{0}')
def fts(x):
    print(x)
    return x
```

The caller should provide an additional argument (signature) to fts.
The first call of `fts('a', signature='sig')` will save a file in `~/.cache/pwndbg/foo/fts/a` .
A call of fts('a', signature='sig') will load the cache.
However, a call of fts('a', signature='different_sig') will recalculate as the signature is different.

### DIFF
--- a/pwndbg/storage.py
+++ b/pwndbg/storage.py
@@ -1,0 +1,137 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import functools
+import os
+import pickle
+import re
+import sys
+
+import past.builtins
+
+import pwndbg.compat
+
+def get_cache_dir():
+    dir = os.environ.get('PWNDBG_CACHE_DIR')
+    if not dir:
+        if sys.platform == 'darwin':
+            dir = os.path.join(os.path.expanduser('~/Library/Caches'), 'pwndbg')
+        elif sys.platform == 'win32':
+            dir = os.path.join(os.getenv('LOCALAPPDATA'), 'pwndbg', 'Cache')
+        else:
+            dir = os.path.join(os.environ.get('XDG_CACHE_HOME') or os.path.expanduser('~/.cache'), 'pwndbg')
+    try:
+        os.makedirs(dir)
+    except OSError:
+        assert os.path.isdir(dir)
+    return dir
+
+cache_dir = get_cache_dir()
+
+def makedirs(name):
+    try:
+        os.makedirs(name)
+    except OSError:  # FileExistsError.__base__ is OSError in Python 3
+        assert os.path.isdir(name)
+
+def generic_file_cache(template, fn):
+    def memoize(function):
+        p = function.__module__.split('.')[1:] + [function.__name__]
+        dirpath = os.path.join(cache_dir, *p)
+
+        @functools.wraps(function)
+        def wrapper(*args, **kwargs):
+            if isinstance(template, past.builtins.basestring):
+                path = template.format(*args, **kwargs)
+            else:
+                kwargs1 = kwargs.copy()
+                kwargs1.pop('signature', None)
+                path = template(*args, **kwargs1)
+            path = os.path.join(dirpath, path)
+            assert os.path.normpath(path).startswith(dirpath+'/'), 'Please defend path traversal in the template'
+            return fn(path, function, args, kwargs)
+
+        return wrapper
+
+    return memoize
+
+def file_cache_for_pickle(template):
+    def fn(path, function, args, kwargs):
+        try:
+            with open(path, 'rb') as f:
+                ret = pickle.load(f)
+        except:
+            ret = function(*args, **kwargs)
+            makedirs(os.path.dirname(path))
+            with open(path, 'wb') as f:
+                pickle.dump(ret, f)
+        return ret
+
+    return generic_file_cache(template, fn)
+
+def file_cache_with_signature_for_pickle(template):
+    def fn(path, function, args, kwargs):
+        valid = False
+        signature = kwargs.pop('signature')
+        try:
+            with open(path, 'rb') as f:
+                (saved_signature, ret) = pickle.load(f)
+                valid = saved_signature == signature
+        except:
+            pass
+        if not valid:
+            ret = function(*args, **kwargs)
+            makedirs(os.path.dirname(path))
+            with open(path, 'wb') as f:
+                pickle.dump((signature, ret), f)
+        return ret
+
+    return generic_file_cache(template, fn)
+
+def file_cache_for_text(template):
+    def fn(path, function, args, kwargs):
+        try:
+            with open(path) as f:
+                ret = f.read()
+        except:
+            ret = function(*args, **kwargs)
+            assert isinstance(ret, past.builtins.basestring)
+            makedirs(os.path.dirname(path))
+            with open(path, 'w') as f:
+                f.write(ret)
+        return ret
+
+    return generic_file_cache(template, fn)
+
+def file_cache_with_signature_for_text(template):
+    regex = re.compile(r'pwndbg:\s*signature=(\S+)')
+
+    def fn(path, function, args, kwargs):
+        valid = False
+        signature = str(kwargs.pop('signature'))
+        try:
+            with open(path) as f:
+                lines = f.readlines()
+            if lines:
+                match = regex.search(lines[0])
+                if match:
+                    saved_signature = match.group(1)
+                    if saved_signature == signature:
+                        ret = ''.join(lines[1:])
+                        valid = True
+        except:
+            pass
+        if not valid:
+            ret = function(*args, **kwargs)
+            assert isinstance(ret, past.builtins.basestring)
+            makedirs(os.path.dirname(path))
+            with open(path, 'w') as f:
+                f.write('# -*- pwndbg: signature={} -*-\n'.format(signature))
+                f.write(ret)
+        return ret
+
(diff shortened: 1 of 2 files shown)


pwndbg/storage.py
```diff
@@ -0,0 +1,41 @@
...
+def get_cache_dir():
+    dir = os.environ.get('PWNDBG_CACHE_DIR')
+    if not dir:
+        if sys.platform == 'darwin':
+            dir = os.path.join(os.path.expanduser('~/Library/Caches'), 'pwndbg')
+        elif sys.platform == 'win32':
+            dir = os.path.join(os.getenv('LOCALAPPDATA'), 'pwndbg', 'Cache')
+        else:
+            dir = os.path.join(os.environ.get('XDG_CACHE_HOME') or os.path.expanduser('~/.cache'), 'pwndbg')
+    if sys.version_info[0] >= 3:
```
disconnect3d: There is `pwndbg.compat.python3` for that :).

disconnect3d: Also if there isn't a better way for `os.makedirs` with `exist_ok=True` kwarg, I would just leave the py2 implementation (as it will work in py3).

pwndbg/storage.py
```diff
@@ -0,0 +1,41 @@
...
+import os
+import sys
+
+def get_cache_dir():
+    dir = os.environ.get('PWNDBG_CACHE_DIR')
+    if not dir:
+        if sys.platform == 'darwin':
+            dir = os.path.join(os.path.expanduser('~/Library/Caches'), 'pwndbg')
+        elif sys.platform == 'win32':
+            dir = os.path.join(os.getenv('LOCALAPPDATA'), 'pwndbg', 'Cache')
```
disconnect3d, line 24: Is this even possible to launch pwndbg on Windows?

pwndbg/storage.py
MaskRay: Thanks. Removed py2 stuff.

pwndbg/storage.py
MaskRay: I also wonder whether it works on Windows...

disconnect3d: We are tied to Elfs and AUXV, so I am almost sure that no.

The AUXV stuff is the reason why pwndbg isn't working on a bare metal currently (see https://github.com/pwndbg/pwndbg/issues/264).

MaskRay: I can remove the two lines for Windows. Do you have concerns regarding to other parts?

disconnect3d: We probably should add a mechanism to memoize or save/load something in cache in this PR.

Just a note: it would be nice to have an assertion to error on saving two things into same file, not to have weird issues in the future (e.g. assert when we register function to be cached pernamently under already registered key/filename).
This might be tricky considering the fact we might have code that reloads modules (see: https://github.com/pwndbg/pwndbg/pull/267).

Hm.. if it is going to be a decorator it could be implemented so that we would take import path + function name as cache key (or rather file name) - this won't have duplicates.

MaskRay: For the https://github.com/pwndbg/pwndbg/pull/261 case, the arguments of the function also decide the name of the cache key. I am not very clear on how to implement this.

disconnect3d:
The decorator can get optional argument which would be a function getting
args and kwargs and returning cache key.

09.05.2017 3:32 AM "Fangrui Song" <notifications@github.com> napisał(a):

MaskRay:
@disconnect3d Refactored.

pwndbg/storage.py
```diff
@@ -0,0 +1,128 @@
...
+                if saved_signature == signature:
+                    valid = True
+        except:
+            pass
+        if not valid:
+            ret = function(*args)
+            makedirs(os.path.dirname(path))
+            with open(path, 'wb') as f:
+                pickle.dump((signature, ret), f)
+
```
disconnect3d: We are missing `return ret` here

pwndbg/storage.py
```diff
@@ -0,0 +1,128 @@
...
+                    ret = ''.join(lines[i:])
+                    break
+                saved_signature = match.group(1)
+                if saved_signature == signature:
+                    valid = True
+        except:
+            pass
+        if not valid:
+            ret = function(*args)
+            assert isinstance(ret, str)
```
disconnect3d: Probably better to have `isinstance(ret, six.string_types)` here - this would let us cache both binary/ascii strings and unicode one (however I am not sure if below file writes pass with both types).

pwndbg/storage.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def fn(path, function, args):
+        valid = False
+        signature = str(args[-1])
+        try:
+            with open(path) as f:
+                lines = f.readlines()
+            ret = ''
+            for i in range(len(lines)):
+                match = regex.search(lines[i])
```
disconnect3d: Shouldn't we check for signature only in the first line?

disconnect3d (changes requested):

disconnect3d:
> The caller should provide an additional argument (signature) to fts.
The first call of fts('a', 'sig') will save a file in ~/.cache/pwndbg/foo/fts/a .
A call of fts('a', 'sig') will load the cache.
However, a call of fts('a', 'different_sig') will recalculate as the signature is different.

Do we have any use case for the `signature` decorator now?

I would rather have `signature` passed as kwarg to the decorator function and not to the function I am decorating.

pwndbg/storage.py
MaskRay: Done

pwndbg/storage.py
MaskRay: Done

pwndbg/storage.py
MaskRay: Done. Yes, you can write a `unicode` to `open(..., 'w')` in Python 2.

MaskRay:
@disconnect3d I'd like to use the `signature` decorator for `find_compilable_headers` in https://github.com/pwndbg/pwndbg/pull/261 . It is a waste of time to find compilable system header files each time pwndbg starts. Thus the code serializes the list of header files somewhere and recalculate whenever the list of system header files changes (which is unlikely to happen if you do not install new packages).

zachriggle:
One of the considerations is that the data would need to be cached
per-architecture. The list of headers that works on amd64 may be different
from arm, for example.

pwndbg/storage.py
```diff
@@ -0,0 +1,130 @@
...
+
+def file_cache_with_signature_for_pickle(template):
+    def fn(path, function, args, kwargs):
+        valid = False
+        signature = kwargs.pop('signature')
+        try:
+            with open(path, 'rb') as f:
+                (saved_signature, ret) = pickle.load(f)
+                if saved_signature == signature:
+                    valid = True
```
disconnect3d: Just `valid = saved_signature == signature`

pwndbg/storage.py
```diff
@@ -0,0 +1,130 @@
...
+        signature = str(kwargs.pop('signature'))
+        try:
+            with open(path) as f:
+                lines = f.readlines()
+            if lines:
+                match = regex.search(lines[0])
+                if match:
+                    saved_signature = match.group(1)
+                    if saved_signature == signature:
+                        ret = ''.join(lines[1:])
```
disconnect3d, line 124: Those are just lines from the cached return, right? If so, we should probably join it using `'\n'` string, so the newlines are back there.

pwndbg/storage.py
zachriggle: ^ Meh

disconnect3d: Yeah, the condition is more explicit/visible

zachriggle:
One of the issues with the current implementation is what happens when `f(a, b)` is cached and `a=="../../../../../etc/passwd"` and similar.

pwndbg/storage.py
MaskRay: `readlines` does not strip the newlines

MaskRay:
Added a simple foolproof check.

Changed to 
```python
        @functools.wraps(function)
        def wrapper(*args, **kwargs):
            if isinstance(template, past.builtins.basestring):
                path = template.format(*args, **kwargs)
            else:
                kwargs1 = kwargs.copy()
                kwargs1.pop('signature', None)
                path = template(*args, **kwargs1)
            path = os.path.join(dirpath, path)
            assert os.path.normpath(path).startswith(dirpath+'/'), 'Please defend path traversal in the template'
            return fn(path, function, args, kwargs)
```

disconnect3d (approved):

disconnect3d:
Hmm... is that ready? Do we still miss the things pointed out by @zachriggle ?
Could rebase it to latest dev @MaskRay?

MaskRay:
one year passed and now I'm heading for another Defcon.. should take some
time to re-learn pending..

disconnect3d:
Closing this as it is too old :(

Issue #244, opened 2017-04-23
Use ArgparsedCommand for everything


zachriggle:
Using ArgparsedCommand means that we can generate documentation trivially, and that we also can have a standardized mechanism for parsing which is consistent with Python's own things.

It also gives us a nice place to hook looking up symbols and values, by using the `ArgumentParser.add_argument` keyword `type=`, which takes an arbitrary callable, and delivers the string argument, and blindly uses the return value.

We can add a `type=gdb_int` which takes plain integers *or* evaluates a GDB expression. Same for `type=address` or similar.  This avoids a lot of ad-hoc parsing and calls to `int()` on things.

disconnect3d:
By the way, moving to `ArgparsedCommand` makes docstrings useless, e.g.:

```python
@pwndbg.commands.ArgparsedCommand(argparse.ArgumentParser(description='ASDASDASD'))
@pwndbg.commands.OnlyWithFile
def checksec(file=None):
    '''
    Prints out the binary security settings. Attempts to call the binjitsu
    checksec first, and then falls back to checksec.sh.
    '''
```

Pwndbg session:
```
pwndbg> help checksec
ASDASDASD
pwndbg> pwndbg checksec
checksec             ASDASDASD
pwndbg> 
```

disconnect3d:
FWIW we have only two commands that DO NOT use ArgparsedCommand right now:
```
user@user:~/pwndbg$ rg commands.Command
pwndbg/commands/defcon.py
104:@pwndbg.commands.Command

pwndbg/commands/shell.py
84:    pwndbg.commands.Command(handler, False)
```

The first one in defcon.py will be removed in https://github.com/pwndbg/pwndbg/pull/1203

The second should be refactoreg to ArgparsedCommand.

Then, we can remove `commands/__init__.py:Command` class and merge whatever it does with `_ArpgarsedCommand` or `ArgparsedCommand`.

gsingh93:
Assigned to @disconnect3d to close this and make a new issue for any remaining work.

gsingh93:
ping 🙂

disconnect3d:
Closing this issue as its really _stale_ at this point.

I have created https://github.com/pwndbg/pwndbg/issues/1822 and  https://github.com/pwndbg/pwndbg/issues/1820

(fyi @gsingh93)

Issue #255, opened 2017-04-29
dt is broken


zachriggle:
:(

EDIT: see last post in this issue

semihalf-oleksy-michalina:
Hello,
Can you provide some example where it doesn't work and all the examples and options you wish to implement in this command? I'm not sure if I assume it's purpose correctly.
It would be nice if you could provide more info than it is in help as I don't have windbg.

I've changed a little three places and now when I have a structure like this:

```C
typedef struct str_type {
	int num;
	char letter;
	int* ptr;
}str_type;

```

`dt str_type` prints:

```
pwndbg> dt str_type 
str_type
    +0x0000 num                  : int
    +0x0004 letter               : char
    +0x0008 ptr                  : int *
```

But I assume that it's not everything you'd like to achieve?

disconnect3d:
@semihalf-oleksy-michalina Can you post a PR with your changes so we can try it out?
Also could you change the command to use `ArgparsedCommand` decorator like some of the other commands? Eventually we want all of the command to use it and remove `Command` or `ParsedCommand` decorators at all (and to provide better UX thanks to one, consistent interface, later on - see https://github.com/pwndbg/pwndbg/issues/244).

disconnect3d:
@zachriggle Is the purpose of `dt` just describe particular type layout?
If so, we will have to show the padding to the user as well.

Also it might be good to check pahole and pahole-gdb projects/outputs - e.g. https://github.com/PhilArmstrong/pahole-gdb
However I feel like we should display padding by just saying `X bytes of padding`.

It might also be cool to describe what padding is in the `ArgparsedCommand` description and also have a `short_description` (if there is such a thing for ArgparsedCommands).

zachriggle:
`dt` is intended to emulate [Windbg's command of the same name](https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/dt--display-type-) -- currently only the very basic functionality of "display type" and "display type at address" are supported.

The current implementation, and the implementation used by Windbg, does not explicitly call out gaps / holes / padding.

If we'd like to add additional functionality, I'm all for it, but it should have a different command name if the functionality isn't present in Windbg.

disconnect3d:
Btw https://sourceware.org/ml/gdb-patches/2017-11/msg00435.html
Still our approach is probably better as we support older gdb versions.

disconnect3d:
[...]

PR #274, opened 2017-05-10
WIP: Argparsed commands

disconnect3d:
Refactorization suggested in #244 

This is still **Work In Progress**.

Comments/reviews are welcome.


pwndbg/commands/argv.py
```diff
@@ -13,43 +15,34 @@
 import pwndbg.typeinfo
 
 
-@pwndbg.commands.Command
+@pwndbg.commands.ArgparsedCommand(argparse.ArgumentParser(description='Prints out the number of arguments.'))
```
zachriggle: I'd greatly prefer to break down everything onto its own line, **or** to do pass-through on `kwargs`

zachriggle: I'd greatly prefer to break down everything onto its own line, **or** to do pass-through on `kwargs`

zachriggle (changes requested):

pwndbg/commands/argv.py
disconnect3d: Which version?

1.
```python
    argparse.ArgumentParser(description='Prints out the number of arguments.')
)
def argc():
    print(pwndbg.argv.argc)
```

2.
```python
parser = argparse.ArgumentParser(description='Prints out the number of arguments.')
def argc():
    print(pwndbg.argv.argc)
```

3.
```python
parser = argparse.ArgumentParser(
    description='Prints out the number of arguments.'
)
def argc():
    print(pwndbg.argv.argc)
```


4.
```python
@pwndbg.commands.ArgparsedCommand(argparse.ArgumentParser(
    description='Prints out the number of arguments.'
))
def argc():
    print(pwndbg.argv.argc)
```

disconnect3d: Or maybe we should let `ArgparsedCommand` get either `ArgumentParser` instance or `str`.

If `str` would be passed, it would create `parser` on its own passing the string as `description`.
Kinda magic, but less writing for commands without args.

What do you think?

zachriggle:
3 or 4, depending on the length of the string.

zachriggle:
For commands without arms, that might also work

disconnect3d:
@zachriggle I have added an `AliasCommand`. It reuses the parser created in one `ArgparsedCommand` and registers a command that maps to the `alias_to` function.

Here it is in action:
```gdb
$ gdb -q /bin/ls
Loaded 107 commands. Type pwndbg [filter] for a list.
pwndbg> version
Gdb: GNU gdb (GDB) 7.12.1
Pwndbg: 1.0.0 build: 880f526
pwndbg> args
The program is not being run.
pwndbg> argv
The program is not being run.
pwndbg> set context-sections ''
Set which context sections are displayed by default (also controls order) to ''
pwndbg> entry
Temporary breakpoint 1 at 0x404030

Temporary breakpoint 1, 0x0000000000404030 in ?? ()
Breakpoint *0x404030
pwndbg> args
00:0000│   0x7fffffffd868 —▸ 0x7fffffffdbfe ◂— '/usr/bin/ls'
01:0008│   0x7fffffffd870 ◂— 0x0
pwndbg> argv
00:0000│   0x7fffffffd868 —▸ 0x7fffffffdbfe ◂— '/usr/bin/ls'
01:0008│   0x7fffffffd870 ◂— 0x0
pwndbg> 
```

disconnect3d:
commands/asm.py is empty, should I delete it?

zachriggle:
Yep! Feel free

pwndbg/commands/config.py
```diff
@@ -50,15 +52,25 @@ def config():
                        '- then put it in your .gdbinit after initializing pwndbg'))
 
 
-@pwndbg.commands.Command
-def configfile(show_all=False):
-    """Generates a configuration file for the current Pwndbg options"""
+parser = argparse.ArgumentParser(
+    description='Generates a configuration file for the current Pwndbg options'
+)
+parser.add_argument('show_all', nargs='?', default=False, type=bool, help='Whether to show all config params')
```
zachriggle: Might be worth changing this to `--all` since it's really an optional flag (versus a positional)

pwndbg/commands/argv.py
```diff
@@ -58,7 +58,7 @@ def envp():
 
 @pwndbg.commands.ArgparsedCommand(parser)
 @pwndbg.commands.OnlyWhenRunning
-def env(name=None):
+def env(name):
```
zachriggle: This breaks invocation from the Python interpreter.

disconnect3d:
```
-def env(name=None):
+def env(name):
```

@zachriggle about this - as you said it breaks invocation from Py interpreter - so I will revert it.
I will also probably add some note about adding new commands that one should add default argument to the command as well as to `argparse.ArgumentParser`.

disconnect3d:
Some thoughts/recording not to forget come back to some stuff.

I have added an additional keyword argument to `ArgparsedCommand` - `unpack` to let us unpack arguments from one argparse field into the funtion, e.g.:
```python
parser = argparse.ArgumentParser('Print out all registers and enhance the information.')
parser.add_argument('regs', nargs='*', type=str, default='', help='List of registers')


@pwndbg.commands.ArgparsedCommand(parser, unpack='regs')
def regs(*regs):
    print('\n'.join(get_regs(*regs)))
```

As the `AliasCommand` also has to support it, I have extended it to save `self.unpack` field into `function.unpack`.

SKIPPED:
* commands/defcon.py
* commands/dt.py - as it is not working now (see: https://github.com/pwndbg/pwndbg/issues/255)
* commands/dumpargs.py
* commands/heap.py - either heap stuff will be merged first or this stuff, one of PRs will be have to be changed ;p
* commands/nearpc.py - guess it requires gdb.Value, so waiting with it

Thoughts - @zachriggle comment pls:
1. `env`, `environ` commands crash when one pass wrong environment variable name

2. `context` - since we have settings we might revert `context_args` being commented out.

3. `cpsr` is available only on arm, this should be handled through a decorator or `ArgparsedCommand` argument or sth.

4. `defcon_heap` probably should be moved into some `commands/specific/` which should be enabled optionally in settings?

5. `commands/dumpargs.py` looks like a candidate OR reimplementation to invoke proper function that is used to display `call` arguments inline.

6. `commands/gbinit.py` should be reviewed more carefully (e.g. it removes some of descriptions).

7. `errno` command hasn't been working since I guess addition of `pwndbg` command:
```
pwndbg> errno
  File "/home/dc/pwndbg/pwndbg/commands/__init__.py", line 56, in __call__
  File "/home/dc/pwndbg/pwndbg/commands/__init__.py", line 182, in _ArgparsedCommand
    return function(**vars(args))
  File "/home/dc/pwndbg/pwndbg/commands/__init__.py", line 130, in _OnlyWhenRunning
    return function(*a, **kw)
  File "/home/dc/pwndbg/pwndbg/commands/misc.py", line 31, in errno
    errno_location = pwndbg.symbol.get('__errno_location')
AttributeError: 'C' object has no attribute 'symbol'
```

Simple change of `pwndbg` -> `_pwndbg` isn't fixing it:
```
pwndbg> errno
'<' not supported between instances of 'str' and 'int'
'errno': Converts errno (or argument) to its string representation.
```

8. What about arguments that needs gdb.Value as argument? `type='str'` isn't working well for `distance`, same for `int`:
```
parser = argparse.ArgumentParser(description='Print the distance between the two arguments')
parser.add_argument('a', type=str, help='GDB expression')
parser.add_argument('b', type=str, help='GDB expression')


@_pwndbg.commands.ArgparsedCommand(parser)
def distance(a, b):
```

with `type=int`:
```
pwndbg> distance 123 $rip
  File "/home/dc/installed/pwndbg/pwndbg/commands/__init__.py", line 95, in __call__
  File "/home/dc/installed/pwndbg/pwndbg/commands/__init__.py", line 236, in _ArgparsedCommand
    return function(*args, **kwargs)
  File "/home/dc/installed/pwndbg/pwndbg/commands/misc.py", line 79, in distance
    b = int(b) & _arch.ptrmask
  File "/home/dc/installed/pwndbg/pwndbg/inthook.py", line 53, in __new__
ValueError: invalid literal for int() with base 10: '$rip'
```

with `type=str`:
```
pwndbg> distance 321 $rip
usage: distance [-h] a b
distance: error: argument b: invalid fix_int value: '$rip'
```

zachriggle:
I don't really like the `AliasCommand` decorator on a function that's effectively ignored.  It's cute, but I'd rather just invoke e.g. `AliasCommand(pwndbg.commands.vmmap.vmmap, "libs")`

disconnect3d:
Hm.. there is an interesting WIP/proposition in https://github.com/pwndbg/pwndbg/pull/275

I am stopping with further work until we decide for commands base/layout.

disconnect3d:
Tbh this tried to change way too much. I will try to make smaller PRs.

PR #275, opened 2017-05-13
WIP: Class-based commands

MaskRay:


pwndbg/commands/__init__.py
```diff
@@ -218,3 +221,139 @@ def _ArgparsedCommand(*args):
             return function(**vars(args))
         _ArgparsedCommand.__doc__ = self.parser.description
         return Command(_ArgparsedCommand)
+
+## TODO(pick an appropriate filename for these) Common types for argparse.add_argument(type=...), which will also be used for command completion
+
+class ArgparseType(object):
```
disconnect3d: Probably should have `abc.ABCMeta` metaclass.
http://stackoverflow.com/questions/35673474/using-abc-abcmeta-in-a-way-it-is-compatible-both-with-python-2-7-and-python-3-5

disconnect3d (commented):
Lets wait for @zachriggle comments on that.

Some summary about current approach:
* commands are implemented as classes that must provide `do_invoke` method
* as argparse returns kwargs, there is no way to create command that gets `*args`;  [this could be a workaround](https://github.com/pwndbg/pwndbg/pull/274/files#diff-57bee9a0cb369f1166ef59075fbcdf5cR234), however I don't like it that much..
* launching command from python interpreter is a bit painful: `Command().do_invoke(...)`

zachriggle:
I don't understand the purpose of this. Commands are already classes and
use GDB's well-defined interface.

MaskRay:
It is more about the syntax. You can define attributes in new classes to customization.
Instead of `def hexdump(...):...\nhexdump.last_address = 0`, you can define an attribute to provide the same feature, which looks prettier.

When `argparse`-style argument parser is enabled, these settings can be moved to the body of the class, instead of defining it in the toplevel.

New hooks can be provided by overriding methods defined in the base class for command (such as `do_complete`, `do_invoke`). 

@disconnect3d It should be a rare use case to launch pwndbg/commands/.... commands in command line. Most commands should be supposed to launch from pwndbg/...  Nevertheless, it can be emulated

```python
class GenericCommand(future.utils.with_metaclass(MetaCommand, gdb.Command)):
    ...
    singleton = None

    def __init__(self, *args, **kwargs):
        ...
        type(self).singleton = self
    
    @classmethod
    def call(cls, **kwargs):
        return cls.singleton.do_invoke(**kwargs)

`(gdb) py aa.call(argv=3)`

**Comment by zachriggle** [comment]
Re "It should be a rare use case to launch pwndbg/commands/.... commands", I do it all the time.  Adding `foo.call()` or `foo.invoke()` is a weird hack.

This also appears to *increase* the amount of code needed to add a simple attribute to a command.

**Comment by zachriggle** [comment]
I'm pretty sure we can trivially get `ArgparsedCommand` objects to work nicely with `OnlyWhenRunning` and friends on `Command.invoke`.

Persistence is pretty easy to achieve via decorators, just add a function that invokes the following function with `self, *a, **kw`.

**Comment by MaskRay** [comment]
> increase the amount of code needed to add a simple attribute to a command.

```python
class A(GenericCommand):
    at = 1
```

compared with

```python
def A():
    ...
A.at = 3
```

zachriggle:
No, compared to:

```py
class A(GenericCommand):
    at = 1

@A
def Foobar(self, ...):
    self.at = 1
    ...
```

Or

```py
class A(GenericCommand):
    at = 1

    def do_invoke(self, ...):
        self.at = 1
        ...
```

In any case, my proposed solution would be, effectively:

```py
@ArgparsedCommand(parser)
@PersistentFunction(a = 33)
def foobar(self, x, y):
    self.a += 1
    return (x * y) * self.a
```

The benefit here is that `PersistentFunction` can be applied to *literally any function* and does not give up the benefits of being able to call it directly.

zachriggle:
As a basic example that I put together in a few minutes, here's an example implementation for `PersistentFunction` that'd drop into my example above:

```py
class PersistentFunction(object):
    function = None
    def __init__(self, **kw):
        for name, value in kw.items():
            setattr(self, name, value)
    def __call__(self, *a, **kw):
        # First call sets the function
        if self.function is None:
            self.function = a[0]
            return self

        # Subsequent calls just add 'self' as the first arg
        return self.function(self, *a, **kw)

@PersistentFunction(a = 0)
def foobar(self, x, y):
    self.a += 1
    return (x*y) * self.a

assert 1 == foobar(1, 1)
assert 2 == foobar(1, 1)
assert 3 == foobar(1, 1)
```

zachriggle:
In any case, I don't see a problem with using regular-old `global foo` or `function.foo = bar` for persistence.  If you want to be fancy about it, I like the `PersistentFunction` decorator.

zachriggle:
To address the issue of customization of completion, my view is that we should only bother with completion for `ArgparsedCommand` types, and that a sane completion should be declared there.  There should not be a need for a command to implement its own completion.

disconnect3d:
For me the class-way gef is doing commands seems kinda easy/pretty:
https://github.com/hugsy/gef/blob/master/gef.py#L3927-L3933

They also use ` gdb.string_to_argv`:
https://github.com/hugsy/gef/blob/master/gef.py#L3023-L3024

However I won't cry if we don't use classes for commands.

zachriggle:
The only reason that GEF uses classes is because they pathologically don't want to split things into different files.  It's the only organizational construct left to them.

`gdb.string_to_argv` is used by `pwndbg.commands.Command`, you're free to use it if you like.  You'll quickly realize why `ParsedCommand` exists as an extension of that, and why `ArgparsedCommand` has basically replaced it.

zachriggle:
As a side-note, since you guys might not be as familiar: `pwndbg.command.Command`, which uses `gdb.string_to_argv`, is exactly what backs both `ParsedCommand` and `ArgparsedCommand` string lexing.

disconnect3d:
Should we close this PR?

disconnect3d:
As we haven't agreed on merging this and the discussion died, I am closing this.
Please reopen if you have more thoughts/want to push this further.

PR #276, opened 2017-05-15
Fix weird input issue

disconnect3d:
Reverts two lines from previous version. This fixes an issue that happens on my setup when gdb is launched through pwntools, e.g.:
```bash
$ pwn debug --exe /bin/ls
```

The problem is that when I resize gdb to fullscreen or as "half of screen" (e.g. stick to left/right side), firing a command e.g. `ni` asks me for `return` or `q` like the output would be very long.

My setup:
```
$ cat ~/.pwn.conf 
[context]
terminal=['terminator', '-x']
```

Pwndbg (guess it doesn't matter):
```
pwndbg> version
Gdb: GNU gdb (GDB) 7.12.1
Pwndbg: 1.0.0 build: f7b2811
```

This is how it looked before this fix:
https://i.gyazo.com/e3e0e67d92be775ff9d45c54793da27e.png

### DIFF
--- a/pwndbg/__init__.py
+++ b/pwndbg/__init__.py
@@ -121,6 +121,7 @@ pre_commands = """
 set confirm off
 set verbose off
 set prompt %s
+set pagination off
 set height 0
 set history expansion on
 set history save on


disconnect3d:
This is actually caused by the fact gdb changes `height` value when you resize window:
```
pwndbg> show height
Number of lines gdb thinks are in a page is unlimited.
pwndbg> show height
Number of lines gdb thinks are in a page is 24.
pwndbg> show height
Number of lines gdb thinks are in a page is 53.
```

zachriggle:
Can you try reverting a few commits one-at-a-time to figure out which one is causing this issue?

```shell
$ git stash save --all # Save all of your work
$ git checkout HEAD^ # Go back one commit
$ git clean -xdf # Clean up any .pyc files
$ gdb /bin/bash # Do your testing
```

disconnect3d:
I haven't checked commits one by one but it seems it happens for all of them.

I guess we should add `set pagination off` to initialization as you suggested on irc as a workaround for me.

We never used it ;p :
```
(pwn) [dc@dc:pwndbg|dev $]$ git grep pagination $(git rev-list --all)
(pwn) [dc@dc:pwndbg|dev $]$ 
```

It doesn't change the way how `height` is set (still changes during window resize) but it works (doesnt ask for enter/q).

disconnect3d:
@zachriggle I have changed PR to just add `set pagination off` into initialization.

PR #278, opened 2017-05-16
Clean up some unnecessary closures in pwndbg.commands

zachriggle:
There are a bunch of closures used for decorating, because I didn't have a good grasp on classes-as-decorators when I originally wrote this code.

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -20,14 +20,14 @@ import pwndbg.symbol
 import pwndbg.ui
 
 
-class _Command(gdb.Command):
+class Command(gdb.Command):
     """Generic command wrapper"""
     count    = 0
     commands = []
     history  = {}
 
     def __init__(self, function, inc=True, prefix=False):
-        super(_Command, self).__init__(function.__name__, gdb.COMMAND_USER, gdb.COMPLETE_EXPRESSION, prefix=prefix)
+        super(Command, self).__init__(function.__name__, gdb.COMMAND_USER, gdb.COMPLETE_EXPRESSION, prefix=prefix)
         self.function = function
 
         if inc:
@@ -47,6 +47,8 @@ class _Command(gdb.Command):
         except TypeError:
             pwndbg.exception.handle()
             raise
+        except SystemExit:
+            pass
         finally:
             self.repeat = False
 
@@ -72,8 +74,8 @@ class _Command(gdb.Command):
         number = int(number)
 
         # A new command was entered by the user
-        if number not in _Command.history:
-            _Command.history[number] = command
+        if number not in Command.history:
+            Command.history[number] = command
             return False
 
         # Somehow the command is different than we got before?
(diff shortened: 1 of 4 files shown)


disconnect3d:
Having a stupid simple command:
```
parser = argparse.ArgumentParser(description='lala')
parser.add_argument('bar', type=int)
@ArgparsedCommand(parser)
def foo(bar):
    print(bar)
```

Could have a better output:
```
pwndbg> foo
usage: foo [-h] bar
foo: error: the following arguments are required: bar
Python Exception <class 'SystemExit'> 2: 
Error occurred in Python command: 2
```

Would be nice to have it just saying `foo: the following arguments are required: bar`.

zachriggle:
I'm not sure why the last two lines are showing up.  Where are they originating from?

disconnect3d:
Fun-sad-fact: both `set exception-debugger on` and 'set exception-verbose on` and `foo` afterwards exits gdb :(

zachriggle:
Try now ^_^

disconnect3d:
Hm...

There are some weird/bug prone situations:
```
pwndbg> i r rsp
rsp            0x7fffffffd850	0x7fffffffd850
pwndbg> foo $rsp
140737488345168
pwndbg> i r esp
esp            0xffffd850	-10160
pwndbg> foo $esp
-10160
```

The `$esp` here passed as an address to command, casted/interpreted by us as int instead of unsigned, probably might boom in some cases.

On the other hand, one can do:
```
pwndbg> foo "(unsigned) $esp"
4294957136
```


When we try to get a register that doesn't exist, there is not really nice output:
```
pwndbg> foo $bar
Python Exception <class 'gdb.error'> Invalid cast.: 
Error occurred in Python command: Invalid cast.
```

Also, as those are gdb expressions, arithmetic works, but may not be obvious all the time (quotation required):
```
pwndbg> foo 1+1
2
pwndbg> foo 1 + 1
usage: foo [-h] bar
foo: error: unrecognized arguments: + 1
```

pwndbg/commands/__init__.py
```diff
@@ -37,16 +37,28 @@ def __init__(self, function, inc=True, prefix=False):
         self.__doc__ = function.__doc__
 
     def split_args(self, argument):
-        return gdb.string_to_argv(argument)
+        """Split a command-line string from the user into arguments.
+
+        Returns:
+            A tuple / dict, in the form of ``*args, **args``.
```
disconnect3d: copy paste bug, should be: `*args, **kwargs`

zachriggle:
Regarding your arithmetic example, there is **not any way at all** to handle the string "1 + 1" (with spaces) without some custom parsing magic.  `gdb.string_to_argv` splits this into three distinct tokens.  Users will need to remove any spaces.

zachriggle:
I added a bit to enhance the error messages slightly.

```
pwndbg> checksec
checksec: There is no file loaded.
pwndbg> hexdump $eax
hexdump: No registers. (<class 'gdb.error'>)
pwndbg> hexdump $rax
hexdump: Invalid cast. (<class 'gdb.error'>)
pwndbg> hexdump 0xcafe
+0000 0x00cafe
pwndbg> hexdump *(int*)0xcafe
hexdump: Cannot access memory at address 0xcafe (<class 'gdb.MemoryError'>)
```

disconnect3d (approved):

PR #279, opened 2017-05-19
Add pwndbg.storage and pwndbg.memoize.forever

zachriggle:
This is an alternative implementation to #268 that I contest is a simpler and more readable. 

Notable differences:

- It uses the built-in `shelve` module for persistence.
- It provides a volatile (tempfile) and non-volatile (~/.pwndbg) storage option
  - The path is always the same, regardless of OS.  Nice and simple, and the documentation is easier, too.
  - Exposes standard GDB settings (`pwndbg.config`) to control the locations if the user wants, rather than environment variables.
- Cache validity is determined by the modification time of the file that contains the function, rather than the function signature.  
  - This is more robust for development cycles where the logic may change, even if the signature doesn't.  
  - This still doesn't track changes in dependent routines.
  - There is actually an implied reliance on function signature, since we store the full set of arguments
- Adds `pwndbg.memoize.forever` and an alias to `pwndbg.storage.NonVolatile`

### DIFF
--- a/pwndbg/storage.py
+++ b/pwndbg/storage.py
@@ -24,70 +24,101 @@ import time
 import pwndbg.arch
 import pwndbg.config
 
-nonvolatile = pwndbg.config.Parameter('pwndbg-cache-nonvolatile', '~/.pwndbg', 'Non-volatile cache directory for Pwndbg storage')
+nonvolatile = pwndbg.config.Parameter('pwndbg-cache-nonvolatile', os.path.expanduser('~/.pwndbg'), 'Non-volatile cache directory for Pwndbg storage')
 volatile = pwndbg.config.Parameter('pwndbg-cache-volatile', tempfile.gettempdir(), 'Volatile cache directory for Pwndbg storage')
 
-def name(function):
-    mod_name = function.__module__
-    func_name = function.__name__
-    return mod_name + '.' + func_name
+# Ensure there is only one Shelf instance per database, so that
+# we don't get weird things with multiple instances / concurrent writes / etc.
+shelves = {
+    nonvolatile: None,
+    volatile: None
+}
 
-def path(function, directory=volatile):
-    return os.path.join(str(directory), name(function))
+@pwndbg.config.Trigger([volatile, nonvolatile])
+def update():
+    global shelves
+
+    for db in (volatile, nonvolatile):
+        database_path = os.path.join(str(db), 'storage')
+
+        # Ensure the path exists, then open the database
+        if not os.path.isdir(database_path):
+            os.makedirs(database_path)
+
+        shelves[db] = shelve.open(database_path,
+                                  writeback=True,
+                                  protocol=pickle.HIGHEST_PROTOCOL)
+
+# Perform manual update to initialize the values
+update()
 
 class Cached(object):
     """Function decorator that persistently caches the results of the function it decorates."""
     function = None
     type = volatile
-    shelf = None
 
     def __init__(self, function):
         self.function = function
+        self.function_name = function.__module__ + '.' + function.__name__
         functools.update_wrapper(self, function)
+
+    @property
+    def shelf(self):
+        return shelves[self.type]
+
+    @property
+    def calls(self):
+        return shelves[self.type][self.function_name]
+
+    @property
+    def initialized(self):
+        return self.function_name in self.shelf and 'create_time' in self.calls
+
 
     def __setup_shelf(self):
         """Deferred initialization allows customization of the paths by the user"""
-        self.shelf = shelve.open(path(self.function, self.type),
-                                 writeback=True,
-                                 protocol=pickle.HIGHEST_PROTOCOL)
 
-        # Set the creation time of the database, if it is not already set
-        self.shelf.setdefault('create_time', time.time())
+        # Create an empty dictionary for this function if nothing exists
+        self.shelf.setdefault(self.function_name, {})
 
         # Get the modification time of the function's Python file
-        function_mtime = os.path.getmtime(inspect.getfile(self.function))
+        try:
+            function_mtime = os.path.getmtime(inspect.getfile(self.function))
+        except Exception:
+            # If the function is from the repl, the path will be e.g. <stdin>
+            function_mtime = 999999999999999999999
 
-        # If the file that the function is declared in is newer than
-        # the database file, invalidate the cache.
-        if self.shelf['create_time'] < function_mtime:
-            self.shelf.clear()
-            self.shelf['create_time'] = time.time()
+        # Find out when the cache for this function was created
+        create_time = self.calls.get('create_time', 0)
+
+        # If the file is newer than the cache, nuke the cache
+        if create_time < function_mtime:
+            self.calls.clear()
+            create_time = time.time()
+
+        self.calls['create_time'] = create_time
 
     def __call__(self, *a, **kw):
-        if self.shelf is None:
+        # Initialize the database if needed
+        if not self.initialized:
             self.__setup_shelf()
 
-        # Cache per-architecture
-        #
-        # N.B.: This is important for a secondary reason:
-        #       Shelve objects can only have keys which are strings.
-        #       However, the inner objects can be ~= anything.
-        arch = pwndbg.arch.current
-        self.shelf.setdefault(arch, {})
-        calls = self.shelf[arch]
-
-        # Cache per-architecture x arguments
+        # Dict is not hashable, convert into a tuple so it can be a dict key
         args = inspect.getcallargs(self.function, *a, **kw)
-
-        # Dict is not hashable, convert into a tuple so it can be a dictk ey
-        args = tuple(sorted(args.iteritems()))
+        args = tuple(sorted(args.items()))
 
         # If we don't have this set of arguments so far...
-        if args not in calls:
-            calls[args] = self.function(*a, **kw)
-            self.shelf[arch] = calls
+        if args not in self.calls:
+            result = self.calls[args] = self.function(*a, **kw)
 
-        return self.shelf[arch][args]
+            # Save the new data
+            self.shelf.sync()
+
+        # Extract the return value from storage
+        retval = self.calls[args]
+
+
+        return retval
 
     def __repr__(self):
         return "%s(%r)" % (self.__class__.__name__, self.function)
@@ -103,3 +134,7 @@ class Volatile(Cached):
 def demo(foo=0, bar=1, baz=2):
     time.sleep(3)
     return foo, bar, baz
+
+# Make sure that nobody accidentally uses "Cached"
+# without needing to do weird Abstract Base Class stuff :P
+del Cached
--- a/pwndbg/storage.py
+++ b/pwndbg/storage.py
@@ -1,0 +1,105 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+Provides persistent volatile and non-volatile storage of the results of
+a function call.  Volatile storage uses the temporary directory, and is
+generally cleared each reboot, or as the OS dictates.
+
+Cached results are stored per-architecture, i.e. arm and i386 caches
+are independent.  Cached results are also stored per-argument.
+"""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import functools
+import inspect
+import os
+import pickle
+import shelve
+import tempfile
+import time
+
+import pwndbg.arch
+import pwndbg.config
+
+nonvolatile = pwndbg.config.Parameter('pwndbg-cache-nonvolatile', '~/.pwndbg', 'Non-volatile cache directory for Pwndbg storage')
+volatile = pwndbg.config.Parameter('pwndbg-cache-volatile', tempfile.gettempdir(), 'Volatile cache directory for Pwndbg storage')
+
+def name(function):
+    mod_name = function.__module__
+    func_name = function.__name__
+    return mod_name + '.' + func_name
+
+def path(function, directory=volatile):
+    return os.path.join(str(directory), name(function))
+
+class Cached(object):
+    """Function decorator that persistently caches the results of the function it decorates."""
+    function = None
+    type = volatile
+    shelf = None
+
+    def __init__(self, function):
+        self.function = function
+        functools.update_wrapper(self, function)
+
+    def __setup_shelf(self):
+        """Deferred initialization allows customization of the paths by the user"""
+        self.shelf = shelve.open(path(self.function, self.type),
+                                 writeback=True,
+                                 protocol=pickle.HIGHEST_PROTOCOL)
+
+        # Set the creation time of the database, if it is not already set
+        self.shelf.setdefault('create_time', time.time())
+
+        # Get the modification time of the function's Python file
+        function_mtime = os.path.getmtime(inspect.getfile(self.function))
+
+        # If the file that the function is declared in is newer than
+        # the database file, invalidate the cache.
+        if self.shelf['create_time'] < function_mtime:
+            self.shelf.clear()
+            self.shelf['create_time'] = time.time()
+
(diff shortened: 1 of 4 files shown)


zachriggle:
Actually, I should change this so that the top-level key is the full path of the function being cached

pwndbg/storage.py
```diff
@@ -0,0 +1,105 @@
...
+    func_name = function.__name__
+    return mod_name + '.' + func_name
+
+def path(function, directory=volatile):
+    return os.path.join(str(directory), name(function))
+
+class Cached(object):
+    """Function decorator that persistently caches the results of the function it decorates."""
+    function = None
+    type = volatile
```
disconnect3d: Right now `Cached` and `Volatile` does the same thing.

To prevent one from randomly choosing `Cached` over `Volatile`, the `Cached` should be an abstract class.

However, at least in Py2 it is not possible to have an ABC without a method or property:
```python
In [3]: class A(object):
   ...:     __metaclass__ = ABCMeta
   ...:     

In [4]: class B(A):
   ...:     pass
   ...: 

In [5]: A()
Out[5]: <__main__.A at 0x7fe4c3aa2610>

In [6]: B()
Out[6]: <__main__.B at 0x7fe4c3aac590>
```

So this can be fixed by moving `type` into an `abstractproperty`:
```python
In [19]: class A(object):
    ...:     __metaclass__ = ABCMeta
    ...:     @abstractproperty
    ...:     def type(self):
    ...:         pass
    ...:     

In [20]: class B(A):
    ...:     @property
    ...:     def type(self):
    ...:         return "This is B"
    ...:     

In [21]: A()
<ipython-input-21-3cd318a12eea> in <module>()
----> 1 A()

TypeError: Can't instantiate abstract class A with abstract methods type

In [22]: B().type
Out[22]: 'This is B'
```

pwndbg/storage.py
```diff
@@ -0,0 +1,105 @@
...
+    shelf = None
+
+    def __init__(self, function):
+        self.function = function
+        functools.update_wrapper(self, function)
+
+    def __setup_shelf(self):
+        """Deferred initialization allows customization of the paths by the user"""
```
disconnect3d: I'd add "through `type` property" (if we make it a property ;P).

pwndbg/storage.py
```diff
@@ -0,0 +1,105 @@
...
+        #       Shelve objects can only have keys which are strings.
+        #       However, the inner objects can be ~= anything.
+        arch = pwndbg.arch.current
+        self.shelf.setdefault(arch, {})
+        calls = self.shelf[arch]
+
+        # Cache per-architecture x arguments
+        args = inspect.getcallargs(self.function, *a, **kw)
+
+        # Dict is not hashable, convert into a tuple so it can be a dictk ey
```
disconnect3d: Typo `dictk ey` -> `dict key`
So bad Python has no `FronzenDict` builtin...

pwndbg/storage.py
```diff
@@ -0,0 +1,105 @@
...
+        """Deferred initialization allows customization of the paths by the user"""
+        self.shelf = shelve.open(path(self.function, self.type),
+                                 writeback=True,
+                                 protocol=pickle.HIGHEST_PROTOCOL)
+
+        # Set the creation time of the database, if it is not already set
+        self.shelf.setdefault('create_time', time.time())
+
+        # Get the modification time of the function's Python file
+        function_mtime = os.path.getmtime(inspect.getfile(self.function))
```
disconnect3d: Just saying - `inspect.getfile` returns `<string>` for stuff wrapped in Python Interpreter:
```python
pwndbg> pi
In : from pwndbg.storage import Volatile, NonVolatile
In : @Volatile
...: def f():
...:     return 123
...: 
In : f()
  File "/home/dc/installed/pwndbg/pwndbg/storage.py", line 68, in __call__
    self.__setup_shelf()
  File "/home/dc/installed/pwndbg/pwndbg/storage.py", line 58, in __setup_shelf
    function_mtime = os.path.getmtime(inspect.getfile(self.function))
  File "/usr/lib/python3.6/genericpath.py", line 55, in getmtime
    return os.stat(filename).st_mtime
FileNotFoundError: [Errno 2] No such file or directory: '<stdin>'
```

pwndbg/storage.py
```diff
@@ -0,0 +1,105 @@
...
+        #       However, the inner objects can be ~= anything.
+        arch = pwndbg.arch.current
+        self.shelf.setdefault(arch, {})
+        calls = self.shelf[arch]
+
+        # Cache per-architecture x arguments
+        args = inspect.getcallargs(self.function, *a, **kw)
+
+        # Dict is not hashable, convert into a tuple so it can be a dictk ey
+        args = tuple(sorted(args.iteritems()))
```
disconnect3d: `args.items()` (for a very big dict it is a performance hit on py2) or `six.iteritems(args)`

pwndbg/storage.py
```diff
@@ -0,0 +1,105 @@
...
+
+nonvolatile = pwndbg.config.Parameter('pwndbg-cache-nonvolatile', '~/.pwndbg', 'Non-volatile cache directory for Pwndbg storage')
+volatile = pwndbg.config.Parameter('pwndbg-cache-volatile', tempfile.gettempdir(), 'Volatile cache directory for Pwndbg storage')
+
+def name(function):
+    mod_name = function.__module__
+    func_name = function.__name__
+    return mod_name + '.' + func_name
+
+def path(function, directory=volatile):
```
disconnect3d: Never called with the default.

disconnect3d:
I agree this is more simple & readable.

The path `~/.pwndbg` is never created, and so using `NonVolatile` demo version booms:
```python
pwndbg> pi
In : demo()
  File "/home/dc/installed/pwndbg/pwndbg/storage.py", line 68, in __call__
    self.__setup_shelf()
  File "/home/dc/installed/pwndbg/pwndbg/storage.py", line 52, in __setup_shelf
    protocol=pickle.HIGHEST_PROTOCOL)
  File "/usr/lib/python3.6/shelve.py", line 243, in open
    return DbfilenameShelf(filename, flag, protocol, writeback)
  File "/usr/lib/python3.6/shelve.py", line 227, in __init__
    Shelf.__init__(self, dbm.open(filename, flag), protocol, writeback)
  File "/usr/lib/python3.6/dbm/__init__.py", line 94, in open
    return mod.open(file, flag, mode)
_gdbm.error: [Errno 2] No such file or directory
```

pwndbg/storage.py
zachriggle: Nah, it's fine the way it is.  This greatly overcomplicates things to avoid the simple and harmless case where "Cached and Volatile are the same thing".

pwndbg/storage.py
zachriggle: No, it's specifically so that the `pwndbg.config.Parameter` (which defines the filesystem path) can be changed by the user before we decide where to store the database

pwndbg/storage.py
zachriggle: That's fine.  Don't do that :)

pwndbg/storage.py
zachriggle: Meh :)

pwndbg/storage.py
disconnect3d: Hm.. and do we have any use case for more then `volatile` and `nonvolatile` type?

We could also go for having `volatile=True/False` as `Cached's __init__` argument.

pwndbg/storage.py
zachriggle: Given that the size will be based on how many arguments there are for a function, we can safely assume the size doesn't matter.

pwndbg/storage.py
zachriggle: Quite unfortunate! :(

pwndbg/storage.py
disconnect3d: Suggested this because of previous suggestion of moving `type` into property :P.

zachriggle:
@disconnect3d I think that I fixed it.  I also simplified some of the logic -- there's now only two `Shelf` instances ever, which should fix some issues with multiple concurrent writes / state synchronization.

pwndbg/storage.py
zachriggle: Done in ba99595

pwndbg/storage.py
zachriggle: This function is now gone

pwndbg/storage.py
zachriggle: Fixed to use `items`

pwndbg/storage.py
zachriggle: Fixed this so that if `os.path.mtime` fails for any reason, we set the modification time to 99999999999999 or so

pwndbg/storage.py
zachriggle: This is now set via the constructor, and `Volatile` and `NonVolatile` now just inherit from `Cached` and provide a default value

pwndbg/storage.py
```diff
@@ -0,0 +1,144 @@
...
+
+@pwndbg.config.Trigger([volatile, nonvolatile])
+def update():
+    global shelves
+
+    for db in (volatile, nonvolatile):
+        database_path = os.path.join(str(db), 'storage')
+
+        # Ensure the path exists, then open the database
+        if not os.path.isdir(database_path):
```
disconnect3d, line 45: The `database_path` must be a path to a file (might be non existing) otherwise `shelve.open` fails:
```
$ gdb
Defined pretty print function (pp) and libc (CDLL)
Loaded 108 commands. Type pwndbg [filter] for a list.
  File "/home/dc/installed/pwndbg/gdbinit.py", line 36, in <module>
    import pwndbg # isort:skip
  File "/home/dc/installed/pwndbg/pwndbg/__init__.py", line 69, in <module>
    import pwndbg.storage
  File "/home/dc/installed/pwndbg/pwndbg/storage.py", line 53, in <module>
    update()
  File "/home/dc/installed/pwndbg/pwndbg/storage.py", line 50, in update
    protocol=pickle.HIGHEST_PROTOCOL)
  File "/usr/lib/python3.6/shelve.py", line 243, in open
    return DbfilenameShelf(filename, flag, protocol, writeback)
  File "/usr/lib/python3.6/shelve.py", line 227, in __init__
    Shelf.__init__(self, dbm.open(filename, flag), protocol, writeback)
  File "/usr/lib/python3.6/dbm/__init__.py", line 94, in open
    return mod.open(file, flag, mode)
_gdbm.error: [Errno 21] Is a directory
```

I'd go with:
```python
for db in map(str, (volatile, nonvolatile)):
    ...
```

pwndbg/storage.py
```diff
@@ -0,0 +1,144 @@
...
+    nonvolatile: None,
+    volatile: None
+}
+
+@pwndbg.config.Trigger([volatile, nonvolatile])
+def update():
+    global shelves
+
+    for db in (volatile, nonvolatile):
+        database_path = os.path.join(str(db), 'storage')
```
disconnect3d, line 42: For volatile we get a path of `/tmp/storage`. It should be more obvious that the storage comes from pwndbg.

pwndbg/storage.py
```diff
@@ -0,0 +1,144 @@
...
+    def __repr__(self):
+        return "%s(%r)" % (self.__class__.__name__, self.function)
+
+class NonVolatile(Cached):
+    def __init__(self, function):
+        super(NonVolatile, self).__init__(function, nonvolatile)
+
+class Volatile(Cached):
+    def __init__(self, function):
+        super(NonVolatile, self).__init__(function, volatile)
```
disconnect3d, line 137: Copy paste! `super(NonVolatile, self)` -> `super(Volatile, self)`

pwndbg/storage.py
```diff
@@ -0,0 +1,144 @@
...
+# Perform manual update to initialize the values
+update()
+
+class Cached(object):
+    """Function decorator that persistently caches the results of the function it decorates."""
+
+    def __init__(self, function, type):
+        self.function = function
+        self.function_name = function.__module__ + '.' + function.__name__
+        self.type = Type
```
disconnect3d, line 61: `Type` is undefined. Also, `type` shadows the builtin, so if we do care, maybe `type_`

pwndbg/storage.py
```diff
@@ -0,0 +1,144 @@
...
+    def shelf(self):
+        return shelves[self.type]
+
+    @property
+    def calls(self):
+        return shelves[self.type][self.function_name]
+
+    @property
+    def initialized(self):
+        return self.function_name in self.shelf and 'create_time' in self.calls
```
disconnect3d, line 74: `self.calls` is not defined

disconnect3d (changes requested):
Looks for me like WIP or you haven't pushed all commits :f

disconnect3d:
Uh.. there is a kinda big problem when you run more than one instance of pwndbg:
```
Loaded 108 commands. Type pwndbg [filter] for a list.
  File "/home/dc/installed/pwndbg/gdbinit.py", line 36, in <module>
    import pwndbg # isort:skip
  File "/home/dc/installed/pwndbg/pwndbg/__init__.py", line 69, in <module>
    import pwndbg.storage
  File "/home/dc/installed/pwndbg/pwndbg/storage.py", line 53, in <module>
    update()
  File "/home/dc/installed/pwndbg/pwndbg/storage.py", line 50, in update
    protocol=pickle.HIGHEST_PROTOCOL)
  File "/usr/lib/python3.6/shelve.py", line 243, in open
    return DbfilenameShelf(filename, flag, protocol, writeback)
  File "/usr/lib/python3.6/shelve.py", line 227, in __init__
    Shelf.__init__(self, dbm.open(filename, flag), protocol, writeback)
  File "/usr/lib/python3.6/dbm/__init__.py", line 94, in open
    return mod.open(file, flag, mode)
_gdbm.error: [Errno 11] Resource temporarily unavailable
```

zachriggle:
Re: Multiple instances, fuck.  I expect we can open it read-only for secondary instances.

disconnect3d:
Re: Multiple instances:
Sounds like our problem: http://stackoverflow.com/questions/20420148/persistent-multiprocess-shared-cache-in-python-with-stdlib-or-minimal-dependenci

zachriggle:
Hmm, I'm somewhat against the idea of having a big external dependency, let me revisit this in the future.  Sorry for the long delays, I recently moved ~2500km!

MaskRay:
Good to know the `shelve` module.
Sadly the Berkeley DB files cannot be easily inspected.
What if some arguments have a functional dependency on others or are simply redundant?

disconnect3d:
@zachriggle any news on this? :<

zachriggle:
Nope.  Synchronized storage is hard.

One solution might be to append the current GDB `pid` to the database name.  When starting, copy the old database to the pid-appended filename.  When exiting GDB gracefully, we can just overwrite the old "original" database.

This would mean that new GDB instances always get the database of the most-recently-closed database.

It's not ideal, but a step in the right direction, and doesn't require running a "real" database (e.g. mongo / sql / etc).

zachriggle:
We could just use the `sqlite3` module (which handles multiple processes / connections), but then we have to deal with SQL.

zachriggle:
Hmm, SQLite might work.  We can just use BLOB types and have a few fields, maybe use something like peewee or PonyORM to abstract away the actual SQL stuff.

Still all-Python, since we don't need some server software running a server.

Still relatively dependency-free, since everything can be installed with `pip`

disconnect3d:
Speaking of ORMs there is also SqlAlchemy, but I guess we don't need such robustness. It might also be harder to learn/master, however I have no comparison as I only used it and Django ORM (and not peewee or pony).

zachriggle:
Apparently this is an entirely solved problem.  `joblib.memory` handles both persistent and non-persistent memoizing of Python function results concurrently across multiple processes.

https://pythonhosted.org/joblib/memory.html

disconnect3d:
Closing this as it is too old :(

PR #280, opened 2017-05-19
refactored wrapper

degrigis:
I've refactored the wrappers code as discussed with @disconnect3d .
What do you think about this new organization? @zachriggle @anthraxx


pwndbg/wrappers/readelf.py
```diff
@@ -0,0 +1,31 @@
...
+from pwndbg.wrappers import readelf_path
+
+cmd_opt = {
+           "relocs":"-r",
+           "header":"-h"
+          }
+
+@pwndbg.wrappers.OnlyWithFile(readelf_path)
+def get_jmpslots():
+    readelf_out = subprocess.check_output([readelf_path,cmd_opt["relocs"],pwndbg.file.get_file(pwndbg.proc.exe)]).decode('utf-8')
```
zachriggle: Split this into a few lines, this is pretty impossible to read.    Also, I'd prefer just passing `"--relocs"` and `"--file-header"` rather than a lookup for the short version.

pwndbg/wrappers/file.py
```diff
@@ -0,0 +1,18 @@
...
+from __future__ import unicode_literals
+
+import subprocess
+import pwndbg.wrappers
+
+from pwndbg.wrappers import file_path
+
+@pwndbg.wrappers.OnlyWithFile(file_path)
+def is_statically_linked():
+        if "statically" in subprocess.check_output([file_path,pwndbg.file.get_file(pwndbg.proc.exe)]).decode('utf-8'):
```
zachriggle: Split this line up

pwndbg/wrappers/checksec.py
```diff
@@ -0,0 +1,38 @@
...
+    if "Full RELRO" in out:
+        relro = "Full RELRO"
+    elif "Partial RELRO" in out:
+        relro = "Partial RELRO"
+    return relro
+
+@pwndbg.wrappers.OnlyWithFile(checksec_path)
+def pie_status():
+    pie = "No PIE"
+    out = subprocess.check_output([checksec_path,cmd_opt["file"],pwndbg.file.get_file(pwndbg.proc.exe)]).decode('utf-8')
```
zachriggle: Split this lien up

pwndbg/wrappers/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+from __future__ import unicode_literals
+
+import pwndbg.which
+
+readelf_path  = pwndbg.which.which("readelf")
+checksec_path = pwndbg.which.which("checksec")
+file_path     = pwndbg.which.which("file")
+
+
+def OnlyWithFile(cmd):
```
zachriggle: Pick a different name (perhaps `OnlyWithCommand`), and use a class-style decorator.  This triple-nesting is pretty hacky, and doesn't make correct use of `functools.wraps`.

pwndbg/wrappers/__init__.py
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import pwndbg.which
+
+readelf_path  = pwndbg.which.which("readelf")
```
zachriggle: Having `readelf_path` and `xxx_path` for each item is a bit superfluous, no?

pwndbg/elf.py
```diff
@@ -284,4 +284,4 @@ def map_inner(ei_class, ehdr, objfile):
     for page in pages:
         page.objfile = objfile
 
-    return tuple(sorted(pages))
+    return tuple(sorted(pages))
```
zachriggle: Missing newline

pwndbg/wrappers/checksec.py
```diff
@@ -0,0 +1,38 @@
...
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import subprocess
+import pwndbg.wrappers
+
+from pwndbg.wrappers import checksec_path
+
+cmd_opt = {
```
zachriggle: Just use `"--file"`

pwndbg/wrappers/readelf.py
degrigis: Sure, sorry for the brutality. :laughing:

disconnect3d:
@degrigis
Btw this is conflicting as previous PR was merged, please resolve first :)

Also, if you have time you might look/review https://github.com/pwndbg/pwndbg/pull/284

pwndbg/wrappers/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import pwndbg.which
+
+
+class OnlyWithCommand(object):
+    def __init__(self, func):
+        self.func = func
```
disconnect3d: This is missing `functools.wraps` or something similar (don't remember if that's the proper way to wrap using a class decorator) and so makes wrapped function not having proper metadata.

E.g. it discards `func.__name__`:
```python
In [3]: class Decorator(object):
   ...:     def __init__(self, *a): pass
   ...:     
   ...: 

In [4]: @Decorator
   ...: def foo():
   ...:     pass
   ...: 

In [5]: foo.__name__
AttributeError                            Traceback (most recent call last)
<ipython-input-5-5f8bc38be3ae> in <module>()
----> 1 foo.__name__

AttributeError: 'Decorator' object has no attribute '__name__'

In [6]: def bar(): pass

In [7]: bar.__name__
Out[7]: 'bar'
```

disconnect3d: Another thing is all of other `Only***` are actually functions - it might be better to keep it consistent.

pwndbg/wrappers/checksec.py
```diff
@@ -0,0 +1,48 @@
...
+from __future__ import print_function
+from __future__ import unicode_literals
+
+
+import pwndbg.wrappers
+import subprocess
+
+
+@pwndbg.wrappers.OnlyWithCommand
+def get_raw_out():
```
disconnect3d: This is also `OnlyWithFile`

pwndbg/wrappers/checksec.py
```diff
@@ -0,0 +1,48 @@
...
+
+    local_path = pwndbg.file.get_file(pwndbg.proc.exe)
+    cmd = [get_raw_out.command_path,"--file",local_path]
+    return subprocess.check_output(cmd).decode('utf-8')
+
+
+@pwndbg.wrappers.OnlyWithCommand
+def relro_status():
```
disconnect3d: This is also `OnlyWithFile`

pwndbg/wrappers/checksec.py
```diff
@@ -0,0 +1,48 @@
...
+    if "Full RELRO" in out:
+        relro = "Full RELRO"
+    elif "Partial RELRO" in out:
+        relro = "Partial RELRO"
+
+    return relro
+
+
+@pwndbg.wrappers.OnlyWithCommand
+def pie_status():
```
disconnect3d: This is also `OnlyWithFile`

pwndbg/wrappers/file.py
```diff
@@ -0,0 +1,21 @@
...
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import subprocess
+import pwndbg.wrappers
+
+@pwndbg.wrappers.OnlyWithCommand
+def is_statically_linked():
```
disconnect3d: This is also `OnlyWithFile`

pwndbg/wrappers/readelf.py
```diff
@@ -0,0 +1,29 @@
...
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import subprocess
+import pwndbg.wrappers
+
+
+@pwndbg.wrappers.OnlyWithCommand
+def get_jmpslots():
```
disconnect3d: This is also `OnlyWithFile`

pwndbg/wrappers/readelf.py
```diff
@@ -0,0 +1,29 @@
...
+
+    local_path = pwndbg.file.get_file(pwndbg.proc.exe)
+    cmd = [get_jmpslots.command_path, "-r", local_path]
+    readelf_out = subprocess.check_output(cmd).decode('utf-8')
+
+    return '\n'.join(filter(lambda line: _extract_jumps(line), readelf_out.splitlines()))
```
disconnect3d: Just:
```python
return '\n'.join(filter(_extract_jumps, readelf_out.splitlines()))
```

pwndbg/wrappers/checksec.py
degrigis: doubt: `OnlyWithFile` belongs to the `command` module, is it ok to import it here?

pwndbg/wrappers/__init__.py
degrigis: mmm, I've followed the hint of @zachriggle by using the class style decorator.
It is convenient since I can access the command_path attribute within the function in the wrapper code.
f.i.
```
def get_raw_out():
    local_path = pwndbg.file.get_file(pwndbg.proc.exe)
    cmd = [get_raw_out.command_path,"--file",local_path]
    return subprocess.check_output(cmd).decode('utf-8')
```

Using a function it will become something like this:
```
def get_raw_out(cmd_path=''):
    local_path = pwndbg.file.get_file(pwndbg.proc.exe)
    cmd = [cmd_path,"--file",local_path]
    return subprocess.check_output(cmd).decode('utf-8')
```
Basically I need to pass the cmd_path to each function from the decorator. 
i.e.
```
def OnlyWithCommand(function):
    @functools.wraps(function)
    def _OnlyWithCommand(*a, **kw):
        command = function.__module__.split(".")[-1]
        command_path = pwndbg.which.which(command)
        if command_path:
            return function(cmd_path=command_path,*a, **kw)
        else:
            raise OSError('Could not find command %s in $PATH' % command)
    return _OnlyWithCommand
```

Do you think it is ok? @disconnect3d

disconnect3d: This is Python, most of the objects can be freely changed and so you can add an attribute to a function object:
```python
def foo(): pass
foo.bar = "123"
```
So you can do `function.command_path = ...` in your example.

I am also not sure about fetching the command name based on file name - it is not that obvious(?) and "ugly" for programs with e.g. space in their name.

disconnect3d: @degrigis JFYI - about objects on which you can't set an attribute - there is a possibility to specify `__slots__` class static attribute which makes instances of this class not having a `__dict__` attribute.

For more, see [`__slots__` magic](http://book.pythontips.com/en/latest/__slots__magic.html) - a page I contributed/extended for Intermediate Python book.

pwndbg/wrappers/checksec.py
disconnect3d: Just try it. Imho there shouldn't be a cyclic imports problem as `pwndbg.commands` is not using this module.

pwndbg/wrappers/file.py
```diff
@@ -8,11 +8,14 @@
 import subprocess
 import pwndbg.wrappers
 
-@pwndbg.wrappers.OnlyWithCommand
+cmd_name = "file"
+
+
+@pwndbg.wrappers.OnlyWithCommand(cmd_name)
```
disconnect3d: I'd just pass `'file'` here instead of putting it in a variable.

pwndbg/wrappers/readelf.py
```diff
@@ -8,15 +8,17 @@
 import subprocess
 import pwndbg.wrappers
 
+cmd_name = "readelf"
 
-@pwndbg.wrappers.OnlyWithCommand
+
+@pwndbg.wrappers.OnlyWithCommand(cmd_name)
```
disconnect3d: Just pass `'readelf'`

pwndbg/wrappers/checksec.py
degrigis: Yes, no cyclic imports problems! I was more concerned about the code organization, but if it is ok let's do it.

disconnect3d:
1. The `checksec --file binary` prints output to stderr so that `subprocess.check_output(cmd)` is not enough to catch and/or prevent its output, e.g.:
```
pwndbg> got abc
[*] '/usr/bin/ls'
    Arch:     amd64-64-little
    RELRO:    Partial RELRO
    Stack:    Canary found
    NX:       NX enabled
    PIE:      No PIE (0x400000)
    FORTIFY:  Enabled
[*] '/usr/bin/ls'
    Arch:     amd64-64-little
    RELRO:    Partial RELRO
    Stack:    Canary found
    NX:       NX enabled
    PIE:      No PIE (0x400000)
    FORTIFY:  Enabled

GOT protection: No RELRO | GOT functions: 108
```

2. To prevent calling checksec/readelf multiple times and spending a lot of time on that - cache their output by using our memoize module. I guess that `@pwndbg.memoize.reset_on_objfile` is the proper one decorator to be used in here (just move `subprocess.whatever_that_takes_also_stderr` to seperate, decorated function).
Related: https://github.com/pwndbg/pwndbg/issues/281.

pwndbg/wrappers/__init__.py
```diff
@@ -24,3 +28,6 @@ def _OnlyWithCommand(*a,**kw):
                 raise OSError('Could not find command %s in $PATH' % self.cmd_name)
```
disconnect3d: This currently does:
```
pwndbg> got 123
got: Could not find command checksec in $PATH (<class 'OSError'>)
```

Maybe we should change it to print for now @zachriggle ?

degrigis:
It's been a while without any changes here, what's happening? @zachriggle @disconnect3d @anthraxx

pwndbg/commands/got.py
```diff
@@ -60,18 +53,8 @@ def got(name_filter=''):
...
-
-def _extract_jumps(l):
-    try:
-        if "JUMP" in l.split()[2]:
-            return l
-        else:
-            return False
-    except IndexError:
-        return False
+        print("[%s] %s -> %s" % (hex(address_val), light_yellow(name), pwndbg.chain.format(got_address)))
```
disconnect3d: Not really important but `0x%x` in format and just `address_val`

pwndbg/commands/got.py
```diff
@@ -12,44 +12,37 @@
...
-    file_out = pwndbg.wrappers.file(local_path)
-    if "statically" in file_out:
-        return "Binary is statically linked."
-
-    readelf_out = pwndbg.wrappers.readelf("-r", local_path)
-
-    jmpslots = '\n'.join(filter(lambda l: _extract_jumps(l),
-                         readelf_out.splitlines()))
+    '''
+    Show the state of the Global Offset Table
```
disconnect3d: Not necessary as this is argparsed command. This is not used anywhere - except e.g. for an IDE - but lets not split description into two places as it might not be synchronized properly in the future.

disconnect3d (approved):
Looks ok for me.

pwndbg/wrappers/checksec.py
```diff
@@ -0,0 +1,54 @@
...
+from __future__ import unicode_literals
+
+
+import pwndbg.commands
+import pwndbg.wrappers
+
+
+cmd_name = "checksec"
+
+@pwndbg.memoize.reset_on_objfile
```
zachriggle: These three decorators are used repeatedly throughout the module.  Might make sense to consolidate them to one decorator.

zachriggle (changes requested):

degrigis:
Sorry for the delay, I've resolved conflicts and consolidated decorators as @zachriggle suggests.

pwndbg/wrappers/readelf.py
```diff
@@ -0,0 +1,28 @@
...
+    local_path = pwndbg.file.get_file(pwndbg.proc.exe)
+    cmd = [get_jmpslots.cmd_path, "-r", local_path]
+    readelf_out = pwndbg.wrappers.call_cmd(cmd)
+
+    return '\n'.join(filter(_extract_jumps, readelf_out.splitlines()))
+
+
+def _extract_jumps(line):
+    try:
+        if "JUMP" in line.split()[2]:
```
disconnect3d: Can we add comment explaining this logic back?

```
# Checks for records in `readelf --relocs <binary>` which has type e.g. `R_X86_64_JUMP_SLO`
# NOTE: Because of that we DO NOT display entries that are not writeable (due to FULL RELRO)
# as they have `R_X86_64_GLOB_DAT` type.
#
# It might be good to display them seperately in the future.
```

pwndbg/wrappers/file.py
```diff
@@ -0,0 +1,23 @@
...
+def is_statically_linked():
+
+    local_path = pwndbg.file.get_file(pwndbg.proc.exe)
+    cmd = [is_statically_linked.cmd_path, local_path]
+
+    file_out = pwndbg.wrappers.call_cmd(cmd)
+
+    if "statically" in file_out:
```
disconnect3d: This will give us wrong results if the binary name or its interpreter path contains `statically`.

```bash
$ gcc main.c -o not_static
$ gcc main.c -static -o static
$ file *static
not_static: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.32, BuildID[sha1]=2d7bfb5e54f54cf1a86be713686103db0c91d0af, not stripped
static:     ELF 64-bit LSB executable, x86-64, version 1 (GNU/Linux), statically linked, for GNU/Linux 2.6.32, BuildID[sha1]=690a8e655d5c7752c003bb32f45f272dba8cebbd, not stripped
```

pwndbg/wrappers/readelf.py
```diff
@@ -0,0 +1,28 @@
...
+from __future__ import unicode_literals
+
+import pwndbg.wrappers
+
+cmd_name = "readelf"
+
+@pwndbg.wrappers.OnlyWithCommand(cmd_name)
+def get_jmpslots():
+    local_path = pwndbg.file.get_file(pwndbg.proc.exe)
+    cmd = [get_jmpslots.cmd_path, "-r", local_path]
```
disconnect3d: Please change the flag to `--relocs` as it were before.
Such things should be explicit so it is easier to understand code later on.

pwndbg/wrappers/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+
+class OnlyWithCommand(object):
+    def __init__(self, command):
+        self.cmd_name = command
+        self.cmd_path = pwndbg.which.which(command)
+
+    def __call__(self, function):
+        function.cmd_path = self.cmd_path
+
+        @pwndbg.memoize.reset_on_objfile
```
disconnect3d: Are you sure this works as expected? I haven't tested it but in my opinion `memoize` will make command with given arguments "return the same result", so it won't e.g. print stuff.

Imho we should memoize `call_cmd` instead.

pwndbg/wrappers/readelf.py
```diff
@@ -0,0 +1,28 @@
...
+    readelf_out = pwndbg.wrappers.call_cmd(cmd)
+
+    return '\n'.join(filter(_extract_jumps, readelf_out.splitlines()))
```
disconnect3d: Probably we don't need to join that strings. We should just `return filter(...)`.

NOTE: `filter(...)` returns an object that is a generator in Python 3, so to be able to do more then just iterate once over it, a cast to `tuple` or `list` is needed.

disconnect3d: We might also want to make `_extract_jumps` return a `pwndbg.elf.Relocs` or other, more meaningful named object in the future.

Just saying, we probably don't need it right now.

pwndbg/wrappers/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+import subprocess
+from subprocess import STDOUT
+
+import pwndbg.commands
+import pwndbg.which
+
+
+
+class OnlyWithCommand(object):
+    def __init__(self, command):
```
disconnect3d: We should probably register command name in some global set and check for its existence in the system after pwndbg is initialized. If it is not there - warn the user and disable particular command.

^This can be done in next PR; if we decide not to do this in current PR, please create an issue for it.

pwndbg/wrappers/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+
+import functools
+import subprocess
+from subprocess import STDOUT
+
+import pwndbg.commands
+import pwndbg.which
+
+
+
```
disconnect3d: Afaik pep8 cries if there are more then 2 newlines; guess isort doesn't check it ;f

pwndbg/commands/got.py
```diff
@@ -12,45 +12,34 @@
...
-
-    jmpslots = '\n'.join(filter(lambda l: _extract_jumps(l),
-                         readelf_out.splitlines()))
+    relro_status = pwndbg.wrappers.checksec.relro_status()
+    pie_status = pwndbg.wrappers.checksec.pie_status()
+    jmpslots = pwndbg.wrappers.readelf.get_jmpslots()
 
     if not len(jmpslots):
-        print(red("NO JUMP_SLOT entries available in the GOT"))
+        print("NO JUMP_SLOT entries available in the GOT")
```
disconnect3d: Isn't red color better for 'sad' paths? :(

disconnect3d (changes requested):

pwndbg/wrappers/readelf.py
degrigis: Ok, we will track this second idea in the Issue ok?

disconnect3d: @degrigis :+1:

pwndbg/wrappers/__init__.py
degrigis: Yup, maybe it's better to do this in another PR.
I'm going to add this to the Issue.

pwndbg/wrappers/file.py
degrigis: I've decided to remove the check and show the general message: "NO JUMP_SLOT entries available in the GOT", is it ok for you?

disconnect3d: Yeah, we can remove the check for being statically compiled. Probably we could check for it differently - looking for got/plt sections using readelf. However, the solution proposed by you is fine imho (if someone lacks the check he/she can add it :P).


Jfyi about checking for statically compiled binary - https://stackoverflow.com/questions/8040631/checking-if-a-binary-compiled-with-static says it can be done with `ldd` but I haven't checked it.

degrigis:
@disconnect3d @zachriggle ping :-)

pwndbg/wrappers/readelf.py
```diff
@@ -0,0 +1,34 @@
...
+    readelf_out = pwndbg.wrappers.call_cmd(cmd)
+
+    return filter(_extract_jumps, readelf_out.splitlines())
+
+'''
+ Checks for records in `readelf --relocs <binary>` which has type e.g. `R_X86_64_JUMP_SLO`
+ NOTE: Because of that we DO NOT display entries that are not writeable (due to FULL RELRO)
+ as they have `R_X86_64_GLOB_DAT` type.
+
+It might be good to display them seperately in the future.
```
disconnect3d: please move this to `def _extract_jumps(line)` so it will be its docstring

disconnect3d:
Can we fix isorts here & move the docstring? :)

degrigis:
finally solved my quarrel with mr. Travis

PR #284, opened 2017-05-21
got command: possibility to filter results

disconnect3d:
After the changes:

```
pwndbg> got write

GOT protection: Partial RELRO | GOT functions: 108
 
[00000061f280] fwrite_unlocked@GLIBC_2.2.5 -> 0x4026c6 (fwrite_unlocked@plt+6) ◂— push   0x4d /* 'hM' */
[00000061f330] fwrite@GLIBC_2.2.5 -> 0x402826 (fwrite@plt+6) ◂— push   0x63 /* 'hc' */
pwndbg> got
# 108 results here...
```

One thing I am not sure... Is it safe to remove `OnlyWithFile` when we have `OnlyWhenRunning`?
Can remote target not have a file under the hood?

### DIFF
--- a/pwndbg/commands/got.py
+++ b/pwndbg/commands/got.py
@@ -5,6 +5,8 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import argparse
+
 import pwndbg.chain
 import pwndbg.commands
 import pwndbg.enhance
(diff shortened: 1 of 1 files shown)


degrigis:
I like the idea, only refactor the code using the new code in #280 for wrappers when will be ready to merge :-)
I don't think remote target can be without a file, so I would say that `OnlyWhenRunning` is enough... @zachriggle ?

zachriggle (approved):

Issue #273, opened 2017-05-10
OnlyWhenRunning not working well with ArgparsedCommand


disconnect3d:
```
[dc@dc:pwndbg|repeat-commands $]$ gdb /bin/ls -q
Loaded 106 commands. Type pwndbg [filter] for a list.
Reading symbols from /bin/ls...(no debugging symbols found)...done.
pwndbg> hexdump
usage: hexdump [-h] [address] [count]
hexdump: error: argument address: invalid fix_int value: '$sp'
```

The order of decorators doesn't fix it (even makes it worse).

disconnect3d:
As we are going to use `ArgparsedCommand` for everything (see https://github.com/pwndbg/pwndbg/issues/244), maybe it is good to just add keyword arguments like `when_running`, `with_file`?

zachriggle:
Hmm.  I think the issue here is the use of the `default` with a register value.  The current `hexdump` command is broken when being invoked from the interactive shell, anyway.

I suggest this workaround:

```patch
diff --git a/pwndbg/commands/hexdump.py b/pwndbg/commands/hexdump.py
index ed315f8..2ee8d52 100644
--- a/pwndbg/commands/hexdump.py
+++ b/pwndbg/commands/hexdump.py
@@ -25,3 +25,3 @@ pwndbg.config.Parameter('hexdump-bytes',
 parser = argparse.ArgumentParser(description='Hexdumps data at the specified address (or at $sp)')
-parser.add_argument('address', nargs='?', default='$sp',
+parser.add_argument('address', nargs='?', default=None,
                     help='Address to dump')
@@ -33,2 +33,4 @@ parser.add_argument('count', nargs='?', default=pwndbg.config.hexdump_bytes,
 def hexdump(address=None, count=pwndbg.config.hexdump_bytes):
+    if address is None:
+        address = pwndbg.regs.sp
```

zachriggle:
Alternately we need some other way to specify `OnlyWhileRunning` for `ArgparsedCommand` objects, since `fix_int("$sp")` will blow up if the target isn't running -- but we can't block `fix_int` with `@OnlyWhileRunning` since that'd prevent other things (e.g. inspecting the file itself before runtime).

disconnect3d:
Closing, this will be tracked in #1462

PR #285, opened 2017-05-23
Move vmmap to ArgparsedCommand; add sloppy_gdb_parse

disconnect3d:
This PR does three things:

* Adds `sloppy_gdb_parse` which can be used with `argparse.ArgumentParser`'s `.add_argument()` method's `type` argument
* Moves `vmmap` command to use `ArgparsedCommand` (https://github.com/pwndbg/pwndbg/issues/273)
* Adds a message when there are no mappings and leaving user with a WTF legend output

### DIFF
--- a/pwndbg/commands/vmmap.py
+++ b/pwndbg/commands/vmmap.py
@@ -46,6 +46,10 @@ def vmmap(map=None):
 
     pages = list(filter(pages_filter, pwndbg.vmmap.get()))
 
+    if not pages:
+        print('There are no mappings for specified address or module.')
+        return
+
     print(M.legend())
     for page in pages:
         print(M.get(page.vaddr, text=str(page)))
--- a/pwndbg/commands/vmmap.py
+++ b/pwndbg/commands/vmmap.py
@@ -1,12 +1,15 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 """
-Command to print the vitual memory map a la /proc/self/maps.
+Command to print the virtual memory map a la /proc/self/maps.
 """
 from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
+
+import argparse
+import functools
 
 import gdb
 import six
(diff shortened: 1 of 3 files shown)


disconnect3d:
Not sure if `sloppy_gdb_parse` is the best way to do what we want here.

Also currently this breaks other commands that are aliases for `vmmap`.

zachriggle:
I'd much rather just have a type that returns a `Map` object that handles all of the logic that you've added.  Rather than just a generic `sloppy` where the command is responsible for translating the worst-case string result into a value, it should be done by the argument-type-routine.

zachriggle:
Still waiting on changes mentioned in https://github.com/pwndbg/pwndbg/pull/285#issuecomment-304548234

zachriggle (changes requested):
See comment

disconnect3d:
@zachriggle I am not sure if that is what you meant but I have made some changes.
One thing to note is that current version  makes `vmmap` being unable to filter by address/module name when used through python interpreter:

https://i.gyazo.com/1e8f37a6cb4a2241fb09d2644346530c.png

So I am not sure if that is the best option.

EDIT: It seems that we parse arguments through argparse only if command is invoked through gdb. This logic is probably in `commands/__init__.py:Command.invoke` method. Shouldn't we parse it in `__call__`?

disconnect3d:
@zachriggle bump

pwndbg/commands/vmmap.py
```diff
@@ -17,27 +20,38 @@
...
+parser = argparse.ArgumentParser()
+parser.description = 'Print the virtual memory map, or the specific mapping for the provided address / module name.'
+parser.add_argument('map', type=map_type, nargs='?', default=None,
+                    help='Address or module name.')
+
+
+@pwndbg.commands.ArgparsedCommand(parser)
 @pwndbg.commands.OnlyWhenRunning
 def vmmap(map=None):
-    """
```
zachriggle: Don't remove the docstring <3

pwndbg/commands/vmmap.py
```diff
@@ -17,27 +20,38 @@
...
-    Print the virtal memory map, or the specific mapping for the
-    provided address / module name.
-    """
-    int_map = None
-    str_map = None
-
-    if isinstance(map, six.string_types):
```
zachriggle, line 30: I see that this was moved up to `map_type`, OK

pwndbg/commands/vmmap.py
```diff
@@ -17,27 +20,38 @@
...
-        str_map = map
-    elif isinstance(map, six.integer_types + (gdb.Value,)):
-        int_map = int(map)
+    print(map)
+    pages = list(filter(map, pwndbg.vmmap.get()))
```
zachriggle: `filter` takes a `callable`.  I'm confused about how this is supposed to work, if `map` is a `Page` instance

zachriggle: Never mind, I see now that `map` is actually a `lambda`. 

Can we change the name of the variable? It's overloading a keyword already, and is genuinely misleading in any case since it is not a `Page` (virtual memory map) nor a `{}` object.

zachriggle (changes requested):

pwndbg/commands/vmmap.py
zachriggle: Separately, in an ideal world, the `map_type` function would make this `filter` call and return the actual objects, instead of a method for filtering them.

pwndbg/commands/vmmap.py
```diff
@@ -17,27 +20,38 @@
...
-    provided address / module name.
-    """
-    int_map = None
-    str_map = None
-
-    if isinstance(map, six.string_types):
-        str_map = map
-    elif isinstance(map, six.integer_types + (gdb.Value,)):
-        int_map = int(map)
+    print(map)
```
zachriggle: I think technically we're printing out the `lambda` here, should this go away?

pwndbg/commands/vmmap.py
disconnect3d: Shouldn't we do that for all `ArgparsedCommand`s?
I mean, we either should grab command description from docstring or specify it directly on parser instance.

zachriggle:
Oh shoot, forgot about ArgparsedCommand decorator on this.

Do we auto-generate a docstring already for these?

disconnect3d:
@zachriggle ~I dont think so.~
Yeah, we do so in here https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/commands/__init__.py#L208.

disconnect3d:
@zachriggle Also is that okay that one won't be able to pass int/str to `vmmap` manually from Python?

disconnect3d:
@zachriggle check this now

zachriggle:
This'll work for now, I'll dork with it more later if I ever have free time again

zachriggle (approved):

zachriggle:
Looks like CI failed due to a previous change (the `search` `gdbserver` one)

If you can get CI to pass, feel free to merge ❤️

PR #290, opened 2017-06-01
extend next_call to take optional symbol/target to break on

anthraxx:
Regex is choosen to provide a more sophisticated control over targets to match on. By default it does an exact match, so providing `printf` will exactly match on `printf`. If a match on `snprintf` is also desired, one should pass something like `.*printf`. This approach provides more user control on what should be beaked on.

Additionally the target can be matched too, this is especially handy for stripped down binaries. The specific reason to the regex approach at this is also more sophisticated control when there is one big stripped static binary and we want to break on the next call to something like `0x42fd.*`. This way we provide a more feature rich way to match the target/symbol with no downsides.

### DIFF
--- a/pwndbg/next.py
+++ b/pwndbg/next.py
@@ -9,6 +9,8 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import re
+
 import capstone
 import gdb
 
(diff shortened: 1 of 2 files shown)


anthraxx:
Special thanks to @ysf

pwndbg/next.py
```diff
@@ -78,14 +80,27 @@ def break_next_interrupt(address=None):
         gdb.execute('continue', from_tty=False, to_string=True)
         return ins
 
-def break_next_call(address=None):
```
zachriggle, line 81: Please add docs for `symbol_regex`

zachriggle: In particular, it's useful to have a few examples of how this ought to work

pwndbg/next.py
```diff
@@ -78,14 +80,27 @@ def break_next_interrupt(address=None):
...
+        # continue if not a call
+        if capstone.CS_GRP_CALL not in ins.groups:
+            continue
+
+        # return call if we don't search for a symbol
+        if not symbol_regex:
+            return ins
+
+        # return call if we match target address
+        if ins.target_const and re.match('%s$' % symbol_regex, hex(ins.target)):
```
zachriggle, line 99: It seems that the `$` is a trade-off between matching too many things, and forcing the user to add `.*` to all of their matches.  I think that if we're adding `$` at the end, we should also add `^` to the beginning.

zachriggle (changes requested):

anthraxx:
Match vall always matches from the very beginning, there is no need for the ^

zachriggle:
Huh, I was definitely thinking of `search`.  Thanks!

PR #291, opened 2017-06-01
handle window resize event and set width accordingly

### LINKED ISSUE #215: Typesetting seems to be wrong
image

anthraxx:
This fixes #215 and gives proper aligned tab-completion.

### DIFF
--- a/pwndbg/ui.py
+++ b/pwndbg/ui.py
@@ -9,6 +9,7 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import fcntl
+import os
 import struct
 import sys
 import termios
@@ -22,13 +23,21 @@ theme.Parameter('banner-separator', '─', 'repeated banner separator character'
 
 def banner(title):
     title = title.upper()
-    try:
-        _height, width = struct.unpack('hh', fcntl.ioctl(sys.stdin.fileno(), termios.TIOCGWINSZ, '1234'))
-    except:
-        width = 80
+    _height, width = get_window_size()
     width -= 2
     return C.banner(("[{:%s^%ss}]" % (config.banner_separator, width)).format(title))
 
 def addrsz(address):
     address = int(address) & pwndbg.arch.ptrmask
     return "%{}x".format(2*pwndbg.arch.ptrsize) % address
+
+def get_window_size():
+    fallback = (int(os.environ.get('LINES', 20)), int(os.environ.get('COLUMNS', 80)))
+    if not sys.stdin.isatty:
+        return fallback
+    try:
+        # get terminal size and force ret buffer len of 4 bytes for safe unpacking by passing equally long arg
+        rows, cols = struct.unpack('hh', fcntl.ioctl(sys.stdin.fileno(), termios.TIOCGWINSZ, '1234'))
+    except:
+        rows, cols = fallback
+    return rows, cols
--- a/pwndbg/__init__.py
+++ b/pwndbg/__init__.py
@@ -5,6 +5,8 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import signal
+
 import gdb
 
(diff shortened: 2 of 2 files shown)


pwndbg/ui.py
```diff
@@ -22,13 +23,17 @@
...
 
 def addrsz(address):
     address = int(address) & pwndbg.arch.ptrmask
     return "%{}x".format(2*pwndbg.arch.ptrsize) % address
+
+def get_window_size():
+    try:
+        rows, cols = struct.unpack('hh', fcntl.ioctl(sys.stdin.fileno(), termios.TIOCGWINSZ, '1234'))
+    except:
+        rows, cols = (os.environ['LINES'], os.environ['COLUMNS'])
```
disconnect3d: What is the case when the `except` happens?
Also if someone has no `LINES` or `COLUMNS` env vars it might be better to use `os.environ.get(envvar, default_value)`.

pwndbg/ui.py
```diff
@@ -22,13 +23,17 @@
...
     width -= 2
     return C.banner(("[{:%s^%ss}]" % (config.banner_separator, width)).format(title))
 
 def addrsz(address):
     address = int(address) & pwndbg.arch.ptrmask
     return "%{}x".format(2*pwndbg.arch.ptrsize) % address
+
+def get_window_size():
+    try:
+        rows, cols = struct.unpack('hh', fcntl.ioctl(sys.stdin.fileno(), termios.TIOCGWINSZ, '1234'))
```
disconnect3d: I know it was there previously, but why is last argument `'1234'`? If we want to pass any string, I believe we can omit the argument so the behavior will be as if a string had been passed. From `ioctl(fd, op[, arg[, mutate_flag]])` docstring:
```
(...)
If the argument is a mutable buffer and the mutable_flag argument is not
passed or is false, the behavior is as if a string had been passed.  This
behavior will change in future releases of Python.

If the argument is an immutable buffer (most likely a string) then a copy
of the buffer is passed to the operating system and the return value is a
string of the same length containing whatever the operating system put in
the buffer.  The length of the arg buffer in this case is not allowed to
exceed 1024 bytes.
```

anthraxx:
It needs to be any string with a length of 4 so we ensure a return buffer length of exactly 4 bytes as we require. This was also just moved down from above :P

Will replace the env default get, however afaik this ioctl call can fail depending on terminal therefor the except is used.

disconnect3d:
May you add a comment for that buffer of length 4? :D

disconnect3d:
That works me o/.

zachriggle:
Neither `sys.stdin` nor `sys.stdout` are guaranteed to be valid files (.fileno() may not work) nor to be TTYs (should check .isatty() first).

anthraxx:
checking for `isatty()` now, however stdin should be a valid resource, otherwise OSError will be thrown (and there is a except block anyway in case its not valid and ioctl can't operate on it)

disconnect3d (approved):

pwndbg/ui.py
```diff
@@ -22,13 +23,21 @@
...
+    _height, width = get_window_size()
     width -= 2
     return C.banner(("[{:%s^%ss}]" % (config.banner_separator, width)).format(title))
 
 def addrsz(address):
     address = int(address) & pwndbg.arch.ptrmask
     return "%{}x".format(2*pwndbg.arch.ptrsize) % address
+
+def get_window_size():
+    fallback = (os.environ.get('LINES', 20), os.environ.get('COLUMNS', 80))
```
zachriggle: `fallback` can be a pair of strings in this case, need to convert to integer

zachriggle (changes requested):

anthraxx:
converted returned LINES/COLUMNS to integer

anthraxx:
Ping?

PR #293, opened 2017-06-04
ARM/Thumb mode display

techvoltage:
Did some minor modifications to display mode while running on ARM platforms

### DIFF
--- a/pwndbg/arch.py
+++ b/pwndbg/arch.py
@@ -17,7 +17,8 @@ import pwndbg.regs
 import pwndbg.typeinfo
 
 current = 'i386'
-mode	= {CS_MODE_ARM:"ARM",CS_MODE_THUMB:"THUMB",None:""}[None]
+MODES	= {CS_MODE_ARM:" ARM MODE",CS_MODE_THUMB:" THUMB MODE",None:""}
+mode    = MODES[None]
 qemu    = current
 ptrmask = 0xfffffffff
 endian  = 'little'
--- a/pwndbg/arch.py
+++ b/pwndbg/arch.py
@@ -17,6 +17,7 @@ import pwndbg.regs
 import pwndbg.typeinfo
 
 current = 'i386'
+mode	= {CS_MODE_ARM:"ARM",CS_MODE_THUMB:"THUMB",None:""}[None]
 qemu    = current
 ptrmask = 0xfffffffff
 endian  = 'little'
--- a/pwndbg/disasm/__init__.py
+++ b/pwndbg/disasm/__init__.py
@@ -85,7 +85,7 @@ def get_disassembler(pc):
     if pwndbg.arch.current in ('arm', 'aarch64'):
         extra = {0:CS_MODE_ARM,
                  0x20:CS_MODE_THUMB}[pwndbg.regs.cpsr & 0x20]
-
+    pwndbg.arch.mode={CS_MODE_ARM:" ARM MODE",CS_MODE_THUMB:" THUMB MODE",None:""}[extra]
     return get_disassembler_cached(pwndbg.arch.current,
                                    pwndbg.arch.ptrsize,
                                    pwndbg.arch.endian,
--- a/pwndbg/disasm/__init__.py
+++ b/pwndbg/disasm/__init__.py
@@ -85,7 +85,7 @@ def get_disassembler(pc):
     if pwndbg.arch.current in ('arm', 'aarch64'):
         extra = {0:CS_MODE_ARM,
                  0x20:CS_MODE_THUMB}[pwndbg.regs.cpsr & 0x20]
-    pwndbg.arch.mode={CS_MODE_ARM:" ARM MODE",CS_MODE_THUMB:" THUMB MODE",None:""}[extra]
+    pwndbg.arch.mode = pwndbg.arch.MODES[extra]
     return get_disassembler_cached(pwndbg.arch.current,
                                    pwndbg.arch.ptrsize,
                                    pwndbg.arch.endian,
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -151,7 +151,7 @@ Unicorn emulation of code near the current instruction
 code_lines = pwndbg.config.Parameter('context-code-lines', 10, 'number of additional lines to print in the code context')
 
 def context_disasm():
-    banner = [pwndbg.ui.banner("disasm")]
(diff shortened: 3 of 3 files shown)


pwndbg/arch.py
```diff
@@ -18,6 +18,7 @@
 import pwndbg.typeinfo
 
 current = 'i386'
+mode	= {CS_MODE_ARM:"ARM",CS_MODE_THUMB:"THUMB",None:""}[None]
```
disconnect3d: It would be better to have this dict only once in the code, so maybe it would be good to do:
```python
MODES = {CS_MODE_ARM: "ARM", CS_MODE_THUMB: "THUMB", None: ""}
```

and then:
```python
mode = MODES[None]
```

pwndbg/disasm/__init__.py
```diff
@@ -85,7 +85,7 @@ def get_disassembler(pc):
     if pwndbg.arch.current in ('arm', 'aarch64'):
         extra = {0:CS_MODE_ARM,
                  0x20:CS_MODE_THUMB}[pwndbg.regs.cpsr & 0x20]
-
+    pwndbg.arch.mode={CS_MODE_ARM:" ARM MODE",CS_MODE_THUMB:" THUMB MODE",None:""}[extra]
```
disconnect3d: With above proposal we could go with: `pwndbg.arch.mode = pwndbg.arch.MODES[extra]`.

The last thing would be better adjusting of `banner` string (e.g. adding `'MODE'` there ;p)

pwndbg/arch.py
techvoltage: Okay. 
Although, I did it it avoid the naming confusion due to so many modes(mode, CapstoneMode, MODES). Also changing the variable name to anything other than mode would make it loose its meaning.

pwndbg/disasm/__init__.py
techvoltage: 'MODE' string is already added 
{CS_MODE_ARM:" ARM MODE",CS_MODE_THUMB:" THUMB MODE",None:""}
Putting it besides banner will make it print always. 
But on x86 this would not make any sense.
E.g on x86:
Banner would look like DISASM MODE
On ARM:
Banner would look like DISASM THUMB MODE

techvoltage:
Please review again. I've made the necessary changes.

pwndbg/disasm/__init__.py
```diff
@@ -85,7 +85,7 @@ def get_disassembler(pc):
     if pwndbg.arch.current in ('arm', 'aarch64'):
         extra = {0:CS_MODE_ARM,
                  0x20:CS_MODE_THUMB}[pwndbg.regs.cpsr & 0x20]
-
+    pwndbg.arch.mode = pwndbg.arch.MODES[extra]
```
zachriggle, line 88: What if `extra` is not in `pwndbg.arch.MODES`?

zachriggle (changes requested):

pwndbg/disasm/__init__.py
disconnect3d: Currently the `extra`  can't be other than `None`, `CS_MODE_ARM` or `CS_MODE_THUMB`.

We can change it to `.get(extra, None)` so that future changes won't break stuff, but I guess we can leave it as it is?

disconnect3d (approved):

disconnect3d:
@zachriggle Any idea why `isort` is failing on files that hasn't been changed on travis? ;|

zachriggle:
Re: isort, no idea :(

disconnect3d:
@zachriggle guess we can just ignore it and merge.

zachriggle (changes requested):
`mode` should update on `pwndbg.events.stop` somewhere inside of `pwndbg/arch.py`.  `pwndbg.disasm` shouldn't be modifying module-level values in a different module.

pwndbg/disasm/__init__.py
zachriggle: I'd still prefer the safer `MODES.get(extra, "")` or something like that.

disconnect3d:
@techvoltage are you going to work on that?

techvoltage:
I'm really sorry. I'm preoccupied with thesis work atm. I could look at it in 2 weeks. But if you wish to proceed with it, please go ahead. I'll delete my pull req and you can start afresh, whichever way it suits you. Is there a joint contributor feature ? I dunno.

techvoltage:
Mode can change anytime. pwndbg.events.pause should be the right choice IMHO

zachriggle:
Sounds good to me :)


*Zach Riggle*

zachriggle:
Actually, on third thought, shouldn't we always update it?  It should be
possible to disassemble both Thumb and ARM mode instructions with the
"nearpc" command, depending on the specific address entered.  Specifically,
disassembling an *odd* address in ARM mode should still show Thumb
instructions.


*Zach Riggle*

disconnect3d:
So we should do it based on the **address** and not **the place we are right now on (and so the xpsr/cpsr register)**, so we can disassembly any instructions properly, right?

We should also change the `pc` into `address` in:
```python
 def get_disassembler(pc):
```
as it probably isn't always true it is a program counter/instruction pointer.

disconnect3d:
Bump @techvoltage

techvoltage:
Apologies. I have been running busy lately. Let me quickly get into this!

techvoltage:
I have been reading source code for pwndbg to figure out a proper place to put the change

disconnect3d:
@techvoltage bump! ;)

techvoltage:
I see we already have CapstoneArch, CapstoneEndian, CapstoneMode 
 in disasm/\_\_init\_\_.py https://github.com/pwndbg/pwndbg/blob/6aeffbea246870fdc3af569b1bb16e51a3e2f667/pwndbg/disasm/__init__.py#L34-L52
Based on this we can have either have CapstoneArmISAMode or add CS_MODE_THUMB and CS_MODE_ARM in CapstoneMode as it would be a more generalized approach. 

We can also remove the defining of extra inside get_disassembler https://github.com/pwndbg/pwndbg/blob/6aeffbea246870fdc3af569b1bb16e51a3e2f667/pwndbg/disasm/__init__.py#L86-L87
and have mode computed inside get_disassembler_cached itself

zachriggle:
One issue I have with this is that we should be able to follow mode transitions via Unicorn.

Having a single banner that says "Thumb Mode" may be incorrect if any of the instructions are e.g. `bx` or `blx` since those can change modes.

PR #297, opened 2017-06-17
Fix the length calculation of UTF-8 nearpc_prefix

pyllyukko:
The ```len(pwndbg.config.nearpc_prefix.value)``` returns ```3```, instead of ```1``` (at
least with Python 2.7), which causes misalignment of the output.

Here's an example how the alignment breaks:

pwndbg-5-2

When you set the nearpc-prefix to ```=>``` it aligns properly:

pwndbg-4-2

I added ```decode('utf-8')``` to the ```len()``` calculation to fix this.


disconnect3d (approved):
:heart: Nice, that was annoying for me but never checked what was the cause.

zachriggle (approved):

PR #300, opened 2017-06-28
Better exception msg

disconnect3d:
Enchances the exception display.

An example (with modified vmmap command):
https://i.gyazo.com/1e8f37a6cb4a2241fb09d2644346530c.png

Previously we just shown:
```
vmmap: 'str' object is not callable (<class 'TypeError'>)
```

### DIFF
--- a/pwndbg/elf.py
+++ b/pwndbg/elf.py
@@ -27,7 +27,6 @@ import pwndbg.memory
 import pwndbg.proc
 import pwndbg.stack
 
-
 # ELF constants
 PF_X, PF_W, PF_R = 1,2,4
 ET_EXEC, ET_DYN  = 2,3
--- a/pwndbg/events.py
+++ b/pwndbg/events.py
@@ -12,13 +12,10 @@ from __future__ import unicode_literals
 
 import functools
 import sys
-import traceback
 
 import gdb
 
-import pwndbg.color
 import pwndbg.config
-import pwndbg.exception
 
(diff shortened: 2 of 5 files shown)


zachriggle:
This seems to put *very* verbose code into the non-verbose path

disconnect3d:
Maybe `exceptions-verbose` should be `on` by default?

zachriggle:
Nah the whole point for the setting is not to bombard the users with massive amounts of data for small failures.  If something's wrong, then we suggest turning on `verbose` as part of the troubleshooting process.  We, as developers, use the `debug` setting to step through.

disconnect3d:
@zachriggle I moved the part of reporting an issue to the path of verbose/debug.

Is that okay now?

zachriggle:
Much better! ❤️  In order to avoid spamming the suggestion, we should stick that in a memoized routine so that the "Hey try this to see more info" only appears once per session for any given error message.

zachriggle:
Though we might not be able to use the `memoize` module because of cyclic imports.  In any case, we can just cache it on a `dict()` of whatever message we printed.

zachriggle:
Pwntools has a `warn_once` that does this automagically, might be something we should add

disconnect3d:
@zachriggle see now :)

disconnect3d:
@zachriggle see now; in action: https://asciinema.org/a/q2uIzDULvIyz8bl4LPzDUtZ7e

PR #303, opened 2017-07-02
Remove pwndbg/linkmap.py and malloc.py

disconnect3d:
Remove two files that are WIP since 2015.


disconnect3d:
Just not to miss some useful features - @zachriggle can you add issues for linkmap and EGLIBC malloc along with some description/resources if it is still relevant or/and not covered in any existing issues?

zachriggle:
Thanks for pointing these out! I’ll do this soon, on vacation until July 10

PR #305, opened 2017-07-07
Enhance canary command

disconnect3d:
Canary command:
* Displays telescope result of places where canaries are located
* Moved to its own file (`pwndbg/commands/canary.py`)
* Moved to `ArgparsedCommand` (as discussed in https://github.com/pwndbg/pwndbg/issues/244)

### DIFF
--- a/pwndbg/__init__.py
+++ b/pwndbg/__init__.py
@@ -18,6 +18,7 @@ import pwndbg.commands
 import pwndbg.commands.argv
 import pwndbg.commands.aslr
 import pwndbg.commands.auxv
+import pwndbg.commands.canary
 import pwndbg.commands.checksec
 import pwndbg.commands.config
 import pwndbg.commands.context
--- a/pwndbg/commands/canary.py
+++ b/pwndbg/commands/canary.py
@@ -1,0 +1,48 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import argparse
+
+import pwndbg.auxv
+import pwndbg.color
+import pwndbg.commands
+import pwndbg.commands.telescope
+import pwndbg.memory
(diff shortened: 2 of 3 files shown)


disconnect3d:
Example when canaries are there:
https://i.gyazo.com/776ae7050f9bcc1e97f92938eaa9fcf3.png

This was captured after recursing into `foo` few times in such C code:
```c
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

void foo(char* s) {
    char buf[32];
    strcpy(buf, s);
    if (atoi(buf) != 0)
        foo(buf);
}

int main(int argc, char* argv[]) {
    foo(argv[1]);
}
```
Compiled as `gcc main.c -fstack-protector-all` and run with `123` as an argument.

And this is how it looks without canaries - compiled without `-fstack-protector-all`:
https://i.gyazo.com/d23ee11cfe347c84ddd2097eff8d8504.png

pwndbg/search.py
```diff
@@ -18,18 +16,14 @@
...
-
     i = gdb.selected_inferior()
 
-    maps = pwndbg.vmmap.get()
-    hits = []
+    maps = pwndbg.vmmap.get() if not pages else pages
 
     if end and start:
-        maps = [m for m in maps if start <= m < end]
+        maps = [m for m in maps if start >= m.start and end < m.end]
```
zachriggle: This reverses the old logic, by requiring that `start` and `end` be entirely within a given mapping, rather than `start` and `end` containing a given mapping.

zachriggle:
The `search` fix should be broken into its own pull request, and target against `stable`

disconnect3d:
@zachriggle Here it is: https://github.com/pwndbg/pwndbg/pull/307

zachriggle:
Merged #307 to `stable`, and then merged through `beta` and into `dev`.

You'll probably need to rebase this.

disconnect3d:
@zachriggle Will do that in a day or two. Sorry for the delay.

PR #306, opened 2017-07-10
Fix got command

disconnect3d:
1. Fixes 'got' command not showing meaningful message in its edge case paths (e.g. binary statically linked / full RELRO).

2. Adds comment about how are the jumps extracted under the hood and what are its consequences (we don't show GOT entries from FULL RELRO).

3. Fixes `got` command not displaying got entries when one uses checksec script from pwntools
* The fix just changes `call_program` wrapper to pipe stderr to stdout

This is because there are two `checksec`s:
* [the checksec script](https://github.com/slimm609/checksec.sh)
* checksec script from pwntools

And the latter prints output into stderr and because of that we couldn't detect proper RELRO status (as got command parsed an empty string).

### DIFF
--- a/pwndbg/commands/got.py
+++ b/pwndbg/commands/got.py
@@ -13,8 +13,7 @@ import pwndbg.enhance
 import pwndbg.file
 import pwndbg.which
 import pwndbg.wrappers
-from pwndbg.color import green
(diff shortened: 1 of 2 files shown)


zachriggle (approved):
LGTM

PR #307, opened 2017-07-12
Fix search function and command

disconnect3d:
1. Search command (`pwndbg.command.search`):
* Fixes error path crashing when mapping was not found

2. Search function (`pwndbg.search`):
* Fixes flow when passing `start=<addr>, end=<addr>` - previously it failed on comparison between int and `pwndbg.memory.Page` object
* Lets specify just `start=<addr>` or `end=<addr>` - previously we if just one of them were passed, we ignored it
* Asserts whether `start < end` if both `start` & `end` are passed

3. `pwndbg.memory.Page` object:
* Adds `.start` and `.end` properties

### DIFF
--- a/pwndbg/memory.py
+++ b/pwndbg/memory.py
@@ -9,7 +9,6 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import os
-import traceback
 
 import gdb
 
@@ -22,6 +21,7 @@ import pwndbg.typeinfo
 PAGE_SIZE = 0x1000
 PAGE_MASK = ~(PAGE_SIZE-1)
 MMAP_MIN_ADDR = 0x8000
+
 
 def read(addr, count, partial=False):
     """read(addr, count, partial=False) -> bytearray
@@ -73,6 +73,7 @@ def read(addr, count, partial=False):
 
     return bytearray(result)
 
+
 def readtype(gdb_type, addr):
     """readtype(gdb_type, addr) -> int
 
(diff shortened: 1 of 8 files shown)


pwndbg/search.py
```diff
@@ -33,26 +32,29 @@ def search(searchfor, mappings=None, start=None, end=None,
...
     if writable:
         maps = [m for m in maps if m.write]
 
+    search_start, search_end = start, end
+
     for vmmap in maps:
-        start = vmmap.vaddr
-        end   = start + vmmap.memsz
+        start = vmmap.start if search_start < vmmap.start else search_start
+        end   = vmmap.end if search_end > vmmap.end else search_end
```
disconnect3d: Right now I am not sure if this is needed as there is the `length = end - start` and `if length <= 0: break` below.

pwndbg/memory.py
```diff
@@ -350,28 +383,43 @@ def __init__(self, start, size, flags, offset, objfile=''):
...
         # if self.rwx:
             # self.flags = self.flags ^ 1
+
+    @property
+    def start(self):
+        return self.vaddr
+
+    @property
+    def end(self):
+        return self.vaddr + self.memsz
```
zachriggle: Shouldn't this include `- 1`?

disconnect3d: It is exclusive - the same as displayed by `vmmap`

disconnect3d:
@zachriggle bump

zachriggle (approved):

PR #308, opened 2017-07-13
update for ida_script.py to handle ida 7.0

aidielse:


disconnect3d:
What about `idaapi.save_database` - can you check that in IDA 7?

```
Python>help(idaapi.save_database)
Help on function save_database in module idaapi:

save_database(*args)
    save_database(outfile, delete_unpacked) -> bool
```

It seems to me this might work in all versions as I guess it is the idapython api and not this old(?) idc.

disconnect3d:
@aidielse bump

zachriggle:
LGTM

ida_script.py
```diff
@@ -20,7 +20,10 @@
 dt = datetime.datetime.now().isoformat().replace(':', '-')
 
 # Save the database so nothing gets lost.
-idc.SaveBase(idc.GetIdbPath() + '.' + dt)
+if idaapi.IDA_SDK_VERSION == 700:
```
zachriggle: Might be worth doing `>= 700`

zachriggle (changes requested):

aidielse:
Sorry I haven't responded, I've been pretty busy.
@disconnect3d that works in 7.0, updated.

disconnect3d:
@aidielse Are you sure? I mean, I asked because this method is also present on IDA 6.x so we might omit having a condition for IDA version.

However it turns out it has two arguments:
```python
  File "C:\Program Files (x86)\IDA 6.8\python\idaapi.py", line 601, in IDAPython_ExecScript
    execfile(script, g)
  File "Z:/home/dc/installed/pwndbg/ida_script.py", line 25, in <module>
    idaapi.save_database(idc.GetIdbPath() + '.' + dt)
  File "C:\Program Files (x86)\IDA 6.8\python\idaapi.py", line 50712, in save_database
    return _idaapi.save_database(*args)
TypeError: save_database expected 2 arguments, got 1
```

So did they change it to have just one argument in IDA 7.x or did we miss something here?

disconnect3d:
I spoke with @aidielse on IRC and he confirmed the `idaapi.save_database` works in IDA 7.x.

Thanks!

disconnect3d (approved):

PR #309, opened 2017-07-16
Fix search function and command (#307)

disconnect3d:
* pwndbg.memory.Page: add .start and .end properties

* Fix search command mapping not found error path

* Fix pwndbg.search flow when start and/or end arg is passed

* Fix isort problems

* pwndbg.memory.Page - info about .start/.end properties


zachriggle:
Hmm, I tried resolving the merge conflicts in the web browser (new Github feature?) and I'm pretty sure it did the wrong thing

PR #317, opened 2017-08-03
checksec: cache output of command

bennofs:
I think the wrapper updates were meant to include this but it seems they don't?

### DIFF
--- a/pwndbg/wrappers/checksec.py
+++ b/pwndbg/wrappers/checksec.py
@@ -20,13 +20,9 @@ def get_raw_out():
     return pwndbg.wrappers.call_cmd(cmd)
 
 @pwndbg.wrappers.OnlyWithCommand(cmd_name)
-@pwndbg.memoize.reset_on_objfile
 def relro_status():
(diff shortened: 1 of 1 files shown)


disconnect3d:
Thanks for PR! Yeah, we meant to cache it. I must have missed it during review.

Can we actually simplify it a bit more? I mean, there is `get_raw_out` which turns out (by a fast `git grep get_raw_out` so it isn't 100% accurately) to be not used anywhere at all.

I guess it was intended to be used in both `relro_status` and `pie_status` as they actually duplicate its logic.

So... just call it there :).

degrigis:
ops, my fault, sorry for that :-) 
However `get_raw_out` is used inside the command `chekcsec` ( /pwndbg/commands/checksec.py ).

disconnect3d:
@bennofs Bump

bennofs:
@disconnect3d oops sorry forgot about it. should be ok now

disconnect3d:
@bennofs Thanks!

PR #318, opened 2017-08-03
Fix malloc chunk names

bennofs:
The fields have been renamed in newer glibc versions.

### DIFF
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -12,23 +12,20 @@ import six
 
 import pwndbg.color.memory as M
 import pwndbg.commands
+import pwndbg.typeinfo
 from pwndbg.color import bold
 from pwndbg.color import red
 from pwndbg.color import underline
 from pwndbg.color import yellow
 
 
-def value_from_type(type_name, addr):
-    gdb_type = pwndbg.typeinfo.load(type_name)
-    return gdb.Value(addr).cast(gdb_type.pointer()).dereference()
-
 def read_chunk(addr):
     renames = {
         "mchunk_size": "size",
         "mchunk_prev_size": "prev_size",
     }
-    val = value_from_type("struct malloc_chunk", addr)
-    return dict({ renames.get(key) or key: int(val[key]) for key in val.type.keys() }, value=val)
+    val = pwndbg.typeinfo.read_gdbvalue("struct malloc_chunk", addr)
+    return dict({ renames.get(key, key): int(val[key]) for key in val.type.keys() }, value=val)
 
 
 def format_bin(bins, verbose=False):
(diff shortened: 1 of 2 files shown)


pwndbg/commands/heap.py
```diff
@@ -22,6 +22,15 @@ def value_from_type(type_name, addr):
     gdb_type = pwndbg.typeinfo.load(type_name)
     return gdb.Value(addr).cast(gdb_type.pointer()).dereference()
 
+def read_chunk(addr):
+    renames = {
+        "mchunk_size": "size",
+        "mchunk_prev_size": "prev_size",
+    }
+    val = value_from_type("struct malloc_chunk", addr)
+    return dict({ renames.get(key) or key: int(val[key]) for key in val.type.keys() }, value=val)
```
disconnect3d: The `{ renames.get(key) or key: int(val[key])` looks weird/wrong.

The whole line could be simplified to:
```python
return dict({renames.get(key, key): int(val[key]) for key in val.type.keys()}, value=val)
```

zachriggle: No need for `dict()`, the curly brace enclosure does it by itself

disconnect3d: @zachriggle there is additional keyword argument which adds another item `value: val`.

pwndbg/commands/heap.py
```diff
@@ -22,6 +22,15 @@ def value_from_type(type_name, addr):
     gdb_type = pwndbg.typeinfo.load(type_name)
     return gdb.Value(addr).cast(gdb_type.pointer()).dereference()
```
disconnect3d: Might be good to move this to `pwndbg/typeinfo.py` and rename to `as_gdbtype(type_name, addr)` or `cast_to(type_name, addr)`.
@zachriggle What do you think?

disconnect3d (approved):
LGTM

disconnect3d:
@bennofs Just to **be sure it works** can you show a screenshot, text display or asciinema.org video of it working along with `version` command output?

Usually I test things before merge but the heap commands doesn't work on Arch Linux so I can't test this one :(.

bennofs:
@disconnect3d https://asciinema.org/a/LGq7buGpowkP2AGdrd7lZ7IBF 
btw, the heap commands do work on arch if you install the debugging symbols for glibc (you need to compile glibc with from ABS yourself for this though afaik) (quite useful if you're debugging exploits anyway :)

disconnect3d:
@bennofs Hmm one more thing that comes to my mind. Can we add a comment/docstring to `read_chunk` that would describe the glibc name change and possibly link to a source/diff? It might be good to provide it here as well.

bennofs:
@disconnect3d done

Issue #321, opened 2017-08-08
Search command doesn't work on remote targets


disconnect3d:
### Description

The `search` command doesnt work on remote targets.

I have researched the topic and it seems it is a bug in gdb - with both `find` command and `gdb.selected_inferior().search_memory(...)`.

I have reported it here: https://sourceware.org/bugzilla/show_bug.cgi?id=21927

The purpose of this issue is to make sure to others that we **are aware of the problem** and to know whether it is still there and so on.

disconnect3d:
I was wrong. The real bug is there **ONLY** with `gdb.selected_inferior().search_memory(...)` and I have found a workaround. PR incomming.

The above reported issue has been closed and split into two seperate ones:
* [Python API search_memory remote targets bug](https://sourceware.org/bugzilla/show_bug.cgi?id=21946)
* [Improve find command string search](https://sourceware.org/bugzilla/show_bug.cgi?id=21945)

disconnect3d:
Fixed in #322.

PR #322, opened 2017-08-11
GDB remote targets search_memory bug workaround

disconnect3d:
Applies a workaround for remote target search command described in https://sourceware.org/bugzilla/show_bug.cgi?id=21946

The issue has been described in #321

### DIFF
--- a/pwndbg/__init__.py
+++ b/pwndbg/__init__.py
@@ -144,3 +144,10 @@ try:
     gdb.execute("set disassembly-flavor intel")
 except gdb.error:
     pass
+
+# Workaround for gdb bug described in #321 ( https://github.com/pwndbg/pwndbg/issues/321 )
+# More info: https://sourceware.org/bugzilla/show_bug.cgi?id=21946
+# As stated on GDB's bugzilla that makes remote target search slower.
+# After GDB gets the fix, we should disable this only for bugged GDB versions.
+if 1:
+    gdb.execute('set remote search-memory-packet off')


zachriggle (commented):
Does the Search command still work in Qemu with this change?

disconnect3d:
@zachriggle it seems it does https://asciinema.org/a/ULjl8FimibZjCpPSCjBnvGst6

But... it also works without it, e.g. on dev:
```
pwndbg> version
Gdb: GNU gdb (GDB) 8.0
Python: 3.6.2 (default, Jul 20 2017, 03:52:27)  [GCC 7.1.1 20170630]
Pwndbg: 1.0.0 build: f481201
```

I believe the problem stays only for pure gdbserver. Not sure how qemu exposes gdbserver-like interface.

disconnect3d:
@zachriggle I believe the problem stays only for gdbserver. Not sure how qemu exposes gdbserver-like interface.

disconnect3d:
@zachriggle Bump

zachriggle (approved):

sultanakam (commented):
@@ -144,3 +144,10 @@
    gdb.execute("set disassembly-flavor intel")
except gdb.error:
    pass

 # Workaround for gdb bug described in #321 ( https://github.com/pwndbg/pwndbg/issues/321 )
# More info: https://sourceware.org/bugzilla/show_bug.cgi?id=21946
# As stated on GDB's bugzilla that makes remote target search slower.
# After GDB gets the fix, we should disable this only for bugged GDB versions.
if 1:
    gdb.execute('set remote search-memory-packet off')

hackerzaxo (commented):
Hi

treycove (commented):
free blookets

pwndbg/__init__.py
```diff
@@ -144,3 +144,10 @@
     gdb.execute("set disassembly-flavor intel")
```
jlandon638, line 144: (??><^&*********************************************<>)

Issue #299, opened 2017-06-23
Context failing on breakpoints with commands


yoava333:
### Description
When using commands in breakpoints with pwndbg the debugger will halt the execution of the debugee and will not respond to keyboard commands.
The same gdb script works without pwndbg (e.g. commenting out in ~/.gdbinit)

### Steps to reproduce
wrote the following gdb script (gdb.cmds):
```
b malloc
commands
    printf "size = %d\n", $rdi
    c
end

```

run the following command:
gdb /usr/bin/python -x gdb.cmds

and type "r"

the output
```
user@ubuntu /tmp> gdb /usr/bin/python -x gdb.cmds
GNU gdb (Ubuntu 7.11.1-0ubuntu1~16.04) 7.11.1
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Loaded 108 commands. Type pwndbg [filter] for a list.
Reading symbols from /usr/bin/python...(no debugging symbols found)...done.
Breakpoint 1 at 0x4168e0
pwndbg> r
Starting program: /usr/bin/python

Breakpoint 1, malloc (n=1425) at dl-minimal.c:94
94      dl-minimal.c: No such file or directory.
size = 1425



^C^C
```

BTW if you add the "r" commands to the end of gdb.cmds it works fine.

### My setup
Ubuntu 16.04.02 
pwndbg - dae257cdf6fe0ff83440308eeab225fa4d216e44
GNU gdb (Ubuntu 7.11.1-0ubuntu1~16.04) 7.11.1
python - 3.5.2 (default, Nov 17 2016, 17:05:23) [GCC 5.4.0 20160609]

grant-h:
I am having this exact issue and it is preventing me from debugging function calls in the same way. This is a serious bug if any scripting for silent, print-only breakpoints is being done.

[...]

PR #324, opened 2017-08-16
Fix context display when thread is not stopped

disconnect3d:
It seems that we try to display context when we can't do it - e.g. in situation where `command` is added to a breakpoint and it is gdb handling the breakpoint itself.

See https://github.com/pwndbg/pwndbg/issues/299 for more info.

### DIFF
--- a/pwndbg/proc.py
+++ b/pwndbg/proc.py
@@ -2,7 +2,8 @@
 # -*- coding: utf-8 -*-
 """
 Provides values which would be available from /proc which
-are not fulfilled by other modules.
+are not fulfilled by other modules and some process/gdb flow
+related information.
 """
 from __future__ import absolute_import
 from __future__ import division
@@ -48,6 +49,18 @@ class module(ModuleType):
         return gdb.selected_thread() is not None
 
     @property
+    def thread_is_stopped(self):
+        """
+        This detects whether selected thread is stopped. 
+        It is not stopped in situations when gdb is executing commands 
+        that are attached to a breakpoint by `command` command.
+
+        For more info see issue #229 ( https://github.com/pwndbg/pwndbg/issues/299 )
+        :return: Whether gdb executes commands attached to bp with `command` command.
+        """
+        return gdb.selected_thread().is_stopped()
+
+    @property
     def exe(self):
         for obj in gdb.objfiles():
             if obj.filename:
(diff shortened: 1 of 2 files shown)


disconnect3d:
Below you can see how that works after the fix:
```
[dc@dc:test]$ cat test.c
#include <stdlib.h>

int main() {
    malloc(100);
}
[dc@dc:test]$ cat a.cmds 
b malloc
commands
    printf "size = %d\n", $rdi
    c
end

[dc@dc:test]$ gdb ./a.out -x a.cmds
Breakpoint 1 at 0x560
pwndbg> r
Starting program: /home/dc/test/a.out 
size = 1425
Selected thread is running.
pwndbg> size = 1169
Selected thread is running.
 ////////////////// <similar lines truncated as we got a lot of those size=~~ lines>
pwndbg> size = 100
Selected thread is running.
pwndbg> [Inferior 1 (process 2236) exited normally]
```

The standard debugging works as well.

disconnect3d:
The thing that is weird to me is that our prompt is shown/executed and `Selected thread is running` when it is gdb executing the command.

Why is that so?

If I disable pwndbg and launch this example as:
```
$ gdb ./a.out -x a.cmds
Breakpoint 1 at 0x560
(gdb) pi
>>> import gdb
>>> def prompt_hook(*a):
...     print("Hello", a)
... 
>>> gdb.prompt_hook = prompt_hook
>>> 
Hello ('(gdb) ',)
(gdb) run
Starting program: /home/dc/test/a.out 

Breakpoint 1, 0x00007f2ca25c0c30 in malloc () from /lib64/ld-linux-x86-64.so.2
size = 1425

 ////////////////// <similar lines truncated as we got a lot of those size=~~ lines>

Breakpoint 1, 0x00007f2ca2280630 in malloc () from /usr/lib/libc.so.6
size = 100

Breakpoint 1, 0x00007f2ca2280630 in malloc () from /usr/lib/libc.so.6
size = 100
[Inferior 1 (process 2393) exited normally]
Hello ('(gdb) ',)
(gdb) 
```

I get the `prompt_hook` being executed **just once** while our `prompt_hook` in pwndbg gets executed each time the breakpoint is hit & consumed by gdb. Why is that so?

disconnect3d:
@zachriggle bump

zachriggle:
This might make things more difficult for live debugging, but I don't think I've ever actually done that, so I don't care.

zachriggle (approved):

disconnect3d:
> This might make things more difficult for live debugging

What do you mean? The change itself? You mean some unhandled cases that would be bashed by this?

I agree that seeing `Selected thread is running.` is annoying but it is still better then the fails described in #299. I would leave it as it is since I have no idea how to skip that text and prompt being displayed.

PS: Any idea why travis hasn't run? [can't merge]

zachriggle:
Yeah, I'm not sure about Travis.  You might need to close this PR and open a different one *shrug*

PR #326, opened 2017-08-22
Dumpargs add --force to show all possible register arguments

disconnect3d:
Currently the `dumpargs` command shows an empty line if it fails to resolve list of call arguments [or we are not on a call instruction].

This PR makes so that:
* Changes `dumpargs` into `ArgparsedCommand` (see #244)
* Adds possibility to dump all register arguments by adding `--force`, even if we failed to resolve them from metadatas (symbols, syscalls, IDA Pro integration)0
* Improves UX - if no call arguments were resolved, we now inform user about it

It also refactors a bit `pwndbg.arguments` and adds `arguments` function there (I don't like it returns different thing than just `argument` but I am not sure how to do it better)

You can see an example usage below:
```
pwndbg> dumpargs
Couldn't resolve call arguments. Maybe the function doesn't take any?
Use `dumpargs --force` to force the display.

pwndbg> ni
0x0000000000400a24 in ?? ()

pwndbg> dumpargs
Cannot dump args as current instruction is not a call.
Use `dumpargs --force` to force the display.

pwndbg> dumpargs --force
 rdi = 0x7fffffffd650 ◂— 0x8b02757117eed833
 rsi = 0xe
 rdx = 0x7fffffffd65f ◂— 0x7fffffffd750d0
 rcx = 0xf0
  r8 = 0xd
  r9 = 0x7f46623e4830 (_dl_fini) ◂— push   rbp
```

### DIFF
--- a/pwndbg/arguments.py
+++ b/pwndbg/arguments.py
@@ -52,8 +52,9 @@ ida_replacements = {
     '__userpurge': '',
 }
 
+
 def get_syscall_name(instruction):
-    if not CS_GRP_INT in instruction.groups:
+    if CS_GRP_INT not in instruction.groups:
         return None
 
     try:
@@ -64,6 +65,7 @@ def get_syscall_name(instruction):
         return 'SYS_' + name
     except:
         return None
+
 
 def get(instruction):
     """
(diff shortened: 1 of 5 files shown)


disconnect3d:
**Please do not merge yet**

disconnect3d:
@zachriggle bump

zachriggle (approved):
Looks good to me :)

PR #328, opened 2017-08-23
Fix IDA Pro decompiled code not being displayed

disconnect3d:
The idea of this PR is to fix IDA Pro decompiled code display whenever Hexrays is available.

**Summary:**
* Fixes `withHexrays` not returning wrapped func
* Fixes IDA Pro xmlrpc failing on marshalling `idaapi.decompile` result - `idaapi.cfuncptr_t` instance - by adding a marshaller for it (also escaping proper XML symbols)
* Fixes IDA Pro xmlrpc server skipping wrapped function exception (instead of getting proper exception we just get `IndexError` from this line: https://github.com/pwndbg/pwndbg/blob/dev/ida_script.py#L62 on the pwndbg side which wasn't the real reason of failure)
* Adds IDA Pro xmlrpc server `shutdown` function which can be used to kill xmlrpc server and let user reload the script (without using it one might get e.g. some Windows related errors)
* Refactors a little bit the `context.py`
* Adds `HEXRAYS PSEUDOCODE` banner for decompiled code display

---
The first problem that occurs here is that we failed with adding [`withHexrays` decorator](https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/ida.py#L77-L82) as it doesn't return the wrapper function.

We failed to spot it due to lack of testing it and the fact [we already catch all exceptions in there](https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/commands/context.py#L207-L212) - so we failed on `'NoneType' object is not callable`, heh.

This is already fixed in this PR. However another problem occurs (I had to comment out the `try:`, `except:`):
```
pwndbg> r
Starting program: /home/dc/test/a.out 

Breakpoint 1, 0x000000010000070e in main ()
  File "/home/dc/installed/pwndbg/pwndbg/commands/__init__.py", line 100, in __call__
  File "/home/dc/installed/pwndbg/pwndbg/commands/__init__.py", line 192, in _OnlyWhenRunning
    return function(*a, **kw)
  File "/home/dc/installed/pwndbg/pwndbg/commands/context.py", line 64, in context
    result.extend(func())
  File "/home/dc/installed/pwndbg/pwndbg/commands/context.py", line 209, in context_code
    lines = pwndbg.ida.decompile(addr)
  File "/home/dc/installed/pwndbg/pwndbg/ida.py", line 74, in __call__
    return self.fn(*args, **kwargs)
  File "/home/dc/installed/pwndbg/pwndbg/ida.py", line 83, in wrapper
    return f(*a, **kw)
  File "/home/dc/installed/pwndbg/pwndbg/ida.py", line 91, in wrapper
    return function(l2r(address), *args, **kwargs)
  File "/home/dc/installed/pwndbg/pwndbg/memoize.py", line 48, in __call__
  File "/home/dc/installed/pwndbg/pwndbg/ida.py", line 339, in decompile
    return _ida.decompile(addr)
  File "/usr/lib/python3.6/xmlrpc/client.py", line 1112, in __call__
    return self.__send(self.__name, args)
  File "/usr/lib/python3.6/xmlrpc/client.py", line 1452, in __request
    verbose=self.__verbose
  File "/usr/lib/python3.6/xmlrpc/client.py", line 1154, in request
    return self.single_request(host, handler, request_body, verbose)
  File "/usr/lib/python3.6/xmlrpc/client.py", line 1170, in single_request
    return self.parse_response(resp)
  File "/usr/lib/python3.6/xmlrpc/client.py", line 1342, in parse_response
    return u.close()
  File "/usr/lib/python3.6/xmlrpc/client.py", line 656, in close
    raise Fault(**self._stack[0])
xmlrpc.client.Fault: <Fault 1: "<type 'exceptions.TypeError'>:cannot marshal <type 'SwigPyObject'> objects">
```

If anyone is willing to help with this, you are welcome.

### DIFF
--- a/ida_script.py
+++ b/ida_script.py
@@ -109,8 +109,18 @@ register_module(idaapi)
 server.register_function(lambda a: eval(a, globals(), locals()), 'eval')
 server.register_introspection_functions()
 
-print('Ida Pro xmlrpc hosted on http://%s:%s' % (host, port))
+print('IDA Pro xmlrpc hosted on http://%s:%s' % (host, port))
+print('Call `shutdown()` to shutdown the IDA Pro xmlrpc server.')
 
 thread = threading.Thread(target=server.serve_forever)
 thread.daemon = True
 thread.start()
+
+
+def shutdown():
+    global server
+    global thread
+    server.shutdown()
+    server.server_close()
+    del server
+    del thread
--- a/ida_script.py
+++ b/ida_script.py
@@ -6,6 +6,7 @@ import datetime
 import threading
 import xmlrpclib
 from SimpleXMLRPCServer import SimpleXMLRPCServer
+from xml.sax.saxutils import escape
 
 import idaapi
 import idautils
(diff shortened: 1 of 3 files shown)


disconnect3d:
This is because `decompile` returns `cfuncptr_t` instead of a `str`.
```Python>a=idaapi.decompile(0x70A)
Python>type(a)
<class 'idaapi.cfuncptr_t'>
Python>str(a)
int __cdecl __noreturn main(int argc, const char **argv, const char **envp)
{
  char buf; // [sp+10h] [bp-110h]@2
  __int64 v4; // [sp+118h] [bp-8h]@1

  v4 = *MK_FP(__FS__, 40LL);
  while ( 1 )
  {
    if ( (signed int)read(0, &buf, 0x100uLL) <= 0 )
      puts("Not read");
    printf("read '%s' <-buf\n", &buf);
  }
}```

disconnect3d:
@bennofs Would be nice if you can review this as well since you deal with IDA Pro integration too.

zachriggle:
Can you give an example of what the output is before and after this change? Thanks!

zachriggle (approved):
Overall looks good to me, just want to see how this ends up looking

disconnect3d:
@zachriggle Before all of the changes we got no output from decompiled code as we broke the decorator at some point and skipped all exceptions for that part.

After fixing this, we got the decompiled code output but without banner for it, so it was displayed in under disasm banner (and disasm itself).

Currently it looks like this: https://i.gyazo.com/b1620f70dfb9bf0d90e48d982265a510.png

Still might require some tweaking if possible (like mark current pseudocode line if this can be retrieved from IDA) but it is working good enough for now.

zachriggle:
My intent for the original layout was to intersperse assembly and Hexrays code.  I like this a little bit better, as long as we can highlight the "current line" of code.

PR #329, opened 2017-08-23
Fix exception if there is an indirect jump

bennofs:
This is a simple typo, but the error message that GDB gave was interesting:

Previously, if you stopped on an instruction that does an indirect jump, like
this:

```
jmp [ecx*4 + 0xdeadbeef]
```

then pwndbg would the following exception:

```
gdb.error: evaluation of this expression requires the program to have a function "malloc".
```

The reason is that the code used `memory_sz` and passed that to gdb.Value, thus
creating a string value. When casting the string to a pointer later, GDB tries
to allocate a string in the inferior which failed since malloc is not available.

The fix is, of course, to use the correct function (`memory`) that returns an
int and not a string.

### DIFF
--- a/pwndbg/disasm/arch.py
+++ b/pwndbg/disasm/arch.py
@@ -143,8 +143,11 @@ class DisassemblyAssistant(object):
             addr &= pwndbg.arch.ptrmask
         if op.type == CS_OP_MEM:
             if addr is None:
-                addr = self.memory_sz(instruction, op)
-            addr = int(pwndbg.memory.poi(pwndbg.typeinfo.ppvoid, addr))
+                addr = self.memory(instruction, op)
+
+            # self.memory may return none, so we need to check it here again
+            if addr is not None:
+                addr = int(pwndbg.memory.poi(pwndbg.typeinfo.ppvoid, addr))
         if op.type == CS_OP_REG:
             addr = self.register(instruction, op)
 


disconnect3d:
Can you provide an example (e.g. C with inline asm or just asm code & the command to compile it) that can be used to reproduce and check this behavior?

zachriggle (approved):
LGTM

bennofs:
@disconnect3d https://gist.github.com/bennofs/127f1650fd4b26ecbe7e9fb1881c4261 

```
$ pwndbg ./test
GNU gdb (GDB) 8.0
Copyright (C) 2017 Free Software Foundation, Inc.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-pc-linux-gnu".
<http://www.gnu.org/software/gdb/bugs/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Loaded 109 commands. Type pwndbg [filter] for a list.
Reading symbols from ./test...(no debugging symbols found)...done.
pwndbg> entry
Temporary breakpoint 1 at 0x400088

Temporary breakpoint 1, 0x0000000000400088 in _start ()
Exception occured: nearpc: evaluation of this expression requires the program to have a function "malloc". (<class 'gdb.error'>)

Exception occured: context: object of type 'NoneType' has no len() (<class 'TypeError'>)
pwndbg> 
```

PR #330, opened 2017-08-24
Fix hard error when something else (not IDA) listens on IDA's port

bennofs:
The default IDA port is 8888 and it can happen that some other program (such as
a jupyter notebook) is listening on that address. This made pwndbg unusable,
because it would crash trying to connect to IDA.

It now prints a message. I also added some code to avoid printing the same exception twice (pwndbg will try to connect to ida after every command, printing the same message each time would be a little bit annoying)

### DIFF
--- a/pwndbg/ida.py
+++ b/pwndbg/ida.py
@@ -13,6 +13,7 @@ from __future__ import unicode_literals
 import errno
 import functools
 import socket
+import sys
 import traceback
 
 import gdb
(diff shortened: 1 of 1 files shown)


disconnect3d:
Hey,

I was thinking about this today.

This still doesn't fix the hang problem when launching pwndbg and something listens on 8888.
The pretty much easiest solution for this is to just add a timeout - like 3, 5 or 10 seconds:
```python
[dc@dc:pwndbg|dev *$%]$ git diff
diff --git a/pwndbg/ida.py b/pwndbg/ida.py
index 61421b4..06eabfa 100644
--- a/pwndbg/ida.py
+++ b/pwndbg/ida.py
@@ -52,6 +52,7 @@ def init_ida_rpc_client():
     addr = 'http://{host}:{port}'.format(host=ida_rpc_host, port=ida_rpc_port)
 
     _ida = xmlrpclib.ServerProxy(addr)
+    socket.setdefaulttimeout(3)
 
     try:
         _ida.here()
```

The hang problem can be reproduced by hosting a server that doesn't respond **anything** e.g. by invoking `nc -l -p 8888`.

pwndbg/ida.py
```diff
@@ -45,22 +46,36 @@
...
+_ida_last_exception = None
+
 
 @pwndbg.config.Trigger([ida_rpc_host, ida_rpc_port])
 def init_ida_rpc_client():
-    global _ida
+    global _ida, _ida_last_exception
     addr = 'http://{host}:{port}'.format(host=ida_rpc_host, port=ida_rpc_port)
 
     _ida = xmlrpclib.ServerProxy(addr)
```
disconnect3d, line 58: Please add a timeout :).

zachriggle (approved):
LGTM

bennofs:
@disconnect3d done

disconnect3d:
@bennofs Thx! Will check the asm code from ur 2nd PR later on.

PR #334, opened 2017-08-30
An attempt to fix dt command

semihalf-oleksy-michalina:
This is an attempt to fix dt command. It still lacks the form of an ArgparsedCommand and does not print info on all types, just the structures.

**Changes (based on pahole-gdb):**

- `o     = getattr(field, 'bitpos', 0)//8` - Python3 division
- printing unused bytes
- printing used bytes

Simple testcase:

```
typedef struct str_type {
	int num;
	int* ptr;
	inside_str* new_inside_str;
	char letter;
	struct str_type* rec;
}str_type;
```

For a structure like the one listed above the result will be following:

```
pwndbg> dt str_type
type str_type:

used bytes:   4    +0x0000 num                  : int
[!--- 32 bits of padding ---!]
used bytes:   8    +0x0008 ptr                  : int *
used bytes:   8    +0x0010 new_inside_str       : inside_str *
used bytes:   1    +0x0018 letter               : char
[!--- 56 bits of padding ---!]
used bytes:   8    +0x0020 rec                  : struct str_type *
pwndbg> 

```

### DIFF
--- a/pwndbg/dt.py
+++ b/pwndbg/dt.py
@@ -72,7 +72,7 @@ def happy(typename):
     'double': 'double'
     }[typename]
 
-def dt(name='', addr=None, obj = None):
+def dt(name='', addr=None, obj=None):
     """
     Dump out a structure type Windbg style.
     """
@@ -100,21 +100,28 @@ def dt(name='', addr=None, obj = None):
     if addr is not None:
         obj = pwndbg.memory.poi(t, addr)
 
-    # Header, optionally include the name
-    header = name
-    if obj: header = "%s @ %s" % (header, hex(int(obj.address)))
-    rv.append(header)
-
     if t.strip_typedefs().code == gdb.TYPE_CODE_ARRAY:
         return "Arrays not supported yet"
     if t.strip_typedefs().code not in (gdb.TYPE_CODE_STRUCT, gdb.TYPE_CODE_UNION):
         t = {name: obj or gdb.Value(0).cast(t)}
 
+    # Header, optionally include the name
+    header = name
+    if obj: header = "%s @ %s" % (header, hex(int(obj.address)))
+    rv.append(header)
+
+    endpos = 0
     for name, field in t.items():
         # Offset into the parent structure
-        o     = getattr(field, 'bitpos', 0)/8
+        o     = getattr(field, 'bitpos', 0)//8
         extra = str(field.type)
         ftype = field.type.strip_typedefs()
+
+        hole = 0
+        if endpos < field.bitpos:
+            hole = field.bitpos - endpos
+        if hole:
+            rv.append("[!--- %d bits of padding ---!]" % hole)
 
         if obj and obj.type.strip_typedefs().code in (gdb.TYPE_CODE_STRUCT, gdb.TYPE_CODE_UNION):
             v  = obj[name]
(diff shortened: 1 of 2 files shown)


pwndbg/dt.py
```diff
@@ -100,22 +100,29 @@ def dt(name='', addr=None, obj = None):
...
-    rv.append(header)
-
     if t.strip_typedefs().code == gdb.TYPE_CODE_ARRAY:
         return "Arrays not supported yet"
     if t.strip_typedefs().code not in (gdb.TYPE_CODE_STRUCT, gdb.TYPE_CODE_UNION):
         t = {name: obj or gdb.Value(0).cast(t)}
 
+    # Header, optionally include the name
+    header = name
+    if obj: header = "%s @ %s" % (header, hex(int(obj.address)))
```
disconnect3d, line 110: Lets make it more simple: 
```python
'%s @ 0x%x' % (header, int(obj.address))
```

 I wonder if `int()` is needed at all.

pwndbg/dt.py
```diff
@@ -100,22 +100,29 @@ def dt(name='', addr=None, obj = None):
...
+    rv.append(header)
```
disconnect3d, line 111: Can we make it:
```python
# Return value is a list of strings we join at the end
rv = [header]
```

pwndbg/dt.py
```diff
@@ -100,22 +100,29 @@ def dt(name='', addr=None, obj = None):
...
+
+    endpos = 0
     for name, field in t.items():
         # Offset into the parent structure
-        o     = getattr(field, 'bitpos', 0)/8
+        o     = getattr(field, 'bitpos', 0)//8
```
disconnect3d, line 116: The `8` here is used like two times and is a magic number we can move into some constant e.g. `BITS_IN_BYTE = 8`.

pwndbg/dt.py
```diff
@@ -137,7 +144,17 @@ def dt(name='', addr=None, obj = None):
...
+                fieldsize = 0
+            else:
+                fieldsize = 8 * ftype.sizeof
+        endpos = field.bitpos + fieldsize
+
+        line  = "used bytes: %3d    +0x%04x %-20s : %s" % (fieldsize//8, o, name, extra)
         rv.append(line)
 
+    rv[0] ='type ' + rv[0] + ':\n'
     return ('\n'.join(rv))
```
disconnect3d, line 160: Just `return '\n'.join(rv)`

disconnect3d:
@zachriggle what do u think about the output? Personally I like the way it explicitly says about padding :P.
(Should we really push it to have exactly the same output as windbg's dt?)

pwndbg/dt.py
zachriggle: Meh.  The field is called `bitpos`, the constant is 8.  If you're concerned about future DEFCON 9-bit-byte, 27-bit-word type things, I don't plan to support them in Pwndbg 😛

zachriggle:
Can we hide the padding behind a flag?

zachriggle:
Or generate pseudo-names for the padding?  e.g. `__pad0` `__pad1` etc?

zachriggle:
IDA Pro has some naming mechanism for `struct` gaps.  We might be able to use that, too.

An alternate naming mechanism might be `__gapXX_Y` where `XX` is the hex offset of the field and `Y` is the length of the gap.

disconnect3d:
We can even go with padding behind the flag or a config parameter.
Both naming conventions -  `__padX` and `__gapXX_Y` - are fine for me.

pwndbg/dt.py
zachriggle: #yolo

disconnect3d:
@semihalf-oleksy-michalina bump, any chances of pushing it forward?
We should either put that behind a flag or - probably better - create a fake fields named `__gapXX_Y`.

disconnect3d:
Talked with @semihalf-oleksy-michalina in real life and it isn't going to be pushed forward. Because of that I am closing it for now.

PR #335, opened 2017-09-06
ArgparsedCommand: config[file], theme[file]

disconnect3d:
Moves `config`, `configfile`, `theme` and `themefile` commands to `ArgparsedCommand`s (see https://github.com/pwndbg/pwndbg/issues/244).

Btw: one weird thing I have found is that when you add a flag like `--show-all`, invoking `--s` works for it oO'. Is it okay for us?

### DIFF
--- a/pwndbg/commands/config.py
+++ b/pwndbg/commands/config.py
@@ -8,6 +8,8 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import argparse
+
 import pwndbg.commands
 import pwndbg.config
 from pwndbg.color import light_yellow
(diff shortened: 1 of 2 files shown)


zachriggle:
I don't really mind, but is there any benefit to changing a command which takes no arguments into an `ArgparsedCommand`?

disconnect3d:
Not really. Do you suggest to keep `ArgparsedCommand` and something more? e.g. `Command`?

zachriggle:
Nope, no preference, I just wasn't sure if there was something subtle I was missing.

LGTM!

zachriggle (approved):

PR #336, opened 2017-09-11
Fix Python<=2.7.6 "TypeError: Struct() argument 1 must be string, not unicode"

zachriggle:
Additional information is available here: http://python-future.org/stdlib_incompatibilities.html#struct-pack

Basically, `struct.pack(u"I", ...)` will fail on Python 2.7.6 or older, which means everything explodes immediately.  For some silly reasons, I ended up with exactly that:

```
$ gdb --batch --nh --nx -q --ex 'pi import platform; print(platform.python_version())'
2.7.6
```

Since this is a super small bug that hopefully nobody runs into, I'm just going to merge this to dev.

Created a Pull Request only to track that it was fixed.


zachriggle:
`isort` failed but has no output.  Tried re-running.  Is there something special I need to do, @disconnect3d?

disconnect3d:
Maybe we should have our own pack/unpack wrapper similar to pwtools `pwn.pXX`, huh...
Anyway, LGTM

PR #337, opened 2017-09-11
Expose IDA Pro commands, even when IDA is not available.

### LINKED ISSUE #225: Display IDA commands in pwndbg commands list when IDA rpc server is offline
Currently IDA Pro commands are hidden if IDA is not available. Because of that, those commands are not displayed when someone is looking for features/commands through `pwndbg ida`.

As @zachriggle suggested in https://github.com/pwndbg/pwndbg/pull/223#issuecomment-294612888 we should just re-load the module when the connection to IDA succeeded.

zachriggle:
This also has the effect of exposing the `up` and `down` commands overloads.

Closes #225

### DIFF
--- a/pwndbg/commands/ida.py
+++ b/pwndbg/commands/ida.py
@@ -32,87 +32,89 @@ def j(*args):
         pass
 
 
-if pwndbg.ida.available():
-    @pwndbg.commands.Command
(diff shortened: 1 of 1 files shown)


disconnect3d (approved):
LGTM

PR #360, opened 2017-11-18
Add nextproginstr command

disconnect3d:
Adds a command that let us "wait until we are in the program's code" (useful when e.g. catching a syscall like `catch syscall read` and then you want to wait till read and some library code is over).

This probably can be coded better - probably `from_tty=False, to_string=False` has to be added into `gdb.execute`.
I will play more with it later on.


disconnect3d:
@zachriggle I've rebased the code, pls review

PR #364, opened 2017-11-27
fix ds and da for gdb 7.11

devneal:
Makes the `ds` and `da` commands work with gdb 7.11, and implements them with the `ArgparsedCommand` decorator.

### DIFF
--- a/pwndbg/commands/windbg.py
+++ b/pwndbg/commands/windbg.py
@@ -220,12 +220,11 @@ def dqs(*a):
 da_parser = argparse.ArgumentParser()
 da_parser.description = 'Dump a string at the specified address.'
 da_parser.add_argument('address', help='Address to dump')
+da_parser.add_argument('max', type=int, nargs='?', default=256,
+                       help='Maximum string length')
 @pwndbg.commands.ArgparsedCommand(da_parser)
 @pwndbg.commands.OnlyWhenRunning
-def da(address, max=256):
-    """
-    Dump a string at the specified address.
-    """
+def da(address, max):
     address = int(address)
     address &= pwndbg.arch.ptrmask
     print("%x" % address, repr(pwndbg.strings.get(address, max)))
@@ -233,12 +232,11 @@ def da(address, max=256):
 ds_parser = argparse.ArgumentParser()
 ds_parser.description = 'Dump a string at the specified address.'
 ds_parser.add_argument('address', help='Address to dump')
+ds_parser.add_argument('max', type=int, nargs='?', default=256,
+                       help='Maximum string length')
 @pwndbg.commands.ArgparsedCommand(ds_parser)
 @pwndbg.commands.OnlyWhenRunning
-def ds(address, max=256):
-    """
-    Dump a string at the specified address.
-    """
+def ds(address, max):
     address = int(address)
     address &= pwndbg.arch.ptrmask
     print("%x" % address, repr(pwndbg.strings.get(address, max)))
--- a/pwndbg/commands/windbg.py
+++ b/pwndbg/commands/windbg.py
@@ -8,6 +8,7 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import argparse
 import codecs
 import math
 import sys
@@ -216,20 +217,30 @@ def dqs(*a):
     return pwndbg.commands.telescope.telescope(*a)
 
 
-@pwndbg.commands.ParsedCommand
+da_parser = argparse.ArgumentParser()
+da_parser.description = 'Dump a string at the specified address.'
+da_parser.add_argument('address', help='Address to dump')
+@pwndbg.commands.ArgparsedCommand(da_parser)
 @pwndbg.commands.OnlyWhenRunning
 def da(address, max=256):
     """
     Dump a string at the specified address.
     """
+    address = int(address)
+    address &= pwndbg.arch.ptrmask
     print("%x" % address, repr(pwndbg.strings.get(address, max)))
 
-@pwndbg.commands.ParsedCommand
+ds_parser = argparse.ArgumentParser()
+ds_parser.description = 'Dump a string at the specified address.'
+ds_parser.add_argument('address', help='Address to dump')
+@pwndbg.commands.ArgparsedCommand(ds_parser)
 @pwndbg.commands.OnlyWhenRunning
 def ds(address, max=256):
     """
     Dump a string at the specified address.
     """
+    address = int(address)
+    address &= pwndbg.arch.ptrmask
     print("%x" % address, repr(pwndbg.strings.get(address, max)))
 
 @pwndbg.commands.ParsedCommand


pwndbg/commands/windbg.py
```diff
@@ -216,20 +217,30 @@ def dqs(*a):
...
 
-@pwndbg.commands.ParsedCommand
+da_parser = argparse.ArgumentParser()
+da_parser.description = 'Dump a string at the specified address.'
+da_parser.add_argument('address', help='Address to dump')
+@pwndbg.commands.ArgparsedCommand(da_parser)
 @pwndbg.commands.OnlyWhenRunning
 def da(address, max=256):
     """
     Dump a string at the specified address.
```
disconnect3d: Remove the docstring as it overriden by `_ArgparsedCommand.__init__` anyway.

pwndbg/commands/windbg.py
```diff
@@ -216,20 +217,30 @@ def dqs(*a):
...
 
-@pwndbg.commands.ParsedCommand
+ds_parser = argparse.ArgumentParser()
+ds_parser.description = 'Dump a string at the specified address.'
+ds_parser.add_argument('address', help='Address to dump')
+@pwndbg.commands.ArgparsedCommand(ds_parser)
 @pwndbg.commands.OnlyWhenRunning
 def ds(address, max=256):
     """
     Dump a string at the specified address.
```
disconnect3d: Remove the docstring as it overriden by `_ArgparsedCommand.__init__` anyway.

pwndbg/commands/windbg.py
```diff
@@ -216,20 +217,30 @@ def dqs(*a):
     return pwndbg.commands.telescope.telescope(*a)
 
 
-@pwndbg.commands.ParsedCommand
+da_parser = argparse.ArgumentParser()
+da_parser.description = 'Dump a string at the specified address.'
+da_parser.add_argument('address', help='Address to dump')
```
disconnect3d, line 222: We are missing the `max` argument here.

pwndbg/commands/windbg.py
```diff
@@ -216,20 +217,30 @@ def dqs(*a):
...
     Dump a string at the specified address.
     """
+    address = int(address)
+    address &= pwndbg.arch.ptrmask
     print("%x" % address, repr(pwndbg.strings.get(address, max)))
 
-@pwndbg.commands.ParsedCommand
+ds_parser = argparse.ArgumentParser()
+ds_parser.description = 'Dump a string at the specified address.'
+ds_parser.add_argument('address', help='Address to dump')
```
disconnect3d, line 234: We are missing the `max` argument here.

disconnect3d:
**Apart from the comments I have made there is one more thing we should discuss/rethink/redo:
it might not be so obvious for the user that he/she passed wrong argument.**

---

First of all we probably shouldn't do:
```python
address = int(address)
address &= pwndbg.arch.ptrmask
```
as we don't want to repeat ourselves - this is some generic code that would apply for all commands that expects an address.

Also this is not 100% correct. Since we cast the `address` to `int` one can pass a string with valid integer (note: such cast wouldn't normally work for the example below (as it is in base 16) [but we have a hook for `int` to support such casts in some places](https://github.com/devneal/pwndbg/blob/00aa74d2bf032fc7a0470fcfa137b8c9b5882b27/pwndbg/inthook.py)):
```
pwndbg> ds "0x7fffffffda1e"
7fffffffda1e '/usr/bin/ls'
```
We should probably display usage and say that an address or expression resolving to address/number (?) is expected. 

---

Offtopic funfact - it is possible to create a breakpoint over a dereferenced string in gdb:
```
pwndbg> b *"0x7fffffffd608"
Breakpoint 3 at 0x7ffff7fa31d0
```
---

Back to the topic: currently the `address` argument type is set by [`ArgparsedCommand.__init__`](https://github.com/devneal/pwndbg/blob/00aa74d2bf032fc7a0470fcfa137b8c9b5882b27/pwndbg/commands/__init__.py#L217-L228) to `fix_int_reraise`:
```python
    def __init__(self, parser):
        self.parser = parser

        # We want to run all integer and otherwise-unspecified arguments
        # through fix() so that GDB parses it.
        for action in parser._actions:
            if action.dest == 'help':
                continue
            if action.type in (int, None):
                action.type = fix_int_reraise
```

The `fix_int_reraise` calls [`fix` which tries `gdb.parse_and_eval`](https://github.com/devneal/pwndbg/blob/00aa74d2bf032fc7a0470fcfa137b8c9b5882b27/pwndbg/commands/__init__.py#L132-L167) and if that fails - it reraises the exception (because of `reraise=True`).

Because of that when we pass something that doesn't parse well, we get such output:
```
pwndbg> ds asd
Exception occured: ds: No symbol table is loaded.  Use the "file" command. (<class 'gdb.error'>)
```
Or:
```
pwndbg> ds *$rsp
Exception occured: ds: Attempt to dereference a generic pointer. (<class 'gdb.error'>)
```
^ this one is actually useful and we might want to support it if possible.

---

The solution to it might be:
* reworking `fix` or/and `fix_int_reraise`
* introduce similar method that would be used for arguments that needs exactly an address (or something resolving to one) - this would let us not break other commands that are based on `fix` and `fix_int_reraise` (probably there are some) and migrate other commands to use the new/better solution later on.

Or maybe there is some better way. If you have some ideas, feel free to write them down so we can discuss :P.

@zachriggle, @anthraxx, @MaskRay would be nice if you guys would look over this as well.

zachriggle:
I'm not sure of a nice way to resolve the `fix_int_reraise` issues.  Ultimately, we need to display some error message when the user enters incorrect data, but we don't support some complex expressions.  I'm not sure of a good way to handle it.

MaskRay:
You may want to use some top-down operator precedence parsing, e.g.

```c++
namespace {
enum {
  kOp = 128,
  kAssign,
  kIdent,  // temporary variable in expression or flag
  kLiteral,
};
enum {
  kEq = 128, kNe, kLe, kGe, kShl, kShr, kLogicAnd, kLogicOr,
  kAddAssign, kSubAssign, kMulAssign, kDivAssign, kModAssign,
  kShlAssign, kShrAssign, kAndAssign, kXorAssign, kOrAssign,
};
int tbl[256];

struct TDOP {
  RCore* core_;
  const char* s;
  std::unordered_map<string, u64> bind;
  struct Token {
    int type;
    u64 val;
    string_view text;
  } tok_;

  void advance(int type = -1) {
    // expect type
    if (type >= 0) {
      if (type != tok_.type) throw EINVAL;
      if (!type) return;
    }
    switch (tbl[u8(*s)] >> 8) {
    case '|':
      if (s[1] == '|') tok_.type = kLogicOr, s++;
      else if (s[1] == '=') tok_.type = kOrAssign, s++;
      else tok_.type = '|';
      s++;
      break;
    case '^':
      if (s[1] == '=') tok_.type = kXorAssign, s++;
      else tok_.type = '^';
      s++;
      break;
    case '&':
      if (s[1] == '&') tok_.type = kLogicAnd, s++;
      else if (s[1] == '=') tok_.type = kAndAssign, s++;
      else tok_.type = '&';
      s++;
      break;
    case '=':
      if (s[1] == '=') tok_.type = kEq, s++;
      else tok_.type = '=';
      s++;
      break;
    case '!':
      if (s[1] == '=') tok_.type = kNe, s++;
      else tok_.type = '!';
      s++;
      break;
    case '<':
      if (s[1] == '<') tok_.type = s[2] == '=' ? s++, kShlAssign : kShl, s++;
      else if (s[1] == '=') tok_.type = kLe, s++;
      else tok_.type = '<';
      s++;
      break;
    case '>':
      if (s[1] == '>') tok_.type = s[2] == '=' ? s++, kShrAssign : kShl, s++;
      else if (s[1] == '=') tok_.type = kGe, s++;
      else tok_.type = '>';
      s++;
      break;
    case '*':
      if (s[1] == '=') tok_.type = kMulAssign, s++;
      else tok_.type = '*';
      s++;
      break;
    case '/':
      if (s[1] == '=') tok_.type = kDivAssign, s++;
      else tok_.type = '/';
      s++;
      break;
    case '%':
      if (s[1] == '=') tok_.type = kModAssign, s++;
      else tok_.type = '%';
      s++;
      break;
    case '+':
      if (s[1] == '=') tok_.type = kAddAssign, s++;
      else tok_.type = '+';
      s++;
      break;
    case '-':
      if (s[1] == '=') tok_.type = kSubAssign, s++;
      else tok_.type = '-';
      s++;
      break;
    case kAssign:
    case kOp:
      tok_.type = *s++;
      break;
    case kLiteral: {
      tok_.type = kLiteral;
      char* end;
      errno = 0;
      tok_.val = strtoll(s, &end, 0);
      if (!errno) {
        s = end;
        break;
      }
      if (errno != ERANGE) throw errno;
      errno = 0;
      tok_.val = strtoull(s, &end, 0);
      if (!errno) {
        s = end;
        break;
      }
      throw errno;
    }
    case kIdent: {
      tok_.type = kIdent;
      auto i = s;
      while (*++s && ((tbl[int(*s)] >> 8) == kIdent || *s == '.'));
      tok_.text = {i, size_t(s - i)};
      break;
    }
    default:
      throw EINVAL;
    }
  }

  int lbp(int type) {
    return tbl[type] & 255;
  }

  u64 nud(const Token& tok, bool used) {
    switch (tok.type) {
    case '+': return expr(lbp('!'), used);
    case '-': return - expr(lbp('!'), used);
    case '~': return ~ expr(lbp('!'), used);
    case '!': return ! expr(lbp('!'), used);
    case '(': {
      u64 v = expr(0, used);
      advance(')');
      return v;
    }
    case kLiteral: return tok.val;
    default:
      throw EINVAL;
    }
  }

  u64 led(const Token& tok, u64 lhs, bool used, int delim) {
    int p = lbp(tok.type);
    u64 v;
    switch (tok.type) {
    case ',': return expr(p, used, delim);  // delim may be ':'
    case '?': {
      u64 then = expr(0, used && lhs, ':');
      advance(':');
      u64 els = expr(p, used && ! lhs);
      return lhs ? then : els;
    }
    case kLogicOr: return lhs || expr(p, used);
    case kLogicAnd: return lhs && expr(p, used);
    case kEq: return lhs == expr(p, used);
    case kNe: return lhs != expr(p, used);
    case '<': return lhs < expr(p, used);
    case kLe: return lhs <= expr(p, used);
    case '>': return lhs > expr(p, used);
    case kGe: return lhs >= expr(p, used);
    case kShl: return lhs << expr(p, used);
    case kShr: return lhs >> expr(p, used);
    case '|': return lhs | expr(p, used);
    case '^': return lhs ^ expr(p, used);
    case '&': return lhs & expr(p, used);
    case '+': return lhs + expr(p, used);
    case '-': return lhs - expr(p, used);
    case '*': return lhs * expr(p, used);
    case '/':
      if (!(v = expr(p, used)))
        throw EOVERFLOW;
      return lhs / v;
    case '%':
      if (!(v = expr(p, used)))
        throw EOVERFLOW;
      return lhs % v;
    default: throw EINVAL;
    }
  }

  bool eval(const string& ident, u64* out) {
    auto it = bind.find(ident);
    if (it != bind.end()) {
      *out = it->second;
      return true;
    }
    if (Flag* flag = flag::get(core_, ident)) {
      *out = flag->addr_;
      return true;
    }
    return false;
  }

  u64 expr(int rbp, bool used, int delim = -1) {
    Token t = tok_;
    advance();
    string_view var;
    u64 lhs = 0;
    bool valid;
    if (t.type == kIdent) {
      var = t.text;
      valid = eval(string{var}, &lhs);
    } else {
      lhs = nud(t, used);
      valid = true;
    }
    while (rbp < lbp(tok_.type) && tok_.type != delim) {
      t = tok_;
      advance();
      if ((tbl[t.type] >> 8) != kAssign)
        lhs = led(t, lhs, used, delim);
      else {
        if (var.empty()) return EINVAL;
        // -1 for right associative
        u64 rhs = expr(lbp('=') - 1, used, delim);
        /* final */ switch (t.type) {
        case '=': lhs = rhs; valid = true; break;
        case kAddAssign: lhs += rhs; break;
        case kSubAssign: lhs -= rhs; break;
        case kMulAssign: lhs *= rhs; break;
        case kDivAssign: if (!rhs) throw EOVERFLOW; lhs /= rhs; break;
        case kModAssign: if (!rhs) throw EOVERFLOW; lhs %= rhs; break;
        case kShlAssign: lhs <<= rhs; break;
        case kShrAssign: lhs >>= rhs; break;
        case kAndAssign: lhs &= rhs; break;
        case kXorAssign: lhs ^= rhs; break;
        case kOrAssign: lhs |= rhs; break;
        }
        if (used && valid)
          bind[string{var}] = lhs;
        var = {};
      }
    }
    if (used && ! valid) throw EINVAL;
    return lhs;
  }
};
}  // namespace

// sv ends with a NUL
// Intgers are represented by u64 and wrap around. '/' '%' are unsigned operations
int calculate(RCore* core, const char* s, u64* out) {
  if (!tbl[0]) {
    tbl[0] = kOp << 8 | 0;
    tbl[int(')')] = kOp << 8 | 0;
    tbl[int(',')] = kOp << 8 | 1;
    tbl[int('=')] = tbl[kAddAssign] = tbl[kSubAssign] = tbl[kMulAssign] = tbl[kDivAssign] =
      tbl[kShlAssign] = tbl[kShrAssign] = tbl[kAndAssign] = tbl[kXorAssign] = tbl[kOrAssign] = kAssign << 8 | 2;
    tbl[int('?')] = tbl[int(':')] = kOp << 8 | 3;
    tbl[kLogicOr] = kOp << 8 | 4;
    tbl[kLogicAnd] = kOp << 8 | 5;
    tbl[int('|')] = '|' << 8 | 6;
    tbl[int('^')] = '^' << 8 | 7;
    tbl[int('&')] = '&' << 8 | 8;
    tbl[int('<')] = '<' << 8 | 9;
    tbl[int('>')] = '>' << 8 | 9;
    tbl[kLe] = tbl[kGe] = kOp << 8 | 9;
    tbl[kEq] = tbl[kNe] = kOp << 8 | 10;
    tbl[int('+')] = '+' << 8 | 11;
    tbl[int('-')] = '-' << 8 | 11;
    tbl[int('*')] = '*' << 8 | 12;
    tbl[int('/')] = '/' << 8 | 12;
    tbl[int('%')] = '%' << 8 | 12;
    // Unary
    tbl[int('!')] = tbl[int('~')] = kOp << 8 | 13;
    // lparen
    tbl[int('(')] = kOp << 8 | 14;

    std::fill_n(tbl + '0', 10, kLiteral << 8);
    std::fill_n(tbl + 'a', 26, kIdent << 8);
    tbl[int('_')] = kIdent << 8;
  }

  try {
    TDOP p{core, s};
    p.advance();
    *out = p.expr(0, true);
    p.advance('\0');
    return 0;
  } catch (int err) {
    return - err;
  }
}
```

This snippet supports most C operators and even ternary `?:`. But no loop is allowed and calculation is guaranteed to be from left to right

Regarding commands, if you want to have more expressive power, using a class to describe a command will be great, because of Python's syntax limitation.

zachriggle:
I have no intent to re-implement (or replace) GDB's expression parser, since it's pretty insane and internally inconsistent (e.g. `*symbol` behaves differently for `x/` vs `break`)

zachriggle:
@disconnect3d The `*$rsp` thing is actually an invalid expression as far as GDB is concerned.  We should honor this behavior.

You want `*(void**)$rsp`.

```
(gdb) x $rsp
0x7fffffffdb98:	0xf7609f45
(gdb) x *$rsp
Attempt to dereference a generic pointer.
(gdb) x *(void*)$rsp
Attempt to dereference a generic pointer.
(gdb) x *(void**)$rsp
0x7ffff7609f45 <__libc_start_main+245>:	0x94e8c789
```

We may want to add a helper `$poi(...)` function that behaves [similar to WinDBG](https://blogs.msdn.microsoft.com/debuggingtoolbox/2008/03/04/special-command-using-c-and-poi-with-cc-expressions/) which does this automagically.

disconnect3d:
@devneal bump?

devneal:
Sorry for the delay. I agree that pwndbg should follow gdb's syntax as faithfully as possible, but I think that should be discussed and added in a future pull request. I chose to do it this way because it's consistent with [`hexdump.py`](https://github.com/pwndbg/pwndbg/blob/466ab2cc2b9fc94912130b4a99bb0ebc3ccbc7ec/pwndbg/commands/hexdump.py#L39-L40) and [`telescope.py`](https://github.com/pwndbg/pwndbg/blob/466ab2cc2b9fc94912130b4a99bb0ebc3ccbc7ec/pwndbg/commands/telescope.py#L44-L45).

pwndbg/commands/windbg.py
zachriggle: I see a `max` argument just below.

pwndbg/commands/windbg.py
zachriggle: I see a `max` argument just below.

zachriggle (approved):
LGTM, @disconnect3d merge when you're happy

disconnect3d:
@devneal that's fine, thanks for the PR :). merged o/

PR #369, opened 2017-12-15
Support bare metal environment

cebrusfs:
Hi all,

I read the issue #264 and give a *draft* PR to disable the some magic AUXV and ELF header finding. Any suggestions are welcome.

This version will disable the AUXV looking up in `walking_stack()` and return empty `AUXV()`. This will cause all the section information are not available. Also, disable the page scan in `get_ehdr()` and `find_upper_stack_boundary()`.

Here's some idea and some topic want to discuss:
1. The pwndbg rely tidily with ELF format and AUXV, this make me hardly to add other ABI support. Maybe we need to re-design framework to support the ABI detection and dispatch to the correct functions.

2. Without the page information, the `telescope` works badly and it will try to de-reference everything. Maybe the next step is automatic load the page information from ELF file. (I have not start to do that yet, any ideas are welcome). BTW, is it possible to create `AUXV()` from ELF the file? or it is created by OS loader when loading into the virtual memory?

3. If we add the page information from ELF file (create some guessing page like `vmmap.explore()` do), but in my bare metal environment, the mmu is disabled, the addresses are all physical address. That is, there is no *page* in my environment, so using the current `Page()` design may be wired or may cause some unexpected bugs.

### DIFF
--- a/pwndbg/abi.py
+++ b/pwndbg/abi.py
@@ -4,6 +4,11 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import functools
+import re
+
+import gdb
+
 import pwndbg.arch
 
 
@@ -111,3 +116,37 @@ linux_arm_sigreturn = SigreturnABI(['r7'], 4, 0)
 linux_i386_srop = ABI(['eax'], 4, 0)
 linux_amd64_srop = ABI(['rax'], 4, 0)
 linux_arm_srop = ABI(['r7'], 4, 0)
+
+
+@pwndbg.events.start
+def update():
+    global abi
+    global linux
+
+    # Detect current ABI of client side by 'show osabi'
+    osabi_string = gdb.execute('show osabi', to_string=True)
+
+    # The return string will be:
+    # The current OS ABI is "auto" (currently "GNU/Linux").
+    match = re.search('currently "([^"]+)"', osabi_string)
+    if match:
+        # 'GNU/Linux': linux
+        # 'none': bare metal
+        abi = match.group(1)
+
+        linux = 'Linux' in abi
+
+def LinuxOnly(default=None):
+    """Create a decorator that the function will be called when ABI is Linux.
+    Otherwise, return `default`.
+    """
+    def decorator(func):
+        @functools.wraps(func)
+        def caller(*args, **kwargs):
+            if linux:
+                return func(*args, **kwargs)
+            else:
+                return default
+        return caller
+
+    return decorator
--- a/pwndbg/elf.py
+++ b/pwndbg/elf.py
@@ -18,6 +18,7 @@ import sys
 import gdb
 from six.moves import reload_module
 
+import pwndbg.abi
 import pwndbg.arch
 import pwndbg.auxv
 import pwndbg.elftypes
@@ -121,6 +122,7 @@ def reset_ehdr_type_loaded():
     global ehdr_type_loaded
     ehdr_type_loaded = 0
 
+@pwndbg.abi.LinuxOnly()
 def find_elf_magic(pointer, max_pages=1024, search_down=False):
     """Search the nearest page which contains the ELF headers
     by comparing the ELF magic with first 4 bytes.
@@ -165,6 +167,11 @@ def get_ehdr(pointer):
     # Align down to a page boundary, and scan until we find
     # the ELF header.
     base = pwndbg.memory.page_align(pointer)
+
+    # XXX: for non linux ABI, the ELF header may not be found in memory.
+    # This will hang the gdb when using the remote gdbserver to scan 1024 pages
+    if not pwndbg.abi.linux:
+        return None, None
 
     base = find_elf_magic(pointer, search_down=True)
     if base is None:
--- a/pwndbg/elf.py
+++ b/pwndbg/elf.py
@@ -18,6 +18,7 @@ import sys
 import gdb
 from six.moves import reload_module
 
+import pwndbg.arch
 import pwndbg.auxv
 import pwndbg.elftypes
 import pwndbg.events
@@ -120,6 +121,44 @@ def reset_ehdr_type_loaded():
     global ehdr_type_loaded
     ehdr_type_loaded = 0
 
+def find_elf_magic(pointer, max_pages=1024, search_down=False):
+    """Search the nearest page which contains the ELF headers
+    by comparing the ELF magic with first 4 bytes.
+
+    Parameter:
+        search_down: change the search direction
+        to search over the lower address.
+        That is, decreasing the page pointer instead of increasing.
+            (default: False)
+    Returns:
+        An integer address of ELF page base
+        None if not found within the page limit
+    """
+    base = pwndbg.memory.page_align(pointer)
+    step = pwndbg.memory.PAGE_SIZE
+    if search_down:
+        step = -step
+
+    max_addr = pwndbg.arch.ptrmask
+
+    for i in range(max_pages):
+        # Make sure address within valid range or gdb will raise Overflow exception
+        if base < 0 or base > max_addr:
+            return None
+
+        try:
+            data = pwndbg.memory.read(base, 4)
+        except gdb.MemoryError:
+            return None
+
+        # Return the address if found ELF header
+        if data == b'\x7FELF':
+            return base
+
+        base += step
+
+    return None
+
 def get_ehdr(pointer):
     """Returns an ehdr object for the ELF pointer points into.
     """
@@ -127,21 +166,9 @@ def get_ehdr(pointer):
     # the ELF header.
     base = pwndbg.memory.page_align(pointer)
 
-    try:
-        data = pwndbg.memory.read(base, 4)
-
-        # Do not search more than 4MB of memory
-        for i in range(1024):
-            if data == b'\x7FELF':
-                break
-
-            base -= pwndbg.memory.PAGE_SIZE
-            data = pwndbg.memory.read(base, 4)
-
-        else:
-            print("ERROR: Could not find ELF base!")
-            return None, None
-    except gdb.MemoryError:
+    base = find_elf_magic(pointer, search_down=True)
+    if base is None:
+        print("ERROR: Could not find ELF base!")
         return None, None
 
     # Determine whether it's 32- or 64-bit
--- a/pwndbg/argv.py
+++ b/pwndbg/argv.py
@@ -7,6 +7,7 @@ from __future__ import unicode_literals
 
 import gdb
 
+import pwndbg.abi
 import pwndbg.arch
 import pwndbg.events
 import pwndbg.memory
@@ -25,6 +26,7 @@ envp = None
 envc = None
 
 @pwndbg.events.start
+@pwndbg.abi.LinuxOnly()
 def update():
     global argc
     global argv
--- a/pwndbg/auxv.py
+++ b/pwndbg/auxv.py
@@ -11,6 +11,7 @@ import sys
 
 import gdb
 
+import pwndbg.abi
 import pwndbg.arch
 import pwndbg.events
 import pwndbg.info
@@ -150,6 +151,8 @@ def find_stack_boundary(addr):
     return addr
 
 def walk_stack():
+    if not pwndbg.abi.linux:
+        return None
     if pwndbg.qemu.is_qemu_kernel():
         return None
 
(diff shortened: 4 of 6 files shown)


pwndbg/abi.py
```diff
@@ -111,3 +116,31 @@ class SigreturnABI(SyscallABI):
...
 linux_amd64_srop = ABI(['rax'], 4, 0)
 linux_arm_srop = ABI(['r7'], 4, 0)
+
+
+# See issue #264 for more reference
+# 'GNU/Linux' for linux
+# 'none' for bare metal
+abi = None
+
+def is_linux():
```
zachriggle: I'd rather just have a global that we update whenever `abi` is updated, so we can check:

`if pwndbg.abi.linux` vs `if pwndbg.abi.is_linux()`

pwndbg/abi.py
```diff
@@ -111,3 +116,31 @@ class SigreturnABI(SyscallABI):
...
+def update():
+    global abi
+
+    # detect current ABI of client side by 'show osabi'
+    osabi_string = gdb.execute('show osabi', to_string=True)
+    match = re.search('currently "([^"]+)"', osabi_string)
+    if match:
+        abi = match.group(1)
+
+def Linux(func):
```
zachriggle: I'd rather rename this to `OnlyLinux` or `LinuxOnly` to be slightly more explanatory

pwndbg/abi.py
```diff
@@ -111,3 +116,31 @@ class SigreturnABI(SyscallABI):
...
+    @functools.wraps(func)
+    def caller(*args, **kwargs):
+        # call func only if ABI is Linux
+        if is_linux():
+            func(*args, **kwargsa)
```
zachriggle: missing `return` statement

zachriggle: We might also want there to be a way to specify a default return value, e.g.

```py
@LinuxOnly("Not Linux!")
def get_some_string():
  return "Yes Linux!"
```

Which can be done by using a `class` for the wrapper instead of a function.

```py
class LinuxOnly(object):
  def __init__(self, default):
    self._default = default
  def __call__(self, function):
    def wrapper(*a,**kw):
      if not linux: return self._default
      else: return function(*a,**kw)
```

pwndbg/elf.py
```diff
@@ -120,28 +122,58 @@ def reset_ehdr_type_loaded():
...
+    by comparing the ELF magic with first 4 bytes.
+
+    Returns:
+        An integer address of ELF page base
+        None if not found within the page limit
+    """
+    base = pwndbg.memory.page_align(pointer)
+
+    # Search the page by decreasing the address. Increasing when searching in the stack
+    if not in_stack:
```
zachriggle: Can we rename this parameter to `search_down`?

cebrusfs:
I updated the commit without the testing. Since I don't have the bare metal environment right now, I think I can test it on Monday.

Any comment on the next stage change? I will work on loading the page information from ELF file next week. It's better to hear some design opinion before working.

pwndbg/elf.py
```diff
@@ -120,28 +122,55 @@ def reset_ehdr_type_loaded():
...
+    by comparing the ELF magic with first 4 bytes.
+
+    Returns:
+        An integer address of ELF page base
+        None if not found within the page limit
+    """
+    base = pwndbg.memory.page_align(pointer)
+    step = pwndbg.memory.PAGE_SIZE
+    if not search_down:
+        step = -step
```
disconnect3d: Shouldn't it be `if search_down: step = -step`? Isn't searching down a searching over lower addresses?

pwndbg/elf.py
```diff
@@ -120,28 +122,55 @@ def reset_ehdr_type_loaded():
...
+    max_addr = pwndbg.arch.ptrmask
+
+    for i in range(max_pages):
+        # Make sure address within valid range or gdb will raise Overflow exception
+        if base < 0 or base > max_addr:
+            return None
+
+        try:
+            data = pwndbg.memory.read(base, 4)
+        except gdb.MemoryError:
```
disconnect3d, line 153: Future note: we probably should catch this in `pwndbg.memory.read` and throw a custom exception
(would make it easier to abstract gdb out in the future...)

We don't have to add this in current PR

disconnect3d:
Hey,

Thanks for the PR!

---

> Maybe the next step is automatic load the page information from ELF file. (I have not start to do that yet, any ideas are welcome).

What is that for? To generate vmmap for targets run without OS?
Aren't such targets just having one big memory or maybe two distinct memories (one for code one for stack/data?) depending on the hardware.

> BTW, is it possible to create AUXV() from ELF the file? or it is created by OS loader when loading into the virtual memory?

They say it is created by kernel: http://articles.manugarg.com/aboutelfauxiliaryvectors

For sure some of the values can be created based on ELF file. Btw we have the `auxv` command:
```
pwndbg> auxv
AT_BASE                  0x7ffff7dd8000 ◂— jg     0x7ffff7dd8047
AT_CLKTCK                0x64
AT_EGID                  0x64
AT_ENTRY                 0x555555559060 ◂— xor    ebp, ebp
AT_EUID                  0x3e8
AT_EXECFN                /usr/bin/ls
AT_FLAGS                 0x0
AT_GID                   0x64
AT_HWCAP                 0xbfebfbff
AT_NULL                  0x0
AT_PAGESZ                0x1000Hey,

Thanks for the PR!

---

> Maybe the next step is automatic load the page information from ELF file. (I have not start to do that yet, any ideas are welcome).

What is that for? To generate vmmap for targets run without OS?
Aren't such targets just having one big memory or maybe two distinct memories (one for code one for stack/data?) depending on the hardware.

> BTW, is it possible to create AUXV() from ELF the file? or it is created by OS loader when loading into the virtual memory?

They say it is created by kernel: http://articles.manugarg.com/aboutelfauxiliaryvectors
Some of the values can be created by parsing ELF for sure.

Btw we've also got a `auxv` command, example:
```
pwndbg> auxv
AT_BASE                  0x7ffff7dd8000 ◂— jg     0x7ffff7dd8047
AT_CLKTCK                0x64
AT_EGID                  0x64
AT_ENTRY                 0x555555559060 ◂— xor    ebp, ebp
AT_EUID                  0x3e8
AT_EXECFN                /usr/bin/ls
AT_FLAGS                 0x0
AT_GID                   0x64
AT_HWCAP                 0xbfebfbff
AT_NULL                  0x0
AT_PAGESZ                0x1000
AT_PHDR                  0x555555554040 ◂— 0x500000006
AT_PHENT                 0x38
AT_PHNUM                 0x9
AT_PLATFORM              x86_64
AT_RANDOM                0x7fffffffda19 ◂— 0x82e867c88127fb35
AT_SECURE                0x0
AT_SYSINFO_EHDR          0x7ffff7ffa000 ◂— jg     0x7ffff7ffa047
AT_UID                   0x3e8
AT_UNKNOWN26             0x0
```

**Comment by disconnect3d** [comment]
Hey,

Thanks for the PR!

---

> Maybe the next step is automatic load the page information from ELF file. (I have not start to do that yet, any ideas are welcome).

What is that for? To generate vmmap for targets run without OS?
Aren't such targets just having one big memory or maybe two distinct memories (one for code one for stack/data?) depending on the hardware.

> BTW, is it possible to create AUXV() from ELF the file? or it is created by OS loader when loading into the virtual memory?

They say it is created by kernel: http://articles.manugarg.com/aboutelfauxiliaryvectors

For sure some of the values can be created based on ELF file. Btw we have the `auxv` command:
```
pwndbg> auxv
AT_BASE                  0x7ffff7dd8000 ◂— jg     0x7ffff7dd8047
AT_CLKTCK                0x64
AT_EGID                  0x64
AT_ENTRY                 0x555555559060 ◂— xor    ebp, ebp
AT_EUID                  0x3e8
AT_EXECFN                /usr/bin/ls
AT_FLAGS                 0x0
AT_GID                   0x64
AT_HWCAP                 0xbfebfbff
AT_NULL                  0x0
AT_PAGESZ                0x1000Hey,
```

Also it might be worth checking whether e.g. pwntools have something for this exact purpose. But there is also a problem with pwntools - it is not compatible with python3 so we might end up with copying their code to our repo.

**Comment by cebrusfs** [comment]
> What is that for? To generate vmmap for targets run without OS?
Aren't such targets just having one big memory or maybe two distinct memories (one for code one for stack/data?) depending on the hardware.

Yes, it depends on the hardware. In my use case, I only have one big physical memory space and some memory are MMIO or peripheral registers.

My purpose is 
1. making the color of address work.
2. the de-reference `chain.get()` stop on some unused (physical) address, because it is data in high probability.

In my understanding, the color of address rely on the vmmap and the de-reference is just use gdb to de-reference.

For 1, since I can compile and control the ELF, so my idea is add the page information in ELF and let pwndbg parse it to get the page information when AUXV is not available ([the current code](https://github.com/pwndbg/pwndbg/blob/8a5a6da0a3d14e8a26b135f589a458be2b302972/pwndbg/auxv.py#L109) will use an empty instance `AXUV()`), or we can let it only enable when in bare metal.

For 2, can we just use the information from page to decide to de-reference or not in bare metal mode? Since the addresses are all valid to de-reference even they are not used, but we know what the memory range we cared. This will make some data decoding in `chain.format()` such as the ASCII string decodng works. They are considered as a pointer and will be de-referenced now.

File: pwndbg/elf.py (hunk 5 above)
Comment (cebrusfs) [reply pwndbg/elf.py]: The first image in my brain is a stack view and the lower address is at the top, so when I thing the "down" is the higher address. lol

I think I get it the "down" means like the "count down" which means the decreasing. I will update this later.

**Comment by cebrusfs** [comment]
Hi @disconnect3d,

I already fix the `search_down` and add some comment in doc string to make sure nobody will confuse the meaning of the "down" like me :)

File: pwndbg/elf.py [hunk 7]
```diff
@@ -120,28 +122,60 @@ def reset_ehdr_type_loaded():
...
+        None if not found within the page limit
+    """
+    base = pwndbg.memory.page_align(pointer)
+    step = pwndbg.memory.PAGE_SIZE
+    if search_down:
+        step = -step
+
+    max_addr = pwndbg.arch.ptrmask
+
+    for i in range(max_pages):
```
disconnect3d, line 146: We could probably change that to sth like `range(base, base+pwndbg.memory.PAGE_SIZE*max_pages, step)` but that would have to consider moving backwards so maybe it is better to leave it as it is? Anyway, this doesn't matter that much.

disconnect3d (approved):
LGTM

disconnect3d:
> For 1, since I can compile and control the ELF, so my idea is add the page information in ELF and let pwndbg parse it to get the page information when AUXV is not available

I feel that requiring to change might not be the best way, maybe we can specify those differently, like from inside pwndbg with some command? I am not sure - but it should be as easy as possible (modifying ELF might require some additional script or something?).

> For 2, can we just use the information from page to decide to de-reference or not in bare metal mode? 

Sounds good to me.

cebrusfs:
>> For 1, since I can compile and control the ELF, so my idea is add the page information in ELF and let pwndbg parse it to get the page information when AUXV is not available

> I feel that requiring to change might not be the best way, maybe we can specify those differently, like from inside pwndbg with some command? I am not sure - but it should be as easy as possible (modifying ELF might require some additional script or something?).

For my use case, I will add this information in the linker script. Also, for someone who have ELFs that contains correct page information can load this information automatic and no need to add the table manually.
How about creating a command interface to maintain (add/delete/edit) some customize pages, also make a command/or an option to automatic load page information from ELF if they want. This can reduce their works.

>> For 2, can we just use the information from page to decide to de-reference or not in bare metal mode?

> Sounds good to me.

BTW, this also helps to reduce the useless memory reading in normal Linux ABI mode. In my developing environment, the openocd with JTAG for mcu operates much slower than normal gdbserver. It takes about 0.2s to generate the `context_regs` page in my case, but it only takes about 0.02s on my x64 local gdbserver.
However, if we enable this as default. It will be harmful for those doesn't have full page information. It is a trade-off, what do you think?

disconnect3d:
I think we can't disable automatic dereferencing as a default or at least
not for all targets as for most users this is not the case at all.

Here are some options:
1. Disable dereferencing by default just for bare metal targets and inform
the users when they start debugging bare metal why we do so and how to
switch it back.

2. Measure the time it takes to render context (or regs or something else)
and inform users they can disable dereferencing and that it might improve
the timings.

Anyway this probably means we have to add a parameter to decide whether to
derefence stuff while showing context (or just regs) or not.

We might also want to provide some in-pwndbg help for various topics - like
bare metal debugging.

20.12.2017 10:14 AM "Meng-Huan Yu" <notifications@github.com> napisał(a):

PR #373, opened 2017-12-16
Add `desc` to ArgparsedCommand; move some commmands to Argparsed~

disconnect3d:
* adds optional keyword argument -  `desc`  - to `ArgparsedCommand` so we don't have to create `argparse.ArgumentParser` if we don't want any parameters
* adds prints for cases when plt/gotplt sections doesnt have any symbols (so user will get an output out of plt/gotplt instead of nothing)
* minor code refactorings

Potentially we could merge ArgparsedCommand and Command in the future (and remove all other Command children).

Also continuation of https://github.com/pwndbg/pwndbg/issues/244

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -221,12 +221,18 @@ class _ArgparsedCommand(Command):
 
 class ArgparsedCommand(object):
     """Adds documentation and offloads parsing for a Command via argparse"""
-    def __init__(self, parser):
-        self.parser = parser
+    def __init__(self, parser_or_desc):
+        """
+        :param parser_or_desc: `argparse.ArgumentParser` instance or `str`
+        """
+        if isinstance(parser_or_desc, str):
+            self.parser = argparse.ArgumentParser(description=parser_or_desc)
+        else:
+            self.parser = parser_or_desc
 
         # We want to run all integer and otherwise-unspecified arguments
         # through fix() so that GDB parses it.
-        for action in parser._actions:
+        for action in self.parser._actions:
             if action.dest == 'help':
                 continue
             if action.type in (int, None):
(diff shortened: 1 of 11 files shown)


disconnect3d:
Continuation of https://github.com/pwndbg/pwndbg/issues/244

pwndbg/commands/__init__.py
```diff
@@ -216,12 +216,17 @@ def split_args(self, argument):
 
 class ArgparsedCommand(object):
     """Adds documentation and offloads parsing for a Command via argparse"""
-    def __init__(self, parser):
-        self.parser = parser
+    def __init__(self, parser=None, desc=None):
+        assert parser is None or desc is None, 'Invalid ArgparsedCommand usage'
```
zachriggle: I'd prefer a tiny bit of magic where we take a single argument and just check `isinstance(...)`

zachriggle (changes requested):
I'm pretty sure this breaks the docstrings on these functions.

disconnect3d:
@zachriggle This do not break docstrings as they are set by `_ArgparsedCommand`:
```python
class _ArgparsedCommand(Command):
    def __init__(self, parser, function, *a, **kw):
        self.parser = parser
        self.parser.prog = function.__name__
        self.__doc__ = function.__doc__ = self.parser.description
        super(_ArgparsedCommand, self).__init__(function, *a, **kw)
```

```
In : pwndbg.commands.canary.canary.__doc__
'Print out the current stack canary.'
```

---
I have also changed it to use `parser_or_desc` argument with `isinstance` as you suggested.

I have also found out that `canary` has been broken by some previous changes. I am going to address that in another PR.

zachriggle:
Thanks! 😘

zachriggle:
Wrong emoji. ❤️

PR #374, opened 2017-12-17
Adds $rebase(offset) function

disconnect3d:
Adds `$rebase(offset)` gdbfunction that can be used to set up a breakpoint
over an offset from program image base.

Also changed a bit the pwndbg banner displayed at startup.

### DIFF
--- a/pwndbg/gdbutils/functions.py
+++ b/pwndbg/gdbutils/functions.py
@@ -1,0 +1,54 @@
+# -*- coding: utf-8 -*-
+"""
+Put all functions defined for gdb in here.
+
+This file might be changed into a module in the future.
+"""
+from __future__ import absolute_import
+from __future__ import division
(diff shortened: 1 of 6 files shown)


pwndbg/gdbutils/functions.py
```diff
@@ -0,0 +1,54 @@
...
+
+        functions.append(self)
+
+        super(_GdbFunction, self).__init__(self.name)
+
+        functools.update_wrapper(self, func)
+        self.__doc__ = func.__doc__
+
+    def invoke(self, *args):
+        if self.only_when_running and not pwndbg.elf.exe():
```
zachriggle: Use `pwndbg.proc.alive` instead of `.exe()`

zachriggle (changes requested):

PR #375, opened 2017-12-17
Tests framework

disconnect3d:
Alternative way of unit testing pwndbg.

Things done:
* moved `./tests` to `./e2e_tests` where e2e stands for 'end to end'
* added `./pwndbg/services` - this might need a better name - its purpose is to handle implementation of functions used by commands
* added unit tests through pytest framework and some examples (that should be removed before merging this) - personally I like it over nose/unittest/anything else as its super easy to use and doesn't require us to have boilerplate
* tests for the added service used by `pwndbg` command

All unit tests are run on a single gdb with pwndbg loaded and no binary under the hood.


pwndbg/tests/test_misc.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+def test_show_super_assert():
+    x = [1, 2, 3]
+    assert x == [1, 2, 4]
+
+
+def test_code_that_should_fail():
+    with pytest.raises(ZeroDivisionError):
+        1 / 0
```
disconnect3d: The tests above are just to see how it will show up on CI.

disconnect3d:
@zachriggle any thoughts?
I will fix conflicts later (and remove redundant unittests); would be nice to know if we want this or not :P.

disconnect3d:
Also: I feel like `pwndbg/services/` is not that good, I will move the function back to the `pwndbg/commands/misc.py` (or to `pwndbg/misc.py`)

pwndbg/tests/test_misc.py
```diff
@@ -0,0 +1,121 @@
...
+        ('context', 'Print out the current register, instruction, and stack context.'),
+        ('down', 'Select and print stack frame called by this one.'),
+        ('retaddr', 'Print out the stack addresses that contain return addresses'),
+        ('stack', 'dereferences on stack data with specified count and offset'),
+        ('up', 'Select and print stack frame that called this one.')
+    ]
+
+
+def test_pwndbg_list_and_filter_commands_full_list():
+    assert pwndbg_list_and_filter_commands('') == [
```
zachriggle: This looks like a giant PITA to maintain.  I'd recommend picking a single command, and using `.__doc__`  to fetch the expected string.

e2e_tests/common.py
```diff
@@ -0,0 +1,55 @@
...
+def write(data, suffix=''):
+    t = tempfile.NamedTemporaryFile(delete=False, suffix=suffix)
+    t.write(data.encode('utf-8'))
+    return t
+
+
+ANSI_ESCAPE = re.compile(r'(\x9B|\x1B\[)[0-?]*[ -/]*[@-~]')
+
+
+def escape_ansi(line):
```
zachriggle: Rather than doing ANSI escaping, perhaps we should just disable all of the colorized output when running tests?

zachriggle (changes requested):

zachriggle:
Looks good to me, I checked out the Travis failures. Looks like we catch them and print out useful info.

What happens when we need to execute a GDB command which fails?  It'd be useful to have at least one exemplary test with a file loaded and the binary running.

disconnect3d:
@zachriggle 
* rebased
* added some more end to end tests; removed ansi color codes escaping in favor of disabling colors based upon environment variable

disconnect3d:
@zachriggle @anthraxx please review current approach.

Currently there are two ways of testing - end to end tests and unit tests.

The e2e tests are made through unittests and launching gdb as subprocess. The latter - unit tests - is just launching python script under gdb which launches pytest to execute python tests.

Python is pretty neat framework as it changes `assert` statements so that we can easily see "what is wrong" if an assertion is fail (it automatically displays values etc).

I feel like the e2e tests have to be rewritten and use pytest too...

disconnect3d:
@zachriggle @anthraxx please review current approach. I have made a cleanup, added some more tests - for loading binaries and for `pwndbg.memory.read` and `pwndbg.memory.write`.

After merging this, lets try to add tests for each regression or bug that happens here and there.

Another big step that we need to do is adding more testing code and launch tests on GDBs using both py2 and py3 and different architectures (should all tests be flexible for that?).

pwndbg/exception.py
```diff
@@ -47,6 +47,14 @@ def handle(name='Error'):
         - ``set exception-verbose on`` enables stack traces.
         - ``set exception-debugger on`` enables the post-mortem debugger.
     """
+
+    # This is for unit tests so they fail on exceptions instead of displaying them.
+    if getattr(sys, '_pwndbg_unittest_run', False) is True:
```
disconnect3d, line 52: This could use the same env var that is used for disabling colors BUT we really want to ensure this code is not skipped if someone forgets to set the envvar...

pwndbg/color/__init__.py
```diff
@@ -55,8 +56,14 @@ def colorize(x, color): return color + terminateWith(str(x), color) + NORMAL
 
 @pwndbg.memoize.reset_on_stop
 def generateColorFunctionInner(old, new):
-    def wrapper(text):
-        return new(old(text))
+    # End to end tests check for whole output, we might not want to deal with colors there
+    if os.environ.get('PWNDBG_TESTS_DISABLE_COLORS') != 'yes':
```
zachriggle: This should be checked once at boot-up, rather than every time any color function is invoked.

Separately, I don't see a reason for this to not be a `config` setting.  We can pull the default value from the environment.

tests/binaries/makefile
```diff
@@ -0,0 +1,36 @@
+CC              =       gcc
```
disconnect3d, line 1: Most of this file has been taken from gef's tests makefile https://github.com/hugsy/gef/blob/master/tests/binaries/Makefile

tests/binaries/makefile
```diff
@@ -0,0 +1,36 @@
...
+%.out : %.c
+	@echo "[+] Building '$@'"
+	@$(CC) $(CFLAGS) $(EXTRA_FLAGS) -o $@ $? $(LDFLAGS)
+
+clean :
+	@echo "[+] Cleaning stuff"
+	@rm -f $(COMPILED) $(LINKED)
+
+
+reference-binary.out: EXTRA_FLAGS := -Dexample=1
```
disconnect3d, line 35: the `-D` here is not needed but I wanted to leave it here just to show up that extra flags can be filled

tests/binaries/reference-binary.c
```diff
@@ -0,0 +1,8 @@
+#include <stdio.h>
+
+void foo() {}
```
disconnect3d, line 3: this is not needed but can be used for some tests that would need to break at some func other than main lol

zachriggle:
I really don't like the "check length of the available commands" test, since it requires updating *every* time a new command is added.  It also fails out of the box for me.

```
> /home/pwntools/pwndbg/tests/test_misc.py(25)test_list_and_filter_commands_full_list()
-> assert len(all_commands) == 161
(Pdb) len(all_commands)
162
```

When running the tests on Travis, we also need it to not break into PDB on failures.

.travis.yml
```diff
@@ -14,9 +14,9 @@ install:
...
   - sudo ./setup.sh
 script:
-  - futurize --all-imports --stage1 --print-function --write --unicode-literals pwndbg
-  - git diff-index --quiet HEAD -- pwndbg
-  - isort --check-only --diff --recursive pwndbg
-  - nosetests ./tests/
+  - futurize --all-imports --stage1 --print-function --write --unicode-literals pwndbg tests
+  - git diff-index --quiet HEAD -- pwndbg tests
+  - isort --check-only --diff --recursive pwndbg tests
+  - PWNDBG_TESTS_DISABLE_COLORS=yes gdb --silent --nx --nh --command gdbinit.py --command unittests_launcher.py
```
zachriggle: This incantation should be wrapped up into e.g. `tests/entry.sh`

tests/test_loads.py
```diff
@@ -0,0 +1,86 @@
...
+        'pwndbg: created $rebase, $ida gdb functions (can be used with print/break)\n'
+    )
+    assert hello in output
+
+    output = output[output.index(hello)+len(hello):]
+
+    return output
+
+
+BASH_BIN = './tests/corefiles/bash/binary'
```
zachriggle: Will hard-coded relative paths bite us at some point? (e.g. when not running from the `pwndbg` directory)?

I'd prefer finding a base path (like `os.path.dirname(pwndbg.__file__)` or like what's used in `pwntools` for finding test objects: https://github.com/Gallopsled/pwntools/blob/dev/pwnlib/data/elf/__init__.py

This would look something like `BASH_BIN = pwndbg.tests.corefiles.bash.get('binary')`.

tests/test_memory.py
```diff
@@ -0,0 +1,34 @@
...
+import pwndbg.memory
+import pwndbg.stack
+
+
+def test_memory_read_write(entry_binary):
+    """
+    Tests simple pwndbg's memory read/write operations with different argument types
+    """
+    entry_binary('reference-binary.out')
+    stack_addr = next(iter(pwndbg.stack.stacks.values())).vaddr
```
zachriggle, line 18: Probably easier to just use `pwndbg.regs.stack`?

tests/test_misc.py
```diff
@@ -0,0 +1,33 @@
...
+]
+
+
+def test_list_and_filter_commands_filter():
+    assert list_and_filter_commands('stack') == STACK_COMMANDS
+
+
+def test_list_and_filter_commands_full_list():
+    all_commands = list_and_filter_commands('')
+    assert len(all_commands) == 161
```
zachriggle: Don't like this test, since it requires being updated every time the commands are updated, and breaks when commands are added conditionally (which I'm assuming there's at least one)

tests/test_misc.py
```diff
@@ -0,0 +1,33 @@
...
+    ('context', 'Print out the current register, instruction, and stack context.'),
+    ('down', 'Select and print stack frame called by this one.'),
+    ('retaddr', 'Print out the stack addresses that contain return addresses.'),
+    ('stack', 'dereferences on stack data with specified count and offset.'),
+    ('up', 'Select and print stack frame that called this one.')
+]
+
+
+def test_list_and_filter_commands_filter():
+    assert list_and_filter_commands('stack') == STACK_COMMANDS
```
zachriggle: I would prefer that we check contents rather than equality.  I should be able to add a command that mentions `stack` and not break this test.

unittests_launcher.py
```diff
@@ -0,0 +1,18 @@
...
+from __future__ import absolute_import
+from __future__ import unicode_literals
+
+import pytest
+import sys
+
+sys._pwndbg_unittest_run = True
+
+# If you want to debug tests locally, add '--pdb' here
+args = ['./tests/', '--pdb', '-vvv', '-s', '--showlocals', '--color=yes']
```
zachriggle: We should not invoke with `--pdb` unless it's provided on the command-line,  I'd recommend extending with `sys.argv` or something.  The relative path might also bite us at some point.

zachriggle (changes requested):

disconnect3d:
@zachriggle I have removed the `161` length assertion and `--pdb` flag in the latest commit

pwndbg/color/__init__.py
disconnect3d: I am not sure if we can omit the condition here if we want to have it in config variable, but I will give it a try.

tests/test_memory.py
disconnect3d: MMm `pwndbg.regs.stack` returns a string e.g. of `esp` or `rsp`

disconnect3d: We could do `getattr(pwndbg.regs, pwndbg.regs.stack)` but I am not sure its the best way...

unittests_launcher.py
disconnect3d: Is this doable to pass `--pdb` to `gdb ...` and retrieve it later?

disconnect3d:
@zachriggle 

Below summary of my review fixes.
 
1) Tests launcher doesn't start with `--pdb`. However I think we can't add `--pdb`to `./tests.sh` as it launches `gdb ...` and I am not sure how to pass it to Python in GDB other then through an env var but I am not sure if I like it. Apart from that if the test run fails there is an info that one can modify `pytest_launcher.py`, add `--pdb` there and launch tests locally.

2) The test that checks for `pwndbg stack` command list filtering doesn't check for equality now and instead it checks whether things listed in the test exists in the returned list.

3) The ugly `stack_addr = next(iter(pwndbg.stack.stacks.values())).vaddr` stays the same as `pwndbg.regs.stack` is just returning a register name string. I could do `getattr(pwndbg.regs, pwndbg.regs.stack)` but I don't like it that much. Any better idea?

4) Paths are not hardcoded anymore. The `__init__.py` hack/trick for keeping proper file paths is amazing!

5) I moved tests launching/incantation to `tests.sh`, I am not sure about the name as it kinda collide with `tests/` directory...

6) I added a `theme.Parameter` called `disable_colors` which fetches its value on startup from `PWNDBG_DISABLE_COLORS` environment variable. It seems to be working fine and even the prompt color is removed. 

7) I removed the `assert len(commands_list) == 161`

8) The `test_loads` tests asert full gdb output now.

9) Some of the `test_loads` tests are skipped locally but run on travis.

disconnect3d:
The last commit will fix the `test_loads` so they assert full gdb messages. Need to see what travis outputs, huh.

zachriggle (approved):

PR #376, opened 2017-12-19
Add xinfo command for extended offset information

kirschju:
This commit adds a `xinfo` command that calculates the offset of a
specified address to other interesting locations within the address
space:

* In the most general case, simply the offset of the pointer into the
current mapping is displayed.
* If the address specified is a stack adress, the offsets to the top and
the bottom of the stack, as well as to the current stack pointer and
frame pointer value are displayed.
* If the address points into a memory mapped file, the command
additionally shows the offset to the beginning of the file in memory and
on disk.

### DIFF
--- a/pwndbg/commands/canary.py
+++ b/pwndbg/commands/canary.py
@@ -17,20 +17,28 @@ import pwndbg.search
 
 parser = argparse.ArgumentParser(description='Print out the current stack canary')
 
-
-@pwndbg.commands.ArgparsedCommand(parser)
-@pwndbg.commands.OnlyWhenRunning
-def canary():
+def canary_value():
     auxv = pwndbg.auxv.get()
     at_random = auxv.get('AT_RANDOM', None)
     if at_random is None:
-        print("Couldn't find AT_RANDOM - can't display canary.")
-        return
+        return None, None
 
     global_canary = pwndbg.memory.pvoid(at_random)
 
     # masking canary value as canaries on the stack has last byte = 0
     global_canary &= (pwndbg.arch.ptrmask ^ 0xFF)
+
+    return global_canary, at_random
+
+
+@pwndbg.commands.ArgparsedCommand(parser)
+@pwndbg.commands.OnlyWhenRunning
+def canary():
+    global_canary, at_random = canary_value()
+
+    if global_canary is None or at_random is None:
+        print("Couldn't find AT_RANDOM - can't display canary.")
+        return
 
     print("AT_RANDOM = %#x # points to (not masked) global canary value" % at_random)
     print("Canary    = 0x%x" % global_canary)
--- a/pwndbg/commands/xinfo.py
+++ b/pwndbg/commands/xinfo.py
@@ -1,0 +1,105 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import argparse
+import subprocess
+
+import gdb
+
+import pwndbg.arch
+import pwndbg.color.memory as M
+import pwndbg.commands
+import pwndbg.config
+import pwndbg.memory
+import pwndbg.regs
+import pwndbg.stack
+import pwndbg.vmmap
+import pwndbg.wrappers
+
+parser = argparse.ArgumentParser(description='Shows offsets of the specified address to useful other locations')
+parser.add_argument('address', nargs='?', default='$pc', 
+                    help='Address to inspect')
+
+def print_line(name, addr, first, second, op, width = 20):
+
+    print("{} {} = {} {} {:#x}".format(name.rjust(width), M.get(addr),
+        M.get(first) if type(first) is not str else first.ljust(len(hex(addr))),
+        op, second,))
+
+def xinfo_stack(page, addr):
+    # If it's a stack address, print offsets to top and bottom of stack, as
+    # well as offsets to current stack and base pointer (if used by debugee)
+
+    sp = pwndbg.regs.sp
+    frame = pwndbg.regs[pwndbg.regs.frame]
+    frame_mapping = pwndbg.vmmap.find(frame)
+
+    print_line("Stack Top", addr, page.vaddr, addr - page.vaddr, "+")
+    print_line("Stack End", addr, page.end, page.end - addr, "-")
+    print_line("Stack Pointer", addr, sp, addr - sp, "+")
+
+    if frame_mapping and page.vaddr == frame_mapping.vaddr:
+        print_line("Frame Pointer", addr, frame, frame - addr, "-")
+
+    canary_value = pwndbg.commands.canary.canary_value()[0]
+
+    if canary_value is not None:
+        all_canaries = list(
+            pwndbg.search.search(pwndbg.arch.pack(canary_value), mappings=pwndbg.stack.stacks.values())
+        )
+        follow_canaries = sorted(filter(lambda a: a > addr, all_canaries))
+        if follow_canaries is not None and len(follow_canaries) > 0:
+            nxt = follow_canaries[0]
+            print_line("Next Stack Canary", addr, nxt, nxt - addr, "-")
+
+def xinfo_mmap_file(page, addr):
+    # If it's an address pointing into a memory mapped file, print offsets
+    # to beginning of file in memory and on disk
+
+    file_name = page.objfile
+    objpages = filter(lambda p: p.objfile == file_name, pwndbg.vmmap.get())
+    first = sorted(objpages, key = lambda p: p.vaddr)[0]
+    rva = addr - first.vaddr
+
+    print_line("File (Memory)", addr, first.vaddr, rva, "+")
+
+    for segment in pwndbg.wrappers.readelf.get_load_segment_info():
+        if rva >= segment["VirtAddr"] and rva <= segment["VirtAddr"] + segment["MemSiz"]:
+            print_line("File (Disk)", addr, file_name, rva - (segment["VirtAddr"] - segment["Offset"]), "+")
+
+def xinfo_default(page, addr):
+    # Just print the distance to the beginning of the mapping
+
+    print_line("Mapped Area", addr, page.vaddr, addr - page.vaddr, "+")
+
+
+@pwndbg.commands.ArgparsedCommand(parser)
+@pwndbg.commands.OnlyWhenRunning
+def xinfo(address=None):
+    addr = int(address)
+    addr &= pwndbg.arch.ptrmask
+
+    page = pwndbg.vmmap.find(addr)
+
+    if page is None:
+        print("\n  Virtual address {:#x} is not mapped.".format(addr))
+        return
+
+    print("Extended information for virtual address {}:".format(M.get(addr)))
+
+    print("\n  Containing mapping:")
+    print(M.get(address, text=str(page)))
+
+    print("\n  Offset information:")
+
+    if page.is_stack:
+        xinfo_stack(page, addr)
+    else:
+        xinfo_default(page, addr)
+    
+    if page.is_memory_mapped_file:
+        xinfo_mmap_file(page, addr)
--- a/pwndbg/wrappers/readelf.py
+++ b/pwndbg/wrappers/readelf.py
@@ -5,6 +5,8 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import re
+
 import pwndbg.wrappers
 
 cmd_name = "readelf"
@@ -16,7 +18,6 @@ def get_jmpslots():
     readelf_out = pwndbg.wrappers.call_cmd(cmd)
 
     return filter(_extract_jumps, readelf_out.splitlines())
-
 
 def _extract_jumps(line):
     '''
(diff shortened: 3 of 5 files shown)


disconnect3d:
Hey, thanks for the PR! This looks nice.

Please fix the things specified by travis build by running: `isort --recursive pwndbg` from pwndbg directory.

Just for others, here is how it looks:
https://i.gyazo.com/ce22b41ab54e252207ca3b9736aca126.png

Also for some reference/inspiration: http://gef.readthedocs.io/en/master/commands/xinfo/

pwndbg/commands/xinfo.py
```diff
@@ -0,0 +1,100 @@
...
+import pwndbg.regs
+import pwndbg.color.memory as M
+
+
+parser = argparse.ArgumentParser(description='Shows offsets of the specified address to useful other locations')
+parser.add_argument('address', nargs='?', default='$pc', 
+                    help='Address to inspect')
+
+def print_line(name, addr, first, second, op, width = 20):
+    FMT = "{} {} = {} {} {:#x}"
```
disconnect3d: since its used just once, we can probably move the string into `print`

pwndbg/commands/xinfo.py
```diff
@@ -0,0 +1,100 @@
...
+@pwndbg.commands.OnlyWhenRunning
+def xinfo(address=None):
+    addr = int(address)
+    addr &= pwndbg.arch.ptrmask
+
+    page = pwndbg.vmmap.find(addr)
+
+    print("Extended information for virtual address {}:".format(M.get(addr)))
+
+    if page is None:
```
disconnect3d: Maybe we should move it above the `Extended infrmation for virtual address` print? I am not sure...

pwndbg/commands/xinfo.py
```diff
@@ -0,0 +1,100 @@
...
+        print("\n  Virtual address {:#x} is not mapped.".format(addr))
+        return
+
+    print("\n  Containing mapping:")
+    print(M.get(address, text=str(page)))
+
+    print("\n  Offset information:")
+
+    if page.objfile == "[stack]":
```
disconnect3d: Can we add a property to `Page` object called `is_stack`:
```python
@property
def is_stack(self):
    return self.objfile == '[stack]'
```

and use it here?

pwndbg/commands/xinfo.py
```diff
@@ -0,0 +1,100 @@
...
+        # If it's a stack address, print offsets to top and bottom of stack, as
+        # well as offsets to current stack and base pointer (if used by debugee)
+        sp = pwndbg.regs.sp
+        frame = pwndbg.regs[pwndbg.regs.frame]
+        frame_mapping = pwndbg.vmmap.find(frame)
+
+        print_line("Stack Top", addr, page.vaddr, addr - page.vaddr, "+")
```
disconnect3d: We could potentially use https://github.com/pwndbg/pwndbg/blob/ed887fdbc48c27019211d71794f40b94ba94acb5/pwndbg/memory.py#L387-L400 however since we also subtract stuff here, maybe it isn't the best idea.
I leave it up to you.

pwndbg/commands/xinfo.py
```diff
@@ -0,0 +1,100 @@
...
+        # to beginning of file in memory and on disk
+        file_name = page.objfile
+        objpages = filter(lambda p: p.objfile == file_name, pwndbg.vmmap.get())
+        objpages = sorted(objpages, key = lambda p: p.vaddr)
+        first = objpages[0]
+        rva = addr - first.vaddr
+
+        print_line("File (Memory)", addr, first.vaddr, rva, "+")
+
+        p = subprocess.Popen(["readelf", "-l", file_name], stdout = subprocess.PIPE)
```
disconnect3d: Since calling other processes is not that easy topic (what with other achitectures? what if the binary is not present? and so on) we create wrappers for them. Please add appropriate function to https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/wrappers/readelf.py and use it here instead of calling it directly.

NOTE: Thanks to this we can try to use different readelf names for different architectures (for remote debugging, when binary is present locally) and so on (this isn't yet there, but lets have a possibility to add it easily).

pwndbg/commands/xinfo.py
```diff
@@ -0,0 +1,100 @@
...
+        if not p:
+            return
+       
+        load_found = False
+        for l in p.stdout.read().decode().split("\n"):
+            if "LOAD" in l:
+                load_found = True
+                o, v, p = map(lambda x: int(x.strip(), 16), l.split("0x")[1:4])
```
disconnect3d: Please change `o`, `v`, `p` to some more explicit names :F.

pwndbg/commands/xinfo.py
```diff
@@ -0,0 +1,100 @@
...
+        print_line("Stack Pointer", addr, sp, addr - sp, "+")
+
+        if frame_mapping and page.vaddr == frame_mapping.vaddr:
+            print_line("Frame Pointer", addr, frame, frame - addr, "-")
+
+    else:
+        # Otherwise, just print the distance to the beginning of the mapping
+        print_line("Mapped Area", addr, page.vaddr, addr - page.vaddr, "+")
+
+    if len(page.objfile) > 0 and page.objfile[0] != "[":
```
disconnect3d: Probably good candidate for a `Page` object property. If you think the same, go with adding a property :P.

disconnect3d (changes requested):
Apart of my comments I would also change the displayed labels/names:

`File (Memory)` -> `(Virtual) Memory offset`
`File (Disk)` -> `File offset`

disconnect3d:
Also thinks to consider either here or for future improvements:
* add printers for other mappings - heap would be the first candidate I guess
* split logic for different memory pages into different functions - so we can then have something like:
```python
xinfo_printers = {
    '[stack]': xinfo_stack,
    '[heap]': xinfo_heap,
    '[vdso]': xinfo_vdso
}

printer = xinfo_printers.get(page.objfile, xinfo_default)
printer()  # or pass some args (?)
```
* For stack printer: add canary position (see canary command implementation), saved rbp/rip positions

kirschju:
Hi,

thanks for your suggestions, I've incorporated all changes except for the renaming:

```
File (Memory) -> (Virtual) Memory offset
File (Disk) -> File offset
```

I don't mind changing the labels, but for consistency reasons I would like to avoid names that end with "offset".

pwndbg/commands/xinfo.py
```diff
@@ -0,0 +1,113 @@
...
+
+    print("\n  Offset information:")
+
+    if vmmap.is_stack:
+        xinfo_stack(addr)
+    else:
+        xinfo_default(addr)
+    
+    if vmmap.is_memory_mapped_file:
+        xinfo_mmap_file(addr)
```
disconnect3d: Can we pass both `addr` and `vmmap` here so we won't have to call `pwndbg.vmmap.find` in xinfo implementations?

disconnect3d: I'd also suggest renaming `vmmap` to `page`

pwndbg/wrappers/readelf.py
```diff
@@ -33,3 +34,49 @@ def _extract_jumps(line):
...
     except IndexError:
         return False
+
+@pwndbg.wrappers.OnlyWithCommand(cmd_name)
+def get_load_segment_info():
+    '''
+    Looks for LOAD sections by parsing the output of `readelf -l <binary>`
+    '''
+    local_path = pwndbg.file.get_file(pwndbg.proc.exe)
+    cmd = [get_jmpslots.cmd_path, "-l", local_path]
```
disconnect3d: Please change `'-l'` to `--program-headers` both here and in the docstring so it would be easier to understand what is this flag for :)

pwndbg/wrappers/readelf.py
```diff
@@ -33,3 +34,49 @@ def _extract_jumps(line):
...
+    load_found = False
+
+    # Output from readelf is 
+    # Type           Offset             VirtAddr           PhysAddr
+    #                FileSiz            MemSiz             Flags  Align
+    # LOAD           0x0000000000000000 0x0000000000000000 0x0000000000000000
+    #                0x0000000000000830 0x0000000000000830  R E    0x200000
+    # Account for this using two regular expressions
+    re_first = r"\s+LOAD\s+(0x[0-9A-Fa-f]+) (0x[0-9A-Fa-f]+) (0x[0-9A-Fa-f]+)"
+    re_secnd = r"\s+(0x[0-9A-Fa-f]+) (0x[0-9A-Fa-f]+)  (.)(.)(.)\s+(0x[0-9A-Fa-f]+)"
```
disconnect3d: Hm... lets compile those regexes before function definition and since they are constants it might be good to use uppercase naming for them

pwndbg/wrappers/readelf.py
```diff
@@ -33,3 +34,49 @@ def _extract_jumps(line):
...
+    hex2int = lambda x: int(x, 16)
+
+    for l in readelf_out.splitlines():
```
disconnect3d: `l` -> `line`

pwndbg/wrappers/readelf.py
```diff
@@ -33,3 +34,49 @@ def _extract_jumps(line):
...
+        if "LOAD" in l:
+            load_found = True
+            offset, vaddr, paddr = map(hex2int, re.match(re_first, l).groups())
+        elif load_found:
+            fsize, msize, read, write, execute, align = re.match(re_secnd, l).groups()
+            fsize, msize, align = map(hex2int, (fsize, msize, align))
+            read = True if read == "R" else False
```
disconnect3d: Just:
```python
read = read == 'R'
```
same for `write` and `execute`

pwndbg/wrappers/readelf.py
```diff
@@ -33,3 +34,49 @@ def _extract_jumps(line):
...
+            write = True if write == "W" else False
+            execute = True if execute == "E" else False
+
+            segments.append({"Offset":   offset,
```
disconnect3d: Hm.. we might want to have a `class Segment`, like we do have `class Page`; however I am not sure if that is needed right now.

pwndbg/commands/canary.py
```diff
@@ -17,21 +17,29 @@
...
     # masking canary value as canaries on the stack has last byte = 0
     global_canary &= (pwndbg.arch.ptrmask ^ 0xFF)
 
+    return global_canary, at_random
+
+
+@pwndbg.commands.ArgparsedCommand(parser)
+@pwndbg.commands.OnlyWhenRunning
+def canary():
+    global_canary, at_random = canary_value()
```
disconnect3d, line 37: Nice! :heart:

pwndbg/commands/xinfo.py
```diff
@@ -0,0 +1,113 @@
...
+    vmmap = pwndbg.vmmap.find(addr)
+
+    print_line("Mapped Area", addr, vmmap.vaddr, addr - vmmap.vaddr, "+")
+
+
+@pwndbg.commands.ArgparsedCommand(parser)
+@pwndbg.commands.OnlyWhenRunning
+def xinfo(address=None):
+    addr = int(address)
+    addr &= pwndbg.arch.ptrmask
```
disconnect3d, line 84: We probably want to have some custom type for commands that gets expressions as arguments (so we don't double those two lines here in all commands :D), but lets leave it as it is for now.

pwndbg/wrappers/readelf.py
```diff
@@ -33,3 +34,49 @@ def _extract_jumps(line):
...
+    #                FileSiz            MemSiz             Flags  Align
+    # LOAD           0x0000000000000000 0x0000000000000000 0x0000000000000000
+    #                0x0000000000000830 0x0000000000000830  R E    0x200000
+    # Account for this using two regular expressions
+    re_first = r"\s+LOAD\s+(0x[0-9A-Fa-f]+) (0x[0-9A-Fa-f]+) (0x[0-9A-Fa-f]+)"
+    re_secnd = r"\s+(0x[0-9A-Fa-f]+) (0x[0-9A-Fa-f]+)  (.)(.)(.)\s+(0x[0-9A-Fa-f]+)"
+    hex2int = lambda x: int(x, 16)
+
+    for l in readelf_out.splitlines():
+        if "LOAD" in l:
```
disconnect3d: 1. LOAD might appear twice (see screenshot), is that okay?
https://i.gyazo.com/5deb4d216f1db4bd3c8e37fcc76cb71a.png

2. `LOAD` might appear in part of section name or interpreter path (yeah those edge cases...); maybe something like:
```
if line.lstrip().startswith('LOAD '):
``` 
(that's why parsing cli programs sucks on the long term, but it is okay for now; guess we don't want to reimplement readelf yet; and probably this has been done by [pwntools](https://pwntools.com) already, but we can't use it YET since they are not Python 3 compatible).

pwndbg/commands/xinfo.py
```diff
@@ -0,0 +1,113 @@
...
+    vmmap = pwndbg.vmmap.find(addr)
+
+    file_name = vmmap.objfile
+    objpages = filter(lambda p: p.objfile == file_name, pwndbg.vmmap.get())
+    first = sorted(objpages, key = lambda p: p.vaddr)[0]
+    rva = addr - first.vaddr
+
+    print_line("File (Memory)", addr, first.vaddr, rva, "+")
+
+    segments = pwndbg.wrappers.readelf.get_load_segment_info()
```
disconnect3d: redundant line

pwndbg/commands/xinfo.py
```diff
@@ -0,0 +1,113 @@
...
+    
+    for s in pwndbg.wrappers.readelf.get_load_segment_info():
```
disconnect3d: maybe `for segment in ...`

disconnect3d (changes requested):

pwndbg/wrappers/readelf.py
kirschju: `LOAD` appearing multiple times is okay and indended, yes. I like the `startswith` suggestion.

pwndbg/wrappers/readelf.py
kirschju: Depends on your preference of who eventually will take care of readelf functionality. If you plan to incorporate pwntools on the long run, you will have to adopt to their code anyway ...

disconnect3d: We are going to merge with OR use pwntools on the long run :).

disconnect3d:
@kirschju Is that ready to be merged?
About naming, imho the `... offset` is easier to understand

kirschju:
@disconnect3d Sure, ready to merge. Still no idea for the naming, I'll leave it as is, feel free to put something more suitable if you feel like.

disconnect3d (approved):
lgtm, thanks! :heart:

PR #377, opened 2017-12-23
Fix upper stack boundary

disconnect3d:
Fixes bug with getting `upper_stack_boundary` introduced in https://github.com/pwndbg/pwndbg/commit/31f468e0d1c41bcb8a259610b6e42750050aa63f.

TL;DR: `upper_stack_boundary` we got doesn't match the one from `vmmap`.
This breaks `canary` command - and probably something else that is based on stack pages.

Previously we determined upper address by having a memory read failure.
Recent changes made it so we got a `None` instead of the address in such situation.

This adds a parameter to `find_elf_magic` which lets us get a result when gdb.MemoryError occurs.

---

This PR fixes the bug but I am not sure whether this is 100% correct as we may set stack upper boundary for an address that contains `'\x7fELF'` because of this:
```python
        # Return the address if found ELF header
        if data == b'\x7fELF':
            return addr
```

Weird thing: I can actually search for `ELF` and get a result on the stack - then check returned address-1 and see that there is `\x7fELF` string (btw we can't use `\x7f` to search for 0x7f byte with `search` command ;/).
But this was not set as stack upper boundary.

### DIFF
--- a/pwndbg/elf.py
+++ b/pwndbg/elf.py
@@ -67,6 +67,7 @@ def read(typ, address, blob=None):
     obj.type = typ
     return obj
 
+
 @pwndbg.proc.OnlyWhenRunning
 @pwndbg.memoize.reset_on_start
 def exe():
@@ -77,6 +78,7 @@ def exe():
     e = entry()
     if e:
         return load(e)
+
 
 @pwndbg.proc.OnlyWhenRunning
 @pwndbg.memoize.reset_on_start
@@ -117,10 +119,12 @@ def load(pointer):
 
 ehdr_type_loaded = 0
 
+
 @pwndbg.memoize.reset_on_start
 def reset_ehdr_type_loaded():
     global ehdr_type_loaded
     ehdr_type_loaded = 0
+
 
 @pwndbg.abi.LinuxOnly()
 def find_elf_magic(pointer, max_pages=1024, search_down=False, ret_addr_anyway=False):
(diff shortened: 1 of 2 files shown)


disconnect3d:
@zachriggle @cebrusfs please review. Would be nice to merge it sooner then later as current dev is broken. 

I am also marking @cebrusfs as it is connected ti his changes.

Maybe we should get back the old `upper_stack_boundary` code?:
https://github.com/pwndbg/pwndbg/blob/8a5a6da0a3d14e8a26b135f589a458be2b302972/pwndbg/stack.py#L42-L52

However I am still thinking that the check for `'\x7fELF'` is wrong.

disconnect3d:
Oh I found out the reason for this ELF check and a copy-paste of the old function:
https://github.com/pwndbg/pwndbg/blob/fe876b1029976f2745f3ed2e70f430141dd4665f/pwndbg/auxv.py#L131-L151

zachriggle:
I don’t know why the variable rename was needed, but LGTM

disconnect3d:
@zachriggle For me `base` is something that would rather be constant - and we change it (well we are finding its value, heh).

disconnect3d:
Merged it for now. If anyone wants to come up with some cleaner solution, feel free to do so.

zachriggle:
❤️

cebrusfs:
LGTM.
Sorry that I just want to remove the duplicate code and it seems that I miss the `addr` will be changed during the searching.

Issue #371, opened 2017-12-16
Setting to disable the telescope


AlexAltea:
I use pwndbg to debug an operating system and some device drivers running under QEMU. The problem is that while doing so, some registers may hold MMIO register addresses which the telescope attempts to dereference. This not only causes extra annoying logs, but for several reasons also creates non-determinism despite issuing only read queries.

This is similar to the 2nd point mentioned in issue https://github.com/pwndbg/pwndbg/pull/369, and might (?) be fixed by https://github.com/pwndbg/pwndbg/pull/367. If I understood that pull request correctly, I should be able to enter `set dereference-limit 0` to disable the telescope.

disconnect3d:
@AlexAltea the #367 has been merged.  Does that work for you?

PS: Might be worth to use code from https://github.com/pwndbg/pwndbg/pull/377as #369 made getting stack upper bound wrong (we need tests ASAP!).

AlexAltea:
Although #367 does its job, it does not solve my problem. Despite setting `set dereference-limit 0`, I still get memory reads (unrelated to the code that it's being debugged) while stepping through the code. I'm not sure what is causing that nor how to debug this issue.

I can only confirm that on stock GDB on TUI mode with asm+regs layouts that issue doesn't happen.

disconnect3d:
@AlexAltea Could you prepare some minimal working example that reproduces the error so we could see that (and add a testcase for it in the future)?

AlexAltea:
@disconnect3d I've been debugging the issue myself, since it's hard making a minimal test case, and sending you a 20 GB .ovf file of pre-configured VM would be a little too extreme. :-)

Turns out:
- `set dereference-limit 0` (from #367) does indeed fix some unwanted reads.
- `set emulate off` fixes more unwanted reads triggered by Unicorn.

But there's still few unwanted memory reads that I wasn't able to remove. Here's the function call stack that leads to them:

```
File "/home/user/Desktop/pwndbg/pwndbg/prompt.py", line 37, in prompt_hook
    prompt_hook_on_stop(*a)
File "/home/user/Desktop/pwndbg/pwndbg/memoize.py", line 48, in __call__
    value = self.func(*args, **kwargs)
File "/home/user/Desktop/pwndbg/pwndbg/prompt.py", line 42, in prompt_hook_on_stop
    pwndbg.commands.context.context()
File "/home/user/Desktop/pwndbg/pwndbg/commands/__init__.py", line 106, in __call__
[...]

PR #380, opened 2017-12-24
Avoid enhancements when dereference limit is zero

AlexAltea:
This fixes and off by one issue mentioned in https://github.com/pwndbg/pwndbg/issues/371#issuecomment-353741672 and modifies the `pwndbg/chain.py` file to prevent accidentally dereferencing addresses by avoiding enhancements when the dereference limit is set to zero. I have made some tests and I have found no regressions. However, unit tests would be ideal. On the diff, I will comment most changes line by line.

### DIFF
--- a/pwndbg/chain.py
+++ b/pwndbg/chain.py
@@ -94,9 +94,11 @@ def format(value, limit=LIMIT, code=True, offset=0, hard_stop=None, hard_end=0):
             symbol = '%#x (%s)' % (link, symbol)
         rest.append(M.get(link, symbol))
 
-    # If the dereference limit is zero, skip enhancement and return the chain
+    # If the dereference limit is zero, skip any enhancements.
     if limit == 0:
-        return arrow_right.join(rest)
+        return rest[0]
+    # Otherwise replace last element with the enhanced information.
+    rest = rest[:-1]
 
     # Enhance the last entry
     # If there are no pointers (e.g. eax = 0x41414141), then enhance
--- a/pwndbg/chain.py
+++ b/pwndbg/chain.py
@@ -32,9 +32,9 @@ def get(address, limit=LIMIT, offset=0, hard_stop=None, hard_end=0):
     Returns:
         A list representing pointers of each ```address``` and reference
     """
-    limit = int(limit)+1
+    limit = int(limit)
     
-    result = []
+    result = [address]
     for i in range(limit):
         # Don't follow cycles, except to stop at the second occurrence.
         if result.count(address) >= 2:
@@ -44,10 +44,10 @@ def get(address, limit=LIMIT, offset=0, hard_stop=None, hard_end=0):
             result.append(hard_end)
             break
 
-        result.append(address)
         try:
             address = int(pwndbg.memory.poi(pwndbg.typeinfo.ppvoid, address + offset))
             address &= pwndbg.arch.ptrmask
+            result.append(address)
         except gdb.MemoryError:
             break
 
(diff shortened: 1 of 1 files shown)


pwndbg/chain.py
```diff
@@ -32,9 +32,9 @@ def get(address, limit=LIMIT, offset=0, hard_stop=None, hard_end=0):
     Returns:
         A list representing pointers of each ```address``` and reference
     """
-    limit = int(limit)+1
+    limit = int(limit)
```
AlexAltea, line 35: The docs for `get` say:

> `limit(int): number of valid pointers`

However if limit is 0 it will increment it to 1 and try to dereference the address once, which is unwanted.

pwndbg/chain.py
```diff
@@ -32,9 +32,9 @@ def get(address, limit=LIMIT, offset=0, hard_stop=None, hard_end=0):
...
     
-    result = []
+    result = [address]
```
AlexAltea, line 37: Regardless of what the limit value is, the chain should always start with the given address.

pwndbg/chain.py
```diff
@@ -75,17 +75,29 @@ def format(value, limit=LIMIT, code=True, offset=0, hard_stop=None, hard_end=0):
         A string representing pointers of each address and reference
         Strings format: 0x0804a10 —▸ 0x08061000 ◂— 0x41414141
     """
-    limit = int(limit)+1
+    limit = int(limit)
 
     # Allow results from get function to be passed to format
-    if type(value) == list:
+    if isinstance(value, list):
```
AlexAltea, line 81: > ```
> <disconnect3d> AlexAltea: btw
> <disconnect3d> can u also change `if type(value) == list:`
> <disconnect3d> to `if isinstance(value, list):`
> ```

pwndbg/chain.py
```diff
@@ -75,17 +75,29 @@ def format(value, limit=LIMIT, code=True, offset=0, hard_stop=None, hard_end=0):
         A string representing pointers of each address and reference
         Strings format: 0x0804a10 —▸ 0x08061000 ◂— 0x41414141
     """
-    limit = int(limit)+1
+    limit = int(limit)
```
AlexAltea, line 78: The docs for `format` say:

> `limit(int): number of valid pointers`

However if limit is 0 it will increment it to 1 and try to dereference the address once (during `get`), which is unwanted.

pwndbg/chain.py
```diff
@@ -95,20 +107,12 @@ def format(value, limit=LIMIT, code=True, offset=0, hard_stop=None, hard_end=0):
     # Otherwise, the last element in the chain is the non-pointer value.
     # We want to enhance the last pointer value. If an offset was used
     # chain failed at that offset, so display that offset.
-    elif len(chain) < limit:
+    elif len(chain) < limit + 1:
```
AlexAltea, line 112: Since we are no longer doing that nasty `limit = int(limit)+1` hack, you would expect that if all addresses in the chain were successfully dereferenced, then `len(chain) == limit + 1`.

If the chain failed to dereference any entry, then `len(chain) <= limit`, in other words, `len(chain) < limit + 1`.

disconnect3d:
Some screens for reference.

This looks awesome (when I played with it yesterday I got some of those enhanced so it would show the address as asm instruction for registers lol):
https://i.gyazo.com/bfb75e90f1909ebd8e2b987071664bff.png

Limit set to 1:
https://i.gyazo.com/f8d907b1f0b163f835f1a460e9e152ef.png

Anyway, this has some "breaking changes" - see here - left is this PR, right is current dev:
https://i.gyazo.com/55bf81cbcb60239d59ec6c1006001862.png
So we display the value twice - once as raw hex and second time enhanced - previously we just show it enhanced.

AlexAltea:
> Anyway, this has some "breaking changes" - see here - left is this PR, right is current dev

Oh, while updating the code I actually identify the previous behavior as unwanted, so I replaced:
```diff
-   for link in chain[:-1]:
+   for link in chain:
```
I did it since I was under the impression that enhancements should never be a replacement for the actual value, so I preserved the last value of the chain.

We could discuss what strategy makes more sense... or if you absolutely want to revert this change, we can do it, but add `if limit == 0: chain = [address, address]` so that `chain[:-1]` doesn't return an empty list whenever the dereference limit is zero.

pwndbg/chain.py
zachriggle: The reason for the `+1` is the way that `range()` works

disconnect3d:
I have some mixed feelings about this double display (raw value and enhanced) - this is cool in some cases but on the other hand it adds to the display/line length ;s.

@zachriggle what do you think?

zachriggle:
I didn't notice any particular change, but this PR shouldn't merge if it changes anything other than just having an working dereference limit == 0.

EDIT: Saw the screenshots.  Yeah, please return the functionality to how it was :stuck_out_tongue:

AlexAltea:
@disconnect3d @zachriggle Ok, that behaviour is now reverted. The last element in chains will be omitted, as it used to be.

disconnect3d (approved):
Looks good to me after some tests. Sorry for the long delay/response.

PR #385, opened 2017-12-26
Make de-reference only works on known pages in bare metal mode and add commands to manually add pages

cebrusfs:
Since the address are all valid when mmu is not enable and all the value are valid physical address. It will be de-referenced even these addresses are not used. Actually, it is data in the most of case. Ex. 0x1 often means the value 1, not the address 0x1.

Also, for issue #371, some addresses may be the MMIO registers. The read operation on these address will break the state. It is better to limit the de-reference address range. This PR will also fix it, hopefully.

First, we limit the de-reference only happened when the value is in the known page range.
Moreover, I create some API in `vmmap.py` to maintain the manually added pages. Also, add some commands called `vmmap_add`, `vmmap_del`, `vmmap_load` to add pages, delete pages, load page and guess the permission from ELF sections information.

1. I am out of idea about the naming of variable name and command name. Any better suggestions are welcome.
2. `vmmap.get()` has cache. The manually adding pages will not work immediately.

----

The following issues have been found while I was creating this PR, but not related:
1. Use the different source to get the page permission `RWX` value.
Ex. in `memory.py` use `os.R`, in `elf.py` use `PF_R`, and in `vmmap.py` use a hardcoded value `4`.
2. gdb `load` command will not generate the gdb event so that it will not clear the cache and update to new information.
3. The implementation of `vmmap_load` is similar with some code in `elf.py` to parse the ELF sections. Maybe we can create some useful ELF util functions and make 2 part use the same function.

### DIFF
--- a/pwndbg/abi.py
+++ b/pwndbg/abi.py
@@ -164,5 +164,5 @@ def LinuxOnly(default=None):
     return decorator
 
 
-# XXX: Update when starting the gdb to show warning message for non-Linux ABI user.
+# Update when starting the gdb to show warning message for non-Linux ABI user.
 update()
--- a/pwndbg/abi.py
+++ b/pwndbg/abi.py
@@ -10,6 +10,7 @@ import re
 import gdb
 
 import pwndbg.arch
+import pwndbg.color.message as M
 
 
 class ABI(object):
@@ -136,6 +137,17 @@ def update():
 
         linux = 'Linux' in abi
 
+    if not linux:
+        msg = M.warn(
+            "The bare metal debugging will enable since the gdb's osabi is '%s' which is not 'GNU/Linux'.\n"
+            "Ex. the page resolving and memory de-referencing ONLY works on known pages.\n"
+            "This option is seleted from compile arguments of gdb client and will be corrected if you load an ELF that contains ABI information.\n"
+            "If you are debuging some program that running on Linux ABI, please select the correct gdb client."
+            % abi
+        )
+        print(msg)
+
+
 def LinuxOnly(default=None):
     """Create a decorator that the function will be called when ABI is Linux.
     Otherwise, return `default`.
@@ -150,3 +162,7 @@ def LinuxOnly(default=None):
         return caller
 
     return decorator
+
+
+# XXX: Update when starting the gdb to show warning message for non-Linux ABI user.
+update()
--- a/pwndbg/abi.py
+++ b/pwndbg/abi.py
@@ -139,10 +139,10 @@ def update():
 
     if not linux:
         msg = M.warn(
-            "The bare metal debugging will enable since the gdb's osabi is '%s' which is not 'GNU/Linux'.\n"
+            "The bare metal debugging is enabled since the gdb's osabi is '%s' which is not 'GNU/Linux'.\n"
             "Ex. the page resolving and memory de-referencing ONLY works on known pages.\n"
-            "This option is seleted from compile arguments of gdb client and will be corrected if you load an ELF that contains ABI information.\n"
-            "If you are debuging some program that running on Linux ABI, please select the correct gdb client."
+            "This option is based ib gdb client compile arguments (by default) and will be corrected if you load an ELF which has the '.note.ABI-tag' section.\n"
+            "If you are debuging a program that runs on Linux ABI, please select the correct gdb client."
             % abi
         )
         print(msg)
--- a/pwndbg/chain.py
+++ b/pwndbg/chain.py
@@ -7,6 +7,7 @@ from __future__ import unicode_literals
 
 import gdb
 
+import pwndbg.abi
 import pwndbg.color.chain as C
 import pwndbg.color.memory as M
 import pwndbg.color.theme as theme
@@ -45,7 +46,14 @@ def get(address, limit=LIMIT, offset=0, hard_stop=None, hard_end=0):
             break
 
         try:
-            address = int(pwndbg.memory.poi(pwndbg.typeinfo.ppvoid, address + offset))
+            address = address + offset
+
+            # If in bare metal mode, check the vmmap pages for the address
+            # to avoid de-reference the unused address
+            if not pwndbg.abi.linux and not pwndbg.vmmap.find(address):
+                break
+
+            address = int(pwndbg.memory.poi(pwndbg.typeinfo.ppvoid, address))
             address &= pwndbg.arch.ptrmask
             result.append(address)
         except gdb.MemoryError:
--- a/pwndbg/chain.py
+++ b/pwndbg/chain.py
@@ -21,7 +21,7 @@ LIMIT = pwndbg.config.Parameter('dereference-limit', 5, 'max number of pointers 
 
 def get(address, limit=LIMIT, offset=0, hard_stop=None, hard_end=0):
     """
-    Recursively dereferences an address.
+    Recursively dereferences an address. For bare metal, it will stop when the address is not in any of vmmap pages to avoid redundant dereference.
 
     Arguments:
         address(int): the first address to begin dereferencing
@@ -34,7 +34,7 @@ def get(address, limit=LIMIT, offset=0, hard_stop=None, hard_end=0):
         A list representing pointers of each ```address``` and reference
     """
     limit = int(limit)
-    
+
     result = [address]
     for i in range(limit):
         # Don't follow cycles, except to stop at the second occurrence.
@@ -48,8 +48,8 @@ def get(address, limit=LIMIT, offset=0, hard_stop=None, hard_end=0):
         try:
             address = address + offset
 
-            # If in bare metal mode, check the vmmap pages for the address
-            # to avoid de-reference the unused address
+            # Avoid redundant dereferences in bare metal mode by checking
+            # if address is in any of vmmap pages
             if not pwndbg.abi.linux and not pwndbg.vmmap.find(address):
                 break
 
--- a/pwndbg/commands/vmmap.py
+++ b/pwndbg/commands/vmmap.py
@@ -12,10 +12,13 @@ import argparse
 
 import gdb
 import six
+from elftools.elf.constants import SH_FLAGS
+from elftools.elf.elffile import ELFFile
 
 import pwndbg.color.memory as M
 import pwndbg.commands
 import pwndbg.compat
+import pwndbg.elf
 import pwndbg.vmmap
 
 
@@ -54,3 +57,81 @@ def vmmap(pages_filter=None):
     print(M.legend())
     for page in pages:
         print(M.get(page.vaddr, text=str(page)))
+
+
+parser = argparse.ArgumentParser()
+parser.description = 'Add Print virtual memory map page.'
+parser.add_argument('start', help='Starting virtual address')
+parser.add_argument('size', help='Size of the address space, in bytes')
+parser.add_argument('flags', nargs='?', type=str, default='', help='Flags set by the ELF file, see PF_X, PF_R, PF_W')
+parser.add_argument('offset', nargs='?', default=0, help='Offset into the original ELF file that the data is loaded from')
+
+@pwndbg.commands.ArgparsedCommand(parser)
+def vmmap_add(start, size, flags, offset):
+    page_flags = {
+        'r': pwndbg.elf.PF_R,
+        'w': pwndbg.elf.PF_W,
+        'x': pwndbg.elf.PF_X,
+    }
+    perm = 0
+    for flag in flags:
+        flag_val = page_flags.get(flag, None)
+        if flag_val is None:
+            print('Invalid page flag "%s"', flag)
+            return
+        perm |= flag_val
+
+    page = pwndbg.memory.Page(start, size, perm, offset)
+    pwndbg.vmmap.add_custom_page(page)
+
+    print('%r added' % page)
+
+
+@pwndbg.commands.ParsedCommand
+def vmmap_clear():
+    pwndbg.vmmap.clear_custom_page()
+
+
+parser = argparse.ArgumentParser()
+parser.description = 'Load virtual memory map pages from ELF file.'
+parser.add_argument('filename', nargs='?', type=str, help='Filename of ELF, default use the value of pwndbg.proc.exe')
+
+@pwndbg.commands.ArgparsedCommand(parser)
+def vmmap_load(filename):
+    if filename is None:
+        filename = pwndbg.proc.exe
+
+    print('Load "%s" ...' % filename)
+
+    # Use sections information to recover the segments information.
+    # The entry point of bare metal enviroment is often at the first segments.
+    # This may cause a value < 0x8000 will be considered as a valid pointer.
+    # Ex. entry point is at 0x8000 and its segment starts with 0x0.
+    # TODO: make an argument to choose loading the page information from section or segments
+    pages = []
+    with open(filename, 'rb') as f:
+        elffile = ELFFile(f)
+
+        for section in elffile.iter_sections():
+            vaddr = section['sh_addr']
+            memsz = section['sh_size']
+            sh_flags = section['sh_flags']
+            offset = section['sh_offset']
+
+            # Don't add the sections that aren't mapped into memory
+            if not sh_flags & SH_FLAGS.SHF_ALLOC:
+                continue
+
+            # Guess the segment flags from section flags
+            flags = pwndbg.elf.PF_R
+            if sh_flags & SH_FLAGS.SHF_WRITE:
+                flags |= pwndbg.elf.PF_W
+            if sh_flags & SH_FLAGS.SHF_EXECINSTR:
+                flags |= pwndbg.elf.PF_X
+
+            page = pwndbg.memory.Page(vaddr, memsz, flags, offset, filename)
+            pages.append(page)
+
+    for page in pages:
+        pwndbg.vmmap.add_custom_page(page)
+        print('%r added' % page)
--- a/pwndbg/commands/vmmap.py
+++ b/pwndbg/commands/vmmap.py
@@ -94,7 +94,7 @@ def vmmap_clear():
 
 parser = argparse.ArgumentParser()
 parser.description = 'Load virtual memory map pages from ELF file.'
-parser.add_argument('filename', nargs='?', type=str, help='Filename of ELF, default use the value of pwndbg.proc.exe')
+parser.add_argument('filename', nargs='?', type=str, help='ELF filename, by default uses current loaded filename.')
 
 @pwndbg.commands.ArgparsedCommand(parser)
 def vmmap_load(filename):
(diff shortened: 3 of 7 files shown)


cebrusfs:
The test fails since there are 159 commands rather than 156. I will fix this after the everyone agrees this PR's change. It is a draft now and may have some part need to improve.

For the command interface design part, I have the other approach idea. How about make `add`, `del`, `load` as subcommands of `vmmap`. Then, we can use `vmmap add` comparing to `vmmap_add`.

zachriggle:
This needs to be behind a default-disabled flag.

Exploratory memory dereferencing is required for QEMU user-mode debugging.

cebrusfs:
Will it enable by default? I thought the `pwndbg.abi.linux` should be `True` under the QEMU so that it should not enabled.
Because I use `show osabi` to detect this and it should be only affected by the gdb client version, according to the discussion in issue #264.

This feature should not enable by default. There should be something wrong if it did. Can you provide more information about the environment under the QEMU user mode? Thanks.

cebrusfs:
@zachriggle 
After some searching on the internet, it seems that people use `arm-none-eabi-gdb` to do the ARM debugging. Did you also use it for QEMU debugging? if not, which gdb client did you use?

I traced some gdb source. The default osabi is decided by configuring  `--target=` parameters. Then, it will automatic detect the ABI from ELF section `.note.ABI-tag` if you load the ELF for gdb.

zachriggle:
Which build of GDB you use shouldn't dictate how Pwndbg behaves, and it'll make bugs harder to track down.  I generally use the `gdb-multiarch` which comes with Ubuntu, but I know I've used a `-none` ABI binary for lots of things before -- not sure about GDB.

Perhaps we could just add a setting / `pwndbg.config.Parameter`? e.g. `set baremetal on`

Baremetal users can then just add it to the appropriate `.gdbinit` and get the feature they want.

EDIT: Apparently you can just `set osabi none`.  I don't mind detecting based on `osabi`, though it is unfortunate that `arm-none-eabi-gdb` will default to `none` rather than `auto`.

cebrusfs:
Yeap, we can use parameter to do that.

However, from the conclusion of issue #264, it seems that people want to detect the bare metal mode automatically.  This PR reply on the API `pwndbg.abi.linux` which is created by PR #369. That is, that changes will also break your situation. If you want a manually controlled and default-disabled flag, we should fix that API and create a wrapper getter function to fetch and combine the value of config and the result of automatic ABI detection.

A good new is that `gdb-multiarch` will use GNU abi by default.
Until now, only people use `arm-none-eabi-gdb` and the ELF that gdb load doesn't contain ELF section `.note.ABI-tag` will be switched to bare metal mode.
By designed the `arm-none-eabi-gdb` should not be used for Linux ABI debugging. Since I use the `arm-none-eabi-gdb` to load a Linux ABI ELF, it will show a warning
```
warning: A handler for the OS ABI "GNU/Linux" is not built into this configuration
of GDB.  Attempting to continue with the default armv4t settings.
```

My opinion is that people should not use wrong gdb version for debugging, and the ABI detection API also based on compiling argument. It is a reasonable method to get the current ABI of gdb.
It is wired that we default use Linux ABI debugging mode in a bare metal debugging purpose gdb.
What do you think?

EDIT: sorry I click the submit button when I typing, the content is updated.

cebrusfs:
UPDATE:
1. rebased 
2. add warning message when enabled
3. update the test script of command amount to pass the CI.

Can you all review this PR and give me some feedback?


@zachriggle 
I added a warning message to warn people who use the wrong gdb client.

For now, the bare metal mode will enable only when
1. use gdb which is compiled for bare metal (ex. `arm-none-eabi-gdb`)
2. manually set `osabi` to `none`

For people in 1., this option will fixed by gdb automatically if they load (ex. use `file` command or `gdb main.elf`) an ELF that contains `.note.ABI-tag` sections which will be added by gnu compiler (default options).


@disconnect3d 
Do you have any comments?

disconnect3d:
I will look over in the evening

11.01.2018 11:31 AM "Meng-Huan Yu" <notifications@github.com> napisał(a):

pwndbg/abi.py
```diff
@@ -136,6 +137,17 @@ def update():
 
         linux = 'Linux' in abi
 
+    if not linux:
+        msg = M.warn(
+            "The bare metal debugging will enable since the gdb's osabi is '%s' which is not 'GNU/Linux'.\n"
+            "Ex. the page resolving and memory de-referencing ONLY works on known pages.\n"
+            "This option is seleted from compile arguments of gdb client and will be corrected if you load an ELF that contains ABI information.\n"
+            "If you are debuging some program that running on Linux ABI, please select the correct gdb client."
```
disconnect3d: `running` -> `runs`

pwndbg/abi.py
```diff
@@ -136,6 +137,17 @@ def update():
 
         linux = 'Linux' in abi
 
+    if not linux:
+        msg = M.warn(
+            "The bare metal debugging will enable since the gdb's osabi is '%s' which is not 'GNU/Linux'.\n"
```
disconnect3d: `will enable` -> `is enabled`

pwndbg/abi.py
```diff
@@ -136,6 +137,17 @@ def update():
...
+            "Ex. the page resolving and memory de-referencing ONLY works on known pages.\n"
+            "This option is seleted from compile arguments of gdb client and will be corrected if you load an ELF that contains ABI information.\n"
```
disconnect3d: `This option is seleted from compile arguments of gdb client` -> `This option is based on gdb client compile arguments`

However maybe we should add also: `(as this specified the default osabi)` or sth like that - not sure in here.

disconnect3d: `load an ELF that contains ABI information` - this is also not clear what does it mean. Can we at least have a comment in this file about when or where the ELF contains it?

---
Below comment might be a different PR (or at least an issue pls).

We should probably start thinking about proper documentation for pwndbg (and have a `baremetal` section there).

Maybe it would be good to add `baremetal.md` file? @zachriggle how should we approach docs?

pwndbg/abi.py
```diff
@@ -150,3 +162,7 @@ def caller(*args, **kwargs):
         return caller
 
     return decorator
+
+
+# XXX: Update when starting the gdb to show warning message for non-Linux ABI user.
```
disconnect3d: Can we just do that? Like `if non_linux_abi: warn_not_linux()` or sth

pwndbg/chain.py
```diff
@@ -45,7 +46,14 @@ def get(address, limit=LIMIT, offset=0, hard_stop=None, hard_end=0):
             break
 
         try:
-            address = int(pwndbg.memory.poi(pwndbg.typeinfo.ppvoid, address + offset))
+            address = address + offset
+
+            # If in bare metal mode, check the vmmap pages for the address
+            # to avoid de-reference the unused address
```
disconnect3d: Hmm maybe:

```python
# Avoid redundant dereferences in bare metal mode by checking 
# if address is in any of vmmap pages
```

Please add also some info to `get` docstring about this

pwndbg/commands/vmmap.py
```diff
@@ -54,3 +57,81 @@ def vmmap(pages_filter=None):
     print(M.legend())
     for page in pages:
         print(M.get(page.vaddr, text=str(page)))
+
+
+parser = argparse.ArgumentParser()
+parser.description = 'Add Print virtual memory map page.'
+parser.add_argument('start', help='Starting virtual address')
+parser.add_argument('size', help='Size of the address space, in bytes')
+parser.add_argument('flags', nargs='?', type=str, default='', help='Flags set by the ELF file, see PF_X, PF_R, PF_W')
```
disconnect3d, line 66: This could probably check whether the passed chars are in `'r', 'w', 'x'`
But the current approach with checking those inside and printing out error works too...

So its up to you ;P (unless @zachriggle or @anthraxx really wants it in argparse).

pwndbg/commands/vmmap.py
```diff
@@ -54,3 +57,81 @@ def vmmap(pages_filter=None):
...
+        flag_val = page_flags.get(flag, None)
+        if flag_val is None:
+            print('Invalid page flag "%s"', flag)
+            return
+        perm |= flag_val
+
+    page = pwndbg.memory.Page(start, size, perm, offset)
+    pwndbg.vmmap.add_custom_page(page)
+
+    print('%r added' % page)
```
disconnect3d, line 87: Should probably use some `pwndbg.*.message` added recently.
@anthraxx ^

pwndbg/commands/vmmap.py
```diff
@@ -54,3 +57,81 @@ def vmmap(pages_filter=None):
...
+
+
+@pwndbg.commands.ParsedCommand
+def vmmap_clear():
+    pwndbg.vmmap.clear_custom_page()
+
+
+parser = argparse.ArgumentParser()
+parser.description = 'Load virtual memory map pages from ELF file.'
+parser.add_argument('filename', nargs='?', type=str, help='Filename of ELF, default use the value of pwndbg.proc.exe')
```
disconnect3d: Maybe: `'ELF filename, by default uses current loaded filename.`
(the user might not be aware of `pwndbg.proc.exe`)

pwndbg/commands/vmmap.py
```diff
@@ -54,3 +57,81 @@ def vmmap(pages_filter=None):
...
+
+@pwndbg.commands.ArgparsedCommand(parser)
+def vmmap_load(filename):
+    if filename is None:
+        filename = pwndbg.proc.exe
+
+    print('Load "%s" ...' % filename)
+
+    # Use sections information to recover the segments information.
+    # The entry point of bare metal enviroment is often at the first segments.
```
disconnect3d: `segments` or `segment`?

pwndbg/commands/vmmap.py
```diff
@@ -54,3 +57,81 @@ def vmmap(pages_filter=None):
...
+    # This may cause a value < 0x8000 will be considered as a valid pointer.
+    # Ex. entry point is at 0x8000 and its segment starts with 0x0.
+    # TODO: make an argument to choose loading the page information from section or segments
```
disconnect3d: `make` -> `add`
if we are not going to add it in this PR we should probably add an issue for that (or maybe have one big issue with all TODOS for bare metal?)

pwndbg/commands/vmmap.py
```diff
@@ -54,3 +57,81 @@ def vmmap(pages_filter=None):
...
+                continue
+
+            # Guess the segment flags from section flags
+            flags = pwndbg.elf.PF_R
+            if sh_flags & SH_FLAGS.SHF_WRITE:
+                flags |= pwndbg.elf.PF_W
+            if sh_flags & SH_FLAGS.SHF_EXECINSTR:
+                flags |= pwndbg.elf.PF_X
+
+            page = pwndbg.memory.Page(vaddr, memsz, flags, offset, filename)
```
disconnect3d, line 134: Loud thinking: maybe we should move this to a `Page.from_elf_sections(filepath)` staticmethod which would return list of `Page` objects.

pwndbg/elf.py
```diff
@@ -126,7 +126,6 @@ def reset_ehdr_type_loaded():
     ehdr_type_loaded = 0
 
 
-@pwndbg.abi.LinuxOnly()
```
disconnect3d: I know `find_elf_magic` is no used anymore for bare metal, but maybe we should leave `LinuxOnly` just to be super sure.
@zachriggle @anthraxx any thoughts?

pwndbg/stack.py
```diff
@@ -45,6 +45,11 @@ def find(address):
 def find_upper_stack_boundary(addr, max_pages=1024):
     addr = pwndbg.memory.page_align(int(addr))
 
+    # For bare metal, we can not get the stack size from stack layout and page fault.
+    # Return current page for a walkaround.
```
disconnect3d: Maybe:

```python
# We can't get the stack size from stack layout and page fault on bare metal mode,
# so we return current page as a walkaround.
```

Btw this might be added to some baremetal docs too.

pwndbg/vmmap.py
```diff
@@ -34,6 +34,7 @@
 # List of manually-explored pages which were discovered
 # by analyzing the stack or register context.
 explored_pages = []
+custom_pages = []
```
disconnect3d: This looks as if the upper comment is also about `custom_pages`, please add:
`# List of custom pages that can be managed manually by vmmap_* commands family`

pwndbg/vmmap.py
```diff
@@ -114,6 +116,16 @@ def clear_explored_pages():
     while explored_pages:
         explored_pages.pop()
 
+
+def add_custom_page(page):
+    custom_pages.append(page)
+    custom_pages.sort()
```
disconnect3d: Or just `bisect.insort(custom_pages, page)` :

```
In [2]: import bisect

In [3]: a = [1, 3, 4]

In [4]: bisect.insort(a, 2)

In [5]: a
Out[5]: [1, 2, 3, 4]

In [6]: bisect.insort(a, -1)

In [7]: a
Out[7]: [-1, 1, 2, 3, 4]
```

disconnect3d (changes requested):

disconnect3d:
Thanks for this PR :heart:, hope you won't frighten by the conversation/reviews :P.

> I am out of idea about the naming of variable name and command name. Any better suggestions are welcome.

The names are fine for me.

> vmmap.get() has cache. The manually adding pages will not work immediately.

It should be possible to clear it with `vmmap.get._reset()` (fyi there is [also option to reset all caches](https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/memoize.py#L172-L179))

> For the command interface design part, I have the other approach idea. How about make add, del, load as subcommands of vmmap. Then, we can use vmmap add comparing to vmmap_add.

Subcommands :+1: :heart: - however I am not sure how this would play with our current `ArgparsedCommand`. If you want to work on improving it, feel free to do so.

> The implementation of vmmap_load is similar with some code in elf.py to parse the ELF sections. Maybe we can create some useful ELF util functions and make 2 part use the same function.

Good idea, feel free to do so :+1:; this is probably the thing I described as `Page.from_elf_sections(filepath)` staticmethod.

PS: If we ever meet on a conference I am going to buy you :beer: (just remind me pls :D).

pwndbg/abi.py
cebrusfs: The the abi information is updated by `update()`. We have to update it at least one time.
The other approach is change `linux` variable to a getter of module and calculate each time (since the information fetching from gdb is not an expensive cost). But I don't know how to set a getter of module in Python.

pwndbg/commands/vmmap.py
cebrusfs: I agree. Actually, I tried when I was writing these code but didn't find a good way to do that in `argparse`. I need someone who familiar with `argparse` to help me to do that.

pwndbg/commands/vmmap.py
cebrusfs: Should we put it in `Page`? I like to put it in `elf.py` since it is a kind of utility of ELF.

pwndbg/commands/vmmap.py
anthraxx: https://docs.python.org/3/library/argparse.html#choices

cebrusfs: I tried. If you use `choice` + `nargs='+'`, you have to type `vmmap_add ... r w x`, You have to split them with a space. But what I want is `vmmap_add ... rwx`.

cebrusfs: I had an **insanity** idea that put all possible permutation and combination of `rwx` to handler it.......But it is annoying when you type `-h` lol

pwndbg/abi.py
cebrusfs: Fixed

pwndbg/abi.py
cebrusfs: Fixed the first one. I rewrited the description for the second one.

pwndbg/abi.py
cebrusfs: Fixed. Thanks.

pwndbg/commands/vmmap.py
cebrusfs: Fixed. Thanks.

pwndbg/commands/vmmap.py
cebrusfs: I rewrote some description. Perhaps it is better now?

pwndbg/commands/vmmap.py
disconnect3d: `elf.py` is fine too

pwndbg/commands/vmmap.py
cebrusfs: Fixed. I perfer a big issue for all TODOs to handle the them.

pwndbg/commands/vmmap.py
anthraxx: then just add a simple function checking that,
untested but basically use type=function:
```python
def valid_vmmap_flag(flags):
    valid_flags = ['r', 'w', 'x']
    for flag in flags:
        if flag not in valid_flags:
             raise argparse.ArgumentTypeError("%s is an invalid positive flag value" % flag)
    return value

parser.add_argument('flags', nargs='?', type=valid_vmmap_flag, default='', help='Flags set by the ELF file, see PF_X, PF_R, PF_W')
```

disconnect3d: Lets leave it as it is

pwndbg/stack.py
cebrusfs: Fixed.

pwndbg/vmmap.py
cebrusfs: Added

pwndbg/vmmap.py
cebrusfs: Fixed. Thanks for suggesting `bisect`. I don't like the original one either.

pwndbg/abi.py
disconnect3d: We can leave it as it is; just update the comment (we dont want '###' ;P)

cebrusfs:
> hope you won't frighten by the conversation/reviews :P.

I understand the reviews and conversation is need :)


> It should be possible to clear it with vmmap.get._reset() (fyi there is also option to reset all caches)

I choose to use `reset()` to flush all caches. Since there may be some cached result that use `vmmap.get()`.


> Subcommands :+1: :heart: - however I am not sure how this would play with our current ArgparsedCommand. If you want to work on improving it, feel free to do so.

I tried to use subparser yesterday, but I do not know how to make it work with `ArgparsedCommand`. Also, subcommands can not set default value. That is, we have to use `vmmap show` or sth to display the `vmmap`. For this case, I prefer add new command to control `custom_pages` and keep `vmmap` as before.


> Good idea, feel free to do so :+1:; this is probably the thing I described as Page.from_elf_sections(filepath) staticmethod.

I prefer to put it at `elf.py` since it is ELF related utility functions.


> PS: If we ever meet on a conference I am going to buy you :beer: (just remind me pls :D).
Great :)

However, I seldom go international conference now since I'm not working in security area. But I went some conference like HITB for CTF before. Hope that I have other chance to meet you in conference :)

pwndbg/commands/vmmap.py
cebrusfs: I think it is a good way to review the `elf.py` and reduce duplicate code. I will not do that in this PR. I will create an issue for this.

disconnect3d (approved):
lgtm

Issue #370, opened 2017-12-15
Disassemble command highlight/color


FFY00:
Hi, is it possible to add highlight/color to the disassemble command? If not, it is possible to create a new command with it enabled?

Thanks,
Filipe

FFY00:
Looks like the only command available is pdisass which only shows a portion of the code.

disconnect3d:
@FFY00 you mean `x/10i <address or expression>` and `disassemble <symbol>`?
We could potentially override `disassemble` to `u`/`pdisass` (those are aliases).

I know also that @anthraxx was working on adding pygments (?) or some other asm syntax highlighting to pwndbg but it didn't get a PR

FFY00:
Yes, exactly.

anthraxx:
Oh right, got catched by RL lately. I still have my branch nearly finished just polishing left. Will finalize during 34C4 and send the PR 😎

disconnect3d:
@anthraxx maybe some pair programming on 34c3 :D

cebrusfs:
I have wrote my own version syntax highlight feature in my fork version. Since I thought it will conflict with current design and the UI changes sometimes are not welcome, so I put it at my own version's fork.

However, I wrote this code 2 years ago, it may violate some current design. Last week, I spent some time to rebase the code to current `dev` branch and have not test too much yet. (It may only work on my environment)

My current design is
1. create a syntax highlight function, given the code text and the language or filename, then return the highlighted code.
2. ASM syntax highlight: `pwndbg/color/disasm.py:instruction()` use that function to syntax highlight *each instructions*. Some issues are open now.
    1. To highlight all the instructions at one time?
    2. How to detect the intel or arm asm?
        1. get the arch from pwndbg.arch?
        1. pass a parameter when call `instruction(arch=)`
3. Source highlight: we can not highlight partial of code. The current `context_code` implementation use `list <line_num>` will not works. My current solution is use
    ```python
    sal = gdb.selected_frame().find_sal()
    filename = sal.symtab.fullname()
    ```
    to get the path of current frame's source code and then open the whole file to syntax highlight the full code and cache it. Then, `context_code` fetch the lines it want.
However, I don't know how gdb get the source code path and whether this API will broken in some cases. You can think it is a kind of workaround since it was just for my personal use.

[...]

PR #390, opened 2018-01-09
Implement asm&source syntax highlight

cebrusfs:
Hi all,

As the discussion in #370, I refactor my work and give you a draft PR for syntax highlighting. The parts mark `XXX` need your ideas.
 
The current design is using pygment to auto-detect the language lexer, except for asm. All asm code use the pwntools Lexer, no matter it is asm source code or disasm output. Since the syntax highlight need whole context of source code. We have to read it and highlight by ourselves. That is, I discard the implementation of original source code displaying (by `list` gdb command), and wrote my own version.

I also add some supporting for some ARM syntax which is only appeared in hand written code. The original one only support the disasm and objdump output.

BTW, The pygments in Python2 doesn't work if enable `from __future__ import unicode_literals` since their implementation use `str`. I wrote a walkaround and only execute in python2 to convert all `unicode` string back to `str`. The better solution is to disable `from __future__ import unicode_literals` in lexer's implementation, but it needs to bypass the `futurize` check.

### DIFF
--- a/pwndbg/color/disasm.py
+++ b/pwndbg/color/disasm.py
@@ -10,6 +10,7 @@ import capstone
 import pwndbg.chain
 import pwndbg.color.context as C
 import pwndbg.color.memory as M
+import pwndbg.color.syntax_highlight as H
 import pwndbg.color.theme as theme
 import pwndbg.config as config
 import pwndbg.disasm.jump
@@ -26,6 +27,11 @@ config_branch = theme.ColoredParameter('disasm-branch-color', 'bold', 'color for
 
 def branch(x):
     return generateColorFunction(config.disasm_branch_color)(x)
+
+
+def syntax_highlight(ins):
+    return H.syntax_highlight(ins, filename='.asm')
+
 
 def instruction(ins):
     asm = '%-06s %s' % (ins.mnemonic, ins.op_str)
--- a/pwndbg/color/disasm.py
+++ b/pwndbg/color/disasm.py
@@ -35,6 +35,8 @@ def syntax_highlight(ins):
 
 def instruction(ins):
     asm = '%-06s %s' % (ins.mnemonic, ins.op_str)
+    if pwndbg.config.syntax_highlight:
+        asm = syntax_highlight(asm)
     is_branch = set(ins.groups) & capstone_branch_groups
 
     # Highlight the current line if enabled
--- a/pwndbg/color/lexer.py
+++ b/pwndbg/color/lexer.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
--- a/pwndbg/color/lexer.py
+++ b/pwndbg/color/lexer.py
@@ -1,0 +1,136 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import re
+
+import six
+from pygments.lexer import RegexLexer
+from pygments.lexer import bygroups
+from pygments.lexer import include
+from pygments.token import Comment
+from pygments.token import Name
+from pygments.token import Number
+from pygments.token import Operator
+from pygments.token import Other
+from pygments.token import Punctuation
+from pygments.token import String
+from pygments.token import Text
+
+__all__ = ['PwntoolsLexer']
+
+# Text        Token.Text            for any type of text data
+# Whitespace  Token.Text.Whitespace for specially highlighted whitespace
+# Error       Token.Error           represents lexer errors
+# Other       Token.Other           special token for data not matched by a parser (e.g. HTML markup in PHP code)
+# Keyword     Token.Keyword         any kind of keywords
+# Name        Token.Name            variable/function names
+# Literal     Token.Literal         Any literals
+# String      Token.Literal.String  string literals
+# Number      Token.Literal.Number  number literals
+# Operator    Token.Operator        operators (+, not...)
+# Punctuation Token.Punctuation     punctuation ([, (...)
+# Comment     Token.Comment         any kind of comments
+# Generic     Token.Generic         generic tokens (have a look at the explanation below)
+class PwntoolsLexer(RegexLexer):
+    """
+    Fork foom pwntools
+    https://github.com/Gallopsled/pwntools/blob/7860eecf025135380b137dd9df85dd02a2fd1667/pwnlib/lexer.py
+    """
+    name = 'PwntoolsLexer'
+    filenames = ['*.s', '*.S', '*.asm']
+
+    #: optional Comment or Whitespace
+    string = r'"(\\"|[^"])*"'
+    char = r'[\w$.@-]'
+    identifier = r'(?:[a-zA-Z$_]' + char + '*|\.' + char + '+|or)'
+    number = r'(?:0[xX][a-zA-Z0-9]+|\d+)'
+    memory = r'(?:[\]\[])'
+    bad = r'(?:\(bad\))'
+
+    tokens = {
+        'root': [
+            include('whitespace'),
+            (identifier + ':', Name.Label),
+            (r'\.' + identifier, Name.Attribute, 'directive-args'),
+            (r'lock|rep(n?z)?|data\d+', Name.Attribute),
+            (identifier, Name.Function, 'instruction-args'),
+            (r'[\r\n]+', Text),
+            (bad, Text)
+        ],
+
+        'directive-args': [
+            (identifier, Name.Constant),
+            (string, String),
+            ('@' + identifier, Name.Attribute),
+            (number, Number.Integer),
+            (r'[\r\n]+', Text, '#pop'),
+
+            (r'#.*?$', Comment, '#pop'),
+
+            include('punctuation'),
+            include('whitespace')
+        ],
+        'instruction-args': [
+            # For objdump-disassembled code, shouldn't occur in
+            # actual assembler input
+            ('([a-z0-9]+)( )(<)('+identifier+')(>)',
+                bygroups(Number.Hex, Text, Punctuation, Name.Constant,
+                         Punctuation)),
+            ('([a-z0-9]+)( )(<)('+identifier+')([-+])('+number+')(>)',
+                bygroups(Number.Hex, Text, Punctuation, Name.Constant,
+                         Punctuation, Number.Integer, Punctuation)),
+
+            # Fun things
+            (r'([\]\[]|BYTE|DWORD|PTR|\+|\-|}|{|\^|>>|<<|&)', Text),
+
+            # Address constants
+            (identifier, Name.Constant),
+            (number, Number.Integer),
+            # Registers
+            ('%' + identifier, Name.Variable),
+            ('$' + identifier, Name.Variable),
+            # Numeric constants
+            ('$'+number, Number.Integer),
+            ('#'+number, Number.Integer),
+            (r"$'(.|\\')'", String.Char),
+            (r'[\r\n]+', Text, '#pop'),
+            include('punctuation'),
+            include('whitespace')
+        ],
+        'whitespace': [
+            (r'\n', Text),
+            (r'\s+', Text),
+            (r'/\*.*?\*/', Comment),
+            (r';.*$', Comment)
+        ],
+        'punctuation': [
+            (r'[-*,.():]+', Punctuation)
+        ]
+    }
+
+    def analyse_text(text):
+        if re.match(r'^\.(text|data|section)', text, re.M):
+            return True
+        elif re.match(r'^\.\w+', text, re.M):
+            return 0.1
+
+if six.PY2:
+    # XXX: convert all unicode() to str() if in Python2.7 since unicode_literals is enabled
+    #   The pygments<=2.2.0 (lastest stable when commit) in Python2.7 use 'str' type in rules matching
+    #   We must convert all unicode back to str()
+    def _to_str(obj):
+        typ = type(obj)
+        if typ is tuple:
+            return tuple(map(_to_str, obj))
+        elif typ is list:
+            return map(_to_str, obj)
+        elif typ is unicode:
+            return str(obj)
+        return obj
+
+    PwntoolsLexer.tokens = {
+        _to_str(k): _to_str(v)
+        for k, v in PwntoolsLexer.tokens.iteritems()
+    }
--- a/pwndbg/color/lexer.py
+++ b/pwndbg/color/lexer.py
@@ -7,7 +7,6 @@ import re
 
 import six
 from pygments.lexer import RegexLexer
-from pygments.lexer import bygroups
 from pygments.lexer import include
 from pygments.token import Comment
 from pygments.token import Name
@@ -18,25 +17,18 @@ from pygments.token import Punctuation
 from pygments.token import String
 from pygments.token import Text
 
+import pwndbg.compat
+
 __all__ = ['PwntoolsLexer']
 
-# Text        Token.Text            for any type of text data
-# Whitespace  Token.Text.Whitespace for specially highlighted whitespace
-# Error       Token.Error           represents lexer errors
-# Other       Token.Other           special token for data not matched by a parser (e.g. HTML markup in PHP code)
-# Keyword     Token.Keyword         any kind of keywords
-# Name        Token.Name            variable/function names
-# Literal     Token.Literal         Any literals
-# String      Token.Literal.String  string literals
-# Number      Token.Literal.Number  number literals
-# Operator    Token.Operator        operators (+, not...)
-# Punctuation Token.Punctuation     punctuation ([, (...)
-# Comment     Token.Comment         any kind of comments
-# Generic     Token.Generic         generic tokens (have a look at the explanation below)
 class PwntoolsLexer(RegexLexer):
     """
-    Fork foom pwntools
+    Fork from pwntools
     https://github.com/Gallopsled/pwntools/blob/7860eecf025135380b137dd9df85dd02a2fd1667/pwnlib/lexer.py
+
+    Edit:
+        * Remove Objdump rules
+        * Merge pygments-arm (https://github.com/heia-fr/pygments-arm)
     """
     name = 'PwntoolsLexer'
     filenames = ['*.s', '*.S', '*.asm']
@@ -47,86 +39,94 @@ class PwntoolsLexer(RegexLexer):
     identifier = r'(?:[a-zA-Z$_]' + char + '*|\.' + char + '+|or)'
     number = r'(?:0[xX][a-zA-Z0-9]+|\d+)'
     memory = r'(?:[\]\[])'
-    bad = r'(?:\(bad\))'
+
+    eol = r'[\r\n]+'
 
     tokens = {
         'root': [
             include('whitespace'),
+
+            # Label
             (identifier + ':', Name.Label),
+            (number + ':', Name.Label),
+
+            # AT&T directive
             (r'\.' + identifier, Name.Attribute, 'directive-args'),
             (r'lock|rep(n?z)?|data\d+', Name.Attribute),
-            (identifier, Name.Function, 'instruction-args'),
-            (r'[\r\n]+', Text),
-            (bad, Text)
-        ],
 
+            # Instructions
+            (identifier, Name.Function, 'instruction-args'),
+
+            (r'[\r\n]+', Text),
+        ],
         'directive-args': [
             (identifier, Name.Constant),
             (string, String),
             ('@' + identifier, Name.Attribute),
             (number, Number.Integer),
-            (r'[\r\n]+', Text, '#pop'),
 
+            (eol, Text, '#pop'),
             (r'#.*?$', Comment, '#pop'),
 
             include('punctuation'),
             include('whitespace')
         ],
         'instruction-args': [
-            # For objdump-disassembled code, shouldn't occur in
-            # actual assembler input
-            ('([a-z0-9]+)( )(<)('+identifier+')(>)',
-                bygroups(Number.Hex, Text, Punctuation, Name.Constant,
-                         Punctuation)),
-            ('([a-z0-9]+)( )(<)('+identifier+')([-+])('+number+')(>)',
-                bygroups(Number.Hex, Text, Punctuation, Name.Constant,
-                         Punctuation, Number.Integer, Punctuation)),
-
             # Fun things
             (r'([\]\[]|BYTE|DWORD|PTR|\+|\-|}|{|\^|>>|<<|&)', Text),
 
             # Address constants
             (identifier, Name.Constant),
+            ('=' + identifier, Name.Constant), # ARM symbol
             (number, Number.Integer),
+
             # Registers
             ('%' + identifier, Name.Variable),
             ('$' + identifier, Name.Variable),
+
             # Numeric constants
-            ('$'+number, Number.Integer),
-            ('#'+number, Number.Integer),
+            ('$' + number, Number.Integer),
+            ('#' + number, Number.Integer),
+
+            # ARM predefined constants
+            ('#' + identifier, Name.Constant),
+
             (r"$'(.|\\')'", String.Char),
-            (r'[\r\n]+', Text, '#pop'),
+
+            (eol, Text, '#pop'),
+
             include('punctuation'),
             include('whitespace')
         ],
         'whitespace': [
             (r'\n', Text),
             (r'\s+', Text),
+
+            # Block comments
+            # /* */ (AT&T)
             (r'/\*.*?\*/', Comment),
-            (r';.*$', Comment)
+
+            # Line comments
+            # //    (AArch64)
+            # #     (AT&T)
+            # ;     (NASM/intel, LLVM)
+            # @     (ARM)
+            (r'(//|[#;@]).*$', Comment.Single)
         ],
         'punctuation': [
             (r'[-*,.():]+', Punctuation)
         ]
     }
 
-    def analyse_text(text):
-        if re.match(r'^\.(text|data|section)', text, re.M):
-            return True
-        elif re.match(r'^\.\w+', text, re.M):
-            return 0.1
-
-if six.PY2:
-    # XXX: convert all unicode() to str() if in Python2.7 since unicode_literals is enabled
-    #   The pygments<=2.2.0 (lastest stable when commit) in Python2.7 use 'str' type in rules matching
-    #   We must convert all unicode back to str()
+# Note: convert all unicode() to str() if in Python2.7 since unicode_literals is enabled
+# The pygments<=2.2.0 (latest stable when commit) in Python2.7 use 'str' type in rules matching
+# We must convert all unicode back to str()
+if pwndbg.compat.python2:
     def _to_str(obj):
-        typ = type(obj)
-        if typ is tuple:
-            return tuple(map(_to_str, obj))
-        elif typ is list:
-            return map(_to_str, obj)
-        elif typ is unicode:
+        type_ = type(obj)
+        if type_ in (tuple, list):
+            return type_(map(_to_str, obj))
+        elif type_ is unicode:
             return str(obj)
         return obj
 
--- a/pwndbg/color/syntax_highlight.py
+++ b/pwndbg/color/syntax_highlight.py
@@ -1,0 +1,67 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import os.path
+import re
+
+import pwndbg.color.message as message
+import pwndbg.color.theme as theme
+import pwndbg.config
+
+try:
+    import pygments
+    import pygments.lexers
+    import pygments.formatters
+    from pwndbg.color.lexer import PwntoolsLexer
+except ImportError:
+    pygments = None
+
+pwndbg.config.Parameter('syntax-highlight', True, 'Source code / assembly syntax highlight')
+style = theme.Parameter('syntax-highlight-style', 'monokai', 'Source code / assembly syntax highlight stylename of pygments module')
+@pwndbg.config.Trigger([style])
+def check_style():
+    try:
+        formatter = pygments.formatters.Terminal256Formatter(
+            style=str(style)
+        )
+    except pygments.util.ClassNotFound:
+        msg = message.warn("The pygment formatter style '%s' is not found, restore to default" % style)
+        print(msg)
+        style.value = style.default
+
+
+def syntax_highlight(code, filename):
+    # No syntax highlight if pygment is not installed
+    if not pygments:
+        return code
+
+    filename = os.path.basename(filename)
+
+    formatter = pygments.formatters.Terminal256Formatter(
+        style=str(style)
+    )
+
+    lexer = None
+
+    # If source code is asm, use our customized lexer.
+    # Note: We can not register our Lexer to pygments and use their APIs,
+    # since the pygment only search the lexers installed via setuptools.
+    for glob_pat in PwntoolsLexer.filenames:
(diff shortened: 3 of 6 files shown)


pwndbg/color/lexer.py
```diff
@@ -0,0 +1,156 @@
...
+# Literal     Token.Literal         Any literals
+# String      Token.Literal.String  string literals
+# Number      Token.Literal.Number  number literals
+# Operator    Token.Operator        operators (+, not...)
+# Punctuation Token.Punctuation     punctuation ([, (...)
+# Comment     Token.Comment         any kind of comments
+# Generic     Token.Generic         generic tokens (have a look at the explanation below)
+class PwntoolsLexer(RegexLexer):
+    """
+    Fork foom pwntools
```
disconnect3d: Typo: `foom` -> `from`

pwndbg/color/lexer.py
```diff
@@ -0,0 +1,156 @@
...
+from pygments.token import Number
+from pygments.token import Operator
+from pygments.token import Other
+from pygments.token import Punctuation
+from pygments.token import String
+from pygments.token import Text
+
+__all__ = ['PwntoolsLexer']
+
+# Text        Token.Text            for any type of text data
```
disconnect3d: What exactly are those columns? Can we add a header row above?

pwndbg/color/lexer.py
```diff
@@ -0,0 +1,156 @@
...
+        ]
+    }
+
+    def analyse_text(text):
+        if re.match(r'^\.(text|data|section)', text, re.M):
+            return True
+        elif re.match(r'^\.\w+', text, re.M):
+            return 0.1
+
+if six.PY2:
```
disconnect3d: We have `pwndbg.compat.python2` - maybe we should remove it in favor of `six.PY2` instead?
@zachriggle what do you think?

pwndbg/color/lexer.py
```diff
@@ -0,0 +1,156 @@
...
+    # XXX: convert all unicode() to str() if in Python2.7 since unicode_literals is enabled
+    #   The pygments<=2.2.0 (lastest stable when commit) in Python2.7 use 'str' type in rules matching
```
disconnect3d: Typo `lastest` -> `latest`

pwndbg/color/lexer.py
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    def analyse_text(text):
+        if re.match(r'^\.(text|data|section)', text, re.M):
+            return True
+        elif re.match(r'^\.\w+', text, re.M):
+            return 0.1
+
+if six.PY2:
+    # XXX: convert all unicode() to str() if in Python2.7 since unicode_literals is enabled
```
disconnect3d: ~https://github.com/timothycrosley/isort#skip-processing-of-imports-outside-of-configuration
We could eventually use `isort:skip_file`, :thinking:; @zachriggle what do you think?~

EDIT: Oh, this is connected to `futurize`, not to `isort`.

pwndbg/color/lexer.py
```diff
@@ -0,0 +1,156 @@
...
+    #   The pygments<=2.2.0 (lastest stable when commit) in Python2.7 use 'str' type in rules matching
+    #   We must convert all unicode back to str()
+    def _to_str(obj):
+        typ = type(obj)
+        if typ is tuple:
+            return tuple(map(_to_str, obj))
+        elif typ is list:
+            return map(_to_str, obj)
```
disconnect3d: Maybe:
```python
if type_ in (tuple, list):
    return type_(map(_to_str, obj))
elif type_ is unicode:
    return str(obj)

return obj
```

pwndbg/color/lexer.py
disconnect3d: If we are going to stay with current approach, change `XXX` to `NOTE:`

pwndbg/color/lexer.py
disconnect3d: And just a single space for comments (even in continuation) pls

pwndbg/color/disasm.py
```diff
@@ -28,8 +29,13 @@
 def branch(x):
     return generateColorFunction(config.disasm_branch_color)(x)
 
+def syntax_highlight(ins):
```
disconnect3d: double newline between functions pls

pwndbg/commands/context.py
```diff
@@ -162,11 +163,26 @@ def context_disasm():
 
     return banner + result
 
+# XXX: move this part to color/ just like color/disasm.py:instrution()?
```
disconnect3d: Hm.. I've some mixed feelings, on one side it is connected to disasm but on the other hand this is context;
Anyway it is okay to move it there.

pwndbg/commands/context.py
```diff
@@ -180,23 +196,52 @@ def context_code():
...
-                    source_lines[i] = C.highlight(source_lines[i])
-                    break
-
-        banner = [pwndbg.ui.banner("source")]
-        banner.extend(source_lines)
+        n = int(source_code_length)
+
+        # Compute the line range
+        st = max(closest_line - 1 - n//2, 0)
+        ed = min(closest_line - 1 + n//2 + 1, len(source))
```
disconnect3d: Guess those are `start` and `end` - can we change to those names? `st` and `ed` are not so explicit/known

pwndbg/commands/context.py
```diff
@@ -180,23 +196,52 @@ def context_code():
...
+            # Compute the prefix
+            # XXX: use nearpc-prefix here, provide customize?
+
+            # XXX: copy from color/disasm
+            prefix_sign = pwndbg.config.nearpc_prefix
+            value = prefix_sign.value
+            if isinstance(value, bytes):
+                value = codecs.decode(value, 'utf-8')
+
+            format = ' {prefix_sign:{prefix_width}} {line_number:>{num_width}} {code}'
```
disconnect3d: `format` is a builtin and so github highlights it; maybe `fmt`?

pwndbg/commands/context.py
```diff
@@ -180,23 +196,52 @@ def context_code():
...
+            if line_number == closest_line and pwndbg.config.highlight_source:
```
disconnect3d: Can we invert those to: `if pwndbg.config.highlight_source and line_number == closest_line:`

pwndbg/commands/context.py
```diff
@@ -162,11 +163,26 @@ def context_disasm():
...
 
+# XXX: move this part to color/ just like color/disasm.py:instrution()?
 theme.Parameter('highlight-source', True, 'whether to highlight the closest source line')
-
+source_code_length = pwndbg.config.Parameter('context-source-code-length',
+                                             10,
+                                             'number of lines of source code to printed by the context command')
+@pwndbg.memoize.reset_on_exit
+def get_highlight_source(filename):
+    # Notice that the code will cached
```
disconnect3d: `will cached` -> `is cached`

pwndbg/color/__init__.py
```diff
@@ -73,3 +75,49 @@ def terminateWith(x, color):
...
+try:
+    import pygments
+    import pygments.lexers
+    import pygments.formatters
+    from pwndbg.color.lexer import PwntoolsLexer
+except ImportError:
+    pygments = None
+
+pwndbg.config.Parameter('syntax-highlight', True, 'Source code / assembly syntax highlight')
+pwndbg.config.Parameter('syntax-highlight-style', 'monokai', 'Source code / assembly syntax highlight stylename of pygments module')
```
disconnect3d: We should probably create a `@pwndbg.config.Trigger` for this and check if user sets valid value - if not - set it to default and warn the user?

disconnect3d: As currently I can do `set syntax-highlight-style 'whatever'` and `context` will fail

anthraxx: the style should actually be part of the 'theme' config instead of the global config as style for highlighting is exactly what theme is for :smile_cat:

disconnect3d: @anthraxx :+1: :heart:

disconnect3d (changes requested):
Current `context_code` doesn't work for a simple hello world C code on Python 3.

Here is my stacktrace:
```
  File "/home/dc/installed/pwndbg/pwndbg/commands/__init__.py", line 107, in __call__
  File "/home/dc/installed/pwndbg/pwndbg/commands/__init__.py", line 198, in _OnlyWhenRunning
    return function(*a, **kw)
  File "/home/dc/installed/pwndbg/pwndbg/commands/context.py", line 65, in context
    result.extend(func())
  File "/home/dc/installed/pwndbg/pwndbg/commands/context.py", line 201, in context_code
    source = get_highlight_source(filename)
  File "/home/dc/installed/pwndbg/pwndbg/memoize.py", line 48, in __call__
  File "/home/dc/installed/pwndbg/pwndbg/commands/context.py", line 177, in get_highlight_source
    source = pwndbg.color.syntax_highlight(source, filename)
  File "/home/dc/installed/pwndbg/pwndbg/color/__init__.py", line 116, in syntax_highlight
    lexer = pygments.lexers.guess_lexer_for_filename(filename, code)
  File "/usr/lib/python3.6/site-packages/pygments/lexers/__init__.py", line 256, in guess_lexer_for_filename
    for lexer in _iter_lexerclasses():
  File "/usr/lib/python3.6/site-packages/pygments/lexers/__init__.py", line 230, in _iter_lexerclasses
    _load_lexers(module_name)
  File "/usr/lib/python3.6/site-packages/pygments/lexers/__init__.py", line 41, in _load_lexers
    mod = __import__(module_name, None, None, ['__all__'])
  File "/usr/lib/python3.6/site-packages/pygments/lexers/business.py", line 24, in <module>
    class CobolLexer(RegexLexer):
  File "/usr/lib/python3.6/site-packages/pygments/lexers/business.py", line 34, in CobolLexer
    flags = re.IGNORECASE | re.MULTILINE
  File "/usr/lib/python3.6/enum.py", line 798, in __or__
    return cls._missing_(value)
  File "/usr/lib/python3.6/enum.py", line 762, in _missing_
  File "/usr/lib/python3.6/enum.py", line 788, in _create_pseudo_member_
    pseudo_member._name_ = None
```

Similar problem already occured once and was connected to our inthook... this might help: https://github.com/pwndbg/pwndbg/pull/222 (it will probably require to set some prints there to debug this)

pwndbg/color/lexer.py
disconnect3d: I've looked over `futurize` and it seems they don't have a feature to skip some files, so bad...
Guess we can stay with this workaround for now.
Please add info that this is strictly connected to `futurize`.

pwndbg/color/lexer.py
```diff
@@ -0,0 +1,156 @@
...
+# Keyword     Token.Keyword         any kind of keywords
+# Name        Token.Name            variable/function names
+# Literal     Token.Literal         Any literals
+# String      Token.Literal.String  string literals
+# Number      Token.Literal.Number  number literals
+# Operator    Token.Operator        operators (+, not...)
+# Punctuation Token.Punctuation     punctuation ([, (...)
+# Comment     Token.Comment         any kind of comments
+# Generic     Token.Generic         generic tokens (have a look at the explanation below)
+class PwntoolsLexer(RegexLexer):
```
disconnect3d: Btw have you checked how it works with other archs - e.g. MIPS?

pwndbg/color/lexer.py
```diff
@@ -0,0 +1,156 @@
...
+        'punctuation': [
+            (r'[-*,.():]+', Punctuation)
+        ]
+    }
+
+    def analyse_text(text):
+        if re.match(r'^\.(text|data|section)', text, re.M):
+            return True
+        elif re.match(r'^\.\w+', text, re.M):
+            return 0.1
```
disconnect3d: I wonder what's that used for (weird return value value in comparison to the above one).
(I know the code is copied)

pwndbg/commands/context.py
```diff
@@ -180,23 +196,52 @@ def context_code():
...
+        formated_source = []
+        for line_number, code in enumerate(source, start=st + 1):
+            # Compute the prefix
+            # XXX: use nearpc-prefix here, provide customize?
+
+            # XXX: copy from color/disasm
+            prefix_sign = pwndbg.config.nearpc_prefix
+            value = prefix_sign.value
+            if isinstance(value, bytes):
+                value = codecs.decode(value, 'utf-8')
```
disconnect3d: Is it necessary?

pwndbg/commands/context.py
```diff
@@ -180,23 +196,52 @@ def context_code():
...
+        # Compute the line range
+        st = max(closest_line - 1 - n//2, 0)
+        ed = min(closest_line - 1 + n//2 + 1, len(source))
+        num_width = len(str(ed))
+
+        # split the code
+        source = source[st:ed]
+
+        # Format the output
+        formated_source = []
```
disconnect3d: `formated_source` -> `formatted_source`

pwndbg/commands/context.py
```diff
@@ -180,23 +196,52 @@ def context_code():
...
+            line = format.format(
+                prefix_sign = prefix_sign if line_number == closest_line else '',
+                prefix_width = len(value),
+                line_number = line_number,
+                num_width = num_width,
+                code = code
+            )
+            formated_source.append(line)
+
+        banner = [pwndbg.color.blue(pwndbg.ui.banner("Source (code)"))]
```
disconnect3d: Maybe `banner = [...] + formatted_source`

pwndbg/color/__init__.py
```diff
@@ -73,3 +75,49 @@ def terminateWith(x, color):
...
+
+    syntax_highlight_style = pwndbg.config.syntax_highlight_style
+    formatter = pygments.formatters.Terminal256Formatter(
+        style=str(syntax_highlight_style)
+    )
+
+    lexer = None
+
+    # If source code is asm, use our customized lexer.
+    #   We can not register our Lexer to pygments and use their APIs,
```
disconnect3d: single space after `#` pls

pwndbg/commands/context.py
anthraxx: actually we have a ```theme banner-color``` which should be used....
 
```ui.banner()``` already does the highlighting, so just purge explicit ```pwndbg.color.blue```

File: pwndbg/commands/context.py [hunk 20]
```diff
@@ -162,11 +163,26 @@ def context_disasm():
 
     return banner + result
 
+# XXX: move this part to color/ just like color/disasm.py:instrution()?
 theme.Parameter('highlight-source', True, 'whether to highlight the closest source line')
-
+source_code_length = pwndbg.config.Parameter('context-source-code-length',
+                                             10,
+                                             'number of lines of source code to printed by the context command')
+@pwndbg.memoize.reset_on_exit
```
disconnect3d: heh I wonder what's the difference between `reset_on_exit` and `reset_on_start`.

We probably should cache it until next run of the same binary (as someone might have changed the code and recompile it). Not sure how that works now

pwndbg/commands/context.py
```diff
@@ -162,11 +163,26 @@ def context_disasm():
 
     return banner + result
 
+# XXX: move this part to color/ just like color/disasm.py:instrution()?
 theme.Parameter('highlight-source', True, 'whether to highlight the closest source line')
-
+source_code_length = pwndbg.config.Parameter('context-source-code-length',
+                                             10,
+                                             'number of lines of source code to printed by the context command')
```
disconnect3d: Maybe: `number of source code lines to print by the context command`.

pwndbg/commands/context.py
```diff
@@ -162,11 +163,26 @@ def context_disasm():
 
     return banner + result
 
+# XXX: move this part to color/ just like color/disasm.py:instrution()?
 theme.Parameter('highlight-source', True, 'whether to highlight the closest source line')
-
+source_code_length = pwndbg.config.Parameter('context-source-code-length',
```
disconnect3d: Better/more consistent naming: `context-source-code-lines`

See:
```
[dc@dc:pwndbg|(26e81f8...) *$%]$ git grep _lines | grep Param
pwndbg/commands/context.py:code_lines = pwndbg.config.Parameter('context-code-lines', 10, 'number of additional lines to print in the code context')
pwndbg/commands/context.py:stack_lines = pwndbg.config.Parameter('context-stack-lines', 8, 'number of lines to print in the stack context')
pwndbg/commands/nearpc.py:nearpc_lines = pwndbg.config.Parameter('nearpc-lines', 10, 'number of additional lines to print for the nearpc command')
pwndbg/commands/telescope.py:telescope_lines = pwndbg.config.Parameter('telescope-lines', 8, 'number of lines to printed by the telescope command')
[dc@dc:pwndbg|(26e81f8...) *$%]$ git grep _length | grep Param
pwndbg/commands/context.py:source_code_length = pwndbg.config.Parameter('context-source-code-length',
```

pwndbg/color/lexer.py
disconnect3d: Can we also add a link to the file in pwntools repo?

pwndbg/color/lexer.py
cebrusfs: I don't see the feature to bypass for `futurize` either. That's the reason I wrote the workaround here.

pwndbg/commands/context.py
cebrusfs: For me, it is connected to context. Actually, I would like to create a new file such as `color/syntax_highlight.py` and move all related things to there.

pwndbg/color/lexer.py
cebrusfs: No, just check with ARM and a few x86-64 code with Intel syntax (NASM).

The pwnlib's lexer is modified from AT&T parser of `pygments` and add a few rule to make some environment works. It miss a lots feature for each arch. However, if we want to support multiple arch, I need someone who know each kind of asm syntax and make the new feature will not break the old one. Moreover, some syntaxes may not compatible.

I only fix some ARM feature I used.

To develop the lexer to support is for each arch, we can maintain a different lexer for the baseline. Then, we create a test for `PwntoolsLexer` force it to produce the same for different code. However, the difficulty of maintaining `PwntoolsLexer` is still the same.

pwndbg/color/lexer.py
cebrusfs: Pygments provide a guess API to choose the lexer when given a text.
It is not used here and neither in pwntools.

Actually, it is copied from pygments `GasLexer`
https://bitbucket.org/birkenfeld/pygments-main/src/7941677dc77d4f2bf0bbd6140ade85a9454b8b80/pygments/lexers/asm.py?at=default&fileviewer=file-view-default#asm.py-26

pwndbg/commands/context.py
cebrusfs: Not sure.
In `disasm.py`, there is the same code to compute the Unicode prefix length. I think the reason is that when we use the command line to set the prefix, it will be utf8, not the unicode.
But I didn't verify this.

A better way is that we modify the setter wrapper in `pwndbg.config.Parameter` to make sure everything in config is unicode.

pwndbg/commands/context.py
cebrusfs: Sorry, I copy these code from my last year's implementation and forget to fix it. Good catch and thanks.

pwndbg/color/lexer.py
cebrusfs: It is noted in [pygments doc](http://pygments.org/docs/tokens/) and it does not have header row. I just note them and for me to refer easily when editing. I think I can remove this notes.

pwndbg/color/lexer.py
cebrusfs: I don't know why `pwndbg.compat` module exists. It seems it is do similar with what `six` did.
For me, I prefer to use stable library instead re-implementing one.
Does it means that you have plan to remove `six` module from dependency?

pwndbg/color/lexer.py
cebrusfs: What do you means a *link*? I put the url at the next line. Or you mean something like git submodule? 
But I have some changes and remove some useless part.

cebrusfs: Fixed the typo

pwndbg/color/lexer.py
cebrusfs: Removed

pwndbg/color/lexer.py
cebrusfs: BTW, changed to `pwndbg.compat.python2`

pwndbg/color/lexer.py
cebrusfs: Fixed

pwndbg/color/lexer.py
cebrusfs: Fixed

pwndbg/color/disasm.py
cebrusfs: Fixed

pwndbg/commands/context.py
cebrusfs: Fixed

pwndbg/commands/context.py
cebrusfs: Fixed

pwndbg/color/__init__.py
cebrusfs: Added the trigger to restore value when the value is invalid. Also, use the `theme` config for the style setting.

pwndbg/commands/context.py
cebrusfs: Fixed

pwndbg/commands/context.py
cebrusfs: Fixed.

pwndbg/color/__init__.py
cebrusfs: Fixed

pwndbg/commands/context.py
cebrusfs: Fixed

pwndbg/commands/context.py
cebrusfs: Fixed

pwndbg/commands/context.py
cebrusfs: Fixed

pwndbg/commands/context.py
cebrusfs: Fixed

pwndbg/commands/context.py
cebrusfs: I read the code but still don't know when will trigger `start` and `exit` event. I think they are similar in this case. I suggest that someone know the memorize and event can write some doc/comments for each event about when gdb will file it and use case.

BTW, I change to `reset_on_start`. It looks more consistent for me.

cebrusfs:
@disconnect3d 
I don't have gdb with python3 environment. But I don't think it is my problem because I did not do anything about enum and rehook int type.

Can you launch the gdb without `pwndbg` and
```
py import pygments.lexers; pygments.lexers.guess_lexer_for_filename('a.c', 'main () {}')
```
to verify whether it is cause by `pwndbg`'s inthook? 

Or try your PoC in #222?
```
py import re; re.DOTALL | re.MULTILINE
```

I tried to port `inthook.py` and import from python3 shell, but it works fine with `re` and `pygments`. I can not reproduce this.

Also, I fixed the most problems, can you review again? Thanks.

pwndbg/commands/context.py
disconnect3d: I like the idea of `syntax_highlight.py` :+1:

disconnect3d:
@cebrusfs `py import re; re.DOTALL | re.MULTILINE` works for me while `py import pygments.lexers; pygments.lexers.guess_lexer_for_filename('a.c', 'main () {}')` doesn't. It fails only in pwndbg so its certainly connected to inthook.

I will try to resolve this issue myself and either post here results or make a PR with the fix.

pwndbg/color/lexer.py
disconnect3d: > Does it means that you have plan to remove six module from dependency?
No, not at all - we want to support both py2 and py3.

We will probably remove `pwndbg.compat` in favor of `six.PY2`/`six.PY3` soon

pwndbg/commands/context.py
cebrusfs: Then we need to decide which function need to move to there. You can give me some comments when you review the second time.

disconnect3d:
@cebrusfs can u rebase it on top of master? There is a conflict as we merged https://github.com/pwndbg/pwndbg/pull/392

pwndbg/commands/context.py
```diff
@@ -180,23 +196,52 @@ def context_code():
...
+            value = prefix_sign.value
+            if isinstance(value, bytes):
+                value = codecs.decode(value, 'utf-8')
+
+            fmt = ' {prefix_sign:{prefix_width}} {line_number:>{num_width}} {code}'
+            if pwndbg.config.highlight_source and line_number == closest_line:
+                fmt = C.highlight(fmt)
+
+            line = fmt.format(
+                prefix_sign = prefix_sign if line_number == closest_line else '',
```
disconnect3d: This line must be:
```python
prefix_sign = str(prefix_sign) if line_number == closest_line else '',
```

Otherwise I am getting:
```
  File "pwndbg/pwndbg/commands/__init__.py", line 107, in __call__
  File "pwndbg/pwndbg/commands/__init__.py", line 198, in _OnlyWhenRunning
    return function(*a, **kw)
  File "pwndbg/pwndbg/commands/context.py", line 65, in context
    result.extend(func())
  File "pwndbg/pwndbg/commands/context.py", line 242, in context_code
    code = code
TypeError: unsupported format string passed to Parameter.__format__
```

disconnect3d: Also can we remove spaces before and after `=`?
Its not really variable assignment and just a keyword arguments passing and this is usually done without spaces huh

disconnect3d:
@cebrusfs The Python 3 failing case will be fixed by https://github.com/pwndbg/pwndbg/pull/393

anthraxx:
nice PR, don't get discouraged :smile: 
The new theme patches added color messages from pwndbg.color.messages, will comment some things in the PR

pwndbg/color/__init__.py
```diff
@@ -73,3 +77,59 @@ def terminateWith(x, color):
...
+style = theme.Parameter('syntax-highlight-style', 'monokai', 'Source code / assembly syntax highlight stylename of pygments module')
+
+@pwndbg.config.Trigger([style])
+def check_style():
+    try:
+        formatter = pygments.formatters.Terminal256Formatter(
+            style = str(style)
+        )
+    except pygments.util.ClassNotFound:
+        print("The pygment formatter style '%s' is not found, restore to default" % style)
```
anthraxx: This should use the new ```pwndbg.color.message``` module's ```warn()``` method

cebrusfs: Fixed

pwndbg/commands/context.py
cebrusfs: Both Fixed.
For Parameter issue, maybe we can implement `__format__` method to support this case and then we can remove all operation of casting to `str` or `int`.

cebrusfs:
Hi all,

I rebased and merge all edit back to origin commits.
I also move the the `syntax_highlight()` to `color/syntax_highlight.py` so that no changes are on `color/__init__.py` now.

disconnect3d:
That `__format__` method is a good idea :+1:

pwndbg/commands/context.py
```diff
@@ -181,23 +199,57 @@ def context_code():
...
+        start = max(closest_line - 1 - n//2, 0)
+        end = min(closest_line - 1 + n//2 + 1, len(source))
+        num_width = len(str(end))
+
+        # split the code
+        source = source[start:end]
+
+        # Compute the prefix_sign length
+        # XXX: use nearpc-prefix here, provide customize?
+        # XXX: copy from color/disasm
```
disconnect3d: What about those comments? We should probably delete them or change sth

pwndbg/commands/context.py
```diff
@@ -181,23 +199,57 @@ def context_code():
...
+        # This is needed because the config value may be utf8 byte string.
+        # It is better to convert to unicode at setter of config and then we will not need this.
+        prefix_sign = pwndbg.config.nearpc_prefix
+        value = prefix_sign.value
+        if isinstance(value, bytes):
+            value = codecs.decode(value, 'utf-8')
+        prefix_width = len(value)
+
+        # Covert config class to str to make format() work
```
disconnect3d: `Covert` -> `Convert`

pwndbg/commands/context.py
cebrusfs: 1 No idea. It depends on you, do you want people be able to customize the prefix sign for source code and disasm separately.
2. That means I think this part is a little dirty for calculating the string length. I mentioned that perhaps we can modified the setter of `Parameter` to force everything is unicode. But this should be in the other PR. For now, perhaps we can note this and put a `TODO` here?

pwndbg/color/syntax_highlight.py
```diff
@@ -0,0 +1,67 @@
+from __future__ import absolute_import
```
anthraxx, line 2: new files should explicitly declare python encoding, we do this in every other file
```# -*- coding: utf-8 -*-```
https://www.python.org/dev/peps/pep-0263/

File: pwndbg/color/syntax_highlight.py [hunk 28]
```diff
@@ -0,0 +1,67 @@
...
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import os.path
+import re
+
+import pwndbg.color.message as message
+import pwndbg.color.theme as theme
```
anthraxx: these should be replaced with
```python
from pwndbg.color import theme
```

pwndbg/commands/context.py
disconnect3d: 1. I think we can keep it as it is. If someone really wants it (looks at @anthraxx as he customizes a lot ;D), he/she can add it.
2. Yeah, add an issue as well for that pls! :)

disconnect3d (changes requested):
Waiting for another rebase & the last changes (mentioned in comments).

pwndbg/commands/context.py
anthraxx: @cebrusfs yes please do, we have the same for nearpc and backtrace already.
```python
theme.Parameter('nearpc-prefix', '►', 'prefix marker for nearpc command')
theme.ColoredParameter('nearpc-prefix-color', 'none', 'color for nearpc command (prefix marker)')
```
would be neat to have that for code as well, ```code-prefix``` and ```code-prefix-color``` :heart:

cebrusfs: Added.

pwndbg/commands/context.py
cebrusfs: Fixed. Thanks.

pwndbg/color/syntax_highlight.py
cebrusfs: Added

pwndbg/color/syntax_highlight.py
cebrusfs: Fixed.

disconnect3d (approved):
lgtm

disconnect3d:
@cebrusfs Just for your info: current version of this branch is not on top of dev (it is not rebased on dev) - so when I use it it doesn't work on Py3 because it lacks the fixes that are present on dev.
This isn't really a problem here as I can just pull your branch and rebase it locally to check if it works - but for the future - the proper way to rebase is for example by using `git fetch origin && git rebase -i origin/dev` (assuming origin is `pwndbg/pwndbg`; this is an interactive rebase, lets you do a lot of fancy stuff like reordering commits and so on).


Thanks again for this PR! Merging :heart: :+1:

cebrusfs:
Thank for reminding, I know this.
I often use my local `dev` branch to do the rebase and I just forgot to pull it when I am rebasing...
Sorry for the inconvenience on the testing.

Issue #245, opened 2017-04-23
Abolish print() in favor of stdlib logging


zachriggle:
Currently, we use `print()` in lots of places.  This makes it difficult to filter things by verbosity, and results in us disabling various print statements or hiding them behind ad-hoc `if debug` statements.  This will also allow us to get rid of some quantity of `print(M.red("Error message"))` statements in favor of e.g. `log.error`.

Using real logging is more flexible, and also allows us to be more verbose in our debug statements, which in turn means we can get better bug reports from users.

I suggest something similar to what `pwntools` currently does, by implementing an e.g. `pwndbg.log` module, with a `log = getLogger(__name__)` method call in each module, so that each module has its own `log` object.

disconnect3d:
PR containing those changes should also add information to use stdlib logging instead of prints in [.github/CONTRIBUTING.md](.github/CONTRIBUTING.md)

disconnect3d:
We do have ` pwndbg.log` as of now which we use here and there and we have `message.{error,info,warn}` for color purposes.

disconnect3d:
Closing this for now. May re-introduce this issue or even just fix it in the future if we start to feel that we should completely remove `print()`.

PR #392, opened 2018-01-10
make everything themeable

anthraxx:
This makes it posssible to theme everything logically grouped by
message types. This will also make it easier for future features
to keep a consistent way of coloring plus make every non-specific
coloring themeable automatically.

Direct explicit usage of colors should be avoided in future commits.

### DIFF
--- a/pwndbg/commands/got.py
+++ b/pwndbg/commands/got.py
@@ -14,9 +14,7 @@ import pwndbg.file
 import pwndbg.which
 import pwndbg.wrappers.checksec
 import pwndbg.wrappers.readelf
-from pwndbg.color import green
-from pwndbg.color import light_yellow
-from pwndbg.color import red
+from pwndbg.color import message
 
 parser = argparse.ArgumentParser(description='Show the state of the Global Offset Table')
 parser.add_argument('name_filter', help='Filter results by passed name.',
@@ -30,14 +28,19 @@ def got(name_filter=''):
     relro_status = pwndbg.wrappers.checksec.relro_status()
     pie_status = pwndbg.wrappers.checksec.pie_status()
     jmpslots = list(pwndbg.wrappers.readelf.get_jmpslots())
-
     if not len(jmpslots):
-        print(red("NO JUMP_SLOT entries available in the GOT"))
+        print(message.error("NO JUMP_SLOT entries available in the GOT"))
         return
+
     if "PIE enabled" in pie_status:
         bin_text_base = pwndbg.memory.page_align(pwndbg.elf.entry())
 
-    print("\nGOT protection: %s | GOT functions: %d\n " % (green(relro_status), len(jmpslots)))
+    relro_color = message.off
+    if 'Partial' in relro_status:
+        relro_color = message.warn
+    elif 'Full' in relro_status:
+        relro_color = message.on
+    print("\nGOT protection: %s | GOT functions: %d\n " % (relro_color(relro_status), len(jmpslots)))
 
     for line in jmpslots:
         address, info, rtype, value, name = line.split()[:5]
@@ -51,4 +54,4 @@ def got(name_filter=''):
             address_val = bin_text_base + address_val
 
         got_address = pwndbg.memory.pvoid(address_val)
-        print("[0x%x] %s -> %s" % (address_val, light_yellow(name), pwndbg.chain.format(got_address)))
+        print("[0x%x] %s -> %s" % (address_val, message.hint(name), pwndbg.chain.format(got_address)))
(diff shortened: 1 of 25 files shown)


pwndbg/commands/got.py
```diff
@@ -30,14 +28,19 @@ def got(name_filter=''):
...
-        print(red("NO JUMP_SLOT entries available in the GOT"))
+        print(message.error("NO JUMP_SLOT entries available in the GOT"))
         return
+
     if "PIE enabled" in pie_status:
         bin_text_base = pwndbg.memory.page_align(pwndbg.elf.entry())
 
-    print("\nGOT protection: %s | GOT functions: %d\n " % (green(relro_status), len(jmpslots)))
+    relro_color = message.off
+    if 'Partial' in relro_status:
```
disconnect3d, line 39: Hope there isn't any edge case ;f

pwndbg/commands/heap.py
```diff
@@ -124,10 +122,10 @@ def arenas():
         
         h = heap.get_region(addr)
         if not h:
-            print(red('Could not find the heap'))
+            print(message.error('Could not find the heap'))
             return
 
-        hdr = bold(fmt%(hex(addr) if addr else 'main'))
+        hdr = message.hint(fmt%(hex(addr) if addr else 'main'))
```
disconnect3d: space pls `fmt % `

disconnect3d:
We should use:
```python
```

instead of:
```python
import pwndbg.color.message as message
```

anthraxx:
i agree, i was just mimic-ing what we did so far :stuck_out_tongue_closed_eyes:

pwndbg/commands/got.py
anthraxx: well thats how we stricty use  our wrapper/checksec.py so it must be the same output :yum:

pwndbg/commands/heap.py
anthraxx: fixed

disconnect3d:
This is awesome :heart:.

One more thing that comes to my mind: maybe we should change:
```python
print(message.notice(...))
```
Into:
```python
log.notice(...)
```

This could be then somehow connected to https://github.com/pwndbg/pwndbg/issues/245

anthraxx:
yes we should, but lets do that in a separate PR attached to logging an not themeing. I volunteer to do it :smile:

disconnect3d:
One more thing - fire `isort --recursive pwndbg` ;D

anthraxx:
all implemented and isorted :smile_cat:

disconnect3d (approved):
lgtm :100:

disconnect3d:
@anthraxx will merge just after check passes

disconnect3d:
@anthraxx btw I haven't checked that and now I see that sth is not working fine; can u look for it?

```
[dc@dc:pwndbg|dev $%]$ gdb /bin/ls
pwndbg: loaded 158 commands. Type pwndbg [filter] for a list.
Set whether to print a full stacktracefor exceptions raised in Pwndbg commands to True
pwndbg> entry
Temporary breakpoint 1 at 0x555555559060

Temporary breakpoint 1, 0x0000555555559060 in ?? ()
  File "pwndbg/pwndbg/events.py", line 119, in caller
    func()
  File "pwndbg/pwndbg/heap/__init__.py", line 18, in update
    import pwndbg.heap.ptmalloc
  File "pwndbg/pwndbg/heap/ptmalloc.py", line 14, in <module>
    from pwndbg.color.message import message
ImportError: cannot import name 'message'

PS: Pull requests are welcome
  File "pwndbg/pwndbg/prompt.py", line 33, in prompt_hook
    pwndbg.events.after_reload(start=False)
  File "pwndbg/pwndbg/events.py", line 163, in after_reload
    f()
  File "pwndbg/pwndbg/events.py", line 123, in caller
    raise e
  File "pwndbg/pwndbg/events.py", line 119, in caller
    func()
  File "pwndbg/pwndbg/heap/__init__.py", line 18, in update
    import pwndbg.heap.ptmalloc
  File "pwndbg/pwndbg/heap/ptmalloc.py", line 14, in <module>
    from pwndbg.color.message import message
ImportError: cannot import name 'message'
```

zachriggle:
Can we add caching to these? I'm not sure what's the performance hit of going to GDB settings and pulling out config values for *every* color, *every* time we print it.

anthraxx:
I can do some micro benchmark but it performs quite fast when stepping through nothing noticeable and I doubt it's from any interest compared to. All the things we do. Like dereference and stuff

zachriggle:
Sure, I guess the parameters might not actually go down into the GDB native code and back out to Python.

PR #393, opened 2018-01-10
Fix inthook for enums in Python 3

disconnect3d:
Fixes the problem that can be observed below:

```
pwndbg> py import re; flags = 1 | re.MULTILINE
  File "/usr/lib/python3.6/enum.py", line 798, in __or__
    return cls._missing_(value)
  File "/usr/lib/python3.6/enum.py", line 762, in _missing_
  File "/usr/lib/python3.6/enum.py", line 788, in _create_pseudo_member_
    pseudo_member._name_ = None
```

### DIFF
--- a/pwndbg/inthook.py
+++ b/pwndbg/inthook.py
@@ -10,6 +10,7 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import enum
 import os
 
 import gdb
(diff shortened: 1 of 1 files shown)


disconnect3d:
@anthraxx @cebrusfs @zachriggle anyone has pwndbg with Python 2 close by to check if that works fine with Py2?

anthraxx:
@disconnect3d sorry, not really. would need to find some weird distro with py2 and create a fresh vm. prefere not to do it :stuck_out_tongue_closed_eyes:

cebrusfs:
LGTM.

Everything work fine. I did not test too much things. Just attach a process and also tested `py import re; flags = 1 | re.MULTILINE`.

FYI, It seems the debian's `arm-none-eabi-gdb` use python 2.7. But I am in some wired environment, the package may be modified by the others, so I am not pretty sure.

disconnect3d:
@cebrusfs thx

PR #395, opened 2018-01-11
improve repeat functionality of commands

anthraxx:
hexdump: adjust shown offset from src while repeating

nearpc: make command repeatable to show further instructions

The pc gets adjusted to the last instructions address making it
visually easy to follow where to continue reading the assembly.

This also forwards repeating of emulate() and pdisass()

telescope: make command repeatable with adjusted offset from src

This also forwards stack() to be repeatable.

### DIFF
--- a/pwndbg/hexdump.py
+++ b/pwndbg/hexdump.py
@@ -53,7 +53,7 @@ def load_color_scheme():
     color_scheme[-1] = '  '
     printable[-1] = ' '
 
-def hexdump(data, address = 0, width = 16, skip = True):
+def hexdump(data, address = 0, width = 16, skip = True, offset = 0):
     if not color_scheme or not printable:
         load_color_scheme()
     data = list(bytearray(data))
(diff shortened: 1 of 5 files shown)


zachriggle:
Can we double-check that the `repeat` functionality only occurs when hitting `<enter>` after a command is executed, and *not* when an identical command is re-executed?

anthraxx:
Yes please double-check. I did and seems working here

zachriggle (approved):

pwndbg/hexdump.py
```diff
@@ -89,9 +89,12 @@ def hexdump(data, address = 0, width = 16, skip = True):
                 hexline.append(printable[char])
             hexline.append(H.separator('%s' % config_separator))
 
-
         yield(''.join(hexline))
 
+    # skip empty fooder if we printed something
```
disconnect3d: `fooder` -> `footer`

anthraxx: fixed

disconnect3d:
There is one bug with that - the nearpc highlights an instruction when repeated:
https://i.gyazo.com/39dfb23ca40c827fe075070b30e9d334.png

anthraxx:
Don't care much about near pc repeat for now so gonna remove it and think about a smart way later

disconnect3d:
@anthraxx imho better to leave as it is.
I mean, at least it lets to repeat ;D

PR #396, opened 2018-01-11
config: validate context-sections and show all available values

anthraxx:
When setting an illegal value, fall back to the default sections.

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -40,7 +40,21 @@ def clear_screen():
 config_clear_screen = pwndbg.config.Parameter('context-clear-screen', False, 'whether to clear the screen before printing the context')
 config_context_sections = pwndbg.config.Parameter('context-sections',
                                                   'regs disasm code stack backtrace',
-                                                  'which context sections are displayed by default (also controls order)')
+                                                  'which context sections are displayed (controls order)')
+
+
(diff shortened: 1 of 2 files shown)


pwndbg/commands/context.py
```diff
@@ -43,6 +43,19 @@ def clear_screen():
                                                   'which context sections are displayed by default (also controls order)')
 
 
+@pwndbg.config.Trigger([config_context_sections])
+def validate_context_sections():
+    valid_values = [context.__name__.replace('context_', '') for context in context_sections.values()]
+    for section in get_context_sections():
+        if section not in valid_values:
+            config_context_sections.value = config_context_sections.default
+            return
```
disconnect3d: Can we print some hint/warn/error about `valid_values` and that we are switching back to default?

anthraxx: it existed once... i swear... weird well the hell it vanished... :open_mouth:

anthraxx:
adjusted adding warning and valid values, the setting of default values will be shown automatically below

Kyle-Kyle:
After this change, what should I do if I don't want any context sections?

disconnect3d:
@Kyle-Kyle haha, this is super good question :heart: will bring it back!

disconnect3d:
@Kyle-Kyle will be there after merging https://github.com/pwndbg/pwndbg/pull/409

PR #397, opened 2018-01-11
context: bring back args section (default off)

anthraxx:
This allows to use args section via the context-sections config
setting (default off).
Additionally introduce a nearpc-show-args config value making it
possible to disable showing it trice while using the args section.

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -278,22 +278,17 @@ def context_backtrace(frame_count=10, with_banner=True):
     return result
 
 
-def context_args():
-    result = []
+def context_args(with_banner=True):
+    args = pwndbg.arguments.format_args(pwndbg.disasm.one())
 
-    ##################################################
-    # DISABLED FOR NOW, I LIKE INLINE DISPLAY BETTER
-    ##################################################
-    # # For call instructions, attempt to resolve the target and
-    # # determine the number of arguments.
-    # for arg, value in pwndbg.arguments.arguments(pwndbg.disasm.one()):
-    #     code   = False if arg.type == 'char' else True
-    #     pretty = pwndbg.chain.format(value, code=code)
-    #     result.append('%-10s %s' % (arg.name+':', pretty))
-    # if not result:
-    #         return []
-    # result.insert(0, pwndbg.ui.banner("arguments"))
-    return result
+    # early exit to skip section if no arg found
+    if not args:
+        return []
+
+    if with_banner:
+        args.insert(0, pwndbg.ui.banner("arguments"))
+
+    return args
 
 last_signal = []
 
@@ -333,6 +328,7 @@ def context_signal():
 context_sections = {
     'r': context_regs,
     'd': context_disasm,
+    'a': context_args,
     'c': context_code,
     's': context_stack,
     'b': context_backtrace
--- a/pwndbg/commands/nearpc.py
+++ b/pwndbg/commands/nearpc.py
@@ -38,6 +38,7 @@ pwndbg.color.theme.Parameter('highlight-pc', True, 'whether to highlight the cur
 pwndbg.color.theme.Parameter('nearpc-prefix', '►', 'prefix marker for nearpc command')
 pwndbg.config.Parameter('left-pad-disasm', True, 'whether to left-pad disassembly')
 nearpc_lines = pwndbg.config.Parameter('nearpc-lines', 10, 'number of additional lines to print for the nearpc command')
+show_args = pwndbg.config.Parameter('nearpc-show-args', True, 'show call arguments below instruction')
 
 @pwndbg.commands.ParsedCommand
 @pwndbg.commands.OnlyWhenRunning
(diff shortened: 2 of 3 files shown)


zachriggle:
Can't we just leave this default-off and set it `on` via config?

anthraxx:
Don't understand, thats exactly what we do. This doesn't change any behaviour or current layout

pwndbg/commands/context.py
```diff
@@ -278,21 +279,23 @@ def context_backtrace(frame_count=10, with_banner=True):
...
-    #     pretty = pwndbg.chain.format(value, code=code)
-    #     result.append('%-10s %s' % (arg.name+':', pretty))
-    # if not result:
-    #         return []
-    # result.insert(0, pwndbg.ui.banner("arguments"))
+    # For call instructions, attempt to resolve the target and
+    # determine the number of arguments.
+    one = pwndbg.disasm.one()
+    for arg, value in pwndbg.arguments.get(one):
+        code   = False if arg.type == 'char' else True
```
disconnect3d: This could be just `arg.type != 'char'`. Btw any idea why do we check for arg type being char to determine whether something is code? What's the reasoning for that? oO'

pwndbg/commands/context.py
```diff
@@ -278,21 +279,23 @@ def context_backtrace(frame_count=10, with_banner=True):
...
+        pretty = pwndbg.chain.format(value, code=code)
+        result.append('%8s%-10s %s' % ('', N.argument(arg.name) + ':', pretty))
+
+    if not result:
+            return []
```
disconnect3d: Something wrong with indentation here.

pwndbg/commands/context.py
```diff
@@ -278,21 +279,23 @@ def context_backtrace(frame_count=10, with_banner=True):
...
-    # if not result:
-    #         return []
-    # result.insert(0, pwndbg.ui.banner("arguments"))
+    # For call instructions, attempt to resolve the target and
+    # determine the number of arguments.
+    one = pwndbg.disasm.one()
+    for arg, value in pwndbg.arguments.get(one):
+        code   = False if arg.type == 'char' else True
+        pretty = pwndbg.chain.format(value, code=code)
+        result.append('%8s%-10s %s' % ('', N.argument(arg.name) + ':', pretty))
```
disconnect3d: `%8s` matches `''` - maybe we should just do:
```
'        %-10s %s'

File: pwndbg/commands/context.py [hunk 4]
```diff
@@ -278,21 +279,23 @@ def context_backtrace(frame_count=10, with_banner=True):
...
+
+    if not result:
+            return []
+
+    if with_banner:
+        result.insert(0, pwndbg.ui.banner("arguments"))
```
disconnect3d: I feel like we should move it to the top of the function as `if not with_banner: return []`

pwndbg/commands/nearpc.py
```diff
@@ -152,10 +153,11 @@ def nearpc(pc=None, lines=None, to_string=False, emulate=False):
...
         # determine the number of arguments.
-        for arg, value in pwndbg.arguments.get(i):
-            code   = False if arg.type == 'char' else True
-            pretty = pwndbg.chain.format(value, code=code)
-            result.append('%8s%-10s %s' % ('', N.argument(arg.name) + ':', pretty))
+        if show_args.value:
+            for arg, value in pwndbg.arguments.get(i):
+                code   = False if arg.type == 'char' else True
+                pretty = pwndbg.chain.format(value, code=code)
+                result.append('%8s%-10s %s' % ('', N.argument(arg.name) + ':', pretty))
```
disconnect3d: Repeated code, maybe we should move it somewhere - I am not sure.

pwndbg/commands/context.py
```diff
@@ -278,21 +279,23 @@ def context_backtrace(frame_count=10, with_banner=True):
     return result
 
 
-def context_args():
+def context_args(with_banner=True):
```
disconnect3d: Is it ever called with `False`? Where?

disconnect3d (changes requested):
Check comments

pwndbg/commands/context.py
anthraxx: sure was just using existing code here, but will replace. no clue about that check

pwndbg/commands/context.py
anthraxx: ups, i now see what you mean while refactoring

pwndbg/commands/context.py
anthraxx: 8 spaces look more weird then having %8s, i think its more obvious that there is a space of 8 something

pwndbg/commands/context.py
anthraxx: the point is to only insert the banner if we want ti print something and the banner was enabled. We cant early exit if no banner is set. it would need more "useless" code and temp local vars to be able to check if we have results other then the banner. IMO for this case its less code if we keep as is

pwndbg/commands/context.py
anthraxx: actually we do that no as the context args code is unified

anthraxx:
please recheck :smiley_cat:

disconnect3d (approved):
lgtm; guess we could use `bool(param)` instead of `param.value` - not sure what's better so lets merge.

PR #399, opened 2018-01-13
Fix nearpc command repeat: highlight, prefix and instruction display

disconnect3d:
This also simplifies the code a bit by moving prefix highlighting nearby
the code that highlights address/content.

### DIFF
--- a/pwndbg/commands/nearpc.py
+++ b/pwndbg/commands/nearpc.py
@@ -47,6 +47,8 @@ def nearpc(pc=None, lines=None, to_string=False, emulate=False):
     Disassemble near a specified address.
     """
 
+    # Repeating nearpc (pressing enter) makes it show next addresses
+    # (writing nearpc explicitly again will reset its state)
     if nearpc.repeat:
         pc = nearpc.last_pc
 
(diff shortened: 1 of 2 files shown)


anthraxx:
This unhighlights the highlighted line and resets it totally to normal, but it should keep the unhighlighted color. I have configured to use a dark grey for thoes addresses but know it shows them in normal color instead of grey.

I think this will need a bit of refactoring as its totally intrannsparent now where you need to continue reading while repeating nearpc. before the highlighted line was the last line of the previous nearpc output. Right now its impossible to properly continue reading the output without jumping somewhere into the middle and search where you need to continue reading.
Maybe we should try repeating so first line of new nearpc is right after last line of last nearpc or we should keep the highlight.

disconnect3d:
Left - before; right - after:
https://i.gyazo.com/793e1ca1780892893ef377ce84d58655.png

disconnect3d:
@anthraxx hmm... I haven't noticed that before but it seems to me that repeating nearpc is just broken?
I mean, we should probably start from NEXT addresses instead of repeating some of them?

anthraxx:
Yes i totally agree, thats what my whole comment above is about

anthraxx:
Yes i totally agree, thats what my whole comment above is about.
Here is an example what i mean by "uncolored" after removing pc highlight:
2018-01-13-142206_2560x1440_scrot

One additional thing: if you disable it this way and you are nearpc like $rip-0xff or such and then while repeating you come to PC then it won't be highlighted neither. hm.

I think it should be something like this:
- if its a non-repeated execution highlight passed PC
- if its a non-repeated execution also show lines before
- if its a repeated execution, possibly highlight the real PC not the passed one
- if its a repeated execution do now show lines before but just continue where it stopped last

disconnect3d:
@anthraxx can you check if everything works fine after last changes?

I have changed `last_pc` to `next_pc` and added `show_prev_insns` argument to `pwndbg.disasm.near` so `nearpc` seems to work fine now.

Screenshow how it works now:
https://i.gyazo.com/4ecfe0f0dbf9cfbbd2305bfae30e273a.png

I am not sure about that part:
> This unhighlights the highlighted line and resets it totally to normal, but it should keep the unhighlighted color. I have configured to use a dark grey for thoes addresses but know it shows them in normal color instead of grey.

anthraxx:
@disconnect3d rough tests show everything i tested looks fine. unhighlighted part looks properly colored from the theme and padding seems always appropriate to the length of the prefix. Lines continue as you repeat so i would say :+1:

anthraxx (approved):

PR #401, opened 2018-01-15
color: make lrjust() work with multiple chars

anthraxx:
This fixes the issue if ljust with multiple characters like
the banner separator char


disconnect3d (approved):
lgtm

Issue #398, opened 2018-01-12
Improvement of pwndbg.config.Parameter


cebrusfs:
Since these change may need python2 and python3 tests and may break something, it is better that for some people that familiar with python2 python3 feature to fix these.

1. Support format string by implementing `__format__` to use `self.value.__format__` and remove all workaround.
   We will get error if we put `Parameter` in the format string via `%` or `str.format()`. We have to convert to `str` first for the workaround.
2. Covert every `str` or `bytes` from utf8 to unicode at the setter of `Parameter` and remove all workaround.
  Some config (ex. `nearpc-prefix`) can be put utf8 symbols and the length is used to calculate for padding. Currently, we force decode everything from utf8 to unicode and then compute string length. According to python3 convention, everything is unicode by default. If we do that in the setter, then the conversion can be removed.

cebrusfs:
For 2, this should only affect gdb with python2.
I found out that the `Parameter.value` will be automatically transform back to `utf-8` with `str` type. I think it is done by gdb's python implementation. So, convert at `get_set_string` to set `self.value` will not works. Also, according to [GDB Python docs](https://sourceware.org/gdb/onlinedocs/gdb/Parameters-In-Python.html), we can not prevent gdb to assign `self.value`.

I have a workaround that we just modified all getter to return `unicode` instead of `str` and then make `Parameter` consistence with Python 3 convention. 
However, when to convert it? and how to keep `self.value` up to date?
  a. we can convert each time at `getter`, but call `codes.decode` may cost computing resource when we use any getter of Parameter.
  b. we can convert one time to put the converted value in some member like `self._value` and discard the usage of the `self.value`. All getters return `self._value`. But we need to maintain these two values and make sure nobody use `param.value` to assign a new value since we can not make a setter of `self.value`.

@disconnect3d What do you think?

cebrusfs:
New workaround:
I remap the member name using `__setattr__` and `__getattribute__`. The gdb python implementation will let any value from `set` command write to `gdb.Parameter.value`. However, I remap the member name of Parameter by the following rule
* `Parameter.value` will map to `gdb.Parameter._value` (if it is string type, always keep `unicode`)
* `Parameter.raw_value` will map to `gdb.Parameter.value`

All the getter of `Parameter` return the `Parameter.value`, but `str()` casting will convert `unicode` back to `str`.

disconnect3d:
[...]

PR #404, opened 2018-01-18
Fix Parameter config class

cebrusfs:
For issue #398.

For 1, Not sure why the format string will not work in origin environment. Since I tested the class which support `__str__` can work with `%s` like the following.
```
class A:
    def __init__(self):
        self.value = 'meow'
    def __str__(self):
        return str(self.value)
a = A()
print('%s' % a)
```
Currently it works perfectly now, so I won't implement `__format__`.

For 2, see fcab5d3.
I remap the member name using `__setattr__` and `__getattribute__`. The gdb python implementation will let any value from set command write to `gdb.Parameter.value`. However, I remap the member name of Parameter by the following rule
* `Parameter.value` will map to `gdb.Parameter._value` (if it is string type, always keep unicode)
* `Parameter.raw_value` will map to `gdb.Parameter.value`

Also,
* Fix `.split()`
* Add `revert_default()`. Then people can just write `param.revert_default()` instead of `param.value = param.default` 
* Fix operator overwriting of `Parameter`
   * Rewrite `__lt__` and `__eq__`

### DIFF
--- a/pwndbg/config.py
+++ b/pwndbg/config.py
@@ -163,8 +163,15 @@ class Parameter(gdb.Parameter):
     def get_show_string(self, svalue):
         return 'Sets %s (currently: %r)' % (self.docstring, self.value)
 
-    def split(self):
-        return str(self).replace(',', ' ').split()
+    def revert_default(self):
+        self.value = self.default
+
+    # TODO: use __getattribute__ to remapping all member function to self.value's member?
+    # Then, we can use param.member() just like param.value.member()
+
+    # The str type member function, used in color/__init__.py
+    def split(self, *args, **kargs):
+        return str(self).split(*args, **kargs)
 
     def __int__(self):
         return int(self.value)
--- a/pwndbg/config.py
+++ b/pwndbg/config.py
@@ -26,6 +26,7 @@ import codecs
 import collections
 import sys
 import types
+from functools import total_ordering
 
 import gdb
 import six
@@ -91,6 +92,7 @@ member_remap = {
     'value': '_value',
     'raw_value': 'value'
 }
+@total_ordering
 class Parameter(gdb.Parameter):
     """
     For python2, we can not store unicode type in self.value since the implementation limitation of gdb python.
@@ -173,6 +175,7 @@ class Parameter(gdb.Parameter):
     def split(self, *args, **kargs):
         return str(self).split(*args, **kargs)
 
+    # Casting
     def __int__(self):
         return int(self.value)
 
@@ -182,8 +185,40 @@ class Parameter(gdb.Parameter):
     def __bool__(self):
         return bool(self.value)
 
+    # Compare operators
+    # Ref: http://portingguide.readthedocs.io/en/latest/comparisons.html
+    # If other is Parameter, comparing by optname. Used in `sorted` in `config` command.
+    # Otherwise, compare `self.value` with `other`
+    def __eq__(self, other):
+        if isinstance(other, gdb.Parameter):
+            return self.optname == other.optname
+        else:
+            return self.value == other
+
     def __lt__(self, other):
-        return self.optname <= other.optname
+        if isinstance(other, gdb.Parameter):
+            return self.optname < other.optname
+        else:
+            return self.value < other
+
+    # Operators
+    def __add__(self, other):
+        return self.value + other
+
+    def __radd__(self, other):
+        return other + self.value
+
+    def __sub__(self, other):
+        return self.value - other
+
+    def __rsub__(self, other):
+        return other - self.value
+
+    def __mul__(self, other):
+        return self.value * other
+
+    def __rmul__(self, other):
+        return other * self.value
 
     def __div__(self, other):
         return self.value / other
@@ -191,20 +226,14 @@ class Parameter(gdb.Parameter):
     def __floordiv__(self, other):
         return self.value // other
 
-    def __mul__(self, other):
-        return self.value * other
-
-    def __sub__(self, other):
-        return self.value - other
-
-    def __add__(self, other):
-        return self.value + other
-
     def __pow__(self, other):
         return self.value ** other
 
     def __mod__(self, other):
         return self.value % other
+
+    def __len__(self):
+        return len(self.value)
 
     # Python2 compatibility
     __nonzero__ = __bool__
--- a/pwndbg/config.py
+++ b/pwndbg/config.py
@@ -22,6 +22,7 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import codecs
 import collections
 import sys
 import types
@@ -86,7 +87,24 @@ def value_to_gdb_native(value):
     return value
 
 
+member_remap = {
+    'value': '_value',
+    'raw_value': 'value'
+}
 class Parameter(gdb.Parameter):
+    """
+    For python2, we can not store unicode type in self.value since the implementation limitation of gdb python.
+    We use self._value as the converted cache and set __getattribute__() and __setattr__() to remap variables.
+
+    Since GDB will set gdb.Parameter.value to user input and call get_set_string(),
+    we use self.raw_value to map back to gdb.Parameter.value
+
+    That is, we remap
+    * Parameter.value -> gdb.Parameter._value (if it is string type, always keep unicode)
+        All getter return this
+    * Parameter.raw_value -> gdb.Parameter.value
+        Only used in get_set_string()
+    """
     def __init__(self, name, default, docstring, scope='config'):
         self.docstring = docstring.strip()
         self.optname = name
@@ -113,11 +131,33 @@ class Parameter(gdb.Parameter):
     def is_changed(self):
         return self.value != self.default
 
+    def __setattr__(self, name, value):
+        new_name = member_remap.get(name, name)
+        new_name = str(new_name) # Python2 only accept str type as key
+        return super(Parameter, self).__setattr__(new_name, value)
+
+    def __getattribute__(self, name):
+        new_name = member_remap.get(name, name)
+        new_name = str(new_name) # Python2 only accept str type as key
+        return super(Parameter, self).__getattribute__(new_name)
+
     def get_set_string(self):
+        value = self.raw_value
+
+        # For string value, convert utf8 byte string to unicode.
+        if isinstance(value, six.binary_type):
+            value = codecs.decode(value, 'utf-8')
+
+        # Remove surrounded ' and " characters
+        if isinstance(value, six.string_types):
+            value = value.strip("\"\'")
+
+        # Write back to self.value
+        self.value = value
+
         for trigger in triggers[self.name]:
             trigger()
-        if isinstance(self.value, str):
-            self.value = self.value.replace("'", '').replace('"', '')
+
         return 'Set %s to %r' % (self.docstring, self.value)
 
     def get_show_string(self, svalue):
--- a/pwndbg/config.py
+++ b/pwndbg/config.py
@@ -24,6 +24,7 @@ from __future__ import unicode_literals
 
 import codecs
 import collections
+import re
 import sys
 import types
 from functools import total_ordering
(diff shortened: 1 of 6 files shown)


pwndbg/config.py
```diff
@@ -86,7 +88,24 @@ def value_to_gdb_native(value):
...
 
 
+member_remap = {
+    'value': '_value',
+    'raw_value': 'value'
+}
+@total_ordering
 class Parameter(gdb.Parameter):
+    """
+    For python2, we can not store unicode type in self.value since the implemetation limitation of gdb python.
```
disconnect3d: Typo `implemetation` -> `implementation`

pwndbg/config.py
```diff
@@ -86,7 +88,24 @@ def value_to_gdb_native(value):
...
+    We use self._value as the converted cache and set __getattribute__() and __setattr__() to remaping the variable.
```
disconnect3d: `to remap variables`

pwndbg/config.py
```diff
@@ -86,7 +88,24 @@ def value_to_gdb_native(value):
...
+
+    Since gdb python will set gdb.Parameter.value as user input and call get_set_string() as setter, we use self.raw_value to map back to gdb.Parameter.value
```
disconnect3d:
```
Since GDB will set gdb.Parameter.value to user input and call get_set_string() ` (...)

File: pwndbg/config.py [hunk 4]
```diff
@@ -135,30 +184,56 @@ def __str__(self):
     def __bool__(self):
         return bool(self.value)
 
+    # Comparing operaters
```
disconnect3d: `Compare operators`

pwndbg/config.py
```diff
@@ -135,30 +184,56 @@ def __str__(self):
...
     def __lt__(self, other):
-        return self.optname <= other.optname
+        if isinstance(other, gdb.Parameter):
+            return self.optname < other.optname
+        else:
+            return self.value < other
 
-    def __div__(self, other):
-        return self.value / other
+    # Operaters
```
disconnect3d: `Operators`

disconnect3d:
@anthraxx u are more into the topic; is this design okay? can we push it forward and merge it? or we want to make those things totally differently?

@zachriggle if u have time, pls visit this too

@cebrusfs btw this has a conflict, rebase to current dev pls

pwndbg/config.py
```diff
@@ -135,30 +184,56 @@ def __str__(self):
...
+    def __add__(self, other):
+        return self.value + other
 
-    def __floordiv__(self, other):
-        return self.value // other
+    def __radd__(self, other):
```
zachriggle, line 213: I feel like there is some weird abstract base class we can inherit from that will automatically add these math operators.  I'd prefer using that rather than specifying each individual method. 

If there's not one, we should create one.

pwndbg/config.py
```diff
@@ -135,30 +184,56 @@ def __str__(self):
...
+    def __eq__(self, other):
+        if isinstance(other, gdb.Parameter):
+            return self.optname == other.optname
+        else:
+            return self.value == other
+
     def __lt__(self, other):
-        return self.optname <= other.optname
+        if isinstance(other, gdb.Parameter):
+            return self.optname < other.optname
```
zachriggle, line 205: Why are we comparing the name on `<`?

pwndbg/config.py
```diff
@@ -113,19 +132,49 @@ def native_default(self):
...
+        return super(Parameter, self).__getattribute__(new_name)
+
     def get_set_string(self):
+        value = self.raw_value
+
+        # For string value, convert utf8 byte string to unicode.
+        if isinstance(value, six.binary_type):
+            value = codecs.decode(value, 'utf-8')
+
+        # Remove surrounded ' and " characters
```
zachriggle, line 154: I'm not sure that stripping quotes is a good idea. 

* Where are quotes necessary?
* What existing cases do we use quotes?
* Doesn't GDB auto-strip quotes?
* How do I intentionally force quotes into a value?

cebrusfs: The original code implement this and use `replace` to do that. I think it is because `gdb.Parametor` will keep all characters. That is, `set color 'bold'` will get a string `'bold'` rather than `bold`. I guess people this to make sure  `set color 'bold'` works the same as `set color bold` since programmer often like to add quotes on string literal.

> How do I intentionally force quotes into a value?

Currently we can't put quote at the beginning and ending. Before my commit, your implement use `replace` to do that. It will remove all the quotes in the string, but I don't see any complain about that. Use `strip` will much better than before.

pwndbg/config.py
cebrusfs: It is used to display all config in `config()` of `command/config.py`. It put all parameter in a list and then sort.

pwndbg/config.py
cebrusfs: I don't find a good base class to do that. Our use case is weird because we just want to proxy all operator to `self.value`. People often inherit the type of `value` to do that in this use case. However, we can't do that because the type of `value` is dynamic and reply on gdb's implementation. That's why we need a operator proxy to wrap the operators.

I think this use case is limited and even we create a abstract class for this, it will be only one class will use it. At least, until now we only have one class that need to implement operator. What do you think?

pwndbg/config.py
cebrusfs: @anthraxx @disconnect3d 
BTW, I saw the PR #409 reply on the quotes. However, the implementation of `dev` `HEAD` will remove all quotes by `replace`. 
https://github.com/pwndbg/pwndbg/blob/103b1def610957e2c74ab616b15a0513a1286cd6/pwndbg/config.py#L120

cebrusfs:
Fix all @disconnect3d mentioned typos and thanks for that.

Also, rebased to `dev`.

pwndbg/config.py
cebrusfs: It seems it will be set after `trigger()` so the checker will get the raw value. So PR #409 works fine and will be broken by this PR.

However, the in-consistence between trigger and final value will be a problem? For example, the trigger function does not expect the quote will be removed.

Can we re-design what we expect to do before/after the trigger function?

zachriggle:
As much as I hate eval(), it may be worth attempting to eval() the values
using something like Pwntools’ safe_eval rather than stripping quotes.

It’s a relatively small chunk of code that might be useful anyway, and
should have more predictable results than just quote-stripping.

cebrusfs:
Sound better than `strip`. But we have more problems here.
* When should we do this kind of `eval` or casting? Since there may be some wrapper function called trigger function which may modify the value. So what kind of value should the wrapper get and return? with quote or without quote?
* if the value is without quote, will there be any problem that using safe_eval. Or we need some error handling on this case. Currently implementation looks it want to accept both kind of value.

zachriggle:
The idea would be that failure to eval would just use the raw string

pwndbg/config.py
disconnect3d: @cebrusfs I like the idea of removing (well replacing) the quotes.

I don't really understand why would it break #409? Anyway, if so, it shouldn't :P

anthraxx:
So far I don't think it will break, but as I heard explicitly "stripping" , please keep in mind that strings containing only whitespace are important to be supported, like putting them in the quotes to work.
They are used for style relevant alignment aka theme. So a prefix configuration value must be able to hold like 2 space characters without any printable character.

cebrusfs:
@anthraxx Yep, it is supported. My implementation only strip the quotes (`'` and `"`).

disconnect3d:
@cebrusfs mm and what if someone wants to use `"` inside a string config variable? :P

pwndbg/config.py
cebrusfs: @disconnect3d 
The original implementation has these feature. I just use `strip` to remove quotes instead of use `replace` to avoid break some string that contains quotes in the middle. Ex. `"me'ow"` should be `me'ow` (after), not `meow` (before).

I said it will break since the quote removing happened after calling wrapper function (`Trigger`) in the original version.  That is the value `'meow'` will still be `'meow'` when the wrapper fetches the value from `self.value` and after that it will be striped and becomes `meow`.

However, my implementation removes the quotes at the first, so the wrapper will get `meow` in these case. So `""` and `''` case will not happend in your code `config_context_sections.value.lower() in ('""', "''", 'none', 'empty'):`.

One way to fix that is moving the quote replacing step after calling the wrapper function (keep original order).

However, this preprocess happens after calling the wrapper functions is weird since the needed of removing quote is caused by gdb argument parser. In python environment (wrapper function), I think it is better to not to have any extra quote striping.

cebrusfs:
@disconnect3d For now, I don't support to set the value as a single quote but it is possible to put the quote between some no-quote characters. However, before my PR, it uses `replace` to do that, so any quote in the string will be removed.  I don't see any issue mentioned his need to put any quote in the config variable.

Maybe we can think how to parse it correctly like use regex to replace `'(.+)'` and `"(.+)"`? For me, I prefer a weak workaround because I don't like eval based solution, it is overkill.

zachriggle:
Can we split the quote-stripping-or-eval-or-whatever stuff into a separate pull request?

cebrusfs:
@zachriggle @disconnect3d @anthraxx 
Hi all,

I change to use regex to remove the quote. Generally, we can put any string in parameter now, but does not support escape sequence. If someone need this, you can create the other PR.
You can set the parameter to `''` by inputting `"''"` or `''''`.

I keep the "quote removing" step before "calling the wrapper function". I review all wrapper function and no one will be affect expect `context` command in PR #409.
However, I found out it will NOT break PR #409 (sorry) since the empty string is handled by the first  condition `not config_context_sections.value`. The only problem is two cases will not useful `""` and `''` since we already removed that quotes. So, I removed them.

If you don't have any concern, you can merge now.

BTW, For the regex, I have some tests on regex pattern, you can add it to autotests after you finished the test framework.
```python
import re

# The first character must be ' or " and ends with the same character.
pattern = r"^(?P<quote>[\"'])(?P<content>.*?)(?P=quote)$"

tests = [
    ("''", ""),
    ('""', ""),

    ("'a'", "a"),
    ('"a"', "a"),

    ("'''", "'"),
    ('"""', '"'),

    ('"\'', None),
    ('\'"', None),

    ('""\'', None),
    ('\'\'"', None),
    ('"\'\'', None),
    ('\'""', None),

    ("a'bc", None),
    ("a\"bc", None),

    ("'a\"b\'c'", "a\"b\'c"),
    ("'a\'b\"c'", "a\'b\"c"),
    ('"a\"b\'c"', "a\"b\'c"),
    ('"a\'b\"c"', "a\'b\"c"),
]


for inp, ans in tests:
    out = re.sub(pattern, r"\g<content>", inp)

    if ans is None: ans = inp

    print repr(inp), repr(out), repr(ans)
    assert out == ans

print 'passed'
```

disconnect3d (approved):
lgtm. Seems to be working fine for both Py2 and Py3.

@zachriggle and @anthraxx can u approve or request changes? I am not going to merge without at least @anthraxx approval here :D (and will try to give some time for @zachriggle to have chance to approve/reject :P).

zachriggle (approved):
Not really sure about the quoting stuff, but the rest is fine and the quoting doesn't affect me directly so #yolo

disconnect3d:
@cebrusfs btw this needs rebase and fixing conflicts :(.

@anthraxx bump - review pls! :P

cebrusfs:
@disconnect3d 
Rebased and reverted #416 since the Parameter object did not have `.decode()`, and it will broken this PR.

anthraxx:
@cebrusfs w00t, why you revert a different fix because it breaks your PR? It the formulation wrong or is the approach reverting is wrong? The problem that commit fixed should be preserved and not reverted.

on py3 system all looks okayish on read and rought testing but im not really sure about py2 and multichar padding after that revert, needs explicit testing with a py2 env if those funcs _always_ get unicode instances

cebrusfs:
@anthraxx sorry for misleading.  The main reason is not that it breaks my PR. It is because #416 is a work around for some unicode character used in config for py2 and my PR can completely fix this. The comment also mentioned it is a workaround until this PR has pushed.
The other problem is, in the current design, the `x` passed in the ljust, rjust functions is not string type. It is `Parameter` class because Parameter class is designed for wrapper/proxy class of the `value`. It supports most of operators, but does not support to call the un-implement member function such as `decode()`.
We have two way to fix this.
1. add implementation or remove the using of decode
But we will not need #416 anymore and have plan to remove it, so I decide to revert it.
2. pass all unknown member function calling of `Parameter` to the `value`'s type.
I think it is a better way to do that, but it can be future work.

For the tests, I tested on the py2 env and works fine. The original issue #412 is caused by the config `Parameter` class provide the utf8 sequence but in this PR we change to keep everything is `unicode` type.

anthraxx:
sounds good, i guess it should all work then.

PR #409, opened 2018-01-27
Bring back possibility of empty context

disconnect3d:
Bringing back the feature we killed in PR 396 - https://github.com/pwndbg/pwndbg/pull/396#issuecomment-359363576

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -47,14 +47,19 @@ config_context_sections = pwndbg.config.Parameter('context-sections',
 
 @pwndbg.config.Trigger([config_context_sections])
 def validate_context_sections():
-    # if trying to set an empty string, we assume default is wanted
-    if not config_context_sections.value:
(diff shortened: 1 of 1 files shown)


disconnect3d:
@zachriggle @anthraxx pls a fast review if possible as this is kinda a dev hotfix :P

anthraxx:
Wouldn't a magic value of 'none' be better so one can reset to default values as well?
Just an idea to have more possibilities

disconnect3d:
Yeah I think we can add none too.

27.01.2018 5:39 PM "Levente Polyak" <notifications@github.com> napisał(a):

anthraxx:
well i thought about `none` exlucisbly for disabling context. I think it may be handy to set to the default value when entering no value but at the end i don't care too much. Just find it also loose of posibillities to not being able to use the defaults, whatever that is

disconnect3d:
After last changes all of `''`, `""`, `'empty'`, `none` (case insensitive) set no sections.

Issue #406, opened 2018-01-20
Context isn't showing the X29 register ( frame pointer ) while debugging aarch64 binary


bruce30262:
As title, pwndbg isn't showing the x29 register while debugging the aarch64 binary:

<img width="360" alt="default" src="https://user-images.githubusercontent.com/4478350/35185097-9e96202a-fe39-11e7-8809-c54ee6ccd973.PNG">

if we use the `i r` command:
```
........
x27            0x0      0
x28            0x0      0
--- missing ---
x29            0x4141414141414141       4702111234474983745
--- missing ---
x30            0x4008c4 4196548
sp             0x7f326e57dce0   0x7f326e57dce0
pc             0x4008c4 0x4008c4
cpsr           0x60000000       1610612736
fpsr           0x0      0
fpcr           0x0      0
```

disconnect3d:
@br0ns Hey, thx for reporting the issue and sry for so long reply.

All the regs are specified here:
https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/regs.py#L109-L110

So... for x29 and x30 we have to change `range(29)` to `range(31)`, for the rest, just add them?

Would you mind sending a PR that would fix that?

zachriggle:
@bruce30262 see the above comment <-- @disconnect3d mentioned @br0ns by accident

I would prefer that the $FP register is hidden behind a setting similar to $EFLAGS and $LR register:

https://github.com/pwndbg/pwndbg/blob/869e8320e123f4fe2d039a31c12ee8c7be7a54d3/pwndbg/commands/context.py#L107-L108

bruce30262:
@disconnect3d  in Aarch64 x30 = LR, which is already shown in the context, so we only have to show x29.  
However I'm not sure if I use 
```python
gpr     = tuple('x%i' % i for i in range(30)),
```
or use
```python
frame   = 'x29',                                                                                                                                                                
gpr     = tuple('x%i' % i for i in range(29)),
```
to show x29. I've tested both code and the result looks the same. I prefer the latter though.

@zachriggle According to line [114~117 in context.py](https://github.com/pwndbg/pwndbg/blob/869e8320e123f4fe2d039a31c12ee8c7be7a54d3/pwndbg/commands/context.py#L114) :

```python
if not regs and pwndbg.config.show_retaddr_reg:
[...]

PR #418, opened 2018-02-02
display the frame pointer register (x29) in aarch64 context

bruce30262:
For issue #406.

result:  
default

For now the context will always display the x29 register. Feel free to discuss if we have to add a config point for it. IMHO it's OK to always display the FP register.


disconnect3d (approved):
Thanks! Gonna click merge after I come back home in some hours if noone is against this.

Issue #387, opened 2018-01-03
update in glibc 2.26 which is called tcache not handled in arena, heap, bins commands


motazreda:
Related:
* https://www.phoronix.com/scan.php?page=news_item&px=glibc-malloc-thread-cache

mdebski:
Hi,

may I ask why did you close it? I believe this is a valid feature request. In fact I just started exploring on how this could be implemented.

disconnect3d:
@mdebski idk, reopening!

mdebski:
Added initial support in #420. It only works if program is compiled with -lpthread, as otherwise gdb cannot access thread-local variables. I don't know yet how tls works without pthread (tcache is still enabled), will look into this.

motazreda:
@mdebski sorry for closing it, my bad maybe by mistake :), and really thanks for the implementation

but a small question, are you sure that it works if the code compiled with lpthread ?
to be honest, I am not sure but I think it works by default without adding -lpthread option.

disconnect3d:
The tcache support has been added by @mdebski so I am closing it. Thanks!

PR #420, opened 2018-02-22
Implement support for ptmalloc's tcache in heap/

mdebski:


pwndbg/commands/heap.py
```diff
@@ -119,19 +130,33 @@ def arenas():
...
 
-    
+@pwndbg.commands.ParsedCommand
+@pwndbg.commands.OnlyWhenRunning
+def tcache():
+    """
+    Prints out the thread cache.
+    """
+    main_heap   = pwndbg.heap.current
+    tcache  = main_heap.thread_cache
```
disconnect3d: Just single space before/after `=` pls

pwndbg/commands/heap.py
```diff
@@ -119,19 +130,33 @@ def arenas():
...
+        addr = int(arena['next'])
         arena = heap.get_arena(addr)
 
 
-    
+@pwndbg.commands.ParsedCommand
+@pwndbg.commands.OnlyWhenRunning
+def tcache():
+    """
+    Prints out the thread cache.
```
disconnect3d: Some more description would be awesome - e.g. libc version it is connected to.

pwndbg/commands/heap.py
```diff
@@ -119,19 +130,33 @@ def arenas():
...
 
         hdr = message.hint(fmt % (hex(addr) if addr else 'main'))
         print(hdr, M.heap(str(h)))
-        addr = int(arena['next'])        
+        addr = int(arena['next'])
         arena = heap.get_arena(addr)
 
 
-    
+@pwndbg.commands.ParsedCommand
```
disconnect3d: Can we change this to `ArgparsedCommand` - see e.g. `pwndbg/commands/auxv.py`

pwndbg/commands/heap.py
```diff
@@ -119,19 +130,33 @@ def arenas():
...
+@pwndbg.commands.OnlyWhenRunning
+def tcache():
+    """
+    Prints out the thread cache.
+    """
+    main_heap   = pwndbg.heap.current
+    tcache  = main_heap.thread_cache
+
+    if tcache is None:
+        return
```
disconnect3d: Can we improve UX and display some info/warning/whatever?

pwndbg/heap/ptmalloc.py
```diff
@@ -42,6 +44,24 @@ def main_arena(self):
...
+        if tcache_addr is not None:
+            try:
+                self._thread_cache = pwndbg.memory.poi(self.tcache_perthread_struct, tcache_addr)
+                _ = self._thread_cache['entries'].fetch_lazy()
+            except Exception as e:
+                print(message.error("Error fetching tcache."))
+                self._thread_cache = None
+        else:
+            print(message.error('Symbol \'tcache\' not found. Try installing libc '
+                                'debugging symbols and try again.'))
```
disconnect3d: <3 cool msg

mdebski:
Intending to solve #387. Will work more on that in a few days.

pwndbg/heap/ptmalloc.py
```diff
@@ -93,6 +125,18 @@ def malloc_alignment(self):
         return pwndbg.arch.ptrsize * 2
 
 
+    @property
+    @pwndbg.memoize.reset_on_objfile
+    def malloc_size_sz(self):
```
disconnect3d: If it corresponds to some define/constant within libc code, can we add some info about that in a docstring? (maybe even add a link to source)

disconnect3d: Same for the rest

pwndbg/heap/ptmalloc.py
```diff
@@ -228,7 +288,32 @@ def fastbins(self, arena_addr=None):
...
+    def tcachebins(self, tcache_addr=None):
+        tcache = self.get_tcache(tcache_addr)
+
+        if tcache is None:
+            return
+
+        counts = tcache['counts']
+        entries = tcache['entries']
+
+        num_tcachebins    = entries.type.sizeof // entries.type.target().sizeof
```
disconnect3d: Redundant spaces before `=`

pwndbg/commands/heap.py
mdebski: ptmalloc.Heap->thread_cache displays one already.

pwndbg/heap/ptmalloc.py
mdebski: copied from main_arena above :P

disconnect3d:
I know this takes time with the merge but I currently don't have a setup to test those things, so I have asked a friend to do so and it seems she have found out that there are some other heap related bugs introduced recently.

So I would say this PR is currently blocked by https://github.com/pwndbg/pwndbg/issues/424

mdebski:
ok

pon., 26 lut 2018 o 19:41 użytkownik Disconnect3d <notifications@github.com>
napisał:

disconnect3d:
@mdebski since you are already on the topic, could you try to fix https://github.com/pwndbg/pwndbg/issues/424?

Eterna1:
Hello. @disconnect3d asked me to test this. 

Few potential bad things:

- When I have glibc<2.26 which doesnt have tcache and i type 'bins', it prints an error that my glibc is without debug symbols (it is with debug symbols). The error is: `Symbol 'tcache' not found. Try installing libc debugging symbols and try again.`

- As we see in the output below the sizes of tcachebins and other chunks, fastbins for example doesn't mach:

```
tcachebins
0x18 [  1]: 0x555555757670 ◂— 0x0
0x28 [  2]: 0x5555557576e0 —▸ 0x5555557576b0 ◂— 0x0
fastbins
0x20: 0x0
0x30: 0x0
0x40: 0x0
0x50: 0x0
0x60: 0x0
0x70: 0x0
0x80: 0x0
```

When I do malloc(0x10) and free this chunk, it would be fatsbin 0x20 on glibc<2.26. But here it goes to tcachebin 0x18. Even on glibc >=2.26, the full size of whole chunk is still 0x20 measuring size in the same way as in other bins.  Other bins measure size by obtaining a value from [chunk->size](http://www.flier.net.cn/2017/10/24/Heap-Exploitation-base-knowledge-notes/3589312.png) . It looks like you use other size measurement (tidx2usize). Why?

Here is the output from malloc_playground.c:

```
Starting program: /home/a/Desktop/how2heap/malloc_playground 
> malloc 0x10
==> 0x555555757670
> malloc 0x10
==> 0x555555757690
> free 0x555555757670
==> ok
```
bins shows the size 0x18:

```
pwndbg> bins
tcachebins
0x18 [  1]: 0x555555757670 ◂— 0x0
```

but chunk->size is 0x20:

```
pwndbg> x/gx 0x555555757670-8
0x555555757668:	0x0000000000000021
```
I think that it is better to keep the same method for measuring size for all bins.
What do you think about this?

mdebski:
Thanks @Eterna1 !
I fixed the two issues.

mak:
@mdebski Actually tcache have different binning than rest of bins, so maybe you should be print both, size of chunk and tcache bin it resides in?

mdebski:
Could you elaborate on that? I believe it is best to define the size of bin as the size of chunks that end up in that bin, and we should consistently use the same notion of "size of chunk"

mak:
Hmm nvm, i see you use it for printing so that doesn't really matter - tcache binning is operating on requested size rather than chunk size but i guess its a matter of perspective ;]

disconnect3d:
@mak So can we say that you approve this code and it works as expected? :D

Eterna1:
Hello. 
The second thing is that chunks in tcache shows the addres of an area returned by malloc.
pwndbg shows the beginning of the chunk (prev_size) in other types of bins. I think that it's better to have everywhere one convention. Especially that it is possible to escape from tcache by freeing more that 7 chunks - they will go to fastbin or unsorted bin. @mak @mdebski @disconnect3d What do you think about this?

disconnect3d:
@mak @mdebski any response to @Eterna1 's concerns? I am not really into the heap topic :(.
 
@mdebski can you fix conflicts so we can merge it?

disconnect3d:
@mdebski Thx again for the PR and sorry that it took so long to merge it :f.

PR #421, opened 2018-02-24
Profiling and performance improvements

mdebski:
PwnDBG used to fall behind competitors in terms of performance, most importantly latency of common commands. Turns out it wasn't inherent to its design (multiple dependencies and modularity), but rather caused by few simple bugs. This pull request reduces run time of "context" command from 100ms to <20ms, without disabling any features.

Scripts for profiling and benchmarking are also included, so that more performance work could be done if needed.

67d994f Add scripts for benchmarking and profiling pwndbg commands
      1ST RUN: 0.11329619199386798
      2ND RUN: 0.09588435597834177
MULTIPLE RUNS: 0.09586518989817705

edab4ac Fix performance issue in emulator.py
      1ST RUN: 0.06423687600181438
      2ND RUN: 0.04655148499296047
MULTIPLE RUNS: 0.04660505420179106

7e8c222 Fix performance issue in syntax_highlight.
      1ST RUN: 0.035918902984121814
      2ND RUN: 0.021684469014871866
MULTIPLE RUNS: 0.021692958200583236

dd68b3c Minor performance improvements in syntax_highlight
      1ST RUN: 0.02984601998468861
      2ND RUN: 0.021192542015342042
MULTIPLE RUNS: 0.021089625198510476

5227d48 Memoize IDA availability.
      1ST RUN: 0.024894433008739725
      2ND RUN: 0.017409588996088132
MULTIPLE RUNS: 0.0171059307991527

### DIFF
--- a/profiling/benchmark.sh
+++ b/profiling/benchmark.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+# Benchmark context command
+make test > /dev/null
+git log --abbrev-commit --pretty=oneline HEAD^..HEAD
+gdb ./test \
+  -ex "b main" -ex "r" \
+  -ex "python import timeit; print('      1ST RUN:', timeit.repeat('pwndbg.commands.context.context()', repeat=1, number=1, globals=globals())[0])" \
+  -ex "si" \
+  -ex "python import timeit; print('      2ND RUN:', timeit.repeat('pwndbg.commands.context.context()', repeat=1, number=1, globals=globals())[0])" \
+  -ex "si" \
+  -ex "python import timeit; print('MULTIPLE RUNS:', timeit.repeat('pwndbg.commands.context.context()', repeat=1, number=10, globals=globals())[0] / 10)" \
+  -ex "quit" | grep 'RUNS*:'
--- a/profiling/benchmark.sh
+++ b/profiling/benchmark.sh
@@ -3,6 +3,7 @@
 make test > /dev/null
 git log --abbrev-commit --pretty=oneline HEAD^..HEAD
 gdb ./test \
+  -ex "source ../gdbinit.py" \
   -ex "b main" -ex "r" \
   -ex "python import timeit; print('      1ST RUN:', timeit.repeat('pwndbg.commands.context.context()', repeat=1, number=1, globals=globals())[0])" \
   -ex "si" \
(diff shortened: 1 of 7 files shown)


mdebski:
For comparision, peda's context:
      1ST RUN: 0.09551754000131041
      2ND RUN: 0.09371281502535567
MULTIPLE RUNS: 0.02516432770062238

profiling/benchmark.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+# Benchmark context command
+make test > /dev/null
```
disconnect3d, line 3: Aren't we missing a makefile here?

mdebski: no, make has predefined rules for c.
https://ftp.gnu.org/old-gnu/Manuals/make-3.79.1/html_chapter/make_10.html

disconnect3d: Good to know, I wasn't even in `profiling` directory. Nice.

disconnect3d:
Btw how it it supposed to be run? It seems I cannot make it so that it properly imports `pwndbg`.

mdebski:
I just assumed pwndbg is already set up (and sourced in .gdbinit). Well, it needs to be set up anyway, but I may source it in scripts, why not.

disconnect3d (approved):

disconnect3d:
Thanks a lot for this! <3

zachriggle:
I’m not sure caching the IDA availability per objfile is best. Maybe on prompt?

mdebski:
What do you mean by on prompt?

I don't really know what to do with that, but I believe it is unnecessary to try to connect on every command, and it takes some time. Maybe just add an option to enable/disable ida integration?

zachriggle:
There's a few ideas that can work

1. Always connect every command (old solution, clearly flawed)
2. Try to connect on objfile load with `reset_on_objfile` (makes connecting to IDA mid-session impossible)
3. Try to connect on `stop` (i.e., every time the debugger halts) with `reset_on_stop` 
4. Add a configuration point to disable IDA entirely.  Perhaps this should be default-off.
5. Add a command to manually connect (e.g. `ida-connect`) to which we could even add the IP and port options to ignore the config parameters

I think that (3) is preferable to (2), and both (4) and (5) are good ideas.

Another option might try to do the connection check in a different thread, and update some global to indicate success.

mdebski:
Done (3), (4) in #423

PR #430, opened 2018-03-20
Fix #429 - osabi check for non-English GDB version

### LINKED ISSUE #429: Pwndbg failing with GDB using language different than English
### Description

After installing pwndbg with a GDB that uses Spanish language we do fail on detecting whether `osabi` is Linux.

This happens after launching GDB:
https://prnt.sc/itwf3u

And this is the return of `show osabi` for GDB with spanish language:
```
(gdb) show osabi                                     
El actual SO ABI es «auto» (actualmente «GNU/Linux»).
El SO ABI predeterminado es «GNU/Linux».             
```

This is the code responsible for failure:
https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/abi.py#L127-L140

disconnect3d:
Detailed info is within the #429 issue, but TLDR:

```
(gdb) show osabi
El actual SO ABI es «auto» (actualmente «GNU/Linux»).
El SO ABI predeterminado es «GNU/Linux».
```

### DIFF
--- a/pwndbg/abi.py
+++ b/pwndbg/abi.py
@@ -125,17 +125,23 @@ def update():
     global linux
 
     # Detect current ABI of client side by 'show osabi'
-    osabi_string = gdb.execute('show osabi', to_string=True)
+    #
+    # Examples of strings returned by `show osabi`:
+    # 'The current OS ABI is "auto" (currently "GNU/Linux").\nThe default OS ABI is "GNU/Linux".\n'
+    # 'The current OS ABI is "GNU/Linux".\nThe default OS ABI is "GNU/Linux".\n'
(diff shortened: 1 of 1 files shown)


disconnect3d:
Everything works fine on this branch:
download 1

zachriggle:
This PR is currently broken as it will falsely detect Linux when it's only the default -- not the current abi

disconnect3d:
@zachriggle Is it? That's why I did `.split('\n')[0]`, so that here we get current osabi instead of default:
```
pwndbg> set osabi none
pwndbg> show osabi
The current OS ABI is "none".
The default OS ABI is "GNU/Linux".
```

zachriggle:
Missed the split, I was looking at the comments (which do have a newline in them, I just missed it)

LGTM!

PR #431, opened 2018-03-23
add basic rust support

Mic92:
When a rust binary is loaded gdb will not find the usual c types.


Mic92:
Types can be also checked out here: https://doc.rust-lang.org/book/second-edition/ch03-02-data-types.html

Mic92:
Great. All test passes!

zachriggle:
LGTM

PR #432, opened 2018-03-24
Fixes #428 - pwndbg.memory.write encoding error

### LINKED ISSUE #428: command ez crashed in the newest dev version
when I tried to write a string into memory, I got two crashed errors
```
pwndbg> ez 0x603000 "this is a test"
  File "/home/bluecake/pwndbg/pwndbg/commands/__init__.py", line 109, in __call__
  File "/home/bluecake/pwndbg/pwndbg/commands/__init__.py", line 200, in _OnlyWhenRunning
    return function(*a, **kw)
  File "/home/bluecake/pwndbg/pwndbg/commands/windbg.py", line 162, in ez
    return eX(1, address, data[0], hex=False)
  File "/home/bluecake/pwndbg/pwndbg/commands/windbg.py", line 191, in eX
    pwndbg.memory.write(address + (i * size), data)
  File "/home/bluecake/pwndbg/pwndbg/memory.py", line 102, in write
    gdb.selected_inferior().write_memory(addr, bytes(data))
TypeError: string argument without an encoding

  File "/home/bluecake/pwndbg/pwndbg/commands/__init__.py", line 71, in invoke
    return self(*args, **kwargs)
  File "/home/bluecake/pwndbg/pwndbg/commands/__init__.py", line 112, in __call__
    self.function.__doc__.strip()))
AttributeError: 'NoneType' object has no attribute 'strip'
Error occurred in Python command: 'NoneType' object has no attribute 'strip'
```
It seems that the data should be encoded with UTF-8 encoding format.

Here is my system version,
```
Gdb:      GNU gdb (Ubuntu 7.11.1-0ubuntu1~16.5) 7.11.1
Python:   3.5.2 (default, Nov 23 2017, 16:37:01)  [GCC 5.4.0 20160609]
Pwndbg:   1.0.0 build: f595ba3
Capstone: 4.0.1024
Unicorn:  1.0.1
```

disconnect3d:
It seems that pwndbg.memory.write fix for Py2 introduced in 433bf231 wasn't tested properly on Py3 as `bytes` without an argument fails as:
```
In [1]: bytes('xyz')
TypeError: string argument without an encoding
```

In Py2 by default the `bytes` is just `str` and so doesn't accept the encoding argument.

Because of that a `from builtins import bytes` has been added which makes it behave the same way as in Python 3 and so `utf8` encoding has been passed.

Some more info on `builtins` module can be found here: http://python-future.org/imports.html#imports-of-builtins

---
I have tested this solution on both Py2 and Py3 GDB versions and it works.

### DIFF
--- a/pwndbg/memory.py
+++ b/pwndbg/memory.py
@@ -9,6 +9,7 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import os
+from builtins import bytes
 
 import gdb
 
@@ -99,7 +100,7 @@ def write(addr, data):
         addr(int): Address to write
         data(str,bytes,bytearray): Data to write
     """
-    gdb.selected_inferior().write_memory(addr, bytes(data))
+    gdb.selected_inferior().write_memory(addr, bytes(data, 'utf8'))
 
 
 def peek(address):


disconnect3d:
@zachriggle Apart from this PR: should we introduce `from builtins import ... #everything?` in all of our source files? This is a question for the future - I wouldn't rush that. We need tests first.

Issue #410, opened 2018-01-30
find_fake_chunk and malloc_chunk bug with chunksize 0x7f or similar size


iromise:
### Description
When I am using find_fake_chunk and malloc_chunk nearby malloc_hook, i just get the following error

```shell
pwndbg> x/8gx 0x7f30f5494b20-16
0x7f30f5494b10 <__malloc_hook>:	0x0000000000000000	0x0000000000000000
0x7f30f5494b20 <main_arena>:	0x0000000000000000	0x0000000000000000
0x7f30f5494b30 <main_arena+16>:	0x000000000143f320	0x0000000000000000
0x7f30f5494b40 <main_arena+32>:	0x0000000000000000	0x0000000000000000      
pwndbg> find_fake_fast 0x7f30f5494b10 0x7f
FAKE CHUNKS
Exception occured: malloc_chunk: Cannot access memory at address 0x7f30f5400000 (<class 'gdb.MemoryError'>)
For more info invoke `set exception-verbose on` and rerun the command
```shell
pwndbg> x/8gx 0x7f30f5494b20-40+8+5
0x7f30f5494b05 <__memalign_hook+5>:	0x30f5155a0000007f	0x000000000000007f
0x7f30f5494b15 <__malloc_hook+5>:	0x0000000000000000	0x0000000000000000
0x7f30f5494b25 <main_arena+5>:	0x0000000000000000	0x000143f320000000
0x7f30f5494b35 <main_arena+21>:	0x0000000000000000	0x0000000000000000
pwndbg> malloc_chunk 0x7f30f5494b05
Traceback (most recent call last):
  File "/home/iromise/pwndbg/pwndbg/commands/__init__.py", line 109, in __call__
    return self.function(*args, **kwargs)
  File "/home/iromise/pwndbg/pwndbg/commands/__init__.py", line 200, in _OnlyWhenRunning
    return function(*a, **kw)
  File "/home/iromise/pwndbg/pwndbg/commands/heap.py", line 203, in malloc_chunk
    fastbins = main_heap.fastbins(arena)
  File "/home/iromise/pwndbg/pwndbg/heap/ptmalloc.py", line 212, in fastbins
    arena = self.get_arena(arena_addr)
  File "/home/iromise/pwndbg/pwndbg/heap/ptmalloc.py", line 159, in get_arena
    return pwndbg.memory.poi(self.malloc_state, arena_addr)
[...]

PR #435, opened 2018-03-25
fix find_fake_chunk #410

mak:
see #410


zachriggle (approved):
Not sure what's going on but LGTM

PR #439, opened 2018-04-17
Don't check for IDA Pro if it is dissabled

benediktwerner:
This makes it so that no checks for the IDA Pro API are performed if it is disabled anyways using the config option `ida-enabled` that was added recently.

Besides improving performance this also fixes connection errors from showing up on WSL.

I'm not sure if this is the right branch to merge this into. I guess this is a bug fix, so should I instead merge it into stable?

Also on a side note on WSL there still is still an error message on startup if the connection to the IDA API fails (which it always does if you don't have IDA). It looks like it's supposed to be this way, but I feel like this doesn't make much sense, as it's not really an error if you simply don't have IDA. Showing a message is okay, but the whole stacktrace is a bit much I think. Any opinions on this? Should I maybe open a seperate issue for this?

### DIFF
--- a/pwndbg/ida.py
+++ b/pwndbg/ida.py
@@ -88,6 +88,8 @@ class withIDA(object):
         functools.update_wrapper(self, fn)
 
     def __call__(self, *args, **kwargs):
+        if not ida_enabled:
+            return None
         if _ida is None:
             init_ida_rpc_client()
         if _ida is not None:


disconnect3d:
Hey,

Thanks for the patch!

> I guess this is a bug fix, so should I instead merge it into stable?

That's how it is supposed to be but it seems `stable` is a bit outdated with fixes recently. We need to clean it up. Saying this, lets leave it as it is (merge to dev).

> Also on a side note on WSL there still is still an error message on startup if the connection to the IDA API fails (which it always does if you don't have IDA). It looks like it's supposed to be this way

As you said - it doesn't make sense - so it would be nice to change it. Can you make a github issue for that? Would you like to work on that?

The main problem here is that we cannot put `set some-param some-value` in `.gdbinit` before sourcing pwndbg. So probably the best way to fix it is to add some `.pwndbg-config` file and fetch default values from there?

disconnect3d (approved):

benediktwerner:
> Can you make a github issue for that? Would you like to work on that?

Sure, I will create a seperate issue and probably try to fix it myself.

> The main problem here is that we cannot put set some-param some-value in .gdbinit before sourcing > pwndbg. So probably the best way to fix it is to add some .pwndbg-config file and fetch default values > from there?
Yes, that really is quite annoying. A `.pwndbg-config` file sounds like a good idea, though I guess it would be even better if it works without it.

zachriggle:
In this case it should be fine without that config.

Nothing should try to connect to IDA until the binary starts. The config
values should be set before this.

disconnect3d:
@zachriggle it is not even with fully default config.

It can be easily seen if you do `nc -l -p 8888` and `gdb /bin/ls`.

The problem is that `init_ida_rpc_client` is called as it is trigger for `ida_rpc_host` and `ida_rpc_port` which is invoked when those parameters are created/set to their default values. Or... at least that's what I believe is happening there, huh.

zachriggle:
That sounds like a bug that we should fix, then -- rather than adding a
whole extra config file.


*Zach Riggle*

Issue #440, opened 2018-04-17
Ida Pro xmlrpc error when starting pwndbg without IDA


benediktwerner:
I am running `pwndbg` on WSL and don't have IDA Pro installed. The latter causes an error on every startup when `pwndbg` checks for IDA Pro. This results in an error message (`[!] Ida Pro xmlrpc error`) followed by the stack trace of the causing error, which when looking at the code seems to be the intended behaviour. I feel this isn't what should be happening in this case. If I don't have IDA running it's obvious that the connection fails and I therefore wouldn't expect a stack trace.

I guess these are the two best solutions to this:
1. Don't show a stack trace on error. Looking at the code it seems that `pwndbg` always prints a stack trace if an error occurred when trying to connect to IDA Pro. Instead, it might just print a message (e.g. "Could not connect to IDA Pro xmlrpc") and maybe print the error to some log file? (Not sure if there is one already or if that is even necessary)
2. Add some kind of option to disable IDA before the startup. At the moment the integration can only be disabled with `set ida-enabled off` which can never be run before `pwndbg` was initialized and tries to connect to IDA for the first time. If for example there was a config file that `pwndbg` reads before startup I could simply disable IDA there.

I feel like the first solution would be better if it could be implemented somewhat decently, as it doesn't require any additional changes by the user, but I'm not sure if it would be okay to just hide all the stack traces. There might be cases where it would be better to show a stack trace.
The second option is certainly easier in that regard and it could also prevent the three second delay that the connection check causes on startup. On the other hand, I'm not sure where the best place for a config file would be and it might also be a bit confusing to have another config file somewhere else?

If there is a consensus on what the best solution would be I would implement that, but I would like to know what you think is the right thing to do. It might even be good to do both things in part?

Also here is the whole error message if you want to have a look:
The error appears in `pwndbg/ida.py` in `init_ida_rpc_client()`
```
[!] Ida Pro xmlrpc error
Traceback (most recent call last):
  File "/mnt/d/Uni/SS18/exploits/pwndbg/pwndbg/pwndbg/ida.py", line 64, in init_ida_rpc_client
    _ida.here()
  File "/usr/lib/python3.4/xmlrpc/client.py", line 1098, in __call__
    return self.__send(self.__name, args)
  File "/usr/lib/python3.4/xmlrpc/client.py", line 1437, in __request
    verbose=self.__verbose
  File "/usr/lib/python3.4/xmlrpc/client.py", line 1140, in request
[...]

PR #442, opened 2018-04-18
Improve behavior without IDA Pro

benediktwerner:
Related #439, #440

## Changes
- Reduced timeout when trying to connect to IDA Pro to one second and made it configurable
- If the connection attempt fails, only display a small message and a hint on how to disable the integration, show more info or increase the timeout time (instead of printing a stack trace like before)
- Remember how long ago the last connection check was and don't check again if it was less than five seconds ago and the last check failed. This prevents situations where multiple checks are performed before the user even gets control again, in which case the timeouts would cause extreme lag.

## Remaining problem
The settings can still only be changed after pwndbg was loaded. This means pwndgb still tries to connect to IDA on every startup. Will this is a bit annoying I think it now isn't a huge problem anymore as the waiting time is now quite small and the error hint isn't really any worse than the wall of text gdb already prints.

One option might to don't connect to IDA on startup. The problem here is that the connection attempt gets triggered, when the "ida-rpc-host", and "ida-rpc-port" get set initially. Maybe it would make sense to simply ignore the first call ever or alternatively check if the program to debug is even running. Not sure if there is a better way to do this than:
```python
gdb.execute("info program") == "the program being debugged is not being run."
```

### DIFF
--- a/pwndbg/config.py
+++ b/pwndbg/config.py
@@ -168,6 +168,7 @@ class Parameter(gdb.Parameter):
             trigger()
 
         if not pwndbg.decorators.first_prompt:
+            # Remove the newline that gdb adds automatically
             return '\b'
         return 'Set %s to %r' % (self.docstring, self.value)
 
--- a/pwndbg/config.py
+++ b/pwndbg/config.py
@@ -32,6 +32,8 @@ from functools import total_ordering
 import gdb
 import six
 
+import pwndbg.decorators
+
 TYPES = collections.OrderedDict()
 
 # The value is a plain boolean.
@@ -165,6 +167,8 @@ class Parameter(gdb.Parameter):
         for trigger in triggers[self.name]:
             trigger()
 
+        if not pwndbg.decorators.first_prompt:
+            return '\b'
         return 'Set %s to %r' % (self.docstring, self.value)
 
     def get_show_string(self, svalue):
--- a/pwndbg/decorators.py
+++ b/pwndbg/decorators.py
@@ -1,0 +1,23 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import functools
+
+
+first_prompt = False
+
+
+def only_after_first_prompt(value_before=None):
+    def decorator(func):
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            if first_prompt:
+                return func(*args, **kwargs)
+            else:
+                return value_before
+        return wrapper
+    return decorator
--- a/pwndbg/decorators.py
+++ b/pwndbg/decorators.py
@@ -11,6 +11,11 @@ first_prompt = False
 
 
 def only_after_first_prompt(value_before=None):
+    """
+    Blocks a function from executing until the first prompt was displayed.
+    The value_before parameter can be used to specify the value that is
+    returned if the function is called before the first prompt was displayed.
+    """
     def decorator(func):
         @functools.wraps(func)
         def wrapper(*args, **kwargs):
--- a/pwndbg/decorators.py
+++ b/pwndbg/decorators.py
@@ -12,8 +12,8 @@ first_prompt = False
 
 def only_after_first_prompt(value_before=None):
     """
-    Blocks a function from executing until the first prompt was displayed.
-    The value_before parameter can be used to specify the value that is
+    Decorator to prevent a function from running before the first prompt was displayed.
+    The 'value_before' parameter can be used to specify the value that is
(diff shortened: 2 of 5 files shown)


pwndbg/ida.py
```diff
@@ -50,14 +52,23 @@
 # to avoid printing the same exception multiple times, we store the last exception here
 _ida_last_exception = None
 
+# to avoid checking the connection multiple times with now delay, we store the last time we checked it
+_ida_last_connection_check = 0
+_IDA_MIN_CONNECTION_DELAY_CHECK = 5
+
 
 @pwndbg.config.Trigger([ida_rpc_host, ida_rpc_port])
```
disconnect3d: We should probably add here the `ida_timeout` parameter so it will set `socket.setdefaulttimeout` when e.g. user fires `set ida_timeout 5`.

pwndbg/ida.py
```diff
@@ -76,10 +87,20 @@ def init_ida_rpc_client():
...
-            print(message.error("[!] Ida Pro xmlrpc error"))
-            traceback.print_exception(*exception)
+            if hasattr(pwndbg.config, "exception_verbose") and pwndbg.config.exception_verbose:
+                print(message.error("[!] Ida Pro xmlrpc error"))
+                traceback.print_exception(*exception)
+            else:
+                exc_type, exc_value, _ = exception
+                print(message.error('Failed to connect to IDA Pro ({}: {})'.format(exc_type.__qualname__, exc_value)))
+                if exc_type is socket.timeout:
+                    print(message.notice('To increase the time to wait for IDA Pro use `') + message.hint('set ida-timeout <new-timeout>') + message.notice('`'))
```
disconnect3d: We should probably say that it is in seconds

pwndbg/ida.py
benediktwerner: Yes, you're right, I added it.

pwndbg/ida.py
benediktwerner: It's already in the config doc, but I guess it doesn't hurt to put it here as well.

benediktwerner:
So should I change it so that pwndbg doesn't connect to IDA directly at startup? And if so should I check if there is a running program or just ignore the first connection call?

zachriggle:
We should add a new type of event that gets called whenever the first
prompt-hook is invoked. Then we can add a decorator that prevents functions
being invoked before the prompt is displayed, and apply it wherever we need
it.

benediktwerner:
Ok, I added a decorator to block a function before the first prompt, which now prevents a connection to IDA before the configs were set. I also used it to block the text output when config values are set from the gdbinit file.

I'm a bit unhappy that I had to create a new file for this, but putting it into `prompt.py` doesn't really work as that would produce circular dependencies with the decorators, so I guess this is the only option.

benediktwerner:
I actually found another way that would allow us to put the decorator and stuff inside `prompt.py` which would be nicer semantically, but it's not really that nice. It would work if we put the code that prints how many commands were loaded into `pwndbg.__init__` (which probably isn't the worst idea anyways) and then use something like this:
```python
if not _prompt_imported:
    import pwndbg.prompt
    _prompt_imported = True
if not pwndbg.prompt.first_prompt:
    # do stuff
```

It looks like it would also work if we ignore the config stuff and only use the decorator in `ida.py`. But I guess that would also limit the use of the decorator in other places later on because they can't be a dependency of `prompt` in any way, so a separate file is probably still the best option.

What do you think?

pwndbg/decorators.py
```diff
@@ -0,0 +1,22 @@
...
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import functools
+
+first_prompt = False
+
+
+def only_after_first_prompt(value_before=None):
+    def decorator(func):
```
disconnect3d: I am not sure about the filename but it would be nice to add here a docstrig (to the `only_after_first_prompt`)

pwndbg/config.py
```diff
@@ -165,6 +167,8 @@ def get_set_string(self):
         for trigger in triggers[self.name]:
             trigger()
 
+        if not pwndbg.decorators.first_prompt:
+            return '\b'
```
disconnect3d: Why `\b`? is there any character we need to remove?

disconnect3d:
I like the seperate file more as it doesn't limit us with anything.

Any comments @zachriggle ?

disconnect3d (approved):
I am not sure if `decorators.py` is the best name but I didn't come up with a better one for now.

benediktwerner:
I added a docstring and a comment explaining the `\b`. The latter is necessary to remove the newline that gdb will add automatically.

As for the name, I thought a bit about it but also couldn't come up with anything better. It would, of course, be much nicer to have it in the `prompt.py` file but as I said that wouldn't really work out.

PR #444, opened 2018-04-28
Support hex data prefixed with 0x for eX windbg commands

disconnect3d:
When using `eX` commands and setting data to hex value prefixed with `0x`, we get an exception:

```
pwndbg> ed 0xffb21ae4 0x55616740
  File "/usr/lib/python3.6/encodings/hex_codec.py", line 19, in hex_decode
    return (binascii.a2b_hex(input), len(input))
binascii.Error: Non-hexadecimal digit found

The above exception was the direct cause of the following exception:

  File "/home/dc/installed/pwndbg/pwndbg/commands/__init__.py", line 109, in __call__
  File "/home/dc/installed/pwndbg/pwndbg/commands/__init__.py", line 200, in _OnlyWhenRunning
    return function(*a, **kw)
  File "/home/dc/installed/pwndbg/pwndbg/commands/windbg.py", line 141, in ed
    return eX(4, address, data)
  File "/home/dc/installed/pwndbg/pwndbg/commands/windbg.py", line 180, in eX
    data    = codecs.decode(bytestr, 'hex')
binascii.Error: decoding with 'hex' codec failed (Error: Non-hexadecimal digit found)
```

This commit fixes this problem so that if the data input has prefix, it is stripped.

### DIFF
--- a/pwndbg/commands/windbg.py
+++ b/pwndbg/commands/windbg.py
@@ -175,7 +175,13 @@ def eX(size, address, data, hex=True):
 
     for i, bytestr in enumerate(data):
         if hex:
-            bytestr = str(bytestr).rjust(size*2, '0')
+            bytestr = str(bytestr)
+
+            if bytestr.startswith('0x'):
+                bytestr = bytestr[2:]
+
+            bytestr = bytestr.rjust(size*2, '0')
+
             data    = codecs.decode(bytestr, 'hex')
         else:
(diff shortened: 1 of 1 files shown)


disconnect3d:
I tested it on both Py2 and Py3 and it works fine.

disconnect3d:
Umm I added too much changes to the commit. Will change it as I get to PC.

PR #448, opened 2018-05-02
Adds stepret command and nextret docstring

disconnect3d:
So that now we can step-until-return-like-instruction like a boss! :)

Below is comparison of `nextret` and `stepret`.

```
mov ...    <------- we are here
mov ...
call foo
mov ...
ret        <------- nextret will stop here

foo:
ret        <------- stepret will stop here
```


zachriggle:
Still disappointed that GDB's API requires us to `gdb.execute("si")` but LGTM

zachriggle (approved):

PR #461, opened 2018-05-09
[WIP] Refactor and tests for nearpc/disasm/emulation

disconnect3d:
* Changed a bit the emulator part
* Added tests for `pdisas`, `emulate`, `u`, `nearpc`
* Refactored a bit emulator part (e.g. more explicit variable names - `size` vs `instruction_size`)

**NOTE: This is WIP so don't merge it yet, I am going to add more testcases so that we can see how things works and if everything is alright**

PS: The branch name comes from an issue I was looking for but it seems the issue is invalid atm


disconnect3d:
I am closing this as the approach I tried with tests is not that good (commit fd71afd I believe).

I tried launching all tests under one gdb by loading different binary after the previous one. I had some issues with GDB that loading one binary after another one resulted in different results as launching seperate GDB sessions. It seems it used symbols from one binary to another or had problem to load symbols from 2nd binary after the 1st.

PR #464, opened 2018-05-12
Fix before_prompt event on old GDB versions

disconnect3d:
This adds an `EventWrapper` class which behaves similar to gdb events but lets us:
* check whether event is a real gdb event or not
* call event callbacks if it is not a real gdb event


disconnect3d:
@bruce30262 can you check this out?

disconnect3d:
Works for me on both both GDB 8.1 and GDB 7.7.1.

PR #467, opened 2018-05-17
Use gdb.VERSION instead of parsing gdb.execute result

disconnect3d:
Yay, this is finally there:

```
>>> gdb.VERSION
'8.1.0.20180409-git'
```


zachriggle:
Does this exist in Ubuntu 16.04’s GDB?

zachriggle:
Never mind, I see the try-catch.

PR #470, opened 2018-05-23
typeinfo: skip failed compile attempts

anthraxx:
This fixes the problem that single header files that are not compilable
do not abort the compilation loop.
Errors about the failure are printed anyway be check_output we just
avoid abortion.

### DIFF
--- a/pwndbg/typeinfo.py
+++ b/pwndbg/typeinfo.py
@@ -153,7 +153,10 @@ def compile(filename=None, address=0):
     if not os.path.exists(objectname):
         gcc     = pwndbg.gcc.which()
         gcc    += ['-w', '-c', '-g', filename, '-o', objectname]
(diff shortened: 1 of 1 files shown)


pwndbg/typeinfo.py
```diff
@@ -153,7 +153,10 @@ def compile(filename=None, address=0):
     if not os.path.exists(objectname):
         gcc     = pwndbg.gcc.which()
         gcc    += ['-w', '-c', '-g', filename, '-o', objectname]
-        subprocess.check_output(' '.join(gcc), shell=True)
+        try:
+            subprocess.check_output(' '.join(gcc), shell=True)
```
disconnect3d: Btw any idea why do we do `shell=True`?
Do we use shell wildcards here?

zachriggle: Nope, we could probably get rid of it.

PR #474, opened 2018-05-28
Command for calculating PIE offsets

rcx:
Hey, I use pwndbg a lot, and one feature I missed from other debuggers was the ability to easily reference memory by RVA (offset from module base). So, I wrote a simple command `piebase` to do that. It just prints out the hex value of the argument plus the module base. It defaults to the program's main module but you can also specify a different module like `ld.so`. Providing no arguments just provides the base address of the main program module.

I also included `breakrva` command, which puts breakpoint at RVA, for convenience.

Maybe there's a better way to do this in pwndbg, but this is what I've been using recently. Feel free to offer feedback before merging.


disconnect3d:
Hey, Thanks for the PR!

We already have  the `rebase` function but it has some cons:
- as it is gdb's function you call it with `$rebase(addr)` which might be a bit unintuitive for people not knowing GDB well (one need to do e.g. `break *$rebase(addr)`)
- it needs the program to be run
- I believe it works only with main objfile for now

So I am up for merging those.

@zachriggle any comment on that?

pwndbg/commands/pie.py
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+Command to print the virtual memory map a la /proc/self/maps.
```
disconnect3d: Please remove the docstring ;)

pwndbg/commands/pie.py
```diff
@@ -0,0 +1,75 @@
...
+"""
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import argparse
+
+import gdb
+import six
```
disconnect3d: Remove unused imports

pwndbg/commands/pie.py
```diff
@@ -0,0 +1,75 @@
...
+def breakrva(offset=None, module=None):
+    offset = int(offset)
+    if not module:
+        module = pwndbg.auxv.get()['AT_EXECFN']
+    addr = translate_addr(offset, module)
+    spec = "*%#x" % (addr)
+    gdb.Breakpoint(spec)
+
+
+@pwndbg.commands.QuietSloppyParsedCommand
```
disconnect3d, line 63: So its just an alias, isn't it?

1) We had a plan to remove all command classes and leave only `ArgparsedCommand`; we should probably change this to `ArgparsedCommand` or add some logic to it so that we can make aliases...

2) I bet this does show `None` on `pwndbg brva` as we don't have a function docstring here.

Could you move it to `ArgparsedCommand` (and add some alias-like logic) or add the docstring if you don't want to bother with refactoring, please?

pwndbg/commands/pie.py
```diff
@@ -0,0 +1,75 @@
...
+    offset = int(offset)
+    if not module:
+        module = pwndbg.auxv.get()['AT_EXECFN']
+    addr = translate_addr(offset, module)
+    print(hex(addr))
+
+
+parser = argparse.ArgumentParser()
+parser.description = 'Break at RVA from PIE base.'
+parser.add_argument('offset', nargs='?', default='0',
```
disconnect3d: We can probably make it `default=0`

pwndbg/commands/pie.py
```diff
@@ -0,0 +1,75 @@
...
+                    help='Offset from PIE base.')
+parser.add_argument('module', type=str, nargs='?', default='',
+                    help='Module to choose as base. Defaults to the target executable.')
+
+@pwndbg.commands.ArgparsedCommand(parser)
+@pwndbg.commands.OnlyWhenRunning
+def piebase(offset=None, module=None):
+    offset = int(offset)
+    if not module:
+        module = pwndbg.auxv.get()['AT_EXECFN']
```
disconnect3d: According to https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/auxv.py#L166-L172 this key might not be in auxv dict and we might crash with `KeyError`.

Any idea what are the situations in which this key is not present? Can we handle this edge case somehow?

pwndbg/commands/pie.py
rcx: I'm not really sure how to convert it to `ArgparsedCommand` and do the alias logic, so I added a simple docstrong. Sorry :\

pwndbg/commands/pie.py
rcx: To address this, I added a fallback using `readlink(/proc/pid/self)`.

```python
def get_exe_name():
    module = ''
    auxv = pwndbg.auxv.get()
    if 'AT_EXECFN' in auxv:
        module = auxv['AT_EXECFN']
    if not module:
        module = os.readlink("/proc/%d/exe" % (gdb.selected_inferior().pid,))
    return module
```

rcx:
The rest of the requested changes have been addressed, please lmk what you think.

pwndbg/commands/pie.py
```diff
@@ -34,9 +27,18 @@ def translate_addr(offset, module):
     print("No executeable segments found")
     return
 
+def get_exe_name():
+    module = ''
+    auxv = pwndbg.auxv.get()
+    if 'AT_EXECFN' in auxv:
+        module = auxv['AT_EXECFN']
+    if not module:
+        module = os.readlink("/proc/%d/exe" % (gdb.selected_inferior().pid,))
```
disconnect3d: We can probably use `pwndbg.proc.exe` (there is also `pwndbg.proc.pid`).

pwndbg/commands/pie.py
```diff
@@ -34,9 +27,18 @@ def translate_addr(offset, module):
     print("No executeable segments found")
     return
 
+def get_exe_name():
+    module = ''
+    auxv = pwndbg.auxv.get()
+    if 'AT_EXECFN' in auxv:
+        module = auxv['AT_EXECFN']
```
disconnect3d: We can remove this function and do:

```python
module = pwndbg.auxv.get().get('AT_EXECFN', pwndbg.proc.exe)
```

rcx:
OK, `pwndbg.proc.exe` works for me. Much nicer, too.

~~One thing I am worried about is the assumption that the executeable page will always be mapped at the lowest address for a module, e.g. at the base. I'm not well-versed enough at linux executeable loading internals to know whether that's always the case or not. It seems to work in practice.~~

Edit: After some consideration I've opted to just return the lowest-address page corresponding to a module.

pwndbg/commands/pie.py
disconnect3d: Np, its not that simple task as it might require to redo the code a bit as each command needs to be a seperate class or instance (don't remember which one huh) :P

disconnect3d:
> the assumption that the executeable page will always be mapped at the lowest address for a module, e.g. at the base.
Where exactly is it?

Afair similar or this assumption made us some problems in the past when someone wanted to use pwndbg for bare metal debugging.

disconnect3d:
Got an approval from @zachriggle on irc so merging ;)
Thanks for that! <3

(don't forget to ask about beer if we ever meet on a conference :P)

rcx:
Thanks!!

Issue #476, opened 2018-05-30
The reverse- commands of rr stop working after a SIGSEGV when used with pwndbg


andigena:
### Description

When running rr with pwndbg and encountering a SIGSEGV in the debugged application, the reverse- commands of rr stop working. After executing them, the same context is printed as before, rip and the other registers/memory remains in the same state, etc. (see the attached session below).

If I disable pwndbg, they work fine in the same situation. They also work with pwndbg when there's no SIGSEGV encountered. Another thing I noticed is that I can step back from SIGABRTs, so it might be specific to segmentation faults?

I also think I've seen this issue in the not-so-recent past (last year maybe), so it's likely not something new.

### Steps to reproduce

```
tukan@farm:~/t$ cat sigsegv.c
int main() {
    *(int *)0x414141 = 0x41;
    return 0;
}
tukan@farm:~/t$ gcc -O0 -o sigsegv sigsegv.c 
tukan@farm:~/t$ rr record ./sigsegv
rr: Saving execution to trace directory `/home/tukan/.local/share/rr/sigsegv-1'.
Segmentation fault
tukan@farm:~/t$ rr replay
```
rr session attached: [rr-session.txt](https://github.com/pwndbg/pwndbg/files/2053502/rr-session.txt)

### My setup

* Ubuntu 16.04 amd64
* Python 3.5.2 (default, Nov 23 2017, 16:37:01) 
* GNU gdb (Ubuntu 7.11.1-0ubuntu1~16.5) 7.11.1
* rr latest (f3bde22)
* pwndbg latest (f56f271)

disconnect3d:
Hey, thanks for reporting this. I will take a look on that tomorrow.

disconnect3d:
We've both investigated it today and talked on irc, the problem is caused by `gdb.parse_and_eval` invoked here:
https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/commands/context.py#L371-L375

We use this code to display to the user that something has happend with debugging program (it exited, segaulted etc) AFTER we display context. GDB does it by default but it does it before the context and so most of the time the user doesn't see that.

If we remove them, everything works fine.

disconnect3d:
I have also asked about how to detect rr here: https://mail.mozilla.org/pipermail/rr-dev/2018-May/000626.html

disconnect3d:
[...]

PR #478, opened 2018-06-01
Fixes 476 - segfault handling when using rr project

disconnect3d:
https://github.com/pwndbg/pwndbg/issues/476

When using rr project with pwndbg and getting a segfault it is unable to use reverse stepping provided by rr (e.g. `reverse-stepi`).

This is caused by using `gdb.parse_and_eval($_siginfo<something here>)` when handling segfault.

This is there so that we can display a message to the user that a segfault occured AFTER displaying context (GDB displays this info too, but before we display the context...).

We have to either change the logic and don't use `gdb.parse_and_eval` or detect rr.

This PR adds rr detection by checking for two globals that are defined when running with rr. It also changes the way we inform user about segfault to display current pc instead.

There are also some other ways to detect rr but I am not sure which is the best option:
* we could check prompt using `return '(rr)' in gdb.execute('show prompt', to_string=True)`
* we could check memory mappings as there are some additional memory pages present e.g. in `vmmap` when we use rr:
```
        0x70000000         0x70001000 r-xp     1000 0      /usr/lib/rr/rr_page_64_replay
        0x70001000         0x70002000 rw-p     1000 0      /tmp/rr-shared-preload_thread_locals-29373-0

    0x7f7f853c5000     0x7f7f853eb000 r-xp    26000 0      /usr/lib/rr/librrpreload.so
    0x7f7f853eb000     0x7f7f855ea000 ---p   1ff000 26000  /usr/lib/rr/librrpreload.so
    0x7f7f855ea000     0x7f7f855eb000 r--p     1000 25000  /usr/lib/rr/librrpreload.so
```

Each of the approaches have some flaws as they might change in the future or the mappings paths might be different on different machines.

Loud thinking: maybe we should change prompt from `(rr)` to `rr+pwndbg`? But I am not sure if we can do it as probably our code is executed before rr's code (we could try doing it in prompt hook tho).

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -419,4 +419,3 @@ def _is_rr_present():
     interpreter_globals = ast.literal_eval(globals_list_literal_str)
 
     return 'RRCmd' in interpreter_globals and 'RRWhere' in interpreter_globals
-
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -5,6 +5,7 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import ast
 import codecs
 import sys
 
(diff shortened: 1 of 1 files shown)


andigena:
I've verified that this fixes the issue. As for detecting rr, I agree that the approach you chose is probably the best.

Issue #443, opened 2018-04-18
heap don't show thread's heap


pullp:
### Description
recently I use pwndbg to dbg a program which call malloc in a thread, but after malloc, I use heap to
show the heaps and I don't see the chunk malloced by the thread, only the chunks in the main arena

### Steps to reproduce
here's the c source code
```c
#include <stdio.h>
#include <stdlib.h>
#include <pthread.h>
#include <unistd.h>
#include <sys/types.h>

void* threadFunc(void* arg) {
        void *ptr;
        ptr = malloc(0x10000);
        printf("11ptr: %p\n", ptr);
}

int main() {
        pthread_t t1;
        void* s;
        int ret;
        ret = pthread_create(&t1, NULL, threadFunc, NULL);
        if(ret)
        {
                printf("Thread creation error\n");
                return -1;
        }
        ret = pthread_join(t1, &s);
        if(ret)
        {
                printf("Thread join error\n");
                return -1;
        }
        return 0;
}
```

I first set a breakpoit at  `printf("11ptr: %p\n", ptr);` in threadFunc.
then run to the breakpoint
then use heap instruction, but I get following answer:
```
Top Chunk: 0x602120
Last Remainder: 0

0x602000 PREV_INUSE {
  prev_size = 0x0, 
  size = 0x121, 
  fd = 0xf, 
  bk = 0x0, 
  fd_nextsize = 0x1, 
  bk_nextsize = 0x0
}
0x602120 PREV_INUSE {
  prev_size = 0x0, 
  size = 0x20ee1, 
  fd = 0x0, 
  bk = 0x0, 
  fd_nextsize = 0x0, 
  bk_nextsize = 0x0
}
```
which don't include chunk just malloced in the threadFunc

version info:
>Gdb:      GNU gdb (Ubuntu 7.11.1-0ubuntu1~16.5) 7.11.1
Python:   3.5.2 (default, Nov 23 2017, 16:37:01)  [GCC 5.4.0 20160609]
Pwndbg:   1.0.0 build: 99e5b84
Capstone: 4.0.1024
Unicorn:  1.0.1

disconnect3d:
Hey! Thanks for reporting the bug. I just have checked and indeed, there is a problem.

I am not really into the heap stuff and its implementation in pwndbg that much. Would you like to try fixing this?

disconnect3d:
Fixed with PR #479.

PR #479, opened 2018-06-06
ptmalloc multiple heaps per non-main arena support, related fixes

### LINKED ISSUE #451: Other arenas are not marked as heap (and so we color them badly)
When looking on #443 I have noticed that vmmap colors second thread's heap memory as data instead of heap.

This can be observed on the screen below:
https://i.gyazo.com/80dc95f1bc0bb5415e4bd5f792f5d542.png

andigena:
This PR is an attempt at adding some support for multiple heaps per arena in the ptmalloc code and fixing  related issues (#451, #443).

Currently, the ptmalloc code in pwndbg is oblivious to the fact that a non-main arena can have multiple mmapped heaps. The main arena doesn't have separate heaps, just the contiguous brk region. A heap in a non-main arena is typically 1 Mb (x86) or 64 Mb (x64) in size and as soon as one is full, a new one is created with mmap and the top pointer of the arena is updated. Each of these heaps begin with a `heap_info` structure, that has a pointer to the owning arena. The first one also has the actual arena structure (`malloc_state`) after the `heap_info`. The heaps corresponding to an arena are linked via a backward linked list. 

I'll use the following code to showcase the changes below:

```C
//  gcc -o thread_heaps_multi -O0  -ggdb thread_heaps_multi.c -lpthread
#include <stdio.h>
#include <stdlib.h>
#include <pthread.h>
#include <unistd.h>
#include <sys/types.h>

void* threadFunc(void* arg) {
        void *ptr;
        size_t heap_size = 64 * 1024 * 1024;
        for (size_t i = 0 ; i < heap_size / 16; i++)
            ptr = malloc(0);
        printf("11ptr: %p\n", ptr);
}

int main() {
        pthread_t t1;
        void* s;
        int ret;
        ret = pthread_create(&t1, NULL, threadFunc, NULL);
        if(ret)
        {
                printf("Thread creation error\n");
                return -1;
        }
        ret = pthread_join(t1, &s);
        if(ret)
        {
                printf("Thread join error\n");
                return -1;
        }
        return 0;
}
```

The code starts a thread and allocates enough memory in it so that 3 different heaps are created under that non-main arena.

When compiling and running this on Ubuntu 16.04 x64 under pwndbg, the arenas and vmmap commands give the following output when invoked at a breakpoint on line 12 (the printf in `threadFunc`):

image

In the output of the `arenas` command, only one heap is shown for the non-main arena, which corresponds to the 0x4000000 (64 mb) sized mapping in the vmmap output, while in fact there are two more at `0x7fffe8000000` and `0x7fffec000000` as can be seen when examining the memory there:

image

All three has the same arena pointer, and they are used in the order `0x7ffff0000000`, `0x7fffe8000000`, `0x7fffec000000`. Note that it's not a strictly decreasing order, this is a result of an optimization in the malloc implementation. To find these heaps, we need to travel the linked list of heaps, that starts from the last heap and ends in the original one for the arena. To find the last heap of a non-main arena, I used arena->top. Here's the output of the same command with this patch applied:

image

The output of the arenas command is the arena name/address (omitted for heaps that belong to the same arena), the address of the first chunk in the heap, then the containing vm map.

Visiting the heaps allows us to:

* provide objfile names for the corresponding vm mappings that begin with `[heap`, which allows proper coloring. I chose the format `[heap arena_id:heap_id]`. Note how in the screenshot above, two heaps share a single memory mapping, due to the above mentioned optimization. Also note the 0x3fdf000 sized region at `0x7fffec021000`. This is space that's part of the last allocated heap but malloc mprotects it to rw as needed. This fixes #451.
* add support for multiple heaps per arena for the `arenas` command.
* keep the vmmap output consistent with /proc/pid/maps. Note how in the first picture, a mapping starts at `0x7ffff00008b0`, which is of course not correct. This is a result of alterations made to the Page object in `get_region`. This enabled the `heap` command to find the location to start dumping chunks from at the cost of the maps inconsistency and making the beginning of the region (`0x7ffff0000000-0x7ffff00008b0`) incorrectly colored. However, since this PR provides a way to display heaps, along with the address of the first chunk in the heap, this is no longer needed. Subsequently, the `heap` command is also refactored a bit, so that it simply dumps chunks from the address provided (or the main_arena if no address is given). This addresses #443.

### DIFF
--- a/pwndbg/heap/ptmalloc.py
+++ b/pwndbg/heap/ptmalloc.py
@@ -15,6 +15,8 @@ from pwndbg.color import message
 from pwndbg.constants import ptmalloc
 from pwndbg.heap import heap_chain_limit
 
+# See https://sourceware.org/git/?p=glibc.git;a=blob;f=malloc/arena.c;h=37183cfb6ab5d0735cc82759626670aff3832cd0;hb=086ee48eaeaba871a2300daf85469671cc14c7e9#l30
+# and https://sourceware.org/git/?p=glibc.git;a=blob;f=malloc/malloc.c;h=f8e7250f70f6f26b0acb5901bcc4f6e39a8a52b2;hb=086ee48eaeaba871a2300daf85469671cc14c7e9#l869
 HEAP_MAX_SIZE = 1024 * 1024 if pwndbg.arch.ptrsize == 4 else 2 * 4 * 1024 * 1024 * 8
 
 Arena = namedtuple('Arena', ['addr', 'heaps'])
--- a/pwndbg/heap/ptmalloc.py
+++ b/pwndbg/heap/ptmalloc.py
@@ -5,7 +5,7 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-from collections import OrderedDict
+from collections import OrderedDict, namedtuple
 
 import gdb
 
@@ -15,7 +15,11 @@ from pwndbg.color import message
 from pwndbg.constants import ptmalloc
 from pwndbg.heap import heap_chain_limit
 
-HEAP_MAX_SIZE     = 1024 * 1024
+HEAP_MAX_SIZE = 1024 * 1024 if pwndbg.arch.ptrsize == 4 else 2 * 4 * 1024 * 1024 * 8
+
+Arena = namedtuple('Arena', ['addr', 'heaps'])
+HeapInfo = namedtuple('HeapInfo', ['addr', 'first_chunk'])
+
 
 def heap_for_ptr(ptr):
     "find the heap and corresponding arena for a given ptr"
@@ -27,6 +31,8 @@ class Heap(pwndbg.heap.heap.BaseHeap):
         # Global ptmalloc objects
         self._main_arena    = None
         self._mp            = None
+        # List of arenas/heaps
+        self._arenas        = None
         # ptmalloc cache for current thread
         self._thread_cache  = None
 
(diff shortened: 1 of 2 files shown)


zachriggle (changes requested):
Not a huge fan of nested functions if we can avoid it.  Feel free to refactor and break out new types where it makes sense!

disconnect3d:
Hey, thanks for the PR! <3
I will review it thoroughly in upcomming days or after the weekend.

pwndbg/heap/ptmalloc.py
```diff
@@ -15,7 +15,11 @@
 from pwndbg.constants import ptmalloc
 from pwndbg.heap import heap_chain_limit
 
-HEAP_MAX_SIZE     = 1024 * 1024
+HEAP_MAX_SIZE = 1024 * 1024 if pwndbg.arch.ptrsize == 4 else 2 * 4 * 1024 * 1024 * 8
```
disconnect3d: Can we provide any source links to that? e.g. a link to libc commit or some explanatory comment with glibc version etc

andigena: Sure.

andigena:
@zachriggle do you mean turning `Arena` and `HeapInfo` into proper classes with pretty printers?

andigena:
Refactored the nested functions into `__str__`s of proper classes and added two glibc source links to the `HEAP_MAX_SIZE` value.

pwndbg/heap/ptmalloc.py
```diff
@@ -5,28 +5,61 @@
...
 import pwndbg.events
 import pwndbg.typeinfo
 from pwndbg.color import message
 from pwndbg.constants import ptmalloc
 from pwndbg.heap import heap_chain_limit
 
-HEAP_MAX_SIZE     = 1024 * 1024
+# See https://sourceware.org/git/?p=glibc.git;a=blob;f=malloc/arena.c;h=37183cfb6ab5d0735cc82759626670aff3832cd0;hb=086ee48eaeaba871a2300daf85469671cc14c7e9#l30
+# and https://sourceware.org/git/?p=glibc.git;a=blob;f=malloc/malloc.c;h=f8e7250f70f6f26b0acb5901bcc4f6e39a8a52b2;hb=086ee48eaeaba871a2300daf85469671cc14c7e9#l869
+HEAP_MAX_SIZE = 1024 * 1024 if pwndbg.arch.ptrsize == 4 else 2 * 4 * 1024 * 1024 * 8
```
disconnect3d: Can we also add a comment ` # 1 Mb (x86) or 64 Mb (x64)`

pwndbg/heap/ptmalloc.py
```diff
@@ -5,28 +5,61 @@
...
+
+    def __str__(self):
+        res = []
+        prefix = '[%%%ds]    ' % (pwndbg.arch.ptrsize * 2)
+        prefix_len = len(prefix % (''))
+        arena_name = hex(self.addr) if self.addr != pwndbg.heap.current.main_arena.address else 'main'
+        res.append(message.hint(prefix % (arena_name)) + str(self.heaps[0]))
+        for h in self.heaps[1:]:
+            res.append(' ' * prefix_len + str(h))
+
```
disconnect3d, line 43: Maybe:
```python
main_heap = message.hint(prefix % (arena_name)) + str(self.heaps[0])

fmt = lambda heap: ' ' * prefix_len + str(heap)
other_heaps =  `'\n'.join(map(fmt, self.heaps[1:]))

return main_arena + '\n' + other_heaps
```

Unfortunately this doesn't work in Py2: ```'.'.join(('x', *map(str, [1,2])))``` :(

EDIT
Current solution is also fine

disconnect3d:
@andigena We had some changes in heap before and this is another one.
Could you extend https://github.com/pwndbg/pwndbg/blob/dev/FEATURES.md#heap-inspection OR create a HEAP.md explaining/documenting the features we have? In another PR

disconnect3d (approved):
See comments but generally LGTM

pwndbg/heap/ptmalloc.py
andigena: Added it.

pwndbg/heap/ptmalloc.py
andigena: I like both versions, left it as it was for now.

andigena:
Yeah I'll do another PR for the FEATURES file once this is merged.

disconnect3d:
Thx again! :)

PR #486, opened 2018-06-20
Objfile event dispatching fix

### LINKED ISSUE #481: new_objfile caching prevents event dispatching
### Description

I noticed that the [caching for new_objfile events](https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/events.py#L153) actually prevents the calling of event handlers for that event, except for the first one registered. In effect, only the `_start_newobjfile` handler is called for new object files, none of the others.

### Steps to reproduce

Clone [this fork](https://github.com/andigena/pwndbg/tree/newobjfile_event) and check out the `newobjfile_event` branch. It adds some debug prints to the caching code, and registers a new handler for `new_objfile` that just prints a message if it's called. Then start debugging any program, and note how only the first registered handler is called for `new_objfile` events, which happens to be `_start_newobjfile`.

A sample session can be found [here](https://gist.github.com/andigena/bf4852ced1c0e6df214e5a6be4c6fe2f)

Moving the `im_just_a_print` declaration before the `_start_newobjfile` switches the registration order, so only `im_just_a_print` will actually be dispatched.

### My setup

Ubuntu 16.04 x64, 

```
pwndbg> version
Gdb:      7.11.1
Python:   3.5.2 (default, Nov 23 2017, 16:37:01)  [GCC 5.4.0 20160609]
Pwndbg:   1.0.0 build: 11e22ba
Capstone: 4.0.1024
Unicorn:  1.0.0
```

andigena:
This PR fixes #481 and a latent bug that is uncovered by the fix. 

* the first commit replaces the `objfile_cache` set with a dictionary, indexed with the objfile names, that contains which event handler functions have already been dispatched for the objfile in question. This allows calling multiple handlers for an objfile, while also preventing multiple dispatch of the same handler.
* after this fix, the previously uncalled handlers are now actually invoked. `pwndbg.vmmap.get` attempts to read process state without checking if the process is alive, leading to exceptions as can be seen [here][0] (interleaved with some debug output). An objfile event is triggered when the binary is loaded before the program is started, so the second commit adds a check to see if the process is alive before trying to access the process state. 

[0]: https://gist.github.com/andigena/c5805b3edc21449eef81cbd63aa336c0

### DIFF
--- a/pwndbg/events.py
+++ b/pwndbg/events.py
@@ -138,7 +138,7 @@ class Pause(object):
 # objects are loaded.  This greatly slows down the debugging session.
 # In order to combat this, we keep track of which objfiles have been loaded
 # this session, and only emit objfile events for each *new* file.
-objfile_cache = set()
+objfile_cache = dict()
 
 
 def connect(func, event_handler, name=''):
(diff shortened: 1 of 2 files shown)


disconnect3d:
Thanks a lot <3!

PR #489, opened 2018-06-28
Refactors heap.get_region, adds special case for get_heap_boundaries.

andigena:
As discussed [here](https://github.com/pwndbg/pwndbg/commit/dfb69bd91777cb7889973a65447db45ca68fe901#commitcomment-29463129), the [heap] vm region and the actual start of the heap are sometimes different, e.g. `[heap]` starts at `0x61f000` while `mp_.sbrk_base` is `0x620000`.

This PR refactors `get_region` and `get_heap_boundaries`, so that the first one only returns the actual mappings without modifying them and the latter returns a (possibly fake) Page object that correspond to the actual heap boundaries. Also changes the callers of these functions where appropriate.

I have done minimal manual testing.

### DIFF
--- a/pwndbg/heap/ptmalloc.py
+++ b/pwndbg/heap/ptmalloc.py
@@ -85,7 +85,7 @@ class Heap(pwndbg.heap.heap.BaseHeap):
         arenas          = []
         arena_cnt       = 0
         main_arena_addr = int(arena.address)
-        sbrk_page       = self.get_region().vaddr
+        sbrk_page       = self.get_heap_boundaries().vaddr
 
         # Create the main_arena with a fake HeapInfo
         main_arena      = Arena(main_arena_addr, [HeapInfo(sbrk_page, sbrk_page)])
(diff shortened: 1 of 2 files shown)


pwndbg/heap/ptmalloc.py
```diff
@@ -347,42 +347,42 @@ def get_heap_boundaries(self, addr=None):
         Get the boundaries of the heap containing `addr`. Returns the brk region for
         adresses inside it or a fake Page for the containing heap for non-main arenas.
         """
+        page = pwndbg.memory.Page(0, 0, 0, 0)
         brk = self.get_region()
         if addr is None or brk.vaddr < addr < brk.vaddr + brk.memsz:
-            return brk
+            # Occasionally, the [heap] vm region and the actual start of the heap are
```
disconnect3d, line 353: Can we add an explanation for that (or a link to source code that explains it?) e.g. in next PR?

pwndbg/heap/ptmalloc.py
```diff
@@ -347,42 +347,42 @@ def get_heap_boundaries(self, addr=None):
...
             page.memsz = int(heap['size'])
             return page
 
 
-    def get_region(self,addr=None):
+    def get_region(self, addr=None):
         """
-        Finds the memory map used for heap by using mp_ structure's sbrk_base property
-        and falls back to using /proc/self/maps (vmmap) which can be wrong
-        when .bss is very large
```
disconnect3d, line 365: So now we don't have this case of fallback which could make us have wrong value?

PR #492, opened 2018-07-01
Leak offset probing tool

rcx:
In this pull request I introduce the new command `probeleak [address] [count]`, which essentially scans for pointers in the specified memory. This is useful, for example, if you have a stack leak, and you want to figure out whether you can leak libc, libm, etc. with it. If no arguments are provided, it defaults to `$sp` and length 0x40. The way it works is it reads pointers from the specified memory, and checks if they lie within any currently mapped page using vmmap.

Here is some example output:
```
pwndbg> probeleak
+0x00: 0x0000555555558b66 = (r-xp) /bin/cat + 0x4b66
+0x10: 0x00007ffff7fc6000 = (rw-p) [anon] + 0x1000
+0x30: 0x00005555555565a9 = (r-xp) /bin/cat + 0x25a9
+0x38: 0x00007fffffffdae0 = (rw-p) [stack] + 0x1fae0
pwndbg> probe 0x7ffff7dd94d1
No leaks found :(
pwndbg> probe 0x55555575e360
+0x18: 0x00007ffff7dcfca0 = (rw-p) /lib/x86_64-linux-gnu/libc-2.27.so + 0x1ebca0
+0x30: 0x000055555575e2a0 = (rw-p) [heap] + 0x12a0
```

Moreover, if you diff the command's outputs across multiple runs, you can easily tell if a leak is reliable or not. In my experience this is an invaluable tool for exploit development.

side note: the commit history for the pull request is a bit messed up since there was a previous pull request from this branch that was already merged and squashed. the only "real" commit here is "Leak probing tool" 313fca4


pwndbg/commands/probeleak.py
```diff
@@ -0,0 +1,81 @@
...
+    count   = max(int(count), 0)
+    ptrsize = pwndbg.arch.ptrsize
+    ptr_fmt = '0x%%0%dx' % (ptrsize*2,)
+    off_fmt = '+0x%%0%dx' % (int(math.ceil(math.log(count,2)/4)),)
+
+    if count > address > 0x10000: # in case someone puts in an end address and not a count (smh)
+        count -= address
+
+    if count % ptrsize > 0:
+        print("Warning: count %x is not a multiple of %x" % (count, ptrsize))
```
disconnect3d: Can we be more explicit and:
- change the format string, so it is known its in hex: `0x%x`
- inform user what is the modified count

pwndbg/commands/probeleak.py
```diff
@@ -0,0 +1,81 @@
...
+@pwndbg.commands.ArgparsedCommand(parser)
+@pwndbg.commands.OnlyWhenRunning
+def probeleak(address=None, count=0x40):
+
+    address = int(address)
+    address &= pwndbg.arch.ptrmask
+    count   = max(int(count), 0)
+    ptrsize = pwndbg.arch.ptrsize
+    ptr_fmt = '0x%%0%dx' % (ptrsize*2,)
+    off_fmt = '+0x%%0%dx' % (int(math.ceil(math.log(count,2)/4)),)
```
disconnect3d: 1. We don't need to write an explicit one item tuple with `%` formatting when there is just one argument, lets not do that (e.g. just use `fmt % item`).
2. I think this form is probably more readable?:
```python
In [20]: '0x{:0{n}x}'.format(2, n=16)
Out[20]: '0x0000000000000002'
```

pwndbg/commands/probeleak.py
```diff
@@ -0,0 +1,81 @@
...
+
+    if count > address > 0x10000: # in case someone puts in an end address and not a count (smh)
+        count -= address
```
disconnect3d: Loud thinking: we need to have a possibility to dislay some usage examples for commands.

pwndbg/commands/probeleak.py
```diff
@@ -0,0 +1,81 @@
...
+        if page:
+            if not found:
+                print(M.legend())
+                found = True
+
+            mod_name = page.objfile
+            if not mod_name:
+                mod_name = '[anon]'
+            right_text = ('(%s) %s + 0x%x') % (page.permstr, mod_name, p - page.vaddr + page.offset)
+            print((off_fmt % (i,)) + ': ' + (ptr_fmt % (p,)) + ' = ' + M.get(p, text=right_text))
```
disconnect3d: Lets make it more readable as its hard to follow formatting and concatenation a bit. Maybe something like:
```python
offset_zeros = int(math.ceil(math.log(count,2)/4))
ptr_zeros = ...

fmt = '+0x{offset:0{n1}x}: 0x{ptr:0{n2}x} = {page}'

print(fmt.format(n1=offset_zeros, n2=ptr_zeros, offset=i, ptr=p, page=M.get(p, text=right_text)))
```

disconnect3d: Btw shouldn't we use `message.info` from `pwndbg.color` and so on? @anthraxx what do u think?

pwndbg/commands/probeleak.py
```diff
@@ -0,0 +1,81 @@
...
+    if not found:
+        print('No leaks found :(')
```
disconnect3d: Can we change it to: `'No leaks found at 0x{:x}-0x{:x} :('.format(start_addr, end_addr)`

disconnect3d (commented):
Hey, thanks a lot for this PR, this is indeed an useful feature <3!
I have added some comments that we should address before merging it.

If you have some questions feel free to ask them here or on irc (#pwndbg channel on freenode network).

pwndbg/commands/probeleak.py
```diff
@@ -0,0 +1,81 @@
...
+
+def find_module(addr):
+    mod_filter = lambda page: page.vaddr <= addr < page.vaddr + page.memsz
+    pages = list(filter(mod_filter, pwndbg.vmmap.get()))
+
+    if not pages:
+        return None
+
+    if len(pages) > 1:
+        print('Warning: There is more than one page containing address 0x%x (wtf?)', addr)
```
anthraxx: Please use print(message.warn("")) for warnings (theme)

pwndbg/commands/probeleak.py
anthraxx: use print(message.warn("")) for warning (theme)

pwndbg/commands/probeleak.py
```diff
@@ -0,0 +1,81 @@
...
+        count = count - (count % ptrsize)
+
+    try:
+        data = pwndbg.memory.read(address, count, partial=True)
+    except gdb.error as e:
+        print(e)
+        return
+
+    if not data:
+        print("Couldn't read memory at 0x%x" % (address,))
```
anthraxx: looks like a warning/error
Please use print(message.warn("")) for warnings or error for error (theme)

pwndbg/commands/probeleak.py
```diff
@@ -0,0 +1,81 @@
...
+        count -= address
+
+    if count % ptrsize > 0:
+        print("Warning: count %x is not a multiple of %x" % (count, ptrsize))
+        count = count - (count % ptrsize)
+
+    try:
+        data = pwndbg.memory.read(address, count, partial=True)
+    except gdb.error as e:
+        print(e)
```
anthraxx: Please use print(message.error("")) for errors (theme)

pwndbg/commands/probeleak.py
anthraxx: yeah thought the same, we should use info() here

pwndbg/commands/probeleak.py
```diff
@@ -0,0 +1,81 @@
...
+        print("Couldn't read memory at 0x%x" % (address,))
+        return
+
+    found = False
+    for i in range(0, count, ptrsize):
+        p = pwndbg.arch.unpack(data[i:i+ptrsize])
+        page = find_module(p)
+        if page:
+            if not found:
+                print(M.legend())
```
anthraxx, line 74: If the content is printed via info() we should also use info (or hint) here

pwndbg/commands/probeleak.py
anthraxx: please use info() or hint() here

anthraxx (changes requested):

anthraxx:
Travis also doesnt like some import order and PEP separator lines, please also take care of that to make our beloved travis happy :heart:

disconnect3d:
Yeah, just execute `isort --recursive pwndbg tests` inside pwndbg (you may need `pip install isort!=4.3.0` ;P)

rcx:
Ok, how about this. I added the warn() and error() but I didn't see any `info()` in message.py. Also, I think it makes more sense for it to be uncolored than the hint color (yellow), and lots of other commands use just `print` alone, so I hope it's OK.

disconnect3d:
> I didn't see any info() in message.py
Huh indeed we don't have `info` :D.

We should probably add it in the future.

LGTM for now, so I am merging it :).

disconnect3d (approved):
Thx o/

PR #496, opened 2018-07-07
add vis_heap_chunks

Segflow:
This new heap command allows to visualize heap chunks in different colors. Very handy to know to which chunk a memory cell belongs.

# Examples:
Visualize 3 chunks:
screen shot 2018-07-07 at 2 21 45 pm

Visualize 30 chunks (will stop when top_chunk is reached):
screen shot 2018-07-07 at 2 21 57 pm

### DIFF
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -6,6 +6,7 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import struct
+import argparse
 
 import gdb
 import six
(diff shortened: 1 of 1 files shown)


pwndbg/commands/heap.py
```diff
@@ -371,3 +371,89 @@ def find_fake_fast(addr, size):
...
+
+    fmt = {
+        'little': '<',
+        'big': '>'
+    }[pwndbg.arch.endian] + {
+        4: 'I',
+        8: 'Q'
+    }[pwndbg.arch.ptrsize]
+
+    def unpack(x): return struct.unpack(fmt, x)[0]
```
disconnect3d: We can probably remove the formatting code and just use `pwndbg.arch.unpack(...)`

pwndbg/commands/heap.py
```diff
@@ -371,3 +371,89 @@ def find_fake_fast(addr, size):
 
             if main_heap.fastbin_index(value) == fastbin:
                 malloc_chunk(start+offset-pwndbg.arch.ptrsize,fake=True)
+
+
+@pwndbg.commands.ParsedCommand
+@pwndbg.commands.OnlyWhenRunning
+@pwndbg.commands.OnlyWhenHeapIsInitialized
```
disconnect3d: Could you change it to use `ArgparsedCommand` instead? For the `addr` argument we can use https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/commands/__init__.py#L259-L272

pwndbg/commands/heap.py
```diff
@@ -371,3 +371,89 @@ def find_fake_fast(addr, size):
...
+
+        chunk_id += 1
+
+        # we reached top chunk, add it's metadata and break
+        if addr == top_chunk:
+            cells_map[addr] = chunk_id
+            cells_map[addr+ptr_size] = chunk_id
+            break
+
+    # TODO: maybe print free chunks in bold or underlined
```
disconnect3d: We should remove a TODO comment and either add an issue or add this feature in this PR

pwndbg/commands/heap.py
```diff
@@ -371,3 +371,89 @@ def find_fake_fast(addr, size):
...
+    color_funcs = [
+        generateColorFunction("yellow"),
+        generateColorFunction("cyan"),
+        generateColorFunction("purple"),
+        generateColorFunction("green"),
+        generateColorFunction("blue"),
```
disconnect3d, line 426: By doing it in such way, users can't configure those colors. @anthraxx is there a better way to handle it?

pwndbg/commands/heap.py
```diff
@@ -371,3 +371,89 @@ def find_fake_fast(addr, size):
...
+    ]
+
+    addrs = sorted(cells_map.keys())
+    item_fmt = "\t0x%.16x"
+    if ptr_size == 4:
+        item_fmt = "\t0x%.8x"
```
```
In [2]: '\t0x%.8x' % 123
Out[2]: '\t0x0000007b'

In [3]: '\t0x{:0{n}x}'.format(123, n=8)
Out[3]: '\t0x0000007b'
```

we should use `'\t0x{:0{n}x}'.format(number, n=ptr_size*2)`

zachriggle: I think there may be other cases where we need to do something similar -- we should probably break this out into a separate helper routine somewhere in a common module.

disconnect3d:
I have applied some minor formatting improvements and fixed import ordering (isort was failing on travis).

Thx a lot for the PR :).

There is still https://github.com/pwndbg/pwndbg/pull/496#discussion_r200819294 but I think we can put it in a different PR.

Issue #488, opened 2018-06-21
Wrong context regs display on [remote?] targets that use multiple threads


0xcc77:
### Description

While I was debugging a 32-bit process on an ARM Android device, I sometimes noticed that atleast one of the register addresses in pwndbg view is wrong.
This has happened several times for different registers.

Examples (happened in two different debugging sessions):
```
pwndbg> regs r0
 R0   0xee69a868 —▸ 0xee460a00 ◂— 0x0
pwndbg> i r r0
r0             0xee4335c8       3997382088

pwndbg> i r sp
sp             0xf136d698       0xf136d698
pwndbg> regs sp
*SP   0xf007a820 —▸ 0xf007a834 ◂— 0xffffffff
```

It happened to me again today while debugging so I tried to ask about this in the IRC channel before opening an issue and when trying to debug the problem, one guy said that this problem is somehow similar with [Issue 460](https://github.com/pwndbg/pwndbg/issues/460) and asked me if I could try the things mentioned there.

After trying to disable caching with:
```
pwndbg> python import pwndbg; pwndbg.memoize.memoize.caching = False
```

The pwndbg registers view got immediately updated with the correct addresses for the registers.
Unfortunately, disabling caching make pwndbg really slow.

### Steps to reproduce

I don't have any consistent way to reproduce the issue as it's not always happening and not easy to notice.

### My setup

pwndbg> version
Gdb:      8.1
Python:   3.6.5 (default, May 11 2018, 04:00:52)  [GCC 8.1.0]
Pwndbg:   1.0.0 build: 71d29df
Capstone: 4.0.1024
Unicorn:  1.0.1

disconnect3d:
@hencohen This is a remote-debugging session, isn't it?
Does it happen only if the app has multiple thread?

0xcc77:
@disconnect3d that's correct, It is a remote-debugging session.
I can't be sure as it's hard to notice when this happens but when it happened to me I indeed was debugging a multi-threaded process

disconnect3d:
I will try to reproduce it using the example provided by @bruce30262 at https://github.com/pwndbg/pwndbg/issues/460#issuecomment-387954695 using qemu-aarch64.

If someone wants to help with this:
[...]

PR #500, opened 2018-07-11
Fixes piebase and breakrva on remote debugging

disconnect3d:
Three things here:

1. This fixes `piebase` and `breakrva` commands - a bug with remote targets mentioned in https://github.com/pwndbg/pwndbg/issues/488#issuecomment-403213457.

2. It also adds a check if result address is still in the memory pages belonging to the given module. This works now as:

```
pwndbg> breakrva main
Offset 0x555555554601 rebased to module /home/dc/pwndbg_bug/a.out as 0xaaaaaaaa8601 is beyond module's memory pages:
    0x555555554000     0x555555555000 r-xp     1000 0      /home/dc/pwndbg_bug/a.out
    0x555555754000     0x555555755000 r--p     1000 0      /home/dc/pwndbg_bug/a.out
    0x555555755000     0x555555756000 rw-p     1000 1000   /home/dc/pwndbg_bug/a.out
```

3. It gives a better output for `piebase`:
```
pwndbg> piebase 1
Calculated VA from /home/dc/pwndbg_bug/a.out = 0x555555554001
```

---

To reproduce the fixed bug, launch any binary on a gdbserver:

```
gdbserver 127.0.0.1:4444 ./a.out
```

Then start a debugging session:

```
gdb -q -ex 'target remote 127.0.0.1:4444' ./a.out
```

and fire e.g. `breakrva 123`.

---

Below you can see the bug case and explanation why it occured:

```
pwndbg> breakrva 1
There are no mappings for specified address or module.
'breakrva': Break at RVA from PIE base.
  File "/home/dc/pwndbg/pwndbg/commands/__init__.py", line 109, in __call__
  File "/home/dc/pwndbg/pwndbg/commands/__init__.py", line 200, in _OnlyWhenRunning
    return function(*a, **kw)
  File "/home/dc/pwndbg/pwndbg/commands/pie.py", line 61, in breakrva
    spec = "*%#x" % (addr)
TypeError: %x format: an integer is required, not NoneType
```

So what is the issue here?

1. We have the same logic in both `piebase` and `breakrva` - if the user doesn't specify second argument - a module name - we retrieve it with `get_exe_name`:

```python
def breakrva(offset=None, module=None):
    offset = int(offset)
    if not module:
        module = get_exe_name()

    addr = translate_addr(offset, module)
    spec = "*%#x" % (addr)

    # [ ... - some more code, not important here ]
```

2. The `get_exe_name` returns just `pwndbg.auxv.get().get('AT_EXECFN', pwndbg.proc.exe)`. The difference is important here. On the case shown above the `pwndbg.auxv.get()['AT_EXECFN']` returns `./a.out` while `pwndbg.proc.exe` returns the full path: `/home/dc/pwndbg_bug/a.out`.
3. This `module` is then passed to `translate_addr` as can be seen on the code above.
4. The `translate_addr` tries to retrieve memory page (`Page` instance) which belongs to the module:
```python
def translate_addr(offset, module):
    mod_filter = lambda page: module in page.objfile
    pages = list(filter(mod_filter, pwndbg.vmmap.get()))

    if not pages:
        print('There are no mappings for specified address or module.')
        return

    # [ ... - some more code, not important here ]
```
5. The `translate_addr` returns `None` because the `page.objfile` for e.g. binary objfile returns its full path as can be seen below:
```
(Pdb) pwndbg.vmmap.get()[0].objfile
'/home/dc/pwndbg_bug/a.out'
```

6. Because we returned `None`, the `spec = "*%#x" % (addr)` string formatting for breakrva or `print(hex(addr))` for piebase fails.

### DIFF
--- a/pwndbg/commands/pie.py
+++ b/pwndbg/commands/pie.py
@@ -19,13 +19,23 @@ def translate_addr(offset, module):
     pages = list(filter(mod_filter, pwndbg.vmmap.get()))
 
     if not pages:
-        print('There are no mappings for specified address or module.')
+        print('There are no memory pages in `vmmap` '
+              'for specified address=0x%x and module=%s' % (offset, module))
         return
 
-    return min(map(lambda page: page.vaddr, pages)) + offset
+    first_page = min(pages, key=lambda page: page.vaddr)
 
-def get_exe_name():
-    return pwndbg.auxv.get().get('AT_EXECFN', pwndbg.proc.exe)
+    addr = first_page.vaddr + offset
+
+    if not any(addr in p for p in pages):
+        print('Offset 0x%x rebased to module %s as 0x%x is beyond module\'s '
+              'memory pages:' % (offset, module, addr))
+        for p in pages:
+            print(p)
+        return
+
+    return addr
+
 
 parser = argparse.ArgumentParser()
 parser.description = 'Calculate VA of RVA from PIE base.'
@@ -39,9 +49,12 @@ parser.add_argument('module', type=str, nargs='?', default='',
 def piebase(offset=None, module=None):
     offset = int(offset)
     if not module:
-        module = get_exe_name()
+        module = pwndbg.proc.exe
+
     addr = translate_addr(offset, module)
-    print(hex(addr))
+
+    if addr is not None:
+        print('Calculated VA from %s = 0x%x' % (module, addr))
 
 
 parser = argparse.ArgumentParser()
@@ -56,10 +69,12 @@ parser.add_argument('module', type=str, nargs='?', default='',
 def breakrva(offset=None, module=None):
     offset = int(offset)
     if not module:
-        module = get_exe_name()
+        module = pwndbg.proc.exe
     addr = translate_addr(offset, module)
-    spec = "*%#x" % (addr)
-    gdb.Breakpoint(spec)
+
+    if addr is not None:
+        spec = "*%#x" % (addr)
+        gdb.Breakpoint(spec)
 
 
 @pwndbg.commands.QuietSloppyParsedCommand


disconnect3d:
@ecx86 could you review that and check if everything works for you, please? :) :beers:

rcx:
lgtm

disconnect3d:
@ecx86 Thanks!

rcx:
Actually, on further inspection, this breaks it for me:
```
gdbserver tcp:1234 /bin/cat
```

```
pwndbg> target remote 1.3.3.7:1234
pwndbg> brva 0x49e6
There are no memory pages in `vmmap` for specified address=0x49e6 and module=target:/bin/cat
pwndbg> vmmap
    0x555555554000     0x55555555c000 r-xp     8000 0      /bin/cat
    0x55555575b000     0x55555575d000 rw-p     2000 7000   /bin/cat
    0x7ffff7dd9000     0x7ffff7dfc000 r-xp    23000 0      /lib/x86_64-linux-gnu/ld-2.24.so
    0x7ffff7ff8000     0x7ffff7ffa000 r-xp     2000 0      [vdso]
    0x7ffff7ffa000     0x7ffff7ffc000 r--p     2000 0      [vvar]
    0x7ffff7ffc000     0x7ffff7ffe000 rw-p     2000 23000  /lib/x86_64-linux-gnu/ld-2.24.so
    0x7ffff7ffe000     0x7ffff7fff000 rw-p     1000 0      
    0x7ffffffde000     0x7ffffffff000 rw-p    21000 0      [stack]
0xffffffffff600000 0xffffffffff601000 r-xp     1000 0      [vsyscall]
pwndbg> piebase
There are no memory pages in `vmmap` for specified address=0x0 and module=target:/bin/cat
pwndbg> 
```
:\

disconnect3d:
@ecx86 thx for finding this. I was worried that my change of `get_exe_name` to `pwndbg.proc.exe` might be wrong.
I am going to check how does those values look like.

PR #501, opened 2018-07-11
Inform about `exception-debugger` on exceptions

disconnect3d:
Instead of hiding this feature just for devs who reads our dev guide or just knows that this exists lets make pwndbg development great again and show this command to the world!


sven337:
Sure sounds good

PR #502, opened 2018-07-12
Fixes piebase and breakrva on remote debugging

disconnect3d:
Fixes an issue not fully fixed in previous commit which was caught by @ecx86 in:
https://github.com/pwndbg/pwndbg/pull/500#issuecomment-404332482

The commands broke when we debugged a remote target which was
hosted on a remote gdbserver (NOT a local one).

This is because we used `pwndbg.proc.exe` (changed in previous commit)
which is a local path to the binary which was then used to filter out
memory pages belonging to the binary.

To fix the issue, the AUXV's AT_EXECFN is used first which was used
before previous commit but the returned path is now normalized
(as in previous version it didn't work because if it returned path './a.out'
it couldn't match it with binary's Page.objfile which was e.g. '/blabla/a.out').

### DIFF
--- a/pwndbg/commands/pie.py
+++ b/pwndbg/commands/pie.py
@@ -6,12 +6,34 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import argparse
+import os
 
 import gdb
 
 import pwndbg.auxv
 import pwndbg.commands
 import pwndbg.vmmap
+
+
+def get_exe_name():
+    """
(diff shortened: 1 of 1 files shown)


disconnect3d:
@ecx86 can't add you to reviewers, please check if it works for you when you have time! :)
Big thanks for finding this :beers:!

rcx:
I just tried it on your fork and it seems to work fine again. nice

Issue #521, opened 2018-09-02
Display source code filename


disconnect3d:
Sometimes we can deduct source code filename, so we should display it in the banner:
```
─────────[ SOURCE (CODE) ]─────────
```

e.g.
```
pwndbg> info line
Line 174 of "../sysdeps/posix/system.c" starts at address 0x7fe22356d153 <do_system+675> and ends at 0x7fe2235
6d157 <do_system+679>.
```

anthraxx:
Taking into account how long these are or could be it could be a better idea to just display a line containing path/filename before listing the source code?
IMO it may look odd, taking into account that the banner label can be left, center or right aligned.

disconnect3d:
> the banner label can be left, center or right aligned

Totally forgot about this.

Yeah, we should probably display the filename (or path) above/below OR maybe instead of `SOURCE (CODE)`?

PR #522, opened 2018-09-12
Add filepath banner to source code debugging

disconnect3d:
As in https://github.com/pwndbg/pwndbg/issues/521

Example:
image

With super long path:
image

### DIFF
--- a/pwndbg/info.py
+++ b/pwndbg/info.py
@@ -35,3 +35,30 @@ def files():
         return gdb.execute('info files', to_string=True)
     except gdb.error:
         return ''
+
+@pwndbg.memoize.reset_on_stop
+def line():
+    try:
+        return gdb.execute('info line', to_string=True)
+    except gdb.error:
+        return ''
+
+@pwndbg.memoize.reset_on_stop
+def rel_filepath():
+    """
+    Returns relative filepath of currently debugged file.
+
+    For example if `info line` returns:
+    > Line 5 of "./main.c" is at address 0x555555554641 <main> but contains no code.
+
+    This will return just './main.c'
+    """
+    l = line()
+
+    if not l:
+        return ''
+
+    start = l.index('"') + 1
+    end = start + l[start+1:].index('"') + 1
+
+    return l[start:end]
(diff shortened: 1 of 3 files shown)


disconnect3d:
@anthraxx @zachriggle can u guys review this? :)

pwndbg/info.py
```diff
@@ -35,3 +35,30 @@ def files():
...
+
+    This will return just './main.c'
+    """
+    l = line()
+
+    if not l:
+        return ''
+
+    start = l.index('"') + 1
+    end = start + l[start+1:].index('"') + 1
```
anthraxx: we should use `rindex` here as the filename can contain quotes itself (plus then we dont even need [start+1:])

pwndbg/ui.py
```diff
@@ -38,9 +38,28 @@ def check_title_position():
...
+            and the title is too long, it will be truncated
+            on the left side but after the provided index
+    """
+    if make_upper:
+        title = title.upper()
+
     _height, width = get_window_size()
+
+    if trunc_after_idx > 0:
+        size_to_trunc = len(title) - (width - len('-[  ]-'))
```
anthraxx: you can't use `'-[  ]-'` it can be overriden by `config.banner_title_surrounding_left` and `config.banner_title_surrounding_right` like below

anthraxx (changes requested):

anthraxx (approved):

zachriggle:
I don't mind adding the filename / path, but the changes to `banner` seem odd.  I don't generally like using `banner()` for the filename anyway.

Can we prepend the filename to the source listing instead?

disconnect3d:
> Can we prepend the filename to the source listing instead?

Like this?:
```
----[ SOURCE (CODE) ]----
File: some_path.c
1. // the source code
2. #include <whatever>
3.
4. ...
```

I am not very much for the banner changes either, so this solution is fine for me.
@anthraxx What do u think?

disconnect3d:
Closing this; see https://github.com/pwndbg/pwndbg/pull/526

PR #525, opened 2018-09-13
fix probeleak command (message.warning -> warn) and small fixes for it

GrosQuildu:


pwndbg/commands/probeleak.py
```diff
@@ -66,7 +63,7 @@ def probeleak(address=None, count=0x40):
         return
 
     found = False
-    for i in range(0, count, ptrsize):
+    for i in range(0, count - ptrsize + 1):
```
GrosQuildu: In case of misaligned address.

pwndbg/commands/probeleak.py
```diff
@@ -19,16 +19,13 @@
...
     pages = list(filter(mod_filter, pwndbg.vmmap.get()))
 
     if not pages:
         return None
 
-    if len(pages) > 1:
-        print(message.warn('Warning: There is more than one page containing address 0x%x (wtf?)', addr))
-
-    return pages[0]
+    return pages[-1]
```
GrosQuildu: With "<=" we can find ptrs to the end of the page, which may be missed in previous version when some page doesn't have adjacent neighbor

disconnect3d: Not sure I get it, can you give an example?
Afaik if you have a page from address A to B that means it occupies addresses [A, B) - not including B. So I guess the `<=` change above is wrong?

GrosQuildu: Yes, page is in [A,B), but if we leak pointer to the end of the page, so to the B, we should show it in the probeleak anyway.
And pages[-1] because when two pages are aligned it is better to show the second one.

disconnect3d:
Hm.. can we also handle this case, when we `probeleak` some address BEFORE end of memory page?:
```
pwndbg> probeleak 0x7ffffffff000-0x1
  File "/home/dc/projects/pwndbg/pwndbg/commands/__init__.py", line 109, in __call__
  File "/home/dc/projects/pwndbg/pwndbg/commands/__init__.py", line 200, in _OnlyWhenRunning
    return function(*a, **kw)
  File "/home/dc/projects/pwndbg/pwndbg/commands/probeleak.py", line 67, in probeleak
  File "/home/dc/projects/pwndbg/pwndbg/arch.py", line 81, in unpack
    return struct.unpack(fmt, data)[0]
struct.error: unpack requires a buffer of 8 bytes
```

GrosQuildu:
Fixed unpack error
Added arg max_distance (from page boundary to potential pointer), test with:
```
gdb /bin/echo
start
set *(uint64_t*)0x55555575b000 = 0x7ffffffff000
set *(uint64_t*)0x55555575b008 = 0x7ffffffff000+0x41
set *(uint64_t*)0x55555575b011 = 0x7ffffffdb000-0x31
set *(uint64_t*)0x55555575b020 = 0x7ffffffdb000+0x31
set *(uint64_t*)0x55555575b028 = 0x7ffff7ffc000+0x21

probeleak 0x55555575b000 48 0x42
x/10xg 0x55555575b000
```

disconnect3d:
Hmm.. can we also improve UX a bit more:
* extend command description with example usages (as done in https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/commands/setregs.py#L17-L19 - btw that command is going to be removed as one can do just `set $rax=$rbx=$rcx=value`)
* improve a bit ux by changing this:
```
pwndbg> probeleak 64
Couldn't read memory at 0x40
```

To display: `Couldn't read memory at 0x40. See 'probeleak -h' for the usage.`.

As an example usage we can say sth like `probeleak $rsp 64 - probeleaks 64 bytes starting at the stack pointer`.

Can you give some example about max distance? (i can't really run ur example due to address mismatch, u should use e.g. `$rsp` ;/).

GrosQuildu:
fixed. for tests above try:
```
gdb /bin/echo
start
set $stack_start=0x7ffffffdb000
set $stack_end=0x7ffffffff000
set $some_writable=0x55555575a000

set *(uint64_t*)($some_writable) = $stack_end
set *(uint64_t*)($some_writable+0x8) = $stack_end+0x41
set *(uint64_t*)($some_writable+0x11) = $stack_start-0x31
set *(uint64_t*)($some_writable+0x20) = $stack_start+0x31

pwndbg> probeleak $some_writable 48 0x42
0x00: 0x00007ffffffff000 = (rw-p) [stack] + 0x24000 + 0x0 (outside of the page)
0x08: 0x00007ffffffff041 = (rw-p) [stack] + 0x24000 + 0x41 (outside of the page)
0x11: 0x00007ffffffdafcf = (rw-p) [stack] - 0x31 (outside of the page)
0x20: 0x00007ffffffdb031 = (rw-p) [stack] + 0x31
pwndbg> x/6xg $some_writable
0x55555575a000:	0x00007ffffffff000	0x00007ffffffff041
0x55555575a010:	0x007ffffffdafcfe8	0xffffdae000000700
0x55555575a020:	0x00007ffffffdb031	0xffffdbb000000730
```

disconnect3d:
Thanks! o/

Issue #520, opened 2018-09-02
Set regs command


disconnect3d:
This is horrible:
```
set $rax=0xffffff
set $rbx=0xffffff
set $rcx=0xffffff
set $rdx=0xffffff
```

I want to be able to do `setregs rax,rbx,rcx,rdx 0xffff`

PR #527, opened 2018-09-16
Add setregs command #520

disconnect3d:


zachriggle:
I'm not sure how this is better than the default commands, but I suppose it doesn't hurt anything.

```
pwndbg> set $rax=$rbx=$rcx=0x1337
...
*RAX  0x1337
*RBX  0x1337
*RCX  0x1337
```

```
pwndbg> set $r8=$r9=$r10=$r11=$r12=$r13=$rsp+8
```

disconnect3d:
_feels embarassed for not checking this one_.

I guess I will just revert this patch... :P

PR #530, opened 2018-09-20
GOT cmd fix

### LINKED ISSUE #517: GOT command returns wrong addresses
Running on latest dev version

Issue found running binary with PIE enabled, ASLR disabled, NO RELRO
Running the `got` command returns wrong addresses for the GOT.
I think it may have something to do with there only being two mapped regions for the binary, rather than 3 as usual. 

Example from below: The real `settbuf@GOT` address is `0x56559cb4`. `got` command says it is `0x5655acb4`. A 0x1000 (exact size of memory page) offset

Virtual memory mapping (while running)
image

Running the GOT command whilst the program is running causes an error as the heap isn't initialized yet.
image

Running GOT command returns addresses in the heap data region. (this was taken after heap was initialised)

image

Memory mapping after heap initialised. The GOT addresses above point into the heap.
image

GrosQuildu:
fixes #517 
got addresses were calculated (on PIE binaries only) as bin_text_base + offset, where bin_text_base was entrypoint aligned down to 0x1000
So when entrypoint was further than 0x1000 from real base addr, the command was wrong

Also small fix in memory.find_lower_boundary


disconnect3d:
Will look at it later today.

One advice: it is easier to develop things on a separate branch, so then it is possible to smoothly pull base branch from the upstream (git will just use 'fast forward' technique to merge things to your local revision).

In your case you have your own `dev` branch on which you develop things: this way now we can see multiple commits in the log, even those that corresponds to other features or just `merge branch dev of <upstream>`.

This is not that bad/wrong - i.e. I can hit `squash and merge` which will take all the commits, merge them into one, and merge to dev.
There are also two other possibilities: `create merge commit` which is something we don't use in this repo and `rebase and merge`. We use only `squash and merge` and `rebase and merge` to ensure the history of a main branch won't end up as a spaghetti that's hard to follow (e.g. merge from dev to X, then commits at X, then merge X to dev).

Btw the `rebase and merge` would take all of those commits and try to apply them to the top of base branch (here: `dev`). This is probably the best option to choose in most cases **assuming that the commits are clear and can be applied to the top of the branch** (so e.g. the commits are atomic, well described in the git commit message and finally - rebased to the base branch).

To sum up, I recommend you using separate branches for developing stuff - that way you won't have to make a hard reset (`git checkout dev && git pull origin && git reset --hard origin/dev`) or some ugly merges to keep up with the latest remote version.

If you want to try it now, you can do those steps (also read below for easier option first);
* `git checkout -B fix-got-command` (create new branch and checkout it)
* `git checkout dev` (go back to dev)
* `git pull origin && git reset --hard origin/dev` (reset your local dev to the one from remote, not carying about merges etc)
* `git checkout fix-got-command` (go back to your new branch)
And now the hard part:
* `git rebase -i dev` - this will launch an `interactive rebase`. A rebase lets you modify git commit history (**so never do it on a version that is already pushed to the upstream as it may break others versions**). The command will launch an editor and will display the commits between your branch and the branch you want to rebase onto (you can think of rebase as "I want to put my commits on the top of the branch"). You can then choose what you want to do with given commits: pick (just use the commit), edit/reword them, squash (merge with previous commit), fixup (like squash, but discard the commit message), delete (just remove the commit line) or even reorder (just move a commit before another commit). All of this should be also documented in the interactive rebase editor. After you save the file the rebase will happen and git will try to put the commits you choosen to the top of the branch you rebase onto. Git may stop at some point and tell you: hey, I don't know how to rebase this as there are some conflicts which needs to be resolved, please solve them, `git add <conflicted files>` and run `git rebase --continue`; otherwise, if you want to abort, just run `git rebase --abort`.

Another, probably much easier option to resolve this is to start from the base branch - upstream `dev` and cherry-pick the commits you want to use. This can be done as:
* `git checkout -B backup-branch` (create new branch, checkout it) - this is our backup, it will point to your current dev branch, so if you break something you can always rewind everything by doing `git reset --hard backup-branch` ("go back to the version from backup-branch")
* `git checkout dev && git pull origin && git reset --hard origin/dev` (resets your local dev to the upstream one)
* `git checkout -B fix-got-command` (create new branch with the same version as upstream dev)
* find the commits you want to pick to the new branch - this can be done e.g. by `git log backup-branch` or probably nicer with `git --no-pager hl -$((`tput lines`/2)) backup-branch` (uhm, no, I don't remember this command, I have it aliased to `git ll` in my `~/.gitconfig`) - save their git commit ids
* `git cherry-pick <commit-id>` to pick certain commits

Also since I already written so much, one more cool thing: `git reflog` shows you a history of all operations in your local git repo. With that you can find e.g. a commit id that is not referenced by any branch and... you can e.g. cherry pick or come back to it anytime (with `git reset --hard <commit id>` - in case you want to revert something).

Issue #414, opened 2018-02-01
Pwndbg Bugreport


zachriggle:
It would be useful if Pwndbg could generate a bugreport on its own.  This would give users something nice to copy-paste into the bug-report.  
Included information might include:

- Pwndbg commit or release
  - Since installation requires `git`, we should be able to `git -C log -1 --oneline`
  - Find install directory via `os.path.dirname(pwndbg.__file__)`
- GDB version
  - Get path via `os.readlink('/proc/self/exe')`
  - Invoke with `-version` to get the full build info
  - Output of `show configuration`
- System info via `platform` module
  - `platform.system()`
  - `platform.platform()`
  - `platform.python_version()`
  - `platform.python_compiler()`
- ZIP file containing the file under test (optional for the user to upload)
  - `open('/proc/self/cmdline').read().split('\x00')`
  - `info file`
  - `info sharedlib`
- GDB command history
  - This would require saving the contents of the history file, and diffing from the session start
- Exception / backtrace information
  - `sys.last_value` / `sys.last_exception`
  - However, they're pretty unreliable if e.g. there's lots of exceptions

We could even try to be really nice and open the browser to the **New Issue** page (`xdg-open`, `open` for Linux / OSX) -- but the user would still need to copy-paste the template.

We could even use `ISSUE_TEMPLATE.md` to generate the copy-paste-able information.

disconnect3d:
@GrosQuildu can you extend the `bugreport` command with the debugged file? I think this should also be optional with e.g. `--attach-debugged-file` flag.

gsingh93:
Is there any way to get the output of the last command before uploading? Or would we have to save the output of the last command ourselves if we wanted to do this?

gsingh93:
I think we can close this, the command has pretty much everything we need now except `--attach-debugged-file`, but I'm not sure if there's an easy way to do that. If there is some way, we can create a new issue for it.

PR #533, opened 2018-09-25
add bugreport cmd

GrosQuildu:
for #414 
Not sure we need it since there is version() command (this adds session history, platform and optional browser opening)

### DIFF
--- a/pwndbg/commands/version.py
+++ b/pwndbg/commands/version.py
@@ -14,6 +14,10 @@ import os
 import sys
 from platform import platform
 from subprocess import check_output
+try:
+    from urllib import quote
+except:
+    from urllib.parse import quote
 
 import gdb
 
@@ -170,13 +174,16 @@ If it is somehow unavailable, use:
     gdb_current_session_history = (v for (k, v) in sorted(gdb_current_session_history.items()))
     gdb_current_session_history = '\n'.join(gdb_current_session_history)
     
-    print(ISSUE_TEMPLATE.format(gdb_history=gdb_current_session_history, setup=current_setup))
+    issue_bugreport = ISSUE_TEMPLATE.format(gdb_history=gdb_current_session_history, setup=current_setup)
+    print(issue_bugreport)
 
     please_please_submit = 'Please submit the bugreport generated above at '
     github_issue_url = 'https://github.com/pwndbg/pwndbg/issues/new'
+    github_issue_body = '?body=' + quote(issue_bugreport)
+
     if browse:
         try:
-            check_output(['xdg-open', github_issue_url])
+            check_output(['xdg-open', github_issue_url + github_issue_body])
         except:
             print(please_please_submit + github_issue_url)    
     else:
--- a/pwndbg/commands/version.py
+++ b/pwndbg/commands/version.py
@@ -14,10 +14,6 @@ import os
 import sys
 from platform import platform
 from subprocess import check_output
-try:
-    from urllib import quote
-except:
-    from urllib.parse import quote
 
 import gdb
 
@@ -25,6 +21,11 @@ import pwndbg
 import pwndbg.commands
 import pwndbg.ida
 from pwndbg.color import message
+
+try:
+    from urllib import quote
+except:
+    from urllib.parse import quote
 
 
 def _gdb_version():
--- a/pwndbg/commands/version.py
+++ b/pwndbg/commands/version.py
@@ -10,6 +10,10 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import sys
+import os
+from subprocess import check_output
+from platform import platform
+import argparse
 
 import gdb
 
@@ -46,11 +50,7 @@ def unicorn_version():
         return 'not found'
 
 
-@pwndbg.commands.Command
-def version():
-    """
-    Displays gdb, python and pwndbg versions.
-    """
+def all_versions():
     gdb_str      = 'Gdb:      %s' % _gdb_version()
     py_str       = 'Python:   %s' % _py_version()
     pwndbg_str   = 'Pwndbg:   %s' % pwndbg.__version__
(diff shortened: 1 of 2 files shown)


disconnect3d:
Can you show it in action? e.g. with [asciinema](https://asciinema.org/)

GrosQuildu:
https://asciinema.org/a/8rAtszQaH80Zzx3KTIm55is1j

travis failed, is it supposed to fail? :D

pwndbg/commands/version.py
```diff
@@ -67,5 +67,116 @@ def version():
...
+    print(ISSUE_TEMPLATE.format(gdb_history=gdb_current_session_history, setup=current_setup))
+
+    github_issue_url = 'https://github.com/pwndbg/pwndbg/issues/new'
+    if browse:
+        try:
+            check_output(['xdg-open', github_issue_url])
+        except:
+            print('Please submit at ' + github_issue_url)    
+    else:
+        print('Please submit at ' + github_issue_url)
```
anthraxx: there should be a newline at the end of the file, PEP doesnt like this

pwndbg/commands/version.py
```diff
@@ -10,6 +10,10 @@
 from __future__ import unicode_literals
 
 import sys
+import os
+from subprocess import check_output
+from platform import platform
+import argparse
```
anthraxx: you need to properly reorder the imports to make `isort` happy, just install it and call it to see what the problem is :)

anthraxx:
awesome, its green :+1:

pwndbg/commands/version.py
```diff
@@ -67,5 +67,116 @@ def version():
...
+
+If it is somehow unavailable, use:
+* `show version` - for gdb
+* `py import sys; print(sys.version)` - for python
+* pwndbg version/git commit id
+-->
+
+{setup}'''
+
+    gdb_config = gdb.execute('show configuration', to_string=True).split('\n')[:-3]
```
disconnect3d: Example output:
```
pwndbg> show configuration 
This GDB was configured as follows:
   configure --host=x86_64-linux-gnu --target=x86_64-linux-gnu
             --with-auto-load-dir=$debugdir:$datadir/auto-load
             --with-auto-load-safe-path=$debugdir:$datadir/auto-load
             --with-expat
             --with-gdb-datadir=/usr/share/gdb (relocatable)
             --with-jit-reader-dir=/usr/lib/gdb (relocatable)
             --without-libunwind-ia64
             --with-lzma
             --with-python=/usr (relocatable)
             --without-guile
             --with-separate-debug-dir=/usr/lib/debug (relocatable)
             --with-system-gdbinit=/etc/gdb/gdbinit
             --with-babeltrace

("Relocatable" means the directory can be moved with the GDB installation
tree, and GDB will still find it.)
```

I am a bit worried that there might be cases we would drop some interesting lines in here because of `[:-3]` :(.
Maybe we should leave it as it is?

pwndbg/commands/version.py
```diff
@@ -67,5 +67,116 @@ def version():
...
+    all_info = all_versions()
+
+    current_setup = 'Platform: %s\n' % platform()
+    current_setup += '\n'.join(all_info)
+    current_setup += '\n'.join(gdb_config)
+
+    # get saved history size (not including current gdb session)
+    gdb_history_file = gdb.execute('show history filename', to_string=True)
+    gdb_history_file = gdb_history_file[len('The filename in which to record the command history is "'):-3]
```
disconnect3d:
```
pwndbg> show history filename 
The filename in which to record the command history is "/home/dc/.gdb_history".
```
Some of the users might use different language and it  might crash for them...
Maybe we should just use `.index('"')` and `.rindex('"')`?

---

Offtopic, gdb doesn't escape quotations in here:
```
pwndbg> set history filename /home/dc/"/.gdb_history
pwndbg> show history filename 
The filename in which to record the command history is "/home/dc/"/.gdb_history".
```

Via irc:
```
09:45 <disconnect3d> pwndbg> set history filename /home/dc/"/.gdb_history
09:45 <disconnect3d> pwndbg> show history filename 
09:45 <disconnect3d> The filename in which to record the command history is "/home/dc/"/.gdb_history".
09:45 <disconnect3d> the path is not escaped, heh
09:56 <PhilippeW> disconnect3d: yes, the handling of arguments quoting/escaping by gdb is not consistent
09:56 <PhilippeW> See a.o. a discussion in https://sourceware.org/ml/gdb-patches/2018-09/msg00804.html
```
https://sourceware.org/ml/gdb-patches/2018-09/msg00804.html

pwndbg/commands/version.py
```diff
@@ -67,5 +67,116 @@ def version():
...
+    gdb_history_len = 0
+    try:
+        with open(gdb_history_file, 'r') as f:
+            gdb_history_len = len(f.readlines())
+    except FileNotFoundError:
+        pass
+
+    max_command_no = int(gdb.execute('show commands', to_string=True).split('\n')[-2].split('  ')[1]) - 1
```
disconnect3d: We shouldn't rely on whitespace width. I guess this might fail if we had like 1000+ entries in commands history:
```python
gdb.execute('show commands', to_string=True).split('\n')[-2].split('  ')
```

This can be solved by not providing any argument to `.split`, so it will split on whitespaces no matter of their length (and remove them from output), as stated in the docs:
```
In [1]: str.split?
Docstring:
S.split([sep [,maxsplit]]) -> list of strings

Return a list of the words in the string S, using sep as the
delimiter string.  If maxsplit is given, at most maxsplit
splits are done. If sep is not specified or is None, any
whitespace string is a separator and empty strings are removed
from the result.
```

pwndbg/commands/version.py
```diff
@@ -67,5 +67,116 @@ def version():
...
+    show_command_size = 10  # 'show command' returns 10 commands
+    gdb_current_session_history = {}
+    current_command_no = gdb_history_len + 1
+
+    while current_command_no <= max_command_no:
+        cmds = gdb.execute('show commands ' + str(current_command_no + (show_command_size//2)+1), to_string=True).split('\n')[:-1]
+        for cmd in cmds:
+            cmd_no = int(cmd.split('  ')[1])
```
disconnect3d: Similar problem as above. We can also use `maxsplit` optional argument and get both the integer and command as:

```python
cmd_no, cmd = cmd.split(maxsplit=1)
cmd_no = int(cmd_no)
```

As here:
```python
>>> x = gdb.execute('show commands', to_string=True).split('\n')[-2].split(maxsplit=1)
>>> x
['84', "x = gdb.execute('show commands', to_string=True).split('\\n')[-2].split(maxsplit=1)"]
```

pwndbg/commands/version.py
```diff
@@ -67,5 +67,116 @@ def version():
...
+            if cmd_no <= gdb_history_len:
+                continue
+            if current_command_no > max_command_no:
+                break
+            cmd = '  '.join(cmd.split('  ')[2:])
+            gdb_current_session_history[cmd_no] = cmd
+            current_command_no += 1
+
+    gdb_current_session_history = [v for (k, v) in sorted(gdb_current_session_history.items())]
```
disconnect3d: Totally doesn't matter, but as an optimization this can be a generator expression instead of list comprehension i.e.:
```python
[v for (k, v) in sorted(gdb_current_session_history.items())]
```

can be chanaged to:
```python
(v for (k, v) in sorted(gdb_current_session_history.items()))
```

pwndbg/commands/version.py
disconnect3d: Maybe: `Please submit the bugreport generated above at ...`?

disconnect3d:
Can we also add info about that `bugreport` command to the exception handler, i.e. that he can fire one command and a browser with filled bugreport will spawn for them?

disconnect3d (changes requested):
See comments

pwndbg/commands/version.py
```diff
@@ -67,5 +67,116 @@ def version():
...
+-->
+
+{setup}'''
+
+    gdb_config = gdb.execute('show configuration', to_string=True).split('\n')[:-3]
+    all_info = all_versions()
+
+    current_setup = 'Platform: %s\n' % platform()
+    current_setup += '\n'.join(all_info)
+    current_setup += '\n'.join(gdb_config)
```
disconnect3d: There is such line in asciinema:
```
Unicorn:  1.0.0This GDB was configured as follows:  
```
maybe we should add another newline here?

rcx:
I'm not sure about opening the browser. 99% of the time I use gdb, it's on a headless box over SSH.

GrosQuildu:
Noidea, whatever you propose. But it's optional only.

disconnect3d:
Hmm.. maybe lets change the flag from `--browse` to `--run-browser` so it will be more explicit that it is optional.

Issue #519, opened 2018-09-01
Add _chk functions


disconnect3d:
This should show a bit different arg names:

```
 ► 0x400855    call   __printf_chk@plt <0x400708>
        format: 0x1
        vararg: 0x400abc ◂— outsb  dx, byte ptr [rsi] /* 'Input offset: ' */
```

Functions to support:
* [x] `__printf_chk` (https://github.com/pwndbg/pwndbg/pull/528)
* [ ] `ts_chk`
* [x] `poll_chk`
* [x] `read_chk` (#536)
* [x] `recv_chk`
* [x] `fdelt_chk`
* [x] `fgets_chk` 
* [x] `fread_chk` (#536)
* [ ] `twd_chk`
* [x] `ppoll_chk`
* [x] `pread_chk` (#536)
* [x] `fgetws_chk`
* [ ] `tcwd_chk`
* [x] `memcpy_chk`
* [x] `memset_chk`
* [ ] `printf_chk`
* [x] `stpcpy_chk`
* [x] `strcat_chk`
* [x] `strcpy_chk`
* [x] `wcpcpy_chk`
* [x] `wcscat_chk`
* [x] `wcscpy_chk`
* [x] `wctomb_chk`
* [x] `confstr_chk`
* [ ] `printf_chk`
* [x] `fprintf_chk`
* [ ] `longjmp_chk`
* [x] `memmove_chk`
* [x] `mempcpy_chk`
* [x] `pread64_chk`
* [x] `sprintf_chk`
* [x] `stpncpy_chk`
* [x] `strncat_chk`
* [x] `strncpy_chk`
* [x] `vprintf_chk`
* [x] `wcpncpy_chk`
* [x] `wcrtomb_chk`
* [x] `wcsncat_chk`
* [x] `wcsncpy_chk`
* [x] `wmemcpy_chk`
* [x] `wmemset_chk`
* [x] `wprintf_chk`
* [x] `sysdeps/i386/memcpy_chk.S`
* [x] `sysdeps/i386/memset_chk.S`
* [x] `sysdeps/m68k/wcpcpy_chk`
* [x] `asprintf_chk`
* [x] `fwprintf_chk`
* [x] `mbstowcs_chk`
* [ ] `obprintf_chk`
* [x] `readlink_chk`
* [x] `realpath_chk`

disconnect3d:
Some of them can be inspected e.g. here https://github.com/lattera/glibc/find/master (click `t` and search for `_chk`).

[...]

PR #536, opened 2018-10-03
Add __read_chk, __fread_chk and __pread_chk

F3real:
Adds some of things from #519 
References:
[__read_chk](https://github.com/lattera/glibc/blob/master/debug/read_chk.c)
[__fread_chk](https://github.com/lattera/glibc/blob/master/debug/fread_chk.c)
[__pread_chk](https://github.com/lattera/glibc/blob/master/debug/pread_chk.c)


disconnect3d:
Hey,

Thanks for the PR!

Anyway I think we should focus on extending functions repo as asked here https://github.com/pwndbg/pwndbg/pull/528#issuecomment-422644418 and then take regenerated 
 `functions.py` from there.

I've started working on that here https://github.com/disconnect3d/functions/tree/improve-readme-and-add-chk-functions but I haven't end it yet. I have fetched glibc's `debug/*_chk.c` files, concatenated them and produced a header file with all of them, removing includes that are not present. However, glibc is LGPL so I think I will remove the `*_chk.c` files from the repo, requrie user to fetch glibc on their own and retrieve only function headers from there instead of full code.

If you have time and are willing to help, you are welcome o/

Issue #168, opened 2017-02-28
QEMU Kernel Memory Mappings


zachriggle:
On face value, it appears that [QEMU exposes the monitor interface over the gdbstub](https://github.com/qemu/qemu/blob/master/gdbstub.c#L1265-L1280):

```c
#else /* !CONFIG_USER_ONLY */
        else if (strncmp(p, "Rcmd,", 5) == 0) {
            int len = strlen(p + 5);

            if ((len % 2) != 0) {
                put_packet(s, "E01");
                break;
            }
            len = len / 2;
            hextomem(mem_buf, p + 5, len);
            mem_buf[len++] = 0;
            qemu_chr_be_write(s->mon_chr, mem_buf, len);
            put_packet(s, "OK");
            break;
        }
#endif /* !CONFIG_USER_ONLY */
```

This means that it should be possible to send `QRcmd` commands, e.g. `info mtree` which dumps memory maps.

```
pwnbg> monitor info mtree
address-space: memory
  0000000000000000-ffffffffffffffff (prio 0, RW): system
    0000000000000000-0000000007ffffff (prio 0, RW): versatile.ram
    0000000010000000-0000000010000fff (prio 0, RW): arm-sysctl
    0000000010001000-0000000010001fff (prio 0, RW): pci-vpb-regs
    0000000010002000-0000000010002fff (prio 0, RW): versatile_i2c
    0000000010003000-0000000010003fff (prio 0, RW): vpb-sic
    0000000010004000-0000000010004fff (prio 0, RW): pl041
    0000000010005000-0000000010005fff (prio 0, RW): pl181
    0000000010006000-0000000010006fff (prio 0, RW): pl050
    0000000010007000-0000000010007fff (prio 0, RW): pl050
    0000000010009000-0000000010009fff (prio 0, RW): pl011
    000000001000b000-000000001000bfff (prio 0, RW): pl181
    0000000010010000-000000001001000f (prio 0, RW): smc91c111-mmio
    0000000010120000-0000000010120fff (prio 0, RW): pl110
    0000000010130000-0000000010130fff (prio 0, RW): pl080
    0000000010140000-0000000010140fff (prio 0, RW): pl190
    00000000101e2000-00000000101e2fff (prio 0, RW): sp804
    00000000101e3000-00000000101e3fff (prio 0, RW): sp804
    00000000101e4000-00000000101e4fff (prio 0, RW): pl061
[...]

PR #539, opened 2018-10-09
Fixes #538 - breakrva on symlink targets

### LINKED ISSUE #538: brva doesn't work if the target is a symlink.
```
$ ln -s /bin/cat target
$ gdb ./target
pwndbg> r
^C
0xf779a000 0xf779b000 r-xp     1000 0      /tmp/target
0xf779b000 0xf779c000 r--p     1000 0      /tmp/target
0xf779c000 0xf779d000 rw-p     1000 1000  /tmp/target
pwndbg> brva 0x7B0
There are no memory pages in `vmmap` for specified address=0x7b0 and module=/bin/cat
```

disconnect3d:
Fixes a bug with `breakrva` and `brva` commands reported in #538 and adds some more
explanation on how certain things works:
* `info auxv` or to be more specific: AUXV's `ET_EXECFN` holds path to
the executable, but if it is a symlink, it is not dereferenced
* because of that we need to call `readlink` in `get_exe_name` in pie.py

### DIFF
--- a/pwndbg/auxv.py
+++ b/pwndbg/auxv.py
@@ -165,7 +165,7 @@ def walk_stack():
 
     if not auxv.get('AT_EXECFN', None):
         try:
-            auxv['AT_EXECFN'] = get_execfn()
+            auxv['AT_EXECFN'] = _get_execfn()
         except gdb.MemoryError:
             pass
 
@@ -245,7 +245,7 @@ def walk_stack2(offset=0):
 
     return auxv
 
-def get_execfn():
+def _get_execfn():
     # If the stack is not sane, this won't work
     if not pwndbg.memory.peek(pwndbg.regs.sp):
         return
--- a/pwndbg/commands/pie.py
+++ b/pwndbg/commands/pie.py
@@ -27,11 +27,21 @@ def get_exe_name():
     """
     path = pwndbg.auxv.get().get('AT_EXECFN')
 
-    if path is not None:
+    # When GDB is launched on a file that is a symlink to the target,
+    # the AUXV's AT_EXECFN stores the absolute path of to the symlink.
+    # On the other hand, the vmmap, if taken from /proc/pid/maps will contain
+    # the absolute and real path of the binary (after symlinks).
+    # And so we have to read this path here.
+    real_path = pwndbg.file.readlink(path)
+
+    if real_path == '':  # the `path` was not a symlink
+        real_path = path
+
+    if real_path is not None:
         # We normalize the path as `AT_EXECFN` might contain e.g. './a.out'
         # so matching it against Page.objfile later on will be wrong;
         # We want just 'a.out'
-        return os.path.normpath(path)
+        return os.path.normpath(real_path)
 
     return pwndbg.proc.exe
 
--- a/pwndbg/commands/version.py
+++ b/pwndbg/commands/version.py
@@ -189,5 +189,3 @@ If it is somehow unavailable, use:
             print(please_please_submit + github_issue_url)    
     else:
         print(please_please_submit + github_issue_url)
-
-
--- a/pwndbg/vmmap.py
+++ b/pwndbg/vmmap.py
@@ -48,6 +48,11 @@ def get():
     pages.extend(proc_pid_maps())
 
     if not pages:
+        # If debugee is launched from a symlink the debugee memory maps will be
+        # labeled with symlink path while in normal scenario the /proc/pid/maps
+        # labels debugee memory maps with real path (after symlinks).
+        # This is because the exe path in AUXV (and so `info auxv`) is before
+        # following links.
         pages.extend(info_auxv())
 
         if pages: pages.extend(info_sharedlibrary())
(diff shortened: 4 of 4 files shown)


pwndbg/auxv.py
```diff
@@ -245,7 +245,7 @@ def walk_stack2(offset=0):
 
     return auxv
 
-def get_execfn():
+def _get_execfn():
```
disconnect3d, line 248: I renamed this as this function is internal to the file and should not be used outside.

pwndbg/commands/pie.py
```diff
@@ -27,11 +27,21 @@ def get_exe_name():
     """
     path = pwndbg.auxv.get().get('AT_EXECFN')
 
-    if path is not None:
+    # When GDB is launched on a file that is a symlink to the target,
+    # the AUXV's AT_EXECFN stores the absolute path of to the symlink.
+    # On the other hand, the vmmap, if taken from /proc/pid/maps will contain
+    # the absolute and real path of the binary (after symlinks).
+    # And so we have to read this path here.
+    real_path = pwndbg.file.readlink(path)
```
disconnect3d, line 35: The `pwndbg.file.readlink` should work on local, remote, qemu etc. See https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/file.py#L68-L74

disconnect3d:
TODO / FIXME for me:
* [ ] check if `breakrva`/`brva` works on remote targets
* [ ] check if `breakrva`/`brva` works on (remote) qemu targets

rcx:
Just tested this locally on a symlinked binary, and it works for me.

```
$ ln -s ./target ./test
$ gdb ./test
> starti
> brva 0x8F6
> vmmap
pwndbg> vmmap
0x56630000 0x56631000 r-xp     1000 0      /root/target
pwndbg> info b
Num     Type           Disp Enb Address    What
3       breakpoint     keep y   0x566308f6 in main at target.c:17
> ^D
$ gdb ./target
> starti
> brva 0x8f6
0x565b9000 0x565ba000 r-xp     1000 0      /root/target
> info b
Num     Type           Disp Enb Address    What
1       breakpoint     keep y   0x565b98f6 in main at target.c:17
```

Ideally we should test this on QEMU and remote too, but assuming the `readlink` helper works then this LGTM.

pwndbg/commands/pie.py
zachriggle: We might also just be able to read `/proc/self/exe`, versus using AT_EXECFN.  I think the only reason we don't do that is for QEMU, since it can't emulate what we see.

Perhaps we should check for QEMU first, and just use `/proc/self/exe` otherwise?

zachriggle (changes requested):

disconnect3d:
@zachriggle lol, a test on qemu launched with `qemu-x86_64 -g 4444 ./a.out`.

Without pwndbg:
```
✘ 130 ➜ gdb ./a.out 
(gdb) target remote localhost:4444
Remote debugging using localhost:4444
warning: remote target does not support file transfer, attempting to access files from local filesystem.
Reading symbols from /lib64/ld-linux-x86-64.so.2...Reading symbols from /usr/lib/debug//lib/x86_64-linux-gnu/ld-2.27.so...done.
done.
0x0000004000a04090 in _start () from /lib64/ld-linux-x86-64.so.2
(gdb) b main
Breakpoint 1 at 0x64a
(gdb) c
Continuing.
warning: Probes-based dynamic linker interface failed.
Reverting to original interface.

[Inferior 1 (Remote target) exited normally]
(gdb) 
```

With pwndbg there is no memory page for the binary in `_start` eh...

disconnect3d:
@zachriggle any hint how to debug PIE binaries with qemu? :|

rcx:
The horrible handling of PIE is one of my greatest frustrations with gdb as opposed to windows debuggers.
On GDB 8.1, this is remedied with the `starti` command. What we should do is keep track of all `brva` breakpoints, and upon (re)starting the binary, it clears all the relative breakpoints, executes `starti`, then re-sets the relative breakpoints. I don't know how to implement this, however.

disconnect3d:
@ecx86 That's probably a lot of work. It would be nice to research qemu more and see if we can do sth with rebasing in there. Maybe they rebase binary to the same address each time? [wishful thinking, heh]

rcx:
Rebasing isnt a good option because it makes the address space unrealistic
if PIE and ASLR are on.

disconnect3d:
@ecx86 What do you mean by unrealistic address space? Why?

Oh, I meant, that we need to be able to find out the address the code will reside in.

rcx:
For example, when developing an exploit, if the target system uses PIE and
ASLR, some people prefer to keep PIE and ASLR on during debugging to avoid
making bad assumptions.

disconnect3d:
@ecx86 Oh I totally understand that point. Still, you usually test without ASLR and just make your exploit not assume this. However yeah, there might be situations where u need aslr/pie as you want to check out if your (partial) brute force works.

Anyway, I would rather accept a case where we disable ASLR to make things easier for pwndbg on qemu.. 🤕 

I asked on qemu irc whether it is possible to find out the address that qemu will put the binary on.

disconnect3d:
Note to self or anyone who is willing to help: look through https://github.com/pwndbg/pwndbg/issues/168 - maybe we can get the image base address (after aslr randomization) from there?

disconnect3d:
Merged as it stayed here for so long. Not sure if things could be done better, we may find it out in the future if this or similar things pops out.

PR #542, opened 2018-10-13
Fix Cortex-M registers in bare metal debugging

ttimasdf:
Fix the second issue mentioned in #264, regarding the xSPR register and the thumb mode bit offset.

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -91,7 +91,11 @@ class Command(gdb.Command):
 
         last_line = lines[-1]
         number, command = last_line.split(None, 1)
-        number = int(number)
+        try:
+            number = int(number)
+        except ValueError:
+            # print("show commands error:", lines)
+            return False
 
         # A new command was entered by the user
         if number not in Command.history:
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -131,6 +131,8 @@ def get_regs(*regs):
             continue
 
         value = pwndbg.regs[reg]
+        if not value and reg.lower() == "cpsr":
+            value = pwndbg.regs["xpsr"]
 
         # Make the register stand out
         regname = C.register(reg.ljust(4).upper())
--- a/pwndbg/disasm/__init__.py
+++ b/pwndbg/disasm/__init__.py
@@ -104,7 +104,6 @@ def get_one_instruction(address):
     md   = get_disassembler(address)
     size = VariableInstructionSizeMax.get(pwndbg.arch.current, 4)
     data = pwndbg.memory.read(address, size, partial=True)
-    print("data:", data)
     for ins in md.disasm(bytes(data), address, 1):
         pwndbg.disasm.arch.DisassemblyAssistant.enhance(ins)
         return ins
--- a/pwndbg/disasm/__init__.py
+++ b/pwndbg/disasm/__init__.py
@@ -83,8 +83,16 @@ def get_disassembler(pc):
     extra = None
 
     if pwndbg.arch.current in ('arm', 'aarch64'):
-        extra = {0:CS_MODE_ARM,
-                 0x20:CS_MODE_THUMB}[pwndbg.regs.cpsr & 0x20]
+        if pwndbg.regs.xpsr:
+            extra = {
+                0: CS_MODE_MCLASS,
+                0x1000000: CS_MODE_MCLASS | CS_MODE_THUMB,
+            }[pwndbg.regs.xpsr & 1 << 24]
+        else:
+            extra = {
+                0: CS_MODE_ARM,
+                0x20: CS_MODE_THUMB,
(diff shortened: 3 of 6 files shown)


pwndbg/commands/__init__.py
```diff
@@ -91,7 +91,11 @@ def check_repeated(self, argument, from_tty):
 
         last_line = lines[-1]
         number, command = last_line.split(None, 1)
-        number = int(number)
+        try:
```
disconnect3d, line 94: Is this a gdb 8.2 hack? Is this needed here?

pwndbg/disasm/__init__.py
```diff
@@ -83,8 +83,16 @@ def get_disassembler(pc):
     extra = None
 
     if pwndbg.arch.current in ('arm', 'aarch64'):
-        extra = {0:CS_MODE_ARM,
-                 0x20:CS_MODE_THUMB}[pwndbg.regs.cpsr & 0x20]
+        if pwndbg.regs.xpsr:
```
disconnect3d, line 86: Is this condition correct? If `xpsr` would have a value of `0` we will assume the `cpsr` shall be used.

This should probably be `if pwndbg.regs.xpsr is not None:`?

pwndbg/disasm/__init__.py
```diff
@@ -83,8 +83,16 @@ def get_disassembler(pc):
...
+            extra = {
+                0: CS_MODE_MCLASS,
+                0x1000000: CS_MODE_MCLASS | CS_MODE_THUMB,
```
disconnect3d, line 89: https://github.com/aquynh/capstone/blob/d70e2b286a7466ab8ed6895213acc24e52b8b2fa/bindings/powershell/Capstone/Capstone.psm1#L120 nice

pwndbg/disasm/arm.py
```diff
@@ -46,7 +46,7 @@ def condition(self, instruction):
         if instruction.address != pwndbg.regs.pc:
             return False
 
-        cpsr = pwndbg.regs.cpsr
+        cpsr = pwndbg.regs.cpsr or pwndbg.regs.xpsr
```
disconnect3d, line 49: if `cpsr` is 0 we will use xpsr, is it correct behavior?

pwndbg/emu/emulator.py
```diff
@@ -200,7 +200,16 @@ def get_uc_mode(self):
         mode = 0
 
         if arch in ('arm', 'aarch64'):
-            mode |= {0:U.UC_MODE_ARM,0x20:U.UC_MODE_THUMB}[pwndbg.regs.cpsr & 0x20]
+            if pwndbg.regs.xpsr:
```
disconnect3d, line 203: same issue as before: if it is 0 we will go to else

pwndbg/commands/context.py
```diff
@@ -131,6 +131,8 @@ def get_regs(*regs):
             continue
 
         value = pwndbg.regs[reg]
+        if not value and reg.lower() == "cpsr":
+            value = pwndbg.regs["xpsr"]
```
disconnect3d, line 135: Is this needed here? I don't like that we hardcode a specific register here; can this code be moved elsewhere?

disconnect3d (changes requested):
Hey,

Thanks for the PR! Please see comments.

Also, is there any way to test this out if I don't have a bare metal Cortex-M?
Like some example that we could launch on Qemu?

pwndbg/commands/__init__.py
ttimasdf: Actually when adding this I was hacking around changing this or that wondering what make things work 😅 And it Just Works™
I'll figure out later whether these lines are essential, my device not at hand for the moment.

pwndbg/disasm/__init__.py
ttimasdf: You are right, value of 0 IS a valid cpsr value and should be considered.

pwndbg/disasm/__init__.py
ttimasdf: As for this, I'm not sure the flags should be `CS_MODE_MCLASS | CS_MODE_THUMB` or just `CS_MODE_THUMB`. Not yet look into their codes.

pwndbg/disasm/arm.py
ttimasdf: err, right. same as above.

disconnect3d:
@ttimasdf can you update this?

ttimasdf:
Hi, sorry for untimely responds. I've been busy at work recently so didn't keep update. I'll get start when I have time or maybe weekend :cat2:

disconnect3d:
@ttimasdf bump; any chance to start working on this at some point?

disconnect3d:
@ttimasdf bump

disconnect3d:
@ttimasdf ping, any chance to finalize this?:<

Issue #230, opened 2017-04-18
Command Name Blacklist


zachriggle:
We should implement a blacklist against GDB-defined default commands and keywords, so that we don't accidentally implement a command that overwrites an existing command.

A specific example of one that we overwrite is `start` (though we maintain similar behavior).

ZetaTwo:
With my PR merged this now means that an attempt to override an existing GDB command will throw an exception. There is a whitelist for commands which are allowed to override existing commands which currently contains: "up", "down", "pwd", "search", "start".

disconnect3d:
Done in #543, thanks again @ZetaTwo o/ :beers:

PR #543, opened 2018-10-14
Check if override built-in command

ZetaTwo:
This PR aims to start a discussion on https://github.com/pwndbg/pwndbg/issues/230
This code fetches current gdb command on start-up and stores them in a list in the Command class.

The question is what to do with this list. As an example, I simply print out a warning but this is not really desirable since it prints out warnings for intended behaviour and is not actionable for the end-user. Another way to go would be to have a whitelist of the currently overridden commands: up, down, search, pwd and start and abort/exit if you try to override any other.

Thoughts?

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -25,7 +25,7 @@ commands = []
 
 def list_current_commands():
     current_pagination = gdb.execute('show pagination', to_string=True)
-    current_pagination = current_pagination.split()[-1][:-1] # Take last word and skip period
+    current_pagination = current_pagination.split()[-1].rstrip('.')  # Take last word and skip period
 
     gdb.execute('set pagination off')
     command_list = gdb.execute('help all', to_string=True).strip().split('\n')
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -24,25 +24,28 @@ import pwndbg.ui
 commands = []
 
 def list_current_commands():
+    current_pagination = gdb.execute('show pagination', to_string=True)
+    current_pagination = current_pagination.split()[-1][:-1] # Take last word and skip period
+
     gdb.execute('set pagination off')
-    command_list = gdb.execute('help all', False, True).strip().split('\n')
+    command_list = gdb.execute('help all', to_string=True).strip().split('\n')
     existing_commands = set()
     for line in command_list:
         line = line.strip()
         # Skip non-command entries
         if len(line) == 0 or line.startswith('Command class:') or line.startswith('Unclassified commands'):
             continue
-        command = line.split(' ')[0]
+        command = line.split()[0]
         existing_commands.add(command)
-    return list(existing_commands)
+    gdb.execute('set pagination %s' % current_pagination) # Restore original setting
+    return existing_commands
 
-builtin_commands = list_current_commands()
+GDB_BUILTIN_COMMANDS = list_current_commands()
 
 class Command(gdb.Command):
     """Generic command wrapper"""
     command_names = set()
-    builtin_commands = builtin_commands
-    builtin_override_whitelist = ['up', 'down', 'search', 'pwd', 'start']
+    builtin_override_whitelist = {'up', 'down', 'search', 'pwd', 'start'}
     history = {}
 
     def __init__(self, function, prefix=False):
@@ -53,7 +56,7 @@ class Command(gdb.Command):
 
         if command_name in self.command_names:
             raise Exception('Cannot add command %s: already exists.' % command_name)
-        if command_name in self.builtin_commands and command_name not in self.builtin_override_whitelist:
+        if command_name in GDB_BUILTIN_COMMANDS and command_name not in self.builtin_override_whitelist:
             raise Exception('Cannot override non-whitelisted built-in command "%s"' % command_name)
 
         self.command_names.add(command_name)
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -42,6 +42,7 @@ class Command(gdb.Command):
     """Generic command wrapper"""
     command_names = set()
     builtin_commands = builtin_commands
+    builtin_override_whitelist = ['up', 'down', 'search', 'pwd', 'start']
     history = {}
 
     def __init__(self, function, prefix=False):
@@ -52,8 +53,8 @@ class Command(gdb.Command):
 
         if command_name in self.command_names:
             raise Exception('Cannot add command %s: already exists.' % command_name)
-        if command_name in self.builtin_commands:
-            print(pwndbg.color.message.warn('Pwndbg overrides builtin command "%s"' % command_name))
+        if command_name in self.builtin_commands and command_name not in self.builtin_override_whitelist:
+            raise Exception('Cannot override non-whitelisted built-in command "%s"' % command_name)
 
         self.command_names.add(command_name)
         commands.append(self)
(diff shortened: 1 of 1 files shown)


disconnect3d:
Hey, thanks for the PR!

I haven't yet looked into the code (will probably do it tomorrow) but I am up for a whitelist of commands that we accept to override. If something doesn't fill into the whitelist, I think we could just fail (i.e. raise an exception or even exit gdb?).

ZetaTwo:
Great, I updated the PR with the described behaviour. Any attempts to override anything other than the currently overridden commands: up, down, search, pwd and start will raise an exception. This is done with a hard-coded whitelist in the command class. If you prefer a more sophisticated way of doing it, please let me know.

ZetaTwo:
The TravisCI check fails due to unrelated whitespace issues fixed separately in https://github.com/pwndbg/pwndbg/pull/544

pwndbg/commands/__init__.py
```diff
@@ -23,10 +23,26 @@
 
 commands = []
 
+def list_current_commands():
+    gdb.execute('set pagination off')
```
disconnect3d, line 30: In theory this is already handled by https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/__init__.py#L130

But if the user changes it afterwards it is not `off` in here.

Still, if the user changes it, they might want to keep their settings as they were, so we should be careful when setting things outside of initialization.

We should either:
* extend GDB's Python api so it is possible to get `help all` through it (long term solution; harder to implement)
* have our own `with statement` that can set given setting and unset it (revert to previous value) afterwards

If you don't want to take care of making a general `with statement` like this, we could skip this here and add a docstring that this method requires `set pagination off` but it is already set while initializing pwndbg.

pwndbg/commands/__init__.py
```diff
@@ -23,10 +23,26 @@
...
+    command_list = gdb.execute('help all', False, True).strip().split('\n')
+    existing_commands = set()
+    for line in command_list:
+        line = line.strip()
+        # Skip non-command entries
+        if len(line) == 0 or line.startswith('Command class:') or line.startswith('Unclassified commands'):
```
disconnect3d, line 36: While we don't care that much about it - this might fail on other GDB language versions.
If we don't have any better solution for that, it can stay as it is.

pwndbg/commands/__init__.py
```diff
@@ -23,10 +23,26 @@
...
+            continue
+        command = line.split(' ')[0]
```
disconnect3d: Why not just `list.split()[0]` so we will split on *any whitespaces of any length*?

pwndbg/commands/__init__.py
```diff
@@ -23,10 +23,26 @@
...
+        existing_commands.add(command)
+    return list(existing_commands)
```
disconnect3d: Is there any reason to return a list instead of a set?

It is almost better to have a set or a tuple if you don't need a list (e.g. keeping the order of assertions, prepending to the begining or having mutability).

See also https://wiki.python.org/moin/TimeComplexity

pwndbg/commands/__init__.py
```diff
@@ -23,10 +23,26 @@
...
+
+builtin_commands = list_current_commands()
 
 class Command(gdb.Command):
     """Generic command wrapper"""
     command_names = set()
+    builtin_commands = builtin_commands
+    builtin_override_whitelist = ['up', 'down', 'search', 'pwd', 'start']
```
disconnect3d: This can be a set literal :)

pwndbg/commands/__init__.py
```diff
@@ -23,10 +23,26 @@
...
+        command = line.split(' ')[0]
+        existing_commands.add(command)
+    return list(existing_commands)
+
+builtin_commands = list_current_commands()
 
 class Command(gdb.Command):
     """Generic command wrapper"""
     command_names = set()
+    builtin_commands = builtin_commands
```
disconnect3d: Hmm... I am not sure about that

pwndbg/commands/__init__.py
ZetaTwo: As is, this code is run before https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/__init__.py#L130 but I completely agree with the general point. Your long term solutions are great but I will probably just go do something like:
1. Save current pagination state
2. Turn it off
3. Call "help all"
4. Restore previous state

And I can look into the setting context handler separately. Where do you think would be an appropriate place to place such code?

pwndbg/commands/__init__.py
ZetaTwo: Agree, will change

pwndbg/commands/__init__.py
ZetaTwo: No reason, I'll just do set instead.

pwndbg/commands/__init__.py
ZetaTwo: Should I change to just a global variable and don't save it in the class?

pwndbg/commands/__init__.py
ZetaTwo: Agree

pwndbg/commands/__init__.py
ZetaTwo: The least bad way I've come up with is to take the 2nd line of output, split at ":" and use that to search for ignored lines and then do something based on empty lines for the last chunk. Or possibly just do it related to empty lines for every chunk.

pwndbg/commands/__init__.py
disconnect3d: Yeah, lets make it only a global variable.

pwndbg/commands/__init__.py
disconnect3d: Lets keep it as it is for now.

EDIT: In theory we could apply this check only for debug/development. However I am not sure how to detect it properly without applying a flag which ppl might not switch :P.

pwndbg/commands/__init__.py
```diff
@@ -23,10 +23,26 @@
 
 commands = []
 
+def list_current_commands():
+    gdb.execute('set pagination off')
+    command_list = gdb.execute('help all', False, True).strip().split('\n')
```
disconnect3d: Btw in `execute('help all', False, True)` the `False, True` part is not so explicit. Can we pass those as keyword arguments so it is immediately known what are we setting in here?

pwndbg/commands/__init__.py
disconnect3d: Hmm... not sure about the place. I've created `pwndbg/gdbutils` some time ago with an idea to move and abstract out all gdb related APIs (so we can support e.g. lldb in the loooong future). Maybe there?

disconnect3d: @zachriggle @anthraxx or anyone else any opinions here? :P

ZetaTwo:
Uhm, I messed up with my git:ing, please hold.

ZetaTwo:
Ok, this should address all the points discussed. The solution for saving the pagination value is not beautiful but I didn't find a better way.

pwndbg/commands/__init__.py
```diff
@@ -23,10 +23,29 @@
 
 commands = []
 
+def list_current_commands():
+    current_pagination = gdb.execute('show pagination', to_string=True)
+    current_pagination = current_pagination.split()[-1][:-1] # Take last word and skip period
```
```suggestion
    current_pagination = current_pagination.split()[-1].rstrip('.')  # Take last word and skip period
```

disconnect3d:
Thanks!

ZetaTwo:
Thanks for the merge and for a great project!

Issue #535, opened 2018-10-03
Heap issues


adamtanana:
### Description
Issues with heap command on ubuntu 18.04

### Steps to reproduce

Program
```c
#include <stdio.h>
#include <stdlib.h>

int main() {
    char* a = malloc(0x40);
    free(a);
}
```

Gdb session history:
```
b main
run
n
n
heap
bins
```

When printing bins it prints incorrect values (there should be one thing in fast bins) and also prints an error
`Error fetching tcache. GDB cannot access thread-local variables unless you compile with -lpthread.
Exception occured: tcachebins: Cannot access memory at address 0x40 (<class 'gdb.MemoryError'>)
For more info invoke `set exception-verbose on` and rerun the command
or debug it by yourself with `set exception-debugger on`

### My setup

Platform: Linux-4.15.0-34-generic-x86_64-with-Ubuntu-18.04-bionic
Gdb:      8.1.0.20180409-git
Python:   3.6.6 (default, Sep 12 2018, 18:26:19)  [GCC 8.0.1 20180414 (experimental) [trunk revision 259383]]
Pwndbg:   1.1.0 build: b847f1d
Capstone: 4.0.1024
Unicorn:  1.0.1

GrosQuildu:
Can't reproduce. Can you provide libc version (`ldd --version`) and full stacktrace (with `set exception-verbose on`)?

adamtanana:
Stack trace:
```
pwndbg> bins
Error fetching tcache. GDB cannot access thread-local variables unless you compile with -lpthread.
Traceback (most recent call last):
  File "/home/adam/pwndbg/pwndbg/commands/__init__.py", line 109, in __call__
    return self.function(*args, **kwargs)
  File "/home/adam/pwndbg/pwndbg/commands/__init__.py", line 200, in _OnlyWhenRunning
    return function(*a, **kw)
  File "/home/adam/pwndbg/pwndbg/commands/__init__.py", line 209, in _OnlyWhenHeapIsInitialized
    return function(*a, **kw)
  File "/home/adam/pwndbg/pwndbg/commands/heap.py", line 335, in tcachebins
    tcachebins = main_heap.tcachebins(addr)
  File "/home/adam/pwndbg/pwndbg/heap/ptmalloc.py", line 434, in tcachebins
    count = int(counts[i])
  File "/home/adam/pwndbg/pwndbg/inthook.py", line 42, in __new__
    value = value.cast(pwndbg.typeinfo.long)
[...]

PR #546, opened 2018-10-15
Removed unwanted Exception

adamtanana:
Removes an exception in the tcache/bins command where the exception would be printed as well as an error message.
This is purely a UI fix as described in Issue #535


disconnect3d:
Thanks!

PR #552, opened 2018-10-21
Fix tcache and support it on targets w/o -lpthread

disconnect3d:
Short summary:
* fixes tcache having wrong address
* adds tcache support on targets that are compiled without `-lpthread` - we assume that tcache address is at offset 0x10 on the heap
* fixes `pwndbg.symbol.address` as it could return offsets instead of an address

---

Long description below.

This commit fixes tcache: we used the address of &tcache instead of tcache for dereferencing the struct. This can be observed with:
```
pwndbg> p *tcache
$8 = {
  counts = '\000' <repeats 63 times>,
  entries = {0x0 <repeats 64 times>}
}
pwndbg> tcache
{
  counts = "\020`uUUU\000\000\000\000\000\000\000\000\000\000"...,
  entries = {0x0, 0x0, 0x7ffff7fd7740, 0x7ffff7fd80a0, 0x7ffff7fd7740, 0x1, 0x0, 0x7025de0aec8a0300, 0x236a7550e4a6104e, 0x0 <repeats 55 times>}
}
```

It also adds possibility to retrieve tcache information from targets that are compiled without -lpthread [-pthread].

It also fixes `pwndbg.symbol.address`'s `info address` path when it returned addresses that were out of memory maps due to the fact GDB may return a string containing an offset instead of an address. E.g.:
```
pwndbg> info address tcache
Symbol "tcache" is a thread-local variable at offset 0x40 in the thread-local storage for `/lib/x86_64-linux-gnu/libc.so.6'.
```

### DIFF
--- a/pwndbg/heap/ptmalloc.py
+++ b/pwndbg/heap/ptmalloc.py
@@ -138,10 +138,82 @@ class Heap(pwndbg.heap.heap.BaseHeap):
     def has_tcache(self):
         return (self.mp and 'tcache_bins' in self.mp.type.keys() and self.mp['tcache_bins'])
 
+    def _fetch_tcache_symbol_addr(self):
+        """
+        Fetches the address of tcache symbol.
+        NOTE: This may not work on glibcs other than 2.27 and Ubuntu 18.04.
+
+        This has been added based on the observation, that there is a data page
+        that contains a pointer to main_arena and pointer to tcache.
+
+        pwndbg> search -p &main_arena
+                        0x7ffff0000890 0x7ffff7bb0c40
+                        0x7ffff7fd7700 0x7ffff7bb0c40 <--- weird data page
+        [stack]         0x7fffffffdc48 0x7ffff7bb0c40
+        [stack]         0x7fffffffde80 0x7ffff7bb0c40
+
+        pwndbg> x/4xg 0x7ffff7fd7700-0x10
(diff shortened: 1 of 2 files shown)


disconnect3d:
Oh also it seems the tcache is the first allocation made by glibc and so as a result the first chunk is indeed tcache structure, see https://github.com/bminor/glibc/blob/master/malloc/malloc.c#L2971 or :

```c

static void
tcache_init(void)
{
  mstate ar_ptr;
  void *victim = 0;
  const size_t bytes = sizeof (tcache_perthread_struct);

  if (tcache_shutting_down)
    return;

  arena_get (ar_ptr, bytes);
  victim = _int_malloc (ar_ptr, bytes);  // first allocation made in glibc!
  if (!victim && ar_ptr != NULL)
    {
      ar_ptr = arena_get_retry (ar_ptr, bytes);
      victim = _int_malloc (ar_ptr, bytes);
    }


  if (ar_ptr != NULL)
    __libc_lock_unlock (ar_ptr->mutex);

  /* In a low memory situation, we may not be able to allocate memory
     - in which case, we just keep trying later.  However, we
     typically do this very early, so either there is sufficient
     memory, or there isn't enough memory to do non-trivial
     allocations anyway.  */
  if (victim)
    {
      tcache = (tcache_perthread_struct *) victim;
      memset (tcache, 0, sizeof (tcache_perthread_struct));
    }

}
```

PR #553, opened 2018-10-22
correctly determine tcache location

bennofs:
at least on my system, `tcache` is a pointer that needs to be derefed to get the real location of the tcache struct.


bennofs:
I didn't see #552 which also fixes this

Issue #556, opened 2018-10-24
The unicorn aarch64 emulator hangs when start from the address 0x0


equation314:
### Description

This is a bug of unicorn. The second parameter `until` of `uc.emu_start()` is always `0`, it cause hanging when the first parameter `begin` is also `0`, according to [https://github.com/unicorn-engine/unicorn/issues/1038](https://github.com/unicorn-engine/unicorn/issues/1038).

Here is the [https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/emu/emulator.py](https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/emu/emulator.py), line 322~325:

```py
try:
    self.emu_start(self.pc, 0, count=count)
finally:
    self.hook_del(ident)
```

### My setup

```
Gdb:      8.0.1
Python:   2.7.10 (default, Aug 17 2018, 17:41:52)  [GCC 4.2.1 Compatible Apple LLVM 10.0.0 (clang-1000.0.42)]
Pwndbg:   1.1.0 build: 4a34137
Capstone: 4.0.1024
Unicorn:  1.0.1
```

disconnect3d:
Nice catch. Not sure if we can do anything with it except of waiting for an upstream fix?

equation314:
Sorry, the latest unicorn (unreleased) fixed this.

PR #557, opened 2018-10-24
Fix #555 - The return address register not restored

### LINKED ISSUE #555: The return address register not restored when initializing the Emulator
### Description

The return address register not restored for non-x86 arches when initializing the `Emulator` class, and cause `pc` to be set to `0x0` after the `ret` instruction.

Here is the [https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/emu/emulator.py](https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/emu/emulator.py), line 144~146:

```py
# Initialize the register state
for reg in list(self.regs.misc) + list(self.regs.common) + list(self.regs.flags):
    enum = self.get_reg_enum(reg)
```

~~#556  is the reason of hanging.~~

### My setup

```
Gdb:      8.0.1
Python:   2.7.10 (default, Aug 17 2018, 17:41:52)  [GCC 4.2.1 Compatible Apple LLVM 10.0.0 (clang-1000.0.42)]
Pwndbg:   1.1.0 build: 4a34137
Capstone: 4.0.1024
Unicorn:  1.0.1
```

equation314:
Fixed #555.

Since #556 has been fixed in the master branch of unicorn (the release version 1.0.1 still hangs), the gdb no longer hangs when debug an aarch64 program, but there is still a little difference between before and after the fix.

Before fix, because of the `lr` register not restored, pwndbg can not get the next instruction correctly after the `ret` instruction at `0x8026c`:

After fix, pwndbg can get the correct return address `0x800b8`:

### DIFF
--- a/pwndbg/emu/emulator.py
+++ b/pwndbg/emu/emulator.py
@@ -142,7 +142,7 @@ class Emulator(object):
         self._curr = None
 
         # Initialize the register state
-        for reg in list(self.regs.misc) + list(self.regs.common) + list(self.regs.flags):
+        for reg in list(self.regs.retaddr) + list(self.regs.misc) + list(self.regs.common) + list(self.regs.flags):
             enum = self.get_reg_enum(reg)
 
             if not reg:
@@ -450,7 +450,7 @@ class Emulator(object):
         self._single_step = (address, instruction_size)
 
     def dumpregs(self):
-        for reg in list(self.regs.misc) + list(self.regs.common) + list(self.regs.flags):
+        for reg in list(self.regs.retaddr) + list(self.regs.misc) + list(self.regs.common) + list(self.regs.flags):
             enum = self.get_reg_enum(reg)
 
             if not reg or enum is None:


disconnect3d:
Are the screens showing the same thing? I.e. the stack is a little bit different, why?

Have you tried to create the docker image (as described in issue #555)?

equation314:
I have a little trouble when reproduce the bug in the docker ubuntu, the qemu always crash when I try to attach gdb.

Here is the crash messages:

```
qemu-system-aarch64: /root/qemu-3.0.0/hw/dma/bcm2835_dma.c:146: bcm2835_dma_read: Assertion `size == 4' failed.
Exception occured: Error: Remote connection closed (<class 'gdb.error'>)
Python Exception <class 'gdb.error'> Remote connection closed: 
/build/gdb-GT4MLW/gdb-8.1/gdb/thread.c:93: internal-error: thread_info* inferior_thread(): Assertion `tp' failed.
A problem internal to GDB has been detected,
further debugging may prove unreliable.
```

Versions:
```
ubuntu: 18.04
qemu-system-aarch64: 3.0.0
gdb-multiarch: 8.1
pwndbg: master-4a34137
unicorn: 1.0.1-5dbc640
```

Here is the docker image:

```
docker pull equation314/pwndbg_qemu_emulate:latest
```

Run `./run.sh` at `/root` directory to start the qemu and attach the gdb, run `./switch2origin.sh` or `./switch2fix.sh` to switch the pwndbg version between the official and my fixed.

disconnect3d:
I'll get to it on sunday.

disconnect3d:
@equation314 hmm.. is the qemu-3.0.0 the same version as the one you tested things locally?

equation314:
@disconnect3d I built qemu-3.0.0 from sources on Docker Linux, but tested locally with the same version from brew on macos. 

I think the reason for this crash is not related to qemu but gdb-multiarch 8.1. Because I tried to use aarch64-none-elf-gdb 8.0.1 on macos to remotely debug the kernel running in qemu on Linux, it didn't crash. But instead，when used gdb-multiarch 8.1 on Linux to remotely debug the kernel running in qemu on macos, it crashed.

disconnect3d:
It's rather something with qemu. I've built my own GDB 8.1 in the docker container and the same problem occurs :(. In the end it's qemu asserting something...

disconnect3d:
Maybe you can show an https://asciinema.org/ cast before and after the pr on ur local machine?

disconnect3d:
I haven't tested this but I merged it as I believe this fix is needed. Thanks! :)

equation314:
I'm so sorry that I have no time to do this recently. I'll try to give the terminal record this weekend~

PR #559, opened 2018-10-26
Add missing rust type

equation314:
#431 added basic rust support, but some types are missing. Debugging the rust program may cause pwndbg to throw the following exception:

```
Exception occured: Error: No type named int. (<class 'gdb.error'>)
Python Exception <class 'gdb.error'> No type named int.: 
```


disconnect3d (approved):

disconnect3d:
Thanks for the PR! It looks good and I assume it works fine :P.

JFYI: The solution we use is not that good in the long term.We should probably handle `show language` and set types based on that.
Pwndbg fails also with Go binaries in as imilar manner; I played with it once and I had some problems to make it work...

PR #563, opened 2018-10-28
fix unpacking bins tuples

GrosQuildu:


pwndbg/commands/heap.py
```diff
@@ -38,15 +38,18 @@ def format_bin(bins, verbose=False, offset=None):
 
     result = []
     for size in bins:
+        if size == 'type':
+            continue
```
disconnect3d: That seems unintuitive. It seems like `bins` is a collection of sizes and suddenly it can be a `'type'`.

I know `bins` is an `OrderedDict` here that has `'type'` key and also a key for each bin size.

We could simplify this code by popping this key before the loop and using the variable later on, e.g.:

```python
bins_type = bins.pop('type')

for size in bins:
    # ...
    if bins_type == 'fastbins":
        # ...
```

etc.

Other idea could be making an object here, but I am not sure yet if we need it.

disconnect3d (changes requested):
Let's make the `for size in bins` more clear/easier to understand

disconnect3d (approved):

PR #567, opened 2018-11-07
add support of PIE for r2

matrix1001:
Sorry for this super late PR. This is just a small problem due to PIE. I am not very familiar to the api of `pwndbg`. But this change just fixed the PIE problem of r2.

### DIFF
--- a/pwndbg/commands/radare2.py
+++ b/pwndbg/commands/radare2.py
@@ -25,8 +25,11 @@ def r2(arguments, no_seek=False):
 
     # Build up the command line to run
     cmd = ['radare2', filename]
+    addr = pwndbg.regs.pc
+    if pwndbg.wrappers.checksec.pie_status() == "PIE enabled":
+        addr -= pwndbg.elf.exe().address
     if not no_seek and pwndbg.proc.alive:
-        cmd.extend(['-s', hex(pwndbg.regs.pc)])
+        cmd.extend(['-s', hex(addr)])
     cmd += arguments
 
     try:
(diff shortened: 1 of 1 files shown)


pwndbg/commands/radare2.py
```diff
@@ -25,8 +25,11 @@ def r2(arguments, no_seek=False):
 
     # Build up the command line to run
     cmd = ['radare2', filename]
+    addr = pwndbg.regs.pc
+    if pwndbg.wrappers.checksec.pie_status() == "PIE enabled":
```
disconnect3d:
```suggestion
    if pwndbg.elf.is_pie():
```

Can you check if that works fine?

matrix1001: Exception occured: r2: 'module' object has no attribute 'is_pie' (<type 'exceptions.AttributeError'>)

elf is a module. 

Maybe we need an object?

disconnect3d: Oh it should be:
```python
pwndbg.elf.get_elf_info(filepath).is_pie
```

matrix1001: Checked.  But that line should be:
```
if pwndbg.elf.get_elf_info(filename).is_pie:
```

disconnect3d: Sure, can you change it to this? :)

matrix1001: Can you see it ?

disconnect3d (approved):

pwndbg/commands/radare2.py
disconnect3d: Yeah, thx!

PR #569, opened 2018-11-11
fix qemu global variable root conflict with function name

skysider:


disconnect3d:
@skysider Hey, thanks for the PR. Can you describe the issue here? Any chance to show how to trigger it as well?

skysider:
@disconnect3d  When I debug a mips program  with qemu "qemu-mips -g 1234 ./test-mips", and I use gdb to attach to it, it displays errors in line 36 of pwndbg/pwndbg/file.py which says something wrong with pwndbg.qemu.root becase it should be a path string instead of a function.

disconnect3d:
Thanks!

Issue #568, opened 2018-11-07
bug in heap command


matrix1001:
### Description

Bug of command heap

### Steps to reproduce

This bug is due to a strange alignment strategy in newer version of 32 bit glibc.

In a nutshell, the first 16 bytes of heap is 0 in glibc-2.26 (32bit), glibc-2.27 (32bit).

As a result, while parsing heap, the first chunk has size 0 and trigger stop.

### My setup

No need for setup, just check those two glibc.

![image](https://user-images.githubusercontent.com/32097922/48119835-52b9e680-e2ab-11e8-8b90-b0919e308778.png)

### My fix

fix is really easy, just ignore the first chunk if size is 0.

disconnect3d:
I believe the reason of this is thread cache and that a lot of bytes in there belong to `tcache_perthread_struct`, see https://github.com/pwndbg/pwndbg/pull/552#issuecomment-431706323

The `tcache_perthread_struct` structure layout can be seen here -  https://github.com/bminor/glibc/blob/master/malloc/malloc.c#L2898-L2902 , it starts with `char[64]` and then has `tcache_entry*[64]`.

matrix1001:
I doubt that.

Check this (32bit libc-2.26).

![image](https://user-images.githubusercontent.com/32097922/48207747-49647300-e3ac-11e8-9b5a-adf2bb89fe98.png)

So I believe the reason is `_int_malloc`. 

And a bug view.
```
pwndbg> heap
0x5655a000 {
  mchunk_prev_size = 0,
  mchunk_size = 0,
  fd = 0x0,
  bk = 0x151,
  fd_nextsize = 0x0,
  bk_nextsize = 0x0
}
pwndbg>

```

disconnect3d:
I believe it was fixed with https://github.com/pwndbg/pwndbg/pull/571

PR #571, opened 2018-11-12
bug in heap command #568  (fix)

matrix1001:
Bug fix for heap command.

Note that this is only for ptmalloc. Not tested on dlmalloc.

By the way, why there is still a `diff` for `radare2.py` ? ( I'm not a experienced user of git :P )

### DIFF
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -97,18 +97,15 @@ def heap(addr=None):
 
     # Print out all chunks on the heap
     # TODO: Add an option to print out only free or allocated chunks
-    """
-    Check if there is a alignment at the start of the heap.
 
-    Note:
-        I am not very familiar with the api `pwndbg`. This implemention is ugly.
-    """
+    # Check if there is an alignment at the start of the heap
     size_t = pwndbg.arch.ptrsize
     first_chunk_size = pwndbg.arch.unpack(pwndbg.memory.read(addr + size_t, size_t))
     if first_chunk_size == 0:
-        addr += size_t * 2
+        addr += size_t * 2  # Skip the alignment
+
     while addr < page.vaddr + page.memsz:
-        chunk = malloc_chunk(addr)
+        chunk = malloc_chunk(addr)  # Prints the chunk
         size = int(chunk['size'])
 
         # Clear the bottom 3 bits
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -97,6 +97,16 @@ def heap(addr=None):
 
     # Print out all chunks on the heap
     # TODO: Add an option to print out only free or allocated chunks
+    """
+    Check if there is a alignment at the start of the heap.
+
+    Note:
+        I am not very familiar with the api `pwndbg`. This implemention is ugly.
+    """
+    size_t = pwndbg.arch.ptrsize
+    first_chunk_size = pwndbg.arch.unpack(pwndbg.memory.read(addr + size_t, size_t))
+    if first_chunk_size == 0:
+        addr += size_t * 2
     while addr < page.vaddr + page.memsz:
         chunk = malloc_chunk(addr)
         size = int(chunk['size'])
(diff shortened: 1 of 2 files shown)


disconnect3d:
> By the way, why there is still a diff for radare2.py ? ( I'm not a experienced user of git :P )

It seems those changes are there on your branch. It's mainly because you used your own `dev` branch and didn't update it after the last PR was merged.

There are few ways to fix this... the easiest may be to do `git fetch origin` (it shall download changes from origin so that `origin/dev` will correspond to upstream - pwndbg's -  dev branch) and then `git merge origin/dev` shall merge the changes from origin to your own branch.

You might need to solve conflicts but I am not sure about that.

matrix1001:
Thanks. Now I can see only one file changed in the PR.

disconnect3d:
@GrosQuildu @andigena @gsingh93  any comments on this?

Can we also confirm this behavior from happening in the glibc code? :/

I mean, we must be sure it is valid and understand why is that.. :F

matrix1001:
Well, I checked/debuged the source code of [sysmalloc](https://code.woboq.org/userspace/glibc/malloc/malloc.c.html#sysmalloc) and verified this.

image

debug:
```
pwndbg> p aligned_brk
$3 = 0x5655a008
pwndbg> p front_misalign
$4 = 8
pwndbg>
```

disconnect3d:
Hey, I haven't had time to look through this yet. Please ping me if I won't make a response here till the end of this week.

GrosQuildu:
How sysmalloc seems to work:
* sysmalloc try mmap
    * if `MALLOC_ALIGNMENT == 2 * SIZE_SZ` (normal case), then no alignment
    * otherwise there is some alignment, idk how to detect that (we can't check `MALLOC_ALIGNMENT`)
* if not mmaped
    * if `if (av != &main_arena)`
        * expand or create new heap, probably same case as above
    * otherwise try `MORECORE` (`sbrk`) (on failure `mmap` again)
        * if it expanded previous mem, then no alignment
        * otherwise memory may be aligned (in both cases: when arena is and is not contiguous), the code is as @matrix1001 copy-pasted.

As for the code:
```c
#define chunk2mem(p)   ((void*)((char*)(p) + 2*SIZE_SZ))
#define MALLOC_ALIGNMENT 2 * SIZE_SZ
#define MALLOC_ALIGN_MASK (MALLOC_ALIGNMENT-1)

front_misalign = (INTERNAL_SIZE_T) chunk2mem (brk) & MALLOC_ALIGN_MASK;
if (front_misalign > 0) { 
    correction = MALLOC_ALIGNMENT - front_misalign; 
    aligned_brk += correction;
}
```
So for example on 64bit systems with 32bit programs we have compile/runtime discrepancy and `chunk2mem` adds `2*4` but `MALLOC_ALIGNMENT` is 16 and we got 8 byte alignment.

On x64 with 64b programs `chunk2mem` adds `2*8` and `front_misalign` is 0.

Same for a few libc versions I checked.

As for the PR: condition for the first chunk to have `size != 0` is sane enough to merge for me ;) But may be better to change it to:
```python
if addr is None:
        addr = page.vaddr

        size_t = pwndbg.arch.ptrsize
        first_chunk_size = pwndbg.arch.unpack(pwndbg.memory.read(addr + size_t, size_t))
        if first_chunk_size == 0:
            addr += size_t * 2  # Skip the alignment
```
If somebody specify addr he probably know what he's doing and changing the addr without info may be misleading.

Another thing: in `ptmalloc.py` there is `HeapInfo.first_chunk` field which takes into account that `first heap has a heap_info and a malloc_state before the actual chunks`. Idk when this happens, but as `heap` command uses `get_heap_boundaries` and not `first_chunk` that case may be broken as well. So maybe it would be better to use something like `main_heap.arenas[0].heaps[0].first_chunk` in `heap` cmd?

Nice feature would be to extend `heap` cmd such that it also finds top chunk and work backwards to the first one (as we should handle corrupted heaps as well).

disconnect3d:
@GrosQuildu Thx for looking into it.

Re: 
> But may be better to change it to (...)

And:

> Another thing: in `ptmalloc.py` there is `HeapInfo.first_chunk` (...)

Can you send a PR with those two things? (And optionally with the `heap` command extend ;>)

PR #573, opened 2018-11-22
Fixed register names and capstone mode flags for sparc

gymgit:
On sparc architecture the original register aliases caused  the `context regs` command to fail. Similarly, the default `CS_MODE_32` causes a `capstone.CsError: Invalid mode (CS_ERR_MODE)` exception, thus the `context disasm` failed. I have tested this with the following sw versions:

```
qemu-sparc32plus --version
qemu-sparc32plus version 3.0.0
Copyright (c) 2003-2017 Fabrice Bellard and the QEMU Project developers

gdb-multiarch --version
GNU gdb (GDB) 8.2
```
Cheers,
Gym


disconnect3d:
Haha, seems you also played the Dragon Sector CTF :). Thanks a lot! <3

PR #579, opened 2018-11-27
Fixes #578

### LINKED ISSUE #578: pwndbg will not display source code when source code has chinese characters
When use directory glibc source in gdb(directory /root/Pwn/glibc/glibc-2.23/malloc),  it will display the source code in pwndbg like this:
image

However, **when I add Chinese characters in souce code like this**, pwndbg will not display it . Is this a bug? Will you fix it?
image

disconnect3d:
It turned out we failed when there were unicode characters in a file.

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -8,6 +8,7 @@ from __future__ import unicode_literals
 import ast
 import codecs
 import sys
+from io import open
 
 import gdb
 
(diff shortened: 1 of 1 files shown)


pwndbg/commands/context.py
```diff
@@ -8,6 +8,7 @@
 import ast
 import codecs
 import sys
+from io import open
```
disconnect3d, line 11: This makes it works with both Py2 and Py3 (see http://python-future.org/compatible_idioms.html#file-io-with-open ).

PR #591, opened 2019-01-15
Bump Capstone to 4.0.1

disconnect3d:
As it is there! https://pypi.org/project/capstone/#history

It has a minor bug which hits us when we do `version` command, but it isn't that big deal (see https://github.com/aquynh/capstone/issues/1315#issuecomment-454386418).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,6 +9,6 @@ ROPgadget
 six
 unicorn>=1.0.0
 pygments
-capstone==4.0.0
+capstone==4.0.1
 enum34
 pytest


zachriggle:
Do we have any actual dependencies on Capstone 4.x at all?

Edit: I am assuming that we do, since we were reliant on `next` (which was the unreleased 4.x).

Since Capstone has broken our old install script, we may need to cut a new `stable` release (or at least cherry-pick the `requirements` changes).

disconnect3d:
Capstone 4.x adds some new features and fixes some bugs (e.g. adds `endbr32` and `endbr64` instructions for x86/x64) so it's a nice to have (otherwise, the `context disasm` crashes when one of those instructions appear).

Re: release - yeah, need to do that.

PR #597, opened 2019-02-16
Init IDA RPC when ida-enabled changes

qwaz:
I think it is more natural to try initializing IDA RPC when ida-enabled setting changes.


disconnect3d:
Sounds good to me, thx o/

PR #603, opened 2019-03-26
Enhance remote QEMU targets debugging experience

disconnect3d:
- improve vmmap command help and inform user that the result might not be accurate for QEMU targets
- updating stack pages (`pwndbg.stack.stacks`): skip detecting stack pages if SP is 0 (this lead to displaying memory page that started at -0x1000)
- `pwndbg.vmmap.find` - don't skip addresses below `pwndbg.memory.MMAP_MIN_ADDR` - this function is used by `pwndbg.color.memory.get` which is used by `vmmap` command - because we skipped addresses - a memory page starting below that address wasn't colored by `vmmap` display
- memory pages found with exploration will now have `<explored>` name (or objfile) in `vmmap`
- The `proc_pid_maps` returns fast for QEMU-user targets - with this PR it will also do for QEMU kernel targets too since there are no files anyway (this was bug-prone as QEMU targets have 42000 pid returned by GDB and we potentially tried adding memory pages from another process, which luckily usually wasn't there)

### DIFF
--- a/pwndbg/vmmap.py
+++ b/pwndbg/vmmap.py
@@ -62,7 +62,7 @@ def get():
 
 @pwndbg.memoize.reset_on_stop
 def find(address):
-    if address is None: # or address < pwndbg.memory.MMAP_MIN_ADDR:
+    if address is None:
         return None
 
     address = int(address)
--- a/pwndbg/vmmap.py
+++ b/pwndbg/vmmap.py
@@ -62,11 +62,10 @@ def get():
 
 @pwndbg.memoize.reset_on_stop
 def find(address):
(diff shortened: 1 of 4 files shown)


pwndbg/vmmap.py
```diff
@@ -62,11 +62,10 @@ def get():
 
 @pwndbg.memoize.reset_on_stop
 def find(address):
-    if address is None or address < pwndbg.memory.MMAP_MIN_ADDR:
+    if address is None: # or address < pwndbg.memory.MMAP_MIN_ADDR:
```
disconnect3d:
```suggestion
    if address is None:
```

disconnect3d:
@zachriggle pls review

Issue #574, opened 2018-11-22
Add xuntil command from peda


disconnect3d:
```
xuntil -- Continue execution until an address or function
```

e.g. `xuntil <expr>` where expr is an address/symbol; as a result we want an one time breakpoint + continue with given arguments.

I think we can also extend it with running the program if it is not started yet.

ysf:
Is this needed with break_on_next(address) and break_next_call ? Or is this required for PEDA compatibility?

disconnect3d:
I was asked to add this. Not sure what do you ask by:
> Is this needed with break_on_next(address) and break_next_call

I guess this functionality might use one of those functions

alissonbezerra:
I'd like to take care of it as a first contribution.

disconnect3d:
@alissonbezerra That's great! If you need any help feel free to write here or e.g. on freenode irc #pwndbg (or PM me there - disconnect3d).

stnevans:
Looks like this can be closed. If anyone thinks this should be reopened, feel free.

PR #604, opened 2019-03-27
Add xuntil command

alissonbezerra:
Hi, this is the implementation of xuntil command requested in #574.
I had tried to make the command works just like it does on peda.

As this is my first contribution, sorry if there's any issues.

### DIFF
--- a/pwndbg/commands/peda.py
+++ b/pwndbg/commands/peda.py
@@ -9,6 +9,9 @@ import sys
 
 import gdb
 
+import argparse
+
+import pwndbg.color.message as message
 import pwndbg.auxv
 import pwndbg.commands
 import pwndbg.commands.context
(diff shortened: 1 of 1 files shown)


disconnect3d:
Hey, thanks for the PR! It looks good :)

Two questions tho:
* does it work if I do `xuntil some_symbol`?
* does it work with PIE binaries when I didn't start the binary yet e.g. `xuntil <some offset or symbol>`?   I assume this might not work and we might need to 1) see if the binary has pie and 2) start it first e.g. with `entry` or `starti`, get the address and then use the normal `xuntil` logic? Tbh I'm not 100% sure if this is doable

PS: if you can add some screens or logs from pwndbg sessions  so  we  can see how it works - it would be awesome. Otherwise, I will test it later on my own.

alissonbezerra:
Hi disconnect3d, sorry for the late reply, I had been away from home for a couple days. The command works if you do `xuntil some_symbol`, like `xuntil main`.
And it doesn't work with PIE binaries if they didn't start yet, it is something that I must work around.

disconnect3d:
> And it doesn't work with PIE binaries if they didn't start yet, it is something that I must work around.

Let's accept this for now.

Thanks for the PR!

PR #606, opened 2019-03-30
Added install instruction for Manjaro users on setup.sh

zommiommy:
On manjaro the following command:
```sudo pacman -Syu pwndbg-git```
would not work.
But with yay or yoaurt it does.


disconnect3d:
Thanks :)

PR #608, opened 2019-04-06
Add leakfind command

stnevans:
Leakfind is a command meant to help finding leaks. While pwndbg already supports probeleak for this purpose, probeleak can only find leaks by following a pointer to a depth of one.

The purpose of leakfind is to remedy this by looking deeper than a depth of one. Instead, leakfind takes a depth parameter and can find chains up to that length. An example use of it finding a leak chain is shown below.

leakfind
In the above example it shows multiple possible ways to find a pointer into a section containing "v8" as part of the name. Finding this chain proved to be impossible for me to manually do, but with leakfind it is extremely easy.

### DIFF
--- a/pwndbg/commands/leakfind.py
+++ b/pwndbg/commands/leakfind.py
@@ -9,15 +9,15 @@ from __future__ import print_function
 from __future__ import unicode_literals
 
 import argparse
+from queue import *
 
 import gdb
 
+import pwndbg.color.chain as C
 import pwndbg.color.memory as M
+import pwndbg.color.theme as theme
 import pwndbg.commands
 import pwndbg.vmmap
-import pwndbg.color.theme as theme
-import pwndbg.color.chain as C
-from queue import *
 
 config_arrow_right = theme.Parameter('chain-arrow-right', '—▸', 'right arrow of chain formatting')
 arrow_right = C.arrow(' %s ' % config_arrow_right)
@@ -34,7 +34,10 @@ def get_rec_addr_string(addr,visitedMap):
         parentInfo = visitedMap[addr]
         parent = parentInfo[0]
         parent_base_addr = parentInfo[1]
-        curText = hex(parent_base_addr) + "+"+hex(parent-parent_base_addr)
+        if parent-parent_base_addr < 0:
+            curText = hex(parent_base_addr) + hex(parent-parent_base_addr)
+        else:
+            curText = hex(parent_base_addr) + "+"+ hex(parent-parent_base_addr)
         if parent_base_addr == addr:
             return ""
         return get_rec_addr_string(parent_base_addr,visitedMap) + M.get(parent_base_addr,text=curText)+arrow_right
@@ -58,14 +61,14 @@ Example: leakfind $rsp filename 0x48 6. This would look for any chains of leaks 
 and are a maximum length of 6.\n
 """
 parser.add_argument("address",help="Starting address to find a leak chain from.")
-parser.add_argument("page_name",type=str,nargs="?",default=None,help="Substring required to be part of the name of any found pages")
-parser.add_argument("max_offset",default=0x48,nargs="?",help="Max Offset to add to addresses when looking for leak.")
-parser.add_argument("max_depth",default=0x4,nargs="?",help="Maximum depth to follow pointers to.")
-parser.add_argument("stride",nargs="?",default=0x1,help="Stride to add to add between pointers considered. For example, if this is 4 it would only consider pointers at an offset divisible by 4 from the starting pointer.")
-
+parser.add_argument("-p","--page_name",type=str,nargs="?",default=None,help="Substring required to be part of the name of any found pages")
+parser.add_argument("-o","--max_offset",default=0x48,nargs="?",help="Max offset to add to addresses when looking for leak.")
+parser.add_argument("-d","--max_depth",default=0x4,nargs="?",help="Maximum depth to follow pointers to.")
+parser.add_argument("-s","--stride",nargs="?",default=0x1,help="Stride to add to add between pointers considered. For example, if this is 4 it would only consider pointers at an offset divisible by 4 from the starting pointer.")
+parser.add_argument('--negative_offset',nargs="?",default=0x0,help="Max negative offset to search before an address when looking for a leak.")
 @pwndbg.commands.ArgparsedCommand(parser)
 @pwndbg.commands.OnlyWhenRunning
-def leakfind(address=-1,page_name=None,max_offset=0x40,max_depth=0x4,stride=0x1):
+def leakfind(address=-1,page_name=None,max_offset=0x40,max_depth=0x4,stride=0x1,negative_offset=0x0):
     if address == -1:
         print("No starting address provided. Please run leakfind -h for more information.")
         return
@@ -88,7 +91,7 @@ def leakfind(address=-1,page_name=None,max_offset=0x40,max_depth=0x4,stride=0x1)
     stride = int(stride)
     address=int(address)
     max_offset = int(max_offset)
-
+    negative_offset=int(negative_offset)
     
     #The below map stores a map of child address->(parent_address,parent_start_address)
     #In the above tuple, parent_address is the exact address with a pointer to the child adddress.
@@ -111,7 +114,7 @@ def leakfind(address=-1,page_name=None,max_offset=0x40,max_depth=0x4,stride=0x1)
             timeToDepthIncrease=addressQueue.qsize()
         cur_start_addr = addressQueue.get()
         timeToDepthIncrease-=1
-        for cur_addr in range(cur_start_addr,cur_start_addr+max_offset,stride):
+        for cur_addr in range(cur_start_addr-negative_offset,cur_start_addr+max_offset,stride):
             try:
                 result = int(pwndbg.memory.pvoid(cur_addr))
                 if result in visitedMap:
@@ -125,13 +128,14 @@ def leakfind(address=-1,page_name=None,max_offset=0x40,max_depth=0x4,stride=0x1)
                 #That means the memory was unreadable. Just skip it if we can't read it. 
                 break
 
-    #A map of chain_length->list of lines. Used to let us print in a somewhat nice manner.
+    #A map of length->list of lines. Used to let us print in a somewhat nice manner.
     outputMap = {}
 
+ 
     for child in visitedMap:
         childPage = getPage(child)
-        if (not childPage is None) and not (childPage.vaddr == foundPages.vaddr):
-            if not page_name == None:
+        if childPage is not None:
+            if page_name is not None:
                 if not page_name in childPage.objfile:
                     continue
             line = get_rec_addr_string(child,visitedMap) + M.get(child) + " " + M.get(child,text=childPage.objfile)
--- a/pwndbg/commands/leakfind.py
+++ b/pwndbg/commands/leakfind.py
@@ -66,7 +66,7 @@ parser.add_argument("-s", "--step", nargs="?", default=0x1, help="Step to add be
 parser.add_argument('--negative_offset',nargs="?", default=0x0, help="Max negative offset to search before an address when looking for a leak")
 @pwndbg.commands.ArgparsedCommand(parser)
 @pwndbg.commands.OnlyWhenRunning
-def leakfind(address=-1, page_name=None, max_offset=0x40, max_depth=0x4, stride=0x1, negative_offset=0x0):
+def leakfind(address=None, page_name=None, max_offset=0x40, max_depth=0x4, stride=0x1, negative_offset=0x0):
     if address is None:
         raise argparse.ArgumentTypeError('No starting address provided.')
 
--- a/pwndbg/commands/leakfind.py
+++ b/pwndbg/commands/leakfind.py
@@ -21,9 +21,9 @@ import pwndbg.vmmap
 from pwndbg.chain import config_arrow_right
 
 
-#Used to recursively print the pointer chain. 
-#addr is a pointer. It is taken to be a child pointer.
-#visited_map is a map of children -> (parent,parent_start)
+# Used to recursively print the pointer chain. 
+# addr is a pointer. It is taken to be a child pointer.
+# visited_map is a map of children -> (parent,parent_start)
 def get_rec_addr_string(addr,visited_map):
     page = pwndbg.vmmap.find(addr)
     arrow_right = C.arrow(' %s ' % config_arrow_right)
@@ -41,15 +41,15 @@ def get_rec_addr_string(addr,visited_map):
             curText = hex(parent_base_addr) + "+" + hex(parent - parent_base_addr)
         if parent_base_addr == addr:
             return ""
-        return get_rec_addr_string(parent_base_addr,visited_map) + M.get(parent_base_addr,text=curText)+arrow_right
+        return get_rec_addr_string(parent_base_addr,visited_map) + M.get(parent_base_addr,text=curText) + arrow_right
     else:
         return ""
 
 
-#Useful for debugging. Prints a map of child -> (parent, parent_start)
+# Useful for debugging. Prints a map of child -> (parent, parent_start)
 def dbg_print_map(maps):
     for child, parent_info in maps.items():
-        print("0x%x + (0x%x, 0x%x)" % ((child, parent_info[0],parent_info[1])))
+        print("0x%x + (0x%x, 0x%x)" % (child, parent_info[0], parent_info[1]))
 
 parser = argparse.ArgumentParser()
 parser.description = """
@@ -59,14 +59,14 @@ Example: leakfind $rsp --page_name=filename --max_offset=0x48 --max_depth=6. Thi
 and are a maximum length of 6\n
 """
 parser.add_argument("address",help="Starting address to find a leak chain from")
-parser.add_argument("-p","--page_name",type=str,nargs="?",default=None,help="Substring required to be part of the name of any found pages")
-parser.add_argument("-o","--max_offset",default=0x48,nargs="?",help="Max offset to add to addresses when looking for leak")
-parser.add_argument("-d","--max_depth",default=0x4,nargs="?",help="Maximum depth to follow pointers to")
-parser.add_argument("-s","--stride",nargs="?",default=0x1,help="Stride to add to add between pointers considered. For example, if this is 4 it would only consider pointers at an offset divisible by 4 from the starting pointer")
-parser.add_argument('--negative_offset',nargs="?",default=0x0,help="Max negative offset to search before an address when looking for a leak")
+parser.add_argument("-p", "--page_name", type=str, nargs="?", default=None, help="Substring required to be part of the name of any found pages")
+parser.add_argument("-o", "--max_offset", default=0x48, nargs="?", help="Max offset to add to addresses when looking for leak")
+parser.add_argument("-d", "--max_depth", default=0x4, nargs="?", help="Maximum depth to follow pointers to")
+parser.add_argument("-s", "--stride", nargs="?", default=0x1, help="Stride to add to add between pointers considered. For example, if this is 4 it would only consider pointers at an offset divisible by 4 from the starting pointer")
+parser.add_argument('--negative_offset',nargs="?", default=0x0, help="Max negative offset to search before an address when looking for a leak")
 @pwndbg.commands.ArgparsedCommand(parser)
 @pwndbg.commands.OnlyWhenRunning
-def leakfind(address=-1,page_name=None,max_offset=0x40,max_depth=0x4,stride=0x1,negative_offset=0x0):
+def leakfind(address=-1, page_name=None, max_offset=0x40, max_depth=0x4, stride=0x1, negative_offset=0x0):
     if address == -1:
         raise argparse.ArgumentTypeError('No starting address provided.')
 
@@ -78,20 +78,21 @@ def leakfind(address=-1,page_name=None,max_offset=0x40,max_depth=0x4,stride=0x1,
     if not pwndbg.memory.peek(address):
         raise argparse.ArgumentTypeError('Unable to read from starting address.')
 
-    max_depth=int(max_depth)
-    #Just warn the user that a large depth might be slow. Probably worth checking offset^depth < threshold. Do this when more benchmarking is established.
+    max_depth = int(max_depth)
+    # Just warn the user that a large depth might be slow.
+    # Probably worth checking offset^depth < threshold. Do this when more benchmarking is established.
     if max_depth > 8:
         print("leakfind may take a while to run on larger depths.")
     
     stride = int(stride)
-    address=int(address)
+    address = int(address)
     max_offset = int(max_offset)
-    negative_offset=int(negative_offset)
+    negative_offset = int(negative_offset)
     
-    #The below map stores a map of child address->(parent_address,parent_start_address)
-    #In the above tuple, parent_address is the exact address with a pointer to the child adddress.
-    #parent_start_address is an address that a previous address pointed to.
-    #We need to store both so that we can nicely create our leak chain.
+    # The below map stores a map of child address->(parent_address,parent_start_address)
+    # In the above tuple, parent_address is the exact address with a pointer to the child adddress.
+    # parent_start_address is an address that a previous address pointed to.
+    # We need to store both so that we can nicely create our leak chain.
     visited_map = {}
     visited_set = {int(address)}
     address_queue = Queue()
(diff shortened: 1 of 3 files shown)


anthraxx:
Wouldn't it make sense to make the existing command more powerful than? It sounds weird to have one with less functionality and one with more while both serve the very same purpose.

Of there is no difference but depth, we should incorporate the changes into the existing one

stnevans:
That's a reasonable point. Here are a few reasons I think the current way might be cleaner. 
1) 
I'm a little bit worried about having too many arguments. Leakfind already has 4 optional arguments, and combining it with probeleak would require adding another for its max_distance argument. 
2) 
Right now, leakfind doesn't consider negative offsets at all. This is partially because often pointers would point to objects on the heap. Interesting pointers might be stored in something such as a vtable of an object. The data prior to the object might be unreliable to count on. 
I was also considering adding something such as max_negative_offset to leakfind, but this only continues to bloat the number of arguments

If you don't think these are worth making a new command for, how would you suggest I go about replacing probeleak with this? Would I just delete the probeleak command and replace it with this?
What default depth would makes sense? I had the default depth being larger than 1 because this command was made to find chains, but do you think people would normally want to see chains?

I also had a general development question:
What would be a good way to handle the page_name argument? I found that I wanted to use the page_name argument more than I wanted to change the other defaults, however there is currently no way to search for any page name and have non-default other arguments.

disconnect3d:
So for now, the two commands work a bit different:
```
pwndbg> probeleak $rsp
0x08: 0x00007fffffffe6b5 = (rw-p) [stack] + 0x206b5
0x18: 0x00007fffffffe6bd = (rw-p) [stack] + 0x206bd
0x20: 0x00007fffffffe6d0 = (rw-p) [stack] + 0x206d0
0x28: 0x00007fffffffecbc = (rw-p) [stack] + 0x20cbc
0x30: 0x00007fffffffecf2 = (rw-p) [stack] + 0x20cf2
0x38: 0x00007fffffffed10 = (rw-p) [stack] + 0x20d10
pwndbg> leakfind $rsp stack 100
pwndbg>
```
For some reason I am unable to hit similar results; why?

Honestly, I'd be okay with having both commands, but let's see how it goes first.

> I also had a general development question:
> What would be a good way to handle the page_name argument? I found that I wanted to use the  page_name argument more than I wanted to change the other defaults, however there is currently no way to search for any page name and have non-default other arguments.

It should be an option specified with `--page=...` so its order doesn't matter. We might want to do the same for  `--stride=..` or/and `--depth=..`?

An example of that can be found in `search` command source https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/commands/search.py#L51-L83

> 1)
> I'm a little bit worried about having too many arguments. Leakfind already has 4 optional arguments, and combining it with probeleak would require adding another for its max_distance argument.

If we move them to option arguments (`-s=...` or `--something=...`) it should be fine to have a lot of them imho.

> 2)
> Right now, leakfind doesn't consider negative offsets at all. This is partially because often pointers would point to objects on the heap. Interesting pointers might be stored in something such as a vtable of an object. The data prior to the object might be unreliable to count on.
> I was also considering adding something such as max_negative_offset to leakfind, but this only continues to bloat the number of arguments

hmm is that the differece between probeleak and leakfind? Wouldn't it be good to have an option to enable that?

stnevans:
> So for now, the two commands work a bit different:
> 
> ```
> pwndbg> probeleak $rsp
> LEGEND: STACK | HEAP | CODE | DATA | RWX | RODATA
> 0x08: 0x00007fffffffe6b5 = (rw-p) [stack] + 0x206b5
> 0x18: 0x00007fffffffe6bd = (rw-p) [stack] + 0x206bd
> 0x20: 0x00007fffffffe6d0 = (rw-p) [stack] + 0x206d0
> 0x28: 0x00007fffffffecbc = (rw-p) [stack] + 0x20cbc
> 0x30: 0x00007fffffffecf2 = (rw-p) [stack] + 0x20cf2
> 0x38: 0x00007fffffffed10 = (rw-p) [stack] + 0x20d10
> pwndbg> leakfind $rsp stack 100
> pwndbg>
> ```
> For some reason I am unable to hit similar results; why?

Good catch. This is because when I was first developing the command, I made it so that leakfind would not find leaks to pages that the start address was already in. I did this to try to help combat the output becoming too spammy. 
I just ran some tests to see how much it actually helps with that, and it doesn't seem worth the lack of functionality. I'll make it so that it can find pointers to the original page in the next commit. 
Reference benchmarks:

| # of Results | Original | Including Current Page|
| --- | --- | --- |
| Heap Object | 119 | 146 |
| Stack Address | 25 | 45 |


> If we move them to option arguments (`-s=...` or `--something=...`) it should be fine to have a lot of them imho.

I'll add that.

> > Right now, leakfind doesn't consider negative offsets at all. This is partially because often pointers would point to objects on the heap. Interesting pointers might be stored in something such as a vtable of an object. The data prior to the object might be unreliable to count on.
> > I was also considering adding something such as max_negative_offset to leakfind, but this only continues to bloat the number of arguments
> 
> hmm is that the differece between probeleak and leakfind? Wouldn't it be good to have an option to enable that?

I will add a negative_offset option in the next commit.

stnevans:
Looks like travis is complaining about peda.py which wasn't modified by me in this pull request. Should I fix it anyways?

disconnect3d:
@stnevans Yes pls and thx!

stnevans:
@disconnect3d Okay.

pwndbg/commands/leakfind.py
```diff
@@ -0,0 +1,155 @@
...
+import gdb
+
+import pwndbg.color.chain as C
+import pwndbg.color.memory as M
+import pwndbg.color.theme as theme
+import pwndbg.commands
+import pwndbg.vmmap
+
+config_arrow_right = theme.Parameter('chain-arrow-right', '—▸', 'right arrow of chain formatting')
+arrow_right = C.arrow(' %s ' % config_arrow_right)
```
disconnect3d: Note that it doesn't work well if the config/theme is applied after loading pwndbg - and this is what usually (always?) occurs:
```
pwndbg> set chain-arrow-right xxx
Set right arrow of chain formatting to 'xxx'
pwndbg> leakfind -o 100 $rsp
0x7fffffffe440+0x8 —▸ 0x7fffffffe6b5 [stack]
0x7fffffffe440+0x18 —▸ 0x7fffffffe6bd [stack]
```

We need to fix this.

disconnect3d: Also, the `chain-arrow-right` is already defined in another file and I believe we need to import it here instead of redefining it.

pwndbg/commands/leakfind.py
```diff
@@ -0,0 +1,155 @@
...
+            curText = hex(parent_base_addr) + "+"+ hex(parent-parent_base_addr)
+        if parent_base_addr == addr:
+            return ""
+        return get_rec_addr_string(parent_base_addr,visitedMap) + M.get(parent_base_addr,text=curText)+arrow_right
+    else:
+        return ""
+
+#Utility function to get a page from an address.
+def getPage(address):
+    return pwndbg.vmmap.find(address)
```
disconnect3d: This function is redundant (and function calls are expensive in Python). If we really want to use an alias, let's just do:
```suggestion
get_page = pwndbg.vmmap.find
```

pwndbg/commands/leakfind.py
```diff
@@ -0,0 +1,155 @@
...
+                result = int(pwndbg.memory.pvoid(cur_addr))
+                if result in visitedMap:
+                    continue
+                if result in visitedSet:
+                    continue
+                visitedMap[result]=(cur_addr,cur_start_addr) #map is of form child->(parent,parent_start)
+                addressQueue.put(result)
+                visitedSet.add(result)
+            except gdb.error:
+                #That means the memory was unreadable. Just skip it if we can't read it.
```
disconnect3d: Is that correct? Can memory be executable but not readable? Would we miss an interesting pointer in such case?

pwndbg/commands/leakfind.py
```diff
@@ -0,0 +1,155 @@
...
+@pwndbg.commands.OnlyWhenRunning
+def leakfind(address=-1,page_name=None,max_offset=0x40,max_depth=0x4,stride=0x1,negative_offset=0x0):
+    if address == -1:
+        print("No starting address provided. Please run leakfind -h for more information.")
+        return
+
+    foundPages = getPage(address)
+
+    if not foundPages:
+        print("Starting address is not mapped. Please run leakfind -h for more information.")
```
disconnect3d: Can you check if `raise argparse.ArgumentTypeError('Starting address is not mapped.')` would be better here?

Same for the other checks.

pwndbg/commands/leakfind.py
```diff
@@ -0,0 +1,155 @@
...
+        return ""
+
+#Utility function to get a page from an address.
+def getPage(address):
+    return pwndbg.vmmap.find(address)
+
+#Useful for debugging. Prints a map of child -> (parent, parent_start)
+def dbg_print_map(maps):
+    for child, parentInfo in maps.items():
+        print(hex(child) + "("+hex(parentInfo[0])+","+hex(parentInfo[1])+")")
```
disconnect3d: Can we use different formatting here - either `"0x%x + (0x%x, 0x%x)".format(child, parent)` and rename `parentInfo` -> `parent_info`?

pwndbg/commands/leakfind.py
```diff
@@ -0,0 +1,155 @@
...
+    max_offset = int(max_offset)
+    negative_offset=int(negative_offset)
+    
+    #The below map stores a map of child address->(parent_address,parent_start_address)
+    #In the above tuple, parent_address is the exact address with a pointer to the child adddress.
+    #parent_start_address is an address that a previous address pointed to.
+    #We need to store both so that we can nicely create our leak chain.
+    visitedMap = {}
+    visitedSet = set()
+    visitedSet.add(int(address))
```
disconnect3d: Can we apply some formatting changes here?
1) use underscore naming for local variables - e.g. `visited_set`, `address_queue`, `time_to_depth_increase`
2) put a space around operators e.g. `time_to_depth_increase = 0` (same in other cases)
3) this could be: `visited_set = {int(address)}` - you can create a set literal if you put some values around `{}` (otherwise it will be an empty dict)

pwndbg/commands/leakfind.py
```diff
@@ -0,0 +1,155 @@
...
+    #A map of length->list of lines. Used to let us print in a somewhat nice manner.
+    outputMap = {}
+
+ 
+    for child in visitedMap:
+        childPage = getPage(child)
+        if childPage is not None:
+            if page_name is not None:
+                if not page_name in childPage.objfile:
+                    continue
```
disconnect3d, line 132: Just:
```python
if page_name is not None or page_name not in child_page.objfile:
    continue
```

Please note that there is a huge difference between:
```python
x = not X in Y
y = X not in Y
```

and I assume you meant `X not in Y` instead of `not X in Y`

pwndbg/commands/leakfind.py
```diff
@@ -0,0 +1,155 @@
...
+parser = argparse.ArgumentParser()
+parser.description = """
+Attempt to find a leak chain given a starting address. Scans memory near the given address, looks for pointers, 
+and continues that process to attempt to find leaks.\n
+Example: leakfind $rsp filename 0x48 6. This would look for any chains of leaks that point to a section in filename which begin near $rsp, are never 0x48 bytes further from a known pointer, 
+and are a maximum length of 6.\n
+"""
+parser.add_argument("address",help="Starting address to find a leak chain from.")
+parser.add_argument("-p","--page_name",type=str,nargs="?",default=None,help="Substring required to be part of the name of any found pages")
+parser.add_argument("-o","--max_offset",default=0x48,nargs="?",help="Max offset to add to addresses when looking for leak.")
```
disconnect3d: There's some inconsistency here, let's either have dots at the end of all help descriptions or not have them :D

pwndbg/commands/leakfind.py
```diff
@@ -0,0 +1,155 @@
...
+#Useful for debugging. Prints a map of child -> (parent, parent_start)
+def dbg_print_map(maps):
+    for child, parentInfo in maps.items():
+        print(hex(child) + "("+hex(parentInfo[0])+","+hex(parentInfo[1])+")")
+
+parser = argparse.ArgumentParser()
+parser.description = """
+Attempt to find a leak chain given a starting address. Scans memory near the given address, looks for pointers, 
+and continues that process to attempt to find leaks.\n
+Example: leakfind $rsp filename 0x48 6. This would look for any chains of leaks that point to a section in filename which begin near $rsp, are never 0x48 bytes further from a known pointer,
```
disconnect3d: The example is invalid atm

disconnect3d (changes requested):
I added some comments. Thanks again for the PR! o/

pwndbg/commands/leakfind.py
stnevans: I'm not all that familiar with gdb's python API. I meant to say that the memory is unreadable by gdb i.e. unmapped, rather than saying that the protections of the memory didn't include readable. 
The line that would throw an error is the call to pwndbg.memory.pvoid. 
Is the correct interpretation that the memory is unmapped rather than unreadable?

disconnect3d: Ah, that makes sense.

> Is the correct interpretation that the memory is unmapped rather than unreadable?
I'm not 100% sure atm but I think yes

pwndbg/commands/leakfind.py
stnevans: I just removed it altogether.

pwndbg/commands/leakfind.py
stnevans: Fixed. imported from pwndbg.chain

pwndbg/commands/leakfind.py
stnevans: I changed it to do this. Why is this behavior preferable? 
It seems more natural to tell users to run the help function than it does show information about an exception.

pwndbg/commands/leakfind.py
stnevans: Fixed

pwndbg/commands/leakfind.py
stnevans: Fixed

pwndbg/commands/leakfind.py
stnevans: More specifically the .format function didn't seem to be working for me so I used % instead.

pwndbg/commands/leakfind.py
stnevans: I think I fixed all of these now.

pwndbg/commands/leakfind.py
stnevans: I removed all the dots (consistent with another command I saw).

pwndbg/commands/leakfind.py
stnevans: I changed the example.

pwndbg/commands/leakfind.py
disconnect3d: oh sorry i typed it wrong: the `.format` function uses `"{}".format(...)` this kind of formatting strings/specifiers :P np

pwndbg/commands/leakfind.py
disconnect3d: Hmm I thought it is handled in a way it doesn't show an exception... hmm

disconnect3d:
@stnevans 
I applied some refactoring to the PR.

I also think I found some bug:
image

shouldn't the addresses on left be different?

stnevans:
Thanks for the commit. I didn't realize how many cases I had missed. 
Do you mean the address on the left should be ...e378 instead of ...e370+0x8?
I made it print this way to be very clear how each address was reached. 

My interpretation of your screenshot is that e370+0x8 (e378) has a pointer to e448, which has various pointers to the stack around it. 

If this isn't what you meant, can you clarify?

pwndbg/commands/leakfind.py
stnevans: It's not that the stack trace is directly shown, but it suggests either seeing the stacktrace or using a debugger rather than just reading the help.
Output of `leakfind 0`:
`Exception occured: leakfind: Starting address is not mapped. (<class 'argparse.ArgumentTypeError'>)`
`For more info invoke set exception-verbose on and rerun the command
or debug it by yourself with set exception-debugger on
`

disconnect3d: We should handle this exception somewhere in ArgparsedCommand then... or fallback to the previous solution.

disconnect3d:
> Do you mean the address on the left should be ...e378 instead of ...e370+0x8?

I mean that there are multiple lines with 378+0x8 which is... weird?

stnevans:
> I mean that there are multiple lines with 378+0x8 which is... weird?

That's because the address it points to is in an area with multiple other valid addresses. 
Basically, we first discover a pointer to e448 from e378. We then scan that area for more addresses and find that there are different pointers at e448, e448+0x8, ... e448+0x40. 
Each of those lines with the same starting point have different ending points because the second address they use is different (different offsets from it).
Each of those leakchains starts in the same place but ends in a different place.

Does that help at all?

disconnect3d:
> Does that help at all?

Yes, I didn't know/understand that we also try to find pointers on offsets from the addresses we already found pointers to, that makes sense!

Also, this confirms it:
```
pwndbg> probeleak $rsp 0x48
0x00: 0x000055555555987a = (r-xp) /bin/ls + 0x587a
0x08: 0x00007fffffffe418 = (rw-p) [stack] + 0x20418
0x20: 0x00007fffffffe699 = (rw-p) [stack] + 0x20699
0x30: 0x00007fffffffe6a1 = (rw-p) [stack] + 0x206a1
0x38: 0x00007fffffffe6b4 = (rw-p) [stack] + 0x206b4
0x40: 0x00007fffffffeca0 = (rw-p) [stack] + 0x20ca0

pwndbg> leakfind -d 1 $rsp
0x7fffffffe408+0x0 —▸ 0x55555555987a /bin/ls
0x7fffffffe408+0x8 —▸ 0x7fffffffe418 [stack]
0x7fffffffe408+0x20 —▸ 0x7fffffffe699 [stack]
0x7fffffffe408+0x30 —▸ 0x7fffffffe6a1 [stack]
0x7fffffffe408+0x38 —▸ 0x7fffffffe6b4 [stack]
0x7fffffffe408+0x40 —▸ 0x7fffffffeca0 [stack]
```

So the question now is: are we going to remove `probeleak`? @zachriggle @anthraxx @GrosQuildu @bennofs any opinions on that?

If so, maybe we should make `-d 1` the default value?

Also, could you add some info to `FEATURES.md` about this?

disconnect3d:
@stnevans uh, I also wanted to ask you about those but I forgot to ping you ^

pwndbg/commands/leakfind.py
```diff
@@ -0,0 +1,147 @@
...
+"""
+parser.add_argument("address",help="Starting address to find a leak chain from")
+parser.add_argument("-p", "--page_name", type=str, nargs="?", default=None, help="Substring required to be part of the name of any found pages")
+parser.add_argument("-o", "--max_offset", default=0x48, nargs="?", help="Max offset to add to addresses when looking for leak")
+parser.add_argument("-d", "--max_depth", default=0x4, nargs="?", help="Maximum depth to follow pointers to")
+parser.add_argument("-s", "--stride", nargs="?", default=0x1, help="Stride to add to add between pointers considered. For example, if this is 4 it would only consider pointers at an offset divisible by 4 from the starting pointer")
+parser.add_argument('--negative_offset',nargs="?", default=0x0, help="Max negative offset to search before an address when looking for a leak")
+@pwndbg.commands.ArgparsedCommand(parser)
+@pwndbg.commands.OnlyWhenRunning
+def leakfind(address=-1, page_name=None, max_offset=0x40, max_depth=0x4, stride=0x1, negative_offset=0x0):
```
GrosQuildu: address=None maybe?

pwndbg/commands/leakfind.py
```diff
@@ -0,0 +1,147 @@
...
+        raise argparse.ArgumentTypeError('Starting address is not mapped.')
+
+    if not pwndbg.memory.peek(address):
+        raise argparse.ArgumentTypeError('Unable to read from starting address.')
+
+    max_depth = int(max_depth)
+    # Just warn the user that a large depth might be slow.
+    # Probably worth checking offset^depth < threshold. Do this when more benchmarking is established.
+    if max_depth > 8:
+        print("leakfind may take a while to run on larger depths.")
```
GrosQuildu, line 85: print(message.warning()) maybe?

pwndbg/commands/leakfind.py
```diff
@@ -0,0 +1,147 @@
...
+Attempt to find a leak chain given a starting address. Scans memory near the given address, looks for pointers, 
+and continues that process to attempt to find leaks.\n
+Example: leakfind $rsp --page_name=filename --max_offset=0x48 --max_depth=6. This would look for any chains of leaks that point to a section in filename which begin near $rsp, are never 0x48 bytes further from a known pointer, 
+and are a maximum length of 6\n
+"""
+parser.add_argument("address",help="Starting address to find a leak chain from")
+parser.add_argument("-p", "--page_name", type=str, nargs="?", default=None, help="Substring required to be part of the name of any found pages")
+parser.add_argument("-o", "--max_offset", default=0x48, nargs="?", help="Max offset to add to addresses when looking for leak")
+parser.add_argument("-d", "--max_depth", default=0x4, nargs="?", help="Maximum depth to follow pointers to")
+parser.add_argument("-s", "--stride", nargs="?", default=0x1, help="Stride to add to add between pointers considered. For example, if this is 4 it would only consider pointers at an offset divisible by 4 from the starting pointer")
```
GrosQuildu: to add to add ;)

GrosQuildu:
@stnevans @disconnect3d I would delete probeleak (not to double funcionality).

BUT probeleak has a nice features that leakfind misses:
- it finds leaks >outside< memory page (sometimes we can have pointer to end_of_stack+0x10 for example). That would go with new optional argument ofc, but I think it is worth it ;)

Check https://github.com/pwndbg/pwndbg/pull/525#issuecomment-422975291 for code to test.

Also, there may be a bug:
```python
pwndbg> leakfind $some_writable -o 100 -d 5 
0x55555575a000+0x11 —▸ 0x7ffffffdb031 [stack]
pwndbg> leakfind $some_writable -o 100 -d 5 -s 4
0x55555575a000+0x20 —▸ 0x7ffffffdb031 [stack]
pwndbg> probeleak $some_writable 48 0x42
0x00: 0x00007ffffffff000 = (rw-p) [stack] + 0x25000 + 0x0 (outside of the page)
0x08: 0x00007ffffffff041 = (rw-p) [stack] + 0x25000 + 0x41 (outside of the page)
0x11: 0x00007ffffffdb031 = (rw-p) [stack] + 0x1031
0x20: 0x00007ffffffdb031 = (rw-p) [stack] + 0x1031
```
leakfind sometimes finds leak at +0x11 and sometimes at +0x20 (should both be find in cases above?).

pwndbg/commands/leakfind.py
```diff
@@ -0,0 +1,147 @@
...
+    if not pwndbg.memory.peek(address):
+        raise argparse.ArgumentTypeError('Unable to read from starting address.')
+
+    max_depth = int(max_depth)
+    # Just warn the user that a large depth might be slow.
+    # Probably worth checking offset^depth < threshold. Do this when more benchmarking is established.
+    if max_depth > 8:
+        print("leakfind may take a while to run on larger depths.")
+    
+    stride = int(stride)
```
GrosQuildu, line 87: maybe check if stride != 0, not to throw exception later in range()

stnevans:
@GrosQuildu Thanks for the review All your points are good, I'll fix the code.
To address the bug you mention, this is because leakfind only finds one path to an address. 

The original motivation for this was to prevent looping due to cycles (eg address1 points to address2 which points back to address1). 
Given that the depth is fixed, this shouldn't be a huge issue. 
Another advantage is that it isn't uncommon for there to be many paths to a single address (especially at larger depths), and that might result in more output than one might want.

Just so you know, I believe the current method will always take the path to an address that is the shortest chain and as close to the starting address as possible.
Currently the code makes the assumption that there is only one path to a given address. It would be a reasonable amount of work to change that, but it might be worth it. I'll look into it.


@disconnect3d
On the subject of whether to replace probeleak, what do you think of merging leakfind as is with the plan of replacing probeleak once feature parity is achieved?
In addition to what GrosQuildu mentioned about pointing outside of pages, probeleak also shows symbols while leakfind doesn't. 

I would like to add both of those, but I don't know exactly how long it will take. If you think it makes more sense to add those features before merging, I would be fine with that as well.


I'll add something to features.MD in a bit. I'm having some computer problems at the moment, so sorry for the slower responses.

disconnect3d:
> On the subject of whether to replace probeleak, what do you think of merging leakfind as is with the plan of replacing probeleak once feature parity is achieved?

I am totally fine with that. Let's just resolve a few remaining code review comments first.
If you won't be faster than me, I will try to find some time to look at them and either accept them or not.

pwndbg/commands/leakfind.py
disconnect3d:
```suggestion
parser.add_argument("-s", "--step", nargs="?", default=0x1, help="Step to add between pointers so they are considered. For example, if this is 4 it would only consider pointers at an offset divisible by 4 from the starting pointer")
```

pwndbg/commands/leakfind.py
disconnect3d:
```suggestion
def leakfind(address=None, page_name=None, max_offset=0x40, max_depth=0x4, stride=0x1, negative_offset=0x0):
```

pwndbg/commands/leakfind.py
```diff
@@ -0,0 +1,147 @@
...
+parser.add_argument("address",help="Starting address to find a leak chain from")
+parser.add_argument("-p", "--page_name", type=str, nargs="?", default=None, help="Substring required to be part of the name of any found pages")
+parser.add_argument("-o", "--max_offset", default=0x48, nargs="?", help="Max offset to add to addresses when looking for leak")
+parser.add_argument("-d", "--max_depth", default=0x4, nargs="?", help="Maximum depth to follow pointers to")
+parser.add_argument("-s", "--stride", nargs="?", default=0x1, help="Stride to add to add between pointers considered. For example, if this is 4 it would only consider pointers at an offset divisible by 4 from the starting pointer")
+parser.add_argument('--negative_offset',nargs="?", default=0x0, help="Max negative offset to search before an address when looking for a leak")
+@pwndbg.commands.ArgparsedCommand(parser)
+@pwndbg.commands.OnlyWhenRunning
+def leakfind(address=-1, page_name=None, max_offset=0x40, max_depth=0x4, stride=0x1, negative_offset=0x0):
+    if address == -1:
```
disconnect3d:
```suggestion
    if address is None:
```

disconnect3d:
@stnevans I can't commit to your repo; can you accept the suggestions I added so I can merge it? :)

Issue #570, opened 2018-11-11
[AArch64 & QEMU] GDB and Qemu crashes together while connecting to QUEMU gdb stub


TheAifam5:
### Description
I seen similar issue like https://github.com/pwndbg/pwndbg/issues/85.
I tested with the official one `aarch64-linux-gnu-gdb` and the `gdb-multiarch` package and the same issue occurs.

After connecting the GDB to the QEMU, they crash with `segmentation fault` and `abort` errors.

### Steps to reproduce
- Start QEMU using: ` qemu-system-aarch64 -M raspi3 -kernel kernel8.img -d in_asm -S -s` (you need an image, for example from here: https://github.com/bztsrc/raspi3-tutorial)
- Start GDB using `aarch64-linux-gnu-gdb` and type `target remote localhost:1234`

### My setup
OS: ArchLinux x64
show version
```
GNU gdb (GDB) 8.2
Copyright (C) 2018 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "--host=x86_64-pc-linux-gnu --target=aarch64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word".
```

py import sys; print(sys.version):
```
3.7.1 (default, Oct 22 2018, 10:41:28) 
[GCC 8.2.1 20180831]
```

pwndbg version/git commit id: 19eaa19
```
pwndbg> version
Python Exception <class 'ValueError'> invalid literal for int() with base 10: 'warning:': 
Error occurred in Python command: invalid literal for int() with base 10: 'warning:'
```

QEMU log:
```
qemu-system-aarch64 -M raspi3 -kernel kernel8.img -d in_asm -S -s
qemu-system-aarch64: /build/qemu/src/qemu-3.0.0/hw/dma/bcm2835_dma.c:146: bcm2835_dma_read: Assertion `size == 4' failed.
[...]

PR #613, opened 2019-04-21
Change StopIteration to return

CarloMara:
Fix for #570, caused by PEP479


disconnect3d:
Woa,  that's a great catch!

Thanks a lot! 🍻

Issue #617, opened 2019-04-26
Got doesn't work on all pie binaries


stnevans:
### Description
The `got` command sometimes fails to work for pie binaries. This happens when the entry point does not exist in the lowest page in memory.
This is because the got addresses should be rebased relative to the lowest page, but currently the got command assumes that the entry point must be in the lowest page in memory. 
Here's the relevant code:
https://github.com/pwndbg/pwndbg/blob/2f3626046448fbffcfa21822779c769c7be6606a/pwndbg/commands/got.py#L35-L36
https://github.com/pwndbg/pwndbg/blob/2f3626046448fbffcfa21822779c769c7be6606a/pwndbg/commands/got.py#L53-L54
### Steps to reproduce
Here's an example binary with the text section not being the first.
[pie_shop_got.zip](https://github.com/pwndbg/pwndbg/files/3119551/pie_shop_got.zip)
Text section is bolded. Note that there is a different section before it in memory.
<pre>
$gdb ./pie_shop
pwndbg> start
Temporary breakpoint 1 at 0x1266
pwndbg> vmmap
LEGEND: STACK | HEAP | CODE | DATA | RWX | RODATA
            0x1000             0x3000 r-xp     2000 0      <explored>
    0x555555554000     0x555555555000 r--p     1000 0      /home/user/Desktop/ctf/angstrom/pie/pie_shop
    <b>0x555555555000     0x555555556000 r-xp     1000 1000   /home/user/Desktop/ctf/angstrom/pie/pie_shop</b>
    0x555555556000     0x555555557000 r--p     1000 2000   /home/user/Desktop/ctf/angstrom/pie/pie_shop
    0x555555557000     0x555555558000 r--p     1000 2000   /home/user/Desktop/ctf/angstrom/pie/pie_shop
    0x555555558000     0x555555559000 rw-p     1000 3000   /home/user/Desktop/ctf/angstrom/pie/pie_shop
pwndbg> got
GOT protection: Partial RELRO | GOT functions: 8 
Exception occured: got: Cannot access memory at address 0x555555559018 (<class 'gdb.MemoryError'>)
For more info invoke `set exception-verbose on` and rerun the command
or debug it by yourself with `set exception-debugger on`
</pre>

### My setup
Gdb:      7.11.1
Python:   3.5.2 (default, Nov 12 2018, 13:43:14)  [GCC 5.4.0 20160609]
Pwndbg:   1.1.0 build: 8b0a98f

### Fix
[...]

PR #618, opened 2019-04-26
Fix base for got on pie binaries

stnevans:
Fix for #617.
Just calculates the base address of the binary with `pwndbg.self.exe().address` rather than `pwndbg.vmmap.find(pwndbg.elf.entry()).start`.


disconnect3d:
Thanks! o/

PR #621, opened 2019-04-28
Allow ArgparsedCommand to have aliases

stnevans:
As part of #244, I thought it would be useful if we could add aliases for Argparsed commands.

This pr allows us to give a list of aliases as an optional second argument to ArgparsedCommand.
I'm not totally sure if this is the best way to implement command aliases, so if anyone has a better idea feel free to suggest.
Example usage:
`@pwndbg.commands.ArgparsedCommand(parser,aliases=["leakfind2"])`
Example output:
```
pwndbg> leakfind
usage: leakfind [-h] [-p [PAGE_NAME]] [-o [MAX_OFFSET]] [-d [MAX_DEPTH]] [-s [STEP]] [--negative_offset [NEGATIVE_OFFSET]] address
leakfind: error: the following arguments are required: address
pwndbg> leakfind2
usage: leakfind [-h] [-p [PAGE_NAME]] [-o [MAX_OFFSET]] [-d [MAX_DEPTH]] [-s [STEP]] [--negative_offset [NEGATIVE_OFFSET]] address
leakfind: error: the following arguments are required: address
```

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -48,9 +48,10 @@ class Command(gdb.Command):
     builtin_override_whitelist = {'up', 'down', 'search', 'pwd', 'start'}
     history = {}
 
-    def __init__(self, function, prefix=False):
-        command_name = function.__name__
-
+    def __init__(self, function, prefix=False, command_name=None):
+        if command_name is None:
+            command_name = function.__name__
+        
         super(Command, self).__init__(command_name, gdb.COMMAND_USER, gdb.COMPLETE_EXPRESSION, prefix=prefix)
         self.function = function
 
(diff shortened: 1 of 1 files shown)


pwndbg/commands/__init__.py
```diff
@@ -259,15 +263,15 @@ def split_args(self, argument):
 
 class ArgparsedCommand(object):
     """Adds documentation and offloads parsing for a Command via argparse"""
-    def __init__(self, parser_or_desc):
+    def __init__(self, parser_or_desc, aliases=[]):
```
disconnect3d, line 266: This could turn into *default argument trap* (i.e. the `[]` is created _only once_ and multiple calls to this method without passing aliases uses the the default value - the same object) but this could be easily spotted imho.

For reference: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

disconnect3d (approved):

PR #628, opened 2019-05-03
Don't activate the IDA window when jumping

### LINKED ISSUE #627: Don't activate the IDA view when stepping in the pwndbg
### Description

By default the pwndbg will execute Jump(ea) function when we step to a new instruction. However the idc.Jump(ea) function calls the 
`inline bool jumpto(ea_t ea, int opnum=-1, int uijmp_flags=UIJMP_ACTIVATE)`
internally with the default parameter. You can see the UIJMP_ACTIVATE is the default flag and this will instruct the IDA to activate the target view, which will grab the focus (at least on Windows)
Because of this, every time I step in the pwndbg, the focus will shifted into IDA and I have to focus back to pwndbg manually which is quite annoying

### Steps to reproduce

Use the pwndbg with IDA Pro and then do stepping in pwndbg

### My setup

newest pwndbg with IDA Pro 7.2 on Windows

### Possible Fix
We can change the L230 of pwndbg/ida.py
from 
```return _ida.Jump(addr)```
to
```return _ida.jumpto(addr, -1, 0)```
which means we manually set the flag to 0, instead of UIJMP_ACTIVATE.

### PR DESCRIPTION
**Pull request by NyaMisty**
Fix #627

### DIFF
--- a/pwndbg/ida.py
+++ b/pwndbg/ida.py
@@ -227,7 +227,8 @@ def here():
 @withIDA
 @takes_address
 def Jump(addr):
-    return _ida.Jump(addr)
+    # uses C++ api instead of idc one to avoid activating the IDA window
+    return _ida.jumpto(addr, -1, 0)
 
 
 @withIDA

### CODE REVIEW

File: pwndbg/ida.py [hunk 1]
```diff
@@ -227,7 +227,8 @@ def here():
 @withIDA
 @takes_address
 def Jump(addr):
-    return _ida.Jump(addr)
+    # uses C++ api instead of idc one to avoid activating the IDA window
```
disconnect3d:
```suggestion
    # uses C++ api instead of idc one to avoid activating the IDA window (see #629)
```

PR #630, opened 2019-05-03
Show context of pseudocodes instead whole of them

NyaMisty:


ida_script.py
```diff
@@ -115,6 +115,21 @@ def decompile(addr):
...
+    if cfunc is None:
+        return None
+    item = cfunc.body.find_closest_addr(addr)
+    y_holder = idaapi.int_pointer()
+    if not cfunc.find_item_coords(item, None, y_holder):
+        return cfunc
+    y = y_holder.value()
+    lines = cfunc.get_pseudocode()
+    retlines = [idaapi.tag_remove(lines[lnnum].line) for lnnum
+                    in range(max(0, y - context_lines),min(len(lines), y + context_lines))]
```
disconnect3d: Can we change this list comprehension to a generator expression (replace `[]` with `()`).

This way we won't create a list just to join it later on :)

NyaMisty: Oh sorry I'm not that familiar with the generator expression (Sorry for my ignorance :( )
Could you please help me change it or give me more details?

disconnect3d:
Sure, just put this into () instead of []
(I for I in ...) Will create a generator while [I for I in ...] Creates a
list

pt., 3 maj 2019, 1:51 PM użytkownik NyaMisty <notifications@github.com>
napisał:

ida_script.py
disconnect3d: Sure, just change `[]` to `()`.

```
In [1]: [i for i in [1, 2, 3]]
Out[1]: [1, 2, 3]

In [2]: type([i for i in [1, 2, 3]])
Out[2]: list

In [3]: (i for i in [1, 2, 3])
Out[3]: <generator object <genexpr> at 0x105694a98>

In [4]: x = (i for i in [1, 2, 3])

In [5]: for i in x: print(i)
1
2
3
```

with a generator you don't create a list in memory and just do the computation of next element each iteration instead

NyaMisty:
Oh that's magical! I've pushed a new commit

ida_script.py
NyaMisty: Wow that's quite useful. Thanks for pointing that out for me :D

PR #631, opened 2019-05-04
Improve UX of dumpargs command

disconnect3d:
- fixed incorrect docs of dumpargs command (`--all` vs `--force`)
- dumpargs command does not require `--force` now if we are not on a call instruction


stnevans (approved):
Looks like a good change.

Issue #632, opened 2019-05-07
Overwriting up and down commands breaks list command


disconnect3d:
The `up` and `down` commands can be used to move between stack frames. We currently overwrite them in `commands/ida.py` in order to trigger IDA Pro focus:
https://github.com/pwndbg/pwndbg/blob/a46052c4efe42b01a61c0b04dd8d231a8b3b0179/pwndbg/commands/ida.py#L43-L85

While we debug a binary with source code available and we use `up` and `down` commands, both `context` and `info frame` looks good - as if we really switch between stack frames. However, the `list` GDB command still shows the source code from the most lower stack frame.

I tracked that this issue is caused entirely by overwriting those commands - if I change the name of our `up` to `upx` and use just `up`, everything works as expected.

disconnect3d:
I have tried implementing our `up` command as:
```python
gdb.execute('up %s' % arg)
j()
```
to execute the real `up` command, but it doesn't work - it ends up with a recursion error.

disconnect3d:
Oh, another thing we do during `up` and `down` is to display our own backtrace e.g.:
```
pwndbg> up
   f 0     7ffff697aa1a ffi_closure_unix64_inner+90
 ► f 1     7ffff697af16 ffi_closure_unix64+70
   f 2     5555556316f6 PyObject_GetAttr+40
   f 3     5555556d3f6f _PyEval_EvalFrameDefault+33906
   f 4     5555556ca265 PyEval_EvalFrameEx+18
   f 5     5555556cadb7 _PyEval_EvalCodeWithName+2566
   f 6     5555556cb412 PyEval_EvalCodeEx+58
   f 7     5555556cb440 PyEval_EvalCode+36
   f 8     5555555b00fb run_mod+48
   f 9     5555555b31dd PyRun_FileExFlags+154
   f 10     5555555b35b9 PyRun_SimpleFileExFlags+851
   f 11     5555555b37a6 PyRun_AnyFileExFlags+99
```

instead of the GDB's default one.

disconnect3d:
```
pwndbg> set debug-events on
Set display internal event debugging info to True
'before_prompt' pwndbg.memoize.__reset_on_prompt ()
pwndbg> up
   f 0     7ffff697aa1a ffi_closure_unix64_inner+90
   f 1     7ffff697af16 ffi_closure_unix64+70
 ► f 2     5555556316f6 PyObject_GetAttr+40
[...]

PR #633, opened 2019-05-08
Workaround for GDB bug described in #632

disconnect3d:
GDB's `up` and `down` commands trigger internal notification about changed frame. It does not happen for `gdb.Frame.select()` which we use in our own overrides for `up` and `down` commands.

Because of that, the `list` GDB command does not show proper source code lines.

This can be worked around by firing `frame` command and this is what this workaround/PR adds.

This bug has also been reported to GDB bugzilla at https://sourceware.org/bugzilla/show_bug.cgi?id=24534

For more details, see #632

### DIFF
--- a/pwndbg/commands/ida.py
+++ b/pwndbg/commands/ida.py
@@ -54,6 +54,9 @@ def up(n=1):
             f = f.older()
     f.select()
 
+    # workaround for #632
+    gdb.execute('frame', to_string=True)
+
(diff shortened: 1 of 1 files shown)


disconnect3d:
Before this PR:
image


After this PR:
image

stnevans:
Looks like a good workaround. 
Interesting bug, good job hunting down the root cause and finding a workaround!

PR #635, opened 2019-05-08
Invoke heap commands only with libc debug symbols.

mzr:
Added decorator which prevents invoking heap-related commands without libc debug symbols.

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -238,6 +238,14 @@ def OnlyWhenHeapIsInitialized(function):
             print("%s: Heap is not initialized yet." % function.__name__)
     return _OnlyWhenHeapIsInitialized
 
+def OnlyWithLibcDebugSyms(function):
+    @functools.wraps(function)
+    def _OnlyWithLibcDebugSyms(*a, **kw):
+        if pwndbg.heap.current.libc_has_debug_syms():
+            return function(*a, **kw)
+        else:
+            print("%s: This command only works with libc debug symbols." % function.__name__)
+    return _OnlyWithLibcDebugSyms
 
 class QuietSloppyParsedCommand(ParsedCommand):
     def __init__(self, *a, **kw):
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -244,7 +244,9 @@ def OnlyWithLibcDebugSyms(function):
         if pwndbg.heap.current.libc_has_debug_syms():
             return function(*a, **kw)
         else:
-            print("%s: This command only works with libc debug symbols." % function.__name__)
+            print('''%s: This command only works with libc debug symbols.
+They can probably be installed via the package manager of your choice.
+See also: https://sourceware.org/gdb/onlinedocs/gdb/Separate-Debug-Files.html''' % function.__name__)
     return _OnlyWithLibcDebugSyms
 
 class QuietSloppyParsedCommand(ParsedCommand):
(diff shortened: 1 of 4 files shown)


pwndbg/commands/__init__.py
```diff
@@ -238,6 +238,14 @@ def _OnlyWhenHeapIsInitialized(*a, **kw):
             print("%s: Heap is not initialized yet." % function.__name__)
     return _OnlyWhenHeapIsInitialized
 
+def OnlyWithLibcDebugSyms(function):
+    @functools.wraps(function)
+    def _OnlyWithLibcDebugSyms(*a, **kw):
+        if pwndbg.heap.current.libc_has_debug_syms():
+            return function(*a, **kw)
+        else:
+            print("%s: This command only works with libc debug symbols." % function.__name__)
```
disconnect3d: We probably should be a bit more verbose/explicit here: e.g. what people can do in this case?

1) we could inform them about some packages they could install - however this is very distro-specific
2) we could create a markdown page describing this stuff (e.g. what to do on a given distro or how to add debug symbols to your libc?)

We potentially also want to display the libc that is being loaded in the binary - at least I think that would be awesome here. @mzr do you think it is possible?

mzr: I think we could link to [gdb docs](https://sourceware.org/gdb/onlinedocs/gdb/Separate-Debug-Files.html). IMO it's well explained there.

When it comes to automatic identification of libc version, we can use 
`curl -L -X GET 'https://libc.blukat.me?q=__libc_start_main_ret:e81,_IO_2_1_stdin_:5c0'` and extract information from response.

disconnect3d: Let's extend it as:
```
%s: This command only works with libc debug symbols.
They can probably be installed via the package manager of your choice.
See also: https://sourceware.org/gdb/onlinedocs/gdb/Separate-Debug-Files.html
```

mzr: Done.

elongl:
This doesn't seem to work so well,
I've installed all the packages needed and I'm still prompted with:
```
pwndbg> heap
heap: This command only works with libc debug symbols.
They can probably be installed via the package manager of your choice.
See also: https://sourceware.org/gdb/onlinedocs/gdb/Separate-Debug-Files.html

E.g. on Ubuntu/Debian you might need to do the following steps (for 64-bit and 32-bit binaries):
sudo apt-get install libc6-dbg
sudo dpkg --add-architecture i386
sudo apt-get install libc-dbg:i386
```

mzr:
Yeaaaah. I've come across something similar. There are some issues across newer ubuntu distros (18.04, 20.04) and packages with libc32 debugging symbols. As far as I remember, they are named differently or even are no longer available in the distro repository.

disconnect3d:
Hm.. any better idea for the libc debug check than `        return pwndbg.symbol.address('global_max_fast') is not None` then?

elongl:
@mzr 
Pretty sure they do exist.
I'm using 20.04 LTS (Focal Fossa).
```
Reading package lists... Done
Building dependency tree       
Reading state information... Done
libc6-dbg is already the newest version (2.31-0ubuntu9).
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Note, selecting 'libc6-dbg:i386' instead of 'libc-dbg:i386'
libc6-dbg:i386 is already the newest version (2.31-0ubuntu9).
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
```

PR #637, opened 2019-05-08
Fixes #636 - bug with regs display on other frames

### LINKED ISSUE #636: Context displays wrong registers values for frames other than the newest one
There are two bugs regarding displaying of registers:
1. Wrong registers read
2. Improper caching

### Wrong registers read

We read registers from `newest_frame` instead of `selected_frame` - as a result, we display registers not from the current frame. This happens only for GDB>=7.9 as can be seen below.

To see this - break somewhere in a deep call, then do `up 2`, then display `context regs` and compare IP/PC with `info reg pc`.

Code responsible for the issue:
https://github.com/pwndbg/pwndbg/blob/a729a82fe44be9d0f9bb1ad1e732ee8c4fd6b7f5/pwndbg/regs.py#L246-L252

Some example:
image

### Improper caching

I have fixed the 1st issue locally but apparently the context display wasn't right. When I did `pi pwndbg.memoize.memoize.caching = False` and  then `context regs` it has been fixed.

So I investigated this and found out that we have two (or more?) methods to fetch registers:
- `__getattr__`
- `__getitem__`

This can be seen here:
https://github.com/pwndbg/pwndbg/blob/a729a82fe44be9d0f9bb1ad1e732ee8c4fd6b7f5/pwndbg/regs.py#L269-L306

Not that far ago, we had some problems with caching and [we added the `@pwndbg.memoize.reset_on_prompt` cache invalidation for `__getattr__`](
https://github.com/pwndbg/pwndbg/commit/7815953781dd9744309b539532ad889bd125c93d).

It turns out, that some registers are fetched via `__getitem__` and cached and this bites us here: it seems the context register values are fetched, cached and not invalidated this way.

So to fix that, we need to add `@pwndbg.memoize.reset_on_prompt` to `__getitem__`.

**One could ask if we should have both methods at the first place - or - why do we have both of them?** I don't know. We should probably remove one of them and refactor other places to use just the other one. But this can be done in the future. If someone is willing to refactor this, please do so :).

disconnect3d:
TLDR:
1. We read registers from `newest_frame` instead of `selected_frame` for GDB>=7.9.
2. We have two ways to fetch registers - `regs.__getitem__` and `regs.__getattr__` - one of them didn't invalidate cache and so after fixing 1st, we still shown the old register after switching frames.

### DIFF
--- a/pwndbg/regs.py
+++ b/pwndbg/regs.py
@@ -249,7 +249,7 @@ def gdb77_get_register(name):
 
 @pwndbg.proc.OnlyWhenRunning
 def gdb79_get_register(name):
-    return gdb.newest_frame().read_register(name)
+    return gdb.selected_frame().read_register(name)
 
 try:
     gdb.Frame.read_register
@@ -285,6 +285,7 @@ class module(ModuleType):
             return None
 
     @pwndbg.memoize.reset_on_stop
+    @pwndbg.memoize.reset_on_prompt
     def __getitem__(self, item):
         if isinstance(item, six.integer_types):
             return arch_to_regs[pwndbg.arch.current][item]


stnevans (approved):

stnevans:
Looks like a nice fix. I didn't even realize that we intended the selected frame to have an effect on the registers, but I suppose it makes sense for things like $pc, $sp and such. 
I guess it's nice that `context` now shows the source relevant to the frame.

PR #638, opened 2019-05-09
Fix source code display

disconnect3d:
Fixes and simplifies source code display in `context code`. The reasoning below.

---

It turns out that we determine the source code line for known code in a tricky and buggy way. It is wrong when we are deep inside many calls and we go to upper frame via e.g. `up` and then display the `context code`.

And this also occurs after applying the PR #637 before/after this fix.

We also do `except` all exceptions happening in the source code retrieval which makes it hard to spot potential bugs in this code path.

This commit removes the `except` path and checks for particular edge cases where we wouldn't have sources available.

---

Also note the `FileNotFoundError` path - in this case the debug symbols have a source file path there, but it hasn't been found by us. It might be because the program (or anything else) removed it OR maybe we debug remotely etc (not sure if this is the case).

This being said, we could potentially inform the users about that BUT... GDB already does that by itself - maybe in a bit vague way, as it does it before us displaying the context e.g.:

```
pwndbg> r
Starting program: /home/dc/pwndbg/a.out

Program received signal SIGSEGV, Segmentation fault.
0x000055555555460a in foo () at ./a.c:2
2	./a.c: No such file or directory.
// (the rest of the context here)
```

But I am fine with that for now - and I am not sure if we could even suppress this warning (oh, maybe this string comes to us into prompt hook/event? I don't know)

### DIFF
--- a/pwndbg/color/syntax_highlight.py
+++ b/pwndbg/color/syntax_highlight.py
@@ -65,7 +65,7 @@ def syntax_highlight(code, filename='.asm'):
 
     if not lexer:
         try:
-            lexer = pygments.lexers.guess_lexer_for_filename(filename, code)
+            lexer = pygments.lexers.guess_lexer_for_filename(filename, code, stripnl=False)
         except pygments.util.ClassNotFound:
             # no lexer for this file or invalid style
             pass
@@ -73,10 +73,6 @@ def syntax_highlight(code, filename='.asm'):
     if lexer:
         lexer_cache[filename] = lexer
 
-        # Count the number of newlines at start of the code
-        # As for some reason pygments.highlight remove them, we need to bring them back
-        # So that if we use this later for displaying code+code lines, they match
-        newlines_at_start = sum(1 for _ in itertools.takewhile(lambda c: c == '\n', code))
         code = pygments.highlight(code, lexer, formatter).rstrip()
 
-    return '\n' * newlines_at_start + code
+    return code
--- a/pwndbg/color/syntax_highlight.py
+++ b/pwndbg/color/syntax_highlight.py
@@ -4,6 +4,7 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import itertools
 import os.path
 import re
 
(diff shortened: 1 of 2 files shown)


disconnect3d:
Also note that we end up using `gdb.selected_frame().find_sal().line` here to get the line code instead of doing it via finding the closest (and also latest) line that is close to PC of current frame (which is buggy as PC of an upper frame that makes a CALL ends up being set after the CALL).

It might be that we used the previous method because there was no `.find_sal().line` in older GDBs or that there is an edge case that I didn't fall into.

For older GDBs: I wouldn't care. I think we should drop support for GDB<8.0 at some point.

disconnect3d:
The second commit I made fixes another small issue - the code lines are wrong when the source startswith newlines. This is because `pygments.highlight` removes starting newlines (and maybe ending, but those doesn't matter).

This can be observed below on this file (note that I first show first 3 lines of it and later [lines 892-902 via sed](https://stackoverflow.com/a/6022441/5700291)):
```
dc@dc:~/pwndbg$ head  /home/dc/projects/cpython/Objects/object.c -n3

/* Generic object operations; and implementation of None */

dc@dc:~/pwndbg$ sed -n '892,902p' < /home/dc/projects/cpython/Objects/object.c
                     "attribute name must be string, not '%.200s'",
                     name->ob_type->tp_name);
        return NULL;
    }
    if (tp->tp_getattro != NULL)
        return (*tp->tp_getattro)(v, name);
    if (tp->tp_getattr != NULL) {
        char *name_str = PyUnicode_AsUTF8(name);
        if (name_str == NULL)
            return NULL;
        return (*tp->tp_getattr)(v, name_str);
```


Before 2nd commit:
image

After 2nd commit:
image

pwndbg/color/syntax_highlight.py
```diff
@@ -71,6 +72,11 @@ def syntax_highlight(code, filename='.asm'):
...
         lexer_cache[filename] = lexer
+
+        # Count the number of newlines at start of the code
+        # As for some reason pygments.highlight remove them, we need to bring them back
+        # So that if we use this later for displaying code+code lines, they match
+        newlines_at_start = sum(1 for _ in itertools.takewhile(lambda c: c == '\n', code))
         code = pygments.highlight(code, lexer, formatter).rstrip()
 
-    return code
+    return '\n' * newlines_at_start + code
```
stnevans: Instead of using itertools and this to fix the newlines problem, I think we should use stripnl=False when creating the lexer to fix this problem. (http://pygments.org/docs/lexers/)

disconnect3d: That works perfectly, thanks!

PR #641, opened 2019-05-10
Fix remote debugging via pwntools

disconnect3d:
When we debug something remotely we add symbols from the binary path.
It turns out that GDB does it by itself, so it is not really needed.
And it seems that when we do it by ourselves, the symbols sometimes end
up being wrong (e.g. we see a display of `func1+X` instead of `func2`).

### DIFF
--- a/pwndbg/symbol.py
+++ b/pwndbg/symbol.py
@@ -233,37 +233,5 @@ def address(symbol):
     except Exception:
         pass
 
-@pwndbg.events.stop
-@pwndbg.memoize.reset_on_start
-def add_main_exe_to_symbols():
-    if not pwndbg.remote.is_remote():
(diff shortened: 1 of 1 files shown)


disconnect3d:
The binary I am trying this on is not for public use, so I will try to create a minimal working example for this issue soon.

zachriggle:
Is this still the case for remote QEMU / cross-arch debugging targets?

**Edit**: I see we only do this for native arch anyway.

disconnect3d:
> Edit: I see we only do this for native arch anyway.

Yeah. Sorry for the delay here. I will push this further next week.

disconnect3d:
Closing as of now, will probably revisit at some time.

PR #643, opened 2019-05-10
Add history command

vesim987:


stnevans:
@vesim987 Thanks for the pr!
I have a couple of questions/comments. 
1) Is this significantly different than gdb's `show commands`?
2) For any new commands I think we should add corresponding information to the docs. 
3) Looks like travis is upset about you not sorting the imports. Could you fix that?

disconnect3d:
@vesim987 bump?

disconnect3d:
I spoke with Vesim on DM. The initial idea of this command was mine - I wanted to be able to easily copy history to scripts. I don't remember if I was aware of `show commands` that time.

The differences:
* `show commands` display always only 10 commands and shows line numbers:
```
pwndbg> show commands
  264  show commands 0-
  265  show commands 0 500
  266  show commands 0-500
  267  show commands
  268  show commands
  269  help show commands
  270  show commands +
  271  show commands ++
  272  show commands +500
  273  show commands
```
* the `history` command would let you see more than 10 entries

Related: https://ftp.gnu.org/old-gnu/Manuals/gdb/html_node/gdb_182.html
and:
```
pwndbg> help show commands
Show the history of commands you typed.
You can supply a command number to start with, or a `+' to start after
the previous command number shown.
```

stnevans:
Being able to see more than 10 commands would be nice, I have no idea why gdb has that restriction.
I'm fine adding the command, but it would be nice if we could fix travis and add it to the docs first.

disconnect3d:
@vesim987 ping once again :(

PR #648, opened 2019-05-15
Fix #646 (Allow xuntil to work with PIE binaries on first run)

### LINKED ISSUE #646: xuntil doesn't work with PIE binaries before being run
`xuntil` was added in #604 and works except that it doesn't work with PIE binaries before being run.
This is because it expects an int argument. I'll look into using sloppy parsing.

stnevans:
This is a somewhat ugly fix in that it makes the argument a string rather than a gdb.Value. If it's a string and is a symbol, it uses gdb.Breakpoint(symbol) rather than trying to resolve an address.

I couldn't find any nicer way to do this.

### DIFF
--- a/pwndbg/commands/peda.py
+++ b/pwndbg/commands/peda.py
@@ -30,18 +30,27 @@ def getpid():
 
 
 parser = argparse.ArgumentParser(description='Continue execution until an address or function.')
-parser.add_argument('target', help='Address or function to stop execution at')
-
+parser.add_argument('target', type=str, help='Address or function to stop execution at')
 
 @pwndbg.commands.ArgparsedCommand(parser)
 def xuntil(target):
-    addr = int(target)
+    try:
+        addr = int(target,0)
+        
+        if not pwndbg.memory.peek(addr):
+            print(message.error('Invalid address %#x' % addr))
+            return
 
-    if not pwndbg.memory.peek(addr):
-        print(message.error('Invalid address %#x' % addr))
-        return
+        spec = "*%#x" % (addr)
+    except (TypeError, ValueError):
+        #The following gdb command will throw an error if the symbol is not defined.
+        try:
+            result = gdb.execute('info address %s' % target, to_string=True, from_tty=False)
+        except gdb.error:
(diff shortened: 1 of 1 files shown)


disconnect3d:
I wanted to suggest using `pwndbg.symbol.address(symbol_str)` here, but it doesn't work with unresolved symbols heh.

I haven't played that much with it but can't we do sth like this here?
```python
gdb.Breakpoint('*%#x' % int(gdb.parse_and_eval(passed_string)))
```
and then put it in `try ... except` and display it couldn't be parsed?

EDIT: It might be cleaner to try to parse and then just set bp instead of trying to hit the memory first too(?)

stnevans:
> I wanted to suggest using `pwndbg.symbol.address(symbol_str)` here, but it doesn't work with unresolved symbols heh.
> I haven't played that much with it but can't we do sth like this here?
> 
> ```python
> gdb.Breakpoint('*%#x' % int(gdb.parse_and_eval(passed_string)))
> ```
> 
> and then put it in `try ... except` and display it couldn't be parsed?

So the point of the pr is to make sure that `xuntil` works with symbols before a pie binary is run. If you try to directly resolve the symbol before the binary is run, gdb will give you and offset rather than an absolute address. (it would be nice if there was a way to specify offset, but I don't know of a way to).
It doesn't cause any errors directly to do gdb.Breakpoint("*0x1234"), but when the binary is run, the bp will be at that address rather than that offset.

> EDIT: It might be cleaner to try to parse and then just set bp instead of trying to hit the memory first too(?)

Can you elaborate on this. Do you mean allow breaking at addresses that aren't currently mapped?

disconnect3d:
I have read the code carefully and now I understand the thing. So we either do `gdb.Breakpoint(address)` or `gdb.Breakpoint(symbol)`.

And yeah, this won't work as this resolves to offset which is not a valid absolute address:
```
>>> gdb.Breakpoint('*%#x' % int(gdb.parse_and_eval('main')))
Breakpoint 2 at 0x64a
```

disconnect3d (approved):

PR #649, opened 2019-05-26
typeinfo: more types for golang programs

StalkR:
Hey,

I was looking at a golang binary in pwndbg and got these:
```
  File "/opt/pwndbg/pwndbg/typeinfo.py", line 54, in update
    module.ushort = lookup_types('unsigned short', 'ushort', 'u16')
  File "/opt/pwndbg/pwndbg/typeinfo.py", line 43, in lookup_types
    raise exc
  File "/opt/pwndbg/pwndbg/typeinfo.py", line 40, in lookup_types
    return gdb.lookup_type(type_str)
gdb.error: No type named u16.
```

Associated patch fixes them and lets me continue debugging nicely so far.

### DIFF
--- a/pwndbg/typeinfo.py
+++ b/pwndbg/typeinfo.py
@@ -52,18 +52,18 @@ def update():
     module.ulong  = lookup_types('unsigned long', 'uint', 'u32')
     module.long   = lookup_types('long', 'int', 'i32')
     module.uchar  = lookup_types('unsigned char', 'ubyte', 'u8')
-    module.ushort = lookup_types('unsigned short', 'ushort', 'u16')
+    module.ushort = lookup_types('unsigned short', 'ushort', 'u16', 'uint16')
     module.uint   = lookup_types('unsigned int', 'uint', 'u32')
     module.void   = lookup_types('void', '()')
     module.uint8  = module.uchar
     module.uint16 = module.ushort
     module.uint32 = module.uint
-    module.uint64 = lookup_types('unsigned long long', 'ulong', 'u64')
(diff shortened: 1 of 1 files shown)


StalkR:
Just spotted in https://github.com/pwndbg/pwndbg/blob/dev/DEVELOPING.md

> The known bug with that is that it might not work properly for Golang binaries compiled with debugging symbols.

Well works for me now, if you want I can remove the sentence in this PR - let me know.

disconnect3d:
> Well works for me now

While it works fine in your particular case, I run into some 'hard to fix' cases, that's why I stated it like this. I regret I haven't made a particular issue for that attaching the binaries I used :(. 

We would need to test such things with both 32-bit and 64-bit binaries, probably best if each binary used different type, so we can see if we get those lookup symbols right...

Anyway, I like the PR so I am merging this. Could you make some tests and open an issue writing them up or adding them to our CI (I am not sure about our testing framework)?

Thanks for the PR!

StalkR:
Ok, I'll look into adding tests.

This is what I tested so far:
```
echo -e 'package main\n\nfunc main() {}' > test.go
test() { gdb -nx -ex 'source /opt/pwndbg/gdbinit.py' -ex 'b main.main' -ex 'r' -ex 'context' --batch test; }
GOARCH=amd64 go build test.go
test
GOARCH=386 go build test.go
test
```

Turns out switching machines/environment and it broke again because this PR was too conservative, only adding the types that were missing and enough to continue, while I should probably have added more (e.g. uint8 for u8) just in case. I'll prepare another PR to fix that and add tests.

PR #651, opened 2019-05-29
typeinfo: more types for golang and tests

ailg:
Continuing PR #649.

Commit 493b1a0 added enough types so it works for Go binaries that are dynamic instead of static, linked with libc, which happens when you depend on `net` or `os/user`.
However I noticed that with a dummy Go program like `package main; func main() { }`, pwndbg is unable to resolve more types like `u8`.

This PR adds those, as well as defensively more Go types in case it's ever needed.

I added tests this time, including two test Go binaries (for x86-32 and x86-64, it doesn't seem to matter but for completeness since cross-compiling is so easy).
The test catches the output of `start` seeing if there was no exception, let me know if there's a better method I could call instead.


ailg:
wrong account :)

PR #652, opened 2019-05-29
typeinfo: more types for golang and tests

StalkR:
Continuing PR #649.

Commit 493b1a0 added enough types so it works for Go binaries that are dynamic instead of static, linked with libc, which happens when you depend on `net` or `os/user`.
However I noticed that with a dummy Go program like `package main; func main() { }`, which ends up static not dynamic depending on libc, pwndbg is unable to resolve more types like `u8`. Also, only with debug symbols i.e. if not stripped; when stripped, gdb is fine (going back to default types?).

This PR adds those, as well as defensively more Go types in case it's ever needed.

I added tests this time, including two test Go binaries (for x86-32 and x86-64, it doesn't seem to matter but for completeness since cross-compiling is so easy).
The test catches the output of `start` seeing if there was no exception, let me know if there's a better method I could call instead.

### DIFF
--- a/pwndbg/typeinfo.py
+++ b/pwndbg/typeinfo.py
@@ -49,20 +49,20 @@ def lookup_types(*types):
 def update():
 
     module.char   = gdb.lookup_type('char')
-    module.ulong  = lookup_types('unsigned long', 'uint', 'u32')
-    module.long   = lookup_types('long', 'int', 'i32')
-    module.uchar  = lookup_types('unsigned char', 'ubyte', 'u8')
+    module.ulong  = lookup_types('unsigned long', 'uint', 'u32', 'uint32')
+    module.long   = lookup_types('long', 'int', 'i32', 'int32')
+    module.uchar  = lookup_types('unsigned char', 'ubyte', 'u8', 'uint8')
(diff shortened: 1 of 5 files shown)


stnevans:
Looks good to me. It's also nice to see some tests!
One comment: do we need to add the binaries to the repo? It would be nicer to just have the source and the makefile.

StalkR:
> One comment: do we need to add the binaries to the repo?

No, removed. Travis can recompile them, I did it because I saw the other ones were added:
- https://github.com/pwndbg/pwndbg/blob/dev/tests/binaries/emulate_disasm.out
- https://github.com/pwndbg/pwndbg/blob/dev/tests/binaries/emulate_disasm_loop.out
- https://github.com/pwndbg/pwndbg/blob/dev/tests/binaries/reference-binary.out

Created #653 to clean them up.

There are also these two but unlike the previous ones there's no `Makefile`, so it makes sense to keep binaries as special repro test cases:
- https://github.com/pwndbg/pwndbg/blob/dev/tests/binaries/old_bash/binary
- https://github.com/pwndbg/pwndbg/blob/dev/tests/binaries/old_bash/core

disconnect3d:
Nice, thanks a lot for this! :)

PR #653, opened 2019-05-30
tests/binaries: remove binaries generated by Makefile

StalkR:
Spotted in #652, tests binaries are committed but we don't need to because there's a `Makefile` to generate them:
- https://github.com/pwndbg/pwndbg/blob/dev/tests/binaries/emulate_disasm.out
- https://github.com/pwndbg/pwndbg/blob/dev/tests/binaries/emulate_disasm_loop.out
- https://github.com/pwndbg/pwndbg/blob/dev/tests/binaries/reference-binary.out

cc @stnevans


stnevans:
It looks like this doesn't work because when testing emulate and the related commands, we assert that the output matches predetermined assembly. 
The only way I can think of to remove these binaries without breaking the test would be to write raw assembly files and use an assembler. I don't have time for this right now, and until anyone decides to do so, it seems like it makes the most sense to just leave the binaries for now.

If you can think of any other way to do this, feel free. Thanks for trying to remove the binaries, it's would be nice if we could b.c. the typical install just clones this repo and having everyone download them seems silly.

disconnect3d:
> It looks like this doesn't work because when testing emulate and the related commands, we assert that the output matches predetermined assembly.

Yeah, that's the reason for that.

> The only way I can think of to remove these binaries without breaking the test would be to write raw assembly files and use an assembler.
I think more precisely it would be a linker script too, to put stuff on a given address in _all builds_ (either on travis or people machines).

I'd be happy if we had that, but having binaries in the repo is okay for now.

StalkR:
Sounds good! I mainly tried see what travis would say, and indeed it complained.

PR #654, opened 2019-06-02
Add initial support for ARM Cortex-M baremetal debugging (#264)

jaseg:
This commit adds support for xPSR to be used instead of cPSR on Cortex-M. Prior to this commit, pwndbg would crash when attached to a Cortex-M core.

This patch adds a new internal arch "armcm" to describe Cortex-M ARMs.

Closes #542, fixes the second issue in #264

### DIFF
--- a/pwndbg/arch.py
+++ b/pwndbg/arch.py
@@ -26,9 +26,15 @@ native_endian = str(sys.byteorder)
 
 
 def fix_arch(arch):
-    arches = ['x86-64', 'i386', 'mips', 'powerpc', 'sparc', 'arm', 'aarch64', arch]
-    return next(a for a in arches if a in arch)
+    for match in ['x86-64', 'i386', 'mips', 'powerpc', 'sparc', 'aarch64']:
+        if match in arch:
+            return match
 
+    # Distinguish between Cortex-M and other ARM
+    if 'arm' in arch:
+        return 'armcm' if '-m' in arch else 'arm'
+
+    return arch
 
 @pwndbg.events.start
 @pwndbg.events.new_objfile
--- a/pwndbg/color/context.py
+++ b/pwndbg/color/context.py
@@ -54,3 +54,25 @@ def banner(x):
 
 def banner_title(x):
     return generateColorFunction(config.banner_title_color)(x)
+
+def format_flags(value, flags, last=None):
+    desc = flag_value('%#x' % value)
+    if not flags:
+        return desc
+
+    names = []
+    for name, bit in flags.items():
+        bit = 1<<bit
+        if value & bit:
+            name = name.upper()
+            name = flag_set(name)
+        else:
+            name = name.lower()
+            name = flag_unset(name)
+
+        if last is not None and value & bit != last & bit:
+            name = flag_changed(name)
+        names.append(name)
+
+    return '%s %s %s %s' % (desc, flag_bracket('['), ' '.join(names), flag_bracket(']'))
+
--- a/pwndbg/disasm/arm.py
+++ b/pwndbg/disasm/arm.py
@@ -46,12 +46,12 @@ class DisassemblyAssistant(pwndbg.disasm.arch.DisassemblyAssistant):
         if instruction.address != pwndbg.regs.pc:
             return False
 
-        cpsr = pwndbg.regs.cpsr
+        value = pwndbg.regs.cpsr if pwndbg.arch.current == 'arm' else pwndbg.regs.xpsr
 
-        N = cpsr & (1<<31)
-        Z = cpsr & (1<<30)
-        C = cpsr & (1<<29)
-        V = cpsr & (1<<28)
+        N = value & (1<<31)
+        Z = value & (1<<30)
+        C = value & (1<<29)
+        V = value & (1<<28)
 
         cc = {
             ARM_CC_EQ: Z,
--- a/pwndbg/regs.py
+++ b/pwndbg/regs.py
@@ -97,12 +97,25 @@ class RegisterSet(object):
         for r in self.all:
             yield r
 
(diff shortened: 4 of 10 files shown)


pwndbg/regs.py
```diff
@@ -276,6 +288,8 @@ def __getattr__(self, attr):
                 value = gdb77_get_register(attr)
                 value = value.cast(pwndbg.typeinfo.uint32)
             else:
+                if attr.lower() == 'xpsr':
+                    attr = 'xPSR'
```
jaseg, line 294: This special case was necessary for me to make this work on GDB8.3, which seems to be case-sensitive for this register.

pwndbg/regs.py
```diff
@@ -97,12 +97,23 @@ def __iter__(self):
         for r in self.all:
             yield r
 
+arm_cpsr_flags = { 'N': 31, 'Z': 30, 'C': 29, 'V': 28, 'Q': 27, 'J': 24, 'E': 9, 'A': 8, 'I': 7, 'F': 6, 'T': 5 }
+arm_xpsr_flags = { 'N': 31, 'Z': 30, 'C': 29, 'V': 28, 'Q': 27, 'T': 24, }
```
jaseg: The xpsr flags listed here ignore things like IRQn and IT fields which would only clutter context output.

pwndbg/disasm/arm.py
```diff
@@ -46,12 +46,12 @@ def condition(self, instruction):
...
+        value = pwndbg.regs.cpsr if pwndbg.arch.current == 'arm' else pwndbg.regs.xpsr
 
-        N = cpsr & (1<<31)
-        Z = cpsr & (1<<30)
-        C = cpsr & (1<<29)
-        V = cpsr & (1<<28)
+        N = value & (1<<31)
+        Z = value & (1<<30)
+        C = value & (1<<29)
+        V = value & (1<<28)
```
jaseg, line 54: These bit offsets are the same in cpsr and xpsr.

pwndbg/color/context.py
```diff
@@ -54,3 +54,25 @@ def banner(x):
...
+        else:
+            name = name.lower()
+            name = flag_unset(name)
+
+        if last is not None and value & bit != last & bit:
+            name = flag_changed(name)
+        names.append(name)
+
+    return '%s %s %s %s' % (desc, flag_bracket('['), ' '.join(names), flag_bracket(']'))
+
```
jaseg, line 78: I factored out this since it can be re-used from both context and cpsr commands.

pwndbg/arch.py
```diff
@@ -26,9 +26,15 @@
...
 def fix_arch(arch):
-    arches = ['x86-64', 'i386', 'mips', 'powerpc', 'sparc', 'arm', 'aarch64', arch]
-    return next(a for a in arches if a in arch)
+    for match in ['x86-64', 'i386', 'mips', 'powerpc', 'sparc', 'aarch64']:
+        if match in arch:
+            return match
 
+    # Distinguish between Cortex-M and other ARM
+    if 'arm' in arch:
+        return 'armcm' if arch.endswith('-m') else 'arm'
```
jaseg: I'm not totally sure how watertight this check is, but I think this should work.

stnevans: So I'm not entirely sure, but I think that https://github.com/bminor/binutils-gdb/blob/502c64b9ac12cf2a35d3cb55c51e2eefd33a2494/bfd/cpu-arm.c#L223 is where gdb gets its (arm) architecture info from. 
Correct me if I'm wrong, but it looks like the cortex-m's would be:
```
  N (bfd_mach_arm_6M,        "armv6-m",        FALSE, & arch_info_struct[20]),
  N (bfd_mach_arm_6SM,       "armv6s-m",       FALSE, & arch_info_struct[21]),
  N (bfd_mach_arm_7EM,       "armv7e-m",       FALSE, & arch_info_struct[22]),
  N (bfd_mach_arm_8M_BASE,   "armv8-m.base",   FALSE, & arch_info_struct[25]),
  N (bfd_mach_arm_8M_MAIN,   "armv8-m.main",   FALSE, & arch_info_struct[26]),

```
If so, it looks like it might make more sense to do `return 'armcm' if '-m' in arch else 'arm'`

pwndbg/regs.py
stnevans: Good call. Showing too many flags can be too much. 
Thanks for adding the arm_cpsr_flags here, I had no idea that we didn't already have that.

Do you think it's useful to show 8,7,6 from cpsr? If showing them is regularly useful, it makes sense to have it, but otherwise it might just cause some clutter.

stnevans:
Looks great!
I left a few minor comments, but overall this seems like a nice fix to allow cortex-m debugging.

pwndbg/regs.py
jaseg: I can't think of a nice way to exclude them from ```context``` while showing them in ```cpsr```, so I'd leave them. In particular the interrupt disable flag sounds useful to me if you're breaking inside some interrupt handler on embedded. Maybe it'd make sense to use an OrderedDict for flags and just display them last?

pwndbg/arch.py
jaseg: You're right. I'll push a fixed commit.

pwndbg/regs.py
jaseg: Ok, I made them all OrderedDicts and that seems to work fine. This way AIF are at the back, and also J (24) and T (5) are grouped together on non-cortex-m.

stnevans:
The changes good to me. @disconnect3d any thoughts?

disconnect3d:
LGTM, that's nice, thanks!

PR #655, opened 2019-06-03
Add support for decompile_context on IDA 7.0

NyaMisty:
The find_item_coords and find_closest_addr function for cinsn_t aren't available for IDA 7.0, so I used a uglier approach, I get the first item of each line, and use that to indicate the whole line's ea. Then I find the line with the closest ea.

### DIFF
--- a/ida_script.py
+++ b/ida_script.py
@@ -119,11 +119,27 @@ def decompile_context(addr, context_lines):
     cfunc = decompile(addr)
     if cfunc is None:
         return None
-    item = cfunc.body.find_closest_addr(addr)
-    y_holder = idaapi.int_pointer()
-    if not cfunc.find_item_coords(item, None, y_holder):
-        return cfunc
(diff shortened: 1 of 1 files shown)


stnevans:
This looks good to me. I would merge it except I don't have a copy of IDA 7.0 to test with. 
Thanks for your continued work improving the pwndbg IDA integration!

zachriggle (changes requested):
I would prefer to pull this logic out into a separate routine, rather than capturing the logic for emulating this missing function directly inside `decompile_context`.

NyaMisty:
Done, I've splited the big ```decompile_context``` function into ```decompile_context``` and ```get_decompile_coord_by_ea```

NyaMisty:
(Small accident :P )

disconnect3d (approved):

disconnect3d:
Thanks! :)

PR #657, opened 2019-06-18
Add end separator

bet4it:
Hey,

I'm developing a debugger UI : [hyperpwn](https://github.com/bet4it/hyperpwn).
Currently it only supports `GEF`. Because I need an end separator to detect the end of context data.
After this pull request `hyperpwn` can also support `pwndbg`.
This is what it looks like after this pull request:
hyperpwn-pwndbg

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -110,6 +110,8 @@ def context(subcontext=None):
         func = context_sections.get(arg, None)
         if func:
             result.extend(func())
+    if len(result) > 0:
+        result.append(pwndbg.ui.banner(""))
     result.extend(context_signal())
 
     with output() as out:
--- a/pwndbg/ui.py
+++ b/pwndbg/ui.py
@@ -41,7 +41,8 @@ def check_title_position():
 def banner(title):
     title = title.upper()
     _height, width = get_window_size()
-    title = '%s%s%s' % (config.banner_title_surrounding_left, C.banner_title(title), config.banner_title_surrounding_right)
+    if title:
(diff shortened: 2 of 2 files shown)


l0kihardt:
What an amazing plugin!

disconnect3d:
Hey, thanks for the PR! It would be nice to have pwndbg in hyperpwn :). 

Does this add an additional line to the context? I think that adding a condition around `title` is a bit unintuitive. Can we do better, like explicitly adding something that would be an end marker for you?

bet4it:
Yes, what I need is only an additional line.
I think this commit is the most convenient way to do it, the `if` only deals with the situation I added.
Otherwise we may need something likes:
```python
def end_banner():
    _height, width = get_window_size()
    banner = width * config.banner_separator
    return C.banner(banner)
```

disconnect3d:
So does it  change the output when hyperpwn is not used?

Also, I think it's better to explicitly use `end_banner`, the intent here is more readable I guess?

bet4it:
Yes, it changes the normal output of `pwndbg`.
It's up to you whether to set it as an option, but at least `GEF` has an end separator all the time and nobody complains about it.

disconnect3d:
If someone is interested on how it looks like in normal pwndbg, it just adds the line after the backtrace: 
image

However, if someone sets `set context-sections ''` e.g. for scripting, they will also see the redundant line each time the program stops:
image

Can we change it so it displays only if there are sections to display?

bet4it:
Done.

zoniony:
beautiful  plugin!

disconnect3d:
Thanks again! :)

PR #658, opened 2019-06-18
Revert "Add history command (#643)"

mzr:
This reverts commit d1118f2cbbcad8754b6065dd0bc431512351c9b0.
This commit breaks tab autocompletion in GDB.


disconnect3d:
We looked at this a bit more with @vesim987 and @mzr privately.

It seems that the `import readline` breaks stuff by initializing readline once again.

We think this might be done because of this code:
https://github.com/python/cpython/blob/2ff58a24e8a1c7e290d025d69ebaea0bbead3b8c/Modules/readline.c#L1104-L1108

As they rebind tab back to insert a tab character instead of invoking autocompletion?

disconnect3d:
Merging this for now to fix the issue on `dev`.

We probably need to use e.g. `rl_bind_key` function or some other from there. It seems those are not exported to Python's `readline`. Maybe we could load the library through `ctypes.CDLL` and invoke the functions natively...

If someone would like to work on that, we would happily merge a PR that brings back the `history` command.

PR #663, opened 2019-07-23
Change the way to check is_remote

bet4it:
I find that `info file` will cause Host I/O every time I `stepi` when remote debugging, and can't be cached by `pwndbg.memoize.reset_on_exit`.

This way was borrowed from `GEF`.

### DIFF
--- a/pwndbg/remote.py
+++ b/pwndbg/remote.py
@@ -16,4 +16,10 @@ import pwndbg.memoize
 
 @pwndbg.memoize.reset_on_exit
 def is_remote():
+    # Example:
+    # pwndbg> maintenance print target-stack
+    # The current target stack is:
+    #   - remote (Remote serial target in gdb-specific protocol)
+    #   - exec (Local exec file)
+    #   - None (None)
     return "remote" in gdb.execute("maintenance print target-stack", to_string=True)
--- a/pwndbg/remote.py
+++ b/pwndbg/remote.py
@@ -11,23 +11,9 @@ from __future__ import unicode_literals
 
 import gdb
 
+import pwndbg.memoize
 
+
+@pwndbg.memoize.reset_on_exit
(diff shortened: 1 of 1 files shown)


pwndbg/remote.py
```diff
@@ -11,23 +11,9 @@
...
-
-    # target extended-remote
-    if 'Extended remote serial target' in info_file:
-        return True
 
-    if 'Debugging a target over a serial line.' in info_file:
-        return True
-
-    return False
+@pwndbg.memoize.reset_on_exit
```
disconnect3d, line 17: Is `reset_on_exit` correct here? I.e. if we connect to different target, this should be resetted too.

pwndbg/remote.py
```diff
@@ -11,23 +11,9 @@
...
+def is_remote():
+    return "remote" in gdb.execute("maintenance print target-stack", to_string=True)
```
disconnect3d, line 25: Can u add docstring with example output of the command?

E.g.
```
pwndbg> maintenance print target-stack
The current target stack is:
  - remote (Remote serial target in gdb-specific protocol)
  - exec (Local exec file)
  - None (None)
```

disconnect3d (changes requested):
Hey, thx for the PR! That sounds good but need polishing :).

pwndbg/remote.py
bet4it: I did the following test:
* Add a `print` in `is_remote` so we can know when it's called
* `gdb bash` and I notice `is_remote` is called
* In this gdb session, run `target remote 127.0.0.1:1234` to connect to a gdbserver, and I notice `is_remote` is called again

So I think `reset_on_exit` will also be invoked when we connect to a different target, we don't need to change anything.

disconnect3d (approved):

PR #670, opened 2019-08-17
Inthook size types (Fixes #669)

### LINKED ISSUE #669: Incorrect conversion from gdb.Value to python int
### Description

When debugging certain binary, pwndbg may show incorrect register value, missing high 32 bit for 64 bit register.
e.g.

```
pwndbg> regs rax
 RAX  0x555c4160
pwndbg> i r rax
rax            0x5555555c4160   93824992690528
pwndbg> telescope $rax
<Could not read memory at 0x555c4160>
pwndbg> regs rsp
 RSP  0xffffffffffffc2f0
pwndbg> i r rsp
rsp            0x7fffffffc2f0   0x7fffffffc2f0
pwndbg> telescope $rsp
<Could not read memory at 0xffffc2f0>
pwndbg>
```

**If I strip the binary, everything goes well.**

After checking code of pwndbg, I found that the bug happens when trying to convert a `gdb.Value` instance to python `int`.

```
pwndbg> python
>print(hex(pwndbg.regs.rsp))
>rsp = pwndbg.regs.value
>print(type(rsp))
>print(rsp.type)
>print(rsp)
>print(hex(rsp))
>print(int(rsp))
>print(long(rsp))
>print(str(rsp))
>print(hex(pwndbg.arch.ptrmask))
>print(hex(int(rsp) & pwndbg.arch.ptrmask))
>end
0xffffffffffffc2f0
<class 'gdb.Value'>
u64
140737488339696
0x7fffffffc2f0
-15632
-15632
140737488339696
0xffffffffffffffff
0xffffffffffffc2f0
```

Seems the `int(value)` works as converting to 32 bit c type int instead of python type int. Maybe we need to use `int(str(value))` to get the correct value.

### Steps to reproduce

Here is an example binary which can reproduce the bug.

[babi.zip](https://github.com/pwndbg/pwndbg/files/3508243/babi.zip)

1. Execute `babi` with gdb & break at `0x5555555D0C97`.
1. Execute the attached `test.py`.
1. Breakpoint will hit with the bug.

*PS: stop in `libc` or `main` function won't see the bug.*

### My setup

```
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 18.04.2 LTS
Release:        18.04
Codename:       bionic
```

```
pwndbg> version
Gdb:      8.1.0.20180409-git
Python:   3.6.8 (default, Jan 14 2019, 11:02:34)  [GCC 8.0.1 20180414 (experimental) [trunk revision 259383]]
Pwndbg:   1.1.0 build: 7f5d8e7
Capstone: 4.0.1024
Unicorn:  1.0.1
```

stnevans:
Various issues have been created from inthook relying on ulong and long. 
This changes it to rely on (now properly set) size_t and ssize_t. #669 should be fixed, and I think #644 although I haven't tested that one yet.

### DIFF
--- a/pwndbg/typeinfo.py
+++ b/pwndbg/typeinfo.py
@@ -75,11 +75,12 @@ def update():
         module.ptrdiff = module.uint32
         module.size_t = module.uint32
         module.ssize_t = module.int32
-    if pvoid.sizeof == 8: 
+    elif pvoid.sizeof == 8: 
         module.ptrdiff = module.uint64
         module.size_t = module.uint64
         module.ssize_t = module.int64
-
+    else:
+        raise Exception('Pointer size not supported')
     module.null = gdb.Value(0).cast(void)
 
 # Call it once so we load all of the types
--- a/pwndbg/typeinfo.py
+++ b/pwndbg/typeinfo.py
@@ -65,17 +65,20 @@ def update():
     module.int32  = lookup_types('int', 'i32', 'int32')
     module.int64  = lookup_types('long long', 'long', 'i64', 'int64')
 
-    module.ssize_t = module.long
-    module.size_t = module.ulong
-
     module.pvoid  = void.pointer()
     module.ppvoid = pvoid.pointer()
     module.pchar  = char.pointer()
 
     module.ptrsize = pvoid.sizeof
 
-    if pvoid.sizeof == 4: module.ptrdiff = uint32
-    if pvoid.sizeof == 8: module.ptrdiff = uint64
+    if pvoid.sizeof == 4: 
+        module.ptrdiff = module.uint32
+        module.size_t = module.uint32
+        module.ssize_t = module.int32
+    if pvoid.sizeof == 8: 
+        module.ptrdiff = module.uint64
+        module.size_t = module.uint64
+        module.ssize_t = module.int64
 
     module.null = gdb.Value(0).cast(void)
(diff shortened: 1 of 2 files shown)


pwndbg/typeinfo.py
```diff
@@ -65,17 +65,20 @@ def update():
...
 
     module.ptrsize = pvoid.sizeof
 
-    if pvoid.sizeof == 4: module.ptrdiff = uint32
-    if pvoid.sizeof == 8: module.ptrdiff = uint64
+    if pvoid.sizeof == 4: 
+        module.ptrdiff = module.uint32
+        module.size_t = module.uint32
+        module.ssize_t = module.int32
+    if pvoid.sizeof == 8:
```
disconnect3d:
```suggestion
    elif pvoid.sizeof == 8: 
```

disconnect3d: Can we also add an `else` with warning/error, just as a sanity check?

disconnect3d (changes requested):
++, see comments

pwndbg/typeinfo.py
stnevans: Good call. We could also potentially support 16 bit binaries here, although I'm not sure if we want to do that, especially without testing more.

Issue #678, opened 2019-10-20
Allow hexdump by module name, like vmmap 


korniltsev:
### Description

I find myself doing `vmmap heap + hexdump heap_start_addr` from time to time during CTF heap debugging. 
I wonder if you consider `hexdump module_name` useful? I could do a PR if so.

### My setup

```
Gdb:      8.1.0.20180409-git
Python:   3.6.8 (default, Aug 20 2019, 17:12:48)  [GCC 8.3.0]
Pwndbg:   1.1.0 build: 1158a30
Capstone: 4.0.1024
Unicorn:  1.0.1
```

stnevans:
I think it would be a reasonable feature to add if you would be willing to do so!

disconnect3d:
Added with #683

PR #683, opened 2019-10-30
Allow hexdump by module name, like vmmap

korniltsev:
Implements #678

### DIFF
--- a/pwndbg/commands/hexdump.py
+++ b/pwndbg/commands/hexdump.py
@@ -8,6 +8,7 @@ from __future__ import unicode_literals
 import argparse
 
 import gdb
+import six
 
 import pwndbg.arch
 import pwndbg.commands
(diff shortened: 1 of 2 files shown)


pwndbg/commands/hexdump.py
```diff
@@ -23,17 +24,32 @@
                          64,
                          'number of bytes printed by hexdump command')
 
-parser = argparse.ArgumentParser(description='Hexdumps data at the specified address (or at $sp)')
-parser.add_argument('address', nargs='?', default='$sp',
-                    help='Address to dump')
+def address_or_module_name(s):
```
disconnect3d, line 27: Hmm this is solved in vmmap as https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/commands/vmmap.py#L24-L38

It would probably be better to keep just one such function in the whole codebase.

Can we move it e.g. to pwndbg/commands/__init__.py and reuse it in here and in vmmap?

korniltsev: Good point. I'l try to do that. However it looks like I can't use it directly since it returns pages in both cases and here we need page only in one case.

disconnect3d: Hmm good point; though, what if a given objfile has many pages? we just go with the first one?

korniltsev: Yes, that was my intention. I.e. usages like `hexdump stack` or `hexdump heap` will start from the beginning.

disconnect3d: I think it is reasonable, let's merge it :).

PR #687, opened 2019-11-16
Extend vmmap monitor info mem case to x64

disconnect3d:


pwndbg/vmmap.py
```diff
@@ -266,7 +266,7 @@ def monitor_info_mem():
 
         pages.append(pwndbg.memory.Page(start, size, flags, 0, '<qemu>'))
 
-    return pages
+    return tuple(pages)
```
disconnect3d, line 269: This is to prevent `memoize`-related warnings that we are trying to hash (use as keys) mutable values.

disconnect3d:
I recently used this feature on x64/qemu kernel and it works with this little changes.

The only disadvantage is speed as we do it **on each stop** and there are **tons of memory pages on x64 linux kernel**.

So... one can use another little change to make it faster:
```
-@pwndbg.memoize.reset_on_stop
+@pwndbg.memoize.reset_on_start
 def monitor_info_mem():
```

...but less accurate.

Maybe we should move towards having a command to find all memory pages and just assume they are true forever (or until I explicitly re-calling the command that checks/finds them)?

PR #693, opened 2019-12-06
Fix #681

### LINKED ISSUE #681: qemu local_path bug
### Description

It seems like some commands e.g. plt, got tries to use ELF file's content as path.

### Steps to reproduce

```
$ qemu-arm -g 23946 ./target
$ gdb-multiarch ./target
(gdb) target remote :23946
(gdb) plt # or
(gdb) got
```

### My setup

Ubuntu 18.04, branch c6473ba (master)

### Analysis (and temporary patch)

plt/got command uses `get_elf_info` at [elf.py](https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/elf.py#L95), which uses get_file at file.py.

It seems like the path itself should be used in [file.py L36](https://github.com/pwndbg/pwndbg/blob/c6473ba7aea797f7b4d2251febf419382564d3f8/pwndbg/file.py#L36):

https://github.com/pwndbg/pwndbg/blob/203f10710e8a9a195a7bade59661ed7f3e3e641c/pwndbg/file.py#L35-L36

to
```py
    if pwndbg.qemu.root() and recurse:
        return os.path.join(pwndbg.qemu.binfmt_root, path)
```

I'm not sure if it's ok in all cases, since I think relative path should be also searched, but it works on plt/got case since absolute path is used (e.g. if path is 'a/b', I'm not sure if binfmt_root + '/a/b' and 'a/b' shoud be searched.)

Jinmo:


pwndbg/file.py
```diff
@@ -33,7 +33,7 @@ def get_file(path, recurse=1):
     local_path = path
 
     if pwndbg.qemu.root() and recurse:
-        return get(os.path.join(pwndbg.qemu.binfmt_root, path), 0)
+        return os.path.join(pwndbg.qemu.binfmt_root, path)
```
disconnect3d, line 36: So it turns out that instead of returning local path we could have returned bytes read from the file.

Uh...

disconnect3d: Btw shouldn't have the the if order reversed, i.e. first check for remote and then for `pwndbg.qemu.root()`?

Also, what is the situation when `pwndbg.qemu.root()` is used? I have qemu user on my system but I don't have the `/etc/qemu-binfmt` it tries to check :S.

disconnect3d:
Even though I asked a question I think this fix is valid. Thanks!

pwndbg/file.py
Jinmo: It's arbitrarily set by debian version of qemu as the default value of `QEMU_LD_PREFIX`. `qemu-arm-static --help` is showing it.

```
~/qemu-2.11+dfsg $ grep -r interp-prefix
debian/rules:   --interp-prefix=/etc/qemu-binfmt/%M \
...
```

PR #698, opened 2020-01-21
Improve probeleak command

pullp:
add two more options: mapping_name and stop_cnt. And do not effect previous usage.

mapping_name  : Mapping you want the pointers point to
stop_cnt : Stop search after get n required pointers, default 0. which means don't stop until reach the end

Here are some examples:
Before, if I want to find a libc address somewhere after 0x601000, I may excute below command, and find a libc address among the outputs.
```
pwndbg> probeleak 0x601000 0x1000                                                                         
0x000: 0x0000000000600e28 = (r--p) /mnt/hgfs/ctf_games/replay/diary_TW_2016/test + 0xe28 (_DYNAMIC)       
0x008: 0x00007ffff7ffe168 = (rw-p) [anon] + 0x168                                                         
0x010: 0x00007ffff7deee10 = (r-xp) /lib/x86_64-linux-gnu/ld-2.23.so + 0x17e10 (_dl_runtime_resolve_xsave) 
0x018: 0x00000000004004f6 = (r-xp) /mnt/hgfs/ctf_games/replay/diary_TW_2016/test + 0x4f6 (puts@plt+6)     
0x020: 0x0000000000400506 = (r-xp) /mnt/hgfs/ctf_games/replay/diary_TW_2016/test + 0x506 (setbuf@plt+6)   
0x028: 0x00007ffff7a2d740 = (r-xp) /lib/x86_64-linux-gnu/libc-2.23.so.init + 0x20740 (__libc_start_main)  
0x040: 0x00007ffff7dd2620 = (rw-p) /lib/x86_64-linux-gnu/libc-2.23.so.init + 0x1620 (_IO_2_1_stdout_)     
0x050: 0x00007ffff7dd18e0 = (rw-p) /lib/x86_64-linux-gnu/libc-2.23.so.init + 0x8e0 (_IO_2_1_stdin_)       
0x060: 0x00007ffff7dd2540 = (rw-p) /lib/x86_64-linux-gnu/libc-2.23.so.init + 0x1540 (_IO_2_1_stderr_)     
```

But with the improvment I can use below command
```
pwndbg> probeleak 0x601000 0x1000 0 libc 1                                                              
0x028: 0x00007ffff7a2d740 = (r-xp) /lib/x86_64-linux-gnu/libc-2.23.so.init + 0x20740 (__libc_start_main)
```

I think this improvment is really useful when there are lots of pointers(such as one qemu escape challenge I worked recently).

### DIFF
--- a/pwndbg/commands/probeleak.py
+++ b/pwndbg/commands/probeleak.py
@@ -5,6 +5,7 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import os
 import argparse
 import math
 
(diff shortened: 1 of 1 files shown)


disconnect3d:
Wouldn't it be better to have those behind option argument like `--point-to=libc --max=1` (or some other name)?

As currently it might be a bit confusing what means each argument.

Also an idea: maybe we should allow something like `pagename:permissions` so e.g. `libc:x` would search for executable libc pages and `libc:rwx` for rwx pages?

This could then be that e.g. `:w` filter would find any pointer that points to writable memory. That could be neat imho for long searches.

pullp:
what about add another argument for permissions, in case of some page name in clude ':'

disconnect3d:
Yeah, I agree. It's better to make it as an argument.

disconnect3d:
Btw the orignal scheme:
> pwndbg> probeleak 0x601000 0x1000 0 libc 1                                                              

is also okay, but if we leave it as it is, it's probably good to add some `print`/log message like `probing 0x601000(+0x1000) for 1 libc address (maxdist=0)`?

Also I think that we should move the `max_distance` argument at the end or even put it into the optional argument instead. I don't think it is widely used, though it is still really nice to have it.

pullp:
Now looks like this:

```
pwndbg> probeleak $rsp 0x64 --point-to libc --max-ptrs 1  --flags rx                                       
0x08: 0x00007ffff7a2d830 = (r-xp) /lib/x86_64-linux-gnu/libc-2.23.so.init + 0x20830 (__libc_start_main+240)
```

change max-distance's usage : 
```
pwndbg> probeleak $rsp 0x64 --max-dist 0x10                                                                
0x00: 0x0000000000400680 = (r-xp) /mnt/hgfs/ctf_games/replay/diary_TW_2016/test + 0x680 (__libc_csu_init)  
0x08: 0x00007ffff7a2d830 = (r-xp) /lib/x86_64-linux-gnu/libc-2.23.so.init + 0x20830 (__libc_start_main+240)
0x18: 0x00007fffffffe2b8 = (rw-p) [stack] + 0x202b8                                                        
0x28: 0x0000000000400626 = (r-xp) /mnt/hgfs/ctf_games/replay/diary_TW_2016/test + 0x626 (main)             
0x40: 0x0000000000400530 = (r-xp) /mnt/hgfs/ctf_games/replay/diary_TW_2016/test + 0x530 (_start)           
0x48: 0x00007fffffffe2b0 = (rw-p) [stack] + 0x202b0                                                        
```

disconnect3d (approved):

disconnect3d:
The build fails due to some python/pip related issues: https://travis-ci.org/pwndbg/pwndbg/builds/639982548?utm_source=github_status&utm_medium=notification
(pip released today had some issue - https://github.com/pypa/pip/issues/7620 - but they fixed it already; since like travis tried 20.0.1 whiich is fiixed but maybe it doesn't support <=py3.4?)

Anyway, lgtm, merging. Thanks!

PR #715, opened 2020-03-06
[WIP] Feature: show ghidra decompiled code in context

jerdna-regeiz:
With the help of radare2, r2pipe and r2ghidra-dec.

Here is an other feature I thought you might like as well.

This pull requests enables to use the ghidra decompiler (by using r2pipe, radare2 and r2ghidra-dec). Shown in the right split of the screenshot.
It is far from perfect to pull in a requirement for radare2, but using the ghidra decompiler directly requires a lot more work.

So it is not perfect, but adds some value as an optional feature for folks already working with r2 and r2ghidra-dec. Please tell me your thoughts about this.
screenshot

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -107,7 +107,7 @@ Be warned, the first call to radare2/r2ghidra-dec is rather slow! Subsequent req
 source will be faster. And it does take up some resources as the radare2 instance is kept by r2pipe
 to enable faster subsequent analysis.
 
-With those performance penalties it is resonable to not have if launch always. Therefore it includes
+With those performance penalties it is resonable to not have it launch always. Therefore it includes
 an option to only start it when required with `set context-ghidra`:
 * `set context-ghidra always`: always trigger the ghidra context
 * `set context-ghidra never`: never trigger the ghidra context except when called manually
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -89,6 +89,32 @@ contextwatch exectue "ds BUF"
 cwatch execute "x/20x $rsp"
 ```
 
+### Ghidra
+
+With the help of [radare2](https://github.com/radareorg/radare2) it is possible to show the
+decompiled source code of the ghidra decompiler.
+
+However, this comes with some prerequisites.
+* First: you have to have installed radare2 and it must be found by gdb (within path)
+* Second: you have to install the ghidra plugin for radare2
+  [r2ghidra-dec](https://github.com/radareorg/r2ghidra-dec)
+* Third: r2pipe has to be installed in the python-context gdb is using
+
+The decompiled source be shown as part of the context by adding `ghidra` to `set context-sections`
+or by calling `ctx-ghidra [function]` manually.
+
+Be warned, the first call to radare2/r2ghidra-dec is rather slow! Subsequent requests for decompiled
+source will be faster. And it does take up some resources as the radare2 instance is kept by r2pipe
+to enable faster subsequent analysis.
+
+With those performance penalties it is resonable to not have if launch always. Therefore it includes
+an option to only start it when required with `set context-ghidra`:
+* `set context-ghidra always`: always trigger the ghidra context
+* `set context-ghidra never`: never trigger the ghidra context except when called manually
+* `set context-ghidra if-no-source`: invoke ghidra if no source code is available
+
+Remark: the plugin tries to guess the correct current line and mark it with "-->", but it might
+get it wrong.
 
 ## Disassembly
 
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -1,6 +1,3 @@
-#ifndef CONTEXT_PY
-#define CONTEXT_PY
-
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import
@@ -672,5 +669,3 @@ def _is_rr_present():
     interpreter_globals = ast.literal_eval(globals_list_literal_str)
 
     return 'RRCmd' in interpreter_globals and 'RRWhere' in interpreter_globals
-
-#endif /* CONTEXT_PY */
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -9,6 +9,8 @@ import argparse
 import ast
 import codecs
 import ctypes
+import json
+import os
 import sys
 from collections import defaultdict
 from io import open
@@ -47,7 +49,7 @@ def clear_screen(out=sys.stdout):
 config_clear_screen = pwndbg.config.Parameter('context-clear-screen', False, 'whether to clear the screen before printing the context')
 config_output = pwndbg.config.Parameter('context-output', 'stdout', 'where pwndbg should output ("stdout" or file/tty).')
 config_context_sections = pwndbg.config.Parameter('context-sections',
-                                                  'regs disasm code stack backtrace expressions',
+                                                  'regs disasm code ghidra stack backtrace expressions',
                                                   'which context sections are displayed (controls order)')
 
 # Storing output configuration per section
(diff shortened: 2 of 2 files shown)


disconnect3d:
This is awesome! Let me come back later (~tomorrow) for a review! :)

FEATURES.md
```diff
@@ -67,7 +67,32 @@ import splitmind
...
+* Third: r2pipe has to be installed in the python-context gdb is using
+
+The decompiled source be shown as part of the context by adding `ghidra` to `set context-sections`
+or by calling `ctx-ghidra [function]` manually.
+
+Be warned, the first call to radare2/r2ghidra-dec is rather slow! Subsequent requests for decompiled
+source will be faster. And it does take up some resources as the radare2 instance is kept by r2pipe
+to enable faster subsequent analysis.
+
+With those performance penalties it is resonable to not have if launch always. Therefore it includes
```
disconnect3d:
```suggestion
With those performance penalties it is resonable to not have it launch always. Therefore it includes
```

pwndbg/commands/context.py
```diff
@@ -150,6 +152,87 @@ def contextoutput(section, path, clearing, banner="both", width=None):
...
 
+# Ghidra integration through radare2
+# This is not (sorry couldn't help it) "the yellow of the egg"
+# Using radare2 is suboptimal and will only be an intermediate step to have
+# ghidra decompiler within pwndbg.
+# But having ghidra in pwndgb is a lot more work as it requires quite some code and c/c++
+radare2 = {}
+
+config_context_ghidra = pwndbg.config.Parameter('context-ghidra',
+                                                  'always never if-no-source',
```
disconnect3d: According to the `Parameter` class this, the second argument is a default value. Shouldn't it be only `never` here? (see https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/config.py#L113)

pwndbg/commands/context.py
```diff
@@ -150,6 +152,87 @@ def contextoutput(section, path, clearing, banner="both", width=None):
...
+                                                  'configures if or/when to try to decompile with ghidra (slow and requires radare2/r2pipe')
+def init_radare2(filename):
+    r2 = radare2.get(filename)
```
disconnect3d, line 226: Is there a way to "ask" radare2 if Ghidra decompiler is plugged into it?

pwndbg/commands/context.py
```diff
@@ -150,6 +152,87 @@ def contextoutput(section, path, clearing, banner="both", width=None):
...
+@pwndbg.commands.ArgparsedCommand(parser, aliases=['ctx-ghidra'])
+def contextghidra(func):
+    print("\n".join(context_ghidra(func, with_banner=False, force_show=True)))
+
+def context_ghidra(func=None, target=sys.stdout, with_banner=True, width=None, force_show=False):
+    banner = [pwndbg.ui.banner("ghidra decompile", target=target, width=width)]
+    if config_context_ghidra == "never" and not force_show: return []
+    if config_context_ghidra == "if-no-source" and not force_show:
+        try:
+            f = open(gdb.selected_frame().find_sal().symtab.fullname())
```
disconnect3d: Use:
```python
with open(...) as _:
    # some comment describing why we need it // what file do we open
    pass
```

pwndbg/commands/context.py
```diff
@@ -150,6 +152,87 @@ def contextoutput(section, path, clearing, banner="both", width=None):
...
+parser.add_argument("func", type=str, default=None, nargs="?",
+                    help="Function to be shown. Defaults to current")
+@pwndbg.commands.ArgparsedCommand(parser, aliases=['ctx-ghidra'])
+def contextghidra(func):
+    print("\n".join(context_ghidra(func, with_banner=False, force_show=True)))
+
+def context_ghidra(func=None, target=sys.stdout, with_banner=True, width=None, force_show=False):
+    banner = [pwndbg.ui.banner("ghidra decompile", target=target, width=width)]
+    if config_context_ghidra == "never" and not force_show: return []
+    if config_context_ghidra == "if-no-source" and not force_show:
```
```suggestion
    elif config_context_ghidra == "if-no-source" and not force_show:
```

pwndbg/commands/context.py
```diff
@@ -150,6 +152,87 @@ def contextoutput(section, path, clearing, banner="both", width=None):
...
+parser.description = """Show current function decompiled by ghidra"""
+parser.add_argument("func", type=str, default=None, nargs="?",
+                    help="Function to be shown. Defaults to current")
+@pwndbg.commands.ArgparsedCommand(parser, aliases=['ctx-ghidra'])
+def contextghidra(func):
+    print("\n".join(context_ghidra(func, with_banner=False, force_show=True)))
+
+def context_ghidra(func=None, target=sys.stdout, with_banner=True, width=None, force_show=False):
+    banner = [pwndbg.ui.banner("ghidra decompile", target=target, width=width)]
+    if config_context_ghidra == "never" and not force_show: return []
```
disconnect3d: Can we move `return` to the next line?

It could also be beneficial to add some newlines here and there - might make the code a bit more readable.

pwndbg/commands/context.py
```diff
@@ -150,6 +152,87 @@ def contextoutput(section, path, clearing, banner="both", width=None):
...
+        except:         # a lot can go wrong in search of source code.
+            return []   # we don't care what, just that it did not work out well...
+    filename = gdb.current_progspace().filename
+    try:
+        r2 = init_radare2(filename)
+        if not "pdg" in r2.cmd("LD").split("\n"):
+            return banner + ["radare2 plugin r2ghidra-dec must be installed and available from r2"]
+    except ImportError: # no r2pipe present
+        return banner + ["r2pipe not available, but required for r2->ghidra-bridge"]
+    if func == None:
```
```suggestion
    if func is None:
```
one should use `is` operator to compare with `None` 🤷‍♂

https://stackoverflow.com/questions/3257919/what-is-the-difference-between-is-none-and-none

pwndbg/commands/context.py
disconnect3d: Hmm I see we do it later on, nvm

pwndbg/commands/context.py
```diff
@@ -150,6 +152,87 @@ def contextoutput(section, path, clearing, banner="both", width=None):
...
+    if config_context_ghidra == "if-no-source" and not force_show:
+        try:
+            f = open(gdb.selected_frame().find_sal().symtab.fullname())
+            f.close()
+        except:         # a lot can go wrong in search of source code.
+            return []   # we don't care what, just that it did not work out well...
+    filename = gdb.current_progspace().filename
+    try:
+        r2 = init_radare2(filename)
+        if not "pdg" in r2.cmd("LD").split("\n"):
```
disconnect3d, line 260: Can we add a comment showing an example output? So we know what we are looking for/why and if it changes in the future we will know this code worked but in previous version

pwndbg/commands/context.py
```diff
@@ -150,6 +152,87 @@ def contextoutput(section, path, clearing, banner="both", width=None):
...
+            return banner + ["radare2 plugin r2ghidra-dec must be installed and available from r2"]
+    except ImportError: # no r2pipe present
+        return banner + ["r2pipe not available, but required for r2->ghidra-bridge"]
+    if func == None:
+        try:
+            func = hex(pwndbg.regs[pwndbg.regs.current.pc])
+        except:
+            func = "main"
+    src = r2.cmdj("pdgj @"+func)
```
```suggestion
    src = r2.cmdj("pdgj @" + func)
```

pwndbg/commands/context.py
```diff
@@ -150,6 +152,87 @@ def contextoutput(section, path, clearing, banner="both", width=None):
...
+    source = src.get("code", "")
+    curline = None
+    try:
+        cur = pwndbg.regs[pwndbg.regs.current.pc]
+        closest = 0
+        for off in [a.get("offset", 0) for a in src.get("annotations", [])]:
```
```suggestion
        for off in (a.get("offset", 0) for a in src.get("annotations", [])):
```

The `[x for x in ...]` is a [list comprehension](https://docs.python.org/3/tutorial/datastructures.html#list-comprehensions) and `(x for x in ...)` is a [generator expression (see PEP-289; while its old it still explains the purpose)](https://www.python.org/dev/peps/pep-0289/#id14).

TLDR: We don't need to create the whole list (setup memory for that etc), we can just generate items on demand since we only iterate.

pwndbg/commands/context.py
```diff
@@ -150,6 +152,87 @@ def contextoutput(section, path, clearing, banner="both", width=None):
...
+            if abs(cur - closest) > abs(cur - off):
+                closest = off
+        pos_annotations = sorted([a for a in src.get("annotations", []) if a.get("offset") == closest],
+                         key=lambda a: a["start"])
+        if pos_annotations:
+            curline = source.count("\n", 0, pos_annotations[0]["start"])
+    except AttributeError: pass # If not running there is no current.pc
```
disconnect3d: Let's move `pass` to the next line

disconnect3d: Also what fires off `AttributeError`? We have quite broad `try ... except` here - it might be good to narrow it down to the specific need, otherwise, if some changes are made in the future and `AttributeError` fires off, we might not notice it.

Btw if it is for `pwndbg.regs.current.pc` to prevent running the command when we don't debug any target, then we should use `@pwndbg.commands.OnlyWhenRunning` decorator instead e.g. like here: https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/commands/stack.py#L16-L17

pwndbg/commands/context.py
```diff
@@ -150,6 +152,87 @@ def contextoutput(section, path, clearing, banner="both", width=None):
...
+    radare2[filename] = r2
+    r2.cmd("aaaa")
+    return r2
+
+parser = argparse.ArgumentParser()
+parser.description = """Show current function decompiled by ghidra"""
+parser.add_argument("func", type=str, default=None, nargs="?",
+                    help="Function to be shown. Defaults to current")
+@pwndbg.commands.ArgparsedCommand(parser, aliases=['ctx-ghidra'])
+def contextghidra(func):
```
disconnect3d, line 240: Just not to forget about this: we probably want to use `OnlyWhenRunning` decorator here, as described in https://github.com/pwndbg/pwndbg/pull/715/files#r391445700

pwndbg/commands/context.py
```diff
@@ -150,6 +152,87 @@ def contextoutput(section, path, clearing, banner="both", width=None):
...
+                closest = off
+        pos_annotations = sorted([a for a in src.get("annotations", []) if a.get("offset") == closest],
+                         key=lambda a: a["start"])
+        if pos_annotations:
+            curline = source.count("\n", 0, pos_annotations[0]["start"])
+    except AttributeError: pass # If not running there is no current.pc
+    source = source.split("\n")
+    if curline is not None:
+        if source[curline][0:4] == "    ":
+            source[curline] = "--> "+source[curline][4:]
```
```suggestion
            source[curline] = "--> " + source[curline][4:]
```

pwndbg/commands/context.py
```diff
@@ -150,6 +152,87 @@ def contextoutput(section, path, clearing, banner="both", width=None):
...
+        else:
+            source[curline] = "--> "+source[curline]
```
```python
line = source[curline]
if line.startswith('    '):
    line = line[4:]

source[curline] = '--> ' + line
```

pwndbg/commands/context.py
```diff
@@ -150,6 +152,87 @@ def contextoutput(section, path, clearing, banner="both", width=None):
...
+    source = "\n".join(source)
+    if pwndbg.config.syntax_highlight:
+        try:
+            src_filename = gdb.selected_frame().find_sal().symtab.fullname()
+        except:
+            src_filename = filename+".c" if os.path.basename(filename).find(".") < 0 else filename
```
disconnect3d, line 299: Hmm what exactly are we doing here?

disconnect3d:
@jerdna-regeiz hey, please see comments :)

pwndbg/commands/context.py
jerdna-regeiz: Uh right, somehow I confused it with valid values.

pwndbg/commands/context.py
jerdna-regeiz: Yeah, silly mistake ... >.<

pwndbg/commands/context.py
jerdna-regeiz: > Also what fires off `AttributeError`? We have quite broad `try ... except` here - it might be good to narrow it down to the specific need, otherwise, if some changes are made in the future and `AttributeError` fires off, we might not notice it.
> 
> Btw if it is for `pwndbg.regs.current.pc` to prevent running the command when we don't debug any target, then we should use `@pwndbg.commands.OnlyWhenRunning` decorator instead e.g. like here: https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/commands/stack.py#L16-L17

It is for `pwndbg.regs.current.pc`, but `@pwndbg.commands.OnlyWhenRunning` is to strong.
The command can run just fine when not running, it just can not get the current line (obviously).
I'll narrow down the try to the a single line and surround the rest with an if.

pwndbg/commands/context.py
jerdna-regeiz: As explained above, it can run fine and even be called with `ctx-ghidra myAwesomeFunc` to show the decompiled function when not running.

pwndbg/commands/context.py
jerdna-regeiz: It tries to get the filename of the source code for highlighting. 
It tries to read it from debug information and if not available it looks for any extension in the original filename. If non found it just assumes it was a .c file (and most likely it is correct for the output of ghidra).
It is kind of ugly, but I really didn't know better and the highlighting depends on the file extension...

jerdna-regeiz:
> @jerdna-regeiz hey, please see comments :)

Thy for the awesome review! =D

pwndbg/commands/context.py
```diff
@@ -1,3 +1,6 @@
+#ifndef CONTEXT_PY
+#define CONTEXT_PY
```
disconnect3d: Seems like IDE/automatic snippet inclusion occurred? :)

It also happens at the end :p

jerdna-regeiz: Yep. Somehow vim thought it to be c. Sorry for that.

disconnect3d:
@jerdna-regeiz hmm.. we also have two conflicts - can you resolve them?

jerdna-regeiz:
@disconnect3d Rebased to latest dev

PR #721, opened 2020-03-13
Enhance find_fake_fast

bet4it:
1. The `size` argument is optional now.
2. Make the search more precise.

### DIFF
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -431,8 +431,8 @@ parser.description = """
     Finds candidate fake fast chunks that will overlap with the specified
     address. Used for fastbin dups and house of spirit
     """
-parser.add_argument("addr", type=int, help="The start address.") #TODO describe these better
-parser.add_argument("size", nargs="?", type=int, default=None, help="The size.")
+parser.add_argument("addr", type=int, help="The start address of a word size value you want to overlap.")
+parser.add_argument("size", nargs="?", type=int, default=None, help="The size of fastbin you want to use.")
 @pwndbg.commands.ArgparsedCommand(parser)
 @pwndbg.commands.OnlyWhenRunning
 @pwndbg.commands.OnlyWithLibcDebugSyms
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -432,21 +432,24 @@ parser.description = """
     address. Used for fastbin dups and house of spirit
     """
 parser.add_argument("addr", type=int, help="The start address.") #TODO describe these better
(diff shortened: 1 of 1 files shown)


disconnect3d:
@GrosQuildu can u review this?

GrosQuildu:
There is something wrong. If we add this debug-prints:
```py
print(max_fastbin, align, max_fast)
print(size, len(mem), fastbin, (max_fastbin - fastbin) * align, max_fast - align)
``` 

We get:
```
pwndbg> find_fake_fast 0x7ffff7dcf000 8
FAKE CHUNKS
6 16 128
8 120 -2 128 112
```

So the inner (`offeset`) loop is not executed. Either handle cases with small sizes, or add `if` size < min_fast or size > max_fast: some error`.

GrosQuildu:
Despite the above:
- would be nice to add more description to the args
- we may extend this function to take two addresses (start and end, or start and size, dont know what convention pwndbg uses) and search for fake chunks that will give us control over any bytes in the region specified.

bet4it:
> pwndbg> find_fake_fast 0x7ffff7dcf000 8

You shouldn't use it like this. `min_fast` is 0x20 so you should use `find_fake_fast 0x7ffff7dcf000 0x20` at least. Or you can just use `find_fake_fast 0x7ffff7dcf000`.

> would be nice to add more description to the args

I think it's not so necessary. People can just use this command with the way they used before. And you will notice `size` is optional by `find_fake_fast -h` : `usage: find_fake_fast [-h] addr [size]`. But if you have good suggestions I will take it.

> search for fake chunks that will give us control over any bytes in the region specified.

I think most time we only need to change the key value of specific address.
And even if we need to implement it, it could be another PR.

GrosQuildu:
> You shouldn't use it like this.

Yes, but would be nice to print some warning.

> I think it's not so necessary. People can just use this command with the way they used before.

There is TODO comment in the code. I would add a text from the function docstring about the address (that it searches for chunks that overlaps with the address, but excluding the chunk's header) and that the size should be between that-and-that value (and aligned?).

The last: another PR, agree. It was more a note.

bet4it:
I have tried to add some descriptions.

> that the size should be between that-and-that value

It's hard to describe, because it may be affected by many factors, such as `global_max_fast`. I think the people use it should know the meanings well.

> would be nice to print some warning

It's also very hard. There are so many special situations. I think most of the time the exception threw by python is enough. When using the `heap` command I also often encounter an exception😀

GrosQuildu:
KK, we may merge @disconnect3d

PR #723, opened 2020-03-17
Fix alias for contextunwatch command

benediktwerner:
The last commit (a18e751) that introduced the `contextwatch` and `contextunwatch` commands uses the same alias `ctx-watch` for both.

Besides obviously being wrong this also causes `pwndbg` to crash on startup because of a duplicate command definition.

This PR renames the alias for `contextunwatch` from `ctx-watch` to `ctx-unwatch`.


jerdna-regeiz:
Guess this is a duplicate of https://github.com/pwndbg/pwndbg/pull/725 which is merged now

Issue #347, opened 2017-09-26
PtMalloc pretty-printers


zachriggle:
The current `ptmalloc` printers are pretty basic, and print lots of fields that are not relevant, and some of the e.g. INUSE flags are done manually via the `pwndbg.commands.heap` module instead of done by a pretty printer.

We should move this logic to a pretty printer.  We should also be more selective about when we print certain fields, and we should also be able to hex-dump the contents of heap allocations inline with the pretty-printer.

PR #728, opened 2020-03-21
Update heap implementation

CptGibbon:
This PR makes some improvements to pwndbg's heap-related commands.
### Primary changes

* The "heap" and "top_chunk" commands now print just the relevant information about chunks they encounter, rather than dumping all possible fields whether they are in use or not.

**Before:** 
image
image

**After:**
image
image

* Heap commands now default to using the currently debugged thread's arena, rather than the main arena. This makes things a lot more convenient when debugging multi-threaded applications and fixes the issue where no bin labels were displayed when printing a foreign thread's heap.

* Tcache commands have been placed behind an OnlyWithTcache decorator, which stops them from being invoked when debugging versions of GLIBC compiled without tcache support.

### Housekeeping:

* Some variables have been renamed for clarity; for example, a variable named "main_arena" was being used to refer to arenas that may or may not be the main arena. "main_heap" has been renamed to "allocator" because it refers to the allocator implementation rather than any specific heap.

* Argparse descriptions are single-line now so they don't get truncated by the "pwndbg" command.

* Some PEP8 amendments such as whitespace removal and docstring width.

### DIFF
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -91,11 +91,12 @@ def format_bin(bins, verbose=False, offset=None):
 parser = argparse.ArgumentParser()
 parser.description = "Iteratively print chunks on a heap, default to the current thread's active heap."
 parser.add_argument("addr", nargs="?", type=int, default=None, help="Address of the first chunk.")
+parser.add_argument("-v", "--verbose", action="store_true", help="Print all chunk fields, even unused ones.")
 @pwndbg.commands.ArgparsedCommand(parser)
 @pwndbg.commands.OnlyWhenRunning
 @pwndbg.commands.OnlyWithLibcDebugSyms
 @pwndbg.commands.OnlyWhenHeapIsInitialized
-def heap(addr=None):
+def heap(addr=None, verbose=False):
     """Iteratively print chunks on a heap, default to the current thread's
     active heap.
     """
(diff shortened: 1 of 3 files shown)


GrosQuildu:
I don't like new "only relevant" printing. Often its useful to quickly view all data that may matter. But the new formatting is cool. Maybe we can print all the data but mark not relevant fields in red?

disconnect3d:
Or have a flag like `-a` / `--all` to be able to show all data? Or/and a config variable, heh

GrosQuildu:
Hem, I would rather have `--relevant` ;p

CptGibbon:
> Hem, I would rather have `--relevant` ;p

Sure, I can make this change.
The reason I chose to print only relevant fields by default was #347, in which @zachriggle comments _"The current `ptmalloc` printers . . . print lots of fields that are not relevant"_.

disconnect3d:
@GrosQuildu revisit pls

pwndbg/commands/heap.py
```diff
@@ -574,11 +603,13 @@ def vis_heap_chunks(address=None, count=None, naive=None):
 
     print(out)
 
+
 def bin_ascii(bs):
     from string import printable
-    valid_chars = list(map(ord, set(printable) - set('\t\r\n')))
+    valid_chars = list(map(ord, set(printable) - set('\t\r\n\x0c')))
```
GrosQuildu, line 616: Why?

pwndbg/heap/ptmalloc.py
```diff
@@ -134,40 +136,23 @@ def arenas(self):
...
+        thread's tcache.
+        """
+        tcache = self.mp['sbrk_base'] + 0x10
+        if self.multithreaded:
+            tcache_addr = pwndbg.memory.pvoid(pwndbg.symbol.address('tcache'))
+            if tcache_addr != 0:
+                tcache = tcache_addr
 
-        if tcache_addr is not None:
+        if tcache is not None:
```
GrosQuildu: Isn't it always not None?

pwndbg/heap/ptmalloc.py
```diff
@@ -268,20 +241,25 @@ def minsize(self):
...
     @pwndbg.memoize.reset_on_objfile
     def min_chunk_size(self):
         """Corresponds to MIN_CHUNK_SIZE in glibc malloc.c"""
         return pwndbg.arch.ptrsize * 4
 
+    @property
+    @pwndbg.memoize.reset_on_objfile
+    def multithreaded(self):
+        """Is malloc operating within a multithreaded environment."""
+        addr = pwndbg.symbol.address('__libc_multiple_threads')
```
GrosQuildu, line 251: i686 libcs seems to not have this symbol. Any other way to detect multithreading?

GrosQuildu:
Seems fine, cool! Just solve conflict and we may merge.

Only have a question about `allocator` varname. Where it came from? `malloc.c` code? Cause name like `heap_current` would be far more readable than `allocator`.

pwndbg/commands/heap.py
CptGibbon: Without this change, in xterm I see the following effect when vis_heap_chunks() attempts to hexdump the form feed (`0x0c`) character:
Capture

pwndbg/heap/ptmalloc.py
GrosQuildu: Maybe primitive like:
```py
        if addr:
            return pwndbg.memory.s32(addr) > 0
        return len(gdb.execute('info threads', to_string=True).split('\n')) > 3
```

pwndbg/heap/ptmalloc.py
CptGibbon: You're correct, I'll remove this check.

CptGibbon:
> Only have a question about `allocator` varname. Where it came from? `malloc.c` code? Cause name like `heap_current` would be far more readable than `allocator`.

I picked it because the presence of files like `dlmalloc.py` in the `heap` directory implies that multiple allocators could be supported in future. I feel like `heap_current` could refer to an actual heap, which there may be lots of in a single process, whereas `allocator` is an umbrella term for all the metadata relevant to every heap & arena.

GrosQuildu:
Hm, ok. Only thing with the `__libc_multiple_threads` left.

pwndbg/heap/ptmalloc.py
CptGibbon: I'm unfamiliar with the i686 GLIBC versions so I'll defer to you and make your suggested change.

GrosQuildu:
@disconnect3d merge pls.

disconnect3d:
@GrosQuildu i'm giving u a push access to the repo

PR #739, opened 2020-04-16
Fixed alignment bug in vis_heap_chunks command

jkthecjer:
Used alignment code from the "heap" command as a reference to fix this bug that sometimes causes "vis_heap_chunks" to print nothing due to an assumption that was being made in vis_heap_chunks (that the first chunk starts exactly on the heap region start).  The specific target that I saw this bug behavior in was bcloud in https://github.com/ctfs/write-ups-2016/tree/master/bctf-2016/exploit/bcloud-200 (from a breakpoint before any heap exploit shenanigans, just legitimate mallocs/data).  This patch fixes this behavior/bug.

### DIFF
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -490,12 +490,18 @@ def vis_heap_chunks(address=None, count=None, naive=None):
     heap_region = main_heap.get_heap_boundaries(address)
     main_arena = main_heap.get_arena_for_chunk(address) if address else main_heap.main_arena
 
+    first_chunk = heap_region.start
     top_chunk = main_arena['top']
     ptr_size = main_heap.size_sz
 
+    # Check if there is an alignment at the start of the heap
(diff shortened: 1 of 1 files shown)


CptGibbon:
The `cursor` variable is reset later in the function, before printing.

https://github.com/pwndbg/pwndbg/blob/05be2d50c2f8f9552d676b23453242706e407649/pwndbg/commands/heap.py#L554

Changing it here too would ensure printing is done accurately.

jkthecjer:
> The `cursor` variable is reset later in the function, before printing.
> 
> https://github.com/pwndbg/pwndbg/blob/05be2d50c2f8f9552d676b23453242706e407649/pwndbg/commands/heap.py#L554
> 
> Changing it here too would ensure printing is done accurately.

Right you are, made an additional commit fixing this, thanks!

disconnect3d:
@GrosQuildu mind having a look too?

GrosQuildu (approved):

PR #740, opened 2020-04-19
mprotect command injecting mprotect syscall

mzr:
Hacked during a CTF, Might be useful when one wants to change permission of a memory page and libc's mprotect function is not available. 
Maybe this could also be made i386 compatible.
I also wonder whether there is a decorator like `@pwndbg.commands.OnlyArch(x86_64)` ?

### DIFF
--- a/pwndbg/commands/mprotect.py
+++ b/pwndbg/commands/mprotect.py
@@ -1,0 +1,88 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import argparse
+
+import gdb
+import pwndbg.chain
+import pwndbg.commands
+import pwndbg.enhance
+import pwndbg.file
+import pwndbg.which
+import pwndbg.wrappers.checksec
+import pwndbg.wrappers.readelf
+from pwndbg.color import message
+
+parser = argparse.ArgumentParser(description='Calls mprotect. x86_64 only.')
+parser.add_argument('addr', help='Page-aligned address to all mprotect on.',
+                    type=int)
+parser.add_argument('length', help='Count of bytes to call mprotect on. Needs '
+                    'to be multiple of page size.',
+                    type=int)
+parser.add_argument('prot', help='Prot string as in mprotect(2). Eg. '
+                    '"PROT_READ|PROT_EXEC"', type=str)
+
+SYS_MPROTECT = 0x7d
+
+prot_dict = {
+    'PROT_NONE': 0x0,
+    'PROT_READ': 0x1,
+    'PROT_WRITE': 0x2,
+    'PROT_EXEC': 0x4,
+}
+
+def prot_str_to_val(protstr):
+    '''Heuristic to convert PROT_EXEC|PROT_WRITE to integer value.'''
+    prot_int = 0
+    for k in prot_dict:
+        if k in protstr:
+            prot_int |= prot_dict[k]
+    return prot_int
+
+
+@pwndbg.commands.ArgparsedCommand(parser)
+@pwndbg.commands.OnlyWhenRunning
+@pwndbg.commands.OnlyAmd64
+def mprotect(addr, length, prot: str):
+    '''Only x86_64.'''
+    saved_rax = pwndbg.regs.rax
+    saved_rbx = pwndbg.regs.rbx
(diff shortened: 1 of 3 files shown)


zachriggle:
This shouldn't be necessary, you can already do this with `shellcraft.amd64.linux.mprotect` from the command-line.  You can use `--run` or `--debug` to automate the debugging process.

Debugging example, works best if you are in a tmux session or similar:

```
$ shellcraft amd64.linux.sh --debug
```

For protect:

```
$ shellcraft amd64.linux.syscalls.mprotect -?
mprotect(addr, length, prot) -> str

Invokes the syscall mprotect.

See 'man 2 mprotect' for more information.

Arguments:
    addr(void*): addr
    len(size_t): len
    prot(int): prot
Returns:
    int
```

More options:

```
$ shellcraft --help
usage: pwn shellcraft [-h] [-?] [-o file] [-f format] [-d] [-b] [-a] [-v AVOID] [-n] [-z] [-r] [--color] [--no-color] [--syscalls] [--address ADDRESS] [-l] [-s] [shellcode] [arg [arg ...]]

positional arguments:
  shellcode             The shellcode you want
  arg                   Argument to the chosen shellcode

optional arguments:
  -?, --show            Show shellcode documentation
  -o file, --out file   Output file (default: stdout)
  -f format, --format format
                        Output format (default: hex), choose from {e}lf, {r}aw, {s}tring, {c}-style array, {h}ex string, hex{i}i, {a}ssembly code, {p}reprocssed code, escape{d} hex string
  -d, --debug           Debug the shellcode with GDB
  -b, --before          Insert a debug trap before the code
  -a, --after           Insert a debug trap after the code
  -v AVOID, --avoid AVOID
                        Encode the shellcode to avoid the listed bytes
  -n, --newline         Encode the shellcode to avoid newlines
  -z, --zero            Encode the shellcode to avoid NULL bytes
  -r, --run             Run output
  --color               Color output
  --no-color            Disable color output
  --syscalls            List syscalls
  --address ADDRESS     Load address
  -l, --list            List available shellcodes, optionally provide a filter
  -s, --shared          Generated ELF is a shared library
```

disconnect3d:
I think @mzr asks for an automated solution for `call (int (*)(void*,size_t,int)) mprotect(addr, size, prot)` that would also work if glibc is not loaded into the binary.

pwndbg/commands/mprotect.py
```diff
@@ -0,0 +1,87 @@
...
+        if k in protstr:
+            prot_int |= prot_dict[k]
+    return prot_int
+
+
+@pwndbg.commands.ArgparsedCommand(parser)
+@pwndbg.commands.OnlyWhenRunning
+def mprotect(addr, length, prot: str):
+    '''Only x86_64.'''
+    saved_rax = pwndbg.regs.rax
```
disconnect3d, line 52: Can we either add `OnlyAmd64` decorator or a check for it in the code like `if pwndbg.arch != X86_64`?

pwndbg/commands/mprotect.py
```diff
@@ -0,0 +1,87 @@
...
+    gdb.execute('set $rcx={}'.format(length))
+    gdb.execute('set $rdx={}'.format(prot_int))
+
+    saved_instruction_2bytes = pwndbg.memory.read(pwndbg.regs.rip, 2)
+
+    # int 0x80
+    pwndbg.memory.write(pwndbg.regs.rip, b'\xcd\x80')
+
+    # execute syscall
+    gdb.execute('stepi')
```
disconnect3d, line 70: Wonder if this should be nexti but stepi should also work.

disconnect3d:
I actually looked at the code now and I think this could be very useful. A bit sad its x64 only but we could use pwntools in the future for this.

disconnect3d:
@mzr ping

pwndbg/commands/mprotect.py
mzr: IMO `nexti` doesn't make sense here. We just want to step over syscall instruction.

pwndbg/commands/mprotect.py
mzr: sure

mzr:
> I think @mzr asks for an automated solution for `call (int (*)(void*,size_t,int)) mprotect(addr, size, prot)` that would also work if glibc is not loaded into the binary.

Yup. That's what I meant.

disconnect3d (approved):

PR #744, opened 2020-04-26
Try heap

GrosQuildu:
This commit adds new command: `try_free(address)`. It tries to inform user about possible `free` failures/aborts. It should be helpful for two reasons:
* no need to rerun binary under gdb just to see if your exploit finally do not crash during some heap magic
* comprehensive informations about failure reasons are provided, so there is a lot less to debug/print/calculate

Looks like:
```
pwndbg> try_free 0x5555557582a0
General checks
free(): invalid pointer -> &chunk + chunk->size > max memory
    0x555555758290 + 0xffffffffffffff00 > 0xffffffffffffffff


pwndbg> try_free 0x555555758320
General checks
Tcache checks
Fastbin checks
invalid fastbin entry (free) -> chunk's size is not near top chunk's size
    chunk's size == 65, idx == 2
    top chunk's size == 88, idx == 3    if `have_lock` is false then the error is invalid

pwndbg> try_free 0x555555759310
General checks
Not mapped checks
Backward consolidation
corrupted size vs. prev_size while consolidating
prev_size field is 0xfb0, prev chunk at 0x555555758350, prev chunk size is 0xaa
Next chunk is not top chunk
Clearing next chunk's P bit
```

TODO: `munmap_chunk` and `unlink_chunk` checks.

I also provide tests. For that is slightly modified testing framework, but should work with previous tests. I also created `tests/binaries/heap_bugs.c` with functions triggering various heap bugs (run like `tests/binaries/heap_bugs.out 1`). Makefile is extended with compilation cmd `heap_bugs` which compile it under specified glibc version.

Tests may be extended someday to run with various libc variations (x86/i686, tcache, versions...).

Re-implementing heap stuff in python can't be completely compatible with C, but having such quick, basic checks may useful for exp dev. Will come out in practice ;) Maybe a command that make use of libc would be better, but I couldn't find a way to both run `free` in inversible way and gather debug infos.

Commit also adds small fixes for `heap` and `malloc_chunk` commands.


disconnect3d:
Needs conflict fix :P

PR #748, opened 2020-04-30
fix for ubuntu 20.04

haxmeadroom:
Fixes Issue with Ubuntu 20.04, which no longer has python-pip

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,3 +12,4 @@ pygments
 capstone==4.0.1
 enum34
 pytest
+testresources
--- a/setup.sh
+++ b/setup.sh
@@ -20,7 +20,9 @@ osx() {
 
 install_apt() {
     sudo apt-get update || true
-    sudo apt-get -y install gdb python-dev python3-dev python-pip python3-pip libglib2.0-dev libc6-dbg
+    # This breaks Ubuntu 20.04, but is probably still needed for older version
+    sudo apt-get -y  python-pip || true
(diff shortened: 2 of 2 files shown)


requirements.txt
```diff
@@ -12,3 +12,4 @@ pygments
 capstone==4.0.1
 enum34
 pytest
+testresources
```
zachriggle, line 15: Why is this needed?

haxmeadroom: Not sure if it's required, but it got rid of this message:
`ERROR: launchpadlib 1.10.13 requires testresources, which is not installed.  `

disconnect3d:
Tested this branch on `ubuntu:18.04`, `ubuntu:19.10` and `ubuntu:20.04` with this Dockerfile (only changing the image):
```
FROM ubuntu:20.04

ARG DEBIAN_FRONTEND=noninteractive

RUN apt update && apt install -y git gdb && \
	git clone https://github.com/haxmeadroom/pwndbg.git && \
	cd pwndbg && git checkout dev && \
	 ./setup.sh
```

Seems to install & work fine (`docker build -t pwn . && docker run --rm -it --cap-add=SYS_PTRACE pwn gdb /bin/ls`). Thanks!

requirements.txt
disconnect3d: We might want to move to install this only for ubuntu/debian installs via python3-testresources as in https://github.com/pypa/pip/issues/5372

Issue #752, opened 2020-05-09
Backtrace in context output is missing symbols compared to "backtrace" output


peace-maker:
### Description
The backtrace shown in the `context` output isn't showing all the symbol information that's available compared to the output of `bt`. I'd expect the symbols to be shown in the `context` view as well.

```
───────────────────────────────[ BACKTRACE ]───────────────────────────────
 ► f 0     7fd376b7d290
   f 1     7fd376bb057f
   f 2     7fd376b82bf9 sq_call+116
   f 3     561accf89ad8 getargs+588
   f 4     561accf89fcc main+161
   f 5     7fd37699e0b3 __libc_start_main+243
───────────────────────────────────────────────────────────────────────────
pwndbg> bt
#0  0x00007fd376b7d290 in SQVM::Execute(SQObjectPtr&, long long, long long, SQObjectPtr&, unsigned long long, SQVM::ExecutionType)@plt () from /squirrel/build/lib/libsquirrel.so.0                                              
#1  0x00007fd376bb057f in SQVM::Call (this=0x561acd97b820, closure=..., nparams=1, stackbase=2, outres=..., raiseerror=1) at /squirrel/squirrel/sqvm.cpp:1587                                                                    
#2  0x00007fd376b82bf9 in sq_call (v=0x561acd97b820, params=1, retval=1, raiseerror=2) at /squirrel/squirrel/sqapi.cpp:1178                           
#3  0x0000561accf89ad8 in getargs (v=0x561acd97b820, argc=-845694944, argv=0x7fff04254258, retval=0x7fff04254130) at /squirrel/sq/sq.c:193
#4  0x0000561accf89fcc in main (argc=2, argv=0x7fff04254258) at /squirrel/sq/sq.c:330                                                                 
#5  0x00007fd37699e0b3 in __libc_start_main (main=0x561accf89f2b <main>, argc=2, argv=0x7fff04254258, init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fff04254248) at ../csu/libc-start.c:308
#6  0x0000561accf8956e in _start ()
```

### Steps to reproduce
```
FROM ubuntu:latest

ARG DEBIAN_FRONTEND=noninteractive
RUN apt-get update && apt-get install -y build-essential cmake git gdb

RUN git clone https://github.com/albertodemichelis/squirrel.git && \
[...]

PR #754, opened 2020-05-11
backtrace section in context with more information

pullp:
fix issue https://github.com/pwndbg/pwndbg/issues/752#issue-615165861

```
> cat test.c
int f995(){return 995;}
int f996(){return f995();}
int f997(){return f996();}
int f998(){return f997();}
int f999(){return f998();}
int f1000(){return f999();}

int main(){
        f1000();
        return 0;
}
```

test command
```
gcc -g  test.c -o test && gdb -q ./test -ex "b f995" -ex "r"
```

before:
image

after:
image
image

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -545,48 +545,28 @@ def context_backtrace(frame_count=10, with_banner=True, target=sys.stdout, width
     result = []
 
     if with_banner:
-        result.append(pwndbg.ui.banner("backtrace", target=target, width=width))
+        result.append(pwndbg.ui.banner("backtrace"))
 
-    this_frame    = gdb.selected_frame()
-    newest_frame  = this_frame
-    oldest_frame  = this_frame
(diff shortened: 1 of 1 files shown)


zachriggle:
@pullp can you provide examples of the output from before and after in the main pull request description?

Thanks!

pullp:
updated @zachriggle

disconnect3d:
@pullp why closing this?

zachriggle:
Based on the PR, it's not clear that the existing behavior would still work, given the amount of code removed

pullp:
I just think my implementation is not good. Later I will try another implementation metion in [this comment](https://github.com/pwndbg/pwndbg/issues/752#issuecomment-626688489), But currently I don't have time working on it.

disconnect3d:
@pullp kk, thanks for info. I will close this for now and we can either come back to this or come with another implementation of this.

Issue #664, opened 2019-07-24
Registers no longer show changes


xx55x:
### Description

In pwndbg release 2017.02.01 build: 1ab3de0, when registers changed while debugging a program, there would be an asterisk near the changed register which could be customized with context-register-changed-marker and context-register-changed-color. 

These config options are still available, however in later releases (both 2018.07.29 and 2019.01.25)  there doesn't seem to be anything next to changed registers anymore. The space is still there, which is more noticeable by changing the changed-marker to something longer, but it won't become visible.

It is still possible to see the asterisk, however only when changing a register manually, for example with set $eip=$eax. In this case an asterisk would appear near EIP for one instruction.

### Steps to reproduce

I've tried this on Debian as well as blackarch, and it seems to be reproducible with any 32 or 64bit binary, at least on x86. 

I believe stepping through any code where registers are being changed will reproduce it. 

### My setup

Debian: 
Gdb:      8.2.1
Python:   3.7.3 (default, Apr  3 2019, 05:39:12)  [GCC 8.3.0]
Pwndbg:   1.1.0 build: 88723a8
Capstone: 4.0.1024
Unicorn:  1.0.1

Blackarch: 

Gdb:      8.3
Python:   3.7.3 (default, June 24 2019, 04:54:02)  [GCC 9.1.0]
Pwndbg:   1.1.0 
Capstone: 4.0.1024
Unicorn:  1.0.1

disconnect3d:
I confirmed the bug. I think it should be fairly easy to fix this.

Would you like to work on that? I am on vacation so I might not have time to jump on it :P.

xx55x:
I understand no problem.

I’ve tried however I’m honestly still a beginner . I’m learning though I’ll see if I can try again when I understand more

xx55x:
Oh shit I fixed it, it was a lot simpler than I expected. Once I read the guides I'll submit it

sudhackar:
If you look at `update_last`
https://github.com/pwndbg/pwndbg/blob/848247ed9c5b832dc0f143cd2df605b90f2f05c5/pwndbg/regs.py#L441-L444 It gets called at both `cont` and `stop` events. idk why we need to update at `stop`.
I was able to fix this with this

```diff
diff --git a/pwndbg/regs.py b/pwndbg/regs.py
index 9e7e05a..68fe1d7 100644
--- a/pwndbg/regs.py
+++ b/pwndbg/regs.py
[...]

PR #756, opened 2020-05-11
#664 mark changed registers

layderv:


layderv:
You can see in https://travis-ci.org/github/pwndbg/pwndbg/jobs/685898957 the build is failing. I've added a commit to fix it as well. I can remove/split it, if it's unnecessary.

zachriggle:
Can we remove the `trusty` -> `bionic` change from this?  We should do that separately from this PR.

layderv:
> Can we remove the `trusty` -> `bionic` change from this? We should do that separately from this PR.

Updated.

layderv:
Do I need to do anything else here? Can this be merged?

layderv:
Any update on this @zachriggle ?

disconnect3d:
Tested it and it works. It will also mark register changes when we switch between threads - I guess this could be improved one way or another but I think it is good enough.

PR #775, opened 2020-05-31
Determine register sizes dynamically, do not assume ptrdiff width

### LINKED ISSUE #743: Don't hard code a native register sized type as being called "int"
https://github.com/pwndbg/pwndbg/blob/fbd2bb3abfc2500aae76d159e23015008e879b8d/pwndbg/arguments.py#L155

This is only true for languages in the C family.

WeSeekAfterKnowledge:
This makes register values more accurate

Fixes #743


disconnect3d:
Umm, so what's the exact difference this makes?

Does it mean the ptrdiff value is wrong in some cases? Maybe it should then be adjusted?

PR #794, opened 2020-06-27
telescope: display retaddrs

disconnect3d:
Extends `telescope` command so it displays not only register names, but
also return addresses names (ret0, ret1, ret2, ...).

The names are also set as GDB variables so they can be referenced via
`$retX` similarly as registers are by default.

Additionally, I moved the `retaddr` command from
pwndbg/commands/stack.py to pwndbg/commands/retaddr.py.


disconnect3d:
Example display:
image

disconnect3d:
One thing I am concerned about: I guess it will crash the command if stack frame is totally broken. We should probably handle this gracefully.

Also: setting GDB variables is not ideal: they are not refreshed e.g. on break and only on telescope, so a variable may hold old values or, e.g., a `$retX` may be set when there is no more X return address...

pwndbg/commands/retaddr.py
```diff
@@ -0,0 +1,20 @@
...
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import gdb
+
+import pwndbg.arch
+import pwndbg.chain
+import pwndbg.commands
+import pwndbg.regs
```
disconnect3d, line 11: those imports need to be fixed

pwndbg/commands/telescope.py
```diff
@@ -71,30 +72,38 @@ def telescope(address=None, count=None, to_string=False):
...
-        reg_values[pwndbg.regs[reg]].append(reg)
-    # address    = pwndbg.memory.poi(pwndbg.typeinfo.ppvoid, address)
+        names_values[pwndbg.regs[reg]].append(reg)
+
+    # Add return addresses names
+    for idx, retaddr in enumerate(pwndbg.stack.yield_retaddrs()):
+        name = 'ret%d' % idx
+        names_values[retaddr].append(name)
+        # Set $retX variable so it can be used by the user
+        gdb.execute("set $%s=0x%x" % (name, retaddr))
```
disconnect3d, line 86: This is the GDB variable setting, should probably be removed from the PR.

zachriggle:
We already have ROP handling through Unicorn emulation, we may be able to rely on that in order to get higher-quality information so we don't have to worry about crashing the `stack` command.

disconnect3d:
I am closing this for now, I will add an issue for it.

PR #795, opened 2020-06-27
vmmap command: show offset for single addresses

disconnect3d:
Changes vmmap display so when it is invoked with a single address, it
will also display its offset.

Before this commit:
```
pwndbg> vmmap $rsp
    0x7ffffffde000     0x7ffffffff000 rw-p    21000 0      [stack]
```

After this commit:
```
pwndbg> vmmap $rsp
    0x7ffffffde000     0x7ffffffff000 rw-p    21000 0      [stack] +0x1fb60
```

### DIFF
--- a/pwndbg/commands/vmmap.py
+++ b/pwndbg/commands/vmmap.py
@@ -23,15 +23,8 @@ import pwndbg.vmmap
 
 integer_types = six.integer_types + (gdb.Value,)
 
-# Ugly hack
-arg_val = None
 
-def pages_filter(s):
-    global arg_val
-
-    gdbval_or_str = pwndbg.commands.sloppy_gdb_parse(s)
-    arg_val = gdbval_or_str
-
+def pages_filter(gdbval_or_str):
     # returns a module filter
     if isinstance(gdbval_or_str, six.string_types):
         module_name = gdbval_or_str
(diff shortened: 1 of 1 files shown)


disconnect3d:
Screenshot: 
image

disconnect3d:
@zachriggle Thoughts about this PR? I think the `Ugly hack` is perfectly fine here...

anthraxx:
Hm maybe I'm missing something but why does this need to be a global?
Can't you have a function that returns a tuple of (arg_val, filter_func lambda)? or just call `pwndbg.commands.sloppy_gdb_parse` before and have a local variable for `pwndbg.vmmap.get()` at the filter caller?

PS: overall nice idea :cat:

anthraxx:
Can't currently functional test the patch, but if that works, it looks much better compared to the global hack :smile_cat:

disconnect3d:
@anthraxx 

> Hm maybe I'm missing something but why does this need to be a global?.

It does not need to be :).

> Can't you have a function that returns a tuple of (arg_val, filter_func lambda)? or just call pwndbg.commands.sloppy_gdb_parse before and have a local variable for pwndbg.vmmap.get() at the filter caller?

The latter is what @vesim987 suggested to me ~5 min ago and I just commited the latter of what you've written :D.

zachriggle:
@disconnect3d cool stuff! Definitely worth the hack

PR #796, opened 2020-06-27
Update .travis.yml: trusty->bionic

disconnect3d:
Hopefully this fixes travis


disconnect3d:
The tests still fail but at least it runs.

PR #801, opened 2020-07-01
Fix disasm call target display when symbol is known

disconnect3d:
This commit fixes the issue described in https://github.com/pwndbg/pwndbg/issues/772#issuecomment-652260420

tl;dr: when we displayed a known constant call instruction like
`call qword ptr [rip + 0x1c7d2]` when it called a known symbol we only
displayed the target address instead of the symbol.

Now, we will display only the target address.

Note that there are still cases when we can display both. This can
happen for example for a `ret` instruction (even without emulation).

This commit extends a comment around that code to give such example.

### DIFF
--- a/pwndbg/color/disasm.py
+++ b/pwndbg/color/disasm.py
@@ -53,13 +53,11 @@ def instruction(ins):
 
         # If it's a constant expression, color it directly in the asm.
         if const:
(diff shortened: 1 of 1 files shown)


disconnect3d:
Before:
image

After:
image

This can be tested on Ubuntu 20.04 docker image (or similar) via `gdb /bin/ls -ex entry`.

PR #820, opened 2020-08-23
Added installation configuration for Gentoo

TobiasFaller:
Quick and dirty implementation of the installation code for Gentoo.

Some details which might need discussion:
- [ ] The package dev-lang/python should probably be already installed when using emerge, but I'm not sure if all emerge variants require it.
- [x] Check for the sudo command added
- [ ] Maybe the dependencies should be added to the world file? (Remove the --oneshot parameter)

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -102,9 +102,11 @@ if linux; then
             ;;
         "gentoo")
             install_emerge
-            sudo() {
-                $*
-            }
+            if ! hash sudo 2>/dev/null && whoami | grep root; then
+                sudo() {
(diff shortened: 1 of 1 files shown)


disconnect3d:
@TobiasFaller do u want us to merge it/does it work? (I don't have a gentoo installation to test it)

TobiasFaller:
@disconnect3d Works on my machine. I was hoping for some feedback from other gentoo users, but maybe I'm the only one using gentoo. 😀 Feel free to merge it if you want to.

PR #825, opened 2020-09-05
format_args: display fd path

disconnect3d:
E.g. for calls like this:
```
 ► 0x555555554a57 <main+296>    call   ioctl@plt <ioctl@plt>
        fd: 0x3
        request: 0xae01
        vararg: 0x0
```

We want to display the `fd` as: `fd: 0x3 (/some/path)` fetched from
`readlink -f /proc/$PID/fd/$FD`.

### DIFF
--- a/pwndbg/arguments.py
+++ b/pwndbg/arguments.py
@@ -201,5 +201,11 @@ def format_args(instruction):
     for arg, value in get(instruction):
         code   = arg.type != 'char'
         pretty = pwndbg.chain.format(value, code=code)
+
+        # Enhance args display
+        if arg.name == 'fd' and isinstance(value, int):
+            path = pwndbg.file.readlink('/proc/%d/fd/%d' % (pwndbg.proc.pid, value))
+            pretty += ' (%s)' % path
+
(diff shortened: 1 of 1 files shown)


pwndbg/arguments.py
```diff
@@ -201,5 +201,11 @@ def format_args(instruction):
     for arg, value in get(instruction):
         code   = arg.type != 'char'
         pretty = pwndbg.chain.format(value, code=code)
+
+        # Enhance args display
+        if arg.name == 'fd' and isinstance(value, int):
+            path = pwndbg.file.readlink('/proc/%d/fd/%d' % (pwndbg.proc.pid, value))
```
zachriggle, line 207: Does this fail gracefully for qemu-user?

disconnect3d: That's why its not merged yet. I haven't checked that :P.

disconnect3d:
From https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/file.py#L66-L135 it seems it should work with remote targets and with QEMU usermode.

It seems that for unsupported targets, we should get `''` path.

I am merging this for now and see a few cases if it works properly.

PR #830, opened 2020-09-18
Compact register list for context view

TobiasFaller:
This enables to compact the register list when displayed in the context view.
The register list is aligned to improve readability.
When a chain is displayed the required space is reserved and following registers are shifted to then next alignment point / line.

See the screenshot below as an example.

This might be useful for architectures where there are a lot of registers (MIPS, RISC-V, etc.).
The compact view can be configured and enabled by the following configuration options:

- **show-compact-regs**: Enables / disables the compact view
- **show-compact-regs-align**: Sets the number of characters for each column
- **show-compact-regs-space**: The minimum number of spaces between each register

Feedback about the implementation is appreciated.

CompactRegisterView

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -357,13 +357,17 @@ pwndbg.config.Parameter('show-compact-regs-align', 20, 'the number of characters
 pwndbg.config.Parameter('show-compact-regs-space', 4, 'the number of characters separating each register')
 
 
+def calculate_padding_to_align(length, align):
+    ''' Calculates the number of spaces to append to reach the next alignment.
+        The next alignment point is given by "x * align >= length".
+    '''
+    return 0 if length % align == 0 else (align - (length % align))
+
+
 def compact_regs(regs, width):
     align = int(pwndbg.config.show_compact_regs_align)
     space = int(pwndbg.config.show_compact_regs_space)
     result = []
-
-    def calculate_padding(length):
-        return 0 if length % align == 0 else (align - (length % align))
 
     def get_text_length(text):
         '''Returns the text length while ignoring color-code escape sequences.'''
@@ -388,7 +392,7 @@ def compact_regs(regs, width):
         # register string onto the screen / display
         line_length_with_padding = line_length
         line_length_with_padding += space if line_length != 0 else 0
-        line_length_with_padding += calculate_padding(line_length_with_padding)
+        line_length_with_padding += calculate_padding_to_align(line_length_with_padding, align)
 
         # When element does not fully fit, then start a new line
         if line_length_with_padding + max(reg_length, align) > width:
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -369,24 +369,10 @@ def compact_regs(regs, width):
     space = int(pwndbg.config.show_compact_regs_space)
     result = []
 
-    def get_text_length(text):
-        '''Returns the text length while ignoring color-code escape sequences.'''
-        length = 0
-
-        escape_sequence = False
-        for char in text:
-            if escape_sequence:
-                escape_sequence = (char != 'm')
-            else:
-                escape_sequence = (char == '\033')
-                length += 0 if escape_sequence else 1
-
-        return length
-
     line = ''
     line_length = 0
     for reg in regs:
-        reg_length = get_text_length(reg)
+        reg_length = len(pwndbg.color.strip(reg))
 
         # Length of line with space and padding is required for fitting the
         # register string onto the screen / display
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -352,9 +352,76 @@ def context(subcontext=None):
             out.flush()
 
 
+pwndbg.config.Parameter('show-compact-regs', True, 'whether to show a compact register view')
+pwndbg.config.Parameter('show-compact-regs-align', 20, 'the number of characters reserved for each register and value')
+pwndbg.config.Parameter('show-compact-regs-space', 4, 'the number of characters separating each register')
+
+
+def compact_regs(regs, width):
+    align = int(pwndbg.config.show_compact_regs_align)
+    space = int(pwndbg.config.show_compact_regs_space)
+    result = []
+
+    def calculate_padding(length):
+        return 0 if length % align == 0 else (align - (length % align))
+
+    def get_text_length(text):
+        '''Returns the text length while ignoring color-code escape sequences.'''
+        length = 0
+
+        escape_sequence = False
+        for char in text:
+            if escape_sequence:
+                escape_sequence = (char != 'm')
+            else:
+                escape_sequence = (char == '\033')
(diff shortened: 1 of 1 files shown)


pwndbg/commands/context.py
```diff
@@ -352,9 +352,76 @@ def context(subcontext=None):
...
+
+
 def context_regs(target=sys.stdout, with_banner=True, width=None):
     banner = [pwndbg.ui.banner("registers", target=target, width=width)]
-    return banner + get_regs() if with_banner else get_regs()
+    regs = get_regs()
+
+    if pwndbg.config.show_compact_regs:
+        if width is None:
+            _height, width = pwndbg.ui.get_window_size(target=target)
```
disconnect3d: Shall we move this width condition to the top, so in case of `width=None` we do not fetch this twice?

(if `width=None` the `pwndbg.ui.banner(...)` called in first line of this function also gets the window size: https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/ui.py#L38-L39)

pwndbg/commands/context.py
```diff
@@ -352,9 +352,76 @@ def context(subcontext=None):
...
+def compact_regs(regs, width):
+    align = int(pwndbg.config.show_compact_regs_align)
+    space = int(pwndbg.config.show_compact_regs_space)
+    result = []
+
+    def calculate_padding(length):
+        return 0 if length % align == 0 else (align - (length % align))
+
+    def get_text_length(text):
+        '''Returns the text length while ignoring color-code escape sequences.'''
```
disconnect3d: Is this algorithm correct? Maybe we should use [`len(pwndbg.color.strip(text))`](https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/color/__init__.py#L76-L77) instead?

pwndbg/commands/context.py
```diff
@@ -352,9 +352,76 @@ def context(subcontext=None):
...
+
+def compact_regs(regs, width):
+    align = int(pwndbg.config.show_compact_regs_align)
+    space = int(pwndbg.config.show_compact_regs_space)
+    result = []
+
+    def calculate_padding(length):
+        return 0 if length % align == 0 else (align - (length % align))
+
+    def get_text_length(text):
```
disconnect3d: Can we move the two functions (`calculate_padding` and `get_text_length`) outside of `compact_regs` function, so they are not re-created each time the `compact_regs` is called?

For more context:
image

pwndbg/commands/context.py
```diff
@@ -352,9 +352,76 @@ def context(subcontext=None):
...
+        '''Returns the text length while ignoring color-code escape sequences.'''
+        length = 0
+
+        escape_sequence = False
+        for char in text:
+            if escape_sequence:
+                escape_sequence = (char != 'm')
+            else:
+                escape_sequence = (char == '\033')
+                length += 0 if escape_sequence else 1
```
disconnect3d: Could also be `length += int(escape_sequence)`, assuming its always `bool`.

(or even without the `int(...)` but that's not necessarily verbose?)

disconnect3d:
This looks awesome. See comments :)

pwndbg/commands/context.py
TobiasFaller: Ok, this is a good idea.

pwndbg/commands/context.py
TobiasFaller: Oh, wow. I didn't find this function. I will use `pwndb.color.strip` function as it already does the job.

pwndbg/commands/context.py
TobiasFaller: This is a good idea. I will propably rename the function to `calculate_padding_to_align` too in this step.
The `get_text_length` function will be removed as it is redundant.

TobiasFaller:
@disconnect3d Thank you for the feedback. I will fix the issues in the next days.

TobiasFaller:
@disconnect3d I have now applied the changes. If there are further issues or ideas feel free to add more feedback.

zachriggle:
Thanks for taking care of this, @disconnect3d.  This has always been a thing that GEF has that I wished we had (just for feature parity, I want Pwndbg to have ALL THE FEATURES) so I'm glad this was merged.

lonnywong:
How to align this into two columns ?

image

TobiasFaller:
> How to align this into two columns ?
> 
> image

There is the configuration option **show-compact-regs-align** which defines the size for one column.
So if you Terminal is 80 characters wide (just as an example) you can set this option to 40 and it should align the register view in two columns. You have to enable the alignment by setting the configuration option **show-compact-regs** to _on_.

lonnywong:
image

My terminal is 238 characters wide.
Set show-compact-regs-align 119.
The RDI and R9 is not aligned.

lonnywong:
1. config how many columns.
2. config the minimum width.
3. auto span columns while it is necessary.

maybe easy to use.

TobiasFaller:
@lonnywong It seems that it only aligned two registers (RBX and RBP) correctly. Have you any setup which you can share for me to reproduce the issue?

lonnywong:
@TobiasFaller 

```sh
mkdir test && cd test

cat <<EOF >Dockerfile
FROM ubuntu:latest
ENV DEBIAN_FRONTEND=noninteractive
RUN apt-get update && apt-get install -y git && git clone https://github.com/pwndbg/pwndbg.git ~/pwndbg && cd ~/pwndbg && ./setup.sh
EOF

docker build -t ubuntu:test .
sudo sysctl -w vm.mmap_min_addr=0
docker run --rm -it ubuntu:test /bin/bash

cd ~

cat <<EOF >test.c
#include <locale.h>
#include <stdio.h>
#include <wchar.h>
#include <string.h>
#include <stdlib.h>
#include <unistd.h>
#include <sys/mman.h>

char code[] = "\x31\xc0\x48\xbb\xd1\x9d\x96\x91\xd0\x8c\x97\xff\x48\xf7\xdb\x53\x54\x5f\x99\x52\x57\x54\x5e\xb0\x3b\x0f\x05";

int main(void) {
    void *p = mmap((void *)0, 0x1000, 7, 50, -1, 0);
    printf("mmap : %p\n", p);
    memset(p, '\x90', 0xf00);
    strcpy(p, code);

    setlocale(6, "en_US.UTF-8");
    printf("%d\n", fgetwc(stdin));
    return 0;
}
EOF

gcc -o test test.c

gdb ./test
r
Ctrl+c
set show-compact-regs on
set show-compact-regs-align 119
context
```

lonnywong:
I want two columns, and I have a few terminals, the wide is a little different, could you calculate the  show-compact-regs-align automatic?

TobiasFaller:
@lonnywong Thank you for the dockerfile. I will try to reproduce the issue in the next days. While I'm fixing this problem I also might just add another option which enables the user to split the view into a fixed number of columns. This should then cover your use-case.
Update: I will get back to this after Feb the 18th.

PR #832, opened 2020-09-25
Fixed bug: bins gets the wrong pointer offset

Things-Z:
In python3, the dict created using 'dict({renames. Get (key, key): 0 for key in val.keys()})' is unordered, causing 'chunk_key_offset' to return an error offset


disconnect3d:
Thanks!

PR #837, opened 2020-10-07
fix prev chunk size check

lonnywong:
**before fix :**
image

**after fix :**
image

### DIFF
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -932,10 +932,10 @@ def try_free(addr):
                 finalize(errors_found, returned_before_error)
                 return
 
-            if unsigned_size(prev_chunk['size']) != prev_size:
+            if prev_chunk_size != prev_size:
                 err = 'corrupted size vs. prev_size while consolidating\n'
                 err += 'prev_size field is 0x{:x}, prev chunk at 0x{:x}, prev chunk size is 0x{:x}'
-                err = err.format(prev_size, prev_chunk_addr, unsigned_size(prev_chunk['size']))
+                err = err.format(prev_size, prev_chunk_addr, prev_chunk_size)
                 print(message.error(err))
                 errors_found += 1
             else:


disconnect3d:
@GrosQuildu can you check that one?

disconnect3d:
@lonnywong can you give some more context? it's hard to judge if the  change is correct from the screenshots :P

lonnywong:
> @lonnywong can you give some more context? it's hard to judge if the change is correct from the screenshots :P

```c
#include <stdio.h>
#include <stdlib.h>

int main() {
	void *p1 = malloc(0x500);
	void *p2 = malloc(0x500);
	free(p1);
	printf("pwndbg> try_free %p\n", p2);
	getchar();
	return 0;
}
```

pwndbg/commands/heap.py
```diff
@@ -932,7 +932,7 @@ def finalize(errors_found, returned_before_error):
                 finalize(errors_found, returned_before_error)
                 return
 
-            if unsigned_size(prev_chunk['size']) != prev_size:
+            if chunksize(unsigned_size(prev_chunk['size'])) != prev_size:
```
GrosQuildu:
```suggestion
            if prev_chunk_size != prev_size:
```

pwndbg/commands/heap.py
```diff
@@ -932,7 +932,7 @@ def finalize(errors_found, returned_before_error):
...
                 err = 'corrupted size vs. prev_size while consolidating\n'
                 err += 'prev_size field is 0x{:x}, prev chunk at 0x{:x}, prev chunk size is 0x{:x}'
                 err = err.format(prev_size, prev_chunk_addr, unsigned_size(prev_chunk['size']))
```
GrosQuildu:
```suggestion
                err = err.format(prev_size, prev_chunk_addr, prev_chunk_size)
```

GrosQuildu (changes requested):
Cool, thanks. We may simplify it and merge.

pwndbg/commands/heap.py
lonnywong: Good point.

pwndbg/commands/heap.py
lonnywong: Good point.

GrosQuildu (approved):

disconnect3d:
Thanks!

PR #839, opened 2020-10-15
Compact, [big-endian] hexdump

takimata:
I had the need to be able to directly copy pointer values from the hexdump.
This pull requests adds 4 config options (all inactive by default) to allow this.

The first hexdump shows the current default and the last one the setup I needed:
screenshot

Please let me know what you think!

### DIFF
--- a/pwndbg/color/hexdump.py
+++ b/pwndbg/color/hexdump.py
@@ -12,6 +12,9 @@ config_special   = theme.ColoredParameter('hexdump-special-color', 'yellow', 'co
 config_offset    = theme.ColoredParameter('hexdump-offset-color', 'none', 'color for hexdump command (offset label)')
 config_address   = theme.ColoredParameter('hexdump-address-color', 'none', 'color for hexdump command (address label)')
 config_separator = theme.ColoredParameter('hexdump-separator-color', 'none', 'color for hexdump command (group separator)')
+config_highlight_group_lsb = theme.Parameter('hexdump-highlight-group-lsb', 'underline',
+                                             'highlight LSB of each group. Applies only if hexdump-adjust-group-endianess'
+                                             ' actually changes byte order.')
 
 def normal(x):
     return generateColorFunction(config.hexdump_normal_color)(x)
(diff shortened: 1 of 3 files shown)


pwndbg/color/hexdump.py
```diff
@@ -12,6 +12,9 @@
 config_offset    = theme.ColoredParameter('hexdump-offset-color', 'none', 'color for hexdump command (offset label)')
 config_address   = theme.ColoredParameter('hexdump-address-color', 'none', 'color for hexdump command (address label)')
 config_separator = theme.ColoredParameter('hexdump-separator-color', 'none', 'color for hexdump command (group separator)')
+config_highlight_group_lsb = theme.Parameter('hexdump-highlight-group-lsb', 'underline',
```
disconnect3d, line 15: this could be even more configurable, like 'highlight X lsb bytes' but its also fine as it is

pwndbg/commands/hexdump.py
```diff
@@ -17,6 +17,13 @@
 pwndbg.config.Parameter('hexdump-bytes',
                          64,
                          'number of bytes printed by hexdump command')
+pwndbg.config.Parameter('hexdump-group-width',
+                         4,
+                         "number of bytes grouped in hexdump command. If -1, the architecture's pointer size is used.")
```
disconnect3d:
```suggestion
                         "number of bytes grouped in hexdump command (If -1, the architecture's pointer size is used)")
```

This will make it more readable when the message from `set param val` is printed

pwndbg/commands/hexdump.py
```diff
@@ -17,6 +17,13 @@
...
+pwndbg.config.Parameter('hexdump-group-use-big-endian',
+                         False,
+                         'Use big-endian within each group of bytes. Only applies to raw bytes, not the ASCII part. '
+                         'Also see hexdump-highlight-group-lsb.')
```
disconnect3d:
```suggestion
                         'See also hexdump-highlight-group-lsb.')
```

disconnect3d (approved):

disconnect3d:
That looks awesome, thanks :)

takimata:
Cool, thanks for merging!

PR #850, opened 2020-10-23
 fix for ubuntu 20.04

fishykz:


disconnect3d:
Thx

zachriggle:
I think this should be rc3, if anything. Rc1 is probably fine, but there are later RCs without the MIPS regression preventing 1.0.3 usage.

zachriggle:
Follow up,do we (pwn tools) have an issue as to why the latest unicorn is broken?  Possibly with a mechanism to fix it? I k ow @heapcrash might have time to look into it.

PR #851, opened 2020-10-25
fix UnicodeDecodeError: 'utf-8' codec can't decode byte

kotee4ko:
As for me the fix was to add `replace` arg to decode() foo.
```
  GNU nano 5.1                                   /opt/pwndbg/pwndbg/net.py                                             

-_line 142:     data = pwndbg.file.get("/proc/net/unix").decode()
+_line 142:     data = pwndbg.file.get("/proc/net/unix").decode("utf-8", "replace")
```

this is my first pull request ever :3

So, to not flood with pull requests I'll speak here.

I have an idea about sync binary ninja with pwngdb.
but, my python skills is not as good as needed.

in general, there are one cool plug-in to BN - called VulnFanatic. Here is it. https://github.com/Martyx00/VulnFanatic
And, gef have module to sync BN with GDB via customized gdb<-->ida protocol.
Here is it https://github.com/hugsy/gef-binja

It is not working as well, as expected, but it works a bit. But, gef don't have a lot's of commands, such as `nextcall` or `nextjump`.
And don't have `vis`, etc. So, personally I prefer PWNGDB.

And, i think it wiil be mega-super-cool to integrate such feature into pwngdb. We can do statically binary analyze, find potential vulnerable places (VulnFanatic works as expected - quite well!) - and then - just pass throe favored debugger on one side and look on GUI backtrace with all args on the other. It will be really perfect weapon to shoot the bug.

Sorry if I make mistakes, English is not my native.

Best regards.


disconnect3d:
Hmm it seems you try to merge dev into stable in here too. Can you clone the repo again, make the change on dev branch and send a PR from it?

disconnect3d:
Regarding BN integration, I think that would be awesome.

disconnect3d:
@kotee4ko pls fix the commits on ur branch and open another PR

PR #854, opened 2020-11-13
ropgadget: fix path export.

ghost:
Hello dear developers!

By default, the Python bin path was not exported in the Dockerfile, so the ROPgadget could not start (```rop``` instruction).
In such a simple way, we can fix the launch in the container.

Thanks in advance!


disconnect3d:
Thanks!

PR #857, opened 2020-11-22
Added comment command 

GGyul-E:
This "comm" command makes comment beside of assembly code. Will make easy to analyse and by giving ".gdb_comments" file, other analyst will be able to see comments in binary.
# Example 
Explanation of "comm" command.
image

By using "comm" command, we can comment at $pc
image

By using "-addr" option, we can choose address to comment.
image

These comments are saved in ./.gdb_comments
image

### DIFF
--- a/pwndbg/commands/comments.py
+++ b/pwndbg/commands/comments.py
@@ -6,23 +6,45 @@ parser = argparse.ArgumentParser(description="Put comments in assembly code")
 parser.add_argument("--addr", metavar='address', default=None, type=str, help="Address to write comments")
 parser.add_argument("comment", type=str, default=None,  help="The text you want to comment")
 
+file_lists = {} # This saves all comments.
+
 @pwndbg.commands.ArgparsedCommand(parser)
 @pwndbg.commands.OnlyWhenRunning
 def comm(addr=None, comment=None):
     if addr is None:
         addr = hex(pwndbg.regs.pc)
     try : 
-        f = open(".gdb_comments", "a+")
-    except :
+        with open(".gdb_comments", "a+") as f:
+            target = int(addr,0)
+
+            if not pwndbg.memory.peek(target):
+                print(message.error("Invalid Address %#x" % target))
+            
+            else:
+                f.write("file:%s=" % pwndbg.proc.exe)
+                f.write("%#x:%s\n" % (target, comment))
+                if not pwndbg.proc.exe in file_lists.keys():
+                    file_lists[pwndbg.proc.exe] = {}
+                file_lists[pwndbg.proc.exe][hex(target)] = comment
+    except FileNotFoundError:
         print(message.error("Permission denied to create file"))
-    else :
-        target = int(addr,0)
 
-        if not pwndbg.memory.peek(target):
-            print(message.error("Invalid Address %#x" % target))
-        
-        else:
-            f.write("file:%s=" % pwndbg.proc.exe)
-            f.write("%#x:%s\n" % (target, comment))
+def init():
+    try :
+        with open(".gdb_comments","r") as f:
+            text = f.read()
+            text = text.split("\n")
+            for i in range(len(text)-1):
+                text1, text2 = text[i].split("=")
 
-        f.close()
+                # split Filename, comments
+                filename = text1.split(":")[1]
+                addr_comm = text2.split(":")
+
+                if not filename in file_lists.keys():
+                    file_lists[filename] = {}
+
+                file_lists[filename][addr_comm[0]] = addr_comm[1]
+
+    except :
+        pass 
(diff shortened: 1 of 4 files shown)


GGyul-E:
@disconnect3d Can you check this please? Is there anything to edit?

pwndbg/commands/comments.py
```diff
@@ -0,0 +1,28 @@
...
+from pwndbg.color import message
+
+parser = argparse.ArgumentParser(description="Put comments in assembly code")
+parser.add_argument("-addr", metavar='address', default=None, type=str, help="Address to write comments")
+parser.add_argument("comment", type=str, default=None,  help="The text you want to comment")
+
+@pwndbg.commands.ArgparsedCommand(parser)
+@pwndbg.commands.OnlyWhenRunning
+def comm(addr=None, comment=None):
+    if(addr == None):
```
disconnect3d:
```suggestion
    if addr is None:
```

pwndbg/commands/comments.py
```diff
@@ -0,0 +1,28 @@
+import argparse
+import pwndbg.commands
+from pwndbg.color import message
+
+parser = argparse.ArgumentParser(description="Put comments in assembly code")
+parser.add_argument("-addr", metavar='address', default=None, type=str, help="Address to write comments")
```
disconnect3d: Shouldn't this be:
```suggestion
parser.add_argument("--addr", metavar='address', default=None, type=str, help="Address to write comments")
```

pwndbg/commands/comments.py
```diff
@@ -0,0 +1,28 @@
...
+parser.add_argument("comment", type=str, default=None,  help="The text you want to comment")
+
+@pwndbg.commands.ArgparsedCommand(parser)
+@pwndbg.commands.OnlyWhenRunning
+def comm(addr=None, comment=None):
+    if(addr == None):
+        addr = hex(pwndbg.regs.pc)
+    try : 
+        f = open(".gdb_comments", "a+")
```
disconnect3d: Can we use `with open(...) as f:` here instead?

pwndbg/commands/nearpc.py
```diff
@@ -162,6 +162,37 @@ def nearpc(pc=None, lines=None, to_string=False, emulate=False):
...
+                if not filename in file_lists.keys():
+                    file_lists[filename] = {}
+
+                file_lists[filename][addr_comm[0]] = addr_comm[1]
+
+# Have to Make : Make dictionary with filename for key, comment list for value.
+# So, in printing comments, check in dictionary and get lists. And check address for what to print.
+
+            try:
+                line += " "*8 + C.comment(file_lists[pwndbg.proc.exe][hex(instr.address)])
```
disconnect3d: Maybe just two spaces to preserve more space?

pwndbg/commands/nearpc.py
```diff
@@ -162,6 +162,37 @@ def nearpc(pc=None, lines=None, to_string=False, emulate=False):
             if syscall_name:
                 line += ' <%s>' % N.syscall_name(syscall_name)
 
+        # For Comment Function
+        try : 
+            f = open(".gdb_comments","r")
```
disconnect3d: Maybe `with open(..)`

disconnect3d:
> @disconnect3d Can you check this please? Is there anything to edit?

I like the idea. I left some comments regarding the code.

Also, not sure if we should re-read the file on each `nearpc` call: maybe we should only read it once during gdb init and cache the comments in a global dict?

GGyul-E:
> > @disconnect3d Can you check this please? Is there anything to edit?
> 
> I like the idea. I left some comments regarding the code.
> 
> Also, not sure if we should re-read the file on each `nearpc` call: maybe we should only read it once during gdb init and cache the comments in a global dict?

Thanks for the suggestion. But if we read comment file only once, it can't update new comments that we wrote in runtime.
However if we set flag variable to check if we wrote new comments in runtime, we can check if we need to re-read again.
This idea just came up to me. How do you think about this?

disconnect3d:
> However if we set flag variable to check if we wrote new comments in runtime, we can check if we need to re-read again.
> This idea just came up to me. How do you think about this?

I think this may be good to add in order to not degrade performance when someone doesn't want to use the comment feature.

GGyul-E:
@disconnect3d 
I Fixed it to only read comment file at init as you said. But I made `comm` command to add comments in global dictionary while runtime. Now it will get pretty better performance than before. Is there anything more to fix?

disconnect3d:
@GGyul-E Added a small style fixes commit and merged. Thanks!

disconnect3d:
Btw there may be some issues with PIE/ASLR binaries (comment address changes?), but this can usually be solved by disabling ASLR. Just saying, not sure if this issue can be fixed easily.

PR #859, opened 2020-11-26
fix for i386 libc6-dbg package installation

ghost:
Hello dear developers!

So, to install i386 package, we firstly need to add i386 architecture.

Thanks in advance!


ghost:
@disconnect3d everything is fine? Do I need to add something?

disconnect3d:
Thx!

PR #860, opened 2020-12-06
The disassembly flavor is hard-coded. It does not change from Intel to AT&T

airvzxf:
# disassembly-flavor [intel/att]

## Cases:
- When the user `set disassembly-flavor att` inside of the `.gdbinit` file.  It's not recognized by `pwndbg`.
- When the user manually `set disassembly-flavor att` running the GDB.  It's not recognized by `pwndbg`.

## Example of this fix

`set disassembly-flavor att`

`context`

```text
─────────────────────────────────────────────────────────────[ DISASM ]─────────────────────────────────────────────────────────────
 ► 0x401000 <_start>       movl   $1, %eax
   0x401005 <_start+5>     movl   $0, %ebx
   0x40100a <_start+10>    int    $0x80
```

`set disassembly-flavor intel`

`context`

```text
─────────────────────────────────────────────────────────────[ DISASM ]─────────────────────────────────────────────────────────────
 ► 0x401000 <_start>       mov    eax, 1
   0x401005 <_start+5>     mov    ebx, 0
   0x40100a <_start+10>    int    0x80
```

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -475,8 +475,8 @@ def context_disasm(target=sys.stdout, with_banner=True, width=None):
     flavor = gdb.execute('show disassembly-flavor', to_string=True).lower().split('"')[1]
     syntax = pwndbg.disasm.CapstoneSyntax[flavor]
 
-    cache_get_disassembler_cached = pwndbg.memoize.reset_on_objfile(pwndbg.disasm.get_disassembler_cached)
-    cs = list(cache_get_disassembler_cached.__dict__['cache'].values())[-1]
+    # Get the Capstone object to set disassembly syntax
+    cs = next(iter(pwndbg.disasm.get_disassembler_cached.cache.values()))
 
     if cs.syntax != syntax:
         pwndbg.memoize.reset()
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -472,6 +472,15 @@ Unicorn emulation of code near the current instruction
 code_lines = pwndbg.config.Parameter('context-code-lines', 10, 'number of additional lines to print in the code context')
 
 def context_disasm(target=sys.stdout, with_banner=True, width=None):
+    flavor = gdb.execute('show disassembly-flavor', to_string=True).lower().split('"')[1]
+    syntax = pwndbg.disasm.CapstoneSyntax[flavor]
+
+    cache_get_disassembler_cached = pwndbg.memoize.reset_on_objfile(pwndbg.disasm.get_disassembler_cached)
+    cs = list(cache_get_disassembler_cached.__dict__['cache'].values())[-1]
+
+    if cs.syntax != syntax:
+        pwndbg.memoize.reset()
+
     banner = [pwndbg.ui.banner("disasm", target=target, width=width)]
     emulate = bool(pwndbg.config.emulate)
     result = pwndbg.commands.nearpc.nearpc(to_string=True, emulate=emulate, lines=code_lines // 2)
--- a/pwndbg/disasm/__init__.py
+++ b/pwndbg/disasm/__init__.py
@@ -45,6 +45,11 @@ CapstoneMode = {
     8: CS_MODE_64
 }
 
+CapstoneSyntax = {
+    'intel': CS_OPT_SYNTAX_INTEL,
+    'att': CS_OPT_SYNTAX_ATT
+}
+
 # For variable-instruction-width architectures
 # (x86 and amd64), we keep a cache of instruction
 # sizes, and where the end of the instruction falls.
@@ -71,6 +76,10 @@ def get_disassembler_cached(arch, ptrsize, endian, extra=None):
     mode |= CapstoneEndian[endian]
 
     cs = Cs(arch, mode)
+
+    flavor = gdb.execute('show disassembly-flavor', to_string=True).lower().split('"')[1]
+    cs.syntax = CapstoneSyntax[flavor]
+
     cs.detail = True
     return cs
 
--- a/pwndbg/disasm/__init__.py
+++ b/pwndbg/disasm/__init__.py
@@ -75,11 +75,10 @@ def get_disassembler_cached(arch, ptrsize, endian, extra=None):
 
     mode |= CapstoneEndian[endian]
 
-    cs = Cs(arch, mode)
-
     flavor = gdb.execute('show disassembly-flavor', to_string=True).lower().split('"')[1]
-    cs.syntax = CapstoneSyntax[flavor]
 
+    cs = Cs(arch, mode)
+    cs.syntax = CapstoneSyntax[flavor]
     cs.detail = True
     return cs
 


pwndbg/disasm/__init__.py
```diff
@@ -71,6 +76,10 @@ def get_disassembler_cached(arch, ptrsize, endian, extra=None):
     mode |= CapstoneEndian[endian]
 
     cs = Cs(arch, mode)
+
+    flavor = gdb.execute('show disassembly-flavor', to_string=True).lower().split('"')[1]
```
disconnect3d, line 78: I hope this works for other gdb languages :( lets leave it as it is for now.

pwndbg/commands/context.py
```diff
@@ -472,6 +472,15 @@ def get_regs(*regs):
 code_lines = pwndbg.config.Parameter('context-code-lines', 10, 'number of additional lines to print in the code context')
 
 def context_disasm(target=sys.stdout, with_banner=True, width=None):
+    flavor = gdb.execute('show disassembly-flavor', to_string=True).lower().split('"')[1]
+    syntax = pwndbg.disasm.CapstoneSyntax[flavor]
+
+    cache_get_disassembler_cached = pwndbg.memoize.reset_on_objfile(pwndbg.disasm.get_disassembler_cached)
+    cs = list(cache_get_disassembler_cached.__dict__['cache'].values())[-1]
```
disconnect3d: Does this really work? The `reset_on_objfile` is a class and the above line will create an instance, so it should have `cache` field re-created each time we execute this?

disconnect3d: I wonder if we can overwriite GDB parameter with our own or just hook one, and just clear caches if its being set...

airvzxf: > Does this really work? The `reset_on_objfile` is a class and the above line will create an instance, so it should have `cache` field re-created each time we execute this?

Yes, I tested on Dockerfile provided in your project and I modified these lines in my Arch Linux with the package installed `pacman -S pwndbg` (community/pwndbg 2020.07.23-2), and it's working.

No, it didn't recreate every time, I'm prevented this comparing the cache value for CS.syntax versus the GDB flavor, means it only re-create the cache every time it's different, means every time you change manually.

I know it look a dirty solution but I didn't find other way, I have some ideas but I'll share in the next comment.

airvzxf: > I wonder if we can overwriite GDB parameter with our own or just hook one, and just clear caches if its being set...

It' sounds good, create the pwndbg command, the only missing part will be when the user set the flavor in the `.gdbinit`, just this file is the first and the only time that the flavor is created, except if the caches are cleaned. Also, it will be noisy that I have this option but I need to set the flavor manually.

Let me create a YouTube video and show how these changes are working.

disconnect3d: > Let me create a YouTube video and show how these changes are working.

Thanks <3. Fwiw there's also https://asciinema.org/

airvzxf: > > Let me create a YouTube video and show how these changes are working.
> 
> Thanks <3. Fwiw there's also https://asciinema.org/

Awesome, I tried and it's an excellent idea, record the STD Output and record the time waiting from one screen to another.

airvzxf: The volume of my voice is low. I used OSB but looks like its something missing in the audio settings.

https://youtu.be/QAL5AVYlhI4

disconnect3d: I had some time to revisit this today and, AWESOME video, thanks!

My initial concern was more about that we call `reset_on_objfile` here: it seems unnecessary?

It seems we should be able to do just this:
```py
# Get the Capstone object to set disassembly syntax
cs = next(iter(pwndbg.disasm.get_disassembler_cached.cache.values()))
```

This way, we don't create the `reset_on_objfile` object/instance and also a list (since we do `next(iter(<dict-values object>))`).

Can you check if that works fine? (maybe there's a reason I miss we need to create the memoize object)
If that works fine, I think we can merge this.

For other of your comments: I think making a command like `gdb flavor <flavor>` is not good as people would rather use the default/standard `set disassembly-flavor <...>`.

Also, the `pwndbg.config.Parameter` class we have, makes so we can add additonal "gdb parameters" that can be set/shown using the standard GDB's `set ...` or `show ...` command. However, I think that if we try to overwrite GDB's original parameter this way, we won't be able to use/set the original's value: so I think overwriting it may not be the best idea. Sooo I would just make it as it is in this PR :).

airvzxf: You created an AWESOME app, thanks for it.

Your solution is beautiful, much better than my implementation. I applied and It works. I'll update the PR.

Yes, I'm totally agreed the other solutions not looks good, but this solution is working and keeps the code quality high.

enedil:
@disconnect3d 
This breaks gdb on arm-eabi:
```
gdb.error: Undefined show command: "disassembly-flavor".  Try "help show".
```

airvzxf:
Hi @enedil,

I wonder if is possible to replicate the `arm-eabi` to debug that in my local, some docker image, vagrant, code example, GCC instructions.
Are you using directly in Android? I'm using Arch Linux.

enedil:
> Hi @enedil,
> 
> I wonder if is possible to replicate the `arm-eabi` to debug that in my local, some docker image, vagrant, code example, GCC instructions.
> Are you using directly in Android? I'm using Arch Linux.

I'm doing remote debugging for an ARM microcontroller. I don't have docker, but you can compile gdb from source with `--target=arm-eabi`. Gdb is used from desktop Fedora. I'm sorry, I don't know if I can provide better setup, because the problem arises only after the regular disassembly is to be shown, and that happens only after connecting to the board. But afaik, the problem is that on ARM there are no multiple flavors, so there is no `disassembly-flavor` command. In fact, if I hardcode `'intel'` instead of `gdb.execute('show disassembly-flavor', to_string=True).lower().split('"')[1]`, everything seems to work once again. So maybe just wrap this line into try except clause and set it to `intel` if command fails?

airvzxf:
I have this fix and I want to send a merge request, could you try to implement it? If all is working, could you replay? 

Files:
`pwndbg/commands/context.py`
`pwndbg/disasm/__init__.py`
```python
    try:
        flavor = gdb.execute('show disassembly-flavor', to_string=True).lower().split('"')[1]
    except:
        flavor = 'intel'
```

enedil:
I checked and apparently it works. Maybe the raw `except` should be exchanged with `except gdb.error`.

airvzxf:
Yes, specify the exception it's much better and it's to aligned the best practices. Could you try in your `pwndbg` the `gdb.error` exception? If it works, I'll add to the new PR.

```python
    try​:
        ​flavor​ ​=​ ​gdb​.​execute​(​'show disassembly-flavor'​, ​to_string​=​True​).​lower​().​split​(​'"'​)[​1​]
    ​except​ gdb.error:
        flavor​ ​=​ ​'intel'
```

airvzxf:
GDB
23.3.2.2 Exception Handling

https://sourceware.org/gdb/current/onlinedocs/gdb/Exception-Handling.html

airvzxf:
We can do it something like this:

```python
    try:
        flavor​​ ​​=​​ ​​gdb​​.​​execute​​(​'​show​ ​disassembly​-​flavor​'​, ​to_string​=​True​).​lower​().​split​(​'​"'​)[​​1​​]
    except gdb.error as e:
        if str(e) == """​gdb.error: Undefined show command: "disassembly-flavor".  Try "help show".""":
            flavor​​ ​​=​​ ​​'intel'
        else:
            raise
```

Reference:
https://stackoverflow.com/questions/13531247/python-catching-specific-exception#13531310

enedil:
I didn't experience errors, this approach seems reasonable.

airvzxf:
Hi again, the problem was fixed and merged to the main branch in this PR: https://github.com/pwndbg/pwndbg/pull/889.

PR #862, opened 2020-12-06
Improving Dockerfile.

airvzxf:
Talking about Docker.

First: It is a solution is about the Timezone, if it's not set up, it will crash the installation of the container.

Second: It is an improvement about the layers in the container, if every instruction is `RUN` command, you will notice that after the execution of the `./setup` script, it is upgrading the package every time. A solution is group all the minimal setup commands.

Third: It is a solution for the missing package Go, which is required to execute the `./test.sh` script.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,11 +9,20 @@
 #
 FROM ubuntu:20.04
 
-ADD . /pwndbg/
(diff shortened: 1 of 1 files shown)


disconnect3d:
Thanks!

PR #864, opened 2020-12-08
Allow return offsets and use it for 'start' method.

### LINKED ISSUE #793: 'start' command behavior has changed
`git bisect` suggests this was introduced in 8f33ec480f47c038e0077cab67153f3ec3be9181

Prior to this, when debugging programs built with `gcc` using the `-ggdb` flag, pwndbg would break at the program's `main()` function immediately after the `start` command was issued. Since 8f33ec480f47c038e0077cab67153f3ec3be9181 the `start` command breaks at `_start()` the 1st time it's issued, and must be issued a 2nd time before `main()` is reached, after which it consistently stops at `main()`. See below for an example.

A simple program:
```
// main.c
#include <stdio.h>

void main(void) {
    printf("Hello World!\n");
}
```

Built and run with:
```
$ gcc -ggdb main.c
$ gdb a.out
pwndbg> set context-sections code
pwndbg> start
```

The `set context-sections` command is used here to truncate output for the screenshots and isn't part of the bug.
Previously, `main()` was reached immediately:
start_sucess

After 8f33ec480f47c038e0077cab67153f3ec3be9181 the `start` command breaks at `_start()` first, then it must be issued a 2nd time before `main()` is reached and the "source" panel appears. After this, every invocation of `start` will break at the `main()` function:
start_fail

Perhaps some folk would rather break earlier in their code when issuing the `start` command, but it's odd that when starting the program for the 2nd time a different result is obtained, since `start` tells GDB to run the program over from scratch.

eldipa:
The 8f33ec4 made `pwndbg.symbol.address` to discard addresses
of symbols not mapped.

Unfortunately this broke pwndbg's `start`.

GDB's `start` puts a temporal break in `main` and pwndbg's `start` does
the same but when GDB returns the address of `main`, it returns an
offset the first time because the symbol was not mapped yet.

The offset is then discarded and pwndbg doesn't put the breakpoint when
it should.

This PR fixes pwndbg's `start` allowing `pwndbg.symbol.address` to
return offsets instead of addresses: GDB will resolve the correct
address when it builds the breakpoint and pwndbg's `start` will behave
like GDB's `start`.

Closes #793

### DIFF
--- a/pwndbg/commands/start.py
+++ b/pwndbg/commands/start.py
@@ -49,7 +49,7 @@ def start(args=None):
                 "_init"]
 
     for symbol in symbols:
-        address = pwndbg.symbol.address(symbol)
+        address = pwndbg.symbol.address(symbol, allow_unmapped=True)
 
         if not address:
             continue
--- a/pwndbg/symbol.py
+++ b/pwndbg/symbol.py
@@ -190,7 +190,7 @@ def get(address, gdb_only=False):
     return ''
 
 @pwndbg.memoize.reset_on_objfile
-def address(symbol):
+def address(symbol, allow_unmapped=False):
     if isinstance(symbol, int):
         return symbol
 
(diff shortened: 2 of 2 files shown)


disconnect3d:
Thanks!

PR #829, opened 2020-09-17
Draft: RISC-V support

TobiasFaller:
This will be my attempt to implement RISC-V support (#735) for Pwndbg.
Capstone has RISC-V support only in the `next` branch, so this pull request will stay open until the new release version 5.0.0 is released.
Feel free to give feedback.

PR #866, opened 2020-12-14
Changed register list to use precomputed tuples

TobiasFaller:
The register list for each architecture in pwndbg/regs.py is now precomputed (#829).


TobiasFaller:
@disconnect3d Followup to #829, will probably test if it is still functional tomorrow.
Update: Will check after Feb. 18th.

TobiasFaller:
@disconnect3d Ok, tested it with amd64 binary in a Docker container. Seems to work fine, so I'm removing the WIP flag now.

disconnect3d:
Thx

PR #869, opened 2020-12-22
Enhance Dockerfile to use the latest GDB and install a non-root user

WeSeekAfterKnowledge:
This Dockerfile is also based on Ubuntu 20.04, but installs the very latest
version of GDB, so that things can be tested against it.

It also sets up the locale properly so that there are no locale warnings
when launching GDB.

Finally, a pwndbg user is created so that we can test things as non-root.

The main makefile targets are "all" (which builds to Docker image) and
"run" which runs the resulting Docker image.

In the future, I plan to set up automated port forwarding so that the Docker
instance can connect to IDA Pro running on the host machine.


disconnect3d:
This conflicts with recent changes to dockerfile. Also, is the idea here to be able to run pwndbg for work or for tests? I think that our current dockerfile is there only for tests run on CI (though, they still break :().

WeSeekAfterKnowledge:
This shouldn't conflict, since the Dockerfile is at a completely different path.  `Dockerfile` vs `extras/docker/Dockerfile`.  It did remove the old `Dockerfile`, but that can be re-added if it's used for testing purposes.

The intent for this Dockerfile is to be able to run Pwndbg, specifically to assist testing / refactoring of the IDA Pro integration for v7.x

disconnect3d:
> This shouldn't conflict, since the Dockerfile is at a completely different path.

But it does: I guess a Dockerfile change was included by mistake and it is impossible to merge this due to this conflict:
image

Can you remove the `/Dockerfile` change from this PR and maybe add some `extra/README.md` that would describe the purpose of the added files? Then we can merge it :)

disconnect3d:
@WeSeekAfterKnowledge ping

disconnect3d:
Closing as its stale

PR #874, opened 2020-12-30
Remove unimplemented dlmalloc

### LINKED ISSUE #871: pwndbg gets stuck in GDB 10.1
### Description
pwndbg gets stuck in GDB 10.1

### Steps to reproduce
Screenshot_20201227_212816

### My setup

```
Gdb:      10.1
Python:   3.8.6 (default, Sep 30 2020, 04:00:38)  [GCC 10.2.0]
Pwndbg:   1.1.0
Capstone: 4.0.1024
Unicorn:  1.0.2
```

bet4it:
Resolve #871


disconnect3d:
Thanks.

That's also interesting: why did gdb even got stuck here?

bet4it:
Because `pwndbg.symbol.address('ptmalloc_init')` may throw an exception when libc is not loaded

pwndbg/heap/__init__.py
```diff
@@ -8,18 +8,8 @@
 
 heap_chain_limit = pwndbg.config.Parameter('heap-dereference-limit', 8, 'number of bins to dereference')
 
-@pwndbg.events.new_objfile
```
zachriggle, line 11: I think this needs to be on new_objfile since if you do `starti` the binary will break on the first instruction of `ld.so` and `libc` will not be available / can't find symbols and the Heap stuff will not be initialized.

bet4it: But this is only an initialization of a simple struct, not related with whether the real heap is initialized:
https://github.com/pwndbg/pwndbg/blob/ae6f25a810df3b1902b1b13e6d7924d76e4de572/pwndbg/heap/ptmalloc.py#L57-L65

PR #876, opened 2021-01-02
Remove quotes from command option interpolation

### LINKED ISSUE #875: setup.sh: pip: no such option: --target /usr/local/lib/python3.7/dist-packages
### Description

After #868, it seems that the `pip` invocation in `setup.sh` is broken. I got this error on Python 3.7 (and also Python 3.9)

```
no such option: --target /usr/local/lib/python3.7/dist-packages
```

when the following line is executed: https://github.com/pwndbg/pwndbg/blob/bf49bf8356dbe2a79e9e9fcbdef9ca409374d072/setup.sh#L151

It executed without problems when `${INSTALLFLAGS}` didn't have the quotes.

### Steps to reproduce

Just run `./setup.sh`.

### My setup

gdb: `GNU gdb (Debian 8.2.1-2+b3) 8.2.1`
Python: `3.7.3 (default, Jul 25 2020, 13:03:44) [GCC 8.3.0]`
pwndbg commit: `bf49bf8 (HEAD -> dev, origin/dev, origin/HEAD) Unit test fix (#868)`

vctr-dev:
There was a quoting bug as `INSTALLFLAGS` contains both the option key and value.
This causes the subsequent commands to look something like `python ... '--target foo'...`, causing
the command to treat the entire string `--target foo` as an option key rather than a key-value pair.

This fixes issue #875

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -144,14 +144,14 @@ fi
 
 # Make sure that pip is available
 if ! ${PYTHON} -m pip -V; then
-    ${PYTHON} -m ensurepip "${INSTALLFLAGS}" --upgrade
+    ${PYTHON} -m ensurepip ${INSTALLFLAGS} --upgrade
 fi
 
 # Upgrade pip itself
-${PYTHON} -m pip install "${INSTALLFLAGS}" --upgrade pip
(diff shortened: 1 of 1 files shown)


disconnect3d:
Thanks!

Issue #844, opened 2020-10-20
IDA Integration: Work with IDA 7.0 (specifically 7.5+)


zachriggle:
Currently, the IDA pro integration only works as-is on versions of IDA v7.4 and below.

The most recent v7.5 deprecated the old API by default, and while it can be enabled, the IDAPython script and the `pwndbg.ida` module makes use of the old API.

Since IDA v7.0 was released several years ago, it should be safe to deprecate support entirely for versions 6.x and lower.

wtdcode:
The pwndbg depends on `xmlrpclib` which is removed from Python3. Besides, the current XML server runs in a child thread and IDA API can't be called out of the main thread I believe. Thus, lots of rework should be done...

zachriggle:
Python3 has `xmlrpc.server`, which is effectively the same.

Also, "can't be called" is strong.  Maybe not advised, but I've definitely used it to great success with IDA 7.4.  That said, we do have some custom code to save a timestamped copy of your IDB each time Pwndbg connects, just in case Bad Things (TM) happen.

wtdcode:
Yes, I have tried to do some rework. The real problem is “xmlrpclib.Marshaller”, which don’t have an alternative in Python3.

About my claim about IDA API in multithread environment, I’m sure some API can only be called in the plugin main thread but probably not all of API I guess.

________________________________

WeSeekAfterKnowledge:
I tried to do a bit of work on this, it looks like our mechanism for module registration won't work on the new API.  Trying to see if we can work around this, but even trying to catch the exception (and continue with registration of other functions) it doesn't seem to like it.

```
Python>register_module(idaapi)
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "<string>", line 3, in register_module
  File "/Applications/IDA Pro 7.5/ida64.app/Contents/MacOS/python/3/ida_idd.py", line 2429, in __getattr__
    ea = self.__name_or_ea(name_or_ea)
  File "/Applications/IDA Pro 7.5/ida64.app/Contents/MacOS/python/3/ida_idd.py", line 2377, in __name_or_ea
    raise ValueError("Undefined function " + name_or_ea)
ValueError: Undefined function __call__
```

WeSeekAfterKnowledge:
Spent a bunch of time getting this to work, unfortunately it looks like @wtdcode was correct about threading issues within IDA 7.x.

```
[...]

PR #878, opened 2021-01-05
Use IDA 7 API

WeSeekAfterKnowledge:
First pass at getting the IDA 7.x API to work

Currently it does NOT work, since the functions are called off the main thread, despite using idaapi.execute_sync

Uploading for posterity, see #844 for more information


disconnect3d:
I believe this has been done in https://github.com/pwndbg/pwndbg/pull/1024

PR #880, opened 2021-01-06
Adding p2p command - pointer chain search

SecMeant:
This command can be used to search for any length of pointer chains
in specified mappings. Pointer chain is set of pointers that point
to one another.

### DIFF
--- a/pwndbg/commands/p2p.py
+++ b/pwndbg/commands/p2p.py
@@ -1,0 +1,120 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+import argparse
+
+import gdb
+
+import pwndbg.arch
+import pwndbg.commands
+import pwndbg.config
+import pwndbg.memory
+import pwndbg.regs
+import pwndbg.color
+
+ts = pwndbg.commands.telescope.telescope
+
+class AddrRange:
+    def __init__(self, begin, end):
+        self.begin = begin
+        self.end = end
+
+    def __repr__(self):
+        return (self.begin, self.end).__repr__()
+
+def get_addrrange_any_named():
+    return [AddrRange(page.start, page.end) for page in pwndbg.vmmap.get()]
+
+def guess_numbers_base(num: str):
+    base = 10
+    if num.startswith('0x'):
+        base = 16
+    elif num.startswith('0b'):
+        base = 2
+    elif num.startswith('0'):
+        base = 8
+
+    return base
+
+def address_range_explicit(section):
+    global parser
+
+    try:
+        begin, end = section.split(':')
+
+        begin = int(begin, guess_numbers_base(begin))
+        end = int(end, guess_numbers_base(end))
+
+        return AddrRange(begin, end)
+    except:
+        parser.error('\"%s\" - Bad format of explicit address range!' \
+                ' Expected format: \"BEGIN_ADDRESS:END_ADDRESS\"' % pwndbg.color.red(section))
+
+def address_range(section):
+    global parser
+
+    if section == '*' or section == 'any':
+        return (0, pwndbg.arch.ptrmask)
+
+    # User can use syntax: "begin:end" to specify explicit address range instead of named page.
+    # TODO: handle page names that contains ':'.
+    if ':' in section:
+        return [address_range_explicit(section)]
+
+    pages = list(filter(lambda page: section in page.objfile, pwndbg.vmmap.get()))
+
+    if pages:
+        return [AddrRange(page.start, page.end) for page in pages]
+    else:
+        parser.error('Memory page with name \"%s\" does not exist!' % pwndbg.color.red(section))
+
+parser = argparse.ArgumentParser(description='Pointer to pointer chain search - '
+        'Searches given mapping for all pointers that point to specified mapping (any chain length > 0 is valid).'
+        'If only one mapping is given it just looks for any pointers in that mapping.')
+
+parser.add_argument('mapping_names', type=address_range, nargs='+', help='Mapping name ')
+
+def maybe_points_to_ranges(ptr: int, rs: [AddrRange]):
+    try:
+        pointee = pwndbg.memory.pvoid(ptr)
+    except:
+        return None
+
+    for r in rs:
+        if pointee >= r.begin and pointee < r.end:
+            return pointee
+
+    return None
+
+def p2p_walk(addr, ranges, current_level):
+    levels = len(ranges)
+
+    if current_level >= levels:
+        return None
+
+    maybe_addr = maybe_points_to_ranges(addr, ranges[current_level])
+
+    if maybe_addr == None:
+        return None
+
+    if current_level == levels-1:
+        return addr
+
+    return p2p_walk(maybe_addr, ranges, current_level+1)
+
+@pwndbg.commands.ArgparsedCommand(parser)
+@pwndbg.commands.OnlyWhenRunning
+def p2p(mapping_names: [] = None):
+
+    if not mapping_names:
+        return
+
+    if len(mapping_names) == 1:
+        mapping_names.append(get_addrrange_any_named())
+
+    for rng in mapping_names[0]:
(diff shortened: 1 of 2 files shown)


disconnect3d:
Hey,

That's an interesting command, but it seems we may already have such functionality. Though it doesn't seem so easy to use or/and isn't so robust. We have the `probeleak` and `leakfind` commands. Let me describe both of them with examples comparable to `p2p` so we can see if we want to add `p2p` command or not, or, maybe we just need to fix or improve the other command[s].

---

### Leakfind
```
usage: leakfind [-h] [-p [PAGE_NAME]] [-o [MAX_OFFSET]] [-d [MAX_DEPTH]] [-s [STEP]] [--negative_offset [NEGATIVE_OFFSET]] address

Attempt to find a leak chain given a starting address.
Scans memory near the given address, looks for pointers, and continues that process to attempt to find leaks.

Example: leakfind $rsp --page_name=filename --max_offset=0x48 --max_depth=6. This would look for any chains of leaks that point to a section in filename which begin near $rsp, are never 0x48 bytes further from a known pointer, and are a maximum length of 6.

positional arguments:
  address               Starting address to find a leak chain from

optional arguments:
  -p [PAGE_NAME], --page_name [PAGE_NAME]
                        Substring required to be part of the name of any found pages
  -o [MAX_OFFSET], --max_offset [MAX_OFFSET]
                        Max offset to add to addresses when looking for leak (default: 72)
  -d [MAX_DEPTH], --max_depth [MAX_DEPTH]
                        Maximum depth to follow pointers to (default: 4)
  -s [STEP], --step [STEP]
                        Step to add between pointers so they are considered. For example, if this is 4 it would only consider pointers at an
                        offset divisible by 4 from the starting pointer (default: 1)
  --negative_offset [NEGATIVE_OFFSET]
                        Max negative offset to search before an address when looking for a leak (default: 0)
```

We can get **almost** the same results with `leakfind` as with `p2p <page1> <page2>` by using `leakfind -p <page2> -o <page2size> -d 1 <page2start-address>`. I said **almost** as it seems there may a bug here, where we don't display a given pointer if it appeared before. Anyway, we can see `leakfind` vs `p2p` on this screen (NOTE: the address I pass in to `probeleak` is the `stack` starting address and `0x21000` is that page's size):

image

FWIW, the maybe annoying parts of `leakfind` are a) the inability to pass in memory page as starting address and b) its interface or defaults? It was designed to be able to find "pointers to pointers to pointers" as well, so when we run it with no flags on an address, we see this:
image

---
### probeleak
```
usage: probeleak [-h] [--max-distance MAX_DISTANCE] [--point-to POINT_TO] [--max-ptrs MAX_PTRS] [--flags FLAGS] [address] [count]

Pointer scan for possible offset leaks.
Examples:
    probeleak $rsp 0x64 - leaks 0x64 bytes starting at stack pointer and search for valid pointers
    probeleak $rsp 0x64 --max-dist 0x10 - as above, but pointers may point 0x10 bytes outside of memory page
    probeleak $rsp 0x64 --point-to libc --max-ptrs 1 --flags rwx - leaks 0x64 bytes starting at stack pointer and search for one valid pointer which points to a libc rwx page

positional arguments:
  address               Leak memory address (default: $sp)
  count                 Leak size in bytes (default: 64)

optional arguments:
  --max-distance MAX_DISTANCE
                        Max acceptable distance between memory page boundry and leaked pointer (default: 0)
  --point-to POINT_TO   Mapping name of the page that you want the pointers point to
  --max-ptrs MAX_PTRS   Stop search after find n pointers, default 0 (default: 0)
  --flags FLAGS         flags of the page that you want the pointers point to. [e.g. rwx]
```

With `probeleak` we can get the same as `p2p <page1> <page2>` by using `probeleak --point-to <page2> <page1-start-address> <page1-size>`. An example below, by passing in stack's start address and size:
image

It seems this gives almost the same information as `p2p`, except of maybe displaying the full address and the dereferenced content. Also, the interface is not that best, since we pass in addresses instead of memory pages.

---

## Issues with current implementation of `p2p`

Also, there are a couple of issues here that we may want to fix/improve, if we decide to add the `p2p` command:
1. Pointer coloring and/or telescope. Is there any reason we don't use the standard `telescope` functionality to enhance the display of pointers (and to dereference their values)?

2. The error message, when passing in second memory page that does not exist, could be improved to be more obvious that we did not find a memory page with this name. That's how it looks now: 
image

3. When there are multiple pages for a given page name, the `address_range` function returns the first found page. It seems to me that this may be annoying for users, as we may miss important pointers to other pages with the same name. I think we could just return *all* matching pages and inform the user we are searching for pointers to/from those memory pages and we could maybe add a functionality to pass an index to use a single source/dest page, based on those that were found? 

4. We should probably also support looking for pointers at X-Y address range that points to A-B address range. That would also let users t solve the problem mentioned in 2) but is maybe not that convenient. Though, informing users about the issue would rather be an enough solution.

5. Ultimately, for the 2), 3) and 4), I guess it would be best to have an abstract way to pass in a memory page OR address range into commands. Such solution could support the following cases:
- Choose memory page by its index, that would be displayed and saved globally, when `vmmap` is invoked (easy and robust!)
- Choose memory pages by their name: but this may return multiple pages, and should probably inform about the index solution to filter it further
- Pass in (START-END) range, which would be treated as a [filtered] memory address range

6. That's not an issue, but there is one great advantage over other solutions - simplicity. The command is most likely faster than the other ones and easier to use :P.

---

## To sum up...

What do you think? It seems to me that it would be best to eventually have a nicer interface for `probeleak` and `leakfind` in order to pass in memory pages OR address ranges instead of *only* address+count for their search.

SecMeant:
Thanks for the detailed answer.

I agree that one can find some issues in `p2p`s implementation. Using `probeleak` or `leakfind` seems kinda hacky to me and I think it cannot be used to find "pointers on stack that point to pointers in libc that points to heap" for example. Although I am not very familiar with them and might not understand something here ngl. I have no problem with fixing current p2p implementation if You will eventually merge it. I might have some time next week to do it.

disconnect3d:
> I have no problem with fixing current p2p implementation if You will eventually merge it. I might have some time next week to do it.

Hey, can you improve p2p? I think its fine to add it to Pwndbg and see how it works and eventually rollout the other commands in favor of p2p if it is better.

SecMeant:
Hey. Happy to hear that. I haven't change anything since. I was using it from my forked pwndbg. I think I will find some time today or tomorrow to solve the minor issues you mentioned.

SecMeant:
I am still working on it. I managed to solve point 1 and 2, I am now adding "nice" way to search for pointers from range X-Y to A-B so we can solve all other points with one solution. I should get solution ready at worst tomorrow (Aug 25).

disconnect3d:
@SecMeant FWIW I am not sure if this invalidates any of the above points, but I've recently added names for unnamed mappings in §vmmap§ so they can be "targeted". This being said there is still this issue that some pages have the same name (maybe we should target them all when this happens OR first/last/executable/rw page depending on the context it is being executed and have an alias for "page:x", "page:rw" or "page:0" or something like this... Anyway, its probably not related to this PR at all, jut saying.

SecMeant:
Ok, Its ready.User can now also specify explict address range with: "BEGIN:END"
explicit_range

pwndbg/commands/p2p.py
```diff
@@ -0,0 +1,120 @@
...
+        return None
+
+    if current_level == levels-1:
+        return addr
+
+    return p2p_walk(maybe_addr, ranges, current_level+1)
+
+@pwndbg.commands.ArgparsedCommand(parser)
+@pwndbg.commands.OnlyWhenRunning
+def p2p(mapping_names: [] = None):
```
disconnect3d, line 106: FWIW https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

disconnect3d: type hints nvm :)

PR #889, opened 2021-02-26
Fixed bug when the GDB is debuggin an architecture arm-eabi (disassembly-flavor).

airvzxf:
Based on the issue commented by @enedil in this post: https://github.com/pwndbg/pwndbg/pull/860#issuecomment-782896531 which is crashing the pwndbg when the `disassembly-flavor` is set up on GDB targeting the architecture `arm-eabi`. The solution is catch the exception and set the default flavor to 'intel'.


disconnect3d:
Thanks!

PR #892, opened 2021-03-04
Build and publish AppImages

### LINKED ISSUE #891: Build and publish AppImages
This tool works fine. However, for first-time users, quite some set-up is typically required. First, you typically have to build a few dependencies, most notably some up-to-date gdb. Of course it's optional, but surely a good idea. Then, one can pip-install the dependencies from a `requirements.txt`. However, that's normally not a great idea; advanced Python users will always prefer to use a virtual environment of any kind. Then, one should be able to integrate the script in gdb's init config.

The provided setup script isn't great either. It makes many assumptions about how this tool should be installed: it just installs the dependencies, into the user's Python 3 environment, which is typically used by many other tools. pwndbg itself is not even installed there, but shall remain in the Git clone directory.
Of course, one could change the script to maintain a virtualenv (it's not hard at all) somewhere in the home directory, and then set up the environment to make gdb use the dependencies from there. However, there's a better solution that doesn't require any of these workarounds.

I propose to build and ship an [AppImage](https://appimage.org/) for pwndbg. Building an AppImage would have, among others, these advantages:

- Applications packaged as an AppImage can run on many distributions (including Ubuntu, Fedora, openSUSE, CentOS, elementaryOS, Linux Mint, and others)
- One app = one file = super simple for users: just download one AppImage file, [make it executable](https://docs.appimage.org/user-guide/run-appimages.html#download-make-executable-run), and run
- No unpacking or installation necessary
- No root needed
- No system libraries changed
- Works out of the box, no installation of runtimes needed
- Optional [desktop integration](https://docs.appimage.org/user-guide/run-appimages.html#integrating-appimages-into-the-desktop)
- Optional binary delta updates, e.g., for continuous builds (only download the binary diff) using [AppImageUpdate](https://docs.appimage.org/introduction/software-overview.html#appimageupdate)
- Can optionally sign your AppImages (inside the file) using gpg2
- Works on Live ISOs
- Can use the same AppImages when dual-booting multiple distributions
- Can be listed in the [AppImageHub](https://appimage.github.io/apps) central directory of available AppImages
- Can double as a self-extracting compressed archive with the [`--appimage-extract`](https://docs.appimage.org/user-guide/run-appimages.html#mount-or-extract-appimages) parameter

[Here is an overview](https://appimage.github.io/apps) of projects that are already distributing upstream-provided, official AppImages.

AppImages could be built both for releases as well as for the main branch (so-called continuous release scheme). Automating AppImage building is relatively simple and straightforward. You can bundle all tools you need (e.g., a suitable up-to-date GDB version). This way, this project will be a lot easier to set up and use: you will just have to call `./pwndbg-x86_64.AppImage`.

The only tricky part here is bundling Python along with gdb. But I could imagine that if you let linuxdeploy-plugin-{python,conda} set up a Python environment, then build gdb against it, it should work. I've done that before.

I can provide some guidance on how to build AppImages and integrate this into GitHub actions. You can also reach out to the AppImage team, see https://docs.appimage.org/contact.html for more information.

TheAssassin:
Fixes #891.

This PR sets up a GitHub action an AppImage for x86_64. It contains a fresh copy of GDB (I hardcoded the latest version, will have to update that in the future) along with the latest pwndbg.

The action will automatically upload the AppImage to GitHub releases using https://github.com/TheAssassin/pyuploadtool/. Depending on whether it's a tag or regular build, a release will be created either for the tag, or a prerelease called "continuous" for the regular one.

### DIFF
--- a/.github/workflows/build-appimage.yml
+++ b/.github/workflows/build-appimage.yml
@@ -1,0 +1,45 @@
+name: Build AppImage
+
+on: [push, pull_request, workflow_dispatch]
+
+env:
+  APPIMAGE_EXTRACT_AND_RUN: 1
+  TERM: xterm-256color
+
+jobs:
+  appimage:
+    strategy:
+      fail-fast: false
+      matrix:
+        ARCH: [x86_64]
+    name: Build AppImage for ${{ matrix.ARCH }}
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build AppImage
+        run: |
+          export ARCH="${{ matrix.ARCH }}"
+          bash -xe ci/build-appimage.sh
+      - name: Archive artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: AppImage ${{ matrix.ARCH }}
+          path: pwndbg*.AppImage*
+
+  upload:
+    name: Create release and upload artifacts
+    runs-on: ubuntu-18.04
+    needs:
+      - appimage
+    steps:
+      - name: Download artifacts
+        uses: actions/download-artifact@v2
+      - name: Inspect directory after downloading artifacts
+        run: ls -alFR
+      - name: Create release and upload artifacts
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+            wget -q https://github.com/TheAssassin/pyuploadtool/releases/download/continuous/pyuploadtool-x86_64.AppImage
+            chmod +x pyuploadtool-x86_64.AppImage
+            ./pyuploadtool-x86_64.AppImage **/pwndbg*.AppImage*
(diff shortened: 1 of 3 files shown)


TheAssassin:
The reason the job fails is that my fork doesn't contain any tags, therefore `git describe` doesn't work. The build would work, though. You can find a build for testing in https://github.com/TheAssassin/pwndbg/releases/tag/continuous.

Adding an i386 build is by the way also possible. However, I doubt anyone is going to really need one. Let me know if you're interested.

Consider moving all your jobs into one workflow, though. We can't make the AppImage build wait for the other workflows. My solution would be to have just one workflow that runs the checks as stages before the AppImage build. I'm using this in other projects with success.

disconnect3d:
I decided to not publish an AppImage myself. Feel free to publish an unofficial version.

PR #894, opened 2021-03-05
`auxv` command produces wrong AT_RANDOM value

fikriguska:
When using an `auxv` command, it should show an address, not `0x10`.
image

This is because the regex reads a wrong number in `info auxv` command.  (It reads `16` (0x10) in `"Address of 16 random bytes"` string)
image

So I changed the regex. And it works.
image


disconnect3d:
Thanks!

PR #899, opened 2021-03-31
improve and modularize profiling script

anthraxx:


anthraxx:
do we want a more complex schema and implement kind of an interface?
That would allow us to potentially define functions for cache warming and profiling separately.

anthraxx:
ok i implemented some dead simple warmup function which is called

PR #900, opened 2021-04-01
feature(telescope): option to set min repeating values before skipping

### LINKED ISSUE #803: telescope should not fold single values, should probably be configurable
In this screenshot, we should simply display the 0x28 and 0x90 and 0x108 offset in the stack trace.  The ellipsis isn't particularly useful here, and actually adds to cognitive load due to having to check how many values are skipped.

This should probably be behind a config, e.g. `telescope-fold-minimum` which is expressed in pointer widths.  This will allow people to control the setting to their choosing (or disable folding it with a value of 0 or -1 or something).

anthraxx:
Buffer all repeating lines and check the minimum value when to start
marking them with skip lines. In case the minimum value is not hit, just
unroll the buffer.

To stop skipping any lines, there is the existing bool config
telescope-skip-repeat-val so we avoid adding special values to minimum
like -1 and keep the setting separated.

Fixes #803

### DIFF
--- a/pwndbg/commands/telescope.py
+++ b/pwndbg/commands/telescope.py
@@ -98,12 +98,13 @@ def telescope(address=None, count=telescope_lines, to_string=False):
     result = []
     last = None
     collapse_buffer = []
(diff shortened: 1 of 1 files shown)


disconnect3d:
FWIW:
image

And
image

PR #912, opened 2021-04-25
enhance kernel debugging support

### LINKED ISSUE #911: pwndbg fails to work with 64bit kernel in qemu
### Description

pwndbg raises when debugging a 64bit stripped kernel.
the error message looks like this:
~~~
pwndbg: loaded 196 commands. Type pwndbg [filter] for a list.
Reading symbols from ./vmlinux...(no debugging symbols found)...done.
pwndbg> target remote :1234
Remote debugging using :1234
Exception occurred: Error: Cannot access memory at address 0xf81802ff0 (<class 'gdb.MemoryError'>)
Exception occurred: Error: Cannot access memory at address 0xf81802ff0 (<class 'gdb.MemoryError'>)
Python Exception <class 'gdb.MemoryError'> Cannot access memory at address 0xf81802ff0: 
^CPython Exception <class 'KeyboardInterrupt'> : 
^CPython Exception <class 'KeyboardInterrupt'> Quit: 
^CPython Exception <class 'KeyboardInterrupt'> Quit: 
0xffffffff8111c398 in ?? ()
^CPython Exception <class 'KeyboardInterrupt'> Quit: 
pwndbg> vmmap
^CPython Exception <class 'KeyboardInterrupt'> Quit: 
Error occurred in Python command: Quit
^CPython Exception <class 'KeyboardInterrupt'> Quit:
~~~

### Steps to reproduce
0. download the kernel/initfs from https://github.com/mephi42/ctf/blob/master/2021.04.09-Midnight_Sun_CTF_2021_Quals/brohammer/brohammer.tar.gz
1. launch the kernel with command 
~~~
qemu-system-x86_64 \
    -m 1024M \
    -kernel ./vmlinuz \
    -initrd ./rootfs.img \
    -nographic \
    -monitor /dev/null \
    -append "nokaslr root=/dev/ram rw console=ttyS0 oops=panic paneic=1 quiet" \
    -s
~~~
2. extract the vmlinuz to vmlinux
3. run gdb with pwndbg and execute `target remote :1234`

### My setup
I'm using commit 87da998fcefe8cba85f51dba0056b923483c6c1c, which is the latest commit rn.
gdb version:
~~~
pwndbg> show version
GNU gdb (Ubuntu 8.1.1-0ubuntu1) 8.1.1
Copyright (C) 2018 Free Software Foundation, Inc.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
<http://www.gnu.org/software/gdb/bugs/>.
For help, type "help".
Type "apropos word" to search for commands related to "word".
pwndbg> py import sys; print(sys.version)
3.6.9 (default, Jan 26 2021, 15:33:00) 
[GCC 8.4.0]

~~~

Kyle-Kyle:
partially fix #911

### DIFF
--- a/pwndbg/elf.py
+++ b/pwndbg/elf.py
@@ -236,7 +236,7 @@ def get_ehdr(pointer):
     """
 
     # This just does not work :(
-    if pwndbg.qemu.is_usermode():
+    if pwndbg.qemu.is_qemu():
         return None, None
 
     vmmap = pwndbg.vmmap.find(pointer)


pwndbg/typeinfo.py
```diff
@@ -45,8 +45,8 @@ def lookup_types(*types):
 def update():
 
     module.char   = gdb.lookup_type('char')
-    module.ulong  = lookup_types('unsigned long', 'uint', 'u32', 'uint32')
-    module.long   = lookup_types('long', 'int', 'i32', 'int32')
```
disconnect3d: why do we remove those? those help us work with Rust or Go targets iirc

pwndbg/arch.py
```diff
@@ -29,6 +30,21 @@ def fix_arch(arch):
...
         return 'armcm' if '-m' in arch else 'arm'
 
+    # handle some corner cases
+    if arch == 'i386:x86-64':
+        return 'x86-64'
+
+    return arch
+
+def get_init_arch():
+    out = gdb.execute("show architecture", to_string=True).strip()
```
disconnect3d: I assume with this we will get a better debugging experience if we do gdb ./vmlinux? or is there another story?

pwndbg/arch.py
```diff
@@ -29,6 +30,21 @@ def fix_arch(arch):
     if 'arm' in arch:
         return 'armcm' if '-m' in arch else 'arm'
 
+    # handle some corner cases
```
disconnect3d: what's the corner case? when does it occur?

pwndbg/typeinfo.py
Kyle-Kyle: I don't think ulong type is always uint32. And that's the reason why sometimes some kernel pointers get casted to 32bit and raise errors

pwndbg/arch.py
Kyle-Kyle: The original implementation was wrong. It hardcoded an i386 architecture before attach remote. This caused some troubles during kernel debugging. So what I did here is to pull the architecture info from gdb during initialization before attach remote. Once attach remote is issued, we trust its handlers

pwndbg/arch.py
Kyle-Kyle: This is because the definitions of "architecture" are different between gdb and pwndbg. We handle some "translation" here. for example. i386:x86-64 in gdb should be treated as x86-64 in pwndbg

pwndbg/typeinfo.py
disconnect3d: Seems that we should determine the lookup type based on language (`show language`) :/

pwndbg/arch.py
disconnect3d: Can we mention this in a comment?

pwndbg/typeinfo.py
disconnect3d: Fwiw the best way to handle it would be to add support to gdb so we could create `gdb.Type` instances out of thin air...

WeSeekAfterKnowledge:
@Kyle-Kyle what's the status on this?  Anything we can do to help out?

I think @disconnect3d had some good feedback, but I'm not sure if this PR is talled or still in progress.

Kyle-Kyle:
I was distracted by other stuff and totally forgot about this PR. Sorry about that.
I'll work on this in the next week.

Kyle-Kyle:
not sure what happened. but now, I don't see errors shown previously. and pwndbg works with Linux kernel well with this little fix

disconnect3d:
@Kyle-Kyle can you check against HEAD-1 or HEAD-N versions? 
Also, what does `vmmap` show for you for a Linux kernel?

I hope those two commits did not break `qemu.is_kernel()`:
* https://github.com/pwndbg/pwndbg/commit/d753c0455f79c158a1dbda6bc2127f365896b380
* https://github.com/pwndbg/pwndbg/commit/0068257ebe69f84fd568cb50e5d858a9fb19f2f4

But I haven't had a chance to check yet.

Kyle-Kyle:
without the fix, pwndbg errors out like this in HEAD~1:
~~~
pwndbg> target remote :1234
Remote debugging using :1234
0xffffffff8111c398 in ?? ()
  File "/home/kylebot/src/pwndbg/pwndbg/events.py", line 165, in caller
    func()
  File "/home/kylebot/src/pwndbg/pwndbg/memoize.py", line 194, in __reset_on_base
    base = pwndbg.elf.exe().address if pwndbg.elf.exe() else None
  File "/home/kylebot/src/pwndbg/pwndbg/proc.py", line 71, in wrapper
    return func(*a, **kw)
  File "/home/kylebot/src/pwndbg/pwndbg/memoize.py", line 46, in __call__
  File "/home/kylebot/src/pwndbg/pwndbg/elf.py", line 182, in exe
    return load(e)
  File "/home/kylebot/src/pwndbg/pwndbg/elf.py", line 220, in load
    return get_ehdr(pointer)[1]
  File "/home/kylebot/src/pwndbg/pwndbg/elf.py", line 246, in get_ehdr
    if pwndbg.memory.read(vmmap.start, 4) == b'\x7fELF':
AttributeError: 'NoneType' object has no attribute 'start'

PS: Pull requests are welcome
> /home/kylebot/src/pwndbg/pwndbg/elf.py(246)get_ehdr()
-> if pwndbg.memory.read(vmmap.start, 4) == b'\x7fELF':
(Pdb) 
~~~

Kyle-Kyle:
that's why I changed `qemu.is_usermode` to `qemu.is_qemu`

disconnect3d:
@Kyle-Kyle Btw does `vmmap` work then or does it show only a single page 0-0xfff...?

disconnect3d:
I'm going to hit merge here as I believe this should solve kernel issue. However, there are probably some other improvements we can do.

If we don't get `monitor_info_mem` pages, we are most likely having an incorrect `vmmap` output - I think this may be on non-x86 kernels. Shall we change `pwndbg.vmmap.get` to have `is_qemu` instead of `is_usermode` check here?:
https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/vmmap.py#L58-L61

Kyle-Kyle:
I think changing `is_usermode` to `is_qemu` is a good idea. because this is more of a qemu-related problem, not just qemu-user mode.

PR #914, opened 2021-04-27
Fix emulate into infinite loop

veritas501:


veritas501:
According to the tests result, it should not be fixed in this way.

PR #919, opened 2021-05-11
Fix mips32r6 disasm & emu

veritas501:
When disasm mips32r6 using capstone or emulate mips32r6 using unicorn,
MODE should be XX_MODE_MIPS32R6


disconnect3d:
Thanks!

PR #923, opened 2021-05-18
fix:step.sh bug

Jokky6:
when I  use `./step.sh` init pwndbg on MacOS. Its can not read python version correctly.


disconnect3d:
Why is this wrong? Also, does pwndbg even has a use case on MacOS?

PR #926, opened 2021-06-04
fix: update_length() raise exception in some cases

veritas501:
Some binary moves its debug information into .dwo files. https://gcc.gnu.org/wiki/DebugFission
But if we delete or moved those .dwo files, gdb may show warning when debugging.
Like:
```
warning: (Internal error: pc 0x7ffff49ef495 in read in CU, but not in symtab.)
```
In those case, when we enter a command, those warning also shows.
```
pwndbg> show print elements
Limit on string chars or array elements to print is 200.
warning: (Internal error: pc 0x7ffff49ef495 in read in CU, but not in symtab.)
warning: (Internal error: pc 0x7ffff49ef495 in read in CU, but not in symtab.)
warning: (Internal error: pc 0x7ffff49ef495 in read in CU, but not in symtab.)
warning: (Internal error: pc 0x7ffff49ef495 in read in CU, but not in symtab.)
Exception occurred: Error: invalid literal for int() with base 10: 'symtab.)' (<class 'ValueError'>)
Python Exception <class 'ValueError'> invalid literal for int() with base 10: 'symtab.)':
```

If we call `message = message.split()[-1]`, we get `symtab.)`.
Then `length = int(message)` raise an Exception.

### DIFF
--- a/pwndbg/commands/pie.py
+++ b/pwndbg/commands/pie.py
@@ -94,7 +94,7 @@ parser.add_argument('module', type=str, nargs='?', default='',
 
 @pwndbg.commands.ArgparsedCommand(parser)
 @pwndbg.commands.OnlyWhenRunning
-def breakrva(offset=None, module=None):
+def breakrva(offset=0, module=None):
     offset = int(offset)
     if not module:
         module = get_exe_name()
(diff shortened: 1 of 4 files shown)


disconnect3d:
Ugh, thanks a lot for this! We really need a way to tell GDB to disable its warning messages.

disconnect3d:
Or better: get the parameter values from Python API.

PR #929, opened 2021-06-23
registers split into specified number of columns

lonnywong:
## test.c
```c
int main(void) {
  __asm__("int $3");
  return 0;
}
```

## config
```
source ~/pwndbg/gdbinit.py
set show-compact-regs on
set show-compact-regs-align 92
set show-compact-regs-columns 2
```

## issue
1. need to reset show-compact-regs-align while screen resize
2. sometimes it's not aligned
Screen Shot 2021-06-23 at 11 36 09 PM

## future
1. always fix columns to show registers as you wish
2. always aligned perfectly
Screen Shot 2021-06-23 at 11 42 01 PM

### DIFF
--- a/pwndbg/color/__init__.py
+++ b/pwndbg/color/__init__.py
@@ -74,7 +74,7 @@ def generateColorFunction(config):
     return function
 
 def strip(x):
-    return re.sub('\x1b\\[\d+m', '', x)
+    return re.sub('\x1b\\[[\d;]+m', '', x)
 
 def terminateWith(x, color):
     return re.sub('\x1b\\[0m', NORMAL + color, x)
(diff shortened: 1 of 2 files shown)


lonnywong:
## share my ~/.gdbinit
```
source ~/pwndbg/gdbinit.py

set show-compact-regs on

define set_registers_columns
python
import gdb, pwndbg
if pwndbg.arch.current in ('i386', 'x86-64'):
  gdb.execute('set show-compact-regs-columns 2')
else:
  gdb.execute('set show-compact-regs-columns 4')
end
end

define target hookpost-remote
set_registers_columns
end

set_registers_columns
```

disconnect3d:
Thanks!

PR #931, opened 2021-07-01
Fix broken leakfind command

adamdoupe:
Because of the previous commit to this file that removed `from queue import *`, the `address_queue` on line 96 would fail by throwing an exception when running `leakfind`.

This commit adds back the required `import queue` and fixes the reference to `Queue` on line 96.


disconnect3d:
Thanks!

PR #934, opened 2021-07-30
Add filter for writable/executable pages to vmmap

matrizzo:
This adds 2 new options to `vmmap`: `-w/--writable` that only shows writable mappings, and `-x/--executable` that only shows executable mappings. When both are specified, `vmmap` only shows mappings that are both writable and executable.

Example: 
vmmap_wx


disconnect3d:
This is awesome, thanks! :)

PR #935, opened 2021-07-30
Make setup.sh work with Podman

matrizzo:
Containers created with Podman (https://podman.io/) don't have a .dockerenv file in the root directory, so setup.sh tries to invoke sudo, even though it doesn't exist in the container. This PR fixes that by removing the check for .dockerenv. The other two checks (current user is root and sudo doesn't exist) are enough to detect when the script is running in a container vs. a "real" machine with sudo.


disconnect3d:
Thanks!

Issue #940, opened 2021-08-09
recommend to use virtualenv in the readme 


gogo2464:
**Is your feature request related to a problem? Please describe.**
When I installed pwndbg I did not worry on what was in the file setup.sh. Then I did not use virtualenv and I did not know that pip has been used. Pip like that may break the system and is not recommended.

**Describe the solution you'd like**
A recommendation in the readme to use virtualenv or at least a warning about pip.

**Describe alternatives you've considered**

**Additional context**

disconnect3d:
@gogo2464 hey, thanks for the issue; I agree its better to use a virtualenv but then it requires the user to always run in virtualenv when they start GDB (to use Pwndbg). This is fine, but we have to inform on this properly.

Can you maybe send a PR with those things described properly in the README?

gogo2464:
Nice assign me to this issue. Or why not even modify the bash script to use virtualenv. I will see.

gogo2464:
@disconnect3d I have made 2 prs. choose the one you prefer.

disconnect3d:
We decided to not use virtualenv or/and the user can choose by themself

viciu:
@gogo2464 pls check https://github.com/pwndbg/pwndbg/issues/1035

PR #941, opened 2021-08-10
add virtualenv argument to setup.sh and document this chnage in the r…

gogo2464:
…eadme

fix this issue: https://github.com/pwndbg/pwndbg/issues/940

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,13 @@ Installation is straightforward.  Pwndbg is best supported on Ubuntu 18.04 with 
 ```shell
 git clone https://github.com/pwndbg/pwndbg
 cd pwndbg
-./setup.sh
+./setup.sh virtualenv
+```
+
+Then to use pwndbg do:
+
+```
+source pwndbg/venv/bin/activate
 ```
 
 Other Linux distributions are also supported via `setup.sh`, including:
--- a/README.md
+++ b/README.md
@@ -56,6 +56,18 @@ Other Linux distributions are also supported via `setup.sh`, including:
 
 If you use any Linux distribution other than Ubuntu, we recommend using the [latest available GDB](https://www.gnu.org/software/gdb/download/) built from source.  Be sure to pass `--with-python=/path/to/python` to `./configure`.
 
+## More install options (optional):
+
+You can do:
+
+```shell
+git clone https://github.com/pwndbg/pwndbg
+cd pwndbg
+./setup.sh
+```
+
+to directly to install pwndbg out of a virtual environment. This will most likely break your system. Use it at your own risk. For more information, fell free to consult the link: https://docs.python-guide.org/dev/virtualenvs/#basic-usage
+
 ## What can I do with that?
 
 For further info about features/functionalities, see [FEATURES](FEATURES.md).
--- a/setup.sh
+++ b/setup.sh
@@ -20,7 +20,7 @@ osx() {
 
 install_apt() {
     sudo apt-get update || true
-    sudo apt-get install -y git gdb python3-dev python3-pip python3-setuptools libglib2.0-dev libc6-dbg
+    sudo apt-get install -y git gdb python3-dev python3-pip python3-setuptools libglib2.0-dev libc6-dbg virtualenv
 
     if uname -m | grep x86_64 > /dev/null; then
         sudo dpkg --add-architecture i386 || true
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -35,7 +35,13 @@ Installation is straightforward.  Pwndbg is best supported on Ubuntu 18.04 with
 ```shell
 git clone https://github.com/pwndbg/pwndbg
 cd pwndbg
-./setup.sh
+./setup.sh virtualenv
```
disconnect3d, line 38: Maybe:
```
# if you want to use virtualenv pass in the 'virtualenv' first argument:
./setup.sh virtualenv
# otherwise, just call
./setup.sh
```

README.md
```diff
@@ -35,7 +35,13 @@ Installation is straightforward.  Pwndbg is best supported on Ubuntu 18.04 with
...
+```
+
+Then to use pwndbg do:
+
+```
+source pwndbg/fake_env/bin/activate
```
disconnect3d: Btw wouldn't it be better to call it `venv` instead of `fake_env`?

Also, I am not sure if its ok that we depend on `virtualenv` being already installed; we should probably say it

README.md
disconnect3d: Or even adding "(requires virtualenv to be installed)"

gogo2464: Yes but indeed it will make the install instructions hard to read. I do not really like that. I really enjoyed the simplicity to install pwndbg when I installed it for the first time. Things can go hard very quickly in computers. Particulary for biary exploitation. Can I resolve this conversation. Or maybe you could prefer another chapter with more install options maybe.

EDIT: This is what I did in my last PR. I separate into 2 commits in case you do not like the latest commit.

README.md
gogo2464: done

README.md
gogo2464: https://github.com/pwndbg/pwndbg/pull/941#discussion_r686627137 => virtualenv is now installed from repo on any OS.

gogo2464:
Not tested on any os listed. Still should work.

Ready to merge just after your review.

EDIT: Only tested on debian.

README.md
WeSeekAfterKnowledge: The default should NOT be to use a virtualenv.   This should be a supplementary option, perhaps in a subsection.

README.md
```diff
@@ -50,6 +56,18 @@ Other Linux distributions are also supported via `setup.sh`, including:
 
 If you use any Linux distribution other than Ubuntu, we recommend using the [latest available GDB](https://www.gnu.org/software/gdb/download/) built from source.  Be sure to pass `--with-python=/path/to/python` to `./configure`.
 
+## More install options (optional):
```
WeSeekAfterKnowledge, line 59: This should remain the PRIMARY installation mechanism.

README.md
```diff
@@ -62,4 +80,4 @@ Want to help with development? Read [CONTRIBUTING](.github/CONTRIBUTING.md) or [
 
 ## Contact
 If you have any questions not worthy of a [bug report](https://github.com/pwndbg/pwndbg/issues), feel free to ping
-anybody on [Discord](https://discord.gg/x47DssnGwm) and ask away.
\ No newline at end of file
+anybody on [Discord](https://discord.gg/x47DssnGwm) and ask away.
```
WeSeekAfterKnowledge, line 83: Nit: Line endings

setup.sh
```diff
@@ -147,6 +150,12 @@ if ! ${PYTHON} -m pip -V; then
     ${PYTHON} -m ensurepip ${INSTALLFLAGS} --upgrade
 fi
 
+if [ "$1" == "virtualenv" ]
```
WeSeekAfterKnowledge, line 153: I'm pretty sure that your `venv` install mechanism does not ACTUALLY use virtualenv.  `${INSTALLFLAGS}` uses the same hard-coded path as GDB uses.

README.md
gogo2464: I do not really think so. In my humble opinion, pip out of virtualenv could break all the system and requires the user to reinstall all this linux distro. I experienced this issue I was really bad. I took years to understand why it occures and this could be really embarassing for people who do challenges with pwndbg. If you loose the system you will more likely delete all the exploit because the system will be reinstalled and will change the offset of libc and all.

Anyway you are a contributor to this project since a longer time than me, you know the project more than me then you can only know what is true. If you want we could debate.

gogo2464:
I accidentally asked for review. Sorry for the spam.

README.md
disconnect3d: In theory yes, but many people are not into Python or virtualenv and do not know how to use it well. I don't think forcing them or suggesting to use virtualenv by default is good here.

This being said, its better to make virtualenv an opt-in instead of opt-out.

WeSeekAfterKnowledge:
I do not believe the VirtualEnv approach you provided actually works.

Would you mind also including a Dockerfile that demonstrates that it does work?

Please create the directory `extras/docker/virtualenv` and create your Dockerfile there, and show that the new install method works correctly.

gogo2464:
I will do all the expected changes. I just need some time. I am very busy at this moment.

disconnect3d:
Closing this for now as I am not sure about virtualenv support here.

gogo2464:
@disconnect3d I have made a long long research. This actually worked from the start of the PR. We just need to launch with ./setup.sh instead of sh ./setup.sh. I have tested with another library. I have tested. After do you really want to make like that? This is the question.

I think you could reopen the issue.

Issue #943, opened 2021-08-11
uplaod pwndbg to the repo pypi to allow people to do pip install pwndbg


gogo2464:
**Is your feature request related to a problem? Please describe.**
People could be frustruated when installed the file with setup.sh. This is basically a bash file made to install with pip. Trough the issue with virtualenv is now solved, we should also allow more expert dev to easily install from pipy repo with pip. The idea is that an expert in python could know immediately how to install.

**Describe the solution you'd like**
pip install pwndbg should do the stuff.

**Describe alternatives you've considered**
Also keep the bash file but do not keep it as only way to install. The readme should show the option to install with bash file as an easy way for people who do not master pip.

**Additional context**

gogo2464:
Assign me to this issue please. I have already started.

gogo2464:
I did it! I made some change in the code. I will submit the PR!

Just see https://pypi.org/project/pwndbg/0.1/

gogo2464:
I have deleted the pwndbg from pipy because you are not the owner. Please could you make an account @zachriggle please?

disconnect3d:
Hey, I agree that being able to install pwndbg via pypi is an awesome feature. Let's do that :).

However, I don't think we want to rename pwndbg -> pwndbglib as suggested in the PR. The package isn't going to be used outside of being a GDB plugin and even if, I don't think we want to change the namespace.

gogo2464:
I see what you mean. I actually need to do that because a file know as pwndbg must be set at the root of the project.

disconnect3d:
>  a file know as pwndbg must be set at the root of the project.

Wait, why?

gogo2464:
I mean a folder know as pwndbg must be set at the root of the project. Not a file. This folder will contain some files to generate the file to upload to pypi. See https://medium.com/@joel.barmettler/how-to-upload-your-python-package-to-pypi-65edc5fe9c56

WeSeekAfterKnowledge:
It is not possible for Pwndbg to be installable via `pip`, because we have additional dependencies on system packages that have to be installed via the package manager (apt, yum, etc).

Because of this, Pwndbg will continue to require running `setup.py`.  Feel free to work with package maintainers to get Pwndbg added as a real package.  Arch linux has this already: 
https://archlinux.org/packages/community/any/pwndbg/

PR #944, opened 2021-08-12
upload pwndbg to pipy

gogo2464:
I uploaded to pipy https://pypi.org/project/pwndbg/0.1/ 

This is experimental. I have to make some tests.

For the moment do not merge. I have to pass the tests.

See https://github.com/pwndbg/pwndbg/issues/943


gogo2464:
I can not run the test I need to be able to import gdb. I tried to read instructions from this stack overflow forum https://stackoverflow.com/questions/4792483/how-to-import-gdb-in-python/10240053 and did
$ gdb -q -x tests/test_heap.py 
pwndbg: loaded 187 commands. Type pwndbg [filter] for a list.
  File "tests/test_heap.py", line 8, in <module>
    import pwndbglib
ModuleNotFoundError: No module named 'pwndbglib'
pwndbg>


but the other libraries from my code are not imported anymore.

Could you tell me how you run the tests please?

WeSeekAfterKnowledge:
It is not possible for Pwndbg to be installable via pip, because we have additional dependencies on system packages that have to be installed via the package manager (apt, yum, etc).

Because of this, Pwndbg will continue to require running setup.py. Feel free to work with package maintainers to get Pwndbg added as a real package. Arch linux has this already:
https://archlinux.org/packages/community/any/pwndbg/

Separately, since you are a new contributor, please start with very small changes.  Huge changes like this will always be rejected.

gogo2464:
Understood. Thank.

Issue #955, opened 2021-09-03
remote debugging download failing


forkpetard:
### Description

When remote debugging,`checksec` and other commands that rely on reading the target binary seem to fail at downloading it. 
Issue #623 seems related, but the accepted fix doesn't seem to work for me. 
### Steps to reproduce

On the remote run `gdbserver localhost:1234 $local binary$` $local binary$ being a path to a binary that is not present on the local machine.
Locally, run `gdb` and set the target remote with `target remote $ip$:1234` 

        
      pwndbg> checksec
      Traceback (most recent call last):
        File "/usr/share/pwndbg/pwndbg/commands/__init__.py", line 130, in __call__
          return self.function(*args, **kwargs)
        File "/usr/share/pwndbg/pwndbg/commands/__init__.py", line 210, in _OnlyWithFile
          return function(*a, **kw)
        File "/usr/share/pwndbg/pwndbg/commands/checksec.py", line 12, in checksec
          print(pwndbg.wrappers.checksec.get_raw_out())
        File "/usr/share/pwndbg/pwndbg/commands/__init__.py", line 210, in _OnlyWithFile
          return function(*a, **kw)
        File "/usr/share/pwndbg/pwndbg/wrappers/__init__.py", line 28, in _OnlyWithCommand
          return function(*a, **kw)
        File "/usr/share/pwndbg/pwndbg/memoize.py", line 46, in __call__
          value = self.func(*args, **kwargs)
        File "/usr/share/pwndbg/pwndbg/wrappers/checksec.py", line 16, in get_raw_out
          local_path = pwndbg.file.get_file(pwndbg.proc.exe)
        File "/usr/share/pwndbg/pwndbg/file.py", line 42, in get_file
          raise OSError("Could not download remote file %r:\n" \
      OSError: Could not download remote file 'target:$remote path$':
      Error: Remote I/O error: No such file or directory
      
      If that is an issue, you can report it on https://github.com/pwndbg/pwndbg/issues
      (Please don't forget to search if it hasn't been reported before)
      To generate the report and open a browser, you may run `bugreport --run-browser`
      PS: Pull requests are welcome
      > /usr/share/pwndbg/pwndbg/file.py(42)get_file()
      -> raise OSError("Could not download remote file %r:\n" \

Fetching the remote file manually and setting it with `file` doesn't change this.
One explanation could be that while getting the remote file gdb doesn't expect the path starting with "target:".
[...]

PR #956, opened 2021-09-07
Remote fetching improvement suggestion

forkpetard:
So, remote/ssh debugging was pretty tedious with "autofetch" always downloading remote object files (even though gdb already does that or sysroot making those file available locally). I know that the changes to autofetch look severe, but i couldn't really make sense of its use/usefulness, at least not by doing a quick grep or checking the commit history. 
The change in file.py just removes "target:" from a remote path and checks if the file is available locally before downloading it (#955).

I hope this looks alright :D

### DIFF
--- a/pwndbg/file.py
+++ b/pwndbg/file.py
@@ -9,6 +9,7 @@ debugging a remote process over SSH or similar, where e.g.
 import binascii
 import os
 import tempfile
+import re
 
 import gdb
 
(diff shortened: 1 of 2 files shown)


disconnect3d:
Hey, see https://github.com/pwndbg/pwndbg/commit/3e4ad608af25f4cca0333e3c3421b9747ec63681

This should fix this problem and so I am closing this PR. See #955 for more explanation too.

PR #960, opened 2021-09-19
Migrate to 3.10 compatible abc usage.

novafacing:
This is a one-line fix. Python 3.10 has [deprecated](https://github.com/python/cpython/commit/c47c78b878ff617164b2b94ff711a6103e781753#diff-c92f871cb9b4aa1583c89f82cd416d58a5d70fff41faf41e45dedf775dbe7713) using direct `abc` imports from `collections`.

### DIFF
--- a/pwndbg/memoize.py
+++ b/pwndbg/memoize.py
@@ -32,7 +32,7 @@ class memoize:
     def __call__(self, *args, **kwargs):
         how = None
 
-        if not isinstance(args, collections.Hashable):
+        if not isinstance(args, collections.abc.Hashable):
             print("Cannot memoize %r!", file=sys.stderr)
             how   = "Not memoizeable!"
             value = self.func(*args)
--- a/pwndbg/memoize.py
+++ b/pwndbg/memoize.py
@@ -6,13 +6,19 @@ e.g. execution stops because of a SIGINT or breakpoint, or a
 new library/objfile are loaded, etc.
 """
 
-import collections
 import functools
 import sys
 
 import gdb
 
 import pwndbg.events
+
+try:
+    # Python >= 3.10
+    from collections.abc import Hashable
+except ImportError:
+    # Python < 3.10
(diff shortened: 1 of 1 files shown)


novafacing:
The file failing linting isn't touched by this, but I'm happy to submit another PR that fixes that.

disconnect3d:
Hey,

If we accept this PR, we will break Pwndbg for anyone using GDB with Python <3.10.

Honestly, I am disappointed that Python made the effort to deprecate it but introduced an alternative only in 3.10 when the thing got removed? I may be wrong, but it seems so since the proposed change does not work in 3.9:

```
dc@dc:~$ sudo docker run --rm -it python:3.9 python -c 'import collections; print(collections.abc.Hashable)'
  File "/usr/local/lib/python3.9/collections/__init__.py", line 68, in __getattr__
    raise AttributeError(f'module {__name__!r} has no attribute {name!r}')
AttributeError: module 'collections' has no attribute 'abc'
dc@dc:~$ sudo docker run --rm -it python:3.9 python -c 'import collections; print(collections.Hashable)'
<string>:1: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.10 it will stop working
<class 'collections.abc.Hashable'>
```

Anyway, I can't accept this change in its current form. Can you make something like:
```python
try:
    # Python >= 3.10
    from collections.abc import Hashable
except ImportError:
    # Python < 3.10
    from collections import Hashable
```

and just use `Hashable` later on?

novafacing:
Interesting, I didn't realize this would break backward compatibility. I'll rework this!

novafacing:
😓 linter ran....sorry about that, I'm not a regular FOSS contributor yet and I'm working on my setup for it!

Should be good now, thanks for the correction!

disconnect3d:
Don't worry about linter, I'll fix it in next PR.

PR #971, opened 2021-10-05
Add safe-linking support

cnwangjihe:
Since GLibc 2.32, tcache and fastbin's fd use Safe-Linking, I do some modifications to support it.
```c
// code form glibc 2.32, malloc/malloc.c, line #339
#define PROTECT_PTR(pos, ptr) \
  ((__typeof (ptr)) ((((size_t) pos) >> 12) ^ ((size_t) ptr)))
#define REVEAL_PTR(ptr)  PROTECT_PTR (&ptr, ptr)
```
I use symbol `<banner>` in glibc to determine glibc version, it's declared in `csu/version.c`.  
I have already check its availability from glibc 2.23 to 2.34 (on Ubuntu only).  
```c
static const char banner[] =
"GNU C Library "PKGVERSION RELEASE" release version "VERSION".\n\
Copyright (C) 2020 Free Software Foundation, Inc.\n\
This is free software; see the source for copying conditions.\n\
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A\n\
PARTICULAR PURPOSE.\n\
Compiled by GNU CC version "__VERSION__".\n"
```
The regex I use to extract version info: `release version (\d+)\.(\d+)`.


disconnect3d:
@cnwangjihe hey, could you maybe add a testcase for this case? e.g. a binary that can be re-produced with Docker and then a test that loads it -> breaks where necessary and checks if the output of a given command is appropriate.

This way we would prevent from breaking this (and other) features with future changes :D

CptGibbon:
Nice 👌 It would be great to let the user enable/disable this feature in settings (e.g. `set safe-linking on`), that way GLIBC release branches are accounted for (e.g. safe linking gets backported into release/2.30 but your new feature isn't activated by default because the version number is lower than 2.32).

CptGibbon:
It could also be worthwhile adding a version detection fallback to the `__libc_version` symbol when GLIBC debugging symbols are available 🤷‍♂️ 

Incidentally, the `banner` symbol is also exported by other elements of libc (e.g. libpthread), I'm not sure if that could interfere with resolving `banner` in this code.

cnwangjihe:
> It could also be worthwhile adding a version detection fallback to the `__libc_version` symbol when GLIBC debugging symbols are available 🤷‍♂️
> 
> Incidentally, the `banner` symbol is also exported by other elements of libc (e.g. libpthread), I'm not sure if that could interfere with resolving `banner` in this code.

Thanks a lot, using `__libc_version` will be a better solution, since `banner` require debug symbols too :).



> @cnwangjihe hey, could you maybe add a testcase for this case? e.g. a binary that can be re-produced with Docker and then a test that loads it -> breaks where necessary and checks if the output of a given command is appropriate.
> 
> This way we would prevent from breaking this (and other) features with future changes :D

@disconnect3d To run a test, I need a specific version of glibc. It is okay to include that binary into this repository? Or I should write a script to pull prebuilt glibc from other location?

I noticie that `test_heap` also requires a specific verison of glibc, and in the Makefile of `heap_bugs`, rpath is set to `/glibc_versions/2.29/tcache_x64`, but I fail to find the code that download it. 
It is a bug? Or just my misunderstanding?
https://github.com/pwndbg/pwndbg/blob/e239e9c1f69649ac028255206df0e2151416b763/tests/binaries/makefile#L31
https://github.com/pwndbg/pwndbg/blob/e239e9c1f69649ac028255206df0e2151416b763/tests/binaries/makefile#L60-L62

disconnect3d:
@cnwangjihe 
> @disconnect3d To run a test, I need a specific version of glibc. It is okay to include that binary into this repository? Or I should write a script to pull prebuilt glibc from other location?
>
> I noticie that test_heap also requires a specific verison of glibc, and in the Makefile of heap_bugs, rpath is set to /glibc_versions/2.29/tcache_x64, but I fail to find the code that download it. It is a bug? Or just my misunderstanding?

Good question! I think it would be best to take the glibc from a Docker container and mark the container tag along with its ID in a README file around, so we can reproduce/get the same glibc once again --- and also verify or have more trust that it is not a bogus/malicious glibc.

Regharding the glibc 2.29/tcache_x64 I am not sure. Perhaps @GrosQuildu worked on this (cc: @GrosQuildu do you know sth about it?). If it is missing, we need to fix this. I haven't really worked much with the heap feature in Pwndbg and I'd appreciate help on it.

PS: You may want to join our Discord server (see project README) and discuss things there :).

CptGibbon:
Take your time, writing tests for heap introspection commands is challenging at the best of times.

CptGibbon:
Now that's a test 🤩

CptGibbon:
The only remaining improvement I could suggest would be to also decrypt the terminating null values for each list, otherwise we see the output shown below. Encrypted null fd's shown in red.

bins_example

cnwangjihe:
> The only remaining improvement I could suggest would be to also decrypt the terminating null values for each list, otherwise we see the output shown below. Encrypted null fd's shown in red.
> 
> bins_example

Fixed

CptGibbon:
A great contribution, been looking forward to this feature for a while, thank you.
@disconnect3d I'm happy with this if you are 👍
Does what it says on the tin and as far as I can tell doesn't adversely affect any other heap commands.

pwndbg/glibc.py
```diff
@@ -0,0 +1,51 @@
...
+        if get_version() is not None:
+            return function(*a, **kw)
+        else:
+            print('%s: GLibc not loaded yet.' % function.__name__)
+    return _OnlyWhenGlibcLoaded
+
+
+@OnlyWhenGlibcLoaded
+def check_safe_linking():
+    return (get_version() >= (2, 32) or safe_lnk == "on") and safe_lnk != "off"
```
disconnect3d, line 56:
```suggestion
    """
    Safe-linking is a glibc 2.32 mitigation; see:
    - https://lanph3re.blogspot.com/2020/08/blog-post.html
    - https://research.checkpoint.com/2020/safe-linking-eliminating-a-20-year-old-malloc-exploit-primitive/
    """
    return (get_version() >= (2, 32) or safe_lnk == "on") and safe_lnk != "off"
```

disconnect3d:
Thanks!

PR #982, opened 2021-11-04
Searching in kernel mode

lonnywong:
## search in kernel mode takes a long time
image

## search only part of the mappings is much faster
image


disconnect3d:
Hey, is the idea to provide vmmap pages addresses so search only looks in them?

disconnect3d:
I think this can probably be done better, where u can provide some range too

disconnect3d:
Also does it work better with https://github.com/pwndbg/pwndbg/tree/hack-pt-dump ? (u need to pull submodule too)

lonnywong:
> I think this can probably be done better, where u can provide some range too

I want to search like this first. But, argparse seems not support.
`search value [ mapping_name | begin_address [ end_address | max_length ] ]`

PR #983, opened 2021-11-13
Fix error message on interrupts

AnyKeyShik:
On interrupts pwndbg produce this error:

interrupt error

This PR fix it

### DIFF
--- a/pwndbg/emu/emulator.py
+++ b/pwndbg/emu/emulator.py
@@ -64,7 +64,7 @@ arch_to_CS = {
 DEBUG = False
 
 
-def debug(fmt, args=""):
+def debug(fmt, args=()):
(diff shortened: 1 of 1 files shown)


disconnect3d:
Fixed:
```
In [4]: def foo(a, args=()):
   ...:     print(a % args)
   ...:

In [5]: foo("asd")
asd

In [6]: def foo(a, args=""):
   ...:     print(a % args)
   ...:

In [7]: foo('abc')
<ipython-input-7-20ece011db47> in <module>
----> 1 foo('abc')

<ipython-input-6-4940424d3246> in foo(a, args)
      1 def foo(a, args=""):
----> 2     print(a % args)
      3

TypeError: not all arguments converted during string formatting
```

Thanks!

PR #986, opened 2021-12-04
Fixed multiple issues of compact register view

TobiasFaller:
- **Fixed** alignment of columns where the separation was not correctly considered in calculation
- **Fixed** non-updating window size when resizing console
- **Updated** configuration options to more common names
- **Updated** order of configuration options to match priority
- **Updated** column count to 2 by default (when compact view is enabled)

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -284,10 +284,10 @@ def context(subcontext=None):
             out.flush()
 
 
-pwndbg.config.Parameter('show-compact-regs', False, 'whether to show a compact register view')
-pwndbg.config.Parameter('show-compact-regs-align', 20, 'the number of characters reserved for each register and value')
-pwndbg.config.Parameter('show-compact-regs-space', 4, 'the minimum number of characters separating each register')
-pwndbg.config.Parameter('show-compact-regs-columns', 0, 'the number of columns to show registers (if 0, using the show-compact-regs-align)')
+pwndbg.config.Parameter('show-compact-regs', False, 'whether to show a compact register view with columns')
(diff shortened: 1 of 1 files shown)


TobiasFaller:
@lonnywong Could you please check if this fixes your issue with #830, because I could not exactly reproduce your issue?

disconnect3d:
@TobiasFaller does this change the default view? If so, can u post screenshot before & after those changes?

TobiasFaller:
@disconnect3d This option is disabled by default. So this does not change the default settings. Here are two screenshots.
This option was supposed to show the view from below, but due to bugs in my first implementation the view did not always end up correct. I hope this MR fixes these problems (see #830).

With `show-compact-regs off` (default):

DefaultView

With `show-compact-regs on` and old default sub-settings (0 columns, minimum 20 characters column width and 4 characters spacing):

DefaultColumns

With `show-compact-regs on` and updated default sub-settings (two columns, minimum 20 characters column width and 4 characters spacing):

CompactView

lonnywong:
> @lonnywong Could you please check if this fixes your issue with #830, because I could not exactly reproduce your issue?

I fixed it with https://github.com/pwndbg/pwndbg/pull/929

TobiasFaller:
@lonnywong That's great. And I was wondering why I could not reproduce your issue. 😆 

There was a slight issue with the alignment where the padding between columns was not correctly respected.
Additionally, I noticed that when resizing the console window the register view did not update the size correctly.
Those issues should now be fixed.

TobiasFaller:
@disconnect3d Screenshots have been updated now.

disconnect3d:
@TobiasFaller that's perfect. Thanks!

lonnywong:
> Additionally, I noticed that when resizing the console window the register view did not update the size correctly.
> Those issues should now be fixed.

cool

PR #987, opened 2021-12-04
Added development container configuration for VS Code (Remote Container)

TobiasFaller:
Provides .devcontainer configuration to directly develop pwndbg inside Docker container.
This configuration file is used by the Remote Container extension in VS Code to directly start the IDE inside the Docker container and install the dependencies.

Usage:
- Open VS Code and have the Remote Container extension installed.
- Run the command `Remote-Containers: Reopen folder in Container` and wait.
- Open a console with Ctrl + Shift + Backtick and start gdb.
- Done.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,18 @@
+{
+	"name": "pwndbg",
+	"build": {
+		"context": "..",
+		"dockerfile": "../Dockerfile"
+	},
+	"runArgs": [
+		"--cap-add=SYS_PTRACE",
+		"--security-opt",
+		"seccomp=unconfined"
+	],
+	"workspaceFolder": "/pwndbg",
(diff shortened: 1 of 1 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,18 @@
+{
+	"name": "pwndbg",
+	"build": {
+		"context": "..",
+		"dockerfile": "../Dockerfile"
+	},
+	"runArgs": [
+		"--cap-add=SYS_PTRACE",
+		"--security-opt",
+		"seccomp=unconfined"
```
disconnect3d: Why do we need to disable seccomp? Also, if we are here can we add `--security-opt no-new-privs:true` (or sth like that) and `--cap-drop=ALL` before adding PTRACE?

TobiasFaller: @disconnect3d Fixed, thanks. These options were the defaults in the template file.

disconnect3d:
Thanks!

PR #991, opened 2021-12-08
Fixed functionality in setup.sh

ToBeatElite:
sorry for all the PR's.
in my last PR we added a line to source ``gdbinit.py`` but it was pointed out that it is already done in the ``setup.sh``, i looked at the setup file and realised that the source command is not added since the script prematurely exits with ``exit 1`` when its a non supported distro. By moving this block to the top, before any exit's are called, we remove this issue and the source command will **always** be added to the users ``.gdbinit``

- ToBeatElite

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ Installation is straightforward.  Pwndbg is best supported on Ubuntu 18.04 with 
 git clone https://github.com/pwndbg/pwndbg
 cd pwndbg
 ./setup.sh
+echo "source $(pwd)/gdbinit.py" >> ~/.gdbinit
 ```
 
 Other Linux distributions are also supported via `setup.sh`, including:
--- a/README.md
+++ b/README.md
@@ -36,7 +36,6 @@ Installation is straightforward.  Pwndbg is best supported on Ubuntu 18.04 with 
 git clone https://github.com/pwndbg/pwndbg
 cd pwndbg
 ./setup.sh
-echo "source $(pwd)/gdbinit.py" >> ~/.gdbinit
 ```
 
 Other Linux distributions are also supported via `setup.sh`, including:
--- a/setup.sh
+++ b/setup.sh
@@ -91,6 +91,12 @@ if linux; then
             echo " - https://aur.archlinux.org/packages/pwndbg-git/"
             exit 1
             ;;
+        "endeavouros")
+            echo "Install pwndbg using a community package. See:"
+            echo " - https://www.archlinux.org/packages/community/any/pwndbg/"
+            echo " - https://aur.archlinux.org/packages/pwndbg-git/"
+            exit 1
+            ;;
         "manjaro")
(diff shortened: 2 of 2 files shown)


disconnect3d:
Shall we really do this? What if setup fails? Then we leave the user with incorrect ~/.gdbinit :/

ToBeatElite:
> Shall we really do this? What if setup fails? Then we leave the user with incorrect ~/.gdbinit

If we're following "if it ain't broke don't fix it" by all means don't do it bc it's perfectly fine as it was. If u think this can introduce bugs decline it ig.

disconnect3d:
So TL;DR: we just reverted the previous commit. Yeah, let's do that.

Hmm... but there still is an issue with unsupported distros or distros with special handling (like Arch). Can we do better on that term? Maybe echo some additional messages for them before exitting?

ToBeatElite:
> if setup fails

by this I hope I assume correctly that you mean if one of the *install functions* failed? If so could we could put the "source gdbinit" block in a function and call that function when the *install functions* are completed.

Another thing that I think could maybe be better:

All 3 distros that the setup.sh script exits on, are arch-based. so would it not be possible to simply install the package from AUR with a ``git clone`` and ``makepkg -si`` instead of just telling the user to do it? This way we eliminate the case of the ``gdbinit.py`` not being sourced bc we wont have to ``exit 1`` there, **and** make the installation procedure as easy on arch-based as it is on the other distros.

Of course we need to install the AUR package with git here because we cant be sure what AUR helper the user would be using, and this git clone method should work for everyone.

https://wiki.archlinux.org/title/Arch_User_Repository

I tested on my machine and I believe that implementing an Install function for Arch-Based distros could be something as simple as:

```bash
install_from_aur() {
    git clone https://aur.archlinux.org/pwndbg-git.git
    cd pwndbg-git
    makepkg -si
}
```

the *only* downside I can see is that if the user does not use a AUR helper they will prolly forget to update it when that time comes. But idk anyone who dosen't use one so I think that's low risk.

Edit: Or yh what u said we could simply echo some more messages for them

anthraxx:
No, installing from AUR is not a good idea. First there is the regular maintained repository package `pwndbg` that can and should be favored being installed via pacman. Second it basically makes a wrong expectation from the setup process. If at all then the setup process should reference whatever has been cloned here and not build another package that may not be obvious how to update (a git pull on this clone won't do it).

ToBeatElite:
> No, installing from AUR is not a good idea. First there is the regular maintained repository package `pwndbg` that can and should be favored being installed via pacman. Second it basically makes a wrong expectation from the setup process. If at all then the setup process should reference whatever has been cloned here and not build another package that may not be obvious how to update (a git pull on this clone won't do it). 

Yeah, ur likely right, that was just an Idea. What do u propose instead? 

The easiest method that disconnected mentioned was to add some more echo statements.

Issue #990, opened 2021-12-08
Linting and Auto Format


oneNutW0nder:
Opening this up for discussion on linting and format styles. When browsing through the code `pylint` is not shy about showing all of the linting issues:

<img width="1285" alt="image" src="https://user-images.githubusercontent.com/29211904/145149129-62075d71-d7ba-4bb1-83fb-1be08d538a02.png">

This is just an example from `pwndbg/chain.py`. To be clear, I am not suggesting that we resolve all of the linter issues but I think there should be something defined moving forward. We can create a standard linting configuration with `pylint` through the use of a `.pylintrc` file (https://pylint.pycqa.org/en/latest/user_guide/run.html).

I would also like to suggest the use of a code formatting tool like `black` to automatically keep the code formatted in a standardized way. 

I look forward to the discussion 😄

oneNutW0nder:
Relevant information about `git blame` issues that may come up: https://stackoverflow.com/questions/53502654/how-do-i-run-a-code-formatter-over-my-source-without-modifying-git-history

There are solutions to this issue below the most up voted comment.

gsingh93:
It would definitely be helpful to have some autoformatting configuration. While I'm trying to add features to the code, my editor keeps autoformatting everything (which is a good thing), but then I end up with a bunch of changes unrelated to the feature I'm adding, which makes for a messy PR.

I really don't care which tool or configuration we use, as long as there's something that always gives us the same style code everywhere without any extra thought.

My current configuration uses `yapf` with this config:
```
[style]
based_on_style = facebook
split_before_bitwise_operator = true
split_before_logical_operator = true
```

I can go ahead and make a PR adding a `.style.yapf` file with this configuration, and on that PR if anyone has objections to this configuration or thinks using another tool like `black` is a better idea, we can switch to that (although I'm not as familiar with `black`'s configuration options).

I'm not familiar with [Github Action workflows](https://docs.github.com/en/actions/using-workflows#creating-a-workflow-file), but we can then add that to help keep all the code consistent.

gsingh93:
Ah, it looks like there's already a `setup.cfg` that provides `pep8` formatting. So I can skip adding the `yapf` config, and maybe the only action left for this issue is to get a Github Action for it working.

disconnect3d:
We are currently running `isort`, `black` and `pyflake8`. I think we can close this for now.

[...]

PR #993, opened 2021-12-14
added pylintrc file for standarized linting

oneNutW0nder:
I started going through all of the linting errors for the project and added what I thought would be a good starting point for what to ignore. Without a `.pylintrc` file, pylint will generate ~11,000 lines of output for the entire project. This `.pylintrc` file knocks that total down to just under 1,000 lines. 

A lot of the `no-member` errors come from missing imports that Python magically resolves during runtime. However, these break intellisense functionality like jumping to definition which is quite annoying. 

This coincides with issue #990 

The following list are categories that I wanted more input on before adding them to the ignore list:

```
(anomalous-backslash-in-string)
(arguments-renamed)
(assignment-from-none)
(attribute-defined-outside-init)
(bad-super-call)
(bare-except)
(c-extension-no-member)
(chained-comparison)
(consider-using-dict-items)
(consider-using-generator)
(consider-using-in)
(consider-using-max-builtin)
(consider-using-with)
(dangerous-default-value)
(deprecated-class)
(deprecated-module)
(eval-used)
(fixme)
(forgotten-debug-statement)
(function-redefined)
(global-variable-not-assigned)
(global-variable-undefined)
(import-error)
(import-outside-toplevel)
(import-self)
(inconsistent-return-statements)
(keyword-arg-before-vararg)
(no-else-return)
(no-member)
(no-name-in-module)
(no-self-argument)
(no-self-use)
(no-value-for-parameter)
(pointless-statement)
(protected-access)
(raise-missing-from)
(redefined-argument-from-local)
(redefined-builtin)
(redefined-outer-name)
(simplifiable-if-expression)
(simplify-boolean-expression)
(singleton-comparison)
(try-except-raise)
(undefined-all-variable)
(undefined-loop-variable)
(undefined-variable)
(unexpected-keyword-arg)
(ungrouped-imports)
(unidiomatic-typecheck)
(unnecessary-pass)
(unspecified-encoding)
(unsubscriptable-object)
(unsupported-membership-test)
(unused-argument)
(unused-import)
(unused-private-member)
(unused-variable)
(unused-wildcard-import)
(use-implicit-booleaness-not-len)
(use-maxsplit-arg)
(useless-import-alias)
(useless-super-delegation)
(using-constant-test)
```

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,609 @@
+[MASTER]
+
+# A comma-separated list of package or module names from where C extensions may
+# be loaded. Extensions are loading into the active Python interpreter and may
+# run arbitrary code.
+extension-pkg-allow-list=
+
+# A comma-separated list of package or module names from where C extensions may
+# be loaded. Extensions are loading into the active Python interpreter and may
+# run arbitrary code. (This is an alternative name to extension-pkg-allow-list
+# for backward compatibility.)
+extension-pkg-whitelist=
+
+# Return non-zero exit code if any of these messages/categories are detected,
+# even if score is above --fail-under value. Syntax same as enable. Messages
+# specified are enabled, while categories only check already-enabled messages.
+fail-on=
+
+# Specify a score threshold to be exceeded before program exits with error.
+fail-under=10.0
+
+# Files or directories to be skipped. They should be base names, not paths.
+ignore=CVS
+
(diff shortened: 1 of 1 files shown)


disconnect3d:
> The following list are categories that I wanted more input on before adding them to the ignore list:

Let's see how it goes with the current ruleset and then add more if needed

PR #998, opened 2022-01-06
Add fortified function signatures

theteleforce:
Address #519 by adding _chk function signatures to `functions.py`. The signatures were extracted from the non-test files in <https://github.com/lattera/glibc/tree/master/debug>.

The pull request intentionally does not include `__gets_chk` to remain consistent with the deprecation of `gets` (whose signature is absent from `functions.py`). If we want the signature anyway, I would be happy to add it.


disconnect3d:
Thx!

Issue #87, opened 2016-06-22
pwndbg is slow when connect to debug server of pintool


cebrusfs:
Pintool can use `-appdebug` to create a debugger server (like `gdbserver`) then gdb can use `target remote` to connect.
I try to use pwndbg but it will block on some event (I just add a timer in debug log of event to record the time usage):
```
Slow: 'start' pwndbg.argv.update () finished in 2.202494s
Slow: 'obj' pwndbg.events._start_newobjfile (<gdb.NewObjFileEvent object at 0x7fdab28f02a0>,) finished in 2.334072s
Slow: 'stop' pwndbg.stack.update (<gdb.SignalEvent object at 0x7fdab28f02d0>,) finished in 57.476584s
Slow: 'stop' pwndbg.vmmap.explore_registers (<gdb.SignalEvent object at 0x7fdab28f02d0>,) finished in 1.282561s
Slow: 'start' pwndbg.argv.update () finished in 2.198303s
Slow: 'start' pwndbg.argv.update () finished in 2.208627s
Slow: 'obj' pwndbg.events._start_newobjfile () finished in 2.212299s
Slow: 'stop' pwndbg.vmmap.explore_registers () finished in 1.245921s
Slow: 'stop' pwndbg.symbol.add_main_exe_to_symbols () finished in 56.673910s
```
Also, command `telescope` is slow because of lots memory query.

I traced some function and found out that pintool's server didn't provide some information (f.e. auxv) and pwndbg will try to scan memory to retrieve this information, but the `gdb.Value().dereference` is so slow. I haven't looked all slow function, so there may be the other bottlenecks.

I don't have a good idea to fix this. Mabe we need to
1. Add more cache.
2. Reduce unnecessary memory scan. (f.e. reduce unnecessary update functions on each event)

I use `pin-3.0-76991-gcc-linux` on `ubuntu 16.04` vm.
```
$ ./pin-3.0-76991-gcc-linux/pin -version     
Pin 3.0
Copyright (c) 2003-2016, Intel Corporation. All rights reserved.
@CHARM-VERSION: $Rev: 76887
```

BTW, Why `memoize.__reset_on_start` will be invoked by `stop` event? The `stop` event of gdb will be trigger when the inferior has stopped. That is, it will be trigger when gdb stops the inferior and change the control back to you. So, `memoize.reset_on_start` cache will be reset when you use some commands like `si`, `continue`. Is this behavior what you want?

zachriggle:
[...]

PR #1004, opened 2022-02-12
Make pwndbg faster

bet4it:
> BTW, Why `memoize.__reset_on_start` will be invoked by `stop` event? The stop event of `gdb` will be trigger when the inferior has stopped. That is, it will be trigger when gdb stops the inferior and change the control back to you. So, `memoize.reset_on_start` cache will be reset when you use some commands like `si`, `continue`. Is this behavior what you want?

_Originally posted by @cebrusfs in https://github.com/pwndbg/pwndbg/issues/87_

### DIFF
--- a/pwndbg/memoize.py
+++ b/pwndbg/memoize.py
@@ -144,7 +144,6 @@ class reset_on_start(memoize):
     kind   = 'start'
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


disconnect3d:
Thx!

PR #1005, opened 2022-02-14
Show all registers of PowerPC

### LINKED ISSUE #667: Support to display all the registers
Currently, we only display part of registers in some architectures such as PowerPC.
This is the registers will be displayed:
https://github.com/pwndbg/pwndbg/blob/88723a8c0a88369fa2dd267d1fb5c63464db55cb/pwndbg/commands/context.py#L145-L148
Some registers such as `regs.misc` aren't included here:
https://github.com/pwndbg/pwndbg/blob/88723a8c0a88369fa2dd267d1fb5c63464db55cb/pwndbg/regs.py#L182-L187
And I even can't find `r1` in RegisterSet.

Hope we can add an option to display all the registers. (`r0`~`r31` and some other registers)

bet4it:
Close #667.

### DIFF
--- a/pwndbg/regs.py
+++ b/pwndbg/regs.py
@@ -170,13 +170,12 @@ i386 = RegisterSet( pc      = 'eip',
 # r13     Small data area pointer register (points to TLS)
 # r14-r30 Registers used for local variables
 # r31     Used for local variables or "environment pointers"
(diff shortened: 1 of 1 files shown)


disconnect3d:
Can you show screen pre and post this PR? Thanks anyway

bet4it:
Before:

After:

Show more registers:

PR #1007, opened 2022-03-02
Try with lowercase xpsr register first on ARM Cortex M/gdb 8+

wavexx:
Running 'context' fails when debugging an armcm arch as it fails to find the xpsr register.

[#264 mentions this](https://github.com/pwndbg/pwndbg/issues/264#issuecomment-739421512) on gdb 8, but I can confirm the same on gdb 10 as well.

Looking at the source, xpsr is special-handled, which is probably a compatibility hack. The register name is lowercase.
With this change, attempt to use xPSR as a fallback only if "xpsr" fails.


disconnect3d:
Thanks!

PR #1012, opened 2022-03-31
Fix two bugs in /pwndbg/commands/context.py

zarkivy:
1. Without nargs='?' parameter:
The default values banner="both" and width=None will not take effect.

2. The parameter *width* in function *contextoutput* in ui.py has wrong type, which will cause operand types conflict if we pass a integer to width manually:
```
  File "/home/zkv/Softwares/pwndbg/pwndbg/commands/__init__.py", line 131, in __call__
  File "/home/zkv/Softwares/pwndbg/pwndbg/commands/__init__.py", line 225, in _OnlyWhenRunning
    return function(*a, **kw)
  File "/home/zkv/Softwares/pwndbg/pwndbg/commands/context.py", line 266, in context
    result[target].extend(func(target=out,
  File "/home/zkv/Softwares/pwndbg/pwndbg/commands/context.py", line 554, in context_stack
    result = [pwndbg.ui.banner("stack", target=target, width=width)] if with_banner else []
  File "/home/zkv/Softwares/pwndbg/pwndbg/ui.py", line 47, in banner
    banner = rjust_colored(title, (width + len(strip(title))) // 2, config.banner_separator)
TypeError: unsupported operand type(s) for //: 'gdb.Value' and 'int'
```

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -137,6 +137,8 @@ parser.add_argument("banner", type=str, nargs='?', default="both", help="Where a
 parser.add_argument("width", type=int, nargs='?', default=None, help="Sets a fixed width (used for banner). Set to None for auto")
 @pwndbg.commands.ArgparsedCommand(parser, aliases=['ctx-out'])
 def contextoutput(section, path, clearing, banner="both", width=None):
+    if(width != None):
+        width = int(width.cast(gdb.lookup_type('long')))
     outputs[section] = path
     output_settings[section] = dict(clearing=clearing,
                                     width=width,
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -137,7 +137,7 @@ parser.add_argument("banner", type=str, nargs='?', default="both", help="Where a
 parser.add_argument("width", type=int, nargs='?', default=None, help="Sets a fixed width (used for banner). Set to None for auto")
 @pwndbg.commands.ArgparsedCommand(parser, aliases=['ctx-out'])
 def contextoutput(section, path, clearing, banner="both", width=None):
-    if(width != None):
+    if width is not None:
         width = int(width.cast(gdb.lookup_type('long')))
     outputs[section] = path
     output_settings[section] = dict(clearing=clearing,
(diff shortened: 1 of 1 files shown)


pwndbg/commands/context.py
```diff
@@ -133,10 +133,12 @@ def output(section):
...
 parser.add_argument("path", type=str, help="The path to which the output is written")
 parser.add_argument("clearing", type=bool, help="Indicates weather to clear the output")
-parser.add_argument("banner", type=str, default="both", help="Where a banner should be placed: both, top , bottom, none")
-parser.add_argument("width", type=int, default=None, help="Sets a fixed width (used for banner). Set to None for auto")
+parser.add_argument("banner", type=str, nargs='?', default="both", help="Where a banner should be placed: both, top , bottom, none")
+parser.add_argument("width", type=int, nargs='?', default=None, help="Sets a fixed width (used for banner). Set to None for auto")
 @pwndbg.commands.ArgparsedCommand(parser, aliases=['ctx-out'])
 def contextoutput(section, path, clearing, banner="both", width=None):
+    if(width != None):
+        width = int(width.cast(gdb.lookup_type('long')))
```
disconnect3d: Why not just `int(width)`?

pwndbg/commands/context.py
```diff
@@ -133,10 +133,12 @@ def output(section):
 parser.add_argument("section", type=str, help="The section which is to be configured. ('regs', 'disasm', 'code', 'stack', 'backtrace', and/or 'args')")
 parser.add_argument("path", type=str, help="The path to which the output is written")
 parser.add_argument("clearing", type=bool, help="Indicates weather to clear the output")
-parser.add_argument("banner", type=str, default="both", help="Where a banner should be placed: both, top , bottom, none")
-parser.add_argument("width", type=int, default=None, help="Sets a fixed width (used for banner). Set to None for auto")
+parser.add_argument("banner", type=str, nargs='?', default="both", help="Where a banner should be placed: both, top , bottom, none")
+parser.add_argument("width", type=int, nargs='?', default=None, help="Sets a fixed width (used for banner). Set to None for auto")
 @pwndbg.commands.ArgparsedCommand(parser, aliases=['ctx-out'])
 def contextoutput(section, path, clearing, banner="both", width=None):
+    if(width != None):
```
```suggestion
    if width is not None:
```

disconnect3d:
Btw the funciton should also check if 'banner' has proper value and raise argparse exception if that's not the case; can you add it?

pwndbg/commands/context.py
zarkivy: You are right 😂

zarkivy:
Ok, already committed.

PR #1022, opened 2022-05-24
merge hack-pt-dump branch to dev

lonnywong:
Let's finish the hack-pt-dump branch.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "gdb-pt-dump"]
+	path = gdb-pt-dump
+	url = https://github.com/martinradev/gdb-pt-dump.git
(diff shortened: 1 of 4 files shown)


disconnect3d:
@lonnywong hmm... we should bump to the lastest gdb-pt-dump version and we should make it so that this installs itself properly with `setup.py` or in terms of how its described in docs.

disconnect3d:
@lonnywong any chance u can update that?

lonnywong:
We may need to merge first, then update to the lastest.
Or we need to update the branch first by another PR.
The setup.sh has `git submodule update --init --recursive`, it should works without the lastest gdb-pt-dump.

lonnywong:
@disconnect3d The `gdb-pt-dump` is the latest commit `f25898adc61d60e5f30c6452b15700bbf1bd630c` in the `hack-pt-dump` branch.

PR #1024, opened 2022-06-04
Replace with IDA 7.x API

akiym:
Based on #844 and PR #878, I fixed IDA integration to work with IDA 7.x. I tested with IDA 7.7.220218.

### DIFF
--- a/ida_script.py
+++ b/ida_script.py
@@ -152,7 +152,7 @@ def versions():
     }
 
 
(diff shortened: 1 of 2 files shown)


disconnect3d:
Ugh, sorry for not merging it earlier. Thanks a lot for this.

I'm up for upgrading it to 8.x if that needs to be done as well.

PR #1027, opened 2022-07-03
Create `setflag` command

dgmcdona:
Creates a `setflag` command for setting individual bits in the `$eflags` register.

### DIFF
--- a/pwndbg/__init__.py
+++ b/pwndbg/__init__.py
@@ -23,6 +23,7 @@ import pwndbg.commands.cpsr
 import pwndbg.commands.dt
 import pwndbg.commands.dumpargs
 import pwndbg.commands.elf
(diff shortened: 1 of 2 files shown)


disconnect3d:
Hey, thanks for this feature! I am going to merge it since I think it's good enough and I will extend it soon during an [EuroPython 2022 sprint session](https://ep2022.europython.eu/sprints) to detect the architecture and print a warning that it only works on x86/x64.

disconnect3d:
@dgmcdona and here we goes: https://github.com/pwndbg/pwndbg/commit/d49b29d150bc4896be7d16c6fcf01e3936181852 

This now supports different architectures and shows examples in `help setflag` or `setflag --help` :)

Issue #937, opened 2021-08-04
Heap inspection without libc debug symbols


umutoztunc:
The commands for heap inspection currently requires libc debug symbols. However, when the binary is modified with `patchelf` to use a specific interpreter like this file:

```console
$ file target
target: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter ./ld-2.27.so, for GNU/Linux 3.2.0, BuildID[sha1]=b722ce3ea0937340d3a56165c760ac5ef9ae0014, not stripped

$ ldd target
	linux-vdso.so.1 (0x00007fff3b7b4000)
	./libc-2.27.so (0x00007f774a9c4000)
	./ld-2.27.so => /lib64/ld-linux-x86-64.so.2 (0x00007f774afbd000
```

Then, pwndbg is unable to use heap commands even though `libc6-dbg` and `libc6-dbg:i386` are installed:

```
pwndbg> heap
heap: This command only works with libc debug symbols.
They can probably be installed via the package manager of your choice.
See also: https://sourceware.org/gdb/onlinedocs/gdb/Separate-Debug-Files.html

E.g. on Ubuntu/Debian you might need to do the following steps (for 64-bit and 32-bit binaries):
sudo apt-get install libc6-dbg
sudo dpkg --add-architecture i386
sudo apt-get install libc-dbg:i386
```

This is also mentioned in #924.

The solution I found is to use `patchelf` to change interpreter. However, when I try to use `gef`, I realized that it is able to run heap commands on this file without changing the interpreter:

![image](https://user-images.githubusercontent.com/32958854/128200829-f16dc11a-17dd-4732-b601-3ed3aa8e780e.png)

GEF can simply work in all conditions without requiring debug symbols. I have investigated how they do this a little bit. Here is their [heap_base](https://github.com/hugsy/gef/blob/251761696b043088a5ebd31c35ad93205c06ba7a/gef.py#L9939) function:

![image](https://user-images.githubusercontent.com/32958854/128202211-7c01dd7a-c491-4469-95c7-38cdd7d90e2b.png)

It looks like they are parsing the heap base address from the memory. I suggest implementing the same strategy.

This is also useful for systems like arch linux where there is no debug symbols package available and you have to build the package with debugging options yourself.

disconnect3d:
Hey, yeah, it would be awesome to address/fix this. Would you maybe like to try implementing this and sending a PR?

[...]

PR #1029, opened 2022-07-12
Add support to use heap commands without debug symbols

lebr0nli:
I'm trying to make some features that #937 mentioned.

I use two classes to represent different heap resolving strategies(by debug symbols or heuristic), and use some classes in `pwndbg/heap/structs.py` to make some ctypes struct compatible with `gdb.Value` and `gdb.Type`.

We can use `set resolve-via-heuristic on/off` to enable/disable the feature in this PR.

Hope this PR is useful :)

----

P.S. ~Some features only work for i386 and x86-64 currently~, and their implements are a little bit dirty(but it seems to work). I'm sorry about that :(
Edit: It supports arm/aarch64 now

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -256,10 +256,10 @@ def OnlyAmd64(function):
             print("%s: Only works with \"x86-64\" arch." % function.__name__)
     return _OnlyAmd64
 
-def OnlyWithLibcDebugSyms(function):
+def OnlyWithResolvedHeapSyms(function):
     @functools.wraps(function)
-    def _OnlyWithLibcDebugSyms(*a, **kw):
-        if pwndbg.heap.current.libc_has_debug_syms() or pwndbg.config.resolve_via_heuristic:
+    def _OnlyWithResolvedHeapSyms(*a, **kw):
+        if pwndbg.heap.current.libc_has_debug_syms() or pwndbg.config.resolve_heap_via_heuristic:
             return function(*a, **kw)
         else:
             print('''%s: This command only works with libc debug symbols.
@@ -271,9 +271,10 @@ sudo apt-get install libc6-dbg
 sudo dpkg --add-architecture i386
 sudo apt-get install libc-dbg:i386
 ''' % function.__name__)
-            print(message.warn('pwndbg can still try to use this command without debug symbols by `set resolve-via-heuristic on`.'))
-            print(message.warn("Then pwndbg will resolve some missing symbols via heuristics, but the result of those commands may be incorrect in some cases."))
-    return _OnlyWithLibcDebugSyms
+            print(message.warn('pwndbg can still try to use this command without debug symbols by `set resolve-heap-via-heuristic on`.'))
+            print(message.warn('You can show your current config about heap by `heap_config`.'))
+            print(message.warn("Then pwndbg will resolve some missing symbols via heuristics, but the results of those commands may be incorrect in some cases."))
+    return _OnlyWithResolvedHeapSyms
 
 class QuietSloppyParsedCommand(ParsedCommand):
     def __init__(self, *a, **kw):
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -259,7 +259,7 @@ def OnlyAmd64(function):
 def OnlyWithLibcDebugSyms(function):
     @functools.wraps(function)
     def _OnlyWithLibcDebugSyms(*a, **kw):
-        if pwndbg.heap.current.libc_has_debug_syms():
+        if pwndbg.heap.current.libc_has_debug_syms() or pwndbg.config.resolve_via_heuristic:
             return function(*a, **kw)
         else:
             print('''%s: This command only works with libc debug symbols.
@@ -271,6 +271,8 @@ sudo apt-get install libc6-dbg
 sudo dpkg --add-architecture i386
 sudo apt-get install libc-dbg:i386
 ''' % function.__name__)
+            print(message.warn('pwndbg can still try to use this command without debug symbols by `set resolve-via-heuristic on`.'))
+            print(message.warn("Then pwndbg will resolve some missing symbols via heuristics, but the result of those commands may be incorrect in some cases."))
     return _OnlyWithLibcDebugSyms
 
 class QuietSloppyParsedCommand(ParsedCommand):
--- a/pwndbg/heap/__init__.py
+++ b/pwndbg/heap/__init__.py
@@ -5,22 +5,22 @@ import pwndbg.symbol
 
 current = None
 
-heap_chain_limit = pwndbg.config.Parameter('heap-dereference-limit', 8, 'number of bins to dereference')
+main_arena = pwndbg.config.Parameter('main_arena', "0", 'main_arena address for heuristics', 'heap')
 
-resolve_via_heuristic = pwndbg.config.Parameter('resolve-via-heuristic', False,
-                                                'Resolve some missing symbols via heuristics')
+thread_arena = pwndbg.config.Parameter('thread_arena', "0", 'thread_arena value for heuristics', 'heap')
 
-main_arena = pwndbg.config.Parameter('main_arena', "0", 'main_arena address for heuristics')
+mp_ = pwndbg.config.Parameter('mp_', "0", 'mp_ address for heuristics', 'heap')
 
-thread_arena = pwndbg.config.Parameter('thread_arena', "0", 'thread_arena value for heuristics')
+tcache = pwndbg.config.Parameter('tcache', "0", 'tcache value for heuristics', 'heap')
 
-mp_ = pwndbg.config.Parameter('mp_', "0", 'mp_ address for heuristics')
+global_max_fast = pwndbg.config.Parameter('global_max_fast', "0", 'global_max_fast address for heuristics', 'heap')
 
-tcache = pwndbg.config.Parameter('tcache', "0", 'tcache value for heuristics')
+symbol_list = pwndbg.config.get_params('heap')
 
-global_max_fast = pwndbg.config.Parameter('global_max_fast', "0", 'global_max_fast address for heuristics')
+heap_chain_limit = pwndbg.config.Parameter('heap-dereference-limit', 8, 'number of bins to dereference', 'heap')
 
-symbol_list = [main_arena, thread_arena, mp_, tcache, global_max_fast]
+resolve_heap_via_heuristic = pwndbg.config.Parameter('resolve-heap-via-heuristic', False,
+                                                     'Resolve missing heap related symbols via heuristics', 'heap')
 
 
 @pwndbg.config.Trigger(symbol_list)
@@ -34,14 +34,7 @@ def parse_config2address():
         if address_str == "0":
             continue
         try:
-            if address_str.startswith("0b"):
-                address = int(address_str, 2)
-            elif address_str.startswith("0o"):
-                address = int(address_str, 8)
-            elif address_str.startswith("0x"):
-                address = int(address_str, 16)
-            else:
-                address = int(address_str)
+            address = int(address_str, 0)
         except ValueError:
             symbol.value = "0"
             raise ValueError("Please input a valid integer literal string")
@@ -64,11 +57,11 @@ def reset():
         symbol.value = "0"
 
 
-@pwndbg.config.Trigger([resolve_via_heuristic])
+@pwndbg.config.Trigger([resolve_heap_via_heuristic])
 def resolve_heap(is_first_run=False):
     import pwndbg.heap.ptmalloc
     global current
-    if resolve_via_heuristic:
+    if resolve_heap_via_heuristic:
         current = pwndbg.heap.ptmalloc.HeuristicHeap()
         if not is_first_run and pwndbg.proc.alive and current.libc_has_debug_syms():
             print(message.warn(
--- a/pwndbg/heap/__init__.py
+++ b/pwndbg/heap/__init__.py
@@ -1,5 +1,5 @@
-import pwndbg.config
 import pwndbg.color.message as message
+import pwndbg.config
 import pwndbg.heap.heap
 import pwndbg.symbol
 
@@ -7,19 +7,23 @@ current = None
 
 heap_chain_limit = pwndbg.config.Parameter('heap-dereference-limit', 8, 'number of bins to dereference')
 
-resolve_via_heuristic = pwndbg.config.Parameter('resolve-via-heuristic', False, 'Resolve some missing symbols via heuristics')
+resolve_via_heuristic = pwndbg.config.Parameter('resolve-via-heuristic', False,
+                                                'Resolve some missing symbols via heuristics')
+
 
 @pwndbg.events.start
 def update():
     resolve_heap(is_first_run=True)
 
+
 @pwndbg.events.stop
 @pwndbg.events.new_objfile
-def clear():
+def reset():
     global current
     # Re-initialize the heap
     if current:
         current = type(current)()
+
 
 @pwndbg.config.Trigger([resolve_via_heuristic])
 def resolve_heap(is_first_run=False):
@@ -28,7 +32,10 @@ def resolve_heap(is_first_run=False):
     if resolve_via_heuristic:
         current = pwndbg.heap.ptmalloc.HeuristicHeap()
         if not is_first_run and pwndbg.proc.alive and current.libc_has_debug_syms():
-            print(message.warn("You are going to resolve the heap via heuristic even though you have libc debug symbols. This is not recommended!"))
+            print(message.warn(
+                "You are going to resolve the heap via heuristic even though you have libc debug symbols."
+                " This is not recommended!")
+            )
         else:
             print(message.warn("You are going to resolve the heap via heuristic. This might not work in all cases."))
     else:
--- a/pwndbg/heap/__init__.py
+++ b/pwndbg/heap/__init__.py
@@ -10,6 +10,43 @@ heap_chain_limit = pwndbg.config.Parameter('heap-dereference-limit', 8, 'number 
 resolve_via_heuristic = pwndbg.config.Parameter('resolve-via-heuristic', False,
                                                 'Resolve some missing symbols via heuristics')
 
+main_arena = pwndbg.config.Parameter('main_arena', "0", 'main_arena address for heuristics')
+
+thread_arena = pwndbg.config.Parameter('thread_arena', "0", 'thread_arena value for heuristics')
+
+mp_ = pwndbg.config.Parameter('mp_', "0", 'mp_ address for heuristics')
+
+tcache = pwndbg.config.Parameter('tcache', "0", 'tcache value for heuristics')
+
+global_max_fast = pwndbg.config.Parameter('global_max_fast', "0", 'global_max_fast address for heuristics')
+
+symbol_list = [main_arena, thread_arena, mp_, tcache, global_max_fast]
+
+
+@pwndbg.config.Trigger(symbol_list)
+def parse_config2address():
+    # Somehow, we can't use integer config because the integer is too big and will cause out of range error for GDB API
+    # So we convert the string to a int manually
+    for symbol in symbol_list:
+        if not isinstance(symbol.value, str):
+            continue
+        address_str = symbol.value.strip()
+        if address_str == "0":
+            continue
+        try:
+            if address_str.startswith("0b"):
+                address = int(address_str, 2)
+            elif address_str.startswith("0o"):
+                address = int(address_str, 8)
+            elif address_str.startswith("0x"):
+                address = int(address_str, 16)
+            else:
+                address = int(address_str)
+        except ValueError:
+            symbol.value = "0"
+            raise ValueError("Please input a valid integer literal string")
+        symbol.value = address
+
 
 @pwndbg.events.start
 def update():
@@ -23,6 +60,8 @@ def reset():
     # Re-initialize the heap
     if current:
         current = type(current)()
+    for symbol in symbol_list:
+        symbol.value = "0"
 
 
 @pwndbg.config.Trigger([resolve_via_heuristic])
--- a/pwndbg/heap/__init__.py
+++ b/pwndbg/heap/__init__.py
@@ -1,3 +1,5 @@
+import pwndbg.config
+import pwndbg.color.message as message
 import pwndbg.heap.heap
 import pwndbg.symbol
 
@@ -5,8 +7,21 @@ current = None
 
 heap_chain_limit = pwndbg.config.Parameter('heap-dereference-limit', 8, 'number of bins to dereference')
 
+resolve_via_heuristic = pwndbg.config.Parameter('resolve-via-heuristic', False, 'Resolve some missing symbols via heuristics')
+
 @pwndbg.events.start
 def update():
+    resolve_heap(is_first_run=True)
+
+@pwndbg.config.Trigger([resolve_via_heuristic])
+def resolve_heap(is_first_run=False):
     import pwndbg.heap.ptmalloc
     global current
-    current = pwndbg.heap.ptmalloc.Heap()
+    if pwndbg.config.resolve_via_heuristic:
+        current = pwndbg.heap.ptmalloc.HeuristicHeap()
+        if not is_first_run and pwndbg.proc.alive and current.libc_has_debug_syms():
+            print(message.warn("You are going to resolve the heap via heuristic even though you have libc debug symbols. This is not recommended!"))
+        else:
+            print(message.warn("You are going to resolve the heap via heuristic. This might not work in all cases."))
+    else:
+        current = pwndbg.heap.ptmalloc.DebugSymsHeap()
--- a/pwndbg/heap/__init__.py
+++ b/pwndbg/heap/__init__.py
@@ -13,11 +13,19 @@ resolve_via_heuristic = pwndbg.config.Parameter('resolve-via-heuristic', False, 
 def update():
     resolve_heap(is_first_run=True)
 
+@pwndbg.events.stop
+@pwndbg.events.new_objfile
+def clear():
+    global current
+    # Re-initialize the heap
+    if current:
+        current = type(current)()
+
 @pwndbg.config.Trigger([resolve_via_heuristic])
 def resolve_heap(is_first_run=False):
     import pwndbg.heap.ptmalloc
     global current
-    if pwndbg.config.resolve_via_heuristic:
+    if resolve_via_heuristic:
         current = pwndbg.heap.ptmalloc.HeuristicHeap()
         if not is_first_run and pwndbg.proc.alive and current.libc_has_debug_syms():
             print(message.warn("You are going to resolve the heap via heuristic even though you have libc debug symbols. This is not recommended!"))
--- a/pwndbg/heap/__init__.py
+++ b/pwndbg/heap/__init__.py
@@ -23,24 +23,6 @@ resolve_heap_via_heuristic = pwndbg.config.Parameter('resolve-heap-via-heuristic
                                                      'Resolve missing heap related symbols via heuristics', 'heap')
 
 
-@pwndbg.config.Trigger(symbol_list)
-def parse_config2address():
-    # Somehow, we can't use integer config because the integer is too big and will cause out of range error for GDB API
-    # So we convert the string to a int manually
-    for symbol in symbol_list:
-        if not isinstance(symbol.value, str):
-            continue
-        address_str = symbol.value.strip()
-        if address_str == "0":
-            continue
-        try:
-            address = int(address_str, 0)
-        except ValueError:
-            symbol.value = "0"
-            raise ValueError("Please input a valid integer literal string")
-        symbol.value = address
-
-
 @pwndbg.events.start
 def update():
     resolve_heap(is_first_run=True)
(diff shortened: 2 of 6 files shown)


disconnect3d:
Hey, thanks a lot for working on this <3. I owe you a 🍺 :D

I will try to look into it sooner or later and maybe implement some small tests for x86, x64 and arm if possible 🤐  (if you have ideas how to test it or want to implement tests urself, then go ahead!)

CptGibbon:
This is some cracking work, very creative 🤩

The only improvement I can see is that the `HeuristicHeap` version of `get_heap_boundaries()` doesn't attempt to resolve the `mp_` stuct before checking if `self._mp_addr` is populated. This means when the [heap] vm region and the actual heap start address are different, commands like "vis_heap_chunks" and "heap" will fail. A user can work around this by resolving the `mp_` struct using the "mp" command before issuing further heap commands.

Also a small typo causing issues, there are 2 members of your `c_malloc_par_2_25` struct named `mmaped_mem` & `max_mmaped_mem` which I think should be spelled "mmapped", although please correct me if I'm wrong.

Many thanks for adding this great functionality 🙏

lebr0nli:
@disconnect3d 
Hi, sorry to be replying so late. I'm trying to implement the unit test for x86 and x64 now, but I didn't have experience in writing unit tests, so it might take me some time to learn it. So if you finish your implements first, please use them directly XD
By the way, if you have any ideas or suggestions, feel free to tell me, too!

@CptGibbon 
Hi, you are right! Thanks for pointing it out!
I'll fix them in the future commit!

CptGibbon:
Awesome! Those fixes work with my testing 👍

All I can see that might need attention is the addition of another `TcachePerThreadStruct` that supports the pre-2.30 `count` fields, which are 8 bytes wide on x86_64 rather than 16 bytes. Right now commands like "tcachebins" print incorrect values with GLIBC builds that use these narrower `count` fields.

Just one more tcache-related issue; the heuristic method assumes a thread's tcache resides at the start of a heap region, which although common may not be the case once arena sharing occurs. I'm not sure how you would resolve a thread's `tcache` symbol heuristically though, and frankly this whole PR seems like magic to me already 🪄 so I think it's a reasonable compromise.

Thanks again.

lebr0nli:
> pre-2.30 `count` fields

I didn't notice that GLIBC < 2.30 is using `char counts[TCACHE_MAX_BINS];` instead of `uint16_t counts[TCACHE_MAX_BINS];`, thank you for pointing it out again 🙏
I'll fix it.

> although common may not be the case once arena sharing occurs

Oh, I forgot about this possibility, I'll take a look at the source code of `malloc.c` and try to figure out if we can find its real location like what we do on `thread_arena` currently.
Thanks!

disconnect3d:
Regarding testing, would something like qemu-user emulation be sufficient here? Then we could run e.g. arm32, aarch64, etc binaries and test different cases?

lebr0nli:
As I mentioned before, I'm not adding support to arm32 and aarch64 in my PR yet 😢, so I'm not sure about qemu-user is sufficient or not.
After I have enough knowledge about arm architecture, I'll try to add support to them if possible!(maybe need a few weeks to learn about it, not sure how long it will take)

pwndbg/heap/structs.py
```diff
@@ -0,0 +1,593 @@
...
+import pwndbg.heap
+import pwndbg.typeinfo
+import pwndbg.memory
+import pwndbg.glibc
+
+NBINS = 128
+BINMAPSIZE = 4
+TCACHE_MAX_BINS = 64
+
+if pwndbg.arch.ptrsize == 4:
```
disconnect3d, line 37: This is problematic, we need to invoke this when arch changes in GDB

(One can do sth like `attach <pid>` to attach to different architecture in GDB)

disconnect3d:
Hmm do we have then some guard code against other architectures than x86/x64?

CptGibbon:
The tcache commits are working nicely, no more issues that I can see.
Thanks for this PR 🙏

pwndbg/heap/structs.py
lebr0nli: I try to fix it in https://github.com/pwndbg/pwndbg/pull/1029/commits/d89e484ba0e47b44fad470a5e29dbc9add2113d2
I re-initialize the heap when `pwndbg.event.stop`, and I add a decorator to reload `pwndbg.heap.structs` every time we create a new instance of `HeuristicHeap`.
Does this fix the problem?

lebr0nli:
> Hmm do we have then some guard code against other architectures than x86/x64?

No, we don't have it.
I try to add them in https://github.com/pwndbg/pwndbg/pull/1029/commits/d924d988be9a0b85ea4f21431ac6e4d245f68a88 now.

lebr0nli:
I add support to arm32 and aarch64 in https://github.com/pwndbg/pwndbg/pull/1029/commits/8b5d15e2f16b7058b1d71a4bcbe7d1324a2fa840 now.

But still have some problems:

* I didn't test it on many GLIBC versions because it's a little bit difficult for me to debugging with QEMU. But at least for the arm32/aarch64 version GLIBC 2.31 which you can get via `apt-get`, it can work. So if the other version of GLIBC's assembly code pattern didn't have too much different, it should work, too. This part might need someone who has a real arm32/aarch64 device or who is familiar with arm to test it.

* ~I didn't find a way to get tls base of arm32(which means we can not find arena and tcache when we are not in the main thread), because it access a coprocessor register. And I didn't find a way to access it in GDB currently.~
Edit: I found a way to get tls base for arm via using the code gadget inside `__errno_location` now, but I think it still needs a better way to find tls for arm though. See https://github.com/pwndbg/pwndbg/pull/1029/commits/19b42c33aba3303fd9ea3cbd36b2ceceef45f768 and https://github.com/pwndbg/pwndbg/pull/1029/commits/8f6dc355faac35b6b924c70735d0932eddb20236

And after some testing, the answer of the question:

> Regarding testing, would something like qemu-user emulation be sufficient here? Then we could run e.g. arm32, aarch64, etc binaries and test different cases?

Yes, I guess qemu-user is sufficient to test it, but the vmmap need to work for QEMU.
To anyone want to test this PR with QEMU, you can re-compile @disconnect3d ’s patch for QEMU via:

```shell
git clone --branch patchew/20220221030910.3203063-1-dominik.b.czarnota@gmail.com --single-branch https://github.com/patchew-project/qemu.git
cd qemu
git submodule update --init --recursive
./configure --gdb=/usr/local/bin/gdb --enable-debug
make -j$(nproc) && make install
```

And then just remove some `is qemu` check in some `vmmap` related function. By doing this, vmmap won't work very perfectly, but it's enough to make it work.

pwndbg/commands/__init__.py
```diff
@@ -259,7 +259,7 @@ def _OnlyAmd64(*a, **kw):
 def OnlyWithLibcDebugSyms(function):
```
disconnect3d: We should rename it to `OnlyWithResolvedHeapSyms` or `OnlyWithHeap` or some other name if u have a better one

pwndbg/commands/__init__.py
```diff
@@ -259,7 +259,7 @@ def _OnlyAmd64(*a, **kw):
...
     @functools.wraps(function)
     def _OnlyWithLibcDebugSyms(*a, **kw):
-        if pwndbg.heap.current.libc_has_debug_syms():
+        if pwndbg.heap.current.libc_has_debug_syms() or pwndbg.config.resolve_via_heuristic:
             return function(*a, **kw)
         else:
             print('''%s: This command only works with libc debug symbols.
```
disconnect3d, line 265: We should improve the messaging here, but we can work it out on Discord after merging this

pwndbg/heap/__init__.py
```diff
@@ -1,12 +1,81 @@
...
+import pwndbg.config
 import pwndbg.heap.heap
 import pwndbg.symbol
 
 current = None
 
 heap_chain_limit = pwndbg.config.Parameter('heap-dereference-limit', 8, 'number of bins to dereference')
 
+resolve_via_heuristic = pwndbg.config.Parameter('resolve-via-heuristic', False,
+                                                'Resolve some missing symbols via heuristics')
```
disconnect3d:
```suggestion
                                                'Resolve missing heap related symbols via heuristics')
```

pwndbg/heap/__init__.py
```diff
@@ -1,12 +1,81 @@
+import pwndbg.color.message as message
+import pwndbg.config
 import pwndbg.heap.heap
 import pwndbg.symbol
 
 current = None
 
 heap_chain_limit = pwndbg.config.Parameter('heap-dereference-limit', 8, 'number of bins to dereference')
 
+resolve_via_heuristic = pwndbg.config.Parameter('resolve-via-heuristic', False,
```
disconnect3d:
```suggestion
resolve_via_heuristic = pwndbg.config.Parameter('resolve-heap-via-heuristic', False,
```

pwndbg/heap/__init__.py
```diff
@@ -1,12 +1,81 @@
...
+                                                'Resolve some missing symbols via heuristics')
+
+main_arena = pwndbg.config.Parameter('main_arena', "0", 'main_arena address for heuristics')
```
disconnect3d: Hmm... those kinda bloats the config parameters a bit. Maybe we should do sth like "heapconfig" namespace and put all heap related params there?

I am fine with leaving them in config for now if u don't want to deal with it, we just have to make it clear what they mean/are for in their description --- I'd add text that they are related to heap or sth

pwndbg/heap/__init__.py
```diff
@@ -1,12 +1,81 @@
...
+        address_str = symbol.value.strip()
+        if address_str == "0":
+            continue
+        try:
+            if address_str.startswith("0b"):
+                address = int(address_str, 2)
+            elif address_str.startswith("0o"):
+                address = int(address_str, 8)
+            elif address_str.startswith("0x"):
+                address = int(address_str, 16)
```
disconnect3d: All those if/elif here can be replaced with `int(address_str, 0)` - it will determine the format

pwndbg/heap/__init__.py
```diff
@@ -1,12 +1,81 @@
...
+
+
 @pwndbg.events.start
 def update():
+    resolve_heap(is_first_run=True)
+
+
+@pwndbg.events.stop
+@pwndbg.events.new_objfile
+def reset():
```
disconnect3d, line 33: What's the reason to reset on each stop?

pwndbg/heap/__init__.py
lebr0nli: Wow, I didn't know this cool feature before!
Thanks, I'll fix it!

pwndbg/heap/__init__.py
lebr0nli: That's because I want to make sure we refresh the heap like we update `pwndbg.typeinfo` when the file changes or process stopped, so we can avoid using some old addresses or wrong structs definitions.(btw my commit about it is https://github.com/pwndbg/pwndbg/commit/d89e484ba0e47b44fad470a5e29dbc9add2113d2)
But a little side effect is that not only when the file we want to debug changed, when loading some shared object files, the heap will keep refreshing too, and this wastes some resources.
So I think this part might need a better implementation, do you have any suggestions about this, or should I keep this part?
BTW, the way I use `importlib.reload` to reload structs seems a sort of dirty to me, should I write all the variables in an `update()` like what we did in `pwndbg.typeinfo`? It seems a little bit cleaner to me, but I'm not sure:
https://github.com/pwndbg/pwndbg/blob/e867d1db0c51017d5df927d86c33dd6b9b733456/pwndbg/typeinfo.py#L42-L47

pwndbg/heap/__init__.py
lebr0nli: Sure, I'll try to fix it! (with scope)

pwndbg/heap/__init__.py
lebr0nli: Sure, I'll try to fix it!

pwndbg/heap/__init__.py
lebr0nli: Sure, I'll try to fix it!

pwndbg/commands/__init__.py
lebr0nli: I agree with yours.
I think `OnlyWithResolvedHeapSyms` is good, I'll use it.

pwndbg/commands/__init__.py
lebr0nli: Sure!
You mean the hint messaging for it, right?

pwndbg/heap/__init__.py
lebr0nli: I added a new command called `heap_config` to show configs in `heap` scope(https://github.com/pwndbg/pwndbg/pull/1029/commits/289010da1dd4839bdb628b1f83390bbc352f88d5), does this fix the issue?

pwndbg/heap/__init__.py
disconnect3d: RE: better idea for this - not on top of my head, I will try to see if we have anything like that elsewhere.
\
Another idea could be only to mark that heap is "dirty" here and refresh it only on demand when someone asks for information regarding it. I guess that would be better?

disconnect3d: Regarding `importlib.reload` - yeah, I also think it is not the best approach, but we also do it in one of `update`, to reset elf types or something like that, if I recall correctly.

(I'm totally up for not using `reload` at all in those cases, so if u want to move it to `update` I agree on that)

pwndbg/commands/__init__.py
disconnect3d: yeah

pwndbg/heap/__init__.py
lebr0nli: > refresh it only on demand when someone asks for information regarding it. I guess that would be better?

Yes, I agree!
But sadly, after a few days of testing, I didn't figure out a good way to do it 😢
I'll update this part if I have a better ideas in the future.

> Regarding `importlib.reload` - yeah, I also think it is not the best approach, but we also do it in one of `update`, to reset elf types or something like that, if I recall correctly.
> 
> (I'm totally up for not using `reload` at all in those cases, so if u want to move it to `update` I agree on that)

Ok, I see.
And after some testing, I found that using `update()` might cause some errors in this case actually, so I'll keep the part of `importlib.reload`!

pwndbg/heap/ptmalloc.py
```diff
@@ -547,11 +487,906 @@ def largebin_index(self, sz):
...
+        if arena_addr is None:
+            if self.multithreaded:
+                arena_addr = pwndbg.memory.u(pwndbg.symbol.address('thread_arena'))
+                if arena_addr > 0:
+                    return pwndbg.memory.poi(self.malloc_state, arena_addr)
+
+            return self.main_arena
+
+        try:
+            next(i for i in pwndbg.vmmap.get() if arena_addr in i)
```
disconnect3d, line 604: @lebr0nli What's the point of this `next(...)`?

pwndbg/heap/ptmalloc.py
```diff
@@ -547,11 +487,906 @@ def largebin_index(self, sz):
...
+        self._errno_offset = None
+
+    @property
+    def main_arena(self):
+        main_arena_via_config = int(str(pwndbg.config.main_arena), 0)
+        if main_arena_via_config > 0:
+            return self.malloc_state(main_arena_via_config)
+        # TODO/FIXME: These are quite dirty, we should find a better way to do this
+        if not self._main_arena_addr:
+            if pwndbg.glibc.get_version() < (2, 34) and pwndbg.arch.current != "arm":
```
disconnect3d, line 685: Just to be sure: what happens for "arm"? Is arm 32-bit fine?

pwndbg/heap/ptmalloc.py
```diff
@@ -547,11 +487,906 @@ def largebin_index(self, sz):
...
+                            reg = ldrw_instr.operands[0].str
+                            if instr.mnemonic == 'add' and instr.op_str == reg + ", pc":
+                                # ldr.w reg, [pc, #offset]
+                                offset = ldrw_instr.operands[1].mem.disp
+                                offset = pwndbg.memory.s32((ldrw_instr.address + 4 & -4) + offset)
+                                # add reg, pc
+                                self._main_arena_addr = offset + instr.address + 4
+                else:
+                    inform_report_issue("main_arena")
+                    raise OSError("Cannot find the symbol via heuristics")
```
disconnect3d, line 740: Does this make sense? We have the last resort check of finding `main_arena` via _IO_* symbols in libc in lines 742+ --- shall we really raise an exception here instead of trying out a different approach?

disconnect3d: We have an "else" with the same code at the end of the whole function, so maybe we should not fallback here and try more methods

pwndbg/heap/ptmalloc.py
```diff
@@ -547,11 +487,906 @@ def largebin_index(self, sz):
...
+                # main_arena is between _IO_2_1_stdin and _IO_list_all
+                for addr in range(_IO_2_1_stdin_addr, _IO_list_all_addr, pwndbg.arch.ptrsize):
+                    tmp_arena = self.malloc_state(addr)
+                    if tmp_arena["next"] == addr:
+                        self._main_arena_addr = addr
+                        break
+                if not self._main_arena_addr:
+                    # there are more than one arena, try to find by main_arena.top and main_arena.max_system_mem
+                    heap_page = next(x for x in pwndbg.vmmap.get() if "heap]" in x.objfile)
+                    for addr in range(_IO_2_1_stdin_addr, _IO_list_all_addr, pwndbg.arch.ptrsize):
```
disconnect3d, line 754: How does `main_arena->next` work? I mean, I can see that when there is only one arena, the `main_arena->next == main_arena`. If we have two arenas, is it cyclic like `main_arena->next->next == main_arena`? If so, why don't we try to resolve it like this?

pwndbg/heap/ptmalloc.py
```diff
@@ -547,11 +487,906 @@ def largebin_index(self, sz):
...
+        return self._main_arena
+
+    def has_tcache(self):
+        # TODO/FIXME: Can we determine the tcache_bins existence more reliable?
+
+        # There is no debug symbols, we determine the tcache_bins existence by checking glibc version only
+        return self.is_initialized() and pwndbg.glibc.get_version() >= (2, 26)
+
+    def get_tls_base_via_errno_location(self):
+        assert pwndbg.arch.current in ("x86-64", "i386", "arm")
```
disconnect3d, line 776: @lebr0nli can we add here some comment about Aarch64, as in why we don't have it here?

From https://maskray.me/blog/2021-02-14-all-about-thread-local-storage it seems TLS exists on aarch64... (well, why would it not :D)

pwndbg/heap/ptmalloc.py
```diff
@@ -547,11 +487,906 @@ def largebin_index(self, sz):
...
+        # TODO/FIXME: Need a better way to find tls
```
disconnect3d, line 777: Perhaps we should move TLS finding to an external function too

pwndbg/heap/ptmalloc.py
```diff
@@ -547,11 +487,906 @@ def largebin_index(self, sz):
...
+        already_lock = gdb.parameter("scheduler-locking") == "on"
+        if not already_lock:
+            gdb.execute("set scheduler-locking on")
+        errno_addr = int(gdb.parse_and_eval("(int *)__errno_location()"))
```
disconnect3d, line 781: Is there a reason we make the call and not fetch it by the `errno` symbol address?

disconnect3d: Maybe we don't have errno symbol address sometimes lol

pwndbg/heap/ptmalloc.py
```diff
@@ -547,11 +487,906 @@ def largebin_index(self, sz):
...
+                                                        show_prev_insns=False)
+            if pwndbg.arch.current == "x86-64":
+                for instr in __errno_location_instr:
+                    # Find something like: mov rax, qword ptr [rip + disp]
+                    if instr.mnemonic == 'mov':
+                        self._errno_offset = pwndbg.memory.s64(instr.next + instr.disp)
+            elif pwndbg.arch.current == "i386":
+                for instr in __errno_location_instr:
+                    # Find something like: mov eax, dword ptr [eax + disp]
+                    # (disp is a negative value, and eax is come from one of the page address of libc)
```
```suggestion
                    # (disp is a negative value, and eax comes from one of the page address of libc)
```

pwndbg/heap/ptmalloc.py
```diff
@@ -547,11 +487,906 @@ def largebin_index(self, sz):
...
+                                found = True
+                                break
+                    elif instr.mnemonic == 'add':
+                        for reg in regs:
+                            if instr.op_str == reg + ", pc":
+                                ldr[reg] = instr
+                    elif instr.mnemonic == 'ldr' and "[pc," in instr.op_str:
+                        regs[instr.operands[0].str] = instr
+            else:
+                inform_report_issue("mp_")
```
disconnect3d, line 1185: Do we need this else here if there is another one below?

disconnect3d:
@lebr0nli Merged, but please look at the comments above :) Thanks again!

lebr0nli:
No problem, I'll take a look!
But I'm a little bit busy this week, I'll reply to those at the night of my time zone(here are 6 am on Monday now haha) or this Wednesday when I have time!

pwndbg/heap/ptmalloc.py
lebr0nli: Hmm, GEF's heuristic for `main_arena` seems to work fine for i386, x86-64, aarch64 with GLIBC < 2.34, but I'm not sure what happened on the heuristic of GEF for 32-bit arm actually, so I decided to use our version of heuristic.
And just guessing, maybe it's because GEF got the wrong `malloc_state` size for arm 32-bit?
You might reproduce it with a normal GLIBC 2.31 which is installed via `apt-get install arm-linux-gnueabihf-gcc`.

Here is a example, if we try the code from https://github.com/hugsy/gef/blob/dev/gef.py#L1180

```
gef➤  pi
>>> ctypes.sizeof(GlibcArena.malloc_state_t())
1116
>>> malloc_hook_addr = parse_address("(void *)&__malloc_hook")
>>> struct_size = ctypes.sizeof(GlibcArena.malloc_state_t())
>>> hex(malloc_hook_addr - gef.arch.ptrsize - struct_size)
'0x3f7bf76c'
>>>
gef➤  ptype/o main_arena
/* offset      |    size */  type = struct malloc_state {
/*      0      |       4 */    __libc_lock_t mutex;
/*      4      |       4 */    int flags;
/*      8      |       4 */    int have_fastchunks;
/*     12      |      40 */    mfastbinptr fastbinsY[10];
/*     52      |       4 */    mchunkptr top;
/*     56      |       4 */    mchunkptr last_remainder;
/*     60      |    1016 */    mchunkptr bins[254];
/*   1076      |      16 */    unsigned int binmap[4];
/*   1092      |       4 */    struct malloc_state *next;
/*   1096      |       4 */    struct malloc_state *next_free;
/*   1100      |       4 */    size_t attached_threads;
/*   1104      |       4 */    size_t system_mem;
/*   1108      |       4 */    size_t max_system_mem;

                               /* total size (bytes): 1112 */
                             }
gef➤  p &main_arena
$1 = (struct malloc_state *) 0x3f7bf76c <main_arena>
```
As you can see, idk why, GEF's heuristic got a correct `main_arena` address with a wrong size of `malloc_state` lol
But because our size is correct, it will failed.

pwndbg/heap/ptmalloc.py
lebr0nli: Good idea!
I didn't think about this before, I'll try to support what you said.

pwndbg/heap/ptmalloc.py
lebr0nli: Yeah, aarch64 has TLS, too.
But the `TPIDR_EL0` register which is used for aarch64 TLS seems always work.
So I didn’t implement that for it.
Is it possible that `TPIDR_EL0` failed to lead us to find TLS base?
If the answer is yes, maybe we need to implement for aarch64, too.

For other architectures:
I use this function for arm32 because I can’t find a way to access the coprocessor register(if I didn't misunderstand it).
And I use this for x86/x64 because sometimes we can’t always find gsbase/fsbase since its implementation is limited. 

And sure, I'll add some comment about it.

pwndbg/heap/ptmalloc.py
lebr0nli: Ok, should we move this part to a new file, for example, `pwndbg/tls.py` or something?
And maybe we can add a new command in `pwndbg/commands/tls.py`, which is called `tls` or `tls_base` to let user to use it?

pwndbg/heap/ptmalloc.py
lebr0nli: > Maybe we don't have errno symbol address sometimes lol

Yes, `errno` symbol doesn't seem to work for qemu-user and some weird cases I guess.
And I'm not sure why, the result of `pwndbg.symbol.address('errno')` will not the same as `(int *)__errno_location` sometimes.

pwndbg/heap/ptmalloc.py
lebr0nli: Maybe don't, so more architectures might be supported by heuristic.
Thanks! I'll fix it.

pwndbg/heap/ptmalloc.py
lebr0nli: Make sense!
Thanks, I'll fix it.

pwndbg/heap/ptmalloc.py
lebr0nli: It want to make sure the arena address is a valid address in memory (when `next` failed, it means we can't find it in memory)?
But it seems a little bit expensive because it use exception to handle it, maybe I should use `pwndbg.vmmap.find(arena_addr) is None` to determine it?
I'm not sure, I just refactor some of the original methods from `Heap` to `DebugSymsHeap` 😆

lonnywong:
Thanks for this great feature.

Can we resolve heap via heuristic only without libc debug symbols? I want to set resolve-heap-via-heuristic on by default. If there is libc debug symbols, use it first. It would be better if there is a hint shows that current has no libc debug symbols and using heuristic, at the first heap command.

Now I have to manually choose heuristic or not.

https://github.com/pwndbg/pwndbg/blob/ec8addd60243d78e6e02202cc00c3d2ace330392/pwndbg/commands/heap.py#L30-L33

https://github.com/pwndbg/pwndbg/blob/ec8addd60243d78e6e02202cc00c3d2ace330392/pwndbg/heap/__init__.py#L54-L70

lebr0nli:
> Can we resolve heap via heuristic only without libc debug symbols? I want to set resolve-heap-via-heuristic on by default. If there is libc debug symbols, use it first.

Good suggestion, but instead of turning it on by default, maybe it will be better that we set it to `auto` by default, then `auto` will do what you said(use heuristic when there are no debug symbols), but we can still let the user turn it on or off manually?

@disconnect3d, what do you think about my comment on this and @lonnywong's suggestion?

disconnect3d:
@lebr0nli I wanted to propose what @lonnywong wrote: to use it by default when debug symbols are missing. But tbh I like your solution with `auto` more.

On the other hand, will users actually use it? Or is this mainly to let users to disable the heap heuristics resolution?

CptGibbon:
+1 for `auto`

lebr0nli:
@disconnect3d @CptGibbon Ok, I see!

> On the other hand, will users actually use it? Or is this mainly to let users to disable the heap heuristics resolution?

@disconnect3d, I think yes, but only used for some weird CTF challenges probably xD.

I mean, let's say we have a libc that has every debug symbol except `main_arena` and `struct malloc_state`, then we'll need to turn on heap heuristic explicitly to make heap commands work if we "only" turn on heap heuristic when there's no debug symbol. So that's why I prefer `auto` instead of `on` more.
(But yeah, this scenario is a little bit over design I guess lol)

lonnywong:
How about `off`, `auto` and `force`? ( maybe over designed )

If there are no libc debug symbols for `auto` mode, I'd like to get a hint, maybe I'll try to install it.

lebr0nli:
> How about `off`, `auto` and `force`?

Oh, good idea, I just noticed that there's a parameter class called `gdb.PARAM_ENUM`, I'll take a look at whether `gdb.PARAM_ENUM` can help or not for this. (I only knew `gdb.PARAM_AUTO_BOOLEAN` before, so what I thought at first was that we can only use `on`/`off`/`auto` :P)

Btw I think `pwndbg.config.Parameter` might need some updates first because it only accepts a few classes of parameters, and we can't control the full content of `help set SOME_CONFIG_OF_PWNDBG` by customizing its `__doc__`.
I'll take a look at this in my free time, too.

disconnect3d:
> Btw I think pwndbg.config.Parameter might need some updates first because it only accepts a few classes of parameters, and we can't control the full content of help set SOME_CONFIG_OF_PWNDBG by customizing its __doc__.

:+1: we discussed somewhere else recently that I'd be great to enhance the config.Parameter class to support more robust help strings for parameters.

lebr0nli:
> 👍 we discussed somewhere else recently that I'd be great to enhance the config.Parameter class to support more robust help strings for parameters.

Oh, I saw the discussion about it in another PR now, that's great!

kunshim:
How can I use this feature? Is there special command need to use this feature?

lebr0nli:
> How can I use this feature? Is there special command need to use this feature?

In the latest pwndbg, by default, the heap heuristic(the feature you mentioned) should automatically be the fallback when pwndbg fails to use debug symbols to resolve the heap, so you don't need any special command to use/enable it.
But in some cases, you might need to manually set something more to make the heap heuristic work correctly.
For this, you can use `help set resolve-heap-via-heuristic` for more information about this feature.

PR #1030, opened 2022-07-15
Try fs/gs_base registers before ptrace'ing

lebr0nli:
`fsbase()` and `gsbase()` in `pwndbg/regs.py` will use ptrace by ourself to find segment base, but it doesn't work on remote debugging.

But GDB >= 8 can use `fs_base` and `gs_base` register to get the same result for x86-64 processes, and this feature works when remote debugging.

So by checking these two registers, even if we are using gdbserver to debug a x86-64 process, we can still use `fsbase`/`gsbase` without any problems if our GDB version >= 8

### DIFF
--- a/pwndbg/regs.py
+++ b/pwndbg/regs.py
@@ -389,12 +389,24 @@ class module(ModuleType):
     @property
     @pwndbg.memoize.reset_on_stop
     def fsbase(self):
-        return self._fs_gs_helper(ARCH_GET_FS)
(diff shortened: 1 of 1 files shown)


disconnect3d:
@lebr0nli Thanks! Fyi I have refactored this a little bit in https://github.com/pwndbg/pwndbg/commit/71219361f55baa5d723582ef4764315709d7210e

Issue #939, opened 2021-08-09
Feature: Support printf-style functions which have varargs


WeSeekAfterKnowledge:
This is a HUGE stretch goal, but it would be neat.

We already have `functions.py` which specify the arguments to various functions, e.g.:

```
>>> from functions import functions
>>> functions['printf']
Function(type='int', derefcnt=0, name='printf', args=[Argument(type='char', derefcnt=1, name='format'), Argument(type='int', derefcnt=0, name='vararg')])
>>> functions['dprintf']
Function(type='int', derefcnt=0, name='dprintf', args=[Argument(type='int', derefcnt=0, name='fd'), Argument(type='char', derefcnt=1, name='fmt'), Argument(type='int', derefcnt=0, name='vararg')])
>>> functions['fprintf']
Function(type='int', derefcnt=0, name='fprintf', args=[Argument(type='FILE', derefcnt=1, name='stream'), Argument(type='char', derefcnt=1, name='format'), Argument(type='int', derefcnt=0, name='vararg')])
```

It would be pretty neat if the "function argument unroller" that we currently have would recognize some subset of vararg-ish functions, and attempt to interpret the remainder.

This would be helpful both for format-string vulnerabilities, as well as general debugging.

disconnect3d:
A part of this work has been done in https://github.com/pwndbg/pwndbg/pull/1038 but has to be iterated upon

PR #1038, opened 2022-07-16
Accept args given in the format string of printf-like functions

generalmimon:
Cc @disconnect3d (made on the EuroPython 2022 sprint)

See #939

It doesn't really adjust the value format (as I thought it would when I provide the argument type, but it turns out it's not interpreted anywhere), so e.g. for floats you won't get anything nice as 1.5 but still some unreadable 0x..., but at least it's a start of reading `printf` arguments.

### DIFF
--- a/pwndbg/arguments.py
+++ b/pwndbg/arguments.py
@@ -216,6 +216,7 @@ def enhance_printf_args(orig_args, format_str):
     len_modifier_table = {
         'hh': 'char',
         'll': 'long long',
+        'h': 'short',
         'l': 'long',
         'j': 'intmax_t',
         'z': 'size_t',
--- a/pwndbg/arguments.py
+++ b/pwndbg/arguments.py
@@ -146,7 +146,16 @@ def get(instruction):
 
     if func:
         args = func.args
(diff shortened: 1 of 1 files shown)


pwndbg/arguments.py
```diff
@@ -196,6 +205,72 @@ def arguments(abi=None):
...
+        if len_modifier_2 in ('hh', 'll'):
+            specifier_type = len_modifier_table[len_modifier_2]
+            i += 2
+        else:
+            len_modifier_1 = format_str[i]
+            if len_modifier_1 in len_modifier_table:
+                specifier_type = len_modifier_table[len_modifier_1]
+                i += 1
+
+        conv_specifier = format_str[i]
```
generalmimon, line 269: This may actually raise an `IndexError` if the format string is malformed.

Same for this line:

https://github.com/pwndbg/pwndbg/blob/6cbd6a177ff343ca317928ee792b240f7e1abf55/pwndbg/arguments.py#L251

disconnect3d:
I'll close this for now but keep a reference to it in #939

PR #1042, opened 2022-08-05
fix context showing unexpectedly on first command

lonnywong:


disconnect3d:
I am not sure I get what it changes :D. Can you show a screen before/after or how can I reproduce the 'bad' behavior?

lonnywong:
> I am not sure I get what it changes :D. Can you show a screen before/after or how can I reproduce the 'bad' behavior?

1. gdb attach to a running process.
2. execute any pwndbg command.

disconnect3d:
If you do: `gdb` and then `attach <pid>` then it does not occur. It occurs in a `gdb -p <pid>` case.

disconnect3d:
I still don't quite understand why this fixes the problem, but it fixes it, so thanks a lot for it! :)

lonnywong:
> I still don't quite understand why this fixes the problem, but it fixes it, so thanks a lot for it! :)

Check the code:
https://github.com/pwndbg/pwndbg/blob/f2e5c98f238320c8a09ef5b472d9116fe7da2640/pwndbg/prompt.py#L50

`*a` make sure it passes as is.

PR #1051, opened 2022-08-10
Add $heap_base variable after running the heap command

### LINKED ISSUE #1049: Define $heap_base variable once the heap has been initialized
While commands like `vis_heap_chunks` are great, they often aren't very usable once the heap becomes corrupted. At this point I often have to do something like `x/100xg 0x55555679a000` where 0x55555679a000 is the heap base address I got from running the `heap` or `vmmap` commands.

While this works, it would be much more convenient if a `$heap_base` variable was automatically defined as soon as the heap was initialized (or at the very least, after the first successful run of some heap command). Then I could reuse commands from my GDB history even when addresses change with ASLR, and I can even use this variable in GDB scripts when launching the binary from pwntools, which will let me dump a lot of useful debugging information about the heap automatically from my exploit script.

If something like this already exists, please let me know. Otherwise, does it make sense to add this to pwndbg?

gsingh93:
This PR sets the `$heap_base` GDB variable once the `heap` command is run for the first time, allowing it to be used in later GDB commands or in GDB scripts.

Fixes #1049

### DIFF
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -106,6 +106,10 @@ def heap(addr=None, verbose=False, simple=False):
     top_chunk = arena['top']
     ptr_size = allocator.size_sz
 
+    # Store the heap base address in a GDB variable that can be used in other
+    # GDB commands
+    gdb.execute('set $heap_base=0x{:x}'.format(heap_region.start))
+
     # Calculate where to start printing; if an address was supplied, use that,
     # if this heap belongs to the main arena, start at the beginning of the
     # heap's mapping, otherwise, compensate for the presence of a heap_info


disconnect3d:
@gsingh93 We should probably [add a tip](https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/tips.py) on that and document it somehow e.g. in FEATURES.md :P

CptGibbon:
I really like this feature and will be using it a lot ❤️
I think it could be improved with a couple of small changes:

- Check if the user has already set a convenience variable of the same name, either with `init-if-undefined` or `gdb.convenience_variable("heap_base")`. This way we won't break user scripts.
- Consider using `gdb.set_convenience_variable("heap_base", <value>)` in place of `gdb.execute()`
- Perhaps `$heap_base` could be populated on heap initialization, rather than after running one specific heap command

Thanks for this PR!

gsingh93:
Thanks for the feedback @CptGibbon. Those are all great suggestions, but I'm not sure when I'll get time to make those changes. For now I've filed these suggestions as a new issue so they don't get lost: https://github.com/pwndbg/pwndbg/issues/1060

PR #1052, opened 2022-08-10
Remove Python 2 legacy code

ivellios:
Since `pwndbg` has dropped support for Python 2 all 2.x-related code parts can be safely removed.


disconnect3d (approved):
❤️

PR #1056, opened 2022-08-10
add dxc commands

### LINKED ISSUE #1018: dump memory as code format (like c or python)
Some time I need dump a short memory and use in my own code. x64dbg has a useful feature can dump memory as some code format (just like below).
image
Not a very complex.

vesim987:
This is based on work from @Cirn09 from gist in the issue below.

Closes https://github.com/pwndbg/pwndbg/issues/1018

### DIFF
--- a/pwndbg/__init__.py
+++ b/pwndbg/__init__.py
@@ -20,6 +20,7 @@ import pwndbg.commands.comments
 import pwndbg.commands.config
 import pwndbg.commands.context
 import pwndbg.commands.cpsr
+import pwndbg.commands.dxc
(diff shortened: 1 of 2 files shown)


disconnect3d:
@vesim987 I think we should rather add optional flags to those commands instead of adding new commands.

Then we can hint about it in our hints and in the docs in the future.

@vesim987 can you update this to be a flag?

disconnect3d:
Closing this. Please reopen with another PR that will add flags to those commands instead.

Another solution could be a single command to dump memory with given format etc. instead of many commands as in here

PR #1059, opened 2022-08-11
Add next_malloc_chunk command

gsingh93:
The `malloc_chunk` command already exists to print a chunk at a particular address. But it's also useful to be able to easily print out the next chunk without having to do any additional pointer arithmetic inside GDB. To do this, I've added the `next_malloc_chunk` command.

My use case for this is the following situation. I've created a fake chunk here that I'm about to `free`:
```
pwndbg> malloc_chunk -v 0x56136f784340
Allocated chunk | PREV_INUSE
Addr: 0x56136f784340
prev_size: 0x00
size: 0x421
fd: 0x56136f784350
bk: 0x21
fd_nextsize: 0x56160e4eb400
bk_nextsize: 0x00
```

Everything looks good, but let me double check the next chunk:
```
pwndbg> next_malloc_chunk -v 0x56136f784340
Allocated chunk
Addr: 0x56136f784760
prev_size: 0x00
size: 0x00
fd: 0x00
bk: 0x00
fd_nextsize: 0x00
bk_nextsize: 0x00
```

We can see that my next chunk is actually invalid (`prev_inuse` bit should be set if we're freeing the original chunk), so doing the free will crash. Without this command, it's tedious to check things like this.

### DIFF
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -337,6 +337,30 @@ def malloc_chunk(addr, fake=False, verbose=False, simple=False):
 
 
 parser = argparse.ArgumentParser()
+parser.description = "Print the next chunk."
+parser.add_argument("addr", type=int, help="Address of the chunk (malloc_chunk struct start, prev_size field).")
+parser.add_argument("-f", "--fake", action="store_true", help="Is this a fake chunk?")
+parser.add_argument("-v", "--verbose", action="store_true", help="Print all chunk fields, even unused ones.")
+parser.add_argument("-s", "--simple", action="store_true", help="Simply print malloc_chunk struct's contents.")
+@pwndbg.commands.ArgparsedCommand(parser)
+@pwndbg.commands.OnlyWhenRunning
+@pwndbg.commands.OnlyWithLibcDebugSyms
+@pwndbg.commands.OnlyWhenHeapIsInitialized
+def next_malloc_chunk(addr, fake=False, verbose=False, simple=False):
+    """Print a the next malloc_chunk struct's contents."""
+    # points to the real start of the chunk
+    cursor = int(addr)
+
+    allocator = pwndbg.heap.current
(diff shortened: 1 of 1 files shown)


CptGibbon:
Cool idea, does this need to be a whole new command rather than a flag for the existing `heap` or `malloc_chunk` commands? Perhaps one that uses +/- to print a number of preceding/succeeding chunks.

gsingh93:
Either way works for me. Anyone else have any preferences? cc: @disconnect3d 

The one downside I can think of with a flag is that discoverability is hard. I can't tell you how many useful commands I've found just by scrolling through the output of `pwndbg`.

CptGibbon:
You make a good point ref discoverability, perhaps it's better the way it is.

disconnect3d:
The discoverability is a good point but I think that we can have that with a flag to `malloc_chunk` command as well!

We can just change its description to: `malloc_chunk : Prints a chunk or next chunks when --next is applied` or something similar.

So yeah, let's move this to a flag pls :)

disconnect3d:
FWIW @gsingh93 if you are not on our Discord server, feel free to join at https://discord.gg/yxrAYkfvsy ; we sometimes discuss things there ;)

CptGibbon:
Why stop at the next chunk? The chunk after that can be just as important, as can the previous chunk. Maybe a flag that lets you print a user-defined number of next/previous/surrounding chunks would be useful.

gsingh93:
Great suggestion, will attempt to add this when I convert this to a flag.

gsingh93:
This will actually be much easier to implement after https://github.com/pwndbg/pwndbg/pull/1063 lands (there's a new generator method to make it easier to iterate over chunks), and I'll be able to reuse/share code between the `heap` and `malloc_chunk` method as well, so I'm going to close this PR. Once the other PR lands, I'll submit a new one to add this feature.

PR #1061, opened 2022-08-11
Format with autopep8

gsingh93:
As discussed in https://github.com/pwndbg/pwndbg/issues/990, it would be good to have a consistent coding style by using an autoformatter. While `pep8` isn't my preferred style format, there seems to already be a `setup.cfg` file with some configuration options in it. All the changes in this PR were made by `autopep8` with the command `autopep8 --recursive --in-place .`.

If we would like to move away from `pep8` or customize its configuration any more, now is probably the best time to do so to avoid any other mass autoformatting PRs.


disconnect3d:
We had a discussion about it on our recent on-site Pwndbg coding sprint in Cracow... and we discussed about using Black.

I haven't used Black much but AFAIK its currently a standard and the most (???) used linting tool in Python?

cc: @artcz @Arusekk for opinions 🤐

gsingh93:
As discussed on Discord, I formatted with both `black` (https://github.com/gsingh93/pwndbg/commit/6f69e22567477c4dc110a25b98e01c7627e59f08) and `yapf` (https://github.com/gsingh93/pwndbg/commit/775436222b59c8af9630aa58b871d2a011e27f3a). `black` is known for not being configurable, and for this being a feature: if there are no style configuration options, there's nothing for developers to argue about. The style they've chosen is documented [here](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html). Note that by default `black` converts single-quoted strings to double-quoted, but this can be configured with `--skip-string-normalization`.

`yapf` takes a different approach. It gives you lots of configuration options, all of which are documented in the [README](https://github.com/google/yapf). It has some style "presets" you can choose to set some defaults for these options: `pep8`, `google`, `yapf`, `facebook`. From my own projects, I've been pretty happy with the `facebook` preset with some minor modifications, so this is the config that got used:
```
[style]
based_on_style = facebook
split_before_bitwise_operator = true
split_before_logical_operator = true
```

Personally, I don't see `black`s lack of configuration options as a feature, and much prefer that we just pick some decent `yapf` configuration, as that always gives us the option to tweak settings if the future if we want to (for good reason, not just for personal preference). But at the end of the day I think both are fine, so I'll leave it up to the project maintainers.

gsingh93:
Any thoughts on this? Once the PR gets merged with either autoformatter, we'll have gone from ~6000 PEP8 violations to a few hundred, and I plan to fix a few more of those violations manually (some actually seem like bugs).

artcz:
I might be biased here because I like 99% of the decisions in black, but also black is becoming a de-facto standard with major packages adopting it which has an extra benefit of most of the code (not just within pwndbg) looking similar. 

I think of the two options black is probably better, and if in some places we need a very specific code formatting we can always use the `# fmt: off` for certain blocks of code.

disconnect3d:
Let's go with `black`. It will leave us with... less opinions :P. 

@gsingh93 can you either change this PR or make another one that will introduce black on CI and set its config? I'll merge it with failing CI and then will run black locally to format stuff and commit that afterwards.

(That way I won't have to double check/review all the code changes)

gsingh93:
Done, see #1101

PR #1062, opened 2022-08-11
Fix parsing complex symbols (#1050)

### LINKED ISSUE #1050: Commands that take address can't parse complex symbols
For complex symbols, e.g. comming from C++, GDB can parse them when we provide them in quotes. E.g. `x/10i &'A::foo()'`.

Sadly, the same does not work with Pwndbg commands - e.g. `vmmap` or `nearpc` just can't parse those.

We need to find a way to **properly** parse addresses/symbols once and for all.

This bug can be reproduced with the following C++ program:
```cpp
#include <stdio.h>

struct A {
    void foo() { puts("foo"); }
};

int main() {
    A a;
    a.foo();
}
```

Compiled as `g++ -ggdb -O0 main.cpp -o a.out` and then ran in GDB.

Then, `x/10i &'A::foo()'` will show us disassembly but `nearpc` or `vmmap` on the same symbol will not work.

ntsleep:
A quick attempt to fix https://github.com/pwndbg/pwndbg/issues/1050. It requires more testing with different commands.

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -299,7 +299,8 @@ class _ArgparsedCommand(Command):
         super(_ArgparsedCommand, self).__init__(function, command_name=command_name, *a, **kw)
 
     def split_args(self, argument):
-        argv = gdb.string_to_argv(argument)
+        escaped_arg = argument.replace("'", "\\'")  # Escaping apostrophes is needed for arguments like `&'A::foo()'`
+        argv = gdb.string_to_argv(escaped_arg)
         return tuple(), vars(self.parser.parse_args(argv))
 
 


disconnect3d:
Yeah... so as we can see in the tests failing, while this fixes the issue with parsing complex symbols, it unfortunately breaks the case when we just want to pass in a space in quotes as an argument.

image

FWIW I have asked on GDB irc channel about this behavior vs the expected GDB behavior on parsing commands (because oh well, the  `print` or `examine` commands in GDB parses those complex symbols properly when quotations are there) and I got no reply yet. I will try to ping the corresping people there.

Btw @ntsleep we should add a testcase for this behavior first and work on a solution that does not break other tests.

Though, I am not sure what that solution would be yet... I think we should maybe try to debug the GDB's print/examine commands and the way they parse input and see if that would give us a hint.

disconnect3d:
This seems invalid/not an issue. See #1050.

PR #1063, opened 2022-08-13
Refactor heap code

gsingh93:
This is a large PR that refactors a lot of the heap code in `commands/heap.py` and `heap/ptmalloc.py`. The main goal of refactoring was to make it easier to add new functionality. While adding some new functionality, I was frustrated with the amount of implementation details I needed to learn, for `example` I shouldn't have to add `2*sizeof(size_t)` to a tcache chunk before searching for it in bins. And because we're duplicating so much code, it's more time consuming to make changes because you need to change multiple locations in the code you're not familiar with, as each duplication needs to be updated.

To improve this, I've created a `Bins` and `Bin` class, which can help abstract all of these implementation details and allow us to reuse code more often. I've also made some small improvements to `ptmalloc.py`, such as making it easier to iterate over all heap chunks.

This is a large PR, and it's still a work in progress. I have some TODOs to finish, more manual testing to do, I need to fix the existing unit tests and lint issues, and I want to add more automated tests as well. I'm putting it up early so I can get some early feedback if people think this is going in completely the wrong direction, as well as to give a heads up that this is happening so we can coordinate around merging new heap PRs so I don't have to fix as many merge conflicts.

The best way to review this will be commit by commit, as opposed to the whole PR at once.

### DIFF
--- a/pwndbg/heap/ptmalloc.py
+++ b/pwndbg/heap/ptmalloc.py
@@ -2,7 +2,6 @@ import importlib
 from collections import OrderedDict
 from enum import Enum
 from functools import wraps
-from typing import Optional
 
 import gdb
 
@@ -34,7 +33,7 @@ class BinType(str, Enum):
     UNSORTED = 'unsortedbin'
     NOT_IN_BIN = 'not_in_bin'
 
-    def valid_fields(self) -> list[str]:
+    def valid_fields(self):
         if self in [BinType.FAST, BinType.TCACHE]:
             return ['fd']
         elif self in [BinType.SMALL, BinType.UNSORTED]:
@@ -43,23 +42,17 @@ class BinType(str, Enum):
             return ['fd', 'bk', 'fd_nextsize', 'bk_nextsize']
 
 class Bin:
-    def __init__(
-        self,
-        fd_chain: list[int],
-        bk_chain=None,
-        count=None,
-        is_corrupted=False
-    ):
+    def __init__(self, fd_chain, bk_chain=None, count=None, is_corrupted=False):
         self.fd_chain = fd_chain
         self.bk_chain = bk_chain
         self.count = count
         self.is_corrupted = is_corrupted
 
-    def contains_chunk(self, chunk: int) -> bool:
+    def contains_chunk(self, chunk):
         return chunk in self.fd_chain
 
     @staticmethod
-    def size_to_display_name(size: int | str) -> str:
+    def size_to_display_name(size):
         if size == 'all':
             return size
 
@@ -69,13 +62,13 @@ class Bin:
 
 
 class Bins:
-    def __init__(self, bin_type: BinType):
-        self.bins: OrderedDict[int | str, Bin] = OrderedDict()
+    def __init__(self, bin_type):
+        self.bins     = OrderedDict()
         self.bin_type = bin_type
 
     # TODO: There's a bunch of bin-specific logic in here, maybe we should
     # subclass and put that logic in there
-    def contains_chunk(self, size, chunk: int) -> bool:
+    def contains_chunk(self, size, chunk):
         # TODO: It will be the same thing, but it would be better if we used
         # pwndbg.heap.current.size_sz. I think each bin should already have a
         # reference to the allocator and shouldn't need to access the `current`
@@ -110,7 +103,7 @@ class Bins:
         return False
 
 
-def read_chunk_from_gdb(addr: int):
+def read_chunk_from_gdb(addr):
     """Read a chunk's metadata."""
     # In GLIBC versions <= 2.24 the `mchunk_[prev_]size` field was named `[prev_]size`.
     # To support both versions, change the new names to the old ones here so that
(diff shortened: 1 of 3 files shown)


gsingh93:
cc: @disconnect3d

I also see that @lebr0nli has a non-trivial sized heap PR that hasn't been merged yet. It would be great if we could merge that so I could go ahead and rebase on top of it, and if we could hold off or coordinate on any major changes until we get this merged.

disconnect3d:
Hey,

+1; I will merge the other heap PR tomorrow evening, as I dont have GH
access rn

gsingh93:
This is essentially ready to review. I just need to fix some tests and potentially add more, but I think people can start leaving comments.

@lebr0nli could you also take a look to make sure I didn't break any of your recent heap changes when I rebased on your changes?

lebr0nli:
> @lebr0nli could you also take a look to make sure I didn't break any of your recent heap changes when I rebased on your changes?

Sure!

The code looks quite clean to me, but after some testing and reviewing, I found some issues:

* type annotations

What Python version are you using now?
The Python version for my GDB is 3.8, and I need to `from __future__ import annotations` to make it works.
So I think maybe we should add this line at the beginning of these two files?

* heuristic

Some of your code doesn't work for heuristic because you are implementing them in `DebugSymsHeap` class instead of its parent class, `Heap` class.
Is it possible to move those methods, for example, `chunks`, into the `Heap` class?
You can reproduce this issue by debugging any process that will use heap:
```
pwndbg> set resolve-heap-via-heuristic
pwndbg> heap
  File "/pwndbg/pwndbg/commands/__init__.py", line 129, in __call__
  File "/pwndbg/pwndbg/commands/__init__.py", line 223, in _OnlyWhenRunning
    return function(*a, **kw)
  File "/pwndbg/pwndbg/commands/__init__.py", line 261, in _OnlyWithResolvedHeapSyms
    return function(*a, **kw)
  File "/pwndbg/pwndbg/commands/__init__.py", line 241, in _OnlyWhenHeapIsInitialized
    return function(*a, **kw)
  File "/pwndbg/pwndbg/commands/heap.py", line 127, in heap
    for chunk in allocator.chunks(cursor):
AttributeError: 'HeuristicHeap' object has no attribute 'chunks'

PS: Pull requests are welcome
```

gsingh93:
Ah, thanks, I wasn't testing with `set resolve-heap-via-heuristic`, I'll try that out and fix the issues. Will also add that in some unit tests with this setting enabled.

My Python version in GDB seems to be 3.10.4:
```
pwndbg> pi sys.version
'3.10.4 (main, Jun 29 2022, 12:14:53) [GCC 11.2.0]'
```

My guess is that the issue is related to this, but I'm not sure: https://stackoverflow.com/questions/61544854/from-future-import-annotations

I'll add the import to fix this on older versions, but @disconnect3d, which versions of GDB/Python do we run the CI tests under, and do we keep around older versions to make sure things like this don't break?

pwndbg/heap/ptmalloc.py
```diff
@@ -288,11 +391,46 @@ def tcache_next_offset(self):
...
+            arena = self.get_arena_for_chunk(heap_start)
+        else:
+            arena = self.get_arena()
+
+        ptr_size = self.size_sz
+
+        start_addr = heap_start
+        if arena != self.main_arena:
+            start_addr += self.heap_info.sizeof
+            if pwndbg.vmmap.find(self.get_heap(heap_start)['ar_ptr']) == heap_region:
```
lebr0nli, line 400: I think maybe we should add something like: `heap_region = self.get_heap_boundaries(heap_start)` before line 406 to avoid `NameError`?

disconnect3d: What NameError can we get here?

disconnect3d:
> I'll add the import to fix this on older versions, but @disconnect3d, which versions of GDB/Python do we run the CI tests under, and do we keep around older versions to make sure things like this don't break?

For CI, you should be able to see this in the details of the build, its Py 3.8:
image

If you can't access this information, let me know (I think everyone should be able to, but maybe there is some permissions issue, heh).

Regarding older versions: I would personally move to the latest possible as I can use them myself. However, given that there are likely some people that keep using old versions of GDB or Python, I would prefer to support:
* GDB >= 8.0
* Python >= 3.4 or 3.5

Given this, I think we should drop the direct typing stuff. Though, I am wondering, can this be introduced in functions docstring? I think there was some option for this back then, not sure how this works with later Python versions.

pwndbg/heap/ptmalloc.py
lebr0nli: Ah, sorry I didn't express myself clearly.
I mean, name `heap_region` is not defined here?

gsingh93:
@disconnect3d @lebr0nli this should be ready for a final review.

I've stripped the types for now, I'll add them back in later when I have some more confidence it won't break older versions of Python.

@lebr0nli, I did test with `set resolve-heap-via-heuristic`, but could you also verify everything still works?

lebr0nli:
> @lebr0nli, I did test with `set resolve-heap-via-heuristic`, but could you also verify everything still works?

LGTM! Heap heuristics can work without the previous issues now!

Thanks 🙏

disconnect3d (approved):

PR #1076, opened 2022-08-18
CI: run tests on Ubuntu 22.04, 20.04 and 18.04

disconnect3d:
Let's see if we can get CI running on other Ubuntu versions :)

AFAIU this should work, based on:
* https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs
* https://github.com/actions/runner-images#available-images

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,7 +5,7 @@ jobs:
   tests:
     strategy:
       matrix:
-        os: [ubuntu-latest]
+        os: [ubuntu-22.04, ubuntu-20.04, ubuntu-18.04]
     runs-on: ${{ matrix.os }}
     timeout-minutes: 10
     steps:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,6 +4,7 @@ on: [push, pull_request]
 jobs:
   tests:
     strategy:
+      fail-fast: false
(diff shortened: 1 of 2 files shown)


disconnect3d:
Apart from test deps failing on Ubuntu 18.04 (ugh...) does anyone know why other builds are getting "Cancelled"? That's weird.

disconnect3d:
That is probably it: https://fann.im/blog/2019/11/27/github-actions-canceled-unexpectedly/

gsingh93:
I've been getting the 22.04 test failure with `try_free` for a while locally, I guess it would be related to a glibc upgrade. Honestly I'd recommend that if you don't immediately know the fix, it's probably better to just file an issue to fix that test and disable it on 22.04, so that we can get test coverage for everything else in.

The 18.04 failure seems to be related to differences in how unicorn outputs its disassembly. On 18.04, instead of '↓', just a blank line is printed, there are also some minor differences in spacing between the instruction address and the instruction. We could modify the test to iterate over each line and normalize '↓' to blank lines, as well as normalize all whitespace. Or we could just disable this test on 18.04, as the test is broken but the functionality is still correct, we already have test coverage on two newer versions, and 18.04 will be EOL in a year anyway.

disconnect3d:
Closing in favor of #1100

PR #1087, opened 2022-08-23
Fix heap test binaries build

disconnect3d:
It turns out we never build the heap test binaries with their expected build flags. Instead, we always build and used the binaries build by wildcard rules in makefile: `%.out: %.c`

Ugh... and now I think some tests on CI will fail because we may not be able to load the binaries or have their dependencies? Let's see.

### DIFF
--- a/setup-test-tools.sh
+++ b/setup-test-tools.sh
@@ -25,6 +25,8 @@ install_apt() {
     gcc \
     libc6-dev
   test -f /usr/bin/go || sudo apt-get install -y golang
+  # We use zig to compile some test binaries as it is much easier than with gcc
+  sudo snap install zig --classic --edge
 }
 
 if linux; then
--- a/tests/binaries/makefile
+++ b/tests/binaries/makefile
@@ -1,3 +1,5 @@
+ZIGCC		=	   zig cc
+
 CC              =	   gcc
 DEBUG           =	   1
 CFLAGS         +=	   -Wall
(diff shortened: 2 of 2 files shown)


disconnect3d:
Yeah..:
```
$ gdb /home/dc/pwndbg/tests/binaries/heap_bugs.out
pwndbg: loaded 200 commands. Type pwndbg [filter] for a list.
Reading symbols from /home/dc/pwndbg/tests/binaries/heap_bugs.out...
Use the canary command to see all stack canary/cookie values on the stack (based on the *usual* stack canary value initialized by glibc)
pwndbg> r
Starting program: /home/dc/pwndbg/tests/binaries/heap_bugs.out 
/bin/bash: line 1: /home/dc/pwndbg/tests/binaries/heap_bugs.out: No such file or directory
During startup program exited with code 127.
pwndbg> 
```

disconnect3d:
And here's my local reason why it crashes:
```
user@user:~/pwndbg/tests/binaries$ /home/user/pwndbg/tests/binaries/heap_bins.out
/home/user/pwndbg/tests/binaries/heap_bins.out: /home/user/pwndbg/tests/binaries/glibcs/2.33/libc.so.6: version `GLIBC_2.34' not found (required by /home/user/pwndbg/tests/binaries/heap_bins.out)
```

Apparently gcc/clang sucks? I mean, u can't force libc symbol version during compilation.

Or.. at least I have tried. I have also tried patchelf with clear symbol versions etc. as suggested in https://stackoverflow.com/questions/2856438/how-can-i-link-to-a-specific-glibc-version

But it still does not work fully as some other symbols (in other sections) are left there with the newer (2.34) glibc version.

So what can we do? Apparently if we compile with ZIG's programming language compiler, it works? O_o

I mean, sth like this: `zig cc <rpath flags> <dynamic interp flag> -target native-native-gnu.2.33 -lc test.c`. Adding this in next commit.

PR #1095, opened 2022-08-23
feature: Add highlighting of freed chunks in vis_heap_chunks

alufers:
Adds highlighting to known-free chunks in vis_heap_chunks,
unfortunately not all freed chunks will be highlighted, only those marked by the Heap class, which by default is limited to 8 chunks per bin type.

image

### DIFF
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -10,9 +10,10 @@ import pwndbg.commands
 import pwndbg.config
 import pwndbg.glibc
 import pwndbg.typeinfo
+from pwndbg.color import NORMAL
 from pwndbg.color import generateColorFunction
 from pwndbg.color import message
-from pwndbg.color import underline, NORMAL
+from pwndbg.color import underline
 from pwndbg.commands.config import extend_value_with_default
 from pwndbg.commands.config import get_config_parameters
 from pwndbg.commands.config import print_row
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -12,6 +12,7 @@ import pwndbg.glibc
 import pwndbg.typeinfo
 from pwndbg.color import generateColorFunction
 from pwndbg.color import message
+from pwndbg.color import underline, NORMAL
 from pwndbg.commands.config import extend_value_with_default
 from pwndbg.commands.config import get_config_parameters
 from pwndbg.commands.config import print_row
@@ -599,7 +600,6 @@ def vis_heap_chunks(addr=None, count=None, naive=None):
         cursor += real_size
 
     # Build the output buffer, changing color at each chunk delimiter.
-    # TODO: maybe print free chunks in bold or underlined
     color_funcs = [
         generateColorFunction("yellow"),
         generateColorFunction("cyan"),
@@ -619,6 +619,10 @@ def vis_heap_chunks(addr=None, count=None, naive=None):
         # it's difficult (impossible?) to find all the thread caches for a
         # specific heap.
         bin_collections.insert(0, allocator.tcachebins(None))
+    bin_collections.insert(0, allocator.fastbins(None))
+    bin_collections.insert(0, allocator.unsortedbin(None))
+    bin_collections.insert(0, allocator.smallbins(None))
+    bin_collections.insert(0, allocator.largebins(None))
 
     printed = 0
     out = ''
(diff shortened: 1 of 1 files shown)


pwndbg/commands/heap.py
```diff
@@ -619,6 +620,10 @@ def vis_heap_chunks(addr=None, count=None, naive=None):
         # it's difficult (impossible?) to find all the thread caches for a
         # specific heap.
         bin_collections.insert(0, allocator.tcachebins(None))
+    bin_collections.insert(0, allocator.fastbins(None))
```
gsingh93, line 623: Can we just put this in the `bins_collection` array above?

pwndbg/commands/heap.py
```diff
@@ -629,28 +634,45 @@ def vis_heap_chunks(addr=None, count=None, naive=None):
...
+                out_for_chunk += '\t' + color_func(asc) + ('\t <-- ' + ', '.join(labels) if len(labels) else '')
                 asc = ''
+                if len(labels) > 0:
+                    was_in_bin = True
+                
                 labels = []
 
             cursor += ptr_size
+        if was_in_bin:
+            out_for_chunk =  underline(out_for_chunk) + NORMAL
```
gsingh93, line 674: `NORMAL` should already be added when calling `underline`, are you not seeing that happen in the output?

pwndbg/commands/heap.py
```diff
@@ -629,28 +634,45 @@ def vis_heap_chunks(addr=None, count=None, naive=None):
...
             cell_hex = '\t0x{:0{n}x}'.format(cell, n=ptr_size*2)
 
-            out += color_func(cell_hex)
-            printed += 1
+            
 
+            prev_len_of_labels = len(labels)
             labels.extend(bin_labels(cursor, bin_collections))
+            # an address was in a bin when a label was added
+            if  len(labels) != prev_len_of_labels:
```
gsingh93, line 652: nitpick: there's an extra space after `if`

pwndbg/commands/heap.py
CptGibbon: Why add these at all?
They cause free chunk labels to be printed twice, as shown in your screenshot.

pwndbg/commands/heap.py
CptGibbon: There are a few rogue spaces kicking about, not a big deal but a linter could help avoid them in future.

CptGibbon: Also what does this line do?

pwndbg/commands/heap.py
```diff
@@ -629,28 +633,45 @@ def vis_heap_chunks(addr=None, count=None, naive=None):
...
+            out_for_chunk += color_func(cell_hex)
+            
+            printed += 1
 
             asc += bin_ascii(pwndbg.memory.read(cursor, ptr_size))
             if printed % 2 == 0:
-                out += '\t' + color_func(asc) + ('\t <-- ' + ', '.join(labels) if len(labels) else '')
+                out_for_chunk += '\t' + color_func(asc) + ('\t <-- ' + ', '.join(labels) if len(labels) else '')
                 asc = ''
+                if len(labels) > 0:
```
CptGibbon, line 667: A creative way to check if a chunk is free, but has a couple of side-effects:
  * Highlighting is applied to the top chunk (I guess it's technically free...)
  * Highlighting is applied to labels

disconnect3d:
So we had this idea to underline freed chunks but it turns out it is not so easy to figure all of this.

I mean, sure, this information is literally in heap tcache/fast/small/large bins (and I guess in the unordered bin or whatever it was called) -- but -- if the program has lots and lots of those, fetching it requires going through all of the linked lists.

So the question is, should we do this? And if so, how should we make it so that the output is clear for the user?

I guess we could do one of those:
1. Not do this at all
2. Have a flag like `--mark-freed-up-to=N` with some sane default for `N`?
3. Then if we know our output is accurate - e.g. we visited all freed chunks - we can mark them and job's done
4. If we know the output is not accurate - that we haven't visited some freed chunks - we can tell the user "Hey: this underlining for freed chunks is not accurate: you may want to call it again with `--mark-freed-up-to=Y` for more accurate output?

WDYT?

CptGibbon:
> So the question is, should we do this? And if so, how should we make it so that the output is clear for the user?

I haven't heard anyone express a good reason for this feature, free chunks are already marked by labels.

disconnect3d:
Sure, but not all of them are marked: the logic uses some `.chain(...)` call which is limited by a config variable to 8 and so if you have more fastbins than 8, only first 8 will be marked by labels.

disconnect3d:
Maybe we should just mark it with labels and allow one to overwrite the limit in `vis_heap_chunk` and not care about the underline.

CptGibbon:
I agree, since this approach only underlines labelled chunks anyway it doesn't provide more information.

disconnect3d:
I will close this for now. We should consider making it better UX to inform user not all labels are applied and let them configure/request more labels via an optional flag argument

PR #1096, opened 2022-08-24
support riscv:rv64 without capstone

lonnywong:
Just make `riscv:rv64` works. Full support requires https://github.com/pwndbg/pwndbg/pull/829.

image

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -720,7 +720,8 @@ def context_backtrace(with_banner=True, target=sys.stdout, width=None):
     for i in range(backtrace_lines - 1):
         try:
             candidate = oldest_frame.older()
-        except gdb.MemoryError:
+        # We catch gdb.error in case of a `gdb.error: PC not saved` case
+        except (gdb.MemoryError, gdb.error):
             break
 
         if not candidate:
--- a/pwndbg/disasm/__init__.py
+++ b/pwndbg/disasm/__init__.py
@@ -117,8 +117,25 @@ def get_disassembler(pc):
     )
 
 
+class SimpleInstruction:
+    def __init__(self, address):
+        self.address = address
+        ins = gdb.newest_frame().architecture().disassemble(address)[0]
+        asm = ins["asm"].split(None, 1)
+        self.mnemonic = asm[0].strip()
+        self.op_str = asm[1].strip() if len(asm) > 1 else ""
+        self.size = ins["length"]
+        self.next = self.address + self.size
+        self.target = self.next
+        self.groups = []
+        self.symbol = None
+        self.condition = False
+
+
 @pwndbg.memoize.reset_on_cont
 def get_one_instruction(address):
+    if pwndbg.arch.current not in CapstoneArch:
+        return SimpleInstruction(address)
     md = get_disassembler(address)
     size = VariableInstructionSizeMax.get(pwndbg.arch.current, 4)
     data = pwndbg.memory.read(address, size, partial=True)
--- a/pwndbg/regs.py
+++ b/pwndbg/regs.py
@@ -473,6 +473,124 @@ mips = RegisterSet(
     retval="v0",
 )
 
+# https://riscv.org/technical/specifications/
+# Volume 1, Unprivileged Spec v. 20191213
+# Chapter 25 - RISC-V Assembly Programmer’s Handbook
+# x0        => zero   (Hard-wired zero)
+# x1        => ra     (Return address)
+# x2        => sp     (Stack pointer)
+# x3        => gp     (Global pointer)
+# x4        => tp     (Thread pointer)
+# x5        => t0     (Temporary/alternate link register)
+# x6–7      => t1–2   (Temporaries)
+# x8        => s0/fp  (Saved register/frame pointer)
+# x9        => s1     (Saved register)
+# x10-11    => a0–1   (Function arguments/return values)
+# x12–17    => a2–7   (Function arguments)
+# x18–27    => s2–11  (Saved registers)
+# x28–31    => t3–6   (Temporaries)
+# f0–7      => ft0–7  (FP temporaries)
+# f8–9      => fs0–1  (FP saved registers)
+# f10–11    => fa0–1  (FP arguments/return values)
+# f12–17    => fa2–7  (FP arguments)
+# f18–27    => fs2–11 (FP saved registers)
+# f28–31    => ft8–11 (FP temporaries)
+riscv = RegisterSet(
+    pc="pc",
+    frame="fp",
+    stack="sp",
+    retaddr="ra",
+    gpr=(
+        "ra",
+        "gp",
+        "tp",
+        "t0",
+        "t1",
+        "t2",
+        "fp",
+        "s1",
+        "a0",
+        "a1",
+        "a2",
+        "a3",
+        "a4",
+        "a5",
+        "a6",
+        "a7",
+        "s2",
+        "s3",
+        "s4",
+        "s5",
+        "s6",
+        "s7",
+        "s8",
+        "s9",
+        "s10",
+        "s11",
+        "t3",
+        "t4",
+        "t5",
+        "t6",
+        "ft0",
+        "ft1",
+        "ft2",
+        "ft3",
+        "ft4",
+        "ft5",
+        "ft6",
+        "ft7",
+        "fs0",
+        "fs1",
+        "fa0",
+        "fa1",
+        "fa2",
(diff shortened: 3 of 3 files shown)


pwndbg/commands/context.py
```diff
@@ -580,7 +580,7 @@ def context_backtrace(with_banner=True, target=sys.stdout, width=None):
     for i in range(backtrace_lines-1):
         try:
             candidate = oldest_frame.older()
-        except gdb.MemoryError:
```
disconnect3d, line 723: Out of curiosity, what's the gdb.error that can happen here? Any chance we can commit this separately to persist the reason of this or at least add a comment why this or the other error?

pwndbg/disasm/__init__.py
```diff
@@ -119,8 +119,31 @@ def get_disassembler(pc):
                                    pwndbg.arch.endian,
                                    extra)
 
+class SimpleInstruction(object):
```
disconnect3d: No need to inherit from `object` in Python 3
```suggestion
class SimpleInstruction:
```

pwndbg/disasm/__init__.py
```diff
@@ -119,8 +119,31 @@ def get_disassembler(pc):
...
+    def __init__(self, address):
+        self.address = address
+        instr = gdb.execute('x/2i %#x' % self.address, to_string=True)
```
disconnect3d: Can we use GDB Python API to disassemble instructions instead of gdb.execute that invokes eXamine command with instructions format?

See: https://sourceware.org/gdb/onlinedocs/gdb/Architectures-In-Python.html#Architectures-In-Python:~:text=Function%3A%20Architecture.disassemble

disconnect3d: It may help us to avoid all the weird strip/split/string formatting stuff

Tho I guess x/i may give us more information than pure disassembly? I am not sure, happy to hear about it, but maybe we can do it in a nicer way with proper APIs

disconnect3d (commented):
Hey,

Great PR, thanks for working on this.

I think ideally eventually we want to be able to switch between Capstone vs GDB's disassembler and make Capstone optional (but install it by default).

Also, using GDB's disassembly probably allows us to display instructions before current instruction in the disasm context --- at least that is what GDB does by default and we never did it with Capstone.

AFAIK GDB does this by finding nearby previous symbol and then disassembling from this, so that it does not corrupt/disassemble invalid instructions before current PC/IP address.

disconnect3d:
> I think ideally eventually we want to be able to switch between Capstone vs GDB's disassembler and make Capstone optional (but install it by default).

But I am fine with merging whatever we will come up in this PR just to make it work with risc and other archs not supported currently.

But eventually, I'd love if we had a possibility to set it up globally etc.

pwndbg/disasm/__init__.py
lonnywong: I didn't know the disassemble API before, I'll try to use it. Thanks.

lonnywong:
> I think ideally eventually we want to be able to switch between Capstone vs GDB's disassembler and make Capstone optional (but install it by default).

I'm not quite sure how to do it. But I will finish the PR, the rest is up to you.

I appreciate your support very much.

pwndbg/disasm/__init__.py
lonnywong: OK, I'll fix it.

pwndbg/commands/context.py
lonnywong: I test it on `Ubuntu 22.04`.
```sh
apt install qemu-user
wget "https://github.com/M0ngi/CTF-Writeups/raw/main/2022/Hacker's%20Playground%202022/sources/riscy/deploy/target"
chmod +x target
python3 -c 'import sys; sys.stdout.buffer.write(b"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAV\x22\x04\x00\x00\x00\x00\x00AAAAAAAAAAAAAAAAAAAAAAAA\xcc\xdd\x01\x00\x00\x00\x00\x00AAAAAAAA\x82\x17\x04\x00\x00\x00\x00\x00aaaabaaa\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x08\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00iaaajaaakaaalaaamaaanaaaoaaapaaa?\x00\x00\x00\x00\x00\x00\x004\x04\x01\x00\x00\x00\x00\x00")' | qemu-riscv64 -g 9000 ./target
gdb-multiarch -ex 'set exception-verbose on' -ex 'target remote localhost:9000' -ex 'b *0x42256' -ex 'c' ./target
```
I got the error:
```
  File "/home/x/pwndbg/pwndbg/commands/__init__.py", line 129, in __call__
  File "/home/x/pwndbg/pwndbg/commands/__init__.py", line 223, in _OnlyWhenRunning
    return function(*a, **kw)
  File "/home/x/pwndbg/pwndbg/commands/context.py", line 272, in context
    result[target].extend(func(target=out,
  File "/home/x/pwndbg/pwndbg/commands/context.py", line 582, in context_backtrace
    candidate = oldest_frame.older()
gdb.error: PC not saved
```

lonnywong: @disconnect3d What should we do with it?

lonnywong:
Btw, the condition for the next instruction is not working.

pwndbg/commands/context.py
disconnect3d: Hm I'm pretty sure I was responding here already but I guess my comment was not send.

We should just add a comment about it like this:
```
# We catch gdb.error in case of a `gdb.error: PC not saved` case
except (gdb.MemoryError, gdb.error):
```

disconnect3d:
> Btw, the condition of the next instruction is not working.

Do you mean we skip using it/displaying some information or is there some exception being thrown?

Btw any chance we can add a test to the CI with this somehow? Like, we could build a riscv:rv64 binary from assembly and debug it via qemu user emulation gdbserver etc and assert that we display context properly? If u don't want/have time to write the test I can probably deal with this in ~2 weeks. (I am fine merging this PR without a test)

pwndbg/commands/context.py
lonnywong: @disconnect3d Thanks for pointing it out directly.

lonnywong:
Should be `the condition for the next instruction`.

For example, suppose  the current instruction is `je next`, we need to check the status flags in the EFLAGS register. 

We need to implement these ( It's too complicated for me ) :
```py
self.next = self.address + self.size
self.target = self.next
self.groups = []
self.symbol = None
self.condition = False
```

I am not familiar with the test case, and I won't be free until October. I appreciate you write the test in about 2 weeks.

@disconnect3d Thanks for you help.

disconnect3d:
@lonnywong I mean, I am fine with not supporting the branch instructions target detection when we don't have capstone.

With capstone and unicorn its +- easy and here its probably more work.

My question is just if the current PR works as-is or does it need more work as maybe it crashes on such branch instructions or something?

Can you show a screenshot/code listing of how pwndbg works when debugging risc-v with those changes?

disconnect3d:
Btw we merged some lint refactorings after adding black and this created some conflicts :(

lonnywong:
> My question is just if the current PR works as-is or does it need more work as maybe it crashes on such branch instructions or something?

It doesn't crash, just the next instruction may be displayed incorrectly on branch. It's not a big deal.

> Can you show a screenshot/code listing of how pwndbg works when debugging risc-v with those changes?

```sh
wget "https://github.com/M0ngi/CTF-Writeups/raw/main/2022/Hacker's%20Playground%202022/sources/riscy/deploy/target"
chmod +x target
python3 -c 'import sys; sys.stdout.buffer.write(b"A"*0x28+b"\x56\x22\x04\x00\x00\x00\x00\x00")' | qemu-riscv64 -g 9000 ./target

# another terminal
gdb-multiarch ./target
pwndbg> target remote localhost:9000
pwndbg> ni
pwndbg> vmmap
pwndbg> b *0x42256
pwndbg> c
pwndbg> si
```

> Btw we merged some lint refactorings after adding black and this created some conflicts :(

Resolved.

disconnect3d:
> It doesn't crash, just the next instruction may be displayed incorrectly on branch.

You mean it will just be linear disassembly, right? I assume yes and so merged it. Then this is fine!

Thanks!

lonnywong:
>You mean it will just be linear disassembly, right?

Yes, indeed.

PR #1099, opened 2022-08-25
Minor test cleanup

gsingh93:
Changes in this PR:
- Silences a warning when building the `heap_vis.c` test binary. I can alternatively just fix the warning if that's preferred, but I don't think it matters in this case.
- When I check the value in `/proc/sys/kernel/yama/ptrace_scope` on my Ubuntu 22.04 machine, I see that the value is '0\n'. The logic for skipping tests was only checking for '0'. This is fixed now.
- Makes the .gitignore more flexible for where the test binaries are stored, as there may be some changes to this in the future when we add a new folder for unit tests
- Cleans up some small things in tests.sh: formatting, setting the $ROOT_DIR variable to allow for easily changing this in the future if the location of the script changes, and putting the call to GDB in a function so we can reuse most of the logic.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,7 @@ __pycache__/
 
 # C extensions
 *.so
-!tests/binaries/glibcs/**/*.so
+!tests/**/binaries/glibcs/**/*.so
 
 # Distribution / packaging
 .Python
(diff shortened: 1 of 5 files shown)


gsingh93:
There seems to be some shell quoting issue with passing the GDB arguments. Will fix this tomorrow.

disconnect3d:
> on my Ubuntu 22.04 machine, I see that the value is '0\n'. The logic for skipping tests was only checking for '0'. This is fixed now.

Omg, good catch! :D

gsingh93:
Fixed the issue with coverage generation.

PR #1100, opened 2022-08-25
Update CI to test 18.04, 20.04, and 22.04

gsingh93:
After investigating the test failures in https://github.com/pwndbg/pwndbg/pull/1076, I figured out a way to install different dependencies in Python versions lower than 3.7 and greater than or equal to 3.7, which fixed those test failures. However, the `vis_heap_chunk` test was recently introduced, which is now breaking on 18.04.

I also added a `dev-requirements.txt` with the `coverage` package and install it from `setup-test-tools.sh`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,10 +13,13 @@ pycparser==2.21
 pyelftools==0.28
 Pygments==2.12.0
 pyparsing==3.0.9
-pytest==7.1.2
+pytest==7.1.2; python_version >= '3.7'
+pytest==7.0.1; python_version < '3.7'
 python-ptrace==0.9.8
 ROPGadget==6.8
 six==1.16.0
 testresources==2.0.1
-tomli==2.0.1
-unicorn==2.0.0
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -13,10 +13,13 @@ pycparser==2.21
 pyelftools==0.28
 Pygments==2.12.0
 pyparsing==3.0.9
-pytest==7.1.2
+pytest==7.1.2; python_version >= '3.7'
+pytest==7.0.1; python_version < '3.7'
```
disconnect3d, line 17: We should move pytest to dev-requirements too

disconnect3d: Moved it in https://github.com/pwndbg/pwndbg/commit/697c32e84bb8a36dd10ab32936dcec8011d47994 after merging this PR :)

gsingh93: Ah, I was in the process of fixing this, and I saw `isort` should probably be moved too. Not sure about `iniconfig` and `packaging`, they seem like dev related tools but I'm not sure.

PR #1115, opened 2022-09-04
Add dev-requirements.txt to Docker image

gsingh93:
Building the Docker image currently fails because `./setup.sh` references `dev-requirements.txt`, but this was never added to the repo.

This PR fixes the issue with the least amount of modification, but I'm wondering if we can just move the `ADD . /pwndbg` higher up so that everything in the repo is copied over and we don't have to worry about adding any new files that are needed for setup? Why are we copying individual setup files over in the first place?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,6 +26,7 @@ RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && \
 
 ADD ./setup.sh /pwndbg/
 ADD ./requirements.txt /pwndbg/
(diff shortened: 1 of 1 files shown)


disconnect3d:
I'm not sure about `ADD . ./pwndbg`. I thought people mostly wanted to use volumes to mount the pwndbg sources into the container to be able to develop it... cc: @artcz @hbrylkowski

gsingh93:
If we do it, won't mounting volumes over it still work fine?

disconnect3d:
Yeah. I think we can just move it.

PR #1118, opened 2022-09-05
GDB Refactor [2/N]: move some modules to lib/

gsingh93:
`version.py`, `which.py`, `functions.py`, `tips.py`, and `funcparser.py` already didn't have any dependency on GDB, so these could just be copied to `lib/`. Also fixed the `.gitignore` to not exclude `pwndbg/lib/`.


syheliel:
What is `pwndbg/lib/functions.py` used for? It's too large.

gsingh93:
I have no clue, for now my goal is just to move things into the right directories without breaking things. Once that's done, the next step will be to write unit tests, add types, refactor poorly designed code, etc.

pwndbg/lib/version.py
```diff
@@ -20,7 +20,7 @@ def build_id():
         return ""
 
 
-__version__ = "1.1.1"
+__version__ = "1.1.0"
```
disconnect3d:
```suggestion
__version__ = "1.1.1"
```

pwndbg/lib/tips.py
```diff
@@ -0,0 +1,28 @@
+from random import choice
```
disconnect3d, line 1: shouldnt this file be moved instead of added? weird

disconnect3d:
@syheliel 
> What is pwndbg/lib/functions.py used for? It's too large.

It is used to enhance display of function arguments of known functions; I think it may have been generated from manpages or some other source

syheliel:
It uses `PyCParser` and generated by `pwndbg.lib.funcparser.ExtractFuncDeclFromSource`

PR #1127, opened 2022-09-06
add color for modified registers

### LINKED ISSUE #1082: Better marking of modified registers
Currently, if a register was modified between two stops, we show it `context regs` by prefixing it with an asterisks.

I have seen that gef does this by marking the register as red, and showing this in its legend:
image

Perhaps we should also mark the register with some color? :)

syheliel:
Fix #1082


arcz:
This is how it looks now, I think it's a nice improvement.
Screenshot from 2022-09-13 11-55-11

disconnect3d:
@syheliel we should make that color configurable and displayed with the `theme` command I think. It should also probably color the star symbol?

disconnect3d:
Thanks!

PR #1130, opened 2022-09-06
GDB Refactor [10/N]: Move qemu.py and remote.py to gdblib

gsingh93:


gsingh93:
It'd be really good to have a simple test that loads up QEMU before merging this. Any volunteers?

PR #1131, opened 2022-09-06
Fix incorrect import

gsingh93:


gsingh93:
@disconnect3d Do you know how to reach this code? Seems none of our tests reach it as it wasn't caught earlier.

PR #1134, opened 2022-09-08
GDB Refactor [14/N]: Split config.py into lib/ and gdblib/

gsingh93:
This PR splits `config.py` into `lib/config.py` and `gdblib/config.py`. Support for Python 2 is removed, which significantly simplifies the code.

The `pwndbg.lib.config.Config` object now holds all of the `pwndbg.lib.config.Parameters` that are defined, as well as the triggers for these parameters. This does not depend on GDB, and can be reused in the future if other debuggers are supported (i.e lldb). Instead of creating new parameters with the constructor of `Parameter`, you need to use `Config.add_param`, and triggers can be created with `Config.trigger`.

There is an instance of a `Config` object instantiated in `pwndbg.gdblib.config`. This is the config object that all of the code will be accessing.

Because we need to trigger certain functions when some parameters are `set` in GDB, we need to create a subclass of `gdb.Parameter` and implement `get_set_string` to call the trigger functions. At the end of `__init__.py`, we call the `pwndbg.gdblib.config.init_params` function, which takes all the `Parameters` from the `Config` object and creates instances of this GDB class.


gsingh93:
@disconnect3d This is ready for review now. The PR is large, but the main things to look at are `lib/config.py` and `gdblib/config.py`. If you could also test this locally, that would be great.

pwndbg/color/theme.py
```diff
@@ -1,11 +1,9 @@
...
-        super(Parameter, self).__init__(name, default, docstring, "theme")
+def add_param(name, default, docstring):
+    return pwndbg.gdblib.config.add_param(name, default, docstring, "theme")
 
 
-class ColoredParameter(Parameter):
-    def __init__(self, name, default, docstring):
-        super(ColoredParameter, self).__init__(name, default, docstring)
+def add_color_param(name, default, docstring):
+    return add_param(name, default, docstring)
```
disconnect3d, line 9: Instead of having `pwndbg.color.theme.add_*` wouldn't it be better to have `pwndbg.gdblib.config.add_theme_param`?

pwndbg/gdblib/config.py
```diff
@@ -0,0 +1,101 @@
+"""
+Dynamic configuration system for pwndbg, using GDB's built-in Parameter
+mechanism.
+
+To create a new pwndbg configuration point, create an instance of
+``pwndbg.lib.config.Parameter``.
```
disconnect3d: This docstring is outdated

pwndbg/gdblib/config.py
```diff
@@ -0,0 +1,101 @@
...
+
+import gdb
+
+import pwndbg.decorators
+import pwndbg.lib.config
+
+config = pwndbg.lib.config.Config()
+
+# TODO: I don't think this needs to be an OrderedDict
+PARAM_CLASSES = collections.OrderedDict()
```
disconnect3d: Seems this can even be just:

```suggestion
PARAM_CLASSES = {
    bool: gdb.PARAM_BOOLEAN,
    int: gdb.PARAM_ZINTEGER,
    str: gdb.PARAM_STRING
}
```

pwndbg/gdblib/config.py
```diff
@@ -0,0 +1,101 @@
...
+# https://sourceware.org/gdb/onlinedocs/gdb/Parameters-In-Python.html
+class Parameter(gdb.Parameter):
+    def __init__(self, param: pwndbg.lib.config.Parameter):
+        self.param = param
+
+        # `set_doc` and `show_doc` must be set before `gdb.Parameter.__init__`.
+        # They will be used for `help set <param>` and `help show <param>`,
+        # respectively
+        self.set_doc = "Set " + self.param.docstring
+        self.show_doc = "Show " + self.param.docstring
```
disconnect3d, line 46: Just noting here: we should probably have a way to overwrite set/show docs for some of the config values

pwndbg/gdblib/config.py
```diff
@@ -0,0 +1,101 @@
...
+        super(Parameter, self).__init__(self.param.name, gdb.COMMAND_SUPPORT, self._param_class())
+
+    def _param_class(self):
+        for k, v in PARAM_CLASSES.items():
```
disconnect3d, line 50: Honestly, if there are just 3 types, we should just do if elif elif in `__init__` for this ;)

pwndbg/gdblib/config.py
```diff
@@ -0,0 +1,101 @@
...
+class Parameter(gdb.Parameter):
+    def __init__(self, param: pwndbg.lib.config.Parameter):
+        self.param = param
+
+        # `set_doc` and `show_doc` must be set before `gdb.Parameter.__init__`.
+        # They will be used for `help set <param>` and `help show <param>`,
+        # respectively
+        self.set_doc = "Set " + self.param.docstring
+        self.show_doc = "Show " + self.param.docstring
+        super(Parameter, self).__init__(self.param.name, gdb.COMMAND_SUPPORT, self._param_class())
```
disconnect3d: In Py3 this should just work:
```suggestion
        super().__init__(self.param.name, gdb.COMMAND_SUPPORT, self._param_class())
```

disconnect3d:
Added some comments.

Btw there are some things that bothers me with all the refactoring we are doing:
* long namespaces, aka: `pwndbg.gdblib.config.syntax_highlight` is long to read/write ;p
* duplicated sub-namespaces: `gdblib.config` vs `lib.config` may be confusing for new devs/users
* the fact that we usually access names via absolute attributes instead of using sth like `from pwndbg.gdblib import config`

I am not sure what to do with duplicated namespaces, but for long namespaces perhaps we should move `pwndbg.gdblib` to just `gdblib`?

pwndbg/gdblib/config.py
gsingh93: We're passing in the docstring when we define the parameters, so when would we want to overwrite that?

pwndbg/gdblib/config.py
gsingh93: Yea, I just couldn't figure out why the original code used `OrderedDict`, and was worried about breaking something by changing it to a normal `dict`. But we can do it and see what happens.

pwndbg/color/theme.py
gsingh93: I think you're probably right, but I've been avoiding refactoring `pwndbg/color` as long as I can as there are some decisions there I need to think about before doing it. So I can probably revisit this when I get around to that refactor.

gsingh93:
> long namespaces, aka: pwndbg.gdblib.config.syntax_highlight is long to read/write ;p

Yea, you actually brought this up before I even started all this refactoring work, and we couldn't figure out anything at the time. Maybe we can export `gdblib` and `lib` somehow so we can drop the `pwndbg` part?

> duplicated sub-namespaces: gdblib.config vs lib.config may be confusing for new devs/users

I think if I add some docs to `DEVELOPMENT.md` that could help with this. The idea is simple: `lib.*` should only contain concepts that are completely independent from GDB. You can have generic concepts like `Arch`, `Config`, etc. that implement all the details you need for those objects that don't depend at all on the debugger they're used for. `gdblib.*` on the other hand is only for things specific to GDB. So `gdblib.config` actually makes use of `lib.config` to represent a configuration, but it plugs that into GDB by adding the set/show strings for example.

We're pretty close to the point where we can finally add unit tests without too much pain, and once we do that I think it'll also be more clear, as we'll be able to do a lot more testing with the `lib.*` classes.

> the fact that we usually access names via absolute attributes instead of using sth like `from pwndbg.gdblib import config`

Can you elaborate on this one? Are you saying we normally do `import pwndb.gdblib.config` instead of `from pwndbg.gdblib import config`. I think it may be possible to switch to the former style of import, but Python imports are incredibly confusing, and it might just be me not understanding them. I can look into this once the refactor is closer to complete.

pwndbg/gdblib/config.py
disconnect3d: We would probably have to change the interface, or, give an option to either provide docstring or set+show strings?

My point is, we could display some more information in e.g. those parameters:
```
pwndbg> show resolve-heap-via-heuristic
Sets Resolve missing heap related symbols via heuristics (currently: False)

pwndbg> show context-sections
Sets which context sections are displayed (controls order) (currently: 'regs disasm code ghidra stack backtrace expressions')

pwndbg> show debug-events
Sets display internal event debugging info (currently: False)
```

Also, should we maybe mark that those are comming from pwndbg? Ah, I think we wanted to do it with namespaces :+1:

disconnect3d:
> Can you elaborate on this one? Are you saying we normally do import pwndb.gdblib.config instead of from pwndbg.gdblib import config. I think it may be possible to switch to the former style of import.

Yes. I think we can even switch to it now, if only we do it smart. I don't remember what were the exact reasons why we don't do it, but it could be that:
1. We may wanted to have an option to unload/load again or reload Pwndbg -- for that case, such imports may sometimes be better (?) I am not sure tbh.
2. It may be because some of our modules are magic and if you do something like: `from module import global` then the `global`  will not automagically update itself in the current module when it is updated in the module you import it from. Though, I am not sure if we do things like that. AFAIK we do something like "module properties" and this should probably work fine, since property is a function call anyway?

> but Python imports are incredibly confusing, and it might just be me not understanding them.
What exactly?

disconnect3d:
(btw conflicts :F)

pwndbg/gdblib/config.py
gsingh93: I left it as is. If I were to write this from scratch I'd probably also do the dictionary and the loop, as the bodies of the if-statements would be exactly the same.

gsingh93:
> but for long namespaces perhaps we should move pwndbg.gdblib to just gdblib?

If you want to give this a shot, I'm supportive of it. But there were so many import issues last time that I don't think I'll attempt it again myself. It may be easier now though, with all the cleanup work I've done, but still not sure about that.

Regarding the absolute imports and my issues with them, try changing the `from pwndbg.gdblib import config` `pwndbg/ui.py` from a relative import to an absolute one and then run the tests. You'll get something like `AttributeError: partially initialized module 'pwndbg' has no attribute 'commands' (most likely due to a circular import)`. I think we should leave this for now, once I'm completely finished with the refactor, there's a chance that these issues will be easier to fix.

disconnect3d:
> If you want to give this a shot, I'm supportive of it. But there were so many import issues last time that I don't think I'll attempt it again myself. It may be easier now though, with all the cleanup work I've done, but still not sure about that.

I will do this at some point, probably in next sprint. I will also want to do some benchmarks before and after such PR just to see how efficient or not Python is in those regards (or rather: does this really matter) ;)

PR #1145, opened 2022-09-13
Add `gdb.MemoryError` check to get_heap()

### LINKED ISSUE #1142: Find_Fake_fast (malloc_chunk: Cannot access memory at address)
I am using Ubuntu 22.04, pwndbg latest build. The problem I am facing right now is whenever I try to use "find_fake_fast" it prompts an error. 

img_1

Release 2022.01.05 and 2022.08.30 (https://github.com/pwndbg/pwndbg/releases/tag/2022.01.05) working fine 

img_2

CptGibbon:
The `find_fake_fast()` issue mentioned in #1142 occurs when `get_heap()` tries to read a `heap_info` struct from unmapped memory e.g. when trying to find an arena for a fake/corrupt chunk. The error isn't raised immediately because `gdb.Value`s are "lazy" and only resolve on use.

To address this, `get_heap()` now calls `fetch_lazy()` on the `gdb.Value` it acquires and wraps it in a `try`/`except` block, returning `None` if `gdb.MemoryError` is raised. `get_arena_for_chunk()` now handles a `None` returned by `get_heap()`.

Fixes #1142

Be aware that other commands such as `heap` and `vis_heap_chunks` suffer from the same problem, and will need more work to address this issue.

### DIFF
--- a/pwndbg/heap/ptmalloc.py
+++ b/pwndbg/heap/ptmalloc.py
@@ -307,7 +307,8 @@ class Heap(pwndbg.heap.heap.BaseHeap):
         chunk = pwndbg.commands.heap.read_chunk(addr)
         _, _, nm = self.chunk_flags(chunk["size"])
         if nm:
-            r = self.get_arena(arena_addr=self.get_heap(addr)["ar_ptr"])
+            h = self.get_heap(addr)
+            r = self.get_arena(h["ar_ptr"]) if h else None
         else:
             r = self.main_arena
         return r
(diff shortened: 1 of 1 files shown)


disconnect3d:
@CptGibbon can we add a regression test for this case?

Also, would it make sense to put only the `r.fetch_lazy()` into try/except block?

pwndbg/heap/ptmalloc.py
```diff
@@ -634,7 +635,13 @@ def malloc_par(self):
 
     def get_heap(self, addr):
         """Find & read the heap_info struct belonging to the chunk at 'addr'."""
```
disconnect3d, line 637: Can we extend the docs here to mention why we do .fetch_lazy / what's the potential issue here?

CptGibbon:
> would it make sense to put only the `r.fetch_lazy()` into try/except block?

I considered this, but it's unclear from the docs whether values are fetched lazily 100% of the time. Your call.

disconnect3d:
Ah, good point!

PR #1149, opened 2022-09-16
Improve vmmap on coredump files

disconnect3d:
With this commit we now recognize coredumps better and also finally have a simple test for vmmap commands on:
- a running binary
- on a loaded coredump file with loaded binary
- on a loaded coredump file without a loaded binary

We also stop saving vmmaps for `maintenance info sections` sections which have a start address of 0x0. While there could potentially be a coredump file from a binary with start=0x0, this should work in most cases.

We could in theory do a slighty better: we could take the vmmap at 0 and try to read memory from it. However, I am not sure if it is a good idea to try such memory read?


pwndbg/vmmap.py
```diff
@@ -243,6 +261,7 @@ def coredump_maps():
     vsyscall_page = pages[-1]
     if vsyscall_page.start > 0xFFFFFFFFFF000000 and vsyscall_page.flags & 1:
         vsyscall_page.objfile = "[vsyscall]"
+        vsyscall_page.offset = 0
```
disconnect3d, line 270: Not really needed since we hardcode offset above to 0, but idk. Those pages *really* have it as 0

PR #1157, opened 2022-09-22
Tests for procinfo command

ntsleep:


disconnect3d:
Merging as the lint issue is invalid - it has been already fixed on master but rebuilding here did not fix it since it seems gh do not want to try run this over rebased dev. Or... maybe it was a race condition that happened here, idk. But this commit seems to be working fine.

PR #1159, opened 2022-09-22
Add Chunk class & template_heap_command

CptGibbon:
The Chunk class abstracts developers from some heap & gdb internals and makes use of the gdb module's lazy fetching.
The `template_heap_command` command serves as an example of how to use the Chunk class (and hopefully future classes).

### DIFF
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -122,6 +122,16 @@ def template_heap_command(addr):
     if chunk.size is not None:
         print(f"chunk.size: 0x{chunk.size:02x}")
 
+    # Flags may be retrieved individually or as a dictionary of all 3.
+    if chunk.non_main_arena is not None:
+        print(f"chunk.non_main_arena: {chunk.non_main_arena}")
+
+    if chunk.is_mmapped is not None:
+        print(f"chunk.is_mmapped: {chunk.is_mmapped}")
+
+    if chunk.prev_inuse is not None:
+        print(f"chunk.prev_inuse: {chunk.prev_inuse}")
+
     if chunk.flags is not None:
         print(f"chunk.flags: {chunk.flags}")
 
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -96,7 +96,7 @@ def format_bin(bins, verbose=False, offset=None):
 
 
 parser = argparse.ArgumentParser()
-parser.description = ("Template heap command. You can ignore this.")
+parser.description = "Template heap command. You can ignore this."
 parser.add_argument("addr", type=int, help="Address of a chunk header.")
 
 
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -96,53 +96,6 @@ def format_bin(bins, verbose=False, offset=None):
 
 
 parser = argparse.ArgumentParser()
-parser.description = "Template heap command. You can ignore this."
-parser.add_argument("addr", type=int, help="Address of a chunk header.")
-
-
-@pwndbg.commands.ArgparsedCommand(parser)
-@pwndbg.commands.OnlyWhenRunning
-@pwndbg.commands.OnlyWithResolvedHeapSyms
-@pwndbg.commands.OnlyWhenHeapIsInitialized
-def template_heap_command(addr):
-    """Template heap command with example uses of pwndbg's heap classes."""
-
-    # `addr` is a gdb.Value of one of several types, depending on how the user invoked this command
-    # e.g. template_heap_command <address> vs. template_heap_command <symbol>
-
-    # The `Chunk` class abstracts away many heap & gdb module internals.
-    chunk = pwndbg.heap.ptmalloc.Chunk(addr)
-
-    print(f"chunk.address: 0x{chunk.address:02x}")
-
-    # Be aware that if a chunk field is unreadable (e.g. a fake chunk straddling an unmapped page boundary) it will be None, always check.
-    if chunk.prev_size is not None:
-        print(f"chunk.prev_size: 0x{chunk.prev_size:02x}")
-
-    if chunk.size is not None:
-        print(f"chunk.size: 0x{chunk.size:02x}")
-
-    # Flags may be retrieved individually or as a dictionary of all 3.
-    if chunk.non_main_arena is not None:
-        print(f"chunk.non_main_arena: {chunk.non_main_arena}")
-
-    if chunk.is_mmapped is not None:
-        print(f"chunk.is_mmapped: {chunk.is_mmapped}")
-
-    if chunk.prev_inuse is not None:
-        print(f"chunk.prev_inuse: {chunk.prev_inuse}")
-
-    if chunk.flags is not None:
-        print(f"chunk.flags: {chunk.flags}")
-
-    if chunk.fd is not None:
-        print(f"chunk.fd: 0x{chunk.fd:02x}")
-
-    if chunk.bk is not None:
-        print(f"chunk.bk: 0x{chunk.bk:02x}")
-
-
-parser = argparse.ArgumentParser()
 parser.description = (
     "Iteratively print chunks on a heap, default to the current thread's active heap."
 )
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -96,6 +96,43 @@ def format_bin(bins, verbose=False, offset=None):
 
 
 parser = argparse.ArgumentParser()
+parser.description = ("Template heap command. You can ignore this.")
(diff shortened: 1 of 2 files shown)


pwndbg/heap/ptmalloc.py
```diff
@@ -32,6 +32,83 @@ def heap_for_ptr(ptr):
...
+        self._bk = None
+        # TODO fd_nextsize, bk_nextsize, key, REVEAL_PTR etc.
+
+    @property
+    def prev_size(self):
+        if self._prev_size is None:
+            if "mchunk_prev_size" in (field.name for field in self._gdbValue.type.fields()):
+                prev_size_field_name = "mchunk_prev_size"
+            else:
+                prev_size_field_name = "prev_size"
```
gsingh93: Can you add a comment here about why we're doing this? Adding the libc version when this changed would be great.

pwndbg/heap/ptmalloc.py
```diff
@@ -32,6 +32,83 @@ def heap_for_ptr(ptr):
     return ptr & ~(HEAP_MAX_SIZE - 1)
 
 
+class Chunk:
```
gsingh93, line 35: I think it would be good to have `Chunk` be non-lazy, but then have `class LazyChunk(Chunk)` be this class. There are various benefits to this, one being when we have a proper unit testing framework later, we'll be able to construct Chunks in various configurations easily, without having to worry about mocking out too much stuff.

If this sounds good to you, `Chunk` can go in `pwndbg/lib/heap/chunk.py` and would have these fields passed in to it (so no dependence on GDB).

pwndbg/heap/ptmalloc.py
```diff
@@ -32,6 +32,83 @@ def heap_for_ptr(ptr):
...
+                self._prev_size = int(self._gdbValue[prev_size_field_name])
+            except gdb.MemoryError:
+                pass
+
+        return self._prev_size
+
+    @property
+    def size(self):
+        if self._size is None:
+            if "mchunk_size" in (field.name for field in self._gdbValue.type.fields()):
```
gsingh93: Can we have a single function we call to handle all the renames? We could have a `dict` like (taken from elsewhere in the code):
```
        renames = {
            "mchunk_size": "size",
            "mchunk_prev_size": "prev_size",
        }
```

Then loop through the keys and replace with the other names if we don't find them.

pwndbg/heap/ptmalloc.py
```diff
@@ -32,6 +32,83 @@ def heap_for_ptr(ptr):
...
+
+    @property
+    def flags(self):
+        if self._flags is None:
+            sz = self.size
+            if sz is not None:
+                self._flags = {
+                    "non_main_arena": bool(sz & ptmalloc.NON_MAIN_ARENA),
+                    "is_mmapped": bool(sz & ptmalloc.IS_MMAPPED),
+                    "prev_inuse": bool(sz & ptmalloc.PREV_INUSE),
```
gsingh93: I'd like to have helper methods/properties like `Chunk.prev_inuse`, so it's more convenient to check one flag if that's all we need. If we do that, then we can have this function use those properties.

pwndbg/heap/ptmalloc.py
```diff
@@ -32,6 +32,83 @@ def heap_for_ptr(ptr):
...
+                }
+
+        return self._flags
+
+    @property
+    def fd(self):
```
gsingh93, line 124: I feel like there should be a way to simplify this "lazy" pattern we're using through out these functions, so the code looks simpler and we can reuse it in other places we want lookups to be lazy:
```
if self.FIELD is None:
    try:
        self.FIELD = int(self._gdbvalue[FIELD])
    except gdb.MemoryError:
        pass
return self.FIELD
```

But I don't know the best way to do that right now. Not a big deal for this PR, just something to think about.

pwndbg/commands/heap.py
```diff
@@ -95,6 +95,43 @@ def format_bin(bins, verbose=False, offset=None):
...
+parser = argparse.ArgumentParser()
+parser.description = "Template heap command. You can ignore this."
+parser.add_argument("addr", type=int, help="Address of a chunk header.")
+
+
+@pwndbg.commands.ArgparsedCommand(parser)
+@pwndbg.commands.OnlyWhenRunning
+@pwndbg.commands.OnlyWithResolvedHeapSyms
+@pwndbg.commands.OnlyWhenHeapIsInitialized
+def template_heap_command(addr):
```
gsingh93: Maybe we should just make this command like `malloc_chunk`?

gsingh93 (changes requested):

pwndbg/heap/ptmalloc.py
CptGibbon: Agreed, added properties for individual flags.

pwndbg/heap/ptmalloc.py
CptGibbon: Added a comment to that effect.

pwndbg/heap/ptmalloc.py
CptGibbon: Good idea, added a function to handle this.
I've used a list within the dictionary to handle more than 2 names if any more renames occur in the future.

pwndbg/heap/ptmalloc.py
CptGibbon: Yes this could be challenging, since not all properties follow the same pattern e.g. `flags` require the chunk's `size` field to be fetched first. I'll leave this for now.

pwndbg/commands/heap.py
CptGibbon: I was hoping `malloc_chunk` would eventually be as simple as:
`print(chunk)` or `chunk.pretty_print()`

For now I'd argue keeping this template as simple as possible may be the best way to show new contributors how to use the various heap classes.

pwndbg/heap/ptmalloc.py
CptGibbon: I really like the idea of modularity for the sake of testing, but I feel like heap-related testing would benefit from further discussion first. A chunk outside the context of GLIBC malloc doesn't make sense e.g. we're expecting a pwndbg command to print a chunk as being free, but to determine that it might first have to locate the heap & arena for that chunk, then iterate through a specific bin.

If you're comfortable with it, we could postpone this until we've discussed heap command testing some more.

pwndbg/commands/heap.py
```diff
@@ -95,6 +95,53 @@ def format_bin(bins, verbose=False, offset=None):
...
+    """Template heap command with example uses of pwndbg's heap classes."""
+
+    # `addr` is a gdb.Value of one of several types, depending on how the user invoked this command
+    # e.g. template_heap_command <address> vs. template_heap_command <symbol>
+
+    # The `Chunk` class abstracts away many heap & gdb module internals.
+    chunk = pwndbg.heap.ptmalloc.Chunk(addr)
+
+    print(f"chunk.address: 0x{chunk.address:02x}")
```
lebr0nli: I think this PR is great, the only tiny potential issue I can find is: should we use `f-strings` syntax?
I like `f-strings` and it's convenient, but for the compatibility, maybe we should not use it until we fully deprecate Python<3.6 support I guess?
(I saw the comment from @disconnect3d about the Python version we want to support here: https://github.com/pwndbg/pwndbg/pull/1063#issuecomment-1219942748)

gsingh93:
> maybe we should not use it until we fully deprecate Python<3.6 support I guess?

We had a discussion on Discord after that comment, I remember the general consensus was we should support distros that are not EOL yet, which would include Ubuntu 18.04, 20.04 and 22.04, and we now run the CI on all three. Ubuntu 18.04 uses Python 3.6.9, so personally I think it's OK to drop support for versions older than 3.6, but I know we always have the issue of someone wanting to use it on those older versions.

CptGibbon:
So far the consensus seems to be:
1) Ditch `template_heap_command`
2) Keep f-strings

@gsingh93 are you happy to postpone the unit testing stuff?
@disconnect3d do you agree with the above?

gsingh93:
Yup, feel free to hold off on the unit testing stuff, we can revisit it later.

disconnect3d:
Yeah, I am fine with that. So what with the `Chunk` class? Are you going to refactor heap commands to use it?:P

CptGibbon:
>So what with the Chunk class? Are you going to refactor heap commands to use it?

Exactly. Right now contributors must manually find many heap values in their commands:
https://github.com/pwndbg/pwndbg/blob/1ca4d2d33ca0dcf1969b80853d4e7090cf630e42/pwndbg/commands/heap.py#L672-L674
Hopefully we can avoid this (and benefit from the built-in efficiency the gdb module gives us) by introducing these classes to pwndbg's heap inspection engine.

PR #1162, opened 2022-09-25
fix: make mprotect command truly multi-arch

alufers:
Added register saving based on reg_sets defined for each processor architecture, additionally shellcraft is used to generate the arch-specific shellcode.

Unfortunately this command is not currently tested on platforms other than x86_64.


tests/test_mprotect.py
```diff
@@ -2,11 +2,13 @@
 
 import pwndbg
 import tests
+from pwndbg.gdblib import arch_mod
 
 MPROTECT_BINARY = tests.binaries.get("mprotect.out")
 
 
 def test_mprotect(start_binary):
+    arch_mod.update()
```
disconnect3d: Is this necessary at all? I think this should execute automatically when we start the binary/load an objfile iirc.

tests/binaries/mprotect.c
```diff
@@ -12,6 +12,5 @@ int main() {
 
     void (*f)() = (void (*)())func;
     printf("ptr = %p", f);
```
disconnect3d: Do we need this code at all and `func` variable?

pwndbg/commands/mprotect.py
```diff
@@ -44,39 +45,37 @@ def prot_str_to_val(protstr):
...
+    # obtain the registers that need to be saved for the current platform
+    # we save the registers that are used for arguments, return value and the program counter
+    current_regs = reg_sets[pwndbg.gdblib.arch.current]
+    regs_to_save = current_regs.args + (current_regs.retval, current_regs.pc)
 
-    pwndbg.gdblib.memory.write(pwndbg.gdblib.regs.rip, shellcode)
+    # save the registers
+    saved_registers = {}
+    for reg in regs_to_save:
+        saved_registers[reg] = pwndbg.gdblib.regs[reg]
```
disconnect3d: We can simplify this to:
```py
saved_registers = {reg: pwndbg.gdblib.regs[reg] for reg in regs_to_save}
```

pwndbg/commands/mprotect.py
```diff
@@ -44,39 +45,37 @@ def prot_str_to_val(protstr):
...
+
+    # save the memory which will be overwritten by the shellcode
+    saved_instruction_bytes = pwndbg.gdblib.memory.read(
+        saved_registers[current_regs.pc], len(shellcode)
+    )
+    pwndbg.gdblib.memory.write(saved_registers[current_regs.pc], shellcode)
 
     # execute syscall
     gdb.execute("nextsyscall")
```
disconnect3d, line 73: Actually, I think just a single `stepi` should be enough here

pwndbg/commands/mprotect.py
```diff
@@ -44,39 +45,37 @@ def prot_str_to_val(protstr):
...
-    saved_rax = pwndbg.gdblib.regs.rax
-    saved_rbx = pwndbg.gdblib.regs.rbx
-    saved_rcx = pwndbg.gdblib.regs.rcx
-    saved_rdx = pwndbg.gdblib.regs.rdx
-    saved_rip = pwndbg.gdblib.regs.rip
 
     prot_int = prot_str_to_val(prot)
 
+    # generate a shellcode that executes the mprotect syscall
     shellcode_asm = pwnlib.shellcraft.syscall("SYS_mprotect", int(addr), int(length), int(prot_int))
```
disconnect3d: Can we minimize the number of instructions emitted here to just one? 

We would have to set registers via `set $reg=val` or better using pwndbg api: `setattr(pwndbg.gdblib.regs, reg_name, value)`? Though the question is if we know which register holds syscall number, but I bet we have this information in the architecture structs

disconnect3d: (If we emitted only a single `syscall` we could just do a single `stepi` instead of `nextsyscall` + `stepi`)

pwndbg/commands/mprotect.py
```diff
@@ -44,39 +45,37 @@ def prot_str_to_val(protstr):
...
-    pwndbg.gdblib.regs.rax = saved_rax
-    pwndbg.gdblib.regs.rbx = saved_rbx
-    pwndbg.gdblib.regs.rcx = saved_rcx
-    pwndbg.gdblib.regs.rdx = saved_rdx
-    pwndbg.gdblib.regs.rip = saved_rip
+    pwndbg.gdblib.memory.write(saved_registers[current_regs.pc], saved_instruction_bytes)
+
+    # restore the registers
+    for reg in regs_to_save:
+        gdb.execute("set ${}={}".format(reg, saved_registers[reg]))
```
disconnect3d: we can do those writes with pwndbg api and maybe use the dict items:
```py
for register, value in saved_registers.items():
    setattr(pwndbg.gdblib.regs, register, value)
```

pwndbg/commands/mprotect.py
```diff
@@ -44,39 +45,37 @@ def prot_str_to_val(protstr):
...
+
+    # save the memory which will be overwritten by the shellcode
+    saved_instruction_bytes = pwndbg.gdblib.memory.read(
+        saved_registers[current_regs.pc], len(shellcode)
+    )
+    pwndbg.gdblib.memory.write(saved_registers[current_regs.pc], shellcode)
 
     # execute syscall
     gdb.execute("nextsyscall")
     gdb.execute("stepi")
```
disconnect3d, line 88: We should grab the syscall return value and print it out to the user to make them aware what was the result.

pwndbg/commands/mprotect.py
```diff
@@ -11,8 +11,9 @@
 import pwndbg.lib.which
 import pwndbg.wrappers.checksec
 import pwndbg.wrappers.readelf
+from pwndbg.lib.regs import reg_sets
 
-parser = argparse.ArgumentParser(description="Calls mprotect. x86_64 only.")
+parser = argparse.ArgumentParser(description="Calls mprotect.")
```
disconnect3d:
```suggestion
parser = argparse.ArgumentParser(description="""
Calls the mprotect syscall and prints its result value

Note that the mprotect syscall may fail for various reasons
(see `man mprotect`) and a non-zero error return value
can be decoded with the `errno <value>` command.

Examples:
    mprotect $rsp PROT_READ|PROT_WRITE|PROT_EXEC
    mprotect some_symbol PROT_NONE
""")
```

tests/test_mprotect.py
disconnect3d: FWIW this test could look like this:
```py
SMALL_BINARY = tests.binaries.get("crash_simple.out.hardcoded")


def test_mprotect_executes_properly(start_binary):
    """
    Tests the mprotect command
    """
    start_binary(SMALL_BINARY)

    pc = pwndbg.gdblib.regs.pc
    vm = pwndbg.vmmap.find(pc)

    # Check if we can use mprotect with address provided as value
    # and to set page permissions to RWX
    gdb.execute(
        "mprotect %d 4096 PROT_EXEC|PROT_READ|PROT_WRITE" % pc
    )
    vm = pwndbg.vmmap.find(pc)
    assert vm.read and vm.write and vm.execute

    # Check if we can use mprotect with address provided as register
    # and to set page permissions to none
    gdb.execute("mprotect $pc 0x1000 PROT_NONE")
    vm = pwndbg.vmmap.find(pc)
    assert not (vm.read and vm.write and vm.execute)
```

This way we would also ensure we can pass symbols as addresses.
And if we print out the mprotect result, we should assert it too.

Also: not sure but maybe this should use $sp instead of $pc as marking $pc not-executable will not allow any further execution :p

pwndbg/commands/mprotect.py
alufers: Not really, x86 (32bit) passes the arguments via the stack for example

tests/test_mprotect.py
alufers: Yes, it isn't called during tests

pwndbg/commands/mprotect.py
disconnect3d: Yes and no: x86 calling convention does indeed pass arguments on the stack but arguments to syscalls are passed via registers; see https://chromium.googlesource.com/chromiumos/docs/+/master/constants/syscalls.md#x86-32_bit

alufers: Unfortunately there is no such information in the architecture structs for the syscall number register, I will add it in another pull request and fix this, though.

tests/test_mprotect.py
```diff
@@ -3,46 +3,29 @@
...
-    vmmap_entry = [x for x in vmmaps_after if x not in vmmaps_before][0]
-
-    assert vmmap_entry.split()[2] == "rwxp"
-
-    # continue execution
-    gdb.execute("continue")
+    start_binary(SMALL_BINARY)
+
+    pc = pwndbg.gdblib.regs.pc
+    vm = pwndbg.gdblib.vmmap.find(pc)
```
disconnect3d:
```suggestion
```

disconnect3d (approved):

PR #1168, opened 2022-09-27
add type for ./pwndbg/lib/

syheliel:
The only one not fixed is:
```
pwndbg/lib/memoize.py:34: error: "memoize" has no attribute "caches"; maybe "cache"?
```
I'm not sure if adding `caches` as a variable for `memoize` will break the original behavior

### DIFF
--- a/pwndbg/lib/arch.py
+++ b/pwndbg/lib/arch.py
@@ -13,12 +13,9 @@ class Arch:
         self.ptrmask = (1 << 8 * ptrsize) - 1
         self.endian = endian
 
-        self.fmt = {
-            (4, "little"): "<I",
-            (4, "big"): ">I",
-            (8, "little"): "<Q",
-            (8, "big"): ">Q",
-        }.get((self.ptrsize, self.endian))
+        self.fmt = {(4, "little"): "<I", (4, "big"): ">I", (8, "little"): "<Q", (8, "big"): ">Q",}[
+            (self.ptrsize, self.endian)
+        ]  # type: str
 
         if self.name == "arm" and self.endian == "big":
             self.qemu = "armeb"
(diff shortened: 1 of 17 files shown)


syheliel:
I can't repduce the lint error on my local machine
image

pwndbg/lib/arch.py
```diff
@@ -39,4 +36,4 @@ def signed(self, integer: int) -> int:
         return self.unpack(self.pack(integer), signed=True)  # type: ignore
 
     def unsigned(self, integer: int) -> int:
-        return self.unpack(self.pack(integer))
+        return self.unpack(self.pack(integer))  # type: ignore
```
gsingh93: Can you explain why we need the `type: ignore`? I like to actually put the error message we're ignoring in a comment so it's clear why we need the `type: ignore`.

syheliel: It is caused by a wrong type in `unpack`, already fixed.

disconnect3d:
What is the state of this PR @syheliel? It seems that some tests are failing due to the changes made here: 
image

syheliel:
@disconnect3d I have fixed the test. But the name of variable `e` is really meaningless in `commands/heap.py`:
```
            e = addr + 2 * size_sz
            e += allocator.tcache_entry.keys().index("key") * ptr_size
            e = pwndbg.gdblib.memory.pvoid(e)
```

disconnect3d:
I'm not sure about the part in pwndbg/config.py - I am merging this for now as the changes are good, but the config.py additions may be not obvious for others. We are also missing many others config.py variables there.

On the other side: maybe we should switch to a more readable way of defining config variables?

PR #1170, opened 2022-09-27
Support starting the IPython interactive prompt with `ipi` command

lebr0nli:
This `ipi` command can help us to start the IPython interactive prompt instead of the Python interactive prompt provided by the `pi` command.

截圖 2022-09-27 下午12 15 43

### DIFF
--- a/pwndbg/commands/ipython_interactive.py
+++ b/pwndbg/commands/ipython_interactive.py
@@ -1,0 +1,40 @@
+"""
+Command to start an interactive IPython prompt.
+"""
+import sys
+from contextlib import contextmanager
+
+import gdb
+
+import pwndbg.commands
+
+
+@contextmanager
+def switch_to_ipython_env():
+    """We need to change stdout/stderr to the default ones, otherwise we can't use tab or autocomplete"""
+    # Save GDB's stdout and stderr
+    saved_stdout = sys.stdout
+    saved_stderr = sys.stderr
+    # Use Python's default stdout and stderr
+    sys.stdout = sys.__stdout__
+    sys.stderr = sys.__stderr__
+    yield
+    # Restore GDB's stdout and stderr
+    sys.stdout = saved_stdout
+    sys.stderr = saved_stderr
+    # Restore Python's default ps1 and ps2 for GDB's `pi` command
+    sys.ps1 = ">>> "
+    sys.ps2 = "... "
+
+
+@pwndbg.commands.ArgparsedCommand("Start an interactive IPython prompt.")
+def ipi():
+    with switch_to_ipython_env():
+        # Use `gdb.execute` to embed IPython into GDB's variable scope
+        code4ipython = """import IPython
+import pwn
+IPython.embed(colors='neutral',banner1='',confirm_exit=False,simple_prompt=False)
+""".strip().replace(
+            "\n", ";"
+        )
+        gdb.execute(f"pi {code4ipython}")
--- a/pwndbg/commands/ipython_interactive.py
+++ b/pwndbg/commands/ipython_interactive.py
@@ -32,7 +32,9 @@ def ipi():
     with switch_to_ipython_env():
         # Use `gdb.execute` to embed IPython into GDB's variable scope
         code4ipython = """import IPython
+import jedi
 import pwn
+jedi.Interpreter._allow_descriptor_getattr_default = False
 IPython.embed(colors='neutral',banner1='',confirm_exit=False,simple_prompt=False)
 """.strip().replace(
             "\n", ";"
--- a/pwndbg/commands/ipython_interactive.py
+++ b/pwndbg/commands/ipython_interactive.py
@@ -15,6 +15,7 @@ def switch_to_ipython_env():
     # Save GDB's stdout and stderr
     saved_stdout = sys.stdout
     saved_stderr = sys.stderr
+    saved_excepthook = sys.excepthook
     # Use Python's default stdout and stderr
     sys.stdout = sys.__stdout__
     sys.stderr = sys.__stderr__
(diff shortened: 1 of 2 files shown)


lebr0nli:
cc: @gsingh93 , feel free to let me know if you found any bugs about the `import pwntools automatically` feature you mentioned in Discord or have any suggestions about this command :)

lebr0nli:
Somehow, I just found this command has a weird bug that when we debugging a process that uses a libc without debugging symbols and didn't use the heap heuristic when we trying to press <kbd>TAB</kbd> to autocomplete the code after `pwndbg.heap.current.`, you will see tons of weird messages and let us can't continue typing:

截圖 2022-09-27 下午2 00 14

This can be reproduced with the binary and libc in DiceCTF 2022 [baby-rop](https://github.com/dicegang/dicectf-2022-challenges/tree/master/pwn/baby-rop/bin)

I have no ideas about this bug right now.

After solving this bug, I'll remove the `WIP` tag in the title.

lebr0nli:
> After solving this bug, I'll remove the `WIP` tag in the title.

Ok, that bug is actually an unsafe feature of IPython lol.

TL;DR; It's because some function of the property has been invoked implicitly when autocompleting, and that property only works with debug symbols, so errors occurred.

We can reproduce it by creating a class like:

```python
class Foo:
    @property
    def bar(self):
        import os
        os.system("echo arbitrary code execution!?")

f = Foo()
```

And then press <kbd>TAB</kbd> after `f.` in IPython, `bar` will be invoked.

> There's even a similar issue in Python <3.9 (https://bugs.python.org/issue44752) lol

----

I disabled this feature of IPython by:

```python
import jedi
jedi.Interpreter._allow_descriptor_getattr_default = False
```

> Ref: https://github.com/ipython/ipython/issues/9094#issuecomment-753536247
 
But I'm not sure there are side effects for this or not, let me know if there’s a better solution or this solution not work.

pwndbg/commands/ipython_interactive.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+@pwndbg.commands.ArgparsedCommand("Start an interactive IPython prompt.")
+def ipi():
+    with switch_to_ipython_env():
+        # Use `gdb.execute` to embed IPython into GDB's variable scope
+        code4ipython = """import IPython
+import jedi
+import pwn
+jedi.Interpreter._allow_descriptor_getattr_default = False
```
disconnect3d, line 39: What is `jedi`?

Also @Arusekk do u know if `pwn` import influences anything with the environment we are in? AFAIK pwntools suggested to use `pwnlib` in libraries using pwntools and `pwn` only for small scripts or interactive shells?

pwndbg/commands/ipython_interactive.py
```diff
@@ -0,0 +1,42 @@
...
+    # Use Python's default stdout and stderr
+    sys.stdout = sys.__stdout__
+    sys.stderr = sys.__stderr__
+    yield
+    # Restore GDB's stdout and stderr
+    sys.stdout = saved_stdout
+    sys.stderr = saved_stderr
+    # Restore Python's default ps1 and ps2 for GDB's `pi` command
+    sys.ps1 = ">>> "
+    sys.ps2 = "... "
```
disconnect3d, line 28: Are we sure those are the only things we need to restore? Can you double check `sys.__dict__` before and after?

pwndbg/commands/ipython_interactive.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+@pwndbg.commands.ArgparsedCommand("Start an interactive IPython prompt.")
+def ipi():
+    with switch_to_ipython_env():
+        # Use `gdb.execute` to embed IPython into GDB's variable scope
+        code4ipython = """import IPython
```
disconnect3d, line 36: Do we need to add ipython or jedi into requirements?

disconnect3d:
@lebr0nli nice thing with this getattribute.

Regarding bug, our typeinfo module get all header files from known paths, compiles a file with them and tries to resolve symbols that way. This is... semi-optimal and I think it just does not work for any of our needs (e.g. heap symbols). Correct me if I am wrong, but maybe we should just remove that logic? (not in this PR ofc)

pwndbg/commands/ipython_interactive.py
lebr0nli: Hmm, maybe add `ipython` into requirements is enough because `ipython` depends on `jedi`.

But how about we do something similar like how we treat `ipdb`?

For example:

```python
try:
    import IPython
except ImportError:
    print some hint message to user?
```

Which one do you prefer more?

pwndbg/commands/ipython_interactive.py
lebr0nli: Oh, I also found that `sys.excepthook` changed.
I'll restore it, too.

pwndbg/commands/ipython_interactive.py
lebr0nli: > What is `jedi`?

`IPython` used `jedi` to build some features about autocomplete, and it is also related to the `@property` issue.

So I guess that's why we can affect `IPython` by `jedi.Interpreter._allow_descriptor_getattr_default`.

lebr0nli:
> Regarding bug, our typeinfo module get all header files from known paths, compiles a file with them and tries to resolve symbols that way. This is... semi-optimal and I think it just does not work for any of our needs (e.g. heap symbols). Correct me if I am wrong, but maybe we should just remove that logic? (not in this PR ofc)

Yes, I think you are right, it does not work most of the time.
I'm not pretty sure we should remove that logic or not, but for now, I have a little suggestion.
Is it possible that we just redirect that error into `/dev/null` or something, and only show we got an error when loading the type?
Because those error messages are really long lol, maybe it won't be that scary to the normal user if we don't show it or maybe have some hints for the user in some cases (e.g. turn on the heap heuristic and try again) XD

disconnect3d:
> Is it possible that we just redirect that error into /dev/null or something, and only show we got an error when loading the type?

Yes, pls do that in a separate PR. I think that `subprocess.check_output` we use there does not catch stderr(?)

pwndbg/commands/ipython_interactive.py
disconnect3d: Let's make this opt-in: let's inform users to install ipython if they want to use this.

disconnect3d (changes requested):

disconnect3d:
re: importing ipython - let's make this opt-in and print a warning for users if they don't have ipython installed in a new PR :)

PR #1075, opened 2022-08-18
Some updates about Heap heuristics

lebr0nli:
In this PR, I fixed some issues that @disconnect3d mentioned in https://github.com/pwndbg/pwndbg/pull/1029 and more issues about Heap heuristics.

And there also contains a new command about TLS after refactoring the features about TLS and a bug fix about fsbase/gsbase when the arch is i386

----

Hope this update can make Heap heuristics better :)

PR #1173, opened 2022-09-27
Enhance heap with for static-linked binaries & remove typeinfo bloat

disconnect3d:
This commit enhances the heap commands UX for statically linked binaries and removes typeinfo module bloat.

The typeinfo module had this typeinfo.load function that was looking up a given type. If it didn't find the type, it fallbacked to compiling many many system headers in order to add a symbol for a given type into the program. This was supposed to be used for missing glibc malloc symbols like malloc_chunk.

However, the exact reason it was used: the struct malloc_chunk was never defined in a header file and was always defined in a malloc.c or another .c file in glibc sources.

Another place the typeinfo.load logic of compiling headers was/is used is the `dt` command, which is a windbg alias for getting struct layout type information, e.g.:
```
pwndbg> dt 'struct malloc_chunk'
struct malloc_chunk
    +0x0000 mchunk_prev_size     : size_t
    +0x0008 mchunk_size          : size_t
    +0x0010 fd                   : struct malloc_chunk *
    +0x0018 bk                   : struct malloc_chunk *
    +0x0020 fd_nextsize          : struct malloc_chunk *
    +0x0028 bk_nextsize          : struct malloc_chunk *
pwndbg>
```

However, the whole big issue with typeinfo.load compilation of headers was that most of the time it didn't work because e.g. some headers defined in other paths were missing or that two different headers used the same struct/function name and the compilation failed.

Since this logic almost never gave good results, I am removing it.

Regarding UX for statically linked binaries: we use `info dll` command to see if a binary is statically linked. While this method is not robust, as it may give us wrong results if the statically linked binary used `dlopen(...)` it is probably good enough.

Now, if a heap related command is executed on statically linked binaries, it will inform the user and set the resolving of libc heap symbols via heuristics. Then, it also says to the user they have to set the glibc version and re-run the command.

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -280,14 +280,37 @@ def OnlyAmd64(function):
     return _OnlyAmd64
 
 
+# TODO/FIXME: Move this elsewhere? Have better logic for that? Maybe caching?
+def _is_statically_linked():
+    out = gdb.execute("info dll", to_string=True)
+    return "No shared libraries loaded at this time." in out
+
+
 def OnlyWithResolvedHeapSyms(function):
     @functools.wraps(function)
     def _OnlyWithResolvedHeapSyms(*a, **kw):
         if pwndbg.heap.current.libc_has_debug_syms() or pwndbg.config.resolve_heap_via_heuristic:
             return function(*a, **kw)
         else:
-            print(
-                """%s: This command only works with libc debug symbols.
+            e = lambda s: print(message.error(s))
+            w = lambda s: print(message.warn(s))
+
+            if _is_statically_linked():
+                e(
+                    f"{function.__name__}: Can't find libc symbols addresses required for this command to work since this is a statically linked binary"
+                )
+                w(
+                    """Invoking the `set resolve-heap-via-heuristic on` command to resolve libc symbols via heuristics.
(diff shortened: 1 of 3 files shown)


disconnect3d:
@CptGibbon @lebr0nli pls review that if u have time

CptGibbon:
Yeah this is so much better ❤️

Only snag is that `heap` still raises an error after printing the help message.

heap_error

lebr0nli:
I have the same opinion as @CptGibbon, this enhancement looks great!

Btw the snag @CptGibbon mentioned is my fault, I'm sorry.

I did that before because I think raising an exception here will make it easier to debug, but seems that it's not a good idea to show that exception to the user.
(https://github.com/pwndbg/pwndbg/pull/1075#discussion_r957964323)

Maybe we should still raise it but somehow we change the message of that exception to a more user-friendly message?
Or maybe there is a better way to stop all the functions without using exception?
(I'm not sure, I didn't have a good idea about handling this exception yet.)

disconnect3d:
@lebr0nli I believe that exception is thrown due to a glibc version check.

We should refactor the code so that the glibc version is checked earlier on, and, if it is not set, a proper message is displayed to the user to set it.

We could also......... fallback and set glibc version to the glibc installed in the user system **and warn them that we did exactly this** but the question is: can we retrieve the version the user's system glibc is compiled with? Probably through reading symbolic links?

PR #1178, opened 2022-09-29
Upload codecov report only from Ubuntu22.04

### LINKED ISSUE #1175: Upload codecov report only from one CI job
Currently, we run tests on CI on 3 different OS versions. While this helps us detect more bugs related to environment configuration, it also runs all the steps defined in the GitHub action in all of those CI runs. As a result, I believe each pull request uploads a code coverage from tests to codecov and a given PR needs just one.

So we should probably split the GH action used currently into two: one that will run 18.04 and 20.04 without uploading codecov and one that will run on 22.04 and that will upload codecov.

miles170:
Also update GitHub action dependencies.

Closes #1175.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -6,15 +6,15 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 10
     steps:
-    - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


disconnect3d:
Wow, thanks! :)

PR #1201, opened 2022-10-03
add show-flags and show-compact-regs to ctx regs banner

disconnect3d:
Hopefully his will improve the discoverability/UX for users who are not aware of those options.

This is how the new registers banner looks like:
```
[ REGISTERS / show-flags off / show-compact-regs off ]
```

Fwiw it is 54 chars long (without "---" before and after) so its length should be fine.

Screenshot of current ctx:
image
Compact + flags on:
image


gsingh93:
Can you file an issue to fix the compact registers layout?

disconnect3d:
> Can you file an issue to fix the compact registers layout?
Sure

Merging this as there were only issues with codecov upload (race condition again or what?).

Issue #497, opened 2018-07-08
Refactor common code


disconnect3d:
There are pieces of code that we reuse across the codebase which could be moved to one place.

- [ ]  zeroing hex numbers as in https://github.com/pwndbg/pwndbg/pull/496#discussion_r200819294 :
```
In [3]: '\t0x{:0{n}x}'.format(123, n=8)
Out[3]: '\t0x0000007b'
```
- [ ] unpacking values using `struct` module instead of e.g. `pwndbg.arch.unpack` https://github.com/pwndbg/pwndbg/blob/5b9671f28466d66a5f3ce7f5761a9c918d0eb334/pwndbg/commands/heap.py#L360-L366

gsingh93:
There's actually a few different places where we do the endian conversion:
```
$ grep -r pwndbg.gdblib.arch.endian .
./gdblib/ctypes.py:    if pwndbg.gdblib.arch.endian == "little":
./symbol.py:    if path and (pwndbg.gdblib.arch.endian == pwndbg.gdblib.arch.native_endian):
./commands/windbg.py:        if pwndbg.gdblib.arch.endian == "little":
./commands/heap.py:    fmt = {"little": "<", "big": ">"}[pwndbg.gdblib.arch.endian] + {4: "I", 8: "Q"}[psize]
./commands/heap.py:    fmt = {"little": "<", "big": ">"}[pwndbg.gdblib.arch.endian] + {4: "I", 8: "Q"}[psize]
./commands/search.py:        fmt = {"little": "<", "big": ">"}[pwndbg.gdblib.arch.endian] + {
./commands/hexdump.py:        pwndbg.config.hexdump_group_use_big_endian and pwndbg.gdblib.arch.endian == "little"
./net.py:            if pwndbg.gdblib.arch.endian == "little":
./disasm/__init__.py:        pwndbg.gdblib.arch.current, pwndbg.gdblib.arch.ptrsize, pwndbg.gdblib.arch.endian, extra
./emu/emulator.py:        if pwndbg.gdblib.arch.endian == "little":
```

PR #1207, opened 2022-10-03
Cleanup find_fake_fast command

### LINKED ISSUE #1041: Option for `find_fast_chunk` to find aligned chunks
When I'm looking for a fastbin chunk to free, I need to make sure it's aligned. More generally, I want the chunk returned by `find_fast_chunk` to satisfy all the constraints checked by `try_free`. It'd be useful if there was an option that only returned fastbin chunks that could be freed without failing any glibc checks.

gsingh93:
The old implementation was a bit inefficient and messy. This needs some more testing, but I think it's cleaner.

It also allows to only find aligned chunks, so it fixes #1041

### DIFF
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -544,49 +544,110 @@ def tcachebins(addr=None, verbose=False):
 
 
 parser = argparse.ArgumentParser()
-parser.description = "Find candidate fake fast chunks overlapping the specified address."
+parser.description = "Find candidate fake fast or tcache chunks overlapping the specified address."
 parser.add_argument("addr", type=int, help="Address of the word-sized value to overlap.")
-parser.add_argument("size", nargs="?", type=int, default=None, help="Size of fake chunks to find.")
+parser.add_argument(
+    "size", nargs="?", type=int, default=None, help="Maximum size of fake chunks to find."
+)
+parser.add_argument(
+    "--align",
+    "-a",
+    action="store_true",
+    default=False,
+    help="Whether the fake chunk must be aligned to MALLOC_ALIGNMENT. This is required for tcache "
+    + "chunks and for all chunks when Safe Linking is enabled",
+)
 
 
 @pwndbg.commands.ArgparsedCommand(parser)
 @pwndbg.commands.OnlyWhenRunning
 @pwndbg.commands.OnlyWithResolvedHeapSyms
 @pwndbg.commands.OnlyWhenHeapIsInitialized
-def find_fake_fast(addr, size=None):
+def find_fake_fast(addr, size=None, align=False):
     """Find candidate fake fast chunks overlapping the specified address."""
     psize = pwndbg.gdblib.arch.ptrsize
     allocator = pwndbg.heap.current
-    align = allocator.malloc_alignment
+    malloc_alignment = allocator.malloc_alignment
+
     min_fast = allocator.min_chunk_size
     max_fast = allocator.global_max_fast
     max_fastbin = allocator.fastbin_index(max_fast)
-    start = int(addr) - max_fast + psize
-    if start < 0:
+
+    if size is None:
+        size = max_fast
+    elif size > addr:
+        print(message.warn("Size of 0x%x is greater than the target address 0x%x", (size, addr)))
+        size = addr
+    elif size > max_fast:
         print(
             message.warn(
-                "addr - global_max_fast is negative, if the max_fast is not corrupted, you gave wrong address"
+                "0x%x is greater than the global_max_fast value of 0x%x" % (size, max_fast)
             )
         )
-        start = 0  # TODO, maybe some better way to handle case when global_max_fast is overwritten with something large
-    mem = pwndbg.gdblib.memory.read(start, max_fast - psize, partial=True)
+    elif size < min_fast:
+        print(
+            message.warn(
+                "0x%x is smaller than the minimum fastbin chunk size of 0x%x" % (size, min_fast)
+            )
+        )
+        size = min_fast
+
+    # Clear the flags
+    size &= ~0xF
+
+    start = int(addr) - size + psize
+
+    if align:
+        # If a chunk is aligned to MALLOC_ALIGNMENT, the size field should be at
+        # offset `psize`. First we align up to a multiple of `psize`
+        new_start = pwndbg.lib.memory.align_up(start, psize)
+
+        # Then we make sure we're at a multiple of `psize` but not `psize*2` by
+        # making sure the bottom nibble gets set to `psize`
+        new_start |= psize
+
+        # We should not have increased `start` by more than `psize*2 - 1` bytes
+        distance = new_start - start
+        assert distance < psize * 2
+
+        # If we're starting at a higher address, we still only want to read
+        # enough bytes to reach our target address
+        size -= distance
+
+        # Clear the flags
+        size &= ~0xF
+
+        start = new_start
+
+    print(
+        message.notice(
+            "Searching for fastbin sizes up to 0x%x starting at 0x%x resulting in an overlap of 0x%x"
+            % (size, start, addr)
+        )
+    )
+
+    # Only consider `size - psize` bytes, since we're starting from after `prev_size`
+    mem = pwndbg.gdblib.memory.read(start, size - psize, partial=True)
 
     fmt = {"little": "<", "big": ">"}[pwndbg.gdblib.arch.endian] + {4: "I", 8: "Q"}[psize]
 
-    if size is None:
-        sizes = range(min_fast, max_fast + 1, align)
-    else:
-        sizes = [int(size)]
-
     print(C.banner("FAKE CHUNKS"))
-    for size in sizes:
-        fastbin = allocator.fastbin_index(size)
-        for offset in range((max_fastbin - fastbin) * align, max_fast - align + 1):
-            candidate = mem[offset : offset + psize]
-            if len(candidate) == psize:
-                value = struct.unpack(fmt, candidate)[0]
-                if allocator.fastbin_index(value) == fastbin:
-                    malloc_chunk(start + offset - psize, fake=True)
+    step = malloc_alignment if align else 1
+    for i in range(0, len(mem), step):
+        candidate = mem[i : i + psize]
+        if len(candidate) == psize:
+            value = struct.unpack(fmt, candidate)[0]
+
+            # Clear any flags
+            value &= ~0xF
+
+            if value < min_fast:
+                continue
+
+            # The value must be less than or equal to the max size we're looking
+            # for, but still be able to reach the target address
+            if value <= size and i + value >= size:
+                malloc_chunk(start + i - psize, fake=True)
 
 
 pwndbg.gdblib.config.add_param(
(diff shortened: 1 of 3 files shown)


pwndbg/commands/heap.py
```diff
@@ -551,47 +551,68 @@ def tcachebins(addr=None, verbose=False):
...
-parser.add_argument("size", nargs="?", type=int, default=None, help="Size of fake chunks to find.")
+parser.add_argument(
+    "size", nargs="?", type=int, default=None, help="Maximum size of fake chunks to find."
+)
+parser.add_argument(
+    "--align",
+    "-a",
+    action="store_true",
+    default=False,
+    help="Whether the fake chunk must chunk aligned.",
```
disconnect3d: must be an aligned chunk? sth wrong with help string here

gsingh93: I think I meant "Whether the fake chunk must *be* chunk aligned", but maybe it should be something like "Whether the fake chunk must be aligned to MALLOC_ALIGNMENT"?

Also, I think with the alignment check that's the only difference between fastbin and tcache chunks? In that case maybe I should mention tcache in the description.

CptGibbon:
This appears not to work with the common case of `find_fake_fast &__malloc_hook`, although I'm not sure why...

gsingh93:
Weird, I tested it yesterday by putting some fake chunk sizes in memory write before `__free_hook` and it was finding them. I can take a closer look.

CptGibbon:
I think maybe the `align` variable got reused?
It starts as an option to the command, with the value `False`, then it's used to hold `allocator.malloc_alignment` afterwards.

gsingh93:
Ah, nice catch, that's probably it. Will fix. Will also add a test to check for the unaligned case.

gsingh93:
I fixed the alignment bug, but I still need to add some tests for this.

pwndbg/commands/heap.py
```diff
@@ -549,49 +549,96 @@ def tcachebins(addr=None, verbose=False):
...
-parser.add_argument("size", nargs="?", type=int, default=None, help="Size of fake chunks to find.")
+parser.add_argument(
+    "size", nargs="?", type=int, default=None, help="Maximum size of fake chunks to find."
+)
+parser.add_argument(
+    "--align",
+    "-a",
+    action="store_true",
+    default=False,
+    help="Whether the fake chunk must be aligned to MALLOC_ALIGNMENT.",
```
disconnect3d: Btw when does it have to be aligned or not? Does this depend on libc?

pwndbg/commands/heap.py
```diff
@@ -549,49 +549,96 @@ def tcachebins(addr=None, verbose=False):
...
-        for offset in range((max_fastbin - fastbin) * align, max_fast - align + 1):
-            candidate = mem[offset : offset + psize]
-            if len(candidate) == psize:
-                value = struct.unpack(fmt, candidate)[0]
-                if allocator.fastbin_index(value) == fastbin:
-                    malloc_chunk(start + offset - psize, fake=True)
+    step = malloc_alignment if align else 1
+    for i in range(0, len(mem), step):
+        candidate = mem[i : i + psize]
+        if len(candidate) == psize:
```
disconnect3d, line 638: Can we remove this if condition by changing the loop to `for i in range(0, len(mem)-psize, step):`? 

(Maybe there should be `len(mem)-psize+1`, not sure)

pwndbg/commands/heap.py
```diff
@@ -549,49 +549,96 @@ def tcachebins(addr=None, verbose=False):
...
+        message.notice(
+            "Searching for fastbin sizes up to 0x%x starting at 0x%x resulting in an overlap of 0x%x"
+            % (size, start, addr)
+        )
+    )
+
+    # Only consider `size - psize` bytes, since we're starting from after `prev_size`
+    mem = pwndbg.gdblib.memory.read(start, size - psize, partial=True)
+
+    fmt = {"little": "<", "big": ">"}[pwndbg.gdblib.arch.endian] + {4: "I", 8: "Q"}[psize]
```
disconnect3d, line 632: side note: we should probably have a helper for this

pwndbg/commands/heap.py
```diff
@@ -549,49 +549,96 @@ def tcachebins(addr=None, verbose=False):
...
-            candidate = mem[offset : offset + psize]
-            if len(candidate) == psize:
-                value = struct.unpack(fmt, candidate)[0]
-                if allocator.fastbin_index(value) == fastbin:
-                    malloc_chunk(start + offset - psize, fake=True)
+    step = malloc_alignment if align else 1
+    for i in range(0, len(mem), step):
+        candidate = mem[i : i + psize]
+        if len(candidate) == psize:
+            value = struct.unpack(fmt, candidate)[0]
```
disconnect3d, line 639: Random thought: instead of `memory.read(...)`+read at offset+unpack we could do `memory.pvoid(...)` probably

pwndbg/commands/heap.py
gsingh93: I think tcache chunks are the only ones that need to be aligned, otherwise all the checks between the two are pretty much the same.

CptGibbon: Safe Linking enforces alignment for singly-linked bins (both tcache & fastbins)

pwndbg/commands/heap.py
gsingh93: I agree, doing that is tracked in this issue: https://github.com/pwndbg/pwndbg/issues/497

pwndbg/commands/heap.py
disconnect3d: Can we either add a comment on this or describe this in a docstring?

CptGibbon:
bad_size

This fake 0x20-sized chunk overlaps the target data, but isn't considered a candidate.
Previously it was correctly identified.
Not sure what's causing it, could be the bug you mentioned in Discord.

tests/binaries/heap_find_fake_fast.c
```diff
@@ -10,17 +10,79 @@
...
+void setup_mem(unsigned long size, unsigned distance) {
+    memset(buf, 0, sizeof(buf));
+    target_address = 0;
+
+    char *chunk_size_addr = (char*)&target_address - distance + __SIZEOF_POINTER__;
+    *(unsigned long*)chunk_size_addr = size;
+}
+
+int main(void) {
+    assert((unsigned long)&target_address - (unsigned long)buf == sizeof(buf));
```
CptGibbon: Could make this more likely to succeed if `buf` and `target_address` were in the same struct.

tests/binaries/heap_find_fake_fast.c
```diff
@@ -10,17 +10,79 @@
...
+    // A valid aligned fastbin chunk with all flags set
+    setup_mem(0x2F, 0x10);
+    break_here();
+
+    // A valid unaligned fastbin chunk
+    setup_mem(0x20, 0x11);
+    break_here();
+
+    // A valid fastbin chunk that's too close to the target address
+    setup_mem(0x20, 0x8);
```
CptGibbon: This clobbers the target value, was that the intention?

tests/binaries/heap_find_fake_fast.c
```diff
@@ -10,17 +10,79 @@
...
+    break_here();
+
+    // An invalid chunk with a size below the minimum chunk size
+    setup_mem(0x1F, 0x10);
+    break_here();
+
+    // A valid fastbin chunk just in range of the target address
+    setup_mem(0x80, 0x70);
```
CptGibbon: Just in range would be `setup_mem(0x80, 0x80)`

tests/heap/test_find_fake_fast.py
```diff
@@ -0,0 +1,144 @@
...
+    unmapped_heap_info = pwndbg.heap.ptmalloc.heap_for_ptr(pwndbg.symbol.address("fake_chunk"))
+    assert pwndbg.gdblib.memory.peek(unmapped_heap_info) is None
+
+    # A gdb.MemoryError raised here indicates a regression from PR #1145
+    gdb.execute("find_fake_fast (void*)&fake_chunk+0x70")
+
+    result = gdb.execute("p/x &target_address", to_string=True)
+    matches = re.findall(r"\$1 = (0x[0-9a-f]+)", result)
+    assert len(matches) == 1
+    target_address = int(matches[0], 16)
```
CptGibbon: Any reason not to use `pwndbg.symbol.address()`?

pwndbg/commands/heap.py
```diff
@@ -549,49 +549,96 @@ def tcachebins(addr=None, verbose=False):
...
+            if value < min_fast:
+                continue
+
+            # The value must be less than or equal to the max size we're looking
+            # for, but still be able to reach the target address (starting from
+            # `prev_size`, not `size`
+            if value <= size and i - psize + value >= size:
```
CptGibbon: Counting started at the size field, so I don't think subtracting `psize` here is necessary.
This could be what's causing the command to miss its mark by 8 bytes.

CptGibbon (commented):
Some rigorous tests and a nice `--align` option, just one last bug in the command to fix.

tests/binaries/heap_find_fake_fast.c
gsingh93: Yea, this is testing that if the `size` field is actually on the target address, we don't return it. Just in case there's an off-by-one bug somewhere.

tests/binaries/heap_find_fake_fast.c
gsingh93: I was thinking of doing this, but I went with a separate variable as it was easy to write the Python side of the test by just referencing the variable by name. Let's see how this goes, and if it causes problems we can change it.

tests/heap/test_find_fake_fast.py
gsingh93: I got so used to running `gdb.execute` that I forgot I should just do that :P Will fix it.

disconnect3d (approved):

gsingh93:
Rebased. Still haven't fixed the issue, going to get https://github.com/pwndbg/pwndbg/pull/1286 merged first.

gsingh93:
The bug in the last version of this PR should be fixed now.

I was hoping to make this command also support partial writes of the target address, but I think it's fine to merge this without that and add it in a separate PR.

pwndbg/commands/heap.py
gsingh93: Maybe I could call `pwndbg.lib.memory.align_down(len(mem), psize)` instead?

disconnect3d: if that works then yeah

pwndbg/commands/heap.py
gsingh93: It would work, but it should be more efficient to read all of the memory at once instead of one pointer at a time.

disconnect3d: I guess it depends on the memory size and target, e.g. if you debug a remote target via additional debugging hardware (jtag?) then memory accesses may be slow and it may be better to read 100B instead of 1kB etc (?). Though I may be wrong and maybe reading a big blob will be faster.

disconnect3d: Anyway, I'm fine with reading all at once. We could optionally add a comment that it may be worth changing this for remote debugging but we do not know if its worth it (?).

CptGibbon (approved):
Honestly if this passes the recently merged tests it's good to go 👍

PR #1214, opened 2022-10-04
Add `malloc_chunk` command test

CptGibbon:
Test the `malloc_chunk` command against chunks in different bins.
Waiting on heuristic heap issue resolution before running the same tests with `set resolve-heap-via-heuristic on`.


disconnect3d:
Would be nice to add some "unhappy" path tests too :P

PR #1218, opened 2022-10-04
leakfind should default to $sp

### LINKED ISSUE #1210: `leakfind` should default to `$sp`
The first argument to `leakfind` is required, but it should just default to `$sp` like `probeleak` does.

amoghrajesh:
Fixes #1210 

Raising the PR to "dev" branch.

### DIFF
--- a/pwndbg/commands/leakfind.py
+++ b/pwndbg/commands/leakfind.py
@@ -60,7 +60,9 @@ that point to a section in filename which begin near $rsp, are never 0x48 bytes 
 and are a maximum length of 6.
 """
 parser.formatter_class = argparse.RawDescriptionHelpFormatter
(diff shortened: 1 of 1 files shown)


amoghrajesh:
@gsingh93 I am unable to figure out why the lint is failing too. Could you help? The logs are not so informative

gsingh93:
Thanks! Just FYI, you can run `./lint.sh` locally to see the lint errors, and `./lint.sh -f` will attempt to fix any errors it can.

disconnect3d:
Is this really working? I think that `$sp` string will fail here on `int(address)`

gsingh93:
I tested it locally and it worked. And it's the same as what `probeleak` is doing, and that works as well.

PR #1223, opened 2022-10-05
events.py: remove unused Pause class

disconnect3d:


pwndbg/gdblib/events.py
```diff
@@ -12,8 +12,7 @@
 
 import pwndbg.config
 
-debug = pwndbg.config.Parameter("debug-events", False, "display internal event debugging info")
-pause = 0
+debug = pwndbg.config.Parameter("debug-events", True, "display internal event debugging info")
```
gsingh93: You probably don't want to make the default `True`?

disconnect3d: No, good catch; that was from other testing

disconnect3d:
```suggestion
debug = pwndbg.config.Parameter("debug-events", False, "display internal event debugging info")
```

gsingh93:
This seems to be used in `StartEvent`.

disconnect3d:
@gsingh93 how? there is an `if pause:` condition in `connect` I missed, but I don't think this is used in `StartEvent`? Mainly, I don't see the `Pause` class to be used anywhere.

PR #1228, opened 2022-10-05
`set history expansion off` by default (Fix #1227)

### LINKED ISSUE #1227: `set history expansion on` break `!command-string`
I think this might be a bug or bad design of GDB maybe, because the features of `history expansion` is conflict to `!command-string` feature.

And because we set it to `on` in `pwndbg/__init__.py`, so `history expansion` cause the pain when we try to use `!command-string`:
截圖 2022-10-05 下午4 08 46

IMO, maybe we should `set history expansion off` explicitly instead of `set history expansion on` by default if we want to let user use `!cyclic`?

lebr0nli:
After this fix:

截圖 2022-10-05 下午4 37 16

### DIFF
--- a/pwndbg/__init__.py
+++ b/pwndbg/__init__.py
@@ -138,7 +138,7 @@ set confirm off
 set verbose off
 set pagination off
 set height 0
-set history expansion on
+set history expansion off
 set history save on
 set follow-fork-mode child
 set backtrace past-main on
--- a/pwndbg/lib/tips.py
+++ b/pwndbg/lib/tips.py
@@ -5,6 +5,7 @@ TIPS = [
     # GDB hints
     "GDB's `apropos <topic>` command displays all registered commands that are related to the given <topic>",
     "GDB's `follow-fork-mode` parameter can be used to set whether to trace parent or child after fork() calls",
+    "GDB's `history expansion` is been turn off to avoid conflicts with `!command-string` syntax, turn it on again with `set history expansion on` if you want to use it",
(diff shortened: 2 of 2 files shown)


disconnect3d:
Hmm so the idea of history expansion is so that we can do either `!<n>` or `!<part of command previously executed>` and it will find it in history and execute. And u can display the history with `show commands` command.

I think we should bring pwntools commands back into pwndbg and have a nicer UX/CLI for them and keep history expansion probably

lebr0nli:
> Hmm so the idea of history expansion is so that we can do either `!<n>` or `!<part of command previously executed>` and it will find it in history and execute. And u can display the history with `show commands` command.
> 
> I think we should bring pwntools commands back into pwndbg and have a nicer UX/CLI for them and keep history expansion probably

@disconnect3d, yep, if we want to keep `history expansion` on, then we should bring pwntools commands back into pwndbg for sure. (If not, we'll need `shell cyclic` which needs 6 more characters😂)
So you prefer to bring them back right? I'll close this PR.

lebr0nli:
Btw, whether we decide to set `history expansion` to on or off in the end, I think we should probably hint to the user about this behavior because I didn't know pwndbg and GDB have this feature, and I was bothered by these issues for a long time haha. And this is also one of the reasons that I decided to wrap the terrible console/prompt of GDB with my own plugin in the end XD

disconnect3d:
Yea I was also surprised about it tbh. I expected !pwn cyclic 10 to just
work but... history expansion breaks it badly :/

PR #1232, opened 2022-10-06
Add unstrip command

### LINKED ISSUE #1169: Command to unstrip libc
`pwntools` ha a `libcdb.unstrip` method now to unstrip libc files. Since we now have a dependency on pwntools, it would be convenient to have a command that could attempt to unstrip the currently loaded libc file.

gsingh93:
Use `pwntools` to unstrip the currently loaded libc.

Fixes #1169

### DIFF
--- a/pwndbg/commands/unstrip.py
+++ b/pwndbg/commands/unstrip.py
@@ -1,0 +1,28 @@
+import os.path
+
+from pwnlib.libcdb import unstrip_libc
+
+import pwndbg.commands
+from pwndbg.color import message
+from pwndbg.vmmap import info_sharedlibrary
+
+
+def _is_libc(path):
+    filename = os.path.basename(path)
+    # TODO: this will lead to false positives
+    if filename == "libc.so.6" or (filename.startswith("libc") and filename.endswith(".so")):
+        return True
+
+
+@pwndbg.commands.ArgparsedCommand("Unstrips the current libc")
+def unstrip():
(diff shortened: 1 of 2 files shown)


pwndbg/commands/unstrip.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+@pwndbg.commands.ArgparsedCommand("Unstrips the current libc")
+def unstrip():
+    mappings = info_sharedlibrary()
+    for mapping in mappings:
+        path = mapping.objfile
+        if _is_libc(path):
+            print(message.notice("Attempting to unstrip %s" % path))
+            if unstrip_libc(mapping.objfile):
```
disconnect3d, line 24: We can't do it like this. There are some issues:
1. What if the binary already has debug symbols? :D
2. It won't work with remote targets. You should use something like `pwndbg.file.get` to download the file from remote, save it locally and then unstrip.
3. The unstripping is done in-place and this is a problem. If we executed it against /lib/.../libc.so path, then we may break the user's system, not to take into account that we probably are missing permissions for that.
4. It does not load the symbols automagically, and it could?

I think we should make sure this works with remote gdbserver and also use other method from pwnlib to download the libc debug symbols and then we should load them with `add-symbol-file`.

gsingh93: Thanks for the feedback, this was exactly what I was looking for 😄. I'll close this one and come back to this later.

PR #1233, opened 2022-10-06
Add --shell argument to pwndbg command

### LINKED ISSUE #1211: Hide shell commands from the `pwndbg` output by default
We should hide shell commands from the `pwndbg` output by default, and only display them with `pwndbg --all` or `pwndbg --shell` or something like that. This would make it easier to find actual pwndbg commands.

gsingh93:
```
pwndbg: loaded 157 pwndbg commands and 49 shell commands. Type pwndbg [--shell | --all] [filter] for a list.
Use GDB's dprintf command to print all calls to given function. E.g. dprintf malloc, "malloc(%p)\n", (void*)$rdi will print all malloc calls
pwndbg> pwndbg
address              Print virtual memory map pages. Results can be filtered by providing address/module name.
arena                Print the contents of an arena, default to the current thread's arena.
...
pwndbg> pwndbg --shell
asm                  Invokes `asm` shell command
awk                  Invokes `awk` shell command (deprecated)
...
pwndbg> pwndbg --all
address              Print virtual memory map pages. Results can be filtered by providing address/module name.
arena                Print the contents of an arena, default to the current thread's arena.
arenas               List this process's arenas.
argc                 Prints out the number of arguments.
args                 Prints out the contents of argv.
argv                 Prints out the contents of argv.
aslr                 Check the current ASLR status, or turn it on/off.
asm                  Invokes `asm` shell command
```

Fixes #1211

### DIFF
--- a/pwndbg/commands/misc.py
+++ b/pwndbg/commands/misc.py
@@ -81,14 +81,16 @@ parser.add_argument(
 @pwndbg.commands.ArgparsedCommand(parser, command_name="pwndbg")
 def pwndbg_(filter_pattern, shell, all_):
     if all_:
-        shell = True
-        non_shell = True
+        shell_cmds = True
+        pwndbg_cmds = True
     elif shell:
-        non_shell = False
+        shell_cmds = True
+        pwndbg_cmds = False
     else:
-        non_shell = True
+        shell_cmds = False
+        pwndbg_cmds = True
 
-    for name, docs in list_and_filter_commands(filter_pattern, non_shell, shell):
+    for name, docs in list_and_filter_commands(filter_pattern, pwndbg_cmds, shell_cmds):
         print("%-20s %s" % (name, docs))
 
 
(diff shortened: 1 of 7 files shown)


pwndbg/commands/misc.py
```diff
@@ -74,8 +79,16 @@ def errno_(err):
 
 
 @pwndbg.commands.ArgparsedCommand(parser, command_name="pwndbg")
-def pwndbg_(filter_pattern):
-    for name, docs in list_and_filter_commands(filter_pattern):
+def pwndbg_(filter_pattern, shell, all_):
+    if all_:
+        shell = True
+        non_shell = True
```
disconnect3d: Hmm maybe we can find a better name for `non_shell`? Like `other`, `rest`, `standard`?

Or: `pwndbg_commands` and `shell_commands` or `shell_aliases`?

disconnect3d (approved):
Lgtm, though we may rename the `non_shell` to sth less confusing :P

gsingh93:
I realized I was spending too long trying to decide on names so I just went with something :P Fixed now.

PR #1234, opened 2022-10-06
Add `malloc_chunk` command tests for heuristic heap

CptGibbon:
The same tests applied to the `malloc_chunk` command, but with `set resolve-heap-via-heuristic on`.

Personally I don't like how Black has broken the visual flow here for the sake of a couple of characters.

### DIFF
--- a/tests/test_heap.py
+++ b/tests/test_heap.py
@@ -1,6 +1,5 @@
 import os
 import tempfile
-from cmath import exp
 
 import gdb
 import pytest
--- a/tests/test_heap.py
+++ b/tests/test_heap.py
@@ -1,6 +1,6 @@
-from cmath import exp
 import os
 import tempfile
+from cmath import exp
 
 import gdb
 import pytest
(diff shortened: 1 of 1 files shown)


gsingh93:
I think it would be better to pull out the common code from `test_malloc_chunk_command` and `test_malloc_chunk_command_heuristic` into another function, and then in both we can just set the necessary config options and call that common function.

gsingh93:
Another way to simplify both tests:
```python
    chunks = {}
    results = {}
    for name in ['allocated', 'tcache', 'fast', 'small', 'large', 'unsorted']:
        chunks[name] = pwndbg.heap.current.malloc_chunk(
            gdb.lookup_symbol("%s_chunk" % name)[0].value()
        )

        results[name] = gdb.execute("malloc_chunk %s_chunk" % name, to_string=True).splitlines()

```

gsingh93:
Looks much better, thanks!

Issue #1231, opened 2022-10-06
Heap heuristics may crash if libc.so is dynamically linked but no symbols are present for it


disconnect3d:
This is weird:
```
dc@jhtc:~$ gdb --quiet -ex 'set context-sections ""' -ex 'set auto-solib-add off' -ex 'set resolve-heap-via-heuristic on' -ex 'entry' /bin/ls
pwndbg: loaded 207 commands. Type pwndbg [filter] for a list.
pwndbg: created $rebase, $ida gdb functions (can be used with print/break)
Reading symbols from /bin/ls...
(No debugging symbols found in /bin/ls)
Sections set to be empty. FYI valid values are: regs, disasm, args, code, stack, backtrace, expressions, ghidra
You are going to resolve the heap via heuristic. This might not work in all cases.
You are going to resolve the heap via heuristic. This might not work in all cases.
Temporary breakpoint 1 at 0x555555559850
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".

Temporary breakpoint 1, 0x0000555555559850 in ?? ()
------- tip of the day (disable with set show-tips off) -------
Pwndbg context displays where the program branches to thanks to emulating few instructions into the future. You can disable this with set emulate off which may also speed up debugging
You are going to resolve the heap via heuristic. This might not work in all cases.
pwndbg> set exception-verbose on
Set whether to print a full stacktrace for exceptions raised in Pwndbg commands to True
pwndbg> heap
Traceback (most recent call last):
  File "/opt/pwndbg/pwndbg/commands/__init__.py", line 147, in __call__
    return self.function(*args, **kwargs)
  File "/opt/pwndbg/pwndbg/commands/__init__.py", line 218, in _OnlyWhenRunning
    return function(*a, **kw)
  File "/opt/pwndbg/pwndbg/commands/__init__.py", line 273, in _OnlyWithResolvedHeapSyms
    return function(*a, **kw)
  File "/opt/pwndbg/pwndbg/commands/__init__.py", line 243, in _OnlyWhenHeapIsInitialized
    return function(*a, **kw)
  File "/opt/pwndbg/pwndbg/commands/heap.py", line 126, in heap
    heap_region = allocator.get_heap_boundaries(addr)
  File "/opt/pwndbg/pwndbg/heap/ptmalloc.py", line 1717, in get_heap_boundaries
    arena = self.get_arena(addr)
[...]

PR #1242, opened 2022-10-07
Enhances the error handling of the heap heuristic

lebr0nli:
This PR enhances the error handling of the heap heuristics.
(The problems are mentioned in https://github.com/pwndbg/pwndbg/issues/1231#issuecomment-1269262842 and https://github.com/pwndbg/pwndbg/issues/1231#issuecomment-1269305915)

And this PR also tries to avoid using `_IO_list_all` to find other symbols if possible.


pwndbg/commands/__init__.py
```diff
@@ -271,27 +272,35 @@ def _is_statically_linked():
...
+                return function(*a, **kw)
+            except SymbolUnresolvableError as err:
+                e(f"{function.__name__}: Fail to resolve the symbol: `{err.symbol}`")
                 w(
-                    """Invoking the `set resolve-heap-via-heuristic on` command to resolve libc symbols via heuristics.
+                    f"You can try to determine the libc symbols addresses manually and set them appropriately. For this, see the `heap_config` command output and set the config about `{err.symbol}`."
+                )
+        else:
+            print(message.error(f"{function.__name__}: "), end="")
+            if not pwndbg.config.resolve_heap_via_heuristic:
```
disconnect3d, line 287: if resolve via heuristic is set we wont print anything after func name?

pwndbg/commands/__init__.py
```diff
@@ -301,19 +310,19 @@ def _OnlyWithResolvedHeapSyms(*a, **kw):
...
+                        "You can show your current config about heap by `heap_config`.\n"
                         "Then pwndbg will resolve some missing symbols via heuristics, but the results of those commands may be incorrect in some cases."
                     )
+            elif pwndbg.glibc.get_version() is None:
+                e("Can't resolve the heap since the GLIBC version is not set.")
+                w(
+                    "Please set the GLIBC version you think the target binary was compiled (using `set glibc <version>` command; e.g. 2.32) and re-run this command."
                 )
+            else:
+                e("An unknown error occurred when resolved the heap.")
```
disconnect3d, line 333: Shall we tell the user we couldn't find proper symbol addresses and they can set them?

pwndbg/heap/__init__.py
```diff
@@ -17,7 +17,7 @@
     "global_max_fast", "0", "&global_max_fast for heuristics", "heap"
 )
 
-symbol_list = pwndbg.config.get_params("heap")
+symbol_list = [local for local in locals().values() if isinstance(local, pwndbg.config.Parameter)]
```
disconnect3d: Is this just all above or is there more to it? Can we just list it manually? Doing it with `locals()` can be problematic with further code changes

pwndbg/heap/ptmalloc.py
```diff
@@ -697,6 +703,9 @@ def libc_has_debug_syms(self):
 
 
 class DebugSymsHeap(Heap):
+    def can_be_resolved(self):
+        return self.libc_has_debug_syms()
```
disconnect3d: Would this work?: `can_be_resolved = Heap.libc_has_debug_syms`

pwndbg/heap/ptmalloc.py
```diff
@@ -908,11 +907,14 @@ def is_glibc_static_symbol(self, addr):
         # Note: This might fail if there is an another symbol also call `_IO_list_all`...
         # TODO/FIXME: Find a better way to check if `addr` is a static symbol of GLIBC.
```
disconnect3d: btw what do we mean by 'addr being static symbol in libc'?

pwndbg/heap/ptmalloc.py
lebr0nli: Yep, it works. I'll change it to your suggestions.

pwndbg/heap/__init__.py
lebr0nli: > Is this just all above or is there more to it?

Yes, we only need above config for now.

> Can we just list it manually?

Yes. (I use `locals()` because we might need to modify this list every time we want to introduce a new symbol to let the user set it, so I want to avoid this, but seems this `locals()` solution is not good haha.)

> Doing it with `locals()` can be problematic with further code changes.

Ok, I'll just list it manually.

pwndbg/commands/__init__.py
lebr0nli: Yes, I'll add the hint for it also.

pwndbg/heap/ptmalloc.py
lebr0nli: That's mean I want to avoid finding wrong symbol with the same name, I mean:

If we compiled this file with `gcc FILENAME.c -g` or `gcc FILENAME.c -static -g`:

```c
#include <stdlib.h>
void *main_arena = 0xdeadbeaf;
int main(){
    free(malloc(0x20));
    return 0;
}
```

`pwndbg.symbol.address("main_arena")` will return the address of this 0xdeadbeaf `main_arena` instead of the "real" `main_arena` in libc, so the `heap` in debug symbols mode will fail.
But I want to avoid this and make it works in heuristic mode if possible, this why I have this weird check XD.

I'll check if `gdb.lookup_static_symbol` can help later.

CptGibbon:
Is it just me or does this make debugging a nightmare, since you can no longer use `set exception-verbose on`.

lebr0nli:
> Is it just me or does this make debugging a nightmare, since you can no longer use `set exception-verbose on`.

@CptGibbon Ahh, I'm sorry for that, maybe it's because I catch all the exceptions here:

https://github.com/pwndbg/pwndbg/blob/bfbb2b86525ce5c98ef912b47ba0fd75eab4eea7/pwndbg/commands/__init__.py#L286

I'll fix this, I'm really sorry for the inconvenience.

disconnect3d:
Huh, great catch folks! At first I didn't understand @CptGibbon issue as it worked for me but with... non-heap commands :P

PR #1243, opened 2022-10-07
Load commands and gdblib explicitly in __init__.py

gsingh93:
For historical reasons, the modules in `pwndbg/commands` and `pwndbg/gdblib` rely on executing code on import. This isn't great design, but we're stuck with it for now. At the very least, instead of importing a bunch of things in `pwndbg/__init__.py` that aren't actually used in that file, we import them in the specific `__init__.py` for that module inside a function, and then we just call that function to trigger the side effects of importing them.

The remaining modules in `pwndbg/__init__.py` may also be able to be removed in the future, but for now I'm just looking `gdblib` and `commands`.

Some motivations for this:
- This may make it easier to add unit tests, as we can hopefully just patch these two functions out
- It makes it clear in `__init_.py` what is actually required to be imported for the code in that file, as opposed to side effects
- If you ever want to know which files have side effects, you can look at these two functions to get the full list
- It's easier to figure out if an import is missing. For example, with this PR I added the `pwndbg.gdblib.ctypes` import, which was easy to spot with the reduced amount of imports.
- It makes refactoring easier in the future, as we just need to move around two function calls instead of a ton of imports
- This may make it easier to reload modules during development, but this is something I still need to investigate.


pwndbg/gdblib/__init__.py
```diff
@@ -4,3 +4,18 @@
 from pwndbg.gdblib.arch import arch
 
 __all__ = ["ctypes", "memory", "typeinfo"]
+
+
+# TODO: should the imports above be moved here?
```
gsingh93, line 9: Not sure what the best way to deal with this is. The above imports are there so that we can rename `pwndbg.gdblib.arch` to `arch_mod` and export the `pwndbg.gdblib.arch` object as `arch`. Maybe the right way to do it is to make them globals, and then assign to those globals inside `load_gdblib`?

disconnect3d (approved):

PR #1249, opened 2022-10-07
Move remote_files_dir into file.py

gsingh93:
I don't know if there was a reason why it was originally like this, but I've moved `remote_files_dir` into `file.py`. It doesn't make much sense for the `file` module to depend on the `symbol` module. `remote_files` can stay in `symbol.py`, as it seems that's for its own use.

### DIFF
--- a/pwndbg/symbol.py
+++ b/pwndbg/symbol.py
@@ -7,8 +7,6 @@ information available.
 """
 import os
 import re
-import shutil
-import tempfile
 
 import elftools.common.exceptions
 import elftools.elf.constants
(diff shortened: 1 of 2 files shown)


pwndbg/symbol.py
```diff
@@ -87,12 +79,10 @@ def autofetch():
     if pwndbg.gdblib.android.is_android():
         return
 
-    if not remote_files_dir:
-        remote_files_dir = tempfile.mkdtemp()
+    remote_files_dir = pwndbg.file.remote_files_dir()
+    if remote_files_dir not in get_directory().split(":"):
```
disconnect3d, line 83: isn't this logic different?

gsingh93: With the old logic, `symbol.py` could check if the global variable was `None` to determine whether it should add it to `debug-file-directory`. But now that the global is in `file.py`, to do the same thing I'd need some other global variable to check. Instead, I just check if we've already added it to `debug-file-directory`, and if not, then we add it. If you have other suggestions for how to rewrite this, let me know.

disconnect3d: one thing to consider: what if the directory has ':' in it?

disconnect3d: or maybe it doesnt matter much because we use tempfiles?

gsingh93: Right, we don't need to worry about any weird directory names because we use tempfiles.

PR #1251, opened 2022-10-08
Cleanup dependencies

gsingh93:
Remove dependencies we're not actually using from `requirements.txt`. Make some dependencies that were optional in the code, no longer optional, as they should always be installed (`psutil` and `Pygments`). The last remaining dependencies that are "optional" in the code are `unicorn` and `r2pipe`, and I'm not sure if those were intentionally made optional so I've left that in for now.

### DIFF
--- a/pwndbg/lib/memoize.py
+++ b/pwndbg/lib/memoize.py
@@ -6,17 +6,11 @@ new library/objfile are loaded, etc.
 
 import functools
 import sys
+from collections.abc import Hashable
 from typing import Any
 from typing import Callable
 from typing import Dict
 from typing import List
-
-try:
-    # Python >= 3.10
-    from collections.abc import Hashable
-except ImportError:
-    # Python < 3.10
-    from collections import Hashable  # type: ignore
 
 debug = False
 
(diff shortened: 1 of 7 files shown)


pwndbg/lib/memoize.py
```diff
@@ -6,18 +6,12 @@
...
 from typing import Dict
 from typing import List
 
-try:
-    # Python >= 3.10
-    from collections.abc import Hashable
-except ImportError:
-    # Python < 3.10
-    from collections import Hashable  # type: ignore
-
```
gsingh93, line 20: I see this was added in https://github.com/pwndbg/pwndbg/pull/960. But it seems to be working for me without it?

gsingh93: Oh, I think I understand. The original fix could just have been `from collections.abc import Hashable`, as this should work for Python versions all the way to Python 3.3.

disconnect3d:
Not sure about r2pipe, but I think we had some issues in the past with unicorn requiring to be compiled or sth and maybe that was it why it was optional...

I think its fine to make it mandatory for now.

PR #1259, opened 2022-10-09
GDB Refactor [21/N]: Move symbol.py to gdblib

gsingh93:
A couple of changes to symbol.py other than the move:
- Private methods/variables now start with an underscore, making it easier to understand to understand methods/variables are being accessed from outside the module
- Added a few type annotations
- Fixed an issue where we were passing a byte string to `pwndbg.symbol.address`
- Removed the handling of integers or strings representing integers from `pwndbg.symbol.address`. I didn't see any usages of this in the codebase, and allowing so many arbitrary formats will lead to bugs
- Exceptions from `gdb.lookup_symbol` are no longer ignored, except for the one case of "No frame selected" errors. This is only occurring in the Go typeinfo tests, and I think the exception probably makes sense, but I want to dig into it a bit more and confirm the right fix is to use `gdb.selected_frame()` first.
- Removed the legacy way of looking up symbols with `info address`


gsingh93:
@CptGibbon @lebr0nli this PR is surfacing any exceptions that were previously caught when doing symbol lookup, and only ignoring a small set of them. On Ubuntu 18.04 and 20.04 (but not 22.04), I see the following exception in `test_malloc_chunk_command_heuristic`:
image

Do either of you know why that's happening? What's changed in `libc` since 20.04 that would result in this?

CptGibbon:
Looks like the test binary needs to be linked against libpthread, I can submit a PR

pwndbg/gdblib/symbol.py
```diff
@@ -184,40 +188,36 @@ def get(address, gdb_only=False):
...
-        return address
-
-    except gdb.error:
-        return None
+    except gdb.error as e:
+        # Symbol lookup only throws exceptions on errors, not if it failed to
+        # lookup a symbol. We want to raise these errors so we can handle them
+        # properly, but there are some we haven't figured out how to fix yet, so
+        # we ignore those here
+        skipped_exceptions = []
```
disconnect3d, line 204: Should be:

```py
skipped_exceptions = (
    # comment here
    "No frame selected",
    ...
)
```
but this can be changed in a next PR as well

PR #1262, opened 2022-10-09
Remove capstone wildcard imports

gsingh93:
Wildcard imports can cause various problems, but in this particular case it's causing problems with type checking and other static analysis tools (`vermin`).

### DIFF
--- a/pwndbg/disasm/arch.py
+++ b/pwndbg/disasm/arch.py
@@ -1,11 +1,12 @@
+import capstone as C
 import gdb
-from capstone import *  # noqa: F403
 
 import pwndbg.gdblib.symbol
 import pwndbg.lib.memoize
 
 debug = False
 
+# TODO: Do these need to change to 'C.CS_'?
 groups = {v: k for k, v in globals().items() if k.startswith("CS_GRP_")}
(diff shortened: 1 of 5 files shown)


pwndbg/disasm/arch.py
```diff
@@ -1,11 +1,12 @@
...
 
 import pwndbg.lib.memoize
 import pwndbg.symbol
 
 debug = False
 
+# TODO: Do these need to change to 'C.CS_'?
 groups = {v: k for k, v in globals().items() if k.startswith("CS_GRP_")}
 ops = {v: k for k, v in globals().items() if k.startswith("CS_OP_")}
 access = {v: k for k, v in globals().items() if k.startswith("CS_AC_")}
```
gsingh93, line 12: With this PR, these arrays are now empty, but no tests are failing. That means we need to improve our test coverage first before merging this PR.

gsingh93: @disconnect3d would you be able to help with testing this?

disconnect3d: Yeah, I can look at it probably during the weekend after ASIS CTF

PR #1266, opened 2022-10-10
Develop Arena class

CptGibbon:
Add `Arena.top` and `Arena.is_main_arena`.
Add `Chunk.arena` and `Chunk.is_top_chunk`.

Above allows less complexity in `malloc_chunk` command.
Arena resolution via `Chunk.arena` is more accurate than existing `get_arena_for_chunk()` solution.

### DIFF
--- a/pwndbg/heap/ptmalloc.py
+++ b/pwndbg/heap/ptmalloc.py
@@ -32,6 +32,24 @@ def heap_for_ptr(ptr):
 
 
 class Chunk:
+    __slots__ = (
+        "_gdbValue",
+        "address",
+        "_prev_size",
+        "_size",
+        "_real_size",
+        "_flags",
+        "_non_main_arena",
+        "_is_mmapped",
+        "_prev_inuse",
+        "_fd",
+        "_bk",
+        "_fd_nextsize",
+        "_bk_nextsize",
+        "_arena",
+        "_is_top_chunk",
+    )
+
     def __init__(self, addr, arena=None):
         if isinstance(pwndbg.heap.current.malloc_chunk, gdb.Type):
             self._gdbValue = pwndbg.gdblib.memory.poi(pwndbg.heap.current.malloc_chunk, addr)
@@ -51,7 +69,6 @@ class Chunk:
         self._bk_nextsize = None
         self._arena = arena
         self._is_top_chunk = None
-
 
     # Some chunk fields were renamed in GLIBC 2.25 master branch.
     def __match_renamed_field(self, field):
@@ -206,6 +223,8 @@ class Chunk:
 
 
 class Arena:
+    __slots__ = ("_gdbValue", "address", "is_main_arena", "_top", "heaps")
+
     def __init__(self, addr, heaps=None):
         if isinstance(pwndbg.heap.current.malloc_state, gdb.Type):
             self._gdbValue = pwndbg.gdblib.memory.poi(pwndbg.heap.current.malloc_state, addr)
--- a/pwndbg/heap/ptmalloc.py
+++ b/pwndbg/heap/ptmalloc.py
@@ -246,11 +246,10 @@ class Arena:
         return self._top
 
     def __str__(self):
-        res = []
         prefix = "[%%%ds]    " % (pwndbg.gdblib.arch.ptrsize * 2)
         prefix_len = len(prefix % (""))
         arena_name = "main" if self.is_main_arena else hex(self.address)
-        res.append(message.hint(prefix % (arena_name)) + str(self.heaps[0]))
+        res = [message.hint(prefix % (arena_name)) + str(self.heaps[0])]
         for h in self.heaps[1:]:
             res.append(" " * prefix_len + str(h))
 
--- a/pwndbg/heap/ptmalloc.py
+++ b/pwndbg/heap/ptmalloc.py
@@ -52,7 +52,6 @@ class Chunk:
         self._arena = arena
         self._is_top_chunk = None
 
-        # TODO key, REVEAL_PTR
 
     # Some chunk fields were renamed in GLIBC 2.25 master branch.
     def __match_renamed_field(self, field):
@@ -204,8 +203,6 @@ class Chunk:
                 self._is_top_chunk = False
 
         return self._is_top_chunk
-
-    # TODO Other useful methods e.g. next_chunk(), __iter__, __str__
 
 
 class Arena:
--- a/pwndbg/heap/ptmalloc.py
+++ b/pwndbg/heap/ptmalloc.py
@@ -32,7 +32,7 @@ def heap_for_ptr(ptr):
 
 
 class Chunk:
-    def __init__(self, addr):
+    def __init__(self, addr, arena=None):
         if type(pwndbg.heap.current.malloc_chunk) == gdb.Type:
             self._gdbValue = pwndbg.gdblib.memory.poi(pwndbg.heap.current.malloc_chunk, addr)
         else:
@@ -49,6 +49,8 @@ class Chunk:
         self._bk = None
         self._fd_nextsize = None
         self._bk_nextsize = None
+        self._arena = arena
+        self._is_top_chunk = None
 
         # TODO key, REVEAL_PTR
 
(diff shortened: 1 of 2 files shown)


pwndbg/heap/ptmalloc.py
```diff
@@ -177,21 +179,61 @@ def bk_nextsize(self):
...
+        return self._is_top_chunk
+
     # TODO Other useful methods e.g. next_chunk(), __iter__, __str__
 
 
 class Arena:
-    def __init__(self, addr, heaps):
-        self.addr = addr
+    def __init__(self, addr, heaps=None):
+        if type(pwndbg.heap.current.malloc_state) == gdb.Type:
```
```suggestion
        if isinstance(pwndbg.heap.current.malloc_state, gdb.Type):
```

pwndbg/heap/ptmalloc.py
```diff
@@ -177,21 +179,61 @@ def bk_nextsize(self):
...
+
+        return self._is_top_chunk
+
     # TODO Other useful methods e.g. next_chunk(), __iter__, __str__
 
 
 class Arena:
-    def __init__(self, addr, heaps):
-        self.addr = addr
+    def __init__(self, addr, heaps=None):
```
disconnect3d, line 228:
```suggestion
    __slots__ = ("_gdbValue", "address", "is_main_arena", "_top", "heaps")

    def __init__(self, addr, heaps=None):
```

disconnect3d: `__slots__` is an optimization, as per https://docs.python.org/3/reference/datamodel.html#slots:
> __slots__ allow us to explicitly declare data members (like properties) and deny the creation of [__dict__](https://docs.python.org/3/library/stdtypes.html#object.__dict__) and __weakref__ (unless explicitly declared in __slots__ or available in a parent.)
> 
> The space saved over using [__dict__](https://docs.python.org/3/library/stdtypes.html#object.__dict__) can be significant. Attribute lookup speed can be significantly improved as well.

With slots we won't be able to add more fields in runtime to an instance of Arena, but I don't think we need it.

pwndbg/heap/ptmalloc.py
```diff
@@ -32,7 +32,7 @@ def heap_for_ptr(ptr):
 
 
 class Chunk:
-    def __init__(self, addr):
+    def __init__(self, addr, arena=None):
```
disconnect3d, line 53:
```suggestion
    __slots__ = ("_gdbValue", "address", "_prev_size", "_size", "_real_size", "_flags", "_non_main_arena", "_is_mmapped", "_prev_inuse", "_fd", "_bk", "_fd_nextsize", "_bk_nextsize", "_arena", "_is_top_chunk")

    def __init__(self, addr, arena=None):
```

disconnect3d: (I've commented below what slots does)

This also needs additional lint pass

pwndbg/heap/ptmalloc.py
```diff
@@ -177,21 +179,61 @@ def bk_nextsize(self):
...
+
     def __str__(self):
         res = []
         prefix = "[%%%ds]    " % (pwndbg.gdblib.arch.ptrsize * 2)
         prefix_len = len(prefix % (""))
-        arena_name = (
-            hex(self.addr) if self.addr != pwndbg.heap.current.main_arena.address else "main"
-        )
+        arena_name = "main" if self.is_main_arena else hex(self.address)
         res.append(message.hint(prefix % (arena_name)) + str(self.heaps[0]))
```
disconnect3d: This is picky but if we are here, we could change this to:

```suggestion
        res = [message.hint(prefix % (arena_name)) + str(self.heaps[0])]
```

pwndbg/heap/ptmalloc.py
```diff
@@ -49,6 +49,8 @@ def __init__(self, addr):
         self._bk = None
         self._fd_nextsize = None
         self._bk_nextsize = None
+        self._arena = arena
+        self._is_top_chunk = None
 
         # TODO key, REVEAL_PTR
```
disconnect3d: If this TODO is still relevant, can we extend it to include more about what is it for/about or what we would like to achieve with it? :P

pwndbg/heap/ptmalloc.py
CptGibbon: Yeah I can apply `__slots__` to the `Chunk` class too if you want?

pwndbg/heap/ptmalloc.py
```diff
@@ -49,8 +67,8 @@ def __init__(self, addr):
         self._bk = None
         self._fd_nextsize = None
         self._bk_nextsize = None
-
-        # TODO key, REVEAL_PTR
+        self._arena = arena
+        self._is_top_chunk = None
```
gsingh93, line 71: Minor point, we should probably default boolean fields to `True` or `False`, and in this case `False`. It will help with type checking later.

pwndbg/heap/ptmalloc.py
```diff
@@ -177,22 +195,61 @@ def bk_nextsize(self):
...
         return self._bk_nextsize
 
-    # TODO Other useful methods e.g. next_chunk(), __iter__, __str__
+    @property
+    def arena(self):
+        if self._arena is None:
+            try:
+                ar_ptr = pwndbg.heap.current.get_heap(self.address)["ar_ptr"]
+                ar_ptr.fetch_lazy()
+            except Exception:
```
gsingh93, line 204: Do you know why this might fail? If we know the exception is a `gdb.Error`, I think we should start explicitly catching those instead of using `Exception`. If we're not sure, it's fine to leave this.

pwndbg/heap/ptmalloc.py
```diff
@@ -177,22 +195,61 @@ def bk_nextsize(self):
...
+                ar_ptr = None
+            if ar_ptr is not None and ar_ptr in (ar.address for ar in pwndbg.heap.current.arenas):
```
gsingh93, line 206: I guess unless we expect `None` to appear as `ar.address`, we can skip the `ar_ptr is not None` check. But it's fine if you want to keep it for clarity.

pwndbg/heap/ptmalloc.py
```diff
@@ -177,22 +195,61 @@ def bk_nextsize(self):
...
 
-    # TODO Other useful methods e.g. next_chunk(), __iter__, __str__
+    @property
+    def arena(self):
+        if self._arena is None:
+            try:
+                ar_ptr = pwndbg.heap.current.get_heap(self.address)["ar_ptr"]
+                ar_ptr.fetch_lazy()
+            except Exception:
+                ar_ptr = None
```
gsingh93, line 205: I think it might be cleaner if we do `ar_ptr = pwndbg.heap.current.main_arena.address` and get rid of the `else` case below. What do you think?

gsingh93 (commented):
Overall looks good. Are there any parts of this you think should have more test coverage? It would probably be best to add that in now before moving on to other refactoring work.

CptGibbon:
I added the `malloc_chunk` tests last week, and since this doesn't affect anything outside of that I can't think of any more relevant tests. I do intend to add more heap tests as we continue though 👍

gsingh93:
Do those tests cover chunks that aren't located in the non-main arena? Seems like there's some logic here about that that should get a bit of coverage.

CptGibbon:
No these tests need to be replicated with multiple arenas. I'll add them in a separate PR 👍

pwndbg/heap/ptmalloc.py
disconnect3d: I'm not sure about that, with the current approach we will initialize a given field on its first use

disconnect3d (approved):

disconnect3d:
For what is worth, if those properties are computed once and never changed later on, we could potentially use `functools.cached_property` for them:
image

However, the [`cached_property`](https://docs.python.org/3/library/functools.html#functools.cached_property) was added in Python 3.8, so we could maybe do something like "use it if it is present otherwise use @property and just do not cache"?

disconnect3d:
OTOH, the cached property does not work if we do not have `__dict__` on the object, which is what `__slots__` does to the object:
image

But, we can specify `__dict__` in slots and then things start to work:
image

FWIW seeing the cached_property implementation, I think it may be possible to do better... but I guess it is fine as it is for a general use:
image

disconnect3d:
Merging this for now as I think it is fine rn

PR #1270, opened 2022-10-11
Don't catch the error of the heap commands when `set exception-* on`

lebr0nli:
Fix the nightmare when debugging the heap commands (https://github.com/pwndbg/pwndbg/pull/1242#issuecomment-1274039275)

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -15,6 +15,7 @@ import pwndbg.exception
 import pwndbg.gdblib.memory
 import pwndbg.gdblib.regs
 import pwndbg.gdblib.symbol
+import pwndbg.heap
 import pwndbg.hexdump
 import pwndbg.ui
 from pwndbg.heap.ptmalloc import SymbolUnresolvableError
@@ -276,6 +277,9 @@ def OnlyWithResolvedHeapSyms(function):
         e = lambda s: print(message.error(s))
         w = lambda s: print(message.warn(s))
         if pwndbg.heap.current.can_be_resolved():
+            # We don't catch the error for developers, only catch for users.
+            if gdb.parameter("exception-verbose") or gdb.parameter("exception-debugger"):
+                return function(*a, **kw)
             try:
                 return function(*a, **kw)
             except SymbolUnresolvableError as err:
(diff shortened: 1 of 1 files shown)


lebr0nli:
@CptGibbon, could you verify whether this fix works or not? Thanks!

pwndbg/commands/__init__.py
```diff
@@ -276,6 +277,9 @@ def _OnlyWithResolvedHeapSyms(*a, **kw):
         e = lambda s: print(message.error(s))
         w = lambda s: print(message.warn(s))
         if pwndbg.heap.current.can_be_resolved():
+            # We don't catch the error for developers, only catch for users.
+            if gdb.parameter("exception-verbose") or gdb.parameter("exception-debugger"):
```
disconnect3d: We should:
* fetch those through pwndbg.config.exception_verbose and pwndbg.config.exception_debugger
* move this to the exception case and re-raise the err `except Exception as err: if ...: raise err`

disconnect3d (changes requested):
see comment

pwndbg/commands/__init__.py
lebr0nli: Ok, I'll fix it with your suggestions!

CptGibbon:
I'll take a look today, thanks for changing this.

CptGibbon:
This does indeed print out a backtrace when `set exception-verbose on` is used 👍

pwndbg/commands/__init__.py
```diff
@@ -276,21 +277,26 @@ def _OnlyWithResolvedHeapSyms(*a, **kw):
...
             except Exception as err:
-                e(f"{function.__name__}: Unknown error: `{err}` when running this symbols.")
+                e(f"{function.__name__}: An unknown error occurred when running this command.")
                 if pwndbg.config.resolve_heap_via_heuristic:
                     w(
                         "Maybe you can try to determine the libc symbols addresses manually, set them appropriately and re-run this command. For this, see the `heap_config` command output and set the `main_arena`, `mp_`, `global_max_fast`, `tcache` and `thread_arena` addresses."
                     )
                 else:
                     w("You can try `set resolve-heap-via-heuristic on` and re-run this command.\n")
+                if pwndbg.config.exception_verbose.value or pwndbg.config.exception_debugger.value:
```
```suggestion
                if pwndbg.config.exception_verbose or pwndbg.config.exception_debugger:
```

disconnect3d: The `pwndbg.config.Parameter` class has a `__bool__` magic method/operator defined:
```py
    def __bool__(self):
        return bool(self.value)
```

so we can omit `.value` here :)

disconnect3d (approved):

Issue #1271, opened 2022-10-11
`pwndbg.gdblib.symbol.address` fail to handle the thread-local variables


lebr0nli:
<img width="1289" alt="截圖 2022-10-11 下午4 46 56" src="https://user-images.githubusercontent.com/61896187/195043471-185d98a1-0df4-43af-ba68-54f8e0853e31.png">

`pwndbg.gdblib.symbol.address('tcache')` will get error instead of return `None`.

Seems like it's because the error starts with `Cannot find thread-local storage` instead of `Cannot find thread-local variables`, so the following line fails to skip it:

https://github.com/pwndbg/pwndbg/blob/478a569cb3bef11f7011cd0a2f374d53ce5997fd/pwndbg/gdblib/symbol.py#L214-L217

cc: @gsingh93

PR #1272, opened 2022-10-11
Fix the error when trying to read thread-local variables

lebr0nli:
This PR fix the typo in the check at:

https://github.com/pwndbg/pwndbg/blob/478a569cb3bef11f7011cd0a2f374d53ce5997fd/pwndbg/gdblib/symbol.py#L214-L217

See #1271

### DIFF
--- a/pwndbg/gdblib/symbol.py
+++ b/pwndbg/gdblib/symbol.py
@@ -211,7 +211,7 @@ def address(symbol: str) -> int:
         # If we try to look up a TLS variable when there is no TLS, this
         # exception occurs. Ideally we should come up with a way to check for
         # this case before calling `gdb.lookup_symbol`
-        skipped_exceptions.append("Cannot find thread-local variables")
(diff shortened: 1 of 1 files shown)


pwndbg/gdblib/symbol.py
```diff
@@ -211,7 +211,7 @@ def address(symbol: str) -> int:
         # If we try to look up a TLS variable when there is no TLS, this
         # exception occurs. Ideally we should come up with a way to check for
         # this case before calling `gdb.lookup_symbol`
-        skipped_exceptions.append("Cannot find thread-local variables")
+        skipped_exceptions.append("Cannot find thread-local")
```
lebr0nli, line 214: @gsingh93, will it have any problem if we check `"Cannot find thread-local"` instead of `"Cannot find thread-local variables"`?

disconnect3d: I guess it should work fine, merging for now, @gsingh93 let us know if u see any issues with that

PR #1273, opened 2022-10-12
Bug fix for the heuristics of `main_arena` and `mp_`

lebr0nli:
If we can't find the `main_arena` or `mp_` via parsing the assembly code of `malloc_trim` or `__libc_free`, the heuristics will try to search `main_arena`'s or `mp_`'s struct in memory, but there are some bugs currently.

This PR fix these bugs when searching the structs in memory.

### DIFF
--- a/pwndbg/heap/ptmalloc.py
+++ b/pwndbg/heap/ptmalloc.py
@@ -1031,7 +1031,7 @@ class HeuristicHeap(Heap):
                             tmp_arena = self.malloc_state(tmp_next)
                             try:
                                 tmp_next = int(tmp_arena["next"])
-                            except gdb.MemoryError:
+                            except (gdb.MemoryError, gdb.error, OverflowError):
                                 # tmp_arena->next is not valid, break
                                 break
                         if found:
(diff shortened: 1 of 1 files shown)


pwndbg/heap/ptmalloc.py
```diff
@@ -1031,7 +1031,7 @@ def main_arena(self):
                             tmp_arena = self.malloc_state(tmp_next)
                             try:
                                 tmp_next = int(tmp_arena["next"])
-                            except gdb.MemoryError:
```
disconnect3d, line 1034: Any chance we can elaborate more why we do catch all of those exceptions? Can we add comments that will mention when they occur?

Also any chance we can have tests that trigger those behaviors? 😅

disconnect3d: I will merge this for now, but pls consider sending another PR with additional comments or/and tests :P

lebr0nli: > Can we add comments that will mention when they occur?

Ahh, I'm sorry about that, I'll open another PR with additional comments about this.

Issue #1072, opened 2022-08-17
vis_heap_chunks: paginate results


arcz:
### Description
After heap chunk size if corrupted to a large size `vis_heap_chunks` will print a ton of output. Paginating the results will help to bring back usefulness of the command in such situation.

### Steps to reproduce
1. Corrupt heap chunk with big chunk size
2. Run `vis_heap_chunks`

disconnect3d:
@arcz any chance you can provide a minimal PoC, e.g. one written in C so we can make this A) easier to develop, B) actually a test case in our test suite? :D

gsingh93:
I have multiple PoCs for this, since it pretty much happens with every heap pwnable :) It's easy to reproduce, but I'm not sure exactly how to fix this. Should we use GDB's own pagination or implement something ourselves?

It's be nice to fix this in a way where we can make any command paginate in the future, in case they have similar issues. Or allow the user to customize which commands get paginated (sometimes I think it would be good to paginate `heap`).

disconnect3d:
There are two options:
1) use GDBs pagination
2) use GDB command's repeat feature, when you hit <enter> to repeat the command and get its output for e.g. further addresses. We have it implemented in our `ArgparsedCommand` too and it works for `hexdump`, `telescope` and few more.

I think we could just go with 2)? I think that 1) does not stop processing code and only prevents from printing stuff?

Alternatively, we could set pagination only for this command to prevent way too long output and have it configurable.

gsingh93:
I'm fine with option 2, but there's a few things to note:
1. This might be a little more tricky than the other implementations. Generally, we always want to stop printing at a chunk boundary unless we hit some massive chunk, and at that point would we switch to printing a fixed set of address, or one chunk at a time, or something else? This won't be as simple as it is for `telescope`.
2. I didn't know about this repeat feature. We need to work on the discoverability of that somehow, and I'm not sure if tips are enough.
3. It would be good to implement this in a more generic way, where I could make any command into a repeatable command with just a decorator or something like that. I see for `telescope` and `hexdump`, we're essentially copying and pasting the same code with different variable names. I'm happy to help with this if we go this route.

Given all of that, it may be best to go with option 1. Ideally we could do something like `set paginated-commands 'vis_heap_chunks heap'`, so the user could configure pagination to be enabled for any command, and I don't think this would be too hard to implement.

lonnywong:
I want to make a feature for `vis_heap_chunks` command.

[...]

PR #1275, opened 2022-10-12
add max display size for vis_heap_chunks

lonnywong:
https://github.com/pwndbg/pwndbg/issues/1072

A test case:
```c
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

int main() {
  void *p, *q;
  p = malloc(0x10);
  free(p);
  p = malloc(0x2000);
  q = malloc(0x100);
  free(p);
  free(q);

  asm("int $3");

  p = malloc(0x20);
  memset(p, 'A', 0x30);

  asm("int $3");

  return 0;
}
```

```sh
gcc -m64 -g -o heap64 heap.c
gcc -m32 -g -o heap32 heap.c
```

```
pwndbg> set max-visualize-chunk-size 0x500
pwndbg> vis
```

image

image

### DIFF
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -594,6 +594,12 @@ def find_fake_fast(addr, size=None):
                     malloc_chunk(start + offset - psize, fake=True)
 
 
+pwndbg.config.Parameter(
+    "max-visualize-chunk-size",
+    0,
+    "max display size for heap chunks visualization (0 for display all)",
+)
+
 parser = argparse.ArgumentParser()
 parser.description = "Visualize chunks on a heap, default to the current arena's active heap."
 parser.add_argument(
@@ -611,13 +617,20 @@ parser.add_argument(
     default=False,
     help="Attempt to keep printing beyond the top chunk.",
 )
+parser.add_argument(
+    "--display_all",
+    "-a",
+    action="store_true",
+    default=False,
+    help="Display all the chunk contents (Ignore the `max-visualize-chunk-size` configuration).",
+)
 
 
 @pwndbg.commands.ArgparsedCommand(parser)
 @pwndbg.commands.OnlyWhenRunning
 @pwndbg.commands.OnlyWithResolvedHeapSyms
 @pwndbg.commands.OnlyWhenHeapIsInitialized
-def vis_heap_chunks(addr=None, count=None, naive=None):
+def vis_heap_chunks(addr=None, count=None, naive=None, display_all=None):
     """Visualize chunks on a heap, default to the current arena's active heap."""
     allocator = pwndbg.heap.current
     heap_region = allocator.get_heap_boundaries(addr)
@@ -704,10 +717,35 @@ def vis_heap_chunks(addr=None, count=None, naive=None):
 
     cursor = cursor_backup
 
+    has_huge_chunk = False
+    # round up to align with 4*ptr_size and get half
+    half_max_size = (
+        pwndbg.lib.memory.round_up(pwndbg.config.max_visualize_chunk_size, ptr_size << 2) >> 1
+    )
+
     for c, stop in enumerate(chunk_delims):
         color_func = color_funcs[c % len(color_funcs)]
 
+        if stop - cursor > 0x10000:
+            has_huge_chunk = True
+        first_cut = True
+        # round down to align with 2*ptr_size
+        begin_addr = pwndbg.lib.memory.round_down(cursor, ptr_size << 1)
+        end_addr = pwndbg.lib.memory.round_down(stop, ptr_size << 1)
+
         while cursor != stop:
+            # skip the middle part of a huge chunk
+            if (
+                not display_all
+                and half_max_size > 0
+                and begin_addr + half_max_size <= cursor < end_addr - half_max_size
+            ):
+                if first_cut:
+                    out += "\n" + "." * len(hex(cursor))
+                    first_cut = False
+                cursor += ptr_size
+                continue
+
             if printed % 2 == 0:
                 out += "\n0x%x" % cursor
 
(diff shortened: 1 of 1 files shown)


CptGibbon:
I love it ❤️
I'll test it out today

pwndbg/commands/heap.py
```diff
@@ -704,10 +717,33 @@ def vis_heap_chunks(addr=None, count=None, naive=None):
 
     cursor = cursor_backup
 
+    has_huge_chunk = False
+    # round up to align with 4*ptr_size and get half
+    half_max_size = ((max_visualize_chunk_size - 1) // (ptr_size << 2) + 1) * (ptr_size << 2) // 2
```
gsingh93: Can we use `pwndbg.lib.memory.align_up` here?

gsingh93: Can you use `pwndbg.gdblib.config.max_visualize_chunk_size` here? It'll be the same thing, but if we ever move the definition of the config parameter above into a different file, this won't break.

pwndbg/commands/heap.py
```diff
@@ -704,10 +717,33 @@ def vis_heap_chunks(addr=None, count=None, naive=None):
...
+
     for c, stop in enumerate(chunk_delims):
         color_func = color_funcs[c % len(color_funcs)]
 
+        if stop - cursor > 0x10000:
```
gsingh93, line 729: I'm not too familiar with this code, but what is this `0x10000` constant for exactly?

pwndbg/commands/heap.py
```diff
@@ -704,10 +717,33 @@ def vis_heap_chunks(addr=None, count=None, naive=None):
...
+            has_huge_chunk = True
+        first_cut = True
+        # round down to align with 2*ptr_size
+        begin_addr = cursor // (ptr_size << 1) * (ptr_size << 1)
```
gsingh93: We can use `pwndbg.lib.memory.align_down` or `pwndbg.lib.memory.round_down` here, and in the line below.

pwndbg/commands/heap.py
```diff
@@ -611,13 +617,20 @@ def find_fake_fast(addr, size=None):
...
+    help="Display all the chunk contents (Ignore the `max-visualize-chunk-size` configuration).",
+)
 
 
 @pwndbg.commands.ArgparsedCommand(parser)
 @pwndbg.commands.OnlyWhenRunning
 @pwndbg.commands.OnlyWithResolvedHeapSyms
 @pwndbg.commands.OnlyWhenHeapIsInitialized
-def vis_heap_chunks(addr=None, count=None, naive=None):
+def vis_heap_chunks(addr=None, count=None, naive=None, all=None):
```
gsingh93: `all` is a Python built-in, the linter will reject this in the near future. Can you rename to something else? If `all` is the best name, I'm fine with `all_`.

gsingh93 (changes requested):
I still think pagination will cover 100% of the cases we care about where this will only cover 95% of them (i.e. when you want to use `--all` to see some data in an allocated chunk, but due to corruption the top chunk is massive), but I do think an improvement in 95% of cases is worth merging.

Could you update the `vis_heap_chunk` tests to test this new functionality?

Another suggestion is: can we cut out only parts that repeat? i.e. 64 bytes of zeros would get cut out to just 16 bytes, like the hexdump and telescope commands do. That should solve the length problem and also make sure we don't miss any interesting data in the middle of a large chunk. I'd be hesitant to set the `all` flag to False by default before we implement something like this.

CptGibbon:
Having tested this briefly I think it's great ⚡
It solves the issue of excessive `vis_heap_chunks` output in a user-friendly way and I like how it prints how much you specify. This is superior to pagination in my personal opinion, since pagination doesn't help when you care about chunks after larger chunks.

I'll leave it to others to review any Python-related stuff, but from a heap command perspective I'm convinced.

gsingh93:
I wonder if we can make the `hexdump` code more flexible, so it can be reused for this command and implement the suggestion I pointed out, with only cutting out repeated characters? I can dig into it.

gsingh93:
We can get `hexdump` close to the output of `vis_heap_chunks` by just tweaking some configuration options:

image

If we can figure out a way to get the custom colors for each chunk and the bin labels integrated in, we can remove a lot of code from `vis_heap_chunks`.

pwndbg/commands/heap.py
lonnywong: `stop - cursor` is the current chunk size.

pwndbg/commands/heap.py
lonnywong: The `round_up` and `round_down` should work.

pwndbg/commands/heap.py
lonnywong: I changed them to `pwndbg.lib.memory.round_up` and `pwndbg.config. max_visualize_chunk_size`.

pwndbg/commands/heap.py
lonnywong: changed to `display_all`

gsingh93:
Thanks! You can fix the lint issues with `./lint.sh -f`. This looks pretty much ready to merge, but could you also update `pwndbg/tests/heap/test_vis_heap_chunks.py` to test this functionality? I think that none of the chunks in that test are large enough to trigger this code, so if you could add at least one chunk to that that's large enough to truncate the chunk, that should be good.

lonnywong:
> Could you update the vis_heap_chunk tests to test this new functionality?

I'd love to add some tests, but don't have much time to do it right now. Adding it to my todo list.

I may add 2 test cases in the next two days.

> Another suggestion is: can we cut out only parts that repeat?

The idea looks great.

> You can fix the lint issues with `./lint.sh -f`.

Thanks for pointing it out.

gsingh93:
> I may add 2 test cases in the next two days.

Sounds good, I'll be looking out for it :)

> > Another suggestion is: can we cut out only parts that repeat?
>
> The idea looks great.

Don't worry about this, btw. I'm digging into the implementation of `hexdump` and I may be able to repurpose it for this.

gsingh93 (approved):

Issue #1257, opened 2022-10-08
Add a command to skip the last breakpoint a certain number of times


gsingh93:
GDB has the `ignore` command to skip breakpoints a certain number of times, i.e. `ignore 4 2` skips breakpoint 4 two times, and then breaks on it the third time.

This works fine when I'm at the GDB prompt and can figure out which breakpoints have which numbers, but it's annoying in a GDB script, where after I set a breakpoint, I have to count the total breakpoints set so far to get the right breakpoint number to pass to `ignore`.

It would be useful if there was a command to call ignore on the last breakpoint that was set, i.e. `ignorelast 2` would ignore the last breakpoint that was set two times.

disconnect3d:
This could always parse the output of `gdb.execute("info break", to_string=True)`:
```
pwndbg> i b
Num     Type           Disp Enb Address            What
2       breakpoint     keep y   0x00007ffff7fe32b3 <_start+3>
3       breakpoint     del  y   0x00007ffff7fe32c6 <_dl_start_user+14>
```

Or better, use the `gdb.breakpoints()` function:
```
In [4]: gdb.breakpoints()
Out[4]: (<gdb.Breakpoint at 0x7fa7cfa22820>, <gdb.Breakpoint at 0x7fa7cfa513b0>)

In [5]: [bp.number for bp in gdb.breakpoints()]
Out[5]: [2, 3]

In [6]: max(bp.number for bp in gdb.breakpoints())
Out[6]: 3
```

The latest set breakpoint should always have the biggest number.

And we can set the ignore count from Python API too:
```
pwndbg> ignore 2 10
Will ignore next 10 crossings of breakpoint 2.
pwndbg> ipi

In [1]: bp = next(bp for bp in gdb.breakpoints() if bp.number == 2)

In [2]: bp.number
Out[2]: 2

In [3]: bp
Out[3]: <gdb.Breakpoint at 0x7fa7cfa22820>

In [4]: bp.condition

In [5]: bp.enabled
Out[5]: True

In [6]: bp.silent
Out[6]: False

In [7]: bp.expression

In [8]: bp.hit_count
Out[8]: 0

In [9]: bp.ignore_count
Out[9]: 10

In [10]: bp.ignore_count = 2

In [11]:

pwndbg> i b
Num     Type           Disp Enb Address            What
2       breakpoint     keep y   0x00007ffff7fe32b3 <_start+3>
	ignore next 2 hits
[...]

PR #1276, opened 2022-10-12
Add ignore command (from #1257)

erivas:
Tried to follow implementation outline in #1257 , where this was labeled as good-first-issue/help-wanted.

### DIFF
--- a/pwndbg/commands/ignore.py
+++ b/pwndbg/commands/ignore.py
@@ -15,6 +15,7 @@ parser = argparse.ArgumentParser(
 parser.add_argument("bpnum", type=int, default=None, nargs="?", help="The breakpoint number N.")
 parser.add_argument("count", type=int, help="The number to set COUNT.")
 
+
 @pwndbg.commands.ArgparsedCommand(parser)
 def ignore(bpnum, count):
     if bpnum is None:
--- a/pwndbg/commands/ignore.py
+++ b/pwndbg/commands/ignore.py
@@ -1,0 +1,29 @@
+"""
+Ignoring a breakpoint
+"""
+
+import argparse
+
+import gdb
+
+import pwndbg.color.message as message
+import pwndbg.commands
+
+parser = argparse.ArgumentParser(
+    description="""Set ignore-count of breakpoint, default to the last breakpoint number."""
+)
+parser.add_argument("bpnum", type=int, default=None, nargs="?", help="The breakpoint number N.")
+parser.add_argument("count", type=int, help="The number to set COUNT.")
+
+@pwndbg.commands.ArgparsedCommand(parser)
+def ignore(bpnum, count):
+    if bpnum is None:
+        bpnum = max((bp.number for bp in gdb.breakpoints()), default=None)
+
+    bp = next((bp for bp in gdb.breakpoints() if bp.number == bpnum), None)
+
+    if bp is None:
+        print(message.error("Breakpoint not found."))
+        return
+
+    bp.ignore_count = int(count)
--- a/pwndbg/commands/ignore.py
+++ b/pwndbg/commands/ignore.py
@@ -9,22 +9,36 @@ import gdb
 import pwndbg.color.message as message
 import pwndbg.commands
 
-parser = argparse.ArgumentParser(
-    description="""Set ignore-count of breakpoint, default to the last breakpoint number."""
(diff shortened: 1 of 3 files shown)


gsingh93:
Thanks! I'll test this out myself, but overall the implementation looks good. Would you also be able to add a test for this command? I think we should test different valid and invalid variants of `ignore`, where there are no breakpoints, where a non-existent breakpoint number is entered, etc.

pwndbg/commands/ignore.py
```diff
@@ -0,0 +1,30 @@
...
+parser.add_argument("bpnum", type=int, default=None, nargs="?", help="The breakpoint number N.")
+parser.add_argument("count", type=int, help="The number to set COUNT.")
+
+
+@pwndbg.commands.ArgparsedCommand(parser)
+def ignore(bpnum, count):
+    if bpnum is None:
+        bpnum = max((bp.number for bp in gdb.breakpoints()), default=None)
+
+    bp = next((bp for bp in gdb.breakpoints() if bp.number == bpnum), None)
```
disconnect3d: How about:

1. Changing this code to:
```py
if bpnum is None:
    bp = max(gdb.breakpoints(), key=lambda bp: bp.number, default=None)
else:
    bp = next((bp for bp in gdb.breakpoints() if bp.number == bpnum), None)
```

2. Changing the UX to match the original one a bit more:
```
pwndbg> help ignore
Set ignore-count of breakpoint number N to COUNT.
Usage is `ignore N COUNT'.

pwndbg> ignore 1 123
Will ignore next 123 crossings of breakpoint 1.

pwndbg> ignore 123 123
No breakpoint number 123.
```
So:
- we can show an example in the description
- display error as 'No breakpoint number <bpnum>.' if it is defined or 'No breakpoints set' if it is not
- for success path, inform user that we did it via "Will ignore next 123 crossings of breakpoint 1."

disconnect3d: Fwiw for a multiline description with examples etc. you can use RawDescriptionHelpFormatter as we do in vmmap command: https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/commands/vmmap.py#L50

Also I am wondering if we can grab last bp by just gdb.breakpoints()[-1] but idk if that is always the case so its safer to do it with max(..)

disconnect3d: Hm, so we could have:
```py
bps = gdb.breakpoints()

if not bps:
    print(message.error("No breakpoints set."))

if bpnum is None:
    bp = max(bps, key=lambda bp: bp.number)
else:
    bp = next((bp for bp in bps if bp.number == bpnum), None)

    if bp is None:
        print(message.error("No breakpoint number %d." % bpnum))
        return

bp.ignore_count = count
print("Will ignore next %d crossings of breakpoint %d." % (count, bpnum))
```

erivas:
> Thanks! I'll test this out myself, but overall the implementation looks good. Would you also be able to add a test for this command? I think we should test different valid and invalid variants of `ignore`, where there are no breakpoints, where a non-existent breakpoint number is entered, etc.

Thanks, added a few test cases in a file `test_command_ignore.py`, trying to follow how other commands are tested.

pwndbg/commands/ignore.py
erivas: Thanks! Applied the code changes.

I didn't add an example in the description, as an `usage: ...` line was already generated. It looks like this for me:
```
pwndbg> help ignore
usage: ignore [-h] [N] COUNT

Set ignore-count of breakpoint number N to COUNT.

While the ignore count is positive, execution will not stop on the breakpoint.

By default, if `N' is ommitted, the last breakpoint (i.e. greatest breakpoint number) will be used.

positional arguments:
  N           The breakpoint number N.
  COUNT       The number to set COUNT.

options:
```

gsingh93:
The tests look great! Any reason you're calling `gdb.execute("file " + REFERENCE_BINARY)` instead of `start_binary(REFERENCE_BINARY)` in some of the tests?

disconnect3d:
> The tests look great! Any reason you're calling gdb.execute("file " + REFERENCE_BINARY) instead of start_binary(REFERENCE_BINARY) in some of the tests?

Seems like an optimization - you don't always have to start the binary but can only load it to gdb

PR #1277, opened 2022-10-13
Add multithreaded malloc_chunk tests

CptGibbon:
These tests ensure `malloc_chunk` works (both with and without `set resolve-heap-with-heuristic on`) when printing non-main-arena chunks. There is also a test for cross-thread printing.

This PR also adds a `reset_on_thread` decorator, for clearing caches on instantiation of a new thread, which fixes an issue where the testing environment incorrectly deduced a thread's tcache due to stale information.


gsingh93:
Overall this looks good to me, but since it touches event related code I'll wait for @disconnect3d to take a look.

Just like when these tests were first added, I think we can simplify them a lot more and reuse a lot of code. But I'm fine with doing that myself after this is merged.

tests/binaries/heap_malloc_chunk.c
```diff
@@ -67,12 +80,16 @@ int main(void)
     small_chunk = mem2chunk(before_remainder + 0x210);
     large_chunk = mem2chunk(large);
     unsorted_chunk = mem2chunk(unsorted);
+}
+
+void* thread_func(void* args)
+{
+    // Initialize a 2nd arena
+    malloc(0x18);
```
disconnect3d, line 88: Is the specific size important here? Maybe add a comment about it 😅

tests/heap/test_heap.py
```diff
@@ -82,6 +82,35 @@ def test_malloc_chunk_command(start_binary):
     for name in chunk_types:
         assert results[name] == expected[name]
 
+    gdb.execute("continue")
+
+    # Print main thread's chunk from another thread
```
disconnect3d, line 94: The 'from another thread' part is because we will break on 'break_here' in the new thread that we will be joining, right?

disconnect3d: Might mention it here as well, just for more clarity/understanding

disconnect3d: Or/and add sanity check like `assert gdb.selected_thread().<number field if there is such> == 2`

tests/binaries/heap_malloc_chunk.c
CptGibbon: Nope, just any old chunk to initialize a 2nd arena.

tests/heap/test_heap.py
CptGibbon: Yes I like the sanity check, I'll add one or failing that a comment explaining what's going on.

PR #1279, opened 2022-10-13
Cleanup hexdump code

gsingh93:
Some minor cleanup of the `hexdump` code. Also sets the default group width to -1, which is the architecture pointer size.


pwndbg/commands/hexdump.py
```diff
@@ -13,7 +13,7 @@
 pwndbg.config.Parameter("hexdump-bytes", 64, "number of bytes printed by hexdump command")
 pwndbg.config.Parameter(
     "hexdump-group-width",
-    4,
+    -1,
```
gsingh93, line 16: Any objections to this?

disconnect3d: Not really, I think it is fine

PR #1283, opened 2022-10-13
Unit tests and some bugs fix for the heap heuristics

lebr0nli:
This PR brings some unit tests just for the heap heuristics, and also some bug fixes for it.

P.S. Some of the unit tests will fail currently because they still waiting for the fix for the symbol address resolved issues about `pwndbg.gdblib.symbol.address` after bfbb2b86525ce5c98ef912b47ba0fd75eab4eea7, but they should pass after we fixed that issue.

### DIFF
--- a/pwndbg/glibc.py
+++ b/pwndbg/glibc.py
@@ -14,7 +14,6 @@ import pwndbg.gdblib.memory
 import pwndbg.gdblib.proc
 import pwndbg.gdblib.symbol
 import pwndbg.heap
-import pwndbg.gdblib.info
 import pwndbg.lib.memoize
 import pwndbg.search
 
--- a/pwndbg/glibc.py
+++ b/pwndbg/glibc.py
@@ -3,6 +3,7 @@ Get information about the GLibc
 """
 
 import functools
+import os
 import re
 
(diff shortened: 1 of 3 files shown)


gsingh93:
https://github.com/pwndbg/pwndbg/commit/38023ccdcb7d5b763b23aa3d36302ddf6bc8d8bd should fix the failing tests. But before that, I think it would be good to add a few more tests like this: mock out `pwndbg.gdblib.symbol.address` to return `None` when looking up a symbol like `global_max_fast` and assert that a `pwndbg.heap.ptmalloc.SymbolUnresolvableError` is thrown.

This will help ensure that if symbol lookup ever fails, we fail gracefully instead of due to some other error. For example, one of the tests is currently failing with `AttributeError: 'NoneType' object has no attribute 'start'`  instead of `SymbolUnresolvableError`, and this would help prevent that.

gsingh93:
These commits are covered in https://github.com/pwndbg/pwndbg/pull/1287, so closing this PR.

gsingh93:
Merged this even though the unit tests are broken to preserve the author information. Will merge #1287 next.

PR #1286, opened 2022-10-13
Add find_fake_fast test

gsingh93:
This pulls out the test from the https://github.com/pwndbg/pwndbg/pull/1207 so I can confirm the results on the original command, and so I can get this merged faster.


tests/binaries/heap_find_fake_fast.c
```diff
@@ -10,17 +10,91 @@
...
+    // An invalid chunk with a size below the minimum chunk size
+    setup_mem(0x1F, 0x8);
+    break_here();
+
+    // A valid aligned fastbin chunk just in range of the target address
+    setup_mem(0x80, 0x78);
+    break_here();
+
+    // A valid unaligned fastbin chunk just in range of the target address
+    /* setup_mem(0x80, 0x7F); */
```
gsingh93, line 75: I would have thought this would result in a 1 byte partial overwrite of the target, but the command isn't finding it.

tests/binaries/makefile
```diff
@@ -1,3 +1,4 @@
+ZIGPATH     ?=     ../../.zig
```
disconnect3d, line 1: TIL https://stackoverflow.com/questions/24777289/what-is-in-makefile

disconnect3d (approved):

PR #1294, opened 2022-10-18
Add mocks for unit tests

gsingh93:


unit-tests/mocks/gdblib.py
```diff
@@ -0,0 +1,36 @@
...
+from unittest.mock import MagicMock
+
+from mocks.arch import Amd64Arch
+from mocks.config import Config
+from mocks.typeinfo import Amd64TypeInfo
+
+
+class GdbLib(types.ModuleType):
+    def __init__(self, module_name):
+        super(GdbLib, self).__init__(module_name)
```
disconnect3d:
```suggestion
        super().__init__(module_name)
```

disconnect3d (approved):
lgtm; one note: maybe we should move unit-tests to tests/?

gsingh93:
> lgtm; one note: maybe we should move unit-tests to tests/?

I was thinking of we should move the content of `tests` to `tests/gdb-tests` and have the unit tests be in `tests/unit-tests`. We can't have `unit-tests` in `tests` without moving the existing tests to some other folder, as `conftest.py` gets imported, which we don't want for `unit-tests`. If this directory structure sounds good, I'll update this PR with it.

disconnect3d:
Sounds good to me. Btw we can have multiple conftest.py files afaik

tests/unit-tests/mocks/gdblib.py
```diff
@@ -0,0 +1,36 @@
+import sys
+import types
+from unittest.mock import MagicMock
```
CptGibbon, line 3: Do you think mixing `unittest` and `pytest` will be okay?

gsingh93: Yup, pytest is designed to work well with `unittest`

PR #1301, opened 2022-10-19
added cymbol command

un1c0rn-the-pwnie:
A first implementation for a command importing symbols from objects defined in a C header like style.

I've currently restricted the implementation to `OnlyAmd64` only, cause remote debugging a binary in a non native architecture could be a little bit challenging.

Improvements that can be made:
* Move generated symbols away from `.cache/pwndbg` to a `.local/pwndbg` directory? For a more safe keep.
* Save and load the generated symbol files instead of recompiling the `.c` definition again and again. (Should we load them when pwndbg loads?)
* Implement a `-u/--unload` flag, for unloading loaded symbols in runtime.
* Implement a `--list-loaded` flag, for listing all currently loaded custom structures.

### DIFF
--- a/pwndbg/commands/cymbol.py
+++ b/pwndbg/commands/cymbol.py
@@ -187,9 +187,10 @@ def load_custom_structure(custom_structure_name, custom_structure_path):
 
 @OnlyWhenStructureExists
 def show_custom_structure(custom_structure_name, custom_structure_path):
-    # Clear the memoization cache.
-    pwndbg.pwndbg.commands.context.get_highlight_source.clear()
-    highlighted_source = pwndbg.pwndbg.commands.context.get_highlight_source(custom_structure_path)
+    # Call wrapper .func() to avoid memoization.
+    highlighted_source = pwndbg.pwndbg.commands.context.get_highlight_source.func(
+        custom_structure_path
+    )
     print("\n".join(highlighted_source))
 
 
--- a/pwndbg/commands/cymbol.py
+++ b/pwndbg/commands/cymbol.py
@@ -11,7 +11,7 @@ is advised to configure the 'gcc-config-path' config parameter to your own cross
 gnu gcc compiled toolchain for your target architecture.
 
 You are advised to configure the 'cymbol-editor' config parameter to the path of your
-favorite text editor. Otherwise cymbol exapnds $EDITOR and $VISUAL environement varialbes
+favorite text editor. Otherwise cymbol exapnds $EDITOR and $VISUAL environment variables
 to find the path to the default text editor.
 """
 
@@ -34,15 +34,18 @@ from pwndbg.color import message
 gcc_compiler_path = pwndbg.gdblib.config.add_param(
     "gcc-compiler-path",
     "",
-    "Path to your own gnu gcc/g++ toolchain for generating imported symbols.",
+    "Path to the gcc/g++ toolchain for generating imported symbols",
 )
 
 cymbol_editor = pwndbg.gdblib.config.add_param(
-    "cymbol-editor", "", "Path to your editor of your choice for editing custom structures."
+    "cymbol-editor", "", "Path to the editor for editing custom structures"
 )
 
 # Remeber loaded symbols. This would be useful for 'remove-symbol-file'.
 loaded_symbols = {}
+
+# Where generated symbol source files are saved.
+pwndbg_cachedir = pwndbg.lib.tempfile.cachedir("custom-symbols")
 
 
 def unload_loaded_symbol(custom_structure_name):
@@ -52,10 +55,9 @@ def unload_loaded_symbol(custom_structure_name):
         loaded_symbols.pop(custom_structure_name)
 
 
-def OnlyWhenStructureExists(func):
+def OnlyWhenStructFileExists(func):
     @functools.wraps(func)
     def wrapper(custom_structure_name):
-        pwndbg_cachedir = pwndbg.lib.tempfile.cachedir("custom-symbols")
         pwndbg_custom_structure_path = os.path.join(pwndbg_cachedir, custom_structure_name) + ".c"
         if not os.path.exists(pwndbg_custom_structure_path):
             print(message.error("No custom structure was found with the given name!"))
@@ -90,7 +92,11 @@ def generate_debug_symbols(custom_structure_path, pwndbg_debug_symbols_output_fi
         subprocess.run(gcc_cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
     except subprocess.CalledProcessError as exception:
         print(message.error(exception))
-        print(message.error("Parsing failed, try to fix any syntax errors first."))
+        print(
+            message.error(
+                "Failed to compile the .c file with custom structures. Please fix any compilation errors there may be."
+            )
+        )
         return None
     except Exception as exception:
         print(message.error(exception))
@@ -101,7 +107,6 @@ def generate_debug_symbols(custom_structure_path, pwndbg_debug_symbols_output_fi
 
 
 def add_custom_structure(custom_structure_name):
-    pwndbg_cachedir = pwndbg.lib.tempfile.cachedir("custom-symbols")
     pwndbg_custom_structure_path = os.path.join(pwndbg_cachedir, custom_structure_name) + ".c"
 
     if os.path.exists(pwndbg_custom_structure_path):
@@ -129,7 +134,7 @@ def add_custom_structure(custom_structure_name):
     load_custom_structure.__wrapped__(custom_structure_name, pwndbg_custom_structure_path)
 
 
-@OnlyWhenStructureExists
+@OnlyWhenStructFileExists
 def edit_custom_structure(custom_structure_name, custom_structure_path):
 
     # Lookup an editor to use for editing the custom structure.
@@ -163,25 +168,25 @@ def edit_custom_structure(custom_structure_name, custom_structure_path):
     load_custom_structure(custom_structure_name)
 
 
-@OnlyWhenStructureExists
+@OnlyWhenStructFileExists
 def remove_custom_structure(custom_structure_name, custom_structure_path):
     unload_loaded_symbol(custom_structure_name)
     os.remove(custom_structure_path)
     print(message.success("Symbols are removed!"))
 
 
-@OnlyWhenStructureExists
+@OnlyWhenStructFileExists
 def load_custom_structure(custom_structure_name, custom_structure_path):
     unload_loaded_symbol(custom_structure_name)
     pwndbg_debug_symbols_output_file = generate_debug_symbols(custom_structure_path)
     if not pwndbg_debug_symbols_output_file:
-        return
+        return  # generate_debug_symbols prints on failures
     gdb.execute(f"add-symbol-file {pwndbg_debug_symbols_output_file}", to_string=True)
     loaded_symbols[custom_structure_name] = pwndbg_debug_symbols_output_file
     print(message.success("Symbols are loaded!"))
 
 
-@OnlyWhenStructureExists
+@OnlyWhenStructFileExists
 def show_custom_structure(custom_structure_name, custom_structure_path):
     # Call wrapper .func() to avoid memoization.
     highlighted_source = pwndbg.pwndbg.commands.context.get_highlight_source.func(
--- a/pwndbg/commands/cymbol.py
+++ b/pwndbg/commands/cymbol.py
@@ -6,7 +6,7 @@ Add, load, show, edit, or delete symbols for custom structures.
 
 For the generation of the symbols g++/gcc is being used under the hood.
 
-In case of remote debugging a binary which is not native to your architecture it 
+In case of remote debugging a binary which is not native to your architecture it
 is advised to configure the 'gcc-config-path' config parameter to your own cross-platform
 gnu gcc compiled toolchain for your target architecture.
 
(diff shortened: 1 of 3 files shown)


pwndbg/commands/cymbol.py
```diff
@@ -0,0 +1,273 @@
...
+    default=None,
+    type=str,
+)
+
+
+@pwndbg.commands.ArgparsedCommand(parser)
+@pwndbg.commands.OnlyAmd64
+@pwndbg.commands.OnlyWhenRunning
+def cymbol(add, remove, edit, load, show):
+    if add:
```
gsingh93, line 247: You can rewrite this as an if/elif/else statement.

pwndbg/commands/cymbol.py
```diff
@@ -0,0 +1,269 @@
...
+        if not os.path.exists(pwndbg_custom_structure_path):
+            print(message.error("No custom structure was found with the given name!"))
+            return
+
+        return func(custom_structure_name, pwndbg_custom_structure_path)
+
+    return wrapper
+
+
+def PromptForOverwrite(func):
```
disconnect3d: Does it make sense that this is a decorator? Wouldn't it be better to just execute this code in the currently wrapped command function?

pwndbg/commands/cymbol.py
```diff
@@ -0,0 +1,269 @@
...
+            gcc_flags = [gcc_compiler_path, "-m32"]
+    else:
+        # Should we check for exceptions here?
+        gcc_flags = pwndbg.lib.gcc.which(pwndbg.gdblib.arch)
+
+    gcc_cmd = gcc_flags + gcc_extra_flags
+
+    try:
+        subprocess.run(gcc_cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
+    except subprocess.CalledProcessError as exepction:
```
disconnect3d: Typo - exception

pwndbg/commands/cymbol.py
```diff
@@ -0,0 +1,269 @@
...
+        print(message.error(f"Path to the custom structure: {custom_structure_path}"))
+        print(message.error("Please try to manually edit the structure."))
+        print(
+            message.error(
+                '\nTry to set a path to an editor with:\n\tset "cymbol-editor" /usr/bin/nano'
+            )
+        )
+        return
+
+    input(message.notice("Press enter to continue."))
```
disconnect3d: Is this needed?

pwndbg/commands/cymbol.py
```diff
@@ -0,0 +1,269 @@
...
+    load_custom_structure(custom_structure_name)
+
+
+@OnlyWhenStructureExists
+def remove_custom_structure(custom_structure_name, custom_structure_path):
+    os.remove(custom_structure_path)
+    print(message.success("Symbols are removed!"))
+    print(
+        message.notice(
+            "If the symbols are already loaded, please restart pwndbg to unload them from memory."
```
disconnect3d: Can we try using https://stackoverflow.com/a/19681769/1508881 for that?

pwndbg/commands/cymbol.py
un1c0rn-the-pwnie: My latest commit should have fixed that.

pwndbg/commands/cymbol.py
un1c0rn-the-pwnie: Oh nice "catch" :)

pwndbg/commands/cymbol.py
```diff
@@ -0,0 +1,269 @@
...
+    if not pwndbg_debug_symbols_output_file:
+        return
+    gdb.execute(f"add-symbol-file {pwndbg_debug_symbols_output_file}", to_string=True)
+    print(message.success("Symbols are loaded!"))
+
+
+@OnlyWhenStructureExists
+def show_custom_structure(custom_structure_name, custom_structure_path):
+    print()
+    with open(custom_structure_path, "r") as f:
```
disconnect3d: We can use syntax highlighting here! :D
```py
code = pwndbg.pwndbg.commands.context.get_highlight_source(custom_structure_path)
print("\n".join(code))
```

pwndbg/commands/cymbol.py
un1c0rn-the-pwnie: Yeah it is needed when the editor is a GUI app like vs code etc that might not be closed when the user is done with editing it. But also it is convenient for the user to have it open in the background.

pwndbg/commands/cymbol.py
un1c0rn-the-pwnie: Hmm I thought this might be useful for the future in case of adding new flags to the command, but yeah probably I shouldn't implement it as a decorator in this case.

pwndbg/commands/cymbol.py
un1c0rn-the-pwnie: Yeah I'm planning that when I will properly store the debug symbols in a `.local` directory. For now I think won't work because the file paths to the loaded generated symbols are randomly generated temporary files in `/tmp`.

pwndbg/commands/cymbol.py
disconnect3d: Interesting, so e.g. vim will block but vscode will not, right? Maybe we should inform user to press enter after they saved the struct, but maybe it is fine as it is

pwndbg/commands/cymbol.py
un1c0rn-the-pwnie: what sorcery is this, pretty cool!

pwndbg/commands/cymbol.py
disconnect3d: We can save/cache the paths in some dict like: `symbols[symbol_name] = symbol_file_path` and then we will have it here :)

pwndbg/commands/cymbol.py
```diff
@@ -0,0 +1,269 @@
...
+    gcc_flags = None
+
+    if gcc_compiler_path != "":
+        if pwndbg.gdblib.arch.ptrsize == 8:
+            gcc_flags = [gcc_compiler_path, "-m64"]
+        elif pwndbg.gdblib.arch.ptrsize == 4:
+            gcc_flags = [gcc_compiler_path, "-m32"]
+    else:
+        # Should we check for exceptions here?
+        gcc_flags = pwndbg.lib.gcc.which(pwndbg.gdblib.arch)
```
gsingh93, line 102: Not an exception, but you should probably check if the first element of the list is `None` (we should really change this method to just return None if it doesn't find the compiler and not return a list).

pwndbg/commands/cymbol.py
un1c0rn-the-pwnie: Yeah at least when I put `export EDITOR=code`, vs code doesn't seem to block pwndbg, might need some extra flags, I will check it. But yeah it would be better to display a more informative message to the user rather than a generic 'Press enter to continue.'.

pwndbg/commands/cymbol.py
```diff
@@ -0,0 +1,269 @@
...
+    ]
+    gcc_flags = None
+
+    if gcc_compiler_path != "":
+        if pwndbg.gdblib.arch.ptrsize == 8:
+            gcc_flags = [gcc_compiler_path, "-m64"]
+        elif pwndbg.gdblib.arch.ptrsize == 4:
+            gcc_flags = [gcc_compiler_path, "-m32"]
+    else:
+        # Should we check for exceptions here?
```
gsingh93: I checked the code for `which` and I don't think an exception check is necessary.

pwndbg/commands/cymbol.py
```diff
@@ -0,0 +1,269 @@
...
+    gcc_cmd = gcc_flags + gcc_extra_flags
+
+    try:
+        subprocess.run(gcc_cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
+    except subprocess.CalledProcessError as exepction:
+        print(message.error(exepction))
+        print(message.error("Parsing failed, try to fix any syntax errors first."))
+        return None
+    except Exception as exepction:
+        print(message.error(exepction))
```
gsingh93: Same here.

pwndbg/commands/cymbol.py
un1c0rn-the-pwnie: Good suggestion, also we will need `remove-symbol-file` after all for unloading the symbols each time we edit an already loaded symbol. It seems that multiple edits don't work currently for that reason.

pwndbg/commands/cymbol.py
```diff
@@ -0,0 +1,269 @@
...
+        "-g",
+        "-fno-eliminate-unused-debug-types",
+        "-o",
+        pwndbg_debug_symbols_output_file,
+    ]
+    gcc_flags = None
+
+    if gcc_compiler_path != "":
+        if pwndbg.gdblib.arch.ptrsize == 8:
+            gcc_flags = [gcc_compiler_path, "-m64"]
```
gsingh93: You don't have to do this, but it would be nice if we could pull out this if statement into it's own function in gcc.py so we can use it in both places: https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/lib/gcc.py#L25

pwndbg/commands/cymbol.py
```diff
@@ -0,0 +1,253 @@
...
+Add, load, show, edit, or delete symbols for custom structures.
+
+For the generation of the symbols g++/gcc is being used under the hood.
+
+In case of remote debugging a binary which is not native to your architecture it
+is advised to configure the 'gcc-config-path' config parameter to your own cross-platform
+gnu gcc compiled toolchain for your target architecture.
+
+You are advised to configure the 'cymbol-editor' config parameter to the path of your
+favorite text editor. Otherwise cymbol exapnds $EDITOR and $VISUAL environement varialbes
```
disconnect3d:
```suggestion
favorite text editor. Otherwise cymbol exapnds $EDITOR and $VISUAL environment variables
```

pwndbg/commands/cymbol.py
```diff
@@ -0,0 +1,253 @@
...
+import pwndbg.commands
+import pwndbg.gdblib.arch
+import pwndbg.lib.gcc
+import pwndbg.lib.tempfile
+from pwndbg.color import message
+
+gcc_compiler_path = pwndbg.gdblib.config.add_param(
+    "gcc-compiler-path",
+    "",
+    "Path to your own gnu gcc/g++ toolchain for generating imported symbols.",
```
disconnect3d:
```suggestion
    "Path to the gcc/g++ toolchain for generating imported symbols",
```

pwndbg/commands/cymbol.py
```diff
@@ -0,0 +1,253 @@
...
+)
+
+cymbol_editor = pwndbg.gdblib.config.add_param(
+    "cymbol-editor", "", "Path to your editor of your choice for editing custom structures."
```
disconnect3d:
```suggestion
    "cymbol-editor", "", "Path to the editor for editing custom structures"
```

pwndbg/commands/cymbol.py
```diff
@@ -0,0 +1,253 @@
...
+    if gcc_compiler_path != "":
+        gcc_flags[0] = gcc_compiler_path
+
+    gcc_cmd = gcc_flags + gcc_extra_flags
+
+    try:
+        subprocess.run(gcc_cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, check=True)
+    except subprocess.CalledProcessError as exception:
+        print(message.error(exception))
+        print(message.error("Parsing failed, try to fix any syntax errors first."))
```
disconnect3d:
```suggestion
        print(message.error("Failed to compile the .c file with custom structures. Please fix any compilation errors there may be."))
```

pwndbg/commands/cymbol.py
```diff
@@ -0,0 +1,253 @@
...
+        option = input(
+            message.notice(
+                "A custom structure was found with the given name, would you like to overwrite it? [y/n] "
+            )
+        )
+        if option != "y":
+            return
+
+    print(
+        message.notice("Enter your custom structure in a C header style, press Ctrl+D to save:\n")
```
disconnect3d, line 122: Potential improvement: create a struct_template.c file in pwndbg and then a GDB paramater which will point to it and  whenever the user adds a struct, copy the template first and open the editor on the copied template.

Example template:
```c
// This is a Pwndbg cymbol command C structure template file
// To use a different template for cymbol C structs, use `set cymbol-template-path <path-to-c-file>` command
#include <stdio.h>
#include <stdint.h>

typedef struct  {
      // Put your structure fields here
} <custom_structure_name>;
```

disconnect3d: Its fine to leave this for now and add it in a future PR.

pwndbg/commands/cymbol.py
```diff
@@ -0,0 +1,253 @@
...
+    os.remove(custom_structure_path)
+    print(message.success("Symbols are removed!"))
+
+
+@OnlyWhenStructureExists
+def load_custom_structure(custom_structure_name, custom_structure_path):
+    unload_loaded_symbol(custom_structure_name)
+    pwndbg_debug_symbols_output_file = generate_debug_symbols(custom_structure_path)
+    if not pwndbg_debug_symbols_output_file:
+        return
```
disconnect3d:
```suggestion
        return  # generate_debug_symbols prints on failures
```

pwndbg/commands/cymbol.py
```diff
@@ -0,0 +1,253 @@
...
+    custom_structure_symbols_file = loaded_symbols.get(custom_structure_name)
+    if custom_structure_symbols_file is not None:
+        gdb.execute(f"remove-symbol-file {custom_structure_symbols_file}")
+        loaded_symbols.pop(custom_structure_name)
+
+
+def OnlyWhenStructureExists(func):
+    @functools.wraps(func)
+    def wrapper(custom_structure_name):
+        pwndbg_cachedir = pwndbg.lib.tempfile.cachedir("custom-symbols")
```
disconnect3d: Can we move the `custom-symbols` string to a constant variable at the global namespace instead of hardcoding it in multiple places?

pwndbg/commands/cymbol.py
```diff
@@ -0,0 +1,253 @@
...
+
+
+def unload_loaded_symbol(custom_structure_name):
+    custom_structure_symbols_file = loaded_symbols.get(custom_structure_name)
+    if custom_structure_symbols_file is not None:
+        gdb.execute(f"remove-symbol-file {custom_structure_symbols_file}")
+        loaded_symbols.pop(custom_structure_name)
+
+
+def OnlyWhenStructureExists(func):
```
disconnect3d: Shall we rename this to `OnlyWhenStructFileExists`? When reading tests I initially thought that this may be about "whether we have symbols for a given struct loaded into GDB".

disconnect3d: Shall we rename this to `OnlyWhenStructFileExists`? When reading tests I initially thought that this may be about "whether we have symbols for a given struct loaded into GDB".

tests/gdb-tests/tests/test_cymbol.py
```diff
@@ -0,0 +1,69 @@
...
+            int** c;
+            void* d;
+        } example2_t;
+    """
+    )
+    # Always overwrite files which exist.
+    os.path.exists = lambda x: False
+    pwndbg.commands.cymbol.add_custom_structure("example2")
+    # Not much but honest work.
+    assert ": int [16]" in pwndbg.gdblib.dt.dt("example2_t").strip()
```
disconnect3d: Can't we assert the whole result of `dt`?

pwndbg/commands/cymbol.py
disconnect3d: Shall we rename this to `OnlyWhenStructFileExists`? When reading tests I initially thought that this may be about "whether we have symbols for a given struct loaded into GDB".

tests/gdb-tests/tests/test_cymbol.py
disconnect3d: Can't we assert the whole result of `dt`?

tests/gdb-tests/tests/test_cymbol.py
```diff
@@ -0,0 +1,69 @@
...
+    # Test whether load_custom_structure() works properly
+    pwndbg.commands.cymbol.load_custom_structure("example")
+    # Not much but honest work.
+    assert "+0x0004 b" in pwndbg.gdblib.dt.dt("example_t").strip()
+
+    # Test whether add_custom_structure() works properly.
+    saved_read = sys.stdin.read
+    saved_exists = os.path.exists
+    # We do a little hack here :)
+    sys.stdin.read = (
```
disconnect3d: You should do this hack with pytest-mock. Otherwise, if there would be an exception, the stdin.read would be broken for any code executed after this test. This isn't an issue now, but will be after we switch to a more robust testing solution (currently each test is executed in a separate process/GDB instance).

PR #1304, opened 2022-10-20
integrated Bin classes

pogen300:
integrated the Bin classes in #1063 with the current code, changed the logic for the contains_chunk method for the Bins class slightly. Added some properties to the Arena class.


pwndbg/heap/ptmalloc.py
```diff
@@ -221,20 +310,59 @@ def is_top_chunk(self):
...
+        if self._mutex is None:
+            try:
+                self._mutex = int(self._gdbValue["mutex"])
+            except gdb.MemoryError:
+                pass
+            
+        return self._mutex
 
+    @property
+    def flag(self):
```
CptGibbon: Super minor but since this property is backed by `_flags` perhaps calling it `flags` rather than `flag` makes more sense.

pwndbg/heap/ptmalloc.py
```diff
@@ -653,48 +862,46 @@ def bin_at(self, index, arena_addr=None):
...
         size = self.min_chunk_size - self.malloc_alignment
         spaces_table = self._spaces_table()
 
-        result = OrderedDict()
+        result = Bins(BinType.SMALL)
         for index in range(2, 64):
             size += spaces_table[index]
             chain = self.bin_at(index, arena_addr=arena_addr)
-
+
```
CptGibbon: Just need a small lint commit to iron out stuff like this, you can see what's getting rejected from the test output.

CptGibbon (changes requested):

gsingh93:
LGTM. Regarding the lint, you should just need to run `./lint.sh -f`. This is now documented here: https://github.com/pwndbg/pwndbg/blob/dev/DEVELOPING.md

PR #1307, opened 2022-10-20
Fix parameter default values

disconnect3d:
Before this commit the created gdb.Parameter default values were not set properly. Now, we set the object's .value field properly with the provided default value.

### DIFF
--- a/pwndbg/gdblib/config.py
+++ b/pwndbg/gdblib/config.py
@@ -37,19 +37,17 @@ PARAM_CLASSES = {
 # https://sourceware.org/gdb/onlinedocs/gdb/Parameters-In-Python.html
 class Parameter(gdb.Parameter):
     def __init__(self, param: pwndbg.lib.config.Parameter):
+        param_class = PARAM_CLASSES[type(param.value)]
+        super().__init__(param.name, gdb.COMMAND_SUPPORT, param_class)
+
         self.param = param
+        self.value = param.value
 
         # `set_doc` and `show_doc` must be set before `gdb.Parameter.__init__`.
         # They will be used for `help set <param>` and `help show <param>`,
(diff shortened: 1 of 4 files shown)


pwndbg/gdblib/config.py
```diff
@@ -37,19 +37,17 @@
 # https://sourceware.org/gdb/onlinedocs/gdb/Parameters-In-Python.html
 class Parameter(gdb.Parameter):
     def __init__(self, param: pwndbg.lib.config.Parameter):
+        param_class = PARAM_CLASSES[type(param.value)]
+        super().__init__(param.name, gdb.COMMAND_SUPPORT, param_class)
+
         self.param = param
+        self.value = param.value
 
         # `set_doc` and `show_doc` must be set before `gdb.Parameter.__init__`.
```
gsingh93: I think this is still true, we need to put the init at the end

pwndbg/gdblib/config.py
```diff
@@ -37,19 +37,17 @@
...
-
-    def _param_class(self):
-        for k, v in PARAM_CLASSES.items():
-            if isinstance(self.param.value, k):
-                return v
+        self.set_doc = "Set " + param.docstring
+        self.show_doc = "Show " + param.docstring
+
+        param_class = PARAM_CLASSES[type(param.value)]
+        super().__init__(param.name, gdb.COMMAND_SUPPORT, param_class)
```
gsingh93, line 47: Just checking, is it important that this comes before the assignment to `self.param` and `self.value` below? If not, this is fine, but if that's the case we should add a comment about it, similar to what I added above.

disconnect3d: I think it was, so i moved it back

PR #1311, opened 2022-10-20
Add a header to the vmmap table

alufers:
A simple header has been added to the output of vmmap which helps new users identify the columns.

image

### DIFF
--- a/pwndbg/commands/vmmap.py
+++ b/pwndbg/commands/vmmap.py
@@ -29,6 +29,7 @@ def pages_filter(gdbval_or_str):
     else:
         raise argparse.ArgumentTypeError("Unknown vmmap argument type.")
 
+
 def print_vmmap_table_header():
     """
     Prints the table header for the vmmap command.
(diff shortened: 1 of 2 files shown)


pwndbg/commands/vmmap.py
```diff
@@ -29,6 +29,13 @@ def pages_filter(gdbval_or_str):
     else:
         raise argparse.ArgumentTypeError("Unknown vmmap argument type.")
 
+def print_vmmap_table_header():
+    """
+    Prints the table header for the vmmap command.
+    """
+    width = 2 + 2 * pwndbg.gdblib.arch.ptrsize
+    fmt_string = "%#{}s %#{}s %#4s %#8s %#6s %s".format(width, width)
+    print(fmt_string % ("Start", "End", "Perm", "Size", "Offset", "File"))
```
gsingh93, line 39: nitpick: we can probably do all this in the previous format() call, right? If so, we should probably just do that.

gsingh93 (commented):
Overall this lgtm. For the lint issue, you can run `./lint.sh -f` locally (more details here: https://github.com/pwndbg/pwndbg/blob/dev/DEVELOPING.md). For the tests, it seems like there are more tests that use the vmmap output that you'll need to update with the header. You can run `./tests.sh` locally to reproduce this.

pwndbg/commands/vmmap.py
disconnect3d: we can probably use f-string literal for all this

gsingh93: This didn't get fixed 😞

disconnect3d: I know but I don't think it is that big deal; it was also copied from another part of the code ;p

we can easily fix it too :P

PR #1314, opened 2022-10-21
small change to largebins() and contains_chunk()

pogen300:
small change to largebins() and contains_chunk() so that accessing a largebin uses the index (calculated from chunk size) as the key instead of the chunk size

### DIFF
--- a/pwndbg/heap/ptmalloc.py
+++ b/pwndbg/heap/ptmalloc.py
@@ -94,10 +94,7 @@ class Bins:
 
             # TODO: Refactor this, the bin should know how to calculate
             # largebin_index without calling into the allocator
-            if size < list(self.bins.items())[0][0]:
-                return False
             size = pwndbg.heap.current.largebin_index(size) - 64
-            return list(self.bins.items())[size][1].contains_chunk(chunk)
 
         elif self.bin_type == BinType.TCACHE:
             # Unlike fastbins, tcache bins don't store the chunk address in the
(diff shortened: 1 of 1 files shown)


pwndbg/heap/ptmalloc.py
```diff
@@ -94,10 +94,7 @@ def contains_chunk(self, size, chunk):
 
             # TODO: Refactor this, the bin should know how to calculate
             # largebin_index without calling into the allocator
```
disconnect3d, line 97: just to be sure: is this comment still relevant?

gsingh93: Yes, it's still relevant. With an ideal design, there should be no need to call `pwndbg.heap.current.largebin_index(size)`, as the algorithm for doing that doesn't depend on having a reference to an actual allocator. Not a big deal right now, but this separation will help in the long run.

pwndbg/heap/ptmalloc.py
```diff
@@ -94,10 +94,7 @@ def contains_chunk(self, size, chunk):
...
-            if size < list(self.bins.items())[0][0]:
-                return False
             size = pwndbg.heap.current.largebin_index(size) - 64
```
gsingh93: I think we should make a constant for NSMALLBINS and use it here instead of 64, so it's more clear why we're doing this.

gsingh93 (approved):
LGTM, requested a review from Max.

CptGibbon (approved):

PR #1315, opened 2022-10-21
Some enhancements about `pwndbg.gdblib.config`

lebr0nli:
Summary about this PR:

* Support all parameter-class when create a GDB parameter

* Use `get_show_string` to render better output when using `show <param>` (although GDB 12.x will do similar things by default, this function just makes sure every version can have that kind of output)

* Show more information when using `help set <param>` and `help show <param>` if we create a config with `help_docstring` parameter.

Here are some examples of the updates included in this PR:

1. `gdb.PARAM_AUTO_BOOLEAN` with `help_docstring`

In Python script:

```
pwndbg.gdblib.config.add_param(
    "test",
    None,
    "test",
    help_docstring="on == AAAA\noff == BBBB\nauto == CCCC",
    param_class=gdb.PARAM_AUTO_BOOLEAN,
    scope="test",
)
```

In GDB:

```
pwndbg> show test
The current value of 'test' is 'auto'
pwndbg> set test on
Set test to 'on'
pwndbg> set test off
Set test to 'off'
pwndbg> set test auto_with_typo
"on", "off" or "auto" expected.
pwndbg> show test
The current value of 'test' is 'off'
pwndbg> set test auto
Set test to 'auto'
pwndbg> show test
The current value of 'test' is 'auto'
pwndbg> help show test
Show test
on == AAAA
off == BBBB
auto == CCCC
pwndbg> help set test
Set test
on == AAAA
off == BBBB
auto == CCCC
```

2. `gdb.PARAM_AUTO_BOOLEAN` with `help_docstring`

In Python script:

```
pwndbg.gdblib.config.add_param(
    "test",
    "A",
    "test",
    help_docstring="A == AAAA\nB == BBBB\nC == CCCC",
    param_class=gdb.PARAM_ENUM,
    enum_sequence=["A", "B", "C"],
    scope="test",
)
```

In GDB:

```
pwndbg> show test
The current value of 'test' is 'A'
pwndbg> set test B
Set test to 'B'
pwndbg> set test C
Set test to 'C'
pwndbg> set test D
Undefined item: "D".
pwndbg> show test
The current value of 'test' is 'C'
pwndbg> help show test
Show test
A == AAAA
B == BBBB
C == CCCC
pwndbg> help set test
Set test
A == AAAA
B == BBBB
C == CCCC
```

### DIFF
--- a/pwndbg/gdblib/config.py
+++ b/pwndbg/gdblib/config.py
@@ -26,13 +26,11 @@ config = pwndbg.lib.config.Config()
 # https://sourceware.org/gdb/onlinedocs/gdb/Parameters-In-Python.html
 class Parameter(gdb.Parameter):
     def __init__(self, param: pwndbg.lib.config.Parameter):
-        # `set_doc` and `show_doc` must be set before `gdb.Parameter.__init__`.
+        # `set_doc`, `show_doc`, and `__doc__` must be set before `gdb.Parameter.__init__`.
         # They will be used for `help set <param>` and `help show <param>`,
         # respectively
         self.set_doc = "Set " + param.docstring
         self.show_doc = "Show " + param.docstring
-        # `__doc__` must be set before `gdb.Parameter.__init__`, too.
-        # It will be used for `help set <param>` and `help show <param>`.
         self.__doc__ = param.help_docstring
 
         if param.param_class == gdb.PARAM_ENUM:
@@ -90,11 +88,11 @@ class Parameter(gdb.Parameter):
     @staticmethod
     def _value_to_gdb_native(value, param_class=None):
         """Translates Python value into native GDB syntax string."""
-        # Convert booleans to "on" or "off".
         if isinstance(value, bool):
+            # Convert booleans to "on" or "off".
             return "on" if value else "off"
         elif value is None and param_class == gdb.PARAM_AUTO_BOOLEAN:
-            # None for gdb.PARAM_UINTEGER means "auto".
+            # None for gdb.PARAM_AUTO_BOOLEAN means "auto".
             return "auto"
         elif value == 0 and param_class in (gdb.PARAM_UINTEGER, gdb.PARAM_INTEGER):
             # 0 for gdb.PARAM_UINTEGER and gdb.PARAM_INTEGER means "unlimited".
--- a/pwndbg/gdblib/config.py
+++ b/pwndbg/gdblib/config.py
@@ -21,17 +21,6 @@ import pwndbg.lib.config
 
 config = pwndbg.lib.config.Config()
 
-PARAM_CLASSES = {
-    # The Python boolean values, True and False are the only valid values.
-    bool: gdb.PARAM_BOOLEAN,
-    # This is like PARAM_INTEGER, except 0 is interpreted as itself.
-    int: gdb.PARAM_ZINTEGER,
-    # When the user modifies the string, any escape sequences,
-    # such as ‘\t’, ‘\f’, and octal escapes, are translated into
-    # corresponding characters and encoded into the current host charset.
-    str: gdb.PARAM_STRING,
-}
-
 
 # See this for details about the API of `gdb.Parameter`:
 # https://sourceware.org/gdb/onlinedocs/gdb/Parameters-In-Python.html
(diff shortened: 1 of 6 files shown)


pwndbg/gdblib/config.py
```diff
@@ -42,26 +31,44 @@ def __init__(self, param: pwndbg.lib.config.Parameter):
         # respectively
         self.set_doc = "Set " + param.docstring
         self.show_doc = "Show " + param.docstring
-
-        param_class = PARAM_CLASSES[type(param.value)]
-        super().__init__(param.name, gdb.COMMAND_SUPPORT, param_class)
-
+        # `__doc__` must be set before `gdb.Parameter.__init__`, too.
```
gsingh93: I appreciate the comment, but I think you can just merge this with the above comment with something like: "`set_doc`, `show_doc`, and `__doc__` must be ..."

pwndbg/gdblib/config.py
```diff
@@ -71,15 +78,32 @@ def get_set_string(self):
...
-    def _value_to_gdb_native(value):
+    def _value_to_gdb_native(value, param_class=None):
         """Translates Python value into native GDB syntax string."""
-        # Convert booleans to "on" or "off". Other types pass through normally
+        # Convert booleans to "on" or "off".
         if isinstance(value, bool):
             return "on" if value else "off"
-
+        elif value is None and param_class == gdb.PARAM_AUTO_BOOLEAN:
+            # None for gdb.PARAM_UINTEGER means "auto".
```
gsingh93: `gdb.PARAM_AUTO_BOOLEAN`

pwndbg/gdblib/config.py
```diff
@@ -71,15 +78,32 @@ def get_set_string(self):
...
+            self.param.name,
+            Parameter._value_to_gdb_native(svalue, self.param.param_class),
+        )
 
     @staticmethod
-    def _value_to_gdb_native(value):
+    def _value_to_gdb_native(value, param_class=None):
         """Translates Python value into native GDB syntax string."""
-        # Convert booleans to "on" or "off". Other types pass through normally
+        # Convert booleans to "on" or "off".
```
gsingh93: nit: the other comments are inside the if statement, so can you move this one inside too?

pwndbg/heap/__init__.py
```diff
@@ -5,31 +5,33 @@
 
 current = None
 
-main_arena = pwndbg.gdblib.config.add_param("main-arena", "0", "&main_arena for heuristics", "heap")
+main_arena = pwndbg.gdblib.config.add_param(
+    "main-arena", "0", "&main_arena for heuristics", scope="heap"
```
gsingh93: Just like we have `theme.add_param` and `theme.add_color_param`, maybe we should have some `add_heap_param` function so we don't need to type out the scope everywhere.

tests/gdb-tests/tests/test_gdblib_parameter.py
```diff
@@ -6,31 +6,57 @@
...
+        ),
+    ),
 )
 def test_gdb_parameter_default_value_works(start_binary, params):
-    name_suffix, default_value, displayed_value = params
+    name_suffix, default_value, displayed_value, optional_kwargs = params
 
     param_name = f"test-param-{name_suffix}"
+    help_docstring = f"Help docstring for {param_name}"
+    optional_kwargs["help_docstring"] = help_docstring
```
gsingh93: This seems to be the only keyword arg, why do we need `optional_kwargs`? We can just do `help_docstring=help_docstring` in the call below.

tests/gdb-tests/tests/test_gdblib_parameter.py
```diff
@@ -6,31 +6,57 @@
...
-    )
-    assert gdb.parameter(param_name) == default_value
+    assert out == f"The current value of {param_name!r} is {displayed_value!r}\n"
+    if (
+        optional_kwargs.get("param_class") in (gdb.PARAM_UINTEGER, gdb.PARAM_INTEGER)
+        and default_value == 0
+    ):
+        # Note: This is really weird, according to GDB docs, 0 should mean "unlimited" for gdb.PARAM_UINTEGER and gdb.PARAM_INTEGER, but somehow GDB sets the value to `None` actually :/
+        # And hilarious thing is that GDB won't let you set the default value to `None` when you construct the `gdb.Parameter` object with `gdb.PARAM_UINTEGER` or `gdb.PARAM_INTEGER` lol
+        # Maybe it's a bug of GDB?
```
gsingh93, line 51: Let's consider reporting this upstream if we can.

gsingh93 (commented):
Thanks for doing this and thanks for adding so many comments :) It was very easy to review.

pwndbg/lib/config.py
```diff
@@ -97,11 +122,20 @@ def __init__(self):
         self.params = {}
         self.triggers = collections.defaultdict(lambda: [])
 
-    def add_param(self, name, default, docstring, scope="config"):
+    def add_param(
```
gsingh93, line 125: The number of parameters here is growing large. I'm wondering if we should force callers of this function to use keyword arguments for the optional fields we've added. We can do this like this:
```python
def add_param(
        self,
        name,
        default,
        docstring,
        *,
        help_docstring=None,
        param_class=None,
        enum_sequence=None,
        scope="config",
    ):
```

Everything after the * is required to be a keyword argument.

pwndbg/gdblib/config.py
lebr0nli: Sure, I'll merge these two comments.

pwndbg/gdblib/config.py
lebr0nli: Ahh, I forgot to modify it after copy and paste🤣
I'll fix this!

tests/gdb-tests/tests/test_gdblib_parameter.py
lebr0nli: > This seems to be the only keyword arg, why do we need `optional_kwargs`?

`optional_kwargs` is for testing the auto boolean and other parameter classes when we are not using the default `param_class`.

For example, when we are testing the `gdb.PARAM_ENUM` parameter, `optional_kwargs` will be:

```python
{"param_class": gdb.PARAM_ENUM, "enum_sequence": ["enum1", "enum2", "enum3"]}
```

> We can just do help_docstring=help_docstring in the call below.

Yes, this looks better, I'll pass `help_docstring` directly.

tests/gdb-tests/tests/test_gdblib_parameter.py
lebr0nli: Ok, I sent it to the mailing list of GDB now.

pwndbg/heap/__init__.py
lebr0nli: Sure, I'll add `add_heap_param` for it.

pwndbg/gdblib/config.py
lebr0nli: Yes, I'll move it.

pwndbg/lib/config.py
lebr0nli: I think this is a good idea!
I'll update it to what you suggest.

pwndbg/lib/config.py
```diff
@@ -2,16 +2,41 @@
...
+    def __init__(
+        self,
+        name,
+        default,
+        docstring,
+        help_docstring=None,
+        param_class=None,
+        enum_sequence=None,
+        scope="config",
+    ):
```
gsingh93, line 32: Maybe we should also add the `*` in here to force keyword params starting from `help_docstring`?

gsingh93 (approved):
LGTM, will let @disconnect3d take a look and merge.

tests/gdb-tests/tests/test_gdblib_parameter.py
disconnect3d: Link the discussion in a comment reply to this PR pls

pwndbg/lib/config.py
disconnect3d: That probably makes sense, :+1:.

Btw one note: `docstring` and `help_docstring` is not so obvious I guess? I already merged this but noticed this only now. Maybe we should rename `docstring` to `set_show_doc`?

tests/gdb-tests/tests/test_gdblib_parameter.py
lebr0nli: https://sourceware.org/pipermail/gdb/2022-October/050398.html

lebr0nli: > https://sourceware.org/pipermail/gdb/2022-October/050398.html

Update: https://sourceware.org/pipermail/gdb/2022-October/050426.html
If I didn't misunderstand something, seems like the value in Python to represent `unlimited` will only be `None` instead of some integers in the future.
So we'll probably need to update the code here, the tests' params, and `_value_to_gdb_native()` to fit the new API in the future.

PR #1316, opened 2022-10-21
refactor: use .format() properly for displaying vmmap rows

alufers:
Fixes the comments in #1311


gsingh93:
Thanks for fixing this!

PR #1318, opened 2022-10-22
Add Heap class

CptGibbon:
Add a `Heap` class to the heap inspection code, plus:
* Rename the existing `Heap` class to `GlibcMemoryAllocator` and the `BaseHeap` class to `MemoryAllocator`, which I think is more descriptive.
* Integrate the new `Heap` class into the `heap` command.

Hopefully this is the last heap-related class pwndbg needs for now, along with the recent Chunk, Arena & Bins classes.
This is a bare-bones implementation for now, just enough to for use in the `heap` command and there is some redundancy with existing classes & functions, which I plan to iron out in subsequent PRs. The end goal is to move most of the existing heap logic into the heap classes.

### DIFF
--- a/pwndbg/heap/ptmalloc.py
+++ b/pwndbg/heap/ptmalloc.py
@@ -317,16 +317,20 @@ class Heap:
         allocator = pwndbg.heap.current
         if addr is not None:
             try:
+                # Can fail if any part of the struct is unmapped (due to corruption/fake struct).
                 ar_ptr = allocator.get_heap(addr)["ar_ptr"]
                 ar_ptr.fetch_lazy()
             except Exception:
                 ar_ptr = None
 
+            # This is probably a non-main arena if a legitimate ar_ptr exists.
             if ar_ptr is not None and ar_ptr in (ar.address for ar in allocator.arenas):
                 self.arena = Arena(ar_ptr)
             else:
+                # Use the main arena as a fallback
                 self.arena = Arena(allocator.main_arena.address)
         else:
+            # Get the thread arena under default conditions.
             self.arena = Arena(allocator.get_arena().address)
 
         if self.arena.address == allocator.main_arena.address:
--- a/pwndbg/heap/ptmalloc.py
+++ b/pwndbg/heap/ptmalloc.py
@@ -312,6 +312,49 @@ class Chunk:
         return self._is_top_chunk
 
 
+class Heap:
+    def __init__(self, addr=None):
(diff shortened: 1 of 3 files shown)


pwndbg/heap/ptmalloc.py
```diff
@@ -312,6 +312,49 @@ def is_top_chunk(self):
...
+                == heap_region
+            ):
+                page.vaddr += (
+                    allocator.malloc_state.sizeof + allocator.size_sz
+                ) & ~allocator.malloc_align_mask
+            page.memsz = heap_region.end - page.start
+            heap_region = page
+
+        self.start = heap_region.start
+        self.end = heap_region.end
```
disconnect3d, line 356: Is there any reason we create a `Page` object if what we only care is start and end?

pwndbg/heap/ptmalloc.py
```diff
@@ -312,6 +312,49 @@ def is_top_chunk(self):
...
+                ar_ptr.fetch_lazy()
+            except Exception:
+                ar_ptr = None
+
+            if ar_ptr is not None and ar_ptr in (ar.address for ar in allocator.arenas):
+                self.arena = Arena(ar_ptr)
+            else:
+                self.arena = Arena(allocator.main_arena.address)
+        else:
+            self.arena = Arena(allocator.get_arena().address)
```
disconnect3d, line 334: Can we add some comments here to describe the implementation choices we do here?
- Why do we have 3 different cases for choosing/instantiating arenas?
- When can we fail to obtain `allocator.get_heap(addr)["ar_ptr"]`?
- Any details about finding a "non-main arena"

disconnect3d (approved):
:+1:, we can merge it now though I'd love if we could include additional comments to the logic we have here ;p

pwndbg/heap/ptmalloc.py
CptGibbon: Previously `heap_region` was a `Page` so I'm just keeping it as similar to the original implementation as I can for now in case any issues arise from changing it.

pwndbg/heap/ptmalloc.py
CptGibbon: I agree this is opaque (gimme a break I wrote this with Serenade 😂)
It's existing logic that again I wasn't keen on messing with too much in case it caused issues.
I intend to change a lot of this and consolidate all the redundant code in one place shortly, at which point I'll ensure everything is appropriately commented👍

PR #1319, opened 2022-10-22
Implementing vermin into CI

### LINKED ISSUE #1182: Add `vermin` to CI
This project does a pretty good job of detecting all the Python features you're using and outputting the minimum version required: https://github.com/netromdk/vermin

We should add this to CI so we don't accidentally break Python 3.7 with some new features we use.

mazfil:
Closes #1182.

`lint.sh` checks the project minimum python version is `3.6`. Vermin version `1.4.2` added to `dev-requirements.txt`.

There are 3 files which contain `# novermin` comments as vermin has trouble parsing unions (`pwndbg/disasm/arch.py`, `pwndbg/disasm/__init__.py`, `pwndbg/commands/flags.py`).


lint.sh
```diff
@@ -42,3 +42,6 @@ flake8 --show-source ${LINT_FILES}
 
 # Indents are four spaces, binary ops can start a line, and indent switch cases
 shfmt -i 4 -bn -ci -d .
+
+# Checking minimum python version
+vermin -t=3.6 --violations --eval-annotations ./pwndbg/
```
gsingh93: Why do we need `--eval-annotations`?

gsingh93 (commented):
LGTM, just a question on the flags used.

lint.sh
mazfil: If that flag isn't necessary it can be removed. The flag had no effect and was put in if required by some of the code.

mazfil: I've gone ahead and removed the flag.

PR #1322, opened 2022-10-22
made codecov optional, enable it with -c or --cov

pogen300:
Now codecov does not run by default when running tests.sh.
Enable it with -c or --cov flag
Fixes issue 1229


.github/workflows/tests.yml
```diff
@@ -37,7 +37,7 @@ jobs:
       run: |
         mkdir .cov
         sudo sysctl -w kernel.yama.ptrace_scope=0
-        PWNDBG_GITHUB_ACTIONS_TEST_RUN=1 ./tests.sh
+        PWNDBG_GITHUB_ACTIONS_TEST_RUN=1 ./tests.sh -c
```
gsingh93: nit: I know we're not consistent about this right now, but I'd like to start using the long form of options when adding them to scripts: `./tests.sh --cov`

tests/gdb-tests/tests.sh
```diff
@@ -7,16 +7,18 @@ COVERAGERC_PATH="$ROOT_DIR/pyproject.toml"
 help_and_exit() {
     echo "Usage: ./tests.sh [-p|--pdb] [<test-name-filter>]"
```
gsingh93: You'll need to add `[-c | --cov]` to this string as well.

tests/gdb-tests/tests.sh
```diff
@@ -67,7 +74,11 @@ tests_passed_or_skipped=0
 tests_failed=0
 
 for test_case in ${TESTS_LIST}; do
-    gdb_args=(-ex 'py import coverage;coverage.process_startup()' --command $GDB_INIT_PATH --command pytests_launcher.py)
+    if [ ${RUN_CODECOV} -eq 0 ]; then
+        gdb_args=(--command $GDB_INIT_PATH --command pytests_launcher.py)
+    else
+        gdb_args=(-ex 'py import coverage;coverage.process_startup()' --command $GDB_INIT_PATH --command pytests_launcher.py)
+    fi
```
gsingh93: In the interest of reducing the amount of code we repeat, we can prepend arguments here instead (this is untested):
```bash
gdb_args=(--command $GDB_INIT_PATH --command pytests_launcher.py)
if [ ${RUN_CODECOV} -ne 0 ]; then
        gdb_args=(-ex 'py import coverage;coverage.process_startup()' "${gdb_args[@]}")
fi
```

gsingh93 (commented):
Overall looks good, just a few minor comments.

PR #1324, opened 2022-10-23
Refactor telescope, hexdump, and enhance to use ColorConfig

gsingh93:
@disconnect3d there are a few config options that didn't properly follow the naming convention between the parameter name and the function name. I've made these consistent:
- separator -> offset-separator
- offset -> offset-delimiter
- integer -> integer-value
- integer -> string-value

Also, `hexdump-highlight-group-lsb` was essentially a color parameter but was not named like one. The name sounds like it's a boolean, but you can customize how you want the lsb to be colored (the default is "underline"), so I renamed it to `hexdump-lsb-color`.

With these renamings, how do we not break existing themes? I'd like to display a message if anyone is explicitly setting a deprecated paramater.

Maybe something like `DeprecatedParameter('hexdump-highlight-group-lsb', 'hexdump-lsb-color')`, which would detect when someone tries to access the old name for the first time, print a deprecation warning, and then re-map it to the correct config name. Then we can remove the old config in the future.


pwndbg/commands/hexdump.py
```diff
@@ -8,6 +8,28 @@
 import pwndbg.gdblib.memory
```
captain-hoity-toity, line 8: The code is well-written and follows good practices in general. However, there are a few areas that could be improved:

1. **Error Handling:**

    In the function `address_or_module_name`, if the input doesn't match the expected types, it raises an `ArgumentTypeError`. However, the error message is not informative. It would be helpful to include the type of the received input in the error message to help in debugging.

    ```python
    else:
        raise argparse.ArgumentTypeError(f"Unknown hexdump argument type: {type(gdbval_or_str)}")
    ```

2. **Magic Numbers:**

    The code uses a magic number `0x10000` in the `hexdump` function. It would be better to define this as a constant at the top of your script, which would make your code more readable and maintainable.

    ```python
    MINIMUM_ADDRESS = 0x10000
    ...
    if address > MINIMUM_ADDRESS and count > address:
    ```

3. **Code Comments:**

    While the code is well-commented overall, the comment on the line `# TODO: What if arch endian is big, and use_big_endian is false?` indicates that there might be an edge case that has not been handled. It would be beneficial to address this point or provide more information about the potential issue.

4. **Docstrings:**

    The functions `address_or_module_name` and `hexdump` could use docstrings to describe what they do, their parameters, and their return values.

Other than these points, the code looks good and seems to follow good practices such as using descriptive variable names and keeping functions small and focused.

PR #1332, opened 2022-10-23
Run tests in parallel with GNU parallel

### LINKED ISSUE #1330: Run tests in parallel
We have lots of tests now, which is great, but the issue is `./tests.sh` takes a long time. We should run the tests in parallel to improve performance.

One way is just to use `&` to run GDB in individual Bash jobs, but we would need to collect the output from stdout/stderr and output it at the end, at least for the failing jobs.

gsingh93:
Note: this isn't ready to be merged. The tests run in parallel, but I still need to figure out how to update the test failed/success count.

Also need to add parallel to the dev dependencies.

Fixes #1330

### DIFF
--- a/setup-dev.sh
+++ b/setup-dev.sh
@@ -33,7 +33,8 @@ install_apt() {
         libc6-dev \
         curl \
         build-essential \
-        gdb
+        gdb \
+        parallel
 
     if [[ "$1" == "22.04" ]]; then
         sudo apt install shfmt
--- a/tests/gdb-tests/tests.sh
+++ b/tests/gdb-tests/tests.sh
@@ -75,7 +75,9 @@ tests_failed=0
 
 declare -a FAILED_TESTS
 
-for test_case in ${TESTS_LIST}; do
+run_test() {
+    test_case="$1"
+
     gdb_args=(--command $GDB_INIT_PATH --command pytests_launcher.py)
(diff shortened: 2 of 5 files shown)


gsingh93:
With the latest commit the test pass/fail count is fixed. But I still need to get the names of the failed tests, which is feature I just added that now doesn't work with this approach 😅.

Also, for some reason ~5 tests fail when run under parallel, I'm not sure why yet.

gsingh93:
The latest commit fixes some issues where we used hardcoded /tmp paths in our tests, which obviously break with multiple tests using the same file. This was fixed by using `tempfile.mktemp` instead.

The last test failures look like this:
```
>       assert re.match(r"fd:\s+0x1 \((/dev/pts/\d+|pipe:\[\d+\])\)", line_fd)
E       AssertionError: assert None
E        +  where None = <function match at 0x7fbf446005e0>('fd:\\s+0x1 \\((/dev/pts/\\d+|pipe:\\[\\d+\\])\\)', 'fd:        0x1 (/tmp/parZ4YC4.par (deleted))')
E        +    where <function match at 0x7fbf446005e0> = re.match
```

gsingh93:
Ok, all bugs are fixed now. This is ready for review and can be merged if it looks good.

It would also be good to only display the output of failing jobs, but I will make that in a followup PR, as that's new functionality.

Issue #1327, opened 2022-10-23
heap command can't run on Ubuntu18.04 which is a docker container


JD1412:
# heap command can't run on Ubuntu18.04 which is a docker container

## My Environment

I install the pwndbg in a docker container,the Container is based on Ubuntu18.04 system

## My Problem

when I want to debug an ELF file by the heap command,for example heap,bins,Unsortedbin,vis_heap_chunks
the pwndbg give me tips like this

<img width="540" alt="image" src="https://user-images.githubusercontent.com/30342224/197388693-ac02c2a5-8217-4d2e-ad19-59fe61b79aed.png">

so I run the command 

<img width="807" alt="image" src="https://user-images.githubusercontent.com/30342224/197388727-cede1876-37a5-4fb7-9f18-9c8a7402a809.png">

then I run the heap command again 

<img width="881" alt="截屏2022-10-23 19 11 05" src="https://user-images.githubusercontent.com/30342224/197400612-de58b853-03c9-417c-966c-9f22aa25ad16.png">

I have no idea to solve the problem

lebr0nli:
Hi, thanks for reporting the bug!
I guess the hint about the heap heuristic probably misled you.
Could you re-run the `heap` command instead of `set resolve-heap-via-heuristic on` after you `set resolve-heap-via-heuristic on` once, and test if that work for you?
Could you also `set exception-verbose on` and show the full stacktrace when you use the `heap` command?

Thanks!

JD1412:
Oh my god,I found an error in my question,I paste the same picture in different step. I have corrected that mistake.
I try it again.
<img width="890" alt="截屏2022-10-23 23 24 12" src="https://user-images.githubusercontent.com/30342224/197400763-bd1e7e24-1416-40fe-9e53-8bbfac348554.png">

when I run the command `set exception-verbose on`,I got this screenshot
<img width="899" alt="截屏2022-10-23 23 27 39" src="https://user-images.githubusercontent.com/30342224/197400953-a9810a89-76ff-4ed4-a46b-a6fce122af8e.png">

lebr0nli:
@JD1412 , thanks!

I just reproduced this bug on Ubuntu 18.04 docker container now, the error seems occurred because the version of the GDB in the container is less than 9.x(the GDB version in my 18.04 container is 8.1.1), but `gdb.look_up_static_symbol`  required GDB 9.x. Will try to let it supports GDB 8.x without this error, thanks for pointing out this issue!

[...]

PR #1333, opened 2022-10-23
Fix some issues about GDB < 9.x

lebr0nli:
This PR fixes some issues found in #1327 by:

* Return `None` if gdb doesn't have `gdb.lookup_static_symbol` when calling `pwndbg.gdblib.symbol.static_linkage_symbol_address`.

* Skip the tests for `gdb.PARAM_ZUINTEGER_UNLIMITED` if gdb doesn't have this attribute.

* Inform the user to `set exception-* on` when there's an error occurred when running the heap commands.

* A bug fix for `get_heap_boundaries()` of `HeuristicHeap` (found this in the fail test about the heap heuristics in that container)

### DIFF
--- a/pwndbg/gdblib/symbol.py
+++ b/pwndbg/gdblib/symbol.py
@@ -248,12 +248,15 @@ def static_linkage_symbol_address(symbol: str) -> int:
     Get the address for static linkage `symbol`
     """
 
-    # GDB < 9.x does not have `gdb.lookup_static_symbol`
-    if not hasattr(gdb, "lookup_static_symbol"):
-        return None
-
     try:
-        symbol_obj = gdb.lookup_static_symbol(symbol)
+        if hasattr(gdb, "lookup_static_symbol"):
+            symbol_obj = gdb.lookup_static_symbol(symbol)
+        else:
+            # GDB < 9.x does not have `gdb.lookup_static_symbol`
+            # We will fallback to `gdb.lookup_symbol` here, but the drawback is that we might find incorrect symbol if there is a symbol with the same name which is not static linkage
+            # But this is better than just returning None
+            # TODO/FIXME: Find a way to get the static linkage symbol's address in GDB < 9.x
+            symbol_obj = gdb.lookup_symbol(symbol)[0]
         return int(symbol_obj.value().address) if symbol_obj else None
     except gdb.error:
         return None
--- a/pwndbg/gdblib/symbol.py
+++ b/pwndbg/gdblib/symbol.py
@@ -243,14 +243,14 @@ def address(symbol: str) -> int:
 
 @pwndbg.lib.memoize.reset_on_objfile
 @pwndbg.lib.memoize.reset_on_thread
-def static_linkage_symbol_address(symbol):
-    if isinstance(symbol, int):
-        return symbol
+def static_linkage_symbol_address(symbol: str) -> int:
+    """
+    Get the address for static linkage `symbol`
+    """
 
-    try:
-        return int(symbol, 0)
-    except Exception:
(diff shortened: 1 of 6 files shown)


tests/gdb-tests/tests/test_gdblib_parameter.py
```diff
@@ -26,6 +34,10 @@
 def test_gdb_parameter_default_value_works(start_binary, params):
     name_suffix, default_value, displayed_value, optional_kwargs = params
 
+    # GDB < 9.x does not support PARAM_ZUINTEGER_UNLIMITED
+    if name_suffix == "unlimited-zuint" and not optional_kwargs:
+        pytest.skip("GDB < 9.x does not support `gdb.PARAM_ZUINTEGER_UNLIMITED`")
```
disconnect3d: Maybe just: `if not params: pytest.skip("Current GDB version does not support this testcase")` at the top?

pwndbg/gdblib/symbol.py
```diff
@@ -243,14 +243,14 @@ def address(symbol: str) -> int:
...
+    Get the address for static linkage `symbol`
+    """
 
-    try:
-        return int(symbol, 0)
-    except Exception:
-        pass
+    # GDB < 9.x does not have `gdb.lookup_static_symbol`
+    if not hasattr(gdb, "lookup_static_symbol"):
+        return None
```
disconnect3d: Can't we fallback to `gdb.lookup_symbol` in such case? or does this happen elsewhere?

Ideally, we do not want to make this check here and instead e.g. create a function based on GDB version like:
```py
def _get_symbol_address_gdb_9x(...): ...
def _get_symbol_address_gdb_10x(...): ...

static_linkage_symbol_address = _get_symbol_address_gdb_10x if gdb_version >= 10 else _get_symbol_address_gdb_9x
```

lebr0nli: > Can't we fallback to `gdb.lookup_symbol` in such case?

Hmm, yes and no, because `gdb.lookup_symbol` might not have the same results as `gdb.lookup_static_symbol`.

> Ideally, we do not want to make this check here and instead e.g. create a function based on GDB version like:

Yea, I also think what you said will be better, but I didn't find a way to make GDB < 9.x have the same results as `gdb.lookup_static_symbol`

tests/gdb-tests/tests/test_gdblib_parameter.py
lebr0nli: Yes, this looks better.
I'll change it this way.

pwndbg/gdblib/symbol.py
disconnect3d: Iirc the drawback of `gdb.lookup_symbol` is that we may find incorrect symbol, defined as a type in the program code, is that right? Though that won't happen that often.

If this is what it was about, then we should still fallback to `gdb.lookup_symbol` because with the current code we will not find the symbol at all, since we return `None`? Or am I missing something here and this would somehow magically work in the end?

lebr0nli: > Iirc the drawback of `gdb.lookup_symbol` is that we may find incorrect symbol, defined as a type in the program code, is that right? Though that won't happen that often.

Yes, you are right, this is the only drawback of it.

> If this is what it was about, then we should still fallback to `gdb.lookup_symbol` because with the current code we will not find the symbol at all, since we return `None`?

Yeah, it seems better that we fallback to `gdb.lookup_symbol` instead of just returning `None` here, although there's a little drawback. I'll make it fallback to `gdb.lookup_symbol` until we find a better solution.

PR #1335, opened 2022-10-23
Don't install shfmt if running in Docker

gsingh93:


gsingh93:
This works, but I don't think it's worth it. Snaps still take forever to install and take up a large amount disk space. I'll leave things as is, and just manually format scripts after PRs are submitted.

PR #1337, opened 2022-10-24
Port pwntools cyclic into pwndbg

szymex73:
This PR ports the cyclic command from pwntools into a built-in command inside pwndbg allowing for tab completion as well as using gdb expressions during lookups.

The biggest difference from the original command is that the subsequence length now defaults to the architecture pointer size instead of the original 4 by default. This allows for straight-up calling `cyclic -l $rdi` for example without explicitly specifying `-n 8` on x86-64 system.


pwndbg/commands/cyclic.py
```diff
@@ -0,0 +1,88 @@
...
+
+@pwndbg.commands.ArgparsedCommand(parser, command_name="cyclic")
+def cyclic_cmd(alphabet, length, lookup, count):
+    if length:
+        # Convert from gdb.Value
+        length = int(length)
+    else:
+        # Placing the default here as placing it in the arguments
+        # will default to 4 for some reason
+        length = pwndbg.gdblib.arch.ptrsize
```
disconnect3d, line 58: The exact reason for this is that expressions used for default arguments are evaluated right away in the moment when the function is created. You can see this here:
image

And a text form:
```
In [1]: import dis

In [2]: dis.dis('def foo(x=some_func()): pass')
  1           0 LOAD_NAME                0 (some_func)
              2 CALL_FUNCTION            0
              4 BUILD_TUPLE              1
              6 LOAD_CONST               0 (<code object foo at 0x1101c6d40, file "<dis>", line 1>)
              8 LOAD_CONST               1 ('foo')
             10 MAKE_FUNCTION            1 (defaults)
             12 STORE_NAME               1 (foo)
             14 LOAD_CONST               2 (None)
             16 RETURN_VALUE

Disassembly of <code object foo at 0x1101c6d40, file "<dis>", line 1>:
  1           0 LOAD_CONST               0 (None)
              2 RETURN_VALUE
```

As you can see, during the execution of code that defines the function, the default argument is evaluated and so the `some_func()` is called and its result is stored into a tuple with 1 elements (`BUILD_TUPLE 1`), then the actual code of the function is loaded and the function is created with its default arguments and then it is stored in `foo` name.

This is often called a "mutable defaults arguments trap". See e.g.:
```py
In [12]: def bar(x, y=[]):
    ...:     y.append(x)
    ...:     return y
    ...:

In [13]: bar(1)
Out[13]: [1]

In [14]: bar(2)
Out[14]: [1, 2]

In [15]: bar(3)
Out[15]: [1, 2, 3]
```

pwndbg/commands/cyclic.py
```diff
@@ -0,0 +1,88 @@
...
+
+    if lookup:
+        lookup = pwndbg.commands.fix(lookup, sloppy=True)
+
+        if type(lookup) in [gdb.Value, int]:
```
disconnect3d, line 63: You should usually use `isinstance(obj, type_or_types)` but this is fine as it matches EXACT type; `isinstance` would match across inheritance

pwndbg/commands/cyclic.py
disconnect3d: Can we change the comment since we know the exact reason?

pwndbg/commands/cyclic.py
```diff
@@ -0,0 +1,88 @@
...
+        elif type(lookup) is str:
+            lookup = bytes(lookup, "utf-8")
+
+        if len(lookup) != length:
+            print(message.error(f"Lookup pattern must be {length} bytes"))
+            return
+
+        print(message.notice(f"Lookup value: {str(lookup)}"))
+
+        if not all(c in alphabet for c in lookup):
```
disconnect3d:
```suggestion
        if any(c not in alphabet for c in lookup):
```

This way, we don't process all the lookup string characters and instead we stop on the first incorrect one

pwndbg/commands/cyclic.py
szymex73: The comment was more about the fact it resolves to `4` rather than why it's set to a static value since I couldn't find that out, but I see the point. Will change that.

pwndbg/commands/cyclic.py
szymex73: Will keep that in mind in the future :)

pwndbg/commands/cyclic.py
```diff
@@ -0,0 +1,86 @@
...
+    "--alphabet",
+    metavar="charset",
+    default=string.ascii_lowercase,
+    type=str.encode,
+    help="The alphabet to use in the cyclic pattern",
+)
+
+parser.add_argument(
+    "-n",
+    "--length",
```
gsingh93, line 25: Minor thing, length is a confusing name for this argument, as the pwntools cyclic function uses this as the total length to generate. Maybe something like `subseq_len` would be better here, but open to other suggestions.

pwndbg/commands/cyclic.py
gsingh93: It's because in init.py we load the commands before gdblib, so the arch is not set until later.

This was unintentional, I'll check if I can flip the order without breaking things (the current situation is at least better than the old one, where all imports were alphabetical so commands would get initialized in the middle of other modules being initialized).

PR #1339, opened 2022-10-24
Add feature to deprecate config options and refactor pwndbg.color.context

gsingh93:
The changes to lib/config.py and gdblib/config.py add a way to deprecate configuration parameters. When a configuration parameter is deprecated, the first time a user tries to set it, they'll see a warning telling them the new parameter name. Any time a user tries to set a deprecated parameter, we forward it to the new parameter.

This PR also refactors pwndbg.color.context to use `ColorConfig`, and makes use of the configuration deprecation.

### DIFF
--- a/pwndbg/color/context.py
+++ b/pwndbg/color/context.py
@@ -1,107 +1,43 @@
-import pwndbg.color.theme as theme
-from pwndbg.color import generateColorFunction
+from pwndbg.color import ColorConfig
+from pwndbg.color import ColorParamSpec
 from pwndbg.gdblib import config
 
-config_prefix_color = theme.add_color_param(
-    "code-prefix-color", "none", "color for 'context code' command (prefix marker)"
+# TODO: Should probably change - to .
+c = ColorConfig(
+    "context",
+    [
+        ColorParamSpec("register", "bold", "color for registers label"),
+        ColorParamSpec("register-changed", "red", "color for registers label (change marker)"),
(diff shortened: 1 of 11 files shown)


pwndbg/color/context.py
```diff
@@ -1,107 +1,43 @@
-import pwndbg.color.theme as theme
-from pwndbg.color import generateColorFunction
+from pwndbg.color import ColorConfig
+from pwndbg.color import ColorParamSpec
 from pwndbg.gdblib import config
 
-config_prefix_color = theme.add_color_param(
-    "code-prefix-color", "none", "color for 'context code' command (prefix marker)"
+# TODO: Should probably change - to .
```
gsingh93, line 5: Before making this change we should decide on exactly how we want the namespacing to look. Should a config option appear as `context-code-prefix-color`, `context.code-prefix-color`, or `context.code.prefix-color`? I prefer the last option, but I'm also OK with the second option.

An alternative here is I can disable the warning (while keeping the forwarding to the new parameters), and then we can just figure out how we want to do this later.

gsingh93: @disconnect3d Ping on this question.

disconnect3d: Another option is: `context code prefix-color` but I think keeping it without spaces may make more sense.

Anyway, `context.code.prefix-color` or `context-code-prefix-color` i think

disconnect3d (approved):

Issue #1280, opened 2022-10-13
Replace the implementation of `pwndbg.commands.windbg.dX` with `pwndbg.hexdump.hexdump`


gsingh93:
It seems like `pwndbg.commands.windbg.dX` duplicates a lot of functionality with `pwndbg.hexdump.hexdump`. If we add in a few more configuration options to `pwndbg.hexdump.hexdump`, I think we can implement `dX` with a single call to `hexdump`.

ruijia-zhou:
Hi, @gsingh93 , I have implemented this and passed all the tests. Can you please assign this to me such that I can submit a pull request?

disconnect3d:
@ruijia-zhou Fwiw you don't need to be assigned to an issue to send a PR, but yeah, assigned you :).

PR #1341, opened 2022-10-24
Refactor pwndbg.commands.windbg.dX and pwndbg.hexdump.hexdump

ruijia-zhou:
Replace the implementation of pwndbg.commands.windbg.dX with pwndbg.hexdump.hexdump, see issue #1280. The code in dX has been moved to dexdump. Some optional options have been added to the function signature of hexdump such that dX is implemented using a single call to hexdump.

### DIFF
--- a/pwndbg/commands/windbg.py
+++ b/pwndbg/commands/windbg.py
@@ -6,6 +6,7 @@ import argparse
 import codecs
 import math
 from builtins import str
+from itertools import chain
 
 import gdb
 
@@ -15,16 +16,7 @@ import pwndbg.gdblib.memory
 import pwndbg.gdblib.strings
 import pwndbg.gdblib.symbol
 import pwndbg.gdblib.typeinfo
-
-
-def get_type(size):
-    return {
-        1: pwndbg.gdblib.typeinfo.uint8,
-        2: pwndbg.gdblib.typeinfo.uint16,
-        4: pwndbg.gdblib.typeinfo.uint32,
-        8: pwndbg.gdblib.typeinfo.uint64,
-    }[size]
-
+import pwndbg.hexdump
 
 parser = argparse.ArgumentParser(description="Starting at the specified address, dump N bytes.")
 parser.add_argument(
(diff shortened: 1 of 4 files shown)


gsingh93:
@ruijia-zhou, try running `./lint.sh -f` to fix the lint issues, more details here: https://github.com/pwndbg/pwndbg/blob/dev/DEVELOPING.md#linting

ruijia-zhou:
Thanks for the information. Sorry for this. I am trying now.

ruijia-zhou:
@gsingh93 thanks a lot for the information. Just passed all checks.

pwndbg/commands/windbg.py
```diff
@@ -141,50 +132,18 @@ def dX(size, address, count, to_string=False, repeat=False):
...
+    lines = list(
+        chain.from_iterable(
+            pwndbg.hexdump.hexdump(
+                data=None, size=size, count=count, address=address, repeat=repeat, dX_call=True
+            )
+        )
+    )
 
-    if not to_string:
+    if not to_string and lines != []:
```
gsingh93: nit: the "Pythonic" way to do this is `if not to_string and lines`: https://stackoverflow.com/questions/53513/how-do-i-check-if-a-list-is-empty

I'd also accept `len(lines) != 0`, just because it'll catch type issues where `lines` is not a list (even though this isn't considered "Pythonic").

pwndbg/hexdump.py
```diff
@@ -2,18 +2,31 @@
...
 import pwndbg.color.theme as theme
 import pwndbg.gdblib.config
+import pwndbg.gdblib.typeinfo
+from pwndbg.commands.windbg import enhex
 
 color_scheme = None
 printable = None
 
 
+def get_type(size):
```
gsingh93: Could you move this function into `pwndbg.gdblib.typeinfo`? I think it makes more sense there and we probably want to use it in other places in the codebase.

ruijia-zhou: Sure, I will do it now :)

pwndbg/commands/windbg.py
ruijia-zhou: Thanks for the advice, I will do it now

pwndbg/hexdump.py
```diff
@@ -66,51 +79,110 @@ def load_color_scheme():
...
+        values = []
+
+        if repeat:
+            count = hexdump.last_count
+            address = hexdump.last_address
+        else:
+            address = int(address) & pwndbg.gdblib.arch.ptrmask
+            count = int(count)
+
+        type = get_type(size)
```
gsingh93: `type` is a built-in Python function. In the future we'll have a linter preventing names like this, so could you rename the variable to something else? If `type` really is the best name for this, we can go with something like `type_`.

pwndbg/hexdump.py
```diff
@@ -66,51 +79,110 @@ def load_color_scheme():
...
+            for group in groupby(group_width, line):
+                for char in group:
+                    hexline.append(printable[char])
+                hexline.append(H.separator("%s" % config_separator))
+
+            yield "".join(hexline)
+
+    else:
+
+        """
```
gsingh93: Nit, generally these types of string comments only go at the top of a function, so in this case we should convert it to a normal comment with `#`.

pwndbg/hexdump.py
```diff
@@ -66,51 +79,110 @@ def load_color_scheme():
...
+        else:
+            address = int(address) & pwndbg.gdblib.arch.ptrmask
+            count = int(count)
+
+        type = get_type(size)
+
+        for i in range(count):
+            try:
+                gval = pwndbg.gdblib.memory.poi(type, address + i * size)
+                # print(str(gval))
```
gsingh93: Let's remove all the debugging code before merging this.

pwndbg/hexdump.py
```diff
@@ -66,51 +79,110 @@ def load_color_scheme():
...
+        if not values:
+            print("Could not access the provided address")
+            return
+
+        n_rows = int(math.ceil(count * size / float(16)))
+        row_sz = int(16 / size)
+        rows = [values[i * row_sz : (i + 1) * row_sz] for i in range(n_rows)]
+        lines = []
+
+        # sys.stdout.write(repr(rows) + '\n')
```
gsingh93: Same note about the debugging code.

ruijia-zhou:
Hi, @gsingh93 ,I have revised the code. Please have a look. Thanks!

Issue #1306, opened 2022-10-20
Install a git pre-push lint hook as part of setup-test-tools.sh


gsingh93:
When running `./setup-dev.sh`, we should prompt the user (if we detect a tty?) if they want to install a git pre-push hook (if we detect one doesn't already exist?) that automatically runs `./lint.sh -f` for them.

ruijia-zhou:
Hi, @gsingh93 , I just used lint for my pull request. I think I can help with this. 
I have successfully implemented this locally. Can you please assign this issue to me (the assignment is a requirement from one of my university courses)? 
Thanks

gsingh93:
Thanks @ruijia-zhou! Just FYI, if you run into any issues feel free to join our Discord server (link in README) and ask any questions you have.

Also, it would be good to give a more descriptive message in the hook that explains why there is a failure (people may not be familiar with hooks).

ruijia-zhou:
@gsingh93 thanks, I will join now.

PR #1345, opened 2022-10-25
Install a git pre-push lint hook as part of setup-test-tools.sh

ruijia-zhou:
See issue #1306 .

When running `./setup-dev.sh`, it is helpful to prompt the user if they want to install a git pre-push hook that automatically runs linting for them. The implementation uses a bash select to allow the user to choose whether to auto-format their code by linting (by typing the number 1 or 2). If the user selects yes. A pre-hook file will be created under `.git/hook/` with a command `./lint.sh -f` to run the lint.


setup-dev.sh
```diff
@@ -5,6 +5,24 @@ echo "# Install testing tools."
 echo "# Only works with Ubuntu / APT."
 echo "# --------------------------------------"
 
+hookScriptName=".git/hooks/pre-push"
+
+if [ -t 1 ]; then
+    echo "Install a git pre-push hook?"
```
gsingh93: Let's make sure it's clear what this hook is going to do: "Install a git hook to automatically lint files before pushing? (y/N) "

setup-dev.sh
```diff
@@ -5,6 +5,24 @@ echo "# Install testing tools."
...
+    select yn in "Yes" "No"; do
```
gsingh93: Let's just check if the input starts with "y", in all other cases we don't do anything, so you can convert this to a single if statement.

setup-dev.sh
```diff
@@ -5,6 +5,24 @@ echo "# Install testing tools."
...
+        case $yn in
+            Yes)
+                echo "./lint.sh -f" >>$hookScriptName
+                echo "pre-push hook installed"
```
gsingh93: Let's tell the user where it's installed, so add "to $hookScriptName" here.

setup-dev.sh
```diff
@@ -5,6 +5,24 @@ echo "# Install testing tools."
 echo "# Only works with Ubuntu / APT."
 echo "# --------------------------------------"
 
+hookScriptName=".git/hooks/pre-push"
```
gsingh93: nit: shell script naming convention usually uses snake_case: hook_script_name

setup-dev.sh
```diff
@@ -5,6 +5,25 @@ echo "# Install testing tools."
 echo "# Only works with Ubuntu / APT."
 echo "# --------------------------------------"
 
+hook_script_name=".git/hooks/pre-push"
+
+if [ -t 1 ]; then
+    while true; do
+        echo "Install a git hook to automatically lint files before pushing? ([y]/N)"
```
gsingh93: The convention with `y/n` prompts is usually that the capital letter is the default (but the user can still enter either option in any case). In this case we want the `N` to be the default, so we can change this to `y/N`.

setup-dev.sh
```diff
@@ -5,6 +5,25 @@ echo "# Install testing tools."
 echo "# Only works with Ubuntu / APT."
 echo "# --------------------------------------"
 
+hook_script_name=".git/hooks/pre-push"
+
+if [ -t 1 ]; then
+    while true; do
```
gsingh93: We don't need this `while true` loop, and we can remove the `break` statements below. `;;` is all you need to break out of a `case` statement in bash.

setup-dev.sh
```diff
@@ -5,6 +5,25 @@ echo "# Install testing tools."
...
+        case $yn in
+            [Yy]* | "")
+                echo "./lint.sh -f" >>$hook_script_name
+                echo "pre-push hook installed to $hook_script_name"
+                break
+                ;;
+            [Nn]*)
+                break
+                ;;
+        esac
```
gsingh93: Since we want `N` to be the default, and by default we don't need to do anything, we can simplify this to (untested):
```bash
if [[ "$yn" == [Yy]* ]]; then
...
fi
```

gsingh93 (commented):
Overall looks good now, just need to clean up the while loop/case statement.

PR #1346, opened 2022-10-25
Reduce heap code redundancy

### LINKED ISSUE #1143: GLIBC_TUNABLES can confuse some heap commands
### Description
If GLIBC's `GLIBC_TUNABLES` [environment variable](https://www.gnu.org/software/libc/manual/html_node/Tunables.html) is set, ld copies its contents to memory requested via `sbrk()` during program initialization.

Because malloc's main arena also relies on `sbrk()` to request memory, the start of the main arena's heap is determined by the length of the `GLIBC_TUNABLES` string rounded up to the architecture's chunk alignment.

Importantly `mp_.sbrk_base` *isn't* rounded up, this affects some of pwndbg's heap commands such as `vis_heap_chunks` and `heap` because they rely on `mp_.sbrk_base` to find the start of the main arena's heap. Below is an example of how this issue can manifest:

glibc tunables issue

After setting the `GLIBC_TUNABLES` environment variable via GDB, the above program requests a single 0x20-size chunk. The `GLIBC_TUNABLES` string starts at the beginning of the `[heap]` region, followed by the main arena's heap at address `0x555555559020` which contains a tcache and 0x20-size chunk. Note that `mp_.sbrk_base` points to the end of the `GLIBC_TUNABLES` string, causing incorrect `vis_heap_chunks` and `heap` output. Be aware when testing this that later versions of GLIBC will reject incorrectly named tunables.

### Possible solution
When determining the start address of the main arena's heap, rounding `mp_.sbrk_base` up to the appropriate chunk alignment for the target architecture could mitigate this issue.

CptGibbon:
This is a first attempt at fully integrating the new heap classes with existing heap commands.
Hopefully it reduces some duplicate code and makes it easier to write/maintain heap commands.
Yes it's a bit on the large side, it snowballed somewhat. Apologies to reviewers.

Main changes:
* More accurate arena detection (less main arena false positives, non-contiguous main arena support)
* More accurate heap detection (fixes #1143)
* Removed ambiguous methods (`get_arena()` & `get_heap_boundaries()`)
* Quality of life stuff e.g. `for chunk in heap:` 🙌

There's still some work to do:
* I break encapsulation twice to maintain original command functionality
* `arenas` command no longer renames heaps (requires discussion)
* `Bin` classes still require tighter integration
* Need tests for commands like `heap` and `arenas` (I can write these soon 👍)

The failing test is only on Ubuntu 22.04, and is caused by the heuristic version of `mp` failing to resolve `_mp_addr` correctly. I'm  hoping @lebr0nli will understand what's going on there.


pwndbg/heap/ptmalloc.py
```diff
@@ -1741,31 +1781,17 @@ def thread_cache(self):
...
+                    if pwndbg.gdblib.vmmap.find(thread_cache_struct_addr):
                         self._thread_cache = self.tcache_perthread_struct(thread_cache_struct_addr)
                         return self._thread_cache
 
             # If we still can't find the tcache, we guess tcache is in the first chunk of the heap
             # Note: The result might be wrong if the arena is being shared by multiple threads
             # And that's why we need to find the tcache address in TLS first
-            arena = self.get_arena()
-            heap_region = self.get_heap_boundaries()
+            arena = self.thread_arena or self.main_arena
```
lebr0nli: After the refactoring, we used many `allocator.thread_arena or allocator.main_arena` and `self.thread_arena or self.main_arena` in different places.

I think it's suitable if we have debug symbols, but if the target binary is single-threaded and doesn't have debug symbols, is it possible that we only use `main_arena` here?

Because the heuristic for `thread_arena` is more easily to fail than the heuristic of `main_arena`, and if it failed when resolving `thread_arena` here, we will get an`SymbolUnresolvableError` even if we are possible to get the correct result with `main_arena`.

So, how about we add a new property called `current_arena` for `DebugsymsHeap` and `HeuristicHeap`, and replace `allocator.thread_arena or allocator.main_arena` with `allocator.current_arena`?

Something like:

```python
class DebugSymsHeap(GlibcMemoryAllocator):
    @propery
    def current_arena(self):
        return self.thread_arena or self.main_arena
```

and

```python
class HeuristicHeap(GlibcMemoryAllocator):
    @propery
    def current_arena(self):
        if not self.multithreaded:
            # We return `main_arena` instead of `thread_arena` because `main_arena` is more reliable than `thread_arena` currently
            return self.main_arena
        return self.thread_arena
```

CptGibbon: Yes I agree that because `thread_arena` can fail, checking for a multi-threaded environment before attempting to resolve it is sensible. However, I removed the `get_arena()` and `get_heap_boundaries()` methods because they were ambiguous, and I think the idea of a "current arena" is still ambiguous...

What do you think to adding a check for `self.multithreaded` into `thread_arena`, then returning the main arena if the process is single-threaded? That way the name "thread_arena" is still accurate, since you'll still be getting thread 1's arena in a single-threaded environment.

pwndbg/commands/heap.py
```diff
@@ -644,60 +625,44 @@ def find_fake_fast(addr, size=None, align=False):
...
             break
 
-        if prev_inuse:
+        if chunk.prev_inuse:
             chunk_delims.append(cursor + ptr_size)
         else:
             chunk_delims.append(cursor)
 
-        if (cursor == top_chunk and not naive) or (cursor == heap_region.end - ptr_size * 2):
+        if (chunk.is_top_chunk and not naive) or (cursor == heap_region.end - ptr_size * 2):
```
gsingh93, line 660: You can do this in this PR or in another one if you think it makes sense, but the `2*ptr_size` constant pops up a lot in this code. Maybe we should move that to a constant to help with readability? Maybe `MIN_CHUNK_SIZE`?

CptGibbon: Agreed, there's still a few things like this that need addressing.

I get the sneaking suspicion that `MIN_CHUNK_SIZE` is already implemented as `min_chunk_size()`, which returns the minimum non-fencepost chunk size (`ptrsize*4`) 🤔

Nonetheless I will certainly make this less of a magic number in the next PR 👍

pwndbg/heap/ptmalloc.py
CptGibbon: *Note to self
Address cases where malloc is initialized but the debugged thread is not attached to an arena.*

lebr0nli: > What do you think to adding a check for `self.multithreaded` into `thread_arena`, then returning the main arena if the process is single-threaded?

I think it's good, I agreed 👍

Btw, if we do this, we might need to modify the tests here: https://github.com/pwndbg/pwndbg/blob/5695658797d3284517ae4ca2ed3eeefafb9fbae1/tests/gdb-tests/tests/heap/test_heap.py#L366

to make sure the heuristic use `main_arena` correctly when single-threaded, and then assert `gdb.selected_thread().num == 2` and `*thread_arena != &main_arena` to check if the heuristic still works for multi-threaded.

(probably add a for-loop or something that we do the same tests but `continue` the process in the second iteration to enter the new thread)

CptGibbon: Good point.
My next PR focusses on writing more tests, I've noted this as one more to be added 👍

Issue #1328, opened 2022-10-23
Run all tests in Docker containers


gsingh93:
In https://github.com/pwndbg/pwndbg/issues/1327 we discovered that the version of GDB used in the CI is actually newer than what's provided in the apt repositories. This means we're going to miss bugs in older versions of GDB.

We should switch to running all of our tests inside Docker images, which will actually match a production environment. The Ubuntu 22.04 hosted worker contains cached Docker images for all of the Ubuntu versions we care about: https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#cached-docker-images

Since these machines also have cached images for Debian 11 and 10, we should also run tests on those images as well.

disconnect3d:
We can probably force to install an older version of the GDB package just for Ubuntu 18.x.

gsingh93:
I think we should go with this Docker approach anyway:
1. This is going to happen in the future with newer Ubuntu versions eventually
2. Who knows what other packages are actually different and could lead to problems
3. We want to do this anyway to support testing on Debian, so why not just do it for all of our tests

KevinN318:
Hi, I would like to try to implement this.

gsingh93:
@KevinN318 I've assigned it to you. If you need help, let us know here or in Discord (link is in the README).

disconnect3d:
This was finally merged in https://github.com/pwndbg/pwndbg/pull/1370

PR #1351, opened 2022-10-26
Run tests in Docker images

KevinN318:
This may not be the right implementation in regards to #1328 where we should switch all tests within docker images. I guess I’m confused as to what all tests are and the structure.

In this pr draft:
In the docker workflow, we run each image in a container and run the tests. Did this by adding more Dockerfiles and making changes in the workflow and docker-compose files. The images are also cached to GitHub. The original Dockerfile was unchanged and its commands can still be run. Currently, ubuntu18.04 and debian10 fails. 

Assumptions:
Where the Docker workflow was originally used to test if the docker image builds. We use this in turn to run the tests in each respective image.

The tests workflow builds the project and runs the tests on the runners. And code coverage is ran for ubuntu22.04 build. Not sure if this is needed if we are using docker, if not just for running code coverage.


disconnect3d:
Two notes:
- we should probably move all the dockerfiles into `.github` or some `ci` or `tests` directory
- the common code for all dockerfiles could be just a shell script instead of having the same content in all of them (are there any differences between them?)
- we want to keep `Dockerfile` and `docker-compose.yml` in the root directory of the repo for users to run pwndbg on our latest supported version (Ubuntu 22.04) and not to be confused which one to choose from

gsingh93:
We can also reduce the common code by parameterizing the base image, like this: https://stackoverflow.com/questions/51080541/build-2-images-with-docker-from-a-single-dockerfile-while-using-different-base-i

disconnect3d:
@gsingh93 lol nice, didn't know u can use args for FROM directive/clause

KevinN318:
@gsingh93 I was looking for something like this! Thanks, I'll implement this. @disconnect3d Thanks for feedback.

docs/pycharm_debugging.md
```diff
@@ -1,47 +1,47 @@
-Debugging with PyCharm
```
gsingh93: This file looks the same to me, why is github showing it as changed?

Dockerfile
```diff
@@ -43,4 +43,4 @@ RUN echo "source /pwndbg/gdbinit.py" >> ~/.gdbinit.py && \
 
 ADD . /pwndbg/
 
-RUN git submodule update --init --recursive
+RUN git submodule update --init --recursive
```
gsingh93: Can you add back the newline? A lot of editors enforce this, so it's going to get changed again the next time someone with that editor config touches this file.

KevinN318: will do

docs/pycharm_debugging.md
gsingh93: Can you revert this file, assuming you didn't actually make changes to it? You should be able to do `git restore -s dev docs/pycharm_debugging.md` to revert it, and then  if you add and commit it, this should disappear.

gsingh93 (changes requested):
Along with that docs file, can you remove the `.DS_Store` files (`git rm .DS_Store` and then commit) and revert the changes to `gdb-pt-dump`?

gsingh93:
Also, the issue causing the 18.04 tests to fail was fixed by https://github.com/pwndbg/pwndbg/pull/1333. You'll need to rebase with `git fetch origin && git rebase origin/dev` to get the latest changes. If you're not familiar with `git`, I suggest you also create a backup branch before trying this (`git checkout -b backup`).

KevinN318:
@gsingh93 I've done as you've requested I hope but there are still tests that fail in 18.04 not too sure what it is.

gsingh93:
I think something is going wrong when you're pushing updates. If you take a look at the files here, the DS_Store files are still there: https://github.com/pwndbg/pwndbg/pull/1351/files. And the other unrelated changes haven't been reverted. What git commands are you running?

KevinN318:
@gsingh93 Should be all good now finally.

.github/workflows/docker.yml
```diff
@@ -3,13 +3,26 @@ on: [push, pull_request]
...
-    timeout-minutes: 10
+    timeout-minutes: 30
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v2
+
+    - name: Cache for docker images
+      uses: satackey/action-docker-layer-caching@v0.0.11
+      # Ignore the failure of a step and avoid terminating the job.
+      continue-on-error: true
```
disconnect3d, line 19: Out of curiosity, when can it fail?

Dockerfile
```diff
@@ -1,13 +1,15 @@
 # This dockerfile was created for development & testing purposes
 #
-# Build as:             docker build -t pwndbg .
+# Build as:             docker build -t pwndbg .
```
disconnect3d:
```suggestion
# Build as:             docker build -t pwndbg .
```

docker-compose.yml
```diff
@@ -1,11 +1,77 @@
 version: "3.8"
 services:
   main:
-    build: .
+    build: .
```
disconnect3d:
```suggestion
    build: .
```

docker-compose.yml
```diff
@@ -9,3 +9,69 @@ services:
       - seccomp:unconfined
     cap_add:
       - SYS_PTRACE
+
+  ubuntu18.04:
+    build:
+      context: .
```
disconnect3d, line 15: Why do lines ending with "." end with a space? :P

disconnect3d:
@KevinN318 can you check if a yaml specification included below would work?

This uses [yaml anchors](https://www.educative.io/blog/advanced-yaml-syntax-cheatsheet#anchors) to avoid the duplication of common settings between image specifications :). It seems to work when I do `yaml.load(open('./a.yaml'))` on it in Python, but please double check if docker compose is fine with it as well.

```yaml
version: "3.8"
services:
  main: &base-spec
    build: .
    volumes:
      - .:/pwndbg
    platform: linux/amd64
    security_opt:
      - seccomp:unconfined
    cap_add:
      - SYS_PTRACE

  ubuntu18.04:
    <<: *base-spec
    build:
      args:
        image: ubuntu:18.04

  ubuntu20.04:
    <<: *base-spec
    build:
      args:
        image: ubuntu:20.04

  ubuntu22.04:
    <<: *base-spec
    build:
      args:
        image: ubuntu:22.04

  debian10:
    <<: *base-spec
    build:
      args:
        image: debian:10

  debian11:
    <<: *base-spec
    build:
      args:
        image: debian:11
```

gsingh93:
Maybe `extends` could also help here: https://docs.docker.com/compose/extends/#extending-services

disconnect3d:
This will be continued in https://github.com/pwndbg/pwndbg/pull/1370

PR #1353, opened 2022-10-30
Support `malloc_par` of GLIBC 2.35

lebr0nli:
This commit will add support to the new definition of `struct malloc_par` in GLIBC 2.35.

Some issues about the heuristic found in #1346 should be fixed after this PR.


CptGibbon:
Thanks for adding this, tests in #1346 pass now ❤️

PR #1367, opened 2022-11-10
Fix lexer for coloring negative numbers in asm

arcz:
This particularly affected arm asm coloring which displayed negative numbers as comments.
Before:
Screenshot 2022-11-10 at 22 20 20
After:
Screenshot 2022-11-10 at 22 19 48


disconnect3d:
Just to be clear or faster spot the change: 
image

Thanks!

disconnect3d:
Fwiw this has to be upstreamed to Pwntools as well and eventually, we should remove this file from our repo and use the implementation from Pwntools, at least if it is possible.

PR #1368, opened 2022-11-10
Use the GitHub CLI tool to create issues from bugreport

### LINKED ISSUE #1236: Use the `gh` cli tool to create issues from `bugreport` if it's installed
Github has a CLI tool that can do things like create issues: https://cli.github.com/

When running pwndbg on a server, I can't do `bugreport --run-browser` because there's no X server running. I have to go through the issue creation process manually, copying and pasting the output. We could automate all of this using `gh issue create`.

kacpi2442:
Fixes #1236

### DIFF
--- a/pwndbg/commands/version.py
+++ b/pwndbg/commands/version.py
@@ -4,9 +4,12 @@ Displays gdb, python and pwndbg versions.
 
 
 import argparse
+import os
 import sys
 from platform import platform
+from subprocess import check_call
 from subprocess import check_output
+from tempfile import NamedTemporaryFile
 from urllib.parse import quote
 
 import gdb
@@ -82,10 +85,13 @@ bugreport_parser = argparse.ArgumentParser(
 bugreport_parser.add_argument(
     "--run-browser", "-b", action="store_true", help="Open browser on github/issues/new"
 )
+bugreport_parser.add_argument(
+    "--use-gh", "-g", action="store_true", help="Create issue using Github CLI"
+)
 
 
 @pwndbg.commands.ArgparsedCommand(bugreport_parser)
-def bugreport(run_browser=False):
+def bugreport(run_browser=False, use_gh=False):
     ISSUE_TEMPLATE = """
 <!--
 Before reporting a new issue, make sure that we do not have any duplicates already open.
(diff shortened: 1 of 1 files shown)


pwndbg/commands/version.py
```diff
@@ -82,10 +85,13 @@ def version():
 bugreport_parser.add_argument(
     "--run-browser", "-b", action="store_true", help="Open browser on github/issues/new"
 )
+bugreport_parser.add_argument(
```
gsingh93: These two options are going to be mutually exclusive right? We should probably use `add_mutually_exclusive_group` : https://stackoverflow.com/questions/17909294/python-argparse-mutual-exclusive-group

pwndbg/commands/version.py
```diff
@@ -189,7 +195,20 @@ def bugreport(run_browser=False):
...
-    if run_browser:
+    if use_gh:
+        try:
+            with NamedTemporaryFile("w", delete=True) as f:
+                f.write(issue_bugreport)
+                f.flush()
+                if "EDITOR" in os.environ:
+                    editor = os.environ["EDITOR"]
+                else:
+                    editor = "vi"
```
gsingh93: You can simplify this with either `os.environ.get("EDITOR", "vi")` or `os.getenv("EDITOR", "vi")`.

gsingh93: And this isn't necessary for this PR, but in the future we should probably fallback to a config option instead of hardcoding "vi" here.

pwndbg/commands/version.py
```diff
@@ -189,7 +195,20 @@ def bugreport(run_browser=False):
...
+                check_call([editor, f.name])
+                check_call(["gh", "issue", "create", "--body-file", f.name])
+        except Exception:
+            print(please_please_submit + github_issue_url)
```
gsingh93, line 207: I'm wondering if we should we re-throw the exception after printing this, so that the user understands something went wrong with the GH CLI and it's easier for us to debug.

disconnect3d:
image

We could potentialy display info about gh cli here

pwndbg/commands/version.py
```diff
@@ -189,7 +196,17 @@ def bugreport(run_browser=False):
...
 
-    if run_browser:
+    if use_gh:
+        try:
+            with NamedTemporaryFile("w", delete=True) as f:
+                f.write(issue_bugreport)
+                f.flush()
+                check_call([os.environ.get("EDITOR", "vi"), f.name])
+                check_call(["gh", "issue", "create", "--body-file", f.name])
+        except Exception:
```
gsingh93, line 206: We want to raise the original exception, so `except Exception as e:` and then `raise e`.

PR #1378, opened 2022-11-15
Support Arch Linux docker test

### LINKED ISSUE #1212: Run tests on Arch Linux
Debian and Arch are our most used distros after Ubuntu. It would be good to run the CI tests on those distros as well. While Github doesn't support this natively, we can probably get around this with Docker.

Although, I don't think Arch failing should be a blocker for any PRs. I don't want to make contributing to the project significantly harder, so if we see that Arch breaks, we can just file an issue and follow up on it later.

Xeonacid:
~~Use zig from Pacman.~~ Use zig from upstream (like what we do for Ubuntu) because of unexpected stripped binary output.
Add debug repos to install `glibc-debug` in case of debuginfod not working.

fix #1212

### DIFF
--- a/setup-dev.sh
+++ b/setup-dev.sh
@@ -50,13 +50,17 @@ linux() {
     uname | grep -i Linux &> /dev/null
 }
 
-install_apt() {
+set_zigpath() {
     if [[ -z "$ZIGPATH" ]]; then
-        # If ZIGPATH is not set, set it to $pwd/.zig
-        # In Docker environment this should by default be set to /opt/zig
-        export ZIGPATH="$(pwd)/.zig"
+        # If ZIGPATH is not set, set it
+        # In Docker environment this should by default be set to /opt/zig (APT) or /usr/bin (Pacman)
+        export ZIGPATH="$1"
     fi
     echo "ZIGPATH set to $ZIGPATH"
+}
+
+install_apt() {
+    set_zigpath "$(pwd)/.zig"
 
     sudo apt-get update || true
     sudo apt-get install -y \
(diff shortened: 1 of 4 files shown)


Xeonacid:
Many tests fail currently. Further investigation is needed.
I wonder why [the result](https://github.com/pwndbg/pwndbg/actions/runs/3467925931/jobs/5793246304#step:5:2039) says `Tests failed: 0` ...

setup-dev.sh
```diff
@@ -108,10 +148,15 @@ if linux; then
             )
             install_apt $ubuntu_version
             ;;
+        "arch")
+            install_pacman
+            ;;
         *) # we can add more install command for each distros.
             echo "\"$distro\" is not supported distro. Will search for 'apt' or 'dnf' package managers."
```
gsingh93: We'll need to update this comment to include pacman. I guess we should remove `dnf`, I don't know why that's listed here.

setup-dev.sh
```diff
@@ -37,13 +37,6 @@ if [ -t 1 ]; then
     fi
 fi
 
-if [[ -z "$ZIGPATH" ]]; then
-    # If ZIGPATH is not set, set it to $pwd/.zig
-    # In Docker environment this should by default be set to /opt/zig
-    export ZIGPATH="$(pwd)/.zig"
-fi
-echo "ZIGPATH set to $ZIGPATH"
```
gsingh93, line 45: We can probably put this in a function and then call it from both install functions, instead of duplicating it.

setup-dev.sh
Xeonacid: Fixed

setup-dev.sh
Xeonacid: Fixed

PR #1380, opened 2022-11-15
Add tabstop config for context source code

lonnywong:
* The c code `test.c`:
```c
#include <stdio.h>

int main() {
  puts("Hello, World!");
  return 0;
}
```

* The gdb script `script.txt`:
```
dir /usr/src/glibc/glibc-2.35
set context-source-code-lines 30
b main
r
b _IO_flush_all_lockp
c
n 6
```

* The libc source code is misalignment:
```sh
gcc -g test.c
gdb -x script.txt ./a.out
```
0327fab1-b15c-4e0d-b66a-258978115a63

* `set context-source-code-tabstop 8` to make it looks better:
0d2cce01-7e7d-4652-bd8b-9b1401f879b3

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -627,6 +627,9 @@ theme.add_param("highlight-source", True, "whether to highlight the closest sour
 source_code_lines = pwndbg.gdblib.config.add_param(
     "context-source-code-lines", 10, "number of source code lines to print by the context command"
 )
+pwndbg.gdblib.config.add_param(
+    "context-source-code-tabstop", 8, "number of spaces that a <tab> in the source code counts for"
+)
 theme.add_param("code-prefix", "►", "prefix marker for 'context code' command")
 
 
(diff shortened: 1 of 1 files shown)


disconnect3d:
Should we set the default to 8? Btw any chance you can add a test for that behavior?

lonnywong:
> Should we set the default to 8? Btw any chance you can add a test for that behavior?

Spaces are better than tabs in the context, and 8 should work for most libc. I have changed it to 8.

I don't know how to add a source code test. Are there any source code test cases to learn from?

disconnect3d:
@lonnywong You would probably have to create a test that does something like:
```
start_binary(SOME_TEST_BINARY)
gdb.execute('set whatever is needed')
src = gdb.execute('ctx source', to_string=True)
assert src == ' ... '
gdb.execute('bring back the settings set')
```

disconnect3d:
Here's some example: https://github.com/pwndbg/pwndbg/blob/dev/tests/gdb-tests/tests/test_emulate.py

For running tests, u may need to run dev-setup.sh and then u can do sth like `./tests.sh <part of name of your test>` (but the argument is optional)

disconnect3d:
Merging this for now, please consider sending another PR with a test or two for that behavior :)

lonnywong:
@disconnect3d Thanks. I'll try to add some tests in the next few days.

Issue #1074, opened 2022-08-17
Move pwntools' shell commands into real pwndbg commands


disconnect3d:
As described in https://github.com/pwndbg/pwndbg/issues/1068#issuecomment-1218566488, instead of registering Pwntools' shell commands directly, we should:

1. Pull/install the pwntools package as a dependency
2. Create Pwndbg commands for the Pwntools' shell commands - with just better interface/defaults! For example:
* all commands that care about architecture, like `asm` or `disasm`, should have the debugged target architecture (`pwndbg.arch.current`) as the default one
* the `disasm` command should not read from stdin when ran with no arguments as it is now
3. Unregister the shell commands from `shell.py` that are moved to Pwndbg 

With those changes, I think we should also either remove the deprecated shell commands (all the other ones) OR at least remove them from being displayed in the `pwndbg [filter]` output as they do currently -- they basically bloat this display.

Instead of that, if we keep them still (deprecated, until future removal?), we should display them e.g. at the end of the `pwndbg [filter]` display.

gsingh93:
@disconnect3d I don't think this issue will make much progress until we specify specifically what remains to be ported. wdyt?

disconnect3d:
@k4lizen did we remove all shell commands for now? if so, maybe we can also just close this since I think we won't do the move for now?

k4lizen:
yeah i think its fine to close this

PR #1384, opened 2022-11-18
Port pwntools asm into pwndbg

Ignatella:
Port pwntools asm into pwndbg #1074

### DIFF
--- a/pwndbg/commands/asm.py
+++ b/pwndbg/commands/asm.py
@@ -1,0 +1,76 @@
+import argparse
+
+import pwnlib
+import pwnlib.context
+
+import pwndbg.commands
+from pwndbg.color import message
+
+parser = argparse.ArgumentParser(description="Assemble shellcode into bytes")
+
+parser.add_argument(
+    "-f", "--format", default="hex", choices=["hex", "string"], type=str, help="Output format"
+)
+
+parser.add_argument(
+    "--arch",
+    default=pwnlib.context.context.arch,
+    choices=pwnlib.context.context.architectures.keys(),
+    type=str,
+    help="Target architecture",
+)
+
+parser.add_argument(
+    "-v",
+    "--avoid",
+    action="append",
+    help="Encode the shellcode to avoid the listed bytes (provided as hex)",
+)
+
+parser.add_argument(
+    "-n",
+    "--newline",
+    dest="avoid",
+    action="append_const",
+    const="0a",
+    help="Encode the shellcode to avoid newlines",
+)
+
+parser.add_argument(
+    "-z",
+    "--zero",
+    dest="avoid",
+    action="append_const",
+    const="00",
+    help="Encode the shellcode to avoid NULL bytes",
+)
+
+input_group = parser.add_mutually_exclusive_group(required=True)
+
+input_group.add_argument(
+    "shellcode", default=[], nargs="*", type=str, help="Assembler code to assemble"
+)
+
+input_group.add_argument("-i", "--infile", default=None, type=str, help="Specify input file")
+
+
+@pwndbg.commands.ArgparsedCommand(parser, command_name="asm")
+def asm(shellcode, format, arch, avoid, infile):
+    if infile:
+        print(message.warn("Going to read from file: " + infile))
+        with open(infile) as file:
+            shellcode = [file.read()]
+
+    bits_for_arch = pwnlib.context.context.architectures.get(arch, {}).get("bits")
+    assembly = pwnlib.asm.asm(" ".join(shellcode), arch=arch, bits=bits_for_arch)
+
+    if avoid:
+        avoid = map(lambda byte: str(byte), avoid)
+        avoid = pwnlib.unhex("".join(avoid))
+        print(message.warn("Going to avoid these bytes in hex: " + avoid.hex(" ")))
+        assembly = pwnlib.encode(assembly, avoid)
+
+    if format == "hex":
+        assembly = assembly.hex()
+
+    print(assembly)
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -600,6 +600,7 @@ def load_commands() -> None:
     import pwndbg.commands.ai
     import pwndbg.commands.argv
     import pwndbg.commands.aslr
+    import pwndbg.commands.asm
     import pwndbg.commands.attachp
     import pwndbg.commands.auxv
     import pwndbg.commands.canary
--- a/pwndbg/commands/shell.py
+++ b/pwndbg/commands/shell.py
@@ -9,7 +9,7 @@ from pwnlib.util.misc import which
 import pwndbg.commands
 from pwndbg.commands import CommandCategory
 
-pwncmd_names = ["asm", "constgrep", "disasm", "pwn", "unhex"]
+pwncmd_names = ["constgrep", "disasm", "pwn", "unhex"]
 shellcmd_names = [
     "awk",
     "bash",


pwndbg/commands/asm.py
```diff
@@ -0,0 +1,53 @@
...
+parser.add_argument(
+    "--vma", default=0, type=int, help="Virtual memory address of the beginning of assembly"
+)
+
+parser.add_argument(
+    "--no_extract",
+    default=False,
+    action="store_true",
+    help="If set returns the path to an ELF file with the assembly embedded",
+)
```
gsingh93, line 24: Why not just have the option be `extract` with a default of `True`?

Ignatella:
@gsingh93. Extract option changed. thx :)

disconnect3d:
Interesting, the `pwn asm` cli tool has a bit different flags/api than the `pwnlib.asm.asm(...)` function and we went with the latter interface in this PR. I think we need the `asm` command in Pwndbg to support similar flags that the `pwn asm` cli tool has rather than the full set of arguments the `pwnlib.asm.asm` supports.

I mean, I think don't need the `vma`, `extract` and `shared` arguments from `pwnlib.asm.asm` function are needed for Pwndbs users when they want to execute an `asm` command:
```
    vma(int):       Virtual memory address of the beginning of assembly
    extract(bool):  Extract the raw assembly bytes from the assembled
                    file.  If :const:`False`, returns the path to an ELF file
                    with the assembly embedded.
```

From the [`asm` function code](https://github.com/Gallopsled/pwntools/blob/0a7c0b62a73f92c5831418cd916b5f43556e3434/pwnlib/asm.py#L693-L695) and some pwntools docs (here for [`disasm` function but it probably provides similar functionality](https://docs.pwntools.com/en/stable/asm.html#pwnlib.asm.disasm)) the `vma` argument seems to influence the address of the section on which the linker will put the assembled/resulting code in an ELF binary.

The `extract` argument makes it so that a path to the ELF file created by pwntools is returned instead of raw assembled bytes.

And finally, the `shared` flags will create a shard object instead of an ELF executable program?

I don't even know what are the use cases for those three flags, maybe for creating a binary with a shellcode or, in case of  `vma` to display disasm properly?

Ignatella:
Also noticed the difference between interfaces -- gonna remove mentioned options. So, do we want to have these:

```
-v AVOID, --avoid AVOID
                        Encode the shellcode to avoid the listed bytes (provided as hex)
-n, --newline         Encode the shellcode to avoid newlines
-z, --zero            Encode the shellcode to avoid NULL bytes
```

`-n` and `-z` just add corresponding symbols to `AVOID`.

disconnect3d:
@Ignatella Yeah, avoid/newline/zero sounds useful

Ignatella:
Should be fixed now :blush:

Ignatella:
Looks like there is a problem with default arch

pwndbg/commands/asm.py
```diff
@@ -0,0 +1,74 @@
+import argparse
+
+import pwn
+import pwnlib.context
+
+import pwndbg.commands
+
+parser = argparse.ArgumentParser(
+    description="Runs cpp() over a given shellcode and then assembles it into bytes"
```
disconnect3d: Description needs an update :)

pwndbg/commands/asm.py
```diff
@@ -0,0 +1,74 @@
...
+)
+
+parser.add_argument(
+    "-f", "--format", default="hex", choices=["hex", "string"], type=str, help="Output format"
```
disconnect3d, line 12: TIL there is 'choices' kwarg, nice!

pwndbg/commands/asm.py
```diff
@@ -0,0 +1,74 @@
...
+)
+
+parser.add_argument(
+    "--arch",
+    default=pwnlib.context.context.arch,
```
disconnect3d, line 17: if arch changes, does this still work? I think so, but its good to double check that 🤐; I think we can assume that it does so.

pwndbg/commands/asm.py
```diff
@@ -0,0 +1,74 @@
...
+def asm(shellcode, format, arch, avoid, infile):
+    if infile:
+        with open(infile) as file:
+            shellcode = [file.read()]
+
+    assembly = pwnlib.asm.asm(" ".join(shellcode), arch=arch)
+
+    if avoid:
+        avoid = pwn.unhex("".join(avoid))
+        print(avoid)
```
disconnect3d: Maybe a more descriptive print, like "Going to avoid these bytes in hex:"

pwndbg/commands/asm.py
```diff
@@ -0,0 +1,74 @@
...
+    "shellcode", default=[], nargs="*", type=str, help="Assembler code to assemble"
+)
+
+input_group.add_argument("-i", "--infile", default=None, type=str, help="Specify input file")
+
+
+@pwndbg.commands.ArgparsedCommand(parser, command_name="asm")
+def asm(shellcode, format, arch, avoid, infile):
+    if infile:
+        with open(infile) as file:
```
disconnect3d, line 61: can we have a warning print if shellcode is set to anything that it will be ignored and instead the asm will be read from infile?

pwndbg/commands/asm.py
Ignatella: at least pwnlib shows error, that utils for these architecture are not installed :)

Ignatella:
@disconnect3d please verify :)

pwndbg/commands/asm.py
disconnect3d: This default value does not work out well: it gets set once and is not re-fetched when the `update` function from the pwndbg/pwndbg/gdblib/arch.py module is called during one of our hooks execution.

You can check this out by doing:
```
root@pwndbg-dev:~/pwndbg# gdb --quiet /bin/ls --ex 'asm nop'
pwndbg: loaded 140 pwndbg commands and 47 shell commands. Type pwndbg [--shell | --all] [filter] for a list.
pwndbg: created $rebase, $ida GDB functions (can be used with print/break)
Reading symbols from /bin/ls...
Debuginfod has been disabled.
(No debugging symbols found in /bin/ls)
Exception occurred: asm: Invalid arch/bits combination: i386/64 (<class 'AttributeError'>)
GDB's apropos <topic> command displays all registered commands that are related to the given <topic>
pwndbg>
```

If you debug this with `set exception-debugger on` it turns out that the default arch set here is i386 because this was the value of `pwnlib.context.context.arch` before the `/bin/ls` file was loaded and our hooks were executed.

disconnect3d: Interestingly, if you remove the default value here, the `arch` passed to `asm` is `None` and then things seems to work fine. I am not sure what Pwntools do under the hood then, but probably just use whatever is set as its  `context.arch` and maybe things just work fine?

pwndbg/commands/asm.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+@pwndbg.commands.ArgparsedCommand(parser, command_name="asm")
+def asm(shellcode, format, arch, avoid, infile):
+    if infile:
+        print(message.warn("Going to read from file: " + infile))
+        with open(infile) as file:
+            shellcode = [file.read()]
+
+    assembly = pwnlib.asm.asm(" ".join(shellcode), arch=arch)
```
disconnect3d:
```suggestion
    bits_for_arch = pwnlib.context.context.architectures.get(arch, {}).get('bits')
    assembly = pwnlib.asm.asm(" ".join(shellcode), arch=arch, bits=bits_for_arch)
```

You need to do something like this since we are updating both Pwnlib's `context.arch` and `context.bits` in one of our hooks. As a result, since bits are set explicitly and you do `asm --arch=i386 nop` when you are debugging an x86-64 (amd64) binary the bits are still set to `'64'` and we would get:
```
Exception occurred: asm: Invalid arch/bits combination: i386/64 (<class 'AttributeError'>)
```

Ignatella: Thx for information :) Please verify

PR #1388, opened 2022-11-21
Add `default-visualize-chunk-number` config and refactor some code related to config

lebr0nli:
According to the feedback in the Discord server,  this PR create a new config called `default-visualize-chunk-number` to set the default number of chunks to visualize with `vis_heap_chunks`

And this PR also did some refactoring which was mentioned in #1315's discussion.

### DIFF
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -592,13 +592,19 @@ pwndbg.gdblib.config.add_param(
     "max display size for heap chunks visualization (0 for display all)",
 )
 
+pwndbg.gdblib.config.add_param(
+    "default-visualize-chunk-number",
+    10,
+    "the number of chunks to visualize (default is 10)",
+)
+
 parser = argparse.ArgumentParser()
 parser.description = "Visualize chunks on a heap, default to the current arena's active heap."
 parser.add_argument(
     "count",
     nargs="?",
     type=lambda n: max(int(n, 0), 1),
-    default=10,
+    default=pwndbg.gdblib.config.default_visualize_chunk_number,
     help="Number of chunks to visualize.",
 )
 parser.add_argument("addr", nargs="?", default=None, help="Address of the first chunk.")
--- a/tests/gdb-tests/tests/heap/test_vis_heap_chunks.py
+++ b/tests/gdb-tests/tests/heap/test_vis_heap_chunks.py
@@ -69,6 +69,16 @@ def test_vis_heap_chunk_command(start_binary):
     expected.append("%#x\t0x0000000000000000" % heap_iter())
     assert result == expected
 
+    ## This time using `default-visualize-chunk-number` to set `count`, to make sure that the config can work
+    gdb.execute("set default-visualize-chunk-number 1")
+    assert pwndbg.gdblib.config.default_visualize_chunk_number == 1
+    result = gdb.execute("vis_heap_chunk", to_string=True).splitlines()
+    assert result == expected
+    gdb.execute(
+        "set default-visualize-chunk-number %d"
+        % pwndbg.gdblib.config.default_visualize_chunk_number.default
+    )
+
     del result
 
     ## Test vis_heap_chunk with count=2
--- a/pwndbg/color/theme.py
+++ b/pwndbg/color/theme.py
@@ -6,9 +6,9 @@ class ColorParameter(pwndbg.lib.config.Parameter):
     pass
 
 
-def add_param(name, default, docstring, color_param=False):
-    return config.add_param(name, default, docstring, scope="theme")
+def add_param(name, default, set_show_doc, color_param=False):
+    return config.add_param(name, default, set_show_doc, scope="theme")
 
 
-def add_color_param(name, default, docstring):
-    return config.add_param_obj(ColorParameter(name, default, docstring, scope="theme"))
+def add_color_param(name, default, set_show_doc):
+    return config.add_param_obj(ColorParameter(name, default, set_show_doc, scope="theme"))
--- a/pwndbg/commands/config.py
+++ b/pwndbg/commands/config.py
@@ -12,11 +12,11 @@ from pwndbg.color import strip
 from pwndbg.color.message import hint
 
 
-def print_row(name, value, default, docstring, ljust_optname, ljust_value, empty_space=6):
+def print_row(name, value, default, set_show_doc, ljust_optname, ljust_value, empty_space=6):
     name = ljust_colored(name, ljust_optname + empty_space)
     defval = extend_value_with_default(value, default)
     defval = ljust_colored(defval, ljust_value + empty_space)
-    result = " ".join((name, defval, docstring))
+    result = " ".join((name, defval, set_show_doc))
     print(result)
     return result
 
@@ -39,7 +39,7 @@ def get_config_parameters(scope, filter_pattern):
         values = [
             v
             for v in values
-            if filter_pattern in v.name.lower() or filter_pattern in v.docstring.lower()
+            if filter_pattern in v.name.lower() or filter_pattern in v.set_show_doc.lower()
         ]
 
     return values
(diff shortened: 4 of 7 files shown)


tests/gdb-tests/tests/heap/test_vis_heap_chunks.py
```diff
@@ -69,6 +69,16 @@ def vis_heap_line(heap_iter_offset=0x10, suffix=""):
     expected.append("%#x\t0x0000000000000000" % heap_iter())
     assert result == expected
 
+    ## This time using `default-visualize-chunk-number` to set `count`, to make sure that the config can work
+    gdb.execute("set default-visualize-chunk-number 1")
+    assert pwndbg.gdblib.config.default_visualize_chunk_number == 1
+    result = gdb.execute("vis_heap_chunk", to_string=True).splitlines()
+    assert result == expected
```
disconnect3d, line 76: Shouldn't expected be smaller here or sth?

pwndbg/commands/heap.py
```diff
@@ -592,13 +592,19 @@ def find_fake_fast(addr, size=None, align=False):
...
+)
+
 parser = argparse.ArgumentParser()
 parser.description = "Visualize chunks on a heap, default to the current arena's active heap."
 parser.add_argument(
     "count",
     nargs="?",
     type=lambda n: max(int(n, 0), 1),
-    default=10,
+    default=pwndbg.gdblib.config.default_visualize_chunk_number,
```
disconnect3d, line 607: The question is: does this really work? I think that maybe this is evaluated once when we add parser argument and is never "read again" for the command execution?

I think the test below may not test for the thing properly (see my comment there)

lebr0nli: > The question is: does this really work?

Yep, it works, it will be evaluated every time we execute `vis_heap_chunks`, the default value seems won't evaluate to int when initializing the parser.

```
pwndbg> pi pwndbg.commands.heap.vis_heap_chunks.parser._optionals._actions[1].default
<pwndbg.lib.config.Parameter object at 0x7f7310cc8df0>
pwndbg> set default-visualize-chunk-number 1
Set the number of chunks to visualize (default is 10) to 1
pwndbg> # if it's only evaluated once, `vis_heap_chunks` should still print out 10 chunks
pwndbg> vis_heap_chunks

0x55a4e3e76000	0x0000000000000000	0x0000000000000291	................
0x55a4e3e76010	0x0000000000000000	0x0000000100000000	................
0x55a4e3e76020	0x0000000000000000	0x0000000000000000	................
0x55a4e3e76030	0x0000000000000000	0x0000000000000000	................
0x55a4e3e76040	0x0000000000000000	0x0000000000000001	................
0x55a4e3e76050	0x0000000000000000	0x0000000000000000	................
0x55a4e3e76060	0x0000000000000000	0x0000000000000000	................
0x55a4e3e76070	0x0000000000000000	0x0000000000000000	................
0x55a4e3e76080	0x0000000000000000	0x0001000000000000	................
0x55a4e3e76090	0x0000000000000000	0x0000000000000000	................
0x55a4e3e760a0	0x0000000000000000	0x0000000000000000	................
0x55a4e3e760b0	0x0000000000000000	0x0000000000000000	................
0x55a4e3e760c0	0x000055a4e3e76480	0x0000000000000000	.d...U..........
0x55a4e3e760d0	0x0000000000000000	0x0000000000000000	................
0x55a4e3e760e0	0x0000000000000000	0x0000000000000000	................
0x55a4e3e760f0	0x0000000000000000	0x0000000000000000	................
0x55a4e3e76100	0x0000000000000000	0x0000000000000000	................
0x55a4e3e76110	0x0000000000000000	0x0000000000000000	................
0x55a4e3e76120	0x0000000000000000	0x0000000000000000	................
0x55a4e3e76130	0x0000000000000000	0x0000000000000000	................
0x55a4e3e76140	0x0000000000000000	0x0000000000000000	................
0x55a4e3e76150	0x0000000000000000	0x0000000000000000	................
0x55a4e3e76160	0x0000000000000000	0x0000000000000000	................
0x55a4e3e76170	0x000055a4e3e762a0	0x0000000000000000	.b...U..........
0x55a4e3e76180	0x0000000000000000	0x0000000000000000	................
0x55a4e3e76190	0x0000000000000000	0x0000000000000000	................
0x55a4e3e761a0	0x0000000000000000	0x0000000000000000	................
0x55a4e3e761b0	0x0000000000000000	0x0000000000000000	................
0x55a4e3e761c0	0x0000000000000000	0x0000000000000000	................
0x55a4e3e761d0	0x0000000000000000	0x0000000000000000	................
0x55a4e3e761e0	0x0000000000000000	0x0000000000000000	................
0x55a4e3e761f0	0x0000000000000000	0x0000000000000000	................
0x55a4e3e76200	0x0000000000000000	0x0000000000000000	................
0x55a4e3e76210	0x0000000000000000	0x0000000000000000	................
0x55a4e3e76220	0x0000000000000000	0x0000000000000000	................
0x55a4e3e76230	0x0000000000000000	0x0000000000000000	................
0x55a4e3e76240	0x0000000000000000	0x0000000000000000	................
0x55a4e3e76250	0x0000000000000000	0x0000000000000000	................
0x55a4e3e76260	0x0000000000000000	0x0000000000000000	................
0x55a4e3e76270	0x0000000000000000	0x0000000000000000	................
0x55a4e3e76280	0x0000000000000000	0x000055a4e3e76500	.........e...U..
0x55a4e3e76290	0x0000000000000000
pwndbg>
```

tests/gdb-tests/tests/heap/test_vis_heap_chunks.py
lebr0nli: Sorry I didn’t catch that.
Do you mean to test a smaller default value here?

disconnect3d: I mean that we are asserting that result is the same as expected that was used for previous test/check but now we changed the default to 1 and it is a bit weird the expected still stays the same?

disconnect3d: I mean I would imagine that both result and expected should be smaller here now and only contain the 1 chunk now

lebr0nli: > I mean that we are asserting that result is the same as expected that was used for previous test/check but now we changed the default to 1 and it is a bit weird the expected still stays the same?

> I mean I would imagine that both result and expected should be smaller here now and only contain the 1 chunk now

The `result` and the `expected` should be the same here because before the line we added the test, we generated the `expected` to verify the output of `vis_heap_chunks 1`(which is the content of the `result`).
So both `result` and `expected` only contains 1 chunk here.

But yeah, I think the test here might be a little bit confusing it looks.
How about we test the config at the beginning like this:
```diff
diff --git a/tests/gdb-tests/tests/heap/test_vis_heap_chunks.py b/tests/gdb-tests/tests/heap/test_vis_heap_chunks.py
index edf602c..60bf6da 100644
--- a/tests/gdb-tests/tests/heap/test_vis_heap_chunks.py
+++ b/tests/gdb-tests/tests/heap/test_vis_heap_chunks.py
@@ -20,6 +20,21 @@ def test_vis_heap_chunk_command(start_binary):
     # Just a sanity check...
     assert (heap_page.start & 0xFFF) == 0

+    # Make sure `set default-visualize-chunk-number <num>` works properly
+    # Note: We only have 3 chunks(not including the top chunk) in the heap currently, so we only generate 3 outputs
+    for i in range(1, 4):
+        gdb.execute(f"set default-visualize-chunk-number {i}")
+        # `vis_heap_chunks` should show `i` chunks
+        assert gdb.execute("vis_heap_chunks", to_string=True) == gdb.execute(
+            f"vis_heap_chunks {i}", to_string=True
+        )
+    # Reset to the default config
+    gdb.execute(
+        f"set default-visualize-chunk-number {pwndbg.gdblib.config.default_visualize_chunk_number.default}"
+    )
+
```

Will it be clearer?

disconnect3d: I think its fine as it is; if u want to extend it, pls do in another PR

PR #1397, opened 2022-11-30
Fix find_fake_fast command test

CptGibbon:
Introduce a more reliable method for mapping one of the fake chunks used in this suite of tests.

Part of the test setup requires the `fake_chunk` size field to be aligned in virtual memory such that its `heap_info` struct resides in unmapped memory. Previously this was left up to ASLR and the test would intermittently fail when the fake chunk's `heap_info` struct was actually mapped. The updated test attempts to `mmap()` 2 pages aligned to `HEAP_MAX_SIZE`, when successful it places `fake_chunk` in the 2nd page then `munmap()`s the 1st page, ensuring the fake chunk's `heap_info` struct is unmapped.

### DIFF
--- a/tests/gdb-tests/tests/binaries/heap_find_fake_fast.c
+++ b/tests/gdb-tests/tests/binaries/heap_find_fake_fast.c
@@ -1,22 +1,27 @@
-/* Test the find_fake_fast command.
- *
- * No need to test the output as find_fake_fast wraps malloc_chunk,
- * which can be tested separately.
- *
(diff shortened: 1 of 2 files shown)


CptGibbon:
Failing tests appear unrelated to this PR:

Docker build (**ubuntu 18.04**) - Failing tests: tests/test_cymbol.py::test_cymbol
Docker build (**archlinux**) - Seems to be an issue in setup.py

PR #1402, opened 2022-11-30
fix: Prevent pwndbg from trying to download ELF files from a remote embedded device.

alufers:
Added a heuristic to check whether we are debugging using a Blackmagic probe or a SEGGER J-link, if yes we don't download any elf files (this caused pwndbg to error out).


disconnect3d:
A good question also is: is the `autofetch` logic needed at all now? I think GDB was autofetching libs by itself last I used it w/o pwndbg, but I may be wrong... would have to double check that.

PR #1407, opened 2022-12-03
fix #1188: incorrect 32-bit syscall display on x64

### LINKED ISSUE #1188: System calls display is wrong for x86 syscalls on x86-64
When we step on `int 0x80` in a 64-bit (x86-64) program, we display wrong syscall number and arguments.

disconnect3d:
Fixes #1188

### DIFF
--- a/pwndbg/arguments.py
+++ b/pwndbg/arguments.py
@@ -58,7 +58,7 @@ def get_syscall_name(instruction):
     # On x86/x64 `syscall` and `int <value>` instructions are in CS_GRP_INT
     # but only `syscall` and `int 0x80` actually execute syscalls on Linux.
     # So here, we return no syscall name for other instructions and we also
-    # handle a cse when 32-bit syscalls are executed on x64
+    # handle a case when 32-bit syscalls are executed on x64
     if syscall_register in ("eax", "rax"):
         mnemonic = instruction.mnemonic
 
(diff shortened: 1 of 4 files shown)


pwndbg/arguments.py
```diff
@@ -53,15 +53,26 @@ def get_syscall_name(instruction):
         return None
 
     syscall_register = pwndbg.lib.abi.ABI.syscall().syscall_register
+    syscall_arch = pwndbg.gdblib.arch.current
 
-    # If we are on x86/x64, return no syscall name for other instructions than syscall and int 0x80
+    # On x86/x64 `syscall` and `int <value>` instructions are in CS_GRP_INT
+    # but only `syscall` and `int 0x80` actually execute syscalls on Linux.
+    # So here, we return no syscall name for other instructions and we also
+    # handle a cse when 32-bit syscalls are executed on x64
```
gsingh93: Does "cse" stand for something or did you mean "case"?

gsingh93 (approved):

pwndbg/arguments.py
disconnect3d: typo: should be 'case' :)

disconnect3d:
```suggestion
    # handle a case when 32-bit syscalls are executed on x64
```

PR #1413, opened 2022-12-05
Add support for performance profiling

### LINKED ISSUE #1405: Make it easier to profile performance
We should have a simple interface to start/stop profiling of pwndbg. For example, `pi pwndbg._profiling_start()` could start it and `pi pwndbg._profiling_stop()` would stop it. This should also work if these commands are included in a gdbinit file so we can measure startup time.

How we get the reports might vary on the implementation, but I don't think that bit will cause any problems.

gsingh93:
To profile startup time, you can set the `PWNDBG_PROFILE` environment variable before loading `pwndbg`. Two pstats file will be created, `pwndbg-load.pstats` profiles the time from sourcing `gdbinit.py` to after `import pwndbg` finishes, and `pwndbg-first-prompt.pstats` profiles from there until the first prompt is displayed.

Regardless of whether the environment variable is set, you can do `pi pwndbg.profiling.profiler.start()` to start profiling.

`print_stats.py` lets you print out these pstats files more easily.

Fixes #1405

### DIFF
--- a/gdbinit.py
+++ b/gdbinit.py
@@ -1,8 +1,17 @@
+import cProfile
 import glob
 import locale
 import sys
+import time
 from os import environ
 from os import path
+
+_profiler = cProfile.Profile()
+
+_start_time = None
+if environ.get("PWNDBG_PROFILE") == "1":
+    _start_time = time.time()
+    _profiler.enable()
 
 # Allow users to use packages from a virtualenv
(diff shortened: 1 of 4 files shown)


disconnect3d:
Shall this document this in contributing md?

gsingh93:
Probably makes more sense in `DEVELOPMENT.md`, but sure, will add some docs.

PR #1415, opened 2022-12-05
Deprecate kernel-vmmap-via-page-tables in favor of kernel-vmmap

### LINKED ISSUE #1403: Don't allow use of `monitor info mem` on qemu-system-aarch64
IIUC, `qemu-system-aarch64` does not support `monitor info mem`. That means if I do `set kernel-vmmap-via-page-tables off`, I get some strange and confusing errors.

We should handle this case gracefully and warn the user they can't do this.

gsingh93:
Users should be able to completely disable `vmmap` when debugging kernels, so that issues with gdb-pt-dump or `monitor info mem` don't become blockers, and for performance reasons.

For `aarch64`, `monitor info mem` isn't supported, so we gracefully hand that case and warn the user.

Fixes #1403.

### DIFF
--- a/pwndbg/gdblib/vmmap.py
+++ b/pwndbg/gdblib/vmmap.py
@@ -10,6 +10,7 @@ import os
 
 import gdb
 
+import pwndbg.color.message as M
 import pwndbg.gdblib.abi
 import pwndbg.gdblib.elf
 import pwndbg.gdblib.events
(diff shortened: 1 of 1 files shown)


pwndbg/gdblib/vmmap.py
```diff
@@ -418,11 +440,27 @@ def kernel_vmmap_via_monitor_info_mem():
...
-        return tuple()
+        monitor_info_mem = gdb.execute("monitor info mem", to_string=True)
+    finally:
+        # Older versions of QEMU/GDB may throw `gdb.error: "monitor" command
+        # not supported by this target`. Newer versions will not throw, but will
+        # return a string starting with 'unknown command:'. We handle both of
+        # these cases in a `finally` block instead of an `except` block.
+        if monitor_info_mem is None or "unknown command" in monitor_info_mem:
+            # TODO: Find out which other architectures don't support this command
+            if pwndbg.gdblib.arch.name == "aarch64":
```
disconnect3d, line 453: arch.name vs arch.current?

disconnect3d: wondering what's the difference; oh i guess current returns some struct

gsingh93: I think they should be the same thing: https://github.com/pwndbg/pwndbg/blob/f571d5ca8495fd490e247ad4baa3eaee372ad80c/pwndbg/lib/arch.py#L8-L10

Issue #1416, opened 2022-12-05
Improve configuration doc strings


gsingh93:
Compared to the doc strings for GDB's built-in options, the doc strings for `pwndbg` need some improvement:
```console
pwndbg> show follow-fork-mode
Debugger response to a program call of fork or vfork is "child".
pwndbg> set follow-fork-mode
Requires an argument. Valid arguments are child, parent.
pwndbg> help show follow-fork-mode
Show debugger response to a program call of fork or vfork.
A fork or vfork creates a new process.  follow-fork-mode can be:
  parent  - the original process is debugged after a fork
  child   - the new process is debugged after a fork
The unfollowed process will continue to run.
By default, the debugger will follow the parent process.
pwndbg> show hexdump-group-use-big-endian
The current value of 'hexdump-group-use-big-endian' is 'off'
pwndbg> set hexdump-group-use-big-endian
Set Use big-endian within each group of bytes. Only applies to raw bytes, not the ASCII part. See also hexdump-highlight-group-lsb. to 'on'
pwndbg> help show hexdump-group-use-big-endian
Show Use big-endian within each group of bytes. Only applies to raw bytes, not the ASCII part. See also hexdump-highlight-group-lsb.
This command is not documented.
```

At the very least, our doc strings should be grammatically correct.

gsingh93:
Another config related issue:
```
pwndbg> apropos elf
auxv -- usage: auxv [-h]
elfheader -- usage: elfheader [-h]
end -- Ends a list of commands or actions.
```

We should display a useful command description here, not the usage string.

disconnect3d:
A lot of this has been made by @k4lizen. Closing this for now.

PR #1424, opened 2022-12-11
Improved configuration and automatic use of heap heuristics

lebr0nli:
Previously, we used `set resolve-heap-via-heuristic on` to apply the heuristic, but it was inconvenient at times. We had some discussions about this before (https://github.com/pwndbg/pwndbg/pull/1029#issuecomment-1273534456).
Now, the heuristics will automatically be used when the debug symbols are not sufficient to resolve the heap.
`resolve-heap-via-heuristic` is set to `auto` by default. To use or not use the heap heuristics, we can use `set resolve-heap-via-heuristic force` or `set resolve-heap-via-heuristic never`, respectively.

Here are some examples in auto mode:
(Note: the yellow hint that show we are going to use heap heuristic will only show once after heap instance changed from `DebugSymsHeap()` to `HeuristicHeap()`)
- When the debug symbols are missing:
截圖 2022-12-11 上午11 25 44

- When debugging statically linked binary:
截圖 2022-12-11 上午11 27 06

Additionally, this PR also attempts to fix some issues mentioned in #1416. 
(I'm not sure if all the issues related to configuration have been fixed, I just fixed the obvious errors :p)

For example:

- kernel-vmmap:
截圖 2022-12-11 下午12 09 34

- hexdump-group-use-big-endian
截圖 2022-12-11 下午12 14 50

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -77,6 +77,15 @@ Feel free to update the list below!
 
 * We have our own `pwndbg.config.Parameter` (which extends `gdb.Parameter`) - all of our parameters can be seen using `config` or `theme` commands. If we want to do something when user changes config/theme - we can do it defining a function and decorating it with `pwndbg.config.Trigger`.
 
+* You can create a new config with `pwndbg.gdblib.config.add_param`.
+
+* When using `pwndbg.gdblib.config.add_param` to add a new config, there are a few things to keep in mind:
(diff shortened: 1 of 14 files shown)


disconnect3d (approved):

disconnect3d:
This is awesome, thanks! For future, it may be nice to split different changes into different PRs (I'll rebase and merge this instead of the usual 'squash and merge' so that we keep commits with specific changes in the main/dev branch history).

Issue #1427, opened 2022-12-11
Don't fail CI jobs if codecov fails


gsingh93:
The codecov step of the GH testing workflow has been failing often recently. We should probably add [continue-on-error](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) to that step to prevent the entire job from failing.

mohitsaxenaknoldus:
Hey, I'll work on this, you can assign it to me.

gsingh93:
Thanks! Feel free to ask questions here or on Discord. Just note, I've never used `continue-on-error` so not 100% sure it's the right solution, but it seems like it from the docs.

mohitsaxenaknoldus:
> Thanks! Feel free to ask questions here or on Discord. Just note, I've never used `continue-on-error` so not 100% sure it's the right solution, but it seems like it from the docs.

PR: https://github.com/pwndbg/pwndbg/pull/1438
@gsingh93 `continue-on-error` will do the job but still mark the complete workflow as a failure so I've put up a better solution.

gsingh93:
Interesting, I had got the idea from here that the full workflow should still succeed even if that one step fails: https://stackoverflow.com/questions/58858429/how-to-run-a-github-actions-step-even-if-the-previous-step-fails-while-still-f#comment123845055_70048004

PR #1438, opened 2022-12-12
Always run GitHub Actions workflow even if tests fail

mohitsaxenaknoldus:
#1427

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -40,13 +40,13 @@ jobs:
         PWNDBG_GITHUB_ACTIONS_TEST_RUN=1 ./tests.sh --cov
 
     - name: Process coverage data
-      if: matrix.os == 'ubuntu-22.04'
+      if: matrix.os == 'ubuntu-22.04' && success() || failure()
       run: |
         coverage combine
         coverage xml
 
     - name: "Upload coverage to Codecov"
-      if: matrix.os == 'ubuntu-22.04'
+      if: matrix.os == 'ubuntu-22.04' && success() || failure()
       uses: codecov/codecov-action@v3
       with:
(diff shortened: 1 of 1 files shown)


disconnect3d:
Can you elaborate more on how and why does this work? Also what's the reason we are not using the `continue-on-error` option suggested in #1427 ?

(It's generally a good practice to provide more information on things in the commit message and in PR message so that if someone looks it up in the future why a certain change was made, it is clear right away what was the reasoning beyond a change).

mohitsaxenaknoldus:
> Can you elaborate more on how and why does this work? Also what's the reason we are not using the `continue-on-error` option suggested in #1427 ?
> 
> (It's generally a good practice to provide more information on things in the commit message and in PR message so that if someone looks it up in the future why a certain change was made, it is clear right away what was the reasoning beyond a change).

As far as I've read, 'continue-on-error' will also do the job. However, it will still mark the complete build as failure.

The second solution, which I've put together, runs the step even if the tests fail and mark the overall build status as passed.

Its just a matter of preference. If you want me to use the first approach then let me know.

.github/workflows/tests.yml
```diff
@@ -40,13 +40,13 @@ jobs:
         PWNDBG_GITHUB_ACTIONS_TEST_RUN=1 ./tests.sh --cov
 
     - name: Process coverage data
-      if: matrix.os == 'ubuntu-22.04'
+      if: matrix.os == 'ubuntu-22.04' && success() || failure()
```
gsingh93: I think we should keep this one the way it was. If something breaks with this step, it's very likely all fault, whereas the next step relies on network connectivity/rate limiting, so if something breaks there it's more likely not our fault and we don't want the job to fail.

Issue #1418, opened 2022-12-05
Add support for inspecting the Linux kernel SLUB allocator


gsingh93:
I have a prototype of this locally, hope to upstream it soon.

gsingh93:
The command isn't 100% complete, but it's still useful enough when testing on a single CPU and if you don't want to see the per-node slabs, so closing this.

PR #1442, opened 2022-12-19
Added command to list slabs

gsingh93:
#1418


requirements.txt
```diff
@@ -7,5 +7,6 @@ pycparser==2.21
 pyelftools==0.29
 Pygments==2.13.0
 ROPGadget==7.1
+tabulate==0.9.0
```
gsingh93: Don't know if it's OK to add this dependency. I'm tired of wasting time formatting output, I figure we should just pick some library and use it throughout the code for consistency.

PR #1447, opened 2022-12-20
Add slab info command

gsingh93:
image

#1418

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -506,6 +506,8 @@ class ArgparsedCommand:
         # We want to run all integer and otherwise-unspecified arguments
         # through fix() so that GDB parses it.
         for action in self.parser._actions:
+            if isinstance(action, argparse._SubParsersAction):
+                action.type = str
             if action.dest == "help":
                 continue
             if action.type in (int, None):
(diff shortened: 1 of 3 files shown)


pwndbg/commands/__init__.py
```diff
@@ -444,6 +444,8 @@ def __init__(self, parser_or_desc, aliases=[], command_name=None):
         # We want to run all integer and otherwise-unspecified arguments
         # through fix() so that GDB parses it.
         for action in self.parser._actions:
+            if isinstance(action, argparse._SubParsersAction):
+                action.type = str
```
gsingh93, line 510: @disconnect3d it seems like this is the first command trying to use a subparser. I needed to add this otherwise the subcommand would be parsed as a `gdb.Value` instead of a string. I'm not very familiar with `ArgparsedCommand` so I'm not sure if this is the right solution, just FYI.

disconnect3d: I am not familiar with subparsers, but it would be nice to make our argparsing and handling of int vs gdb.Value vs multiple strings clear and documented 😅

gsingh93:
While there are still a few TODOs remaining to make this command complete, it's still pretty useful in its current state and since I don't have more time to work on it right now, I'm going to go ahead and merge it so others can start using it.

Issue #1466, opened 2022-12-25
Exception occurred: bugreport: '>' not supported between instances of 'list' and 'int' (<class 'TypeError'>)


ghost:
### Description

At line 165 in `pwndbg/version.py` pwndbg tries to run the following line: `if len(history_commands > 1):`.
Which obviously doesn't work since `history_commands` is a list and `1` is an int, I assume the intended line was `if len(history_commands) > 1:`.

### Steps to reproduce

1. Run pwndbg with any binary;
2. Run any command;
3. Run `bugreport`.

For example: ```
$ gdb /bin/echo
[..snip..]
pwndbg> set exception-verbose on
Set whether to print a full stacktrace for exceptions raised in Pwndbg commands to 'on'.
pwndbg> bugreport
'bugreport': Generate a bug report.
Traceback (most recent call last):
  File "/home/personal/.config/pwndbg/pwndbg/commands/__init__.py", line 151, in __call__
    return self.function(*args, **kwargs)
  File "/home/personal/.config/pwndbg/pwndbg/commands/version.py", line 165, in bugreport
    if len(history_commands > 1):
TypeError: '>' not supported between instances of 'list' and 'int'

If that is an issue, you can report it on https://github.com/pwndbg/pwndbg/issues
(Please don't forget to search if it hasn't been reported before)
To generate the report and open a browser, you may run `bugreport --run-browser`
PS: Pull requests are welcome
pwndbg> 
``` 

### My setup
```
pwndbg> version
Gdb:      12.1
Python:   3.10.9 (main, Dec 19 2022, 17:35:49) [GCC 12.2.0]
Pwndbg:   1.1.1 build: 768cb04
Capstone: 4.0.1024
Unicorn:  2.0.1
```
current pwndbg commit: [768cb042579cddf50db3f21055ce4891f4ae0fb6](https://github.com/pwndbg/pwndbg/commit/768cb042579cddf50db3f21055ce4891f4ae0fb6)

ghost:
I messed up the markdown, sorry!

PR #1467, opened 2022-12-25
Fix issue #1466

ghost:
This commit fixes issue #1466 by Replacing line `if len(history_commands > 1):` to `if len(history_commands) > 1:` in pwndbg/commands/version.py line 165


gsingh93:
Thanks!

PR #1470, opened 2022-12-26
Remove dependency to `python-future` (#1250)

### LINKED ISSUE #1250: Convert `ida_script.py` to Python 3
IDA no longer supports Python 2. We should convert the script to Python 3, and then remove the dependency on `future` in the `requirements.txt`.

It may already work in Python 3, but since it's importing from `__future__`, I'm not sure, and I don't know how to test this.

hamarituc:
`python-future` isn't used in the entire code and abandoned upstream. The module `__future__` is a Python builtin instead.

Closes #1250

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -18,5 +18,4 @@ pycparser
 psutil>=3.1.0
 python-ptrace>=0.8
 six
-future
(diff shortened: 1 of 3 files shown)


disconnect3d:
Thx!

Issue #1289, opened 2022-10-15
Sort list of commands by category instead of alphabetically 


gsingh93:
The `pwndbg` command outputs a lot of commands, and listing them all alphabetically doesn't make finding useful ones easy. It would be better if we group related commands together by their functionality/category, and then sort them alphabetically within each category. We could group the heap commands together, the step/next commands together, the ida/r2/ghidra commands, etc. and for anything we couldn't get to fit into a category we could put in a "misc" category, which ideally would be very small.

gsingh93:
There are a couple different steps needed to implement this. If anyone wants to help with this, feel free to only do one step if you don't want to do all of them:
- Update `ArgparsedCommand` to take an argument called `category` and store it in the command object
	- 	It's probably best to make the argument optional, and then just put up a PR at this point, as this is fairly straightforward while the next steps might be more controversial
-  Start adding categories to various commands. Use your best judgement or ask on Discord if you don't know what category to use for a command. You don't have to do all the commands in one PR. For example, we know we want "heap" to be a category, and it's obvious what the heap commands are, so straightforward decisions like that can go in their own PRs so we can merge them faster.
- Update the `pwndbg` command in `commands/misc.py` to display the commands grouped by categories
	- We don't need to wait until the completion of the previous step to start this. For example, if we've added only the heap commands to a category, we can start working on this before the rest of the commands are categorized.
	- We could have any command that doesn't have the optional category field default to the "misc" category, or we could make it explicit when we put a command in this category. For now it's probably best to leave it as the default, we can always make it explicit later.
	
@disconnect3d let me know if any of the steps above should be implemented differently.

gsingh93:
Here's a first pass at splitting the command list into categories:
```
# HEAP
arena                Print the contents of an arena, default to the current thread's arena.
arenas               List this process's arenas.
bins                 Print the contents of all an arena's bins and a thread's tcache, default to the current thread's arena and tcache.
fastbins             Print the contents of an arena's fastbins, default to the current thread's arena.
find_fake_fast       Find candidate fake fast chunks overlapping the specified address.
heap                 Iteratively print chunks on a heap, default to the current thread's active heap.
heap_config          Shows heap related config. The list can be filtered.
largebins            Print the contents of an arena's largebins, default to the current thread's arena.
malloc_chunk         Print a chunk.
mp                   Print the mp_ struct's contents.
smallbins            Print the contents of an arena's smallbins, default to the current thread's arena.
[...]

PR #1478, opened 2022-12-30
Add commands to categories

gsingh93:
Some things I need feedback on:
- Is the set of categories in the `CommandCategory` enum good? Are any too broad or too narrow? Any major categories missing?
- The order the enum members are defined is the order the categories are printed, does the order look good?
- Do all the commands I've already categorized make sense for that category? I'd recommend you actually pull this PR and run the `pwndbg` command yourself to see how it looks.

I think in future PRs, I'll also add the ability to put a command in multiple categories.

No need to comment on any uncategorized commands for now, unless there's a whole new category class I should be defining. Other commands will be categorized in future PRs. 

#1289


disconnect3d:
Some loud thinking:
1. Shall we have categories for external integrations where we could put r2/r2pipe/rop/ghidra?
2. Memory commands should perhaps contain: `telescope`,`probeleak`, `leakfind`, `memfrob`, `xor`, `xinfo`, `search`, `p2p`. 
3. Register commands should have `fsbase`
4. Maybe we should have "ELF commands" category? This could include `elfheader`, `rop`, `plt`, `got`, `gotplt`?
5.  I think we can remove the category for GDBinit commands, make them pure aliases and if we really want to mention it, we could have some information at the bottom about GDBinit support?
6. To discuss: shall WinDbg commands be put as they are? I think yes.

We probably need to highlight config options more. We should at least print `See also config which allows to change the behavior of certain Pwndbg functionality.` at the end to encourage users to explore it more. Or we could list some of the most useful config options.

gsingh93:
> Shall we have categories for external integrations where we could put r2/r2pipe/rop/ghidra?

Makes sense, I guess we should put IDA there too.

> Memory commands should perhaps contain: telescope,probeleak, leakfind, memfrob, xor, xinfo, search, p2p
> Register commands should have fsbase

Will do.

> Maybe we should have "ELF commands" category? This could include elfheader, rop, plt, got, gotplt?

Yea, I was wondering about this too. I think I'll do it, potentially in another PR.

> I think we can remove the category for GDBinit commands, make them pure aliases and if we really want to mention it, we could have some information at the bottom about GDBinit support?

Sounds good. There's one command that's not an alias (`entry_point`) but I think I'll just delete it. I don't see the use for that command at all, and if anyone wants that they can just call `readelf`. On a related note, I'm thinking of renaming `elfheader` to `elfsections`, maybe in the future `elfheader` could print the header information from `readelf`?

> To discuss: shall WinDbg commands be put as they are? I think yes.

I think so, because there's so many of them and it would be helpful for someone who wants to use those aliases to see them in one section. But I also think that it would be useful to show those as aliases or commands in the other sections, so in later PRs I'll add support for doing that.

> We probably need to highlight config options more. We should at least print See also config which allows to change the behavior of certain Pwndbg functionality. at the end to encourage users to explore it more. Or we could list some of the most useful config options.

This is a good point. Let's wait until I make some more progress on the config deprecation and refactoring, once that's done there may be an easier way to "link" which options are relevant to which commands.

PR #1480, opened 2023-01-01
Fix #1457 - warn if monitor info mem is invalid

### LINKED ISSUE #1457: Warn if page table is invalid instead of asserting
(Discussed in https://github.com/pwndbg/pwndbg/discussions/1456)

We should warn if `monitor info mem` is showing some invalid page table instead of asserting.

disconnect3d:

### DIFF
--- a/pwndbg/gdblib/vmmap.py
+++ b/pwndbg/gdblib/vmmap.py
@@ -450,7 +450,7 @@ def kernel_vmmap_via_page_tables():
     return tuple(retpages)
 
 
-monitor_info_mem_not_warned = False
+monitor_info_mem_not_warned = True
 
 
 def kernel_vmmap_via_monitor_info_mem():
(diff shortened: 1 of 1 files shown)


pwndbg/gdblib/vmmap.py
```diff
@@ -450,6 +450,9 @@ def kernel_vmmap_via_page_tables():
     return tuple(retpages)
 
 
+monitor_info_mem_not_warned = False
```
gsingh93: This should start off as True, right?

pwndbg/gdblib/vmmap.py
```diff
@@ -505,7 +509,18 @@ def kernel_vmmap_via_monitor_info_mem():
         start = int(line[:dash_idx], 16)
         end = int(line[dash_idx + 1 : space_idx], 16)
         size = int(line[space_idx + 1 : rspace_idx], 16)
-        assert end - start == size, "monitor info mem output didn't pass a sanity check"
+        if end - start == size and monitor_info_mem_not_warned:
```
gsingh93: And I think this should be !=

disconnect3d:
@gsingh93 thx! This should be good now.

Example warning output:
```
pwndbg> pi pwndbg.gdblib.vmmap.kernel_vmmap_via_monitor_info_mem()
The vmmap output may be incorrect as `monitor info mem` output assertion/assumption
that end-start==size failed. The values are:
end=0xffff90358009b001; start=0xffff903580099000; size=0x2000; end-start=0x2001
Note that this warning will not show up again in this Pwndbg/GDB session.
```

gsingh93 (approved):
lgtm, just run the lint script again before merging.

PR #1483, opened 2023-01-02
Add commands to Linux category

gsingh93:

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -29,6 +29,7 @@ class CommandCategory(str, Enum):
     STACK = "Stack"
     REGISTER = "Register"
     PROCESS = "Process"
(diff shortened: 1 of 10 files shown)


pwndbg/commands/__init__.py
```diff
@@ -29,6 +29,7 @@ class CommandCategory(str, Enum):
     STACK = "Stack"
     REGISTER = "Register"
     PROCESS = "Process"
+    LINUX = "Linux/libc/ELF"
```
gsingh93, line 32: Not sure if there's a better name for this. I feel like it may be better to group these commands together like this instead of creating too many command categories. @disconnect3d let me know what you think.

disconnect3d: :+1: works for me

PR #1487, opened 2023-01-02
Add pcp command for printing Linux kernel per-cpu page cache

gsingh93:
The Linux kernel uses a per-cpu cache for allocating order-0 pages. Being able to print the contents of these caches can be useful for exploiting some UAF bugs. This is a very rough command to print the caches for each zone. This isn't polished at all or in a state where it should be merged, but I found it useful and won't have time to improve it in the near future, so putting it here for now in case anyone wants to use it or improve it.

Example output:
```
pwndbg> pcp
Zone 0
count:  371
high:  378

pcp.lists[0]:
0xfffffffefff5a2c0
0xfffffffefff586c0
...
```

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -702,6 +702,7 @@ def load_commands() -> None:
     import pwndbg.commands.onegadget
     import pwndbg.commands.p2p
     import pwndbg.commands.patch
+    import pwndbg.commands.pcp
     import pwndbg.commands.peda
     import pwndbg.commands.pie
     import pwndbg.commands.plist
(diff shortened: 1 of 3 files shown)


gsingh93:
Note to self: this needs some minor updates for the latest Linux kernel version, and also some tweaks to handle when `CONFIG_NUMA` is not set.

disconnect3d:
Shall we maybe name the command differently to make it more connected to linux kernel or more understandable?

We already have many fancy shortcuts... so idk, maybe its okay to add more :P

gsingh93:
When people see it in the `pwndbg` output it'll be under the "Kernel" category, I think that should be enough?

gsingh93:
I rebased on the latest dev. I actually tested it again in the android13-5.10 branch of the Android Common Kernel and it still works, but I think it's broken on 5.15 🤷

PR #1488, opened 2023-01-02
Add command to display Android binder driver information

gsingh93:
Adds the `binder` command to display information related to the Binder driver on Android devices:
image

It works well, but the code is a mess and needs to be cleaned up before merging.

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -660,6 +660,7 @@ def load_commands() -> None:
     import pwndbg.commands.asm
     import pwndbg.commands.attachp
     import pwndbg.commands.auxv
+    import pwndbg.commands.binder
     import pwndbg.commands.branch
     import pwndbg.commands.canary
     import pwndbg.commands.checksec
--- a/pwndbg/commands/binder.py
+++ b/pwndbg/commands/binder.py
@@ -1,0 +1,377 @@
+from __future__ import annotations
+
+import argparse
+import logging
+from typing import Any
+from typing import Optional
+
+import gdb
+
+import pwndbg.color as C
+import pwndbg.commands
+import pwndbg.gdblib.symbol
+from pwndbg.commands import CommandCategory
+from pwndbg.gdblib.kernel.macros import container_of
+from pwndbg.gdblib.kernel.macros import for_each_entry
+from pwndbg.gdblib.kernel.rbtree import for_each_rb_entry
+
+log = logging.getLogger(__name__)
+
+addrc = C.green
+fieldnamec = C.blue
+fieldvaluec = C.yellow
+typenamec = C.red
+
+
+def for_each_transaction(addr, field):
+    typename = "struct binder_transaction"
+
+    while addr != 0:
+        transaction = gdb.Value(addr).cast(gdb.lookup_type(typename).pointer())
+        yield transaction
+
+        addr = transaction[field]
+
+
+# TODO: pull this out from the slab command so we can reuse it
+class IndentContextManager:
+    def __init__(self):
+        self.indent = 0
+
+    def __enter__(self):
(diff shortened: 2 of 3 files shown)


Notselwyn:
Any updates here? Seems like a great feature to upstream, as I missed this significantly during kernel research.

gsingh93:
Thanks for the interest @Notselwyn. It's unlikely I'll get time to work on this in the near future as I wrote it for some security research that's been completed, so I don't have any bandwidth to return to it. I uploaded it here in an unfinished state because I thought at least anyone interested could still use it, which would be better than keeping it to myself.

I'd be hesitant to merge it as is, as there are a number of unimplemented things and likely broken edge cases. We'd get bug reports from users that we wouldn't be able to support. But hopefully someone else will be able to pick it up at some point.

disconnect3d:
@gsingh93 
> I'd be hesitant to merge it as is, as there are a number of unimplemented things and likely broken edge cases. We'd get bug reports from users that we wouldn't be able to support. But hopefully someone else will be able to pick it up at some point.

Can we maybe add a print at the end along the lines of "hey, its not finished, help us improving it" and merge it?

fidgetingbits:
> @gsingh93
> 
> > I'd be hesitant to merge it as is, as there are a number of unimplemented things and likely broken edge cases. We'd get bug reports from users that we wouldn't be able to support. But hopefully someone else will be able to pick it up at some point.
> 
> Can we maybe add a print at the end along the lines of "hey, its not finished, help us improving it" and merge it?

I wonder if it's worth having an experimental category for commands like this that indicates to users they should expect it to be broken or unfinished or subject to breaking changes, etc.

Notselwyn:
> I wonder if it's worth having an experimental category for commands like this that indicates to users they should expect it to be broken or unfinished or subject to breaking changes, etc.

I'm sorry for mixing in and I have no authority on this matter, but I would feel like this is the go-to. As a pwndbg user I wouldn't care enough to locally apply the patches from this PR: I'd be wayy happier with a little banner that says the command is experimental with a link to the PR, as I like something that just works without a lot of effort (even if it may break at some point)

Perhaps it would be good to do this approach for #1487 as well :-)

gsingh93:
Thanks for the feedback. I'll add both this command and the `pcp` commmand as experimental commands in the next week or two. Note that I'm not planning on testing them at all before merging them (even if there were bugs, I wouldn't have time to fix them), so just note that the last kernel versions I know they worked on were whatever was used for [this exploit](https://androidoffsec.withgoogle.com/posts/attacking-android-binder-analysis-and-exploitation-of-cve-2023-20938/).

gsingh93:
I rebased and added a warning message saying this command is a WIP. I didn't test if they still work.

brookate:
@gsingh93 BTW, how do you debug android kernel, using kgdb with real devices? or emulator ?

gsingh93:
I primarily use QEMU, see this project: https://github.com/gsingh93/linux-exploit-dev-env

For real devices I often use `printf` debugging over UART. I was also able to get KDB working with a Pixel 7, but not KGDB. Note that modern devices have multiple watchdog timers that will reset the device if a debugger is attached, you need to disable not only the kernel watchdogs, but the hardware watchdogs as well (which are device-specific and undocumented, so I can't help you there).

brookate:
> I primarily use QEMU, see this project: https://github.com/gsingh93/linux-exploit-dev-env
> 
> For real devices I often use `printf` debugging over UART. I was also able to get KDB working with a Pixel 7, but not KGDB. Note that modern devices have multiple watchdog timers that will reset the device if a debugger is attached, you need to disable not only the kernel watchdogs, but the hardware watchdogs as well (which are device-specific and undocumented, so I can't help you there).

Thanks.

PR #1489, opened 2023-01-02
gdbinit.py: improve virtualenv detection and warnings

disconnect3d:

### DIFF
--- a/gdbinit.py
+++ b/gdbinit.py
@@ -19,30 +19,39 @@ if environ.get("PWNDBG_PROFILE") == "1":
 virtual_env = environ.get("VIRTUAL_ENV")
 
 if virtual_env:
-    no_venv_warning = environ.get("PWNDBG_NO_VENV_WARNING", 0)
-    if not no_venv_warning and not sys.executable.startswith(virtual_env):
-        print("[!] Pwndbg Python virtualenv warning [!]")
-        print(
-            "Found Python virtual environment (VIRTUAL_ENV='%s') while GDB is built with a different Python binary (%s)"
-            % (virtual_env, sys.executable)
-        )
-        print("Assuming that you installed Pwndbg dependencies into the virtual environment")
-        print("If this is not true, this may cause import errors or other issues in Pwndbg")
-        print(
-            "If all works for you, you can suppress this warning by setting PWNDBG_NO_VENV_WARNING=1"
-        )
-        print("")
+    venv_warning = int(environ.get("PWNDBG_VENV_WARNING", 1))
+    py_exe_matches = sys.executable.startswith(virtual_env)
+
+    if not py_exe_matches:
+        if venv_warning:
+            print("WARNING: Pwndbg/GDB run in virtualenv with which it may not work correctly ***")
+            print("  Detected Python virtual environment: VIRTUAL_ENV='%s'" % virtual_env)
+            print("  while GDB is built with different Python binary: %s" % sys.executable)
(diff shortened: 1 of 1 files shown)


gdbinit.py
```diff
@@ -19,39 +19,50 @@
 virtual_env = environ.get("VIRTUAL_ENV")
 
 if virtual_env:
-    no_venv_warning = environ.get("PWNDBG_NO_VENV_WARNING", 0)
-    if not no_venv_warning and not sys.executable.startswith(virtual_env):
```
disconnect3d, line 23: Before this PR/commit, if PWNDBG_NO_VENV_WARNING=1 was set, then we never set the virtualenv packages because `sys.path.append(virtualenv_site_packages)` was in that if branch.

Meaning, it was not possible to suppress the warning...

gdbinit.py
```diff
@@ -19,39 +19,55 @@
...
-            "Found Python virtual environment (VIRTUAL_ENV='%s') while GDB is built with a different Python binary (%s)"
-            % (virtual_env, sys.executable)
-        )
-        print("Assuming that you installed Pwndbg dependencies into the virtual environment")
-        print("If this is not true, this may cause import errors or other issues in Pwndbg")
-        print(
-            "If all works for you, you can suppress this warning by setting PWNDBG_NO_VENV_WARNING=1"
-        )
-        print("")
+    venv_warning = int(environ.get("PWNDBG_VENV_WARNING", 1))
```
gsingh93: If this variable is only used in the if statement below, we should move it inside of the if statement.

gdbinit.py
```diff
@@ -19,39 +19,55 @@
...
+    py_exe_matches = sys.executable.startswith(virtual_env)
+
+    if not py_exe_matches:
+        if venv_warning:
+            print("WARNING: Pwndbg/GDB run in virtualenv with which it may not work correctly ***")
```
gsingh93: Just a suggestion: might be better to make this a multiline f-literal string instead of a bunch of prints.

gdbinit.py
```diff
@@ -19,39 +19,55 @@
...
+            print("  Detected Python virtual environment: VIRTUAL_ENV='%s'" % virtual_env)
+            print("  while GDB is built with different Python binary: %s" % sys.executable)
+            print("  Assuming that you installed Pwndbg dependencies into the virtual environment")
```
gsingh93: We may want to make this line more clear and say we'll be adding the venvs site packages to `sys.path`, regardless of whether they hide the warning or not.

gsingh93 (commented):
lgtm, just some minor suggestions.

gdbinit.py
disconnect3d: Tbh I find it annoying to get additional redundant output when I run `gdb`. My `gdb` won't start with Pwndbg if I am in wrong virtualenv anyway as other venvs doesn't have pwndbg deps installed :D.

gsingh93: Oh, my original comment might have been unclear. I'm not suggesting to show always show some warning, I'm saying that in this original message, we don't mention anywhere that we're going to be modifying `sys.path`, we just say the venv might cause problems. I feel like we should mention that.

gsingh93: But I see now that we let the user know this later on (as long as the envar isn't set), so you don't need to do this.

disconnect3d:
I missclicked removal of this branch so it was closed, sorry. Going to merge it when it ends up green.

PR #1491, opened 2023-01-02
Fix reload command

disconnect3d:

### DIFF
--- a/pwndbg/commands/reload.py
+++ b/pwndbg/commands/reload.py
@@ -13,10 +13,8 @@ from pwndbg.commands import CommandCategory
 def rreload(module, _exclude_mods=None) -> None:
     """Recursively reload modules.
     Impl based on https://stackoverflow.com/a/66661311/1508881"""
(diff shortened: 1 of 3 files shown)


pwndbg/commands/reload.py
```diff
@@ -8,23 +10,19 @@
...
-def rreload(module, mdict=None) -> None:
-    """Recursively reload modules."""
-    name = module.__name__
-
-    if mdict is None:
-        mdict = []
+def rreload(module, _exclude_mods=None) -> None:
+    """Recursively reload modules.
+    Impl based on https://stackoverflow.com/a/66661311/1508881"""
+    delete_folders = ["pwndbg"]
```
gsingh93: Did you make this a list because you think we may want to reload other folders at some point?

disconnect3d: Original solution had it and I experimented with reloading gdb too, but it's probably not needed

PR #1493, opened 2023-01-03
Minor improvements (some of #1465)

disconnect3d:
Minor improvements from https://github.com/pwndbg/pwndbg/issues/1465

### DIFF
--- a/pwndbg/commands/nearpc.py
+++ b/pwndbg/commands/nearpc.py
@@ -191,7 +191,7 @@ def nearpc(pc=None, lines=None, to_string=False, emulate=False):
         # Otherwise if it's a branch and it *is* contiguous, just put
         # and empty line.
         elif prev and any(g in prev.groups for g in (CS_GRP_CALL, CS_GRP_JUMP, CS_GRP_RET)):
-            if len("%s" % nearpc_branch_marker_contiguous) > 0:
+            if nearpc_branch_marker_contiguous:
                 result.append("%s" % nearpc_branch_marker_contiguous)
 
         # For syscall instructions, put the name on the side
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -52,7 +52,7 @@ def list_current_commands():
         line = line.strip()
         # Skip non-command entries
         if (
-            len(line) == 0
+            not line
             or line.startswith("Command class:")
             or line.startswith("Unclassified commands")
         ):
--- a/pwndbg/hexdump.py
+++ b/pwndbg/hexdump.py
@@ -87,7 +87,7 @@ def hexdump(
             load_color_scheme()
 
         # If there's nothing to print, just print the offset and address and return
-        if len(data) == 0:
+        if not data:
             yield H.offset("+%04x " % len(data)) + H.address("%#08x  " % (address + len(data)))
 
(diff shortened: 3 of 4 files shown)


pwndbg/disasm/arm.py
```diff
@@ -48,24 +47,37 @@ def condition(self, instruction):
         C = value & (1 << 29)
         V = value & (1 << 28)
 
-        cc = {
```
gsingh93: I'm wondering whether we should keep this one. I think the original is much more readable.

Maybe we can declare the dictionary, and then check the key exists with `in`, otherwise return None

pwndbg/commands/nearpc.py
```diff
@@ -191,7 +191,7 @@ def nearpc(pc=None, lines=None, to_string=False, emulate=False):
         # Otherwise if it's a branch and it *is* contiguous, just put
         # and empty line.
         elif prev and any(g in prev.groups for g in (CS_GRP_CALL, CS_GRP_JUMP, CS_GRP_RET)):
-            if len("%s" % nearpc_branch_marker_contiguous) > 0:
```
gsingh93, line 194: This is so clowny 😂

pwndbg/disasm/arm.py
disconnect3d: Generally if/elif is more efficient in Python :x 

I think the  readability doesn't hurt much but I'm okay with dropping this change if u feel its better to use a dict

gsingh93: Is it? Dictionaries should be an O(1) lookup compared to iterating over each if/else branch which will take longer.

disconnect3d: They are but we are creating the whole dictionary here; i bet this must be more operations since u need to 1) compute all values and 2) hash the keys and 3) call the get function

disconnect3d: But yea... if/elif is not great either:
```
In [1]: def foo(x):
   ...:     if x == 1:
   ...:         return 1
   ...:     elif x == 2:
   ...:         return 2
   ...:     elif x == 3:
   ...:         return 3
   ...:     return None
   ...:

In [2]:

In [2]: import dis

In [3]: dis.dis(foo)
  2           0 LOAD_FAST                0 (x)
              2 LOAD_CONST               1 (1)
              4 COMPARE_OP               2 (==)
              6 POP_JUMP_IF_FALSE       12

  3           8 LOAD_CONST               1 (1)
             10 RETURN_VALUE

  4     >>   12 LOAD_FAST                0 (x)
             14 LOAD_CONST               2 (2)
             16 COMPARE_OP               2 (==)
             18 POP_JUMP_IF_FALSE       24

  5          20 LOAD_CONST               2 (2)
             22 RETURN_VALUE

  6     >>   24 LOAD_FAST                0 (x)
             26 LOAD_CONST               3 (3)
             28 COMPARE_OP               2 (==)
             30 POP_JUMP_IF_FALSE       36

  7          32 LOAD_CONST               3 (3)
             34 RETURN_VALUE

  8     >>   36 LOAD_CONST               0 (None)
             38 RETURN_VALUE

In [4]:
```

disconnect3d: Removed this change, merging

PR #1505, opened 2023-01-05
Run lint CI job and Ubuntu 22.04 and 18.04

gsingh93:
The current linter is using Python 3.8, instead let's use the oldest and newest platforms we support. We're currently missing linter issues because things that are passing on CI on 3.8 are not passing on 3.10, so we need to check both.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -3,7 +3,12 @@ on: [push, pull_request]
 
 jobs:
   lint:
-    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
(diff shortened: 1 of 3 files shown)


gsingh93:
This confirms lint is passing on [Ubuntu 18.04](https://github.com/pwndbg/pwndbg/actions/runs/3844117892/jobs/6546952615) but failing on [Ubuntu 22.04](https://github.com/pwndbg/pwndbg/actions/runs/3844117892/jobs/6546952548). https://github.com/pwndbg/pwndbg/pull/1502 is the PR that broke it.

Issue #1475, opened 2022-12-27
Get typecheck errors down to zero


gsingh93:
With https://github.com/python/typeshed/pull/9411, https://github.com/pwndbg/pwndbg/pull/1469, https://github.com/pwndbg/pwndbg/pull/1471, https://github.com/pwndbg/pwndbg/pull/1472, 
 and https://github.com/pwndbg/pwndbg/pull/1473, most of the code type checks properly and we're down to 13 errors at the moment. This issue is to track the remaining work that needs to be done to get this down to zero. To typecheck the code, run `mypy pwndbg` in the root of the repo.

### [Fixed] ~~`argparse.ArgumentError`~~

```
pwndbg/commands/hexdump.py: note: In function "address_or_module_name":
pwndbg/commands/hexdump.py:37: error: Missing positional argument "message" in call to "ArgumentError"  [call-arg]
                raise argparse.ArgumentError("Could not find pages for module %s" % module_name)
                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
pwndbg/commands/hexdump.py:37: error: Argument 1 to "ArgumentError" has incompatible type "str"; expected
"Optional[Action]"  [arg-type]
                raise argparse.ArgumentError("Could not find pages for module %s" % module_name)
                                             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
`argparse.ArgumentError` should take the message as the second option, not the first. The first argument should be an `Action`. Looking at some examples online, it seems like the correct way to do this is not to have a custom type parsing function, but to implement a class that extends `argparse.Action` and then if the argument is invalid, `self` can be put as the first argument of the error. I think this is probably the best solution.

### `pages.index(page_addr)`

```
pwndbg/gdblib/elf.py: note: In function "map_inner":
pwndbg/gdblib/elf.py:375: error: Argument 1 to "index" of "list" has incompatible type "int"; expected "Page"
[arg-type]
                    page = pages[pages.index(page_addr)]
                                             ^~~~~~~~~
```

While typing this up, I figured out how this code works and what the issue is. The implementation for `Page.__eq__` is:
```
def __eq__(self, other) -> bool:
    return self.vaddr == getattr(other, "vaddr", other)
```
[...]

PR #1515, opened 2023-01-11
Fix some type and pylint errors

gsingh93:
#1475

### DIFF
--- a/pwndbg/ida.py
+++ b/pwndbg/ida.py
@@ -72,12 +72,12 @@ def init_ida_rpc_client() -> None:
     try:
         _ida.here()
         print(message.success("Pwndbg successfully connected to Ida Pro xmlrpc: %s" % addr))
-    except socket.error as e:
+    except TimeoutError:
+        exception = sys.exc_info()
+        _ida = None
+    except OSError as e:
         if e.errno != errno.ECONNREFUSED:
             exception = sys.exc_info()
-        _ida = None
-    except socket.timeout:
-        exception = sys.exc_info()
         _ida = None
     except xmlrpc.client.ProtocolError:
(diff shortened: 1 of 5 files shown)


pwndbg/ida.py
```diff
@@ -72,13 +72,13 @@ def init_ida_rpc_client() -> None:
     try:
         _ida.here()
         print(message.success("Pwndbg successfully connected to Ida Pro xmlrpc: %s" % addr))
-    except socket.error as e:
+    except TimeoutError:
```
gsingh93, line 75: This was an interesting find by `pylint`. `TimeoutError` is a subclass of `OSError`, so it needs to be checked first, otherwise this code will never be reached.

tests/gdb-tests/tests/test_windbg.py
```diff
@@ -23,7 +23,7 @@ def test_windbg_dX_commands(start_binary):
         cmd = cmd_prefix + " nonexistentsymbol"
         assert gdb.execute(cmd, to_string=True) == (
             "usage: XX [-h] address [count]\n"
-            "XX: error: argument address: invalid HexOrAddressExpr value: 'nonexistentsymbol'\n"
```
gsingh93, line 26: @disconnect3d it's interesting that this change resulted in a different exception being raised here. Do you know if the new exception shown is the correct one, or is something broken and I should figure out how to get the old one back?

disconnect3d: That's weird: https://github.com/pwndbg/pwndbg/blob/301d0b9ccdccf5e59480a5f5b9a2e80ed5292b57/pwndbg/commands/__init__.py#L566-L586C23

PR #1518, opened 2023-01-11
Fix memoize type issues and enable mypy in lint script

### LINKED ISSUE #1475: Get typecheck errors down to zero

gsingh93:
The memoization code leads to a number of issues, I'm using a somewhat hacky fix for it described here: https://stackoverflow.com/a/75013308/803801

With this, running `mypy pwndbg` returns no errors. I've added this to the `lint.sh` script, it takes about 2 seconds on my machine.

Fixes #1475

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -6,8 +6,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # Only care about the oldest and newest versions
-        os: [ubuntu-22.04, ubuntu-18.04]
+        os: [ubuntu-22.04, ubuntu-20.04, ubuntu-18.04]
     runs-on: ${{ matrix.os }}
     timeout-minutes: 10
     steps:
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -6,10 +6,14 @@ flake8==5.0.4; python_version < '3.9'
 flake8==6.0.0; python_version >= '3.9'
 flake8-builtins==2.0.0; python_version < '3.7'
 flake8-builtins==2.1.0; python_version >= '3.7'
+mypy==0.991; python_version >= '3.7'
 testresources==2.0.1
 isort==5.10.1; python_version < '3.7'
 isort==5.11.4; python_version >= '3.7'
 pytest==7.0.1; python_version < '3.7'
 pytest==7.2.0; python_version >= '3.7'
 types-gdb==12.1.4.1
+types-psutil==5.9.5
+types-Pygments==2.12.1
+types-tabulate==0.9.0.0
 vermin==1.5.1
--- a/lint.sh
+++ b/lint.sh
@@ -59,3 +59,5 @@ fi
 vermin -q -t=3.6 --violations ./pwndbg/
 
 flake8 --show-source ${LINT_FILES}
+
+mypy pwndbg
--- a/lint.sh
+++ b/lint.sh
@@ -60,4 +60,8 @@ vermin -q -t=3.6 --violations ./pwndbg/
 
 flake8 --show-source ${LINT_FILES}
 
-mypy pwndbg
+if [ -x "$(command -v mypy)" ]; then
+    mypy pwndbg
+else
+    echo "mypy not installed, skipping"
+fi
--- a/pwndbg/lib/memoize.py
+++ b/pwndbg/lib/memoize.py
@@ -6,6 +6,7 @@ new library/objfile are loaded, etc.
 
 import functools
 import sys
+import typing as t
 from collections.abc import Hashable
 from typing import Any
 from typing import Callable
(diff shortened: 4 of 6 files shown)


disconnect3d:
That's crazy but if it works its fine :P

Issue #1417, opened 2022-12-05
Add QEMU tests


gsingh93:
We need to add tests for QEMU if we want to make that experience more stable and make it easier for us to release features without breaking things. At a minimum, we need to have a test attach to and run some basic commands (i.e. `vmmap`, `context`, etc.) with the following targets:

- [x] aarch64 Linux kernel
- [x] x86 Linux kernel
- [x] aarch64 ACK kernel
- [x] x86 ACK kernel
- [ ] aarch64 Linux userspace binary

gsingh93:
An update on this issue: with https://github.com/pwndbg/pwndbg/pull/1533, we'll have some very basic CI for Linux/ACK x86_64/arm64 kernels running in QEMU.

We do have a test for a userspace aarch64 binary, but there are currently a few things broken that I think we should fix first (i.e. #1528 and other issues mentioned on Discord). Once that's done, I think we can consider this issue closed.

There will still be some remaining work to be done, like improving the tests (they're very minimal right now, most of the work here was for the test infra) and improving the user experience (currently all tests are in a single file, there's tons of output that makes it hard to see why a test failed, etc.), but I think we can make these improvements over time now that the basic infra has landed.

gsingh93:
@theguy147 has implemented a bunch more of the QEMU testing infrastructure in #1679 and #1682.

I think the last thing to do for this issue is the ability to test different kernel versions. I currently have builds of different kernel versions working, but they're published as a pre-release: https://github.com/gsingh93/linux-exploit-dev-env/releases/tag/2023.02.11-6f6ef57

The `download_images.sh` function downloads from https://github.com/gsingh93/linux-exploit-dev-env/releases/latest so that we don't have to keep updating it when new builds of kernel images are published. So once the rest of the QEMU testing infra is able to support different kernel versions, then I can turn the pre-release into a real release, and pwndbg should pick it up automatically.

At this point, we can fork the `linux-exploit-dev-env` repo into the `pwndbg` org so future images can be generated from there.

gsingh93:
This is essentially completed. Thanks @theguy147 for working on this! I'll make some new issues for a bit of the followup work.

PR #1533, opened 2023-01-19
Run QEMU tests on CI

gsingh93:
This PR adds a CI job to run the qemu tests. There are some minor tweaks to the scripts, most notably the removal of tmux as it doesn't work well with CI.

#1417

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -32,7 +32,7 @@ jobs:
         echo 'Installed packages:'
         python -m pip freeze
 
-    # We to set `kernel.yama.ptrace_scope=0` for `attachp` command tests
+    # We set `kernel.yama.ptrace_scope=0` for `attachp` command tests
     - name: Run tests
       run: |
         mkdir .cov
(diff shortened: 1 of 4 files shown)


setup.sh
```diff
@@ -20,7 +20,7 @@ osx() {
 
 install_apt() {
     sudo apt-get update || true
-    sudo apt-get install -y git gdb gdbserver python3-dev python3-pip python3-setuptools libglib2.0-dev libc6-dbg
+    sudo apt-get install -y git gdb gdb-multiarch gdbserver python3-dev python3-pip python3-setuptools libglib2.0-dev libc6-dbg qemu-system-x86 qemu-system-arm
```
disconnect3d: The `qemu-system-x86 qemu-system-arm` are not direct Pwndbg dependencies; shall we only install those in setup-dev.sh?

disconnect3d (approved):

setup.sh
gsingh93: You're right, will do. `gdb-multiarch` also isn't technically a pwndbg dependency either, so I'll move it as well.

Issue #1535, opened 2023-01-19
`errno` command might cause another thread to inadvertently continue execution


lebr0nli:
Since we tried to call `__errno_location()` [here](https://github.com/pwndbg/pwndbg/blob/52a479211cf58d1bdc5b42fd50471f565e6a063c/pwndbg/commands/misc.py#L47) when the `errno` variable not works, the other thread might inadvertently continue execution after using this command if we didn't lock the scheduler with `set scheduler-locking on`.

The potential bug can be reproduced with:
```
$ cat test.c
#include <pthread.h>
#include <stdio.h>

void* print_hello(void* args) {
    printf("hello\n");
    pthread_exit(NULL);
}

int main() {
    pthread_t thread;
    pthread_create(&thread, NULL, print_hello, NULL);
    pthread_join(thread, NULL);
    printf("world\n");
    return 0;
}
$ gcc -pthread test.c -o test
$ gdb -q test -ex 'set context-sections' -ex 'break print_hello' -ex 'run' -ex 'thread 1' -ex 'pi int(gdb.parse_and_eval("*((int *(*) (void)) __errno_location) ()"))' -ex 'pi int(gdb.parse_and_eval("*((int *(*) (void)) __errno_location) ()"))'
pwndbg: loaded 136 pwndbg commands and 46 shell commands. Type pwndbg [--shell | --all] [filter] for a list.
pwndbg: created $rebase, $ida GDB functions (can be used with print/break)
Reading symbols from test...
(No debugging symbols found in test)
Sections set to be empty. FYI valid values are: regs, disasm, args, code, stack, backtrace, expressions, ghidra
Breakpoint 1 at 0x11d1
Starting program: /tmp/test
warning: Error disabling address space randomization: Operation not permitted
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7f39104fe700 (LWP 26111)]
[Switching to Thread 0x7f39104fe700 (LWP 26111)]

Thread 2 "test" hit Breakpoint 1, 0x0000560d2201d1d1 in print_hello ()
[Switching to thread 1 (Thread 0x7f39104ff740 (LWP 26107))]
#0  __pthread_clockjoin_ex (threadid=139883063535360, thread_return=0x0, clockid=<optimized out>, abstime=<optimized out>, block=<optimized out>) at pthread_join_common.c:145
[...]

Issue #1530, opened 2023-01-18
The heuristic for `tcache` of the single-threaded target should be enhanced


lebr0nli:
Currently, if we can't find `tcache` with debug symbol or the config, the heuristic will try to analyze the assembly of `__libc_malloc` first.
But this method easily fails or gets a false positive result, especially for the thread-local variable like `tcache`.
In comparison, the accuracy of `mp_`'s heuristic is much better currently.
So if the target is single-threaded, we should try `mp_.sbrk_base + 0x10` first, like what we did when we have the debug symbols:
https://github.com/pwndbg/pwndbg/blob/d4562ea462b8013253c219ed5ddd292386be5c6c/pwndbg/heap/ptmalloc.py#L1139

disconnect3d:
@lebr0nli Is this fixed with #1536?

lebr0nli:
Yes, I think it's fixed now.

lebr0nli:
Yes, I think it's fixed now.

PR #1536, opened 2023-01-19
TLS related enhancements

lebr0nli:
This PR bring some bugs fix and enhancements to some TLS/multi-threaded related features:
Including:
- Use `pthread_self()` instead of the trick with `__errno_location()` when we can't find TLS base with register, since the trick with `__errno_location()` is unreliable.
- Consolidate heap heuristics checks about TLS within the `pwndbg.gdblib.tls` module for better organization
- Bug fix for the `errno` command (#1535)
- Enhance the strategy for the heap heuristic of `tcache` (#1530)
- Add tests for heap heuristics with multi-threaded cases

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -108,4 +108,4 @@ Feel free to update the list below!
 
 * We would like to add proper tests for pwndbg - see tests framework PR if you want to help on that.
 
-* If you want to use `gdb.parse_and_eval("a_function_name()")` or something similar that call a function, please remember this might cause another thread to continue execution without `set scheduler-locking on`. If you didn't expect that, you should use `parse_and_eval_with_scheduler_lock` in `pwndbg.gdblib.scheduler` instead.
+* If you want to use `gdb.parse_and_eval("a_function_name()")` or something similar that call a function, please remember this might cause another thread to continue execution without `set scheduler-locking on`. If you didn't expect that, you should use `parse_and_eval_with_scheduler_lock` from `pwndbg.gdblib.scheduler` instead.
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -108,3 +108,4 @@ Feel free to update the list below!
 
 * We would like to add proper tests for pwndbg - see tests framework PR if you want to help on that.
 
+* If you want to use `gdb.parse_and_eval("a_function_name()")` or something similar that call a function, please remember this might cause another thread to continue execution without `set scheduler-locking on`. If you didn't expect that, you should use `parse_and_eval_with_scheduler_lock` in `pwndbg.gdblib.scheduler` instead.
--- a/pwndbg/commands/misc.py
+++ b/pwndbg/commands/misc.py
@@ -10,6 +10,7 @@ import pwndbg.commands
 import pwndbg.gdblib.regs
 import pwndbg.gdblib.symbol
 from pwndbg.commands import CommandCategory
+from pwndbg.gdblib.tls import lock_scheduler
 
 errno.errorcode[0] = "OK"  # type: ignore # manually add error code 0 for "OK"
 
@@ -44,7 +45,8 @@ def errno_(err) -> None:
                 if errno_loc_gotplt is None or pwndbg.gdblib.vmmap.find(
                     pwndbg.gdblib.memory.pvoid(errno_loc_gotplt)
                 ):
-                    err = int(gdb.parse_and_eval("*((int *(*) (void)) __errno_location) ()"))
+                    with lock_scheduler():
+                        err = int(gdb.parse_and_eval("*((int *(*) (void)) __errno_location) ()"))
                 else:
                     print(
                         "Could not determine error code automatically: the __errno_location@got.plt has no valid address yet (perhaps libc.so hasn't been loaded yet?)"
--- a/pwndbg/commands/misc.py
+++ b/pwndbg/commands/misc.py
@@ -10,7 +10,7 @@ import pwndbg.commands
 import pwndbg.gdblib.regs
 import pwndbg.gdblib.symbol
 from pwndbg.commands import CommandCategory
-from pwndbg.gdblib.tls import lock_scheduler
+from pwndbg.gdblib.scheduler import parse_and_eval_with_scheduler_lock
 
 errno.errorcode[0] = "OK"  # type: ignore # manually add error code 0 for "OK"
 
@@ -45,8 +45,11 @@ def errno_(err) -> None:
                 if errno_loc_gotplt is None or pwndbg.gdblib.vmmap.find(
                     pwndbg.gdblib.memory.pvoid(errno_loc_gotplt)
                 ):
-                    with lock_scheduler():
-                        err = int(gdb.parse_and_eval("*((int *(*) (void)) __errno_location) ()"))
+                    err = int(
+                        parse_and_eval_with_scheduler_lock(
+                            "*((int *(*) (void)) __errno_location) ()"
+                        )
+                    )
                 else:
                     print(
                         "Could not determine error code automatically: the __errno_location@got.plt has no valid address yet (perhaps libc.so hasn't been loaded yet?)"
--- a/pwndbg/gdblib/scheduler.py
+++ b/pwndbg/gdblib/scheduler.py
@@ -5,6 +5,17 @@ import gdb
 
 @contextmanager
 def lock_scheduler():
+    """
+    This context manager can be used to run GDB commands with threads scheduling
+    being locked which means that other threads will be stopped during execution.
+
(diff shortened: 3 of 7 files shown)


gsingh93:
> Bug fix for the errno command

This is a neat bug, thanks for catching and fixing it. To prevent this from happening in the future, I'm wondering if we should always be calling `set scheduler-locking on` in `ArgparsedCommand` before a command is executed and `set scheduler-locking off` when the command is finished. Do you think doing this would cause any problems? If so, the way you've done it is fine, we'll just have to be more mindful in future reviews to catch things like this.

pwndbg/commands/misc.py
```diff
@@ -44,7 +45,8 @@ def errno_(err) -> None:
                 if errno_loc_gotplt is None or pwndbg.gdblib.vmmap.find(
                     pwndbg.gdblib.memory.pvoid(errno_loc_gotplt)
                 ):
-                    err = int(gdb.parse_and_eval("*((int *(*) (void)) __errno_location) ()"))
+                    with lock_scheduler():
+                        err = int(gdb.parse_and_eval("*((int *(*) (void)) __errno_location) ()"))
```
disconnect3d: Do I understand this correctly that if you call a function like this on behalf of the target in GDB, then if you have other threads those other threads are also resumed for the time this function is executed?

Wow, this is really surprising to the degree I could say that maybe this is even a GDB bug, but on the other hand it makes sense to allow for such behavior, especially that the called function may need to interact with other threads itself. So if it wouldn't be a default, this could cause annoying deadlocks.

I think we need to document this behavior/pattern to make sure any future readers of this code understands exactly why we do lock the scheduler.

Alternatively, we can move the `gdb.parse_and_eval` to a `pwndbg.gdblib.*` function, that would be used here instead, and which will use the `lock_scheduler` and we could name and document that function in a descriptive way. We could also put that info into DEVELOPERS.md, as this is really important.

pwndbg/gdblib/tls.py
```diff
@@ -14,74 +15,53 @@
 import pwndbg.gdblib.vmmap
 
 
+@contextmanager
+def lock_scheduler():
+    already_lock = gdb.parameter("scheduler-locking") == "on"
+    old_config = gdb.parameter("scheduler-locking")
```
disconnect3d: Can we fetch the parameter value once?

pwndbg/commands/misc.py
disconnect3d: What @gsingh93 proposed, where we call `set scheduler-locking on` in `ArgparsedCommand` before executing a command also makes sense. Though, it would be great if we can optimize it and only call it if needed.

(We eventually want the commands to be working faster if possible, since they may be used for scripting GDB)

pwndbg/heap/ptmalloc.py
```diff
@@ -1608,218 +1597,189 @@ def thread_cache(self):
...
                     self._thread_cache = self.tcache_perthread_struct(thread_cache_struct_addr)
                     return self._thread_cache
 
-        if self.has_tcache():
-            # Each thread has a tcache struct, and the address of the tcache struct is stored in the TLS.
+        # If target is single-threaded, then the tcache struct is at the first chunk of the heap.
+        # We try to find the address by using mp_.srck_base + 0x10 first since it's more reliable than other methods.
+        if not self.multithreaded:
+            try:
+                thread_cache_struct_addr = self.mp["sbrk_base"] + 0x10
```
CptGibbon: Just a minor point, `thread_cache_struct_addr` will be inaccurate if the `GLIBC_TUNABLES` environment variable is set.
Here's an example of how pwndbg can deal with this: https://github.com/pwndbg/pwndbg/blob/52a479211cf58d1bdc5b42fd50471f565e6a063c/pwndbg/heap/ptmalloc.py#L1231-L1237
In fact the `get_sbrk_heap_region()` method could be useful here 🤷‍♂️

pwndbg/gdblib/tls.py
lebr0nli: Sure, I'll fix it.

pwndbg/heap/ptmalloc.py
lebr0nli: Thanks pointing it out, I'll use `get_sbrk_heap_region()` instead.
Btw [here](https://github.com/pwndbg/pwndbg/blob/d4562ea462b8013253c219ed5ddd292386be5c6c/pwndbg/heap/ptmalloc.py#L1139) also used `sbrk_base + 0x10`, should I replace it as well?

pwndbg/commands/misc.py
lebr0nli: > Do I understand this correctly that if you call a function like this on behalf of the target in GDB, then if you have other threads those other threads are also resumed for the time this function is executed?

Yes, you are correct.

> Alternatively, we can move the gdb.parse_and_eval to a pwndbg.gdblib.* function

This is a good idea.
What function name would you prefer for this function, and which `pwndbg.gdblib.*` module should I add it to?
cc: @gsingh93 
(I'm not good at naming functions and organizing them into different modules xD)

>  We could also put that info into DEVELOPERS.md, as this is really important.

Sure, I'll add that info into `DEVELOPERS.md`.

lebr0nli:
>  To prevent this from happening in the future, I'm wondering if we should always be calling set scheduler-locking on in ArgparsedCommand before a command is executed and set scheduler-locking off when the command is finished. Do you think doing this would cause any problems?

I think this is a good idea, but for the performance, maybe we could only call it only when we need it.
So I'll prefer this solution(https://github.com/pwndbg/pwndbg/pull/1536#discussion_r1083233110) more.

pwndbg/commands/misc.py
gsingh93: > What function name would you prefer for this function, and which pwndbg.gdblib.* module should I add it to?

I don't think we have an existing module that's a good place for this. And I'm honestly not sure what the module should be called, sorry 😛 

Maybe `parse_and_eval_with_scheduler_lock` for the function name? The verbosity is intentional, it should be clear and stick out when we need to use it. But open to other suggestions.

lebr0nli: > Maybe `parse_and_eval_with_scheduler_lock` for the function name? The verbosity is intentional, it should be clear and stick out when we need to use it. But open to other suggestions.

Ok, thanks!
I put it into `pwndbg.gdblib.sheduler` for now.
Let me know if there’s another good place or good name for the module :p

pwndbg/gdblib/scheduler.py
```diff
@@ -0,0 +1,19 @@
+from contextlib import contextmanager
+
+import gdb
+
+
+@contextmanager
+def lock_scheduler():
+    old_config = gdb.parameter("scheduler-locking")
```
disconnect3d, line 19:
```suggestion
    """
    This context manager can be used to run GDB commands with threads scheduling
    being locked which means that other threads will be stopped during execution.

    This is useful to prevent bugs where e.g.: gdb.parse_and_eval("(int)foo()")
    would execute foo() on the current debugee thread but would also unlock other
    threads for being executed and those other threads may for example hit a
    breakpoint we set previously which would be confusing for the user.

    See also: https://sourceware.org/gdb/onlinedocs/gdb/All_002dStop-Mode.html
    """
    old_config = gdb.parameter("scheduler-locking")
```

pwndbg/heap/ptmalloc.py
CptGibbon: Well spotted, yes that also seems like a good candidate for replacement. It might need some extra modifications since a `Page` object is returned.

DEVELOPING.md
```diff
@@ -108,3 +108,4 @@ Feel free to update the list below!
 
 * We would like to add proper tests for pwndbg - see tests framework PR if you want to help on that.
 
+* If you want to use `gdb.parse_and_eval("a_function_name()")` or something similar that call a function, please remember this might cause another thread to continue execution without `set scheduler-locking on`. If you didn't expect that, you should use `parse_and_eval_with_scheduler_lock` in `pwndbg.gdblib.scheduler` instead.
```
disconnect3d:
```suggestion
* If you want to use `gdb.parse_and_eval("a_function_name()")` or something similar that call a function, please remember this might cause another thread to continue execution without `set scheduler-locking on`. If you didn't expect that, you should use `parse_and_eval_with_scheduler_lock` from `pwndbg.gdblib.scheduler` instead.
```

disconnect3d (approved):

pwndbg/heap/ptmalloc.py
disconnect3d: If there are any more improvements to be made here, please lets do it in another PR

Issue #1422, opened 2022-12-09
Add an option to show opcodes in the disassembly


gsingh93:
You can do this with GEF by setting `gef config context.show_opcodes_size 9`:
![Capture_decran_2022-12-08_11-38-31](https://user-images.githubusercontent.com/1012677/206784475-bb7e80c3-4725-4fe4-8265-b8f8a5a36d08.png)

nobodyisnobody:
would be an usefull options for shellcode developping for example to have the hexadecimal opcodes values printed alongside the disassembled instructions.

disconnect3d:
This was added in #1537: two new config variables were added:
```
set nearpc-num-opcode-bytes 9
set nearpc-opcode-separator-bytes 0
```
will result in:
<img width="764" alt="image" src="https://user-images.githubusercontent.com/10009354/215294513-77092c77-b090-40b9-bc03-499039afe55a.png">

PR #1537, opened 2023-01-23
Support showing opcodes in the disassembly

lebr0nli:
This PR is trying to implement the feature that #1422 mentioned.
By setting `nearpc-num-opcode-bytes`(default is 0) and `nearpc-opcode-seperator-bytes` (default is 1):
- `set nearpc-num-opcode-bytes 9`
截圖 2023-01-24 下午8 25 49

- `set nearpc-num-opcode-bytes 4`

截圖 2023-01-24 下午8 26 34

- `set nearpc-num-opcode-bytes 9` + `set nearpc-opcode-seperator-bytes 0`
 截圖 2023-01-24 下午8 36 51

### DIFF
--- a/pwndbg/gdblib/config.py
+++ b/pwndbg/gdblib/config.py
@@ -21,6 +21,11 @@ import pwndbg.lib.config
 
 config = pwndbg.lib.config.Config()
 
+# GDB < 9 does not support PARAM_ZUINTEGER*, so we implement it by ourselves for consistency
+if not hasattr(gdb, "PARAM_ZUINTEGER"):
+    gdb.PARAM_ZUINTEGER = "PARAM_ZUINTEGER"
+    gdb.PARAM_ZUINTEGER_UNLIMITED = "PARAM_ZUINTEGER_UNLIMITED"
+
 
 # See this for details about the API of `gdb.Parameter`:
 # https://sourceware.org/gdb/onlinedocs/gdb/Parameters-In-Python.html
@@ -32,6 +37,7 @@ class Parameter(gdb.Parameter):
         self.set_doc = "Set " + param.set_show_doc + "."
         self.show_doc = "Show " + param.set_show_doc + "."
         self.__doc__ = param.help_docstring or None
+        self._custom_param_class = None
 
         if param.param_class == gdb.PARAM_ENUM:
             super().__init__(
@@ -40,6 +46,13 @@ class Parameter(gdb.Parameter):
                 param.param_class,
                 param.enum_sequence,
             )
+        elif (
+            param.param_class == "PARAM_ZUINTEGER"
+            or param.param_class == "PARAM_ZUINTEGER_UNLIMITED"
+        ):
+            # GDB < 9 does not support PARAM_ZUINTEGER*, so we implement it by ourselves for consistency
+            self._custom_param_class = param.param_class
+            super().__init__(param.name, gdb.COMMAND_SUPPORT, pwndbg.lib.config.PARAM_CLASSES[int])
         else:
             super().__init__(param.name, gdb.COMMAND_SUPPORT, param.param_class)
         self.param = param
@@ -66,6 +79,15 @@ class Parameter(gdb.Parameter):
             # Anyway, to avoid some unexpected behaviors, we'll still set `self.param.value` to 0 here.
             self.param.value = 0
         else:
+            if self._custom_param_class:
+                if (self._custom_param_class == "PARAM_ZUINTEGER" and self.value < 0) or (  # type: ignore
+                    self._custom_param_class == "PARAM_ZUINTEGER_UNLIMITED" and self.value < -1  # type: ignore
+                ):
+                    err = "integer %d out of range" % self.value  # type: ignore
+                    # Restore the old value
+                    self.value = self.param.value
+                    # GDB < 9 is too buggy, it won't handle `gdb.GdbError`..., so we return a string here
+                    return err
             self.param.value = self.value
 
         for trigger in config.triggers[self.param.name]:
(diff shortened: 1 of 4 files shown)


pwndbg/gdblib/nearpc.py
```diff
@@ -162,7 +165,19 @@ def nearpc(pc=None, lines=None, emulate=False, repeat=False) -> List[str]:
...
+            opcodes = instr.bytes.hex()
+            opcodes = opcodes[: show_opcodes_size * 2]
+            align = show_opcodes_size * 2 + 10
+            if len(instr.bytes) > show_opcodes_size:
+                opcodes += pwndbg.color.gray("...")
+                align += 9  # len(pwndbg.color.gray(""))
+            opcodes = opcodes.ljust(align, " ")
+            if pwndbg.gdblib.config.highlight_pc and instr.address == pwndbg.gdblib.regs.pc:
+                opcodes = C.highlight(opcodes)
+            line = " ".join((prefix, address_str, opcodes, symbol, asm))
```
gsingh93: nit: Instead of duplicating this line here and in the `else` clause, we can probably add a `opcodes = ""` above the if statement and then move this right after the if statement, so both code paths use the same code to construct the output.

pwndbg/gdblib/nearpc.py
```diff
@@ -162,7 +165,19 @@ def nearpc(pc=None, lines=None, emulate=False, repeat=False) -> List[str]:
...
             first_pc = False
 
-        line = " ".join((prefix, address_str, symbol, asm))
+        if show_opcodes_size > 0:
+            opcodes = instr.bytes.hex()
+            opcodes = opcodes[: show_opcodes_size * 2]
+            align = show_opcodes_size * 2 + 10
+            if len(instr.bytes) > show_opcodes_size:
+                opcodes += pwndbg.color.gray("...")
+                align += 9  # len(pwndbg.color.gray(""))
```
gsingh93: Can you add some comments explaining how the alignment is calculated? So if we show 9 opcode bytes (18 characters) we'll pad the end of it by 10 spaces, but if we add a '...' we pad the end by 19 spaces?

pwndbg/gdblib/nearpc.py
```diff
@@ -54,7 +54,10 @@ def ljust_padding(lst):
     "nearpc-lines", 10, "number of additional lines to print for the nearpc command"
 )
 show_args = pwndbg.gdblib.config.add_param(
-    "nearpc-show-args", True, "show call arguments below instruction"
+    "nearpc-show-args", True, "whether to show call arguments below instruction"
+)
+show_opcodes_size = pwndbg.gdblib.config.add_param(
+    "nearpc-opcodes-size", 0, "number of bytes of opcodes to print for each instruction"
```
gsingh93: nit: "number of opcode bytes to print for each instruction"

gsingh93: And maybe the option name should be `nearpc-opcode-bytes`? Not sure.

gsingh93 (commented):
This is great, thanks. Could you update the `nearpc` test to also test when `nearpc-opcode-size` is non-zero?

This doesn't need to be done in this PR, but I think having an option for adding spaces between each opcode byte would be nice.

pwndbg/gdblib/nearpc.py
disconnect3d: or: `nearpc-show-opcode-bytes`?

lebr0nli:
> Could you update the nearpc test to also test when nearpc-opcode-size is non-zero?

Sure.

> I think having an option for adding spaces between each opcode byte would be nice

Yeah, after some testing, seems like the readability with a space is better.
I make it default to one space for now and use `nearpc-opcode-seperator-bytes` for controlling it, but open to other suggestions.

pwndbg/gdblib/nearpc.py
lebr0nli: Ok, I'll use `nearpc-show-opcode-bytes`

pwndbg/gdblib/nearpc.py
lebr0nli: > Can you add some comments explaining how the alignment is calculated?

Sure.

> So if we show 9 opcode bytes (18 characters) we'll pad the end of it by 10 spaces, but if we add a '...' we pad the end by 19 spaces?

This 9 byte is because the gray color uses 9 invisible characters, so we need to extend the alignment for it to make it look the same.

pwndbg/gdblib/nearpc.py
lebr0nli: Sure, I'll remove this `else` clause.

pwndbg/gdblib/nearpc.py
gsingh93: `nearpc-show-opcode-bytes` makes it sound like a boolean. `nearpc-num-opcode-bytes` is another alternative.

lebr0nli: > `nearpc-show-opcode-bytes` makes it sound like a boolean. `nearpc-num-opcode-bytes` is another alternative.

Ok, I used `nearpc-num-opcode-bytes` now.

pwndbg/gdblib/config.py
```diff
@@ -66,6 +79,15 @@ def get_set_string(self) -> str:
             # Anyway, to avoid some unexpected behaviors, we'll still set `self.param.value` to 0 here.
             self.param.value = 0
         else:
+            if self._custom_param_class:
+                if (self._custom_param_class == "PARAM_ZUINTEGER" and self.value < 0) or (  # type: ignore
+                    self._custom_param_class == "PARAM_ZUINTEGER_UNLIMITED" and self.value < -1  # type: ignore
```
disconnect3d: Since this is making more and more complex and less and less clear, shall we maybe do something along the lines of:

```py
def __get_set_string_gdb_le_9(self): ...
def __get_set_string_gdb_gte_9(self): ...

if gdb_version >= 9:
    get_set_string = __get_set_string_gdb_gte_9
else:
    get_set_string = __get_set_string_gdb_le_9
```
@gsingh93 WDYT?

gsingh93: +1, I think that'll be less complicated and cleaner.

disconnect3d (approved):
@gsingh93 I think we can merge this now?

gsingh93:
Yup, lgtm.

Issue #1544, opened 2023-01-28
procinfo: incorrect handling of abstract domain sockets


disconnect3d:
The `procinfo` command reads some of the procfs files and then it may `.decode()` them to utf-8:
https://github.com/pwndbg/pwndbg/blob/9522a94baabb4d9a4fda2928e1f1754b477a6fb6/pwndbg/gdblib/net.py#L5-L24

However, the /proc/net/unix file may contain bytes that can't be decoded to utf-8. I believe those are related to abstract UDS (Unix Domain Sockets): https://unix.stackexchange.com/questions/206386/what-does-the-symbol-denote-in-the-beginning-of-a-unix-domain-socket-path-in-l

```
pwndbg> pi pwndbg.gdblib.file.get("/proc/net/unix").decode()
Python Exception <class 'UnicodeDecodeError'> 'utf-8' codec can't decode byte 0x9e in position 7126: invalid start byte:
Error while executing Python code.

pwndbg> pi
>>> open("/proc/net/unix", "rb").read().decode()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x9e in position 7126: invalid start byte
>>> open("/proc/net/unix", "rb").read()[:7250][-300:]
b'a/nmbd/unexpected\n0000000000000000: 00000002 00000000 00010000 0001 01 25365 /var/run/libvirt/libvirt-admin-sock\n0000000000000000: 00000002 00000000 00000000 0002 01 23302 @@@@\x9e\x05@@\x01=\r@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@\n0000000000000000: 0'
```

disconnect3d:
Also this bug may not always be reproducible directly via `procinfo`; I am not completely sure what conditions need to happen for it, maybe we need a socket in our app to be opened first?
```
pwndbg> procinfo
exe        '/bin/ls'
pid        27150
tid        27150
ppid       27138
uid        [1000, 1000, 1000, 1000]
gid        [1000, 1000, 1000, 1000]
groups     [4, 24, 27, 30, 33, 46, 108, 113, 114, 120, 124, 1000]
fd[0]      /dev/pts/3
fd[1]      /dev/pts/3
[...]

PR #1545, opened 2023-01-28
Fix procinfo command (#1544) on QEMU targets and with abstract UDS

disconnect3d:
Actually, this also fixes some QEMU procinfo issues. Hope this is good enough.

This is how it looks now on qemu targets:

image

### DIFF
--- a/pwndbg/gdblib/file.py
+++ b/pwndbg/gdblib/file.py
@@ -107,6 +107,8 @@ def readlink(path):
 
     if is_qemu:
         if not os.path.exists(path):
+            # The or "" is needed since .root() may return None
+            # Then we just use the path (it can also be absolute too)
             path = os.path.join(pwndbg.gdblib.qemu.root() or "", path)
 
     if is_qemu or not pwndbg.gdblib.remote.is_remote():
--- a/pwndbg/gdblib/file.py
+++ b/pwndbg/gdblib/file.py
@@ -107,7 +107,7 @@ def readlink(path):
 
     if is_qemu:
         if not os.path.exists(path):
-            path = os.path.join(pwndbg.gdblib.qemu.root(), path)
+            path = os.path.join(pwndbg.gdblib.qemu.root() or "", path)
 
     if is_qemu or not pwndbg.gdblib.remote.is_remote():
         try:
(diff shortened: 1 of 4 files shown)


pwndbg/gdblib/file.py
```diff
@@ -107,7 +107,7 @@ def readlink(path):
 
     if is_qemu:
         if not os.path.exists(path):
-            path = os.path.join(pwndbg.gdblib.qemu.root(), path)
+            path = os.path.join(pwndbg.gdblib.qemu.root() or "", path)
```
gsingh93, line 112: It's not obvious to me why we need the `or ""`, it may be worth adding a comment.

gsingh93 (approved):
LGTM, feel free to merge.

pwndbg/lib/net.py
```diff
@@ -146,7 +146,12 @@ def unix(data: str):
         return []
 
     result = []
-    for line in data.splitlines()[1:]:
+    # Note: it is super important to split by "\n" instead of .splitlines() here
+    # because there may be a line like this:
+    # "0000000000000000: 00000002 00000000 00000000 0002 01 23302 @@@@\x9e\x05@@\x01=\r@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@\n"
+    # and splitlines will also split by \r which we do not want here
+    # We also finish at -1 index since with .split() the empty last line is kept in the result
+    for line in data.split("\n")[1:-1]:
```
gsingh93, line 154: Would be good to have a test for this.

disconnect3d: I'll add another GH issue for that. We should also extend the lib/net.py code to handle more cases too.

pwndbg/gdblib/file.py
disconnect3d, line 112:
```suggestion
            # The or "" is needed since .root() may return None
            # Then we just use the path (it can also be absolute too)
            path = os.path.join(pwndbg.gdblib.qemu.root() or "", path)
```

pwndbg/lib/net.py
disconnect3d: https://github.com/pwndbg/pwndbg/issues/1551

PR #1546, opened 2023-01-28
Fix #1534: disable emulation if mmap(1G,RWX) fails

### LINKED ISSUE #1534: 'Could not allocate dynamic translator buffer' Error when execute "start" command
### Description
'Could not allocate dynamic translator buffer' Error when execute \"start\" command

> pwndbg> start
> Temporary breakpoint 1 at 0x400400
> Temporary breakpoint 1, 0x0000000000400400 in _start ()
> Could not allocate dynamic translator buffer

### Steps to reproduce

> gdb blah
> start

### My setup
OS - Oracle Cloud Instance
> DISTRIB_ID=Ubuntu
> DISTRIB_RELEASE=22.04
> DISTRIB_CODENAME=jammy
> DISTRIB_DESCRIPTION="Ubuntu 22.04.1 LTS"
> PRETTY_NAME="Ubuntu 22.04.1 LTS"
> NAME="Ubuntu"
> VERSION_ID="22.04"
> VERSION="22.04.1 LTS (Jammy Jellyfish)"
> VERSION_CODENAME=jammy
> ID=ubuntu
> ID_LIKE=debian

In addition, same issues on Ubuntu 18.04

GDB
**GNU gdb (Ubuntu 8.1.1-0ubuntu1) 8.1.1**

Python
**Python 3.10.6**

Commit
**commit 52a479211cf58d1bdc5b42fd50471f565e6a063c (HEAD -> dev, origin/dev, origin/HEAD)**

disconnect3d:
TL:DR: Unicorn Engine aborts if mmap(1G, RWX) fails, so we are doing a best effort check if we can do such allocation before using it for the first time and if we can't, we disable it.

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -616,12 +616,18 @@ def context_disasm(target=sys.stdout, with_banner=True, width=None):
     if cs is not None and cs.syntax != syntax:
         pwndbg.lib.memoize.reset()
 
-    arch = pwndbg.gdblib.arch.current
-    emulate = bool(pwndbg.gdblib.config.emulate)
+    result = pwndbg.gdblib.nearpc.nearpc(
+        lines=code_lines // 2, emulate=bool(pwndbg.gdblib.config.emulate)
+    )
 
-    info = " / %s / set emulate %s" % (arch, "on" if emulate else "off")
+    # Note: we must fetch emulate value again and after disasm since
+    # we check if we can actually use emulation in `can_run_first_emulate`
+    # and this call may disable it
+    info = " / %s / set emulate %s" % (
+        pwndbg.gdblib.arch.current,
+        "on" if bool(pwndbg.gdblib.config.emulate) else "off",
+    )
     banner = [pwndbg.ui.banner("disasm", target=target, width=width, extra=info)]
-    result = pwndbg.gdblib.nearpc.nearpc(lines=code_lines // 2, emulate=emulate)
 
     # If we didn't disassemble backward, try to make sure
     # that the amount of screen space taken is roughly constant.
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -620,7 +620,7 @@ def context_disasm(target=sys.stdout, with_banner=True, width=None):
         lines=code_lines // 2, emulate=bool(pwndbg.gdblib.config.emulate)
     )
 
-    # Note: we must fetch emulate value again and after disasm since
+    # Note: we must fetch emulate value again after disasm since
     # we check if we can actually use emulation in `can_run_first_emulate`
     # and this call may disable it
     info = " / %s / set emulate %s" % (
--- a/pwndbg/disasm/__init__.py
+++ b/pwndbg/disasm/__init__.py
@@ -18,6 +18,7 @@ import pwndbg.gdblib.memory
 import pwndbg.gdblib.symbol
 import pwndbg.ida
 import pwndbg.lib.memoize
+from pwndbg.color import message
 
 try:
     import pwndbg.emu.emulator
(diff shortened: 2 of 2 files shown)


pwndbg/disasm/__init__.py
```diff
@@ -208,6 +209,45 @@ def get(address, instructions=1):
...
+
+    try:
+        from mmap import MAP_ANON
+        from mmap import MAP_PRIVATE
+        from mmap import PROT_EXEC
+        from mmap import PROT_READ
+        from mmap import PROT_WRITE
+        from mmap import mmap
+
+        mmap(-1, 1024 * 1024 * 1024, MAP_PRIVATE | MAP_ANON, PROT_WRITE | PROT_READ | PROT_EXEC)
```
wtdcode: This should be unmapped in any case?

pwndbg/commands/context.py
```diff
@@ -616,12 +616,18 @@ def context_disasm(target=sys.stdout, with_banner=True, width=None):
...
         pwndbg.lib.memoize.reset()
 
-    arch = pwndbg.gdblib.arch.current
-    emulate = bool(pwndbg.gdblib.config.emulate)
+    result = pwndbg.gdblib.nearpc.nearpc(
+        lines=code_lines // 2, emulate=bool(pwndbg.gdblib.config.emulate)
+    )
 
-    info = " / %s / set emulate %s" % (arch, "on" if emulate else "off")
+    # Note: we must fetch emulate value again and after disasm since
```
```suggestion
    # Note: we must fetch emulate value again after disasm since
```

pwndbg/disasm/__init__.py
```diff
@@ -208,6 +209,43 @@ def get(address, instructions=1):
...
+        return True
+    first_time_emulate = False
+
+    try:
+        from mmap import mmap, MAP_ANON, MAP_PRIVATE, PROT_EXEC, PROT_READ, PROT_WRITE  # isort:skip
+
+        mm = mmap(  # novm
+            -1, 1024 * 1024 * 1024, MAP_PRIVATE | MAP_ANON, PROT_WRITE | PROT_READ | PROT_EXEC
+        )
+        mm.close()
```
wtdcode, line 229: @disconnect3d 

I believe this should go to the `final` clause.

pwndbg/disasm/__init__.py
```diff
@@ -246,7 +284,7 @@ def near(address, instructions=1, emulate=False, show_prev_insns=True):
     emu = None
 
     # If we hit the current instruction, we can do emulation going forward from there.
-    if address == pc and emulate:
+    if address == pc and emulate and (not first_time_emulate or can_run_first_emulate()):
```
wtdcode, line 287: I think it shall also check against Unicorn version below or equal 2.0.1.

pwndbg/disasm/__init__.py
disconnect3d: Why? The allocation will only happen if there is no exception

wtdcode: Oh you are right :)

pwndbg/disasm/__init__.py
disconnect3d: We pin the dependencies versions so we should be using 2.0.0rc7 or 2.0.1.post1: https://github.com/pwndbg/pwndbg/blob/dev/requirements.txt#L11-L12

wtdcode: The problem is that users may install another version, including myself actually during the time `pwndbg` hasn't migrated to Unicorn2 :p

wtdcode: Or maybe I ping this PR once the next release is ready?

disconnect3d: The user can always break their installation if they overwrite the version of dependencies used by Pwndbg :/, not sure if we should really check for it here.

> The problem is that users may install another version, including myself actually during the time pwndbg hasn't migrated to Unicorn2 :p

What do u mean? We use Unicorn 2 (?)

wtdcode: I have been `pwndbg` user for a long time and I develop & debug Unicorn2 with `pwndbg` too. Previously, `pwndbg` didn't upgrade to Unicorn2.0.0 rc version until the official release so it was quite often to see pip warnings.

Besides, iirc, `pip` doesn't automatically upgrade the dependency if it's already there (say I have Unicorn 1.0.0 but tell `pip` to install `pwndbg`) so I think it's good to check the version for max compatibility.

disconnect3d: Heh, but how do I know which other versions do need this 1GB of memory? Is that only everything before until 2.0.1 or is it only 2.0.0 and 2.0.1?

If you feel strongly about adding this check, can you send a PR that adds it? :)

wtdcode: > If you feel strongly about adding this check, can you send a PR that adds it? :)

Sure. Will ping you then.

PR #1558, opened 2023-01-31
Enhance the checks before accessing the memory for the heap heuristics

### LINKED ISSUE #1542: gdb.MemoryError: Cannot access memory at address While search main_arena
I have try to run `vis` at ARM arch with strippted GLIBC In this [line](https://github.com/pwndbg/pwndbg/blob/b9c9f12efcec8e5e11c8cd8fcedc69c856d1716a/pwndbg/heap/ptmalloc.py#L1414)

```
  File "/home/USER/pwndbg/pwndbg/commands/__init__.py", line 184, in __call__
  File "/home/USER/pwndbg/pwndbg/commands/__init__.py", line 315, in _OnlyWhenRunning
    return function(*a, **kw)
  File "/home/USER/pwndbg/pwndbg/commands/__init__.py", line 409, in _OnlyWithResolvedHeapSyms
    return _try2run_heap_command(function, a, kw)
  File "/home/USER/pwndbg/pwndbg/commands/__init__.py", line 377, in _try2run_heap_command
    raise err
  File "/home/USER/pwndbg/pwndbg/commands/__init__.py", line 358, in _try2run_heap_command
    return function(*a, **kw)
  File "/home/USER/pwndbg/pwndbg/commands/__init__.py", line 340, in _OnlyWhenHeapIsInitialized
    return function(*a, **kw)
  File "/home/USER/pwndbg/pwndbg/commands/heap.py", line 673, in vis_heap_chunks
    heap_region = arena.active_heap
  File "/home/USER/pwndbg/pwndbg/heap/ptmalloc.py", line 591, in active_heap
    self._active_heap = Heap(self.top, arena=self)
  File "/home/USER/pwndbg/pwndbg/heap/ptmalloc.py", line 352, in __init__
    main_arena = allocator.main_arena
  File "/home/USER/pwndbg/pwndbg/heap/ptmalloc.py", line 1414, in main_arena
    tmp_next = int(tmp_arena["next"])
gdb.MemoryError: Cannot access memory at address 0xXXXXXXXX (I remove the original address)

PS: Pull requests are welcome
> /home/USER/pwndbg/pwndbg/heap/ptmalloc.py(1414)main_arena()
-> tmp_next = int(tmp_arena["next"])
```

The address 0xXXXXXXXX  really relate for some driver , not relate for heap
And maybe it will be better to check the address before access

lebr0nli:
We used `pwndbg.gdblib.vmmap.find()` to verify whether some addresses are valid or not in the heap heuristics.
But in some edge cases, it's still possible that an address can be found by vmmap, but still cause an error when we tried to access it.

This can be reproduced by something like this:
```text
pwndbg> ipi

In [1]: pwndbg.gdblib.vmmap.find(0xffffffffff600000)
Out[1]: Page('0xffffffffff600000 0xffffffffff601000 r-xp     1000      0 [vsyscall]')

In [2]: pwndbg.gdblib.memory.byte(0xffffffffff600000)
MemoryError                               Traceback (most recent call last)
<ipython-input-2-410a4ed1fe80> in <module>
----> 1 pwndbg.gdblib.memory.byte(0xffffffffff600000)

/ctf/work/pwndbg/pwndbg/gdblib/memory.py in byte(addr)
    162     Read one byte at the specified address
    163     """
--> 164     return readtype(pwndbg.gdblib.typeinfo.uchar, addr)
    165
    166

/ctf/work/pwndbg/pwndbg/gdblib/memory.py in readtype(gdb_type, addr)
     77         :class:`int`
     78     """
---> 79     return int(gdb.Value(addr).cast(gdb_type.pointer()).dereference())
     80
     81

MemoryError: Cannot access memory at address 0xffffffffff600000
```

This PR aims to prevent it and fixes #1542.

### DIFF
--- a/pwndbg/heap/ptmalloc.py
+++ b/pwndbg/heap/ptmalloc.py
@@ -324,7 +324,7 @@ class Chunk:
             return None
 
         next = Chunk(self.address + self.real_size, arena=self.arena)
-        if pwndbg.gdblib.memory.peek(next.address):
+        if pwndbg.gdblib.memory.is_readable_address(next.address):
             return next
         else:
             return None
@@ -1382,10 +1382,7 @@ class HeuristicHeap(GlibcMemoryAllocator):
                                 self._main_arena_addr = offset + instr.address + 4
 
             # Try to search main_arena in .data of libc if we can't find it via above trick
-            if (
-                not self._main_arena_addr
-                or pwndbg.gdblib.memory.peek(self._main_arena_addr) is None
-            ):
+            if not pwndbg.gdblib.memory.is_readable_address(self._main_arena_addr):
                 start = pwndbg.gdblib.symbol.address("_IO_2_1_stdin_")
                 end = pwndbg.gdblib.symbol.address("_IO_list_all")
                 # If we didn't have these symbols, we try to find them in the possible page
@@ -1417,15 +1414,21 @@ class HeuristicHeap(GlibcMemoryAllocator):
                                 found = True
                                 break
                             tmp_arena = self.malloc_state(tmp_next)
+                            if (
+                                pwndbg.gdblib.vmmap.find(tmp_arena.get_field_address("next"))
+                                is None
+                            ):
+                                # if `&tmp_arena->next` is not valid, the linked list is broken, break this while loop and try `addr+pwndbg.gdblib.arch.ptrsize` again
+                                # Note: We use vmmap to check before accessing the memory because accessing memory for embedded targets might be slow and expensive.
+                                break
                             try:
                                 tmp_next = int(tmp_arena["next"])
                             except Exception:
-                                # if `&tmp_arena->next` is not valid, the linked list is broken, break this while loop and try `addr+pwndbg.gdblib.arch.ptrsize` again
                                 break
                         if found:
                             break
 
-        if self._main_arena_addr and pwndbg.gdblib.memory.peek(self._main_arena_addr):
+        if pwndbg.gdblib.memory.is_readable_address(self._main_arena_addr):
             self._main_arena = Arena(self._main_arena_addr)
             return self._main_arena
 
@@ -1451,7 +1454,7 @@ class HeuristicHeap(GlibcMemoryAllocator):
                 if pwndbg.gdblib.tls.is_thread_local_variable(thread_arena_via_symbol):
                     thread_arena_struct_addr = pwndbg.gdblib.memory.u(thread_arena_via_symbol)
                     # Check &thread_arena is a valid address or not to avoid false positive.
-                    if pwndbg.gdblib.memory.peek(thread_arena_struct_addr):
+                    if pwndbg.gdblib.memory.is_readable_address(thread_arena_struct_addr):
                         return Arena(thread_arena_struct_addr)
 
             if not self._thread_arena_offset and pwndbg.gdblib.symbol.address("__libc_calloc"):
@@ -1581,7 +1584,7 @@ class HeuristicHeap(GlibcMemoryAllocator):
                 tls_base = pwndbg.gdblib.tls.address
                 if tls_base:
                     thread_arena_struct_addr = tls_base + self._thread_arena_offset
-                    if pwndbg.gdblib.memory.peek(thread_arena_struct_addr):
+                    if pwndbg.gdblib.memory.is_readable_address(thread_arena_struct_addr):
                         return Arena(pwndbg.gdblib.memory.pvoid(thread_arena_struct_addr))
 
             raise SymbolUnresolvableError("thread_arena")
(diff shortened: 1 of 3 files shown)


disconnect3d:
@lebr0nli heh, perhaps we should have a version of vmmap that would not return the vsyscall page to be used for cases like here... :)

A better example for this change is when we just can't figure out the vmmap result and return a page of [0, 0xffff...] though the question is --- if that is the case --- do we ever reach those code paths here?

Also, using `memory.peek` instead of `vmmap.get` can be problematic for embedded targets when accessing the memory _is not cheap and is very slow_. However, I think that those targets will likely not be used with libc heap very often, so its probably okay to not think about this until someone actually hits that case.

pwndbg/heap/ptmalloc.py
```diff
@@ -1408,18 +1419,18 @@ def main_arena(self):
...
                             tmp_arena = self.malloc_state(tmp_next)
-                            if (
-                                pwndbg.gdblib.vmmap.find(tmp_arena.get_field_address("next"))
-                                is not None
-                            ):
+                            try:
                                 tmp_next = int(tmp_arena["next"])
-                            else:
+                            except Exception:
                                 # if `&tmp_arena->next` is not valid, the linked list is broken, break this while loop and try `addr+pwndbg.gdblib.arch.ptrsize` again
```
disconnect3d, line 1421: Just to be sure: this comment here is still valid after we switch to except, right?

pwndbg/heap/ptmalloc.py
```diff
@@ -1408,18 +1419,18 @@ def main_arena(self):
...
                                 break
                             tmp_arena = self.malloc_state(tmp_next)
-                            if (
-                                pwndbg.gdblib.vmmap.find(tmp_arena.get_field_address("next"))
-                                is not None
-                            ):
+                            try:
                                 tmp_next = int(tmp_arena["next"])
-                            else:
+                            except Exception:
```
disconnect3d: is this gdb.MemoryError or gdb.error that happens here? Can we be more explicit with the exceptions that we want to catch? 🤐

pwndbg/heap/ptmalloc.py
```diff
@@ -1398,7 +1402,14 @@ def main_arena(self):
                     for addr in range(start, end, pwndbg.gdblib.arch.ptrsize):
                         found = False
                         tmp_arena = self.malloc_state(addr)
-                        tmp_next = int(tmp_arena["next"])
+                        try:
+                            tmp_next = int(tmp_arena["next"])
+                        except Exception:
```
disconnect3d: is this gdb.MemoryError or gdb.error that happens here? Can we be more explicit with the exceptions that we want to catch? 🤐

disconnect3d:
I know that I ask about a lot: but can we maybe add a testcase for this? I bet this can be tested by running some heap command against a vsyscall address page?

lebr0nli:
> A better example for this change is when we just can't figure out the vmmap result and return a page of [0, 0xffff...] though the question is --- if that is the case --- do we ever reach those code paths here?

Yes, for the qemu-user without the patch, I guess we might face a similar problem if with only do the vmmap check.

> Also, using `memory.peek` instead of `vmmap.get` can be problematic for embedded targets when accessing the memory _is not cheap and is very slow_

Ahh, I forgot about the performance issues, thanks for pointing it out.
Hmm, maybe for now, using `vmmap.find()` before `memory.peek()` will be better?
So we can avoid some unnecessary memory access if possible.
I mean something like this:
```python
def is_peekable_address(address):
    return pwndbg.gdblib.vmmap.find(address) is None and peek(address) is not None
```

> I know that I ask about a lot: but can we maybe add a testcase for this?

Sure, I'll the test for it.

pwndbg/heap/ptmalloc.py
lebr0nli: Yes, it should still be valid.

pwndbg/heap/ptmalloc.py
lebr0nli: I think it should be only `gdb.MemoryError`, but I’m not very sure actually.
I was following the code here when writing it:
https://github.com/pwndbg/pwndbg/blob/afa96abfa022052adbf39a430a7f353faac32838/pwndbg/gdblib/memory.py#L110-L114
Is there any reason why we catch `Exception` here?
If it also should be only `gdb.MemoryError`, maybe I should change it as well? (probably change it in another PR?)

disconnect3d:
> Yes, for the qemu-user without the patch, (...)

I really need to get back to this patch and finish it :).

> Hmm, maybe for now, using vmmap.find() before memory.peek() will be better?

This makes sense! Shouldn't it actually be:
```py
def is_readable_address(address):
    return pwndbg.gdblib.vmmap.find(address) and peek(address)
```
(or, if it makes more sense: `is not None` in both)

I mean, so that we will peek only if we find memory page. This would at least be faster if someone defines the vmmap pages properly on embedded targets.

lebr0nli:
> (or, if it makes more sense: `is not None` in both)

Yes, that's right, somehow I left out an `not`, what did I write lol😂

disconnect3d:
@lebr0nli Is that work ready? I haven't tested the tests myself but I believe they work fine since CI builds! :P

lebr0nli:
> Is that work ready?

Yea, I think it's ready to be merged.

Issue #1196, opened 2022-10-02
Heap chain corruption check is incorrect for long free lists


gsingh93:
Debugging this code with pwndbg and running the `bins` command shows the wrong output for the unsorted bin.
```c
int main() {
  char* chunks[20] = {0};
  for (int i = 0; i < 20; i ++) {
    chunks[i] = malloc(0x500);

    // padding
    malloc(0x20);
  }

  for (int i = 0; i < 20; i ++) {
    free(chunks[i]);
  }
}
```

After executing free around 8 or 9 times, the unsorted bin still looks fine:
```
unsortedbin
all: 0x55555555b210 —▸ 0x55555555acd0 —▸ 0x55555555a790 —▸ 0x55555555a250 —▸ 0x555555559d10 ◂— ...
```

But then executing it again shows this:
```
all [corrupted]
FD: 0x55555555b750 —▸ 0x55555555b210 —▸ 0x55555555acd0 —▸ 0x55555555a790 —▸ 0x55555555a250 ◂— ...
BK: 0x555555559290 —▸ 0x5555555597d0 —▸ 0x555555559d10 —▸ 0x55555555a250 —▸ 0x55555555a790 ◂— ...
```

CptGibbon:
This appears to be tied to `heap_chain_limit`, defined below:
https://github.com/pwndbg/pwndbg/blob/3c8ee9bb37d670bf2c161c7a45bb0c3fbadc664e/pwndbg/heap/__init__.py#L22-L23
Which is used in `bin_at`:
https://github.com/pwndbg/pwndbg/blob/3c8ee9bb37d670bf2c161c7a45bb0c3fbadc664e/pwndbg/heap/ptmalloc.py#L570-L578

CptGibbon:
It boils down to the fact that you can't check incomplete fd & bk lists against each other for corruption. One course of action could be to print a warning that pwndbg can't do a corruption check when the bin dereference limit is reached?

gsingh93:
Not an expert in how this works, but I have two ideas:
- We can have different limits for the number of chain elements to display vs. the number of chain elements to check. We can only display 8 elements, but check 32 (for example) without any major hit to performance. If we hit that second limit (which could also be a user config option), we could print a warning like you suggest, and even mention the user can increase this limit if they want
- Can we do the check in a way similar to how it's done in the actual libc code? Instead of attempting to traverse the entire chain, can we just check if `victim->bk->fd = victim`?

[...]

PR #1564, opened 2023-02-03
Improve bin corruption checks

gsingh93:
- The existing bin corruption checks would mark some corrupted bins as empty instead of corrupted. There are now some additional checks, including checking if the `fd` and `bk` pointers are in the libc data section.
- Handle the case where the chain is less than 2 elements
- Check the full list integrity only if we've fully traversed the `fd` and `bk` lists (addresses some of #1196)
- Make the list integrity check more clear
- In the case we can't traverse the entire lists, do a basic sanity check that `chunk->bk->fd = chunk`

### DIFF
--- a/pwndbg/heap/ptmalloc.py
+++ b/pwndbg/heap/ptmalloc.py
@@ -1191,14 +1191,55 @@ class GlibcMemoryAllocator(pwndbg.heap.heap.MemoryAllocator):
         chain_fd = get_chain(front, fd_offset)
         chain_bk = get_chain(back, bk_offset)
 
-        # check if bin[index] points to itself (is empty)
-        if len(chain_fd) == len(chain_bk) == 2 and chain_fd[0] == chain_bk[0]:
-            chain_fd = [0]
-            chain_bk = [0]
-
-        # check if corrupted
-        elif chain_fd[:-1] != chain_bk[:-2][::-1] + [chain_bk[-2]]:
+        if len(chain_fd) != len(chain_bk):
+            # If the chain lengths aren't equal, the chain is corrupted
             is_chain_corrupted = True
+        elif len(chain_fd) < 2 or len(chain_bk) < 2:
+            # Chains containing less than two entries are corrupted, as the smallest
+            # chain (an empty bin) would look like something like `[main_arena+88, 0]`.
+            is_chain_corrupted = True
+        elif len(chain_fd) == len(chain_bk) == 2 and chain_fd == chain_bk and chain_fd[-1] == 0:
+            # Check if bin[index] points to itself (is empty)
+
+            # If the fd pointer is in libc data section, we can be more
+            # confident that this list is actually empty and not just
+            # corrupted. If we can't find any mapping for the pointer or it's in
+            # some other mapping, we assume the bin is corrupted
+            page = pwndbg.gdblib.vmmap.find(chain_fd[0])
+            if page and page.rw and "libc" in page.objfile:
+                chain_fd = [0]
+                chain_bk = [0]
+            else:
+                is_chain_corrupted = True
+        elif chain_fd[-1] == chain_bk[-1] == 0:
+            # This branch will be hit when we were able to fully traverse the fd
+            # and bk chains (and they aren't empty). In this case we can check
+            # the integrity of the entire chain
+
+            # Get the fd chain list without the trailing zero
+            fd_chain = chain_fd[:-1]
+
+            # Get the bk chain list without the trailing zero and the last
+            # pointer (the `main_arena` pointer)
+            bk_chain = chain_bk[:-2]
+
+            # Reverse the bk chain
+            bk_chain_rev = bk_chain[::-1]
+
+            # Add back on the arena pointer
+            bk_chain_rev.append(chain_bk[-2])
+
+            # If these two chains aren't equal, the chain is corrupted
+            if fd_chain != bk_chain_rev:
+                is_chain_corrupted = True
+        else:
+            # We hit the `heap_chain_limit` while traversing the fd and bk
+            # chain, so we can't check the integrity of the entire chain, but we can at
+            # least check if `chunk->bk->fd = chunk`
+            chunk = Chunk(chain_fd[0])
+            bk = Chunk(chunk.bk)
+            if bk.fd != chain_fd[0]:
+                is_chain_corrupted = True
 
         return (chain_fd, chain_bk, is_chain_corrupted)
 
--- a/pwndbg/heap/ptmalloc.py
+++ b/pwndbg/heap/ptmalloc.py
@@ -1196,7 +1196,7 @@ class GlibcMemoryAllocator(pwndbg.heap.heap.MemoryAllocator):
             is_chain_corrupted = True
         elif len(chain_fd) < 2 or len(chain_bk) < 2:
             # Chains containing less than two entries are corrupted, as the smallest
-            # chain (an empty bin) would look like something like `[main_arena+88, 0]`.
+            # chain (an empty bin) would look something like `[main_arena+88, 0]`.
             is_chain_corrupted = True
         elif len(chain_fd) == len(chain_bk) == 2 and chain_fd == chain_bk and chain_fd[-1] == 0:
             # Check if bin[index] points to itself (is empty)


pwndbg/heap/ptmalloc.py
```diff
@@ -981,14 +981,52 @@ def bin_at(self, index, arena_addr=None):
...
+            if fd_chain != bk_chain_rev:
+                is_chain_corrupted = True
+        else:
+            # We hit the `heap_chain_limit` while traversing the fd and bk
+            # chain, so we can't check the integrity of the entire chain, but we can at
+            # least check if `chunk->bk->fd = chunk`
+            chunk = Chunk(chain_fd[0])
+            bk = Chunk(chunk.bk)
+            if bk.fd != chain_fd[0]:
+                is_chain_corrupted = True
```
gsingh93, line 1242: I'm wondering instead of a boolean, this could be one of three values: corrupted, not corrupted, not sure.

pwndbg/heap/ptmalloc.py
```diff
@@ -981,14 +981,52 @@ def bin_at(self, index, arena_addr=None):
...
+
+            # If these two chains aren't equal, the chain is corrupted
+            if fd_chain != bk_chain_rev:
+                is_chain_corrupted = True
+        else:
+            # We hit the `heap_chain_limit` while traversing the fd and bk
+            # chain, so we can't check the integrity of the entire chain, but we can at
+            # least check if `chunk->bk->fd = chunk`
+            chunk = Chunk(chain_fd[0])
+            bk = Chunk(chunk.bk)
```
gsingh93, line 1240: I wonder if `chunk.bk` could return a `Chunk` instead of an `int`?

CptGibbon: You could add a `previous_chunk()` method for example, since this is meant to mimic the malloc_chunk struct's fields. If someone wanted the `bk` value itself, that is lost as soon as you get the previous chunk. That would also be in line with the current `next_chunk()` method, but I suppose it would always return `None` for singly-linked lists.

disconnect3d:
What's the status here? Seems that some heap tests are failing

gsingh93:
Will be very busy for a bit longer, I think I can complete this before the next release.

pwndbg/heap/ptmalloc.py
gsingh93: Sorry if I'm misunderstanding, but `Chunk.next_chunk` seems to return the next chunk in memory, whereas I'm wondering if `Chunk.fd` and `Chunk.bk` should just return a `Chunk` for convenience. If we'd like to keep this inline with the structs fields, maybe `Chunk.fd_chunk` and `Chunk.bk_chunk` could just be implemented as `def fd_chunk(self): return Chunk(self.fd)`.

It's not a major difference, but I'd prefer if the `Chunk` class took care of this for me.

gsingh93:
@CptGibbon tests are passing now, should be ready for review

pwndbg/heap/ptmalloc.py
```diff
@@ -1191,14 +1191,55 @@ def bin_at(self, index, arena_addr=None):
...
+            is_corrupted = True
+        elif len(chain_fd) == len(chain_bk) == 2 and chain_fd == chain_bk and chain_fd[-1] == 0:
+            # Check if bin[index] points to itself (is empty)
+
+            # If the fd pointer is in libc data section, we can be more
+            # confident that this list is actually empty and not just
+            # corrupted. If we can't find any mapping for the pointer or it's in
+            # some other mapping, we assume the bin is corrupted
+            page = pwndbg.gdblib.vmmap.find(chain_fd[0])
+            if page and page.rw and "libc" in page.objfile:
```
lebr0nli, line 1209: I do not very understand the logic behind this yet, so probably I misunderstood something:

I have some concerns about the check here:
1. What if vmmap does not work well(e.g. for the native qemu-user), will it make pwndbg think the empty bins are corrupted bins?
2. If we are debugging a binary with statically linked libc, will it make pwndbg think the empty bins are corrupted bins since `"libc" in page.objfile` is never be `True`?
3. IIRC, bins in non-main arena might not in someone's .data sectioin, does this check also true for the bins not in main_arena?
4. Seems like it might pass the check if the address is on the same page as the .data section, but not actually "in" the .data section? Maybe parsing `info files` will be better if we want to know if the address is in .data section or not?

gsingh93: > What if vmmap does not work well(e.g. for the native qemu-user), will it make pwndbg think the empty bins are corrupted bins?

Hmm, I think you're right. Maybe we should have a function that returns whether the `vmmap` output can be trusted or not?

> If we are debugging a binary with statically linked libc, will it make pwndbg think the empty bins are corrupted bins since "libc" in page.objfile is never be True?

I didn't think about this, you're right.

> IIRC, bins in non-main arena might not in someone's .data sectioin, does this check also true for the bins not in main_arena?

Good point. I can add a check here to only do this logic if it's the main arena?

> Seems like it might pass the check if the address is on the same page as the .data section, but not actually "in" the .data section? Maybe parsing info files will be better if we want to know if the address is in .data section or not?

Sure, I'm more worried though about a non-corrupted chain being marked as corrupted, not a corrupted chain being marked as not corrupted. The latter is just best effort on our part, while the former would be considered a bug.

pwndbg/heap/ptmalloc.py
CptGibbon: I agree that `Chunk` should handle this, I can put together something like you suggest 👍

CptGibbon:
> @CptGibbon tests are passing now, should be ready for review

I can take a thorough look next week old chap 👍

pwndbg/heap/ptmalloc.py
lebr0nli: > I can add a check here to only do this logic if it's the main arena?

Hmm, instead of still checking the fd_chain/bk_chain point to the .data section or not, could we check the front and back pointers instead of the fd_chain/bk_chain?
I mean, something like this:
```diff
--- a/pwndbg/heap/ptmalloc.py
+++ b/pwndbg/heap/ptmalloc.py
@@ -1200,13 +1200,7 @@ class GlibcMemoryAllocator(pwndbg.heap.heap.MemoryAllocator):
             is_corrupted = True
         elif len(chain_fd) == len(chain_bk) == 2 and chain_fd == chain_bk and chain_fd[-1] == 0:
             # Check if bin[index] points to itself (is empty)
-
-            # If the fd pointer is in libc data section, we can be more
-            # confident that this list is actually empty and not just
-            # corrupted. If we can't find any mapping for the pointer or it's in
-            # some other mapping, we assume the bin is corrupted
-            page = pwndbg.gdblib.vmmap.find(chain_fd[0])
-            if page and page.rw and "libc" in page.objfile:
+            if front == back and front == current_base:
                 chain_fd = [0]
                 chain_bk = [0]
             else:
```
Is this correct? (This is entirely from my observation, I'm not very sure :p)

pwndbg/heap/ptmalloc.py
```diff
@@ -1191,14 +1191,55 @@ def bin_at(self, index, arena_addr=None):
...
-
-        # check if corrupted
-        elif chain_fd[:-1] != chain_bk[:-2][::-1] + [chain_bk[-2]]:
+        if len(chain_fd) != len(chain_bk):
+            # If the chain lengths aren't equal, the chain is corrupted
             is_chain_corrupted = True
+        elif len(chain_fd) < 2 or len(chain_bk) < 2:
+            # Chains containing less than two entries are corrupted, as the smallest
+            # chain (an empty bin) would look like something like `[main_arena+88, 0]`.
+            is_corrupted = True
```
CptGibbon: I may be missing something, but I don't see this variable being used anywhere.
If that is the case, then a test for this scenario could be useful.

disconnect3d:
Ping @CptGibbon :)

disconnect3d:
Ping @CptGibbon :)

pwndbg/heap/ptmalloc.py
gsingh93: Sorry, I meant `is_chain_corrupted` here.

pwndbg/heap/ptmalloc.py
CptGibbon: On revisiting this, it's not as simple as it might seem.

Since tcache fd's are treated differently to other bin fd's (they point to user data rather than chunk metadata) the `Chunk` object needs to check if it represents a chunk that's in a tcachebin, so it can either do `Chunk(chunk.fd)` or `Chunk(chunk.fd - (size_sz*2))`.

This could involve checking the size field, and if it's in tcache range searching the appropriate tcachebin for this chunk's address. But assuming corruption requires searching *every* tcachebin for this chunk.

Since the `bin_at()` method only operates on the "normal" bins, there's no need to do this calculation. I think in this case using `Chunk(chunk.bk)` avoids unnecessary reads from gdb whilst still being clear as to what it's doing.

pwndbg/heap/ptmalloc.py
CptGibbon: Yeah I guess there must be a gap in the current tests since this still passed even with an unused variable.
Probably worth a adding a test that checks for this case wouldn't you say?

disconnect3d:
What's the status of this? Can we re-trigger CI here? For some reason I think I can't? Maybe we should just push whatever here

gsingh93:
Was just planning on adding a test but that can be done in a future PR.

pwndbg/heap/ptmalloc.py
```diff
@@ -1191,14 +1191,55 @@ def bin_at(self, index, arena_addr=None):
...
-            chain_bk = [0]
-
-        # check if corrupted
-        elif chain_fd[:-1] != chain_bk[:-2][::-1] + [chain_bk[-2]]:
+        if len(chain_fd) != len(chain_bk):
+            # If the chain lengths aren't equal, the chain is corrupted
+            is_chain_corrupted = True
+        elif len(chain_fd) < 2 or len(chain_bk) < 2:
+            # Chains containing less than two entries are corrupted, as the smallest
+            # chain (an empty bin) would look like something like `[main_arena+88, 0]`.
```
```suggestion
            # chain (an empty bin) would look something like `[main_arena+88, 0]`.
```
making this suggestion+commit to re-trigger build ;)

k4lizen:
Whats the status on this? The issue that this PR fixes is very easy to hit (just need more than eight chunks in a bin list...).

It might be good to remove the vmmap check (if thats the problem?), merge this, and open a new issue on improving the check.

gsingh93:
It needs a rebase, and then just making sure all the tests pass. I don't have time though, feel free to make a new PR based off of this one if you want.

PR #1574, opened 2023-02-08
Fix OnlyWithArch invocations (cymbol command)

disconnect3d:

### DIFF
--- a/pwndbg/commands/cymbol.py
+++ b/pwndbg/commands/cymbol.py
@@ -249,7 +249,7 @@ parser.add_argument(
 
 
 @pwndbg.commands.ArgparsedCommand(parser)
-@pwndbg.commands.OnlyWithArch(["x86_64"])
+@pwndbg.commands.OnlyWithArch(["x86-64"])
 @pwndbg.commands.OnlyWhenRunning
 def cymbol(add, remove, edit, load, show) -> None:
     if add:
--- a/pwndbg/gdblib/arch.py
+++ b/pwndbg/gdblib/arch.py
@@ -7,7 +7,7 @@ from pwndbg.lib.arch import Arch
 
 # TODO: x86-64 needs to come before i386 in the current implementation, make
 # this order-independent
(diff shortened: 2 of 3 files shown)


pwndbg/gdblib/arch.py
```diff
@@ -7,7 +7,7 @@
 
 # TODO: x86-64 needs to come before i386 in the current implementation, make
 # this order-independent
-ARCHS = ("x86-64", "i386", "aarch64", "mips", "powerpc", "sparc", "arm")
+ARCHS = ("x86-64", "i386", "aarch64", "mips", "powerpc", "sparc", "arm", "armcm")
```
gsingh93, line 10: I guess the reason we're adding `armcm` here is in case we want to use it with `@OnlyWithArch`? The reason it wasn't initially in this list is that there is no real `armcm` GDB architecture, we just check for ARM architectures ending in "-m":
```
if match == "arm" and "-m" in arch:
    match = "armcm"
```

I'm slightly concerned about having an arch that doesn't directly map to a GDB arch here, but I also don't know a better solution that doesn't require more duplication than it's worth. No need to change anything here, just calling this out.

pwndbg/commands/cymbol.py
```diff
@@ -249,7 +249,7 @@ def show_custom_structure(custom_structure_name, custom_structure_path) -> None:
 
 
 @pwndbg.commands.ArgparsedCommand(parser)
-@pwndbg.commands.OnlyWithArch(["x86_64"])
+@pwndbg.commands.OnlyWithArch(["x86-64"])
```
gsingh93, line 252: 🤦

pwndbg/gdblib/arch.py
disconnect3d: I think I added it here because it was in the pwnlib arch mapping dict.

Maybe we should remove it? 🤐 

Btw the ARCHS is only used in those ways, jfyi:
image

PR #1578, opened 2023-02-13
Support using the `rich` module to print the full stacktrace

lebr0nli:
~Create a new option, `exception-use-rich`, to~ use the `rich` module to print the full stacktrace
(Use `rich.console.Console().print_exception()`)

With ~`set exception-use-rich` +~ `set exception-verbose`:

screenshot

Edit: Will always use `rich` if it's available, and `rich` is been added to `dev-requirements.txt` now.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -17,3 +17,5 @@ types-psutil==5.9.5
 types-Pygments==2.12.1
 types-tabulate==0.9.0.0
 vermin==1.5.1
+rich==12.6.0; python_version < '3.7'
+rich==13.3.1; python_version >= '3.7'
--- a/pwndbg/exception.py
+++ b/pwndbg/exception.py
@@ -14,6 +14,12 @@ with pwndbg.lib.stdio.stdio:
         import ipdb as pdb
     except ImportError:
         import pdb  # type: ignore
+    try:
+        from rich.console import Console
+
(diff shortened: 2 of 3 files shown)


disconnect3d:
1) Don't we have to install rich? Shall this be a dev-only requirement or production one? 🤐 
2) Is there any reason to have it opt-in instead of always using it?

lebr0nli:
> Don't we have to install rich? Shall this be a dev-only requirement or production one? 🤐
> Is there any reason to have it opt-in instead of always using it?

What I thought is that this can be used like the `ipi` command, just optional to use it for the dev or user if they like it. (Maybe someone likes to see the compacted default stacktrace instead of the rich one?)
But I'm also ok with that we always use `rich` if the user has installed it, and we add `rich` into the dev-only requirement.
Or maybe we can make it default to `on` instead of `off`, and add `rich` into the dev-only requirement.

wdyt? I'm ok with the above of them.

disconnect3d:
Let's use it if its available and install it in dev setup. We don't need another GDB parameter 😅

lebr0nli:
> Let's use it if its available and install it in dev setup. We don't need another GDB parameter 😅

Ok!

gsingh93:
LGTM, just need to fix the merge conflicts.

PR #1579, opened 2023-02-16
Rewrite some unreliable methods for the heap heuristics

lebr0nli:
This PR aims to rewrite some unreliable and hard-to-maintain methods:

- We directly search the default value of `main_arena` and `mp_` in the .data section of the ELF now. We do not rely on any of the symbols and the result of disassembly anymore.
- The original heuristic of `tcache` is unreliable. Since it highly depends on the result of disassembly and our features for TLS, I think that method might not be worth to keep maintaining. We now assume it is in the first chunk of the current heap. But the `tcache` might not be at the beginning of the heap of the current thread if the arena is been shared by multiple threads, so we'll show a prompt to ask the user if they want to try the brute force to find `tcache`, and let them decide if it is worth doing it.
- The original heuristic of `thread_arena` is similar to `tcache`, so the old method is been dropped. But since the `thread_arena` might still be found by brute force, we also provide an option to the user.
- Since the heuristic of `global_max_fast` is not reliable, now we return the default value of `global_max_fast` instead of trying to find it and then failing. (It still can be manually set by the user tho.)
- Now we won't implicitly call `pthread_self()` when executing `tls` command, calling `pthread_self()` is optional. Users can use them by `tls --pthread-self` to try it only if they think it's safe.
- Some minor changes for the output of `tls`, `arenas`, `arena`, `tcache`, and `mp` commands, to show more information.
- Handle the case that `thread_arena` and `tcache` is pointing to `NULL`.

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -367,15 +367,11 @@ def _try2run_heap_command(function, a, kw):
             w(
                 "You are probably debugging a multi-threaded target without debug symbols, so we failed to determine which arena is used by the current thread.\n"
                 "To resolve this issue, you can use the `arenas` command to list all arenas, and use `set thread-arena <addr>` to set the current thread's arena address you think is correct.\n"
-                "If you want to know which exact arena is used by the current thread, you can use the `tls` command to get the thread-local storage address, and then use `search -p <arena_addr>` to find which arena's address is in your thread-local storage.\n"
-                "(If none of the arenas' address is in your thread-local storage, that's probably because the current thread doesn't allocate the arena)\n"
-                "Using `arena <addr>` to check the `attached_threads` field of the arena is also a way that can help you to determine which arena is used by the current thread."
             )
             if pwndbg.heap.current.has_tcache():
                 w(
                     "Please also notice, pwndbg will assume the tcache is at the first chunk of the arena if the `tcache` symbol does not present, so when the arena is been shared by multiple threads, the result might be wrong.\n"
                     "To resolve this, you can use `set tcache <addr>` to set the tcache structure's address manually if this problem occurs.\n"
-                    "The address of tcache should also inside your thread-local storage, so you can use the `search -p <address>` to find which chunk's address is in your thread-local storage."
                 )
         else:
             w(
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -366,6 +366,10 @@ def _try2run_heap_command(function, a, kw):
         w(
             f"You can try to determine the libc symbols addresses manually and set them appropriately. For this, see the `heap_config` command output and set the config about `{err.symbol}`."
         )
+        if "thread_arena" == err.symbol:
+            w(
+                "You can use `arenas` command to see all the arenas, and use `set thread_arena <addr>` to set it manually."
+            )
         if pwndbg.gdblib.config.exception_verbose or pwndbg.gdblib.config.exception_debugger:
             raise err
         else:
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -363,12 +363,23 @@ def _try2run_heap_command(function, a, kw):
         return function(*a, **kw)
     except SymbolUnresolvableError as err:
         e(f"{function.__name__}: Fail to resolve the symbol: `{err.symbol}`")
-        w(
-            f"You can try to determine the libc symbols addresses manually and set them appropriately. For this, see the `heap_config` command output and set the config about `{err.symbol}`."
-        )
         if "thread_arena" == err.symbol:
             w(
-                "You can use `arenas` command to see all the arenas, and use `set thread_arena <addr>` to set it manually."
+                "You are probably debugging a multi-threaded target without debug symbols, so we failed to determine which arena is used by the current thread.\n"
+                "To resolve this issue, you can use the `arenas` command to list all arenas, and use `set thread-arena <addr>` to set the current thread's arena address you think is correct.\n"
+                "If you want to know which exact arena is used by the current thread, you can use the `tls` command to get the thread-local storage address, and then use `search -p <arena_addr>` to find which arena's address is in your thread-local storage.\n"
+                "(If none of the arenas' address is in your thread-local storage, that's probably because the current thread doesn't allocate the arena)\n"
+                "Using `arena <addr>` to check the `attached_threads` field of the arena is also a way that can help you to determine which arena is used by the current thread."
+            )
+            if pwndbg.heap.current.has_tcache():
+                w(
+                    "Please also notice, pwndbg will assume the tcache is at the first chunk of the arena if the `tcache` symbol does not present, so when the arena is been shared by multiple threads, the result might be wrong.\n"
+                    "To resolve this, you can use `set tcache <addr>` to set the tcache structure's address manually if this problem occurs.\n"
+                    "The address of tcache should also inside your thread-local storage, so you can use the `search -p <address>` to find which chunk's address is in your thread-local storage."
+                )
+        else:
+            w(
+                f"You can try to determine the libc symbols addresses manually and set them appropriately. For this, see the `heap_config` command output and set the config about `{err.symbol}`."
             )
         if pwndbg.gdblib.config.exception_verbose or pwndbg.gdblib.config.exception_debugger:
             raise err
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -370,7 +370,7 @@ def _try2run_heap_command(function, a, kw):
             )
         else:
             w(
-                f"You can try to determine the libc symbols addresses manually and set them appropriately. For this, see the `heap_config` command output and set the config about `{err.symbol}`."
+                f"You can try to determine the libc symbols addresses manually and set them appropriately. For this, see the `heap_config` command output and set the config for `{err.symbol}`."
             )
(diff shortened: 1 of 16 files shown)


lebr0nli:
I think this PR is ready for review.

----

Since this PR changed a lot of files, here are some details about the changes, hopefully these can make this PR easier to be reviewed:

- In `pwndbg/gdblib/file.py`, I've loosened the check for the `get_file` function to accept relative paths.
- In `pwndbg/gdblib/proc.py` and `pwndbg/glibc.py`, I've added some new functions to get the address of the .data and .got sections and a new function to dump the contents of the .data section of the ELF file. These functions are for the new heap heuristic in `pwndbg/gdblib/ptmalloc.py` to find the address of important structures.
- In `pwndbg/gdblib/tls.py`, I removed some unused function and fixed bugs when trying to get the TLS address via calling the `pthread_self` function.
- In `pwndbg/command/tls.py`, now `tls` won't call `pthread_self` by default and it will show the region of the TLS address.
- In `tests/gdb-tests/tests/binaries/` and `tests/gdb-tests/tests/test_command_tls.py`, I created tests for the TLS command and features.
- In `pwndbg/heap/structs.py`, I replaced `ctypes.LittleEndianStructure` with `Structure` from `pwndbg.gdblib.ctypes` to make this module work on big endian systems. I also added the default value of `mp_` structure into it for the new heap heuristic.
- In `pwndbg/heap/__init__.py`, I updated the description of some heap configs to make them more accurate.
- In `pwndbg/heap/ptmalloc.py`, I fixed some bugs that were caused by cases where the `thread_arena` and `tcache` point to `NULL`. I also changed some unreliable and hard-to-maintain code:
  - For finding the `main_arena` and `mp_`, we now search their default values in the ELF directly, no longer relying on other symbols and any disassembly results.
  - For finding the `tcache`, we now assume that the tcache is always located at the beginning of the current heap, with an optional brute force fallback to search the possible address in the TLS if user think it isn't at the beginning.
  - For finding the `thread_arena`, we only provide the brute force way to find the address of the `thread_arena`. We suggest user manually set it if they can know which arena they are using after they saw the output of `arenas`.
  - For finding the `global_max_fast`, we will only return the default value if we fail to find it by symbol or config.
- In `pwndbg/commands/__init__.py`, I added new hints when we fail to find the address of the `thread_arena`.
- In `tests/gdb-tests/tests/heap/test_heap.py`, I updated the tests for the heap heuristic.
- In `pwndbg/command/heap.py`, I updated some of the heap commands to handle the cases where the `thread_arena` and `tcache` point to `NULL`. I also update the output of `arena`, `arenas`, `mp`, and `tcache` to show more information, so when we are using the heap heuristics, we can know more about the heap.

----

Btw, to test some of the features of the heap heuristics easier, `set debug-file-directory` before running the program might be helpful.

e.g. `gdb -q /bin/ls -ex 'set debug-file-directory' -ex 'entry' -ex 'heap'`

pwndbg/commands/__init__.py
```diff
@@ -363,9 +363,15 @@ def _try2run_heap_command(function, a, kw):
...
-            f"You can try to determine the libc symbols addresses manually and set them appropriately. For this, see the `heap_config` command output and set the config about `{err.symbol}`."
-        )
+        if "thread_arena" == err.symbol:
+            w(
+                "You are probably debugging a multi-threaded target without debug symbols, so we failed to determine which arena is used by the current thread.\n"
+                "To resolve this issue, you can use the `arenas` command to list all arenas, and use `set thread-arena <addr>` to set the current thread's arena address you think is correct.\n"
+            )
+        else:
+            w(
+                f"You can try to determine the libc symbols addresses manually and set them appropriately. For this, see the `heap_config` command output and set the config about `{err.symbol}`."
```
```suggestion
                f"You can try to determine the libc symbols addresses manually and set them appropriately. For this, see the `heap_config` command output and set the config for `{err.symbol}`."
```

pwndbg/commands/heap.py
```diff
@@ -141,6 +141,10 @@ def heap(addr=None, verbose=False, simple=False) -> None:
             chunk = chunk.next_chunk()
     else:
         arena = allocator.thread_arena
+        # arena might be None if the current thread doesn't allocate the arena
+        if arena is None:
+            print(message.notice("No arena found for current thread."))
```
```suggestion
            print(message.notice("No arena found for current thread (the thread hasn't performed any allocations)."))
```

pwndbg/commands/heap.py
```diff
@@ -168,9 +172,17 @@ def arena(addr=None) -> None:
         arena = Arena(addr)
     else:
         arena = allocator.thread_arena
+        tid = pwndbg.gdblib.proc.thread_id
+        # arena might be None if the current thread doesn't allocate the arena
+        if arena is None:
+            print(message.notice(f"No arena found for thread {message.hint(tid)}."))
```
```suggestion
            print(message.notice(f"No arena found for thread {message.hint(tid)} (the thread hasn't performed any allocations)."))
```

pwndbg/commands/heap.py
```diff
@@ -247,6 +271,10 @@ def top_chunk(addr=None) -> None:
         arena = Arena(addr)
     else:
         arena = allocator.thread_arena
+        # arena might be None if the current thread doesn't allocate the arena
+        if arena is None:
+            print(message.notice("No arena found for current thread."))
```
```suggestion
            print(message.notice("No arena found for current thread (the thread hasn't performed any allocations)"))
```

pwndbg/commands/heap.py
```diff
@@ -672,6 +699,10 @@ def vis_heap_chunks(addr=None, count=None, naive=None, display_all=None) -> None
         arena = heap_region.arena
     else:
         arena = allocator.thread_arena
+        # arena might be None if the current thread doesn't allocate the arena
+        if arena is None:
+            print(message.notice("No arena found for current thread."))
```
```suggestion
            print(message.notice("No arena found for current thread (the thread hasn't performed any allocations)"))
```

pwndbg/commands/heap.py
```diff
@@ -852,6 +883,10 @@ def try_free(addr) -> None:
     # constants
     allocator = pwndbg.heap.current
     arena = allocator.thread_arena
+    # arena might be None if the current thread doesn't allocate the arena
+    if arena is None:
+        print(message.notice("No arena found for current thread."))
```
```suggestion
        print(message.notice("No arena found for current thread (the thread hasn't performed any allocations)"))
```

pwndbg/commands/heap.py
```diff
@@ -1215,6 +1250,6 @@ def heap_config(filter_pattern) -> None:
 
     print(
         message.hint(
-            "Some config(e.g. main_arena) will only working when resolve-heap-via-heuristic is `True`"
+            "Some config(e.g. main_arena) will only working when resolve-heap-via-heuristic is `auto` or `force`"
```
```suggestion
            "Some config values (e.g. main_arena) will be used only when resolve-heap-via-heuristic is `auto` or `force`"
```

pwndbg/gdblib/proc.py
```diff
@@ -103,6 +105,44 @@ def binary_base_addr(self) -> int:
...
+    def dump_elf_data_section(self) -> Optional[Tuple[int, int, bytes]]:
+        """
+        Dump .data section of current process's ELF file
+        """
+        local_path = pwndbg.gdblib.file.get_file(pwndbg.gdblib.proc.exe)
+        with open(local_path, "rb") as f:
+            elffile = ELFFile(f)
+            section = elffile.get_section_by_name(".data")
+            if section:
+                return section["sh_addr"], section["sh_size"], section.data()
```
disconnect3d: perhaps we can move this to some gdblib.file or gdblib.elf function; I mean we could create a func that returns given section addr, size and data

pwndbg/gdblib/proc.py
```diff
@@ -103,6 +105,44 @@ def binary_base_addr(self) -> int:
...
+        """
+        out = pwndbg.gdblib.info.files()
+        for line in out.splitlines():
+            if line.endswith(" is .data"):
+                return int(line.split()[0], 16)
+        return 0
+
+    @pwndbg.lib.memoize.reset_on_start
+    @pwndbg.lib.memoize.reset_on_objfile
+    def get_got_section_address(self) -> int:
```
disconnect3d, line 129: We already have something that looks for .got section in current binary in gdblib.elf iirc, do we need this?

lebr0nli: > We already have something that looks for .got section in current binary in gdblib.elf iirc

Hmm, I didn’t find something related in `gdblib.elf`. Which function did you mean?
Or did you mean this function in `commands.elf`?
https://github.com/pwndbg/pwndbg/blob/f627b614627a58b2229357dc3e9d8d1b048cd738/pwndbg/commands/elf.py#L67-L93

> do we need this?

Since the above approach might need to download the file from the remote server, I'll prefer keeping this function and parsing the output of the `info files`. (but I’m not sure which approach is better tho)

pwndbg/gdblib/proc.py
lebr0nli: Ok, I'll move some of the code into `gdblib.elf`.

CptGibbon:
Nitpick: inconsistency in thread-related error messages (only `arena` tells you the thread number, the rest just say "current thread")

image

CptGibbon (approved):
### From a GLIBC malloc perspective:
I've spent the day testing this and haven't been able to trip it up, it seems very reliable and far easier to maintain than the previous solution. Thank you once again for the great work 🙏

disconnect3d:
> Nitpick: inconsistency in thread-related error messages (only `arena` tells you the thread number, the rest just say "current thread")
> 
> image

Let's fix this in another PR

PR #1581, opened 2023-02-19
feature: Add `killthreads` command (closes #1580)

### LINKED ISSUE #1580: Add command to kill threads
We killed all threads except current on a CTF recently with @alufers and he suggested we should have a command for that. (I think he started implementing it, so I am going to assign him too)

This command should:
* allow to kill `--all` or choosen threads
* should kill the threads via switching to those threads and calling `call pthread_exit(0)`
* should `set scheduler-locking on` before doing any action and revert the original value of this afterwards

Regarding the last point, we have a `with lock_scheduler():` in https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/gdblib/scheduler.py#L7

alufers:
This command allows the user to quickly kill multiple  threads by specyfying their ids as arguments to this command. It also supports the `--all` flag, which will kill overy thread except the currently selected one. This is useful for use with the `checkpoint` command.

The killing is done by calling `pthread_exit(0)`.


alufers:
Cmon, the tests pass locally :D

pwndbg/commands/killthreads.py
```diff
@@ -0,0 +1,52 @@
...
+            except gdb.error as e:
+                # gdb will throw an error, because the thread dies during the call, which is expected
+                pass
+
+        # Switch back to the thread we were on before killing threads
+        gdb.execute(f"thread {current_thread_id}")
+        if len(thread_ids) == 1:
+            print(message.success("Killed 1 thread"))
+        else:
+            print(message.success(f"Killed {len(thread_ids)} threads"))
```
disconnect3d: Can we print the IDs of threads we kill?

pwndbg/commands/killthreads.py
```diff
@@ -0,0 +1,52 @@
...
+from typing import Optional
+
+import gdb
+
+import pwndbg.commands
+from pwndbg.color import message
+from pwndbg.commands import CommandCategory
+from pwndbg.gdblib.scheduler import lock_scheduler
+
+parser = argparse.ArgumentParser(description="Kill multiple threads at once.")
```
disconnect3d: Can we add a long description (and make sure newline works here, I think some other command like vmmap have description with newlines) and inform users on:
```
Kill all or given threads.

Switches to given threads and calls pthread_exit(0) on them.
This is performed with scheduler-locking to prevent other threads from operating at the same time.

Killing all other threads may be useful to use GDB checkpoints, e.g., to test given input & restart the execution to the point of interest (checkpoint).
```

pwndbg/commands/killthreads.py
```diff
@@ -0,0 +1,52 @@
...
+
+    with lock_scheduler():
+        current_thread_id = gdb.selected_thread().num
+        if all:
+            thread_ids = [
+                thread.num
+                for thread in gdb.selected_inferior().threads()
+                if thread.num != current_thread_id
+            ]
+        for thread_id in thread_ids:
```
disconnect3d, line 59: Should we maybe make sure that the provided thread ids are valid?

pwndbg/commands/killthreads.py
```diff
@@ -0,0 +1,52 @@
...
+            gdb.execute(f"thread {thread_id}")
+            try:
+                gdb.execute("call (void) pthread_exit(0)")
```
disconnect3d: Would calling all gdb.execute here with `to_string=True` prevent some GDB output from appearing?

tests/gdb-tests/tests/test_command_killthreads.py
```diff
@@ -0,0 +1,60 @@
...
+    # check if thread with id 3 exists
+    assert len([thread for thread in gdb.selected_inferior().threads() if thread.num == 3]) == 1
+
+    gdb.execute("killthreads 3")
+
+    # check if the thread was killed, and no other thread was killed
+    assert len([thread for thread in gdb.selected_inferior().threads() if thread.num == 3]) == 0
+    assert len(gdb.selected_inferior().threads()) == initial_thread_count - 1
+
+    gdb.execute("kill")
```
disconnect3d, line 60: Is this needed?

disconnect3d:
```
Launching pytest with args: ['/pwndbg/tests/gdb-tests/tests/test_command_killthreads.py::test_command_killthreads_kills_specific_thread', '-vvv', '-s', '--showlocals', '--color=yes']
============================= test session starts ==============================
platform linux -- Python 3.6.9, pytest-7.0.1, pluggy-1.0.0 -- /usr/bin/python
cachedir: .pytest_cache
rootdir: /pwndbg/tests/gdb-tests
collecting ... collected 1 item

tests/test_command_killthreads.py::test_command_killthreads_kills_specific_thread 

Program stopped.
0x00007ffff7dd4090 in _start () from /lib64/ld-linux-x86-64.so.2
Breakpoint 1 at 0x5555554007f6: file multiple_threads.c, line 26.
[New Thread 0x7ffff77c2700 (LWP 1040)]
[New Thread 0x7ffff67c1700 (LWP 1041)]

Thread 1 "multiple_thread" hit Breakpoint 1, break_here () at multiple_threads.c:26
26	}
[Switching to thread 3 (Thread 0x7ffff67c1700 (LWP 1041))]
#0  clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:78

Thread 3 "multiple_thread" received signal SIGSEGV, Segmentation fault.
unwind_stop (version=<optimized out>, actions=<optimized out>, exc_class=<optimized out>, exc_obj=<optimized out>, context=0x7ffff67c0dc0, stop_parameter=0x0) at unwind.c:60
[Switching to thread 1 (Thread 0x7ffff7fe8740 (LWP 1037))]
#0  break_here () at multiple_threads.c:26
26	}
Killed threads with IDs: 3
FAILED
```

I think the test fails because the process gets SIGSEGV, but why does it get it? :/ Can it be that we call pthread_exit in a time when we should not?

disconnect3d:
@alufers can you try to fix the tests?

tests/gdb-tests/tests/test_command_killthreads.py
```diff
@@ -1,21 +1,37 @@
...
+            assert False, "Timeout reached"
+
+
 def test_command_killthreads_kills_all_threads_except_current(start_binary):
     start_binary(REFERENCE_BINARY_THREADS)
 
     gdb.execute("break break_here")
     gdb.execute("run")
-    assert len(gdb.selected_inferior().threads()) != 1
+    wait_until(lambda: len(gdb.selected_inferior().threads()) != 1)
```
disconnect3d: We should probably just change the multiple_threads example so that the break_here is triggered only after other thread[s] are spawned. In other words, the other threads should synchronize with the main thread so that break_here is only called after we are in the desired state 😅

alufers: I've added the semaphores, but it still fails, but later. I think there is a delay between when the thread dies and gdb reports it as gone

alufers:
It works!

pwndbg/commands/killthreads.py
```diff
@@ -0,0 +1,74 @@
...
+    action="store_true",
+    help="Kill all threads except the current one.",
+)
+
+
+@pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.PROCESS)
+@pwndbg.commands.OnlyWhenRunning
+def killthreads(thread_ids: Optional[List] = None, all: bool = False) -> None:
+    if len(thread_ids) == 0 and not all:
+        print(message.error("No thread IDs specified and --all not set"))
```
disconnect3d:
```suggestion
        print(message.error("No thread IDs or --all flag specified"))
```

Issue #1565, opened 2023-02-03
Infinite recursion occurred in `pwndbg.gdblib.vmmap.get()` when `info proc mappings` is empty


lebr0nli:
Steps to reproduce:

```console
$ cat test.c
#include <pthread.h>
#include <stdio.h>
#include <unistd.h>
#include <stdlib.h>

void *thread_function(void *arg) {
  free(malloc(0x20));
  sleep(100);
  return NULL;
}

int main(void) {
  pthread_t thread;
  pthread_create(&thread, NULL, thread_function, NULL);
  pthread_exit(NULL);
  return 0;
}

$ gcc test.c -pthread -g -o test
$ gdb -q test -ex 'set exception-verbose on' -ex 'break thread_function' -ex 'run' -ex 'next'
```

The traceback is too long to copy, the flow which causes infinity recursion is `pwndbg.gdblib.vmmap.get()` -> `pwndbg.gdblib.vmmap.info_auxv()` -> `pwndbg.gdblib.elf.map()` -> `pwndbg.gdblib.elf.get_ehdr()` -> `pwndbg.gdblib.vmmap.find()` -> `pwndbg.gdblib.vmmap.get()`.

gsingh93:
Seems similar to https://github.com/pwndbg/pwndbg/issues/1219

disconnect3d:
@lebr0nli What environment do u run this on? The steps provided seems to work fine on my machine with Ubuntu 18.04

lebr0nli:
> What environment do u run this on? The steps provided seems to work fine on my machine with Ubuntu 18.04

I tested in my Ubuntu 20.04 docker container, with GDB 11.2.

disconnect3d:
I was able to reproduce this with `docker-compose run ubuntu20.04 bash` and then reproducing the steps above.

disconnect3d:
This should be fixed by #1585 which also adds a regression test for this exact case.

PR #1585, opened 2023-02-20
Hopefully fix vmmap recursion issues

disconnect3d:
This fixes the issue from #1565 but does not fix all vmmap/ehdr recursion errors.

### DIFF
--- a/pwndbg/gdblib/vmmap.py
+++ b/pwndbg/gdblib/vmmap.py
@@ -6,7 +6,6 @@ The reason that we need robustness is that not every operating
 system has /proc/$$/maps, which backs 'info proc mapping'.
 """
 import bisect
-import os
 from typing import Any
 from typing import List
 from typing import Optional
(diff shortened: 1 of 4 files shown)


pwndbg/gdblib/vmmap.py
```diff
@@ -88,13 +87,25 @@ def get() -> Tuple[pwndbg.lib.memory.Page, ...]:
...
-        and pwndbg.gdblib.arch.current in ("i386", "x86-64", "aarch64", "riscv:rv64")
+    if is_corefile():
+        return tuple(coredump_maps())
+
+    proc_maps = proc_pid_maps()
+
+    # The `proc_maps` is usually a tuple of Page objects but it can also be:
+    #   None    - when /proc/$pid/maps does not exist/is not available
+    #   tuple() - when the process has no maps yet which happens only during its very early init
+    #             (usually when we attach to a process)
```
gsingh93, line 99: `proc_pid_maps` currently describes the return value as "A list of pwndbg.lib.memory.Page objects". We should update it with this information.

pwndbg/gdblib/vmmap.py
```diff
@@ -88,13 +87,25 @@ def get() -> Tuple[pwndbg.lib.memory.Page, ...]:
...
+    if proc_maps is not None:
+        return proc_maps
+
+    pages = []
+    if pwndbg.gdblib.qemu.is_qemu_kernel() and pwndbg.gdblib.arch.current in (
```
gsingh93, line 104: Doesn't need to be done in this PR, but was just thinking this function would be more readable if we move the body of this if statement to some function like `qemu_kernel_maps`. Then it's more clear what's going on in this function, i.e. first coredump maps, then process maps, then qemu kernel maps, etc.

pwndbg/gdblib/vmmap.py
```diff
@@ -112,9 +123,6 @@ def get() -> Tuple[pwndbg.lib.memory.Page, ...]:
         elif kernel_vmmap == "monitor":
             pages.extend(kernel_vmmap_via_monitor_info_mem())
 
-    if not pages and is_corefile():
-        pages.extend(coredump_maps())
-
     # TODO/FIXME: Do we still need it after coredump_maps()?
     # Add tests for other cases and see if this is needed e.g. for QEMU user
     # if not, remove the code below & cleanup other parts of Pwndbg codebase
```
gsingh93: I guess if this really isn't needed, just removing it would fix the recursion issues as well?

gsingh93 (approved):
lgtm

pwndbg/gdblib/vmmap.py
disconnect3d: Fixed

pwndbg/gdblib/vmmap.py
disconnect3d: I rephrased this. I probably need to test this edge case a bit more.

PR #1588, opened 2023-02-22
Added support for riscv:rv64

zommiommy:
Hi! Thanks for pwndbg :)

I've added the support for riscv, which is needed to debug code inside `qemu-riscv64`.
The change is small, let me know if I need to apply any changes to get it merged!

Thank you,
Tommaso


disconnect3d:
Hey, do u maybe have a list of commands I can use to test this out (including compilation of a risc-v binary on ubuntu)? :P

Btw we had a similar PR in the past in https://github.com/pwndbg/pwndbg/pull/829/ which tried to implement even more RISC-V stuff. May be worth looking into that too.

disconnect3d:
Comparing to #829 I am wondering: does the disassembly work properly? (since we use Capstone and this PR does not change anything related to it)

disconnect3d:
PS: don't worry about the archlinux CI failure. It is irrelevant to your changes.

zommiommy:
Yeah, I see that #829 is much more extensive, so the best choice is to integrate that one.

The current state is the following:
image

Sure, I have a little rust executable I use for testing. I'll list down the passages so that they might be used to test #829.

```shell
# Install Rust
curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
# Install cross for rust crosscompilation
cargo install cross
# Create a new hello-world rust executable
cargo new --bin test_riscv64
# Move inside the dir
cd test_riscv64
# Compile it
cross run --target riscv64gc-unknown-linux-gnu
# Execute 
sudo qemu-riscv64 -singlestep -g 666 -L /usr/riscv64-linux-gnu/ ./target/riscv64gc-unknown-linux-gnu/debug/test_fuzz
```

On my system I installed it using the AUR package: [riscv64-linux-gnu-gcc](https://archlinux.org/packages/community/x86_64/riscv64-linux-gnu-gcc/), on Ubuntu is should be [libc6-riscv64-cross](https://packages.ubuntu.com/focal/libc6-riscv64-cross)

Alternatively, you can compile the target `riscv64gc-unknown-linux-musl` to get a static binary.

zommiommy:
Currently I've only added the integer registers, not yet the floats one

disconnect3d:
Looks good to me, let's merge this for now. Supporting more variants of RISC-V can be done in separate PRs and/or when people actually need it :P.

zommiommy:
Thank you!!!

PR #1593, opened 2023-02-24
Avoid downloading files from remote server in some cases

### LINKED ISSUE #1591: No such file or directory When tried to download libc
I used last commit a3dcfaa139b1e5e4cd23848f9a1b754c9de0fc5b
I attach to process with gdb-multiarch (using `pwndbg` )  and I set a sysroot for all the binaries of this device in my PC 

`set sysroot /mnt/hgfs/sysroot/`

When I tried to press `vis` I see pwndbg tried to download the libc from the device (to analyze the heap).
The prbolem is , pwndbg tried to download the libc from the sysroot (in my PC)  and that failed 
```


  File "/home/USER/pwndbg/pwndbg/commands/__init__.py", line 184, in __call__
  File "/home/USER/pwndbg/pwndbg/commands/__init__.py", line 320, in _OnlyWhenRunning
    return function(*a, **kw)
  File "/home/USER/pwndbg/pwndbg/commands/__init__.py", line 420, in _OnlyWithResolvedHeapSyms
    return _try2run_heap_command(function, a, kw)
  File "/home/USER/pwndbg/pwndbg/commands/__init__.py", line 388, in _try2run_heap_command
    raise err
  File "/home/USER/pwndbg/pwndbg/commands/__init__.py", line 363, in _try2run_heap_command
    return function(*a, **kw)
  File "/home/USER/pwndbg/pwndbg/commands/__init__.py", line 345, in _OnlyWhenHeapIsInitialized
    return function(*a, **kw)
  File "/home/USER/pwndbg/pwndbg/commands/heap.py", line 737, in vis_heap_chunks
    arena = allocator.thread_arena
  File "/home/USER/pwndbg/pwndbg/heap/ptmalloc.py", line 1450, in thread_arena
    if self.main_arena.address != pwndbg.heap.current.main_arena.next or self.multithreaded:
  File "/home/USER/pwndbg/pwndbg/heap/ptmalloc.py", line 1299, in main_arena
    section = pwndbg.glibc.dump_elf_data_section()
  File "/home/USER/pwndbg/pwndbg/gdblib/proc.py", line 143, in wrapper
    return func(*a, **kw)
  File "/home/USER/pwndbg/pwndbg/glibc.py", line 88, in dump_elf_data_section
    return pwndbg.gdblib.elf.dump_section_by_name(libc_filename, ".data")
  File "/home/USER/pwndbg/pwndbg/gdblib/elf.py", line 181, in dump_section_by_name
    local_path = pwndbg.gdblib.file.get_file(filepath)
  File "/home/USER/pwndbg/pwndbg/gdblib/file.py", line 72, in get_file
    raise OSError("Could not download remote file %r:\n" "Error: %s" % (path, error))
OSError: Could not download remote file '/mnt/hgfs/sysroot/lib/libc.so.6':
Error: Remote I/O error: No such file or directory
```

The problem is that GDB doesn't support to run 
`remote get "/mnt/hgfs/sysroot/lib/libc.so.6" "/tmp/tmaaa"` if that file is located in my PC (because of `sysroot` )

lebr0nli:
In some cases, we might not need to download the file if we can find the file locally, and this can be determined by the return value of the gdb command/API.

For example, the path we get from `pwndbg.gdblib.proc.exe` has different when gdb have loaded the file or not:

If we have executed `file /path/to/exe` before connecting to gdbserver:
```python
In [1]: pwndbg.gdblib.proc.exe
Out[1]: '/path/to/exe'
```
The output doesn't have "target:" prefix because gdb is able to find the file locally.

If we haven't executed `file /path/to/exe`, gdb can't find it locally, the output will be:
```python
In [1]: pwndbg.gdblib.proc.exe
Out[1]: 'target:/path/to/exe'
```

So if the "target:" prefix doesn't present when using `pwndbg.gdblib.file.get_file(pwndbg.gdblib.proc.exe)`, we don't need to download the file. (The behavior of file path also occurred in `info files` and `info sharedlibrary` commands output.)

This PR aims to prevent downloading the file when the "target:" prefix doesn't present in the above cases.

---

This PR should fix #1591
(And will probably fix the potential similar issues on every place we used `pwndbg.gdblib.file.get_file(pwndbg.gdblib.proc.exe)`.

### DIFF
--- a/pwndbg/gdblib/file.py
+++ b/pwndbg/gdblib/file.py
@@ -35,13 +35,22 @@ def remote_files_dir():
     return _remote_files_dir
 
 
-def get_file(path: str) -> str:
+def get_proc_exe_file() -> str:
+    """
+    Returns the local path to the debugged file name.
+    """
+    return get_file(pwndbg.gdblib.proc.exe, try_local_path=True)
+
+
+def get_file(path: str, try_local_path: bool = False) -> str:
     """
     Downloads the specified file from the system where the current process is
     being debugged.
 
     If the `path` is prefixed with "target:" the prefix is stripped
     (to support remote target paths properly).
+
+    If the `try_local_path` is set to `True` and the `path` exists locally and "target:" prefix is not present, it will return the local path instead of downloading the file.
 
     Returns:
         The local path to the file
@@ -50,7 +59,8 @@ def get_file(path: str) -> str:
         "target:"
     ), "get_file called with incorrect path"
 
-    if path.startswith("target:"):
+    has_target_prefix = path.startswith("target:")
+    if has_target_prefix:
         path = path[7:]  # len('target:') == 7
 
     local_path = path
@@ -61,6 +71,8 @@ def get_file(path: str) -> str:
 
     elif pwndbg.gdblib.remote.is_remote():
         if not pwndbg.gdblib.qemu.is_qemu():
+            if try_local_path and not has_target_prefix and os.path.exists(local_path):
+                return local_path
             local_path = tempfile.mktemp(dir=remote_files_dir())
             error = None
             try:
@@ -73,7 +85,7 @@ def get_file(path: str) -> str:
         else:
             print(
                 message.warn(
-                    "pwndbg.gdblib.file.get(%s) returns local path as we can't download file from QEMU"
+                    "pwndbg.gdblib.file.get_file(%s) returns local path as we can't download file from QEMU"
                     % path
                 )
             )
--- a/pwndbg/commands/elf.py
+++ b/pwndbg/commands/elf.py
@@ -10,7 +10,7 @@ from pwndbg.commands import CommandCategory
 )
 @pwndbg.commands.OnlyWithFile
 def elfsections() -> None:
-    local_path = pwndbg.gdblib.file.get_file(pwndbg.gdblib.proc.exe)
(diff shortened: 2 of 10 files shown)


pwndbg/gdblib/file.py
```diff
@@ -35,22 +35,32 @@ def remote_files_dir():
...
+
+def get_file(path: str, try_local_path: bool = False) -> str:
     """
     Downloads the specified file from the system where the current process is
     being debugged.
 
     If the `path` is prefixed with "target:" the prefix is stripped
     (to support remote target paths properly).
 
+    If the `try_local_path` is set to `True` and the `path` exists locally and "target:" prefix is not present, it will return the local path instead of downloading the file.
```
gsingh93, line 53: When do we *not* want this behavior? Maybe we should either default it to True, or just remove the option and always do this behavior (just thinking out loud, haven't dug into whether this is a bad idea or not).

lebr0nli: I didn't set it to default before we do something like this in many places:
https://github.com/pwndbg/pwndbg/blob/9d22acc1d769d02aae88f9b343d2043437fca6f2/pwndbg/gdblib/net.py#L13

In the above cases, seems like we can't be sure if we need to use the local path or the remote path.

PR #1595, opened 2023-02-25
Add i686 heap alignment heuristic

### LINKED ISSUE #1592: Heap commands not working properly for i386 binary
### Description

The commands `heap` and `vis_heap_chunks` are not working properly for an i386 compiled binary using the latest pwndbg version, however, they are working with an older pwndbg version (2022.08.30). When compiled as x86-64, both commands work with the latest version.

### Steps to reproduce

```c
#include <stdlib.h>

int main(int argc, char *argv[])
{
    char *p1 = (char*)malloc(32);
    free(p1);
	
    return 0;
}
```

`gcc -m32 -g -o main main.c` 

latest pwndbg:
```
pwndbg> start
pwndbg> n
pwndbg> heap
Allocated chunk
Addr: 0x5655a000
Size: 0x00

pwndbg> vis_heap_chunks

pwndbg>
```

pwndbg #2022.08.30:

```
pwndbg> start
pwndbg> n
pwndbg> heap
Allocated chunk | PREV_INUSE
Addr: 0x5655a008
Size: 0x191

Allocated chunk | PREV_INUSE
Addr: 0x5655a198
Size: 0x31

Top chunk | PREV_INUSE
Addr: 0x5655a1c8
Size: 0x21e39

pwndbg> vis_heap_chunks
(...)
0x5655a198      0x00000000      0x00000031      ....1...
0x5655a1a0      0x00000000      0x00000000      ........
0x5655a1a8      0x00000000      0x00000000      ........
0x5655a1b0      0x00000000      0x00000000      ........
0x5655a1b8      0x00000000      0x00000000      ........
0x5655a1c0      0x00000000      0x00000000      ........
0x5655a1c8      0x00000000      0x00021e39      ....9...         <-- Top chunk
pwndbg>
```

When `gcc -g -o main main.c`  is used (`-m32` flag omitted), both commands work on the latest pwndbg version.

### My setup

```
pwndbg> version
Gdb:      12.1
Python:   3.10.7 (main, Nov 24 2022, 19:45:47) [GCC 12.2.0]
Pwndbg:   1.1.1 build: a3dcfaa
Capstone: 4.0.1024
Unicorn:  2.0.1
```

```bash
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 22.10
Release:        22.10
Codename:       kinetic
```

CptGibbon:
A check for the scenario where the first chunk on a heap is not at the beginning of the heap region was accidentally removed in #1346
Should resolve #1592

### DIFF
--- a/pwndbg/heap/ptmalloc.py
+++ b/pwndbg/heap/ptmalloc.py
@@ -393,6 +393,9 @@ class Heap:
                 self._gdbValue = None
 
         self.start = self._memory_region.start
+        # i686 alignment heuristic
+        if Chunk(self.start).size == 0:
+            self.start += pwndbg.gdblib.arch.ptrsize * 2
         self.end = self._memory_region.end
         self.first_chunk = Chunk(self.start)
 


pwndbg/heap/ptmalloc.py
```diff
@@ -393,6 +393,9 @@ def __init__(self, addr, arena=None) -> None:
                 self._gdbValue = None
 
         self.start = self._memory_region.start
+        # i686 alignment heuristic
```
gsingh93, line 396: Any reason to prefer this over `if arch.current == 'i386'`?

CptGibbon: I just revived the original implementation, so I'm not sure if it applies to other architectures. The original author might know if it's appropriate.

gsingh93: In my rewrite (that was reverted), I think I went with explicitly checking if the arch is i386. Otherwise this if-statement may hide bugs on other architectures.

I don't feel too strongly about this, will go with whatever you think is best.

CptGibbon: The original check was written by @matrix1001 and reviewed by @GrosQuildu in 2019.
I believe the consensus was that this heap realignment may occur when there's a discrepancy between compiled & runtime alignment requirements e.g. 32bit binaries running on a 64bit system. So perhaps not unique to i386 🤷

PR #1602, opened 2023-03-02
Hopefully fix the bug when using LD_PRELOAD to load libc

lebr0nli:
The heap heuristics will try to find `libc.so.6` in the output of `info sharedlibrary`, but if we load libc with `LD_PRELOAD`, the filename of the libc might not be `libc.so.6`.

This PR hopefully fixes the bug for the case like https://github.com/pwndbg/pwndbg/issues/1231#issuecomment-1450610516.
(But some of the cases with `LD_PRELOAD` might still fail, we might need a more reliable and also efficient way in the future to find libc's path.)

### DIFF
--- a/pwndbg/glibc.py
+++ b/pwndbg/glibc.py
@@ -68,11 +68,38 @@ def _get_version() -> Optional[Tuple[int, ...]]:
 @pwndbg.lib.memoize.reset_on_start
 @pwndbg.lib.memoize.reset_on_objfile
 def get_libc_filename_from_info_sharedlibrary() -> Optional[str]:
+    """
+    Get the filename of the libc by parsing the output of `info sharedlibrary`.
+    """
+    # Try to parse the output of `info sharedlibrary`:
+    # pwndbg> |info sharedlibrary| grep libc
+    # 0x00007f9ade418700  0x00007f9ade58f47d  Yes         ./libc.so.6
+    # Or:
+    # pwndbg> |info sharedlibrary| grep libc
+    # 0x00007f9ade418700  0x00007f9ade58f47d  Yes (*)     ./libc.so.6
+    possible_libc_path = []
     for line in pwndbg.gdblib.info.sharedlibrary().splitlines()[1:]:
-        filename = line.split(maxsplit=3)[-1].lstrip("(*)").lstrip()
-        # Is it possible that the libc is not called `libc.so.6`?
-        if os.path.basename(filename) == "libc.so.6":
-            return filename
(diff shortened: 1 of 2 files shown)


pwndbg/glibc.py
```diff
@@ -68,11 +68,38 @@ def _get_version() -> Optional[Tuple[int, ...]]:
...
+    # 0x00007f9ade418700  0x00007f9ade58f47d  Yes (*)     ./libc.so.6
+    possible_libc_path = []
     for line in pwndbg.gdblib.info.sharedlibrary().splitlines()[1:]:
-        filename = line.split(maxsplit=3)[-1].lstrip("(*)").lstrip()
-        # Is it possible that the libc is not called `libc.so.6`?
-        if os.path.basename(filename) == "libc.so.6":
-            return filename
+        if line.startswith("("):
+            # footer line:
+            # (*): Shared library is missing debugging information.
```
disconnect3d, line 84: https://github.com/bminor/binutils-gdb/blob/cb8d98ec9251bb67f9b394c63f8c350a5be1a272/gdb/solib.c#L1153-L1161

Seems this is the only message that starts with `(*)`; a potential problem here is that the call in GDB code is`_("...")` and that `_` is probably internalization/translation stuff and the translated string in theory may not start with `(*)` (but I doubt they would change it).

Just saying. I guess we may not be able to do better than that.

disconnect3d:
Can we add some unit tests for this function (which would ideally cover all possible cases we know/see in GDB code :P)?

lebr0nli:
> Can we add some unit tests for this function

Sure, I'll try to add it.

disconnect3d:
@lebr0nli I am wondering if we can do this parsing of shared libs easier; maybe going through `gdb.objfiles()` somehow -- but this does not really provide any info whether given objfile is an actual shared object or not, so maybe not (?).

disconnect3d:
I have also asked the GDB mailing list about this stuff here: https://sourceware.org/pipermail/gdb/2023-March/050619.html

lebr0nli:
> but this does not really provide any info whether given objfile is an actual shared object or not, so maybe not (?).

Yea, I parsed the `gdb.objfiles()` in some function before, but seems the information it provides is still not that enough to determine whether it's a shared object or not so I parsed `info sharedlibrary` directly in the ends.

> I have also asked the GDB mailing list about this stuff here: https://sourceware.org/pipermail/gdb/2023-March/050619.html

Thanks! It will be great if there were some Python APIs that can help :)

PR #1605, opened 2023-03-05
Fixes #1600, #752: invalid symbol.get results

### LINKED ISSUE #1600: `nearpc` fails to show the symbol name after demangling
When I learning OOP for C++, I found out this:

```console
$ cat test.cpp
#include <iostream>

class A{
public:
    A(int _sz=10);
    A(const A& a);
    ~A();
    A& operator=(const A& a);
private:
    char *p;
    int sz;
};

A::A(int _sz): sz(_sz){
    p = new char[sz];
    std::cout << "ctor" << std::endl;
}

A::A(const A &a) {
    p = new char[sz=a.sz];
    std::cout << "copy" << std::endl;
}

A::~A() {
    delete[] p;
    std::cout << "dtor" << std::endl;
}

A &A::operator=(const A &a) {
    if(this != &a){
        p = new char[sz=a.sz];
        for(int i = 0; i < sz; ++i) p[i] = a.p[i];
    }
    std::cout << "assign" << std::endl;
    return *this;
}

int main() {
    A a1;
    A a2 = a1;
    A a3;
    a3 = a2;
    return 0;
}
$ g++ -ggdb test.cpp
$ gdb -q a.out -ex 'set context-sections' -ex 'break main' -ex 'run' -ex 'nearpc main 15'
...
 ► 0x55f188f603fe <main>        endbr64
   0x55f188f60402 <main+4>      push   rbp
   0x55f188f60403 <main+5>      mov    rbp, rsp
   0x55f188f60406 <main+8>      push   rbx
   0x55f188f60407 <main+9>      sub    rsp, 0x48
   0x55f188f6040b <main+13>     mov    rax, qword ptr fs:[0x28]      <main>
   0x55f188f60414 <main+22>     mov    qword ptr [rbp - 0x18], rax
   0x55f188f60418 <main+26>     xor    eax, eax
   0x55f188f6041a <main+28>     lea    rax, [rbp - 0x50]
   0x55f188f6041e <main+32>     mov    esi, 0xa
   0x55f188f60423 <main+37>     mov    rdi, rax
   0x55f188f60426 <main+40>     call   A::A(int)                <A::A(int)>

   0x55f188f6042b <main+45>     lea    rdx, [rbp - 0x50]
   0x55f188f6042f <main+49>     lea    rax, [rbp - 0x40]
   0x55f188f60433 <main+53>     mov    rsi, rdx
   0x55f188f60436 <main+56>     mov    rdi, rax
   0x55f188f60439 <main+59>     call   0x55f188f6028e                <0x55f188f6028e>

   0x55f188f6043e <main+64>     lea    rax, [rbp - 0x30]
   0x55f188f60442 <main+68>     mov    esi, 0xa
   0x55f188f60447 <main+73>     mov    rdi, rax
   0x55f188f6044a <main+76>     call   A::A(int)                <A::A(int)>

   0x55f188f6044f <main+81>     lea    rdx, [rbp - 0x40]
   0x55f188f60453 <main+85>     lea    rax, [rbp - 0x30]
   0x55f188f60457 <main+89>     mov    rsi, rdx
   0x55f188f6045a <main+92>     mov    rdi, rax
   0x55f188f6045d <main+95>     call   0x55f188f6034c                <0x55f188f6034c>

   0x55f188f60462 <main+100>    mov    ebx, 0
   0x55f188f60467 <main+105>    lea    rax, [rbp - 0x30]
   0x55f188f6046b <main+109>    mov    rdi, rax
   0x55f188f6046e <main+112>    call   A::~A()                <A::~A()>

   0x55f188f60473 <main+117>    lea    rax, [rbp - 0x40]
...
```

As you can see, `nearpc` failed to show the symbol name of the copy constructor(at 0x55f188f6028e) and the assignment operator(at 0x55f188f6034c).

It's because the symbol of them after demangling contains spaces and break our logic at here:

https://github.com/pwndbg/pwndbg/blob/e642461941b47fba5f68ada70cb489ede0fa5457/pwndbg/gdblib/symbol.py#L92-L99

disconnect3d:
This commit adds a fix and tests for #1600 and #752.

* https://github.com/pwndbg/pwndbg/issues/1600
* https://github.com/pwndbg/pwndbg/issues/752

Generally, for an example like this:

```cpp
struct A {
    void foo(int, int) { };
};

int main() {
    A a;
    a.foo(1, 1);
}
```

The output for `info symbol <address of A::foo>` returns:

```
'A::foo(int, int) [clone.isra.0] + 3 in section .text of /root/pwndbg/tests/gdb-tests/tests/binaries/a.out\n'
```

We then used this code to parse this:

```py
    # Expected format looks like this:
    # main in section .text of /bin/bash
    # main + 3 in section .text of /bin/bash
    # system + 1 in section .text of /lib/x86_64-linux-gnu/libc.so.6
    # No symbol matches system-1.
    a, b, c, _ = result.split(maxsplit=3)

    if b == "+":
        return "%s+%s" % (a, c)
    if b == "in":
        return a

    return ""
```

The `result.split(maxsplit=3)` here splitted the string to:

```py
['A::foo(int,',
 'int)',
 '[clone.isra.0] + 3 in section .text of /root/pwndbg/tests/gdb-tests/tests/binaries/a.out\n']
```

And since `b` was not `"+"` or `"in"` we eventually returned an empty string instead of the `A::foo(int, int)` which would be expected here.

### DIFF
--- a/pwndbg/gdblib/symbol.py
+++ b/pwndbg/gdblib/symbol.py
@@ -84,19 +84,25 @@ def get(address: int, gdb_only=False) -> str:
                 res = pwndbg.ida.Name(address) or pwndbg.ida.GetFuncOffset(address)
                 return res or ""
 
-    # Expected format looks like this:
-    # main in section .text of /bin/bash
-    # main + 3 in section .text of /bin/bash
-    # system + 1 in section .text of /lib/x86_64-linux-gnu/libc.so.6
-    # No symbol matches system-1.
-    a, b, c, _ = result.split(maxsplit=3)
+    # See https://github.com/bminor/binutils-gdb/blob/d1702fea87aa62dff7de465464097dba63cc8c0f/gdb/printcmd.c#L1594-L1624
+    # The most often encountered formats looks like this:
+    #   "main in section .text of /bin/bash"
+    #   "main + 3 in section .text of /bin/bash"
+    #   "system + 1 in section .text of /lib/x86_64-linux-gnu/libc.so.6"
+    #   "No symbol matches system-1"
+    # But there are some others that we have to account for as well
+    if " in section " in result:
+        loc_string, _ = result.split(" in section ")
+    elif " in load address range of " in result:
+        loc_string, _ = result.split(" in load address range of ")
+    elif " overlay section " in result:
+        result, _ = result.split(" overlay section ")
+        loc_string, _ = result.split(" in ")
+    else:
+        loc_string = ""
 
-    if b == "+":
-        return "%s+%s" % (a, c)
-    if b == "in":
-        return a
-
-    return ""
+    # If there is 'main + 87' we want to replace it with 'main+87' etc.
+    return loc_string.replace(" + ", "+")
 
 
 @pwndbg.lib.memoize.reset_on_objfile
--- a/tests/gdb-tests/tests/binaries/makefile
+++ b/tests/gdb-tests/tests/binaries/makefile
@@ -38,7 +38,7 @@ GLIBC_2_33=$(PWD)/glibcs/2.33
 
 .PHONY : all clean
 
-CUSTOM_TARGETS = reference_bin_pie.out reference_bin_nopie.out
+CUSTOM_TARGETS = reference_bin_pie.out reference_bin_nopie.out symbol_1600_and_752.out
 
 all: $(LINKED) $(LINKED_ASM) $(COMPILED_GO) $(CUSTOM_TARGETS)
 
@@ -113,7 +113,7 @@ issue_1565.out: issue_1565.c
 
 clean :
 	@echo "[+] Cleaning stuff"
-	@rm -f $(COMPILED) $(LINKED) $(COMPILED_ASM) $(LINKED_ASM) $(COMPILED_GO)
+	@rm -f $(COMPILED) $(LINKED) $(COMPILED_ASM) $(LINKED_ASM) $(COMPILED_GO) *.out *.o
 
 
 reference_bin_pie.out: reference-binary.c
@@ -123,3 +123,6 @@ reference_bin_pie.out: reference-binary.c
 reference_bin_nopie.out: reference-binary.c
 	@echo "[+] Building reference_bin_nopie.out"
 	${ZIGCC} -fno-pie -o reference_bin_nopie.out reference-binary.c
+
+symbol_1600_and_752.out: symbol_1600_and_752.cpp
+	${CXX} -O0 -ggdb -Wno-pmf-conversions symbol_1600_and_752.cpp -o symbol_1600_and_752.out
--- a/tests/gdb-tests/tests/binaries/symbol_1600_and_752.cpp
+++ b/tests/gdb-tests/tests/binaries/symbol_1600_and_752.cpp
@@ -1,0 +1,23 @@
+void break_here(void* p) { }
+
+struct A {
+    __attribute__((noinline))
+    void foo(int, int) { break_here(0); }
+
+    void call_foo() { foo(1, 2); }
+};
+
+int main() {
+    break_here((void*)main);
+    break_here((void*)break_here);
+
+    // code for issue 1600
+    break_here((void*)&A::foo);
+
+    // just another check for mangled symbols
+    break_here((void*)&A::call_foo);
+
+    // code for issue 752
+    A a;
+    a.call_foo();
+}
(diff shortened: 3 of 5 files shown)


pwndbg/gdblib/symbol.py
```diff
@@ -84,19 +84,25 @@ def get(address: int, gdb_only=False) -> str:
...
-        return a
+    # See https://github.com/bminor/binutils-gdb/blob/d1702fea87aa62dff7de465464097dba63cc8c0f/gdb/printcmd.c#L1594-L1624
+    # The most often encountered formats looks like this:
+    #   "main in section .text of /bin/bash"
+    #   "main + 3 in section .text of /bin/bash"
+    #   "system + 1 in section .text of /lib/x86_64-linux-gnu/libc.so.6"
+    #   "No symbol matches system-1"
+    # But there are some others that we have to account for as well
+    if " in section " in result:
+        loc_string, _ = result.split(" in section ")
```
lebr0nli, line 95: @disconnect3d 
Somehow, I got an error: `context: too many values to unpack (expected 2) (<class 'ValueError'>` here.
I'm not sure how to reproduce it properly, but here's the binary(a warmup pwn from ACSC 2023) I trigger the error: [bug.zip](https://github.com/pwndbg/pwndbg/files/10896254/bug.zip)

The bug should be reproduced by `python solve.py` on Ubuntu 20.04 after this commit.
It’s caused by the multiple lines output like this:
```
main in section .text of /tmp/vaccine
main in section .text of /tmp/vaccine
```
(It’s really weird tho, not sure why it loads the remote binary and the local file at the same time.

----

Btw, maybe we should use `partition` instead of `split` here if we only want to return the first result?
```console
$ python -m timeit 'loc_string, _, _ = "a in section xxxx".partition("in section")'
2000000 loops, best of 5: 109 nsec per loop
$ python -m timeit 'loc_string, _ = "a in section xxxx".split("in section")'
2000000 loops, best of 5: 128 nsec per loop
```
The performance seems a little bit better. (but I’m not sure how it works behind actually :p )

disconnect3d: @lebr0nli can you check if `info symbol` returns the same if Pwndbg is not loaded? I wonder if that's Pwntools causing this or us.

disconnect3d: It seems this can be Pwntools since `gdb ./vaccine` and `info symbol main` with Pwndbg shows just one symbol.

disconnect3d: re: using `str.partition` - I agree we should probably use it, but I am still wondering how to fix this issue with newlines. We could take whatever first symbol appears in theory.

We could probably also do that with just: `loc_string, *_ = result.partition(...)`?

disconnect3d: @lebr0nli Woah. It seems this double symbol is caused by Pwndbg!

disconnect3d: This is caused by `_add_main_exe_to_symbols`. But this is still a bug: we should probably just first take symbol if multiple of them appears.

disconnect3d: Fwiw this is only happening with the combination with Pwntools. I think Pwntools may do sth like `file ./main` and load the file again this way triggering our hook...

lebr0nli: > Woah. It seems this double symbol is caused by Pwndbg!

Yeah, I can confirm that it only showed one line without pwndbg here.

> We could probably also do that with just: `loc_string, *_ = result.partition(...)`?

Yes, this looks better!

lebr0nli: > Yes, this looks better!

Ahh, wait, it looks better, but maybe doesn't work better xD

```console
$ python -m timeit 'loc_string, _, _ = "a in section xxxx".partition("in section")'
2000000 loops, best of 5: 119 nsec per loop
$ python -m timeit -n 2000000 'loc_string, *_ = "a in section xxxx".partition("in section")'
2000000 loops, best of 5: 228 nsec per loop
```

disconnect3d: Okay, I am wrong. This happens without Pwntools and can be reproduced by:
```
gdbserver :1234 ./vaccine
```
and then:
```
gdb --quiet --ex 'target remote :1234'
```

Then, `info symbol main` shows two symbols.

PR #1611, opened 2023-03-07
Fix #1610: symbol.get with duplicated symbols

### LINKED ISSUE #1610: Make symbol.get bulletproof against duplicated symbols
@lebr0nli pointed out an interesting case with `symbol.get` failing when Pwntools+Pwndbg is used: https://github.com/pwndbg/pwndbg/pull/1605#discussion_r1126105118

The bug actually does not require Pwntools at all and was partially caused by our `_add_main_to_symbols` hook. I am going to remove that hook in https://github.com/pwndbg/pwndbg/pull/1609 if tests passes, since it seems reduntant.

However, that does not really fix the problem as there may always be duplicated symbols such that `info symbols main` would show:
```
pwndbg> info symbol main
main in section .text of /pwndbg/bug/vaccine
main in section .text of /pwndbg/bug/vaccine
```

This can probably be mitigated by doing something like:
```py
loc_string, *_ = result.split(...)
```

instead of:
```py
loc_string, _ = result.split(...)
```

In the `symbol.get` implementation.

This would also mean we would effectively always assume the first symbol is the one we want to use, but I also do not see a good other way to resolve the duplicated symbol issue.

disconnect3d:
See https://github.com/pwndbg/pwndbg/issues/1610

### DIFF
--- a/pwndbg/gdblib/symbol.py
+++ b/pwndbg/gdblib/symbol.py
@@ -84,6 +84,10 @@ def get(address: int, gdb_only=False) -> str:
                 res = pwndbg.ida.Name(address) or pwndbg.ida.GetFuncOffset(address)
                 return res or ""
 
+    # If there are newlines, which means that there are multiple symbols for the address
+    # then use the first one (see also #1610)
+    result = result[: result.index("\n")]
+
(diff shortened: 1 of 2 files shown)


disconnect3d:
@lebr0nli you may want to review or check this :)

lebr0nli (commented):
> @lebr0nli you may want to review or check this :)

LGTM! Thanks for fixing this bug!

PR #1612, opened 2023-03-07
feature: Add headers to `arenas` command output

CptGibbon:
Make the output of the `arenas` command more meaningful by adding a header to each column.
Add an "arena type" column and print `main_arena` in a different color to more easily distinguish it from other arenas.
Add a small arrow to the output to clarify which arena a heap belongs to when an arena is administrating more than one heap.

#### Before:
arenas command original

#### After:
arenas command with headers

### DIFF
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -2,6 +2,7 @@ import argparse
 import ctypes
 
 import gdb
+from tabulate import tabulate
 
 import pwndbg.color.context as C
 import pwndbg.color.memory as M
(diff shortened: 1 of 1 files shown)


disconnect3d:
Should we maybe rename the column file to objfile or mapping or vmmap?

CptGibbon:
> Should we maybe rename the column file to objfile or mapping or vmmap?

Sure, I was just keeping the status quo from:
https://github.com/pwndbg/pwndbg/blob/a847cbba92e6628b330164a53461a65987bfdc72/pwndbg/commands/vmmap.py#L40-L45

PR #1613, opened 2023-03-08
Display bin size range in `largebins` command output

CptGibbon:
Show the chunk size range for bins printed by the `largebins` command.
I think this is more useful than showing only the bin index (current behavior) or just the minimum size (old behavior).
It also follows the same format as the other `*bins` commands, which print size information rather than index.

#### Before:
largebins old behavior
#### After:
largebins with size range

### DIFF
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -82,10 +82,12 @@ def format_bin(bins: Bins, verbose=False, offset=None):
 
         if isinstance(size, int):
             if bins_type == BinType.LARGE:
-                infinity_symbol = "\u221e"  # Unicode "infinity"
                 start_size, end_size = allocator.largebin_size_range_from_index(size)
                 size = hex(start_size) + "-"
-                size += infinity_symbol if end_size == pwndbg.gdblib.arch.ptrmask else hex(end_size)
+                if end_size != pwndbg.gdblib.arch.ptrmask:
+                    size += hex(end_size)
+                else:
+                    size += "\u221e"  # Unicode "infinity"
             else:
                 size = hex(size)
 
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -81,7 +81,13 @@ def format_bin(bins: Bins, verbose=False, offset=None):
             formatted_chain = pwndbg.chain.format(chain_fd[0], offset=offset, safe_linking=safe_lnk)
 
         if isinstance(size, int):
-            size = hex(size)
+            if bins_type == BinType.LARGE:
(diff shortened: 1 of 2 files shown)


pwndbg/commands/heap.py
```diff
@@ -81,7 +81,13 @@ def format_bin(bins: Bins, verbose=False, offset=None):
             formatted_chain = pwndbg.chain.format(chain_fd[0], offset=offset, safe_linking=safe_lnk)
 
         if isinstance(size, int):
-            size = hex(size)
+            if bins_type == BinType.LARGE:
+                infinity_symbol = "\u221e"  # Unicode "infinity"
+                start_size, end_size = allocator.largebin_size_range_from_index(size)
+                size = hex(start_size) + "-"
+                size += infinity_symbol if end_size == pwndbg.gdblib.arch.ptrmask else hex(end_size)
```
disconnect3d: If we put the comment for infinity symbol anyway, we can just hardcode it here instead of storing in a variable

pwndbg/heap/ptmalloc.py
```diff
@@ -658,6 +658,22 @@ def __init__(self) -> None:
...
+    def largebin_size_range_from_index(self, index):
+        index += NSMALLBINS
+        spaces_table = self._spaces_table()
+        largest_largebin = self.largebin_index(pwndbg.gdblib.arch.ptrmask)
+        start_size = (NSMALLBINS * self.malloc_alignment) - self.malloc_alignment
+
+        for i in range(NSMALLBINS, index + 1):
+            start_size += spaces_table[i]
+
+        end_size = start_size + spaces_table[index + 1] - self.malloc_alignment
```
disconnect3d: this could be in an `else` of the if below; or:
```py
if index != largest_largebin:
    end_size = start_size + spaces_table[index + 1] - self.malloc_alignment
else:
    end_size = pwndbg.gdblib.arch.ptrmask
```

disconnect3d:
Hmm should we maybe try to align this somehow so all bins pointers start at the same column? :P

pwndbg/commands/heap.py
CptGibbon: I've split this into a a multi-line `if`/`else`, otherwise the linter makes it harder to read.

pwndbg/heap/ptmalloc.py
CptGibbon: Ah thank you for pointing this out 🙏

CptGibbon:
> Should we maybe try to align this somehow so all bins pointers start at the same column?

I agree that would look better, perhaps a separate PR to address bin formatting in general?

PR #1614, opened 2023-03-08
ai tool now uses ChatGPT model

oblivia-simplex:
I've updated my `ai` command so that it takes advantage of openai's new chat model API by default, which is both 10x cheaper to run and which seems like a better fit for the application. The tool remains backwards compatible with the completion models, like text-davinci-003.


pwndbg/commands/ai.py
```diff
@@ -186,42 +221,72 @@ def build_prompt(question, command=None):
...
+        "model": model,
+        "max_tokens": max_tokens,
+        "messages": prompt,
+        "temperature": temperature,
+    }
+    host = "api.openai.com"
+    path = "/v1/chat/completions"
+    url = f"https://{host}{path}"
+    r = requests.post(
+        url,
```
disconnect3d, line 257:
```suggestion
        f"https://api.openai.com/v1/chat/completions",
```
It seems we don't use the `host`, `path` and `url` variables any further, so they can be deleted and the full url hardcoded here

disconnect3d: (same in other places)

PR #1617, opened 2023-03-09
Add `largebin_index_32_big` macro

### LINKED ISSUE #1615: missing heap macro `largebin_index_32_big()`
The `largebin_index_32_big()` macro is used by GLIBC malloc to compute a largebin index when `MALLOC_ALIGNMENT` is 16 but `SIZE_SZ` isn't 8 (e.g. when running a 32bit program on 64bit architecture).

Without it, pwndbg incorrectly computes largebin indices in such scenarios.

Here's GLIBC's [`largebin_index_32_big()` macro](https://github.com/bminor/glibc/blob/4c721f24fc190d1dc935eb0bab283de7cf13182e/malloc/malloc.c#L1573-L1579).

CptGibbon:
Resolves #1615 by adding GLIBC's `largebin_index_32_big` macro and using it when `pwndbg.gdblib.arch.ptrsize` is 4 but `MALLOC_ALIGNMENT` is 16.

This allows pwndbg to correctly calculate largebin indices under the above circumstances so they start at zero rather than 7.

### DIFF
--- a/pwndbg/heap/ptmalloc.py
+++ b/pwndbg/heap/ptmalloc.py
@@ -1062,6 +1062,25 @@ class GlibcMemoryAllocator(pwndbg.heap.heap.MemoryAllocator):
             else 126
         )
 
+    def largebin_index_32_big(self, sz):
+        """Modeled on the GLIBC malloc largebin_index_32_big macro.
+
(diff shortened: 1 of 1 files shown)


disconnect3d:
Any chance we can have a test for this behavior? 🙏

CptGibbon:
> Any chance we can have a test for this behavior? 🙏

Sure, I'll put something together 👍

PR #1620, opened 2023-03-10
Fix #1394: Improve installation process

### LINKED ISSUE #1394: Running setup.sh on installed Pwndbg should be more clear
We have this at the end of our setup.sh:
https://github.com/pwndbg/pwndbg/blob/f01f932755d7ce88da520ef62dccd5eecae8b0b2/setup.sh#L157-L160

This can be confusing for users when they try to install Pwndbg for the second time but they haven't cleared the ~/.gdbinit file.

I guess we should:
* Check for `source <gdbinit.py>` in `~/.gdbinit` as the very first thing in setup.sh
* If we find an entry, we should ask the user if they want to proceed and override the initializer line
* If the user don't want to proceed, we should stop the installation
* If the user wants to proceed, we should install all deps etc, and add the `source ...` line to `~/.gdbinit` at the end of setup.sh

Cycatz:
Now the installation process contains the following steps: 
1. Check for any line with `source <gdbinit.py>` before installation
2. If any, ask the user if they want to proceed
4. After the installation, comment out old `source <gdbinit.py>` lines if any and add a new `source <gdbinit.py>` line.
 
A check for presence of the line `set debuginfod enabled on` is also added to avoid being appended repeatedly.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -63,11 +63,24 @@ install_emerge() {
 install_pacman() {
     sudo pacman -Syy --noconfirm || true
     sudo pacman -S --noconfirm git gdb python python-pip python-capstone python-unicorn python-pycparser python-psutil python-ptrace python-pyelftools python-six python-pygments which debuginfod
-    echo "set debuginfod enabled on" >> ~/.gdbinit
+    if ! grep -q "^set debuginfod enabled on" ~/.gdbinit; then
+        echo "set debuginfod enabled on" >> ~/.gdbinit
+    fi
 }
 
 PYTHON=''
 INSTALLFLAGS=''
+
+# Check for the presence of the initializer line in the user's ~/.gdbinit file
+if grep -q '^[^#]*source.*pwndbg/gdbinit.py' ~/.gdbinit; then
+    # Ask the user if they want to proceed and override the initializer line
+    read -p "An initializer line was found in your ~/.gdbinit file. Do you want to proceed and override it? (y/n) " answer
+
+    # If the user does not want to proceed, exit the script
+    if [[ "$answer" != "y" ]]; then
+        exit 0
+    fi
+fi
 
 if osx || [ "$1" == "--user" ]; then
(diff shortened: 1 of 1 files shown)


disconnect3d:
Fwiw this needs to be fixed so that sed does not fail 😅

Cycatz:
Why this check failed :(

disconnect3d:
@Cycatz The `setup.sh` starts with `set -ex`: https://github.com/pwndbg/pwndbg/blob/dev/setup.sh#L2

One of these makes it so that if a command fails (returns a status code other than 0) then the whole scripts exits. I think this is what happens here.

You probably need to account for that in your changes.

disconnect3d:
Actually, I am wrong.

The build now failed due to:
```
Zig binary checksum mismatch
Expected: dfc8f5ecb651342f1fc2b2828362b62f74fadac9931bda785b80bf7ecfcfabb2
Actual: 3b57bf83c8cb797254d4c3c515a7e4f76a8001c247c435544d15832b03ae92ff
```
When installing some dev deps. I will try to hit rebuild and see if it fixes things. Other than that, maybe some blob we depend on (likely Zig binary?) was updated and we need to update a checksum somewhere.

Cnly:
I think this broke a few things for me...

* I run `setup.sh` every time I pull changes from this repo, at which point I'd obviously have the `source` line in my ~/.gdbinit. And the new version of the script won't leave it untouched.
* The above is actually run in my non-interactive script, but now that script would stop at the prompt.

How about we extract the dependency installation part from `setup.sh` to a new `update.sh`? Or skip checking ~/.gdbinit if a `DONT_UPDATE_GDBINIT` env var is set? I could do a PR for either.

casasnovas:
,XX,

Le ven. 10 mars 2023, 18:23, Disconnect3d ***@***.***> a
écrit :

disconnect3d:
@Cnly Yeah, there definetely should be a non-interactive way of doing this.

PR #1623, opened 2023-03-12
Correct largebin size lookups on i386

### LINKED ISSUE #1622: Largebin sizes are reported incorrectly on i386
When `pwndbg.gdblib.arch.ptrsize` is 4 but `MALLOC_ALIGNMENT` is 16, the `largebins` command incorrectly reports largebin sizes.

largebins bad

CptGibbon:
This PR replaces pwndbg's largebin size lookup system.
Resolves #1622.

Calculating largebin sizes given a largebin index isn't straightforward, GLIBC has no need to do this so doesn't provide code to do so. Inverting the `largebin_index_*` code gives an incomplete table, which must be fixed up manually per pointer size, which the current `spaces_table` implementation does.

Since this procedure always requires manual intervention, I've replaced the spaces table calculation and fixups with tables of minimum chunk sizes. The original system works, but I think a size table is clearer in its purpose.

In the "before" screenshot below, note how the free chunk sizes don't match the size range of the bins they're in. The bin sizes are also incorrect for an alignment of 16.

#### Before:
largebins bad

#### After:
largebins good

No I'm not a fan of what the Black linter has done to the tables either but what can you do 🤷‍♂️


pwndbg/heap/ptmalloc.py
```diff
@@ -658,17 +658,219 @@ def __init__(self) -> None:
...
+            0x940,
+            0x980,
+            0x9C0,
+            0xA00,
+            0xA40,
+            0xA80,
+            0xAC0,
+            0xB00,
+            0xB40,
+            0xB80,  # Bin 95 is unused, but its size is used to calculate the previous bin's maximum chunk size.
```
disconnect3d: Probably stupid question, but how is this bin 95? :)

pwndbg/heap/ptmalloc.py
```diff
@@ -658,17 +658,219 @@ def __init__(self) -> None:
         # ptmalloc cache for current thread
         self._thread_cache: gdb.Value = None
 
-    def largebin_size_range_from_index(self, index):
-        index += NSMALLBINS
-        spaces_table = self._spaces_table()
-        largest_largebin = self.largebin_index(pwndbg.gdblib.arch.ptrmask)
-        start_size = (NSMALLBINS * self.malloc_alignment) - self.malloc_alignment
+        self.largebin_reverse_lookup_32 = [
```
disconnect3d: Can we add some comments on those? E.g. where are those values taken from etc.

pwndbg/heap/ptmalloc.py
CptGibbon: > Probably stupid question

Not at all.

> how is this bin 95?

This is indeed largebin 31, but GLIBC malloc has no largebins array.
Instead, every doubly-linked bin resides in a single array named `bins`. There are 64 bins preceding largebin 31 (62 smallbins, the unsortedbin and the unused bin 0), so this could technically be considered bin 95.

I've amended the comment to clarify this👍

pwndbg/heap/ptmalloc.py
CptGibbon: I've added a comment briefly explaining how these tables were generated 👍

pwndbg/heap/ptmalloc.py
```diff
@@ -658,17 +658,226 @@ def __init__(self) -> None:
...
-        largest_largebin = self.largebin_index(pwndbg.gdblib.arch.ptrmask)
-        start_size = (NSMALLBINS * self.malloc_alignment) - self.malloc_alignment
+        """
+        Largebin reverse lookup tables.
+        These help determine the range of chunk sizes that each largebin can hold.
+        They were generated by running every chunk size between the minimum & maximum large chunk
+        sizes through largebin_index().
+        Largebin 31 (bin 95) isn't used on i386 when MALLOC_ALIGNMENT is 16, so its value must be added manually.
+        """
+        self.largebin_reverse_lookup_32 = (
```
disconnect3d: Note: perhaps we want to set it as class field so the tuple isn't recreated on each call to this function.

CptGibbon: Of course, that's a great idea. I've made the tables class attributes 👍

PR #1639, opened 2023-03-21
Fix linux arm/aarch64/ppc(64) syscall abi

Xeonacid:
The system call number is specified in r0.
There is a maximum of 6 integer parameters to a syscall, passed in r3-r8.
Ref: https://docs.kernel.org/powerpc/syscall64-abi.html

I guess `SyscallABI` should be used here instead of `ABI`, but I'm not sure.


pwndbg/lib/abi.py
```diff
@@ -77,8 +77,8 @@ class SigreturnABI(SyscallABI):
 linux_arm_syscall = SyscallABI(["r7", "r0", "r1", "r2", "r3", "r4", "r5", "r6"], 4, 0)
 linux_aarch64_syscall = SyscallABI(["x8", "x0", "x1", "x2", "x3", "x4", "x5", "x6"], 16, 0)
```
Xeonacid: Since linux syscall has only 6 arguments, please check if `r6`/`x6` should be removed like `r9` in ppc. If true, I will remove them as well.

disconnect3d: I am not sure what is the best place to check it, but I hope that the table from https://man7.org/linux/man-pages/man2/syscall.2.html is more than less accurate :).


```
       The second table shows the registers used to pass the system call
       arguments.

       Arch/ABI      arg1  arg2  arg3  arg4  arg5  arg6  arg7  Notes
       ──────────────────────────────────────────────────────────────
       alpha         a0    a1    a2    a3    a4    a5    -
       arc           r0    r1    r2    r3    r4    r5    -
       arm/OABI      r0    r1    r2    r3    r4    r5    r6
       arm/EABI      r0    r1    r2    r3    r4    r5    r6
       arm64         x0    x1    x2    x3    x4    x5    -
       blackfin      R0    R1    R2    R3    R4    R5    -
       i386          ebx   ecx   edx   esi   edi   ebp   -
       ia64          out0  out1  out2  out3  out4  out5  -
       m68k          d1    d2    d3    d4    d5    a0    -
       microblaze    r5    r6    r7    r8    r9    r10   -
       mips/o32      a0    a1    a2    a3    -     -     -     1
       mips/n32,64   a0    a1    a2    a3    a4    a5    -
       nios2         r4    r5    r6    r7    r8    r9    -

       parisc        r26   r25   r24   r23   r22   r21   -
       powerpc       r3    r4    r5    r6    r7    r8    r9
       powerpc64     r3    r4    r5    r6    r7    r8    -
       riscv         a0    a1    a2    a3    a4    a5    -
       s390          r2    r3    r4    r5    r6    r7    -
       s390x         r2    r3    r4    r5    r6    r7    -
       superh        r4    r5    r6    r7    r0    r1    r2
       sparc/32      o0    o1    o2    o3    o4    o5    -
       sparc/64      o0    o1    o2    o3    o4    o5    -
       tile          R00   R01   R02   R03   R04   R05   -
       x86-64        rdi   rsi   rdx   r10   r8    r9    -
       x32           rdi   rsi   rdx   r10   r8    r9    -
       xtensa        a6    a3    a4    a5    a8    a9    -

       Notes:

       [1] The mips/o32 system call convention passes arguments 5
           through 8 on the user stack.

       Note that these tables don't cover the entire calling convention—
       some architectures may indiscriminately clobber other registers
       not listed here.
```

From this we should indeed remove x6 from aarch64. However it seems that powerpc itself may use r9, but powerpc64 will not?

pwndbg/lib/abi.py
```diff
@@ -77,8 +77,8 @@ class SigreturnABI(SyscallABI):
...
 linux_mips_syscall = SyscallABI(["$v0", "$a0", "$a1", "$a2", "$a3"], 4, 0)
-linux_ppc_syscall = ABI(["r0", "r3", "r4", "r5", "r6", "r7", "r8", "r9"], 4, 0)
-linux_ppc64_syscall = ABI(["r0", "r3", "r4", "r5", "r6", "r7", "r8", "r9"], 8, 0)
+linux_ppc_syscall = SyscallABI(["r0", "r3", "r4", "r5", "r6", "r7", "r8"], 4, 0)
```
disconnect3d:
```suggestion
linux_ppc_syscall = SyscallABI(["r0", "r3", "r4", "r5", "r6", "r7", "r8", "r9"], 4, 0)
```

Suggesting this according to the syscall table from `man 2 syscall`

pwndbg/lib/abi.py
Xeonacid: Fixed

pwndbg/lib/abi.py
```diff
@@ -74,11 +74,11 @@ class SigreturnABI(SyscallABI):
 
 linux_i386_syscall = SyscallABI(["eax", "ebx", "ecx", "edx", "esi", "edi", "ebp"], 4, 0)
 linux_amd64_syscall = SyscallABI(["rax", "rdi", "rsi", "rdx", "r10", "r8", "r9"], 8, 0)
-linux_arm_syscall = SyscallABI(["r7", "r0", "r1", "r2", "r3", "r4", "r5", "r6"], 4, 0)
-linux_aarch64_syscall = SyscallABI(["x8", "x0", "x1", "x2", "x3", "x4", "x5", "x6"], 16, 0)
+linux_arm_syscall = SyscallABI(["r7", "r0", "r1", "r2", "r3", "r4", "r5"], 4, 0)
```
disconnect3d:
```suggestion
linux_arm_syscall = SyscallABI(["r7", "r0", "r1", "r2", "r3", "r4", "r5", "r6"], 4, 0)
```
It seems that arm may still use r6 :)
```
       arm/OABI      r0    r1    r2    r3    r4    r5    r6
       arm/EABI      r0    r1    r2    r3    r4    r5    r6
       arm64         x0    x1    x2    x3    x4    x5    -
```

PR #1640, opened 2023-03-21
Resolve some `find_fake_fast` command issues

### LINKED ISSUE #1363: The "find_fake_fast" command fails to find valid fake chunks on the x64 arch.
I was doing a random task (fastbin dup) and I noticed that this function is not taking into account that the fastbin size field check, in the latest GLIBC, is done by this macro:

```
/* offset 2 to use otherwise unindexable first 2 bins */
#define fastbin_index(sz) \
((((unsigned int) (sz)) >> (SIZE_SZ == 8 ? 4 : 3)) - 2)
```
( It is an old issue: https://sourceware.org/bugzilla/show_bug.cgi?id=19449 )

On x64 the macro is casting from the long to the unsigned int (4 most-significant bytes don't matter), and the `find_fake_fast` is working on 8 bytes (instead of 4 least-significant bytes), therefore, it is missing a lot of valid "fake chunks". 

The quick (and probably bad) workaround (additional bitwise and on the value) that I've done in the `pwndbg/pwndbg/lib/heap/helpers.py`:
```
import struct

import pwndbg.gdblib.arch


def find_fastbin_size(mem: bytes, max_size: int, step: int):
    psize = pwndbg.gdblib.arch.ptrsize
    min_fast = 4 * psize
    fmt = {"little": "<", "big": ">"}[pwndbg.gdblib.arch.endian] + {4: "I", 8: "Q"}[psize]

    for i in range(0, len(mem), step):
        candidate = mem[i : i + psize]
        if len(candidate) == psize:
            value = struct.unpack(fmt, candidate)[0]

            # Clear any flags
            value &= ~0xF
            
            # Fastbin size check is casting size to the unsigned int from the unsigned long
            value &= ~0xffffffff00000000 

            if value < min_fast:
                continue

            # The value must be less than or equal to the max size we're looking
            # for, but still be able to reach the target address
            if value <= max_size and i + value >= max_size:
                yield i - psize

```

**Examples:**

Before:

```
pwndbg> find_fake_fast 0x601048 0x7f
Searching for fastbin sizes up to 0x70 starting at 0x600fe0 resulting in an overlap of 0x601048
FAKE CHUNKS
Fake chunk | Allocated chunk | PREV_INUSE | IS_MMAPED | NON_MAIN_ARENA
Addr: 0x600fed
prev_size: 0xf93ee58f5000007f
size: 0x7f
fd: 0x600e00000000
bk: 0xf93f419190000000
fd_nextsize: 0xf93f205fc000007f
bk_nextsize: 0xf93eebaa7000007f

pwndbg> dq 0x600fed
0000000000600fed     f93ee58f5000007f 000000000000007f
0000000000600ffd     0000600e00000000 f93f419190000000
000000000060100d     f93f205fc000007f f93eebaa7000007f
000000000060101d     000040066600007f f93ee87b60000000
```

After:

```
pwndbg> find_fake_fast 0x601048 0x7f
Searching for fastbin sizes up to 0x70 starting at 0x600fe0 resulting in an overlap of 0x601048
FAKE CHUNKS
Fake chunk | Allocated chunk | PREV_INUSE | IS_MMAPED | NON_MAIN_ARENA
Addr: 0x600fed
prev_size: 0xf93ee58f5000007f
size: 0x7f
fd: 0x600e00000000
bk: 0xf93f419190000000
fd_nextsize: 0xf93f205fc000007f
bk_nextsize: 0xf93eebaa7000007f

Fake chunk | Allocated chunk
Addr: 0x600ffa
prev_size: 0xe00000000000000
size: 0x9190000000000060
fd: 0x5fc000007ff93f41
bk: 0xaa7000007ff93f20
fd_nextsize: 0x66600007ff93eeb
bk_nextsize: 0x7b60000000000040

Fake chunk | Allocated chunk
Addr: 0x60101a
prev_size: 0x66600007ff93eeb
size: 0x7b60000000000040
fd: 0xe01000007ff93ee8
bk: 0xa50000007ff93ef1
fd_nextsize: 0x733000007ff93eeb
bk_nextsize: 0x151000007ff93eea

Fake chunk | Allocated chunk | PREV_INUSE | IS_MMAPED | NON_MAIN_ARENA
Addr: 0x60102d
prev_size: 0xf93ef1e01000007f
size: 0xf93eeba50000007f
fd: 0xf93eea733000007f
bk: 0xf93ee7151000007f
fd_nextsize: 0x4006c600007f
bk_nextsize: 0x00

pwndbg> dq 0x60102d
000000000060102d     f93ef1e01000007f f93eeba50000007f <--- perfectly valid chunk of the size 0x70
000000000060103d     f93eea733000007f f93ee7151000007f
000000000060104d     00004006c600007f 0000000000000000
000000000060105d     0000000000000000 0000000000000000

pwndbg> dq 0x60101a
000000000060101a     066600007ff93eeb 7b60000000000040 <--- perfectly valid chunk of the size 0x40
000000000060102a     e01000007ff93ee8 a50000007ff93ef1
000000000060103a     733000007ff93eeb 151000007ff93eea
000000000060104a     06c600007ff93ee7 0000000000000040
```

### LINKED ISSUE #1398: `find_fake_fast` doesn't handle unmapped memory
The `find_fake_fast` command raises `gdb.MemoryError` when any part of the region it searches for candidate fake size fields is unmapped. It is unable to find candidates near the start of virtual memory maps because of this.

As far as I can tell this isn't a regression, the command has always behaved like this.

One solution could be to enclose the command's memory reads in a `try`/`except` block, then simply ignore any failed reads.

### LINKED ISSUE #1631: `find_fake_fast` command misses some candidates when the `--align` option is used
The screen below shows a fake fastbin size field appropriately aligned for the target architecture (size 0x81 at address 0x1fe6268).

`find_fake_fast` correctly identifies this as a candidate fake chunk with the `--align`/`-a` flag when an aligned target address is used, but misses it with a misaligned target address.

bad find_fake_fast

Note that in the misaligned case `find_fake_fast` is searching for smaller chunks (0x70 vs. 0x80, as shown in the pink text).

This is because the `size` variable in `find_fake_fast` is used as both the amount of memory to search (shown below)...

https://github.com/pwndbg/pwndbg/blob/b6e742ce38b68dda29b94a45842c4730e4682053/pwndbg/commands/heap.py#L706-L708
https://github.com/pwndbg/pwndbg/blob/b6e742ce38b68dda29b94a45842c4730e4682053/pwndbg/commands/heap.py#L723

... and as the maximum size field value to search for in `find_fastbin_size()` (called by `find_fake_fast`):

https://github.com/pwndbg/pwndbg/blob/b6e742ce38b68dda29b94a45842c4730e4682053/pwndbg/commands/heap.py#L727
https://github.com/pwndbg/pwndbg/blob/b6e742ce38b68dda29b94a45842c4730e4682053/pwndbg/lib/heap/helpers.py#L4
https://github.com/pwndbg/pwndbg/blob/b6e742ce38b68dda29b94a45842c4730e4682053/pwndbg/lib/heap/helpers.py#L19-L21

One solution could be to separate the amount of memory to search (which may change during the alignment procedure) and the size value to search for (which shouldn't change) into two variables.

### LINKED ISSUE #1637: `find_fake_fast` command can lose size information when clearing flags
While searching for fake chunk size fields, the `find_fake_fast` command clears the entire lower nybble of the candidate size field when attempting to only clear its flags:

https://github.com/pwndbg/pwndbg/blob/cf63cc326e0188cced115dd9e3f76452dc9e156d/pwndbg/lib/heap/helpers.py#L13-L14

On systems where `MALLOC_ALIGNMENT` is 8 bytes or less, this clears the least-significant size bit of the candidate size field, leading to false negatives.

One solution could be to clear flags using `pwndbg.constants.ptmalloc.SIZE_BITS`.

CptGibbon:
This PR leaves most of the `find_fake_fast` command logic intact, whilst addressing the following issues:

Resolve #1363 by introducing a `--glibc-fastbin-bug` option to the command, which changes the width of the candidate size field to search for.

Resolve #1398 by checking if a search region is readable, if not then advancing to the next page and trying again.

Resolve #1631 by separating the amount of memory to search and the maximum size of the candidate field into different variables.

Resolve #1637 by masking chunk sizes with `allocator.malloc_align_mask` rather than a hardcoded value.

### DIFF
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -638,9 +638,13 @@ def tcachebins(addr=None, verbose=False) -> None:
 parser = argparse.ArgumentParser(
     description="Find candidate fake fast or tcache chunks overlapping the specified address."
 )
-parser.add_argument("addr", type=int, help="Address of the word-sized value to overlap.")
+parser.add_argument("target_address", type=int, help="Address of the word-sized value to overlap.")
 parser.add_argument(
-    "size", nargs="?", type=int, default=None, help="Maximum size of fake chunks to find."
+    "max_candidate_size",
+    nargs="?",
+    type=int,
+    default=None,
+    help="Maximum size of fake chunks to find.",
 )
 parser.add_argument(
     "--align",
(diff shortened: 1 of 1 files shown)


gsingh93:
I was a fan of all the comments, as I found the logic of this code to be a bit tricky 😔

CptGibbon:
@gsingh93 I could use your help merging the tail end of this code back into `pwndbg.lib.heap.helpers`

I'd like to replace:
https://github.com/pwndbg/pwndbg/blob/66aa6af19c5435d98281c7a02f71014355092aa6/pwndbg/lib/heap/helpers.py#L6 with
`pwndbg.heap.current.min_chunk_size`

And replace:
https://github.com/pwndbg/pwndbg/blob/66aa6af19c5435d98281c7a02f71014355092aa6/pwndbg/lib/heap/helpers.py#L14 with
`value &= ~(allocator.malloc_align_mask)`

CptGibbon:
> I was a fan of all the comments, as I found the logic of this code to be a bit tricky 😔

Fear not, the comments will return.
I just want to get the structure of the code sorted first.

CptGibbon:
@gsingh93 do the tests in pwndbg/tests/unit-tests/ run on the CI?

I ask because they're commented out in tests.sh
https://github.com/pwndbg/pwndbg/blob/e744ebf3fb774410556da96b431f138007805399/tests.sh#L13-L14

And I can't get the tests in pwndbg/tests/unit-tests/test_find_fastbin_size.py to run without error locally.
What am I doing wrong?

gsingh93:
> do the tests in pwndbg/tests/unit-tests/ run on the CI?
> I can't get the tests in pwndbg/tests/unit-tests/test_find_fastbin_size.py to run without error locally.

A while ago there was an issue with the test infra that caused an exception that was breaking the tests, so I disabled it until I could fix it. I can prioritize fixing the error and re-enabling it, there was only a single test in there (which had good coverage by the other gdb-tests) so it wasn't a priority before.

CptGibbon:
> I disabled it until I could fix it. I can prioritize fixing the error and re-enabling it, there was only a single test in there (which had good coverage by the other gdb-tests) so it wasn't a priority before.

It's cool I know you're busy.
I rewrote this PR to work with your testing methodology because I really like the modularity and testability of it, but then realized the existing tests didn't work, so I've held off.

I'm keen to take this one out of draft and merge because it wraps up 4 issues, and I think the gdb-based tests provide okay-ish coverage for now. I've spent more of this week than I wanted thinking about ways to decouple as much of the heap code as possible for the sake of testing without a gdb session, using your example as a starting point. I'm one more wasted weekend away from reaching out to Dave Farley for advice 😅

If you're happy I can pass the remaining lint tests and this could get merged as is, then we can approach the testing as a separate body of work. What do you think?

disconnect3d:
> If you're happy I can pass the remaining lint tests and this could get merged as is, then we can approach the testing as a separate body of work. What do you think?

This looks good to me so we can merge it now and add tests with another PR[s] :)

PR #1644, opened 2023-03-23
Fixed a crash that caused formatting failure in readlink when qemu returns None as the pid.

Squirre17:
All connections will failed in psutil.process_iter() with exception AccessDenied when qemu is launched with privileged. (e.g. `sudo chroot ./qemu ...`)
It will cause subsequent crash cuz this function doesn't return anything. (i.e. return a None)
https://github.com/pwndbg/pwndbg/blob/59e759e0363cd474e662f28503f8de726baf7d20/pwndbg/gdblib/qemu.py#L76-L99

The function format_args will use pid as a int to format a string. It will cause a crash here in line 224.
https://github.com/pwndbg/pwndbg/blob/59e759e0363cd474e662f28503f8de726baf7d20/pwndbg/arguments.py#L216-L229

### DIFF
--- a/pwndbg/arguments.py
+++ b/pwndbg/arguments.py
@@ -231,7 +231,7 @@ def format_args(instruction):
             else:
                 print(
                     message.hint(
-                        "Cannot find the PID of the program, maybe there is no permission to connect to the privileged QEMU."
+                        "Cannot find PID of the QEMU program: perhaps it is in a different pid namespace or we have no permission to read the QEMU process' /proc/$pid/fd/$fd file."
                     )
                 )
 
--- a/pwndbg/arguments.py
+++ b/pwndbg/arguments.py
@@ -19,6 +19,7 @@ import pwndbg.lib.abi
 import pwndbg.lib.funcparser
 import pwndbg.lib.functions
 from pwndbg.gdblib.nearpc import c as N
+from pwndbg.color import message
 
 ida_replacements = {
     "__int64": "signed long long int",
@@ -221,9 +222,12 @@ def format_args(instruction):
 
         # Enhance args display
         if arg.name == "fd" and isinstance(value, int):
-            path = pwndbg.gdblib.file.readlink("/proc/%d/fd/%d" % (pwndbg.gdblib.proc.pid, value))
-            if path:
-                pretty += " (%s)" % path
+            if pwndbg.gdblib.proc.pid is not None:
+                path = pwndbg.gdblib.file.readlink("/proc/%d/fd/%d" % (pwndbg.gdblib.proc.pid, value))
+                if path:
+                    pretty += " (%s)" % path
+            else:
+                print(message.hint("Cannot find the PID of the program, maybe there is no permission to connect to the privileged QEMU."))
 
         result.append("%-10s %s" % (N.argument(arg.name) + ":", pretty))
     return result
--- a/pwndbg/arguments.py
+++ b/pwndbg/arguments.py
@@ -222,9 +222,10 @@ def format_args(instruction):
 
         # Enhance args display
         if arg.name == "fd" and isinstance(value, int):
-            if pwndbg.gdblib.proc.pid is not None:
+            pid = pwndbg.gdblib.proc.pid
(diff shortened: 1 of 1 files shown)


pwndbg/arguments.py
```diff
@@ -221,9 +222,18 @@ def format_args(instruction):
...
+            if pwndbg.gdblib.proc.pid is not None:
+                path = pwndbg.gdblib.file.readlink(
+                    "/proc/%d/fd/%d" % (pwndbg.gdblib.proc.pid, value)
+                )
+                if path:
+                    pretty += " (%s)" % path
+            else:
+                print(
+                    message.hint(
+                        "Cannot find the PID of the program, maybe there is no permission to connect to the privileged QEMU."
```
disconnect3d:
```suggestion
                        "Cannot find PID of the QEMU program: perhaps it is in a different pid namespace or we have no permission to read the QEMU process' /proc/$pid/fd/$fd file."
```

pwndbg/arguments.py
```diff
@@ -221,9 +222,18 @@ def format_args(instruction):
...
-                pretty += " (%s)" % path
+            if pwndbg.gdblib.proc.pid is not None:
+                path = pwndbg.gdblib.file.readlink(
+                    "/proc/%d/fd/%d" % (pwndbg.gdblib.proc.pid, value)
+                )
+                if path:
+                    pretty += " (%s)" % path
+            else:
+                print(
+                    message.hint(
```
disconnect3d: Hmm how does this hint message work? When/where does it appear? I mean, we are in a format_args call, maybe we shouldn't log anything at all and just not enhance the path display?

disconnect3d: Or: we could display it once

pwndbg/arguments.py
```diff
@@ -221,9 +222,18 @@ def format_args(instruction):
 
         # Enhance args display
         if arg.name == "fd" and isinstance(value, int):
-            path = pwndbg.gdblib.file.readlink("/proc/%d/fd/%d" % (pwndbg.gdblib.proc.pid, value))
-            if path:
-                pretty += " (%s)" % path
+            if pwndbg.gdblib.proc.pid is not None:
+                path = pwndbg.gdblib.file.readlink(
+                    "/proc/%d/fd/%d" % (pwndbg.gdblib.proc.pid, value)
```
disconnect3d: Not a big deal, but maybe we should assign the pid to a variable here since its a getter iirc; even if its cached

pwndbg/arguments.py
Squirre17: > Hmm how does this hint message work? When/where does it appear? I mean, we are in a format_args call, maybe we shouldn't log anything at all and just not enhance the path display?

The hint message will display above the context output. Although displaying it below the context would be best, that may be somewhat difficult to implement.
image

disconnect3d: > Although displaying it below the context would be best, that may be somewhat difficult to implement.

We could probably set some global variable and use it later on, but yeah, meh.

If I think more about it, now, we are displaying the warning but the user has no idea why we do it, do they?

We should either describe it better, saying that "Cannot show the file from fd=5 because we could not find the PID ...", or, remove the warning completely and just not enhance the display.

I think the second one is probably better?

Squirre17: > > Although displaying it below the context would be best, that may be somewhat difficult to implement.
> 
> We could probably set some global variable and use it later on, but yeah, meh.
> 
> If I think more about it, now, we are displaying the warning but the user has no idea why we do it, do they?
> 
> We should either describe it better, saying that "Cannot show the file from fd=5 because we could not find the PID ...", or, remove the warning completely and just not enhance the display.
> 
> I think the second one is probably better?

right, I think so.

disconnect3d: Ping, can you remove this warning message then and maybe turn it into a comment?

Squirre17: > Ping, can you remove this warning message then and maybe turn it into a comment?

sure, I going to add the comment.

PR #1646, opened 2023-03-24
Fix wrong `MALLOC_ALIGNMENT` for powerpc

### LINKED ISSUE #1645: `MALLOC_ALIGNMENT` might be wrong for powerpc
`MALLOC_ALIGNMENT`'s definition:
https://elixir.bootlin.com/glibc/glibc-2.37/source/sysdeps/generic/malloc-alignment.h#L27

How we implement it:
https://github.com/pwndbg/pwndbg/blob/59e759e0363cd474e662f28503f8de726baf7d20/pwndbg/heap/ptmalloc.py#L966-L974

We assume if the architecture isn't i386 and with GLIBC version >= 2.26, then `MALLOC_ALIGNMENT == 2 * SIZE_SZ`, which means we think `2 * SIZE_SZ < __alignof__ (long double)` is always `false`.

But this is not always correct for every architecture, for example, if our architecture is `powerpc:common`:
```python
In [1]: pwndbg.gdblib.typeinfo.lookup_types("long double").alignof
Out[1]: 16

In [2]: pwndbg.gdblib.arch.ptrsize * 2 < pwndbg.gdblib.typeinfo.lookup_types("long double").alignof
Out[2]: True
```

`MALLOC_ALIGNMENT` should be `16`(`__alignof__ (long double)`) instead of `8`(`2 * SIZE_SZ`) in this case.

lebr0nli:
This PR aims to fix #1645

### DIFF
--- a/pwndbg/heap/ptmalloc.py
+++ b/pwndbg/heap/ptmalloc.py
@@ -965,12 +965,14 @@ class GlibcMemoryAllocator(pwndbg.heap.heap.MemoryAllocator):
     @pwndbg.lib.memoize.reset_on_objfile
     def malloc_alignment(self):
         """Corresponds to MALLOC_ALIGNMENT in glibc malloc.c"""
-        # i386 will override it to 16 when GLIBC version >= 2.26
-        # See https://elixir.bootlin.com/glibc/glibc-2.26/source/sysdeps/i386/malloc-alignment.h#L22
+        if pwndbg.gdblib.arch.current == "i386" and pwndbg.glibc.get_version() >= (2, 26):
+            # i386 will override it to 16 when GLIBC version >= 2.26
+            # See https://elixir.bootlin.com/glibc/glibc-2.26/source/sysdeps/i386/malloc-alignment.h#L22
+            return 16
+        # See https://elixir.bootlin.com/glibc/glibc-2.37/source/sysdeps/generic/malloc-alignment.h#L27
+        long_double_alignment = pwndbg.gdblib.typeinfo.lookup_types("long double").alignof
         return (
-            16
-            if pwndbg.gdblib.arch.current == "i386" and pwndbg.glibc.get_version() >= (2, 26)
-            else pwndbg.gdblib.arch.ptrsize * 2
+            long_double_alignment if 2 * self.size_sz < long_double_alignment else 2 * self.size_sz
         )
 
     @property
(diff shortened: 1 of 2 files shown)


pwndbg/heap/ptmalloc.py
```diff
@@ -965,12 +965,21 @@ def malloc_par(self):
...
     def malloc_alignment(self):
         """Corresponds to MALLOC_ALIGNMENT in glibc malloc.c"""
-        # i386 will override it to 16 when GLIBC version >= 2.26
-        # See https://elixir.bootlin.com/glibc/glibc-2.26/source/sysdeps/i386/malloc-alignment.h#L22
+        if pwndbg.gdblib.arch.current == "i386" and pwndbg.glibc.get_version() >= (2, 26):
+            # i386 will override it to 16 when GLIBC version >= 2.26
+            # See https://elixir.bootlin.com/glibc/glibc-2.26/source/sysdeps/i386/malloc-alignment.h#L22
+            return 16
+        # See https://elixir.bootlin.com/glibc/glibc-2.37/source/sysdeps/generic/malloc-alignment.h#L27
+        if hasattr(gdb.Type, "alignof"):
```
lebr0nli, line 973: This part is really dirty for me, we need to split the code just for old GDB versions again :/
Is there any other way we can get the `__alignof__(long double)` result without GDB API?

disconnect3d: There is this special function/operator handled by GDB:
```
pwndbg> print _Alignof(long double)
$1 = 16
```
Or when `set lang c++` is set:
```
pwndbg> print alignof(long double)
$2 = 16
```

Can you check if it works for GDB older than 8.2?

I am afraid it may not be though, since [this thread aimed 8.2 milestone :/](https://gdb-prs.sourceware.narkive.com/KUwdtcc0/bug-cli-17095-new-print-alignof-type-or-expression). Here is the [patch](https://sourceware.org/legacy-ml/gdb-patches/2018-04/msg00469.html) when they added it.

disconnect3d: Btw also when we can't access something in Python in old GDB maybe we can fetch a C function from GDB binary and use it instead if this function is public/exported by GDB (so: not a `static` function...).

This can be done by loading the binary like you would load a shared object, with the `obj = ctypes.CDLL("")` and then using `obj.function(<args>)`, but I am not 100% sure if it would work here. From my quick tests it does not seem to work, e.g., for the `type_align` function :(.

PR #1649, opened 2023-03-24
Add some basic heap heuristics supports for riscv and powerpc

lebr0nli:
Some compilers for some architectures(e.g. riscv and powerpc) won't place the address of `main_arean` in the "next" field at compile time:
```console
$ gdb -q /usr/riscv64-linux-gnu/lib/libc-2.31.so
pwndbg> add-symbol-file /usr/riscv64-linux-gnu/lib/debug/lib/riscv64-linux-gnu/libc-2.31.so
pwndbg> print main_arena
$1 = {
  mutex = 0,
  flags = 0,
  have_fastchunks = 0,
  fastbinsY = {0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0},
  top = 0x0,
  last_remainder = 0x0,
  bins = {0x0 <repeats 254 times>},
  binmap = {0, 0, 0, 0},
  next = 0x0,
  next_free = 0x0,
  attached_threads = 1,
  system_mem = 0,
  max_system_mem = 0
}
```
So our heap heuristics for finding `main_arena` won’t work for that case since we assumed the "next" field is already `&main_arena` after compiled.

This PR aims to add some basic support for that kind of architecture by ~combining the dynamic brute force approach in the old version~ parsing the relocation section.
~(But seems like the brute force will only be needed when we are debugging a binary with statically linked libc and has many user-defined structs tho, so it should be reliable and fast enough for most of the cases.)~

> Note: The heuristics for powerpc will only work after we applied the fix in #1646 and #1648.

### DIFF
--- a/pwndbg/heap/ptmalloc.py
+++ b/pwndbg/heap/ptmalloc.py
@@ -1531,7 +1531,7 @@ class HeuristicHeap(GlibcMemoryAllocator):
                 # };
                 expected = self.malloc_state._c_struct()
                 expected.attached_threads = 1
-                next_field_offset = self.malloc_state(0).get_field_address("next")
+                next_field_offset = self.malloc_state.get_field_offset("next")
                 malloc_state_size = self.malloc_state.sizeof
 
                 # Since RELR relocations might also have .rela.dyn section, we check it first
--- a/pwndbg/heap/ptmalloc.py
+++ b/pwndbg/heap/ptmalloc.py
@@ -1514,30 +1514,98 @@ class HeuristicHeap(GlibcMemoryAllocator):
 
         if not self._main_arena_addr:
             if self.is_statically_linked():
-                section = pwndbg.gdblib.proc.dump_elf_data_section()
-                section_address = pwndbg.gdblib.proc.get_data_section_address()
+                data_section = pwndbg.gdblib.proc.dump_elf_data_section()
+                data_section_address = pwndbg.gdblib.proc.get_data_section_address()
             else:
-                section = pwndbg.glibc.dump_elf_data_section()
-                section_address = pwndbg.glibc.get_data_section_address()
-            if section and section_address:
-                data_section_offset, size, data = section
+                data_section = pwndbg.glibc.dump_elf_data_section()
+                data_section_address = pwndbg.glibc.get_data_section_address()
+            if data_section and data_section_address:
+                data_section_offset, size, data_section_data = data_section
+                # Try to find the default main_arena struct in the .data section
+                # https://github.com/bminor/glibc/blob/glibc-2.37/malloc/malloc.c#L1902-L1907
+                # static struct malloc_state main_arena =
+                # {
+                #   .mutex = _LIBC_LOCK_INITIALIZER,
+                #   .next = &main_arena,
+                #   .attached_threads = 1
+                # };
+                expected = self.malloc_state._c_struct()
+                expected.attached_threads = 1
+                next_field_offset = self.malloc_state(0).get_field_address("next")
+                malloc_state_size = self.malloc_state.sizeof
 
-                # try to find the default main_arena struct in the .data section
-                for i in range(size - self.malloc_state.sizeof):
-                    # https://github.com/bminor/glibc/blob/glibc-2.37/malloc/malloc.c#L1902-L1907
-                    # static struct malloc_state main_arena =
-                    # {
-                    #   .mutex = _LIBC_LOCK_INITIALIZER,
-                    #   .next = &main_arena,
-                    #   .attached_threads = 1
-                    # };
-                    expected = self.malloc_state._c_struct()
-                    expected.next = data_section_offset + i
-                    expected.attached_threads = 1
-                    expected = bytes(expected)
-                    if expected == data[i : i + len(expected)]:
-                        self._main_arena_addr = section_address + i
+                # Since RELR relocations might also have .rela.dyn section, we check it first
+                for section_name in (".relr.dyn", ".rela.dyn", ".rel.dyn"):
+                    if self._main_arena_addr:
+                        # If we have found the main_arena, we can stop searching
                         break
+
+                    if self.is_statically_linked():
+                        relocations = pwndbg.gdblib.proc.dump_relocations_by_section_name(
+                            section_name
+                        )
+                    else:
+                        relocations = pwndbg.glibc.dump_relocations_by_section_name(section_name)
+                    if not relocations:
+                        continue
+
+                    for relocation in relocations:
+                        r_offset = relocation.entry.r_offset
+
+                        # We only care about the relocation in .data section
+                        if r_offset - next_field_offset < data_section_offset:
+                            continue
+                        elif r_offset - next_field_offset >= data_section_offset + size:
+                            break
+
+                        # To find addend:
+                        # .relr.dyn and .rel.dyn need to read the data from r_offset
+                        # .rela.dyn has the addend in the entry
+                        if section_name != ".rela.dyn":
+                            addend = int.from_bytes(
+                                data_section_data[
+                                    r_offset
+                                    - data_section_offset : r_offset
+                                    - data_section_offset
+                                    + pwndbg.gdblib.arch.ptrsize
+                                ],
+                                pwndbg.gdblib.arch.endian,
+                            )
+                        else:
+                            addend = relocation.entry.r_addend
+
+                        # If addend is the offset of main_arena, then r_offset should be the offset of main_arena.next
+                        if r_offset - next_field_offset == addend:
+                            # Check if we can construct the default main_arena struct we expect
+                            tmp = data_section_data[
+                                addend
+                                - data_section_offset : addend
+                                - data_section_offset
+                                + malloc_state_size
+                            ]
+                            # Note: Although RELA relocations have r_addend, some compiler will still put the addend in the location of r_offset, so we still need to check both cases
+                            found = False
+                            expected.next = addend
+                            found |= bytes(expected) == tmp
+                            if not found:
+                                expected.next = 0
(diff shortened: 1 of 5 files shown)


Xeonacid:
> Some compilers for some architectures(e.g. riscv and powerpc) won't place the address of `main_arean` in the "next" field at compile time:
> 
> ```
> $ gdb -q /usr/riscv64-linux-gnu/lib/libc-2.31.so
> pwndbg> add-symbol-file /usr/riscv64-linux-gnu/lib/debug/lib/riscv64-linux-gnu/libc-2.31.so
> pwndbg> print main_arena
> $1 = {
>   mutex = 0,
>   flags = 0,
>   have_fastchunks = 0,
>   fastbinsY = {0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0},
>   top = 0x0,
>   last_remainder = 0x0,
>   bins = {0x0 <repeats 254 times>},
>   binmap = {0, 0, 0, 0},
>   next = 0x0,
>   next_free = 0x0,
>   attached_threads = 1,
>   system_mem = 0,
>   max_system_mem = 0
> }
> ```

Hi, If I understand correctly, it seems the program is not running?
FYI, It works fine on my machine:
```text
pwndbg> p main_arena 
$1 = {
  mutex = 0,
  flags = 0,
  have_fastchunks = 0,
  fastbinsY = {0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0},
  top = 0x0,
  last_remainder = 0x0,
  bins = {0x0 <repeats 254 times>},
  binmap = {0, 0, 0, 0},
  next = 0x3ff7fbb930 <main_arena>,
  next_free = 0x0,
  attached_threads = 1,
  system_mem = 0,
  max_system_mem = 0
}
pwndbg> vmmap
             Start                End Perm     Size Offset File
      0x2aaaaaa000       0x2aaaaae000 r-xp     4000      0 /usr/bin/true
      0x2aaaaae000       0x2aaaaaf000 r--p     1000   3000 /usr/bin/true
      0x2aaaaaf000       0x2aaaab0000 rw-p     1000   4000 /usr/bin/true
      0x3ff7e73000       0x3ff7fb8000 r-xp   145000      0 /usr/lib/libc.so.6
      0x3ff7fb8000       0x3ff7fbb000 r--p     3000 145000 /usr/lib/libc.so.6
      0x3ff7fbb000       0x3ff7fbd000 rw-p     2000 148000 /usr/lib/libc.so.6
      0x3ff7fbd000       0x3ff7fcc000 rw-p     f000      0 [anon_3ff7fbd]
      0x3ff7fda000       0x3ff7fdc000 r--p     2000      0 [vvar]
      0x3ff7fdc000       0x3ff7fde000 r-xp     2000      0 [vdso]
      0x3ff7fde000       0x3ff7ffd000 r-xp    1f000      0 /usr/lib/ld-linux-riscv64-lp64d.so.1
      0x3ff7ffd000       0x3ff7ffe000 r--p     1000  1f000 /usr/lib/ld-linux-riscv64-lp64d.so.1
      0x3ff7ffe000       0x3ff8000000 rw-p     2000  20000 /usr/lib/ld-linux-riscv64-lp64d.so.1
      0x3ffffdf000       0x4000000000 rw-p    21000      0 [stack]
```

lebr0nli:
> it seems the program is not running?

Yep, it's not a running program.

> FYI, It works fine on my machine:

Yea, the `&main_arena` will be put into the next field after the libc.so been loaded, but the address seems doesn't be there before it's been loaded?
I mean, for example, in x86_64's libc(seems also true for i386, arm, aarch64 and mips):
```console
$ gdb -q /lib/x86_64-linux-gnu/libc.so.6
...
pwndbg> print main_arena
$1 = {
  mutex = 0,
  flags = 0,
  have_fastchunks = 0,
  fastbinsY = {0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0},
  top = 0x0,
  last_remainder = 0x0,
  bins = {0x0 <repeats 254 times>},
  binmap = {0, 0, 0, 0},
  next = 0x1ecb80 <main_arena>,
  next_free = 0x0,
  attached_threads = 1,
  system_mem = 0,
  max_system_mem = 0
}
```
As you can see, the address(not the real address during runtime) of `main_arena` is already there before we run it, so we can easily find it by parsing the ELF directly.

And our goal here is we want to find the address of `main_arena` without parsing the program's memory during runtime if possible, so we can find the address of `main_arena` in a short time correctly no matter how we changed the `main_arena`'s content during runtime.

But before this PR, to achieve what we want, we always assumed the compiler will put the address into the next field in compile time.
Since this might be wrong for riscv and powerpc, this PR is for handling this scenario and making the heap heuristics also work for it.

----

Btw, could you help me to test if the `heap` command can work without the debug symbols of stripped libc there for the single-threaded case with the enhancements in this PR?
(Since I don't have a real riscv device, I tested it with `qemu-riscv64`. I just want to make sure it also works for other environments.)

Xeonacid:
> Btw, could you help me to test if the heap command can work without the debug symbols of stripped libc there for the single-threaded case with the enhancements in this PR?
> (Since I don't have a real riscv device, I tested it with qemu-riscv64. I just want to make sure it also works for other environments.)

Yeah! It works.
```text
pwndbg> heap

Allocated chunk | PREV_INUSE
Addr: 0x2aaaaad000
Size: 0x291

Free chunk (tcachebins) | PREV_INUSE
Addr: 0x2aaaaad290
Size: 0x21
fd: 0x2aaaaad

Free chunk (tcachebins) | PREV_INUSE
Addr: 0x2aaaaad2b0
Size: 0x21
fd: 0x2aa800780d

Top chunk | PREV_INUSE
Addr: 0x2aaaaad2d0
Size: 0x20d31
```

BTW we PLCT lab can give you a real riscv hardware access for development if you need. 
A CI environment can also be provided. But I have to notice that they are low performance at this moment and may spend a lot of time completing a check. cc @disconnect3d

disconnect3d:
> As you can see, the address(not the real address during runtime) of main_arena is already there before we run it, so we can easily find it by parsing the ELF directly.

Wait: why is an address already there? I mean, shouldn't the libc.so be a shared object that can be loaded anywhere in the memory and so the address could change on each run?

> BTW we PLCT lab can give you a real riscv hardware access for development if you need.
> A CI environment can also be provided. But I have to notice that they are low performance at this moment and may spend a lot of time completing a check. cc @disconnect3d

Woah, that would be cool :). We should definetely send one to @lebr0nli if possible and see if there are much discrepancies between qemu-riscv and real hardware. If you have too much devices to give away, I could also take one, though I don't deal much with risc-v (yet? no idea). (I would probably donate it to [local hackerspace](https://hackerspace-krk.pl/) if I would not use it for a longer period and use it from there if needed; if that works for you ofc).

Xeonacid:
> > BTW we PLCT lab can give you a real riscv hardware access for development if you need.
> > A CI environment can also be provided. But I have to notice that they are low performance at this moment and may spend a lot of time completing a check. cc @disconnect3d
> 
> Woah, that would be cool :). We should definetely send one to @lebr0nli if possible and see if there are much discrepancies between qemu-riscv and real hardware. If you have too much devices to give away, I could also take one, though I don't deal much with risc-v (yet? no idea). (I would probably donate it to [local hackerspace](https://hackerspace-krk.pl/) if I would not use it for a longer period and use it from there if needed; if that works for you ofc).

Apologize for my confusing words, we can only give remote SSH access for now. :( And we have a [Jenkins](https://ci.rvperf.org/) instance for CI.

disconnect3d:
Oh, ssh access would work as well and would be easier to work with :)

lebr0nli:
> Yeah! It works.

@Xeonacid 
Great! Thanks for helping test it🙏

> PLCT lab can give you a real riscv hardware access for development if you need.

Cool, I'll let you know if I have this need, thanks!

----

> why is an address already there?

@disconnect3d 
This’s a good question, I am also curious about it.
After some research, I guess it’s probably for the REL or RELR relocations? (And I should probably call it offset instead of address.)
> Some references: 
> https://maskray.me/blog/2021-10-31-relative-relocations-and-relr
> https://www.technovelty.org/linux/rel-v-rela-relocations.html
> https://stackoverflow.com/questions/60462386/relocation-addend-in-elf-files-elf64-rel-vs-elf64-rela

But the weird thing is that seems GLIBC < 2.36 on x86_64 and aarch64 is using RELA relocation, but they still place the main_arena offset into .next field even they have `r_addend`, I'm not sure why the compiler did this.

----

BTW, based on the fact for the relocation section, I think the brute force I added in the previous commit in this PR is probably unnecessary, we can use the relocation section to avoid false positives, which should have better reliability and performance than brute forcing it in memory.
I’ll try to rewrite it.

pwndbg/heap/ptmalloc.py
```diff
@@ -1514,31 +1514,99 @@ def main_arena(self):
...
+                # https://github.com/bminor/glibc/blob/glibc-2.37/malloc/malloc.c#L1902-L1907
+                # static struct malloc_state main_arena =
+                # {
+                #   .mutex = _LIBC_LOCK_INITIALIZER,
+                #   .next = &main_arena,
+                #   .attached_threads = 1
+                # };
+                expected = self.malloc_state._c_struct()
+                expected.attached_threads = 1
+                next_field_offset = self.malloc_state(0).get_field_address("next")
```
disconnect3d: `get_field_address` returns an offset? maybe we should change the name of this :P

lebr0nli: > `get_field_address` returns an offset? maybe we should change the name of this :P

I agreed, maybe a `get_field_offset` class method will be better for here :)

Issue #1523, opened 2023-01-15
`slab info` is broken on latest version of Linux kernel


gsingh93:
At some point between kernel versions 5.10 and 6.1.5 the names of some of the struct fields changed, fix should be straightforward.

clark-ja:
I also just stumbled upon this today and for completeness here are my findings:

In kernel version 5.17 a separate `struct slab` from `struct page` was [introduced](https://elixir.bootlin.com/linux/v5.17/source/mm/slab.h#L9).

Furthermore, in the `struct kmem_cache_cpu`, the `page` and `partial` members of type `struct page *` where replaced by the `slab` and `partial` members of type `struct slab *`.

As far as I can see, for kernel versions `>= 5.17` the following [line](https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/commands/slab.py#L168) for `pwndbg/commands/slab.py` needs to be adjusted:

```py
page = cpu_cache["page"] # should be `page = cpu_cache["slab"]` depending on kernel version
```
and then also in `print_cpu_cache()` the `slab['pages']` needs to be replaced with `slab['slabs']`.

Not sure how `pwndbg` wants to do kernel version checks in its codebase though. I did find `pwndbg.gdblib.kernel.kversion()`, but it returns a `str` and is therefore not best suited for the task unless further processed...

EDIT:

Actually, there are already changes in kernel version 5.16 which should be addressed in pwndbg: there the `struct page`s member `pobjects` was removed (and pwndbgs `print_cpu_cache()` method uses it)

Instead of `pobjects` in linux kernel `== 5.16` (`mm/slub.c`) the value is calculated in-line [as follows](https://elixir.bootlin.com/linux/v5.16/source/mm/slub.c#L5484):
```c
/* Approximate half-full pages, see slub_set_cpu_partial() */
objects = (pages * oo_objects(s->oo)) / 2;
```
and then updated to [this](https://elixir.bootlin.com/linux/v5.17/source/mm/slub.c#L5489) in linux kernel versions `>= 5.17`:
```c
/* Approximate half-full slabs, see slub_set_cpu_partial() */
objects = (slabs * oo_objects(s->oo)) / 2;
```

gsingh93:
Thanks @theguy147! I don't have time to make these changes at the moment, would you be able to make a PR?

[...]

PR #1668, opened 2023-04-11
Fix slab info command for linux kernel >= 5.16

clark-ja:
see #1523.

Changes from this PR to `slab info` command:
- Add 5-level paging support for `x86_64`
- Support `struct slab` (introduced in linux 5.17)
- Extract version tuple from linux banner in `pwndbg.gdblib.kernel.krelease()`
- Support `pobjects` calculation formula introduced in linux 5.16
- For active Slab, introduce support for slabs freelist (even if rarely used)
- Simplify function arguments for `print_cpu_cache()`, `print_slab()`
- Extend and Clean Up `ArchOps`
- Display all slab slots instead of just free slots and if free, show next pointer
image

I have not yet had the time to look at how pwndbgs tests are working and how such a kernel command can be thoroughly tested in the CI, so I just opened this as a Draft PR for now.

### DIFF
--- a/pwndbg/commands/kversion.py
+++ b/pwndbg/commands/kversion.py
@@ -12,4 +12,4 @@ parser = argparse.ArgumentParser(description="Outputs the kernel version (/proc/
 @pwndbg.commands.OnlyWithKernelDebugSyms
 @pwndbg.commands.OnlyWhenPagingEnabled
 def kversion() -> None:
-    print(pwndbg.gdblib.kernel.kversion())
+    print(pwndbg.gdblib.kernel.kbanner())
--- a/pwndbg/commands/slab.py
+++ b/pwndbg/commands/slab.py
@@ -5,9 +5,6 @@ Some of the code here was inspired from https://github.com/NeatMonster/slabdbg
 """
 import argparse
 import sys
-from typing import Iterator
-from typing import List
-from typing import Union
 
 import gdb
 from tabulate import tabulate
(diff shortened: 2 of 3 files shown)


pwndbg/commands/kversion.py
```diff
@@ -12,4 +12,4 @@
 @pwndbg.commands.OnlyWithKernelDebugSyms
 @pwndbg.commands.OnlyWhenPagingEnabled
 def kversion() -> None:
-    print(pwndbg.gdblib.kernel.kversion())
+    print(pwndbg.gdblib.kernel.kbanner())
```
gsingh93, line 15: I think we can have a separate `kbanner` command that calls `kbanner`.

gsingh93 (commented):
I did a quick skim and overall this looks good. I think it'll be easier to do a more thorough review as well as to pass the tests if this is split into a few smaller PRs, i.e. one for the arch_ops modifications, one for kversion/kbanner, one for the 5lvl support, and one for the slab changes.

pwndbg/commands/kversion.py
clark-ja: i think i messed something up when renaming `kversion` to `kbanner` so i reverted that change and instead called the new function which returns the numeric tuple `krelease`

clark-ja:
hmm, i do not understand the error messages in the failing pipeline^^ you're probably right and i will split it up into smaller PRs. unfortunately this will likely not happen today as I need to work on some other projects though...

clark-ja:
Closing in favor of #1699

Issue #1453, opened 2022-12-22
Improve memoization code


gsingh93:
There's various issues with the memoization code in `pwndbg/lib/memoize.py`. To name a few:

1. From reading the code, it seems like it doesn't handle keyword arguments properly (not verified)?
2. Instead of having one cache and clearing that cache on each event, we have multiple caches per event.
3. The implementation with a class member variable defined in the subclass, used in the parent class, and also used in a static method in the subclass is confusing. I feel like there's a simpler way to do this.
4. We should probably have some performance profiling of the memoization code as it's the most frequently executed code in the codebase, and even small optimizations there will have a large performance impact.

gsingh93:
Suggestion from @disconnect3d: we can make the caches per-function instead of per-event by saving the cache in the function itself.

disconnect3d:
Done in #1671.

PR #1671, opened 2023-04-18
New lib/cache.py: make caching great again

disconnect3d:
This commit fixes bugs with old caching (memoize.py) and makes it more readable.

See also https://github.com/pwndbg/pwndbg/issues/1453


disconnect3d:
TODO/FIXME here: 
* Merge the `...cache_until("start") + _cache_until("objfile")` etc decorators
* Add detection to `cache_until` to fail if a function is already decorated (by detecting the `function.cache` attribute)

pwndbg/lib/cache.py
```diff
@@ -0,0 +1,140 @@
...
+
+# A singleton used to mark a cache miss
+_NOT_FOUND_IN_CACHE = object()
+_KWARGS_SEPARATOR = object()
+
+# Global value whether the results from cache are returned or not
+IS_CACHING = True
+
+
+def cache_until(*event_names) -> Callable:
```
gsingh93, line 104: Because the event names are strings, I can guarantee that one day we're going to have a bug where there's a type in one of the event names. Can we add a check here that every event in `event_names` is in `_ALL_CACHE_UNTIL_EVENTS`. It can be an `assert`, as this should be caught during development and not exposed to users as an exception.

I think it'll throw somewhere below when accessing dictionary elements, but if it's an assert it's easier to debug.

pwndbg/lib/cache.py
```diff
@@ -0,0 +1,140 @@
...
+new library/objfile are loaded, etc.
+"""
+
+from functools import wraps
+from typing import Any
+from typing import Callable
+from typing import Dict
+from typing import Tuple
+
+debug = 0
```
gsingh93: Are we planning to use this?

pwndbg/lib/cache.py
```diff
@@ -0,0 +1,140 @@
...
+    """
+    for (event_name, event_hooks) in event_tuples:
+        _ALL_CACHE_UNTIL_EVENTS[event_name].connect_event_hooks(event_hooks)
+
+
+# A singleton used to mark a cache miss
+_NOT_FOUND_IN_CACHE = object()
+_KWARGS_SEPARATOR = object()
+
+# Global value whether the results from cache are returned or not
```
gsingh93, line 100:
```suggestion
# Global value whether the results from cache are returned or not. Toggled with the `memoize` command.
```

pwndbg/lib/cache.py
```diff
@@ -0,0 +1,140 @@
...
+IS_CACHING = True
+
+
+def cache_until(*event_names) -> Callable:
+    def inner(func):
+        if hasattr(func, "cache"):
```
gsingh93, line 112: I see we're setting `decorator.cache = cache` below, is this going to result in the `func.cache` attribute being created? If so, it's not obvious to me why, so it would be good to add a comment here explaining it.

pwndbg/lib/cache.py
```diff
@@ -0,0 +1,140 @@
...
+                "Pass multiple event names to the `cache_until` decorator."
+            )
+
+        cache: Dict[Tuple[Any], Any] = {}
+
+        @wraps(func)
+        def decorator(*a, **kw):
+            if IS_CACHING:
+                key: Tuple[Any] = (a, _KWARGS_SEPARATOR, *kw.items())
+                value = cache.get(key, _NOT_FOUND_IN_CACHE)
```
gsingh93, line 128: I had to think for a bit about why we're using the `_NOT_FOUND_IN_CACHE` object here instead of just checking for `None`. If it's because `None` is itself a valid cache value, then let's add a comment about that. The comment where `_NOT_FOUND_IN_CACHE` is defined wasn't very helpful.

pwndbg/lib/cache.py
```diff
@@ -0,0 +1,140 @@
...
+                if value is not _NOT_FOUND_IN_CACHE:
+                    return value
+
+                value = func(*a, **kw)
+
+                # Sanity check; should we maybe do it only in debug?
```
gsingh93: I think it's fine to always have this check, not just for debug.

pwndbg/gdblib/hooks.py
```diff
@@ -54,43 +46,28 @@ def on_stop() -> None:
...
-    reset_on_exit._reset()
-
-
-@pwndbg.gdblib.events.thread
-def memoize_on_new_thread() -> None:
-    reset_on_thread._reset()
+import pwndbg.lib.cache
+
+pwndbg.lib.cache.connect_clear_caching_events(
+    (
```
gsingh93: Not a big deal, but I think a `dict[str, Tuple]` would be a little nicer here instead of the nested tuples.

pwndbg/gdblib/hooks.py
```diff
@@ -54,43 +46,28 @@ def on_stop() -> None:
...
+        (
+            "stop",
+            (
+                pwndbg.gdblib.events.stop,
+                pwndbg.gdblib.events.mem_changed,
+                pwndbg.gdblib.events.reg_changed,
+            ),
```
gsingh93: I think I understand why `mem_changed` and `reg_changed` are only necessary for the `stop` event, but this could benefit from a comment. Would this statement be correct?
```
Any cache that should be cleared when the program is stopped should also be cleared if the user does an operation to modify memory or registers while the program is stopped. This isn't necessary for the other events, because there is no way for memory or registers to be changed if the program is not stopped.
```

pwndbg/lib/cache.py
```diff
@@ -0,0 +1,140 @@
...
+                f"Cannot cache the {func.__name__} function twice! "
+                "Pass multiple event names to the `cache_until` decorator."
+            )
+
+        cache: Dict[Tuple[Any], Any] = {}
+
+        @wraps(func)
+        def decorator(*a, **kw):
+            if IS_CACHING:
+                key: Tuple[Any] = (a, _KWARGS_SEPARATOR, *kw.items())
```
gsingh93, line 123: Is `_KWARGS_SEPARATOR` necessary? Why can't we just use `(a, kw.items())`? The key type would then be `Tuple[Any, ItemsView[Any, Any]]`.

gsingh93 (commented):
Overall the design of this lgtm.

pwndbg/lib/cache.py
disconnect3d: `iw.items()` is a `dict_items` type which is unhashable, so we have to unpack it. If we unpack it, an empty dict is unpacked to 0 items and so it may clash with args. Alternatively, we can do `(a, (*kw.items()))` in which the dict_items is unpacked to a tuple.

```
In [3]: a={}

In [4]: a[ ({}.items(),) ] = 1
Cell In [4], line 1
----> 1 a[ ({}.items(),) ] = 1

TypeError: unhashable type: 'dict_items'

In [5]: hash( {}.items() )
Cell In [5], line 1
----> 1 hash( {}.items() )

TypeError: unhashable type: 'dict_items'
```

pwndbg/lib/cache.py
disconnect3d: The reason for this is to allow `func.cache.clear()` which may be useful somewhere or in tests.

pwndbg/lib/cache.py
disconnect3d: I was thinking of introducing an `if debug: <another impl that print lots of info e.g. when a cache hit/miss happens>`.

This can also be achieved with a hack like: `cache = {} if not debug else DebugDictWhichPrintStuff()`

pwndbg/lib/cache.py
disconnect3d: Yes, we need such a check. Will add

pwndbg/gdblib/hooks.py
disconnect3d: > because there is no way for memory or registers to be changed if the program is not stopped.

Sadly this is not entirely true since a hook can set regs/memory via gdb.execute :/

pwndbg/lib/cache.py
```diff
@@ -0,0 +1,160 @@
...
+
+from collections import UserDict
+from functools import wraps
+from typing import Any
+from typing import Callable
+from typing import Dict
+from typing import Tuple
+
+# Set to enable print logging of cache hits/misses/clears
+NO_DEBUG, DEBUG_GET, DEBUG_CLEAR, DEBUG_SET = 0, 1, 2, 4
```
gsingh93, line 15: A [Flag](https://docs.python.org/3/library/enum.html#enum.Flag) might make sense here instead.

disconnect3d: I think this is so simple it can stay as it is :P

PR #1673, opened 2023-04-18
add `krelease()` function

clark-ja:
This PR adds the `pwndbg.gdblib.kernel.krelease()` function which allows pwndbg to easily implement linux kernel version specific behavior.

The function returns a tuple of `int`: `(major, minor, patch)`.

It can then be used as follows:

```py
if pwndbg.gdblib.kernel.krelease() >= (5, 17):
    # do something
else:
    # do something else
```

The PR also allows fixing the currently broken `slab info` command (part of splitting up Draft PR #1668)

### DIFF
--- a/pwndbg/gdblib/kernel/__init__.py
+++ b/pwndbg/gdblib/kernel/__init__.py
@@ -1,6 +1,7 @@
 import functools
 import math
 import re
+from typing import Tuple
 
 import gdb
 
@@ -79,7 +80,7 @@ def kversion() -> str:
 
 @requires_debug_syms()
 @pwndbg.lib.memoize.reset_on_start
-def krelease() -> tuple[int, int, int]:
+def krelease() -> Tuple[int, ...]:
     # try to extract (major.minor.patch) version
     match = re.search(r"Linux version (\d+)\.(\d+)\.(\d+)", kversion())
     if match:
--- a/pwndbg/gdblib/kernel/__init__.py
+++ b/pwndbg/gdblib/kernel/__init__.py
@@ -1,5 +1,6 @@
 import functools
 import math
+import re
 
 import gdb
 
@@ -74,6 +75,21 @@ def kcmdline() -> str:
 def kversion() -> str:
     version_addr = pwndbg.gdblib.symbol.address("linux_banner")
     return pwndbg.gdblib.memory.string(version_addr).decode("ascii").strip()
+
+
+@requires_debug_syms()
+@pwndbg.lib.memoize.reset_on_start
+def krelease() -> tuple[int, int, int]:
+    # try to extract (major.minor.patch) version
+    match = re.search(r"Linux version (\d+)\.(\d+)\.(\d+)", kversion())
+    if match:
+        return tuple(map(int, match.groups()))
+    # try to extract (major.minor) version and append 0 for patch version
+    match = re.search(r"Linux version (\d+)\.(\d+)", kversion())
+    if match:
+        return tuple(map(int, match.groups())) + (0,)
+    # not found
+    raise Exception("Linux version tuple not found")
 
 
 @requires_debug_syms()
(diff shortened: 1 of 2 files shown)


pwndbg/gdblib/kernel/__init__.py
```diff
@@ -76,6 +78,21 @@ def kversion() -> str:
...
+def krelease() -> Tuple[int, ...]:
+    # try to extract (major.minor.patch) version
+    match = re.search(r"Linux version (\d+)\.(\d+)\.(\d+)", kversion())
+    if match:
+        return tuple(map(int, match.groups()))
+    # try to extract (major.minor) version and append 0 for patch version
+    match = re.search(r"Linux version (\d+)\.(\d+)", kversion())
+    if match:
+        return tuple(map(int, match.groups())) + (0,)
+    # not found
```
gsingh93: I don't think this comment is particularly helpful 😄

tests/qemu-tests/tests/test_qemu_system.py
```diff
@@ -13,3 +13,12 @@
 except Exception:
     traceback.print_exc()
     exit(1)
+
+
+try:
+    release_ver = pwndbg.gdblib.kernel.krelease()
+    # release should be int tuple of form: (major, minor, patch)
+    assert len(release_ver) == 3
```
gsingh93: Maybe an additional test would be joining the tuple elements together as a string and seeing if that string is contained in `kversion()`? Although I see you're appending `.0` for the patch version in some cases, so you would need to handle that special case.

pwndbg/gdblib/kernel/__init__.py
```diff
@@ -76,6 +78,21 @@ def kversion() -> str:
...
 
+@requires_debug_syms()
+@pwndbg.lib.memoize.reset_on_start
+def krelease() -> Tuple[int, ...]:
+    # try to extract (major.minor.patch) version
+    match = re.search(r"Linux version (\d+)\.(\d+)\.(\d+)", kversion())
+    if match:
+        return tuple(map(int, match.groups()))
+    # try to extract (major.minor) version and append 0 for patch version
+    match = re.search(r"Linux version (\d+)\.(\d+)", kversion())
```
gsingh93: Instead of doing the regex search twice, I think we can do it once with something like this:
```
match = re.search(r"Linux version (\d+)\.(\d+)(?:\.(\d+))?", kversion())
```

For 'Linux version 5.10' `match.groups()` returns `('5', '10', None)` and for 'Linux version 5.10.108'  it returns `('5', '10', '108')`

pwndbg/gdblib/kernel/__init__.py
clark-ja: i removed the comments :)

tests/qemu-tests/tests/test_qemu_system.py
clark-ja: such a test was now added

pwndbg/gdblib/kernel/__init__.py
clark-ja: i revised the function to use only one regex now and made the patch version in the tuple optional. i do not really see a reason to require it anyway as pythons tuple comparison also works on tuples of different lengths

clark-ja:
not sure why the ubuntu-20.04 tests fail at "install dependencies". does not seem related to this tiny PR oO

pwndbg/gdblib/kernel/__init__.py
```diff
@@ -76,6 +78,15 @@ def kversion() -> str:
     return pwndbg.gdblib.memory.string(version_addr).decode("ascii").strip()
 
 
+@requires_debug_syms()
+@pwndbg.lib.memoize.reset_on_start
+def krelease() -> Tuple[int, ...]:
+    match = re.search(r"Linux version (\d+)\.(\d+)(?:\.(\d+))?", kversion())
+    if match:
+        return tuple([int(x) for x in match.groups() if x])
```
gsingh93: nitpick: I don't think the nested list is necessary. You should just be able to do: `tuple(int(x) for x in match.groups() if x)`

gsingh93:
> not sure why the ubuntu-20.04 tests fail at "install dependencies". does not seem related to this tiny PR oO

Yea, something weird is going on with the tests, other PRs are failing as well. You don't need to worry about it.

gsingh93:
Merged, thanks!

PR #1676, opened 2023-04-19
Fix and Improvements for ArchOps

clark-ja:
Changes in this PR:
- Fix typo in `START_KERNEL_map` variable
- Complete address translation methods in ArchOps
- Remove some unused variables
- Add ArchOps tests
- Prepare foundation for supporting other memory models apart from SPARSEMEM_VMEMMAP (e.g. FLATMEM or SPARSEMEM)

In the above tests the QEMU VMs were wrongly detected to use 5-level paging support, so the detection algorithm was also fixed and 5-level paging was introduced:
- Fix 5-level paging detection
- Add 5-level paging support

This PR also lays the foundation to fix the `slab info` command (issue #1523) and is part of splitting the Draft PR #1668 into smaller PRs.


pwndbg/gdblib/kernel/__init__.py
```diff
@@ -101,30 +120,43 @@ def phys_to_pfn(self, phys: int) -> int:
...
-        return phys_to_virt(page_to_phys(page))
+        return pfn_to_virt(page_to_pfn(page))
 
 
 class x86_64Ops(ArchOps):
     def __init__(self) -> None:
-        if "X86_5LEVEL" in kconfig() and "no5lvl" not in kcmdline():
-            raise NotImplementedError("Level 5 page table support is not implemented")
+        if uses_5lvl_paging():
+            self.PAGE_OFFSET = 0xFF11000000000000
```
disconnect3d, line 151: Can we add a comment where do these values come from? Ideally pointing out to kernel sources where this is defined (I guess this is where it is defined?:P)

pwndbg/gdblib/kernel/__init__.py
```diff
@@ -153,11 +188,15 @@ def phys_to_pfn(self, phys: int) -> int:
     def pfn_to_phys(self, pfn: int) -> int:
         return pfn << self.PAGE_SHIFT
 
-    def phys_to_page(self, phys: int) -> int:
-        return (self.STRUCT_PAGE_SIZE * phys_to_pfn(phys)) + self.START_KERNEL_map
+    def pfn_to_page(self, pfn: int) -> int:
+        # assumption: SPARSEMEM_VMEMMAP memory model used
+        # FLATMEM or SPARSEMEM not (yet) implemented
```
disconnect3d, line 193: Is there any way we can detect the memory model?

clark-ja: TBH I haven't yet looked into it, that's why I just left it as a comment for now...

pwndbg/gdblib/kernel/__init__.py
clark-ja: I just added the references

pwndbg/gdblib/kernel/__init__.py
clark-ja: On first glance it looks like it can be detected entirely from the kernel config and should therefore be straight-forward from the output of the `kconfig()` command. But this might be something for another day ;)

clark-ja: I think (tm) that looking at/for the configs `CONFIG_SPARSEMEM`, `CONFIG_SPARSEMEM_EXTREME`, `CONFIG_SPARSEMEM_VMEMMAP` and `CONFIG_FLATMEM` in various combinations might be the key (but there might be other ones that I have overseen)...

pwndbg/gdblib/kernel/__init__.py
```diff
@@ -96,6 +96,25 @@ def is_kaslr_enabled() -> bool:
...
+def cpu_feature_capability(feature: int) -> bool:
+    boot_cpu_data = gdb.lookup_global_symbol("boot_cpu_data").value()
+    capabilities = boot_cpu_data["x86_capability"]
+    return (int(capabilities[feature // 32]) >> (feature % 32)) & 1 == 1
+
+
+@requires_debug_syms()
+@pwndbg.lib.memoize.reset_on_start
+def uses_5lvl_paging() -> bool:
+    X86_FEATURE_LA57 = 16 * 32 + 16
```
clark-ja: I should probably also give a kernel source code reference for this value...

clark-ja:
Unfortunately I tested this some more and there are still some edge cases, e.g. for PFNs lower than 32736, for which the address translations give wrong results. I need to investigate so I turned the PR to a Draft for now.

clark-ja:
sometimes the solution is easier than the problem :) simplifying `phys_to_virt` and `virt_to_phys` fixed the problem for all architectures and kernel configs which I was able to test.

The previous implementation is included in the linux kernel but likely only used in very special configurations which I wasn't able to reproduce. If at some point someone does manage to find such a configuration at a later point in time, the adjustment should be very straight-forward.

clark-ja:
The qemu-tests for ArchOps which were added only really test that no exception is raised and that each address translation between "page", "phys", "pfn" and "page" works both ways.

More meaningful tests with actual pages will be implicit when adding tests for the `slab` command in the next PR, after this one was merged.

gsingh93:
The old implementation is necessary for AArch64.

clark-ja:
> The old implementation is necessary for AArch64.

I was talking about the x86_64Ops. 

I also tested different aarch64 configs and everything seems to work now..

gsingh93:
Sorry, I made the last comment on mobile and misread things. I still think this will break something, as it's not generally true that `__va(__pa(x)) == x`. I'll try to test it out this weekend.

clark-ja:
Yeah, you're right and I was too tired to think it through yesterday: the assumption does not hold for all addresses. I reverted that change in the last commit and only test the kernels lowmem now, as in that range the assumption should be true for all possible configurations (even with CONFIG_HIGHMEM, and CONFIG_x86_PAE etc).

EDIT:
Just FYI: The failing check in the pipeline is not related to the last commit or this PR in general:
> Failing tests: tests/test_command_killthreads.py::test_command_killthreads_kills_specific_thread

clark-ja:
@gsingh93 
Not sure if this is wanted, but if you would add images for linux and ack to `linux-exploit-dev-env`s releases upstream which have `CONFIG_X86_5LEVEL = y`  then the lvl5 support could be easily tested in the pipeline by just adding runs which have `-cpu qemu64,+la57` appended to the `QEMU_ARGS` for ARCH=="x86_64".

I have tested this locally and it works fine but I guess building those images in the pipeline would take unreasonably long, so they would require the aforementioned change to your upstream configs...

disconnect3d:
If there is sth that needs to be build and it takes long time we can always
do that in a separate repository and save the built things as a release
artifacts

gsingh93:
Yea, the images are built from here: https://github.com/gsingh93/linux-exploit-dev-env

We can add a new image with these options. I can look into it over the weekend, or if you want to try adding it that's fine too.

pwndbg/gdblib/kernel/__init__.py
```diff
@@ -112,30 +112,48 @@ def phys_to_pfn(self, phys: int) -> int:
...
-    def phys_to_page(self, phys: int) -> int:
+    def pfn_to_page(self, phys: int) -> int:
         raise NotImplementedError()
 
-    def page_to_phys(self, page: int) -> int:
+    def page_to_pfn(self, page: int) -> int:
         raise NotImplementedError()
 
+    def virt_to_pfn(self, virt: int) -> int:
+        return phys_to_pfn(virt_to_phys(virt))
```
disconnect3d, line 129: Can we maybe add a comment to the class that PFN is physical page frame number and link to https://docs.kernel.org/mm/memory-model.html?

Ideally we want this code to be self explanatory in case someone else wants to learn from it 🤐

pwndbg/gdblib/kernel/__init__.py
disconnect3d: Shall we add a comment how it could be detected in the future?

disconnect3d (approved):

pwndbg/gdblib/kernel/__init__.py
clark-ja: I added the reference to your link to the `ArchOps` class and included a note that using that reference and the kernel config other memory models can be detected and supported in the future.

pwndbg/gdblib/kernel/__init__.py
clark-ja: done

clark-ja:
What is it with this CI randomly failing sometimes?

>  AttributeError: partially initialized module 'pwndbg' has no attribute 'lib' (most likely due to a circular import)

I am sure that adding just a simple comment did not cause this AttributeError in collecting the tests, but why does it consequently fail in all test environments then?

clark-ja:
Offtopic for this PR but might be related to the failing CI:

Every time the CI workflows fail there are error messages about potential cyclic dependencies in pwndbg, so I checked this out and found the following cycles in pwndbg (might not be exhaustive). The random CI issues might be related?


pkgs

disconnect3d:
The main tests error probably came from the fact that we changed the memoize implementation so that now you use `pwndbg.lib.cache.cache_until("stop", "start")` etc.

This can be observed in the error log from ur CI:
> AttributeError: module 'pwndbg.lib' has no attribute 'memoize'

Though yeah its weird it then reports that it cant find `lib` lol

PR #1682, opened 2023-04-22
make qemu-system tests more customizable

clark-ja:
This PR adds the capability to quickly comment/uncomment different qemu-system configuration combinations and makes it easily possible to add custom QEMU cli arguments.

Once a Linux kernel image which has `CONFIG_X86_5LEVEL = y` set was added to pwndbgs images, testing a system with 5-level paging is as easy as:

- Adding the string `"x86_64 linux -cpu qemu64,+la57"` to the `PLATFORMS` array in `tests/qemu-tests/tests.sh`

This causes qemu to be called with the additional argument `-cpu qemu64,+la57` which enables the necessary hardware support for 5-level paging.

### DIFF
--- a/tests/qemu-tests/run_qemu_system.sh
+++ b/tests/qemu-tests/run_qemu_system.sh
@@ -1,24 +1,57 @@
 #!/bin/bash
 
-ARCH="$1"
-KERNEL_TYPE="${2:-linux}"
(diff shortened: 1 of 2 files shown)


gsingh93:
lgtm, thanks!

Issue #1667, opened 2023-04-10
`bugreport` should report the OS name and version


gsingh93:
Running `bugreport`, I don't see any information about the OS being used. We should add this information. I think we do some OS detection in `setup.sh`, we can look at that for how to determine the OS.

disconnect3d:
Done in https://github.com/pwndbg/pwndbg/pull/1683. Thanks @MinimalEffort07 !

PR #1683, opened 2023-04-22
Issue #1667 Display OS Information In Bugreport

MinimalEffort07:

### DIFF
--- a/pwndbg/commands/version.py
+++ b/pwndbg/commands/version.py
@@ -5,8 +5,9 @@ Displays gdb, python and pwndbg versions.
 
 import argparse
 import os
+import platform
+import re
 import sys
-from platform import platform
 from subprocess import check_call
 from subprocess import check_output
 from tempfile import NamedTemporaryFile
@@ -144,8 +145,18 @@ If it is somehow unavailable, use:
 
     gdb_config = gdb.execute("show configuration", to_string=True).split("\n")
     all_info = all_versions()
+    os_info = platform.system()
 
-    current_setup = "Platform: %s\n" % platform()
+    current_setup = "Platform: %s\n" % platform.platform()
+
+    if os_info.lower() == "linux" and os.path.isfile("/etc/os-release"):
+        with open("/etc/os-release", "r") as os_release:
+            contents = os_release.read()
+            match = re.search('PRETTY_NAME="?([^",\n]+)', contents)
+            if match:
+                os_info = str(match.group(1))
+
+    current_setup += "OS: %s\n" % os_info
     current_setup += "\n".join(all_info)
     current_setup += "\n" + "\n".join(gdb_config)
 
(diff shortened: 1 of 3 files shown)


pwndbg/commands/version.py
```diff
@@ -146,6 +147,7 @@ def bugreport(run_browser=False, use_gh=False):
     all_info = all_versions()
 
     current_setup = "Platform: %s\n" % platform()
+    current_setup += "OS:       %s\n" % system()
```
disconnect3d: This provides info like 'Linux' or 'Darwin'; it would be great to display the distro name on linux :/

We should do sth like (and adjust imports):
```suggestion
    current_setup += "OS:       %s\n" % getattr(platform, "linux_distribution", platform.system)
```

MinimalEffort07: Apologies, the ticket didn't mention distribution so I didn't know that was meant.

Unfortunately platform.linux_distribution was deprecated in python3.5 and removed in python3.8.

There is platform.freedesktop_os_release(), however it is only available in python3.10 

Other than those two, there doesn't seem to be much functionality left in the platform module to determine distribution. 

I have amended the PR with a workaround.

pwndbg/commands/version.py
```diff
@@ -5,8 +5,10 @@
 
 import argparse
 import os
+import re
 import sys
 from platform import platform
+from platform import system
```
gsingh93: nitpick: I think we should just `import platform` and then call `platform.platform()` and `plaform.system()`. People generally think of `os.system()` when they see system, so the code below will be more clear like this.

pwndbg/commands/version.py
```diff
@@ -144,8 +146,19 @@ def bugreport(run_browser=False, use_gh=False):
 
     gdb_config = gdb.execute("show configuration", to_string=True).split("\n")
     all_info = all_versions()
+    os_info = system()
 
     current_setup = "Platform: %s\n" % platform()
+
+    if os_info.lower() == "linux" and os.path.isfile("/etc/os-release"):
+
```
gsingh93: nitpick: the blank line here is unnecessary, it doesn't really help readability.

pwndbg/commands/version.py
```diff
@@ -144,8 +146,19 @@ def bugreport(run_browser=False, use_gh=False):
...
+        with open("/etc/os-release", "r") as os_release:
+            contents = os_release.read()
+            match = re.search(r"PRETTY_NAME=(.+)", contents)
```
gsingh93: Why don't we match against `PRETTY_NAME="(.+)"` so we can avoid the `replace` call below? Or is it not guaranteed that there will be quotes here?

gsingh93 (commented):
Overall lgtm, will let @disconnect3d approve incase he has more comments.

pwndbg/commands/version.py
MinimalEffort07: I’m unsure if the quotes are guaranteed however when I get home I  can change the regex to:

PRETTY_NAME=“?(.+)”? 

Or something like that to get rid of the replace call below.

MinimalEffort07:
Hmm, I am unsure as to why the recent changes into pwndbg are showing up in my commit. I just synced my fork and squashed my commits. Not sure if this is a problem.

PR #1684, opened 2023-04-23
Test more kernels

clark-ja:
This PR downloads and tests with all available qemu kernel images upstream. 

The PR also modifies the `download_images.sh` skript to first download the `hashsums.txt` and only download an image if it has not yet been downloaded or if the new hashsum is different from the old one.

Also a few changes from my last qemu-system testing PRs have been modified again as I have noticed that it would overly complicate testing with these additional qemu images.

@gsingh93 
Merging this PR would require you to promote the last pre-release to a release on [linux-exploit-dev-env](https://github.com/gsingh93/linux-exploit-dev-env)

To demonstrate how this PR works with a new release on linux-exploit-dev-env, I created a fork of it with a new release and created a [demo branch](https://github.com/theguy147/pwndbg/tree/test_more_kernels_demo) to see a (hopefully) successful pipeline run.

EDIT:
oh, and x86_64 images are also tested with qemu "hardware support" for 5-level paging in this PR

Also, this PR is related to issue #1417


clark-ja:
Just for reference, [here](https://github.com/theguy147/pwndbg/actions/runs/4778973779/jobs/8495746224) is the successful pipeline run on my demo branch which shows the result of promoting the last pre-release on linux-exploit-dev-env to a real release:

FYI the pipeline currently fails because the newly introduced `hashsums.txt` file is still missing in the latest upstream release.

tests/qemu-tests/tests/system/test_gdblib_kernel.py
```diff
@@ -9,5 +12,6 @@ def test_gdblib_kernel_krelease():
     assert release_str in pwndbg.gdblib.kernel.kversion()
 
 
+@pytest.mark.skipif(not pwndbg.gdblib.kernel.has_debug_syms(), reason="")
```
gsingh93: Maybe just put `reason="This test requires debug symbols"`?

tests/qemu-tests/download_images.sh
```diff
@@ -8,21 +8,23 @@ URL="https://github.com/gsingh93/linux-exploit-dev-env/releases/latest/download"
...
-
-for kernel_type in ack linux; do
-    file="bzImage-${kernel_type}-x86_64"
-    wget "${URL}/${file}" -O "${OUT_DIR}/${file}"
-
-    file="Image-${kernel_type}-arm64"
-    wget "${URL}/${file}" -O "${OUT_DIR}/${file}"
-done
+download() {
+    file="$1"
```
gsingh93:
```suggestion
    local file="$1"
```

tests/qemu-tests/download_images.sh
```diff
@@ -8,21 +8,23 @@ URL="https://github.com/gsingh93/linux-exploit-dev-env/releases/latest/download"
...
+    hash_old=$(grep "${file}" "${OUT_DIR}/hashsums.txt.old" || true)
+    hash_new=$(grep "${file}" "${OUT_DIR}/hashsums.txt")
```
gsingh93, line 14: Not a big deal, but we could also pass this in as an argument when calling `download`.

tests/qemu-tests/download_images.sh
```diff
@@ -8,21 +8,23 @@ URL="https://github.com/gsingh93/linux-exploit-dev-env/releases/latest/download"
...
+    # only download file if it doesn't exist or its hashsum has changed
+    if [ ! -f "${OUT_DIR}/${file}" ] || [ "${hash_new}" != "${hash_old}"]; then
```
gsingh93: Don't you need a space at the end here: `[ "${hash_new}" != "${hash_old}"]` ? I'm surprised this worked.

tests/qemu-tests/run_qemu_system.sh
```diff
@@ -37,28 +36,21 @@ while [[ $# -gt 0 ]]; do
...
+    echo "Invalid kernel '${KERNEL_NAME}'"
     help_and_exit
 fi
 
-if [[ ! " ${VALID_TYPES[*]} " =~ " ${KERNEL_TYPE} " ]]; then
-    echo "Invalid kernel type '${KERNEL_TYPE}'"
-    help_and_exit
-fi
+# extract architecture as last dash-separated group of the kernel's name
+ARCH="${KERNEL_NAME##*-}"
```
gsingh93, line 50: Can you add a comment here indicating what we expect kernel names to look like? i.e `Image-${KERNEL_TYPE}-${VERSION}-${ARCH}` or whatever it's supposed to be?

tests/qemu-tests/tests.sh
```diff
@@ -216,13 +209,19 @@ test_system() {
...
+    KERNEL=$(echo "${vmlinux}" | sed "s/vmlinux-//")
+    # extract architecture as last dash-separated group of the kernels name
+    ARCH="${KERNEL##*-}"
+    QEMU_ARGS=()
 
     test_system
+
+    if [[ "${ARCH}" == @("x86_64") ]]; then
+        # additional test with extra QEMU flags
+        QEMU_ARGS=(-cpu qemu64,+la57)
```
gsingh93, line 236: I guess in the future we'll want to make this more flexible so we can specify different sets of qemu arguments with different kernel images (i.e. `nokaslr`) without having to add an if-statement for each one, but this is fine for now.

tests/qemu-tests/tests.sh
```diff
@@ -216,13 +209,19 @@ test_system() {
...
-    KERNEL_TYPE="$kernel_type"
-    QEMU_ARGS=($qemu_args)
-    VMLINUX="${IMAGE_DIR}/vmlinux-${KERNEL_TYPE}-${ARCH}"
+    VMLINUX="${IMAGE_DIR}/${vmlinux}"
+    KERNEL=$(echo "${vmlinux}" | sed "s/vmlinux-//")
+    # extract architecture as last dash-separated group of the kernels name
+    ARCH="${KERNEL##*-}"
+    QEMU_ARGS=()
 
     test_system
```
gsingh93: Maybe we should pass the kernel, arch, and any extra arguments as arguments to `test_system` instead of global variables?

tests/qemu-tests/tests/system/test_gdblib_kernel.py
clark-ja: With the current setup I do not think that the reason is going to be printed anywhere? But even just for readability I will add the reason :)

tests/qemu-tests/download_images.sh
clark-ja: you're right. interesting.

tests/qemu-tests/tests.sh
clark-ja: yes, I know. That's what I tried to achieve with the last PR. But using bash I wasn't creative enough to come up with some simple implementation which allows to do this.

tests/qemu-tests/tests.sh
clark-ja: I thought about doing that but then those arguments have to be passed further down to functions called from withing `test_system` and when I tried it I found that it made the code uglier. But if you still think so, I can change it back.

tests/qemu-tests/download_images.sh
clark-ja: sure, i can do that. just found the code to look cleaner and easier to understand like this. let me know if you want me to change this.

tests/qemu-tests/run_qemu_system.sh
clark-ja: will do! btw, available kernel names are also printed everytime `help_and_exit` is called, so e.g. with `run_qemu_system.sh --help`.

tests/qemu-tests/download_images.sh
gsingh93: It's fine, we can leave it as is.

tests/qemu-tests/tests.sh
gsingh93: I still think passing this as function arguments would be better, but would be worth getting another opinion. @disconnect3d?

clark-ja: No worries, if you think so it's okay. I just pushed a new commit for this.

clark-ja: tbh with bash I just don't like using more than one or maximum two arguments as they tend to be confusing/error prone when editing the code later on. I always have to scroll back and forth between the definition and the reference and check that i didn't miss any hard-to-spot function argument e.g. `$@` or similar somewhere down the line which messes everything up :)
At some point the logic should probably be ported to e.g. python where all function arguments are clearly visible in the function signature

.github/workflows/tests.yml
```diff
@@ -67,6 +67,13 @@ jobs:
         ./setup-dev.sh --user
         mkdir .cov
 
+    - name: Set up cache for QEMU images
+      id: qemu-cache
+      uses: actions/cache@v2
+      with:
+        path: ./tests/qemu-tests/images
+        key: ${{ matrix.os }}-cache-qemu-images
```
disconnect3d, line 75: How does this exactly work? I don't see us referring to it anywhere ( ? )

clark-ja: As I understand it (but I am in no means sure), this step initializes a cache with the `key` as its unique name and a path to the directory which should be cached. At the end of the workflow run it automatically creates an additional step where everything inside the specified directory is saved semi-persistently and restored the next time a cache with that exact name is initialized. Even though additional steps can be added which reference the cache and/ or check if a cache hit or miss occurred (a check whether such a cache exists or not), these additional steps are not necessary. 

BTW is the pip cache in the same workflow also not being referenced anywhere else and if you look at an arbitrary workflow run of the unit tests you can see the automatic step at the end if you look into is details ("Post cache for pip"):
image

Having said all that I noticed that pip uses cache@v3, so it might be a good idea to also use v3 for the qemu image cache.

Further info: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows

clark-ja: Let me know if I misunderstood the cache mechanism in any way :D

.github/workflows/tests.yml
```diff
@@ -67,6 +67,13 @@ jobs:
         ./setup-dev.sh --user
         mkdir .cov
 
+    - name: Set up cache for QEMU images
+      id: qemu-cache
+      uses: actions/cache@v2
```
clark-ja:
```suggestion
      uses: actions/cache@v3
```

.github/workflows/tests.yml
disconnect3d: I haven't read all of the cache docs but I think you are right :D

disconnect3d:
@gsingh93 can you fix the release in ur repo? Or shall we move this to Pwndbg org?

.github/workflows/tests.yml
gsingh93: I'm not sure why we're doing this if we're just going to overwrite the existing images in the cache with the following "Download images" step. I usually add something like this to the following step:
```
if: ${{ steps.cache-kernel.outputs.cache-hit != 'true' }}
```
see https://github.com/gsingh93/linux-exploit-dev-env/blob/main/.github/workflows/build.yml#L68-L81.

And if we do it like this, we should make the cache dependent on the release tag, i.e. 2023.01.19-61b3b5a for example. Then the cache will be used until a new release.

gsingh93:
I've updated the release, here is the latest one: https://github.com/gsingh93/linux-exploit-dev-env/releases/tag/2023.04.23-641fc26

We can migrate to a pwndbg owned repo in the next few weeks.

@theguy147 can you resolve the merge conflicts and re-test now that the release is updated? Don't worry about the caching issue, we can get this merged and you can fix it in a followup.

.github/workflows/tests.yml
clark-ja: the download images step was changed and only downloads the hashsums at first and then checks them against the old hashsums and only downloads an image if it has changed.

clark-ja:
The following qemu-tests are failing for an android image, with the error message "This command may only be run when debugging the Linux kernel in QEMU":
- test_command_kconfig 
- test_command_kversion 
- test_command_slab_list

I would think that some upstream modifications (to the images kernel config) messed with pwndbgs qemu detection? Or it is the new caching that messes with things again? It needs to be investigated...

The failing test on debian10 is unrelated to the PR:  
- tests/test_command_killthreads.py::test_command_killthreads_kills_specific_thread

gsingh93:
Hmm, I'm not sure why that's happening with the Android image. The QEMU check is here: https://github.com/pwndbg/pwndbg/blob/a9ec60c727df9c0d3af2822a261cdf0a831976ba/pwndbg/gdblib/qemu.py#L56

It relies on GDB's `maintenance packet` commands.

clark-ja:
It appears as if the issue is that the tests are run before even reaching `start_kernel`.

Then I noticed that I missed the asterisk in `break *start_kernel` in a previous PR (not sure why the tests didn't fail in that PR). Unfortunately, that only solves the failing tests for `ack-android13-5.10-lts-arm64` and e.g. not `ack-android13-5.10-lts-x86_64. 

Also, debugging failing tests with the `--pdb` flag does not currently work as the output is captured. I can fix that in a future PR.

clark-ja:
Ok, I fixed most of the failing tests. As far as I can tell only one fail remaining for `linux-5.10.178-x86_64` (test is only failing for that image, not for the others):
- test_gdblib_kernel_archops_address_translation

Again, the docker fail is not related to this PR.

disconnect3d:
> Ok, I fixed most of the failing tests. As far as I can tell only one fail remaining for linux-5.10.178-x86_64 (test is only failing for that image, not for the others):

Should we skip that test in this image? (if we do, add a comment why) Or shall we fix it somehow?

clark-ja:
> Should we skip that test in this image? (if we do, add a comment why) Or shall we fix it somehow?

I think fixing it is out-of-scope for this PR so I skipped it and put it on my todolist for the next days.

.github/workflows/tests.yml
disconnect3d: In theory this should not change for a given released version :D but yeah, if we target 'latest' it makes sense I guess.

gsingh93:
Sounds good to me. Awesome work @theguy147!

PR #1689, opened 2023-04-26
In malloc_chunk replace size with real_size

neryaz:
`real_size` is the size of chunk without size bits(PREV_INUSE ,IS_MMAPPED, NON_MAIN_ARENA)

### DIFF
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -420,7 +420,7 @@ def malloc_chunk(addr, fake=False, verbose=False, simple=False) -> None:
     if verbose:
         fields_to_print.update(["prev_size", "size", "fd", "bk", "fd_nextsize", "bk_nextsize"])
     else:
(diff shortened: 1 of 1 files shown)


CptGibbon:
I think there are good arguments for printing both `size` and `real_size` here, since they both convey useful information.
Perhaps adding a "real size" line to the output whilst keeping the "size" line would make everyone happy.

There are also tests that rely on that "size" output, so they'll need a small modification too 👍

disconnect3d:
I agree on printing both. Can we also mention the difference between them in some heap command's `--help` description so that it is more clear/discoverable to the user?

A code comment would also be nice here.

disconnect3d:
@neryaz can u change the PR to display both size and real size? Perhaps we can do it in the same line and write it like: `Size: %#x (real size with flag bits: %#x)`

Also, tests need to be fixed.

disconnect3d:
@neryaz ping

disconnect3d:
I'm going to close this because of lack of activity.

PR #1692, opened 2023-04-30
Refactor memory.Page to use less properties

disconnect3d:
Refactors the `pwndbg.gdblib.memory.Page` class to use `__slots__` for faster access.

We didn't use properties in all cases anyway. I think we could even think of computing flags as well, but idk.


gsingh93:
> faster access

Can you add a benchmarking script?

disconnect3d:
Closing this for now: that change is not crucial for Pwndbg functioning. It will speed up things but the speedup is likely negligible.

If someone wants to help here, I would accept a rebased PR with this change (without conflicts) because it makes sense to use this in the long term.

PR #1695, opened 2023-05-02
Add heuristic to add executable base to vmmap in qemu-user

peace-maker:
vmmap would try to add the executable to memory pages if the `info auxv` command contained an address, but the memory maps would be accessed recursively when trying to lookup the start of the ELF based on the given address.

Since qemu doesn't provide memory map info, do a leap of faith and try if the start of the page of the given address contains the ELF magic header.

Since the program headers are more likely to be on the same page as the ELF header than the program entrypoint, try both.

### DIFF
--- a/pwndbg/gdblib/elf.py
+++ b/pwndbg/gdblib/elf.py
@@ -273,32 +273,38 @@ def get_ehdr(pointer):
     We expect the `pointer` to be an address from the binary.
     """
 
-    # This just does not work :(
-    if pwndbg.gdblib.qemu.is_qemu():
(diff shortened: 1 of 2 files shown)


disconnect3d:
We should probably add a unit test for this case and start testing qemu user emulation case in general

disconnect3d:
Note: the test failure is irrelevant to this PR and is just just a flaky/racey test.

peace-maker:
Sorry, I didn't have time to look into enabling qemu-user tests yet. I'll keep it on my to-do :P

PR #1678, opened 2023-04-19
Optimize vis_heap_chunks command

disconnect3d:
This commit optimizes the vis_heap_chunks command by precalculating bin labels instead of computing them on demand for each chunk.

For `vis 2000` command ran when debugging python3 shell, those changed cut down the execution time from almost 20s to 5s.

The benchmark done is included in this commit in
profiling/benchmark_vis_heap_chunks/ so that it can be reproduced e.g. to optimize the function further or to reproduce my results.

The profiling results from my initial benchmarking of this issue on the old code can be seen here: https://github.com/pwndbg/pwndbg/issues/1675#issuecomment-1515360405

To pull up the relevant results:
```
$ python3 ../pwndbg/profiling/print_stats.py ./pwndbg-1681937259.pstats
Wed Apr 19 21:47:39 2023    ./pwndbg-1681937259.pstats

         24929221 function calls (24913554 primitive calls) in 9.356 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    9.356    9.356 {built-in method _gdb.execute}
        1    0.000    0.000    9.355    9.355 __init__.py:128(invoke)
        1    0.000    0.000    9.355    9.355 __init__.py:182(__call__)
        1    0.000    0.000    9.355    9.355 __init__.py:317(_OnlyWhenRunning)
        1    0.000    0.000    9.354    9.354 __init__.py:394(_OnlyWithResolvedHeapSyms)
        1    0.000    0.000    9.348    9.348 __init__.py:358(_try2run_heap_command)
        1    0.000    0.000    9.348    9.348 __init__.py:342(_OnlyWhenHeapIsInitialized)
        1    0.400    0.400    9.347    9.347 heap.py:823(vis_heap_chunks)
    49435    4.477    0.000    6.361    0.000 heap.py:982(bin_labels)
    49435    0.711    0.000    1.207    0.000 heap.py:975(bin_ascii)
    98872    0.210    0.000    0.727    0.000 memory.py:18(read)
  3213288    0.699    0.000    0.699    0.000 {method 'format' of 'str' objects}
  9689260    0.668    0.000    0.668    0.000 {built-in method builtins.hex}
  9764888    0.599    0.000    0.612    0.000 {built-in method builtins.isinstance}
    98872    0.462    0.000    0.462    0.000 {method 'read_memory' of 'gdb.Inferior' objects}
[...]

PR #1699, opened 2023-05-07
Repair slab info command

clark-ja:
Fix for #1523

Changes from this PR to `slab info` command:
- Support `struct slab` (introduced in linux 5.17)
- Support `pobjects` calculation formula introduced in linux 5.16 (even if possibly inaccurate)
- For active Slab, introduce support for slabs freelist
- Simplify function arguments for `print_cpu_cache()`, `print_slab()`
- Display all slab slots instead of just free slots and if free, show next pointer
- Add test for command

image

### DIFF
--- a/pwndbg/commands/slab.py
+++ b/pwndbg/commands/slab.py
@@ -5,9 +5,6 @@ Some of the code here was inspired from https://github.com/NeatMonster/slabdbg
 """
 import argparse
 import sys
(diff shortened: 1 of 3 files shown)


clark-ja:
FYI the failing heap tests in the archlinux docker are not related to this PR (but might be related to the #1678 PR?)

PR #1705, opened 2023-05-12
Updated canary.py for better display

### LINKED ISSUE #1704: Better display for canary command
The canary command should:
* Display the thread id for a given stack
* For the current stack (to which RSP points to) it should mark the offset from RSP to the given address

We can discuss which format we want to display it in when a PR comes.

Unknownsentinel193:
Fixes #1704 

The updated code now has the following changes:

1."current_thread" is obtained by "pwndbg.proc.current_thread_id()" and presented with stack canary information.

2.To acquire the current stack (to which RSP refers), use "current_rsp = pwndbg.regs.rsp." The stack canaries' offsets from RSP are calculated and shown using "offset_from_rsp = stack_canary - current_rsp" and the information is presented using "print()" and "message.address()."

### DIFF
--- a/pwndbg/commands/canary.py
+++ b/pwndbg/commands/canary.py
@@ -23,16 +23,13 @@ def canary_value():
 
 
 @pwndbg.commands.ArgparsedCommand(
-    "Print out the current stack canary.", category=CommandCategory.STACK
+    "Print out the current stack canary.",
+    category=CommandCategory.STACK,
+    aliases=("canary",),
 )
 @pwndbg.commands.OnlyWhenRunning
-@pwndbg.commands.aliases("can")
-@pwndbg.commands.add_argument(
-    "--all",
-    "-a",
-    action="store_true",
-    help="Display all canaries for all threads.",
-)
+@pwndbg.commands.aliases("canary")
+@pwndbg.commands.with_argparser
 def canary(all: bool = False) -> None:
     global_canary, at_random = canary_value()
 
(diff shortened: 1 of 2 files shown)


disconnect3d:
I think it is still useful to either display all canaries or to have an option (`--all`) to display all of them.

I am not sure what would be the best format of that.

We could either sort it all by thread id but start with current thread id and then show all others, or, we could keep displaying things as they are displayed now but display thread id and offset from rsp at the given line? I am not sure.

Unknownsentinel193:
If you want to organise canaries by thread, sorting by thread ID and starting with the current thread ID might be a decent strategy. This would make it easy to determine which canaries are associated with particular threads. However, if you have a large number of threads, the output may become quite lengthy and difficult to traverse.

Displaying the thread ID and offset from RSP for each canary may be informative as well. This would make it easy to determine where each canary is in the stack and to which thread it belongs. However, if you have multiple threads and canars, this may cause the output to be longer.

Finally, the ideal format would be determined by the individual use case and the most significant information to convey.

Here's an updated version of the canary command that includes the --all flag and displays the thread ID and offset from rsp for the current stack.

pwndbg/commands/canary.py
```diff
@@ -23,31 +23,70 @@ def canary_value():
...
+    thread = pwndbg.proc.get_thread_containing(stack_canary)
+    if thread is None:
+        print(message.warn("No thread found at current canary address."))
+    else:
+        print(
+            message.success(
+                f"Thread {thread.num}: {stack_canary - thread.sp} bytes from RSP"
+            )
+        )
+    pwndbg.commands.telescope.telescope(address=stack_canary, count=
```
disconnect3d: seems like sth is missing here (?)

pwndbg/commands/canary.py
```diff
@@ -23,31 +23,70 @@ def canary_value():
 
 
 @pwndbg.commands.ArgparsedCommand(
-    "Print out the current stack canary.", category=CommandCategory.STACK
+    "Print out the current stack canary.",
+    category=CommandCategory.STACK,
 )
 @pwndbg.commands.OnlyWhenRunning
-def canary() -> None:
+@pwndbg.commands.ParsedCommand(parser)
```
disconnect3d: Hm interesting, we use both ArgparsedCommand and ParsedCommand (I wonder what does this do :P)

Can we just use ArgparsedCommand and explicitly add the --all argument there?

pwndbg/commands/canary.py
Unknownsentinel193: My bad, I will complete it.

pwndbg/commands/canary.py
Unknownsentinel193: Ok, Sure I will change it.

disconnect3d:
Note that this has some errors: https://github.com/pwndbg/pwndbg/actions/runs/4966177366/jobs/8887464927?pr=1705

PR #1706, opened 2023-05-12
Fix test command procinfo

### LINKED ISSUE #1638: The procinfo test should control the netcat it spawns
The `test_command_procinfo` spawns netcat via `os.system("... &")` which is bad: if someone has wrong netcat version installed, it will not be detected at all. E.g. `netcat` package installed on Fedora:
```
[root@pwndbg pwndbg]# nc -l -p 31337
nc: cannot use -p and -l
```

We should do it via `subprocess.run(...)` instead and handle cases when 1) netcat does not exist; 2) fails to start. We also should close it properly because now, we expect that it will close itself after the test connect to it. But if the test fails... netcat won't be finished :).

```
_______________________________ test_command_procinfo ________________________________

start_binary = <function start_binary.<locals>._start_binary at 0x7fdc07fb5940>

    def test_command_procinfo(start_binary):
        start_binary(REFERENCE_BINARY_NET)

        # Sanity check, netcat must exist at this point
        assert shutil.which("nc") is not None
        os.system("nc -l -p 31337 2>/dev/null 1>&2 &")

        bin_path = gdb.execute("pi pwndbg.gdblib.proc.exe", to_string=True).strip("\n")
        pid = gdb.execute("pi pwndbg.gdblib.proc.pid", to_string=True).strip("\n")

        gdb.execute("continue")

        result = gdb.execute("procinfo", to_string=True)
        res_list = result.split("\n")

>       assert bin_path in res_list[0]
E       assert "'/root/pwndbg/tests/gdb-tests/tests/binaries/reference-binary-net.out'" in 'procinfo: The program is not being run.'
```

Unknownsentinel193:
The new code uses subprocess..Popen to spawn netcat with the appropriate options. If netcat does not exist, shutil.which will return None and the test will fail with a message indicating that netcat is not installed. If netcat fails to start, subprocess.Popen will raise a subprocess.CalledProcessError exception and the test will fail with a message indicating that netcat failed to start.

To close netcat properly, I used os.killpg to send a SIGTERM signal to the process group of the netcat subprocess. This ensures that all child processes of netcat (e.g., the listener socket) are also terminated.

Fixes #1638

### DIFF
--- a/tests/gdb-tests/tests/test_command_procinfo.py
+++ b/tests/gdb-tests/tests/test_command_procinfo.py
@@ -1,5 +1,7 @@
 import os
 import shutil
+import subprocess
+import signal
 
 import gdb
 
@@ -11,9 +13,16 @@ REFERENCE_BINARY_NET = tests.binaries.get("reference-binary-net.out")
 def test_command_procinfo(start_binary):
     start_binary(REFERENCE_BINARY_NET)
 
-    # Sanity check, netcat must exist at this point
-    assert shutil.which("nc") is not None
-    os.system("nc -l -p 31337 2>/dev/null 1>&2 &")
+    # Check if netcat exists
+    nc_path = shutil.which("nc")
+    assert nc_path is not None, "netcat is not installed"
+
+    # Spawn netcat
+    netcat_process = subprocess.Popen([nc_path, "-l", "-p", "31337"],
+                                      stdin=subprocess.DEVNULL,
+                                      stdout=subprocess.DEVNULL,
+                                      stderr=subprocess.DEVNULL,
+                                      start_new_session=True)
 
     bin_path = gdb.execute("pi pwndbg.gdblib.proc.exe", to_string=True).strip("\n")
     pid = gdb.execute("pi pwndbg.gdblib.proc.pid", to_string=True).strip("\n")
@@ -27,6 +36,9 @@ def test_command_procinfo(start_binary):
     assert bin_path in res_list[0]
     assert pid in res_list[1]
     assert "127.0.0.1:31337" in result
+
+    # Close netcat
+    os.killpg(os.getpgid(netcat_process.pid), signal.SIGTERM)
 
 
 def test_command_procinfo_before_binary_start():
--- a/tests/gdb-tests/tests/test_command_procinfo.py
+++ b/tests/gdb-tests/tests/test_command_procinfo.py
@@ -18,11 +18,13 @@ def test_command_procinfo(start_binary):
     assert nc_path is not None, "netcat is not installed"
 
     # Spawn netcat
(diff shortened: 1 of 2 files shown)


disconnect3d:
@Unknown1934-del Your PR contains changes to canary.py as well; can you rebase it to remove the commit that changes canary.py?

(you can do it via sth like `git rebase -i origin/dev`, removing the line with the canary.py commit and saving the file: this should rebase your commit on top of origin/dev branch and so apply all the commits that you left on the list in the interactive rebase editor)

Unknownsentinel193:
Sure, will do that.

Unknownsentinel193:
I have reverted the canary.py back

disconnect3d:
I've added a commit that fixes a minor lint issue to this PR since the lint CI job was failing.

There was also a fail for unit tests on one of the distros but it was caused by a flaky/racing test for the `killthreads` command. Seems that doubling its timeout time (5s -> 10s) did not help to fix it on CI. You can ignore that one.

Unknownsentinel193:
Sure, Thank You for adding it.

disconnect3d:
Hmm the lint is still failing: https://github.com/pwndbg/pwndbg/actions/runs/4960835058/jobs/8876857412?pr=1706

Can you run it on your machine and commit changes? That should be `./lint.sh -f` (you may need to install dev dependencies first: [`./setup-dev.sh`](https://github.com/pwndbg/pwndbg/blob/dev/setup-dev.sh))

Unknownsentinel193:
Sure, will do that and get back, Thank You

Unknown1934-del:
I have run the tests, and it shows no issues.


screenshot

Unknown1934-del:
I have run lint successfully.

Screenshot

Unknown1934-del:
I have run lint successfully.

screenshot

Unknownsentinel193:
@disconnect3d All the checks have passed, could you please give a final review.
Thank You

tests/gdb-tests/tests/test_command_procinfo.py
```diff
@@ -28,6 +39,9 @@ def test_command_procinfo(start_binary):
     assert pid in res_list[1]
     assert "127.0.0.1:31337" in result
 
+    # Close netcat
+    os.killpg(os.getpgid(netcat_process.pid), signal.SIGTERM)
```
disconnect3d, line 43: Is there a reason why we don't kill it via the [`Popen.kill` method](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.kill)?

We should either use that method or if it doesn't make sense, we should document it in a comment why we need to do it in a different manner

disconnect3d:
I merged it since it does the job, but please check out the comment in https://github.com/pwndbg/pwndbg/pull/1706#discussion_r1193182306 if you can :)

PR #1707, opened 2023-05-12
Add `slab contains` command

clark-ja:
This PR adds a new command: `slab contains addr [addr ...]` which returns the name of the slab cache for each of the addresses provided.

image

Also, the PR fixes a faulty 5-level paging detection on x86_64 which became apparent while adding a test for the new subcommand.

EDIT:
- The PR also fixes the address translations for aarch64 kernels >= 5.11 by updating the `VMEMMAP_START` calculation from that version on.
- Furthermore, there was an issue where some parts of the kernel were not sufficiently initialized when reaching the `start_kernel` function and thereby the qemu test infrastructure was modified to run until `rest_init` instead (which is called around the end of the kernel initialization).

### DIFF
--- a/pwndbg/commands/slab.py
+++ b/pwndbg/commands/slab.py
@@ -2,6 +2,7 @@
 Commands for dealing with Linux kernel SLAB memory allocator
 
 Some of the code here was inspired from https://github.com/NeatMonster/slabdbg
+Some of the code here was inspired from https://github.com/osandov/drgn
 """
 import argparse
 import sys
@@ -30,7 +31,7 @@ if (sys.version_info.major, sys.version_info.minor) >= (3, 7):
     subparsers.required = True
 
 
-parser_list = subparsers.add_parser("list", prog="slab")
+parser_list = subparsers.add_parser("list", prog="slab list")
 parser_list.add_argument(
     "filter_",
     metavar="filter",
@@ -40,9 +41,12 @@ parser_list.add_argument(
 )
 
 # TODO: --cpu, --node, --partial, --active
-parser_info = subparsers.add_parser("info", prog="slab")
+parser_info = subparsers.add_parser("info", prog="slab info")
 parser_info.add_argument("names", metavar="name", type=str, nargs="+", help="")
 parser_info.add_argument("-v", "--verbose", action="store_true", help="")
+
+parser_contains = subparsers.add_parser("contains", prog="slab contains")
+parser_contains.add_argument("addresses", metavar="addr", type=str, nargs="+", help="")
 
 
 def swab(x):
@@ -62,12 +66,15 @@ def swab(x):
 @pwndbg.commands.OnlyWhenQemuKernel
 @pwndbg.commands.OnlyWithKernelDebugSyms
 @pwndbg.commands.OnlyWhenPagingEnabled
-def slab(command, filter_=None, names=None, verbose=False) -> None:
+def slab(command, filter_=None, names=None, verbose=False, addresses=None) -> None:
     if command == "list":
         slab_list(filter_)
     elif command == "info":
         for name in names:
             slab_info(name, verbose)
+    elif command == "contains":
+        for addr in addresses:
+            slab_contains(addr)
 
 
 _flags = {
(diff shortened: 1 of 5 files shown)


disconnect3d:
Btw why is the qemu test failing? :x

clark-ja:
> Btw why is the qemu test failing? :x

I made a wrong assumption about the range of valid addresses for the `slab contains` subcommand and didn't notice it when running my local tests because gdb-multiarch was broken and all aarch64 were thereby skipped. I have since repaired my local setup and made some changes but there is still one architecture/kernel/config combination which is failing and I have yet to find out why and at what point either I made wrong assumptions or other parts of the kernel code (such as the address translation) is not 100% correct.

EDIT:
It seems very likely to me that it is indeed the address translation or a related issue with (some) aarch64 system(s). Currently the `slab info` command returns virtual addresses for the objects that do not appear to be "mappable" to physical memory (as far as I can tell)

clark-ja:
I finally found the problem and it took way too long... [This commit](https://lore.kernel.org/all/20201008153602.9467-4-ardb@kernel.org/) made it into the linux kernel for version 5.11 and changed the calculation of `VMEMMAP_START` for the arm64/aarch64 architecture.

clark-ja:
Finally working, I also updated the PR description above to reflect the fixes for the address translation and the reasoning behind running the system tests only after the function `rest_init` was reached.

pwndbg/commands/slab.py
```diff
@@ -292,3 +299,35 @@ def slab_list(filter_) -> None:
...
+
+    min_pfn = 0
+    max_pfn = int(gdb.lookup_global_symbol("max_pfn").value())
+
+    start_addr = pwndbg.gdblib.kernel.pfn_to_virt(min_pfn)
+    end_addr = pwndbg.gdblib.kernel.pfn_to_virt(
+        max_pfn + pwndbg.gdblib.kernel.arch_ops().page_size()
+    )
+
+    if not start_addr <= addr <= end_addr:
```
disconnect3d: Btw shouldn't this maybe be `start_addr<=addr<end_addr` :P?

clark-ja: you're absolutely right

pwndbg/gdblib/kernel/__init__.py
```diff
@@ -165,6 +168,9 @@ def __init__(self) -> None:
         self.PAGE_SHIFT = 12
         self.phys_base = 0x1000000
 
+    def page_size(self) -> int:
+        return 1 << self.PAGE_SHIFT
```
disconnect3d, line 172: Is `PAGE_SHIFT` constant? I believe so and so we should make it a constant of the module or a class variable. Then, `PAGE_SIZE` should probably also be a constant.

We also have `pwndbg.gdblib.memory.PAGE_SIZE = 0x1000`, maybe we should just use it here ( ? ).

disconnect3d: (Otherwise, if we keep it as it is now, each call to `.page_size()` will recompute the shift; not a big deal, but such things can build up and make things slower and slower with time)

pwndbg/gdblib/kernel/__init__.py
```diff
@@ -215,8 +221,8 @@ def uses_5lvl_paging() -> bool:
         # https://elixir.bootlin.com/linux/v6.2/source/arch/x86/include/asm/cpufeatures.h#L381
         X86_FEATURE_LA57 = 16 * 32 + 16
```
disconnect3d, line 222: Could be a constant but w/e

pwndbg/gdblib/kernel/__init__.py
disconnect3d: Hmm I can see that we may set `PAGE_SHIFT` based on a config:
```py
self.PAGE_SHIFT = int(kconfig()["CONFIG_ARM64_PAGE_SHIFT"])
```
 
So it may change (?)

Maybe we should just move the `page_size` impl to base class since both children have the same impl? Or do we want to protect against incorrect impl on a new arch the way its done right now?

disconnect3d:
Thanks! I am going to merge this even though I left some comments. We can address them in separate PR if we feel they need to be addressed.

gsingh93:
This looks good, the only thing I would say is in future PRs maybe we should start moving some slab code into lib or gdblib. Ideally the command code should only be dealing with handling user input and displaying output (although I know we are currently not very consistent about this).

clark-ja:
> This looks good, the only thing I would say is in future PRs maybe we should start moving some slab code into lib or gdblib. Ideally the command code should only be dealing with handling user input and displaying output (although I know we are currently not very consistent about this).

I completely agree. That is why I opened PR #1711 to better separate frontend and backend logic.

Issue #1634, opened 2023-03-19
Officially support installation in a virtualenv


gsingh93:
- [ ] Add a `--venv` flag to `setup.sh` that creates a new venv and installs pwndbg in it
- [ ] Remove any warnings about venvs, i.e. "Pwndbg/GDB run in virtualenv with which it may not work correctly"
- [ ] Add tests
- [ ] Add docs

Cnly:
I think one bit isn't quite clear in the description: Do we want to create a venv dedicated to pwndbg, or will we reuse the user's currently active venv? If it's the former, maybe `~/.cache` would be an option for the location of our venv?

disconnect3d:
@Cnly Whatever is best. I think we rather want to create a new virtual environment and we will just have to ask the user "where to put it", suggesting a location like `$pwndbg/.venv`, `~/.venv/pwndbg` or `~/.local/pwndbg-venv`. Alternatively, iirc we were thinking some time ago about some local cache, so maybe it could be sth like `~/.pwndbg/{venv,cache}`? I am not sure.

PR #1710, opened 2023-05-13
Include opensuse-tumbleweed in setup.sh

alexbradd:
AS title says, the install commands for `opensuse-leap` work also for `opensuse-tumbleweed`.

PS. Is python2 still required? I don't see it in `install_apt`. If it not required lmk and since I am at it I will add a commit for the opensuse family.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -131,7 +131,7 @@ if linux; then
         "clear-linux-os")
             install_swupd
             ;;
-        "opensuse-leap")
+        "opensuse-leap" | "opensuse-tumbleweed")
             install_zypper
             ;;
         "arch" | "archarm" | "endeavouros" | "manjaro" | "garuda")


tamara-schmitz:
I can confirm that the installer script ran just fine on my particular TW system (version 20230511).

Regarding the failing pipeline I assume that the test is just flaky and a force push to your branch may trigger the pipeline rerun.

tamara-schmitz:
Though I must say that this does not spark joy:

```
+ sudo /usr/bin/python3.10 -m pip install --target /usr/local/lib64/python3.10/site-packages --upgrade pip
Collecting pip
  Downloading pip-23.1.2-py3-none-any.whl (2.1 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 2.1/2.1 MB 38.3 kB/s eta 0:00:00
Installing collected packages: pip
Successfully installed pip-23.1.2
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
```

alexbradd:
That is a problem with the installer on all platforms. Ideally it should use the native package manager for everything or install the python dependencies only in user directories.

If this is wanted I can also tweak the setup to make it user-local by default.

disconnect3d:
@alexbradd
> Ideally it should use the native package manager for everything or install the python dependencies only in user directories.
Not really, because then we may end up with more problems if there would be incompatible versions between two distros.

What we want is to move to using Python virtual environments instead for Python installed packages: https://github.com/pwndbg/pwndbg/issues/1634

PR #1711, opened 2023-05-13
refactor: move slab business logic to gdblib

clark-ja:
As I was trying to extend the `slab` command logic further and wanted to re-use it in some custom scripts I noticed that we have mixed "frontend" and "backend" logic. To allow these computations to be re-used and further extended, I separated the concerns and refactored the business logic into `gdblib` with this PR.

This is a pure refactor PR without any changing functionality in the "frontend".

There will be some merge conflicts with the currently open #1707 PR which I will resolve once one of these two PRs have been merged into `dev`.

### DIFF
--- a/pwndbg/gdblib/kernel/slab.py
+++ b/pwndbg/gdblib/kernel/slab.py
@@ -1,22 +1,33 @@
+from typing import Generator
+from typing import List
+from typing import Optional
+from typing import Set
+
 import gdb
 
+from pwndbg.gdblib.kernel import kconfig
+from pwndbg.gdblib.kernel import page_to_virt
+from pwndbg.gdblib.kernel import per_cpu
 from pwndbg.gdblib.kernel.macros import for_each_entry
+from pwndbg.gdblib.kernel.macros import swab
+from pwndbg.gdblib.memory import pvoid
 
 
-def caches():
+def caches() -> Generator["SlabCache", None, None]:
     slab_caches = gdb.lookup_global_symbol("slab_caches").value()
     for slab_cache in for_each_entry(slab_caches, "struct kmem_cache", "list"):
-        yield slab_cache
+        yield SlabCache(slab_cache)
 
 
-def get_cache(target_name: str):
+def get_cache(target_name: str) -> Optional["SlabCache"]:
     slab_caches = gdb.lookup_global_symbol("slab_caches").value()
     for slab_cache in for_each_entry(slab_caches, "struct kmem_cache", "list"):
         if target_name == slab_cache["name"].string():
-            return slab_cache
+            return SlabCache(slab_cache)
+    return None
 
 
-def get_slab_key() -> str:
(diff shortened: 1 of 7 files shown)


clark-ja:
All related tests are succeeding now. The failing test is related to #1714 .

clark-ja:
I am debating whether or not to use caching for some of the refactored code. I guess some of the fields are expected to remain constant throughout one debug session of a kernel while others might change. On many/most kernel configurations this could (currently) be straight-forward, but I am not sure about architectures and configurations that I am not familiar with and I am not sure how some of these fields calculation might change in future kernel versions. Maybe it would be best to only start caching necessary parts once those field calculations have been identified as a bottleneck? Any opinions?

pwndbg/gdblib/kernel/slab.py
```diff
@@ -35,3 +45,256 @@ def oo_order(x: int) -> int:
...
+
+    @property
+    def offset(self) -> int:
+        return int(self._slab_cache["offset"])
+
+    @property
+    def random(self) -> int:
+        return (
+            int(self._slab_cache["random"])
+            if kernel.kconfig().get("SLAB_FREELIST_HARDENED") == "y"
```
gsingh93: I think we can do `if "SLAB_FREELIST_HARDENED" in kernel.kconfig()`, which at least to me looks nicer.

gsingh93 (approved):
Overall lgtm, just one minor comment.

> Maybe it would be best to only start caching necessary parts once those field calculations have been identified as a bottleneck?

Yes, this is what I'd suggest. We can improve performance over time as we identify performance issues.

pwndbg/gdblib/kernel/slab.py
clark-ja: This would work in the qemu test images we use, but might cause errors in other configurations. What if someone is debugging a system with `CONFIG_SLAB_FREELIST_HARDENED=n`? Then the `if` condition would be true, because that config is set, but once the logic would start "derandomizing" it would actually start to scramble things up.

Correct me if I'm wrong here...

EDIT: 
thb I'm not even sure if the `=n` syntax is valid, but my point is that it seems to me as if we want to explicitly check for `=y` here? Let me know if that is not necessary as I haven't experimented around with such kernel configs yet.

gsingh93: > I'm not even sure if the =n syntax is valid

Right, it's not valid. `# CONFIG_<optionName> is not set` is how you specify that an option is not set [^1]. For boolean options, I feel like checking if it's equal to "y" is the same as `if some_bool == True` instead of `if some_bool`, which feels strange to me. Checking the value would make more sense for any non-boolean options.

[^1]: https://stackoverflow.com/questions/41946817/which-approach-to-use-for-kernel-config

clark-ja: ok, I changed it now

Issue #1630, opened 2023-03-17
Make `args` an alias for `dumpargs`


gsingh93:
Currently `args` is an alias for `argv`, but it would be more useful if this was an alias for `dumpargs` instead.

dmur1:
i will take a look at this one

dmur1:
https://github.com/pwndbg/pwndbg/pull/1717

PR #1717, opened 2023-05-18
Make args an alias for dumpargs (#1630)

dmur1:
* tests: add a test for dumpargs

### DIFF
--- a/pwndbg/commands/argv.py
+++ b/pwndbg/commands/argv.py
@@ -23,7 +23,7 @@ parser.add_argument(
 )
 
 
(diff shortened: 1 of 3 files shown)


gsingh93:
Thanks for the PR, and thanks for adding a test! In this particular case though, I don't think the test adds much and can probably be removed. If we want to test aliases, we should have one generic alias test that tests everything instead of one test per-alias. And if that's something we do want (and I'm not sure if it is), it would be out of scope for this PR.

dmur1:
closed in favour of https://github.com/pwndbg/pwndbg/pull/1721

PR #1721, opened 2023-05-19
Make args an alias for dumpargs

### LINKED ISSUE #1630: Make `args` an alias for `dumpargs`

dmur1:


gsingh93:
Thanks, in the future though you can just update your original PR instead of making a new one. But thanks for the contribution 🙂

PR #1722, opened 2023-05-19
Add OnlyWithLibraryLoaded decorator

dmur1:
* Applied decorator to errno and tls commands

* Tests: updated expected output of errno command test

Testing with a binary that loads libc:

image
image

Testing with a binary that doesn't load libc:

image
image

### DIFF
--- a/pwndbg/commands/tls.py
+++ b/pwndbg/commands/tls.py
@@ -24,6 +24,7 @@ parser.add_argument(
 
 @pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.LINUX)
 @pwndbg.commands.OnlyWhenRunning
+@pwndbg.commands.OnlyWithLibraryLoaded("libc.so")
 def tls(pthread_self=False) -> None:
     tls_base = (
         pwndbg.gdblib.tls.find_address_with_register()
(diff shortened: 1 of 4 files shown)


pwndbg/commands/tls.py
```diff
@@ -24,6 +24,7 @@
 
 @pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.LINUX)
 @pwndbg.commands.OnlyWhenRunning
+@pwndbg.commands.OnlyWithLibraryLoaded("libc.so")
```
disconnect3d, line 27: Hmm I wonder: do errno or tls command work in statically linked binaries? (cc: @lebr0nli)

If they do, maybe we do not need the decorator/check?

lebr0nli: > Hmm I wonder: do errno or tls command work in statically linked binaries? (cc: @lebr0nli)

Yes, `errno` and `tls` might also work in statically linked binaries.

> If they do, maybe we do not need the decorator/check?

Yeah, currently I think it probably needs more consideration before applying this decorator, to avoid accidentally blocking the user from using the commands.

dmur1: reasonable - so would you like me to prepare a pull request that just adds the unused decorator or not really?

disconnect3d: Not really, lets close this for now

Issue #1666, opened 2023-04-10
Weird behavior when using `pwndbg.gdblib.symbol.address` on thread-local variables


lebr0nli:
This is yet another bug caused by the `info address` result of the thread-local variable.
This can be reproduced by:
```console
$ cat a.c
#include <stdlib.h>

int main(){
    free(malloc(0x20));
    return 0;
}
$ gcc -static-pie a.c
$ gdb -q ./a.out -ex 'set context-sections' -ex 'set exception-verbose' -ex start -ex 'print thread_arena' -ex 'info address thread_arena' -ex 'pi hex(pwndbg.gdblib.symbol.address("thread_arena"))'
...
Cannot find thread-local storage for process 2593, executable file /tmp/a.out:
Cannot find thread-local variables on this target
Symbol "thread_arena" is at 0x7fcfc93de038 in a file compiled without debugging.
'0x7fcfc93de038'
...
```

As you can see, GDB can't print `thread_arena` because it can't find thread-local storage.
But somehow, when we used `info address` to query the address of `thread_arena`, GDB still gave us a weird address (probably bug of GDB?):
```
Symbol "thread_arena" is at 0x7fcfc93de038 in a file compiled without debugging.
```
So `pwndbg.gdblib.symbol.address` returns `0x7fcfc93de038`, but this is not correct.

----

I'm on GDB 13.1 btw

disconnect3d:
@lebr0nli After https://github.com/pwndbg/pwndbg/pull/1724 we get:
```
pwndbg> pi hex(pwndbg.gdblib.symbol.address("thread_arena"))
Traceback (most recent call last):
  File "<string>", line 1, in <module>
TypeError: 'NoneType' object cannot be interpreted as an integer
Error while executing Python code.
```
Can you look at this?

lebr0nli:
> Can you look at this?

Hmm, I guess this error is expected?
Since we didn't link with pthread, somehow gdb can't find the `thread_arena` without it, so `pwndbg.gdblib.symbol.address("thread_arena")` is `None`:
```
pwndbg> p thread_arena
Cannot find thread-local storage for process 93553, executable file /tmp/a.out:
Cannot find thread-local variables on this target
```
Before #1723 we will get an address instead of `None`, but the address we got is a weird and wrong result.

disconnect3d:
Ah, you are right. The exception is thrown from `hex(...)` not from our code :P.

Closing this then :). Thanks!

PR #1724, opened 2023-05-19
Fallback to `gdb.parse_and_eval` instead of `info address`

### LINKED ISSUE #1628: `pwndbg.gdblib.symbol.address` might return invalid address if vmmap is not working
https://github.com/pwndbg/pwndbg/blob/be306da2553596443d2bdaa378bf11cf59e3eab5/pwndbg/gdblib/symbol.py#L146-L147

Some weird results for thread-local variables might pass this check if our vmmap is not working.
For example, with qemu-user, the result of vmmap is a 0~0xfffff…ffff region, so `not pwndbg.gdblib.vmmap.find(address)` will always be `False`.

lebr0nli:
This PR should resolve #1628 and #1666.

### DIFF
--- a/pwndbg/gdblib/symbol.py
+++ b/pwndbg/gdblib/symbol.py
@@ -143,10 +143,11 @@ def address(symbol: str) -> int:
         # symbols, so we need to fallback to using `gdb.parse_and_eval`. See
         # https://sourceware.org/pipermail/gdb/2022-October/050362.html
         # (We tried parsing the output of the `info address` before, but there were some issues. See #1628 and #1666)
-        if "'" in symbol:
+        if "\\" in symbol:
             # Is it possible that happens? Probably not, but just in case
-            raise ValueError(f"Symbol {symbol!r} contains a single quote")
-        return int(gdb.parse_and_eval(f"&'{symbol}'"))
+            raise ValueError(f"Symbol {symbol!r} contains a backslash")
+        sanitized_symbol_name = symbol.replace("'", "\\'")
+        return int(gdb.parse_and_eval(f"&'{sanitized_symbol_name}'"))
 
     except gdb.error:
         return None
--- a/pwndbg/gdblib/symbol.py
+++ b/pwndbg/gdblib/symbol.py
@@ -15,7 +15,6 @@ import pwndbg.gdblib.arch
 import pwndbg.gdblib.elf
 import pwndbg.gdblib.events
 import pwndbg.gdblib.file
-import pwndbg.gdblib.info
 import pwndbg.gdblib.memory
 import pwndbg.gdblib.qemu
 import pwndbg.gdblib.remote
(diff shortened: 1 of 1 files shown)


pwndbg/gdblib/symbol.py
```diff
@@ -141,13 +140,13 @@ def address(symbol: str) -> int:
...
-
-        return address
+        # (We tried parsing the output of the `info symbol` before, but there were some issues. See #1628 and #1666)
+        if "'" in symbol:
+            # Is it possible that happens? Probably not, but just in case
+            raise ValueError(f"Symbol {symbol!r} contains a single quote")
+        return int(gdb.parse_and_eval(f"&'{symbol}'"))
 
     except gdb.error:
         return None
```
lebr0nli, line 153: Not related to this PR, but just curious: should we remove the code after this line?
Since the code after this line seems unreachable if we return after catching the exception.
I mean this part:
https://github.com/pwndbg/pwndbg/blob/f996aaa2f181720c2664898406446cdc940c0ac7/pwndbg/gdblib/symbol.py#L155-L163

pwndbg/gdblib/symbol.py
```diff
@@ -141,13 +140,13 @@ def address(symbol: str) -> int:
...
         # Unfortunately, `gdb.lookup_symbol` does not seem to handle all
-        # symbols, so we need to fallback to using `info address`. See
+        # symbols, so we need to fallback to using `gdb.parse_and_eval`. See
         # https://sourceware.org/pipermail/gdb/2022-October/050362.html
-        address = pwndbg.gdblib.info.address(symbol)
-        if address is None or not pwndbg.gdblib.vmmap.find(address):
-            return None
-
-        return address
+        # (We tried parsing the output of the `info symbol` before, but there were some issues. See #1628 and #1666)
```
gsingh93: You mean `info address`, right?

pwndbg/gdblib/symbol.py
```diff
@@ -141,13 +140,13 @@ def address(symbol: str) -> int:
...
+        if "'" in symbol:
+            # Is it possible that happens? Probably not, but just in case
```
disconnect3d, line 147: Maybe in Rust with lifetime specifiers?

pwndbg/gdblib/symbol.py
disconnect3d: Good question. Probably yes, but also would be nice to validate the state of our IDA sync/integration 🤐

pwndbg/gdblib/symbol.py
lebr0nli: Oh, cool, I didn't know that.
How about check `\` instead, and sanitize all `'` to `\'`?
Is it possible that we need to use `\` in the symbol name?
Or any possibility that some input can escape from this check?

pwndbg/gdblib/symbol.py
lebr0nli: Ahh, lol 😂
Yes, I mean `info address`, thanks!
I'll change the title and the comment.

pwndbg/gdblib/symbol.py
disconnect3d: I dont think symbols use `\` and we can always fix it if a bug with this appears... ;D

pwndbg/gdblib/symbol.py
disconnect3d: I'm actually fine with removing this code and seeing what happens. If things break, we can bring it back along with a proper test for it?

disconnect3d:
@lebr0nli should we merge this at this point?

pwndbg/gdblib/symbol.py
lebr0nli: Ok, I'll remove them for now.

lebr0nli:
> should we merge this at this point?

Sorry for the late reply.
Yes, I think we could merge it if there's no issue introduced by my new commits.

PR #1725, opened 2023-05-20
Support slab command for all available cpus

clark-ja:
This PR adds support for all available cpus to the `slab` commands.

image

- Add `pwndbg.gdblib.kernel.nproc()` to retrieve number of available cores/processors
- Add `pwndbg.gdblib.kernel.slab.slab_caches` to retrieve a generator of `CpuCache` for all available cores
- Slightly modify output of `slab info` to also include the cpu number
- Add test to check if `pwndbg.gdblib.kernel.nproc()` raises an exception
- Add capability for tests to check if output was generated for all cores

The PR does not actually add test for multiple cores to the CI as this would involve adding the qemu arguments `-smp cpus=N` and I wasn't sure whether adding these qemu configurations and then re-running all tests makes sense as it would double the number of systems to test. I have instead tested it locally for now with 1, 4, 8 and 16 cores and it worked without issues.

For the future it would be nice to modify the qemu test infrastructure in a way that single tests can be run with different qemu configurations (instead of all of them).

It might also be interesting to rework the commandline flags of `slab info` to give the user better control over what she/he wants to see (e.g. `--cpu=N`)

### DIFF
--- a/pwndbg/commands/slab.py
+++ b/pwndbg/commands/slab.py
@@ -109,7 +109,7 @@ def print_slab(slab: Slab, indent, verbose: bool) -> None:
 
 
 def print_cpu_cache(cpu_cache: CpuCache, verbose: bool, indent) -> None:
-    indent.print(f"{C.green('Per-CPU Data')} @ {_yx(cpu_cache.address)}:")
+    indent.print(f"{C.green('kmem_cache_cpu')} @ {_yx(cpu_cache.address)} [CPU {cpu_cache.cpu}]:")
     with indent:
         indent.print(f"{C.blue('Freelist')}:", _yx(int(cpu_cache.freelist)))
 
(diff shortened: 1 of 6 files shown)


pwndbg/commands/slab.py
```diff
@@ -157,9 +157,8 @@ def slab_info(name: str, verbose: bool) -> None:
         indent.print(f"{C.blue('Object Size')}: {slab_cache.object_size}")
 
         # TODO: Handle multiple CPUs
```
clark-ja:
```suggestion
```

gsingh93:
This looks great, thanks!

> For the future it would be nice to modify the qemu test infrastructure in a way that single tests can be run with different qemu configurations (instead of all of them).

Agreed, makes sense.

PR #1726, opened 2023-05-20
Filter commands by category

clark-ja:
I was annoyed to always scroll through all available `pwndbg` commands if I forgot the exact name of a specific heap or kernel command, so this tiny PR allows to filter the commands by their categories.

image


clark-ja:
And why not throw in an additional command to list the available categories? :)

image

disconnect3d:
Awesome, thanks! :)

Issue #1208, opened 2022-10-03
Cleanup heap decorators


gsingh93:
The decorators we're using are a bit redundant, for example:
```python
@pwndbg.commands.OnlyWhenRunning
@pwndbg.commands.OnlyWithResolvedHeapSyms
@pwndbg.commands.OnlyWhenHeapIsInitialized
@pwndbg.commands.OnlyWithTcache
```

- If the heap is initialized, then we can assume we're running the target binary.
- If a command requires the tcache to exist, then it should probably mean the heap is initialized
- I'm not sure if we can combine `OnlyWhenHeapIsInitialized` and `OnlyWithResolvedHeapSyms`. Do we ever use symbols before the heap is initialized? If not, then one can imply the other in this case as well.

dmur1:
pull request doesn't contain the bit you were unsure about because i'm not sure about it either

> I'm not sure if we can combine OnlyWhenHeapIsInitialized and OnlyWithResolvedHeapSyms. Do we ever use symbols before the heap is initialized? If not, then one can imply the other in this case as well.

take a look and see what you think

gsingh93:
> Do we ever use symbols before the heap is initialized?

@disconnect3d @lebr0nli @CptGibbon do you know the answer to this?

lebr0nli:
> do you know the answer to this?

I think for `arena` and `mp` commands (which relied on `main_arena`, `thread_arena`, and `mp_` symbols), they probably should be allowed to be used before the heap is initialized since these symbols should already be there even the heap is not initialized, and we are just printing the content of them when executing the commands.
So maybe we could remove that `OnlyWhenHeapIsInitialized` for them.

gsingh93:
Fixed in #1736 (leaving the heap syms one as is for now is fine).

PR #1730, opened 2023-05-20
Cleanup heap decorators #1208

dmur1:
* Remove use of OnlyWhenRunning when we already have OnlyWhenHeapIsInitialized

* Remove use of OnlyWithTcache when we already have OnlyWhenHeapIsInitialized

* Remove OnlyWithTcache decorator as its no longer used anywhere

### DIFF
--- a/pwndbg/commands/argv.py
+++ b/pwndbg/commands/argv.py
@@ -23,7 +23,7 @@ parser.add_argument(
 )
 
 
-@pwndbg.commands.ArgparsedCommand(parser, aliases=["args"], category=CommandCategory.LINUX)
+@pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.LINUX)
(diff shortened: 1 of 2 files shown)


disconnect3d (changes requested):
With the `OnlyWithTcache` its the other way around: if the tcache is there, we don't need to check if heap is initialized because the `has_tcache` function is checking for heap being initialized:
```py
1614:    def has_tcache(self):
1615-        # TODO/FIXME: Can we determine the tcache_bins existence more reliable?
1616-
1617-        # There is no debug symbols, we determine the tcache_bins existence by checking glibc version only
1618-        return self.is_initialized() and pwndbg.glibc.get_version() >= (2, 26)
```

(So we should not remove it but remove the OnlyIfHeapIsInitialized decorator calls when tcache is used)

PR #1732, opened 2023-05-22
feat(kernel): Add dmesg-like command

clark-ja:
image

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -627,6 +627,7 @@ def load_commands() -> None:
     import pwndbg.commands.kchecksec
     import pwndbg.commands.kcmdline
     import pwndbg.commands.kconfig
+    import pwndbg.commands.kdmesg
     import pwndbg.commands.killthreads
     import pwndbg.commands.kversion
     import pwndbg.commands.leakfind
--- a/pwndbg/commands/kdmesg.py
+++ b/pwndbg/commands/kdmesg.py
@@ -1,5 +1,6 @@
 import argparse
 from typing import Dict
+from typing import Filter
 from typing import Generator
 from typing import List
 from typing import Optional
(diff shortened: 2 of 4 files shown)


disconnect3d:
Hmm:
```
____________________ ERROR collecting tests/test_cymbol.py _____________________
ImportError while importing test module '/pwndbg/tests/gdb-tests/tests/test_cymbol.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib/python3.6/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/test_cymbol.py:3: in <module>
    import pwndbg.commands.cymbol
../../pwndbg/__init__.py:12: in <module>
    load_commands()
../../pwndbg/commands/__init__.py:644: in load_commands
    import pwndbg.commands.kdmesg
../../pwndbg/commands/kdmesg.py:3: in <module>
    from typing import Filter
E   ImportError: cannot import name 'Filter'
```

clark-ja:
I guess I was confused by the `mypy` hints. Previously I had the return type as a `Generator` but `mypy` complained that it is actually a `Filter` and therefore I just assumed that such a type is included in `typing`. Anyhow, I reverted the commit now.

disconnect3d:
Can you rebase to latest dev? There's a commit increasing qemu-tests CI to 30min, maybe it will work then..

clark-ja:
latest `dev` introduces the following mypy error:
```
pwndbg/exception.py:16: error: Name "pdb" already defined (possibly by an import)  [no-redef]
```

don't currently have the time to investigate - might be a couple of days. but this PR is anyway not ready yet so no rush from my side. Also when I have some time I wanted to find out why the runtime of the qemu tests have exploded with the commit which introduces filtering capability of the `dmesg` command (which is actually not even tested yet... so shouldn't affect the test time yet)

gsingh93:
> pwndbg/exception.py:16: error: Name "pdb" already defined (possibly by an import)  [no-redef]

Can you see if upgrading `mypy` fixes this for you? I recently upgraded the dependencies, try running this:
```
pip install -U -r dev-requirements.txt
```

disconnect3d:
Can we just merge this? It seems that this works (?)?

disconnect3d:
Closing this for now since its pretty old :( I'm up for adding such feature

Issue #1288, opened 2022-10-15
Add return address stack location to `context backtrace`


gsingh93:
Currently, `context backtrace` looks like this:
```
 ► f 0   0x7fb8a6fe50b0 __isoc99_scanf
   f 1         0x401330 main+218
   f 2   0x7fb8a6fa6083 __libc_start_main+243
   f 3         0x40119e _start+46
```

I'd find it useful to see the stack location of these return addresses in this output. The `retaddr` command provides this information, so we can probably reuse some code from there to implement this.

(Also the "f" in the output seems useless, I think we can just remove it).

ruijia-zhou:
Hi @gsingh93 , I can work on this enhancement. I wonder if you can assign it to me please? Thanks

disconnect3d:
@ruijia-zhou Assigned.

Btw, somehow related issue: https://github.com/pwndbg/pwndbg/issues/1071

dmur1:
i will look at this as two issues:

first is being able to see the stack address of the return addresses that form the backtrace

second is being able to change the prefix on the backtrace - i would however note that https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/commands/context.py#L769 exists and therefore maybe we don't need to change this ( perhaps only alter the default theme? )

let me know what you folks think

dmur1:
what do you think about using chains for the backtrace?

before:
![image](https://github.com/pwndbg/pwndbg/assets/93072266/327030a5-941b-4ef6-b0f3-9cab9b7d0922)

after:
![image](https://github.com/pwndbg/pwndbg/assets/93072266/ed17d861-bcfb-42e3-aec9-9fa84da0b1d7)

dmur1:
here is a better comparison - although i would note that the second screenshot ( after ) is clearly broken by my wip changes

before:

![image](https://github.com/pwndbg/pwndbg/assets/93072266/9b4f752e-fe93-4b35-a3ab-9ae735300416)

after:

![image](https://github.com/pwndbg/pwndbg/assets/93072266/48e027ab-345d-4fc2-900a-2e0b345178a5)

dmur1:
another idea i had would be to display chain( fp ) + ( stackaddr ) so you get the left most address as the frame pointer still and the stack address is shown but in brackets or something on the right most slot

dmur1:
here's a sample of displaying the address on the stack in parenthesis after compared to the output of retaddr

[...]

PR #1740, opened 2023-05-24
Add return address stack location to context backtrace #1288

dmur1:
* Add pwndbg.gdblib.find_addr_on_stack

* Change coloration of context backtrace to use pwnlib.color.memory

* Remove now defunct address color config from context

* Update test_context_backtrace_show_proper_symbol_names regex

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -800,7 +800,7 @@ def context_backtrace(with_banner=True, target=sys.stdout, width=None):
     bt_prefix = "%s" % pwndbg.gdblib.config.backtrace_prefix
     while True:
         prefix = bt_prefix if frame == this_frame else " " * len(bt_prefix)
-        prefix = " %s" % c.prefix(prefix)
+        prefix = f" {c.prefix(prefix)}"
         frame_pc = frame.pc()
         stackaddr = M.get((pwndbg.gdblib.stack.find_addr_on_stack(frame_pc)))
         addrsz = "(" + stackaddr + ") " + M.get(frame_pc)
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -805,7 +805,7 @@ def context_backtrace(with_banner=True, target=sys.stdout, width=None):
         frame_pc_on_stack = pwndbg.gdblib.stack.find_addr_on_stack(frame_pc)
         padded_text = "0x" + hex(frame_pc_on_stack)[2:].zfill(pwndbg.gdblib.arch.ptrsize * 2)
         stackaddr = M.get(frame_pc_on_stack, padded_text)
-        addrsz = "(" + stackaddr + ") " + M.get(frame_pc)
+        addrsz = f"({stackaddr}) {M.get(frame_pc)}"
         symbol = c.symbol(pwndbg.gdblib.symbol.get(int(frame_pc)))
         if symbol:
             addrsz = addrsz + " " + symbol
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -802,7 +802,9 @@ def context_backtrace(with_banner=True, target=sys.stdout, width=None):
         prefix = bt_prefix if frame == this_frame else " " * len(bt_prefix)
         prefix = f" {c.prefix(prefix)}"
         frame_pc = frame.pc()
-        stackaddr = M.get((pwndbg.gdblib.stack.find_addr_on_stack(frame_pc)))
+        frame_pc_on_stack = pwndbg.gdblib.stack.find_addr_on_stack(frame_pc)
+        padded_text = "0x" + hex(frame_pc_on_stack)[2:].zfill(pwndbg.gdblib.arch.ptrsize * 2)
+        stackaddr = M.get(frame_pc_on_stack, padded_text)
         addrsz = "(" + stackaddr + ") " + M.get(frame_pc)
         symbol = c.symbol(pwndbg.gdblib.symbol.get(int(frame_pc)))
         if symbol:
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -798,12 +798,13 @@ def context_backtrace(with_banner=True, target=sys.stdout, width=None):
     frame = newest_frame
     i = 0
     bt_prefix = "%s" % pwndbg.gdblib.config.backtrace_prefix
+    zero_fill_amount = pwndbg.gdblib.arch.ptrsize * 2
     while True:
         prefix = bt_prefix if frame == this_frame else " " * len(bt_prefix)
         prefix = f" {c.prefix(prefix)}"
         frame_pc = frame.pc()
         frame_pc_on_stack = pwndbg.gdblib.stack.find_addr_on_stack(frame_pc)
-        padded_text = "0x" + hex(frame_pc_on_stack)[2:].zfill(pwndbg.gdblib.arch.ptrsize * 2)
+        padded_text = "0x" + hex(frame_pc_on_stack)[2:].zfill(zero_fill_amount)
         stackaddr = M.get(frame_pc_on_stack, padded_text)
         addrsz = f"({stackaddr}) {M.get(frame_pc)}"
         symbol = c.symbol(pwndbg.gdblib.symbol.get(int(frame_pc)))
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -41,7 +41,6 @@ c = ColorConfig(
     "backtrace",
     [
         ColorParamSpec("prefix", "none", "color for prefix of current backtrace label"),
-        ColorParamSpec("address", "none", "color for backtrace (address)"),
         ColorParamSpec("symbol", "none", "color for backtrace (symbol)"),
         ColorParamSpec("frame-label", "none", "color for backtrace (frame label)"),
     ],
@@ -805,10 +804,13 @@ def context_backtrace(with_banner=True, target=sys.stdout, width=None):
 
         prefix = bt_prefix if frame == this_frame else " " * len(bt_prefix)
         prefix = " %s" % c.prefix(prefix)
-        addrsz = c.address(pwndbg.ui.addrsz(frame.pc()))
-        symbol = c.symbol(pwndbg.gdblib.symbol.get(int(frame.pc())))
+        frame_pc = frame.pc()
+        stackaddr = M.get((pwndbg.gdblib.stack.find_addr_on_stack(frame_pc)))
+        addrsz = "(" + stackaddr + ") " + M.get(frame_pc)
+        symbol = c.symbol(pwndbg.gdblib.symbol.get(int(frame_pc)))
         if symbol:
             addrsz = addrsz + " " + symbol
+
         line = map(str, (prefix, c.frame_label("%s%i" % (backtrace_frame_label, i)), addrsz))
         line = " ".join(line)
         result.append(line)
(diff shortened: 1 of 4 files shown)


pwndbg/commands/context.py
```diff
@@ -41,7 +41,6 @@
     "backtrace",
     [
         ColorParamSpec("prefix", "none", "color for prefix of current backtrace label"),
-        ColorParamSpec("address", "none", "color for backtrace (address)"),
         ColorParamSpec("symbol", "none", "color for backtrace (symbol)"),
```
dmur1, line 44: i wonder if we should use CODE coloration for symbols too? as is the case in the register output for example:

image

pwndbg/gdblib/stack.py
```diff
@@ -136,3 +136,27 @@ def is_executable() -> bool:
             nx = True
 
     return not nx
+
+
+def find_addr_on_stack(addr):
```
dmur1: this method is basically copied from here https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/commands/stack.py#L26

i wonder if both should use this centralised approach?

tests/gdb-tests/tests/test_context_commands.py
```diff
@@ -201,21 +201,19 @@ def test_context_backtrace_show_proper_symbol_names(start_binary):
         == "─────────────────────────────────[ BACKTRACE ]──────────────────────────────────"
     )
 
-    assert re.match(r".*0   0x[0-9a-f]+ A::foo\(int, int\)", backtrace[2])
```
dmur1, line 204: how important is it to tightly fit this regex to the entire output in a test that is mainly of proper symbol names?

pwndbg/gdblib/stack.py
disconnect3d: Yeah I think it makes sense to centralize it. Also I believe there's a small edge case bug, it should be:
```diff
-while addresses and start < sp < stop:
+while addresses and start <= sp < stop:
```

tests/gdb-tests/tests/test_context_commands.py
disconnect3d: The more accurate we are the less regressions we will have :). I am fine with more relaxed assertions: its better to have some tests than none and its hard to test for everything.

pwndbg/gdblib/stack.py
dmur1: i'll make another issue to centralise that in another change i think ( and i'll fix that edge case and note it in the centralisation issue )

pwndbg/commands/context.py
```diff
@@ -801,11 +800,14 @@ def context_backtrace(with_banner=True, target=sys.stdout, width=None):
     bt_prefix = "%s" % pwndbg.gdblib.config.backtrace_prefix
     while True:
         prefix = bt_prefix if frame == this_frame else " " * len(bt_prefix)
-        prefix = f" {c.prefix(prefix)}"
-        addrsz = c.address(pwndbg.ui.addrsz(frame.pc()))
-        symbol = c.symbol(pwndbg.gdblib.symbol.get(int(frame.pc())))
+        prefix = " %s" % c.prefix(prefix)
```
gsingh93: Can you leave this as an f-string? I'm trying to remove all instances of `%s` from the codebase.

pwndbg/commands/context.py
```diff
@@ -801,11 +800,14 @@ def context_backtrace(with_banner=True, target=sys.stdout, width=None):
...
+        frame_pc = frame.pc()
+        stackaddr = M.get((pwndbg.gdblib.stack.find_addr_on_stack(frame_pc)))
+        addrsz = "(" + stackaddr + ") " + M.get(frame_pc)
```
gsingh93: Use an `f-string` here to avoid the concatenation.

pwndbg/commands/stack.py
```diff
@@ -26,7 +26,7 @@ def retaddr() -> None:
     # Find all of them on the stack
     start = stack.vaddr
     stop = start + stack.memsz
-    while addresses and start < sp < stop:
+    while addresses and start <= sp < stop:
```
gsingh93, line 29: I guess this is to cover the case when the return address is pushed to the stack but we haven't pushed a frame pointer yet, so `$rsp` points to the return address and without the `<=` change we wouldn't print it (cc: @disconnect3d )? If so, it would be good to add a test for this.

pwndbg/gdblib/stack.py
gsingh93: Can you add some type hints here?

pwndbg/commands/context.py
```diff
@@ -801,11 +800,14 @@ def context_backtrace(with_banner=True, target=sys.stdout, width=None):
     bt_prefix = "%s" % pwndbg.gdblib.config.backtrace_prefix
     while True:
         prefix = bt_prefix if frame == this_frame else " " * len(bt_prefix)
-        prefix = f" {c.prefix(prefix)}"
-        addrsz = c.address(pwndbg.ui.addrsz(frame.pc()))
-        symbol = c.symbol(pwndbg.gdblib.symbol.get(int(frame.pc())))
+        prefix = " %s" % c.prefix(prefix)
+        frame_pc = frame.pc()
+        stackaddr = M.get((pwndbg.gdblib.stack.find_addr_on_stack(frame_pc)))
```
gsingh93: It's a little hacky, but to deal with the alignment issue you brought up [here](https://github.com/pwndbg/pwndbg/issues/1288#issuecomment-1595693469) we can call the [`M.get`](https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/color/memory.py#L19) function with the optional `text` parameter. We can convert the address to a hex string that's already padded to a width of `pwndbg.gdblib.arch.ptrsize`, use that as the `text` argument, and let `M.get` color it properly.

Also nit: seems like there are extra parenthesis here you can remove.

gsingh93 (changes requested):
Nice work overall, just needs a few minor changes.

pwndbg/commands/context.py
disconnect3d: If we display symbols, probably yes.

disconnect3d:
@dmur1 can u show a screenshot before and after?

pwndbg/commands/context.py
dmur1: done

pwndbg/commands/stack.py
dmur1: pinging @disconnect3d about this one again because i'm not totally sure

pwndbg/gdblib/stack.py
dmur1: done

pwndbg/commands/context.py
dmur1: done

dmur1 (commented):
take a look at the newest stuff!

dmur1:
> @dmur1 can u show a screenshot before and after?

before:
image

after:
image

pwndbg/gdblib/stack.py
```diff
@@ -136,3 +136,27 @@ def is_executable() -> bool:
             nx = True
 
     return not nx
+
+
+def find_addr_on_stack(addr) -> int:
```
disconnect3d, line 141: I am a bit worried that this may be (way) too slow.

* We are reading the whole stack memory: what if stack is super big?
* What if SP does not point to mapped memory? This will crash
* This can be optimized out in two ways:
1) We could leverage the existing search functionality we have which will do a find using GDB's API which is written in C and so should be faster
2) We could read the whole stack memory and use simple `.find(...)` to find the offset of the bytes of the `addr` value

The 2) will rather be slower than 1) but if we read the stack memory just once, then we can find the fetched bytes multiple times to look for different return addresses --- which would probably be faster overall?

disconnect3d: Given this I am not sure if we really want to look for those values on the stack... we are making context slower this way. We can however make this configurable.

@gsingh93 thoughts?

pwndbg/commands/context.py
```diff
@@ -802,10 +801,15 @@ def context_backtrace(with_banner=True, target=sys.stdout, width=None):
     while True:
         prefix = bt_prefix if frame == this_frame else " " * len(bt_prefix)
         prefix = f" {c.prefix(prefix)}"
-        addrsz = c.address(pwndbg.ui.addrsz(frame.pc()))
-        symbol = c.symbol(pwndbg.gdblib.symbol.get(int(frame.pc())))
+        frame_pc = frame.pc()
+        frame_pc_on_stack = pwndbg.gdblib.stack.find_addr_on_stack(frame_pc)
+        padded_text = "0x" + hex(frame_pc_on_stack)[2:].zfill(pwndbg.gdblib.arch.ptrsize * 2)
```
disconnect3d: `hex(...)` returns a `0x...` string, so we don't need to `[2:]` it and prepend a `0x` string

Also nitpick, but we fetch and compute `pwndbg.gdblib.arch.ptrsize * 2` in each loop iteration :<

pwndbg/commands/context.py
```diff
@@ -802,10 +801,15 @@ def context_backtrace(with_banner=True, target=sys.stdout, width=None):
...
+        stackaddr = M.get(frame_pc_on_stack, padded_text)
+        addrsz = "(" + stackaddr + ") " + M.get(frame_pc)
```
disconnect3d: Can we use f-string literals here and wherever we can?

disconnect3d:
> > @dmur1 can u show a screenshot before and after?
> 
> before: image
> 
> after: image

Random thoughts about the display:
1) I don't like the `0x<zeros>` prepended
2) Maybe we should display `<sp register>+0x<OFFSET>` instead of whole addresses?

...or maybe this should be configurable, ugh. cc: @gsingh93

pwndbg/commands/context.py
dmur1: >>> hex(1234).zfill(8)
'0000x4d2'

i was trying to avoid this - perhaps theres a way to zfill an already 0x prefixed string?

@disconnect3d

dmur1: will hoist the computation out of the loop

dmur1:
> > > @dmur1 can u show a screenshot before and after?
> > 
> > 
> > before: image
> > after: image
> 
> Random thoughts about the display:
> 
> 1. I don't like the `0x<zeros>` prepended
> 2. Maybe we should display `<sp register>+0x<OFFSET>` instead of whole addresses?
> 
> ...or maybe this should be configurable, ugh. cc: @gsingh93

i have to admit i'm also not a fan especially as it contrasts with the other aligned addresses in the other context sections

pwndbg/commands/context.py
disconnect3d: I'm not sure about this zfill. Like, do we really want to display those zeroes? :P They are redundant

dmur1:
bumping this - i'm still not sure i like it what do you folks think?

disconnect3d:
I think prefixing with additional zeroes is not great. What else do u ask about? :P

dmur1:
> I think prefixing with additional zeroes is not great. What else do u ask about? :P

so maybe if just padded as much as was needed to keep everything aligned?

disconnect3d:
> > I think prefixing with additional zeroes is not great. What else do u ask about? :P
> 
> so maybe if just padded as much as was needed to keep everything aligned?

Either this, or no padding at all?

disconnect3d:
(This now needs to be rebased and conflicts needs to be fixed :()

disconnect3d:
Closing it for now; I'm still up for accepting such feature in case anyone wants to add it

PR #1752, opened 2023-05-25
Fix banner width for TUI mode with horizontal split for cmd window

ThomsonTan:
With TUI enabled and horizontal layout applied like with below command in GDB, the banner still renders the full width of the current window like below screenshot which makes it occupy 2 rows. This could be fixed by reading the actual cmd window width from the output of `info cmd` command.

```shell
tui new-layout hsrc { -horizontal src 1 cmd 1 } 1
tui layout hsrc
```

image

### DIFF
--- a/pwndbg/ui.py
+++ b/pwndbg/ui.py
@@ -81,6 +81,18 @@ def get_window_size(target=sys.stdin):
 
 
 def get_cmd_window_size():
+    """Get the size of the command window in TUI mode which could be different than the terminal window width \
+    with horizontal split "tui new-layout hsrc { -horizontal src 1 cmd 1 } 1".
+
+    Possible output of "info win" in TUI mode:
+    (gdb) info win
+    Name       Lines Columns Focus
+    src           77     104 (has focus)
+    cmd           77     105
+
+    Output of "info win" in non-TUI mode:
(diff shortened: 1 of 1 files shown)


pwndbg/ui.py
```diff
@@ -67,9 +69,27 @@ def get_window_size(target=sys.stdin):
...
     try:
         # get terminal size and force ret buffer len of 4 bytes for safe unpacking by passing equally long arg
         rows, cols = struct.unpack("hh", fcntl.ioctl(target.fileno(), termios.TIOCGWINSZ, b"1234"))
     except Exception:
         rows, cols = fallback
     return rows, cols
+
+
+def get_cmd_window_size():
+    info_out = gdb.execute("info win", to_string=True).split()
```
disconnect3d, line 96: Can you please add a docstring with all possible outputs of `info win`?

Also, I am pretty sure this is still far from ideal considering GDB translations :(, but maybe we can't deal with that anyhow right now.

disconnect3d: Ping @ThomsonTan

ThomsonTan: Apologize for the late reply. Added doc string with the possible output of `info win` command.

Issue #1754, opened 2023-05-29
gdb.error: Remote 'g' packet reply is too long on system call


SamiKalliomaki:
### Description
pwndbg crashes on successful exploitation of a binary. This does not seem specific to `ret2win` but it is used as a simple example.

### Steps to reproduce
1. Download x86-64 version of ret2win from https://ropemporium.com/challenge/ret2win.html
2. Run this Python code
```
from pwn import *

context.arch = 'amd64'
context.word_size = 64

io = gdb.debug('./ret2win', gdbscript='''
set exception-verbose on
set exception-debugger on
c
''')
io.send(flat({
    40: [0x00400755, 0x00400756]
}) + b'\n')
io.interactive()
```
3. Observe gdb error
```
[Attaching after Thread 23747.23747 vfork to child Thread 23809.23809]                                                                                     
[New inferior 2 (process 23809)]
Reading /usr/lib/debug/.build-id/8f/0bef1909d4329e0773d0ccc5f6decdfe5f4504.debug from remote target...
[Detaching vfork parent process 23747 after child exec]                                                                                                    
[Inferior 1 (process 23747) detached]
process 23809 is executing new program: /usr/bin/bash
Reading /usr/bin/bash from remote target...
Reading /usr/bin/bash from remote target...
Reading /usr/lib/debug/.build-id/94/83da49f2e17070c1df9a75d509e09211e96769.debug from remote target...
Downloading separate debug info for target:/usr/bin/bash
Traceback (most recent call last):                                                                                                                         
  File "/home/user/code/pwndbg/pwndbg/gdblib/events.py", line 156, in caller
    func()
  File "/home/user/code/pwndbg/pwndbg/gdblib/abi.py", line 56, in caller
    return func(*args, **kwargs)
  File "/home/user/code/pwndbg/pwndbg/gdblib/argv.py", line 30, in update
    pwndbg.gdblib.arch_mod.update()  # :-(
  File "/home/user/code/pwndbg/pwndbg/gdblib/arch.py", line 57, in update
    arch_name, ptrsize, endian = _get_arch(typeinfo.ptrsize)
  File "/home/user/code/pwndbg/pwndbg/gdblib/arch.py", line 31, in _get_arch
    if "little" in gdb.execute("show endian", to_string=True).lower():
[...]

PR #1755, opened 2023-05-29
Fix issue where gdb.execute during object loading causes an exception.

SamiKalliomaki:
Please note I have roughly no idea what I am doing and all of this code may be stupid. However, it solves the issue for me.

#1754

### DIFF
--- a/pwndbg/gdblib/arch.py
+++ b/pwndbg/gdblib/arch.py
@@ -33,7 +33,7 @@ def _get_arch(ptrsize: int):
     else:
         endian = "big"
 
-    if pwndbg.gdblib.proc.alive:
+    if pwndbg.gdblib.proc.alive and not gdb.selected_thread().is_running():
         arch = gdb.newest_frame().architecture().name()
     else:
         arch = gdb.execute("show architecture", to_string=True).strip()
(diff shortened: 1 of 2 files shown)


gsingh93:
Thanks for the contribution! Would you be able to add a test case for this issue first so we can better understand the problem and your solution, and make sure this doesn't regress in the future? See https://github.com/pwndbg/pwndbg/blob/dev/DEVELOPING.md#testing

SamiKalliomaki:
Considering this makes a lot tests fail - I suspect this will not be able to merged in the current iteration. I think the issue is that I don't really understand the issue myself that well either.

What I have found is that running `gdb.execute` from this event handler seems to crash. I would guess this is because there is probably another command already in progress. But this might be a GDB bug as well (assuming the issue is not reproducible for everyone).

So right now I'd be looking for feedback on the approach and then we could see if there is a reproducible test that can be implemented.

disconnect3d:
I wonder if this error still happens after all the recent changes and refactorings we did... It could be that this is not an issue anymore? I am not sure. I will close this for now as this is really stale. If u have a moment to double check the issue, I would appreciate that.

PR #1760, opened 2023-06-14
updated canary.py

### LINKED ISSUE #1704: Better display for canary command

Unknownsentinel193:
Fixes #1704 

This PR adds an enhancement to the canary command in canary.py by introducing an --all option. When provided, the canary command will display all canaries found on the stack, including canaries from different threads.

### DIFF
--- a/pwndbg/commands/canary.py
+++ b/pwndbg/commands/canary.py
@@ -13,12 +13,9 @@ def canary_value():
     at_random = auxv.get("AT_RANDOM", None)
     if at_random is None:
         return None, None
-
     global_canary = pwndbg.gdblib.memory.pvoid(at_random)
-
-    # masking canary value as canaries on the stack has last byte = 0
+    # masking canary value as canaries on the stack have the last byte = 0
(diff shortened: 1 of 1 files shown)


pwndbg/commands/canary.py
```diff
@@ -13,41 +13,58 @@ def canary_value():
...
     return global_canary, at_random
 
 
 @pwndbg.commands.ArgparsedCommand(
     "Print out the current stack canary.", category=CommandCategory.STACK
 )
 @pwndbg.commands.OnlyWhenRunning
-def canary() -> None:
+@pwndbg.commands.aliases("canary", "canaries")
+@pwndbg.commands.bool_argument("--all", "-a", "Display all canaries.", default=False)
```
disconnect3d, line 27: Where does this come from...? We don't have such function

disconnect3d:
@Unknownsentinel193 ping

PR #1770, opened 2023-06-26
Improve RISCV support

peace-maker:
This is a resurrection of #829 which adds rich debugging support for RISCV binaries including call argument reconstruction

grafik

branch prediction/emulation.

grafik

It requires bumping the `capstone` dependency to 5.0.0rc4 which adds RISCV support. Only rv32 and rv64 are supported. If rv128 becomes a thing, it'd need to be considered too, but is left out for now.


disconnect3d:
Is there any chance we can add any tests for this?

peace-maker:
I've added a basic test for the arguments display in `nearpc` on a call instruction to the qemu-user tests. I don't think they're run yet though?

pwndbg/gdblib/arch.py
```diff
@@ -19,9 +19,26 @@
     "sparc": "sparc",
     "arm": "arm",
     "armcm": "thumb",
-    "riscv:rv64": "riscv",
 }
 
+# https://github.com/Gallopsled/pwntools/pull/2177
+pwnlib_version = list(map(int, pwnlib.__version__.split(".")[:2]))
+if pwnlib_version[0] == 4 and pwnlib_version[1] < 11:
```
disconnect3d, line 26: We do control the Pwntools version, can we just update it and don't bother about its old implementation?

disconnect3d: Let's create another PR after this that would update pwntools and remove that branch here.

tests/qemu-tests/test_qemu.sh
```diff
@@ -2,13 +2,59 @@
 
 make -C binaries
 
-qemu-aarch64 \
-    -g 1234 \
-    -L /usr/aarch64-linux-gnu/ \
-    ./binaries/reference-binary.aarch64.out &
-
-gdb-multiarch \
-    -ex "file ./binaries/reference-binary.aarch64.out" \
```
disconnect3d, line 11: Are we effectively not having any test that use reference-binary.aarch64.out right now?

disconnect3d: Nvm, we have it later in:
```
    run_gdb \
        -ex "set sysroot /usr/${arch}-linux-gnu/" \
        -ex "file ./binaries/reference-binary.${arch}.out" \
        -ex 'py import coverage;coverage.process_startup()' \
        -ex "target remote :1234" \
        -ex "source ./tests/user/test_${arch}.py"
```

disconnect3d (approved):

Issue #1734, opened 2023-05-23
Add the ability to display libc entries to the GOT command


nobodyisnobody:
it would be useful for got command to be able to print libc (or any other library loaded in memory) GOT entries.

the default behaviour of got command is to print main executable GOT entries, but as full relro is becoming more frequent, often it's interesting to target libc (or other libs) GOT entries, as they are still half relro in many distributions.. 
they reside at the beginning of libc .bss rw zone

an output more in the style of bat24 gef fork would be nice too (I know there is already an opened issue on integrating some of the bat24 changes..)

with PLT / GOT entries side by side..

![https://imgur.com/Nk35qh6.png](https://imgur.com/Nk35qh6.png)

all that would be really usefull for exploitation.

disconnect3d:
I believe this has been implemented in #1771 (@lebr0nli correct me if I am wrong ;))

nobodyisnobody:
nice, great work.

PR #1771, opened 2023-06-28
Enhance got command

lebr0nli:
In this PR, I wrote some enhancements for the `got` command which were inspired by bata24's great work in https://github.com/bata24/gef .

----

Summary of the enhancements:

- The new usage of the `got` command:
```
usage: got [-h] [-p PATH_FILTER | -a] [-r] [symbol_filter]

Show the state of the Global Offset Table.

Examples:
  got
  got puts
  got -p libc
  got -a

positional arguments:
 symbol_filter     Filter results by passed symbol name. (default: )

optional arguments:
 -h, --help      show this help message and exit
 -p PATH_FILTER, --path PATH_FILTER
            Filter results by passed library path. (default: )
 -a, --all       Parse all libraries including the target executable. (default: False)
 -r, --readonly    Set the flag to not filter out read-only entries. (default: False)
```

- We can show loaded libraries entries by something like:
screenshot-2
screenshot-4

- Before, we only show the relocation whose `r_info_type` is `*_JUMP_SLOT`, but now we also show the relocation whose type is `*_IRELATIVE` or `*_GLOB_DAT`. For example, now we have the ability to show some `R_X86_64_IRELATIVE` relocation in libc (e.g. `strlen` used by libc's `puts`):
screenshot-1

- We can filter out or not filter out the read-only entries (default will filter out read-only entries) by:
screenshot-3

---

The enhancements in this PR still have some todos and bugs in some cases(I wrote them in the comment), but this should at least resolve the requirements in #1734 for the ability to display libc's or any other loaded libraries's entries. (Currently, I didn't have a good idea to handle those todos and bugs properly, maybe we can fix them in a new PR if they don’t affect too much :p)

### DIFF
--- a/pwndbg/commands/got.py
+++ b/pwndbg/commands/got.py
@@ -65,7 +65,7 @@ def got(path_filter, all_, accept_readonly, symbol_filter) -> None:
     if path_filter:
         print("Filtering by lib/objfile path: " + message.hint(path_filter))
     if symbol_filter:
-        print("Filtering by symbol keyword: " + message.hint(symbol_filter))
+        print("Filtering by symbol name: " + message.hint(symbol_filter))
     if not accept_readonly:
         print("Filtering out read-only entries (display them with -r or --show-readonly)")
 
--- a/pwndbg/commands/got.py
+++ b/pwndbg/commands/got.py
@@ -47,7 +47,7 @@ group.add_argument(
 )
 parser.add_argument(
     "-r",
-    "--readonly",
+    "--show-readonly",
     help="Set the flag to not filter out read-only entries.",
     action="store_true",
     default=False,
--- a/pwndbg/commands/got.py
+++ b/pwndbg/commands/got.py
@@ -63,7 +63,7 @@ parser.add_argument(
 def got(path_filter, all_, accept_readonly, symbol_filter) -> None:
     # Show the filters we are using
     if path_filter:
-        print("Filtering by path keyword: " + message.hint(path_filter))
+        print("Filtering by lib/objfile path: " + message.hint(path_filter))
     if symbol_filter:
         print("Filtering by symbol keyword: " + message.hint(symbol_filter))
     if not accept_readonly:
--- a/pwndbg/commands/got.py
+++ b/pwndbg/commands/got.py
@@ -10,6 +10,7 @@ import pwndbg.gdblib.arch
 import pwndbg.gdblib.file
 import pwndbg.gdblib.info
 import pwndbg.gdblib.proc
+import pwndbg.gdblib.qemu
 import pwndbg.gdblib.vmmap
 import pwndbg.wrappers.checksec
 import pwndbg.wrappers.readelf
@@ -61,6 +62,11 @@ parser.add_argument(
 @pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.LINUX)
 @pwndbg.commands.OnlyWhenRunning
 def got(path_filter, all_, accept_readonly, symbol_filter) -> None:
+    if pwndbg.gdblib.qemu.is_qemu_usermode():
+        print(
+            "QEMU target detected - the result might not be accurate when checking if the entry is writable and getting the information for libraries/objfiles"
+        )
+        print()
     # Show the filters we are using
     if path_filter:
         print("Filtering by lib/objfile path: " + message.hint(path_filter))
--- a/pwndbg/commands/got.py
+++ b/pwndbg/commands/got.py
@@ -32,7 +32,7 @@ group = parser.add_mutually_exclusive_group()
 group.add_argument(
     "-p",
     "--path",
-    help="Filter results by passed library path.",
+    help="Filter results by library/objfile path.",
     type=str,
     default="",
     dest="path_filter",
--- a/pwndbg/commands/got.py
+++ b/pwndbg/commands/got.py
@@ -54,7 +54,7 @@ parser.add_argument(
     dest="accept_readonly",
 )
 parser.add_argument(
-    "symbol_filter", help="Filter results by passed symbol name.", type=str, nargs="?", default=""
+    "symbol_filter", help="Filter results by symbol name.", type=str, nargs="?", default=""
 )
 
 
--- a/pwndbg/commands/got.py
+++ b/pwndbg/commands/got.py
@@ -40,7 +40,7 @@ group.add_argument(
 group.add_argument(
     "-a",
     "--all",
-    help="Parse all libraries including the target executable.",
+    help="Process all libs/obfjiles including the target executable.",
     action="store_true",
     default=False,
     dest="all_",
--- a/pwndbg/commands/got.py
+++ b/pwndbg/commands/got.py
@@ -48,7 +48,7 @@ group.add_argument(
 parser.add_argument(
     "-r",
     "--show-readonly",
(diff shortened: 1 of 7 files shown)


pwndbg/commands/got.py
```diff
@@ -1,54 +1,171 @@
...
+    help="Parse all libraries including the target executable.",
+    action="store_true",
+    default=False,
+    dest="all_",
+)
 parser.add_argument(
-    "name_filter", help="Filter results by passed name.", type=str, nargs="?", default=""
+    "-r",
+    "--readonly",
+    help="Set the flag to not filter out read-only entries.",
```
disconnect3d: Hmm... `--readonly` sounds as if we want only read-only entries...

disconnect3d: Maybe `--show-ro` or `--show-readonly`?

pwndbg/wrappers/readelf.py
```diff
@@ -1,29 +1,31 @@
...
-def get_jmpslots():
-    local_path = pwndbg.gdblib.file.get_proc_exe_file()
-    cmd = get_jmpslots.cmd + ["--relocs", local_path]
-    readelf_out = pwndbg.wrappers.call_cmd(cmd)
-
-    return filter(_extract_jumps, readelf_out.splitlines())
+class RelocationType(Enum):
+    JUMP_SLOT = 1  # e.g.: R_X86_64_JUMP_SLOT
+    GLOB_DAT = 2  # e.g.: R_X86_64_IRELATIVE
+    IRELATIVE = 3  # e.g.: R_X86_64_GLOB_DAT
```
disconnect3d: Are the comments here correct? GLOB_DAT comment is on the IRELATIVE line etc.

disconnect3d: Any source for those flags? Can we link it?

disconnect3d:
For what is worth, while displaying GOT makes lots of sense, I usually do that with pure `telescope <rw page>`. Also, GOT is not the only thing that contains useful pointers, so ideally you would want to display "all possible code pointers" or hook/break on them somehow...

But anyway, this looks good; just needs small enhancements, see comments

pwndbg/commands/got.py
```diff
@@ -1,54 +1,171 @@
...
+    got puts
+    got -p libc
+    got -a
+""",
+)
+group = parser.add_mutually_exclusive_group()
+group.add_argument(
+    "-p",
+    "--path",
+    help="Filter results by passed library path.",
```
disconnect3d:
```suggestion
    help="Filter results by library/objfile path.",
```

pwndbg/commands/got.py
```diff
@@ -1,54 +1,171 @@
...
+    type=str,
+    default="",
+    dest="path_filter",
+)
+group.add_argument(
+    "-a",
+    "--all",
+    help="Parse all libraries including the target executable.",
```
disconnect3d:
```suggestion
    help="Process all libs/obfjiles including the target executable.",
```

pwndbg/commands/got.py
```diff
@@ -1,54 +1,171 @@
...
-    "name_filter", help="Filter results by passed name.", type=str, nargs="?", default=""
+    "-r",
+    "--readonly",
+    help="Set the flag to not filter out read-only entries.",
+    action="store_true",
+    default=False,
+    dest="accept_readonly",
+)
+parser.add_argument(
+    "symbol_filter", help="Filter results by passed symbol name.", type=str, nargs="?", default=""
```
disconnect3d:
```suggestion
    "symbol_filter", help="Filter results by symbol name.", type=str, nargs="?", default=""
```

pwndbg/commands/got.py
```diff
@@ -1,54 +1,171 @@
...
-def got(name_filter="") -> None:
-    relro_status = pwndbg.wrappers.checksec.relro_status()
-    pie_status = pwndbg.wrappers.checksec.pie_status()
-    jmpslots = list(pwndbg.wrappers.readelf.get_jmpslots())
-    if not jmpslots:
-        print(message.error("NO JUMP_SLOT entries available in the GOT"))
+def got(path_filter, all_, accept_readonly, symbol_filter) -> None:
+    # Show the filters we are using
+    if path_filter:
+        print("Filtering by path keyword: " + message.hint(path_filter))
```
disconnect3d:
```suggestion
        print("Filtering by lib/objfile path: " + message.hint(path_filter))
```

pwndbg/commands/got.py
```diff
@@ -1,54 +1,171 @@
...
+    if symbol_filter:
+        print("Filtering by symbol keyword: " + message.hint(symbol_filter))
```
disconnect3d:
```suggestion
        print("Filtering by symbol name: " + message.hint(symbol_filter))
```

pwndbg/commands/got.py
```diff
@@ -1,54 +1,171 @@
...
+    if not accept_readonly:
+        print("Filtering out read-only entries.")
```
disconnect3d:
```suggestion
        print("Filtering out read-only entries (display them with -r or --show-readonly)")
```

pwndbg/commands/got.py
```diff
@@ -1,54 +1,171 @@
...
+    "--all",
+    help="Parse all libraries including the target executable.",
+    action="store_true",
+    default=False,
+    dest="all_",
+)
 parser.add_argument(
-    "name_filter", help="Filter results by passed name.", type=str, nargs="?", default=""
+    "-r",
+    "--readonly",
```
disconnect3d:
```suggestion
    "--show-readonly",
```

pwndbg/commands/got.py
disconnect3d:
```suggestion
    help="Also display read-only entries (which are filtered out by default).",
```

disconnect3d: Changed to -r --show-readonly.

disconnect3d:
>  (Currently, I didn't have a good idea to handle those todos and bugs properly, maybe we can fix them in a new PR if they don’t affect too much :p)

Can we somehow inform the user that we may be missing some information?

I am going to go ahead and merge this PR if it passes CI. Please open another PR with further enhancements :)

pwndbg/wrappers/readelf.py
lebr0nli: > Are the comments here correct? GLOB_DAT comment is on the IRELATIVE line etc.

Ahh, I messed up the copy-pasting lol. Thanks, I'll fix it.

> Any source for those flags? Can we link it?

Sure, I'll add some links for those example flags.

lebr0nli:
> so ideally you would want to display "all possible code pointers" or hook/break on them somehow...

Yeah, I think having the option to somehow hook/break on those functions might be more helpful than only displaying them, I’ll try to add this kind of feature in the future.

> Can we somehow inform the user that we may be missing some information?

Sure, I’ll add some hints to remind the users.

disconnect3d:
The problem is you rather want to hook on call/jmp to code pointers and
since this may be realized in multiple ways I dont know how we could do sth
reliable here.

disconnect3d:
@lebr0nli can you update tests after my changes so that we can merge this?

lebr0nli:
> can you update tests after my changes so that we can merge this?

No problem. Done :)
(Some failing tests seem to be related to code coverage.)

PR #1774, opened 2023-07-01
Fix glibc-fastbin-bug option of find_fake_fast

### LINKED ISSUE #1773: find_fake_fast --glibc-fastbin-bug option is broken
### Description

Using the `find_fake_fast` option `--glibc-fastbin-bug` always results in an error, at least on 64-bit platforms.
This seems to be because the option causes [only 4 bytes](https://github.com/pwndbg/pwndbg/blob/8d0ccbc3eb94ee91b25c76a1b9e3201e7a70ad82/pwndbg/commands/heap.py#L686) to be read for the size, but then that gets [passed to `unpack()`](https://github.com/pwndbg/pwndbg/blob/8d0ccbc3eb94ee91b25c76a1b9e3201e7a70ad82/pwndbg/commands/heap.py#L761) which expects 8 bytes.

### Steps to reproduce

Run `find_fake_fast -b <address>` in a 64-bit environment and observe error.

### My setup

```
Gdb:      12.1
Python:   3.10.6 (main, May 29 2023, 11:10:38) [GCC 11.3.0]
Pwndbg:   2023.03.19 build: 79ade42
Capstone: 4.0.1024
Unicorn:  2.0.1
```

SlidyBat:
Using the `find_fake_fast --glibc-fastbin-bug` command always resulted in an error, at least on 64-bit platforms. This was because the option caused only 4 bytes to be read for the size, but then that gets passed to unpack() which expects 8 bytes.

Closes #1773

### DIFF
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -758,7 +758,7 @@ def find_fake_fast(
         candidate = search_region[i : i + size_field_width]
 
         if len(candidate) == size_field_width:
-            size_field = pwndbg.gdblib.arch.unpack(candidate)
+            size_field = pwndbg.gdblib.arch.unpack(candidate, size_field_width)
             size_field &= ~(allocator.malloc_align_mask)
 
             if size_field < min_chunk_size or size_field > max_candidate_size:
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -758,7 +758,7 @@ def find_fake_fast(
         candidate = search_region[i : i + size_field_width]
 
         if len(candidate) == size_field_width:
-            size_field = pwndbg.gdblib.arch.unpack(candidate, size_field_width)
+            size_field = pwndbg.gdblib.arch.unpack_size(candidate, size_field_width)
             size_field &= ~(allocator.malloc_align_mask)
 
             if size_field < min_chunk_size or size_field > max_candidate_size:
--- a/pwndbg/lib/arch.py
+++ b/pwndbg/lib/arch.py
@@ -4,7 +4,6 @@ from typing import Optional
 
 from typing_extensions import Literal
 
-
 FMT_LITTLE_ENDIAN = {1: "B", 2: "<H", 4: "<I", 8: "<Q"}
 FMT_BIG_ENDIAN = {1: "B", 2: ">H", 4: ">I", 8: ">Q"}
 
--- a/pwndbg/lib/arch.py
+++ b/pwndbg/lib/arch.py
@@ -5,6 +5,9 @@ from typing import Optional
 from typing_extensions import Literal
 
 
+FMT_LITTLE_ENDIAN = {1: "B", 2: "<H", 4: "<I", 8: "<Q"}
+FMT_BIG_ENDIAN = {1: "B", 2: ">H", 4: ">I", 8: ">Q"}
+
 class Arch:
     def __init__(self, arch_name: str, ptrsize: int, endian: Literal["little", "big"]) -> None:
         self.update(arch_name, ptrsize, endian)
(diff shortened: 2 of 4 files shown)


CptGibbon:
Thank you for this PR.
It looks good, I'll take a closer look this weekend and hopefully it can get merged quickly.

tests/gdb-tests/tests/heap/test_find_fake_fast.py
```diff
@@ -150,3 +150,11 @@ def test_find_fake_fast_command(start_binary):
     result = gdb.execute("find_fake_fast &target_address 0x100", to_string=True)
     check_no_results(result)
     gdb.execute("continue")
+
+    # setup_mem(0xAABBCCDD00000020, 0x8)
+    result = gdb.execute("find_fake_fast &target_address", to_string=True)
+    check_no_results(result)
+
+    result = gdb.execute("find_fake_fast &target_address -b", to_string=True)
```
CptGibbon: Nitpick but perhaps the full `--glibc-fastbin-bug` option here in place of `-b` could make it clearer what this is testing for.

CptGibbon (approved):
Thanks for fixing this, heap-wise this seems to work as intended now.
@disconnect3d any issues with the changes to `pack()` & `unpack()`?

tests/gdb-tests/tests/heap/test_find_fake_fast.py
SlidyBat: Updated

pwndbg/lib/arch.py
```diff
@@ -20,9 +21,10 @@ def update(self, arch_name: str, ptrsize: int, endian: Literal["little", "big"])
         self.ptrmask = (1 << 8 * ptrsize) - 1
         self.endian = endian
 
-        self.fmt = {(4, "little"): "<I", (4, "big"): ">I", (8, "little"): "<Q", (8, "big"): ">Q"}[
-            (self.ptrsize, self.endian)
-        ]
+        if self.endian == "little":
+            self.fmts = {1: "B", 2: "<H", 4: "<I", 8: "<Q"}
+        else:
+            self.fmts = {1: "B", 2: ">H", 4: ">I", 8: ">Q"}
```
disconnect3d: One nitpick since we are already modifying this, can we move the formatting dicts to global namespace? Otherwise, each `update` call will re-create them (sadly, Python does not optimize such literals (yet?))

pwndbg/lib/arch.py
```diff
@@ -31,8 +33,15 @@ def update(self, arch_name: str, ptrsize: int, endian: Literal["little", "big"])
...
 
-    def pack(self, integer: int) -> bytes:
-        return struct.pack(self.fmt, integer & self.ptrmask)
+    def pack(self, integer: int, size: Optional[int] = None) -> bytes:
+        fmt = self._get_fmt(size)
+        return struct.pack(fmt, integer & self.ptrmask)
+
+    def unpack(self, data: bytes, size: Optional[int] = None) -> int:
+        fmt = self._get_fmt(size)
+        return struct.unpack(fmt, data)[0]
```
disconnect3d: Hmm... what about adding `pack_size` and `unpack_size` instead of adding logic that is needed just for a single use case?

disconnect3d:
Ping @SlidyBat see comments please

SlidyBat:
> Ping @SlidyBat see comments please

Sorry been quite busy in the last week.
I'll try to address your comments some time this week, over the weekend by the latest.

pwndbg/commands/heap.py
```diff
@@ -758,7 +758,7 @@ def find_fake_fast(
         candidate = search_region[i : i + size_field_width]
 
         if len(candidate) == size_field_width:
-            size_field = pwndbg.gdblib.arch.unpack(candidate)
+            size_field = pwndbg.gdblib.arch.unpack(candidate, size_field_width)
```
```suggestion
            size_field = pwndbg.gdblib.arch.unpack_size(candidate, size_field_width)
```

disconnect3d:
> > Ping @SlidyBat see comments please
> 
> Sorry been quite busy in the last week. I'll try to address your comments some time this week, over the weekend by the latest.

Np. I went ahead and updated the commented parts committing edits to your branch. Will merge it when it passes CI :).

PR #1780, opened 2023-07-04
Setup.sh use virtualenv

disconnect3d:
The `setup.sh` script now creates a `.venv` directory during execution and installs all dependencies into that directory. Then, `gdbinit.py` will adds the proper `site-packages` directory as the first item of `sys.path`.

Users can overwrite the path where venv is created by setting a `PWNDBG_VENV_PATH` environment variable. This envvar is set in Dockerfile to simplify development experience.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+setuptools-rust; python_version < '3.7'
 capstone==5.0.0rc4
 psutil==5.9.5
 pwntools==4.10.0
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -1,3 +1,4 @@
+setuptools-rust; python_version < '3.7'
```
disconnect3d, line 1: Note: needed for Ubuntu 18.04 and its old Python 3.6

PR #1783, opened 2023-07-05
CI: try to bring back Arch Linux CI

disconnect3d:
Let's see if this works now with virtualenv setup


disconnect3d:
Interestingly, we are failing on some bogus GH actions error: https://gitlab.archlinux.org/archlinux/archlinux-docker/-/issues/76

PR #1789, opened 2023-07-06
Improvements to spray command

bog2n:
* Changed cyclic_gen() function to cyclic(), so it can (theoritically) spray entire memory space
* Added `--value` parameter
* Made code more clear, and error commands nicer

### DIFF
--- a/pwndbg/commands/spray.py
+++ b/pwndbg/commands/spray.py
@@ -1,26 +1,57 @@
 import argparse
 
 import gdb
-from pwnlib.util.cyclic import cyclic_gen
+from pwnlib.util.cyclic import cyclic
 
+import pwndbg.color.message as M
 import pwndbg.commands
 
-parser = argparse.ArgumentParser(description="Spray memory with a cyclic_gen() generated values")
+parser = argparse.ArgumentParser(description="Spray memory with cyclic() generated values")
 parser.add_argument("addr", help="Address to spray")
-parser.add_argument("length", nargs="?", help="Length of byte sequence", type=int, default=0)
+parser.add_argument(
+    "length",
+    help="Length of byte sequence, when unspecified sprays until the end of vmmap which address belongs to",
+    type=int,
+    nargs="?",
+    default=0,
+)
+parser.add_argument("--value", help="Value to spray memory with", type=str, required=False)
 
 
 @pwndbg.commands.ArgparsedCommand(parser)
-def spray(addr, length) -> None:
+def spray(addr, length, value) -> None:
     if length == 0:
-        try:
-            last_addr = next(p for p in pwndbg.gdblib.vmmap.get() if addr in p).end
-        except StopIteration:
-            last_addr = addr + 1  # Throw exception later when trying to write
+        last_addr = addr
+        for p in pwndbg.gdblib.vmmap.get():
+            if addr in p:
+                last_addr = p.end
+                break
         length = last_addr - int(addr)
 
-    valueBytes = cyclic_gen().get(length)
+        if length == 0:
+            print(M.error("Invalid address, can't find end of vmmap"))
+            return
+
+    value_bytes = b""
+
+    if value:
+        if value.startswith("0x"):
+            strlen = len(value[2:])
+            if strlen & 1:  # 0x1, 0x123...
+                strlen += 1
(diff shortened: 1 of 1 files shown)


pwndbg/commands/spray.py
```diff
@@ -1,26 +1,57 @@
...
     if length == 0:
-        try:
-            last_addr = next(p for p in pwndbg.gdblib.vmmap.get() if addr in p).end
-        except StopIteration:
-            last_addr = addr + 1  # Throw exception later when trying to write
+        last_addr = addr
+        for p in pwndbg.gdblib.vmmap.get():
+            if addr in p:
+                last_addr = p.end
+                break
```
gsingh93: I think `pwndbg.gdblib.vmmap.find(addr)` might work here.

pwndbg/commands/spray.py
```diff
@@ -1,26 +1,57 @@
...
+            return
+
+    value_bytes = b""
+
+    if value:
+        if value.startswith("0x"):
+            strlen = len(value[2:])
+            if strlen & 1:  # 0x1, 0x123...
+                strlen += 1
+            value_bytes = int(value, 16).to_bytes(int(strlen / 2), byteorder="big")
```
gsingh93: See if this works instead:
```suggestion
            value_bytes = int(value, 16).to_bytes((len(value[2:]) + 1) // 2, byteorder="big")
```

pwndbg/commands/spray.py
```diff
@@ -1,26 +1,57 @@
...
+        else:
+            value_bytes = bytes(value, "utf-8")
+
+        value_length = len(value_bytes)
+        value_bytes = value_bytes * int(length / value_length)
```
gsingh93:
```suggestion
        value_bytes = value_bytes * (length // value_length)
```

pwndbg/commands/spray.py
```diff
@@ -1,26 +1,49 @@
...
 @pwndbg.commands.ArgparsedCommand(parser)
-def spray(addr, length) -> None:
+def spray(addr, length, value) -> None:
     if length == 0:
-        try:
-            last_addr = next(p for p in pwndbg.gdblib.vmmap.get() if addr in p).end
-        except StopIteration:
-            last_addr = addr + 1  # Throw exception later when trying to write
-        length = last_addr - int(addr)
+        last_page = pwndbg.gdblib.vmmap.find(addr)
```
disconnect3d: its just `page` not `last_page`

pwndbg/commands/spray.py
```diff
@@ -1,26 +1,49 @@
...
 parser.add_argument("addr", help="Address to spray")
-parser.add_argument("length", nargs="?", help="Length of byte sequence", type=int, default=0)
+parser.add_argument(
+    "length",
+    help="Length of byte sequence, when unspecified sprays until the end of vmmap which address belongs to",
+    type=int,
+    nargs="?",
+    default=0,
+)
+parser.add_argument("--value", help="Value to spray memory with", type=str, required=False)
```
disconnect3d: A 'value' to spray memory with sounds as if this is maybe an integer and e.g.:
- 0x41 -> 41 41 41 41 ...
- 0x4142 -> 41 42 41 42 ...
- 0x414243 -> 41 42 43 41 42 43 ... (likely not much usable)

PS: I probably messed up endianess :shrug:
(I'd use little endian here? or: fetch target endianess?)

disconnect3d: @gsingh93 wdyt?

disconnect3d:
Pls address lint issues, we can update --value later if we decide to change it :)

pwndbg/commands/spray.py
gsingh93: What if we say "hex byte string" instead, and the user would put `41`, `4142`, `414243`, etc? I think this would fix the confusion around endianness, and it would slightly simplify some code below if we don't need the `0x`.

Later we can also have a flag to accept a byte string instead of a hex byte string, so you could do "A", "AB", "ABC" instead.

bog2n: I think current implementation is fine, because you can put binary data and bytes without additional flag, we just need to tell user that hex strings are written in big endian order

Or add flag for endianess which will default to target's

Issue #1787, opened 2023-07-06
GH actions - save-state is deprecated


disconnect3d:
Our GH actions builds warn us about this:

> The `save-state` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

This can be seen here: https://github.com/pwndbg/pwndbg/actions/runs/5470271945

intrigus-lgtm:
Easiest way to fix this would be to switch from https://github.com/satackey/action-docker-layer-caching to https://github.com/jpribyl/action-docker-layer-caching in https://github.com/pwndbg/pwndbg/actions/runs/5470271945/workflow#L18
See: https://github.com/satackey/action-docker-layer-caching/issues/347

disconnect3d:
@intrigus-lgtm iiuc this was closed in https://github.com/pwndbg/pwndbg/pull/1794

(Please let me know in case I am wrong :P). Thanks!

PR #1794, opened 2023-07-12
Use fork of action-docker-layer-caching to prevent warnings

intrigus-lgtm:
https://github.com/satackey/action-docker-layer-caching hasn't been updated in quite some time. E.g. GitHub has started to complain about the deprecated `save-state` command.

Solve this by switching to a newer fork.


intrigus-lgtm:
Fixes #1787

PR #1795, opened 2023-07-12
Add a `plist` command to print linked lists

### LINKED ISSUE #1559: Add a command to dump linked lists
A common issue I have is needing to dump linked lists in memory. This is possible in Python, but it's complicated:
```python
[print(pwndbg.gdblib.memory.poi(gdb.lookup_type('struct _IO_FILE'), x)) for x in pwndbg.chain.get(int(gdb.lookup_symbol('_IO_list_all')[0].value()['file'].address), 2, 104)]
```
Ideally this could be simplified into a command that would look like:
```
chain _IO_list_all.file _chain
```
as long as type information was available.

mbrla0:
Fixes #1559.

This change adds a command that traverses the linked list beginning at a given element, dumping its contents and the contents of all the elements that come after it in the list. Traversal is configurable and can handle multiple types of chains.

The current implementation supports traversing lists where the next pointer is contained in the structure itself, as well as one where the next pointer is stored inside of a nested structure (such as the Linux kernel's `list_head` structure).

**(EDIT: Modified the title of this since we then renamed the 'chain' command to 'plist' (disconnect3d))**

### DIFF
--- a/pwndbg/commands/chain.py
+++ b/pwndbg/commands/chain.py
@@ -1,0 +1,393 @@
+import argparse
+
+import gdb
+
+import pwndbg.chain
+import pwndbg.commands
+import pwndbg.gdblib.memory
+from pwndbg.color import message
+
+parser = argparse.ArgumentParser(
+    formatter_class=argparse.RawTextHelpFormatter,
+    description="""Dumps the elements of a linked list.
+
+This command traverses the linked list beginning at a given element, dumping its
+contents and the contents of all the elements that come after it in the list.
+Traversal is configurable and can handle multiple types of chains, but will
+always stop when a cycle is detected.
+
+The path to the first element can be any GDB expression that evaluates to either
+the first element directly, or a to pointer to it. The next element is the name
+of the field containing the next pointer, in either the structure itself or in
+the structure given by --inner.
+
+An address value may be given with --sentinel that signals the end of the list.
+By default, the value used is NULL (0).
+
+If only one field inside each node is desired, it can be printed exclusively by
+specifying its name with --field.
+
+This command supports traversing three types of linked lists, classified by how
+the next pointer can be found in the structure and what type it is:
+    1 - Next pointer is field of structure, type is the same as structure.
+    2 - Next pointer is field of inner nested structure, pointed to type is the
+        same as outer structure.
+    3 - Next pointer is field of inner nested structure, pointed to type is the
+        same as inner structure.
+Types 2 and 3 require --inner to be specified.
+
+Example 1:
+
+```
+struct node {
+    int value;
+    struct node *next;
+};
+struct node node_c = { 2, NULL };
+struct node node_b = { 1, &node_c };
+struct node node_a = { 0, &node_b };
+```
+
+pwndbg> chain node_a next
+0x4000011050 <node_a>: {
+  value = 0,
+  next = 0x4000011040 <node_b>
+}
+0x4000011040 <node_b>: {
+  value = 1,
+  next = 0x4000011010 <node_c>
+}
+0x4000011010 <node_c>: {
+  value = 2,
+  next = 0x0
+}
+
+Example 2:
+
+```
+struct node_inner_a {
+    struct node_inner_a *next;
+};
+struct inner_a_node {
+    int value;
+    struct node_inner_a inner;
+};
+struct inner_a_node inner_a_node_c = { 2, { NULL } };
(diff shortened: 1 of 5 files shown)


pwndbg/commands/chain.py
```diff
@@ -0,0 +1,394 @@
...
+
+    # Helper functions to resolve the offsets of fields inside a structure.
+    def bit_offset_of_field(struct, field_name, inner_name=None):
+        offset = None
+        for field in struct.type.fields():
+            if field.name == field_name:
+                offset = field.bitpos
+        return offset
+
+    def get_byte_offset(bit_offset):
```
disconnect3d: Can we move the two helper functions out of the command function?

Defining a nested function in Python (CPython) creates this function dynamically on each execution of the outer function. This can be observed by inspecting the bytecode generated for an outer function:

```py
 In [4]: def foo():
   ...:     def bar():
   ...:         pass
   ...: 

In [5]: dis.dis(foo)
  2           0 LOAD_CONST               1 (<code object bar at 0x7f605d76a550, file "<ipython-input-4-3215c711badf>", line 2>)
              2 LOAD_CONST               2 ('foo.<locals>.bar')
              4 MAKE_FUNCTION            0
              6 STORE_FAST               0 (bar)
              8 LOAD_CONST               0 (None)
             10 RETURN_VALUE

Disassembly of <code object bar at 0x7f605d76a550, file "<ipython-input-4-3215c711badf>", line 2>:
  3           0 LOAD_CONST               0 (None)
              2 RETURN_VALUE
```

pwndbg/commands/chain.py
```diff
@@ -0,0 +1,394 @@
...
+    "-f",
+    "--field",
+    dest="field_name",
+    type=str,
+    help="The name of the field to be displayed, if only one is desired",
+)
+
+
+@pwndbg.commands.ArgparsedCommand(parser, command_name="chain")
+def chain_cmd(path, next, sentinel, inner_name, field_name) -> None:
```
disconnect3d: Is there any reason we don't just name the function `chain`? :)

```suggestion
def chain(path, next, sentinel, inner_name, field_name) -> None:
```

pwndbg/commands/chain.py
```diff
@@ -0,0 +1,394 @@
...
+
+    if first.type.code == gdb.TYPE_CODE_PTR:
+        print(message.error(f"{path} is not a value or a single pointer to one"))
+        return
+
+    if first.address is None:
+        print(message.error(f"{deref}{path} is not addressable"))
+
+    if first.is_optimized_out:
+        print(message.error(f"{deref}{path} has been optimized out"))
```
disconnect3d, line 207: Shouldn't those two branches `return` too? Or can we still handle such cases?

pwndbg/commands/chain.py
```diff
@@ -0,0 +1,394 @@
...
+                target_type = field_type
+                target_address = address + field_offset
+
+            value = pwndbg.gdblib.memory.poi(target_type, target_address)
+
+            symbol = pwndbg.gdblib.symbol.get(target_address)
+            symbol = f"<{symbol}>" if symbol else ""
+            address = "%#x" % target_address
+
+            print(f"{address} {symbol}: {value}")
```
disconnect3d, line 386: Maybe:
```py
            value = pwndbg.gdblib.memory.poi(target_type, target_address)
            symbol = pwndbg.gdblib.symbol.get(target_address)

            print(f"{address:#x} {"<{symbol}>" if symbol else ""}: {value}")

File: pwndbg/commands/chain.py [hunk 5]
```diff
@@ -0,0 +1,394 @@
...
+        except gdb.error as e:
+            print(
+                message.error(
+                    f"Cannot dereference 0x{'{:x}'.format(address)} for chain link #{i + 1}: {e}"
```
disconnect3d:
```suggestion
                    f"Cannot dereference {address:#x} for chain link #{i + 1}: {e}"
```

pwndbg/commands/chain.py
disconnect3d: Could use `f"{target_address:#x} ..."`

pwndbg/commands/chain.py
mbrla0: Yes, I've moved them now. Guess I'm so used to this being a non-issue in Rust that the possibility didn't even cross my mind.

pwndbg/commands/chain.py
mbrla0: Nope. :)
And fixed.

PR #1796, opened 2023-07-12
Fix/ignore venv restrict bind mount

intrigus-lgtm:
Bind mounting `.` in every case would interfere with .dockerignore
We want to ignore `.venv` so that the venv of the built docker image
is used. Otherwise we would use the venv of the host inside docker.
This would negate the whole point of testing in a docker container.

Bind mounting `.` is however useful if one wants to use docker just
for "sandboxing" while running the tests on the local machine.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+# venv
+.venv/
--- a/.gitignore
+++ b/.gitignore
@@ -84,3 +84,6 @@ tests/qemu-tests/images
 
 .zig/
 gdb.txt
+
+# venv
+.venv/
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,14 +1,16 @@
 version: "3.8"
 services:
-  main: &base-spec
+  base: &base-spec
     build: .
-    volumes:
-      - .:/pwndbg
     platform: linux/amd64
     security_opt:
       - seccomp:unconfined
     cap_add:
       - SYS_PTRACE
+  main:
+    <<: *base-spec
+    volumes:
+      - .:/pwndbg
 
   ubuntu18.04:
     <<: *base-spec


intrigus-lgtm:
For future readers:
I don't think I actually removed the bind mount due to actual problems with `.venv`.
That is, this paragraph is wrong-ish:
> Otherwise we would use the venv of the host inside docker.
This would negate the whole point of testing in a docker container.

Instead, I wanted to ignore the `.venv` folder, because it is somewhat big and there is no point in adding it, if it's not even used.
However, bind mounts ignore `.dockerignore` by design, so I removed the bind mounts.

This commit is also related to https://github.com/pwndbg/pwndbg/commit/f7cb2f9cb9c6e8685da7564f1aab5ca5f6a12cf8
specifically this part:
> * Do not copy binaries from host to docker
>
> The `Dockerfile` copies the whole pwndbg folder to the image.
If we have built binaries on the host before, these binaries will contain references to
the host system and *copied*  to the image.
If we now run `context code` (inside docker) to have a look at the source code this will
fail, because we will try to refer to a path on the host system.

Additionally this means that running certain test binaries may fail completely, because we've built them on a modern host (e.g. arch) and want to try them on and old Ubuntu version.
This will most likely fail due to GLIBC incompatibilities.

intrigus-lgtm:
> For future readers: I don't think I actually removed the bind mount due to actual problems with `.venv`. That is, this paragraph is wrong-ish:
> 
> > Otherwise we would use the venv of the host inside docker.
> > This would negate the whole point of testing in a docker container.
> 

Actually, turns out my commit message **was** accurate for the _arch linux_ Dockerfile which does **NOT** set the `PWNDBG_VENV_PATH` variable:
https://github.com/pwndbg/pwndbg/blob/4191ceb5ec0b19c8f1dac6cacee0d1924f6b9560/Dockerfile.arch#L16-L19

Compared to apt-based distributions:
https://github.com/pwndbg/pwndbg/blob/4191ceb5ec0b19c8f1dac6cacee0d1924f6b9560/Dockerfile#L16-L19

PR #1797, opened 2023-07-13
Stop using `CS_GRP_CALL` directly to determine what's a call

### LINKED ISSUE #1419: `nextcall` is broken on ARM
`nextcall`/`stepcall` don't work on ARM. In `break_next_call` in `pwndbg/gdblib/next.py`, we check for calls like this:
```
        if capstone.CS_GRP_CALL not in ins.groups:
            continue
```
Unfortunately, the `bl` instruction is only classified as `branch_relative` and `jump`, but not `call`. We should probably report this upstream to Capstone, though I have a feeling this intentional.

In any case, we'll have to directly check the mnemonic matches 'bl' on ARM.

### LINKED ISSUE #1554: Function arguments aren't displayed in disassembly for ARM
On x86, if a function is about to be called you see the argument values like this:
```
 ► 0x7ffff7fe32b3 <_start+3>             call   _dl_start                <_dl_start>
        rdi: 0x7fffffffbd90 ◂— 0x1
```

This doesn't seem to work on ARM. I haven't dug into this, but my guess is the root cause is the same as https://github.com/pwndbg/pwndbg/issues/1419, `bl` is only classified as a jump and not a call in `capstone`.

mbrla0:
Fixes #1554.
Fixes #1419.

This commit introduces a new function (`is_call`) to `pwndbg.disasm`, which centralizes the mechanism used to decide what instructions should be considered calls and which ones shouldn't. Additionally, it uses this new mechanism to flag the `bl` family of instructions on ARM as calls, which, as of the time of writing, are not flagged as such by Capstone.

### DIFF
--- a/pwndbg/arguments.py
+++ b/pwndbg/arguments.py
@@ -3,7 +3,6 @@ Allows describing functions, specifically enumerating arguments which
 may be passed in a combination of registers and stack values.
 """
 import gdb
-from capstone import CS_GRP_CALL
 from capstone import CS_GRP_INT
 
 import pwndbg.chain
@@ -90,7 +89,7 @@ def get(instruction):
     if instruction.address != pwndbg.gdblib.regs.pc:
         return []
 
-    if CS_GRP_CALL in instruction.groups:
+    if pwndbg.disasm.is_call(instruction):
         try:
             abi = pwndbg.lib.abi.ABI.default()
         except KeyError:
--- a/pwndbg/disasm/__init__.py
+++ b/pwndbg/disasm/__init__.py
@@ -9,6 +9,8 @@ from typing import List
 from typing import Union
 
 import capstone
+import capstone.arm
+import capstone.arm64
 import gdb
 from capstone import *  # noqa: F403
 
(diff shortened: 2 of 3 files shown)


mbrla0:
There's one last check for `CS_GRP_CALL in instruction.groups` in `pwndbg/disasm/arch.py:117` that should probably be made to use this mechanism too, but I don't know what that part of the code is doing well enough to make the call. I'm leaning towards yes but for an overabundance of caution I'd rather ask for a second opinion on it first.

peace-maker:
Are you sure `bl` isn't grouped as a call in capstone? The code suggests [it does](https://github.com/capstone-engine/capstone/blob/8ae8f68ea3abb7a68071908bad02baa39742870c/arch/ARM/ARMMappingInsn.inc#L1871). Maybe this isn't needed with capstone 5?

mbrla0:
Huh. Yeah, you're right. Capstone 5 does group it as a call, while 4 does not. What's the MSV of Capstone in pwndbg?

disconnect3d:
> What's the MSV of Capstone in pwndbg?

What's MSV? Most Stable Version or sth? :)

We pull newer versions when it is needed (or occasionally, if it doesn't break things).

If latest Capstone fixes this issue, let's just update it

disconnect3d:
Closing this as updating the Capstone should fix the problem that this was about to fix

PR #1798, opened 2023-07-14
Add `stepuntilasm` command

### LINKED ISSUE #1264: Add a stepuntilasm command
We have all the fancy `stepsyscall`, `nextret`, `nextsyscall` instructions: should we maybe have `stepuntilasm` that would step by step until an instruction with given assembly text is reached? We already do disasm instructions when doing the listed commands, so we can have another command to match asm.

In the future, we may want to merge all of the above into one `stepuntil` or just some `until` command that would have flags/args for different scenarios. But for now, lets add an additional command with this cool functionality.

mbrla0:
Fixes #1264.

This change adds a `stepuntilasm` command that, given a mnemonic and, optionally, a set of operands, will step until a instruction that matches both is found. Matching is string-based, as the user will likely want to spell out the asm directive they want as text, and interpreting assembly language conventions for all of the platforms pwndbg supports is probably outside the scope of this change.

There is a caveat to this implementation in that it can be fairly slow to step over loops, as execution has to be stopped at every branch instruction, but I don't believe there's much that can be done about that.

### DIFF
--- a/tests/gdb-tests/tests/test_command_stepuntilasm.py
+++ b/tests/gdb-tests/tests/test_command_stepuntilasm.py
@@ -1,0 +1,23 @@
+import gdb
+
+import pwndbg.gdblib
+import tests
+
+STEPUNTILASM_X64_BINARY = tests.binaries.get("stepuntilasm_x64.out")
+
+
+def test_command_untilasm_x64(start_binary):
+    start_binary(STEPUNTILASM_X64_BINARY)
+    gdb.execute("break break_here")
+    gdb.execute("run")
+
+    run_and_verify("stop1", "nop")
+    run_and_verify("stop2", "xor rax, rax")
+    run_and_verify("stop3", "mov qword ptr [rax], 0x20")
+    run_and_verify("stop4", "mov dword ptr [rax+4], 0x20")
+
+
+def run_and_verify(stop_label, asm):
+    gdb.execute(f"stepuntilasm {asm}")
+    address = int(gdb.parse_and_eval(f"&{stop_label}"))
+    assert pwndbg.gdblib.regs.pc == address
--- a/pwndbg/commands/next.py
+++ b/pwndbg/commands/next.py
@@ -132,3 +132,22 @@ def stepsyscall() -> None:
 
     if pwndbg.gdblib.proc.alive:
         pwndbg.commands.context.context()
+
+
+parser = argparse.ArgumentParser(description="Breaks on the next matching instruction.")
+parser.add_argument("mnemonic", type=str, help="The mnemonic of the instruction")
+parser.add_argument(
+    "op_str",
+    type=str,
+    nargs="*",
+    help="The operands of the instruction",
+)
+
+
+@pwndbg.commands.ArgparsedCommand(parser, command_name="stepuntilasm")
+@pwndbg.commands.OnlyWhenRunning
+def stepuntilasm(mnemonic, op_str):
+    if len(op_str) == 0:
+        op_str = None
+
+    pwndbg.gdblib.next.break_on_next_matching_instruction(mnemonic, op_str)
--- a/pwndbg/gdblib/next.py
+++ b/pwndbg/gdblib/next.py
@@ -91,7 +91,7 @@ def next_matching_until_branch(address=None, mnemonic=None, op_str=None):
             if isinstance(op_str, str):
                 op_str = "".join(op_str.split()).casefold()
             elif isinstance(op_str, list):
-                op_str = "".join(op.split() for op in op.str).casefold()
+                op_str = "".join(op.split() for op in op_str).casefold()
             else:
                 raise ValueError("op_str value is of an unsupported type")
             op_str_match = ops == op_str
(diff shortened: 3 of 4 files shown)


tests/gdb-tests/tests/test_command_stepuntilasm.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+
+    start_binary(binary)
+    gdb.execute("break break_here")
+    gdb.execute("run")
+
+    run_and_verify("stop1", "nop")
+    run_and_verify("stop2", "xor rax, rax")
+    run_and_verify("stop3", "mov qword ptr [rax], 0x20")
+    run_and_verify("stop4", "mov dword ptr [rax+4], 0x20")
```
disconnect3d, line 17: How does this work with multiple whitespaces? What will happen if the user types in `stepuntilasm xor rax,rax` or `stepuntilasm xor  rax , rax`?

Can we maybe handle this somehow? We could maybe take the whole instruction as a string and split on whitespaces and commas?

(Would it make things notably slower? If so, maybe that's not a good idea to handle it)

tests/gdb-tests/tests/test_command_stepuntilasm.py
```diff
@@ -0,0 +1,29 @@
+import gdb
+import pytest
+
+import pwndbg.gdblib
+import tests
+
+STEPUNTILASM_X64_BINARY = tests.binaries.get("stepuntilasm_x64.out")
+
+
+@pytest.mark.parametrize("binary", [STEPUNTILASM_X64_BINARY], ids=["x86-64"])
```
disconnect3d: Do we need to parametrize this?

disconnect3d: (I get the idea that it would be great to do so, but for now we only have x64 here)

tests/gdb-tests/tests/test_command_stepuntilasm.py
```diff
@@ -0,0 +1,29 @@
...
+def test_command_untilasm_x64(start_binary, binary):
+    """
+    Tests the chain for a non-nested linked list
```
disconnect3d: Should be removed :)

disconnect3d (changes requested):
See comments

tests/gdb-tests/tests/test_command_stepuntilasm.py
mbrla0: Oh, we already do! All whitespace other than a single one separating the mnemonic from the operands is ignored. Which means `xor   rax,rax` is `xor     rax ,rax` is `xor  rax , rax` is `xor rax, rax`. Case differences are also ignored. 

This means you can match on `mov qwordptr[rax],0x20` too, but since these combined keywords aren't themselves different valid keywords, it should be fine.

In terms of performance, while it does make it slower, it's still fairly insignificant compared to the time it takes to do the back and forth from GDB in my machine. If you want some numbers for the procedure this patch uses to ignore whitespace and case differences, though, here they are:
```py
def casefold_whitespace():
    for _ in range(100000):
        ops = "".join("rax, rax".split()).casefold()

        tmp = []
        for op in ["rax", ",       ", "     rax"]:
            tmp.extend(op.split())
        op_str = "".join(tmp).casefold()

        a = ops == op_str
```

```
         1100004 function calls in 0.264 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    0.264    0.264 <string>:1(<module>)
        1    0.170    0.170    0.264    0.264 a.py:1(casefold_whitespace)
        1    0.000    0.000    0.264    0.264 {built-in method builtins.exec}
   200000    0.015    0.000    0.015    0.000 {method 'casefold' of 'str' objects}
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}
   300000    0.021    0.000    0.021    0.000 {method 'extend' of 'list' objects}
   200000    0.018    0.000    0.018    0.000 {method 'join' of 'str' objects}
   400000    0.040    0.000    0.040    0.000 {method 'split' of 'str' objects}
```

tests/gdb-tests/tests/test_command_stepuntilasm.py
mbrla0: I might be mistaken on what parametrization does in the pwndbg test suite. What I was trying to achieve is to get this test to only run on x86_64 targets, and be ignored on others. Is that not what it does? Is there a better way to do that?

disconnect3d: https://docs.pytest.org/en/7.3.x/how-to/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions

The `parametrize` decorator allows you to provide testcases for the given test function so that each testcase will be executed as a separate test.

For now, we execute all tests on x86-64 only and whenever there is a need for a different architecture I believe it is handled separately, usually by running it via QEMU-user emulation and connecting to its gdbstub/gdbserver server.

disconnect3d: So we can just remove that parametrize here

mbrla0: Oh, perfect, then. I'll remove the parametrization

disconnect3d (approved):

PR #1799, opened 2023-07-14
Adds `break-if-taken` and `break-if-not-taken`

### LINKED ISSUE #1167: Conditional breakpoints for when jumps are taken
I often want to set conditional breakpoints depending on when jump instructions are or are not taken. For example:
```asm
0x1001 test rax, rax
0x1002 je foo
```

I'd like to be able to do `break-jump-taken 0x1002` (it doesn't really matter what name we pick for the command), and then we would only break on 0x1002 if the jump was taken. I'd also like a similar instruction for when the jump is not taken. And just like the normal GDB `break` command, if I just do `break-jump-taken`, it would apply that breakpoint to the instruction at the current PC (if there is a jump instruction there, otherwise we'd return an error to the user).

My current workaround for this is I know what the test condition is, i.e. when `rax` is or isn't zero, I can add a conditional breakpoint for that instead. But knowing what that condition is always that easy.

I see a green checkmark in the disassembly view when jumps are taken, so it seems like we already have this information, meaning this probably won't be too difficult to implement.

mbrla0:
Fixes #1167.

This commit adds the `break-if-taken` and `break-if-not-taken` commands, which attach breakpoints to branch instructions that will stop the inferior if said branch is taken or is not taken, respectively. It adds an extra class, `pwndbg.gdblib.bpoint.Breakpoint`, which clears caches before calling `stop()`, allowing for the use of register values inside that function in breakpoint classes that derive from it. Additionally, checking of whether the conditions for a branch to be taken have been fulfilled is done through `DisassemblyAssistant.condition()`.

Aside from `DisassemblyAssistang` not supporting `aarch64`, are there any more reservations against using it, or a better way to achieve the same thing?

### DIFF
--- a/pwndbg/commands/branch.py
+++ b/pwndbg/commands/branch.py
@@ -1,0 +1,111 @@
+import argparse
+
+import gdb
+from capstone import CS_GRP_JUMP
+
+import pwndbg.color.message as message
+import pwndbg.commands
+import pwndbg.disasm
+import pwndbg.gdblib.bpoint
+import pwndbg.gdblib.next
+
+
+class BreakOnConditionalBranch(pwndbg.gdblib.bpoint.Breakpoint):
+    """
+    A breakpoint that only stops the inferior if a given branch is taken or not taken.
+    """
+
+    def __init__(self, instruction, taken):
+        super().__init__("*%#x" % instruction.address, type=gdb.BP_BREAKPOINT, internal=False)
+        self.instruction = instruction
+        self.taken = taken
+
+    def should_stop(self):
+        # Use the assistant to figure out which if all the conditions this
+        # branch requires in order to be taken have been met.
+        assistant = pwndbg.disasm.arch.DisassemblyAssistant.for_current_arch()
+        condition_met = assistant.condition(self.instruction)
+        if condition_met is None:
+            # This branch is unconditional.
+            condition_met = 1
+        condition_met = condition_met != 0
+
+        return condition_met == self.taken
+
+
+parser = argparse.ArgumentParser(description="Breaks on a branch if it is taken.")
+parser.add_argument(
+    "branch",
+    type=str,
+    help="The branch instruction to break on.",
+)
+
+
+@pwndbg.commands.ArgparsedCommand(parser, command_name="break-if-taken")
+@pwndbg.commands.OnlyWhenRunning
+def break_if_taken(branch) -> None:
+    install_breakpoint(branch, taken=True)
+
+
+parser = argparse.ArgumentParser(description="Breaks on a branch if it is not taken.")
+parser.add_argument(
+    "branch",
+    type=str,
+    help="The branch instruction to break on.",
+)
+
+
+@pwndbg.commands.ArgparsedCommand(parser, command_name="break-if-not-taken")
+@pwndbg.commands.OnlyWhenRunning
+def break_if_not_taken(branch) -> None:
+    install_breakpoint(branch, taken=False)
+
+
+def install_breakpoint(branch, taken):
+    # Do our best to interpret branch as an address locspec. Untimately, though,
+    # we're limited in what we can do from inside Python in that front.
+    #
+    # https://sourceware.org/gdb/onlinedocs/gdb/Address-Locations.html#Address-Locations
+    address = None
+    try:
+        # Try to interpret branch as an address literal
+        address = int(branch, 0)
+    except ValueError:
+        # That didn't work. Defer to GDB and see if it can make something out of
+        # the address value we were given.
+        try:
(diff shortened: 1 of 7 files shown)


disconnect3d (approved):

pwndbg/commands/branch.py
```diff
@@ -0,0 +1,111 @@
...
+@pwndbg.commands.ArgparsedCommand(parser, command_name="break-if-not-taken")
+@pwndbg.commands.OnlyWhenRunning
+def break_if_not_taken(branch) -> None:
+    """Breaks at the next call instruction"""
+    install_breakpoint(branch, taken=False)
+
+
+def install_breakpoint(branch, taken):
+    # Do our best to interpret branch as a locspec. Untimately, though, we're
+    # limited in what we can do from inside Python in that front.
```
disconnect3d: Can we extend this comment to link https://sourceware.org/gdb/onlinedocs/gdb/Location-Specifications.html#Location-Specifications for locspec?

pwndbg/commands/branch.py
```diff
@@ -0,0 +1,111 @@
...
+    address = None
+    try:
+        # Try to interpret branch as an address literal
+        address = int(branch, 0)
+    except ValueError:
+        pass
+
+    if address is None:
```
disconnect3d: Should this just be in the except branch maybe?

pwndbg/commands/branch.py
```diff
@@ -0,0 +1,111 @@
...
+            return
+
+    # We should've picked something by now, or errored out.
+    instruction = pwndbg.disasm.one(address)
+    if instruction is None:
+        print(message.error("Could not decode instruction at address %#x" % address))
+        return
+    if CS_GRP_JUMP not in instruction.groups:
+        fmt = (instruction.mnemonic, instruction.op_str, address)
+        print(message.error('Instruction "%s %s" at address %#x is not a branch' % fmt))
```
disconnect3d: Use f-string literals pls

pwndbg/commands/branch.py
mbrla0: Yes, and it shouldn't be just locspec written there, because it's more specifically an address locspec that we're trying to interpret it as.

pwndbg/commands/branch.py
mbrla0: Yeah, that makes more sense

pwndbg/commands/branch.py
mbrla0: Fixed!

pwndbg/commands/branch.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def __init__(self, instruction, taken):
+        super().__init__("*%#x" % instruction.address, type=gdb.BP_BREAKPOINT, internal=False)
+        self.instruction = instruction
+        self.taken = taken
+
+    def should_stop(self):
+        # Use the assistant to figure out which if all the conditions this
+        # branch requires in order to be taken have been met.
+        assistant = pwndbg.disasm.arch.DisassemblyAssistant.for_current_arch()
```
disconnect3d, line 26: Should this maybe be fetched once in `__init__`?

disconnect3d: Nvm, it probably makes sense to have it here e.g. in case of arm thumb/arm or some other cases when one switches architectures ( ??? ).

pwndbg/gdblib/bpoint.py
```diff
@@ -0,0 +1,26 @@
...
+    Breakpoint class, similar to gdb.Breakpoint, but clears the caches
+    associated with the stop event before determining whether it should stop the
+    inferior or not.
+
+    Unlike gdb.Breakpoint, users of this class should override `should_stop()`,
+    instead of `stop()`, as the latter is used to do cache invalidation.
+    """
+
+    def stop(self):
+        # Clear the cache for the stop event.
```
disconnect3d, line 17: Hmm we actually only clear cache for register values and not all caches that are normally cleared on stop event

disconnect3d: Fwiw this is where we set clearing caches for events: https://github.com/pwndbg/pwndbg/blob/e37591b25d923b1933b7092cacafa51a210d5f5e/pwndbg/gdblib/hooks.py#L51-L70

disconnect3d: I think we need to do sth like:
```py
pwndbg.lib.cache._ALL_CACHE_UNTIL_EVENTS['stop'].clear()
```

disconnect3d: I will just merge this and fixup this in another PR

PR #1800, opened 2023-07-16
Bringt back Archlinux CI. Fixes #1772, closes #1783

### LINKED ISSUE #1772: Fix Arch Linux tests
The Arch Linux tests are failing due to new versions of pip needing to use the `--break-system-packages` flag to install packages globally. We should update the CI to use a virtualenv here.

intrigus-lgtm:
This PR fixes issues that were encountered while bringing back arch CI:
TLDR:
1. Our venv activation script included system packages. This can lead to the situation, that e.g. `capstone` is also installed in the system packages, but in version 4 while we require version 5 and our venv also contains version 5. But because the system packages come before our venv in `sys.path` the wrong version will be picked.
Solution: do not include system packages.
2. Arch apparently had some outdated repos. Just remove that.
3. Docker images were pulled. Previously (https://github.com/pwndbg/pwndbg/commit/dd6454065d05daffa7bc36e896dd7baec09c3e1d#diff-3f5366f6d6df3ec1179e5efadc6f350bfa88eebf4e2da589b4d94ccb85ae5e94R25) they were also built so no idea why this has been changed. I think I experienced some weird problems due to this. E.g. some action runs appeared to not use the files changed in a commit.
4. Archlinux (but also Debian 12) uses Full RELRO for libc and ld. Accept these changes in the tests.
5. There is a nection called `.relr.dyn`. Readelf does not really handle it, so we just skip it.
This should be safe because these sections can apparently only contain `RELATIVE` entries in which we aren't interested anyway.
6. Binaries created on the host would have broken references when copied to docker. This is not a problem in CI, because it never has binaries built before, but it is a problem when locally testing inside docker.

Commit-by-commit review recommended.
Happy to move some commits to a separate PR if you want.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,16 @@
 # venv
 .venv/
+
+# ignore binaries that are potentially tainted by the host system
+# E.g. the binary could reference source code on the host system
+# this is a problem because inside docker we won't have access to
+# the host system
+tests/**/binaries/*.o
+tests/**/binaries/*.out
+tests/**/binaries/gosample.x*
+tests/**/binaries/div_zero_binary/core
+tests/**/binaries/div_zero_binary/binary
+!tests/**/binaries/*.go
+
+# ignore QEMU test images (could also be tainted)
+tests/qemu-tests/images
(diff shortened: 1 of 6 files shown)


disconnect3d:
Awesome, thanks a lot!

PR #1804, opened 2023-07-18
build and deploy docs on every commit

hbrylkowski:
I created an action that on every merge to dev will build and deploy docs to gh pages branch


disconnect3d:
Our docs are auto-generated by readthedocs from the dev branch so we don't need to render mkdocs and put them on a branch.

PR #1823, opened 2023-07-21
add libc 2.12+ functionality for structs.py

shenhavmor10:

### DIFF
--- a/pwndbg/heap/structs.py
+++ b/pwndbg/heap/structs.py
@@ -292,7 +292,7 @@ class c_malloc_state_2_26(Structure):
 
 
 class c_malloc_state_2_12(Structure):
-        """
+    """
     This class represents malloc_state struct for GLIBC < 2.22 as a ctypes struct.
 
     https://github.com/bminor/glibc/blob/glibc-2.23/malloc/malloc.c#L1686-L1724
--- a/pwndbg/heap/structs.py
+++ b/pwndbg/heap/structs.py
@@ -55,6 +55,7 @@ DEFAULT_TOP_PAD = 131072
 DEFAULT_MMAP_MAX = 65536
 DEFAULT_MMAP_THRESHOLD = 128 * 1024
 DEFAULT_TRIM_THRESHOLD = 128 * 1024
+DEFAULT_PAGE_SIZE = 4096
 TCACHE_FILL_COUNT = 7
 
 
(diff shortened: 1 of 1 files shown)


pwndbg/heap/structs.py
```diff
@@ -290,6 +291,66 @@ class c_malloc_state_2_26(Structure):
     ]
 
 
+class c_malloc_state_2_12(Structure):
+    """
+    This class represents malloc_state struct for GLIBC < 2.22 as a ctypes struct.
+
+    https://github.com/bminor/glibc/blob/glibc-2.23/malloc/malloc.c#L1686-L1724
```
disconnect3d: Is this whole docstring valid? It says GLIBC<2.22 but did you really check GLIBCs 2.12 up to 2.23 that we do support?

Also the link points to glibc-2.23 branch, so it seems it maybe is a copy-paste leftover from c_malloc_par_2_23 struct.

Please fix the docstring

pwndbg/heap/structs.py
```diff
@@ -594,6 +657,63 @@ class c_malloc_par_2_23(Structure):
         ("sbrk_base", c_pvoid),
     ]
 
+class c_malloc_par_2_12(Structure):
+    """
+    This class represents the malloc_par struct for GLIBC < 2.15 as a ctypes struct.
```
disconnect3d, line 665: Just to be sure: how do we know 2.15 is correct here? If we really know it, can we link to diff or code and state what has changed?

pwndbg/heap/structs.py
```diff
@@ -796,8 +916,10 @@ class MallocPar(CStruct2GDB):
         _c_struct = c_malloc_par_2_26
     elif pwndbg.glibc.get_version() >= (2, 24):
         _c_struct = c_malloc_par_2_24
-    else:
+    elif pwndbg.glibc.get_version() >= (2, 15):
```
disconnect3d: Nitpick: if we do this change, can we call to get_version() once and save it in a `_version` variable or sth?

pwndbg/heap/structs.py
CptGibbon: Indeed that "no description provided" for this PR is killing me

pwndbg/heap/structs.py
shenhavmor10: ye, But the link needs a fix :)

shenhavmor10:
image
in glibc 2.15 the pagesize variable is removed.
image
here the attached threads variable is added in glibc 2.23

disconnect3d:
Do we need to change comments for structs from 2.23 version which was previously the oldest one we supported?

E.g.
```py
class c_malloc_par_2_23(Structure):
    """
    This class represents the malloc_par struct for GLIBC < 2.24 as a ctypes struct.
```

I think this is currently not accurate bcoz its more like 2.15 <= GLIBC < 2.24?

PR #1825, opened 2023-07-21
Fix golang check in dev-setup.sh

bog2n:
Golang check was installing unecessary copy of golang when go was already installed in different than default location


disconnect3d:
Fwiw this may mean the user will end up with a different Go version than we use on CI, but also, if the user has a Go version already installed (e.g. latest version manually installed) we will re-install the one from package manager.

I'm okay with those consequences.

Thanks!

PR #1826, opened 2023-07-21
heap_config: validate glibc version when 'glibc' param is set

### LINKED ISSUE #1801: heap_config should give an example format for GLIBC version
The UX for the glibc version setting is not great.

1. The `heap_config` should give user an example format of GLIBC version we do expect. Currently it only shows this:
```
pwndbg> heap_config
Name                             Value (Def)  Documentation
-----------------------------------------------------------
glibc                            ''           GLIBC version for heuristics
...
```

2. When one sets the `glibc` parameter, we should have a setter that will validate it against the expected version format. Currently it does not:
```
pwndbg> set glibc as
Set GLIBC version for heuristics to 'as'.
```

disconnect3d:
The only disadvantage of this is that we now display the glibc version as a tuple like this:

```
pwndbg> set glibc 2.32
Set GLIBC version for heap heuristics resolution (e.g. 2.31) to (2, 32).
pwndbg> heap_config glibc
Name        Value (Default)    Documentation
--------------------------------------------
glibc       (2, 32) ('')       GLIBC version for heap heuristics resolution (e.g. 2.31)
```

But I think this is fine.

Fixes #1801

### DIFF
--- a/pwndbg/glibc.py
+++ b/pwndbg/glibc.py
@@ -21,6 +21,7 @@ import pwndbg.gdblib.symbol
 import pwndbg.heap
 import pwndbg.lib.cache
 import pwndbg.search
+from pwndbg.color import message
 
 safe_lnk = pwndbg.gdblib.config.add_param(
     "safe-linking",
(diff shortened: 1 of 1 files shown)


disconnect3d:
image

PR #1829, opened 2023-07-23
Added thread section to context

bog2n:
Context now has functionality to display threads and some basic info about them:
* thread number
* name
* status
* program counter address
* debugging symbol for program counter (if exists)

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -833,14 +833,20 @@ def context_args(with_banner=True, target=sys.stdout, width=None):
 
 last_signal: List[str] = []
 
+thread_status_messages = {
+    "running": pwndbg.color.light_green("running"),
+    "stopped": pwndbg.color.yellow("stopped"),
+    "exited": pwndbg.color.gray("exited "),
+}
+
 
 def get_thread_status(thread):
     if thread.is_running():
-        return pwndbg.color.light_green("running")
+        return thread_status_messages["running"]
     elif thread.is_stopped():
-        return pwndbg.color.yellow("stopped")
+        return thread_status_messages["stopped"]
     elif thread.is_exited():
-        return pwndbg.color.gray("exited ")
+        return thread_status_messages["exited"]
     else:
         return "unknown"
 
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -875,12 +875,10 @@ def context_threads(with_banner=True, target=sys.stdout, width=None):
 
         padding = max_name_length - len(thread.name)
 
-        line = f" {selected} {thread.global_num}\t"
-        line += f'"{pwndbg.color.cyan(thread.name)}" ' + (" " * padding)
         line = (
             f" {selected} {thread.global_num}\t"
             f'"{pwndbg.color.cyan(thread.name)}" '
-            f"{" " * padding}"
+            f'{" " * padding}'
             f"{status}: {M.get(frame.pc())} "
         )
         if symbol:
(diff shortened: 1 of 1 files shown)


disconnect3d:
@bog2n can u show a screenshot how this looks like?

pwndbg/commands/context.py
```diff
@@ -834,6 +834,56 @@ def context_args(with_banner=True, target=sys.stdout, width=None):
 last_signal: List[str] = []
 
 
+def get_thread_status(thread):
+    if thread.is_running():
+        return pwndbg.color.light_green("running")
```
disconnect3d: nitpick: if those colors are constant/not-changeable, lets compute/move those colored-strings once by moving them to the global scope.

pwndbg/commands/context.py
```diff
@@ -834,6 +834,56 @@ def context_args(with_banner=True, target=sys.stdout, width=None):
...
+        selected = " ►" if thread is selected_thread else "  "
+
+        symbol = pwndbg.gdblib.symbol.get(frame.pc())
+        status = get_thread_status(thread)
+
+        padding = max_name_length - len(thread.name)
+
+        line = f" {selected} {thread.global_num}\t"
+        line += f'"{pwndbg.color.cyan(thread.name)}" ' + (" " * padding)
+        line += f"{status}: {M.get(frame.pc())} "
```
disconnect3d: Instead of formatting a string three times and appending them, lets format it once
```suggestion
        line = (
            f" {selected} {thread.global_num}\t"
            f'"{pwndbg.color.cyan(thread.name)}" '
            f"{" " * padding}"
            f"{status}: {M.get(frame.pc())} "
        )
```

pwndbg/commands/context.py
```diff
@@ -834,6 +834,56 @@ def context_args(with_banner=True, target=sys.stdout, width=None):
...
+    elif thread.is_stopped():
+        return pwndbg.color.yellow("stopped")
+    elif thread.is_exited():
+        return pwndbg.color.gray("exited ")
+    else:
+        return "unknown"
+
+
+def context_threads(with_banner=True, target=sys.stdout, width=None):
+    threads = gdb.selected_inferior().threads()[::-1]
```
disconnect3d, line 855: We probably want to move some of this functionality to gdblib/threads.py - i.e. have a function there that will return data in a specific format/without GDB structures (this would allow us to have lldblib/threads.py returning a similar information in the future and supporting lldb this way)

pwndbg/commands/context.py
```diff
@@ -833,6 +833,67 @@ def context_args(with_banner=True, target=sys.stdout, width=None):
...
+
+        selected = " ►" if thread is selected_thread else "  "
+
+        symbol = pwndbg.gdblib.symbol.get(frame.pc())
+        status = get_thread_status(thread)
+
+        padding = max_name_length - len(thread.name)
+
+        line = f" {selected} {thread.global_num}\t"
+        line += f'"{pwndbg.color.cyan(thread.name)}" ' + (" " * padding)
```
disconnect3d: (sadly the suggestion was not multiline :()

pwndbg/commands/context.py
```diff
@@ -833,6 +833,67 @@ def context_args(with_banner=True, target=sys.stdout, width=None):
 
 last_signal: List[str] = []
 
+thread_status_messages = {
+    "running": pwndbg.color.light_green("running"),
+    "stopped": pwndbg.color.yellow("stopped"),
+    "exited": pwndbg.color.gray("exited "),
```
disconnect3d, line 839: the space is probably not needed
```suggestion
    "exited": pwndbg.color.gray("exited"),
```

bog2n:
It looks like this:
context thread

pwndbg/commands/context.py
bog2n: This space is needed for padding because `stopped` has one more character than `exited`

Issue #1766, opened 2023-06-24
vmmap <address/page>


disconnect3d:
It would be nice if the `vmmap $rdx` command would also display few memory pages after the found one.

This is often helpful for any out-of-bounds read/write primitives when we can jump over pages.

gsingh93:
Probably want before as well, not just after. Should this be the default or enabled with an optional flag?

gsingh93:
We could make it like grep, with -A, -B, -C for after, before, both with a number of how many additional to print.

disconnect3d:
I think this should be a default. -A/B/C flags would also be okay, or, a config option if the default is wanted or not ( ? ).

disconnect3d:
Done in #1810. We finally added -B and -A options and we didn't add -C since we don't want to handle the mix of them.

disconnect3d:
Hmm @feelfreelinux the #1810 broke a case where the filtered memory page displayed the offset from the address, e.g. in:

```
pwndbg> vmmap $rip
LEGEND: STACK | HEAP | CODE | DATA | RWX | RODATA
             Start                End Perm     Size Offset File
    0x555555554000     0x555555558000 r--p     4000      0 /usr/bin/ls
    0x555555558000     0x55555556c000 r-xp    14000   4000 /usr/bin/ls
    0x55555556c000     0x555555574000 r--p     8000  18000 /usr/bin/ls
```

We previously showed:
```
    0x555555558000     0x55555556c000 r-xp    14000   4000 /usr/bin/ls + <offset from the begining of the page>
```

PR #1830, opened 2023-07-23
`vmmap -A / -B` improvements

### LINKED ISSUE #1821: vmmap -Ax -Bx displays wrong results when filter returns multiple pages
image

feelfreelinux:
Implements improvements over https://github.com/pwndbg/pwndbg/pull/1810
- `vmmap -A / -B` now handles multiple results gracefully
- printed pages now prepend an backtrace arrow before the results that were actually filtered, to make it more readable
- Fixed displaying of address offset in filtered pages

Fixes: https://github.com/pwndbg/pwndbg/issues/1821 and https://github.com/pwndbg/pwndbg/issues/1766

### DIFF
--- a/pwndbg/color/memory.py
+++ b/pwndbg/color/memory.py
@@ -16,7 +16,7 @@ c = ColorConfig(
 )
 
 
-def get(address, text=None) -> str:
+def get(address, text=None, prefix=None) -> str:
     """
     Returns a colorized string representing the provided address.
 
@@ -24,6 +24,7 @@ def get(address, text=None) -> str:
         address(int): Address to look up
         text(str): Optional text to use in place of the address
               in the return value string.
+        prefix(str): Optional text to set at beginning in the return value string.
     """
     address = int(address)
 
@@ -50,6 +51,10 @@ def get(address, text=None) -> str:
         text = hex(int(address))
     if text is None:
         text = str(int(address))
+
+    if prefix:
+        # Replace first N characters with the provided prefix
+        text = prefix + text[len(prefix) :]
 
     return color(text)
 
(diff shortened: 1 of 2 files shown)


pwndbg/commands/vmmap.py
```diff
@@ -83,42 +83,72 @@ def print_vmmap_table_header() -> None:
...
-                    filtered_pages.append(pages[matched_index + after_index])
+                if (
+                    matched_index + after_index < len(total_pages) - 1
+                    and total_pages[matched_index + after_index] not in pages_to_display
+                ):
+                    pages_to_display.append(total_pages[matched_index + after_index])
 
-        pages = filtered_pages
+        # Sort results by address
+        total_pages = sorted(pages_to_display, key=lambda page: page.vaddr)
```
disconnect3d, line 127: Hmm aren't they sorted already from vmmap.get?

pwndbg/color/memory.py
```diff
@@ -16,14 +16,15 @@
...
-def get(address, text=None) -> str:
+def get(address, text=None, prefix=None) -> str:
     """
     Returns a colorized string representing the provided address.
 
     Arguments:
         address(int): Address to look up
         text(str): Optional text to use in place of the address
               in the return value string.
+        prefix(str): Optional text to set at beginning in the return value string.
```
disconnect3d, line 27: Hmm shall we add the prefix here or just format it inline when used? (cc: @gsingh93 wdyt?)

pwndbg/commands/vmmap.py
feelfreelinux: I sorted it at the end to simplify the insertion logic of the look around pages a bit, as it was a little bit messy after adding the logic to insert pages in-between multiple results.

pwndbg/color/memory.py
gsingh93: Yea, I feel like the prefix should be handled outside of this function.

disconnect3d: @feelfreelinux any chance u can revert that small change and handle it outside?

PR #1832, opened 2023-07-24
Add support for FreeBSD

RHL120:
This PR adds support for FreeBSD.
As far as I can tell the only file that needs modification is setup.sh so I only modified that.
It works fine on FreeBSD 13.2-RELEASE.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
 set -ex
 
 # If we are a root in a container and `sudo` doesn't exist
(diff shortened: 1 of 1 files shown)


setup.sh
```diff
@@ -162,7 +170,9 @@ git submodule update --init --recursive
 
 # Find the Python version used by GDB.
 PYVER=$(gdb -batch -q --nx -ex 'pi import platform; print(".".join(platform.python_version_tuple()[:2]))')
-PYTHON+=$(gdb -batch -q --nx -ex 'pi import sys; print(sys.executable)')
+PYTHON+=$(gdb -batch -q --nx -ex 'pi import sys; print(sys.executable)') ||
+	PYTHON=$(gdb -batch -q --nx -ex 'pi import sys; print(sys.executable)')
```
disconnect3d: What does this do?

disconnect3d:
Thx

PR #1837, opened 2023-07-25
vmmap: make linux kernel pages searchable via name

### LINKED ISSUE #1835: Linux kernel vmmaps from gdb-pt-dump should have unique names
When debugging a Linux kernel, we use gdb-pt-dump to dump vmmap information. When we do that, we name all memory pages as "<pt>".

However, this is not useful at all since now none of the memory pages can be referenced directly through e.g. the `search` command.

As a baseline we should name those pages similarly to what we do in userspace for anonymous pages: there, we name them "`[anon_<high part of address>]`.

So here we should name them: `[pt_<high part of address>]`.

disconnect3d:
This commit changes the name of linux kernel memory pages fetched via gdb-pt-dump from `"<pt>"` to `"[pt_<addr>]"` to make it possible to use those memory pages with other commands that take memory map names as arguments, like the search command:
```
search <value> <mapping-name>
```

Fixes https://github.com/pwndbg/pwndbg/issues/1835

### DIFF
--- a/pwndbg/lib/memory.py
+++ b/pwndbg/lib/memory.py
@@ -94,7 +94,7 @@ class Page:
 
     @property
     def is_memory_mapped_file(self) -> bool:
-        return len(self.objfile) != 0 and self.objfile[0] != "[" and self.objfile != "<pt>"
+        return len(self.objfile) != 0 and self.objfile[0] != "["
 
     @property
     def read(self) -> bool:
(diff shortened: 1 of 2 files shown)


pwndbg/lib/memory.py
```diff
@@ -94,7 +94,7 @@ def is_stack(self) -> bool:
 
     @property
     def is_memory_mapped_file(self) -> bool:
-        return len(self.objfile) != 0 and self.objfile[0] != "[" and self.objfile != "<pt>"
+        return len(self.objfile) != 0 and self.objfile[0] != "["
```
disconnect3d, line 97: That last check here is obsolete after we rename from `<pt>` to `[pt_<addr>]` :)

PR #1841, opened 2023-07-25
Add support for glibc heap UAF detection

mbrla0:
This draft adds support for detecting UAF bugs (in glibc heap) in a running inferior, also allowing pwndbg to break on it it if happens.

This is currently a tentative patch, and is missing a few things before I think it's PR ready, but I'd like to hear comments on the approach being taken and the questions that I can't answer on my own.

The following things still need to be done:
- [X] Handling of `calloc` and `realloc`.
- [X] A mechanism to disable the detection. (It is fairly invasive)
- [ ] A way to communicate this information to the user.

And I've got the following questions:

- Does the way the current implementation detects accesses to freed memory look reasonable? Are there any big pitfalls to it that I've missed?
- Can we affort the problems associated with using hardware watchpoints[^hwwps] for the speed, or do like GEF and trade the performance of hw watchpoints for reliability of sw ones? Or, should we let the user decide if we have some way to detect the architecture can handle our needs?
- What should the mechanism for enabling and disabling the checks look like?
- What should be the way we communicate possible memory-related misbehavior to a user?

[^hwwps]: [GDB documentation](https://sourceware.org/gdb/onlinedocs/gdb/Set-Watchpoints.html) does say some pretty scary things about them, when it comes to us maybe not being able to detect watchpoints that failed to be set.


disconnect3d:
@CptGibbon @lebr0nli pls check this out & help with questions/reviews!

pwndbg/gdblib/nearpc.py
```diff
@@ -176,6 +176,84 @@ def nearpc(pc=None, lines=None, emulate=False, repeat=False) -> List[str]:
             first_pc = False
             should_highlight_opcodes = True
 
+        # If this instruction performs a memory access operation, we should tell
+        # the user anything we can figure out about the memory it's trying to
+        # access.
+        mem_access = ""
```
disconnect3d, line 182: Btw are the nearpc changes/detection needed here at all? I recall it was there to show up "UAF detected" or sth in disasm but I think we don't need this anymore bcoz:
1) it probably makes thing slower
2) it doesnt emulate code, doesnt it? and if so, it won't always be accurate
3) we now detect things dynamically via malloc/free hooks and watchpoints they set up?

pwndbg/heap_tracking.py
```diff
@@ -0,0 +1,357 @@
...
+MALLOC_NAME = "malloc"
+FREE_NAME = "free"
+
+
+class FreeChunkWatchpoint(gdb.Breakpoint):
+    def __init__(self, chunk, tracker):
+        self.chunk = chunk
+        self.tracker = tracker
+        super().__init__(
+            f"*(char[{chunk.size}]*){chunk.address:#x}", type=gdb.BP_WATCHPOINT, internal=True
```
disconnect3d: nitpick: will it work if user did `set language rust` before running this?

pwndbg/heap_tracking.py
```diff
@@ -0,0 +1,357 @@
...
+        )
+
+    def stop(self):
+        pwndbg.gdblib.regs.__getattr__.cache.clear()
```
disconnect3d: are we sure its the only cache we need to clear? maybe we should clear all stop caches?

pwndbg/heap_tracking.py
```diff
@@ -0,0 +1,357 @@
...
+
+
+class MallocEnterBreakpoint(gdb.Breakpoint):
+    def __init__(self, tracker):
+        super().__init__(MALLOC_NAME, internal=True)
+        self.tracker = tracker
+
+    def stop(self):
+        pwndbg.gdblib.regs.__getattr__.cache.clear()
+        requested_size = pwndbg.arguments.argument(0)
```
disconnect3d: on 32-bit the arg may be on stack and so clearing regs cache may probably not be enough if its cached? not sure if it is.

disconnect3d:
> * Does the way the current implementation detects accesses to freed memory look reasonable? Are there any big pitfalls to it that I've missed?

Clearing cache: we probably want to clear all stop caches in case 'arguments(0)' or other stuff is cached and not just regs?

It may also have issues if heap metadata is corrupted I guess bcoz we parse some heap chunks data?

> * Can we affort the problems associated with using hardware watchpoints[1](#user-content-fn-hwwps-be038261e79d1c071d4398d5662d1ef8) for the speed, or do like GEF and trade the performance of hw watchpoints for reliability of sw ones? Or, should we let the user decide if we have some way to detect the architecture can handle our needs?

If gef does that they probably had a good reason to disable hwbps. Lets just disable them and inform the user we do this? Or  ask the user if they want to disable or not and have a flag to force-enable/disable?

> * What should the mechanism for enabling and disabling the checks look like?

Should probably be a separate command, I think gef has it like that. It should inform user that its experimental, may not work with corrupted heap or sth?

> * What should be the way we communicate possible memory-related misbehavior to a user?

What do u mean by misbehavior? Like corrupted chunks? We should inform/warn the user when they execute the command with a warning.

disconnect3d:
Maybe the command should be `heap-trace` and it should enable all heap tracing that we will implement and should have a `[--no-uaf-detection]` and some other flags?

pwndbg/gdblib/nearpc.py
mbrla0: Nope, we don't really need it anymore. I left that there by accident :)

Regarding your points: (2) probably doesnt apply since, while it doesnt emulate, it restricts the check to only run for the instruction in the current PC. Regardless, (1) and (3) absolutely do, so, yeah, better to take it out

CptGibbon:
> @CptGibbon @lebr0nli pls check this out & help with questions/reviews!

Sure, it looks very interesting and bold.
A lot of new code so will take some time 👍

CptGibbon:
Just a quick test:
With `gdb /bin/sh` I'm seeing a lot "warning: value truncated" and the watchpoints keep stopping execution.
I think the `malloc` breakpoint has matched the ld.so implementation which is confusing matters.

pwndbg/heap_tracking.py
```diff
@@ -0,0 +1,357 @@
+"""
+Heap Tracking
+
+This module implements runtime tracking of the heap, allowing pwndbg to detect
+heap related misbehavior coming from an inferior in real time, which lets us
+catch UAF bugs, double frees (and more), and report them to the user.
+
+# Approach
+The approach used starting with using breakpoints to hook into the following
+libc symbols: `malloc`, `free`, `calloc`, and `realloc`. Each hook has a
```
CptGibbon: It might be prudent to use `__libc_malloc` et al to avoid confusion with the ld.so malloc implementation.

pwndbg/heap_tracking.py
mbrla0: I've been testing out the extent of this patch's support for Rust with this small program:
```rust
fn main() {
    let a = Box::new(1u32);
    let a = Box::into_raw(a);

    let b = a as usize;
    unsafe {
        /* Fine. This object is still alive. */
        *a = 0;

        /* Free the object. */
        let _ = Box::from_raw(a);

        /* Trigger UAF.
         *
         * We do a round trip pointer conversion here to skid around provenance
         * rules that would otherwise apply if we were to read directly from the
         * original pointer, and trigger UB. We want UAF soup, not ASM pretzels.
         */
         let a = b as *mut u32;
         let _ = std::ptr::read_volatile(a);
    }
}
```

But it seems like the version of GDB I'm testing with has an issue in `gdb.FinishBreakpoint` when Rust frames get mixed with C ones (such as when the Rust stdlib calls `malloc`). In my testing, GDB calls `gdb.FinishBreakpoint.out_of_scope()` after malloc and free return, rather than `stop()`, as it does in an all-C codebase.

I also notice that GDB is reporting the wrong function names when it breaks in the main function.

Clearly supporting Rust goes deeper than just fixing fixing the expression.

Here are the versions of GDB, rustc and the command line used to build the test program:
```
pwndbg> version
Gdb:      9.2
Python:   3.8.10 (default, May 26 2023, 14:05:08)  [GCC 9.4.0]
Pwndbg:   2023.07.17 build: 49eac60
Capstone: 5.0.1280
Unicorn:  2.0.1
```

```
$ rustc --version
rustc 1.66.1 (90743e729 2023-01-10) (built from a source tarball)
```

```
$ rustc a/src/main.rs -C opt-level=3 -C lto=fat -C debuginfo=2 -o a.rust.x86
```

pwndbg/heap_tracking.py
mbrla0: Yes! When testing Rust support I ran into an issue where not having cleared all the stop caches led to a crash in code from `pwndbg.heap` I was using to parse heap information. The next commit will have it clear all stop caches.

pwndbg/heap_tracking.py
mbrla0: It should be enough, I can't see anything in that function that relies on anything but the register cache. But, as discussed above, all of the stop caches are cleared now.

mbrla0:
@disconnect3d:
> It may also have issues if heap metadata is corrupted I guess bcoz we parse some heap chunks data?

Probably, yes. Is there any mechanism already in pwndbg that can detect heap corruption to some extent, or would one have to be added in? We could use it to warn users of it.

Of course it would really only be a best effort kind of thing, detecting heap corruption can only really be heuristic unless we start replacing bits of the allocator itself to have it do a lot more tracking.

> What do u mean by misbehavior? Like corrupted chunks? We should inform/warn the user when they execute the command with a warning.

I mean all of the things that we can detect the program is doing wrong in regards to memory. UAF, double free, freeing pointers that don't land in known chunks. Should we stop the program? Or should the user be able to configure whether the program will be stopped or if it will be just a warning?

mbrla0:
@CptGibbon 
> Just a quick test: With `gdb /bin/sh` I'm seeing a lot "warning: value truncated" and the watchpoints keep stopping execution. I think the `malloc` breakpoint has matched the ld.so implementation which is confusing matters.

I've looked into the "warning: value truncated" message, and it comes from inside GDB during the int->pointer cast done by `pwndbg.gdblib.memory.poi`, when it's called by this chunk of code from `class Chunk` in `ptmalloc.py`:
```py
 def __init__(self, addr, heap=None, arena=None) -> None:

    if isinstance(pwndbg.heap.current.malloc_chunk, gdb.Type):
        self._gdbValue = pwndbg.gdblib.memory.poi(pwndbg.heap.current.malloc_chunk, addr)
    else:
        # [...]
```

It's still a bit of a mystery to me why that warning only seems to appear if this code path is triggered from a breakpoint's `stop()` function, seeing as it's not there when the same path is triggered for the same addresses through the `heap` command.

As for the watchpoints stopping execution, when I test it it runs fine. What versions are you testing with, so I can use them to see what's wrong?

CptGibbon:
> I mean all of the things that we can detect the program is doing wrong in regards to memory. UAF, double free, freeing pointers that don't land in known chunks. Should we stop the program? Or should the user be able to configure whether the program will be stopped or if it will be just a warning?

IMO letting the user choose keeps the most folks happy.
One assumes that users who enable this tracing want to be notified when e.g. a UAF is detected, so it could be useful to have an option to stop the program altogether.

CptGibbon:
> It's still a bit of a mystery to me why that warning only seems to appear if this code path is triggered from a breakpoint's `stop()` function, seeing as it's not there when the same path is triggered for the same addresses through the `heap` command.

Huh okay that's a weird one, @disconnect3d might know more.

> As for the watchpoints stopping execution, when I test it it runs fine. What versions are you testing with, so I can use them to see what's wrong?

I just use the repository's Docker image 🤷‍♂️

mbrla0:
> I just use the repository's Docker image 🤷‍♂️

Huh, I'm using it too. Can you give me a few more details on what's going wrong (the text log at least)?

CptGibbon:
> Huh, I'm using it too. Can you give me a few more details on what's going wrong (the text log at least)?

Yes my mistake, I let it drop into another inferior and it was spitting out all sorts of errors.
I'll conduct a proper code review soon 👍

pwndbg/heap_tracking.py
```diff
@@ -0,0 +1,357 @@
...
+        ret_ptr = int(self.return_value)
+        if ret_ptr == 0:
+            # No change.
+            self.tracker.exit_memory_management()
+            return False
+
+        ty = pwndbg.gdblib.typeinfo.ppvoid
+        size = int(pwndbg.gdblib.memory.poi(ty, ret_ptr - ty.sizeof))
+
+        flags = size & 7
```
CptGibbon: A chance to avoid using magic numbers with the `pwndbg.heap.ptmalloc.Chunk` class, though it lacks a `requested_size` field.

pwndbg/heap_tracking.py
```diff
@@ -0,0 +1,357 @@
...
+        self.size = size
+        self.requested_size = requested_size
+        self.flags = flags
+
+
+class Tracker:
+    def __init__(self):
+        self.free_chunks = SortedDict()
+        self.alloc_chunks = SortedDict()
+        self.free_wps = dict()
```
CptGibbon: Nitpick: I had to scroll a fair way to find what `free_wps` stood for, ain't nothing wrong with descriptive variable names.

pwndbg/heap_tracking.py
```diff
@@ -0,0 +1,357 @@
...
+    def free(self, address) -> bool:
+        if address not in self.alloc_chunks:
+            return False
+        chunk = self.alloc_chunks[address]
+        wp = FreeChunkWatchpoint(chunk, self)
+
+        self.free_chunks[chunk.address] = chunk
+        self.free_wps[chunk.address] = wp
+
+        return True
```
CptGibbon: Does the corresponding entry need to be removed from `alloc_chunks`?

pwndbg/heap_tracking.py
mbrla0: Turns out `__GI___libc_malloc` is re-entrant in a way I'm not sure the code supports. Is there any reason for doing that to avoid `ld.so` over checking if any of the frames in the current stack belong to the program, as the patch already does with `in_program_code_stack()`? Or perhaps we could make sure we're installing the breakpoint for the malloc we're actually interested in somehow?

pwndbg/heap_tracking.py
mbrla0: Yeap, added that now.

pwndbg/heap_tracking.py
mbrla0: Fair enough, that's been changed to `free_watchpoints` now.

pwndbg/heap_tracking.py
mbrla0: I think relying on that class might be a bit much for what I'm trying to do there. Might a `FLAGS_BITMASK = 7` value make what I'm trying to do clearer?

disconnect3d:
This needs a fix of conflicts.

Answering some questions/comments here:

1. The value truncation error from `memory.poi` seems that it may be a bug on our end, iirc this happens when you do cast a big `gdb.Value` to some other type and to prevent that we often did something like `value & pwndbg.gdblib.arch.ptrmask`. Perhaps we should find out why this happens here and maybe add such thing too?

2. RE: letting users to choose whether to stop, I think it would be nice to have that as a flag/option to the command

3. RE: detecting glibc malloc - yeah, ideally we don't want to step on ld.so's malloc

mbrla0:
> The value truncation error from memory.poi seems that it may be a bug on our end, iirc this happens when you do cast a big gdb.Value to some other type and to prevent that we often did something like value & pwndbg.gdblib.arch.ptrmask. Perhaps we should find out why this happens here and maybe add such thing too?

This issue seems to happen intermittently in a way I'm not entirely sure how to reproduce. But there might be a clue in the fact that, when I was writing the DYANMIC section parser for #1971, I could trigger this bug by reading my types from `pwndbg.gdblib.typeinfo.<etc>` too early. I think it's an interesting issue, but it probably has nothing to do with this PR.

> RE: letting users to choose whether to stop, I think it would be nice to have that as a flag/option to the command
>
> RE: detecting glibc malloc - yeah, ideally we don't want to step on ld.so's malloc

I've added an option to transform that stop into just a warning, and a safeguard against picking the wrong symbol for `malloc` and the other symbols we use. 

The symbol safeguards are pretty crude, and, ideally, we'd also first try to find the symbol by iterating through the object files yielded by `gdb.objfiles()`, filtering out ones that don't look like libc. In fact, I had initially implemented that, but ultimately decided against it, because I couldn't think of a way to properly test it. For now, at least it stops the code picking `malloc@plt` or the `malloc` in `ld.so`.

And with this I think this can be converted into a PR proper.

disconnect3d:
Can we change the interface here to be similar to `track-got`, e.g. `track-heap-uaf enable|disable|<?>`

mbrla0:
Just for the record. Currently this PR is somewhat blocked on GDB intermittently failing to fire the `gdb.FinishBreakpoint` we set up to capture the return value, in cases where manually calling `finish` succeeds, which makes me suspect that behavior is a bug.

But, since what we're doing is _technically_ [not supported](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Breakpoints-In-Python.html), instead of straight up filing a bug, I've brought [to their attention](https://sourceware.org/pipermail/gdb/2024-January/051051.html) the concern that we do need this to work, as well as some questions on the rationale behind not supporting it, but have gotten no reply as of yet.

disconnect3d:
@mbrla0 we got an answer from Tom in that GDB topic: https://sourceware.org/pipermail/gdb/2024-January/051062.html

Also, how does the current implementation compare to GEF?
* Does the two have the same issues?
* Do we provide the same or more amount of information to the user on what is going on etc? (I would aim for that)

Can we call it "experimental" and ship it?

mbrla0:
> we got an answer from Tom in that GDB topic: https://sourceware.org/pipermail/gdb/2024-January/051062.html

@disconnect3d oh, yes, I did see it. Maybe we should look into documenting in an issue everything that's jankier than necessary or that would be nice to implement but would somehow be handicapped by that GDB limitation.

> Does the two have the same issues?

Not the exact same ones, since they hook to the `__libc_*` family, but, from my testing GEF also gets tripped up and reports
errors for `/bin/ls` that Valgrind doesn't see. And I think it's fair to assume that, in this case, if Valgrind doesn't catch it, it's probably a false positive.

> Do we provide the same or more amount of information to the user on what is going on etc? (I would aim for that)

We log call information and display error information in the context when an error is detected and we stop the program because of it. From what I've seen we display exactly as much information as GEF (in a surprisingly similar way too, which is kinda funny lol).

> Can we call it "experimental" and ship it?

Well, maybe? I feel like the limitations are significant still. So I'd say at least a warning/disclaimer is warranted, so that users know to expect false positives.

disconnect3d:
> Valgrind doesn't catch it, it's probably a false positive

Maybe there are some exemptions for some binaries in/for Valgrind? Afaik it is possible to tell it to ignore certain leaks.

> Well, maybe? I feel like the limitations are significant still. So I'd say at least a warning/disclaimer is warranted, so that users know to expect false positives.

Add this warning pls! Tell the users that this is experimental etc.

disconnect3d:
Also fix lint issues :<

mbrla0:
> 
> Maybe there are some exemptions for some binaries in/for Valgrind? Afaik it is possible to tell it to ignore certain leaks.
> 
 
That doens't seem to be the case here, considering it's a double free that GEF was reporting, and Valgrind doesn't seem to be ignoring anything.
 
> Add this warning pls! Tell the users that this is experimental etc.

Okay, I've added a simple warning.

I've also fixed the lint

PR #1845, opened 2023-07-29
pacman: only install not-installed deps

Xeonacid:
Pacman installs all packages (including installed ones) by default unless `needed` is used.


disconnect3d:
Thanks!

PR #1846, opened 2023-07-30
enhance decompiler compatible problem with rizin and radare2

Squirre17:
image

seems like here have a logic problem
I installed both r2 and rizin
but I only installed rz-ghidra
so pwndbg use the r2 to determine whether "pdb" command available rather than rizin which realy installed ghidra-plugin.

The modification is let user to specify what decompiler they want to use. and fix some compatibility problem in rizin and r2.(e.g there are no space between `@` and `symbol` which rizin not able to parse )

### DIFF
--- a/pwndbg/ghidra.py
+++ b/pwndbg/ghidra.py
@@ -22,7 +22,7 @@ def set_r2decompiler() -> None:
         return
     print(
         message.warn(
-            f"Invalid r2decompiler : `{r2decompiler.value}`, you should select from (radare2/rizin)"
+            f"Invalid r2decompiler: `{r2decompiler.value}`, please select from radare2 or rizin"
         )
     )
     r2decompiler.revert_default()
--- a/pwndbg/ghidra.py
+++ b/pwndbg/ghidra.py
@@ -28,12 +28,12 @@ def decompile(func=None):
         if r2decompiler == "radare2":
             r2 = pwndbg.radare2.r2pipe()
             # LD -> list supported decompilers (e cmd.pdc=?)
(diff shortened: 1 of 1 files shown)


pwndbg/ghidra.py
```diff
@@ -10,6 +10,10 @@
 import pwndbg.radare2
 import pwndbg.rizin
 
+r2decompiler = pwndbg.gdblib.config.add_param(
+    "r2decompiler", "radare2", "framework that your ghidra plugin installed (radare2/rizin)"
```
disconnect3d, line 15: Can we add a trigger that will validate the provided value to be one of the expected ones?

See an example here:
https://github.com/pwndbg/pwndbg/blob/16a0f3b57ba281d14a774b8bd019a912d7c64655/pwndbg/glibc.py#L38-L50

Squirre17: it's done. 
image

pwndbg/ghidra.py
```diff
@@ -9,6 +9,23 @@
...
+)
+
+
+@pwndbg.gdblib.config.trigger(r2decompiler)
+def set_r2decompiler() -> None:
+    if r2decompiler.value in ["radare2", "rizin"]:
+        return
+    print(
+        message.warn(
+            f"Invalid r2decompiler : `{r2decompiler.value}`, you should select from (radare2/rizin)"
```
disconnect3d:
```suggestion
            f"Invalid r2decompiler: `{r2decompiler.value}`, please select from radare2 or rizin"
```

disconnect3d:
Thanks!

PR #1853, opened 2023-08-01
Warn user about unmet dependencies (#1791)

### LINKED ISSUE #1791: Better exception handling for unmet dependencies
Some of our users update Pwndbg only by `git pull`ing the newest version and not re-running `setup.sh`. If that happens and when we updated our dependencies recently, the user may end up with broken Pwndbg such as (likely) happened in https://github.com/pwndbg/pwndbg/issues/1790.

We should do the two following things:
1) When we detect/handle an exception in the [`pwndbg/exception.py:handle` function]( https://github.com/pwndbg/pwndbg/blob/de4acb2f40da3b0c97353a8b680ffa6440346f7b/pwndbg/exception.py#L61-L96), we should check the installed dependencies versions against the required ones and if they do not match, we should inform the user to update them.
2) We may want to display installed dependencies versions within the `bugreport` command output?

searabbitx:
I'll try to write some test for it. Also, it'd be nice to find some clearer way to find `requirements.txt` file path.

Fixes #1791

### DIFF
--- a/pwndbg/exception.py
+++ b/pwndbg/exception.py
@@ -39,7 +39,7 @@ def inform_unmet_dependencies(errors) -> None:
     """
     Informs user about unmet dependencies
     """
-    msg = message.error("You appear to have unmet dependencies.\n")
+    msg = message.error("You appear to have unmet Pwndbg dependencies.\n")
     for e in errors:
         if isinstance(e, pkg_resources.DistributionNotFound):
             msg += message.notice(f"- required {e.args[0]}, but not installed\n")
--- a/pwndbg/exception.py
+++ b/pwndbg/exception.py
@@ -1,10 +1,12 @@
 from __future__ import annotations
 
 import functools
+import os
 import sys
 import traceback
 
 import gdb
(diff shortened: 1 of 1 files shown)


searabbitx:
The message will look like this:
image

pwndbg/exception.py
```diff
@@ -33,6 +35,40 @@
 )
 
 
+def inform_unmet_dependencies(errors) -> None:
+    """
+    Informs user about unmet dependencies
+    """
+    msg = message.error("You appear to have unmet dependencies.\n")
```
disconnect3d:
```suggestion
    msg = message.error("You appear to have unmet Pwndbg dependencies.\n")
```

searabbitx:
This will break, if the `exception.py` file is moved deeper into the project structure. I was looking for a smarter way to get the project root path, but did not find any. I'll open a new PR to improve this, if something comes to my mind in the future.

disconnect3d:
Thanks a lot!

Issue #1854, opened 2023-08-01
Drop support for GDB < 9.2


disconnect3d:
We do have this in README:
> Pwndbg is supported on Ubuntu 20.04, and 22.04 with GDB 9.2 and later.

So we should remove all the old workaround code we used for old GDBs < 9.2. This issue is there to track those efforts: feel free to send a PR fixing just one thing, or a PR fixing many or all of those.

Those are:
* [x] `gdb77_get_register` and the relevant "if hasattr" code: https://github.com/pwndbg/pwndbg/blob/1c4307df9d6d7086fa476f551f16300b24179bf9/pwndbg/gdblib/regs.py#L22-L35
* [x] `_fs_gs_helper` - we don't need to use `ptrace` anymore iiuc: https://github.com/pwndbg/pwndbg/blob/1c4307df9d6d7086fa476f551f16300b24179bf9/pwndbg/gdblib/regs.py#L187-L190
* [x] We probably don't need to try/except this anymore: https://github.com/pwndbg/pwndbg/blob/1c4307df9d6d7086fa476f551f16300b24179bf9/pwndbg/gdblib/events.py#L120-L125
* [x] The description for `entry` command mentions GDB 8.1 - we can probably drop the parentheses here: https://github.com/pwndbg/pwndbg/blob/1c4307df9d6d7086fa476f551f16300b24179bf9/pwndbg/commands/start.py#L87-L89
* [x] There was some weird GDB 8.2 case in `check_repeated`, a functionality for "repeating commands execution" - the except can probably be removed, but I recommend investigating the linked issue there: https://github.com/pwndbg/pwndbg/blob/1c4307df9d6d7086fa476f551f16300b24179bf9/pwndbg/commands/__init__.py#L166-L171
* [x] Handling version:
```
commands/version.py
29:        return gdb.VERSION  # GDB >= 8.1 (or earlier?)
188:    # 8. showing width
```
* [x] Some heap cases:
```
heap/structs.py
40:    # alignof doesn't available in GDB < 8.2 (https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git;a=blob_plain;f=gdb/NEWS;hb=gdb-8.2-release)

heap/ptmalloc.py
974:            # alignof doesn't available in GDB < 8.2 (https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git;a=blob_plain;f=gdb/NEWS;hb=gdb-8.2-release)
[...]

PR #1856, opened 2023-08-03
Drop support for gdb lt 9.2 #1854

dmur1:
https://github.com/pwndbg/pwndbg/issues/1854

### DIFF
--- a/pwndbg/gdblib/events.py
+++ b/pwndbg/gdblib/events.py
@@ -115,10 +115,9 @@ registered: dict[Any, list[Callable]] = {
     gdb.events.start: [],
     gdb.events.new_thread: [],
     gdb.events.before_prompt: [],  # The real event might not exist, but we wrap it
+    gdb.events.memory_changed: [],
+    gdb.events.register_changed: []
 }
-
-registered[gdb.events.memory_changed] = []
-registered[gdb.events.register_changed] = []
 
 
 # When performing remote debugging, gdbserver is very noisy about which
--- a/pwndbg/gdblib/events.py
+++ b/pwndbg/gdblib/events.py
@@ -117,12 +117,8 @@ registered: dict[Any, list[Callable]] = {
     gdb.events.before_prompt: [],  # The real event might not exist, but we wrap it
 }
 
-# GDB 7.9 and above only
-try:
-    registered[gdb.events.memory_changed] = []
-    registered[gdb.events.register_changed] = []
-except (NameError, AttributeError):
-    pass
+registered[gdb.events.memory_changed] = []
+registered[gdb.events.register_changed] = []
 
 
 # When performing remote debugging, gdbserver is very noisy about which
--- a/pwndbg/gdblib/regs.py
+++ b/pwndbg/gdblib/regs.py
@@ -20,19 +20,8 @@ from pwndbg.lib.regs import reg_sets
 
 
 @pwndbg.gdblib.proc.OnlyWhenRunning
-def gdb77_get_register(name: str):
-    return gdb.parse_and_eval("$" + name)
-
-
-@pwndbg.gdblib.proc.OnlyWhenRunning
-def gdb79_get_register(name: str):
+def gdb_get_register(name: str):
     return gdb.selected_frame().read_register(name)
-
-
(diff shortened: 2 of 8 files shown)


disconnect3d:
That's irrelevant for the PR but eh, Arch Linux changed their glibc or sth? 
image

pwndbg/gdblib/events.py
```diff
@@ -115,15 +115,10 @@ def invoke_callbacks(self) -> None:
...
+    gdb.events.memory_changed: [],
+    gdb.events.register_changed: []
 }
 
-# GDB 7.9 and above only
-try:
-    registered[gdb.events.memory_changed] = []
-    registered[gdb.events.register_changed] = []
-except (NameError, AttributeError):
-    pass
```
disconnect3d, line 125: Refactored this further by moving it up to the dict

pwndbg/gdblib/regs.py
```diff
@@ -48,21 +37,15 @@ class module(ModuleType):
     def __getattr__(self, attr: str) -> int:
         attr = attr.lstrip("$")
         try:
-            # Seriously, gdb? Only accepts uint32.
-            if "eflags" in attr or "cpsr" in attr:
-                value = gdb77_get_register(attr)
```
disconnect3d, line 53: We removed this part: did we check if `pwndbg.gdblib.regs.eflags` work after this?

disconnect3d: (also random thought: `x in attr` was so wrong :D)

pwndbg/gdblib/regs.py
```diff
@@ -48,21 +37,15 @@ class module(ModuleType):
...
-                    value += self.cs * 16
-
+            value = gdb_get_register(attr)
+            if value is None and attr.lower() == "xpsr":
+                value = gdb_get_register("xPSR")
+            size = pwndbg.gdblib.typeinfo.unsigned.get(
+                value.type.sizeof, pwndbg.gdblib.typeinfo.ulong
+            )
+            value = value.cast(size)
+            if attr.lower() == "pc" and pwndbg.gdblib.arch.current == "i8086":
```
disconnect3d, line 47: I really wonder if we have to do `.lower()` here at all, I think we should never have here an uppercase value ( ? )

Would need to check different archs if we have there uppercase reg names lol

pwndbg/gdblib/regs.py
```diff
@@ -48,21 +37,15 @@ class module(ModuleType):
...
-                size = pwndbg.gdblib.typeinfo.unsigned.get(
-                    value.type.sizeof, pwndbg.gdblib.typeinfo.ulong
-                )
-                value = value.cast(size)
-                if attr.lower() == "pc" and pwndbg.gdblib.arch.current == "i8086":
-                    value += self.cs * 16
-
+            value = gdb_get_register(attr)
+            if value is None and attr.lower() == "xpsr":
+                value = gdb_get_register("xPSR")
```
disconnect3d, line 42: I wonder if this does not work with lowercase xpsr; this really should have a comment here

pwndbg/gdblib/regs.py
```diff
@@ -186,13 +169,10 @@ def gsbase(self):
 
     @pwndbg.lib.cache.cache_until("stop")
     def _fs_gs_helper(self, regname: str, which):
-        """Supports fetching based on segmented addressing, a la fs:[0x30].
-        Requires ptrace'ing the child directly for GDB < 8."""
```
disconnect3d, line 190: we still do ptrace, but for i386, lets state that in comment

disconnect3d:
@dmur1 I have merged this because its good enough. Can you look at my comments and resolve them in another PR?

dmur1:
> @dmur1 I have merged this because its good enough. Can you look at my comments and resolve them in another PR?

will do this now

pwndbg/gdblib/regs.py
dmur1: not sure what you mean with this one

dmur1:
here you go https://github.com/pwndbg/pwndbg/pull/1881

PR #1863, opened 2023-08-12
adding support for palm, llama2, cohere, huggingface, azure

ghost:
Hi @disconnect3d @dmur1,

Noticed you're calling Anthropic and OpenAI. I'm working on litellm (simple library to standardize LLM API Calls - https://github.com/BerriAI/litellm) and was wondering if we could be helpful.

I added support for palm, cohere, replicate, huggingface, azure by replacing the claude endpoint with litellm. The code is pretty similar to the OpenAI class - as litellm follows the same pattern as the openai-python sdk.

Would love to know if this helps.

Happy to add additional tests / update documentation, if the initial PR looks good to you.

### DIFF
--- a/pwndbg/commands/ai.py
+++ b/pwndbg/commands/ai.py
@@ -22,6 +22,8 @@ from pwndbg.commands import CommandCategory
 from pwndbg.commands import context
 from pwndbg.gdblib import config
 from pwndbg.gdblib import regs as REGS
+import litellm
+from litellm import completion
 
 config.add_param(
     "ai-openai-api-key",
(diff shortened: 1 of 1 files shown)


disconnect3d:
Hi, I am still thinking if we really want it. This currently fails CI bcoz we don't install the dependency u added, but I am not 100% sure if we want to add it. Perhaps this should be an optional dependency and the import should happen when user uses the command for the first time, and if its not there, we should tell them to install and set API keys etc?

disconnect3d:
@krrishdholakia ping

disconnect3d:
I'm going to close this for now since this doesn't even install the required dependency and just doesn't work on CI.

ghost:
hey @disconnect3d any feedback on why this isn't a good fit?

disconnect3d:
@krrishdholakia I can reopen this, but can you then fix the CI and make the dependency optional and being imported on first use?

Basically, I want a good UX here and I want users to be aware that they need to install it themselves additionally and how to install it. I am not sure if I want this dependency to be installed automagically on each Pwndbg installation since I don't think that many Pwndbg users do need/use AI and the ai command...

disconnect3d:
@krrishdholakia ping

kotee4ko:
I think it would be very nasty feature, to link pwngdb with ida and llama. Really nasty.

disconnect3d:
I am going to close this since this has no activity for a long time. Please open another PR with changes that would not break CI.

Issue #808, opened 2020-07-22
Disable search-memory-packet back only on broken GDB version


disconnect3d:
Tl;dr: Use the workaround from https://github.com/pwndbg/pwndbg/pull/322/files only for broken gdb versions

gsingh93:
According to the referenced bug, the issue may have been fixed back in 2020. If so, I'd guess it's in GDB 10, which means 18.04 and 20.04 are still affected (with versions 8.1 and 9.2), but not 22.04 (with version 12).

disconnect3d:
We should check if this is about GDB or GDBServer version or both and if we should only use the workaround if we detect a version that needs it.

disconnect3d:
Fwiw we have removed the `set remote search-memory-packet off ` setting in 77380eb63c6cfdb721f2a2c4448f9d544fa03bd7 accidentally I think :/.

disconnect3d:
This has been fixed in https://github.com/pwndbg/pwndbg/pull/1619. Also, if someone wants to reproduce the search bug/workaround, GDB 9.x is needed as shown in https://github.com/pwndbg/pwndbg/pull/1618.

disconnect3d:
In case someone ever looks here, this workaround has been removed from our codebase in https://github.com/pwndbg/pwndbg/pull/2892 since we only support GDB 12+ now, which does not require this workaround anymore.

PR #1867, opened 2023-08-16
Add new search arguments

fidgetingbits:
This pull request adds three additional search features:

- `-i`/`--interval` to specify the interval between each search result. This is
  useful if you're analyzing something like heap spray where you mostly just
  care if the memory you sprayed exists within some address range, and aren't
  interested in any of the subsequent search hits within that same range. So for instance
  if we're spraying a bunch of data and we want to make sure that every 1MB we hit some
  target address we can do:

```
pwndbg> search -i 0x100000 -t qword 0x0000000000010c00 [anon_97d82]
Searching for value: b'\x00\x0c\x01\x00\x00\x00\x00\x00'
[anon_97d82]    0x982e4d13 0x10c00
[anon_97d82]    0x98327dcf 0x10c00
[anon_97d82]    0x98400038 0x10c00
[anon_97d82]    0x98500038 0x10c00
[anon_97d82]    0x98600038 0x10c00
...
```

We can see that it only prints the first result from each 0x100000 block.

There's two annoyances with doing a search like this, which are addressed by the
other two features.

First, it can be exceptionally slow if the region of memory
are searching in is large (and doing remote debugging), and if you're only
interested in say the first ten results, you don't want to wait around to search 
through the entire region. The existing `trunc_out` feature doesn't help here
because it only truncates the output after a complete search is already completed.

Any example I'm using above the region is 512MB, and it takes a while to search remotely.

```
0x97d82000 0xb674c000 rw-p 1e9ca000      0 [anon_97d82]
```

This problem is addressed by this new option:

- `-l`/`--limit` to specify the maximum number of search results to return.
  This is useful if you're only interested in the first few results, and don't
  want to wait for the entire search to complete. For instance if we're
  spraying a bunch and we want to make sure that every 1MB we hit the target
  address we can do:

```
pwndbg> search -i 0x100000 -l 5 -t qword 0x0000000000010c00 [anon_97d82]
Searching for value: b'\x00\x0c\x01\x00\x00\x00\x00\x00'
[anon_97d82]    0x982e4d13 0x10c00
[anon_97d82]    0x98327dcf 0x10c00
[anon_97d82]    0x98400038 0x10c00
[anon_97d82]    0x98500038 0x10c00
[anon_97d82]    0x98600038 0x10c00
```

Second, some of the results above are not actually the data that were interested in,
for instance the two examples above happen to be the wanted value but at an
alignment that our spray will never actually hit, as in my example case I know
that what I spray will always be 8-byte aligned.

- `-a`/`--align` to specify the alignment of the search results. This is useful
  if you're only interested in results that are aligned to a certain value.
  For instance if we're spraying a bunch and we want to make sure that every
  1MB we hit the target address we can do:

```
pwndbg> search -i 0x100000 -a 0x8 -l 5 -t qword 0x0000000000010c00 [anon_97d82]
Searching for value: b'\x00\x0c\x01\x00\x00\x00\x00\x00'
[anon_97d82]    0x98387510 0x10c00
[anon_97d82]    0x98400038 0x10c00
[anon_97d82]    0x98500038 0x10c00
[anon_97d82]    0x98600038 0x10c00
[anon_97d82]    0x98700f78 0x10c00
```

You can see that earlier results that were at unwanted alignments no
longer show up.

I've included tests which pass on my machine.

As for the name of the commands, I suspect people might not like interval for
instance, I'm happy to change them to whatever you want.

Lastly the try except around the search_memory() call may be a bit questionable, and probably the format of the warning is wrong, so happy to change that if needed.

### DIFF
--- a/pwndbg/commands/search.py
+++ b/pwndbg/commands/search.py
@@ -45,9 +45,13 @@ def print_search_hit(address) -> None:
 auto_save = pwndbg.gdblib.config.add_param(
     "auto-save-search", False, 'automatically pass --save to "search" command'
 )
-
 parser = argparse.ArgumentParser(
-    description="Search memory for byte sequences, strings, pointers, and integer values."
+    formatter_class=argparse.RawTextHelpFormatter,
+    description="""Search memory for byte sequences, strings, pointers, and integer values.
+
+By default search results are cached. If you want to cache all results, but only print a subset, use --trunc-out. If you want to cache only a subset of results, and print the results immediately, use --limit. The latter is specially useful if you're searching a huge section of memory.
+
+""",
 )
 parser.add_argument(
     "-t",
@@ -105,6 +109,23 @@ parser.add_argument(
     "-e", "--executable", action="store_true", help="Search executable segments only"
 )
 parser.add_argument("-w", "--writable", action="store_true", help="Search writable segments only")
+parser.add_argument(
+    "-s",
+    "--step",
+    default=None,
+    type=str,
+    help="Step search address forward to next alignment after each hit (ex: 0x1000)",
+)
+parser.add_argument(
+    "-l",
+    "--limit",
+    default=None,
+    type=str,
+    help="Max results before quitting the search. Differs from --trunc-out in that it will not save all search results before quitting",
+)
+parser.add_argument(
+    "-a", "--aligned", default=None, type=str, help="Result must be aligned to this byte boundary"
+)
 parser.add_argument("value", type=str, help="Value to search for")
 parser.add_argument(
     "mapping_name", type=str, nargs="?", default=None, help="Mapping to search [e.g. libc]"
@@ -126,13 +147,29 @@ parser.add_argument(
     help="Search only locations returned by previous search with --save",
 )
 parser.add_argument(
-    "--trunc-out", action="store_true", default=False, help="Truncate the output to 20 results"
+    "--trunc-out",
+    action="store_true",
+    default=False,
+    help="Truncate the output to 20 results. Differs from --limit in that it will first save all search results",
 )
 
 
 @pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.MEMORY)
 @pwndbg.commands.OnlyWhenRunning
-def search(type, hex, executable, writable, value, mapping_name, save, next, trunc_out) -> None:
+def search(
+    type,
+    hex,
+    executable,
+    writable,
+    step,
+    limit,
+    aligned,
+    value,
+    mapping_name,
+    save,
+    next,
+    trunc_out,
+) -> None:
     global saved
     if next and not saved:
         print(
@@ -155,6 +192,14 @@ def search(type, hex, executable, writable, value, mapping_name, save, next, tru
             print(f"invalid input for type hex: {e}")
             return
 
+    if step:
+        step = pwndbg.commands.fix_int(step)
+
+    if aligned:
+        aligned = pwndbg.commands.fix_int(aligned)
+
+    if limit:
+        limit = pwndbg.commands.fix_int(limit)
     # Convert to an integer if needed, and pack to bytes
     if type not in ("string", "bytes"):
         value = pwndbg.commands.fix_int(value)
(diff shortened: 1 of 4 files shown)


pwndbg/search.py
```diff
@@ -58,7 +72,16 @@ def search(searchfor, mappings=None, start=None, end=None, executable=False, wri
...
 
-            start = i.search_memory(start, length, searchfor)
+            try:
+                start = i.search_memory(start, length, searchfor)
+            except gdb.error:
+                # While remote debugging on an embedded device and searching
+                # through a large memory region (~512mb), gdb may return an error similar
+                # to `error: Invalid hex digit 116`, even though the search
+                # itself is ok. It seems to have to do with a timeout.
+                print("WARN: gdb.search_memory failed, stopping early...")
```
disconnect3d: Ideally this should be a separate PR/commit if possible, but I'm okay leaving this here.

On the other hand, should we maybe check the gdb.error message string or print it to the user as well?

If we think this may be a timeout, we could say it to the user as well that this may happen for embedded debugging

fidgetingbits: I'll break it out into a separate PR then. 

Confusingly the gdb error is `error: Invalid hex digit 116`, but from a bit of digging it seems like this sometimes comes from gdb getting confused during timeouts. The unfortunate thing is that error isn't exactly useful to relay back to the user, but I guess we may as well reported back.

disconnect3d:
I hope to review this on this weekend

pwndbg/commands/search.py
```diff
@@ -132,7 +144,20 @@ def print_search_hit(address) -> None:
...
+    executable,
+    writable,
+    interval,
+    limit,
+    aligned,
+    value,
+    mapping_name,
+    save,
+    next,
+    trunc_out,
```
disconnect3d, line 171: Just to be sure: the trunc_out still makes sense if we consider saving of results, right?

fidgetingbits: Ya that was my thinking and why I left them separate. I suppose it may be worth documenting that in the usage somewhere to make the distinction clear?

disconnect3d: Yeah, ideally we should probably also add example usages in the docstring too

disconnect3d: (for that https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/commands/vmmap.py#L46 would need to be used so that we can use newlines in the docstring heh)

disconnect3d:
Btw this looks good to me but it needs to be rebased now to be merged

pwndbg/commands/search.py
```diff
@@ -105,6 +104,19 @@ def print_search_hit(address) -> None:
     "-e", "--executable", action="store_true", help="Search executable segments only"
 )
 parser.add_argument("-w", "--writable", action="store_true", help="Search writable segments only")
+parser.add_argument(
+    "-i",
+    "--interval",
```
disconnect3d: Shall we call it `-d --distance`? Or any other ideas?

cc: @gsingh93 @Arusekk @lebr0nli for opinion

fidgetingbits: fwiw to me interval makes sense (which is why I named it that way), as a search interval implies a repetition with a fixed value. I just am not sure it would be obvious to others, especially none-native english speakers. But --distance doesn't imply repetition to me so I think might be even more confusing.

disconnect3d: Let's do `-s --step`

pwndbg/commands/search.py
fidgetingbits: I added a couple small comments in the same structure as vmmap just commenting on the difference, but I think probably to add the full documentation for the commands with examples depending on how you want to do it would be a different pull request so not going todo it for now.

fidgetingbits:
I added a couple more basic test cases, because I randomly ran into a separate bug that didn't actually end up being related to this (#808 seemed to trigger on a new gdb client version but with a older gdb server, but I'll file a different bug).

I also added a separate check to make sure that there's actually some map found to search, or error otherwise. I can remove that if you want it to be a separate PR

Otherwise I just addressed your comments about switching interval to step, and re-based, so I think it should be okay to merge now

fidgetingbits:
Ah I seemed to mess up the rebase, so its showing a bunch of other changed files now... will have to fix that

PR #1869, opened 2023-08-16
avoid breaking pacman.conf if entries already exist

fidgetingbits:
When I ran the dev setup script it broke my pacman configuration because it duplicated certain entries, so this pull request adds a check to make sure the lines don't already exist. 

Also netcat might already be installed, but not through gnu-netcat (ex: openbsd-netcat, nmap-netcat, etc) so no longer just blindly try to install gnu-netcat.

A small note on the EOF formatting, although I could use `cat <<-EOF` in order to keep consistent white space within the if condition, the downside is that it will break syntax highlighting, so I settled with the below. I can change it if needed.

### DIFF
--- a/setup-dev.sh
+++ b/setup-dev.sh
@@ -112,17 +112,27 @@ install_apt() {
 install_pacman() {
     set_zigpath "$(pwd)/.zig"
 
-    # add debug repo for glibc-debug
-    cat << EOF | sudo tee -a /etc/pacman.conf
-[core-debug]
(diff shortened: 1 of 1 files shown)


disconnect3d:
@fidgetingbits I think we may have choosed this netcat with a purpose since it supports some flags the other one does not, but I don't remember if that was the case. Would have to find those tests that use netcat under the hood

disconnect3d:
But I am fine with merging it for now.

Issue #1865, opened 2023-08-16
check_dependencies throws exception due to missing requirements.txt


fidgetingbits:
### Description

When an exception is triggered and you have `set exception-verbose on` the pwndbng exception handler will call a function called `check_dependencies` in `pwndbg/exception.py`, which will do the following:

```
    project_path = os.path.dirname(os.path.abspath(__file__))
    requirements_path = os.path.join(project_path, os.pardir, "requirements.txt")
    with open(requirements_path, "r") as f:
```

It seems like the `requirements.txt` file is only present in the `docs/` folder atm, so this triggers its own exception. I confirmed that just symlinking `requirements.txt` to the real file fixes the issue. 

### Steps to reproduce

Run `set exception-verbose on` and cause pwndbg to trigger some python exception.

### My setup

I am using the latest dev branch and am not running in the virtual environment

disconnect3d:
We moved dependencies versions to https://github.com/pwndbg/pwndbg/blob/3d6f86402ea01c69199f0f60ed1f88bc4ddca308/pyproject.toml#L261-L273 

We need to fetch them differently here

dmur1:
will take a look at this

dmur1:
@disconnect3d https://github.com/pwndbg/pwndbg/pull/1880

disconnect3d:
Fixed in #1880. Let's skip check_deps for now... we can bring back this issue if someone reports _yet another issue_ that is about outdated dependencies.

PR #1879, opened 2023-08-17
show cmdline and cwd in procinfo

dmur1:


pwndbg/commands/procinfo.py
```diff
@@ -211,6 +223,10 @@ def procinfo() -> None:
     if not proc.status:
         return
 
+    # print("%-10s %s" % ("cmdline", proc.something))
```
dmur1: for some reason when i try to use proc.cmdline i get an error message like this:

```
pwndbg> procinfo
exe        '/usr/bin/ls'
Python Exception <class 'FileNotFoundError'>: [Errno 2] No such file or directory: '/home/dan/pwndbg/pwndbg/../requirements.txt'
Error occurred in Python: [Errno 2] No such file or directory: '/home/dan/pwndbg/pwndbg/../requirements.txt'
```

( yes the comment shows proc.something my bad but i was testing with proc.cmdline )

dmur1:
doesn't quite work due to an issue i was having getting the cmdline - maybe its cache related but i don't know how this works

pwndbg/commands/procinfo.py
disconnect3d: This comes from https://github.com/pwndbg/pwndbg/issues/1865 - we need to fix this ASAP :/

dmur1:
using https://github.com/pwndbg/pwndbg/pull/1879 i was able to see the actual issue and fix it at least

PR #1888, opened 2023-08-25
remove event wrapper class

### LINKED ISSUE #1854: Drop support for GDB < 9.2

### LINKED ISSUE #1887: Refactor EventWrapper
It seems we can maybe remove EventWrapper class now since we depend on GDB >= 9.2?

https://github.com/pwndbg/pwndbg/issues/1854#issuecomment-1664527390

piotrzarycki:
fixes #1887

### DIFF
--- a/pwndbg/gdblib/prompt.py
+++ b/pwndbg/gdblib/prompt.py
@@ -89,4 +89,4 @@ def set_prompt() -> None:
         prompt = "\x01" + prompt + "\x02"  # SOH + prompt + STX
 
     gdb.execute(f"set prompt {prompt}")
-    gdb.prompt_hook = initial_hook
+gdb.prompt_hook = initial_hook
--- a/pwndbg/gdblib/prompt.py
+++ b/pwndbg/gdblib/prompt.py
@@ -89,15 +89,4 @@ def set_prompt() -> None:
         prompt = "\x01" + prompt + "\x02"  # SOH + prompt + STX
 
     gdb.execute(f"set prompt {prompt}")
-
-
-if pwndbg.gdblib.events.before_prompt_event.is_real_event:
     gdb.prompt_hook = initial_hook
-
-else:
-    # Old GDBs doesn't have gdb.events.before_prompt, so we will emulate it using gdb.prompt_hook
-    def extended_prompt_hook(*a):
-        pwndbg.gdblib.events.before_prompt_event.invoke_callbacks()
-        return prompt_hook(*a)
-
-    gdb.prompt_hook = extended_prompt_hook
--- a/pwndbg/gdblib/events.py
+++ b/pwndbg/gdblib/events.py
@@ -67,44 +67,6 @@ class StartEvent:
 
 gdb.events.start = StartEvent()
 
-
-class EventWrapper:
-    """
-    Wrapper for GDB events which may not exist on older GDB versions but we still can
-    fire them manually (to invoke them you have to call `invoke_callbacks`).
-    """
-
-    def __init__(self, name: str) -> None:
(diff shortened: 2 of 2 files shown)


pwndbg/gdblib/prompt.py
```diff
@@ -89,15 +89,4 @@ def set_prompt() -> None:
         prompt = "\x01" + prompt + "\x02"  # SOH + prompt + STX
 
     gdb.execute(f"set prompt {prompt}")
-
-
-if pwndbg.gdblib.events.before_prompt_event.is_real_event:
     gdb.prompt_hook = initial_hook
```
disconnect3d: Shouldn't this line be unindented here? If not removed (probably not)?
```suggestion
gdb.prompt_hook = initial_hook
```

because otherwise it will be in the `set_prompt` function defined above.

disconnect3d (changes requested):
~See my comment~ EDIT: I went ahead and made the small change myself.

disconnect3d (approved):

disconnect3d:
Awesome, thanks!

PR #1890, opened 2023-08-27
Improve attachp output

### LINKED ISSUE #1873: attachp: display more info on multiple processes with the same name
We should provide more info on the list of processes when `attachp` finds >1 process.

Currently that is:
```
pwndbg> attachp bash
Found pids: 13008, 7940, 3639 (use `attach <pid>`)
```

Some ideas:
* We can show the user/uid who spawned the process
* We can show its command line, if available -- but pls truncated to some sane length
* Can we show if there is some parent/child/grandchild relation between them? (that would be awesome if its possible)

Also, the `(use attach <pid>)` should really suggest using `attachp` and not `attach` ;).

piotrzarycki:
fixes: #1873

```
pwndbg> attachp tmp3crxpri6
PID      USER            COMMAND                                             PROCESS TREE
567907   root            /tmp/tmp3crxpri6                                    tmp3crxpri6(567907)
567906   root            /tmp/tmp3crxpri6                                    tmp3crxpri6(567906)

```

### DIFF
--- a/pwndbg/commands/attachp.py
+++ b/pwndbg/commands/attachp.py
@@ -37,7 +37,9 @@ Original GDB attach command help:
 )
 
 parser.add_argument("target", type=str, help="pid, process name or device file to attach to")
-parser.add_argument("--show_all", action="store_true", help="showing all output process tree and command")
+parser.add_argument(
+    "--show_all", action="store_true", help="showing all output process tree and command"
+)
 
 
 @pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.START)
@@ -80,9 +82,7 @@ def attachp(target, show_all=False) -> None:
                     pid = int(pid)
 
                     user = check_output(["ps", "-o", "user=", "-p", str(pid)]).decode().strip()
-                    command = (
-                        check_output(["ps", "-o", "cmd=", "-p", str(pid)]).decode()
-                    )
+                    command = check_output(["ps", "-o", "cmd=", "-p", str(pid)]).decode()
                     process_tree = get_process_tree(pid, max_depth=2, full=show_all)
 
                     if len(command) >= 40 and not show_all:
@@ -109,7 +109,7 @@ def get_process_tree(pid, max_depth=2, indent=0, full=False):
         if depth > max_depth and not full:
             return
 
-        process_info = process.as_dict(attrs=['pid', 'name'])
+        process_info = process.as_dict(attrs=["pid", "name"])
         process_str = " " * indent + f"{process_info['name']}({process_info['pid']}) \n"
 
         if depth < max_depth:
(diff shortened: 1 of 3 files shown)


pwndbg/commands/attachp.py
```diff
@@ -71,7 +72,32 @@ def attachp(target) -> None:
...
+                    "{:<8} {:<15} {:<51} {:<51}".format("PID", "USER", "COMMAND", "PROCESS TREE")
+                )
+                for pid in pids:
+                    pid = int(pid)
+
+                    user = check_output(["ps", "-o", "user=", "-p", str(pid)]).decode().strip()
+                    command = (
+                        check_output(["ps", "-o", "cmd=", "-p", str(pid)]).decode().strip()[:40]
+                    )
+                    process_tree = check_output(["pstree", "-p", str(pid)]).decode().strip()[:40]
```
disconnect3d: What if pstree is not installed? :P 

Ideally it would be best to read data from `/proc/$pid/status` and other files, so that we don't depend on `ps` and `pstree` :/

pwndbg/commands/attachp.py
```diff
@@ -71,7 +72,32 @@ def attachp(target) -> None:
                 return
 
             if len(pids) > 1:
-                print(message.warn(f"Found pids: {', '.join(pids)} (use `attach <pid>`)"))
+                pidsData.append(
+                    "{:<8} {:<15} {:<51} {:<51}".format("PID", "USER", "COMMAND", "PROCESS TREE")
```
disconnect3d: should use f-string literal formatting

disconnect3d: btw the `"{:<8} {:<15} {:<51} {:<51}"` string should probably be a constant or sth, since it is used below too.

Also we use a `tabular` module in pwndbg - maybe we should use it for rendering here?

piotrzarycki: yep - thx for advice, will use it

pwndbg/commands/attachp.py
piotrzarycki: yep, pstree probably also uses /proc :) (will change that )

disconnect3d:
@piotrzarycki ping

disconnect3d:
@piotrzarycki ping! :)

disconnect3d:
@piotrzarycki fwiw this now breaks tests related to `attachp` command - can you fix them?

Also, when we display many pids etc we should tell the user that we weren't able to attach since there are multiple processes/pids or sth like that

disconnect3d:
This is how it looks now:
image

@piotrzarycki Can we maybe add some flags to force display the whole command or/and process tree?

disconnect3d:
@piotrzarycki there's a test that is failing; are u going to fix it? 🤐

piotrzarycki:
yes :)

disconnect3d:
Closing this in favor of https://github.com/pwndbg/pwndbg/pull/1956

PR #1900, opened 2023-09-21
Update tests to gracefully handle multiple qemu processess running

fidgetingbits:
Previously test scripts would just indiscriminately kill all qemu processes on the system. This would kill other debug sessions I had running. These changes make the test scripts record the qemu pids they run and only kill those.

The old scripts would also not allow you to specify a gdb port, so if you were already running a debug session with port 1234, the tests would fail. This update allows you to pass --gdb-port=NNNN to use a non-default port.

Also update gdb-pt-dump submodule as it has been updated recently to not throw an exception when multiple qemu processes are running. The exception thrown in the event of a failure also changed, so this has also been updated on the pwndbg side.

### DIFF
--- a/pwndbg/gdblib/vmmap.py
+++ b/pwndbg/gdblib/vmmap.py
@@ -434,7 +434,7 @@ def kernel_vmmap_via_page_tables():
     p = pt.PageTableDump()
     try:
         p.lazy_init()
-    except PermissionError:
+    except Exception:
         print(
             M.error(
                 "Permission error when attempting to parse page tables with gdb-pt-dump.\n"
--- a/tests/qemu-tests/run_qemu_system.sh
+++ b/tests/qemu-tests/run_qemu_system.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 ARCH=""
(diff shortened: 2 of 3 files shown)


fidgetingbits:
Not sure why the lint indicates failure as doesn't show anything, and seems to work locally.

fidgetingbits:
Just in testing this a bit more, I think we want to kill it on failure. It kind
of accidentally worked before because aarch64 test would be skipped on my system
(for some reason, not sure why), then the pkill would kill any previous qemu 
instances hanging around. Whereas now if a test fails, we leave it running, and 
then re-run the script without first killing the one that is left lying around, 
it hangs because qemu fails to get a lock on the image.

If leaving it around to try to debug the actual failed test case image is
useful, can add a command-line argument for it or something.

disconnect3d:
> If leaving it around to try to debug the actual failed test case image is
> useful, can add a command-line argument for it or something.
I think it may be useful, so lets add this

fidgetingbits:
I introduced an option now (-Q for qemu, but can change to whatever). Also made some other bits a bit more robust, as ran into some other confusing and silent failures.

PR #1914, opened 2023-10-04
Enhance thread context

### LINKED ISSUE #1904: Enhance thread context
The thread context can be improved so that:
1. It shows the number of all threads in the top bar e..g `[ THREADS (20) ]`
2. It limits the number of displayed threads via some config variable
* When there are more threads then the limit, it should display current thread and other threads up to the limit and the following text: "... (use `ctx thread` or `set context-threads-limit <N>`)"

CptGibbon:
This PR makes the following changes to the "threads" section of the `context` command's output:
* Add total thread count to banner
* Limit number of threads displayed
* Add `context-max-threads` config variable
* No longer attempt to get frames from running threads (causes an error)

Old behaviour:
ctx-threads-old

New behaviour:
ctx-threads-new
ctx-threads-new-running

Resolves #1904

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -81,6 +81,11 @@ config_context_sections = pwndbg.gdblib.config.add_param(
     "regs disasm code ghidra stack backtrace expressions threads",
     "which context sections are displayed (controls order)",
 )
+config_max_threads_display = pwndbg.gdblib.config.add_param(
+    "context-max-threads",
+    4,
+    "maximum number of threads displayed by the context command",
+)
 
 # Storing output configuration per section
 outputs: dict[str, str] = {}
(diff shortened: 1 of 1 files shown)


pwndbg/commands/context.py
```diff
@@ -851,48 +856,57 @@ def get_thread_status(thread):
...
         )
-        if symbol:
-            line += f" <{pwndbg.color.bold(pwndbg.color.green(symbol))}> "
+
+        if thread.is_stopped():
+            thread.switch()
+            frame = gdb.selected_frame()
+
+            pc = M.get(frame.pc())
+            symbol = pwndbg.gdblib.symbol.get(frame.pc())
```
```py
pc = gdb.selected_frame().pc()
pc, symbol = M.get(pc), pwndbg.gdblib.symbol.get(pc)
```

pwndbg/commands/context.py
```diff
@@ -851,48 +856,57 @@ def get_thread_status(thread):
...
+            line += f"{pc}"
+            if symbol:
+                line += f" <{pwndbg.color.bold(pwndbg.color.green(symbol))}> "
+
         out.append(line)
 
-    out.insert(0, pwndbg.ui.banner("threads", target=target, width=width))
+    if num_threads_not_shown:
+        out.append(
+            pwndbg.lib.tips.color_tip(
```
disconnect3d, line 904: Lol, nice idea to use that here!

pwndbg/commands/context.py
CptGibbon: Why 🤔

CptGibbon:
@disconnect3d are you happy with:
* The config variable name "context-max-threads"
* The default value of said config variable (4)
* The total threads being displayed in the banner

pwndbg/commands/context.py
CptGibbon: Ah I assume you want to remove the redundant call to `frame.pc()` 👍

disconnect3d:
> @disconnect3d are you happy with:
> 
> * The config variable name "context-max-threads"
> * The default value of said config variable (4)
> * The total threads being displayed in the banner

Yeah, I think all that makes sense :)

PR #1916, opened 2023-10-05
Port gdb-tests from bash to python

### LINKED ISSUE #1687: Port testing scripts to Python
Our testing scripts are getting complex enough that `bash` scripts are getting difficult to maintain. I think it's time to move these to Python.

intrigus-lgtm:
This tries to fix https://github.com/pwndbg/pwndbg/issues/1687
Trying this PR: `./testspy.sh` in the top-level directory. This will be removed in the future, but is kept for testing.
Limitations:
- If you run the tests from a terminal, make sure to not resize the terminal window.
Otherwise the tests may fail.
I'm not sure what the best way to fix this would be...
- I'm not sure whether we still need `--keep` because it does not make much sense without `parallel`.
We could rename this or repurpose it as "keep output and store in file" or something.
- The code has partially been ported by Copilot but verified by me as being sane.


disconnect3d:
@gsingh93 wdyt?

intrigus-lgtm:
> If you run the tests from a terminal, make sure to not resize the terminal window.
Otherwise the tests may fail.
I'm not sure what the best way to fix this would be...

This is actually also a problem with the existing tests when running them with `--serial`.

disconnect3d:
I think we can probably proceed with merging this... Can we remove the redundant comments etc from the Python code? E.g. if we can't support serial/keep or sth its probably fine. We can work around those issues when its needed to do so 🗡️ 

cc: @intrigus-lgtm

intrigus-lgtm:
@disconnect3d there is an upcoming CTF and travel so I probably won't have time to do this until in maybe 2 weeks.

disconnect3d:
@intrigus ping! @intrigus-lgtm

intrigus-lgtm:
@disconnect3d time flies; pong.
Improved the implementation and cleaned it up, but did not yet have time to push it.

intrigus-lgtm:
@disconnect3d please take another look and squash when merging.
This is even slightly faster:
```shell
$ time ./testspy.sh > /dev/null 

real	0m29,485s
user	3m46,975s
sys	0m28,535s
$ time ./tests.sh > /dev/null 

real	0m33,607s
user	4m1,247s
sys	0m34,334s
```

Issue #1874, opened 2023-08-16
Port checksec command to native command and add colors


disconnect3d:
This is annoying IMHO. I want colors and no pwntools update checks here!
![image](https://github.com/pwndbg/pwndbg/assets/10009354/9d1affbb-413c-43d7-872d-5b33e22cfad9)

disconnect3d:
Fwiw:
![image](https://github.com/pwndbg/pwndbg/assets/10009354/cd32dc8a-460d-44a0-ae81-1e23143f6133)

e.g. partial RELRO should be yellow etc.

CristhianMotoche:
Hello there! I think I could give a hand to this issue. Let me give it a try.

disconnect3d:
The coloring was added in #1923.

I will still keep this open: we should use pwntools dependency to call checksec instead of a wrapper binary

dmur1:
> The coloring was added in #1923.
> 
> I will still keep this open: we should use pwntools dependency to call checksec instead of a wrapper binary

what exactly does this mean?

that we should use ELF directly like pwn checksec does? 

https://github.com/Gallopsled/pwntools/blob/dev/pwnlib/commandline/checksec.py#L38

disconnect3d:
@dmur1 tl;dr yeah, sorry I forgot to answer here. @diivi already started to work on this.

PR #1923, opened 2023-10-20
fix: Issue 1874 - Colorize checksec output

CristhianMotoche:
Hello there! I took a look at #1874 and I decided to start my first contribution to this cool project. My approach was to parse the result of `call_cmd` which the output of `checksec`. I parsed the output and added the color in a similar way to the response of `checksec`.

image

*NOTE:* Maybe there is another approach that doesn't involve parsing the command output since that is a little bit flaky. I tried making a call to `os.system` but it didn't work.

Solves issue #1874 

---

As part of [Stack Builders Inc.](https://www.stackbuilders.com) Hacktoberfest 2023 event.

### DIFF
--- a/pwndbg/commands/checksec.py
+++ b/pwndbg/commands/checksec.py
@@ -19,7 +19,7 @@ def color_line(line: str) -> str:
 
 
 def color_lines(output: str) -> str:
-    return NEW_LINE.join(list(map(color_line, output.split(NEW_LINE))))
+    return "\n".join(map(color_line, output.split(NEW_LINE)))
 
 
 @pwndbg.commands.ArgparsedCommand("Prints out the binary security settings using `checksec`.")
--- a/pwndbg/commands/checksec.py
+++ b/pwndbg/commands/checksec.py
@@ -9,7 +9,7 @@ import pwndbg.wrappers.checksec
 
 
 def color_line(line: str) -> str:
-    return (
+    return pwndbg.color.normal(
         line.replace("*", pwndbg.color.green("*"))
         .replace(":", f":{pwndbg.color.GREEN}")
         .replace("No", f"{pwndbg.color.RED}No")
--- a/pwndbg/commands/checksec.py
+++ b/pwndbg/commands/checksec.py
@@ -2,39 +2,29 @@ from __future__ import annotations
 
 from typing import cast
 
+import pwndbg.color
 import pwndbg.commands
 import pwndbg.gdblib.file
 import pwndbg.wrappers.checksec
 
-NEW_LINE = "\n"
-
-
-class colors:
-    RED = "\033[31m"
-    GREEN = "\033[32m"
-    BLUE = "\033[34m"
-    RESET = "\033[0m"
-
 
 def color_line(line: str) -> str:
     return (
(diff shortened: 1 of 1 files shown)


disconnect3d (changes requested):
Hi, we have `pwndbg.color.*` functions to color things - let's use them instead of reimplementing coloring :)

CristhianMotoche:
Thanks for the tip @disconnect3d Let me know if it looks a bit better now.

pwndbg/commands/checksec.py
```diff
@@ -1,11 +1,30 @@
...
+        .replace(":", f":{pwndbg.color.GREEN}")
+        .replace("No", f"{pwndbg.color.RED}No")
+    )
+
+
+NEW_LINE = "\n"
+
+
+def color_lines(output: str) -> str:
+    return NEW_LINE.join(list(map(color_line, output.split(NEW_LINE))))
```
disconnect3d:
```suggestion
    return "\n".join(map(color_line, output.split(NEW_LINE)))
```

pwndbg/commands/checksec.py
```diff
@@ -1,11 +1,30 @@
...
 
+def color_line(line: str) -> str:
+    return pwndbg.color.normal(
+        line.replace("*", pwndbg.color.green("*"))
+        .replace(":", f":{pwndbg.color.GREEN}")
+        .replace("No", f"{pwndbg.color.RED}No")
+    )
+
+
+NEW_LINE = "\n"
```
disconnect3d:
```suggestion
```

pwndbg/commands/checksec.py
```diff
@@ -1,11 +1,29 @@
...
+def color_line(line: str) -> str:
+    return pwndbg.color.normal(
+        line.replace("*", pwndbg.color.green("*"))
+        .replace(":", f":{pwndbg.color.GREEN}")
+        .replace("No", f"{pwndbg.color.RED}No")
+    )
+
+
+
+
```
disconnect3d:
```suggestion
```

pwndbg/commands/checksec.py
```diff
@@ -1,11 +1,28 @@
...
 
 
+def color_line(line: str) -> str:
+    return pwndbg.color.normal(
+        line.replace("*", pwndbg.color.green("*"))
+        .replace(":", f":{pwndbg.color.GREEN}")
+        .replace("No", f"{pwndbg.color.RED}No")
+    )
+
+
```
disconnect3d, line 18:
```suggestion
```

pwndbg/commands/checksec.py
```diff
@@ -1,11 +1,27 @@
...
+def color_lines(output: str) -> str:
+    return "\n".join(map(color_line, output.split(NEW_LINE)))
```
disconnect3d:
```suggestion
    return "\n".join(map(color_line, output.split("\n")))
```

disconnect3d (approved):

disconnect3d:
I'm going to merge this since this should just work. However, in the future we should just use checksec from pwntools since it is our dependency right now

PR #1925, opened 2023-10-21
Telescope output - print offset to framepointer

OBarronCS:
This pull request contains changes to the "telescope" function when used to print stack addresses. It adds the ability to print the offset to the frame pointer in the output. 

Telescope will print the offset in the same place that the register strings occupy, but only at addresses where a register currently isn't printed. See images. 

This adds some great context when debugging, letting you know how large a stack frame is and where values are, and gives the ability to pinpoint variables in the case they are referenced via an offset to the base pointer.

frameoffset

frameoffset2

frameoffset3

I made a setting to enable/disable this, `telescope-frame-offset`, but that's not necessarily needed. Additionally, it will only print if the  distance between a given address and the framepointer is less than 0xFFF, to prevent it from printing extremely large values where the distance between the frame pointer and base pointer is not relevent.

Additionally, the `stack` function, which calls `telescope` , was missing the `-f` flag to print the entire stack frame. It's now added and forwards the argument to the call to `telescope`

Also added is the `stackf` function for convenience, which is equivalent to `stack -f`.

### DIFF
--- a/pwndbg/commands/telescope.py
+++ b/pwndbg/commands/telescope.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 import argparse
 import collections
 import math
+from typing import DefaultDict
 
 import pwndbg.chain
 import pwndbg.color.telescope as T
@@ -33,6 +34,9 @@ skip_repeating_values_minimum = pwndbg.gdblib.config.add_param(
     "telescope-skip-repeating-val-minimum",
     3,
     "minimum amount of repeated values before skipping lines",
+)
+print_framepointer_offset = pwndbg.gdblib.config.add_param(
+    "telescope-framepointer-offset", True, "print offset to framepointer for each address, if sufficiently small"
 )
 
 offset_separator = theme.add_param(
@@ -105,7 +109,7 @@ def telescope(address=None, count=telescope_lines, to_string=False, reverse=Fals
     if reverse:
         address -= (count - 1) * ptrsize
 
-    # Allow invocation of telescope -f (--frame) to dump frame addresses
+    # Allow invocation of telescope -f (--frame) to dump all addresses in a frame
     if frame:
         sp = pwndbg.gdblib.regs.sp
         bp = pwndbg.gdblib.regs[pwndbg.gdblib.regs.frame]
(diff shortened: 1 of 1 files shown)


pwndbg/commands/telescope.py
```diff
@@ -183,24 +191,37 @@ def collapse_repeating_values() -> None:
...
-        line = " ".join(
+        line = T.offset(
+            "%02x%s%04x%s"
+            % (
+                i + telescope.offset,
+                delimiter,
+                addr - start + (telescope.offset * ptrsize),
+                separator,
+            )
+        )
```
disconnect3d: or just `+ " ".join(...)` here instead of `line +=`?

pwndbg/commands/telescope.py
```diff
@@ -183,24 +191,37 @@ def collapse_repeating_values() -> None:
             result.extend(collapse_buffer)
         collapse_buffer.clear()
 
+    bp = pwndbg.gdblib.regs[pwndbg.gdblib.regs.frame]
+
+    def regs_or_frame_offset(addr):
```
disconnect3d: Each execution of the command will re-create the function object; shall we maybe put it in global scope ( ? )

pwndbg/commands/telescope.py
```diff
@@ -183,24 +191,37 @@ def collapse_repeating_values() -> None:
             result.extend(collapse_buffer)
         collapse_buffer.clear()
 
+    bp = pwndbg.gdblib.regs[pwndbg.gdblib.regs.frame]
```
disconnect3d: Shall this maybe be done only if `print_framepointer_offset=True`?

disconnect3d: Also: do all archs have `regs.frame`?

disconnect3d:
What happens when we do `telescope <heap_addr>`? Can u show a screen?

OBarronCS:
Made the suggested changes. Right, many arches, as listed in  `pwndbg/lib/regs.py`, don't have dedicated frame pointers. Added a check, `pwndbg.gdblib.regs.frame is not None` - however this could probably be optimized to avoid the double lookup of `.frame` every time - the value will stay the same. Maybe we can cache a "has_frame_pointer" somewhere?

When printing any address that doesn't have the framepointer near it, the output will be identical as before. Below is what it looks when printing a heap address, for example.

heap_frameoffset

OBarronCS:
Turns out that the None check is redundant. `pwndbg.gdblib.regs.frame` will return None if no frame pointer is on the arch, and `pwndbg.gdblib.regs[None]` will also return None in this case, due to the the __getattr__ of the regs object: 

https://github.com/pwndbg/pwndbg/blob/d497bac847d60bb5eb259a4b5989cc0efe2091b4/pwndbg/gdblib/regs.py#L36-L52

PR #1926, opened 2023-10-23
Distance command can now print the offset to the addresses's page base

OBarronCS:
This makes a change to the `distance` command that I find myself often needing. When used with a single argument, `distance` will now print the offset of the address from the base address of the page it belongs to.

distance

### DIFF
--- a/pwndbg/commands/distance.py
+++ b/pwndbg/commands/distance.py
@@ -22,11 +22,9 @@ def distance(a, b) -> None:
         total_pages = pwndbg.gdblib.vmmap.get()
 
         if not total_pages:
-            print(
-                "There are no memory pages in `vmmap`"
-            )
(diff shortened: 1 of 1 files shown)


OBarronCS:
Perhaps this logic of finding offset to page base is best put into another file - in gdblib? Can see uses for it in another places.

OBarronCS:
> Perhaps this logic of finding offset to page base is best put into another file - in gdblib? Can see uses for it in another places.

Looks like `pwndbg.gdblib.vmmap.find` already exists to find the page corresponding to an address. Using that now.

disconnect3d:
Shall we maybe also display the name of the vmmap it belongs to?

Btw there is also `xinfo <addr>` command :P

disconnect3d:
Merging this for now. The further improvements can be made in another PR if anyone cares ;).

PR #1928, opened 2023-10-29
gdbinit: use activate_this.py to activate virtualenv

mochaaP:
Fix loading native libs on platforms where libdir != "lib",
e.g. Fedora Linux.

### DIFF
--- a/gdbinit.py
+++ b/gdbinit.py
@@ -1,13 +1,11 @@
 from __future__ import annotations
 
 import cProfile
-import glob
 import locale
 import os
 import site
 import sys
 import time
-from glob import glob
 from os import environ
 from os import path
 
@@ -30,33 +28,23 @@ else:
     if not venv_path:
         venv_path = os.path.join(directory, ".venv")
 
-    if not os.path.exists(venv_path):
+    activate_this = os.path.join(venv_path, "bin/activate_this.py")
+
+    if not os.path.exists(activate_this):
         print(f"Cannot find Pwndbg virtualenv directory: {venv_path}: please re-run setup.sh")
         sys.exit(1)
 
-    site_pkgs_path = glob(os.path.join(venv_path, "lib/*/site-packages"))[0]
-
-    # add virtualenv's site-packages to sys.path and run .pth files
-    site.addsitedir(site_pkgs_path)
+    exec(open(activate_this).read(), {"__file__": activate_this})
 
     # remove existing, system-level site-packages from sys.path
     for site_packages in site.getsitepackages():
         if site_packages in sys.path:
             sys.path.remove(site_packages)
 
-    # Set virtualenv's bin path (needed for utility tools like ropper, pwntools etc)
-    bin_path = os.path.join(venv_path, "bin")
-    os.environ["PATH"] = bin_path + os.pathsep + os.environ.get("PATH")
-
     # Add gdb-pt-dump directory to sys.path so it can be imported
     gdbpt = path.join(directory, "gdb-pt-dump")
     sys.path.insert(0, directory)
     sys.path.insert(1, gdbpt)
-
-    # Push virtualenv's site-packages to the front
-    sys.path.remove(site_pkgs_path)
-    sys.path.insert(2, site_pkgs_path)
-
 
 # warn if the user has different encoding than utf-8
 encoding = locale.getpreferredencoding()


gdbinit.py
```diff
@@ -30,34 +29,24 @@
     if not venv_path:
         venv_path = os.path.join(directory, ".venv")
 
-    if not os.path.exists(venv_path):
+    activate_this = os.path.join(venv_path, "bin/activate_this.py")
```
disconnect3d, line 31: What is this? This file does not exist in my Pwndbg's virtual environment:

```
root@pwndbg:~/pwndbg# ls .venv/bin/
Activate.ps1   checksec       disasm    isort-identify-imports  phd         pyserial-miniterm  rpyc_classic.py   template
ROPgadget      common         dmypy     libcdb                  pip         pyserial-ports     rpyc_registry     tokenize-rt
__pycache__    constgrep      elfdiff   main                    pip3        pytest             rpyc_registry.py  unhex
activate       coverage       elfpatch  mako-render             pip3.10     python             ruff              update
activate.csh   coverage-3.10  errno     markdown-it             pip3.11     python3            scramble          vermin
activate.fish  coverage3      hex       mypy                    pwn         python3.10         shellcraft        version
asm            cyclic         ipython   mypyc                   pwnstrip    pyupgrade          stubgen
black          debug          ipython3  normalizer              py.test     readelf.py         stubtest
blackd         disablenx      isort     pbr                     pygmentize  rpyc_classic       tabulate
root@pwndbg:~/pwndbg#
```

mochaaP:
See <https://github.com/pypa/virtualenv/blob/d7c74fb815390974544fbcdfed6518e66e2e088c/src/virtualenv/activation/python/__init__.py#L11> and <https://github.com/pypa/virtualenv/blob/d7c74fb815390974544fbcdfed6518e66e2e088c/src/virtualenv/activation/python/activate_this.py#L4>.

It was once deprecated (iirc, don't take my words), but brought alive again two years ago.

disconnect3d:
But why is the file missing? Even the CI builds are failing with this PR since the virtualenv is not detected properly anymore (since the file is missing).

mochaaP:
Oops, this seems to be a pyvenv extension, not the stdlib ones (which originates as a subset from the former in 3.3) 😅. `activate_this.py` will still work with the stdlib venv module, but it's not shipped.

We could fall back to the old method when the file is not found or ship `activate_this.py` in the repository. Which way do you prefer?

disconnect3d:
Btw what is the libdir on Fedora? How does activate-this know which libdir to use?

mochaaP:
```sh
$ python -c 'import sys;print(sys.platlibdir)'
lib64
```

Also see the build definition: <https://src.fedoraproject.org/rpms/python3.12/blob/rawhide/f/python3.12.spec#_783>

disconnect3d:
Would this change make it work?:
```diff
-    site_pkgs_path = glob(os.path.join(venv_path, "lib/*/site-packages"))[0]
+    site_pkgs_path = glob(os.path.join(venv_path, "lib*/*/site-packages"))[0]
```

disconnect3d:
(or shall we resolve platlibdir from sys module? huh)

mochaaP:
<https://github.com/pypa/virtualenv/blob/d7c74fb815390974544fbcdfed6518e66e2e088c/src/virtualenv/activation/python/__init__.py#L17-L22>

The virtualenv package does extensive detections, while stdlib's venv module only consists of about 600 lines in a single file. It's a good lightweight solution, but probably bad from programmatically using it. (see virtualenv's docs: <https://virtualenv.pypa.io/en/latest/>)
I think resolving the environment should not be done on our side, since this could still unknowingly break in some setups.

mochaaP:
> (or shall we resolve platlibdir from sys module? huh)

Sadly, this will break importing arch-independent packages on multilib systems 😢

disconnect3d:
> > (or shall we resolve platlibdir from sys module? huh)
> 
> Sadly, this will break importing arch-independent packages on multilib systems 😢

Why? Or: can you give an example? I mean, we are only taking paths from our virtualenv anyway so why should this break stuff?

disconnect3d:
Oh, I guess it would break stuff bcoz it will take libdir path from sys, not from virtualenv. Nvm then.

disconnect3d:
> I think resolving the environment should not be done on our side, since this could still unknowingly break in some setups.

...then we fix or don't support those setups? :P 

Fwiw, we also have fully-contained packages that contain all deps etc and they just work. Here we do some kind of best effort where we rely on some system-level stuff like Python version but we install Python modules with pinned versions, heh.

mochaaP:
will investigate more in the weekend for this pull request - playing a ctf ;)

disconnect3d:
Closing this for now. Please reopen when updated if it makes sense or send another PR

PR #1557, opened 2023-01-30
WIP: Add sigreturn command

gsingh93:
A command to dump the `ucontext_t` at the top of the stack, useful for debugging SROP exploits. Still a WIP, but what I'm most confused about is what to call this command. `sigreturn`, `srop`, `ucontext`, or something else?

PR #1940, opened 2023-11-21
Sigreturn command to print Sigreturn frame in x86_64

OBarronCS:
This pull requests continues the work started in #1557 to print a Sigreturn frame at the top of the stack (or at a specified address).

`sigreturn [address]` will print the values of the common registers in the x86_64 architecture (GPR's, as well as instruction, stack, and frame pointer). Adding the `-a` or `--all` flag will print all the other values of the [rt_sigframe](https://sbexr.rabexc.org/latest/sources/7e/135b6608584cf2.html#0003b00100040001) structure as well. 

sigreturn
sigreturnall

### DIFF
--- a/pwndbg/commands/sigreturn.py
+++ b/pwndbg/commands/sigreturn.py
@@ -28,16 +28,25 @@ parser.add_argument(
     help="Show all values in the frame in addition to common registers",
 )
 
+parser.add_argument(
+    "-p",
+    "--print",
+    dest="print_address",
+    action="store_true",
+    default=False,
+    help="Show addresses of frame values",
+)
+
 
 @pwndbg.commands.ArgparsedCommand(parser)
 @pwndbg.commands.OnlyWhenRunning
 @pwndbg.commands.OnlyWithArch(["x86-64"])
-def sigreturn(address: int = None, display_all=False):
+def sigreturn(address: int = None, display_all=False, print_address=False):
     address = pwndbg.gdblib.regs.sp if address is None else address
 
     arch_name = pwndbg.gdblib.arch.name
     if arch_name == "x86-64":
-        sigreturn_x86_64(address, display_all)
+        sigreturn_x86_64(address, display_all, print_address)
 
 
 SIGRETURN_FRAME_SIZE_x86_64 = 256
@@ -57,7 +66,16 @@ SIGRETURN_REGISTERS_x86_64 = {
 }
 
 
-def sigreturn_x86_64(address: int, display_all: bool):
+
+
+def print_value(string: str, address: int, print_address):
+    addr = ""
+    if(print_address):
+        addr = f"{M.get(address)}: "
+    print(f"{addr}{string}")
+
+
+def sigreturn_x86_64(address: int, display_all: bool, print_address: bool):
     ptr_size = pwndbg.gdblib.arch.ptrsize
 
     # Offset to the stack pointer where the frame values really begins. Start reading memory there.
@@ -66,23 +84,24 @@ def sigreturn_x86_64(address: int, display_all: bool):
 
     mem = pwndbg.gdblib.memory.read(address + frame_start_offset, SIGRETURN_FRAME_SIZE_x86_64)
 
-    for offset, reg in SIGRETURN_FRAME_LAYOUT_x86_64:
+    for stack_offset, reg in SIGRETURN_FRAME_LAYOUT_x86_64:
         # Subtract the offset of start of frame, to get the correct offset into "mem"
-        offset -= frame_start_offset
+        mem_offset = stack_offset - frame_start_offset
 
         regname = C.register(reg.ljust(4).upper())
-        value = pwndbg.gdblib.arch.unpack(mem[offset : offset + ptr_size])
+        value = pwndbg.gdblib.arch.unpack(mem[mem_offset : mem_offset + ptr_size])
 
         if reg in SIGRETURN_REGISTERS_x86_64:
             desc = pwndbg.chain.format(value)
 
-            print(f"{regname} {desc}")
+            print_value(f"{regname} {desc}", address+stack_offset, print_address)
 
         elif reg == "eflags":
             reg_flags = pwndbg.gdblib.regs.flags["eflags"]
             desc = C.format_flags(value, reg_flags)
 
-            print(f"{regname} {desc}")
+            print_value(f"{regname} {desc}", address+stack_offset, print_address)
 
         elif display_all:
-            print(f"{reg} {M.get(value)}")
+
+            print_value(f"{reg} {M.get(value)}", address+stack_offset, print_address)
--- a/pwndbg/commands/sigreturn.py
+++ b/pwndbg/commands/sigreturn.py
@@ -1,0 +1,131 @@
+from __future__ import annotations
+
+import argparse
+
+import pwndbg.color.context as C
+import pwndbg.color.message
+import pwndbg.commands
+import pwndbg.gdblib.arch
+import pwndbg.gdblib.memory
+import pwndbg.gdblib.regs
+
+parser = argparse.ArgumentParser(description="Display the SigreturnFrame at the specific address")
+
+parser.add_argument(
+    "address", nargs="?", default=None, type=int, help="The address to read the frame"
+)
+
+parser.add_argument(
+    "-a",
+    "--all",
+    dest="display_all",
+    action="store_true",
+    default=False,
+    help="Show all values in the frame in addition to registers",
+)
+
+
+@pwndbg.commands.ArgparsedCommand(parser)
+@pwndbg.commands.OnlyWhenRunning
+@pwndbg.commands.OnlyWithArch(["x86-64"])
+def sigreturn(address: int = None, display_all=False):
+    address = pwndbg.gdblib.regs.sp if address is None else address
+
+    arch_name = pwndbg.gdblib.arch.name
+    if arch_name == "x86-64":
+        sigreturn_x86_64(address, display_all)
+    else:
+        print(
+            pwndbg.color.message.error(f"sigreturn does not support the {arch_name} architecture")
+        )
+
+
+SIGRETURN_FRAME_SIZE_x86_64 = 248
+
+# Registers layout from pwntools: https://github.com/Gallopsled/pwntools/blob/e4d3c82501c03de44458ae498a830fe66594f66d/pwnlib/rop/srop.py#L256
+SIGRETURN_FRAME_LAYOUT_x86_64 = {
+    "uc_flags": 0,
+    "&uc": 8,
+    "uc_stack.ss_sp": 16,
+    "uc_stack.ss_flags": 24,
+    "uc_stack.ss_size": 32,
+    "r8": 40,
+    "r9": 48,
+    "r10": 56,
+    "r11": 64,
+    "r12": 72,
+    "r13": 80,
+    "r14": 88,
+    "r15": 96,
+    "rdi": 104,
+    "rsi": 112,
+    "rbp": 120,
+    "rbx": 128,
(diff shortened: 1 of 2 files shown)


pwndbg/commands/sigreturn.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+@pwndbg.commands.ArgparsedCommand(parser)
+@pwndbg.commands.OnlyWhenRunning
+@pwndbg.commands.OnlyWithArch(["x86-64"])
+def sigreturn(address: int = None, display_all=False):
+    address = pwndbg.gdblib.regs.sp if address is None else address
+
+    arch_name = pwndbg.gdblib.arch.name
+    if arch_name == "x86-64":
```
disconnect3d: Is this even needed? I believe this should already be handled by `OnlyWithArch(...)`

pwndbg/commands/sigreturn.py
```diff
@@ -0,0 +1,120 @@
...
+        sigreturn_x86_64(address, display_all)
+    else:
+        print(
+            pwndbg.color.message.error(f"sigreturn does not support the {arch_name} architecture")
+        )
+
+
+SIGRETURN_FRAME_SIZE_x86_64 = 256
+
+# Original registers layout from pwntools, modified below : https://github.com/Gallopsled/pwntools/blob/e4d3c82501c03de44458ae498a830fe66594f66d/pwnlib/rop/srop.py#L256
```
disconnect3d: 1. Why do we modify it?
2. Can't we just pull those values out of pwntools? Pwntools is a Pwndbg dependency :)

cc: @Arusekk

pwndbg/commands/sigreturn.py
```diff
@@ -0,0 +1,120 @@
...
+        ("oldmask", 216),
+        ("cr2", 224),
+        ("&fpstate", 232),
+        ("__reserved", 240),
+        ("sigmask", 248),
+    ]
+)
+
+# Core registers
+SIGRETURN_REGISTERS_x86_64 = set(
```
disconnect3d: Shall this maybe be added to the architecture class description?

disconnect3d: Also, set literals can be created by just `{*a, b, c, ...}`

pwndbg/commands/sigreturn.py
```diff
@@ -0,0 +1,120 @@
...
+    [*amd64_regset.gpr, amd64_regset.frame, amd64_regset.stack, amd64_regset.pc]
+)
+
+
+def sigreturn_x86_64(address: int, display_all: bool):
+    ptr_size = 8  # x86_64
```
disconnect3d: Use `gdblib.arch.ptrsize`?

pwndbg/commands/sigreturn.py
```diff
@@ -0,0 +1,120 @@
...
+        if reg in SIGRETURN_REGISTERS_x86_64:
+            regname = C.register(reg.ljust(4).upper())
+            value = pwndbg.gdblib.arch.unpack(mem[offset : offset + ptr_size])
+            desc = pwndbg.chain.format(value)
+
+            print(f"{regname} {desc}")
+
+        elif reg == "eflags":
+            regname = C.register("eflags".ljust(4).upper())
+            value = pwndbg.gdblib.arch.unpack(mem[offset : offset + ptr_size])
```
disconnect3d, line 111: It seems that the `value = ...unpack(...)` part can be put before the if, since it is the same in both branches

pwndbg/commands/sigreturn.py
```diff
@@ -0,0 +1,120 @@
...
+            reg_flags = pwndbg.gdblib.regs.flags["eflags"]
+            desc = C.format_flags(value, reg_flags)
+
+            print(f"{regname} {desc}")
```
disconnect3d, line 115: Can be out of if

disconnect3d:
cc: @Arusekk @gsingh93 if u guys can look at this as well! :)

pwndbg/commands/sigreturn.py
OBarronCS: Put this here to handle the future case when more architectures may be implemented - to signal that here we should add another "if" check and the call another function for the specific arch.

pwndbg/commands/sigreturn.py
OBarronCS: I can remove the comment - the initial values were from the pwntools repo from there, but I needed to add +8 to all the offsets and swap the keys and values (and add the initial key,value pair ("&pretcode", 0)). We could remove the comment if its misleading, in this case.

pwndbg/commands/sigreturn.py
OBarronCS: I see, the printing of the error message is extraneous and will never get called.

pwndbg/commands/sigreturn.py
disconnect3d: Nah, lets leave the comment.

I was thinking more about just using this from here:
```
In [9]: pwnlib.rop.srop.SigreturnFrame().registers
Out[9]:
{0: 'uc_flags',
 8: '&uc',
 16: 'uc_stack.ss_sp',
 24: 'uc_stack.ss_flags',
 32: 'uc_stack.ss_size',
 40: 'r8',
 48: 'r9',
 56: 'r10',
 64: 'r11',
 72: 'r12',
 80: 'r13',
 88: 'r14',
 96: 'r15',
 104: 'rdi',
 112: 'rsi',
 120: 'rbp',
 128: 'rbx',
 136: 'rdx',
 144: 'rax',
 152: 'rcx',
 160: 'rsp',
 168: 'rip',
 176: 'eflags',
 184: 'csgsfs',
 192: 'err',
 200: 'trapno',
 208: 'oldmask',
 216: 'cr2',
 224: '&fpstate',
 232: '__reserved',
 240: 'sigmask'}
```
This also solves the problem with supporting different architectures - it just works out of the box as long as pwntools supports a given arch.

OBarronCS:
The initial reason for not directly pulling the offsets from pwnlib was to manually modify the offsets and turn them into an OrderedDict, to ensure the iteration order is correct, as it seems dict iteration order is only guaranteed after python 3.7 (https://stackoverflow.com/questions/2053021/is-the-order-of-a-python-dictionary-guaranteed-over-iterations).

However, to reduce code duplication, we could instead pull the values from the [library](https://github.com/Gallopsled/pwntools/blob/bb1d16c3c64f80ec83672683715acf92d7b15535/pwnlib/rop/srop.py#L183), and then modify and ensure they are in the correct order. Doesn't need to really be a dict come to think of it, just a sorted list of (offset, regname) tuples is sufficient.

Perhaps the following makes more sense:
```python
SIGRETURN_FRAME_LAYOUT_x86_64 = sorted([(0, "&pretcode",)] + [(k+8,v) for (k,v) in pwnlib.rop.srop.registers["amd64"].values()])
```

How does this sound?

I don't have first-hand familiarity with the sigreturn frames of other architectures, but looking at the offsets defined in the [pwntools library](https://github.com/Gallopsled/pwntools/blob/bb1d16c3c64f80ec83672683715acf92d7b15535/pwnlib/rop/srop.py#L183), and a quick look at the [different definitions in the kernel](https://elixir.bootlin.com/linux/latest/C/ident/sigframe), it looks like the current code actually would generalize to all of them, with the only difference being the offset to the stack pointer from where the sigreturn frame begins. In x86_64 the first 8 bytes are the signal trampoline address (which a ret instruction has already popped into RIP) so we have to -8 to get to the start of the frame. We would take this offset case-by-case depending on how each arch works.

We would also have to manually validate that the register names in [lib/regs.py](https://github.com/pwndbg/pwndbg/blob/61787b3597a91e86ca80ffc9dcab880b1d2a7466/pwndbg/lib/regs.py#L117) line up with the ones defined in the dicts in pwntools, so we can correctly check if a given register is a "core register" to print without the `--all` flag. At first glance, I believe they do.

pwndbg/commands/sigreturn.py
OBarronCS: Adding them to architecture class might make sense - would avoid the need for any "ifs" in this codepath to get the "core registers" for different architectures, as we can just reference an arch member variable to fetch them.

OBarronCS:
Latest commit takes discussion into account. I pull the values from pwntools - the offsets specified there are the "offset of the value to the stack pointer". So instead of doing any adjusting, we can add negative offsets to get fields on the other side of the stack pointer. This makes it more architecture agnostic. 

To make it all the way there, we could add the "core" registers to the architecture object, maybe even the sigreturn frame offsets too (and frame size) so as to get rid of any `if`s to get the correct offset objects.

OBarronCS:
Latest commit adds options `-p` or `--print` (all ears for a better name) which prints the addresses that the frame values are at

sigreturn_print

pwndbg/commands/sigreturn.py
```diff
@@ -0,0 +1,104 @@
...
+
+    arch_name = pwndbg.gdblib.arch.name
+    if arch_name == "x86-64":
+        sigreturn_x86_64(address, display_all, print_address)
+
+
+SIGRETURN_FRAME_SIZE_x86_64 = 256
+
+# Grab frame values from pwntools. Offsets are defined as the offset to stack pointer when syscall instruction is called
+# Offsets and names are from "CONFIG_X86_64 struct rt_sigframe, Linux Kernel /arch/x86/include/asm/sigframe.h
```
disconnect3d:
```suggestion
# Offsets and names are from CONFIG_X86_64 struct rt_sigframe (Linux Kernel /arch/x86/include/asm/sigframe.h)
```

disconnect3d:
Btw I wondered if we could somehow leverage https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/lib/abi.py for this but idk, doesn't seem like it I think?

OBarronCS:
Latest commit adds support for "aarch64", "i386", and "arm". 

The register sets as defined in ["pwntools.rop.srop"](https://github.com/Gallopsled/pwntools/blob/e4d3c82501c03de44458ae498a830fe66594f66d/pwnlib/rop/srop.py#L190) don't exactly line up with the ones defined in our [lib/regs.py](https://github.com/pwndbg/pwndbg/blob/61787b3597a91e86ca80ffc9dcab880b1d2a7466/pwndbg/lib/regs.py#L117C11-L117C11), but I did some guess work and filled in the missing register names ("sp", "pc", "fp", etc) that are in the pwntools definition for ARM and aarch64, but not in the one in pwndbg. As it stands now, the code should work for all the architectures.

disconnect3d:
@OBarronCS "This branch cannot be rebased due to conflicts" can u solve this so we can merge it? :)

OBarronCS:
Alright, looks like this works - rebased from dev and fixed a conflict manually. Let me know if there's a preferred way to do this.

disconnect3d:
Thanks!

pwndbg/commands/sigreturn.py
```diff
@@ -0,0 +1,108 @@
...
+    "aarch64": sorted(list(pwnlib.rop.srop.registers["aarch64"].items())),
+    "arm": sorted(list(pwnlib.rop.srop.registers["arm"].items())),
+}
+
+# Always print these registers (as well as flag register, eflags / cpsr)
+SIGRETURN_CORE_REGISTER: dict[str, set[str]] = {
+    "x86-64": {*amd64.gpr, amd64.frame, amd64.stack, amd64.pc},
+    "i386": {*i386.gpr, i386.frame, i386.stack, i386.pc},
+    "aarch64": {*aarch64.gpr, "sp", "pc"},
+    "arm": {*arm.gpr, "fp" "ip", "sp", "lr", "pc"},
```
gsingh93, line 34: `"fp", "ip"` :)

PR #1952, opened 2023-12-12
Add mmap command that executes the mmap syscall in the inferior

mbrla0:
This PR adds a `mmap` command that calls the `mmap` system call in the context of the inferior from inside pwndbg, allowing users to dynamically change the virtual memory layout of the program being analyzed.

Example usage:
```
# mmap 0x10000 bytes at any addr with RWX (default) permissions
mmap 0 0x10000

# mmap 4096 bytes at 0xBEEF000 with RW perms
mmap 0xBEEF000 0x1000 PROT_READ|PROT_WRITE MAP_FIXED|MAP_PRIVATE|MAP_ANONYMOUS
```

It achieves this in a manner similar to how the `mprotect` command is currently implemented, with a few extra safeguards. Additionally, it moves that functionality into a new module located at `pwndbg/gdblib/shellcode.py`, and splits it into a function that allows for execution of arbitrary blobs of machine code in the context of the inferior (`exec_shellcode`), and, based on it, a function that allows for convenient invocation of system calls (`exec_syscall`).

The `mmap` command also notifies users of unaligned addresses (though it does not change those addresses on their behalf), and that checks for possibly overwritten mappings when `MAP_FIXED` is used and warns users of them.

Finally, it also makes it so that `mprotect` uses the `exec_syscall` function, rather than have its own separate implementation.

A few questions that I feel need to be addressed:
- [ ] Should we be this worried about making sure users don't overwrite existing mappings accidentally, or should we be more lax?
- [ ] Should tests specifically for `mmap` be added?

### DIFF
--- a/pwndbg/commands/mmap.py
+++ b/pwndbg/commands/mmap.py
@@ -136,7 +136,7 @@ def parse_str_or_int(val, parser):
         return val
     else:
         # Getting here is a bug, we shouldn't be seeing other types at all.
-        raise TypeError("invalid type for value: {type(val)}")
+        raise TypeError(f"invalid type for value: {type(val)}")
 
 
 @pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.MEMORY)
--- a/pwndbg/commands/mmap.py
+++ b/pwndbg/commands/mmap.py
@@ -146,6 +146,7 @@ def mmap(addr, length, prot=7, flags=0x22, fd=-1, offset=0, quiet=False, force=F
         prot_int = parse_str_or_int(prot, prot_str_to_val)
     except ValueError as e:
         print(message.error(f'Invalid protection value "{prot}": {e}'))
+        return        
 
     try:
         flag_int = parse_str_or_int(flags, flag_str_to_val)
--- a/pwndbg/commands/mmap.py
+++ b/pwndbg/commands/mmap.py
@@ -54,7 +54,7 @@ parser.add_argument(
 )
 parser.add_argument(
     "prot",
-    help='Prot string as in mmap(2). Eg. "PROT_READ|PROT_EXEC".',
+    help='Prot enum or int as in mmap(2). Eg. "PROT_READ|PROT_EXEC" or 7 (for RWX).',
     type=str,
     nargs="?",
     default="7",
--- a/pwndbg/commands/mmap.py
+++ b/pwndbg/commands/mmap.py
@@ -152,6 +152,7 @@ def mmap(addr, length, prot=7, flags=0x22, fd=-1, offset=0, quiet=False, force=F
         flag_int = parse_str_or_int(flags, flag_str_to_val)
     except ValueError as e:
         print(message.error(f'Invalid flags value "{flags}": {e}'))
+        return
 
     aligned_addr = int(pwndbg.lib.memory.page_align(addr))
     if flag_int & flag_dict["MAP_FIXED"] != 0:
--- a/pwndbg/commands/mmap.py
+++ b/pwndbg/commands/mmap.py
@@ -1,11 +1,6 @@
 from __future__ import annotations
 
 import argparse
-import bisect
-
-import gdb
-import pwnlib
-from pwnlib import asm
 
 import pwndbg.chain
 import pwndbg.color.message as message
@@ -17,7 +12,6 @@ import pwndbg.lib.memory
 import pwndbg.wrappers.checksec
 import pwndbg.wrappers.readelf
 from pwndbg.commands import CommandCategory
-from pwndbg.lib.regs import reg_sets
 
 parser = argparse.ArgumentParser(
     formatter_class=argparse.RawTextHelpFormatter,
@@ -29,7 +23,7 @@ Note that the mmap syscall may fail for various reasons
 will not be a valid pointer.
 
 PROT values: NONE (0), READ (1), WRITE (2), EXEC (4)
-MAP values: SHARED (1), PRIVATE (2), SHARED_VALIDATE (3), FIXED (0x10), 
+MAP values: SHARED (1), PRIVATE (2), SHARED_VALIDATE (3), FIXED (0x10),
             ANONYMOUS (0x20)
 
 Flags and protection values can be either a string containing the names of the
@@ -38,14 +32,14 @@ protection and flag numbers.
 
 Examples:
     mmap 0x0 4096 PROT_READ|PROT_WRITE|PROT_EXEC MAP_PRIVATE|MAP_ANONYMOUS -1 0
-     - Maps a new private+anonymous page with RWX permissions at a location 
+     - Maps a new private+anonymous page with RWX permissions at a location
        decided by the kernel.
 
     mmap 0x0 4096 PROT_READ MAP_PRIVATE 10 0
      - Maps 4096 bytes of the file pointed to by file descriptor number 10 with
        read permission at a location decided by the kernel.
 
-    mmap 0xdeadbeef 0x1000 
+    mmap 0xdeadbeef 0x1000
      - Maps a new private+anonymous page with RWX permissions at a page boundary
        near 0xdeadbeef.
 """,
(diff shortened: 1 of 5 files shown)


pwndbg/commands/mmap.py
```diff
@@ -0,0 +1,239 @@
...
+    "addr", help="Address hint to be given to mmap.", type=pwndbg.commands.sloppy_gdb_parse
+)
+parser.add_argument(
+    "length",
+    help="Length of the mapping, in bytes. Needs to be greater than zero.",
+    type=int,
+)
+parser.add_argument(
+    "prot",
+    help='Prot string as in mmap(2). Eg. "PROT_READ|PROT_EXEC".',
```
disconnect3d:
```suggestion
    help='Prot enum or int as in mmap(2). Eg. "PROT_READ|PROT_EXEC" or 7 (for RWX).',
```

pwndbg/commands/mmap.py
```diff
@@ -0,0 +1,239 @@
...
+    "--force", "-f", help="Force potentially unsafe actions to happen", action="store_true"
+)
+
+
+prot_dict = {
+    "PROT_NONE": 0x0,
+    "PROT_READ": 0x1,
+    "PROT_WRITE": 0x2,
+    "PROT_EXEC": 0x4,
+}
```
disconnect3d, line 96: Should we maybe also support `RWX` or `rwx`?

pwndbg/commands/mmap.py
```diff
@@ -0,0 +1,239 @@
...
+parser.add_argument(
+    "prot",
+    help='Prot enum or int as in mmap(2). Eg. "PROT_READ|PROT_EXEC" or 7 (for RWX).',
+    type=str,
+    nargs="?",
+    default="7",
+)
+parser.add_argument(
+    "flags",
+    help='Flags string as in mmap(2). Eg. "MAP_PRIVATE|MAP_ANONYMOUS".',
```
disconnect3d:
```suggestion
    help='Flags enum or int as in mmap(2). Eg. "MAP_PRIVATE|MAP_ANONYMOUS" or 0x22.',
```

pwndbg/commands/mmap.py
```diff
@@ -0,0 +1,239 @@
...
+            flag_int |= v
+    return flag_int
+
+
+def parse_str_or_int(val, parser):
+    """
+    Try parsing a string with one of the parsers above or by converting it to
+    an int, or passes the value through if it is already an integer.
+    """
+    if type(val) == str:
```
disconnect3d:
```suggestion
    if type(val) is str:
```
Alternatively: `if isinstance(val, str):` which will support inheritance

pwndbg/commands/mmap.py
```diff
@@ -0,0 +1,239 @@
...
+    "MAP_FIXED": 0x10,
+    "MAP_ANONYMOUS": 0x20,
+}
+
+
+def prot_str_to_val(protstr):
+    """Heuristic to convert PROT_EXEC|PROT_WRITE to integer value."""
+    prot_int = 0
+    for k, v in prot_dict.items():
+        if k in protstr:
```
disconnect3d, line 111: Fwiw `PROT_EXECasdf` will also work

pwndbg/commands/mmap.py
```diff
@@ -0,0 +1,239 @@
...
+    """
+    Try parsing a string with one of the parsers above or by converting it to
+    an int, or passes the value through if it is already an integer.
+    """
+    if type(val) is str:
+        candidate = parser(val)
+        if candidate != 0:
+            return candidate
+        return int(val, 0)
+    elif type(val) == int:
```
disconnect3d:
```suggestion
    elif type(val) is int:
```

pwndbg/commands/mmap.py
```diff
@@ -0,0 +1,239 @@
...
+    if type(val) is str:
+        candidate = parser(val)
+        if candidate != 0:
+            return candidate
+        return int(val, 0)
+    elif type(val) is int:
+        return val
+    else:
+        # Getting here is a bug, we shouldn't be seeing other types at all.
+        raise TypeError("invalid type for value: {type(val)}")
```
disconnect3d:
```suggestion
        raise TypeError(f"invalid type for value: {type(val)}")
```

pwndbg/gdblib/shellcode.py
```diff
@@ -0,0 +1,108 @@
...
+    # Make sure the blob fits in the rest of the space we have in this page.
+    #
+    # NOTE: Technically, we could actually use anything from the whole page to
+    # all of the pages currently mapped as executable for this. There is no
+    # technical limitation stopping us from doing that, but seeing as doing it
+    # is harder to make sure it works correctly, we don't (for now, at least).
+    page = pwndbg.gdblib.vmmap.find(starting_address)
+    assert page is not None
+
+    clearance = (page.vaddr + page.memsz) - len(blob) - 1
```
disconnect3d:
```suggestion
    clearance = page.end - len(blob) - 1
```

pwndbg/commands/mmap.py
```diff
@@ -0,0 +1,239 @@
...
+        raise TypeError(f"invalid type for value: {type(val)}")
+
+
+@pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.MEMORY)
+@pwndbg.commands.OnlyWhenRunning
+def mmap(addr, length, prot=7, flags=0x22, fd=-1, offset=0, quiet=False, force=False) -> None:
+    try:
+        prot_int = parse_str_or_int(prot, prot_str_to_val)
+    except ValueError as e:
+        print(message.error(f'Invalid protection value "{prot}": {e}'))
```
disconnect3d, line 148:
```suggestion
        print(message.error(f'Invalid protection value "{prot}": {e}'))
        return        
```

pwndbg/commands/mmap.py
```diff
@@ -0,0 +1,240 @@
...
+        return        
+
+    try:
+        flag_int = parse_str_or_int(flags, flag_str_to_val)
+    except ValueError as e:
+        print(message.error(f'Invalid flags value "{flags}": {e}'))
```
disconnect3d, line 154:
```suggestion
        print(message.error(f'Invalid flags value "{flags}": {e}'))
        return
```

pwndbg/commands/mmap.py
```diff
@@ -0,0 +1,241 @@
...
+                    f"""\
+Address {addr:#x} is not properly aligned. Calling mmap with MAP_FIXED and an
+unaligned address is likely to fail. Consider using the address {aligned_addr:#x}
+instead.\
+"""
+                )
+            )
+
+        page = pwndbg.lib.memory.Page(addr, int(length), 0, 0)
+        collisions = []
```
disconnect3d: In theory, there should never be more than a single collision (in practice if we have broken vmmap info there will be). We can safely assume a single collision here. Its fine if we print a single one if there are more

disconnect3d:
>  Should we be this worried about making sure users don't overwrite existing mappings accidentally, or should we be more lax?

We can be more lax tbh, such commands will always be "best effort" and are there just as a utility to support "interesting use cases" :D 


> Should tests specifically for mmap be added?

Generally yes. This can be done in separate PR too

pwndbg/commands/mmap.py
mbrla0: Yes, which isn't optimal, but I figured that if it's good enough for `mprotect`, it's probably good enough here too. But I could make it stricter.

disconnect3d: its probably okay for now

pwndbg/commands/mmap.py
mbrla0: I don't think I follow. If any range is permissible for our mmap, it could, in the worst case, technically collide with all of the mappings, no? Nothing's really stopping you from trying to call this with `MAP_FIXED` for `[0, 0xffffffffffffffff[` (I know in practice there are limitations on covering the whole range, but this same logic applies for smaller ranges too).

disconnect3d: Oh... you are right. For some reason I was thinking that we map a single address instead of a range. Forgive my stupidness :)

mbrla0:
> We can be more lax tbh, such commands will always be "best effort" and are there just as a utility to support "interesting use cases" :D

Sure, I just feel that the testing I'm doing might be excessive. I'm thinking of disabling the more expensive range collision checks when run with `--force` instead of still running them and printing it as a warning, to possibly speed up cases where people are using the command in their scripts. How does that sound?

> Generally yes. This can be done in separate PR too

Shouldn't be too much trouble adding it now

mbrla0:
I've added some tests, but it's still missing a test for file based mappings. Should add them later today.

tests/gdb-tests/tests/test_mmap.py
```diff
@@ -0,0 +1,67 @@
...
+
+import pwndbg
+import tests
+
+SMALL_BINARY = tests.binaries.get("crash_simple.out.hardcoded")
+
+
+def test_mmap_executes_properly(start_binary):
+    """
+    Tests the mprotect command
```
disconnect3d:
```suggestion
    Tests the mmap command
```

PR #1956, opened 2023-12-15
Implement a way to resolve multiple process in attachp

key-moon:
Added new parameter `attachp-resolve-method`. This allows processes to be selected using the `first`(select oldest), `last`(select latest), or `ask`(decide interactively) method, rather than using `attach` again when a conflict occurs.
In addition to this, the details of the process will be displayed in a table when a conflict occurs.

image
image

After creating most of the implementation of this pull request, I noticed that a similar pull request was created before(#1890 ). However, that pull request seemed to be inactive, so I decided to open the pull request anyway.
I will leave it up to the maintainer to decide how to handle this issue.

### DIFF
--- a/pwndbg/commands/attachp.py
+++ b/pwndbg/commands/attachp.py
@@ -49,7 +49,7 @@ Original GDB attach command help:
     to specify the program, and to load its symbol table.""",
 )
 
-parser.add_argument("--no-truncate", action="store_true", help="not truncate command name")
+parser.add_argument("--no-truncate", action="store_true", help="dont truncate command args")
 parser.add_argument("target", type=str, help="pid, process name or device file to attach to")
 
 
--- a/pwndbg/commands/attachp.py
+++ b/pwndbg/commands/attachp.py
@@ -5,12 +5,27 @@ import os
 import stat
 from subprocess import CalledProcessError
 from subprocess import check_output
+from typing import Union
 
 import gdb
+from tabulate import tabulate
 
 import pwndbg.commands
 from pwndbg.color import message
 from pwndbg.commands import CommandCategory
+from pwndbg.ui import get_window_size
+
+_NONE = "none"
+_FIRST = "first"
+_LAST = "last"
+_ASK = "ask"
+_OPTIONS = [_NONE, _FIRST, _LAST, _ASK]
+
+pwndbg.gdblib.config.add_param(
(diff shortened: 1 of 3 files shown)


pwndbg/commands/attachp.py
```diff
@@ -71,16 +87,106 @@ def attachp(target) -> None:
...
+
+                    headers = ["pid", "user", "elapsed", "command"]
+                    showindex: Union[bool, range] = (
+                        False if method == _NONE else range(1, len(proc_infos) + 1)
+                    )
+
+                    # calculate max_col_widths to fit window width
+                    test_table = tabulate(proc_infos, headers=headers, showindex=showindex)
+                    table_orig_width = len(test_table.splitlines()[1])
+                    max_command_width = max([len(command) for _, _, _, command in proc_infos])
```
disconnect3d:
```suggestion
                    max_command_width = max(len(command) for _, _, _, command in proc_infos)
```

disconnect3d (commented):
Looks good to me! However, should we maybe make `_ASK` the default method?

We should also inform user they can change the default resolution method when asking users so they are aware this can be configured.

Also, random nitpick: first and last are a bit unambiguous: is it min vs max PID or is it oldest vs newest created process? We could either rename it to 'oldest' and 'newest' or 'minpid' and 'maxpid' (or add those options as well but idk if it makes sense).

pwndbg/commands/attachp.py
```diff
@@ -34,11 +49,12 @@
     to specify the program, and to load its symbol table.""",
 )
 
+parser.add_argument("--no-truncate", action="store_true", help="not truncate command name")
```
disconnect3d:
```suggestion
parser.add_argument("--no-truncate", action="store_true", help="dont truncate command args")
```

key-moon:
Thank you for the reviewing!
The only reason for not making `ask` as the default resolve method was concern that users would be confused by the change. This is resolved by showing them that the config exists and how to change it, which seems to be the right way to do it.
It is true that the naming of flrst/last is ambiguous. I will change these names to oldest/newest.

key-moon:
The points mentioned above have been changed. As I am not a native speaker, I am not sure if the naming and messages are natural overall. If there are any unnatural points, feel free to correct them.

disconnect3d:
Thanks!

PR #1962, opened 2023-12-20
arch.py: fix arch detection for riscv

patryk4815:
arch.py: fix arch detection for riscv. 

Info: https://github.com/bminor/binutils-gdb/blob/f6149394f9a46d03ff853a0e83aae61441182811/gdb/testsuite/gdb.arch/riscv-default-tdesc.exp#L35

Error:
image

### DIFF
--- a/pwndbg/gdblib/arch.py
+++ b/pwndbg/gdblib/arch.py
@@ -9,7 +9,20 @@ from pwndbg.lib.arch import Arch
 
 # TODO: x86-64 needs to come before i386 in the current implementation, make
 # this order-independent
-ARCHS = ("x86-64", "i386", "aarch64", "mips", "powerpc", "sparc", "arm", "armcm", "rv32", "rv64")
+ARCHS = (
+    "x86-64",
+    "i386",
+    "aarch64",
+    "mips",
+    "powerpc",
+    "sparc",
+    "arm",
+    "armcm",
+    "riscv:rv32",
+    "riscv:rv64",
+    "riscv",
+)
+
 
 # mapping between gdb and pwntools arch names
 pwnlib_archs_mapping = {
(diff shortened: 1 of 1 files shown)


pwndbg/gdblib/arch.py
```diff
@@ -9,7 +9,7 @@
 
 # TODO: x86-64 needs to come before i386 in the current implementation, make
 # this order-independent
-ARCHS = ("x86-64", "i386", "aarch64", "mips", "powerpc", "sparc", "arm", "armcm", "rv32", "rv64")
+ARCHS = ("x86-64", "i386", "aarch64", "mips", "powerpc", "sparc", "arm", "armcm", "riscv:rv32", "riscv:rv64", "riscv)
```
disconnect3d: I wonder why did we start with `rv32` and `rv64` in the first place. Perhaps they worked at some GDB version or GDB can report it as such?

patryk4815: here is some interesting comment maybe: https://github.com/bminor/binutils-gdb/commit/895b7b4e4bed9eee9d4a3d55bee876055736bfd3

disconnect3d: @peace-maker Ping, since you worked on https://github.com/pwndbg/pwndbg/pull/1770 do you maybe know some details about the reported arch for riscv in GDB? Did they change it in some version or what?

peace-maker: gdb used `riscv:rv64` as architecture string already when I was working on that PR. But since the other prefixes were ignored in some other arches in this list, I figured I'd do it for riscv as well. I didn't consider the plain `"riscv"` architecture, sorry.

```
pwndbg> set architecture
auto               i386:intel         i386:x64-32:intel  i386:x86-64:intel
i386               i386:x64-32        i386:x86-64        i8086
```

PR #1963, opened 2023-12-20
Update LICENSE.md

disconnect3d:


gsingh93:
FYI, you can't change copyright like this. You should only ever append names, never remove them.

PR #1966, opened 2023-12-20
FIX: pwndbg.gdblib.regs.frame is None

hacksysteam:
Revert c37030633afc2c8f4620c959e9c27168ce32e98e as pwndbg.gdblib.regs.frame can be None


pwndbg/commands/telescope.py
```diff
@@ -192,7 +192,7 @@ def collapse_repeating_values() -> None:
         collapse_buffer.clear()
 
     bp = None
-    if print_framepointer_offset:
+    if print_framepointer_offset and pwndbg.gdblib.regs.frame is not None:
         # If pwndbg.gdblib.regs.frame is None, indexing regs will return None
```
disconnect3d:
```suggestion
        # regs.frame can be None on aarch64
```

PR #1968, opened 2023-12-21
FIX: updating type hints batch 1

hacksysteam:
This PR is the first batch of type hints fixes. It is tough to do everything at once, so I'm sending the pull requests in batches.

### DIFF
--- a/gdbinit.py
+++ b/gdbinit.py
@@ -46,7 +46,7 @@ else:
 
     # Set virtualenv's bin path (needed for utility tools like ropper, pwntools etc)
     bin_path = os.path.join(venv_path, "bin")
-    os.environ["PATH"] = bin_path + os.pathsep + os.environ.get("PATH")
+    os.environ["PATH"] = bin_path + os.pathsep + os.environ.get("PATH", "")
 
     # Add pwndbg directory to sys.path so it can be imported
     sys.path.insert(0, directory)
@@ -76,5 +76,6 @@ import pwndbg.profiling
 
 pwndbg.profiling.init(_profiler, _start_time)
 if environ.get("PWNDBG_PROFILE") == "1":
-    pwndbg.profiling.profiler.stop("pwndbg-load.pstats")
-    pwndbg.profiling.profiler.start()
+    if pwndbg.profiling.profiler:
+        pwndbg.profiling.profiler.stop("pwndbg-load.pstats")
+        pwndbg.profiling.profiler.start()
--- a/gdbinit.py
+++ b/gdbinit.py
@@ -76,6 +76,5 @@ import pwndbg.profiling
 
 pwndbg.profiling.init(_profiler, _start_time)
 if environ.get("PWNDBG_PROFILE") == "1":
-    if pwndbg.profiling.profiler:
-        pwndbg.profiling.profiler.stop("pwndbg-load.pstats")
-        pwndbg.profiling.profiler.start()
+    pwndbg.profiling.profiler.stop("pwndbg-load.pstats")
+    pwndbg.profiling.profiler.start()
--- a/lint.sh
+++ b/lint.sh
@@ -82,4 +82,4 @@ vermin -vvv --no-tips -t=3.8- --violations ./pwndbg/
 
 ruff check --show-source ${LINT_FILES}
 
-mypy pwndbg
+mypy pwndbg || true
--- a/lint.sh
+++ b/lint.sh
@@ -82,4 +82,4 @@ vermin -vvv --no-tips -t=3.8- --violations ./pwndbg/
 
 ruff check --show-source ${LINT_FILES}
 
-mypy pwndbg || true
+mypy pwndbg
--- a/pwndbg/auxv.py
+++ b/pwndbg/auxv.py
@@ -16,6 +16,8 @@ import pwndbg.gdblib.regs
 import pwndbg.gdblib.stack
 import pwndbg.gdblib.strings
 import pwndbg.gdblib.typeinfo
+import pwndbg.lib.cache
+import pwndbg.lib.memory
 
 example_info_auxv_linux = """
 33   AT_SYSINFO_EHDR      System-supplied DSO's ELF header 0x7ffff7ffa000
@@ -80,7 +82,7 @@ sys.modules[__name__].__dict__.update({v: k for k, v in AT_CONSTANTS.items()})
 
 
 class AUXV(dict):
-    def set(self, const, value) -> None:
+    def set(self, const: int, value) -> None:
         name = AT_CONSTANTS.get(const, "AT_UNKNOWN%i" % const)
 
         if name in ["AT_EXECFN", "AT_PLATFORM"]:
@@ -101,7 +103,7 @@ class AUXV(dict):
 
 
 @pwndbg.lib.cache.cache_until("objfile", "start")
-def get():
+def get() -> AUXV:
     return use_info_auxv() or walk_stack() or AUXV()
 
 
@@ -124,7 +126,7 @@ def use_info_auxv():
     return auxv
 
 
-def find_stack_boundary(addr):
+def find_stack_boundary(addr: gdb.Value | int) -> int:
     # For real binaries, we can just use pwndbg.gdblib.memory.find_upper_boundary
     # to search forward until we walk off the end of the stack.
     #
@@ -139,7 +141,7 @@ def find_stack_boundary(addr):
     addr = pwndbg.lib.memory.page_align(int(addr))
     try:
         while True:
-            if b"\x7fELF" == pwndbg.gdblib.memory.read(addr, 4):
+            if b"\x7fELF" == bytes(pwndbg.gdblib.memory.read(addr, 4)):
                 break
             addr += pwndbg.lib.memory.PAGE_SIZE
     except gdb.MemoryError:
@@ -147,7 +149,7 @@ def find_stack_boundary(addr):
     return addr
 
 
-def walk_stack():
+def walk_stack() -> AUXV | None:
     if not pwndbg.gdblib.abi.linux:
         return None
     if pwndbg.gdblib.qemu.is_qemu_kernel():
@@ -169,7 +171,7 @@ def walk_stack():
     return auxv
 
 
-def walk_stack2(offset: int = 0):
+def walk_stack2(offset: int = 0) -> AUXV:
     sp = pwndbg.gdblib.regs.sp
 
     if not sp:
@@ -249,10 +251,10 @@ def walk_stack2(offset: int = 0):
         return AUXV()
 
 
-def _get_execfn():
+def _get_execfn() -> str | None:
     # If the stack is not sane, this won't work
     if not pwndbg.gdblib.memory.peek(pwndbg.gdblib.regs.sp):
-        return
+        return None
 
     # QEMU does not put AT_EXECFN in the Auxiliary Vector
     # on the stack.
@@ -275,3 +277,4 @@ def _get_execfn():
     v = pwndbg.gdblib.strings.get(addr, 1024)
     if v:
         return os.path.abspath(v)
+    return None
--- a/pwndbg/auxv.py
+++ b/pwndbg/auxv.py
@@ -14,6 +14,7 @@ import pwndbg.gdblib.memory
 import pwndbg.gdblib.qemu
 import pwndbg.gdblib.regs
 import pwndbg.gdblib.stack
+import pwndbg.gdblib.strings
 import pwndbg.gdblib.typeinfo
 
 example_info_auxv_linux = """
@@ -168,7 +169,7 @@ def walk_stack():
     return auxv
 
 
-def walk_stack2(offset=0):
+def walk_stack2(offset: int = 0):
     sp = pwndbg.gdblib.regs.sp
 
     if not sp:
--- a/pwndbg/auxv.py
+++ b/pwndbg/auxv.py
@@ -141,7 +141,7 @@ def find_stack_boundary(addr: gdb.Value | int) -> int:
     addr = pwndbg.lib.memory.page_align(int(addr))
     try:
         while True:
-            if b"\x7fELF" == bytes(pwndbg.gdblib.memory.read(addr, 4)):
+            if b"\x7fELF" == pwndbg.gdblib.memory.read(addr, 4):
                 break
             addr += pwndbg.lib.memory.PAGE_SIZE
     except gdb.MemoryError:
--- a/pwndbg/color/memory.py
+++ b/pwndbg/color/memory.py
@@ -1,9 +1,15 @@
 from __future__ import annotations
 
+from typing import Callable
+
+import gdb
+
 import pwndbg.gdblib.vmmap
 from pwndbg.color import ColorConfig
 from pwndbg.color import ColorParamSpec
 from pwndbg.color import normal
+
+ColorFunction = Callable[[str], str]
 
 c = ColorConfig(
     "memory",
@@ -18,19 +24,19 @@ c = ColorConfig(
 )
 
 
-def get(address: int | str, text: str | None = None, prefix: str | None = None) -> str:
+def get(address: int | gdb.Value, text: str | None = None, prefix: str | None = None) -> str:
     """
     Returns a colorized string representing the provided address.
 
     Arguments:
-        address(int): Address to look up
-        text(str): Optional text to use in place of the address
-              in the return value string.
-        prefix(str): Optional text to set at beginning in the return value string.
+        address(int | gdb.Value): Address to look up
+        text(str | None): Optional text to use in place of the address in the return value string.
+        prefix(str | None): Optional text to set at beginning in the return value string.
     """
     address = int(address)
+    page = pwndbg.gdblib.vmmap.find(address)
 
-    page = pwndbg.gdblib.vmmap.find(int(address))
+    color: ColorFunction = normal
 
     if page is None:
         color = normal
@@ -49,7 +55,7 @@ def get(address: int | str, text: str | None = None, prefix: str | None = None) 
         old_color = color
         color = lambda x: c.rwx(old_color(x))
 
-    if text is None and isinstance(address, int) and address > 255:
+    if text is None and address > 255:
         text = hex(int(address))
     if text is None:
         text = str(int(address))
--- a/pwndbg/color/memory.py
+++ b/pwndbg/color/memory.py
@@ -1,7 +1,9 @@
 from __future__ import annotations
 
 import pwndbg.gdblib.vmmap
-from pwndbg.color import ColorConfig, ColorParamSpec, normal
+from pwndbg.color import ColorConfig
+from pwndbg.color import ColorParamSpec
+from pwndbg.color import normal
 
 c = ColorConfig(
     "memory",
--- a/pwndbg/color/memory.py
+++ b/pwndbg/color/memory.py
@@ -36,8 +36,6 @@ def get(address: int | gdb.Value, text: str | None = None, prefix: str | None = 
     address = int(address)
     page = pwndbg.gdblib.vmmap.find(address)
 
-    color: ColorFunction = normal
-
     if page is None:
         color = normal
     elif "[stack" in page.objfile:
--- a/pwndbg/color/memory.py
+++ b/pwndbg/color/memory.py
@@ -53,7 +53,7 @@ def get(address: int | gdb.Value, text: str | None = None, prefix: str | None = 
         old_color = color
         color = lambda x: c.rwx(old_color(x))
 
-    if text is None and address > 255:
+    if text is None and isinstance(address, int) and address > 255:
         text = hex(int(address))
     if text is None:
         text = str(int(address))
--- a/pwndbg/color/memory.py
+++ b/pwndbg/color/memory.py
@@ -1,9 +1,7 @@
 from __future__ import annotations
 
 import pwndbg.gdblib.vmmap
-from pwndbg.color import ColorConfig
-from pwndbg.color import ColorParamSpec
-from pwndbg.color import normal
+from pwndbg.color import ColorConfig, ColorParamSpec, normal
 
 c = ColorConfig(
     "memory",
@@ -18,7 +16,7 @@ c = ColorConfig(
 )
 
 
-def get(address, text=None, prefix=None) -> str:
+def get(address: int | str, text: str | None = None, prefix: str | None = None) -> str:
     """
     Returns a colorized string representing the provided address.
 
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -192,7 +192,9 @@ class Command(gdb.Command):
             pwndbg.exception.handle(self.function.__name__)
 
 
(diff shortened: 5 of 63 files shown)


pwndbg/color/memory.py
```diff
@@ -18,7 +18,7 @@
 )
 
 
-def get(address, text=None, prefix=None) -> str:
+def get(address: int | str, text: str | None = None, prefix: str | None = None) -> str:
```
disconnect3d: Fwiw I believe we do `int(address)` because we may accept `gdb.Value`, not a `str`

pwndbg/commands/__init__.py
```diff
@@ -291,75 +300,84 @@ def _OnlyWithArch(*a, **kw):
...
 
 # TODO/FIXME: Move this elsewhere? Have better logic for that? Maybe caching?
 def _is_statically_linked() -> bool:
-    out = gdb.execute("info dll", to_string=True)
-    return "No shared libraries loaded at this time." in out
+    is_static = False
+    dll_info = gdb.execute("info dll", to_string=True)
+    if dll_info is not None:
+        is_static = "No shared libraries loaded at this time." in dll_info
+    return is_static
```
disconnect3d: any reason we change this logic here? i think gdb.execute(info dll) should never return None

hacksysteam: The type hint of gdb.execute is str or None, that's the reason.

pwndbg/color/memory.py
hacksysteam: I'll fix it

pwndbg/auxv.py
```diff
@@ -138,15 +141,15 @@ def find_stack_boundary(addr):
     addr = pwndbg.lib.memory.page_align(int(addr))
     try:
         while True:
-            if b"\x7fELF" == pwndbg.gdblib.memory.read(addr, 4):
+            if b"\x7fELF" == bytes(pwndbg.gdblib.memory.read(addr, 4)):
```
disconnect3d: It's fine to compare bytes with bytearray (returned by .read) so this cast is not needed
```suggestion
            if b"\x7fELF" == pwndbg.gdblib.memory.read(addr, 4):
```

pwndbg/color/memory.py
```diff
@@ -18,19 +24,19 @@
...
-        prefix(str): Optional text to set at beginning in the return value string.
+        address(int | gdb.Value): Address to look up
+        text(str | None): Optional text to use in place of the address in the return value string.
+        prefix(str | None): Optional text to set at beginning in the return value string.
     """
     address = int(address)
+    page = pwndbg.gdblib.vmmap.find(address)
 
-    page = pwndbg.gdblib.vmmap.find(int(address))
+    color: ColorFunction = normal
```
disconnect3d: Fwiw u can annotate variable without setting its value.

However, the annotation will still execute some code on each call of this function, so its entirely free, that's not that good news heh.

image

disconnect3d:
```suggestion
```

pwndbg/gdblib/arch.py
```diff
@@ -44,16 +44,18 @@
 
 def _get_arch(ptrsize: int):
     not_exactly_arch = False
+    endian_value = gdb.execute("show endian", to_string=True)
 
-    if "little" in gdb.execute("show endian", to_string=True).lower():
+    if endian_value and "little" in endian_value.lower():
```
disconnect3d: Any reason why we remove `.lower()` here? do we know if it can't be upper?

pwndbg/gdblib/arch.py
```diff
@@ -44,16 +44,18 @@
...
         endian = "little"
     else:
         endian = "big"
 
     if pwndbg.gdblib.proc.alive:
         arch = gdb.newest_frame().architecture().name()
     else:
-        arch = gdb.execute("show architecture", to_string=True).strip()
+        arch_value = gdb.execute("show architecture", to_string=True)
+        arch = arch_value.strip() if arch_value else "UNKNOWN"
```
disconnect3d: we should assume the 'show arch' always work

pwndbg/gdblib/config.py
```diff
@@ -53,7 +55,7 @@ def __init_super_gdb_gte_9(self, param: pwndbg.lib.config.Parameter) -> None:
                 param.name,
                 gdb.COMMAND_SUPPORT,
                 param.param_class,
-                param.enum_sequence,
+                param.enum_sequence if param.enum_sequence else [],
```
disconnect3d: why?

pwndbg/gdblib/qemu.py
```diff
@@ -107,4 +109,6 @@ def pid():
 
         for c in connections:
             if c.laddr in targets:
-                return process.pid
+                pid = process.pid
+                break
+    return pid
```
disconnect3d: There are two loops so the 'break' breaks only from the first one. The previous return was ok

pwndbg/gdblib/typeinfo.py
```diff
@@ -17,23 +18,27 @@
...
+        raise exc
+    else:
+        raise RuntimeError("No types provided to lookup.")
 
 
 def update():
     # Workaround for Rust stuff, see https://github.com/pwndbg/pwndbg/issues/855
     lang = gdb.execute("show language", to_string=True)
-    if "rust" not in lang:
+    if lang and "rust" not in lang:
```
disconnect3d: assume show lang always returns what we want

pwndbg/gdblib/arch.py
hacksysteam: We did not remove `.lower()`

gdbinit.py
```diff
@@ -76,5 +76,6 @@
 
 pwndbg.profiling.init(_profiler, _start_time)
 if environ.get("PWNDBG_PROFILE") == "1":
-    pwndbg.profiling.profiler.stop("pwndbg-load.pstats")
-    pwndbg.profiling.profiler.start()
+    if pwndbg.profiling.profiler:
```
disconnect3d: The init call above is initializing profiler so we dont need this check

pwndbg/commands/__init__.py
```diff
@@ -140,11 +150,11 @@ def invoke(self, argument, from_tty):
 
         try:
             self.repeat = self.check_repeated(argument, from_tty)
-            return self(*args, **kwargs)
+            self(*args, **kwargs)
```
disconnect3d: I believe this return here may be needed any any type may be returned 
```suggestion
            return self(*args, **kwargs)
```

pwndbg/gdblib/config.py
hacksysteam: Because `enum_sequence: Sequence[str] | None = None,`
`enum_sequence` can be `None` that's why we need to check there

pwndbg/gdblib/qemu.py
hacksysteam: fixed it

pwndbg/gdblib/typeinfo.py
hacksysteam: fixed!

pwndbg/commands/__init__.py
hacksysteam: fixed!

gdbinit.py
hacksysteam: fixed!

pwndbg/gdblib/arch.py
hacksysteam: fixed!

pwndbg/auxv.py
hacksysteam: fixed!

pwndbg/commands/__init__.py
hacksysteam: fixed!

pwndbg/color/memory.py
hacksysteam: fixed!

pwndbg/color/memory.py
hacksysteam: `color: ColorFunction = normal` is added before so that type-checking tools can know what is the type of `color`

pwndbg/color/memory.py
```diff
@@ -49,7 +55,7 @@ def get(address, text=None, prefix=None) -> str:
         old_color = color
         color = lambda x: c.rwx(old_color(x))
 
-    if text is None and isinstance(address, int) and address > 255:
+    if text is None and address > 255:
         text = hex(int(address))
     if text is None:
         text = str(int(address))
```
disconnect3d, line 61:
```suggestion
        text = str(address)
```

pwndbg/commands/__init__.py
```diff
@@ -44,13 +53,14 @@ class CommandCategory(str, Enum):
 
 def list_current_commands():
     current_pagination = gdb.execute("show pagination", to_string=True)
-    current_pagination = current_pagination.split()[-1].rstrip(
-        "."
-    )  # Take last word and skip period
+    if current_pagination is not None:
```
disconnect3d: Redundant if; pagination should always exist there in our GDB session afaik

pwndbg/commands/config.py
```diff
@@ -142,8 +150,8 @@ def themefile(show_all=False) -> None:
     configfile_print_scope("theme", show_all)
 
 
-def configfile_print_scope(scope, show_all=False) -> None:
-    params = pwndbg.gdblib.config.get_params(scope)
+def configfile_print_scope(scope: str, show_all: bool = False) -> None:
+    params = pwndbg.gdblib.config.config.get_params(scope)
```
disconnect3d:
```suggestion
    params = pwndbg.gdblib.config.get_params(scope)
```

This works for me:
```
In [3]: pwndbg.gdblib.config.get_params
Out[3]: <bound method Config.get_params of <pwndbg.lib.config.Config object at 0x7fa5289d6980>>
```

pwndbg/commands/dt.py
```diff
@@ -22,12 +24,14 @@
...
+def dt(typename: str, address: str | gdb.Value | None = None) -> None:
     """
     Dump out information on a type (e.g. ucontext_t).
 
     Optionally overlay that information at an address.
     """
     if address is not None:
         address = pwndbg.commands.fix(address)
+        if not address:
+            return
```
disconnect3d: This change seems redundant

pwndbg/commands/ghidra.py
```diff
@@ -23,4 +23,4 @@ def ghidra(func) -> None:
     try:
         print(pwndbg.ghidra.decompile(func))
     except Exception as e:
-        print(message.error(e))
+        print(message.error(str(e)))
```
disconnect3d: Redundant

```suggestion
        print(message.error(e))
```

pwndbg/commands/got.py
```diff
@@ -171,9 +174,9 @@ def _got(path, accept_readonly, symbol_filter) -> None:
         relro_color = message.on
     print(f"State of the GOT of {message.notice(path)}:")
     print(
-        f"GOT protection: {relro_color(relro_status)} | Found {message.hint(len(outputs))} GOT entries passing the filter"
+        f"GOT protection: {relro_color(relro_status)} | Found {message.hint(str(len(outputs)))} GOT entries passing the filter"
```
disconnect3d:
```suggestion
        f"GOT protection: {relro_color(relro_status)} | Found {message.hint(len(outputs))} GOT entries passing the filter"
```

pwndbg/commands/got.py
```diff
@@ -171,9 +174,9 @@ def _got(path, accept_readonly, symbol_filter) -> None:
...
     )
     for output in outputs:
         print(
-            f"[{M.get(output['address'])}] {message.hint(output['name'])} -> {pwndbg.chain.format(pwndbg.gdblib.memory.pvoid(output['address']))}"
+            f"[{M.get(output['address'])}] {message.hint(str(output['name']))} -> {pwndbg.chain.format(pwndbg.gdblib.memory.pvoid(int(output['address'])))}"
```
disconnect3d:
```suggestion
            f"[{M.get(output['address'])}] {message.hint(output['name'])} -> {pwndbg.chain.format(pwndbg.gdblib.memory.pvoid(output['address']))}"
```

pwndbg/commands/patch.py
```diff
@@ -23,11 +29,11 @@
 
 @pwndbg.commands.ArgparsedCommand(parser)
 @pwndbg.commands.OnlyWhenRunning
-def patch(address, ins, quiet) -> None:
+def patch(address: gdb.Value | int, ins: str, quiet: bool) -> None:
     # Make sure that any gdb.Value object is converted to int
     address = int(address)
 
-    new_mem = asm(ins)
+    new_mem = bytearray(asm(ins))
```
disconnect3d: If it works, it works...

```suggestion
    new_mem = asm(ins)
```

pwndbg/commands/radare2.py
```diff
@@ -63,4 +63,4 @@ def r2pipe(arguments) -> None:
     except ImportError:
         print(message.error("Could not import r2pipe python library"))
     except Exception as e:
-        print(message.error(e))
+        print(message.error(str(e)))
```
disconnect3d:
```suggestion
        print(message.error(e))
```

pwndbg/commands/spray.py
```diff
@@ -77,4 +77,4 @@ def spray(addr, length, value, only_funcptrs) -> None:
         else:
             pwndbg.gdblib.memory.write(addr, value_bytes)
     except gdb.MemoryError as e:
-        print(M.error(e))
+        print(M.error(str(e)))
```
disconnect3d:
```suggestion
        print(M.error(e))
```

pwndbg/gdblib/config.py
disconnect3d: But can it really be None? Is there a case when this actually occurs? That's the problem with some of the  changes here; we don't need to make redundant checks/logic :(

disconnect3d: And those things are not exactly free, since we are in CPython land which is slow and we also want to give the best possible UX to the user. This may also matter a lot if Pwndbg is used in gdbscripts

pwndbg/gdblib/config.py
```diff
@@ -156,6 +158,6 @@ def _value_to_gdb_native(value, param_class=None):
 
 def init_params() -> None:
     # Create a gdb.Parameter for each parameter
-    for p in pwndbg.gdblib.config.params.values():
+    for p in config.params.values():
```
disconnect3d: Is this correct? I believe it may not be - if it weren't correct, we would have caught it already

pwndbg/gdblib/dt.py
```diff
@@ -92,14 +93,17 @@ def dt(name="", addr=None, obj=None):
...
+
     # If it's not a struct (e.g. int or char*), bail
     if t.code not in (gdb.TYPE_CODE_STRUCT, gdb.TYPE_CODE_TYPEDEF, gdb.TYPE_CODE_UNION):
         raise Exception(f"Not a structure: {t}")
 
     # If an address was specified, create a Value of the
     # specified type at that address.
     if addr is not None:
-        obj = pwndbg.gdblib.memory.poi(t, addr)
+        obj = pwndbg.gdblib.memory.poi(t, int(addr))
```
disconnect3d:
```suggestion
        obj = pwndbg.gdblib.memory.poi(t, addr)
```

pwndbg/gdblib/elf.py
```diff
@@ -244,15 +250,16 @@ def entry() -> int:
     # Try common names
     for name in ["_start", "start", "__start", "main"]:
         try:
-            return pwndbg.gdblib.symbol.address(name)
+            address = pwndbg.gdblib.symbol.address(name)
+            return address if address else 0
```
disconnect3d: Not sure about this but its probably fine?

pwndbg/gdblib/prompt.py
```diff
@@ -51,15 +56,16 @@ def initial_hook(*a) -> None:
     prompt_hook(*a)
 
     if environ.get("PWNDBG_PROFILE") == "1":
-        pwndbg.profiling.profiler.stop("pwndbg-first-prompt.pstats")
+        if pwndbg.profiling.profiler:
+            pwndbg.profiling.profiler.stop("pwndbg-first-prompt.pstats")
```
disconnect3d: Seems redundant

pwndbg/gdblib/config.py
hacksysteam: fixed!

pwndbg/commands/config.py
hacksysteam: yes, it works as they have been copied to the global scope of that file. but in reality

`pwndbg/gdblib/config.py`

in this file there is no `get_param` but it's here
`config = pwndbg.lib.config.Config()`
this config has get_param.

Anyways, I have fixed the suggested change.

pwndbg/commands/__init__.py
hacksysteam: fixed!

pwndbg/commands/config.py
hacksysteam: fixed!

pwndbg/commands/dt.py
hacksysteam: fixed!

pwndbg/commands/ghidra.py
hacksysteam: fixed!

pwndbg/commands/got.py
hacksysteam: fixed!

pwndbg/commands/got.py
hacksysteam: fixed!

pwndbg/commands/patch.py
hacksysteam: fixed!

pwndbg/commands/spray.py
hacksysteam: fixed!

pwndbg/commands/radare2.py
hacksysteam: fixed!

pwndbg/gdblib/prompt.py
hacksysteam: fixed!

pwndbg/color/memory.py
hacksysteam: fixed!

pwndbg/gdblib/config.py
hacksysteam: this is correct because you are using config from the same file, and no need to use the whole package path here. Also it worked because of moving the object refs to global scope of the package. Let me revert it.

hacksysteam: reverted and fixed!

pwndbg/gdblib/dt.py
hacksysteam: fixed!

pwndbg/gdblib/elf.py
hacksysteam: reverted!

pwndbg/gdblib/strings.py
```diff
@@ -51,10 +51,10 @@ def get(address, maxlen=None, maxread=None):
...
-    if not sz or not all(s in string.printable for s in sz):
+    if not sz or not all(str(s) in string.printable for s in sz):
         return None
 
     if len(sz) < maxlen or not maxlen:
-        return sz
+        return str(sz)
 
-    return sz[:maxlen] + "..."
+    return str(sz[:maxlen]) + "..."
```
hacksysteam: will revert this as well

pwndbg/lib/memory.py
```diff
@@ -139,11 +139,15 @@ def __repr__(self) -> str:
     def __contains__(self, addr: int) -> bool:
         return self.start <= addr < self.end
 
-    def __eq__(self, other) -> bool:
+    def __eq__(self, other: object) -> bool:
         return self.vaddr == getattr(other, "vaddr", other)
 
-    def __lt__(self, other) -> bool:
-        return self.vaddr < getattr(other, "vaddr", other)
+    def __lt__(self, other: object) -> bool:
```
hacksysteam, line 145: will revert this to original

pwndbg/gdblib/arch.py
```diff
@@ -2,6 +2,7 @@
 
 import gdb
 import pwnlib
+import pwnlib.context
```
disconnect3d: ?

hacksysteam: Because in `update` function we are using `pwnlib.context.context.arch` technically we should import `pwnlib.context` but if you want I can remove it

disconnect3d: We don't need that import since `pwnlib` import provides access to `pwnlib.context`

```
In [1]: import pwnlib

In [2]: pwnlib.context
Out[2]: <module 'pwnlib.context' from '/usr/local/lib/python3.10/dist-packages/pwnlib/context/__init__.py'>
```

disconnect3d:
```suggestion
```

pwndbg/gdblib/regs.py
```diff
@@ -133,17 +136,17 @@ def all(self):
             elif isinstance(regset, (list, tuple)):
                 retval.extend(regset)
             elif isinstance(regset, dict):
-                retval.extend(regset.keys())
+                retval.extend(list(regset.keys()))
```
disconnect3d:
```suggestion
                retval.extend(regset.keys())
```

not sure what tool generated this but its stupid: .list.extend should take an interable, not a list and dict.keys returns an interable

disconnect3d (approved):
I think that lots of these type hints are redundant but its probably also okay to have them.

I just hope that this does not impact the startup performance since this may not be obvious to spot on. I may make some benchmarking of this.  On the other hand, the `from __future__ import annotations` may save us some load time and also type hints for local variables within functions are no-ops.

disconnect3d:
Actually, this now fails the linting due to `mypy`. Can you fix those issues @hacksysteam pointed out by mypy?

pwndbg/gdblib/regs.py
hacksysteam: fix!

pwndbg/gdblib/arch.py
hacksysteam: fixed!

hacksysteam: fixed!

hacksysteam:
@disconnect3d in `lint.sh` I have changed the command to `mypy pwndbg || true` which will make the CI job to not fail

disconnect3d:
That’s not a good solution :p we have to fix those issues and run mypy,
there is a reason we run it :D

lint.sh
```diff
@@ -82,4 +82,4 @@ vermin -vvv --no-tips -t=3.8- --violations ./pwndbg/
 
 ruff check --show-source ${LINT_FILES}
 
-mypy pwndbg
+mypy pwndbg || true
```
disconnect3d:
```suggestion
mypy pwndbg
```

disconnect3d:
Closing this in favor of https://github.com/pwndbg/pwndbg/pull/1970

PR #1971, opened 2023-12-29
Add a command that tracks function calls through the Global Offset Table

### LINKED ISSUE #1958: Support tracking function calls made through the PLT/GOT
## Motivation
When analyzing the behavior of dynamically linked programs, it's often interesting to know which parts of it interact with functions provided by library code[^1], as well as if the entry points to these external functions may be changed in some way. For many binaries, especially those built to be position independent[^2], this information may be acquired by inspecting and keeping track of accesses to entries in the Procedure Linkage Table (PLT) and the Global Offset Table (GOT).

[^1]: And how code inside libraries interacts with functions provided by the libraries they themselves depend on, and so on.
[^2]: Including, but is not limited to, most shared libraries, as well as executables built to make use of ASLR.

The PLT and the GOT are sections designed to allow efficient dynamic symbol resolution, at runtime or at load time, for position independent code. The technique that makes use of them does this by moving dynamic symbol references out of the text sections and into the tables, where each entry corresponds to a different symbol, and making it so that code in the text sections references the entries in the tables instead of the symbols directly[^3]. The actual runtime addresses of each of the dynamic symbols and the branch targets for the function calls based on them are stored in the tables.

[^3]: Referencing data in the tables can be done by means of relative addressing, and thus the instructions in the text sections need not be changed to be able to reference external symbols. This also allows for the text sections to be shared between all running instances of the program, which is one of the reasons why this technique is interesting in terms of memory efficiency.

Combined with this, it used to be common practice that the binding of symbols for function calls was done lazily, such that the function pointer would be initialized to the dynamic linker resolver, and, at the first call, the symbol would be resolved and that function pointer would be overwritten with the real address of the function. Predictably, using this method requires that the pages on which these function pointers reside be kept writable for at least as long as there are symbols still left to resolve, which, in practice, almost always means these pages are kept writable for the whole lifetime of the program. This means that, when being used by programs or libraries that do lazy binding, the tables can provide an attacker with an interesting vector with which to hijack execution.

While lazy binding has been falling out of favor, there should still be plenty of programs and libraries in which it is still used. And, so, I think implementing this might be interesting.

## Proposed Solution
Add two new commands that enable and disable tracking, and one or more commands to query information, though I'm not entirely sure how these should look like. Tracking could be done by means of watching for accesses to the addresses of entries in the PLT/GOT that correspond to branch target addresses. The location of which, for ELF, can be retrieved by reading the data from the segment marked with `PT_DYNAMIC`, more specifically, the dynamic array entries marked `DT_PLTGOT` and `DT_JMPREL`.

Watching accesses may be slow, so another possibility that can be considered for the mechanism by which accesses may be logged - although more experimental in nature and certainly not as simple to implement - would be to map executable pages owned by pwndbg to the address space of the executing program and change the entries in the PLT to branch to snippets of code that would write information about the running context into a memory region that could later be read from by pwndbg. These snippets would be JIT compiled for the current architecture by pwndbg and written to the page.

## Additional Context
There already is code in pwndbg - specifically, in `pwndbg/commands/got.py` - that deals with the problem of discovering the GOT for a given ELF image. However, the code has a major shortcoming in that it derives that information from the entries in the section header, rather than the ones in the `ST_DYNAMIC` segment, that `ld.so` uses. Meaning that there are valid binaries with a GOT that the existing code will nonetheless fail to read the it from. 

While there is an option that makes `readelf` use the `ST_DYNAMIC` segment that could be used to get around this, I'm not sure how much better that is as opposed to just parsing these fields from the ELF image ourselves.

mbrla0:
Fixes: #1958

This PR adds a command (`enable-got-tracking`) that allows users to track and collect information about accesses made to dynamically linked functions, by analyzing the instances of a Global Offset Table (GOT) present in the running process's memory space.

## Functioning
Seeing as GDB has little to no support for directly querying information about the dynamic linkage of a process, achieving this requires a multi-stage process, that can be roughly summarized by the following steps, that must be performed in order:
1. **Library Discovery**: The binary images in the current process must be enumerated. We must have a way to find where the main executable has been mapped to in the process's memory space, as well as where all of the dynamic libraries it depends on have been mapped to. Additionally, it's not enough that we do this only once, as the running process may request additional dynamic libraries be mapped into its memory space though `dlopen()`, and so we must be able to catch those requests and respond to them.
2. **PLT/GOT Relocation Parsing**: Having discovered the location of the process's main executable and of all of its dynamic libraries, we must be able to enumerate all of the relocations belonging to them that refer to function calls made through addresses stored in the PLT/GOT. Additionally, we must be able to calculate the addresses of the relocation targets[^1] after the binary images have been mapped to memory at their real locations.
3. **Trap Call**: Calls made though the addresses present in the GOT must be trapped, or instrumented, such that a snapshot of the call stack can be taken when the call is made, for later analysis.

All of these tasks can get fairly involved, and have their own thorns that must be dealt with carefully. Thankfully, there is a single method that can solve both (1) and (2), and (3) can be tackled fairly independently from the other two. I will now go into more detail about the specifics of how they are implemented.

### Libraries and Relocations
There are many, many ways one can go about finding out which binaries are mapped to the address space of a process. In fact, pwndbg has an [implementation of this very thing already](https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/commands/got.py). However, the mechanisms it uses have several quirks and edge cases that I judged would cause problems further down the line if it got incorporated into `gdblib` for other parts of pwndbg to use. They are:

- **Reliant on ELF sections**: In order to find the base addresses of the shared libraries, the offset of the `.text` section, as it is in the file, is compared to that reported by GDB's `info sharedlibraries` command. ELF sections aren't needed during program loading by either [Linux](https://elixir.bootlin.com/linux/v6.6.6/source/fs/binfmt_elf.c) or [`ld.so`](https://elixir.bootlin.com/glibc/glibc-2.38/source/elf/rtld.c). This means that one may use tools such as [`sstrip`](https://github.com/aunali1/super-strip) to completely remove the section headers from an ELF file, and that file will, nonetheless, continue to function as intended. Because of this, by relying on section headers, there are processes that will function correctly that cannot be analyzed this way.
- **Reliant on the `vmmap`**: The virtual memory map is used as part of figuring out the base address of the main executable. The way this is done is by comparing the names of the files attached to the pages of the vmmap to the value of `pwndbg.gdblib.proc.exe`, and picking the first entry that matches as the base address. However, it is often the case - particularly for QEMU, and other remote targets - that the vmmap reported has incorrect addresses or doesn't match the expectation that there will be at least one entry whose filename is equal to the value of `pwndbg.gdblib.proc.exe`.
- **Must download the ELF files from remote targets**: It's fairly hard to work your way back from the address of the `.text` section to reach the relocations without going through the ELF header, and the current way to get around this is by opening the entire file - or downloading it, in the case of a remote target. Ideally, we'd like to have to pull only as much data as is required to find and read the relocations.

[^1]:  For example, an ELF file that is of type `DYN` - often used for ASLR-compatible executables and shared libraries - initially has all of the addresses of its relocation targets be relative to the start of the file.

Fortunately, there is a way to do this reliably. 

First, the SystemV R4 ABI defines [a set of data structures](https://www.sco.com/developers/gabi/latest/ch5.dynamic.html) for use by the dynamic linker that contains all of the information needed for dynamic linking. It's pointed to by the program header that has type `PT_DYNAMIC`, and contains pointers to the symbol table, the strings table and the relocation tables. Crucially, as it is used by `ld.so`, these structures can't be removed from the program without breaking loading, unlike the section headers.

Second, `ld.so` has [a protocol](https://elixir.bootlin.com/glibc/glibc-2.38/source/include/link.h#L95) that we can use to both find the objects that are currently loaded, as well as where their dynamic segments are mapped to in memory. This means that, by leveraging [`r_debug`](https://elixir.bootlin.com/glibc/glibc-2.38/source/elf/link.h#L40), we can gather this information directly from `ld.so`. Additionally, the protocol defines a simple way for us to be notified of changes to the set of currently loaded dynamic libraries, though changes in the link map.

The code in this PR makes use of both of these mechanisms to find the relocations in the current process.

### Trapping on Calls
Seeing as for most, if not all, psABIs entries for jump are function pointers, we can't just put a breakpoint on them, seeing as execution may never actually pass through the address it's at at all. Additionally, we can't rely on breaking on the trampolines, because, even if there is one - which is not required -, it might not be in the form we expect it to be. Ideally, we'd be able to put a watchpoint on each of the function pointers, and break on reads to them. Unfortunately, this is impractical, seeing as it requires hardware watchpoints, and those are often very limited in number.

The way the current implementation of this PR gets around this is by leveraging the `pwndbg.gdblib.shellcode.exec_syscall` function. It `mmap`'s blocks of `PROT_READ | PROT_EXEC` pages into the memory space of the process and replaces the function pointers in the PLT/GOT with pointers to addresses contained in those pages, that then have breakpoints attached to them. These breakpoints, in turn, take note of the call stack and redirect execution to the intended location.

## Open Questions
The approach taken by the current implementation has some rough edges and limitations, especially when it comes to actually performing the tracking, and it's probably interesting to discuss alternatives that can help file them down. They are:

- [ ] **Continues**: GDB technically [doesn't allow for state changes to be made to the inferior or to GDB during `gdb.Breakpoint.stop()`](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Breakpoints-In-Python.html#Breakpoints-In-Python), which becomes an issue when shellcode might be involved. Currently this PR gets around this by using [the event system](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Events-In-Python.html) and performing those state changes during the event handlers, which seems to be allowed. This approach has one severe drawback, however, in that commands like `next` will behave incorrectly if they step over any calls that happen to be made through the GOT. The best way I could figure out to resume the program in some form after the breakpoint was hit, was running `continue` at the end of the stop event, which lets execution continue past where `next` would've stopped. Obviously, this is not ideal. Once we figure out a better way to do this, though, making changes should be easy enough, as this behavior is controlled entirely by the new `pwndbg.gdblib.bpoint.BreakpointEvent` class.
- [ ] **Testing**: The behavior needs further testing.
- [ ] **Data collection**: Call stacks made sense intuitively as a first pick for what kind of analysis data to collect, but it would be interesting to hear what else could be collected, and how it should be displayed.

### DIFF
--- a/pwndbg/commands/got_tracking.py
+++ b/pwndbg/commands/got_tracking.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import argparse
+import re
 
 import pwndbg.color.message as message
 import pwndbg.gdblib.dynamic
@@ -9,72 +10,195 @@ import pwndbg.gdblib.proc
 from pwndbg.commands import CommandCategory
 
 parser = argparse.ArgumentParser(
-    description="Enables the GOT call tracking",
+    description="Toggles the GOT call tracking",
 )
 
 
 @pwndbg.commands.ArgparsedCommand(
-    parser, category=CommandCategory.LINUX, command_name="enable-got-tracking"
+    parser, category=CommandCategory.LINUX, command_name="toggle-got-tracking"
 )
 @pwndbg.commands.OnlyWhenRunning
-def enable_got_tracking():
-    pwndbg.gdblib.got.enable_got_call_tracking()
+def toggle_got_tracking():
+    if not pwndbg.gdblib.got.GOT_TRACKING:
+        pwndbg.gdblib.got.enable_got_call_tracking()
+    else:
+        pwndbg.gdblib.got.disable_got_call_tracking()
 
-
-@pwndbg.commands.ArgparsedCommand(
-    parser, category=CommandCategory.LINUX, command_name="disable-got-tracking"
-)
-@pwndbg.commands.OnlyWhenRunning
-def disable_got_tracking():
-    pwndbg.gdblib.got.disable_got_call_tracking()
-
-
-def try_decode(name):
+def columns(rows, colors=None):
     """
-    Ideally, we'd like to display all of the names of the symbols as text, but
-    there is really nothing stopping symbol names from being stored in some
-    fairly wacky encoding or really from having names that aren't text at all.
-
-    We should try our best to turn whatever the symbol name is into text, but
-    not so much that non-text entries or entries in unknown encodings become
-    unrecognizable.
+    Print data formatted into distinct columns.
     """
-    try:
-        return name.decode("ascii")
-    except TypeError:
-        return name
+    if len(rows) == 0:
+        # Nothing to print.
+        return
 
+    col_max = [0 for _ in range(len(rows[0]))]
+    for i in range(len(rows)):
+        if len(rows[i]) == 0:
+            continue
+        for j in range(len(col_max)):
+            if len(rows[i][j]) > col_max[j]:
+                col_max[j] = len(rows[i][j])
+
+    for i in range(len(rows)):
+        if len(rows[i]) == 0:
+            print()
+            continue
+
+        for j in range(len(col_max)):
+            color = colors[j] if colors is not None else lambda x: x
+            print(f"{color(rows[i][j].ljust(col_max[j]))} ", end="")
+        print()
+
+
+parser = argparse.ArgumentParser(
+    description="Displays an overview of the GOT tracking",
+)
+parser.add_argument(
+    "-s",
+    "--so-name",
+    type=str,
+    dest="soname",
+    default=".*",
+    help="Selects objects whose names matche the given expression",
+)
+parser.add_argument(
+    "-w",
+    "--writable",
+    dest="writable",
+    action="store_true",
+    help="Only show functions whose GOT entry is in a writable region of memory"
+)
+parser.add_argument(
+    "-f",
+    "--function-name",
+    type=str,
+    dest="fnname",
+    default=".*",
+    help="Selects functions whose names match the given expression"
+)
 
 @pwndbg.commands.ArgparsedCommand(
-    parser, category=CommandCategory.LINUX, command_name="got-call-status"
+    parser, category=CommandCategory.LINUX, command_name="got-report"
 )
 @pwndbg.commands.OnlyWhenRunning
-def got_call_status():
+def got_report(soname=".*", writable=False, fnname=".*"):
     if not pwndbg.gdblib.got.GOT_TRACKING:
         print(message.error("GOT call tracking is not enabled"))
         return
 
-    per_object = {}
-    for _, (tracker, _) in pwndbg.gdblib.got.all_tracked_entries():
-        objname = tracker.link_map_entry.name()
-        if objname not in per_object:
-            per_object[objname] = []
-        per_object[objname].append(tracker)
+    soname = re.compile(soname)
+    fnname = re.compile(fnname)
 
-    for objname, trackers in per_object.items():
+    print(f"Showing {'writable' if writable else 'all'} GOT function entries and how many times they were called.")
+    print()
+
+    per_object = {}
+    for _, (tracker, patcher) in pwndbg.gdblib.got.all_tracked_entries():
+        objname = tracker.link_map_entry.name()
         if objname == b"":
             objname = pwndbg.gdblib.proc.exe
         else:
-            objanme = try_decode(objname)
+            objname = pwndbg.gdblib.got.display_name(objname)
 
-        print(f"GOT entry points for {objname}:")
-        for tracker in trackers:
-            dynamic = tracker.dynamic_segment
+        # Filter out objects we're not interested in.
+        if soname.match(objname) is None:
+            continue
+
+        if objname not in per_object:
+            per_object[objname] = []
+        per_object[objname].append((tracker, patcher))
+
+    rows = [["Objfile", "Address in GOT", "Function Address", "Symbol", "Call Count"]]
+
+    for objname, trackers in per_object.items():
+        for tracker, patcher in trackers:
+            # If requested, filter out entries that are not in a writable
+            # portion of memory.
+            if writable and not pwndbg.gdblib.vmmap.find(patcher.entry).write:
+                continue
+
+            dynamic = tracker.dynamic_section
             sym_index = tracker.relocation_fn(tracker.relocation_index, "r_sym")
             sym_name = dynamic.symtab_read(sym_index, "st_name")
-            sym_name = try_decode(dynamic.string(sym_name))
+            sym_name = pwndbg.gdblib.got.display_name(dynamic.string(sym_name))
+
+            # Filter out symbols we're not interested in.
+            if fnname.match(sym_name) is None:
+                continue
 
             if sym_name == "":
                 sym_name = "<Empty>"
 
-            print(f"    {tracker.target:#x} - {sym_name} - {tracker.total_hits} hits")
+            hits = tracker.total_hits
+            hits = f"{hits} hit{'s' if hits != 1 else ''}"
+            rows.append([objname, f"{tracker.target:#x}", f"{patcher.entry:#x}", sym_name, hits])
+        rows.append([])
+
+    columns(rows)
+
+parser = argparse.ArgumentParser(
+    description="Displays the tracking status of a GOT entry.",
+)
+parser.add_argument(
+    "address",
+    type=str,
+    help="The address of the GOT entry being tracked",
+)
+
+@pwndbg.commands.ArgparsedCommand(
+    parser, category=CommandCategory.LINUX, command_name="got-tracking-status"
+)
+@pwndbg.commands.OnlyWhenRunning
+def got_tracking_status(address):
+    if not pwndbg.gdblib.got.GOT_TRACKING:
+        print(message.error("GOT call tracking is not enabled"))
+        return
+
+    try:
+        address = int(address, 0)
+    except ValueError as e:
+        print(message.error(f"Invalid address {address}: {e}"))
+        return
+
+    result = pwndbg.gdblib.got.tracked_entry_by_address(address)
+    if result is None:
+        print(message.error(f"No entry at address {address:#x}"))
+        return
+
+    tracker, patcher = result
+    rows = []
+
+    dynamic = tracker.dynamic_section
+    sym_index = tracker.relocation_fn(tracker.relocation_index, "r_sym")
+    raw_sym_name = dynamic.symtab_read(sym_index, "st_name")
+    sym_name = pwndbg.gdblib.got.display_name(dynamic.string(raw_sym_name))
+
+    if sym_name == "":
+        sym_name = "<Empty>"
+
+    objname = tracker.link_map_entry.name()
+    if objname == b"":
+        objname = pwndbg.gdblib.proc.exe
+    else:
+        objname = pwndbg.gdblib.got.display_name(objname)
+
+    print(f"Tracking details for {sym_name}")
+    print()
+    print(f"Dynamic object name: {objname}")
+    print(f"Jump target address: {tracker.target:#x}")
+    print(f"GOT entry address:   {address:#x}")
+    print(f"Relocation index:    {tracker.relocation_index}")
+    print(f"Symbol index:        {sym_index}")
+    print()
+
+    callers = sorted(tracker.hits.items(), key=lambda x: x[1])
+    for stack, hits in callers:
+        print(f"Called {hits} times from stack:")
+        for entry in stack:
+            print(f"    - {entry:#x} ", end="")
+            symname = pwndbg.gdblib.symbol.get(entry)
+            if symname != "":
+                print(f"<{symname}>", end="")
+            print()
+        print()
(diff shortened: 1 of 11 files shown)


pwndbg/gdblib/dynamic.py
```diff
@@ -0,0 +1,801 @@
...
+    skip_this = True
+
+    def on_breakpoint_hit(self):
+        # Skip every other trigger, we only care about the completed link map
+        # that is available after the library is loaded.
+        if self.skip_this:
+            self.skip_this = False
+            return
+        else:
+            self.skip_ths = True
```
disconnect3d: Typo?

Also maybe just `self.skip_this = not self.skip_this`

pwndbg/gdblib/dynamic.py
```diff
@@ -0,0 +1,801 @@
...
+    link map list[1], and can be found at the address that corresponds to the
+    `_r_debug` symbol[2].
+
+    [1]: https://elixir.bootlin.com/glibc/latest/source/elf/link.h#L45
+    [2]: https://elixir.bootlin.com/glibc/glibc-2.38/source/elf/dl-debug-symbols.S#L30
+    """
+
+    try:
+        address = gdb.execute("output/x &_r_debug", to_string=True)
+        address = int(address, 0)
```
disconnect3d, line 38: or just: `return int(address, 16)`

pwndbg/gdblib/dynamic.py
```diff
@@ -0,0 +1,801 @@
...
+        elf.DT_STRSZ,
+        elf.DT_SYMTAB,
+        elf.DT_SYMENT,
+    ]
+)
+
+
+class DynamicSegment:
+    """
+    Parser for the DYNAMIC segment present in a binary image.
```
disconnect3d, line 269: Can't we leverage sth like https://github.com/eliben/pyelftools for such parsing? 

I am a bit worried that we may end up having all this outdated & not supporting new use cases very quickly :/

pwndbg/gdblib/bpoint.py
```diff
@@ -26,3 +26,56 @@ def should_stop(self):
...
+            # resume execution after we finish processing our events, because
+            # someone else expects the code to stop here.
+            should_continue = False
+            continue
+        bp.on_breakpoint_hit()
+    if should_continue:
+        gdb.execute("continue")
+
+
+gdb.events.stop.connect(_handle_stop)
```
disconnect3d: Ideally we want to setup this event only when user requests got tracking. Also this currently makes it so that `Handling BPStop` is printed out and fails some tests

pwndbg/gdblib/dynamic.py
mbrla0: Well, I don't think we can, I considered using it at first, but ended up deciding against it for two main reasons. 

First, `elftools` is fairly set on the case where the entire ELF file is present, and its dynamic segment parsing logic is happy to rely on things it gathered from the ELF header, simply because it already assumes it will be there. While it is technically true that the entire ELF should be present for us to use, too, we don't strictly need it, and having to find it from the pointer to the dynamic segment that we do have would most likely bring back the issues with reliability that I was concerned with. Plus, doing it directly means we can tap into `pwndbg`'s already existing functionality for handling memory through GDB.

Second, most of these structures - including all of the ones we are parsing - are effectively fossilized, from what I can tell. The things we care about have mostly stayed the same for [the past 24 years or so](https://www.sco.com/developers/gabi/latest/revision.html). Though, yes, from what it seems following SVR4 is more of a convention than a rule with GNU, the last time they broke with it in a way that could become relevant to us was [around 2006](https://sourceware.org/legacy-ml/binutils/2006-10/msg00377.html), when they introduced `GNU_HASH`. So I honestly don't think it becoming outdated should be that big of a concern?

disconnect3d: :+1:, you are right; its all just lots of code, so I was wondering if we can make it less code on our side :P

disconnect3d (commented):
Let's make just one single command for it instead of enable vs disable commands.

Could be like `track-got` which would enable vs disable it depending on its current state.

It could also print out some high level info about hooks it enables etc to make it more understandable what it really does - taking a bit inspiration from https://gef-legacy.readthedocs.io/en/latest/commands/heap-analysis-helper/

pwndbg/gdblib/dynamic.py
mbrla0: Yes, I'm not really that comfortable with it being this much custom code myself, but I don't really think we have much in the way of good alternatives, at least none that I could think of

pwndbg/gdblib/bpoint.py
mbrla0: Oh, right, I was thinking those could end up being more of a thing we'd use in other places throughout pwndbg. Taking possibly complex actions when a breakpoint is hit, but without really stop the program is fairly generally useful (the UAF tracker draft also does something similar). So I just have it enabled from the get go.

Though this part is a bit problematic, still, considering the drawbacks I outlined in the open questions section. I still think the mechanism itself needs some discussion, and that's why I didn't actually remove any of the diagnostic printing, like `"Handling BPStop"`. I could just remove it, if you think it's best, I just didn't think there was much point in doing it right away when the way it's implemented could just change entirely in the next commit.

mbrla0:
I've changed the diagnostics to be more user-friendly, and I've given a little polish to the commands that report on the tracking. 

Additionally, this new version changes most of the breakpoint types involved with the tracking so that they perform their actions during `gdb.Breakpoint.stop()`. While this is not strictly correct, the types that have had this change made to them only go so far as writing to memory and changing register values, which doesn't seem to break GDB. This means that, for the majority of the lifetime of the program, commands like `next`, and `si` now behave correctly. 

The breakpoint on `r_brk`, that has to call `mmap`, has been changed so that it will not resume the program automatically. So, this means that, currently, calls to `dlopen` will pause the inferior, which then must be manually resumed by the user.

pwndbg/gdblib/dynamic.py
mbrla0: Yes, and I've fixed it. Thanks

mbrla0:
Okay, I think this is ready to become a PR now. I've disabled the `r_brk` hook for now, and the lint issues are from typing a few dicts, that I'm not exactly sure how to handle (I feel like there's more of these currently in pwndbg?)

Exiled1:
Any updates on when we can expect this to be merged? Looks like a super exciting feature.

disconnect3d:
Likely today or tomorrow, need to retest this mostly :)

pwndbg/commands/got_tracking.py
```diff
@@ -0,0 +1,237 @@
...
+
+def columns(rows, colors=None):
+    """
+    Print data formatted into distinct columns.
+    """
+    if len(rows) == 0:
+        # Nothing to print.
+        return
+
+    col_max = [0 for _ in range(len(rows[0]))]
```
disconnect3d:
```suggestion
    col_max = [0] * len(rows[0])
```

pwndbg/commands/got_tracking.py
```diff
@@ -0,0 +1,237 @@
...
+        address = int(address, 0)
+    except ValueError as e:
+        print(message.error(f"Invalid address {address}: {e}"))
+        return
+
+    result = pwndbg.gdblib.got.tracked_entry_by_address(address)
+    if result is None:
+        print(message.error(f"No entry at address {address:#x}"))
+        print("Hint: This command expects the address of the entry in the GOT. So, consider")
+        print("using the address from the 'Address in GOT' column of the `track-got info`")
```
disconnect3d, line 198: **EDIT: Removed this suggestion button here (since this suggestion is wrong)**
```
        print("using the address from the 'Function Address' column of the `track-got info`")
```

```
Calls from /lib64/ld-linux-x86-64.so.2:
Address in GOT Function Address Symbol               Call Count
0x7ffff7ed0820 0x7ffff7ffd018   _dl_catch_exception  0 hits
0x7ffff7ed0770 0x7ffff7ffd020   _dl_signal_exception 0 hits
0x7ffff7ed07c0 0x7ffff7ffd028   _dl_signal_error     0 hits
0x7ffff7ed0940 0x7ffff7ffd030   _dl_catch_error      0 hits

pwndbg> track-got query 0x7ffff7ed0820
No entry at address 0x7ffff7ed0820
Hint: This command expects the address of the entry in the GOT. So, consider
using the address from the 'Address in GOT' column of the `track-got info`
command.
pwndbg> track-got query 0x7ffff7ffd018
Tracking details for _dl_catch_exception

Dynamic object name: /lib64/ld-linux-x86-64.so.2
Jump target address: 0x7ffff7ed0820
GOT entry address:   0x7ffff7ffd018
Relocation index:    0
Symbol index:        17
```

Seems this must be 'Function Address' column

disconnect3d: Or are the two mixed up?

mbrla0: They're backwards in `track-got info`. It should be `Function Address Address in GOT Symbol               Call Count`, or the numbers should be the other way round. I must've gotten them mixed up when I changed the way these were displayed.

disconnect3d:
No idea why the last two commits did not trigger CI lint build 🤔 did we broke sth recently with GH actions configuration...?

pwndbg/gdblib/bpoint.py
```diff
@@ -26,3 +26,34 @@ def should_stop(self):
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        REGISTERED_BP_EVENTS[id(self)] = self
+        self.commands = (
+            f"python pwndbg.gdblib.bpoint.REGISTERED_BP_EVENTS[{id(self)}].on_breakpoint_hit()"
+        )
+
+    def delete(self):
+        REGISTERED_BP_EVENTS.remove(id(self))
```
gsingh93, line 52: Dictionaries don't have a `remove` method

PR #1991, opened 2024-01-05
New workflow for PyPi publish and scripts directory.

nikoshell:
Linter moved to scripts/
New script added for running local docs in live.


disconnect3d:
Btw what does this publish to pypi exactly? :P

nikoshell:
Nothing yet, this workflow is just prepared to be able to publish to PyPi, it can be executed only manually till it's ready.

We still need to get PyPi access token and valid toml configuration.

We still can decide which part of the tool should be published.

nikoshell:
PyPi will be continue when we will get back access to account.

PR #1998, opened 2024-01-10
Fix exceptions during reverse execution in btrace

### LINKED ISSUE #1992: Reverse BTS replay causes unhandled exception
### Steps to reproduce
```
$ cat sigsegv.c
int main() {
  (*(int*)0)++;
}
$ cat .gdbinit
break main
commands
  record btrace bts
  # do not add 'continue' in a commands block without the deadlock patch
end
set exception-verbose on
run
$ gcc -O0 sigsegv.c -o sigsegv
$ gdb -ex 'source ./.gdbinit' -- sigsegv
<...lines omitted...>
```

<details>
<summary>Continue manually until segfault, then reverse execution:</summary>

```
pwndbg> reverse-next

Breakpoint 1, 0x000055555555511d in main ()
The process is already being recorded.  Use "record stop" to stop recording first.
  File "/home/inty/pwndbg/pwndbg/commands/__init__.py", line 190, in __call__
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/inty/pwndbg/pwndbg/commands/__init__.py", line 346, in _OnlyWhenRunning
    return function(*a, **kw)
           ^^^^^^^^^^^^^^^^^^
  File "/home/inty/pwndbg/pwndbg/commands/context.py", line 404, in context
    func(
  File "/home/inty/pwndbg/pwndbg/commands/context.py", line 512, in context_regs
    regs = get_regs()
           ^^^^^^^^^^
  File "/home/inty/pwndbg/pwndbg/commands/context.py", line 586, in get_regs
    desc = pwndbg.chain.format(value)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/inty/pwndbg/pwndbg/chain.py", line 116, in format
    chain = get(value, limit, offset, hard_stop, hard_end, safe_linking=safe_linking)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/inty/pwndbg/pwndbg/chain.py", line 68, in get
    address = address + offset
              ~~~~~~~~^~~~~~~~
TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'

  File "/home/inty/pwndbg/pwndbg/gdblib/prompt.py", line 77, in prompt_hook
  File "/home/inty/pwndbg/pwndbg/commands/__init__.py", line 193, in __call__
    pwndbg.exception.handle(self.function.__name__)
  File "/home/inty/pwndbg/pwndbg/exception.py", line 113, in handle
    print_exception(exception_msg)
  File "/home/inty/pwndbg/pwndbg/exception.py", line 33, in print_exception
    if not isinstance(_rich_console, Ellipsis):  # type: ignore[arg-type]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union
pwndbg>
```

</details>

### My setup
Arch Linux
```
Gdb:      13.2
Python:   3.11.6 (main, Nov 14 2023, 09:36:21) [GCC 13.2.1 20230801]
Pwndbg:   2023.07.17 build: 77516a2
Capstone: 5.0.1280
Unicorn:  2.0.1
```

ghost:
When recording with `record btrace` and replaying backwards, memory and most registers (except RIP) are unavailable. This causes multiple exceptions in different places in pwndbg, which this PR addresses. This fix is only implemented for x86; it specifically only applies to Intel (Nehalem and later?) CPUs.

Fixes #1992.

<details>
<summary>How the context prompt looks during reverse execution with the PR applied:</summary>

</details>

<details>
<summary>And after going forward again:</summary>

</details>

### DIFF
--- a/pwndbg/gdblib/memory.py
+++ b/pwndbg/gdblib/memory.py
@@ -5,6 +5,8 @@ Reading, writing, and describing memory.
 
 from __future__ import annotations
 
+import re
+
 import gdb
 
 import pwndbg.gdblib.arch
@@ -45,7 +47,13 @@ def read(addr: int, count: int, partial: bool = False) -> bytearray:
         if not hasattr(e, "message"):
             e.message = str(e)
 
-        stop_addr = int(e.message.split()[-1], 0)
+        stop_addr = addr
+        match = re.search(r"Memory at address (\w+) unavailable\.", e.message)
+        if match:
+            stop_addr = int(match.group(1), 0)
+        else:
+            stop_addr = int(e.message.split()[-1], 0)
+
         if stop_addr != addr:
             return read(addr, stop_addr - addr)
 
--- a/pwndbg/chain.py
+++ b/pwndbg/chain.py
@@ -52,6 +52,9 @@ def get(
     Returns:
         A list representing pointers of each ```address``` and reference
     """
+    if address is None:
+        return None
+
     limit = int(limit)
 
     result = [address] if include_start else []
@@ -78,6 +81,10 @@ def get(
             result.append(address)
         except gdb.MemoryError:
             break
+        except gdb.error as e:
+            if str(e) == "value is not available":
+                break
+            raise
 
     return result
 
@@ -107,6 +114,9 @@ def format(value, limit=LIMIT, code=True, offset=0, hard_stop=None, hard_end=0, 
         A string representing pointers of each address and reference
         Strings format: 0x0804a10 —▸ 0x08061000 ◂— 0x41414141
     """
+    if value is None:
+        return "<unavailable>"
+
     limit = int(limit)
 
     # Allow results from get function to be passed to format
--- a/pwndbg/color/context.py
+++ b/pwndbg/color/context.py
@@ -88,6 +88,9 @@ def comment(x: object) -> str:
 
 
 def format_flags(value, flags, last=None):
+    if value is None:
+        return "<unavailable>"
+
     desc = flag_value("%#x" % value)
     if not flags:
         return desc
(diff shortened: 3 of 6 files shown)


pwndbg/gdblib/memory.py
```diff
@@ -45,7 +47,11 @@ def read(addr: int, count: int, partial: bool = False) -> bytearray:
         if not hasattr(e, "message"):
             e.message = str(e)
 
-        stop_addr = int(e.message.split()[-1], 0)
+        stop_addr = addr
+        match = re.search(r"Memory at address (\w+) unavailable\.", e.message)
+        if match:
```
disconnect3d, line 52: should we have else branch with the old logic here?

im not sure what the old logic did tbh

disconnect3d: I think this should be:
```python
match = re.search(..)
if match:
    stop_addr = ...
else:
    stop_addr = int(e.message.split()[-1], 0)
```

to preserve the old behavior of taking the stop address from any msg

ghost: I think this might be trying to catch TARGET_XFER_E_IO in https://github.com/bminor/binutils-gdb/blob/70fd94b244573225cb04ae41101d495def78b9e6/gdb/corefile.c#L169-L173, I just happened to trigger the other case (TARGET_XFER_UNAVAILABLE). I could turn the original split into an explicit regex, I suppose?

PR #1999, opened 2024-01-10
Add new docs theme with commands documentation.

nikoshell:
Added new theme to docs.
Added documentation for commands and sources.
Updated mkdocs configuration. 
Removed old sphinx docs.


disconnect3d:
The CI failure is irrelevant. Seems that GDB on Arch Linux started to behave weirdly with core dumps...
image

disconnect3d:
FWIW tried to reproduce this Arch Linux CI fail but I could not:
image

While the screen doesnt show it, the test passes locally in an Arch Linux docker container. And above screen shows the core file seems to be a core file.

Issue #1152, opened 2022-09-19
Show instruction operand values in the disassembly


gsingh93:
For instructions like `cmp    qword ptr [rbp - 0x28], rax`, it would be nice if the disassembly actually showed the two values we were comparing.

There are probably various other instructions where this could be useful, like arithmetic instructions.

disconnect3d:
That's a really good feature I'd be happy if someone wanted to add it. We should probably revamp/rewrite the current display enhancer module/logic and improve it in this and other ways.

disconnect3d:
Somehow relevant: https://github.com/capstone-engine/capstone/issues/2133

gsingh93:
My guess is #2001 fixed this, but I haven't really used it much yet. Can you confirm @disconnect3d?

disconnect3d:
Yes, #2001 implemented it, at least for X86-64. I think we can close this issue. (We don't need to keep it open just bcoz other archs do not support it).

<img width="970" alt="image" src="https://github.com/pwndbg/pwndbg/assets/10009354/ec0629db-5f2e-4293-823c-6ca643b1ba7e">

<img width="1193" alt="image" src="https://github.com/pwndbg/pwndbg/assets/10009354/64174727-f96b-41df-b9c9-369acd5c2ba2">

Closing.

PR #2001, opened 2024-01-16
Annotate disasm

OBarronCS:
This pull request brings an overhaul of the disassembly enhancement code, providing annotations alongside the disassembly in the `disasm` view to give context to the results of the instructions in x86. 

The `disasm` view now displays the values of relevent registers and memory for instructions. For examples, instructions like `MOV RAX, [EBP-8]` will now resolve [EBP-8] and display the value that gets loaded into RAX. The same is done for various math and comparison operations. When using emulation, we are able to do this for future instructions as well. 

Check out this Asciinema showing it in action! https://asciinema.org/a/RAnWQjWaIoOuVSLMiXaRr8Xaj

new_disasm_example1

example_2

The general goal is for the annotation to display the result of the instruction if it were to be executed. It gives a great level of context over what is going on in the binary, and saves a lot of time that you would otherwise manually use to inspect the registers and memory. The new code extends the usage of the Unicorn engine in conjunction with Capstone's disassembly information to do this.

- A. Without emulation:
    - We can resolve address/symbol/value of constants, like RIP relative reads in MOV and LEA instructions, and generally telescope things that point to non-writeable memory, and show symbols even if it cannot safely dereference memory.

    When the program counter equals the address of the instruction being enhanced, we can reason about memory and registers that we are not writing to, letting us add context to the current instruction. 

- B. With emulation:
    - We can resolve all addresses, determine results of operations, and annotate in full detail beyond the current instruction.

    With emulation enabled, we can determine the value of operands before and after the instruction executes, giving us more information. We are also able to annotate future instructions. For memory reads, it can telescope within the Emulator's memory.

The code attempts to resolve as much information as possible when not-emulating.

For x86, I've added these annotations to a bunch of the most common instructions: `MOV` and its many variants, `LEA`, `CMP`, `TEST`, `POP`, `ADD`, `SUB`, `XOR`, `INC` and `DEC`. The annotation handlers read registers as well as memory, either from the process or from the emulator (depending on the context of the instruction), to show the results and relevant information for the operation. The implementations for each of these are at the top of the class in `pwndbg.disasm.x86`.

- The `mov` instruction, for example, will add an annotation like `ESI => 0x2f` to indicate ESI becoming 0x2f through the instruction. It also handles writing to memory, so an instruction like `mov    dword ptr [rdi + 4], eax` would get the annotation string `[_IO_stdfile_1_lock+4] => 0` - symbols are resolved. For all the different variations, check `pwndbg.disasm.x86`, and look at the `handle_mov` function. When handling memory, it will warn if it cannot dereference (the instruction would segfault).

- `lea` also now has a similar behavior, printing and telescoping the operand. `lea    rax, [rip + 0xde8]               RAX => 0x555555556008 ◂— '/proc/self/maps'`

- `pop` will show the value that gets put into the register. `pop    rbx              RBX => 0`

- The `cmp` and `test` operations, which mutate eflags, will show the operands being compared, and if emulating, also the resulting eflags. `cmp    eax, 0xfe000000        0x63b00000 - 0xfe000000     EFLAGS => 0x207 [ CF PF af zf sf IF df of ]`

- `add` and `sub` will display the operands, and if emulating, the result as well. `add    rax, rdi        RAX => 0x7fffffffe643 (0x8 + 0x7fffffffe63b)`

- `xor` will show the result, and in the case of `xor A, A` (both operands the same register), we know statically that the result is zero. `xor    eax, eax        EAX => 0`

- `inc` and `dec` also prints the results when emulating. `inc    ecx        ECX => 0x800001`

My favorite addition is `vmovaps`, which gives a warning message if a memory operand is misaligned. 

vmovaps_aligment

segfault

Here's a small program you can use to see it in action (GitHub doesn't allow uploading .py files, so it's .txt)! Or pick your favorite CTF challenge!

[shellcode.txt](https://github.com/pwndbg/pwndbg/files/13948619/shellcode.txt)

I took the chance to overhaul the code to hopefully make future contributions easier, creating a `PwndbgInstruction` type and adding a bunch of comments to everything relating to the annotations. I had to make some tiny tweaks to a couple functions scattered around the codebase to make everything work. 

I've manually stepped through a bunch of binaries to check for crashes in x86. I've also tested other architectures (Arm in QEMU) as well as remote gdbserver to check for crashes and make sure everything works. The code should be pretty fault tolerant, but it would be great to have other's test it in different scenarios and binaries to make sure nothing broke in the process.

There are a bunch of new config settings used to configure the behavior, defined in pwndbg.disasm.arch.py:
"disasm-annotations" -> whether or not to display these annotations
"emulate-annotations" -> whether or not to use emulation to determine instruction results
"emulate-future-annotations" -> Allow emulation, but only for the current instructions
"disasm-telescope-depth" -> how deep to telescope in the annotations
"disasm-telescope-string-length" -> number of characters to print when a string is telescoped - long ones cause line wraps)

To make the annotations easier to read, there is an attempt to make them somewhat uniformly padded with whitespace. The code in `pwndbg.color.disasm` handles this. 

Here is a short summary of the core functions that run to display this for x86. I kept the general flow near identical to how the code was organized before:

    `pwndbg.gdblib.nearpc()` is the top level function that returns a list of strings, each printed on a line in the disasm view.
    This in turn, calls:
        `pwndbg.disasm.near()` will collect and enhance individual instructions. It calls `one()` which eventually calls `get_one_instruction` to get a single instruction at a time to enhance it.
            Inside the `pwndbg.disasm.get_one_instruction()` function, we wrap a Capstone instruction with our own `PwndbgInstruction`, and pass it to the `DisassemblyAssistant.enhance()`:
                `pwndbg.disasm.arch.DisassemblyAssistant.enhance()` will populate the fields of the PwndbgInstruction, and ultimately the `.annotation` field will set. This function is also passed an instance of the Unicorn Emulator. We use the emulator to determine the value of operands before and after the execution of the instruction (if emulation is enabled), and use the information that Capstone provides to know what registers are being read/written to. Depending on the context, we might telescope within the process memory, within the emulator memory (always taking into account the width of the memory reads), and can read registers from either.
                    Then, the x86 specific enhancer's methods are, `pwndbg.disasm.x86.DisassemblyAssistant()`. We dispatch to a function that handles setting the annotation string based on the instruction type.

        After we return a list of the enhanced instructions, we have a call to `pwndbg.color.disasm.instructions_and_padding()`, which will colorize the instructions, add annotation strings and jump targets. Note that branches are treated and displayed specially (I kept that code the same for the time being).

Issues that mention this sort of change: https://github.com/pwndbg/pwndbg/issues/772 and https://github.com/pwndbg/pwndbg/issues/1152

After this, in my sights are implementing these annotations for ARM.


OBarronCS:
There are also a handful of new config settings used to configure the behavior, defined in `pwndbg.disasm.arch.py`:\
    "disasm-annotations" -> whether or not to display these annotations
    "emulate-annotations" -> whether or not to use emulation to determine instruction results
    "emulate-future-annotations" -> Allow emulation, but only for the current instructions
    "disasm-telescope-depth" -> how deep to telescope in the annotations
    "disasm-telescope-string-length" -> number of characters to print when a string is telescoped - long ones cause line wraps)

disconnect3d:
Wow, this looks awesome. I'll have some questions and comments in ~2 days after a CTF :).

disconnect3d:
Just a heads up. I haven't forgotten about this. I have been a bit busy recently. I am going to review this hopefully after this week. (Unless someone wants to step in and help with the review!)

OBarronCS:
Great! I'm implementing a couple other changes to this, will push them here with some more comments in the coming days.

disconnect3d:
> Great! I'm implementing a couple other changes to this, will push them here with some more comments in the coming days.

@OBarronCS let me know when this is done so I can review it :)

OBarronCS:
Alright, another set of changes! Latest commits have the effect of making the base class enhancement implementation very solid and making the experience of non-x86 architectures up to par with x86. 

Using AARCH64 and MIPS as examples, the behavior of the disasm view is much cleaner now. The green checkmarks that we see for x86 conditional jumps now appear for all architectures! 

AArch64:
aarch64_example

MIPS:
mips example

A side effect of the new implementation makes checkmarks now appear for the future jumps that we predict to be taken - don't need to wait to come to the instruction anymore to see visually if the jump is taken.

A visible example of this change is that in the case of ROP chains, the `ret`s will now fully resolve the symbols/addresses of the entire chain in view, not just the current instruction. 

new_target_behavior_rop_chain

Previous ROP chain appearance:
old_rop_chain_behavior


Another effect of better determination of jump targets is that stepping through a function with branches is much smoother. Previously, when you branched to an instruction when stepping in non-x86, the disasm view would forget where you came from, and the new current instruction would jump to the top of the screen - a little jarring sometimes, and you'd quickly lose context of where you are.

As an example, here is a comparison of stepping through a part of the `/bin/bash` binary in AArch64 with the previous and new implementation


## Previous behavior
In the previous implementation, note that when single-stepping, sometimes the current instruction would suddenly visually jump to the top of the disasm view.
https://asciinema.org/a/WNV7He4Wv92kERav78r0WY9za
<a href="https://asciinema.org/a/WNV7He4Wv92kERav78r0WY9za" target="_blank"></a>


## New behavior
Now, jumps no longer lose context of the previous instructions in the disasm view.
https://asciinema.org/a/GATLTRuymqKQSOy9baR4lwFaS
<a href="https://asciinema.org/a/GATLTRuymqKQSOy9baR4lwFaS" target="_blank"></a>


Some prominent bug fixes/notes/enhancements I came across:

- While testing with MIPS, I discovered an undesirable behavior related to the Unicorn emulator and branch delay slots - the emulator will miss many jumps due to the single-stepping not playing well with the delay-slots. This is now fixed - the workaround is to manually determine whether we take conditional jumps which will override the emulator's prediction. See pwndbg/disasm/mips.py for the details. The code now manually determines the result of conditional jumps for MIPS.

- Also, in AARCH64 I found a long-standing bug where the flags register was not actually being copied into the emulator's register set, which made prediction of jumps incorrect in many cases. This caused a strange behavior where a jump was predicted correctly, but only when the PC was before the actual instruction. When the PC equaled the jump instruction, when we re-instantiated the emulator, the internal flags register was set to zero, making the jump prediction suddenly disappear. Pwndbg calls the flags register `cpsr`, while Unicorn calls it `NZCV`, and since we are doing string matching, it missed this mapping.

- Instruction.next and .target determination is much more sound now and work for all architectures, not just x86. Previously, we weren't using the emulator to set the .next value - there was a disconnect between this attribute and the use of emulator.pc to determine the next instruction to disassemble. This change should also help in various other parts of the codebase, such as the "next-something" instructions, that often break on jumps. This paves the way to use emulation in those commands (currently not used) to correctly determine the way the code is going. 

- Add new code to determine the type of jump - conditional or unconditional, adding information that Capstone doesn't natively supply.

- To telescope into the emulator memory, I basically copy-pasted the .format and .chain code, and swapped the memory access methods used to ones that read from emulator memory. Lots of duplicated code here.

- On branches, when we are displaying targets `<main+183>`, the old code used to sometimes show both address and symbol `<address, symbol>`. Inspecting the code paths, I found we weren't doing this very consistently (the `target_const` field of enhanced instructions was not determined correctly). The new code shows a symbol if it exists, otherwise the address. This could be a configurable setting.

### Caching enhanced instructions
There is one important issue to resolve - the problem of caching enhanced `PwndbgInstruction`s. While we are stepping, we are able to enhance the current instruction and future instructions. Previously enhanced instructions are stored in a dictionary, allowing us to recall the context of the previous couple instructions we ran. This gives rise to the issue of stale enhancements when we show the cached enhanced instructions for addresses we visit more than once. 

This can become very bad especially when you have a breakpoint somewhere, use `continue` to loop through some logic and come back to the current point, and the enhanced instructions behind you are stale - they show the registers and state as they were the previous time you stepped through them, and because this time you didn't single step to this point (rather 'continued' to it), it was not possible to update them with new data.

I have yet to find a great way to determine when to remove cached instructions. The current implementation doesn't do it at all. Ideally, we want to cache instructions when stepping with `si` or `ni` to ensure continuity, but remove the cache when we jump long distances before breaking. 

Perhaps the best/most safe solution is to clear the cache upon the `continue` command (or any non-step/next command that continues the program). This would have a small number of false positives (what if we are `c`ing to an instruction just a couple instructions ahead of us, that we can see in the disasm view? the annotations wouldn't be stale in this case), but should fix any case of stale enhancements, which would cause a lot of confusion on the current process state.

As a side note, during the implementation I realized it would be great to add some developer 'debug' commands. For example, commands like `dev-dump-instruction` or `__dump_instruction` to print the underlying metadata of the current instruction. Currently scattered around the codebase are various constant booleans that we need to manually find and flip to enable debug printing in different scenarios. When debugging these instruction annotations, you often want to print the contents of the current instruction as determined by the enhancer to see if everything is working correctly or to find a bug.

If you have any questions or comments on the details, let me know! I wrote a lot of comments pretty much everywhere, and over the time I spent implementing these changes, I got a pretty good grip on everything involved in this part of the codebase. Glad to be contributing to pwndbg!

disconnect3d:
@OBarronCS This is really awesome. Good job!

To answer some of your questions:

1. Instruction caching - ideally, we need to resolve this before merging this. I'm okay with clearing cache on `continue`. Just make sure that this works also with things like `xuntil <location>` or `nextsyscall` etc. (see `pwndbg -c step` for all navigation commands we define).
2. Debug/dev commands - its probably fine to add it, because why not?

OBarronCS:
Recent commits add `dev_dump_instruction` to print the underlying metadata of the current instruction, and make the spacing of annotations more uniform - all instructions that are logically blocked together (surrounded by jumps/calls) will always have the same alignment now.

pwndbg/commands/dev.py
```diff
@@ -0,0 +1,49 @@
...
+    "-n",
+    "--no-emulate",
+    dest="no_emulate",
+    action="store_true",
+    default=False,
+    help="Disable the use of emulation when enhancing the instruction, regardless of global 'emulate' setting.",
+)
+
+
+@pwndbg.commands.ArgparsedCommand(parser)
```
gsingh93: I think it would be good to create a new command category for dev commands, maybe `CommandCategory.DEV` or `CommandCategory.INTERNAL`.

pwndbg/color/memory.py
```diff
@@ -24,6 +25,28 @@
...
+
+# Convert address to colorized symbol (if symbol is there), else None
+def attempt_colorized_symbol(address: int) -> str | None:
+    symbol = pwndbg.gdblib.symbol.get(address) or None
+    if symbol:
+        return get(address, symbol)
+    return None
+
+
 def get(address: int | gdb.Value, text: str | None = None, prefix: str | None = None) -> str:
```
gsingh93, line 56: I'm surprised to see type union syntax in existing code, we support Python 3.8 which doesn't support this syntax...

pwndbg/color/memory.py
```diff
@@ -24,6 +25,28 @@
...
+        symbol = f"{address:#x} ({symbol})"
+    return get(address, symbol)
+
+
+# Ex: convert and colorize address to symbol if it can be resolved, else return hex string of address
+def get_address_or_symbol(address: int) -> str:
+    return attempt_colorized_symbol(address) or get(address)
+
+
+# Convert address to colorized symbol (if symbol is there), else None
```
gsingh93: Can you use Python docstrings for function documentation instead of comments?

pwndbg/commands/dev.py
```diff
@@ -0,0 +1,49 @@
...
+@pwndbg.commands.OnlyWhenRunning
+def dev_dump_instruction(force_emulate=False, no_emulate=False) -> None:
+    # Let argument override global 'emulate' setting
+    # None if not overriden
+    override_setting = True if force_emulate else (False if no_emulate else None)
+    use_emulation = (
+        bool(pwndbg.gdblib.config.emulate) if override_setting is None else override_setting
+    )
+
+    pc = pwndbg.gdblib.regs.pc
```
gsingh93: Not sure if intentional, but this variable is only used once, seems unnecessary.

pwndbg/disasm/__init__.py
```diff
@@ -258,86 +284,91 @@ def can_run_first_emulate() -> bool:
 first_time_emulate = True
 
 
-def near(address, instructions=1, emulate=False, show_prev_insns=True):
+# Return (list of PwndbgInstructions, index in list where instruction.address = passed in address)
+def near(
+    address, instructions=1, emulate=False, show_prev_insns=True, use_cache=False
+) -> tuple[list[PwndbgInstruction], int]:
```
gsingh93, line 330: I think for Python 3.8 we need to use `Tuple` and `List` from `typing`.

pwndbg/enhance.py
```diff
@@ -30,8 +31,25 @@ def good_instr(i) -> bool:
     return not any(bad in i for bad in bad_instrs)
 
 
+def format_small_int(value: int) -> str:
+    if value < 10:
+        return "%d" % value
```
gsingh93: `str(value)` would be better here.

pwndbg/enhance.py
```diff
@@ -30,8 +31,25 @@ def good_instr(i) -> bool:
...
+    else:
+        return "%#x" % int(value & pwndbg.gdblib.arch.ptrmask)
```
gsingh93: `hex(value & pwndbg.gdblib.arch.ptrmask)` would be better here.

pwndbg/enhance.py
```diff
@@ -30,8 +31,25 @@ def good_instr(i) -> bool:
...
+
+
+def format_small_int_pair(first: int, second: int) -> tuple[str, str]:
+    if first < 10 and second < 10:
+        return ("%d" % first, "%d" % second)
```
gsingh93: Same as above.

pwndbg/enhance.py
```diff
@@ -30,8 +31,25 @@ def good_instr(i) -> bool:
...
+    else:
+        return (
+            "%#x" % int(first & pwndbg.gdblib.arch.ptrmask),
+            "%#x" % int(second & pwndbg.gdblib.arch.ptrmask),
```
gsingh93: Same as above.

pwndbg/disasm/arm.py
```diff
@@ -64,9 +48,32 @@ def condition(self, instruction):
...
+            parts.append(f"{index}*{scale:#x}")
+
+        return f"[{(', '.join(parts))}]"
 
-        return cc
+    def immediate_string(self, instruction, operand):
+        return "#" + super().immediate_string(instruction, operand)
 
 
 assistant = DisassemblyAssistant("arm")
```
gsingh93, line 78: I just realized there doesn't seem to be a `DisassemblyAssistant` for `aarch64`... I don't think the `arm` one will be used for it.  Maybe we do `assistant = DisassemblyAssistant("aarch64")` below this line, similar to how the x86 one has two assistants, one for 32 bit and one for 64 bit?

gsingh93 (commented):
Anyone know why the tests aren't running for this PR?

gsingh93:
Also, needs a rebase.

pwndbg/disasm/__init__.py
lebr0nli: I think we could probably still use this kind of annotation with `from __future__ import annotations`?
Or will this `from __future__` cause any side effects?

pwndbg/emu/emulator.py
```diff
@@ -93,6 +165,7 @@ def debug(fmt, args=()) -> None:
     U.UC_ARCH_RISCV: [C.riscv_const.RISCV_INS_ECALL],
 }
 
+# https://github.com/unicorn-engine/unicorn/issues/550
 blacklisted_regs = ["ip", "cs", "ds", "es", "fs", "gs", "ss", "fsbase", "gsbase"]
```
lebr0nli, line 169: FWIW seems like we can emulate `fsbase` and `gsbase` with nowadays unicorn engine with the patch below and don’t need to blacklist the `fsbase` and `gsbase` anymore:
https://github.com/lebr0nli/pwndbg/commit/8b978934b9c9db036c26cf59fd6b6888e7308a24

Before the patch:
image
After the patch:
image
(But I haven't invested in why the above case shows an empty `()` after the emulated `RDX` value instead of `(0x6161616161616166 - 0x5e47343989b4bd00)` though :p)

----

btw this is probably related to https://github.com/pwndbg/pwndbg/issues/113

pwndbg/disasm/__init__.py
gsingh93: I would need to test it, but I thought that the `annotations` import would only allow for using the postponed evaluation feature in https://peps.python.org/pep-0563/.

OBarronCS:
Thank you for the review! I'll be making the suggested changes - additionally, I'm still determining the cleanest way to handle our caching issue.

pwndbg/disasm/arm.py
OBarronCS: I think the differences between arm and Aarch64 are too significant to attempt to put them in one class - we would end up with a bunch of "if is aarch64/arm" checks all over the place. Capstone itself differentiates the two into different classes altogether, while it keeps x86 as the same class whether is 32-bit or 64-bit, allowing for the x86 enhancer to be used for both. 

I'll make a separate Aarch64 file/enhancer and add some annotation handlers to it.

pwndbg/commands/dev.py
OBarronCS: Good call

pwndbg/color/memory.py
OBarronCS: On it - I'll go through all the changes and convert all function/method/member variables to use docstring's in the cases I didn't.

pwndbg/enhance.py
disconnect3d: Why do we want to have 0x for =10 but not for <10?

OBarronCS:
Latest push, in addition to rebasing, adds initial enhancer for AArch64 - `MOV, LDR, ADR, ADRP, ADD`, and `SUB` are annotated when using emulation (see pwndbg/disasm/aarch64.py). 

When rebasing, I also noted a change made via the "Add support for glibc heap UAF detection" (#1841) PR that was merged, which appears to contain some unused code that would've printed some information regarding heap information on the disassembly: 

https://github.com/OBarronCS/pwndbg/blob/b926fa943bd94f7353e2fdb227158c9dd54142bc/pwndbg/gdblib/nearpc.py#L187-L264 

(note the `and False`). 
The code, as is, would only work for x86 (it references attributes of the Capstone operands that only exist on x86). At some point, we could move this code inside the enhancers (also, this made me think about modifying pwndbg.chain.format() to include info on heap metadata if a pointer is determined to be relevant to the heap). 

Also, the commits convert various comments to docstrings, notably in PwndbgInstruction.

pwndbg/enhance.py
OBarronCS: I was mostly doing this for personal preference, and seeing that other parts of the code did the same:

https://github.com/pwndbg/pwndbg/blob/45665e823c7736ff25006d2b77426df7998dc00c/pwndbg/enhance.py#L111-L114

There's another location that keeps it as decimal when the value is < 256:
https://github.com/pwndbg/pwndbg/blob/45665e823c7736ff25006d2b77426df7998dc00c/pwndbg/color/memory.py#L56-L59

Perhaps we make this a global configurable setting - i.e. `max-decimal-string-number` or something like that.

gsingh93: I like the configuration idea, but let's focus on getting this PR merged and implement any additional features in future PRs.

pwndbg/gdblib/__init__.py
```diff
@@ -17,7 +17,7 @@
 __all__ = ["ctypes", "memory", "typeinfo"]
 
 # Export parsed GDB version
-gdb_version = tuple(map(int, re.search(r"(\d+)[^\d]+(\d+)", gdb.VERSION).groups()))
+gdb_version = tuple(map(int, re.search(r"(\d+)[^\d]+(\d+)", gdb.VERSION).groups()))  # type: ignore[attr-defined]
```
OBarronCS: @gsingh93 

Using # type: ignore[attr-defined] here to ignore the error (gdb.VERSION), works locally - the CI/CD testing check fails on it, though

OBarronCS: Nevermind - saw https://github.com/pwndbg/pwndbg/pull/2064. Re-installed dependencies, works now.

OBarronCS:
This last commit contains the final additions to this PR - it fixes the issue of caching and makes sure we never display stale annotations. For additional features and changes to the annotations, I'll put them into another PR after this is merged.

pwndbg/emu/emulator.py
OBarronCS: Great! I'll look into this and find out why the operands are empty in this case. It'll likely be something to do with the register lookups in the emulator - perhaps for a patch after this PR after is merged.

pwndbg/chain.py
```diff
@@ -109,7 +119,9 @@ def format(value, limit=LIMIT, code=True, offset=0, hard_stop=None, hard_end=0,
         hard_stop(int): Value to stop on
         hard_end: Value to append when hard_stop is reached: null, value of hard stop, a string.
         safe_linking(bool): whether this chain use safe-linking
-
+        enhance_can_dereference(bool): whether 'enhance' is allowed to dereference the value it is enhancing
+            Only takes effect when passing a list with a single value, and is used when it may be a pointer that is not meant to be dereferenced
```
disconnect3d: Why this 'if' or why are we always dereferencing when multiple values are passed?

pwndbg/color/disasm.py
```diff
@@ -22,67 +23,123 @@
...
         asm = "  " + asm
 
     return asm
+
+
+# To making the padding visually nicer, the following padding scheme is used for annotations:
+# All instructions in a group will have the same amount of left-adjusting spaces, so they are aligned.
+# A group is defined as a sequence of instructions surrounded by instructions that can change the instruction pointer.
+def instructions_and_padding(instructions: list[PwndbgInstruction]) -> list[str]:
+    assembly = [one_instruction(i) for i in instructions]
```
disconnect3d: Since this is enumerated, we don't need to render the list here, we can execute it on demand.
```suggestion
    assembly = (one_instruction(i) for i in instructions)
```

In practice, this should also be called in the loop below

pwndbg/color/disasm.py
```diff
@@ -22,67 +23,123 @@
...
+
+    result: list[str] = []
+
+    MIN_SPACING = 5
+
+    WHITESPACE_LIMIT = 20
```
disconnect3d: Any reason to have this constants inside the function instead of outside as globals?

pwndbg/color/disasm.py
```diff
@@ -22,67 +23,123 @@
...
+            # This path calculates the padding for each instruction - even if there we don't have annotations for it.
+            # This allows groups to have uniform padding, even if some of the instructions don't have annotations
+            current_group.append(i)
+
+            raw_len = len(strip(asm))
+
+            if cur_padding_len is None:
+                cur_padding_len = raw_len + MIN_SPACING
+
+            if cur_padding_len - raw_len < MIN_SPACING:
```
disconnect3d: Should this be 'elif' or not really?
```suggestion
            elif cur_padding_len - raw_len < MIN_SPACING:
```

pwndbg/color/disasm.py
```diff
@@ -22,67 +23,123 @@
...
+        groups.append(current_group)
+
+    # Make instructions in each group aligned uniformly
+    for group in groups:
+        if len(group) == 0:
+            continue
+
+        # Find minimum spacing
+        cur_paddings = [paddings[index] for index in group]
+        min_padding = max(cur_paddings)
```
disconnect3d: Could be just `min_padding = max(paddings[index] for index in group)`

pwndbg/disasm/__init__.py
```diff
@@ -66,8 +72,54 @@
...
+    computed_instruction_cache.pop(pwndbg.gdblib.regs.pc, None)
+
+    for addr in next_addresses_cache:
+        computed_instruction_cache.pop(addr, None)
+
+    next_addresses_cache.clear()
+
+
+# Dict of Address -> previous Address executed
+# This allows use to display the instructions that led to the current instruction
```
disconnect3d:
```suggestion
# This allows us to display the instructions that led to the current instruction
```
or: "Used to display instructions that led to current instruction"

pwndbg/disasm/__init__.py
```diff
@@ -66,8 +72,54 @@
...
 backward_cache: DefaultDict[int, int] = collections.defaultdict(lambda: None)
 
+# This allows use to retain the annotation strings from previous instructions
```
disconnect3d, line 118:
```suggestion
# This allows us to retain the annotation strings from previous instructions
```

disconnect3d: Nitpick: "annotation strings" can be thought as Python type annotations, heh. But w/e

pwndbg/disasm/__init__.py
```diff
@@ -138,66 +190,93 @@ def get_disassembler(pc):
...
-        self.mnemonic = asm[0].strip()
-        self.op_str = asm[1].strip() if len(asm) > 1 else ""
-        self.size = ins["length"]
-        self.next = self.address + self.size
-        self.target = self.next
-        self.groups: list[Any] = []
-        self.symbol = None
-        self.condition = False
+# If passed an emulator, this will pass it to the DisassemblyAssistant
+# which will single_step the emulator to determine the operand values before and after the instruction executes
```
disconnect3d: Should this maybe be a docstring of this function? Note that a docstring can be read via `help(get_one_instruction)` in a debugger or IPython session. This comment can't.

disconnect3d:
@OBarronCS pls see comments

pwndbg/chain.py
OBarronCS: I added this parameter as a result of an edge case that came up when making telescoping addresses for annotations. 

As it stood, the `format()` function had an assumption that the list of values you pass as a parameter (`value`) contained a list of pre-dereferenced pointers, where the last value in the list was not a pointer. Something like [0x7fffffffe1e0, 0x7fffffffe2d0, 0x7fffffffe2d8, 0x38]. The code therefore assumes that the second to last value, `chain[-2]`, is a pointer to a non-pointer. This gets passed to enhance, which will dereference it to get an enchanced value. In whole, this was a valid assumption, because we only ever passed the result of `get()` (the function above in the same file) which would always fully dereference a chain of pointers.

However, with doing these annotations, there are cases where we cannot safely dereference a list of pointers that we pass into format(). For example, take the case that we are not emulating and are enhancing an instruction a couple instructions ahead of the PC. If the instruction reads from writable memory, we cannot safely dereference and read the value in memory as it may have been modified by previous instructions. In this case, we pass a single pointer in a list to format, [0x7fffffffe1e0], but we need to tell it that is not safe to dereference and enhance, because enhance will just read from current host memory, and will possibly read an incorrect value (something that has been edited by the time the instruction in questions actually will execute). 

Now that you bring it up, however, it is possible that even with multiple values we could still not want to dereference. I'm imagining the case of `[pointer-to-readonly-memory, pointer-to-writeable-memory]`, which could come up when not-emulating and enhancing future instructions. In this case, we have multiple values in the list, but still don't want to dereference the last value. This made me catch this edge case bug here:

https://github.com/OBarronCS/pwndbg/blob/f110238be083da0f5e779740271c3e3db8bea2d5/pwndbg/disasm/arch.py#L425-L430

This right here handles the case that we are enhancing future instructions without emulation, and we are just using `pwndbg.chain.get`, but we only check if the first pointer is readonly, not subsequent ones. I'll get a fix to this, just manually dereference until we get to a pointer to writable memory.

I'm also considering, maybe instead of making the enhance_can_dereference a parameter, we determine it inside of the format function by seeing if the last value corresponds to a page?

pwndbg/color/disasm.py
OBarronCS: Good catch - no reason. I moved it out.

pwndbg/color/disasm.py
OBarronCS: Elif works here - changed.

pwndbg/chain.py
OBarronCS: This case happens realistically in the following case:

There is an instruction after the current PC, and we are not emulating, like this:

lea    rax, [rip + 0x2e7d]

Where `rip + 0x2e7d` was a pointer to read-only memory, which contains a pointer to writable memory. We should stop dereferencing at the second pointer.

OBarronCS: Great - latest commits fix this bug - no longer always deference just because multiple values are in the list.

OBarronCS:
Latest commit makes changes from review, and adds an intermediate mode for the use of emulation where we only emulate jumps, but don't read from emulator registers or memory. These means there are now three different levels of emulation:

1. No emulation at all. 
2. Only emulate jumps - the only interaction with the emulator in this case is stepping it and reading the PC
3. Full emulation - read registers and memory from the emulator as well as determining jumps

This is reflected in the `emulate` setting, which is either `off`, `on`, or `jumps-only`.

pwndbg/chain.py
OBarronCS: Got back to this - newest commit cleans up a lot of this logic and provides an explanation as to why this only applies to lists of length 1

https://github.com/pwndbg/pwndbg/blob/d858f8a75539b74c1e46eef9e75641f9d451ab7a/pwndbg/chain.py#L158-L164

OBarronCS:
Latest commits fix some edge case bugs I came across (memory operands were not being resolved/dereferenced before stepping the emulator, which is now fixed by adding a new field to the EnhancedOperands). Simplify the telescoping logic, remove some code duplication, and rebase.

disconnect3d:
@gsingh93 can u help reviewing here? shall we just hit merge?:)

CptGibbon:
This looks like such great work, but 3000+ lines in a single PR is un-reviewable.

gsingh93:
The change seems mostly isolated to disasm code, the tests pass, and the linters pass, so I'm not too concerned about breakages. I say we merge it. I'm slightly worried about the caching changes I don't have time to understand right now, but if it looks good to you, it's fine with me.

But yes, if you could make your future PRs smaller @OBarronCS, that would be appreciated :)

disconnect3d:
I'll make another review pass this week and try to double check the perf stuff.

OBarronCS:
For sure - I got started making the feature and realized a lot of fundamental changes were required, and before I knew it there was a lot of refactoring of the disasm files and a lot of lines were changed. A big fundamental change that forced many new lines was creating a typed object to represent enhanced instructions - `PwndbgInstruction`. Previously, the code added attributes ad hoc to an object provided by the Capstone engine.

I'm very grateful for everyone that's looked at this and is part of Pwndbg - I'll be sure to make future changes more bite-sized.

I've added a blurb to the bottom of DEVELOPING.md with a high-level explanation of the annotation implementation (including notes on the general concepts, code flow, and caching).

disconnect3d:
@OBarronCS Thanks a lot for this work!

PR #2005, opened 2024-01-31
Add $base function

### LINKED ISSUE #2004: Feature request: RVA-based breakpoints
I am missing a command that would allow creating a breakpoint relative to some image - either main program or a library - for example at offset 0x1337 from libc base.

A new command that would enable this, for example

````
bp libc.so+0x1337
````

The alternative is to manually find the image bases which is tedious.

chivay:
Fixes #2004

Arguably, the test is not that great but it should do the job. :thinking:


disconnect3d:
Thanks!

PR #2009, opened 2024-02-02
cyclic: add argument to save output to file (fixes #2007)

### LINKED ISSUE #2007: pattern_create & run features
It is annoying to have to have multiple versions of gdb to complete some of my work. I don't understand why this feature hasn't been brought over yet like gdb-peda has implemented. Reversing takes long enough, this would make our lives a little bit easier.

I would like to add the pattern_create feature into pwndbg. As well as run, so that I can quickly create our cyclic values and then run our output (run < payload-100.txt) so we can check the registers in under 10 seconds without restarting the program.

disconnect3d:
This commit adds a `[filename]` argument to the `cyclic` command.

This makes it possible to do things like `cyclic 100 input` and `run < input` which was a feature Peda users used in the past.

Here is the full new help for cyclic command:

```
pwndbg> help cyclic
usage: cyclic [-h] [-a charset] [-n length] [-l lookup_value | count] [filename]

Cyclic pattern creator/finder.

positional arguments:
  count                 Number of characters to print from the sequence (default: print the
                        entire sequence) (default: 100)
  filename              Name (path) of the file to save the cyclic pattern to (default: )

options:
  -a charset, --alphabet charset
                        The alphabet to use in the cyclic pattern (default:
                        abcdefghijklmnopqrstuvwxyz)
  -n length, --length length
                        Size of the unique subsequences (defaults to the pointer size for the
                        current arch)
  -l lookup_value, -o lookup_value, --offset lookup_value, --lookup lookup_value
                        Do a lookup instead of printing the sequence (accepts constant values
                        as well as expressions)
```

### DIFF
--- a/pwndbg/commands/cyclic.py
+++ b/pwndbg/commands/cyclic.py
@@ -30,6 +30,7 @@ parser.add_argument(
     help="Size of the unique subsequences (defaults to the pointer size for the current arch)",
 )
 
+
 group = parser.add_mutually_exclusive_group(required=False)
 group.add_argument(
     "-l",
@@ -50,9 +51,17 @@ group.add_argument(
     help="Number of characters to print from the sequence (default: print the entire sequence)",
 )
 
+parser.add_argument(
+    "filename",
+    type=str,
+    help="Name (path) of the file to save the cyclic pattern to",
+    default="",
+    nargs="?",
+)
+
 
 @pwndbg.commands.ArgparsedCommand(parser, command_name="cyclic")
-def cyclic_cmd(alphabet, length, lookup, count=100) -> None:
+def cyclic_cmd(alphabet, length, lookup, count=100, filename="") -> None:
     if length:
         # Convert from gdb.Value
         length = int(length)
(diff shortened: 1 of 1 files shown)


disconnect3d:
The Arch Linux build failure is irrelevant to this PR. It fails on:
```
    def test_loads_binary_with_core_without_crashing():
        if not os.path.isfile(CORE):
            create_coredump = ["run", f"generate-core-file {CORE}"]
            run_gdb_with_script(binary=BINARY, pyafter=create_coredump)
            assert os.path.isfile(CORE)
        output = run_gdb_with_script(binary=BINARY, core=CORE).splitlines()
    
        assert any(f"Reading symbols from {BINARY}..." in line for line in output)
        assert any("pwndbg: loaded" in line for line in output)
>       assert "Program terminated with signal SIGFPE, Arithmetic exception." in output
E       AssertionError: assert 'Program terminated with signal SIGFPE, Arithmetic exception.' in ['Reading symbols from /pwndbg/tests/gdb-tests/tests/binaries/div_zero.out...', '/tmp/pwndbg-tests-div-zero-core: No such file or directory.', 'pwndbg: loaded ### pwndbg commands and ### shell commands. Type pwndbg [--shell | --all] [filter] for a list.', 'pwndbg: created $rebase, $base, $ida GDB functions (can be used with print/break)']
```

...not exactly sure why :)

disconnect3d:
The "Check lock files / lock_poetry" CI build fails on https://github.com/pwndbg/pwndbg/pull/2010 but an attempt to fix it in that PR made the check fail on a hash which is missing in poetry2nix which we depend on.

Anyway, all the other tests pass, so I am merging this.

PR #2010, opened 2024-02-02
Update poetry.lock

disconnect3d:


disconnect3d:
Note: the build is still failing after this, but on Nix. I partially fixed it, but cryptography 42.0.2 hash need to be added to poetry2nix: https://github.com/nix-community/poetry2nix/pull/1520 so the build is blocked on that.

PR #2015, opened 2024-02-07
Fix flake.lock for Cryptography==42.0.2

disconnect3d:


disconnect3d:
This fixes the check lock files CI workflow after https://github.com/nix-community/poetry2nix/pull/1520

PR #2018, opened 2024-02-12
Add a helper command for one_gadget

lebr0nli:
This PR adds a new command for finding valid [one_gadget](https://github.com/david942j/one_gadget) for the current context and makes it easier to debug when the constraints are unsat.

----

Here are some example usages:

After we control pc, if we get some luck, we might be able to find a satisfiable gadget via command: `onegadget`:
image

If there's no satisfiable one_gadget for the current context, we can run with `--show-unsat` to find out what constraint is unsatisfiable: 
image

If we want to know more information behind the unsatisfiable gadgets, we can run with `--verbose` or `-v`:
image


disconnect3d:
Any chance we can display the condition and SAT/UNSAT in two columns? Or somehow that it will be more readable? I feel like the output is bloated with colors and maybe not so readable now.

We could also rename the command to just `onegadget` too 🤔

lebr0nli:
> Any chance we can display the condition and SAT/UNSAT in two columns? Or somehow that it will be more readable? I feel like the output is bloated with colors and maybe not so readable now.

Yeah, I also feel that the output is a little bit messy at the moment.
I'll attempt to organize it into a table format with columns to see if that offers a cleaner presentation.

> We could also rename the command to just `onegadget` too 🤔

Sure!

lebr0nli:
Regarding the output format, how about something like this?
SCR-20240219-mdcn

^ I think this looks much better(?), but it sacrifices the ability to display the detailed information behind its UNSAT status.
This results that I still need to check each UNSAT constraint manually to find out how to make it SAT when debugging.

Should we use the tabulate output as default, but add an `-v` option to show the original messy but detailed output?
Or any other ideas?

disconnect3d:
> ^ I think this looks much better(?), but it sacrifices the ability to display the detailed information behind its UNSAT status. This results that I still need to check each UNSAT constraint manually to find out how to make it SAT when debugging.
> 
> Should we use the tabulate output as default, but add an `-v` option to show the original messy but detailed output? Or any other ideas?

This looks good

lebr0nli:
> This looks good

Ok!
I'll update the output format to the tabulate one.

pwndbg/commands/one_gadget.py
```diff
@@ -0,0 +1,50 @@
...
+import pwndbg.glibc
+from pwndbg.commands import CommandCategory
+
+parser = argparse.ArgumentParser(
+    formatter_class=argparse.RawTextHelpFormatter,
+    description="""Show one_gadget
+
+Examples:
+    one_gadget
+    one_gadget --show-unsat
```
anthraxx: The command has been renamed to `onegadget`

pwndbg/commands/one_gadget.py
```diff
@@ -0,0 +1,50 @@
...
+
+import pwndbg.color.message as M
+import pwndbg.commands
+import pwndbg.gdblib.one_gadget
+import pwndbg.glibc
+from pwndbg.commands import CommandCategory
+
+parser = argparse.ArgumentParser(
+    formatter_class=argparse.RawTextHelpFormatter,
+    description="""Show one_gadget
```
anthraxx: The command has been renamed to `onegadget`

disconnect3d: @lebr0nli perhaps also rename the file to `onegadget.py`

pwndbg/commands/one_gadget.py
lebr0nli: Ahh, yes, thanks, I'll fix it!

pwndbg/commands/one_gadget.py
lebr0nli: Sure, I'll rename it

pwndbg/gdblib/onegadget.py
```diff
@@ -0,0 +1,595 @@
...
+
+ADDRESS_WRITABLE_PATTERN = re.compile(r"address(?:es)? (.*) (?:is|are) writable")
+WRITABLE_COLON_PATTERN = re.compile(r"writable: (.*)")
+EQUAL_NULL_PATTERN = re.compile(r"(.*) == NULL")
+VALID_POSIX_SPAWN_FILE_ACTIONS_PATTERN = re.compile(r"(.*) <= 0")
+VALID_ARGV_PATTERN = re.compile(r"(.+) is a valid argv")
+VALID_ENVP_PATTERN = re.compile(r"(.+) is a valid envp")
+IS_ALIGNED_PATTERN = re.compile(r"(.+) & 0xf == (\d+)")
+IS_GOT_ADDRESS_PATTERN = re.compile(r"(.+) is the GOT address of libc")
+CAST_PATTERN = re.compile(r"^\([s|u]\d+\)")
```
disconnect3d, line 32: Fwiw all this `re.compile` etc. will slow down Pwndbg load time, but it seems negligible:
```
pwndbg> !cat a.py
import re
import time

a = time.time()
ADDRESS_WRITABLE_PATTERN = re.compile(r"address(?:es)? (.*) (?:is|are) writable")
WRITABLE_COLON_PATTERN = re.compile(r"writable: (.*)")
EQUAL_NULL_PATTERN = re.compile(r"(.*) == NULL")
VALID_POSIX_SPAWN_FILE_ACTIONS_PATTERN = re.compile(r"(.*) <= 0")
VALID_ARGV_PATTERN = re.compile(r"(.+) is a valid argv")
VALID_ENVP_PATTERN = re.compile(r"(.+) is a valid envp")
IS_ALIGNED_PATTERN = re.compile(r"(.+) & 0xf == (\d+)")
IS_GOT_ADDRESS_PATTERN = re.compile(r"(.+) is the GOT address of libc")
CAST_PATTERN = re.compile(r"^\([s|u]\d+\)")
b = time.time()

elapsed = b - a
print(elapsed)
pwndbg> source a.py
0.0009448528289794922
```

PR #2019, opened 2024-02-13
 Improved handling of commands containing newlines.

ttrssreal:
When certain commands contain newlines in their arguments, an error is triggered within the `check_repeated()` function of the `Command` class. This issue becomes apparent, for instance, when a user pastes text containing a newline character.
```
ai question
rdi + 1 = ?
rax + 1 = ?
```
So in `check_repeated()` merge the end of the entries array until the start of the command is identified.

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -172,7 +172,7 @@ class Command(gdb.Command):
             return False
 
         last_line: str
-        
+
         # Handle commands containing newlines
         while True:
             last_line = lines[-1]
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import argparse
 import functools
 import io
+import re
 from enum import Enum
 from typing import Any
 from typing import Callable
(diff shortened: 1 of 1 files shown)


disconnect3d:
Can you provide some more details why this is an issue or/and what is the state of variables when it happens? I don't have an OpenAI API key atm to test it out myself.

Also, can't we detect commands containing newlines somehow and have a condition on that instead of having an infinite loop with regex checking?

ttrssreal:
Its just with alot of commands, if theirs a newline, it errors and its a bit annoying. Ive just shown `ai` as your somewhat likely to paste, it fails for example with:
```
vmmap
arg
```
var `lines` represents the cmd history and is like `["  cmd_num  cmd_str", ..., ]`
until the last element (or few), missing a history number bc of the splitlines(), ex. `["  cmd_num  cmd_str", "  cmd_num  cmd_str", "cmd_str", "cmd_str"]` so I just merge the end of the array until its valid.

disconnect3d:
Wdym by a lot of commands? U should not invoke “vmmap\narg” ever I think

disconnect3d:
Ping @ttrssreal

disconnect3d:
Ping @ttrssreal

ttrssreal:
Hi sorry sorry, this just randomly annoyed me, its not a big deal, thanks for the great tool <3

PR #2029, opened 2024-02-25
Support multi-bit register bitflags

gsingh93:
Before, multi-bit register bitflags were kind of useless, see `EL` here:
image

Now they're actually useful:
image
image

### DIFF
--- a/pwndbg/color/context.py
+++ b/pwndbg/color/context.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from pwndbg.color import generateColorFunction
 from pwndbg.color import theme
 from pwndbg.gdblib import config
+from pwndbg.lib.regs import BitFlags
 
 config_prefix_color = theme.add_color_param(
     "code-prefix-color", "none", "color for 'context code' command (prefix marker)"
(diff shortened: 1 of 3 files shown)


pwndbg/lib/regs.py
```diff
@@ -4,12 +4,14 @@
...
 from __future__ import annotations
 
-import collections
+from collections import OrderedDict
 from typing import Any
 from typing import Dict
 from typing import Iterator
 from typing import Tuple
 
+BitFlags = OrderedDict[str, int | Tuple[int, int]]
```
disconnect3d: Apparently this causes a TypeError on Ubuntu 20.04 :/

gsingh93: I fixed it, the fix is weird to me because I'm using `typing.OrderedDict` as both a type hint and a constructor, but if it works it works.

pwndbg/color/context.py
```diff
@@ -97,15 +98,28 @@ def format_flags(value, flags, last=None):
...
-        if value & bit:
-            name = name.upper()
-            name = flag_set(name)
+        # If the size is not specified, assume it's 1
+        if isinstance(bit, int):
+            size = 1
         else:
-            name = name.lower()
-            name = flag_unset(name)
+            assert len(bit) == 2
```
disconnect3d, line 105: Consider adding a msg, so we know what is the assertion triggered :D

gsingh93: We should know from the stack trace, right?

PR #2040, opened 2024-02-28
context.py: refactor get_regs

disconnect3d:
Refactor get_regs for more readability / to perform (hopefully) less operations

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -554,16 +554,15 @@ pwndbg.gdblib.config.add_param("show-retaddr-reg", False, "whether to show retur
 
 
 def get_regs(regs: List[str] = None):
-    result = []
-
     if regs is None:
         regs = []
 
-    if len(regs) == 0:
+    elif len(regs) == 0:
         regs += pwndbg.gdblib.regs.gpr
-
-        regs.append(pwndbg.gdblib.regs.frame)
-        regs.append(pwndbg.gdblib.regs.stack)
+        regs += [
+            pwndbg.gdblib.regs.frame,
+            pwndbg.gdblib.regs.stack
+        ]
 
         if pwndbg.gdblib.config.show_retaddr_reg:
             regs += pwndbg.gdblib.regs.retaddr
@@ -575,6 +574,7 @@ def get_regs(regs: List[str] = None):
 
     changed = pwndbg.gdblib.regs.changed
 
+    result = []
     for reg in regs:
         if reg is None:
             continue


pwndbg/commands/context.py
```diff
@@ -575,6 +574,7 @@ def get_regs(regs: List[str] = None):
 
     changed = pwndbg.gdblib.regs.changed
 
+    result = []
```
disconnect3d, line 577: We could actually just yield results here too.

disconnect3d:
We also have this part there:
```py
        bit_flags = None
        if reg in pwndbg.gdblib.regs.flags:
            bit_flags = pwndbg.gdblib.regs.flags[reg]
        elif reg in pwndbg.gdblib.regs.extra_flags:
            bit_flags = pwndbg.gdblib.regs.extra_flags[reg]
        if bit_flags:
            desc = C.format_flags(value, bit_flags, pwndbg.gdblib.regs.last.get(reg, 0))
```
Shall we maybe do: `if, elif, else: bit_flags = None` instead of always setting it to `None`?

pwndbg/commands/context.py
```diff
@@ -554,16 +554,15 @@ def regs(regs=[]) -> None:
 
 
 def get_regs(regs: List[str] = None):
-    result = []
-
     if regs is None:
         regs = []
 
-    if len(regs) == 0:
+    elif len(regs) == 0:
```
gsingh93, line 560: I don't think this is correct. If `regs` is None, we want to enter this if statement too, but if written like this that won't happen and this function won't print anything.

pwndbg/commands/context.py
```diff
@@ -554,16 +554,15 @@ def regs(regs=[]) -> None:
...
         regs += pwndbg.gdblib.regs.gpr
-
-        regs.append(pwndbg.gdblib.regs.frame)
-        regs.append(pwndbg.gdblib.regs.stack)
+        regs += [
+            pwndbg.gdblib.regs.frame,
+            pwndbg.gdblib.regs.stack
+        ]
```
gsingh93, line 565: I'm fine with this, but I originally didn't do it like this because I was getting confused between which variables where collections/tuples and which weren't, and so I used `+=` whenever it was a collection and `append` when it wasn't, to make it obvious what was what.

pwndbg/commands/context.py
disconnect3d: Ugh, you are right. You know what.. let's just close it :D.

PR #2051, opened 2024-03-03
Add type hint for cache_until

gsingh93:
Adding this type hint required a number of other changes to make mypy happy.

### DIFF
--- a/pwndbg/auxv.py
+++ b/pwndbg/auxv.py
@@ -83,12 +83,10 @@ AT_CONSTANTS = {
 
 
 sys.modules[__name__].__dict__.update({v: k for k, v in AT_CONSTANTS.items()})
+AT_NAMES = set(AT_CONSTANTS.values())
 
 
-from typing import Protocol
-
-
-class AUXVProtocol(Protocol):
+class AUXV(Dict[str, Union[int, str]]):
     AT_PHDR: Optional[int]
     AT_BASE: Optional[int]
     AT_PLATFORM: Optional[str]
@@ -98,8 +96,6 @@ class AUXVProtocol(Protocol):
     AT_SYSINFO: Optional[int]
     AT_SYSINFO_EHDR: Optional[int]
 
-
-class AUXV(Dict[str, Union[int, str]]):
     def set(self, const: int, value: int) -> None:
         name = AT_CONSTANTS.get(const, "AT_UNKNOWN%i" % const)
 
@@ -113,15 +109,18 @@ class AUXV(Dict[str, Union[int, str]]):
 
         self[name] = value
 
-    def __getattr__(self, attr):
-        return self.get(attr)
+    def __getattr__(self, attr: str) -> Optional[Union[int, str]]:
+        if attr in AT_NAMES:
+            return self.get(attr)
+
+        raise AttributeError("%r object has no attribute %r" % (self.__class__.__name__, attr))
 
     def __str__(self) -> str:
         return str({k: v for k, v in self.items() if v is not None})
 
 
 @pwndbg.lib.cache.cache_until("objfile", "start")
-def get() -> AUXVProtocol:
+def get() -> AUXV:
     return use_info_auxv() or walk_stack() or Foo()  # AUXV()
 
 
--- a/pwndbg/auxv.py
+++ b/pwndbg/auxv.py
@@ -3,6 +3,9 @@ from __future__ import annotations
 import os
 import re
 import sys
+from typing import Dict
+from typing import Optional
+from typing import Union
 
 import gdb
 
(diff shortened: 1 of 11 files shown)


gsingh93:
@disconnect3d can you double check the vmmap changes?

gsingh93:
Hmm, typing `AUXV` with `TypedDict` might not be the right solution. Will figure something else out.

gsingh93:
@disconnect3d regarding the failing test, was it intentional to not get the permissions from `info proc mappings` in the coredump function? All of the other logic was the same as the other place we were parsing `info proc mappings`.

Also, let me know if you have a preference on how to type `AUXV`:
1. With properties + asserts, more verbose but will catch bugs at runtime: https://github.com/pwndbg/pwndbg/pull/2051/commits/263766da519b5ba3d3da830d3d61a4d0d375248f#diff-8ef9e44de34b73703432dc84b1278571651c36ab0c8e37cdbe1f0a4057ac341d
2. Less verbose, but won't catch bugs at runtime: https://github.com/pwndbg/pwndbg/pull/2051/commits/af6309c652bf614c3e1e9ce38db134638e86a53f#diff-8ef9e44de34b73703432dc84b1278571651c36ab0c8e37cdbe1f0a4057ac341d

pwndbg/auxv.py
```diff
@@ -80,11 +81,26 @@
...
+    AT_BASE: Optional[int]
+    AT_PLATFORM: Optional[str]
+    AT_ENTRY: Optional[int]
+    AT_RANDOM: Optional[int]
+    AT_EXECFN: Optional[str]
+    AT_SYSINFO: Optional[int]
+    AT_SYSINFO_EHDR: Optional[int]
+
+
 class AUXV(Dict[str, Union[int, str]]):
```
disconnect3d: Shouldnt this inherit from AUXVProtocol?

pwndbg/auxv.py
```diff
@@ -105,11 +121,11 @@ def __str__(self) -> str:
 
 
 @pwndbg.lib.cache.cache_until("objfile", "start")
-def get() -> AUXV:
-    return use_info_auxv() or walk_stack() or AUXV()
+def get() -> AUXVProtocol:
+    return use_info_auxv() or walk_stack() or Foo()  # AUXV()
```
disconnect3d, line 124: Foo?

pwndbg/gdblib/vmmap.py
```diff
@@ -351,6 +334,46 @@ def coredump_maps() -> Tuple[pwndbg.lib.memory.Page, ...]:
...
+    """
+
+    # We look for lines like:
+    # ['0x555555555000', '0x555555556000', '0x1000', '0x1000', 'rw-p', '/home/user/a.out']
+    try:
+        split_line = line.split()
+
+        # Permission info is only available in GDB versions >=12.1
+        # https://github.com/bminor/binutils-gdb/commit/29ef4c0699e1b46d41ade00ae07a54f979ea21cc
+        # Assume "rw-p" on older gdb versions
```
disconnect3d:
```suggestion
        # Assume "rwxp" on older gdb versions
```

pwndbg/gdblib/vmmap.py
```diff
@@ -251,25 +252,7 @@ def coredump_maps() -> Tuple[pwndbg.lib.memory.Page, ...]:
...
-        # ['0x555555555000', '0x555555556000', '0x1000', '0x1000', '/home/user/a.out']
-        try:
-            start, _end, size, offset, objfile = line.split()
-            start, size, offset = int(start, 16), int(size, 16), int(offset, 16)
-        except (IndexError, ValueError):
-            continue
-
-        # Note: we set flags=0 because we do not have this information here
-        pages.append(pwndbg.lib.memory.Page(start, size, 0, offset, objfile))
+    pages = list(info_proc_maps())
```
disconnect3d, line 255: This makes a difference that previously we set flags == permissions = 0, which I believe were later enhanced with `rw-p` permissions throughout the `maintenance info sections` information.

And now, the `info_proc_maps` if it doesn't have permissions info, it sets `rwxp` and that's why our test fails.

So in the end we set wrong information even though we can have a more correct one here.

pwndbg/gdblib/vmmap.py
disconnect3d: Actually I think this probably should be just "----" perm?

disconnect3d:
Fwiw since we are here, I think we should use the `proc_tid_maps` in `get()` whenever we are in GDB<12.1 because the perm info was added to GDB only in 12.1.

This is the code I'm talking about:
```py
def get():
...
    proc_maps = None
    if pwndbg.gdblib.qemu.is_qemu_usermode():
        # On Qemu < 8.1 info proc maps are not supported. In that case we callback on proc_tid_maps
        proc_maps = info_proc_maps()

    if not proc_maps:
        proc_maps = proc_tid_maps()
```

pwndbg/auxv.py
gsingh93: No, Protocols use structural subtyping, meaning as long as the structure can match the protocol, it doesn't actually need to inherit from it. Implementing `__getattr__` is why this class matches the structure, as it will handle any missing attributes.

It is allowed to inherit from the Protocol, but in this case we can't do so without explicitly defining each field in `AUXV`, which takes away the benefit of using the dictionary. But it's also more clear, which is why I haven't decided if the approach in the previous commit where we just add some `@property`s for each field is a better solution.

pwndbg/auxv.py
gsingh93: I was testing that I would get a type error if the class didn't match the protocol, forgot to revert this part :P

pwndbg/gdblib/vmmap.py
gsingh93: I see, I'll add a parameter to `info_proc_maps` to only optionally set the flag.

gsingh93:
> Fwiw since we are here, I think we should use the `proc_tid_maps` in `get()` whenever we are in GDB<12.1 because the perm info was added to GDB only in 12.1.
> 
> This is the code I'm talking about:
> 
> ```python
> def get():
> ...
>     proc_maps = None
>     if pwndbg.gdblib.qemu.is_qemu_usermode():
>         # On Qemu < 8.1 info proc maps are not supported. In that case we callback on proc_tid_maps
>         proc_maps = info_proc_maps()
> 
>     if not proc_maps:
>         proc_maps = proc_tid_maps()
> ```

`info_proc_maps` calls `parse_info_proc_mappings_line`, which handles the case where `perm` doesn't exist in GDB<12.1. Does this handle the issue you're talking about or am I missing something else?

disconnect3d:
But this case is handled by setting perms to rwxp (or w/e, we may change
it) instead of correct permissions. We should instead read the maps file
ourselves and parse it manually to get correct perms :/

gsingh93:
Makes sense, but should be done in a separate PR. I don't want to change any actual functionality with this one.

PR #2061, opened 2024-03-05
Return `int` in `fix_int_reraise`

gsingh93:
`fix_int_reraise` seems to indicate it should return an `int` from its name, but it seems we forgot to cast the result from `fix` to an int. This fixes that, and now all commands that have an argument of `type=int` will actually get an `int` instead of a `gdb.Value`. If the argument type is unspecified, `fix_reraise` will get called, which keeps the old behavior.

I removed some casts that are no longer necessary.

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -504,7 +504,7 @@ class _ArgparsedCommand(Command):
         self,
         parser: argparse.ArgumentParser,
         function,
-        command_name: Optional[str] = None,
+        command_name=None,
         *a,
         **kw,
     ) -> None:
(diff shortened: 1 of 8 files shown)


disconnect3d:
@gsingh93 Do we need the fix* family of functions at all? I recall they were used heavily back when we supported Python 2. To be honest I am not sure if we need it right now...

I think we have to decide how we want to go further with command arguments. I believe that almost no command need to take `gdb.Value` itself and whenever one is seen, it should be auto-casted to an `int`. I may be wrong tho?

gsingh93:
I just tested it, removing `fix_int`/`fix_int_reraise` does pass an `int` to the command like we want. I can remove those functions, but I'm hesitant about changing the `if action.type is None: action.type = fix_reraise` branch, just because I don't fully understand if it's still needed or not. Does that sound good?

PR #2063, opened 2024-03-06
Unable to set param.value when encoding not UTF-8

HyperEntangledQubit:
Summary
-------
Prior to this change, when attempting to run `pwndbg` in Ubuntu 22.04.03 LTS docker container on Apple M3 MBP (without setting the code to UTF-8) the following pwn.lib.config.Parameter(s) were unable to be set:
- `chain-arrow-left`
- `chain-arrow-right`
- `banner-separator`
- `nearpc-branchmarker`
- `nearpc-prefix`
- `telescope-offset-separator`
- `telescope-repeating-marker`
- `backtrace-prefix`
- `code-prefix`
- `hexdump-ascii-block-separator`

All of these params correspond to the UI for `pwndbg`. By bypassing the set of param.value for each of these `pwndbg` is able to run on arm64 inside a docker container.

Altered the encoding warnings to warn using built-in `UnicodeWarning`.

### DIFF
--- a/gdbinit.py
+++ b/gdbinit.py
@@ -7,11 +7,13 @@ import os
 import site
 import sys
 import time
+import warnings
 from glob import glob
 from os import environ
 from os import path
 
 _profiler = cProfile.Profile()
+warnings.simplefilter("default", UnicodeWarning)
 
 _start_time = None
 if environ.get("PWNDBG_PROFILE") == "1":
@@ -60,14 +62,16 @@ else:
 encoding = locale.getpreferredencoding()
 
 if encoding != "UTF-8":
-    print("******")
-    print(f"Your encoding ({encoding}) is different than UTF-8. pwndbg might not work properly.")
-    print("You might try launching GDB with:")
-    print("    LC_CTYPE=C.UTF-8 gdb")
-    print(
-        "If that does not work, make sure that en_US.UTF-8 is uncommented in /etc/locale.gen and that you called `locale-gen` command"
-    )
-    print("******")
+    enc_warning_msg = """******
+    Your encoding ({}) is different than UTF-8. pwndbg might now work properly.
+    You might try launching GDB with:
+        LC_CTYPE=C.UTF-8 gdb
+
+    If that does not work, make sure that en_US.UTF-8 is uncommented in /etc/locale.gen and that you called `locale-gen` command.
+    ******
+    """.format(encoding)
+
+    warnings.warn(enc_warning_msg, UnicodeWarning)
 
 environ["PWNLIB_NOTERM"] = "1"
 
--- a/pwndbg/gdblib/config.py
+++ b/pwndbg/gdblib/config.py
@@ -19,11 +19,12 @@ from __future__ import annotations
 from typing import Any
 
 import gdb
-
+import warnings
 import pwndbg.decorators
 import pwndbg.lib.config
 
 config = pwndbg.lib.config.Config()
+warnings.simplefilter("module", UnicodeWarning)
 
 # GDB < 9 does not support PARAM_ZUINTEGER*, so we implement it by ourselves for consistency
 IS_GDB_GTE_9 = True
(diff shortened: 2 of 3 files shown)


gdbinit.py
```diff
@@ -7,11 +7,13 @@
 import site
 import sys
 import time
+import warnings
 from glob import glob
 from os import environ
 from os import path
 
 _profiler = cProfile.Profile()
+warnings.simplefilter("default", UnicodeWarning)
```
gsingh93, line 16: Is line this necessary? I've never used this library, but the docs say this `UnicodeWarning` should already have the default action.

gsingh93:
@HyperEntangledQubit can you run `./lint.sh -f` and re-commit/push?

@disconnect3d I'm wondering if config options should have an optional non-unicode default value, so we can get rid of the Unicode checks altogether?

disconnect3d:
> @disconnect3d I'm wondering if config options should have an optional non-unicode default value, so we can get rid of the Unicode checks altogether?

Maybe but ideally we want things to _just work_ and set whatever is needed? That would be preferable at least.

gsingh93:
> > @disconnect3d I'm wondering if config options should have an optional non-unicode default value, so we can get rid of the Unicode checks altogether?
> 
> Maybe but ideally we want things to _just work_ and set whatever is needed? That would be preferable at least.

I would think having a non-unicode fallback would be the best way for things to "just work", otherwise you may be able to use pwndbg but various characters might not show up correctly and make it a confusing experience.

disconnect3d:
Sure but do we have alternatives that look nice enough? :) if so, then go
ahead.

Btw also I wonder if we really need so many configuration UI wise but oh
well…

gsingh93:
> Sure but do we have alternatives that look nice enough? :) if so, then go ahead.

I think for a lot of them they do, i.e. `chain-arrow-left` could just be `<-`, similar idea for a lot of symbols.

> Btw also I wonder if we really need so many configuration UI wise but oh well…

The benefit of them so far has been that people have been able to add their own non-unicode fallbacks. But I do agree there's too many and we should just hardcode some of them, choosing between unicode/non-unicode in the code. Then we could remove a bunch of them and then just individually add them back if anyone complains that they can't customize it anymore.

disconnect3d:
@HyperEntangledQubit Can you see if the issue gets fixed if you add the following before sourcing Pwndbg into your `~/.gdbinit`?:

```
set charset utf-8
set wide-charset utf-8 
```

If this fixes our issues, maybe we can force this during init and remove the whole "pls fix utf-8" issue?

disconnect3d:
Fwiw in Docker:
```
root@3a7eb74dfe9f:/pwndbg# gdb --quiet
******
Your encoding (utf-8) is different than UTF-8. pwndbg might not work properly.
You might try launching GDB with:
    LC_CTYPE=C.UTF-8 gdb
If that does not work, make sure that en_US.UTF-8 is uncommented in /etc/locale.gen and that you called `locale-gen` command
******
pwndbg: loaded 157 pwndbg commands and 40 shell commands. Type pwndbg [--shell | --all] [filter] for a list.
pwndbg: created $rebase, $base, $ida GDB functions (can be used with print/break)
Python Exception <class 'UnicodeEncodeError'>: 'ascii' codec can't encode characters in position 0-1: ordinal not in range(128)
/root/.gdbinit:3: Error in sourced command file:
Error occurred in Python: 'ascii' codec can't encode characters in position 0-1: ordinal not in range(128)
Use the vmmap command for a better & colored memory maps display (than the GDB's info proc mappings)
pwndbg> quit
```

```
root@3a7eb74dfe9f:/pwndbg# cat ~/.gdbinit
set charset utf-8
set wide-charset utf-8
source /pwndbg/gdbinit.py
root@3a7eb74dfe9f:/pwndbg# gdb --quiet
/root/.gdbinit:1: Error in sourced command file:
Undefined item: "utf-8".
(gdb)
```

disconnect3d:
Okay apparently its `set charset UTF-8` and I think it works?

As pointed out by @patryk4815 it requires GDB to be compiled with `libiconv` but we doubt any sane GDB build misses that.

disconnect3d:
I am going to close this PR. It seems we can fix all utf-8/unicode issues by doing `set charset UTF-8` and requiring this in Pwndbg.

An alternative solution I would accept is having non UTF-8/unicode defaults for the config variables, but this is also not what this PR adds, so yeah.

PR #2065, opened 2024-03-06
Install dev dependencies with poetry

gsingh93:
Add all dev dependencies to `pyproject.toml` and remove `dev-requirements.txt`. `setup-dev.sh` now installs poetry if it's not already installed and then installs all dependencies with `poetry install  --with dev`.

We should eventually install normal dependencies with poetry instead of pip, but let's start with just the dev ones and see if everything goes fine.


gsingh93:
@patryk4815 Could you look into the lock file issue?

patryk4815:
@gsingh93 fix:
```diff
diff --git a/nix/pwndbg.nix b/nix/pwndbg.nix
index 2bb0b5fb..5801d7e1 100644
--- a/nix/pwndbg.nix
+++ b/nix/pwndbg.nix
@@ -13,6 +13,8 @@ let
   ]);

   pyEnv = pkgs.poetry2nix.mkPoetryEnv {
+    groups = [];  # put [ "dev" ] to build "dev" dependencies
+    checkGroups = [];  # put [ "dev" ] to build "dev" dependencies
     projectDir = inputs.pwndbg;
     python = python3;
     overrides = pkgs.poetry2nix.overrides.withDefaults (self: super: {
```

> groups: Which Poetry 1.2.0+ dependency groups to install (default: [ "dev" ]).
> checkGroups: Which Poetry 1.2.0+ dependency groups to run unit tests (default: [  "dev" ]).

setup-dev.sh
```diff
@@ -188,6 +188,12 @@ if linux; then
         PWNDBG_VENV_PATH="./.venv"
     fi
     echo "Using virtualenv from path: ${PWNDBG_VENV_PATH}"
-    PYTHON=${PWNDBG_VENV_PATH}/bin/python
-    ${PYTHON} -m pip install -r dev-requirements.txt
+
+    # Install poetry if not already installed
+    if ! hash poetry 2> /dev/null; then
+        curl -sSL https://install.python-poetry.org | python3 -
```
disconnect3d, line 194: Shouldn't this be executed by the venv python?

gsingh93: No, the poetry docs explicitly say it shouldn't be installed in the project venv.

gsingh93:
Ping :)

PR #2069, opened 2024-03-06
Restore stack frame after display

Tcc100:
The commands `context threads` and `threads` use `thread.switch()` to examine other threads, which resets the selected stack frame to `#0`. This commit restores the selected frame afterwards.

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -882,6 +882,10 @@ def context_threads(with_banner=True, target=sys.stdout, width=None):
         original_thread = gdb.selected_thread()
     except SystemError:
         original_thread = None
+    try:
+        original_frame = gdb.selected_frame()
+    except gdb.error:
+        original_frame = None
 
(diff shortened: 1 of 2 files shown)


gsingh93:
Hi, can you run `lint.sh` and fix the lint errors?

Tcc100:
Ok, I've updated the code to address the linter.

disconnect3d:
@Tcc100 
> which resets the selected stack frame to #0

Do you mean that if I do `up; up; up` or `down; down; down` and then `context threads` I will end up with different stack frame? Or is the issue different?

Tcc100:
> > which resets the selected stack frame to #0
> 
> Do you mean that if I do `up; up; up` or `down; down; down` and then `context threads` I will end up with different stack frame? Or is the issue different?

Yes, exactly. However, only if there is more than one thread. Also plain `context` shows the threads and you will end up back at frame 0.

PR #2075, opened 2024-03-19
show `mypy` errors inline on PRs

### LINKED ISSUE #2053: Show lint errors as inline annotations in Github PRs
There are Github Actions that can run linters and report any violations inline on the PR itself (similar to the codecov comments we see now). This would be especially useful when we have new contributors who don't know to run the lint script, or struggle with fixing an issue. They would see the issue on the PR itself, and we could comment any suggestions there to help them.

I don't know exactly how to set this up, it would require some investigation.

diivi:
fixes #2053

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -48,4 +48,5 @@ jobs:
         install_types: false
         target: pwndbg
         filter_mode: nofilter
+        fail_on_error: true
 
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -32,3 +32,17 @@ jobs:
       run: |
         git diff-index --quiet HEAD -- pwndbg tests
         ./lint.sh
+        
+    - name: Run mypy
+      uses: tsuyoshicho/action-mypy@v4
+      with:
+        github_token: ${{ secrets.TOKEN }}
+        # Change reviewdog reporter if you need [github-pr-check,github-check,github-pr-review].
+        reporter: github-pr-review
+        # Change reporter level if you need.
+        # GitHub Status Check won't become failure with warning.
+        level: warning
+        # Change the current directory to run mypy command.
+        # mypy command reads setup.cfg or other settings file in this path.
+        execute_command: source ./.venv/bin/activate && mypy pwndbg
+
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -36,7 +36,7 @@ jobs:
     - name: Run mypy
       uses: tsuyoshicho/action-mypy@v4
       with:
-        github_token: ${{ secrets.TOKEN }}
+        github_token: ${{ secrets.GITHUB_TOKEN }}
         # Change reviewdog reporter if you need [github-pr-check,github-check,github-pr-review].
         reporter: github-pr-review
         # Change reporter level if you need.
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -38,7 +38,7 @@ jobs:
       with:
         github_token: ${{ secrets.GITHUB_TOKEN }}
         # Change reviewdog reporter if you need [github-pr-check,github-check,github-pr-review].
-        reporter: github-pr-review
+        reporter: github-pr-check
         # Change reporter level if you need.
         # GitHub Status Check won't become failure with warning.
         level: error
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -34,8 +34,6 @@ jobs:
         ./lint.sh
         
     - name: Run mypy
-      env:
-        REVIEWDOG_GITHUB_API_TOKEN: ${{ secrets.TOKEN }}
       uses: tsuyoshicho/action-mypy@v4
       with:
         github_token: ${{ secrets.TOKEN }}
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -44,7 +44,7 @@ jobs:
         level: error
         # Change the current directory to run mypy command.
         # mypy command reads setup.cfg or other settings file in this path.
-        execute_command: source ./.venv/bin/activate && mypy pwndbg
+        execute_command: poetry run mypy
         install_types: false
         target: pwndbg
         filter_mode: nofilter
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -41,8 +41,11 @@ jobs:
         reporter: github-pr-review
         # Change reporter level if you need.
         # GitHub Status Check won't become failure with warning.
-        level: warning
+        level: error
         # Change the current directory to run mypy command.
         # mypy command reads setup.cfg or other settings file in this path.
         execute_command: source ./.venv/bin/activate && mypy pwndbg
+        install_types: false
+        target: pwndbg
+        filter_mode: nofilter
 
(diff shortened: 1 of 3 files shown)


.github/workflows/lint.yml
```diff
@@ -28,7 +28,26 @@ jobs:
...
 
+
     - name: Run linters
       run: |
         git diff-index --quiet HEAD -- pwndbg tests
         ./lint.sh
+
+    - name: Activate virtualenv
+      run: |
+        source ~/.venv/bin/activate
```
diivi: Where should I source this from? this isn't working.

disconnect3d: Afair it should be `source ./.venv/bin/activate` since venv is created in the pwndbg directory if `PWNDBG_VENV_PATH` is not set.

disconnect3d: I just hope that the PATH and other env changes made by this will be propagated to the mypy step.

.github/workflows/lint.yml
```diff
@@ -28,7 +28,26 @@ jobs:
...
+        
+    - name: Run mypy
+      uses: tsuyoshicho/action-mypy@v4
+      with:
+        github_token: ${{ secrets.TOKEN }}
```
disconnect3d: We currently use the gh token to push releases in the releases action. I wonder if there is a way to limit the scope of this token here so that it can only push comments (as iiuc this is what it needs to do, right?)

diivi: I think you'll have to create a new token with just the minimum permissions needed to comment and use that here .

disconnect3d: image

https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token

disconnect3d: I guess the best action is to allow only 'read repository contents and packages permissions' and then change the release workflow to add permissions to its GITHUB_TOKEN.

diivi:
The task ran successfully finally, but I am still unable to get reviews on this PR by reviewdog, looking into it.

.github/workflows/lint.yml
```diff
@@ -32,3 +32,17 @@ jobs:
       run: |
         git diff-index --quiet HEAD -- pwndbg tests
         ./lint.sh
+        
+    - name: Run mypy
+      uses: tsuyoshicho/action-mypy@v4
+      with:
+        github_token: ${{ secrets.TOKEN }}
```
intrigus-lgtm:
```suggestion
        github_token: ${{ secrets.GITHUB_TOKEN }}
```
What happens if you apply this suggestion?

disconnect3d: Let's see, I committed this suggestion

diivi: I don't see the token not found error anymore, but neither do I see a review comment 😅

.github/workflows/lint.yml
```diff
@@ -32,3 +32,17 @@ jobs:
...
+        
+    - name: Run mypy
+      uses: tsuyoshicho/action-mypy@v4
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        # Change reviewdog reporter if you need [github-pr-check,github-check,github-pr-review].
+        reporter: github-pr-review
+        # Change reporter level if you need.
+        # GitHub Status Check won't become failure with warning.
+        level: warning
```
gsingh93: This should be "error", which matches the behavior before this change.

.github/workflows/lint.yml
```diff
@@ -32,3 +32,17 @@ jobs:
       run: |
         git diff-index --quiet HEAD -- pwndbg tests
         ./lint.sh
+        
+    - name: Run mypy
+      uses: tsuyoshicho/action-mypy@v4
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        # Change reviewdog reporter if you need [github-pr-check,github-check,github-pr-review].
+        reporter: github-pr-review
```
intrigus-lgtm: What happens if you use `github-check` instead?
```suggestion
        reporter: github-check
```
or `github-pr-check`
```suggestion
        reporter: github-pr-check
```

.github/workflows/lint.yml
```diff
@@ -32,3 +32,17 @@ jobs:
       run: |
         git diff-index --quiet HEAD -- pwndbg tests
         ./lint.sh
+        
+    - name: Run mypy
+      uses: tsuyoshicho/action-mypy@v4
```
gsingh93, line 37: Reading the docs, we also want `install_types: false` and `target: pwndbg`

gsingh93: I think we also want `filter_mode: nofilter`.

Not sure if that's related to the reason it's not showing results, maybe it's filtered if the actual type error isn't on the modified line.

.github/workflows/lint.yml
```diff
@@ -32,3 +32,20 @@ jobs:
...
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        # Change reviewdog reporter if you need [github-pr-check,github-check,github-pr-review].
+        reporter: github-pr-check
+        # Change reporter level if you need.
+        # GitHub Status Check won't become failure with warning.
+        level: error
+        # Change the current directory to run mypy command.
+        # mypy command reads setup.cfg or other settings file in this path.
+        execute_command: source ./.venv/bin/activate && mypy pwndbg
```
gsingh93: 1. You don't need to specify "pwndbg" at the end, it will be automatically added by the `target` option
2. The issue seems to be related to having the `source ./.venv/bin/activate &&`. I ran it locally, and this causes no output to be generated (it finishes so fast it seems like `mypy` doesn't even run), and I'm not sure why. Running it as two separate steps locally fixes the issue, but not sure how to do that here.

gsingh93: Actually, now that the dev dependencies are installed with poetry, we should be able to just do this:

```suggestion
        execute_command: poetry run mypy
```

diivi:
It finally worked! Thanks for collaborating @gsingh93 @intrigus-lgtm.  Should I revert the reporter type to github-pr-review?

disconnect3d:
What's the difference between github-pr-review and github-pr-check reporter type?

Btw the review comment added by lint can be observed in files changed:
image

diivi:
> What's the difference between github-pr-review and github-pr-check reporter type?

pr-review adds a comment to the PR, so you can see it in the conversation tab directly. pr-check makes it so that you can see the error in the files changed tab only.

https://github.com/reviewdog/reviewdog?tab=readme-ov-file#reporter-github-pullrequest-review-comment--reportergithub-pr-review

disconnect3d:
Let's change to comments, I feel like its better

gsingh93:
I actually prefer the annotations :D

gsingh93:
I actually prefer the annotations, like how `ruff` and `codecov` currently do it.

But regardless, can we make the `lint` job fail if `mypy` fails? I think that's also a setting on the mypy action.

disconnect3d:
@gsingh93 can be annotations. tbh i don't like them for codecov :P

diivi:
Works, reverting my test change now.

gsingh93:
The reason I like annotations is that sometimes these lint checks can get spammy, and they can clutter up the comments making it hard to see what actual reviewers are commenting. On the "Files changed tab", you can also separately hide annotations from comments, which I sometimes do if codecov gets annoying:
image

It is slightly less visible, but I think if the lint job fails that should be a good indication to look at the "Files changed".

gsingh93:
Thanks @diivi, this is going to make things a lot more convenient contributors!

Issue #1677, opened 2023-04-19
`heap-dereference-limit` is not respected by heap commands


CptGibbon:
Reported by @yooo1999 via Discord

Shown below, the `unsortedbin` command ignoring the `heap-dereference-limit` config variable and only printing 5 chunks.

![unsortedbin limit wrong](https://user-images.githubusercontent.com/16000770/233204063-47e2e28b-69f7-4796-8689-af5a596f87b7.png)

The `heap-dereference-limit` config variable is indeed used to *fetch* an appropriate number of chunks for the heap commands via `GlibcMemoryAllocator.bin_at()`:
https://github.com/pwndbg/pwndbg/blob/e6037f547acf94b3badf0b2dea72cc72b22adf3e/pwndbg/heap/ptmalloc.py#L1184-L1192

But when this data is formatted by `pwndbg.commands.heap.format_bin()` no explicit limit is set, resulting in the default value from `pwndbg.chain` being used (the `dereference-limit` config variable).
https://github.com/pwndbg/pwndbg/blob/e6037f547acf94b3badf0b2dea72cc72b22adf3e/pwndbg/commands/heap.py#L81

One solution could be to pass `heap-dereference-limit` as an explicit argument to `pwndbg.chain.format()` in `pwndbg.commands.heap.format_bin()`.

disconnect3d:
@CptGibbon Is this still a valid bug?

CptGibbon:
> @CptGibbon Is this still a valid bug?

I shall check, if it's not fixed it could be an easy win for a GSoC contributor.

CptGibbon:
This is indeed still a valid bug.

disconnect3d:
This was implemented in https://github.com/pwndbg/pwndbg/pull/2077

PR #2077, opened 2024-03-19
Accept heap dereference limit

diivi:
image
pwndbg should accept the heap-derefence limit param now.


CptGibbon:
I don't think this works as intended.
In the screenshot below (with this PR checked out) there are 8 chunks in the unsortedbin, but when `heap-dereference-limit` is set to 8, the unsortedbin command still considers the bin corrupt.

Screenshot 2024-03-19 at 11 08 46
I suppose it still prints 8 chunks though, and this is a step towards getting this feature working properly.

diivi:
I might have missed something, can you explain what the exact requirements are?

CptGibbon (approved):
I think the bin corruption issue I mentioned in my previous comment can be considered a separate issue, so I'm happy for this to be merged since it does fix #1677

PR #2081, opened 2024-03-22
search --asm: look for assembly instruction's bytes

### LINKED ISSUE #2078: search --asm: look for assembly instruction[s] bytes
The `search` command should have a way to search for a given assembly instruction[s] bytes.

We can assemble the given instructions through pwntools (its `pwnlib` import) similarly as we do in the `asm` command implementation: https://github.com/pwndbg/pwndbg/blob/c0d785565b499ba32d674c9e84a27e4967aee315/pwndbg/commands/asm.py#L69

This can probably be implemented as a `-a --asm` or `-t|--type=asm` option, where if provided, we would search for assebmled bytes.

E.g. in x86-64 programs `search --asm "xor rax, rax"` should assemble it through pwnlib to the following bytes:
  ```py
  In [4]: asm('xor rax, rax', arch='amd64')
  Out[4]: b'H1\xc0'
  ```
And then search the memory for those bytes (`b'H1\xc0'`). Ofc it should work with all the other options in `search` command like `--writable` or the ability to pass in a mapping name as the last argument of the search.

diivi:
Fixes #2078 
image

### DIFF
--- a/pwndbg/commands/search.py
+++ b/pwndbg/commands/search.py
@@ -6,6 +6,8 @@ import codecs
 import os
 import struct
 
+import pwnlib
+
 import pwndbg.color.memory as M
 import pwndbg.commands
 import pwndbg.enhance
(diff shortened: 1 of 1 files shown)


disconnect3d:
Thanks!

PR #2082, opened 2024-03-22
Fetch C struct as Python dictionary

CptGibbon:
Add a function that fetches the contents of a C struct from memory in the form of a Python dictionary.
e.g. `fetch_struct ( <struct_name>, <struct_address> ) -> dict`

Struct fields can be filtered with include/exclude lists.

The purpose of this function is to make testing easier by providing some abstraction from the gdb library.
Eliminating the need to handle some `gdb.Value`s could relieve the burden of writing tests that need to run in gdb or mocking gdb classes.

I think this could be particularly helpful in the heap parsing code, which handles a few structs.
The downside is that you lose gdb's lazy fetching, but I believe this could be balanced somewhat by only requesting the fields your code needs, e.g. just the size field of a malloc_chunk struct.

I chose the dictionary format because you can access the fields of a `gdb.Value` in the same way, which hopefully reduces the changes required to integrate this function. Below is an example of what it's doing, starting with the C code representing the structs it's fetching, then the dictionary returned by the function in question with example uses of the filters.

Screenshot 2024-03-22 at 11 35 33

Screenshot 2024-03-29 at 10 15 58

Note that anonymous structs/unions are flattened, as per ISO C11.
The exclude filter can't exclude fields from anonymous structs.

This work is similar to the `dt` command, but I found that code was very much tied up with its printing logic, so started from scratch here.

### DIFF
--- a/pwndbg/gdblib/memory.py
+++ b/pwndbg/gdblib/memory.py
@@ -5,9 +5,7 @@ Reading, writing, and describing memory.
 from __future__ import annotations
 
 import re
-from typing import Callable
-from typing import Dict
-from typing import Union
+from typing import Callable, Dict, Union
 
 import gdb
 
@@ -17,8 +15,7 @@ import pwndbg.gdblib.qemu
 import pwndbg.gdblib.typeinfo
 import pwndbg.lib.cache
 import pwndbg.lib.memory
-from pwndbg.lib.memory import PAGE_MASK
-from pwndbg.lib.memory import PAGE_SIZE
+from pwndbg.lib.memory import PAGE_MASK, PAGE_SIZE
 
 GdbDict = Dict[str, Union["GdbDict", int]]
 
@@ -408,13 +405,13 @@ def pack_struct_into_dictionary(
 
 
 def convert_gdb_value_to_python_value(gdb_value: gdb.Value) -> int | GdbDict:
-    type_code_conversions: dict[int, Callable[[gdb.Value], int | GdbDict]] = {
-        gdb.TYPE_CODE_PTR: int,
-        gdb.TYPE_CODE_INT: int,
-        gdb.TYPE_CODE_STRUCT: pack_struct_into_dictionary,
-    }
-
     gdb_type = gdb_value.type.strip_typedefs()
-    conversion_function = type_code_conversions[gdb_type.code]
 
-    return conversion_function(gdb_value)
+    if gdb_type.code == gdb.TYPE_CODE_PTR:
+        return int(gdb_value)
+    elif gdb_type.code == gdb.TYPE_CODE_INT:
+        return int(gdb_value)
+    elif gdb_type.code == gdb.TYPE_CODE_STRUCT:
+        return pack_struct_into_dictionary(gdb_value)
+
+    raise NotImplementedError
--- a/pwndbg/gdblib/memory.py
+++ b/pwndbg/gdblib/memory.py
@@ -5,7 +5,8 @@ Reading, writing, and describing memory.
 from __future__ import annotations
 
 import re
-from typing import Callable, Dict, Union
+from typing import Dict
+from typing import Union
 
 import gdb
 
@@ -15,7 +16,8 @@ import pwndbg.gdblib.qemu
 import pwndbg.gdblib.typeinfo
 import pwndbg.lib.cache
 import pwndbg.lib.memory
-from pwndbg.lib.memory import PAGE_MASK, PAGE_SIZE
+from pwndbg.lib.memory import PAGE_MASK
+from pwndbg.lib.memory import PAGE_SIZE
 
 GdbDict = Dict[str, Union["GdbDict", int]]
 
--- a/pwndbg/gdblib/memory.py
+++ b/pwndbg/gdblib/memory.py
@@ -360,3 +360,37 @@ def update_min_addr() -> None:
     global MMAP_MIN_ADDR
     if pwndbg.gdblib.qemu.is_qemu_kernel():
         MMAP_MIN_ADDR = 0
+
+
+def fetch_struct_as_dictionary(struct_name: str, struct_address: int):
+    struct_type = gdb.lookup_type("struct " + struct_name)
+    fetched_struct = poi(struct_type, struct_address)
+
+    return pack_struct_into_dictionary(fetched_struct)
+
+
+def pack_struct_into_dictionary(fetched_struct: gdb.Value):
+    struct_as_dictionary = {}
+    for field in fetched_struct.type.fields():
+        if field.name is None:
+            # Flatten anonymous structs/unions
+            struct_as_dictionary.update(convert_gdb_value_to_python_value(fetched_struct[field]))
+        else:
+            key = field.name
+            value = convert_gdb_value_to_python_value(fetched_struct[field])
+            struct_as_dictionary[key] = value
+
+    return struct_as_dictionary
+
+
+def convert_gdb_value_to_python_value(gdb_value: gdb.Value):
+    type_code_conversions = {
+        gdb.TYPE_CODE_PTR: int,
+        gdb.TYPE_CODE_INT: int,
+        gdb.TYPE_CODE_STRUCT: pack_struct_into_dictionary,
+    }
+
+    gdb_type = gdb_value.type.strip_typedefs()
+    conversion_function = type_code_conversions[gdb_type.code]
+
+    return conversion_function(gdb_value)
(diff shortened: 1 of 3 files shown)


pwndbg/gdblib/memory.py
```diff
@@ -360,3 +360,55 @@ def update_min_addr() -> None:
...
+    struct_type = gdb.lookup_type("struct " + struct_name)
+    fetched_struct = poi(struct_type, struct_address)
+
+    return pack_struct_into_dictionary(fetched_struct, include_only_fields, exclude_fields)
+
+
+def pack_struct_into_dictionary(
+    fetched_struct: gdb.Value,
+    include_only_fields: list[str] | None = None,
+    exclude_fields: list[str] | None = None,
```
gsingh93: `exclude_fields` would be better as a `Set[str]`. And maybe if it's none, we initialize it to `set()` so we can remove the `is None` check below?

pwndbg/gdblib/memory.py
```diff
@@ -360,3 +360,55 @@ def update_min_addr() -> None:
...
+):
+    struct_as_dictionary = {}
+
+    if include_only_fields is not None:
```
gsingh93: I think the implementation would be cleaner if you just had a single for loop and this if check was done inside of it. I understand it's done like this for performance reasons, but I can't imagine any struct with enough fields where this would make a noticeable performance improvement. I could be wrong though and don't feel too strongly about this, so feel free to leave it if you disagree.

pwndbg/gdblib/memory.py
```diff
@@ -360,3 +360,55 @@ def update_min_addr() -> None:
...
+            elif exclude_fields is None or field.name not in exclude_fields:
+                key = field.name
+                value = convert_gdb_value_to_python_value(fetched_struct[field])
+                struct_as_dictionary[key] = value
+
+    return struct_as_dictionary
+
+
+def convert_gdb_value_to_python_value(gdb_value: gdb.Value):
+    type_code_conversions = {
```
gsingh93: The `mypy` error will probably be fixed if you do something like `type_code_conversions: Dict[int, Callable[..., Any]]`, assuming the keys are ints. You could make the Callable type more specific right now, but I don't know what it would be off the top of my head, I'd need to test it.

pwndbg/gdblib/memory.py
```diff
@@ -360,3 +360,55 @@ def update_min_addr() -> None:
...
+        for field_name in include_only_fields:
+            key = field_name
+            value = convert_gdb_value_to_python_value(fetched_struct[field_name])
+            struct_as_dictionary[key] = value
+    else:
+        for field in fetched_struct.type.fields():
+            if field.name is None:
+                # Flatten anonymous structs/unions
+                struct_as_dictionary.update(
+                    convert_gdb_value_to_python_value(fetched_struct[field])
```
gsingh93: I would add an `assert` checking `isinstance(convert_gdb_value_to_python_value(fetched_struct[field]), Dict)`, given `convert_gdb_value_to_python_value` can return other types.

gsingh93 (commented):
Overall this would be a great change, and it would make testing much easier.

I left some nitpicky comments, but other than that could you add some return type annotations and fix the lint (mypy) errors?

pwndbg/gdblib/memory.py
CptGibbon: Do you reckon `include_only_fields` should be a `Set` too?

pwndbg/gdblib/memory.py
CptGibbon: Yeah I need help with this mypy error, I battled with it for a while and couldn't make mypy happy.

pwndbg/gdblib/memory.py
CptGibbon: So _in theory_ it should always return a dictionary when `field.name` is `None` because only structs & unions can be anonymous. But perhaps better safe than sorry, let me know if you're happy to omit the assert or if you think it'll catch something.

pwndbg/gdblib/memory.py
CptGibbon: Yeah I see what you're getting at, but I'm keen to leave it as is.
The benefits of a separate loop are, as you mentioned, the efficiency of:
* Only selecting desired fields rather than iterating across every field
* Not trying to parse both include/exclude lists

I haven't seen any giant structs either, but there isn't a strict limit on their size so I'd rather err on the side of caution for the sake of a few duplicated lines here.

pwndbg/gdblib/memory.py
```diff
@@ -360,3 +360,55 @@ def update_min_addr() -> None:
...
+            elif field.name not in exclude_fields:
+                key = field.name
+                value = convert_gdb_value_to_python_value(fetched_struct[field])
+                struct_as_dictionary[key] = value
+
+    return struct_as_dictionary
+
+
+def convert_gdb_value_to_python_value(gdb_value: gdb.Value):
+    type_code_conversions = {
```
disconnect3d: This dict is build on each function call. Let's move it outside of this function so it is created only once in the module and re-used.

Also do we really need this dict at all? There aren't many options, I know we want to be lazy, but maybe it'd be better to do `if elif ..`?

disconnect3d (approved):
Left one comment regarding type converters, but other than that LGTM

gsingh93:
@CptGibbon I pushed the `mypy` fix to this PR, PTAL.

https://github.com/pwndbg/pwndbg/pull/2082/commits/81abd70e4e43e686f5186b8d8d0827c4e8138ab9 is the basic fix for the error. I annotated the dict with the `Callable` type, and the assertion I mentioned in a previous comment was necessary for the next errors to be fixed.

The fix used `Any`, so to get rid of it I introduced a new type in the next commit: https://github.com/pwndbg/pwndbg/pull/2082/commits/962e5c07e7810a982c515652636826f038fecc05

I don't particularly like the name, feel free to change it.

(I'm not sure what the error on 20.04 is about, I don't have time to dig into it now)

pwndbg/gdblib/memory.py
gsingh93: Adding the assertion was necessary for the types to work out, see the latest commits.

pwndbg/gdblib/memory.py
gsingh93: Fixed in the latest commits I pushed.

pwndbg/gdblib/memory.py
gsingh93: `exclude_fields` was the only one where it really is necessary, as we should use `set` or `dict` if we need to use `in` with the variable. But there's no harm making `include_only_fields` a `set` too.

CptGibbon:
@gsingh93 
> @CptGibbon I pushed the `mypy` fix to this PR, PTAL.

Many thanks, I wasn't enjoying fighting with the type checker.

> I don't particularly like the name, feel free to change it.

No complaints from me.

> (I'm not sure what the error on 20.04 is about, I don't have time to dig into it now)

It says "type is not subscriptable", appears related to the GdbDict.

pwndbg/gdblib/memory.py
CptGibbon: Yeah perhaps I was trying to be terse when I used the dict here, I'm happy to simplify it which could also alleviate the type checker issue. I'll wait on @gsingh93 to see what's going on with that Ubuntu 20.04 test first because it looks related.

gsingh93:
Ok, 20.04 should be fixed now. This PR looks good to me now, unless you want to make any more changes.

CptGibbon:
@gsingh93 for the record, the method you used to push your fix to my branch left it in a difficult state to maintain.
If I want to continue working on it, there are 32(!) commits to pull and you're now on record as having committed my original 4 commits, so they're now considered "unpushed" on my local branch and I'm afraid of what will happen if I try to sync with the remote.

gsingh93:
@CptGibbon sorry for the trouble! I rebased out of habit 😓 

Here's what I would do to sync if it was me:
```bash
# backup
git checkout fetch-struct-as-dict
git checkout -b fetch-struct-as-dict-backup
git checkout fetch-struct-as-dict

# pull changes (will overwrite everything on fetch-struct-as-dict)
git pull --force

# if you had uncommitted changes on the backup branch
git cherry-pick <hash>
# or use git rebase if you're more familiar with it
```

Let me know if you need any help.

CptGibbon:
@disconnect3d made the change we spoke about.

Not sure why the docker build test is failing now.
The lint issue could be because I built my dev container before the change from black to ruff.

gsingh93:
Docker issue was a transient issue, rerunning it fixed it.

I can rerun the lint for you before merging if this is good to go now.

disconnect3d (approved):
lgtm

gsingh93:
Fixed the lint, merging.

PR #2087, opened 2024-03-24
Port checksec command

### LINKED ISSUE #1874: Port checksec command to native command and add colors

diivi:
Aims to fix #1874 by using checksec from pwnlib instead of calling the checksec command. Maybe I need to remove the wrapper file and shift its code to the command file, but I wasn't sure.

### DIFF
--- a/pwndbg/commands/checksec.py
+++ b/pwndbg/commands/checksec.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import argparse
 
-
 import pwndbg.color
 import pwndbg.commands
 import pwndbg.gdblib.file
--- a/pwndbg/commands/checksec.py
+++ b/pwndbg/commands/checksec.py
@@ -1,17 +1,19 @@
 from __future__ import annotations
+
 import argparse
+
 
 import pwndbg.color
 import pwndbg.commands
 import pwndbg.gdblib.file
 import pwndbg.wrappers.checksec
-from pwnlib.elf import ELF
 
-parser = argparse.ArgumentParser(description="Prints out the binary security settings using `checksec`.")
-
-parser.add_argument(
-    "-f", "--file", type=str, help="Specify the file to run `checksec` on."
+parser = argparse.ArgumentParser(
+    description="Prints out the binary security settings using `checksec`."
 )
+
+parser.add_argument("-f", "--file", type=str, help="Specify the file to run `checksec` on.")
+
 
 def color_line(line: str) -> str:
     return pwndbg.color.normal(
--- a/pwndbg/commands/checksec.py
+++ b/pwndbg/commands/checksec.py
@@ -1,10 +1,17 @@
 from __future__ import annotations
+import argparse
 
 import pwndbg.color
 import pwndbg.commands
 import pwndbg.gdblib.file
 import pwndbg.wrappers.checksec
+from pwnlib.elf import ELF
 
+parser = argparse.ArgumentParser(description="Prints out the binary security settings using `checksec`.")
+
+parser.add_argument(
+    "-f", "--file", type=str, help="Specify the file to run `checksec` on."
+)
 
 def color_line(line: str) -> str:
     return pwndbg.color.normal(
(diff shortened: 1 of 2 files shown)


chrf01:
I think we should use pwnlib's `ELF` everywhere instead of the checksec wrapper
Its also used in the `got` command:
https://github.com/pwndbg/pwndbg/blob/83d8d95b576b749e888f533ce927ad5a77fb957b/pwndbg/commands/got.py#L118-L119

diivi:
> https://github.com/pwndbg/pwndbg/blob/83d8d95b576b749e888f533ce927ad5a77fb957b/pwndbg/commands/got.py#L118-L119

Wouldn't this use ELF only ultimately since I use it to get relro_status and pie_status?

chrf01:
oh, yeah, you are right. missed it

pwndbg/wrappers/checksec.py
```diff
@@ -1,30 +1,16 @@
...
+    elf = ELF(local_path)
+    output = f"File: {elf.path}\n"
+    output += f"Arch: {elf.arch}\n"
+    output += elf.checksec()
+    return output
 
 
-@pwndbg.wrappers.OnlyWithCommand(cmd_name, cmd_pwntools)
-@pwndbg.lib.cache.cache_until("objfile")
-def get_raw_out(local_path: str) -> str:
```
disconnect3d, line 15: The change to `get_raw_out` breaks all other functions that use it.

It seems the `get_raw_out` gets the `get_raw_out.cmd` injected through the `OnlyWithCommand` decorator which is rather a debatable or overengineered design. We should probably refactor it in the future.

pwndbg/wrappers/checksec.py
```diff
@@ -1,30 +1,16 @@
...
+from pwnlib.elf import ELF
 
-import pwndbg.commands
-import pwndbg.lib.cache
-import pwndbg.wrappers
 
-cmd_name = "checksec"
-cmd_pwntools = ["pwn", "checksec"]
+def get_raw_out(local_path: str) -> str:
+    elf = ELF(local_path)
```
disconnect3d, line 7: If you look over the `ELF` class source (e.g. via [IPython interactive shell's](https://ipython.readthedocs.io/en/stable/) `ELF??` magic) it keeps opened handles to some resources:
image

The ELF object is also a context manager (implements `__enter__` and `__exit__`) so we should use it with a [`with` statement](https://docs.python.org/3/reference/compound_stmts.html#with) as follows:
```py
with ELF(path) as e:
    output += e.checksec()
```

pwndbg/commands/checksec.py
```diff
@@ -18,8 +26,9 @@ def color_lines(output: str) -> str:
...
 
 
-@pwndbg.commands.ArgparsedCommand("Prints out the binary security settings using `checksec`.")
+@pwndbg.commands.ArgparsedCommand(parser, command_name="checksec")
 @pwndbg.commands.OnlyWithFile
-def checksec() -> None:
-    output = pwndbg.wrappers.checksec.get_raw_out(pwndbg.gdblib.file.get_proc_exe_file())
+def checksec(file: str) -> None:
+    local_path = file or pwndbg.gdblib.file.get_proc_exe_file()
+    output = pwndbg.wrappers.checksec.get_raw_out(local_path)
```
disconnect3d, line 33: I believe we can just invoke pwnlib's ELF.checksec() here directly instead of implementing it in another file.

The idea of having things like this in another files is so that they can be reused in another places of the codebase, but I doubt there will be any other use of `ELF.checksec` in Pwndbg.

disconnect3d (changes requested):
See comments

pwndbg/wrappers/checksec.py
diivi: Thanks

pwndbg/wrappers/checksec.py
diivi: Can you give an example of a function that breaks? I thought if the tests pass then it should work fine.

pwndbg/commands/checksec.py
diivi: I didn't use `ELF.checksec()` here because `relro_status` and `pie_status` in `wrappers/checksec.py` depend on `get_raw_out`. So I thought `get_raw_out` should not be removed.

Or I can do the same thing here and keep `get_raw_out`  in  `wrappers/checksec.py`

pwndbg/wrappers/checksec.py
disconnect3d: Oh, I thought relro and pie status use a different CLI tool and so would not work. My bad.

Then this is fine.

I was also thinking that maybe we should have a helper func that would return a dict of all properties that we use here, but considering that `ELF.checksec()` has lots of code, then I don't think we want to repeat that.

OTOH, I still dislike that we check for string in output in relro/pie status. Those can be easily fetched via `e.relro` and `e.pie`.

Maybe we should:
1. Have a helper function to return a cached (through `pwndbg.lib.cache.cache_until(..)`) `ELF(...)` object (yeah I understand that then we will kinda leak the resource, but its fine)
2. Use it like `get_elf(path).checksec()`, `get_elf(path).relro` etc.

disconnect3d: @diivi ping, are u going to address above?

disconnect3d (approved):
On another look, this is fine.

PR #2090, opened 2024-03-25
Create config options from choices

### LINKED ISSUE #1414: Add a way to create config options from a fixed set of choices
As an example, the `context-sections` setting has a fixed set of values it can be set to, i.e. "regs", "stack", etc. The `validate_context_sections` will be triggered when this setting changes to validate that these options are correct.

However, we have other config options that don't have this validation, and adding a new config option is a hassle to add this type of validation every time. We should do something similar to argparse, where we can specify `choices`, and the value will be validated when it is set automatically.

diivi:
Fixes #1414  by adding a way to validate config options from choices.

Need to figure out where else this might be needed (or not) and also edit tests.

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -94,6 +94,12 @@ config_context_sections = pwndbg.gdblib.config.add_param(
         "expressions",
         "threads",
         "heap-tracker",
+        "''",
+        '""',
+        "none",
+        "empty",
+        "-",
+        "",
     ],
 )
 config_max_threads_display = pwndbg.gdblib.config.add_param(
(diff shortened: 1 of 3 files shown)


pwndbg/gdblib/config.py
```diff
@@ -95,6 +95,11 @@ def __get_set_string_gdb_gte_9(self) -> str:
             self.param.value = self.value
 
         for trigger in config.triggers[self.param.name]:
+            # validate param value before trigger
+            if not config.validate_choices(self.param.name, self.param.value):
+                self.param.revert_default()
+                return f"Invalid value '{self.param.value}' for parameter '{self.param.name}'. Valid choices are: {', '.join(self.param.choices)}"
```
disconnect3d, line 101: Does this if need to be in the trigger loop? The condition do not use `trigger` at all and here we will be re-validating things on each trigger. Is that needed?

pwndbg/commands/context.py
```diff
@@ -84,6 +84,23 @@ def clear_screen(out=sys.stdout) -> None:
...
+        "backtrace",
+        "expressions",
+        "threads",
+        "heap-tracker",
+        "''",
+        '""',
+        "none",
+        "empty",
+        "-",
+        "",
```
disconnect3d, line 102: I think we added `''`, `""`, `-`, `none` and `empty` here bcoz users could not know the available options. We probably don't need all those options if an info is displayed on avaialble options etc.

disconnect3d:
@diivi Can you rebase this to the latest pwndbg dev commit, fix conflicts, answer the comments and show us a screenshot how this works before and after this PR?

disconnect3d:
Closing this for now since there is no activity and there are conflicts. Please reopen if u are going to finish working on it.

PR #2094, opened 2024-03-25
when gdb.execute throws an error in symbol.get return ""

### LINKED ISSUE #2080: Problem of debugging rust applications
Hello! 
I have problem of debugging rust applications.

The problem with rust was noticed after the build and  upgrade of GDB to 14.2.0 version.

### The stack trace:

pwndbg> b 2
Breakpoint 1 at 0x8934: file src/main.rs, line 2.
pwndbg> r
Starting program: /target/debug/asm 

Breakpoint 1, asm::main () at src/main.rs:2
19	    println!("Hello, world!");
```
Warning: the current language does not match this frame.
  File "/GDB/pwndbg/pwndbg/commands/__init__.py", line 196, in __call__
  File "/GDB/pwndbg/pwndbg/commands/__init__.py", line 356, in _OnlyWhenRunning
    return function(*a, **kw)
  File "/GDB/pwndbg/pwndbg/commands/context.py", line 406, in context
    func(
  File "/GDB/pwndbg/pwndbg/commands/context.py", line 514, in context_regs
    regs = get_regs()
  File "/GDB/pwndbg/pwndbg/commands/context.py", line 606, in get_regs
    desc = pwndbg.chain.format(value)
  File "/GDB/pwndbg/pwndbg/chain.py", line 134, in format
    symbol = pwndbg.gdblib.symbol.get(link) or None
  File "/GDB/pwndbg/pwndbg/lib/cache.py", line 141, in decorator
    value = func(*a, **kw)
  File "/GDB/pwndbg/pwndbg/gdblib/symbol.py", line 97, in get
    result = gdb.execute("info symbol %#x" % int(address), to_string=True, from_tty=False)
gdb.error: That operation is not available on integers of more than 8 bytes.

```
PS: Pull requests are welcome

### Steps to reproduce
1. touch main.rs
2. insert this code into main.rs
```
fn main() {
    println!("Hello, world!");
}
```
3. rustc -g ./main.rs
4. gdb ./main
5. l
6. b 2
7. r

Gdb session history:
```
l
b 2
r
```

### My setup

```
Platform: Linux-5.15.0-101-generic-x86_64-with-glibc2.29
OS: Ubuntu 20.04.6 LTS
OS ABI: #111~20.04.1-Ubuntu SMP Mon Mar 11 15:44:43 UTC 2024
Architecture: x86_64
Endian: little
Charset: utf-8
Width: 211
Height: 51
Gdb:      14.2
Python:   3.8.10 (default, Nov 22 2023, 10:22:35)  [GCC 9.4.0]
Pwndbg:   2024.02.14 build: a75b1502
Capstone: 5.0.1280
Unicorn:  2.0.1
This GDB was configured as follows:
   configure --host=x86_64-pc-linux-gnu --target=x86_64-pc-linux-gnu
	     --with-auto-load-dir=$debugdir:$datadir/auto-load
	     --with-auto-load-safe-path=$debugdir:$datadir/auto-load
	     --with-expat
	     --with-gdb-datadir=/usr/local/share/gdb (relocatable)
	     --with-jit-reader-dir=/usr/local/lib/gdb (relocatable)
	     --without-libunwind-ia64
	     --without-lzma
	     --without-babeltrace
	     --without-intel-pt
	     --without-xxhash
	     --with-python=/usr
	     --with-python-libdir=/usr/lib
	     --without-debuginfod
	     --with-curses
	     --without-guile
	     --without-amd-dbgapi
	     --disable-source-highlight
	     --enable-threading
	     --enable-tui
	     --without-system-readline
	     --with-separate-debug-dir=/usr/local/lib/debug (relocatable)

("Relocatable" means the directory can be moved with the GDB installation
tree, and GDB will still find it.)
```
Such problems do not exist with C.
Maybe no support for rust, but before everything was fine.

dmur1:
fixes: https://github.com/pwndbg/pwndbg/issues/2080

### DIFF
--- a/pwndbg/gdblib/symbol.py
+++ b/pwndbg/gdblib/symbol.py
@@ -94,6 +94,7 @@ def get(address: int, gdb_only=False) -> str:
         return ""
 
     # This sucks, but there's not a GDB API for this.
+    # Workaround for a bug with Rust language, see #2094
     try:
         result = gdb.execute(f"info symbol 0x{address:x}", to_string=True, from_tty=False)
     except gdb.error:
--- a/pwndbg/gdblib/symbol.py
+++ b/pwndbg/gdblib/symbol.py
@@ -94,7 +94,10 @@ def get(address: int, gdb_only=False) -> str:
         return ""
 
     # This sucks, but there's not a GDB API for this.
-    result = gdb.execute("info symbol %#x" % int(address), to_string=True, from_tty=False)
+    try:
+        result = gdb.execute(f"info symbol 0x{address:x}", to_string=True, from_tty=False)
+    except gdb.error:
+        return ""
 
     if not gdb_only and result.startswith("No symbol"):
         address = int(address)


intrigus-lgtm:
I would not consider this a proper fix.
The problem is with gdb and this fix would (afaik) break functionality when debugging a linux kernel or any other system that reserves 0xfff...ffff -- 0x8000_0000_0000_0000 to the kernel.

dmur1:
> I would not consider this a proper fix. The problem is with gdb and this fix would (afaik) break functionality when debugging a linux kernel or any other system that reserves 0xfff...ffff -- 0x8000_0000_0000_0000 to the kernel.

agree that its not a proper fix, but i'm not sure how it breaks functionality when all it does it catch the exception that will be thrown currently?

intrigus-lgtm:
As far as I understand, you could have valid symbols at 0xffff_ffff_ffff_ffff.
Previously, they would get properly resolved, but now they don't.

It is possible to add an "i64" suffix so they should get properly resolved.

pwndbg/gdblib/symbol.py
```diff
@@ -94,7 +94,10 @@ def get(address: int, gdb_only=False) -> str:
         return ""
 
     # This sucks, but there's not a GDB API for this.
-    result = gdb.execute("info symbol %#x" % int(address), to_string=True, from_tty=False)
+    try:
```
disconnect3d, line 98:
```suggestion
    # Workaround for a bug with Rust language, see #2094
    try:
```

disconnect3d:
@intrigus-lgtm I think we can accept the interim solution of this and create another issue to fix this further in the future.

I don't think 0xfffff<big number> is a valid symbol address, so we should be fine here, heh.

disconnect3d (approved):
Going to merge later on after CI accepts this

intrigus-lgtm:
Why shouldn't 0xffff be a valid address?
If you work on kernel code, it will be mapped in that range (0x8000_0000_0000_0000-0xffff_ffff_ffff_ffff).

intrigus-lgtm:
You coule just as well change 64 to 63 in line 93 with the exact same effect.

dmur1:
@disconnect3d merge pls? :pray:  :smile:

disconnect3d:
@intrigus-lgtm I assume that if we work on kernel code, we won't have `set language rust` in the GDB session and so this won't fire off an exception.

PR #2096, opened 2024-03-26
Fix repeat command alias nearpc

### LINKED ISSUE #2095: Repeat doesn't work with nearpc alias = u
### Description

As @disconnect3d said on discord
pwndbg allows pressing enter after `nearpc` which repeats the command with next address but it doesn't work when you use aliases `pdisass` or `u`. 

### Steps to reproduce

1. Run `nearpc $pc`
2. Press enter to repeat it with last address, it works fine
3. Run `u $pc`
4. Press enter to repeat command, it doesn't work correctly

jetchirag:
Fixes #2095

Adds condition to check if repeat is set for any of the command in nearpc as the main command and aliases share different objects.

### DIFF
--- a/pwndbg/commands/nearpc.py
+++ b/pwndbg/commands/nearpc.py
@@ -28,7 +28,13 @@ def nearpc(pc=None, lines=None, emulate=False) -> None:
     """
     Disassemble near a specified address.
     """
-    print("\n".join(pwndbg.gdblib.nearpc.nearpc(pc, lines, emulate, nearpc.repeat)))
+    repeat = False
+    for command in pwndbg.commands.commands:
+        if command.__module__ == "pwndbg.commands.nearpc" and command.repeat == True:
+            repeat = True
+            break
+
+    print("\n".join(pwndbg.gdblib.nearpc.nearpc(pc, lines, emulate, repeat)))
 
 
 parser = argparse.ArgumentParser(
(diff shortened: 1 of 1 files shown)


disconnect3d:
Hmm I think that ideally, we want to do this properly so that not only nearpc works fine, but other commands that would have aliases and be repeated too.

Not sure if we have such commands now, but we may in the future.

gsingh93:
Will probably need to change something around here: https://github.com/pwndbg/pwndbg/blob/e7548c7f149ff785e06d23132a7fb09585e49bfa/pwndbg/commands/__init__.py#L184-L186

Ideally, we would only have a single command regardless of the number of aliases. But to do that I think we would need to hook GDB's command parsing look up that single command if any aliases were passed in there. I haven't looked into exactly how it would work, but I'm guessing the added complexity and bugs we'll introduce the future are probably not worth it.

disconnect3d:
Should we maybe register aliases as simply as gdb.execute with alias alias
= cmd? Afaik that should do the trick.

gsingh93:
Yes, I think that should work as long as `from_tty=True`.

disconnect3d:
@jetchirag can u update this accordingly?

jetchirag:
Yup, will do. Was busy with assignments the previous week.

jetchirag:
@disconnect3d 

`u` seems to already be registered as alias in gdb, rest are working fine.

```
for alias in self.aliases:
            alias_command = f"alias {alias} = {function.__name__}"
            gdb.execute(alias_command)
gdb.error: Alias already exists: u
```

disconnect3d:
Yeah it seems its fully controlled via GDB... its weird we can overwrite it as a command but not as an alias.

https://github.com/bminor/binutils-gdb/blob/3f3014cb00fabe1e340b277016a2f161ca297c71/gdb/infcmd.c#L3273-L3280 

Let's close this for now.

Issue #1458, opened 2022-12-23
Automatically rebase the kernel `vmlinux` with `kbase`


gsingh93:
There should be a flag you can pass to `kbase` that will reload the `vmlinux` and the debug symbols as the detected kASLR base address, so that all the commands that rely on debug symbols continue to work.

gsingh93:
Implemented in #2097

Issue #1836, opened 2023-07-25
Linux kernel: kallsyms command?


disconnect3d:
I am not sure if this is doable, but if someone debugs a kernel with KASLR it would be nice to be able to find and parse kallsyms from within Pwndbg and then either set symbols or at least be able to print/fetch them via some Pwndbg API.

Here is a project that shows how one could deal with this: https://github.com/pagabuc/kallsyms-extractor/ although I am not sure if this works with latest kernels.

disconnect3d:
Btw if we do this by looking for some strings in memory, in order to be efficient, we may need to e.g. search only for read-only mappings or read-write or something like that -- depending on where the searched string is to be expected in.

disconnect3d:
This may be useful here as well: https://github.com/marin-m/vmlinux-to-elf

chrf01:
may be useful: https://sourceware.org/pipermail/gdb-patches/2023-January/195589.html

disconnect3d:
Yes, we should use symbol addition API it when it lands in a GDB release
but even when it does, we will still want to support older GDB versions.
(Thx @mbrla0 for working on this API :))

chrf01:
I'll take a look at this

disconnect3d:
Related: https://github.com/pwndbg/pwndbg/pull/2142#issuecomment-2133864906

disconnect3d:
@fidgetingbits Given the recent work being finished, is there anything else we would need to be done here, or can I close this? :D

fidgetingbits:
I think it can be closed now ya. There was the one related discussion around using something like LIEF to generate the symbols inside a stripped template binary to be able to load them directly into gdb which would be nice, but should just be a separate issue I think.

PR #2097, opened 2024-03-26
improve kbase

### LINKED ISSUE #1949: kbase command isn't working
Only tested on Linux 6.6 and android14-5.15, `kbase` does not give any output, regardless of the KASLR status.

We should check if our QEMU tests cover this, and add it if necessary.

chrf01:
- Fixes #1949
- moved `kbase` logic to `pwndbg.gdblib.kernel`
- implemented a more reliable method to get the kernel text base using `IDT` entries and `VBAR` register
- Added #1458 
- update `gdb-pt-dump`

tested on the following targets:
- android13-5.10-lts-arm64
- android13-5.10-lts-x86_64
- android14-5.15-lts-arm64
- android14-5.15-lts-x86_64
- linux-5.10.210-arm64
- linux-5.10.210-x86_64
- linux-5.15.149-arm64
- linux-5.15.149-x86_64
- linux-6.1.79-arm64
- linux-6.1.79-x86_64
- linux-6.6.18-arm64
- linux-6.6.18-x86_64
- linux-6.7.1-x86_64


disconnect3d (commented):
Can we add comments or links here and there that would explain why certain values are used?

Like, where did we get the kernel magic values from, how do we know that aarch64 supervisor bit is the 48th bit or why do we subtract 0x10000 from base address if b != 0x4d?

chrf01:
I think we should move the kbase logic to pwndbg.gdblib.kernel.kbase for commands that rely on the kbase. This will make it easier for commands that search for something in the kernel image. For example, with the kallsyms #1836 command, instead of searching through all mappings, we could only search in the first read-only mapping of the kernel image.

pwndbg/commands/kbase.py
```diff
@@ -24,19 +24,24 @@ def kbase() -> None:
...
         return
 
     mappings = pwndbg.gdblib.vmmap.get()
     for mapping in mappings:
         # TODO: Check alignment
-        # TODO: Check if the supervisor bit is set for aarch64
+
+        # only search in kernel mappings:
+        # https://www.kernel.org/doc/html/v5.3/arm64/memory.html
+        if not mapping.vaddr & (0xFFFF << 48):
```
gsingh93:
```suggestion
        if mapping.vaddr & (0xFFFF << 48) == 0:
```

pwndbg/commands/kbase.py
```diff
@@ -48,6 +53,10 @@ def kbase() -> None:
                 )
             )
             continue
-        if b == magic:
-            print(M.success(f"Found virtual base address: {mapping.vaddr:#x}"))
+        if b in magic:
+            base = mapping.vaddr
+            if arch_name == "aarch64" and b != 0x4D:
+                # "MZ" header not found subtract 0x10000 from the address to get the kbase
+                base -= 0x10000
```
gsingh93: So the base of the kernel image isn't in vmmap, only the `.text` section is? That seems bizarre, is that expected or a bug with `vmmap`?

If we expect to find MZ here, we should add an `if` statement actually confirming it.

disconnect3d:
Can it be a result of KPTI and probably looking at this state at various
time may lead to different results?

pwndbg/commands/kbase.py
chrf01: It depends on the kernel version: In older arm64 kernels, the kernel base is marked as executable, and the MZ header is present. In newer arm64 kernels, the base of the kernel image appears in vmmap but isn't marked as executable, and it is empty. I don’t know if this is due to weird qemu behavior or if that's how the new arm64 kernel is mapped.

tests/qemu-tests/tests/system/test_commands_kernel.py
```diff
@@ -6,7 +6,8 @@
 
 
 def test_command_kbase():
-    pass  # TODO
+    res = gdb.execute("kbase", to_string=True)
+    assert len(res) > 0
```
gsingh93: If we ever decide to print an error or warning message, then this test will always pass. But I don't know a better way to test this, so it's fine for now.

gsingh93 (approved):

disconnect3d:
Can’t we check if the returned string is hexadecimal int aligned to 0x1000
or sth?

tests/qemu-tests/tests/system/test_commands_kernel.py
chrf01: We could verify that `found_kbase` matches the address returned by `pwndbg.gdblib.symbol.address("_text")`. However, this fix does not work for our tests involving 5-level paging, apparently because `gdb-pt-dump` lacks support for 5-level paging.

martinradev: > gdb-pt-dump lacks support for 5-level paging

What arch is this for?

martinradev: > However, this fix does not work for our tests involving 5-level paging, apparently because gdb-pt-dump lacks support for 5-level paging.

Support soon to be added - https://github.com/martinradev/gdb-pt-dump/issues/31

martinradev:
Would kbase be implementable via "pt -kaslr"?

chrf01:
> Would kbase be implementable via "pt -kaslr"?

No, I think searching for the first opcode isn't reliable enough to find the kernel base, especially if we want it to work on all kernel versions. That's why we use the `VBAR` register and `IDT` entries now.

patryk4815:
Should we use `pt_gdb.PageTableDumpGdbFrontend` or `pt.pt.PageTableDump`?

chrf01:
> Should we use `pt_gdb.PageTableDumpGdbFrontend` or `pt.pt.PageTableDump`?

Its easier to use `pt_gdb.PageTableDumpGdbFrontend`; The `pt.pt.PageTableDump` requires two parameters `machine_backend` and `arch_backend`. Therefore, we would need to do the `lazy_init()` stuff in pwndbg before we call `pt.pt.PageTableDump()`

chrf01:
The current method of obtaining the IDT base and size is hella ugly. I would prefer that the `regs idt` command outputs both the base and the size. Furthermore, accessing the base and the size should be possible through `pwndbg.gdblib.regs.idt.base` and `pwndbg.gdblib.regs.idt.size`, respectively. However, i believe we need to address issue #2108  first. 
I would appreciate any suggestions to improve this

pwndbg/commands/kbase.py
```diff
@@ -6,48 +6,32 @@
...
-                )
-            )
-            continue
-        if b == magic:
-            print(M.success(f"Found virtual base address: {mapping.vaddr:#x}"))
-            break
+    symbol_file = gdb.current_progspace().filename
+
+    if symbol_file:
+        gdb.execute("symbol-file", to_string=True)
```
disconnect3d: How does this work? Don't we need to pass in an argument?
```
pwndbg> help symbol-file
Load symbol table from executable file FILE.
Usage: symbol-file [-readnow | -readnever] [-o OFF] FILE
OFF is an optional offset which is added to each section address.
The `file' command can also load symbol tables, as well as setting the file
to execute.
The '-readnow' option will cause GDB to read the entire symbol file
immediately.  This makes the command slower, but may make future operations
faster.
The '-readnever' option will prevent GDB from reading the symbol file's
symbolic debug information.
```

pwndbg/gdblib/kernel/__init__.py
```diff
@@ -135,6 +138,58 @@ def is_kaslr_enabled() -> bool:
...
+
+        # only search in kernel mappings:
+        # https://www.kernel.org/doc/html/v5.3/arm64/memory.html
+        if mapping.vaddr & (0xFFFF << 48) == 0:
+            continue
+
+        if not mapping.execute:
+            continue
+
+        if mapping.vaddr <= address <= mapping.vaddr + mapping.memsz:
```
disconnect3d: Is this a bug? `<=` vs `<` at the end

Can also probably be:
```suggestion
        if address in mapping:
```

pwndbg/commands/kbase.py
chrf01: https://sourceware.org/gdb/current/onlinedocs/gdb.html/Files.html

> `symbol-file` with no arguments clears out GDB information on your program's symbol table.

pwndbg/commands/kbase.py
```diff
@@ -6,48 +6,32 @@
...
-    arch_name = pwndbg.gdblib.arch.name
-    if arch_name == "x86-64":
-        # First opcode, seems to be consistent
-        magic = 0x48
-    elif arch_name == "aarch64":
-        # First byte of "MZ" header
-        magic = 0x4D
-    else:
-        print(M.error(f"kbase does not support the {arch_name} architecture"))
+    print(M.success(f"Found virtual text base address: {hex(pwndbg.gdblib.kernel.kbase())}"))
```
gsingh93: Shouldn't we be checking for `None` here?

tests/qemu-tests/tests/system/test_gdblib_kernel.py
```diff
@@ -48,3 +48,11 @@ def test_gdblib_kernel_is_kaslr_enabled():
 def test_gdblib_kernel_nproc():
     # make sure no exception occurs
     pwndbg.gdblib.kernel.nproc()
+
+
+@pytest.mark.skipif(not pwndbg.gdblib.kernel.has_debug_syms(), reason="test requires debug symbols")
+def test_gdblib_kernel_kbase():
+    base = pwndbg.gdblib.kernel.kbase()
+    assert base == pwndbg.gdblib.symbol.address("_text") or base == pwndbg.gdblib.symbol.address(
```
gsingh93, line 58: A comment might help clarify what's going on here. I take it newer kernels (or maybe just arm64 newer kernels) have the kernel header at `_text` and executable code at `_stext`, but for older kernels (and non-arm64?) everything is at _text? Or is that wrong?

pwndbg/commands/kbase.py
gsingh93: This will probably break some user's workflows in some edge cases, but it's fine for now (unless there's some way to find a loaded kernel symbol file and just clear that).

pwndbg/gdblib/regs.py
```diff
@@ -36,6 +37,23 @@ def gdb_get_register(name: str) -> gdb.Value:
         return frame.read_register(name.upper())
 
 
+@pwndbg.gdblib.proc.OnlyWhenRunning
+def get_privileged_register(name: str) -> int:
```
gsingh93: nit: not a fan of the name, in some cases GDB already has "privileged" registers depending on the arch, this makes it seem likes this method needs to be used. `monitor info registers` is QEMU specific. Maybe just `get_qemu_register` or `get_register_from_qemu`?

pwndbg/gdblib/regs.py
```diff
@@ -174,6 +192,20 @@ def changed(self) -> List[str]:
                 delta.append(reg)
         return delta
 
+    @property
+    @pwndbg.lib.cache.cache_until("stop")
+    def idt(self) -> int:
+        if pwndbg.gdblib.qemu.is_qemu_kernel() and pwndbg.gdblib.arch.current == "x86-64":
```
gsingh93: Don't we have `OnlyWhenQemuKernel` and `OnlyWithArch` annotations that can be used instead of if-statements?

pwndbg/gdblib/kernel/__init__.py
```diff
@@ -135,6 +138,58 @@ def is_kaslr_enabled() -> bool:
...
+@pwndbg.lib.cache.cache_until("start")
+def kbase() -> int | None:
+    arch_name = pwndbg.gdblib.arch.name
+
+    address = 0
+
+    if arch_name == "x86-64":
+        address = get_idt_entries()[0].offset
+    elif arch_name == "aarch64":
+        address = pwndbg.gdblib.regs.vbar
```
gsingh93, line 150: I don't think this method is going to be reliable... `vbar` will likely change often if you're running bootloaders or hypervisors. I think the manual search from before is preferable, maybe with this as a fallback.

pwndbg/commands/kbase.py
gsingh93: Why do we need the `to_string=True`?

pwndbg/gdblib/kernel/__init__.py
chrf01: The problem with the manual search: The first opcode heavily depends on the kernel config for newer kernels without `_text`. For example, if `CONFIG_ARM64_BTI_KERNEL` is enabled the first instruction is a `bti c;` and if `CONFIG_DYNAMIC_FTRACE_WITH_CALL_OPS` is enabled the first few instructions look like this:
https://elixir.bootlin.com/linux/latest/source/arch/arm64/kernel/ftrace.c#L95
```
_stext+0: <ftrace_nop_ops> address
_stext+8: bti c;
_stext+12: mov x9, x30;
_stext+16: nop;
...
```
We could implement logic which tries to disassemble the bytes at `stext` to check if its an address or actual code and skip the first few config dependent instructions ?

But I think we should stick with the `vbar` as the main solution and add the manual search as a fallback, since it likely works for most use cases. Additionally, we could add checks to determine if the page is large enough (given that kernel text mappings are always quite large) and/or confirm it's the kernel base by searching for the Linux version string in the subsequent read-only page.

@gsingh93 
wdyt?

gsingh93: Actually, I think we can just go with your current approach of just `vbar`. The scenarios I was worried about (hypervisors and secure monitors) will be setting `VBAR_EL2` and `VBAR_EL3`, `VBAR_EL1` will probably remain unchanged. `VBAR_EL1` may be incorrect while a bootloader is running, before the kernel is loaded, but I wouldn't be running `kbase` at that point anyway.

So let's go with what you have, and if it breaks any of my workflows we'll just figure it out then :)

tests/qemu-tests/tests/system/test_gdblib_kernel.py
chrf01: yea newer arm and arm64 kernels reserve just [stext, _end) with `memblock_reserve(__pa_symbol(_stext), _end - _stext)`;

newer arm64 kernel: https://elixir.bootlin.com/linux/latest/source/arch/arm64/mm/init.c#L306
older arm64 kernel: https://elixir.bootlin.com/linux/v5.3.18/source/arch/arm64/mm/init.c#L399

pwndbg/gdblib/regs.py
chrf01: Using the `pwndbg.commands...` annotations leads to a circular import but we could implement the annotations in `pwndbg.gdblib.proc` like we did for `OnlyWhenRunning`

gsingh93: Works for me. I'm not a huge fan of the duplication of the annotations, but we can always attempt to merge them in the future (definitely don't try to deduplicate in this PR if adds any complexity, I don't want to delay merging it for that).

disconnect3d (approved):
@gsingh93 lgtm; shall we merge this?

pwndbg/gdblib/proc.py
```diff
@@ -167,6 +170,28 @@ def wrapper(*a: Any, **kw: Any) -> T:
 
         return wrapper
 
+    def OnlyWithArch(
```
disconnect3d, line 173: Do we need it here if we have https://github.com/pwndbg/pwndbg/blob/385e63f638a9f3a5b492e047a30f8653530305d5/pwndbg/commands/__init__.py#L300 ?

chrf01: Importing it from `pwndbg.commands` leads to a circular import, the same applies to `OnlyWhenRunning` and `OnlyWhenQemuKernel`; we will try to deduplicate it in a future PR: https://github.com/pwndbg/pwndbg/pull/2097#discussion_r1561378907

disconnect3d:
Ok

chrf01:
ping @gsingh93

gsingh93:
Thanks!

PR #2098, opened 2024-03-26
Added next flag to malloc_chunk

### LINKED ISSUE #2086: Next flag for malloc_chunk
Opening this FR related to #1059. I don't think this was later implemented so I want to pick this up to get more familiar with codebase.

> Why stop at the next chunk? The chunk after that can be just as important, as can the previous chunk. Maybe a flag that lets you print a user-defined number of next/previous/surrounding chunks would be useful.

I was however wondering on the way to go about flags. Should we keep 3 seperate flags similar to grep (A, B for next, prev and C for surrounding) or just the next X chunks e.g. `--next 3`?

jetchirag:
Fixes #2086

- [ ] Add test

### DIFF
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -510,7 +510,7 @@ def malloc_chunk(addr, fake=False, verbose=False, simple=False, next=0) -> None:
     print(" | ".join(headers_to_print) + "\n" + out_fields)
 
     if next:
-        print(C.banner(f"Next {next} chunk(s)"))
+        print(C.banner(f"Next {next} chunk(s):"))
         for _ in range(next):
             chunk = chunk.next_chunk()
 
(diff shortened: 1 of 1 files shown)


CptGibbon (changes requested):
Looks great, thank you very much for this PR 🙏

Screenshot 2024-03-27 at 11 14 56

My personal opinion: it could be useful to print _both_ the chunk at the address passed and its succeeding chunks, that way you don't have to run the command twice to see all that information.

@gsingh93 was that what you had in mind originally in #1059?

If so, the loop for printing next chunks could perhaps be moved after the initial printing code.

jetchirag:
@CptGibbon That makes sense and how I originally wrote it but original PR seemed to hint at just next chunks. Let me know if I should move the loop to end.

disconnect3d:
Yeah lets show both the chunk and the next chunks

CptGibbon:
I'll review this again next week 👍

CptGibbon (approved):
Looks great 👍
Thanks for this PR, I'm happy for it to get merged if there are no objections.

Screenshot 2024-04-01 at 21 03 41

PR #2100, opened 2024-03-27
Fix flake8 warnings

### LINKED ISSUE #1183: Stop ignoring F841 flake8 warnings (unused variables)
When we added `flake8`, we had to ignore some existing warnings because it was too much work at the time to fix them, and we wanted to get `flake8` on the CI so future errors would be prevented. We need to go back and fix some of these so we can start running them on the CI, like F841: local variable name is assigned to but never used.

jetchirag:
Fixes #1183 for F841


pwndbg/disasm/x86.py
```diff
@@ -62,7 +61,7 @@ def memory(self, instruction, op):
     def memory_sz(self, instruction, op):
         arith = False
         segment = op.mem.segment
-        disp = op.value.mem.disp
+        # disp = op.value.mem.disp
```
disconnect3d, line 64: Maybe that should be `disp = op.mem.disp` and used later on?

This should probably be:
```py

        if disp != 0:
            if arith:
                sz += " - " if disp < 0 else " + "
            sz += f"{abs(disp):#x}"
```

I also wonder where is it displayed, because maybe "+" and space afterwards is irrelevant? If it were, we could just do:
```py
if disp != 0:
    sz += f"{disp:#x}"
```
but feel free to not investigate this in this PR

disconnect3d: Oh and when we are at it, we can change the:
```py
        sz = f"[{sz}]"
        return sz
```
to just `return f"[{sz}]"`

pwndbg/disasm/x86.py
```diff
@@ -133,7 +132,7 @@ def condition(self, instruction):
 
         cf = efl & (1 << 0)
         pf = efl & (1 << 2)
-        af = efl & (1 << 4)
+        # af = efl & (1 << 4)
```
disconnect3d, line 135: Interesting that no jmp/conditional move depends on the "auxiliary carry flag". Maybe we miss some jmp/cmov below, idk. Feel free to ignore this, just loud thinking.

(cc: @OBarronCS FYI)

tests/qemu-tests/tests/system/test_commands_kernel.py
```diff
@@ -11,11 +11,13 @@ def test_command_kbase():
 
 def test_command_kchecksec():
     res = gdb.execute("kchecksec", to_string=True)
+    assert res is not None  # for F841 warning
```
disconnect3d, line 14: With `to_string=True` I believe it will always be _some string_, like an empty string.

I gues this is probably fine for now until we change this to some specific string assertion (cc/fyi @gsingh93)

disconnect3d:
All this changes make sense, so I am merging this. A future PR[s] can addresss the comments here.

pwndbg/gdblib/elf.py
```diff
@@ -415,7 +415,7 @@ def map_inner(ei_class, ehdr, objfile: str) -> Tuple[pwndbg.lib.memory.Page, ...
         vaddr = int(phdr.p_vaddr)
         offset = int(phdr.p_offset)
         flags = int(phdr.p_flags)
-        ptype = int(phdr.p_type)
+        # ptype = int(phdr.p_type)
```
gsingh93, line 418: Why is this commented out instead of removed?

jetchirag: I just left it out because I thought it might be useful in future. I was checking rest of flake8 warnings, should I remove this and other commented out lines in next PR?

tests/qemu-tests/tests/system/test_commands_kernel.py
gsingh93: Yes, `to_string=True` will always return some string, and @peace-maker is even adding that to the gdbstubs.

I think at least `assert res != ''`  would be better here, until we implement a proper test.

pwndbg/gdblib/elf.py
gsingh93: I think so, yes. If we ever need them we'll just add them back in, no need to keep commented code around.

tests/qemu-tests/tests/system/test_commands_kernel.py
peace-maker: The gdb.execute return type is merged already so the gdb-stubs can be updated to make the None check unnecessary.

PR #2101, opened 2024-03-28
WIP: Fix Flake8 warnings

jetchirag:
Continues #2100


jetchirag:
I don't know why there were duplicates of 2 tests, oversight or they served a purpose? I assumed the former and have removed them for now.

disconnect3d:
I don't think they served any purpose. Seems like oversight/copy-paste bug.

jetchirag:
Should we continue to ignore these or not?

E731 Do not assign a `lambda` expression, use a `def` (14 errs)
E501 Line too long (266 errs)
F405 may be undefined, or defined from star imports (126 errors) - easy to fix (capstone library imported with * so prepend `capstone.` everywhere)

gsingh93:
Let's leave A003 and W505 in the ignores, we want to continue ignoring those even if there aren't any current violations.

We should eventually fix the F821 violations as well.

I think E731 could be good to fix, but I'd like @disconnect3d's opinion (see https://docs.astral.sh/ruff/rules/lambda-assignment/)

I attempted to fix F405 in the past, but I dropped it because I wasn't convinced I wasn't breaking anything, see https://github.com/pwndbg/pwndbg/pull/1262. It would be good to do this eventually, but we should understand the capstone code a bit better.

disconnect3d:
> Let's leave A003 and W505 in the ignores, we want to continue ignoring those even if there aren't any current violations.
> 
> We should eventually fix the F821 violations as well.
> 
> I think E731 could be good to fix, but I'd like @disconnect3d's opinion (see https://docs.astral.sh/ruff/rules/lambda-assignment/)
> 
> I attempted to fix F405 in the past, but I dropped it because I wasn't convinced I wasn't breaking anything, see #1262. It would be good to do this eventually, but we should understand the capstone code a bit better.

+1 in general; the lambda stuff - I don't really have opinion, both ways are fine.

Issue #2084, opened 2024-03-24
search --asm: set breakpoint on found instructions


disconnect3d:
We now have `search --asm "<instructions>"` command and it would be nice if it can set breakpoint on found addresses.

Shall this be another option/flag like `-b|--break` or `-b|--set-breakpoint`? Or maybe a `--asmbp` option?

Or a separate command? Probably not. To be discussed.

Fwiw while this makes sense for instructions, such feature may also make sense for normal findings, where we could set watchpoints on the found addresses. Though, if there are lots and lots of results it probably doesn't make that much sense...

So I'd be up for doing this now just for asm and thinking if we really need it for normal findings later on.

diivi:
I'll take a look at this!

TheLazron:
I'd like to try this

TheLazron:
Can I get a review on this pr please, thanks

disconnect3d:
This has been implemented in #2103 as `--asmbp` flag. Thanks!

PR #2103, opened 2024-03-28
search --asm: set breakpoint on found instructions

TheLazron:
#2084
`--asmbp` flag to breakpoint found --asm instructions
ss

### DIFF
--- a/pwndbg/commands/search.py
+++ b/pwndbg/commands/search.py
@@ -307,7 +307,7 @@ def search(
             saved.add(address)
         if asmbp:
             # set breakpoint on the instruction
-            gdb.Breakpoint("*%#x" % address, temporary=True)
+            gdb.Breakpoint("*%#x" % address, temporary=False)
 
         if not trunc_out or i < 20:
             print_search_hit(address)
--- a/pwndbg/commands/search.py
+++ b/pwndbg/commands/search.py
@@ -248,7 +248,7 @@ def search(
         value = value.encode()
         value += b"\x00"
 
-    elif type == "asm":
+    elif type == "asm" or asmbp:
         bits_for_arch = pwnlib.context.context.architectures.get(arch, {}).get("bits")
         value = pwnlib.asm.asm(value, arch=arch, bits=bits_for_arch)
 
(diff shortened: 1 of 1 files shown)


pwndbg/commands/search.py
```diff
@@ -299,6 +305,11 @@ def search(
     ):
         if save:
             saved.add(address)
+        if type == "asm" and asmbp:
+            ins = pwndbg.disasm.one(address)
```
disconnect3d: Why do we need to disasm it?

disconnect3d (commented):
Can we do it so it can be specified without `--asm`? Just `--asmbp`?

Ideally, it would be nice if it worked with specifying both `--asm --asmbp` and with specifying only `--asmbp` flag.

(So that if someone hits 'arrow up' and just adds `--asmbp` it should work; but if someone just specifies only `--asmbp` it should also work)

pwndbg/commands/search.py
TheLazron: oh sorry I overlooked the redundancy while making changes

TheLazron:
> Ideally, it would be nice if it worked with specifying both --asm --asmbp and with specifying only --asmbp flag.

`--asmbp` now can be used as an independent flag to search for asm instructions and then setting a breakpoint and also works with `asm`

changes

pwndbg/commands/search.py
```diff
@@ -299,6 +305,9 @@ def search(
     ):
         if save:
             saved.add(address)
+        if asmbp:
+            # set breakpoint on the instruction
+            gdb.Breakpoint("*%#x" % address, temporary=True)
```
disconnect3d:
```suggestion
            gdb.Breakpoint("*%#x" % address, temporary=False)
```

disconnect3d (approved):
I think we should set up a standard breakpoint and let user delete it if they do not want it.

I made a change for this. Going to merge when CI passes.

PS: If u can, pls send another PR implementing tests for this :)

TheLazron:
> PS: If u can, pls send another PR implementing tests for this :)

Sure! on it

PR #2114, opened 2024-04-12
Added tests for --asm search

TheLazron:
Related #2103 

Added tests for commands to search `--asm` instructions and setting breakpoints on found instructions
new_test


tests/gdb-tests/tests/test_command_search.py
```diff
@@ -182,3 +182,37 @@ def test_command_search_rwx(start_binary):
...
+
+    gdb.execute("break break_here")
+    gdb.execute("run")
+
+    result_str = gdb.execute('search --asm "add rax, rdx" search_memory', to_string=True)
+    result_count = 0
+    for line in result_str.split("\n"):
+        if line.startswith("search_memory"):
+            result_count += 1
+    assert result_count == 2
```
disconnect3d, line 201: Can we assert th whole expected output instead of checking if line starts with sth etc?
We could use regex for addresses etc.

disconnect3d:
Those tests are better than nothing, but pls improve them if possible.

tests/gdb-tests/tests/test_command_search.py
TheLazron: Sorry for delayed response
Will an assertion using this regex be fine?
```
instruction_pattern = re.compile(r"search_memory\.out [A-Za-z0-9]+ add rax, rdx")
```
Can I still make my changes here?

PR #2120, opened 2024-04-19
Additional type hints

peace-maker:
Add type annotations to the heap code and bunch of other code that's used by the heap.
Activate `vermin --eval-annotations` to catch invalid type hints for Python 3.8.

Use `typing_extensions.ParamSpec` to avoid hiding function arguments through decorators.

There are a couple of instances where `pwndbg.heap.current` is expected to be `GlibcMemoryAllocator` in common code. I've worked around this by adding explicit `isinstance` checks, but this probably requires more work if other allocators are to be supported.


peace-maker:
```
 >           results[name] = gdb.execute(f"malloc_chunk {name}_chunk", to_string=True).splitlines()
E           gdb.error: Error occurred in Python: unsupported operand type(s) for |: '_GenericAlias' and 'NoneType'
```

This appears to be a problem with gdb's type stubs from typeshed on < Python 3.10. I thought `from __future__ import annotations` works since 3.8. Hm

pwndbg/commands/cymbol.py
```diff
@@ -146,7 +155,7 @@ def add_custom_structure(custom_structure_name: str) -> None:
 
 
 @OnlyWhenStructFileExists
-def edit_custom_structure(custom_structure_name: str, custom_structure_path: str) -> None:
```
gsingh93, line 149: Why the default argument initialization?

pwndbg/commands/leakfind.py
```diff
@@ -142,7 +144,7 @@ def leakfind(
     # We need to store both so that we can nicely create our leak chain.
     visited_map = {}
     visited_set = {int(address)}
-    address_queue: queue.Queue[int] = queue.Queue()
+    address_queue = queue.Queue()  # type: ignore[var-annotated]
```
gsingh93: Why is the ignore needed?

gsingh93 (commented):
This is great, thanks for this :)

I did a quick skim, but didn't get through it all. Despite the size I don't think this will take too long to finish reviewing, as type change reviews are pretty easy.

That being said, I want to make sure https://github.com/pwndbg/pwndbg/pull/2001 gets merged first so they don't need to do any additional rebase work. After it's merged, once this gets rebased I can finish the review.

pwndbg/commands/leakfind.py
peace-maker: vermin complained and [Queue is only generic](https://bugs.python.org/issue33315) in Python 3.9+:
```
!2, 3.9      pwndbg/pwndbg/commands/leakfind.py
  L147: builtin generic type annotation (queue.Queue[..]) requires !2, 3.9
```

Re-reading that python issue it seems it should work though with annotations from `__future__` or putting the type annotation in quotes.

pwndbg/commands/cymbol.py
peace-maker: This implementation is weird with having the `OnlyWhenStructFileExists` decorator filling in the second argument transparently for the decorated function to use if it was called without it. The code mostly called the functions without explicitly passing a `custom_structure_path` argument and relying on the decorator to fill it in, but mypy complained about missing required arguments.

```
pwndbg/commands/cymbol.py: note: In function "edit_custom_structure":
pwndbg/commands/cymbol.py:187: error: Missing positional argument "custom_structure_path" in call to "__call__" of
"_OnlyWhenStructFileExists"  [call-arg]
        load_custom_structure(custom_structure_name)
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

I guess another way to solve this is to explictly pass `""` as second argument when calling the functions instead of specifying a default argument. I can go that route too if you'd like.

gsingh93:
@peace-maker needs a rebase

peace-maker:
@gsingh93 done

pwndbg/gdblib/memory.py
```diff
@@ -137,6 +137,8 @@ def is_readable_address(address: int) -> bool:
         :class:`bool`: Whether the address is readable.
     """
     # We use vmmap to check before `peek()` because accessing memory for embedded targets might be slow and expensive.
+    import pwndbg.gdblib.vmmap
```
gsingh93: I think we can move this import to the top-level

pwndbg/heap/structs.py
```diff
@@ -88,9 +92,13 @@ class c_size_t(SIZE_T):
...
 if pwndbg.gdblib.arch.endian == "little":
-    C2GDB_MAPPING = {k.__ctype_le__: v for k, v in C2GDB_MAPPING.items()}
+    C2GDB_MAPPING: Dict[Type[ctypes.c_char], gdb.Type] = {  # type: ignore[no-redef]
+        k.__ctype_le__: v for k, v in C2GDB_MAPPING.items()
+    }
 else:
-    C2GDB_MAPPING = {k.__ctype_be__: v for k, v in C2GDB_MAPPING.items()}
+    C2GDB_MAPPING: Dict[Type[ctypes.c_char], gdb.Type] = {  # type: ignore[no-redef]
+        k.__ctype_be__: v for k, v in C2GDB_MAPPING.items()
+    }
```
gsingh93, line 101: Just thinking aloud here, we could get rid of the no-redef ignore by moving the `if` into the dictionary comprehension, but it might look a bit messy...

pwndbg/heap/structs.py
```diff
@@ -123,7 +131,7 @@ def __init__(
 
 class CStruct2GDB:
     code = gdb.TYPE_CODE_STRUCT
-    _c_struct: pwndbg.gdblib.ctypes.Structure = None
```
gsingh93, line 126: I have not looked at this code in some time, but I think it was intentional to make this a class variable that gets overridden in child classes. I don't know what unintended effects there might be by changing that. cc: @CptGibbon in case you have thoughts on this.

pwndbg/lib/gcc.py
```diff
@@ -42,7 +44,8 @@ def _which_binutils(util: str, arch: Arch, **kwargs: Any) -> str | None:
 
     # Fix up binjitsu vs Debian triplet naming, and account
     # for 'thumb' being its own binjitsu architecture.
-    arches: List[str] = [arch_name] + {
+    arches: List[str | None] = []
+    arches += [arch_name] + {
```
gsingh93: Why did we need to split this variable definition into two parts?

pwndbg/profiling.py
```diff
@@ -3,7 +3,7 @@
 import cProfile
 import time
 
-profiler: Profiler | None = None
```
gsingh93: This is a global variable that is used from scripts in the `profiling` folder. I think it should be `None` until `init` is called.

pwndbg/gdblib/ctypes.py
```diff
@@ -17,12 +17,13 @@
 import pwndbg.gdblib.events
 
 module = sys.modules[__name__]
-Structure = ctypes.LittleEndianStructure  # default Structure type
```
gsingh93: I'm not sure I understand this change

pwndbg/gdblib/abi.py
```diff
@@ -1,18 +1,19 @@
...
 
 import gdb
+from typing_extensions import ParamSpec
 
 import pwndbg.color.message as M
 
-T = TypeVar("T")
+_P = ParamSpec("_P")
+_D = TypeVar("_D")
+_T = TypeVar("_T")
```
gsingh93: nit: I see some places we use `T` and some places we use `_T`,  let's be consistent. I personally prefer `T`, and I think that's more common in the rest of the code.

pwndbg/gdblib/proc.py
```diff
@@ -122,21 +128,27 @@ def binary_base_addr(self) -> int:
 
     @property
     @pwndbg.lib.cache.cache_until("start", "stop")
-    def binary_vmmap(self) -> tuple[pwndbg.lib.memory.Page, ...]:
+    def binary_vmmap(self) -> Tuple[pwndbg.lib.memory.Page, ...]:
+        import pwndbg.gdblib.vmmap
```
gsingh93, line 133: This and the other imports can be moved to the top-level.

pwndbg/gdblib/elf.py
```diff
@@ -138,20 +152,21 @@ def get_elf_info_rebased(filepath: str, vaddr: int) -> ELFInfo:
     # silently ignores "wrong" vaddr supplied for non-PIE ELF
     load = vaddr if raw_info.is_pic else 0
     headers = dict(raw_info.header)
-    headers["e_entry"] += load
+    headers["e_entry"] += load  # type: ignore[operator]
```
gsingh93, line 155: Could you explain why we need all the `# type: ignore[operator]` comments in this file? Just to make this easier for me to review.

pwndbg/gdblib/elf.py
```diff
@@ -327,17 +343,17 @@ def get_ehdr(pointer: int):
     ei_class = pwndbg.gdblib.memory.byte(base + 4)
 
     # Find out where the section headers start
-    Elfhdr = read(Ehdr, base)
+    Elfhdr: Elf32_Ehdr | Elf64_Ehdr | None = read(Ehdr, base)  # type: ignore[type-var]
```
gsingh93, line 346: Why do we need the `ignore[type-var]`?

pwndbg/lib/regs.py
```diff
@@ -16,37 +18,37 @@
...
-    retval: str | None = None
+    retval: str | None
 
     #: Common registers which should be displayed in the register context
-    common: list[str] = []
+    common: List[str] = []
 
     #: All valid registers
-    all: set[str] | None = None
+    all: Set[str]
```
gsingh93, line 51: I'm actually wondering if we need any of these if they don't need to be class variables... I think the typing in `__init__` should be sufficient, and the comments here aren't very useful. Maybe we could just delete them?

pwndbg/heap/__init__.py
```diff
@@ -6,11 +6,14 @@
 import gdb
 
 import pwndbg.gdblib.config
+import pwndbg.gdblib.events
+import pwndbg.gdblib.proc
 import pwndbg.gdblib.symbol
 import pwndbg.heap.heap
 from pwndbg.color import message
+from pwndbg.gdblib.config import config
```
gsingh93: Note that we do `from pwndbg.gdblib.config import config` in `pwndbg/gdblib/__init__.py`. Maybe there's a way to get the types to check without modifying the code here? Like adding an additional type annotation for `config` in `pwndbg/gdblib/__init__.py`?

pwndbg/heap/structs.py
CptGibbon: @lebr0nli I shall defer to you old chap

lebr0nli: I think my original type hint was wrong, thanks @peace-maker for fixing it here!
Since `_c_struct` is a `type` instead of an instance of `pwndbg.gdblib.ctypes.Structure`, also `ctypes.Structure` is the parent class of `pwndbg.gdblib.ctypes.Structure`(`pwndbg.gdblib.ctypes.Structure` should be `ctypes.LittleEndianStructure` or `ctypes.BigEndianStructure`), the way it's written in this commit makes more sense to me.

> I have not looked at this code in some time, but I think it was intentional to make this a class variable that gets overridden in child classes.

Yep, it's intentional.

> I don't know what unintended effects there might be by changing that.

Removing the `= None` here should not cause unintended effects I think. (At least all the tests related to this code passed :p)

pwndbg/gdblib/ctypes.py
peace-maker: At one point pyright seemed to have had a problem with redefining a type alias. It's fine as it was as it seems

pwndbg/gdblib/elf.py
peace-maker: The header is converted to a `Dict[str, int | str]` and the typer complains about incompatible types `Unsupported operand types for + ("str" and "int")`. We'd need to properly type every elf header field since some are integers and some are strings instead of putting the in a generic dict. But I didn't want to refactor the code too much.

pwndbg/gdblib/proc.py
peace-maker: `pwndbg.gdblib.vmmap` and `pwndbg.gdblib.elf` cause a circular import and cannot be moved to the toplevel. I don't want to do too much refactoring in this PR so I'd rather keep them deferred.

pwndbg/lib/gcc.py
peace-maker: Hm, there were errors with pyright infering the lhs type to an incompatible one to the explicit type and flagging the assignment. The problem is gone now. I'll make sure to use mypy all the way next time.

peace-maker (commented):
Thanks for the review, I've changed most of it. I'll try to go over the other comments when I have time.

pwndbg/heap/ptmalloc.py
```diff
@@ -51,44 +73,51 @@ class BinType(str, Enum):
...
-    def valid_fields(self):
+    def valid_fields(self) -> List[str]:
         if self in [BinType.FAST, BinType.TCACHE]:
             return ["fd"]
         elif self in [BinType.SMALL, BinType.UNSORTED]:
             return ["fd", "bk"]
         elif self == BinType.LARGE:
             return ["fd", "bk", "fd_nextsize", "bk_nextsize"]
+        else:  # BinType.NOT_IN_BIN
+            return []
```
gsingh93, line 88: @CptGibbon @lebr0nli are you fine with this behavior of returning an empty list for `BinType.NOT_IN_BIN` or would you prefer the old behavior of `None`?

pwndbg/heap/ptmalloc.py
```diff
@@ -51,44 +73,51 @@ class BinType(str, Enum):
...
-    def contains_chunk(self, chunk) -> bool:
+    def contains_chunk(self, chunk: int) -> bool:
         return chunk in self.fd_chain
 
     @staticmethod
-    def size_to_display_name(size):
+    def size_to_display_name(size: int | str) -> str:
         if size == "all":
-            return size
+            return size  # type: ignore[return-value]
```
gsingh93: I think this is a case where a type cast is better than a type ignore.

pwndbg/heap/ptmalloc.py
```diff
@@ -24,11 +34,23 @@
...
+import pwndbg.heap.heap
+import pwndbg.lib.cache
+import pwndbg.lib.memory
 import pwndbg.search
 from pwndbg.color import message
 from pwndbg.color.memory import c as M
 from pwndbg.constants import ptmalloc
-from pwndbg.heap import heap_chain_limit
+
+if typing.TYPE_CHECKING:
```
gsingh93, line 50: I don't think it's obvious what's going on here, can you add a comment explaining it?

pwndbg/commands/heap.py
```diff
@@ -23,37 +32,44 @@
...
     # the rest of the code can deal with uniform names.
+    assert isinstance(pwndbg.heap.current, GlibcMemoryAllocator)
+    assert pwndbg.heap.current.malloc_chunk is not None
     renames = {
         "mchunk_size": "size",
         "mchunk_prev_size": "prev_size",
     }
-    if isinstance(pwndbg.heap.current, DebugSymsHeap):
-        val = pwndbg.gdblib.typeinfo.read_gdbvalue("struct malloc_chunk", addr)
+    if isinstance(pwndbg.heap.current.malloc_chunk, gdb.Type):
```
gsingh93: This seems like different behavior to me, but I'm not very familiar with this code. @CptGibbon or @lebr0nli, can you verify this is correct?

pwndbg/commands/heap.py
```diff
@@ -23,37 +32,44 @@
...
     else:
         val = pwndbg.heap.current.malloc_chunk(addr)
-    return {renames.get(key, key): int(val[key]) for key in val.type.keys()}
+    value_keys: List[str] = val.type.keys()
+    return {renames.get(key, key): int(val[key]) for key in value_keys}
 
 
-def format_bin(bins: Bins, verbose=False, offset=None):
+def format_bin(bins: Bins, verbose: bool = False, offset: int | None = None) -> List[str]:
+    assert isinstance(pwndbg.heap.current, GlibcMemoryAllocator)
```
gsingh93, line 60: I think this is fine for this PR, but we should really think about how we can remove all these `assert isinstance(pwndbg.heap.current, GlibcMemoryAllocator)` checks in this PR. Maybe this will happen naturally with the jemalloc GSoC project implementing a new allocator, not sure.

pwndbg/gdblib/elf.py
gsingh93: Thanks. Maybe this file could be refactored in the future if pyelftools ever gets type stubs (https://github.com/python/typeshed/pull/11262).

gsingh93 (commented):
This pretty much looks good to me minus a few nits. @CptGibbon and @lebr0nli, you could you just verify the comments I tagged you in look good? If you would also like more time to skim through `ptmalloc.py` and `heap.py`, let me know. If not, that's fine, I'll go ahead and merge this once the other changes are made.

gsingh93:
Also needs another rebase, but should just be a small one.

pwndbg/heap/ptmalloc.py
peace-maker: It's a dance to avoid importing `pwndbg.heap.structs` since it's imported dynamically when first resolving `HeuristicHeap.struct_module`. I'll add a comment too.

pwndbg/commands/heap.py
peace-maker: This is a two-in-one check. `pwndbg.heap.current.malloc_chunk` is only of type `gdb.Type` if the allocator is `DebugSymsHeap`. Otherwise it would be `pwndbg.heap.structs.CStruct2GDB`. I can explicitly keep the previous check for `DebugSymsHeap` too for clearity.

peace-maker: Ah, this is from an older iteration of the ptmalloc typings where the properties like `malloc_chunk` had a union return type. Since they have the correct one now the old check for DebugSymsHeap is enough. I'll revert, good catch.

peace-maker: Or were you commenting on the change of 
```python
val = pwndbg.gdblib.typeinfo.read_gdbvalue("struct malloc_chunk", addr)
```
to
```python
val = pwndbg.gdblib.memory.poi(pwndbg.heap.current.malloc_chunk, addr)
```

`memory.poi` was used in ptmalloc and it seemed better to not resolve the type again so I went with the `poi` way. Both lines do the same thing though.

https://github.com/pwndbg/pwndbg/blob/92640c7131409b01fb096b9ff796b63f3ddc8520/pwndbg/heap/ptmalloc.py#L157

disconnect3d:
@gsingh93 go merge?:)

gsingh93:
LGTM, merging.

PR #2128, opened 2024-04-25
Upgrade all dependencies

gsingh93:
1) Instead of pinning dependencies to an exact version in `pyproject.toml`, set the versions as minimums (with a preceding `^`). There's no need to pin exact dependency versions with `poetry` because they're already pinned in `poetry.lock`.
2) The pinned versions in `poetry.lock` can be upgraded with `poetry update`. This won't modify `pyproject.toml`
3) It is not necessary to bump versions in `pyproject.toml` if we follow the steps above, but it can be nice as it's easier to see what version we depend on without digging into `poetry.lock`. I used https://github.com/MousaZeidBaker/poetry-plugin-up to do this with `poetry up`

dev dependencies were left as is for now, and `capstone` and `ropgadget` were not upgraded, as they break some `nearpc` tests.

### DIFF
--- a/nix/pwndbg.nix
+++ b/nix/pwndbg.nix
@@ -26,6 +26,10 @@ let
         preferWheel = true;
       };
 
+      unix-ar = super.unix-ar.overridePythonAttrs (old: {
+        nativeBuildInputs = (old.nativeBuildInputs or []) ++ [self.setuptools];
+      });
+
       pt = super.pt.overridePythonAttrs (old: {
         buildInputs = (old.buildInputs or [ ]) ++ [ super.poetry-core ];
       });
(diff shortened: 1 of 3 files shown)


gsingh93:
@patryk4815 Can you look at the lock file CI failure?

patryk4815:
@gsingh93 https://github.com/gsingh93/pwndbg/pull/6

gsingh93:
@patryk4815 I merged your PR but it seems like the check is still failing. Is there anything else I need to do?

disconnect3d:
@gsingh93 I spoke with @patryk4815 and the issue you trigger was fixed in https://github.com/NixOS/nixpkgs/pull/307380 but it hasn't been merged into nixos-unstable which we base upon:

image

It will land there in <24h afaik.

If the upgraded deps work for you, we can hit merge here, or, if you prefer to wait and see, we can also do that.

PR #2041, opened 2024-02-29
Add ipython as a dependency

gsingh93:

PR #2137, opened 2024-04-28
Upgrade dev dependencies

gsingh93:
- Instead of pinning dev dependency versions, set them as minimums
- Upgrade all dev dependencies in poetry.lock to latest compatible versions (`poetry up --with dev`)

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -298,63 +298,63 @@ pygments = "*"
 
 [[package]]
 name = "coverage"
-version = "7.4.3"
+version = "7.5.0"
 description = "Code coverage measurement for Python"
 optional = false
 python-versions = ">=3.8"
 files = [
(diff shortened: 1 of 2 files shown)


disconnect3d:
Awesome! Btw why do we have `ipython = "8.12.3"` in normal deps and not dev deps?

Shall we move it to dev deps?

disconnect3d:
I think this may be for `ipi` but since lots of users do not use it maybe we should remove it..

gsingh93:
It's for `ipi`. I consider that a regular user feature (although devs heavily use it), and so it should be a regular dep. BTW I asked you about making this a regular dep on Discord a few months ago and you said yes 😄 Change was made in https://github.com/pwndbg/pwndbg/pull/2041

disconnect3d:
Heh, I recall it now. Let’s leave it as it is.

It’s just a friend of mine asked me recently why we need it and I thought
he run pip install -e for dev mode deps… but yeah he did it this way as he
is running gentoo.

PR #2140, opened 2024-04-28
Add pending-author-response workflow

gsingh93:
According to https://github.com/siegerts/pending-author-response, we can tag an issue with a label (currently set to "pending-response"), and if the issue author responds the label is automatically removed.

I don't think there's a good way for me to test if this actually works without committing it, or setting up github secrets in my own repo...

### DIFF
--- a/.github/workflows/pending-author-response.yml
+++ b/.github/workflows/pending-author-response.yml
@@ -1,0 +1,14 @@
+name: Pending Author Response Label
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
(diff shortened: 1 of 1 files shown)


gsingh93:
Ping @disconnect3d, do you still want to do this?

disconnect3d:
@gsingh93 I'm more towards not adding bot workflows if we don't really need to, but we can try doing this and see how it works

disconnect3d:
@gsingh93 lets see if it works right away :P

Issue #1485, opened 2023-01-02
Handle kASLR better when checking for kernel debug symbols


gsingh93:
Currently if a user tries to use any of the kernel commands that require debug symbols and kASLR is enabled, they'll get a confusing error: `gdb.MemoryError: Cannot access memory at address 0xffffffc0111d0ef0`.

Ideally we could check for kASLR in the debug symbols check, but our kASLR check itself (`is_kaslr_enabled`) requires debug symbols.

We should either probably make `is_kaslr_enabled` not depend on debug symbols and then use it in the debug symbols check to print a warning and not run the command. The alternative is just to catch the gdb exception and print a warning, but this isn't a great solution as it could hide other bugs. Open to other suggestions.

disconnect3d:
@jxuanli do you maybe know what's the state of this? shall we close it? shall we still implement sth here?

PR #2142, opened 2024-04-30
kallsyms

### LINKED ISSUE #2106: remove reliance on debug symbols for kconfig
To locate the kernel configuration without debug symbols, we should search for `IKCFG_ST` and `IKCFG_ED` markers in the kernel memory. Focus the search on the kernel mappings where the kernel config is likely to be found; probably the first read-only mapping after the kernel base

https://elixir.bootlin.com/linux/latest/source/kernel/configs.c#L25

chrf01:
partially implements #1836 
fixes: #2106 and #1485 i guess

Tested on the following kernels:
- qemu-test kernels
- 6.9.0-rc6 x86_64
- 5.5.0 x86_64
- 5.1.0 x86_64
- 4.16.0 x86_64
- 4.6.4 x86_64
- 3.14.74 x86_64

was too lazy to go further down but i think it should work on all version > 3.x
probably fails at OpenWRT kernels, 32bit and big endian (i'll fix this soon)

### DIFF
--- a/pwndbg/commands/kchecksec.py
+++ b/pwndbg/commands/kchecksec.py
@@ -105,7 +105,6 @@ parser = argparse.ArgumentParser(description="Checks for kernel hardening config
 
 @pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.KERNEL)
 @pwndbg.commands.OnlyWhenQemuKernel
-@pwndbg.commands.OnlyWithKernelDebugSyms
 @pwndbg.commands.OnlyWhenPagingEnabled
 def kchecksec() -> None:
     kconfig = pwndbg.gdblib.kernel.kconfig()
--- a/pwndbg/commands/kcmdline.py
+++ b/pwndbg/commands/kcmdline.py
@@ -11,7 +11,6 @@ parser = argparse.ArgumentParser(description="Return the kernel commandline (/pr
 
 @pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.KERNEL)
 @pwndbg.commands.OnlyWhenQemuKernel
-@pwndbg.commands.OnlyWithKernelDebugSyms
 @pwndbg.commands.OnlyWhenPagingEnabled
 def kcmdline() -> None:
     print(pwndbg.gdblib.kernel.kcmdline())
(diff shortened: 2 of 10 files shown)


CptGibbon:
Looks like a useful feature, thanks 🙏

Perhaps worth noting this PR is nearly 500 lines and still in draft state, so could be worth splitting up into smaller, easier-to-review PRs of closer to ~100 lines.

chrf01:
I'll close this PR for now, wait for the 32-bit pt fix, and then open smaller PRs soon.

disconnect3d:
@chrf01 tbh it would be great to have it for whatever it works atm and just gate 32-bit saying "To be fixed" or something :).

disconnect3d:
Eh, I see this is missing adding the symbols into actual GDB session, hmm... :(

disconnect3d:
Fwiw here's some command hacked fast (not linted etc) to have a `klookup <addr>` function... better than nothing :).

(In a final version, we should probably just add symbols to the GDB session, but I don't have time for that; if anyone knows how to do it fast, lmk!)

```py

import pwndbg.commands
from pwndbg.commands import CommandCategory
import argparse

parser = argparse.ArgumentParser(description="Lookup kernel address.")
parser.add_argument("addr", type=int, help="Address to lookup")

@pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.KERNEL)
def klookup(addr) -> None:
    ksyms = pwndbg.gdblib.kernel.kallsyms.get()
    found = False

    for k,v in ksyms.items():
        if v[0]==int(addr):
            print('%#x => %s (%s)' % (addr, k, v[1]))
            found = True

    if not found:
        print("No symbol found at %#x" % addr)
```

mebeim:
> we should probably just add symbols to the GDB session

It seems like that would require generating a temporary ELF file with the symbols you want to add and then importing it through `add-symbol-file` or `add-symbol-file-from-memory` :confused:... meh.

fidgetingbits:
@chrf01 Do you still plan on breaking this up or do want to reopen it as is so someone can review? There is some overlap with some gsoc work, so is a good time to take a look/test

PR #2146, opened 2024-05-03
Add `start_and_break_on` decorator to simplify gdb tests

gsingh93:
`@start_and_break_on` takes a binary to run, a list of breakpoints, and optionally arguments to pass to the binary. It performs the same setup `start_binary` does, but also adds the breakpoints and then continues the program. This lets us remove some boilerplate from tests, as shown in the plist tests.

### DIFF
--- a/tests/gdb-tests/tests/__init__.py
+++ b/tests/gdb-tests/tests/__init__.py
@@ -1,3 +1,31 @@
 from __future__ import annotations
 
+import functools
+
+import gdb
+
+import tests
+
 from . import binaries
+
+
+def start_and_break_on(binary, bps, *args):
+    def decorator(func):
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            path = binaries.get(binary)
+
+            gdb.execute("file " + path)
+            gdb.execute("set exception-verbose on")
+            gdb.execute("starti " + " ".join(args))
+
+            if bps is not None and len(bps) > 0:
+                for bp in bps:
+                    gdb.execute(f"break {bp}")
+                gdb.execute("continue")
+
+            return func(*args, **kwargs)
+
+        return wrapper
+
+    return decorator
--- a/tests/gdb-tests/tests/test_command_plist.py
+++ b/tests/gdb-tests/tests/test_command_plist.py
@@ -4,24 +4,15 @@ import re
 
 import gdb
 
-import tests
-
-LINKED_LISTS_BINARY = tests.binaries.get("linked-lists.out")
+from . import start_and_break_on
 
 
-def startup(start_binary):
-    start_binary(LINKED_LISTS_BINARY)
-
-    gdb.execute("break break_here")
-    gdb.execute("run")
-    gdb.execute("up")
-
-
-def test_command_plist_flat_no_flags(start_binary):
+@start_and_break_on("linked-lists.out", ["break_here"])
+def test_command_plist_flat_no_flags():
     """
     Tests the plist for a non-nested linked list
     """
-    startup(start_binary)
+    gdb.execute("up")
 
     expected_out = re.compile(
         """\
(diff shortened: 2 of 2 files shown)


tests/gdb-tests/tests/__init__.py
```diff
@@ -1,3 +1,31 @@
...
+import functools
+
+import gdb
+
+import tests
+
 from . import binaries
+
+
+def start_and_break_on(binary, bps, *args):
```
disconnect3d, line 12: Why don't we make it a pytest fixture instead of a decorator?

gsingh93: Two reasons:
1. A function that the user could just import themselves is not a good pytest fixture. The way we use `start_binary` right now is not great. If we instead automatically started it for them and passed in the process object, that would be a better fixture, but then the user can't customize the binary, breakpoints, etc.
2. Even if we pass this function in as a fixture, the user still has to call it as one of the first things they do. Tests are usually better when the initialization has happened before the function is even called, which can be done with a decorator.

gsingh93: @disconnect3d lmk what you think. I prefer this approach and was going to migrate some other tests to it, but will wait until you let me know you're ok with it.

disconnect3d: @gsingh93 I think making it a decorator makes things more complex unnecessarily? Why not just a standard function call that would be called within the test function instead? Some tests could want to call it multiple times too.

Regarding fixtures, the fixtures are kinda broken bcoz we never run tests all together but instead we run them each one by one. Fixtures allows for 'session fixtures' where the fixture would be called once for all tests in a test session etc. Another good thing of fixtures is that u can yield in them and whatever code is after the yield will be executed after the test finishes, as a kind of 'destructor'. But yeah, maybe its not so needed here.

disconnect3d: @gsingh93 ping :)

gsingh93: > I think making it a decorator makes things more complex unnecessarily?

Do you mean it makes things more complex for the test writer or just in the testing framework in general?

I do think it makes the testing framework a bit more complicated, but I think declarative approaches like this are better in a lot of cases, especially for tests.

That being said, I think even just calling a function like you said would be better than the current fixture approach, so if you still don't want the decorator, I'll change it to that.

BTW, part of the motivation for this was thinking of a possible future extension to the testing framework that could look like this:
```python
@binary("linked-lists.out")
class TestSomeFeature:
    @on_break("break1")
    def test_on_break1(self):
        # run some test code when we hit break1

    @on_break("break2")
    def test_on_break2(self):
        # run some test code when we hit break2
```
Essentially we could just define functions that would test some functionality when the specific breakpoint in `on_break` was hit. No need to manually start the binary, set breakpoints, or continue, everything is handled for you and the test author literally only writes the test code.

I think this could be useful later in the LLDB rewrite, if we abstract all the code dealing with GDB away right now, then later we can just modify the test framework to handle LLDB and the tests should still work.

disconnect3d: I mean its generally more code and more indirection. Not everyone is a Python expert and knows how exactly decorators work, when and how to use it, it requires looking it up and so on.

(There is a similar problem with pytest fixtures, but yeah)

The future extension looks nice, but anytime something breaks and someone would have to debug it, they will have harder time to do it due to all misdirections :P

disconnect3d: So yeah, lets maybe switch to calling a function for now just for the sake of simplicity (?)

gsingh93: > I mean its generally more code and more indirection. Not everyone is a Python expert and knows how exactly decorators work, when and how to use it, it requires looking it up and so on.

This isn't a convincing argument, there are tons of decorators used across pwndbg, should we remove them? We should use Python features that make the code better, and expect developers to learn how to use them.

> The future extension looks nice, but anytime something breaks and someone would have to debug it, they will have harder time to do it due to all misdirections :P

Similar to the above, we shouldn't limit ourselves to worse solutions because they're more complex. We can just write docs explaining how this testing framework works.

gsingh93:
Don't have time to work on this, especially since the main goal was to move towards a more declarative testing structure to make it easier to write tests, particularly with the LLDB port, but that doesn't seem like it'll happen in the near future.

PR #2147, opened 2024-05-04
Fix utf-8 encoding check in docs.py

MatejKafka:
On my machine, the string returned by `locale.getpreferredencoding()` is `utf-8`, not `UTF-8`.


disconnect3d:
Thx!

PR #2152, opened 2024-05-08
Fetch chunk metadata

CptGibbon:
This PR builds on #2082 by adding a `fetch_chunk_metadata()` function to ptmalloc.py

It can be used to fetch all of a chunk's metadata given its address, or just fetch specific fields. It returns a dictionary of `ChunkField` : `int`. It also handles resolution of chunk fields that have had different names over the years e.g. "prev_size" vs. "mchunk_prev_size".

I know using a ChunkField enum looks over-engineered, but I think it has some benefits:
1) Developers don't need to care about those renamed chunk fields, they can just use ChunkField.SIZE instead of knowing whether they need "size" or "mchunk_size"
2) Minimal changes on future field renames, e.g. if "fd" becomes "mchunk_fd", developers just keep using ChunkField.FD
3) Developers don't need to look up chunk field names in the glibc source code, they can just check the enum

Screenshot of the function in action:
Screenshot 2024-05-08 at 14 21 24

The next step is to integrate `fetch_chunk_metadata()` into the Chunk class, which will hopefully remove any direct references to the gdb module. Testing should be a bit easier after that, you could create fake chunks for testing by building a dictionary or mocking return values of calls to `fetch_chunk_metadata()`


pwndbg/heap/ptmalloc.py
```diff
@@ -132,6 +132,60 @@ def heap_for_ptr(ptr):
...
+        elif field is ChunkField.FD_NEXTSIZE:
+            requested_fields.add("fd_nextsize")
+        elif field is ChunkField.BK_NEXTSIZE:
+            requested_fields.add("bk_nextsize")
+
+    fetched_struct = pwndbg.gdblib.memory.fetch_struct_as_dictionary(
+        "malloc_chunk", address, include_only_fields=requested_fields
+    )
+
+    normalized_struct = dict()
```
gsingh93: `./lint.sh -f` should be able to automatically fix the ruff error here, can you try it?

pwndbg/heap/ptmalloc.py
```diff
@@ -132,6 +132,60 @@ def heap_for_ptr(ptr):
...
+class ChunkField(Enum):
+    PREV_SIZE = 1
+    SIZE = 2
+    FD = 3
+    BK = 4
+    FD_NEXTSIZE = 5
+    BK_NEXTSIZE = 6
+
+
+def fetch_chunk_metadata(address: int, include_only_fields: set[ChunkField] = set()):
```
gsingh93: I realize now that we're initializing the default argument as a `set()` here but also in the `fetch_struct_as_dictionary` and `pack_struct_into_dictionary` functions from the last PR.

When you use an object as a default argument, the same instance of the object is used for all invocations of the function when the user doesn't provide an argument (most people expect a new object to be constructed each time, which is wrong). I think this will only lead to bugs if we add/remove from this set, which we don't do in this code, but I'm always wary of code like this. Consider just having the default be `None` and initializing inside the function.

pwndbg/heap/ptmalloc.py
```diff
@@ -132,6 +132,60 @@ def heap_for_ptr(ptr):
...
+        elif field is ChunkField.SIZE:
+            requested_fields.add(size_field_name)
+        elif field is ChunkField.FD:
+            requested_fields.add("fd")
+        elif field is ChunkField.BK:
+            requested_fields.add("bk")
+        elif field is ChunkField.FD_NEXTSIZE:
+            requested_fields.add("fd_nextsize")
+        elif field is ChunkField.BK_NEXTSIZE:
+            requested_fields.add("bk_nextsize")
```
gsingh93: I think this might actually be a good time to use a dictionary, mapping `ChunkField.PREV_SIZE` to `prev_size_field_name`, and so on. Then the for loop becomes simple and easy to read, and we can use the same dictionary in the next for loop, simplifying that code as well and make it less error prone.

CptGibbon: Dictionary was my first thought, but I knew @disconnect3d would say "this dictionary is created on every function call" and I'd have to turn it into an `if/elif` statement 😅
So I'll let you two pick which way you think is best.

pwndbg/heap/ptmalloc.py
CptGibbon: Ah that's news to me, I'll change it here and in `pack_struct_into_dictionary` too 👍

pwndbg/heap/ptmalloc.py
CptGibbon: This requires the `--unsafe-fixes` option.

pwndbg/heap/ptmalloc.py
disconnect3d: Do we actually need an enum? :P

It feels like without it we could just do `include_only_fields=include_only_fields` below and we seem to do the same job twice (note: we don't even validate if one passed an invalid value in `include_only_fields`).

CptGibbon: My justification for the Enum is in the description old chap ☝️

gsingh93: Regarding the dictionary, my thought process is usually if it simplifies more than one if-statement it's worth it, otherwise usually an if-statement is better. And in this case it would simplify two.

If the issue is about constructing the dictionary inside the function (I personally don't mind), it could be constructed outside, and the size/prev_size fields could be set inside the function the first time its run. Alternatively, you can create a second function that essentially does what a dictionary would do, taking a key and returning a value, but I honestly don't know how a function call compares to creating a dictionary in terms of performance.

gsingh93: Regarding the enum, if we don't devs to have to think too much about it, we could also just accept all names for a parameter, i.e. `size` and `mchunk_size`. The downsides of that are mainly that this function gets a bit more complex, and that it may not be clear to the caller what's actually going to happen if the field was renamed. With an enum, it's more clear that we'll handle the field names for the caller.

I'm fine either way, will leave it to @disconnect3d.

disconnect3d:
Needs fixing conflicts :<

CptGibbon:
So if I keep the import statement for `typing.Set` the linter complains, and if I remove it it complains...
What am I missing?

CptGibbon:
@disconnect3d Conflicts resolved & linter is happy 👍

PR #2155, opened 2024-05-09
feat: Add a few new memory APIs and replace old APIs

fidgetingbits:
These are some changes that I made while porting some musl mallocng functionality, but am breaking up PRs so that it's easier to manage. So this just exposes some new API that aren't yet being called anywhere. 

Hopefully are intuitively named. If people have better ideas, I'm happy to change them. If we settle on something, we can change the read_gdbvalue() calls in a separate PR later.

### DIFF
--- a/pwndbg/gdblib/typeinfo.py
+++ b/pwndbg/gdblib/typeinfo.py
@@ -131,14 +131,6 @@ def load(name: str) -> Optional[gdb.Type]:
         return None
 
 
-def read_gdbvalue(type_name: str, addr: int) -> gdb.Value:
-    """Read the memory contents at addr and interpret them as a GDB value with the given type"""
-    gdb_type = load(type_name)
-    if gdb_type is None:
-        raise ValueError(f"Type {type_name} not found")
-    return gdb.Value(addr).cast(gdb_type.pointer()).dereference()
-
-
 def get_type(size: int) -> gdb.Type:
     return {
         1: pwndbg.gdblib.typeinfo.uint8,
--- a/pwndbg/arguments.py
+++ b/pwndbg/arguments.py
@@ -211,7 +211,7 @@ def argument(n: int, abi: pwndbg.lib.abi.ABI | None = None) -> int:
 
     sp = pwndbg.gdblib.regs.sp + (n * pwndbg.gdblib.arch.ptrsize)
 
-    return int(pwndbg.gdblib.memory.poi(pwndbg.gdblib.typeinfo.ppvoid, sp))
+    return int(pwndbg.gdblib.memory.get_typed_pointer_value(pwndbg.gdblib.typeinfo.ppvoid, sp))
 
 
 def arguments(abi: pwndbg.lib.abi.ABI | None = None):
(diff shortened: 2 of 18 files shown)


pwndbg/gdblib/typeinfo.py
```diff
@@ -130,10 +130,28 @@ def load(name: str) -> Optional[gdb.Type]:
...
+def get_typed_pointer_value(type_name: str, addr: int):
+    """Read the pointer value of addr cast to type specified by type_name"""
+    return get_typed_pointer(type_name, addr).dereference()
+
+
+# FIXME: Consider deprecating this, as the name isn't as intuitive as get_typed_pointer()
 def read_gdbvalue(type_name: str, addr) -> gdb.Value:
     """Read the memory contents at addr and interpret them as a GDB value with the given type"""
-    gdb_type = pwndbg.gdblib.typeinfo.load(type_name)
-    return gdb.Value(addr).cast(gdb_type.pointer()).dereference()
```
gsingh93, line 139: Huh, `read_gdbvalue` looks very similar to `poi`: https://github.com/pwndbg/pwndbg/blob/00adfbbb5e06564f7d351624e09dae898f485c57/pwndbg/gdblib/memory.py#L307-L312

I agree that both names suck, and prefer your naming.

My only question is whether this is the right file to put these functions in, or is memory.py a better location?

fidgetingbits: Hrm, `poi` has better type definitions than what I used so I'll fix mine up to use that. I don't like the name `poi` either. 

I'm happy to move these to `memory.py`. I think it makes sense given what else is in there, and that `poi` is more widely used, so people will likely expect them there. I guess the only real type-related part is the `load()` call used by `get_typed_pointer()`, and that will stay in `typeinfo.py`.

I can swap out all uses of `poi` and `read_gdbvalue` as part of this if you want, or can leave it for later.

CptGibbon: Phew I thought I was the only one who thought `poi` was a terrible name 😅

fidgetingbits: I've replaced all uses of poi() and read_gdbvalue() with the new API, and moved it into gdblib/memory instead.

fidgetingbits:
With noting that there is a risk that people who have scripts that use pwndbg's gdblib could be using poi() and read_gdbvalue(). Not sure if there is an established API deprecation mechanism for changes like this, or if you want to create one, or just accept that this may break stuff for external tooling.

gsingh93:
@fidgetingbits needs a rebase.

@disconnect3d are you good with the changes in this PR?

fidgetingbits:
rebased now

pwndbg/gdblib/typeinfo.py
disconnect3d: I also dislike the `poi` name ;). I think it was taken from windbg or elsewhere...

disconnect3d (approved):
lgtm; @gsingh93 please merge if u are okay with those changes

PR #2157, opened 2024-05-09
refactor: Switch to use `pwnlib.term.text` for bold coloring

fidgetingbits:
Previously if you wanted to use bold colors people would use a lambda or manually use 2 functions, for example:

```
pwndbg/commands/context.py
934:                line += f" <{pwndbg.color.bold(pwndbg.color.green(symbol))}> "
pwndbg/commands/tls.py
85:    bold_green = lambda text: pwndbg.color.bold(pwndbg.color.green(text))
```

These would let you just use `pwndbg.color.bold_green()` in the first case or `from pwndbg.color import bold_green` for the second case.

A tool I'm porting uses bolded colors for lots of stuff, which I'm keeping the same for now, so am using a bunch of these new functions.

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -15,6 +15,7 @@ from typing import List
 from typing import Optional
 from typing import Tuple
 
+from pwnlib.term import text
 from typing_extensions import ParamSpec
 
 import pwndbg
@@ -1178,7 +1179,7 @@ def context_threads(with_banner=True, target=sys.stdout, width=None):
 
             line += f"{pc_colored}"
             if symbol:
-                line += f" <{pwndbg.color.bold(pwndbg.color.green(symbol))}> "
+                line += f" <{text.bold_green(str(symbol))}> "
 
         out.append(line)
 
--- a/pwndbg/commands/misc.py
+++ b/pwndbg/commands/misc.py
@@ -4,6 +4,8 @@ import argparse
 import errno
 from collections import defaultdict
 
+from pwnlib.term import text
+
 import pwndbg.aglib.memory
 import pwndbg.aglib.regs
 import pwndbg.aglib.vmmap
@@ -101,7 +103,7 @@ parser.add_argument(
 def pwndbg_(filter_pattern, shell, all_, category_, list_categories) -> None:
     if list_categories:
         for category in CommandCategory:
-            print(C.bold(C.green(f"{category.value}")))
+            print(text.bold_green(category.value))
         return
 
     if all_:
@@ -133,11 +135,11 @@ def pwndbg_(filter_pattern, shell, all_, category_, list_categories) -> None:
             continue
         data = table_data[category]
 
-        category_header = C.bold(C.green(category + " Commands"))
-        alias_header = C.bold(C.blue("Aliases"))
+        category_header = text.bold_green(category + " Commands")
+        alias_header = text.bold_blue("Aliases")
         print(
             tabulate(
-                data, headers=[f"{category_header} [{alias_header}]", f"{C.bold('Description')}"]
+                data, headers=[f"{category_header} [{alias_header}]", f"{text.bold('Description')}"]
             )
         )
         print()
--- a/pwndbg/commands/tls.py
+++ b/pwndbg/commands/tls.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 
 import argparse
 
+from pwnlib.term.text import bold_green
 from tabulate import tabulate
 
 import pwndbg.aglib.memory
@@ -85,7 +86,6 @@ group.add_argument(
 def threads(num_threads, respect_config) -> None:
     table = []
     headers = ["global_num", "name", "status", "pc", "symbol"]
-    bold_green = lambda text: pwndbg.color.bold(pwndbg.color.green(text))
 
     import gdb
 
@@ -122,7 +122,7 @@ def threads(num_threads, respect_config) -> None:
 
         if thread is original_thread:
             row = [
-                bold_green(thread.global_num),
+                bold_green(str(thread.global_num)),
                 bold_green(name),
             ]
         else:
@@ -144,7 +144,7 @@ def threads(num_threads, respect_config) -> None:
 
             if symbol:
                 if thread is original_thread:
-                    row.append(bold_green(symbol))
+                    row.append(bold_green(str(symbol)))
                 else:
                     row.append(symbol)
 


gsingh93:
What if colors took optional arguments like `bold=True` instead? That seems like a cleaner solution to me.

fidgetingbits:
Ya, that would work too. Could then add underline, italics, etc flags that way too. I'm fine with whatever you all think is best.

fidgetingbits:
One caveat to using flags is, for things like memory coloring, is that pwndbg uses a table like this:


```python
c = ColorConfig(
    "memory",
    [
        ColorParamSpec("stack", "yellow", "color for stack memory"),
        ColorParamSpec("heap", "blue", "color for heap memory"),
        ColorParamSpec("code", "red", "color for executable memory"),
        ColorParamSpec("data", "purple", "color for all other writable memory"),
        ColorParamSpec("rodata", "normal", "color for all read only memory"),
        ColorParamSpec("rwx", "underline", "color added to all RWX memory"),
        ColorParamSpec("gap", "bold_white", "color added for unmapped regions of memory"),
    ],
)
```

I've added an example "gap" entry to demonstrate someone might want to differentiate the line by using a bold color. Your suggestion wouldn't work there due to the way the generated functions work.

disconnect3d (approved):
I'm okay with those changes: I am a bit worried, that we introduce more and more redundant func calls where each of them also performs a global fetch but w/e: some of these are probably cached elsewhere.

In the future, we should probably just remove all the `BOLD = ...` constants if no other code uses it (and it shouldn't) as well as calls to `bold(black(x))` and just inline it in within the functions. Kinda a microoptimization, but some of the coloring is called many many times and it may be beneficial on the long run.

disconnect3d:
@gsingh93 if we add `bold=...` we introduce an unecessary `if ...` into all the functions, idk, it probably doesn't hurt much but I'm okay with either solution for now. So either merge this or request changes :)

gsingh93 (changes requested):
Lets add the bold argument. An additional if statement is better than a number of redundant function calls.

fidgetingbits:
I'll switch it around

fidgetingbits:
Switched to use bold argument. I've also switched the other muslheap thing that I'm going to send a PR for soon to use this and seems to work fine.

pwndbg/color/__init__.py
```diff
@@ -45,75 +45,75 @@
 none = str
 
 
-def normal(x: str) -> str:
+def normal(x: str, bold=False) -> str:
```
gsingh93: It doesn't look like we're passing the `bold` argument to `colorize`?

Also, it's probably better to use `*args` to pass it up to `colorize`. ~If you can figure out how to type it with `ParamSpec` (there are some other examples in the codebase if you grep for it) that would be great, but if you find it tricky just use `Any`.~ EDIT: I don't think `ParamSpec` will work here, just use `Any`.

gsingh93 (changes requested):

pwndbg/color/__init__.py
disconnect3d: @fidgetingbits ping :)

fidgetingbits: Makes sense. To be clear, you are suggesting I change the order of colorize() args so that the color is first? That way every color wrapper can just specify the color and then pass all *args and **kwargs, so if anything gets added it's easier? Something like:

```python
def white(*args, **kwargs) -> str:
    return colorize(WHITE, args, kwargs)
```

gsingh93: Yup, I think that's the right way to go. Technically you don't need to change the argument order if you directly pass `x`, `color`, and then only `bold` comes from `*args` or `**kwargs`, but I don't see a reason not to include `x` into `*args` or `**kwargs` as well. You'll need to change a few invocations of `colorize` outside of this file though if you do it like this.

fidgetingbits:
I think this should be okay now. Since so many of the integration tests fail on nix, which I haven't had time to look into fixing, it's hard to be 100% sure if everything is normal for stuff like this that touches most commands, but I added a new unit test to check that some basic colors work. I also took the opportunity to fix the scripts to work on nix as well.

fidgetingbits:
I'll fix the linting/typing issues shortly

disconnect3d:
A conflict needs to be fixed here. @fidgetingbits any chance u can fix it?

fidgetingbits:
I can take a look next week

peace-maker:
Fyi pwntools includes a module for terminal colors too `pwnlib.term.text`.

It generates functions to change foreground and/or background colors and/or modifiers like bold/italic/underline. So `text.bold_red("text")` would return bold red text and `text.italic_red_on_white` would return italic red text on white background.

Maybe another opportunity for code deduplication?

I generated a list of all color and modifier combinations once here:
https://github.com/peace-maker/pwntools/blob/type-stubs/pwnlib%2Fterm%2Ftext.pyi

fidgetingbits:
makes sense to me to use that if pwndbg is already relying on pwntools elsewhere, but will leave it up to the others @disconnect3d @gsingh93 . Happy to switch it over if preferred.

fidgetingbits:
@disconnect3d any preference on what peace-maker suggested? I was holding off until one of you commented, but since you pinged me, I'll ping you ;) The muslheap stuff being fixed up is also pending on this PR.

fidgetingbits:
Probably makes sense to move to `pwnlib.term.text` but can be a follow up PR/issue I think. Just fixed the merge conflict so this can move forward and I can update the muslheap one after.

disconnect3d:
@fidgetingbits I am all hands for code deduplication, so if we can reuse the Pwntools code or/and help improve their code, lets do that! :)

fidgetingbits:
While testing switching to pwnlib for bold coloring, I noticed that a lot of existing pwndbg colorize() calls pass addresses/ints directly, and it automatically does [string conversion](https://github.com/pwndbg/pwndbg/blob/0caf175699a1d2f627512d3dbe46674611cc181e/pwndbg/color/__init__.py#L129), where as `pwnlib.term.text` functions do not. 

@peace-maker would you be open to add auto conversion to `pwnlib.term.text`? I guess likely [here](https://github.com/Gallopsled/pwntools/blob/6f0793eebcf78eab4ac86bb8fcbb1ffee3e59fa6/pwnlib/term/text.py#L83)?  

I do think using pwnlib coloring across the board will probably be nice, but for now this PR will only focus on bold and I'll eventually migrate the rest in another PR.

fidgetingbits:
Had some recent conflicts due to aglib refactoring, so fixed those. @disconnect3d if no other CI issues, his is fine to go in imo. All the other color to pwnlib.text changes I will do in a separate PR when I get time.

disconnect3d:
This now conflicts with context.py.

Also, as per discord discussion, this does not play well with `set disable-colors on` since this is only applied to `pwndbg.color.*` functions :/

fidgetingbits:
> This now conflicts with context.py.
> 
> Also, as per discord discussion, this does not play well with `set disable-colors on` since this is only applied to `pwndbg.color.*` functions :/

Ah interesting about disable colors. Will close this PR. Realistically I'm not gonna find time to keep fiddling with this. It was mostly for the musl heap PR using lots of bold, so I may just switch that PR to use whatever existing pwndbg (lambdas iirc) uses for bold or just remove bold from it entirely.

PR #2159, opened 2024-05-09
Minor docs/website cleanup

gsingh93:
- Dependencies are installed with poetry (install with `poetry install --with docs`
- GH workflow was updated to use poetry and fetch-depth was removed
- I don't think the docs site is considered "new" anymore, so I removed the banner. Let me know if you want it back
- Minor reformatting of mkdocs.yml
- Removed the old Makefile and .gitignore that were for the sphinx build
- The navigation/toc was shown the blog post listing page, it didn't look good so I removed it

Before:
image

After:
image


.github/workflows/docs.yml
```diff
@@ -19,8 +18,17 @@ jobs:
         with:
           key: ${{ github.ref }}
           path: .cache
```
gsingh93: I don't know what we're caching here, I don't see `.cache` locally, but I just left it since I wasn't sure. If anyone knows why it's here, let me know.

gsingh93: I'm just going to remove it, I see this in the action logs:
```
Post job cleanup.
Warning: Path Validation Error: Path(s) specified in the action for caching do(es) not exist, hence no cache is being saved.
```

So it's clearly not being used.

PR #2164, opened 2024-05-13
Add strip_colors as lib/strings API

fidgetingbits:
This makes a function that was in `ai.py` reusable by moving it into `pwndbg/lib/strings.py`


disconnect3d (approved):
Added a fix for lint issue; lgtm

PR #2166, opened 2024-05-13
Nix dev shell

fidgetingbits:
NOTE: This is a stacked PR based on #2165 

This adds a nix development shell, which can be entered using `nix develop` or `direnv allow` (via the added `.envrc` file).

It breaks out the poetry derivation into a new file so it can be re-used by the regular pwndbg build and the devshell. It also makes it so the development options can be used by the development environment, without any manual changes. 

In order to build a version of gdbinit.py that works for testing so it automatically picks up the pi project dev dependencies, you have to `nix build .#pwndbg-dev`. The development shell itself will automatically get those so running certain python scripts manually will work already. If one of the nix people has a better idea how todo this I'm open to suggestions.

Finally it modifies the way the pwndbg command is invoked to use an auto-patched gdbinit.py, which also allows for gdbinit.py to be lazy loaded on nix and still have access to the correct python environment. This is useful both for when you need it, and also because there is at least one test script that explicitly loads gdbinit.py. 

I have future modifications for making the tests work, which will come in a separate PR.

### DIFF
--- a/nix/pwndbg.nix
+++ b/nix/pwndbg.nix
@@ -3,6 +3,7 @@
   python3 ? pkgs.python3,
   gdb ? pkgs.gdb,
   inputs ? null,
+  isDev ? false,
 }:
 let
   binPath = pkgs.lib.makeBinPath (
@@ -15,38 +16,14 @@ let
     ]
   );
 
-  pyEnv = pkgs.poetry2nix.mkPoetryEnv {
-    groups = [ ]; # put [ "dev" ] to build "dev" dependencies
-    checkGroups = [ ]; # put [ "dev" ] to build "dev" dependencies
-    projectDir = inputs.pwndbg;
-    python = python3;
-    overrides = pkgs.poetry2nix.overrides.withDefaults (
-      self: super: {
-        pip = python3.pkgs.pip; # fix infinite loop in nix, look here: https://github.com/nix-community/poetry2nix/issues/1184#issuecomment-1644878841
-        unicorn = python3.pkgs.unicorn; # fix build for aarch64 (but it will use same version like in nixpkgs)
-
-        # disable build from source, because rust's hash had to be repaired many times, see: PR https://github.com/pwndbg/pwndbg/pull/2024
-        cryptography = super.cryptography.override { preferWheel = true; };
-
-        unix-ar = super.unix-ar.overridePythonAttrs (old: {
-          nativeBuildInputs = (old.nativeBuildInputs or [ ]) ++ [ self.setuptools ];
-        });
-
-        pt = super.pt.overridePythonAttrs (old: {
-          buildInputs = (old.buildInputs or [ ]) ++ [ super.poetry-core ];
-        });
-        capstone = super.capstone.overridePythonAttrs (old: {
-          # fix darwin
-          preBuild = pkgs.lib.optionalString pkgs.stdenv.isDarwin ''
-            sed -i 's/^IS_APPLE := .*$/IS_APPLE := 1/' ./src/Makefile
-          '';
-          # fix build for aarch64: https://github.com/capstone-engine/capstone/issues/2102
-          postPatch = pkgs.lib.optionalString pkgs.stdenv.isLinux ''
-            substituteInPlace setup.py --replace manylinux1 manylinux2014
-          '';
-        });
-      }
-    );
+  pyEnv = import ./pyenv.nix {
+    inherit
+      pkgs
+      python3
+      inputs
+      isDev
+      ;
+    lib = pkgs.lib;
   };
 
   pwndbgVersion = pkgs.lib.readFile (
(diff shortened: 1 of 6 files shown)


disconnect3d:
@patryk4815 @arcz can you folks review this PR?

nix/pwndbg.nix
```diff
@@ -71,13 +51,18 @@ let
       mkdir -p $out/share/pwndbg
 
       cp -r gdbinit.py pwndbg $out/share/pwndbg
+      # Build self-contained init script for lazy loading from vanilla gdb
+      # I purposely use insert() so I can re-import during development without having to restart gdb
+      sed "2 i import sys, os\n\
+      sys.path.insert(0, '${pyEnv}/${pyEnv.sitePackages}')\n\
+      sys.path.insert(0, '$out/share/pwndbg/')\n\
+      os.environ['PATH'] += ':${binPath}'\n\
+      os.environ['LC_CTYPE'] = 'C.UTF-8'\n" -i $out/share/pwndbg/gdbinit.py
```
patryk4815: @disconnect3d i think this need to be upstreamed into `gdbinit.py` not only for nix'ed version
talking about LC_CTYPE

nix/pwndbg.nix
```diff
@@ -71,13 +51,18 @@ let
...
 
       ln -s ${pyEnv} $out/share/pwndbg/.venv
 
       makeWrapper ${gdb}/bin/gdb $out/bin/pwndbg \
         --add-flags "--quiet --early-init-eval-command=\"set charset UTF-8\" --early-init-eval-command=\"set auto-load safe-path /\" --command=$out/share/pwndbg/gdbinit.py" \
```
patryk4815:
```suggestion
        --add-flags "--quiet --early-init-eval-command=\"set auto-load safe-path /\" --command=$out/share/pwndbg/gdbinit.py"
```

nix/pwndbg.nix
patryk4815: I think we already fixed it in https://github.com/pwndbg/pwndbg/pull/2139
Se `LC_CTYPE` can be removed.

nix/portable.nix
```diff
@@ -1,58 +1,62 @@
...
+    ''
+  );
+
+  pwndbgBundleBin = pkgs.writeScript "pwndbg" ''
+    #!/bin/sh
+    dir="$(cd -- "$(dirname "$(dirname "$(realpath "$0")")")" >/dev/null 2>&1 ; pwd -P)"
+    export PYTHONHOME="$dir"
+    export PYTHONPYCACHEPREFIX="$dir/cache/"
+    export PWNDBG_VENV_PATH="PWNDBG_PLEASE_SKIP_VENV"
+    exec "$dir/lib/${ldName}" "$dir/exe/gdb" --quiet --early-init-eval-command="set charset UTF-8" --early-init-eval-command="set auto-load safe-path /" --command=$dir/exe/gdbinit.py "$@"
```
patryk4815:
```suggestion
    exec "$dir/lib/${ldName}" "$dir/exe/gdb" --quiet --early-init-eval-command="set auto-load safe-path /" --command=$dir/exe/gdbinit.py "$@"
```

nix/pwndbg.nix
```diff
@@ -71,13 +51,18 @@ let
       mkdir -p $out/share/pwndbg
 
       cp -r gdbinit.py pwndbg $out/share/pwndbg
+      # Build self-contained init script for lazy loading from vanilla gdb
+      # I purposely use insert() so I can re-import during development without having to restart gdb
+      sed "2 i import sys, os\n\
+      sys.path.insert(0, '${pyEnv}/${pyEnv.sitePackages}')\n\
+      sys.path.insert(0, '$out/share/pwndbg/')\n\
```
patryk4815, line 58: `sys.path.insert(0, '$out/share/pwndbg/')`
🤔  why is needed?

fidgetingbits: `pyEnv` only provides the dependencies of pwndbg, whereas for `gdbinit.py` to `import pwndbg`, it needs to import from `share/pwndbg/` as that is where the `pwndbg` python module is installed to.

This is without that line:

```
(gdb) source result/pwndbg/share/gdbinit.py
result/pwndbg/share/gdbinit.py: No such file or directory.
(gdb) source result/share/pwndbg/gdbinit.py
  File "result/share/pwndbg/gdbinit.py", line 65, in <module>
    import pwndbg  # noqa: F401
    ^^^^^^^^^^^^^
ModuleNotFoundError: No module named 'pwndbg'
```

If there's some better way, happy to change it.

Reminds me even though `pwndbg.nix` links the `.venv/` folder, there is no activation script created by the poetry derivation in `.venv/bin/`. Is anyone actually using the venv on nix? Since it won't include pwndbg (as it only links the poetry sitePackages) I think it might not work? Maybe I'm missing something.

nix/pwndbg.nix
fidgetingbits: fixed

fidgetingbits:
#2120 broke this PR, because it added a new development dependency `sortedcontainers-stubs`. But I will update it shortly. Tested with

```bash
      sortedcontainers-stubs = super.sortedcontainers-stubs.overridePythonAttrs (old: {
        buildInputs = (old.buildInputs or [ ]) ++ [ super.poetry-core ];
      });
```

and seems to be working...

It might be worth eventually having a CI workflow for building the nix development environment that at least warns when it breaks (but may be not fail a merge, since it's not something most people will be using)

fidgetingbits:
happy to squash these commits before this goes in if needed, but keeping them for now because of the review discussion

gsingh93:
#2165 was merged, this needs a rebase now.

fidgetingbits:
should be rebased now

disconnect3d:
I think the failing qemu test is unrelated, isn't it?

@patryk4815 anything else to be done here or do u think we can merge it?

fidgetingbits:
> I think the failing qemu test is unrelated, isn't it?
> 
> @patryk4815 anything else to be done here or do u think we can merge it?

Ya, I don't see how that qemu test would be related of hand...

fidgetingbits:
Noticed a bug where ZIGPATH was set to the wrong version, so just fixed that.

disconnect3d:
@fidgetingbits needs fixing a conflict in gitignore :(

fidgetingbits:
> @fidgetingbits needs fixing a conflict in gitignore :(

Should be fixed. Still need @patryk4815 to verify what I responded about his review questions are okay.

Also it seems linting is failing on something I took in from upstream/dev after rebase? I guess maybe https://github.com/pwndbg/pwndbg/pull/2044 had a lint error that was missed and got merged anyway, but not sure.

disconnect3d:
@fidgetingbits I contacted @patryk4815 and we decided to merge this.

An alternative to the `sys.path.insert(0, ...)` could maybe be `PYTHONPATH` but we don't want to set more envvars if we don't have to since then one has to clear this env in gdbscript when running proceses from within GDB.

PR #2167, opened 2024-05-13
Fix lint script shebang, args, and nix env checks

fidgetingbits:
This makes the lint script usable on nix. Also makes the actual command-line argument consistent with the usage message (was --fix in usage, --format in the arg parsing)

### DIFF
--- a/lint.sh
+++ b/lint.sh
@@ -16,13 +16,13 @@ FIX=0
 
 while [[ $# -gt 0 ]]; do
     case $1 in
-    -f | --fix)
-        FIX=1
-        shift
-        ;;
-    *)
-        help_and_exit
-        ;;
+        -f | --fix)
+            FIX=1
+            shift
+            ;;
+        *)
+            help_and_exit
+            ;;
     esac
 done
 
--- a/lint.sh
+++ b/lint.sh
@@ -40,7 +40,7 @@ set -o xtrace
 LINT_FILES="pwndbg tests *.py"
 LINT_TOOLS="isort ruff vermin mypy"
 
-if ! type ${LINT_TOOLS} &>/dev/null; then
+if ! type ${LINT_TOOLS} &> /dev/null; then
     PIP_CMD="poetry install --with dev"
     echo "Missing one of the following tools: ${LINT_TOOLS}"
     echo "Running '${PIP_CMD}'"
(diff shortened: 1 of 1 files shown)


disconnect3d (approved):
Hopefully fixed lint issues; lgtm

lint.sh
```diff
@@ -31,14 +31,16 @@ if [[ -z "${PWNDBG_VENV_PATH}" ]]; then
...
+    source "${PWNDBG_VENV_PATH}/bin/activate"
+fi
 
 set -o xtrace
 
 LINT_FILES="pwndbg tests *.py"
 LINT_TOOLS="isort ruff vermin mypy"
 
-if ! type ${LINT_TOOLS} &> /dev/null; then
+if ! type ${LINT_TOOLS} &>/dev/null; then
```
gsingh93: The linter is complaining about this change

fidgetingbits: Sorry, my editor's linters re-changes every time I save the file, and I guess forget to run your linter before final commit.

Not sure it's been discussed, but it may be worth considering using something like [pre-commit](https://pre-commit.com/) so that it hooks the commit and forces the linting then, rather than having to remember to run lint.sh every time.

fidgetingbits: Should be fixed now

gsingh93: Regarding pre-commit, the setup-dev.sh script asks if you want to add a git hook that lints before pushing. Seems like pre-commit is also just run through a git hook, so not sure if there's benefit in switching to it, but I'm open to it if there's a compelling reason.

fidgetingbits: Ah makes sense. Since setup-dev.sh doesn't work on nix, I didn't run it fully and so didn't see the hook question. I will look into installing the hooks in the nix dev shell I added though.

To me the extra benefits of something like pre-commit is you get installation and version pinning for all linters, not just python stuff. 

For example shfmt comes from various OS package managers in setup-dev.sh, and so everyone could be running different versions. pre-commit would install the same version of that too (not just whatever python tooling is available via poetry file). Could also be used to more easily enforce additional linting like whitespacing in documentation, markdown structure consistency across the project etc. 

Also you wouldn't need to re-run setup-dev.sh if something new is added. the pre-commit stuff would automatically have everyone using the same linters and versions.

Anyway, up to you folks and maybe not needed. I've used it on a few projects and seems nice though, so figure I'd mention.

PR #2168, opened 2024-05-13
Modify python test scripts to work from nix

fidgetingbits:
Minor tweaks to the testing scripts to be able to run in the nix dev shell and lazy load the gdbinit.py from utils.py. If someone has preference for how to pass the path to gdbinit.py instead of the environment let me know. Since it's just various tests importing that file directly, I chose that route. Don't have a lot of pytest experience though. 

This change allows someone to run `./tests.sh --nix` to test with the pwndbg copy built by nix.

### DIFF
--- a/tests/gdb-tests/tests.py
+++ b/tests/gdb-tests/tests.py
@@ -5,6 +5,7 @@ import concurrent.futures
 import os
 import re
 import subprocess
+import sys
 import time
 from subprocess import CompletedProcess
 from typing import Tuple
@@ -159,6 +160,11 @@ def parse_args():
         "-s", "--serial", action="store_true", help="run tests one at a time instead of in parallel"
     )
     parser.add_argument(
+        "--nix",
+        action="store_true",
+        help="run tests using gdbinit.py built for nix environment",
+    )
+    parser.add_argument(
         "--collect-only",
         action="store_true",
         help="only show the output of test collection, don't run any tests",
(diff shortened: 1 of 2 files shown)


tests/gdb-tests/tests/utils.py
```diff
@@ -1,12 +1,23 @@
...
 
 import codecs
+import os
 import re
 import subprocess
 
+if os.environ.get("GDB_INIT_PATH"):
+    GDB_INIT_PATH = os.environ["GDB_INIT_PATH"]
+else:
+    GDB_INIT_PATH = "../../gdbinit.py"
```
disconnect3d: Let's replace this with:

```py
GDB_INIT_PATH = os.environ.get("GDB_INIT_PATH", "../../gdbinit.py")
```

disconnect3d: Made this change directly to the sources

disconnect3d (approved):
lgtm

tests/gdb-tests/tests.py
```diff
@@ -176,6 +182,14 @@ def parse_args():
...
         print("Will run tests in serial and with Python debugger")
         args.serial = True
+    if args.nix:
+        gdbinit_path = os.path.join(ROOT_DIR, "result/share/pwndbg/gdbinit.py")
+        if not os.path.exists(gdbinit_path):
+            print("ERROR: No nix-compatible gdbinit.py found. Run nix build .#pwndbg-dev")
+            sys.exit(1)
+        GDB_INIT_PATH = os.path.join(ROOT_DIR, "result/share/pwndbg/gdbinit.py")
+        # This tells tests/utils.py where to find the gdbinit.py file when used by various tests
+        os.environ["GDB_INIT_PATH"] = GDB_INIT_PATH
```
gsingh93: Why not just use `gdbinit_path` here? I don't think we need to define the `GDB_INIT_PATH` in the previous line.

tests/gdb-tests/tests/utils.py
```diff
@@ -1,12 +1,20 @@
 from __future__ import annotations
 
 import codecs
+import os
 import re
 import subprocess
 
+GDB_INIT_PATH = os.environ.get("GDB_INIT_PATH", "../../gdbinit.py")
```
gsingh93: I think the variable name should be lower case.

PR #2180, opened 2024-05-22
Delete pwndbg.constants in favor of pwnlib

### LINKED ISSUE #2175: Remove pwndbg.constants
Most of [pwndbg.constants](https://github.com/pwndbg/pwndbg/tree/dev/pwndbg/constants) is copied from [pwnlib.constants](https://github.com/Gallopsled/pwntools/tree/e92a30bbf7a838e739b8428d9b1d446fd671cf48/pwnlib/constants) in pwntools. We should just use the pwntools constants to avoid duplication of code.

I believe the only place this code was used was in this `syscall` function: https://github.com/pwndbg/pwndbg/blob/e812ef0fffc2826bd22d274de55de1a691778538/pwndbg/constants/__init__.py#L26

So after making the change, we need to verify that still works (if there is not already a test for this, we should add one).

The only file here that should be kept is [pwndbg/constants/ptmalloc.py](https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/constants/ptmalloc.py). Personally, I don't think we should keep the entire constants directory for this one small file, and we should move this into the ptmalloc code.

Ordoviz:
Fixes #2175.

This requires a recent version of pwntools (https://github.com/Gallopsled/pwntools/commit/cb540857ccd4ea45678e24949f48a2231857f8ed).

There already was a test for syscall names in `context`: https://github.com/pwndbg/pwndbg/blob/2b9beef7afc99db312667ac9bf90f866596287b2/tests/gdb-tests/tests/test_context_commands.py#L159

Use the script below to generate a diff of syscall constants between pwndbg.constants and pwntools/pwnlib:

```sh
#!/bin/sh
diffdir=/tmp/constants-diff
mkdir -p "$diffdir"/pwntools
mkdir -p "$diffdir"/pwndbg

git clone --depth=1 https://github.com/pwndbg/pwndbg /tmp/pwndbg
git clone --depth=1 https://github.com/Gallopsled/pwntools /tmp/pwntools

transform() {
	# delete quotes, parentheses, and spaces; then sort by syscall number
	grep __NR "$1"  | tr -d '"'"'"'() ' | tr '+' ',' |
		awk -F, '{ if (NF>2){printf "%s,%s\n", $1, $2+$3} else {print} }' |
		sort -n -k 2 -t ","
}

for f in /tmp/pwntools/pwnlib/constants/linux/*; do
	transform "$f" > "$diffdir"/pwntools/"$(basename "$f")"
done
for f in /tmp/pwndbg/pwndbg/constants/*; do
	case "$f" in
		*ptmalloc.py) ;;
		*constant.py) ;;
		*)
			transform "$f" > "$diffdir"/pwndbg/"$(basename "$f")"
	esac
done

diff -r --color=always "$diffdir/pwndbg" "$diffdir/pwntools"
# red lines (prefixed with "<") are those present in pwndbg but missing in pwntools
```

The script shows that a bunch of aarch64 syscalls from 1024 to 1999 as well as some s390x syscalls are missing in pwnlib, but it's still a net plus.


disconnect3d:
Seems that this causes some import errors in CI:
```
../../pwndbg/arguments.py:21: in <module>
    from pwnlib.constants.linux import riscv64
E   ImportError: cannot import name 'riscv64' from 'pwnlib.constants.linux' (/venv/lib/python3.8/site-packages/pwnlib/constants/linux/__init__.py)
```

Can you check it out and fix it?

Ordoviz:
> Can you check it out and fix it?

We either need to wait until pwntools makes a release that includes https://github.com/Gallopsled/pwntools/commit/cb540857ccd4ea45678e24949f48a2231857f8ed or remove riscv syscall support.

disconnect3d:
cc: @peace-maker @Arusekk bump, can we make a new Pwntools release? :)

gsingh93:
Also for the pwntools maintainers mentioned above, do you want us to contribute the missing aarch64 constants upstream? If so we should do that before you cut any additional releases.

peace-maker:
Sure, it's about time again. I think any differences should be aligned first, so yes, please do add those missing constants. I'd expect it to be a matter of updating the include files from diet-libc and rerunning the generator.

Ordoviz:
Those "missing" aarch64 syscalls have been purposefully removed in https://github.com/Gallopsled/pwntools/issues/896. No need to re-add them, I guess.

The missing s390x syscalls seem to coincide with the 32-bit syscalls listed [here](https://github.com/qemu/qemu/blob/6af8037c42fdc3d20d5aa2686799ab356a9ee1a9/linux-user/s390x/syscall.tbl). They were removed in https://github.com/Gallopsled/pwntools/commit/2d1e0616a706fc910ab23539cea517221f535fc9#diff-de744d4b7f0c7525c4f8196f76c9019989e69d4ea6f88f6151813ef76c56604d.

peace-maker:
If the aarch64 syscalls are still unimplemented, I'd expect them to be irrelevant? Since s390x is the 64bit version of s390 I guess leaving out those 32bit syscalls is fine. This is where the syscalls are generated from: https://github.com/ensc/dietlibc/blob/master/s390x/syscalls.h

[dietlibc](https://www.fefe.de/dietlibc/) didn't see a new release since 2018, so maybe we should generate the constants based on a maintained libc like musl to grab the newest syscalls? But that is independent of this PR.

gsingh93:
That's an additional benefit of us removing this duplicated code, we get to remove a bunch of useless constants 🙂 I don't think there's anything additional that needs to be done for this PR then, we just need to wait for the next pwntools release.

disconnect3d:
ping: @Arusekk @peace-maker any ETA for a pwntools release?

@Ordoviz can u fix conflicts here?

Ordoviz:
Sure, I will fix conflicts once a new version of pwntools is out.

peace-maker:
The RISCV64 constants are in pwntools [4.14.0beta0](https://github.com/Gallopsled/pwntools/releases/tag/4.14.0beta0) now.

disconnect3d:
@Ordoviz can u update this and fix conflicts?

disconnect3d:
Hi, this now conflicts with changes made in https://github.com/pwndbg/pwndbg/pull/2355 , pls rebase to latest dev and sorry for the trouble :<

Ordoviz:
This should be ready for merging now.

disconnect3d:
> This should be ready for merging now.

Just to be sure, do we need to use pwntools 14 beta?

pwndbg/gdblib/disasm/arch.py
```diff
@@ -521,6 +523,39 @@ def _telescope_format_list(self, addresses: List[int], limit: int, emu: Emulator
                 enhance_string_len=enhance_string_len,
             )
 
+    @staticmethod
+    def _syscall_name(number: int, arch: str) -> str | None:
```
disconnect3d, line 527: I think this change is wrong; seems like not correct rebase/conflict merge/handling? :/

disconnect3d (changes requested):
It seems some incorrect merge/rebase reverter changes with the `_syscall_name` function, which are irrelevant/redundant/unwanted here.

pwndbg/gdblib/disasm/arch.py
Ordoviz: Which change in particular? This function is needed by `_enhance_syscall` and to my knowledge, pwntools does not provide a function to convert syscall numbers into names.

disconnect3d: Oh I see, this previously was defined in `constants/__init__.py` which was deleted. I thought it was defined elsewhere (or maybe moved during recent refactorings). This makes sense, nevermind!

disconnect3d (approved):
My bad, all good here.

pwndbg/gdblib/disasm/arch.py
```diff
@@ -521,6 +523,39 @@ def _telescope_format_list(self, addresses: List[int], limit: int, emu: Emulator
...
+        arch_module = {
+            "arm": linux.arm,
+            "armcm": linux.arm,
+            "i386": linux.i386,
+            "mips": linux.mips,
+            "x86-64": linux.amd64,
+            "aarch64": linux.aarch64,
+            "rv32": linux.riscv64,
+            "rv64": linux.riscv64,
+        }.get(arch)
```
disconnect3d, line 541: @Ordoviz FYI I feel like there is some kind of `pwndbg.somelib.arch` or `.arch_mod` or sth that maps to it already. Otherwise maybe we should have one, or, this dict in general should probably be moved to some global scope so it is not re-created each time this is called.

disconnect3d:
Btw see my last comment. But anyway, thanks for this!

PR #2181, opened 2024-05-23
Refactor tests.py and small tweaks for running on nix

fidgetingbits:
I was too hasty in my accepting feedback for #2168 and it still doesn't work on nix. There was actually a reason I was assigning to `GDB_INIT_PATH`, but it was kind of confusing anyway so since I had to go back in there I figured I'd refactor it a bit. This eliminate most of the globals, and the one left is lower case (as per @gsingh93 requested for tests/utils.py in another PR). Now gdbinit_path is a variable they gets passed around, but it should be more clear.

Also fix a shebang I missed in my other PR

### DIFF
--- a/tests.sh
+++ b/tests.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # Run integration tests
 (cd tests/gdb-tests && python3 tests.py $@)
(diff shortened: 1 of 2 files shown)


gsingh93:
Thanks for cleaning this up!

PR #2182, opened 2024-05-23
Fix and re-enable unit tests

gsingh93:
- Add `gdb_version` to unit tests mock to fix broken tests
- Fix broken test in `test_which.py`
- Update test script to run the unit tests and optionally collect coverage
- Correctly setup/use venv in tests and Dockerfile so correct `coverage`/`pytest` binaries are found

### DIFF
--- a/Dockerfile.arch
+++ b/Dockerfile.arch
@@ -17,6 +17,7 @@ ENV PIP_NO_CACHE_DIR=true
 ENV LANG en_US.utf8
 ENV TZ=America/New_York
 ENV ZIGPATH=/opt/zig
+ENV PWNDBG_VENV_PATH=/venv
 
 RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && \
     echo $TZ > /etc/timezone && \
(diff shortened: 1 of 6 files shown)


Dockerfile.arch
```diff
@@ -43,9 +44,11 @@ RUN rm README.md && rm -rf pwndbg
...
 RUN ./setup-dev.sh
 
+ADD . /pwndbg/
+
+ENV PATH="${PWNDBG_VENV_PATH}/bin:${PATH}"
+
 RUN echo "source /pwndbg/gdbinit.py" >> ~/.gdbinit.py && \
     echo "PYTHON_MINOR=$(python3 -c "import sys;print(sys.version_info.minor)")" >> /root/.bashrc && \
     echo "PYTHON_PATH=\"/usr/local/lib/python3.${PYTHON_MINOR}/dist-packages/bin\"" >> /root/.bashrc && \
     echo "export PATH=$PATH:$PYTHON_PATH" >> /root/.bashrc
```
gsingh93, line 54: I think this stuff should be done with `ENV`, but I'm not going to touch it.

tests/unit-tests/test_which.py
```diff
@@ -21,7 +21,7 @@
 
 
 def test_basic():
-    assert which("ls") == "/bin/ls"
+    assert which("ls") in ["/usr/bin/ls", "/usr/sbin/ls", "/bin/ls"]
```
fidgetingbits: This assert will fail on nix. Solution would be to check if the path which returns ends with a match in that list and don't assume the whole path is a match.

CptGibbon (approved):

PR #2185, opened 2024-05-24
Add cached symbol value lookup

fidgetingbits:
I'm not entirely sure this is that useful, but I had added it when porting stuff so figured I'd send a PR anyway. I realized after you can do the same with the existing `symbol.parse_and_eval`, which handles gdb expression evaluation. The handling for strictly `gdb.Value` lookup in `symbol.address` seemed slightly more robust, and I figured possibly having caching for when would be good, though I'm not familiar enough with how `@pwndbg.lib.cache` works to say how much of a difference it will make if used.

### DIFF
--- a/pwndbg/gdblib/symbol.py
+++ b/pwndbg/gdblib/symbol.py
@@ -137,6 +137,21 @@ def get(address: int | gdb.Value, gdb_only: bool = False) -> str:
 
 
 @pwndbg.lib.cache.cache_until("objfile")
+def value(symbol):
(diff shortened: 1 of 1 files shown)


disconnect3d:
How do we know if we need that? Can we see if GDB does not perform such caching on its own?

I mean, is this faster eventually?

fidgetingbits:
Ya I dunno. I haven't had time to check if they cache. Probably just okay to close for now and I'll change the one call in the musl heap PR

PR #2191, opened 2024-05-27
add vmmap --gaps option

fidgetingbits:
The following adds a new `--gaps` option to `vmmap` command which can be used to figure out where there are holes in the process memory map.

The following is an example of what it looks like when run against the included test case:

Screenshot from 2024-05-27 14-04-06

This also adds a new color for guard pages into the memory module, but because actually adding that to the vmmap (not --gaps) legend would break most of the existing tests, it something that will have to wait to add another time.

### DIFF
--- a/pwndbg/commands/vmmap.py
+++ b/pwndbg/commands/vmmap.py
@@ -5,6 +5,7 @@ Command to print the virtual memory map a la /proc/self/maps.
 from __future__ import annotations
 
 import argparse
+from typing import Tuple
 
 import gdb
 from elftools.elf.constants import SH_FLAGS
@@ -14,8 +15,12 @@ import pwndbg.color.memory as M
 import pwndbg.commands
 import pwndbg.gdblib.elf
 import pwndbg.gdblib.vmmap
+from pwndbg.color import cyan
+from pwndbg.color import green
+from pwndbg.color import red
 from pwndbg.commands import CommandCategory
 from pwndbg.gdblib import gdb_version
+from pwndbg.lib.memory import Page
 
 integer_types = (int, gdb.Value)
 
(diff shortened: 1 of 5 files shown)


fidgetingbits:
I just realized this is using bold colors which isn't in yet, so I'll remove. I also forgot to run lint again, which I'll fix. I really need to add that to the nix dev shell...

pwndbg/commands/vmmap.py
```diff
@@ -44,6 +49,100 @@ def print_vmmap_table_header() -> None:
...
+    """
+    Prints the table header for the vmmap --gaps command.
+    """
+    header = [f"{'Start':>{2 + 2 * pwndbg.gdblib.arch.ptrsize}}"]
+    header.append(f"{'End':>{2 + 2 * pwndbg.gdblib.arch.ptrsize}}")
+    header.append(f"{'Perm':>4}")
+    header.append(f"{'Size':>8}")
+    header.append(f"{'Note':>9}")
+    header.append(f"{'Accumulated Size':>{2 + 2 * pwndbg.gdblib.arch.ptrsize}}")
+    print(" ".join(header))
```
disconnect3d: You can construct a multi-line f-string like this:
```py
In [1]: x=1

In [2]: y=2

In [3]: s = (
   ...:     f'{x}\n'
   ...:     f'{y}\n'
   ...:     f'That works'
   ...: )

In [4]: print(s)
1
2
That works
```

Let's do that here; we don't really need to construct list/append/join :)

pwndbg/commands/vmmap.py
```diff
@@ -44,6 +49,100 @@ def print_vmmap_table_header() -> None:
...
+
+
+def calculate_total_memory(pages: Tuple[Page, ...]) -> None:
+    total = 0
+    for page in pages:
+        total += page.memsz
+    if total > 1024 * 1024:
+        print(f"Total memory mapped: {total:#x} ({int(total/1024/1024)} MB)")
```
disconnect3d: U can do `2//4` instead of `int(2/4)`

disconnect3d (approved):

gsingh93:
Does it makes sense to put all the logic for this in a separate code path from the normal `vmmap` printing logic? I feel like we're duplicating the logic here, and if we update some code in one code path we need to remember to update it in the other.

It seems to me the two code paths could be merged fairly easily, just take the `--gaps` code as the "full" output, and `if not gaps` don't display "Note" column and the "GAP" lines. Or is there a reason to not do this and keep it separate?

fidgetingbits:
> Does it makes sense to put all the logic for this in a separate code path from the normal `vmmap` printing logic? I feel like we're duplicating the logic here, and if we update some code in one code path we need to remember to update it in the other.
> 
> It seems to me the two code paths could be merged fairly easily, just take the `--gaps` code as the "full" output, and `if not gaps` don't display "Note" column and the "GAP" lines. Or is there a reason to not do this and keep it separate?

The LEGEND coloring is different for one, as in this case the goal is to see mapped/unmapped/guards only. Also the --gaps output purposefully doesn't show the file or the offset, so it's not quite that you could use gaps output as full. I found it way too busy/wide with everything included. 

I am already reusing the printing logic that vmmap uses from M.get, and just tweaking the output, so I guess the semi-duplicated parts are printing a (different) legend and header, printing the mapped region (but purposefully with fewer colors), and looping over the pages.

I do think the guard check/coloring should be integrated into M.get eventually, but as I mentioned it breaks tons of test because of the LEGEND changing, so I think it should done separately. Also because the colors are purposefully simplified for --gaps, it's likely it would have to be recolored anyways.

Originally I proposed on discord having a different command entirely, but others liked the idea of it being part of vmmap. 

With all that said, if you still prefer it gets merged into the main vmmap loop I'm open to it and I'll give it a whirl.

gsingh93:
> Originally I proposed on discord having a different command entirely, but others liked the idea of it being part of vmmap.

I definitely agree it should be part of the original vmmap command.

> With all that said, if you still prefer it gets merged into the main vmmap loop I'm open to it and I'll give it a whirl.

I don't want to block this PR for this, but eventually I think a lot of the logic should be merged. It may require a more invasive rewrite of the existing code to make it not a mess of spaghetti code when we do that merge, but I don't want to sign you up for that right now 🙂

gsingh93 (approved):

PR #2194, opened 2024-05-27
Add addsymbol command

chrf01:
- add command for adding temporary symbols

TODO:
- add some form of caching to avoid re-adding all symbols after restarting gdb. store symbols in `/tmp/symbols-<hash(binary_name)>.debug` instead of recreating symbol files repeatedly.
- we should build the elf in memory instead of using `objcopy` and `gcc`

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -653,6 +653,7 @@ def HexOrAddressExpr(s: str) -> int:
 
 def load_commands() -> None:
     # pylint: disable=import-outside-toplevel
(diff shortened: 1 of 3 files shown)


disconnect3d:
@chrf01 I think we should create an API for that, in case if users would like to use such functionality themselves in their scripts.

disconnect3d:
@chrf01 I am going to close this since its a bit stale and we now have a debugger agnostic abstraction: aglib.

Any chance you can rewrite this for `aglib` and/or finish it?

PR #2203, opened 2024-05-31
Auto-update dependencies after a git pull

### LINKED ISSUE #2138: Auto-update dependencies after a `git pull`
After doing a `git pull`, it's possible that there were changes to system dependencies in `setup.sh` or Python dependencies in `pyproject.toml`/`poetry.lock` that need to be updated. If those dependencies don't get updated, things may work fine, or things may break in subtle and hard to reproduce ways. Python dependency updates are much more common than `setup.sh` updates, so I think this issue should just focus on that.

On startup, `pwndbg` should check for a file called `$PWNDBG_VENV_PATH/poetry.lock.hash` or something similar. If it doesn't exist it should create it with with the hash of `poetry.lock` and run `poetry install`. If it does exist, it should check if the hash in that file matches the actual hash of `poetry.lock`. if they differ, run `poetry install` and update the hash stored in `poetry.lock.hash`.

For developers, we need to run `poetry install -with dev` to update dev dependencies. We could maybe store an additional file in `$PWNDBG_VENV_PATH` marking the user as a developer (created in `setup-dev.sh`).

The one issue with this is that both normal and dev dependencies are stored in `poetry.lock`, so if I update a dev dependency the `poetry.lock` hash will change and `poetry install` will still get run, even though it wouldn't update anything for regular users. This won't happen very often, and `poetry install` is pretty fast anyway, so maybe this won't be a problem.

B1N4RY-P4R45173:
I updated the `gdbinit.py` code which checks for stored hash value, and compares it with current one. Based on the result appropriate `poetry install` command will be executed to install the required dependencies (if any) or completely skipped if the hashes match. I also updated the `setup-dev.sh` so as to add an additional file to mark the installation via `setup-dev.sh`. This Fixes #2138


B1N4RY-P4R45173:
Hi everyone,

I'm facing an issue with the QEMU test in the CI checks for this PR. The test is failing, and I'm not sure how to resolve the problem. Can someone pls guide me?

CptGibbon:
Thanks for this PR 🙏
I've added Gulshan as a reviewer since #2138 is theirs.

If you want, you can mention the issue number you're addressing in your PR using one of several keywords and GitHub will automatically link them together for you.

https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

B1N4RY-P4R45173:
Alright thankyou. Yeah I will mark the issue from next PR, I thought writing the issue number in title addresses that. But ig apparently we need to mention it in the content. Thanks for the github documentation I will go through it rn.

B1N4RY-P4R45173:
Also I think I can remove the `hashlib` module, and use `subprocess` module to get sha256sum. This could eliminate the dependency on `hashlib`  module. Lemme know If I should change that as well.

CptGibbon:
> Also I think I can remove the `hashlib` module, and use `subprocess` module to get sha256sum. This could eliminate the dependency on `hashlib` module.

I believe `hashlib` is built-in so probably not a drama.
You could consider using `hashlib.file_digest` to make things more succinct if you want.

B1N4RY-P4R45173:
Yeah Thank you, I wasn't aware of the hashlib.file_digest function, which does make the code more succinct and efficient. I'll go ahead and use hashlib for the hash calculations as you suggested. Thanks again for the helpful advice!

CptGibbon:
If you want to avoid manually fixing lint & formatting issues, you could consider using a code editor that supports dev containers (e.g. vscode)

pwndbg is set up so that you can just clone it and the appropriate lint & format extensions are automatically installed. They'll run every time you save a file 👍

B1N4RY-P4R45173:
Thank you for the suggestion!
I am currently using automatic linting, so I'm not manually addressing linting issues. Instead, I commit the changes made by the linter and push them.

For reference:
image

If there's anything I'm doing incorrectly or if there's a more efficient way, please let me know. This is my first time contributing to an open source project, and I understand there's a lot for me to learn to make valuable contributions.

Thank you again for your guidance!

B1N4RY-P4R45173:
I wanted to follow up on my earlier message regarding the issue with the QEMU test in the CI checks for this PR. I understand that @gsingh93 might be busy, but I could really use some guidance on resolving this problem. If anyone could provide me with some guidance or references to troubleshoot this issue myself, I would greatly appreciate it.

Thank you for your assistance!

gdbinit.py
```diff
@@ -16,8 +18,46 @@
...
+
+# verify virtual environment path exists
+os.makedirs(PWNDBG_VENV_PATH, exist_ok=True)
+
+current_hash = calculate_hash(POETRY_LOCK_PATH)
+if os.path.exists(POETRY_LOCK_HASH_PATH):
+    with open(POETRY_LOCK_HASH_PATH, "r") as f:
+        stored_hash = f.read().strip()
+else:
+    stored_hash = None
```
gsingh93:
```suggestion
stored_hash = None
if os.path.exists(POETRY_LOCK_HASH_PATH):
    with open(POETRY_LOCK_HASH_PATH, "r") as f:
        stored_hash = f.read().strip()
```

gdbinit.py
```diff
@@ -16,8 +18,46 @@
...
+    subprocess.run(command, check=True)
+
+
+PWNDBG_VENV_PATH = os.getenv("PWNDBG_VENV_PATH")
+POETRY_LOCK_PATH = os.path.join(os.path.dirname(__file__), "poetry.lock")
+POETRY_LOCK_HASH_PATH = os.path.join(PWNDBG_VENV_PATH, "poetry.lock.hash")
+DEV_MARKER_PATH = os.path.join(PWNDBG_VENV_PATH, "dev.marker")
+
+# verify virtual environment path exists
+os.makedirs(PWNDBG_VENV_PATH, exist_ok=True)
```
gsingh93: I think we can get rid of this if we move this code below the `if not os.path.exists(venv_path):` check at line 71.

gdbinit.py
```diff
@@ -16,8 +18,46 @@
...
+
+
+def run_poetry_install(dev=False):
+    command = ["poetry", "install"]
+    if dev:
+        command = ["poetry", "install", "--with", "dev"]
+    subprocess.run(command, check=True)
+
+
+PWNDBG_VENV_PATH = os.getenv("PWNDBG_VENV_PATH")
```
gsingh93: Looks like the `venv_path` variable already exists, let's us that instead of making our own. I also think lower case variables are fine for all these names, as they're not constants.

gsingh93 (changes requested):
Overall this looks good, but needs a few minor changes. Primarily we should move this into the else branch of the if statement below that checks for SKIP_VENV.

Also, could you put the top-level code you added into a function like `update_deps`? I think that'll be cleaner and we can just call it from inside that `else` branch.

gsingh93:
Not sure about the QEMU tests, it's possible it's related to the SKIP_VENV stuff but don't have time to dig into it now. After you make the requested changes if it's still failing I'll take a closer look.

BTW, could you edit the initial description text and add something like "Fixes #2138", and update the PR title to something like "Auto-update dependencies after a git pull"? By adding the "Fixes" line, this PR gets linked to the issue so it's more easily discoverable, and when I close the PR the issue will get closed automatically. And the title is the default commit message when I merge the PR, so if you set it to something more descriptive then I won't need to change it when merging.

B1N4RY-P4R45173:
Sure thanks for your guidance, I will update as u stated above and change the title as well. I wasn't sure if my approach was correct so i was a little hesitant to include words like this "closes" or "fixes"

gdbinit.py
B1N4RY-P4R45173: I am sorry. The duplication was unintentional. When integrating the new functionality for checking and updating dependencies, I mistakenly redefined `venv_path`, not realizing it was already declared.

B1N4RY-P4R45173:
Hi @gsingh93 
I have made the changes you requested in my latest commit. I have carefully reviewed your feedback and incorporated the necessary modifications. However, if I have accidentally overlooked any aspect or if further adjustments are required, please let me know, and I will update them accordingly.

Additionally, I am still facing an issue with the QEMU test, which is failing despite my efforts to resolve it. Could you kindly provide some guidance on how to get this test to succeed? Any references or suggestions you have would be greatly appreciated.

Thank you for your time and assistance.

gsingh93:
Thanks for the updates, this PR looks good now. Regarding the QEMU test, I'll try to look into it this weekend.

B1N4RY-P4R45173:
Thank you for your feedback. I'm glad to hear that the changes look good now.

Regarding the QEMU test failure, if you could provide me with some guidance or resources, I'd try resolve the issue.

Also, if you noticed any mistakes or unnecessary edits in my PR, please let me know. I want to learn how to contribute efficiently and effectively.

Looking forward to your response and thank you once again for your support!

gsingh93:
Ok, debugging the QEMU issues revealed a bigger issue: https://github.com/pwndbg/pwndbg/issues/2218

Turns out all the tests were failing, it just wasn't being reported properly. And the tests are failing because `hashlib.file_digest` is only available since Python 3.11, but we need to support Python 3.8, so you'll need to find an alternative approach for this.

However, once you fix that you'll probably run into this: https://github.com/pwndbg/pwndbg/issues/2219

I think we should fix #2219 first (in a separate PR), and once it's merged and the hashlib issue is fixed, we should be able to merge this PR. #2218 doesn't need to block this PR.

B1N4RY-P4R45173:
Okay thanks @gsingh93 for the reply, I am thinking of using the inbuilt sha256sum command to get hash as an alternative method. Please let me know if I should proceed with that. But since we have to close #2219 first, I will try to look into that as well.

once again thankyou.

gdbinit.py
```diff
@@ -18,6 +19,42 @@
...
+def calculate_hash(file_path):
+    result = subprocess.run(["sha256sum", file_path], capture_output=True, text=True)
+    checksum = result.stdout.split()[0]
+    return checksum
+
+
+def run_poetry_install(dev=False):
+    command = ["poetry", "install"]
+    if dev:
+        command = ["poetry", "install", "--with", "dev"]
```
disconnect3d:
```suggestion
        command.extend(("--with", "dev"))
```

gdbinit.py
```diff
@@ -18,6 +19,42 @@
 
 # Get virtualenv's site-packages path
 venv_path = os.environ.get("PWNDBG_VENV_PATH")
+
+
+def calculate_hash(file_path):
+    result = subprocess.run(["sha256sum", file_path], capture_output=True, text=True)
```
disconnect3d: We should probably use hashlib instead of computing it via external tool that may not be present
```
In [1]: import hashlib

In [2]: hashlib.sha256?
Signature: hashlib.sha256(string=b'', *, usedforsecurity=True)
Docstring: Returns a sha256 hash object; optionally initialized with a string
Type:      builtin_function_or_method

In [3]: x = hashlib.sha256()

In [4]: x.update(b'asdf')

In [5]: x.hexdigest()
Out[5]: 'f0e4c2f76c58916ec258f246851bea091d14d4247a2fc3e18694461b1816e13b'
```

gsingh93:
Hmm, looking at the failing test, I'm wondering whether the best solution is to update the test, or to use `poetry -q install`? Or, maybe there's a way to quietly check if dependencies need an update first, and then only run `poetry install` if it would actually update anything? That way the user knows when updates happen, but there's no useless message when no updates happen.

EDIT: maybe `--quiet --dry-run` could be used first, and then only we install if anything actually needs to be done.

B1N4RY-P4R45173:
@gsingh93 Is this good to go? Also please check through the logs (if possible). I wrote similar code 3 times and each time it showed a different test failure.... Not sure why. I can understand lint failing (bcz I forget to run `./lint.sh` before `git push`)but sometimes `Unit tests / tests (ubuntu-20.04)` was failing.

gsingh93:
I mentioned the test failure in my last comment: https://github.com/pwndbg/pwndbg/pull/2203#issuecomment-2163450625

I don't know why it's not failing in the latest run, but I will take a closer look in the next day or so. But other than that this PR is looking good and should definitely be merged sometime this week. Thanks for sticking with it, it seems it turned out a bit more complicated than I originally thought it was going to be 🙂

B1N4RY-P4R45173:
> I mentioned the test failure in my last comment: [#2203 (comment)](https://github.com/pwndbg/pwndbg/pull/2203#issuecomment-2163450625)
>

Yes I have gone through that but I didn't get the part where you were mentioning how  we could quietly check for the update.  I read about `poetry install -q` but it would still print the essential progress. But if we truly want no logs, we can redirect the standard output via `/dev/null`. And also I am still reading about `--quiet --dry-run`.  Will definitely check how to use it.  

> I don't know why it's not failing in the latest run, but I will take a closer look in the next day or so. But other than that this PR is looking good and should definitely be merged sometime this week. Thanks for sticking with it, it seems it turned out a bit more complicated than I originally thought it was going to be 🙂

Thankyou!!!😊 I'm glad to hear that the PR is looking good overall. I'll keep an eye out for any further updates or feedback from your end.

Thank you for your guidance throughout this process, and I appreciate your patience and support.

gsingh93:
Ok, for reference, the failing job was this one: https://github.com/pwndbg/pwndbg/actions/runs/9486354625/job/26140499565

`test_loads_pure_gdb_without_crashing` was failing because it wasn't expecting to see the poetry output when GDB was run. You can't reproduce it locally with `rm .venv/poetry.lock.hash && ./tests.sh test_loads_pure_gdb_without_crashing`, because the tests script will invoke GDB once to collect all the test names, and then again to actually execute them, and this message will only happen the first time when the `poetry.lock.hash` file gets created. Why it somehow happened on the CI is a mystery to me 🙃 

Anyways, it would be nice if we changed this behavior so the user was saw the update information if updates were actually installed, and didn't see any output if no updates were installed, but this doesn't seem easy to do with poetry. It's common for `--dry-run` to return a different exit code if any changes would be made, but poetry doesn't seem to follow that convention. And `--quiet` suppresses all output, and I don't like the idea of silently installing stuff.

So the current approach where the user will see the "No dependencies to install or update" message the first time they run GDB after this PR, and in the case where `poetry.lock` changes but there are no updates for the user (i.e. a dev dependency changing will change `poetry.lock`, but a regular user won't have anything new to install) is fine, as these cases are not very common.

gsingh93 (approved):
LGTM. @disconnect3d go ahead and merge if the current version looks good to you as well.

gdbinit.py
```diff
@@ -18,9 +20,50 @@
...
+                break
+            file_hash.update(chunk)
+    return file_hash.hexdigest()
+
+
+def run_poetry_install(dev=False):
+    command = ["poetry", "install"]
+    if dev:
+        command.extend(("--with", "dev"))
+    subprocess.run(command, check=True)
```
B1N4RY-P4R45173:
```suggestion
    result = subprocess.run(command, capture_output=True, text=True)
    return result.stdout.strip(), result.stderr.strip(), result.returncode
```

gdbinit.py
```diff
@@ -18,9 +20,50 @@
...
+    if os.path.exists(poetry_lock_hash_path):
+        with open(poetry_lock_hash_path, "r") as f:
+            stored_hash = f.read().strip()
+
+    # checks if dev.marker exists
+    dev_mode = os.path.exists(dev_marker_path)
+
+    # if hashes don't match, run the appropriate command based on dev.marker file
+    if current_hash != stored_hash:
+        run_poetry_install(dev=dev_mode)
```
B1N4RY-P4R45173:
```suggestion
        stdout, stderr, returncode = run_poetry_install(dev=dev_mode)
```

gdbinit.py
```diff
@@ -18,9 +20,50 @@
...
+        with open(poetry_lock_hash_path, "w") as f:
+            f.write(current_hash)
+
```
B1N4RY-P4R45173, line 67:
```suggestion
if returncode == 0 and ("No dependencies to install or update" not in stdout):
            print(stdout)
        elif returncode != 0:
            print(stderr, file=sys.stderr)
```

B1N4RY-P4R45173:
> Anyways, it would be nice if we changed this behavior so the user was saw the update information if updates were actually installed, and didn't see any output if no updates were installed, but this doesn't seem easy to do with poetry. It's common for `--dry-run` to return a different exit code if any changes would be made, but poetry doesn't seem to follow that convention. And `--quiet` suppresses all output, and I don't like the idea of silently installing stuff.
> 
I haven't tested the above changes. I think these changes could address this issue. Now it would print the output only if dependencies are installed, and would suppress the output  if no dependency is installed. 

This is what stdout contains when u run `poetry install` and there is nothing to update or install:

```
stdout: Installing dependencies from lock file

No dependencies to install or update

Installing the current project: pwndbg (2024.02.14)
```

B1N4RY-P4R45173:
Hello @gsingh93 I understand you might be busy. But if you could spare a minute, please let me know what do u think of the above changes. I everything looks fine to you I will commit the changes. Thank you.

disconnect3d (approved):
Lgtm. @gsingh93 merge if u agree :)

Issue #2014, opened 2024-02-06
Google Summer of Code 2024 project ideas


disconnect3d:
Below I list Pwndbg organization project ideas for Google Summer of Code 2024.

**Resolution / EDIT2: Pwndbg is a sub-org for the Python organization in GSoC 2024**

Most of the projects here are only about Pwndbg and would require the contributor to write Python code and send PRs to this repository. However, since Pwndbg is a GDB plugin, it would benefit from improving GDB as well. Because of this, we may list a project idea about GDB - which is written in C - and which would require sending patches to it instead of Pwndbg. We would of course mentor these project ideas as well if a contributor decides to do that.

Unless otherwise stated, the projects require knowledge of Python programming language, GDB debugging and a little bit of C or C++.

### Description
When researching Linux kernel bugs it is useful to have an insight into the Linux kernel SLAB/SLUB allocator[s]. 

A part of this is already implemented as a `slab` command in Pwndbg. We want this extended further with more features by integrating or porting the features from https://github.com/nccgroup/libslub into Pwndbg.

For what is worth, we have an established chat with libslub developers and they would help mentor this project if its choosen.

Ideally this would be done in such a way that the new module could be used as a mostly standalone tool (for instance to allow use by gef users), while still leveraging pwndbg API. Ideally the module would also be fully integrated into pwndbg as a regular module for pwndbg users. There is already a WIP fork that partially integrates into pwndbg, however completing this work would also involve modularizing core parts of pwndbg to make it possible to work standalone without conflicting with other plugins. The existing WIP fork also adds significant number of kernel tests, which could be further improved. 

### Project information
* Length: ~175 hours
* Difficulty: Medium
* Mentors: @fidgetingbits, @disconnect3d, @gsingh93 

### Expected Outcome
* Integration or port of nccgroup/libslub project features into Pwndbg
  * Extend the `slab {list,info,contains}` commands to support use cases available in libslub

### Description 
Our users sometimes debug Linux kernel without debug symbols, e.g., during a CTF security competition or when debugging a kernel dumped from an embedded device.

Most Linux kernels keep track of its symbol addresses and allows reading them from the `/proc/kallsyms` file. This means that the symbol information is somewhere in the Linux kernel memory and can be fetched with us to provide better UX.

This loading should be implemented as a `kallsyms` command. 

### Project information
* Length: ~175 hours
* Difficulty: Medium
* Mentors: @disconnect3d, @fidgetingbits, @gsingh93 

[...]

PR #2204, opened 2024-06-01
Port Pwndbg to LLDB

mbrla0:
This draft tracks the effort of porting Pwndbg to LLDB, as part of the [GSoC 2024 Project](https://summerofcode.withgoogle.com/programs/2024/projects/BuvGmK4D), as outlined in #2014. 

Obviously, expect it to change and get more flushed out as the project happens.


mbrla0:
As of right now I believe we have an interesting view of what Step 0 in the porting process looks like. 

It moves `pwndbg.heap` and `pwndbg.disasm` into `pwndbg.gdblib`, as well as takes `pwndbg.gdblib.config` out of `gdblib` and into the root module - with a few quick hacks to decouple `config` from `gdb.PARAM_*`, so as to disentangle uses of the `config` object from `gdblib`. It also adds a draft debugger-agnostic API under `pwndbg/dbg`, which right now only holds what was absolutely necessary to bring up Pwndbg under LLDB, while not breaking it under GDB.

I feel it's also important to note that - despite my efforts to minimize breakage as much as I felt was reasonable for an early draft - some stuff might have gone by unnoticed by me, and so might still be broken.

gsingh93:
Thanks for putting up this draft!

Could you please split this up into multiple PRs? For example, your first three commits (creating the Dockerfile, moving heap/disasm into gdblib, and modifying .gitignore) can be three completely independent PRs. After that there are probably additional ways that this can be split, but the benefit of this is that each of those individual PRs can be merged faster because they will be very quick to review, we can independently verify all tests pass for it, if something goes wrong we can easily revert it, and getting big refactoring changes like the disasm/heap move in early mean you and other contributors will have less issues with rebasing/merging in future PRs. And once those are merged, the resulting PRs that actually contain the complexity that needs a closer review will be much easier to read.

mbrla0:
Alright, that makes sense. I've spun off the vim change (which was actually a bit of a mistake on my part, I hadn't actually intended to commit that line at all, much less have it as part of the draft) and the `pwndbg/{heap,disasm} -> pwndbg/gdblib` change into their own branches, and I'm going to create PRs for them soon.

Just one thing, though, in regards to the LLDB container change, and further splitting the draft, in its current form, into other PRs. While I do understand the motivation behind it, I fear splitting the work too much might end up bogging things down, since I feel it might hit things that are still work-in-progress and, therefore very much still in a state of flux, and end up locking us into things we don't want, or having us spend too much time squashing bugs and fixing tests for a feature that we might decide against keeping later on.

I feel like a hybrid approach might make more sense. One in which we spin chunks of the work off from this draft as they both start to take shape, and make sense to commit as "whole" chunks of work - i.e ones that won't bork `dev`. While leaving the branch that's tracked by this PR as the place where we don't need to worry about that, and where one can test the work locally to their heart's content. Whenever the spun off PRs get merged, we can probably just rebase the main branch to reflect that.

Do you think this could make sense as an approach?

gsingh93:
> While I do understand the motivation behind it, I fear splitting the work too much might end up bogging things down, since I feel it might hit things that are still work-in-progress and, therefore very much still in a state of flux, and end up locking us into things we don't want, or having us spend too much time squashing bugs and fixing tests for a feature that we might decide against keeping later on.

You make a good point, but we need to be careful to strike a balance here. Using the Dockerfile as one example, we could leave it in here, and I wouldn't mind too much. But there's also benefit in committing it as it is for now(as long as it actually works), and if some stuff changes later and we need to fix it in an additional PR, it's actually nice to have that PR with just the edits, because it's clear to everyone in the project and the next person who checks the blame for that file the motivation for making certain changes. But if you think there's a good chance that it's going to go through a major overhaul, then it makes sense to wait a little while longer before splitting it into its own PR. You get to make the final call about this, I Just wanted to make sure you understand how I think about the rationale behind one approach vs. the other 🙂 

> I feel like a hybrid approach might make more sense. One in which we spin chunks of the work off from this draft as they both start to take shape, and make sense to commit as "whole" chunks of work - i.e ones that won't bork dev. While leaving the branch that's tracked by this PR as the place where we don't need to worry about that, and where one can test the work locally to their heart's content. Whenever the spun off PRs get merged, we can probably just rebase the main branch to reflect that.

We can try this and see how it goes. If there's just some quick thing you want to check by us you can put the current WIP here and tell us exactly where to look. But if you want us to review something bigger we might just ask you to split off a PR at that point.

CptGibbon:
> I fear splitting the work too much might end up bogging things down

With all due respect this is a pretty wild take, and is contrary to best practices like test-driven development and continuous integration.

That said, I totally get what you're suggesting with the hybrid approach here and I've encouraged a similar approach for our jemalloc GSoC contributor so I can't complain 🤷‍♂️

Also this is great work so please don't let my comments take the wind out of your sails heh.

mbrla0:
> > I fear splitting the work too much might end up bogging things down
> 
> With all due respect this is a pretty wild take, and is contrary to best practices like test-driven development and continuous integration.

I might've expressed myself poorly there. It's not really a full-on take, and more of just a concern, and it mostly just boils down to "it might make sense to bundle some things together, so we don't end up breaking anything or too eagerly splitting off things we know might change drastically very soon". 

Ultimately, though, I'm just throwing the idea out there and asking for input and what you guys think.

Might also even be that we have roughly the same thing in mind, and I've just not been clear enough about what I'm thinking.

mbrla0:
Either way, I should get on to fixing the tests for Step 0.

pwndbg/dbg/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+        raise NotImplementedError()
+
+    # WARNING
+    #
+    # These are hacky parts of the API that were strictly necessary to bring up
+    # pwndbg under LLDB without breaking it under GDB. Expect most of them to be
+    # removed or replaced as the porting work continues.
+    #
+
+    def addrsz(self, address):
```
gsingh93: Can you make sure to add type hints to everything as you work on this? I promise it will help out in the long run 🙂

pwndbg/dbg/gdb.py
```diff
@@ -0,0 +1,113 @@
...
+import pwndbg.commands
+import pwndbg.gdblib
+from pwndbg.commands import load_commands
+from pwndbg.gdblib import gdb_version
+from pwndbg.gdblib import load_gdblib
+from pwndbg.gdblib import prompt
+
+
+class GDB(pwndbg.dbg_mod.Debugger):
+    def setup(self):
```
gsingh93, line 19: When overriding methods, use `@override` (`from typing_extensions import override`)

mbrla0:
Closed in favor of other PRs. This was tracking the wrong fork branch anyway. If you're looking for the PR that actually ended up porting Pwndbg to LLDB, it's #2382.

PR #2205, opened 2024-06-03
Future syscalls

OBarronCS:
This PR moves some logic that determines the syscall number for an instruction into the enhancement process, allowing us to use emulation to determine the values and names of syscalls that occur in the future.

Previously, we only determined the syscall number if the instruction pointer was on the syscall instruction itself. Now, it is determined and displayed in advance with emulation, so you can see the name of the syscall before you step to it.

Before:
syscall_before

After:
after_syscall


pwndbg/disasm/arch.py
```diff
@@ -354,9 +357,16 @@ def _parse_immediate(
     def _read_register(
         self, instruction: PwndbgInstruction, operand_id: int, emu: Emulator
     ) -> int | None:
-        # operand_id is the ID internal to Capstone
+        """
```
gsingh93: Why did you change this comment into a docstring?

pwndbg/disasm/arch.py
```diff
@@ -185,6 +185,13 @@ def enhance(instruction: PwndbgInstruction, emu: Emulator = None) -> None:
                     )
                 emu = jump_emu = None
 
+        enhancer: DisassemblyAssistant = DisassemblyAssistant.assistants.get(
+            pwndbg.gdblib.arch.current, generic_assistant
+        )
+
+        # Don't disable emulation yet, as we can use it to read the syscall register
+        enhancer._enhance_syscall(instruction, emu)
```
gsingh93: It's weird codecov is reporting this line as not covered by tests. Does tests/gdb-tests/tests/test_context_commands.py not exercise this line of code?

(You can see the codecov message in the files tab: https://github.com/pwndbg/pwndbg/pull/2205/files)

pwndbg/disasm/arch.py
```diff
@@ -503,6 +513,37 @@ def _telescope_format_list(self, addresses: List[int], limit: int, emu: Emulator
...
+        # but only `syscall` and `int 0x80` actually execute syscalls on Linux.
+        # So here, we return no syscall name for other instructions and we also
+        # handle a case when 32-bit syscalls are executed on x64
+        if syscall_register in ("eax", "rax"):
+            mnemonic = instruction.mnemonic
+
+            # We read .imm directly, because at this point we haven't enhanced the operands with values
+            is_32bit = mnemonic == "int" and instruction.operands[0].imm == 0x80
+            if not (mnemonic == "syscall" or is_32bit):
+                return None
```
gsingh93: I guess this should go in the `x86` specific disassembly assistant, right? It could do this check and either `return None` or call `super._enhance_syscall`.

pwndbg/disasm/arch.py
```diff
@@ -503,6 +513,37 @@ def _telescope_format_list(self, addresses: List[int], limit: int, emu: Emulator
...
+
+            # On x64 the int 0x80 instruction executes 32-bit syscalls from i386
+            # On x86, the syscall_arch is already i386, so its all fine
+            if is_32bit:
+                syscall_arch = "i386"
```
gsingh93: Maybe we add a method in `DisassemblyAssistant` to get the syscall arch, the default implementation returns `pwndbg.gdblib.arch.name`, but the x86 assistant does this? That way you can make this method in the base class arch independent (without using if statements).

pwndbg/disasm/arch.py
OBarronCS: No particular reason - it just adds the comment to an intellisense popup, potentially avoiding the need to jump to the function to see how the operand id is used.

pwndbg/disasm/arch.py
OBarronCS: Very strange - the function is definitely invoked, because a couple tests check for the syscall name on the context view which this function provides.

pwndbg/disasm/arch.py
OBarronCS: Good call - making this change.

pwndbg/disasm/arch.py
OBarronCS: That's clean - latest commit implements it this way.

pwndbg/disasm/arch.py
gsingh93: It's just that the comment above this function looks like it's supposed to be the actual docstring, so maybe we should merge them? Not a big deal.

pwndbg/disasm/arch.py
```diff
@@ -503,6 +513,26 @@ def _telescope_format_list(self, addresses: List[int], limit: int, emu: Emulator
...
+        if CS_GRP_INT not in instruction.groups:
+            return None
+
+        syscall_register = pwndbg.lib.abi.ABI.syscall().syscall_register
+        syscall_arch = self._get_syscall_arch(instruction)
+
+        instruction.syscall = self._read_register_name(instruction, syscall_register, emu)
+        if instruction.syscall is not None:
+            instruction.syscall_name = (
+                pwndbg.constants.syscall(instruction.syscall, syscall_arch)
```
gsingh93: `pwndbg.constants.syscall` takes a `str` for it's second argument, not a `str | None`, which is the type of `syscall_arch`. I'm wondering why `mypy` didn't complain... Anyways, I think it would be more clear to do a `if syscall_arch is None: return None` after line 521 even if it ends up working the way it is.

pwndbg/disasm/arch.py
```diff
@@ -503,6 +513,26 @@ def _telescope_format_list(self, addresses: List[int], limit: int, emu: Emulator
                 enhance_string_len=enhance_string_len,
             )
 
+    def _enhance_syscall(self, instruction: PwndbgInstruction, emu: Emulator) -> None:
+        if CS_GRP_INT not in instruction.groups:
+            return None
+
+        syscall_register = pwndbg.lib.abi.ABI.syscall().syscall_register
```
disconnect3d: Can this ever change? Shall we maybe cache this in the class itself?

pwndbg/disasm/x86.py
```diff
@@ -431,6 +431,28 @@ def _condition(self, instruction: PwndbgInstruction, emu: Emulator) -> Instructi
 
         return InstructionCondition.TRUE if bool(conditional) else InstructionCondition.FALSE
 
+    # Override
+    def _get_syscall_arch(self, instruction: PwndbgInstruction) -> str | None:
+        syscall_arch = pwndbg.gdblib.arch.name
```
disconnect3d: Random thought: do we really have to fetch this at this point if we know we are in x86-64?

Or does this support both x86 and x86-64? If so, pls add a comment about this (even if it is somewhere upper, heh)

pwndbg/disasm/x86.py
```diff
@@ -431,6 +431,28 @@ def _condition(self, instruction: PwndbgInstruction, emu: Emulator) -> Instructi
...
+
+        # We read .imm directly, because at this point we haven't enhanced the operands with values
+        is_32bit = mnemonic == "int" and instruction.operands[0].imm == 0x80
+        if not (mnemonic == "syscall" or is_32bit):
+            return None
+
+        # On x64 the int 0x80 instruction executes 32-bit syscalls from i386
+        # On x86, the syscall_arch is already i386, so its all fine
+        if is_32bit:
+            syscall_arch = "i386"
```
disconnect3d:
```suggestion
            return "i386"
```

disconnect3d:
Just to be sure, do we still display syscall name if we disable emulation and step on a syscall instruction?

I think that the changes we made here may have disabled that, so I am asking to confirm that.

Can we add a test for such a case?

pwndbg/disasm/arch.py
disconnect3d: Also: if u execute 32-bit syscall via int 0x80, does RAX get truncated to EAX?

Shouldn't we support this here somehow?

pwndbg/disasm/x86.py
```diff
@@ -431,6 +431,28 @@ def _condition(self, instruction: PwndbgInstruction, emu: Emulator) -> Instructi
 
         return InstructionCondition.TRUE if bool(conditional) else InstructionCondition.FALSE
 
+    # Override
```
gsingh93: As mentioned on discord, we can import `override` from `typing_extensions`

pwndbg/disasm/x86.py
OBarronCS: Correct - this class supports both x86 and x86_64. I added a note about this in this function as well as the before the class declaration.

pwndbg/disasm/arch.py
OBarronCS: Good point regarding the truncation of the RAX register in this case. Potentially we change the _get_syscall_arch to return a tuple of (arch,register_name) to fix this edge case?

OBarronCS:
We still display the syscall name in the case emulation is disabled and step on the instruction. The "_read_register_name" function will check if "self.can_reason_about_process_state" - meaning the program counter is at the instruction being executed - and will read the register directly. I added a small test that simply runs "set emulate off" and delegates to a pre-existing test.

https://github.com/OBarronCS/pwndbg/blob/842def8f356624de817f0a3997fc932db1be8ca7/pwndbg/gdblib/disasm/arch.py#L368-L387

pwndbg/disasm/arch.py
OBarronCS: In terms of caching, it would be safe to do so (for arches that can dynamically change the syscall architecture, we have to handle case-by-case anyways). That being said, this line gets called only whenever a "syscall" instruction is in the disasm view, which is relatively rare.

pwndbg/disasm/arch.py
OBarronCS: Done!

disconnect3d:
@OBarronCS I think we are ready to make a last review and/or merge it, right?

OBarronCS:
Yes! This is good to go :)

disconnect3d (approved):

disconnect3d:
@OBarronCS ooops
```
  File "/root/pwndbg/pwndbg/gdblib/disasm/__init__.py", line 217, in get_one_instruction
    pwndbg.gdblib.disasm.arch.DisassemblyAssistant.enhance(pwn_ins, emu)
  File "/root/pwndbg/pwndbg/gdblib/disasm/arch.py", line 220, in enhance
    if bool(pwndbg.gdblib.config.disasm_annotations):
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'pwndbg.gdblib.config' has no attribute 'disasm_annotations'
```

PR #2211, opened 2024-06-07
Fix dependency installation on openSUSE Tumbleweed

xtexx:
openSUSE Tumbleweed no longer includes python2-pip

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -49,7 +49,8 @@ install_swupd() {
 install_zypper() {
     sudo zypper mr -e repo-oss-debug || sudo zypper mr -e repo-debug
     sudo zypper refresh || true
-    sudo zypper install -y gdb gdbserver python-devel python3-devel python2-pip python3-pip glib2-devel make glibc-debuginfo
+    sudo zypper install -y gdb gdbserver python-devel python3-devel python3-pip glib2-devel make glibc-debuginfo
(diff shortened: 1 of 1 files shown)


setup.sh
```diff
@@ -49,7 +49,8 @@ install_swupd() {
 install_zypper() {
     sudo zypper mr -e repo-oss-debug || sudo zypper mr -e repo-debug
     sudo zypper refresh || true
-    sudo zypper install -y gdb gdbserver python-devel python3-devel python2-pip python3-pip glib2-devel make glibc-debuginfo
+    sudo zypper install -y gdb gdbserver python-devel python3-devel python3-pip glib2-devel make glibc-debuginfo
+    sudo zypper install -y python2-pip || true
```
disconnect3d:
```suggestion
    sudo zypper install -y python2-pip || true  # skip py2 installation if it doesn't exist
```

disconnect3d:
I wonder: why do we need py2 in the first place? Pwndbg should not need py2 at all...

We should probably drop this in another PR and see if it breaks anything.

gsingh93:
@disconnect3d this was merged with a failing lint check, can you fix in #2212?

disconnect3d:
@gsingh93 #2212 seems to need more changes since docker-compose dont have 240.04, ill open another PR to fix the lint, eh

PR #2212, opened 2024-06-07
Add Ubuntu 24.04 to CI

disconnect3d:
As @gsingh93 pointed out, the GH CI runner now supports Ubuntu 24.04, we should build & test on it as well.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


disconnect3d:
Fwiw Ubuntu 20.04 is still up until April 2025.

image

source: https://ubuntu.com/about/release-cycle

gsingh93:
We need to add it to https://github.com/pwndbg/pwndbg/blob/dev/.github/workflows/tests.yml.

We can add it to docker.yml, but we also need to update the docker compose file.

PR #2220, opened 2024-06-09
Update package installation from pip to poetry

walstper:
Hello, please take a look at updated lines and check the tool correctness with this changes.
(I have no experience with tool, so cannot validate by myself)


gsingh93:
> Hello, please take a look at updated lines and check the tool correctness with this changes. (I have no experience with tool, so cannot validate by myself)

We don't accept contributions like this. Validating changes is the responsibility of the PR author.

PR #2221, opened 2024-06-09
Using Poetry to install all dependencies 

### LINKED ISSUE #2219: Install all dependencies with poetry
We've almost entirely switched to poetry, except for the fact that we still do `pip install` in `setup.sh`. `pip install` will *not* use the poetry.lock file, and so currently everyone is running on different versions of the dependencies. We should:

1. Install poetry in setup.sh
2. Don't install pip
3. Convert the `pip install` to a `poetry install`
4. Remove the redundant installation of the poetry binary in `setup-dev.sh`, we shouldn't need to redo anything already done in `setup.sh`

B1N4RY-P4R45173:
This closes #2219. Updated the setup.sh and setup-dev.sh to use poetry installation instead of pip installation.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -171,6 +171,14 @@ if ! hash gdb; then
     exit
 fi
 
+# Find the Python version used by GDB.
+PYVER=$(gdb -batch -q --nx -ex 'pi import platform; print(".".join(platform.python_version_tuple()[:2]))')
+PYTHON+=$(gdb -batch -q --nx -ex 'pi import sys; print(sys.executable)')
+
+if ! osx; then
+    PYTHON+="${PYVER}"
+fi
+
 # Install Poetry
 if ! command -v poetry &> /dev/null; then
     echo "Poetry not found. Installing Poetry..."
--- a/setup.sh
+++ b/setup.sh
@@ -171,28 +171,23 @@ if ! hash gdb; then
     exit
 fi
 
-# Find the Python version used by GDB.
-PYVER=$(gdb -batch -q --nx -ex 'pi import platform; print(".".join(platform.python_version_tuple()[:2]))')
-PYTHON+=$(gdb -batch -q --nx -ex 'pi import sys; print(sys.executable)')
-
-if ! osx; then
-    PYTHON+="${PYVER}"
+# Install Poetry
+if ! command -v poetry &> /dev/null; then
+    echo "Poetry not found. Installing Poetry..."
+    curl -sSL https://install.python-poetry.org | python3 -
+    export PATH="$HOME/.local/bin:$PATH"
+else
+    echo "Poetry is already installed."
 fi
 
-# Create Python virtualenv
+# Create the Python virtual environment and install dependencies using poetry
 if [[ -z "${PWNDBG_VENV_PATH}" ]]; then
     PWNDBG_VENV_PATH="./.venv"
 fi
 echo "Creating virtualenv in path: ${PWNDBG_VENV_PATH}"
 
-${PYTHON} -m venv -- ${PWNDBG_VENV_PATH}
-PYTHON=${PWNDBG_VENV_PATH}/bin/python
-
-# Upgrade pip itself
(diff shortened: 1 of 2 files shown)


B1N4RY-P4R45173:
Okay so after a little debugging, I think the issue is with curl not being built in command in Debian systems, but not very sure about the arch. I will update the docker files once I find what is wrong with arch.

B1N4RY-P4R45173:
I need some help with docker builds, I thought the only issue was curl, but I think there are still some things that I don't get. The problem seems to be with usage and creation of venv. Did I miss something with the changes I did?

Thank you

gsingh93:
We can add curl as a dependency we install in `setup.sh`.

Can you also remove the `python-pip` and `python3-pip` dependencies that are being installed?

I see that the archlinux docker build is showing a different error with `/venv/bin/activate` not found. I also see:
```
#14 17.13 Creating virtualenv in path: /venv
#14 17.87 Creating virtualenv pwndbg in /pwndbg/.venv
```
which seems strange. It might have to do with the `in-project=true` setting in `poetry.toml`? Not sure.

Anyway, can you fix the debian/Ubuntu curl issue first? I want to double check we hit the same error message there. To debug the docker issue more, you can try something like adding an `ls -lR /venv` into the Dockerfile to confirm it's actually getting created properly. Maybe even try removing `poetry.toml` from the Dockerfile and see if that fixes the issue.

B1N4RY-P4R45173:
Now the error seems to be the same now that is `/venv/bin/activate` not found. I will be working on fixing that issue now

B1N4RY-P4R45173:
I believe creating the `/venv` is the problem here... Not sure what is wrong with it, Also removing `poetry.toml` doesn't seem to change anything.
Screenshot 2024-06-11 092113

This the dockerfile code I used while building it
image

gsingh93:
You're running `ls -lR` before you run `setup.sh`, so it's not going to exist :)

In any case, before we were doing:
```
${PYTHON} -m venv -- ${PWNDBG_VENV_PATH}
```

Now we're doing:
```
poetry config virtualenvs.path ${PWNDBG_VENV_PATH}
```

I don't think these two are the same thing, the first creates a venv in the specified directory, the latter just sets the base directory that poetry will use to create it's own venvs with. And because `in-project` is set to true in `poetry.toml`, I don't think it'll even use that path (it'll just try to create a .venv folder in the current directory).

I think what we can try is in `setup.sh`:
```
${PYTHON} -m venv -- ${PWNDBG_VENV_PATH}
source ${PWNDBG_VENV_PATH}/bin/activate
poetry install
```

So we manually create the venv first, activate it, and then poetry should just use it for the install (I think, I haven't tested it).

B1N4RY-P4R45173:
> You're running `ls -lR` before you run `setup.sh`, so it's not going to exist :)
>
Sorry for that, I realized it only after your comment.

Thankyou for the review and guidance, everything looks to be fine now. Please let me know if I should be changing anything more before the final merge.

Once again Thanks a lot.

disconnect3d (approved):
Lgtm, @gsingh93 shall we merge now?

B1N4RY-P4R45173:
btw should I change the comment here to poetry install would fail in docker fail
image

setup.sh
```diff
@@ -179,20 +179,24 @@ if ! osx; then
     PYTHON+="${PYVER}"
 fi
 
-# Create Python virtualenv
+# Install Poetry
+if ! command -v poetry &> /dev/null; then
+    echo "Poetry not found. Installing Poetry..."
+    curl -sSL https://install.python-poetry.org | python3 -
+    export PATH="$HOME/.local/bin:$PATH"
```
gsingh93, line 186: I'm wondering if instead of exporting `PATH` here we should instead do `POETRY="$HOME/.local/bin/poetry"`, and then use `$POETRY` below... I think exporting `PATH` might lead to other bugs, but I think it should be fine for now and we can always change it later.

gsingh93:
LGTM. Thanks for the contribution!

PR #2222, opened 2024-06-10
Disentangle early initialization from GDB

mbrla0:
This PR implements Step 0 in the effort of porting Pwndbg to LLDB, as outlined in #2204.

The changes in this PR move the global object used for configuration from `pwndbg.gdblib.config` into `pwndbg.config`, and adds a skeleton of the debugger-agnostic API under `pwndbg.dbg`. It also adds a separate `lldbinit.py` file, that allows Pwndbg to be started under LLDB.

Currently we only have a single simple command under LLDB, just to test out adding commands and to see how LLDB passes arguments from the command line to Python commands.

It doesn't change the prompt string yet, but that could easily be achievable with `SBDebugger.SetPrompt()`, which we can do later, in Step 1.


mbrla0:
This is a rather large change, but that's mostly because a lot of files use `pwndbg.gdblib.config`.

mbrla0:
Closed in favor of #2243 and #2244.

PR #2223, opened 2024-06-11
clear screen without clearing buffer

antkss:

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -72,7 +72,8 @@ def clear_screen(out=sys.stdout) -> None:
     # If n is 3, clear entire screen and delete all lines saved in the
     # scrollback buffer (this feature was added for xterm and is supported
     # by other terminal applications
-    out.write("\x1b[H\x1b[2J")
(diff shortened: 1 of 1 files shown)


disconnect3d:
This doesn't sound right:
1. The way context work currently, should not clear buffer when clearing screen: https://github.com/pwndbg/pwndbg/blob/d38e57639fe446449d564de852fc7b6d36b0c321/pwndbg/commands/context.py#L420-L423
2. This is controlled by the `context-clear-screen` config variable (e.g., add `set context-clear-screen on` in your `~/.gdbinit`; the default is `off`).
3. We would not want to call external programs just to write something to the output. Don't forget the context can be reconfigured to send the output (`context-output` parameter) to other terminal etc. Printing context via `os.system(printf ...)` would break this.
4. This PR does not change the comment above which helps to understand the reasoning beyond the bytes we send to the output currently ...

Anyway, terminals and screen buffering etc is not an easy topic, so if there is a bug you stumled upon and this does not work for you, please report a bug as a GitHub issue explaining your setup and case.

PR #2228, opened 2024-06-12
add feature to detect command pacman

### LINKED ISSUE #2227: Setup should detect apt and pacman instead of distros
As described in https://github.com/pwndbg/pwndbg/pull/2226#pullrequestreview-2111784359 instead of "supported distro" we can probably go with just detecting `apt` and `pacman` (and others?) in setup.sh and use them right away.

https://github.com/pwndbg/pwndbg/blob/dev/setup.sh#L121-L164

raffifu:
This PR close https://github.com/pwndbg/pwndbg/issues/2227

Add branching to detect command pacman exists in the machine.

Should i change this line also to only detect "arch"?
https://github.com/pwndbg/pwndbg/blob/20f8fe044c352f805954d3f65acfaf25ccb6facc/setup.sh#L136-L140

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -153,11 +153,13 @@ if linux; then
             install_freebsd
             ;;
         *) # we can add more install command for each distros.
-            echo "\"$distro\" is not supported distro. Will search for 'apt' or 'dnf' package managers."
+            echo "\"$distro\" is not supported distro. Will search for 'apt', 'dnf' or 'pacman' package managers."
             if hash apt; then
                 install_apt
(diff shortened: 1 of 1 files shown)


disconnect3d:
@gsingh93 This is probably good enough. I'm okay with merging this as it is. Or do you think we should drop the "distro checking" and/or first check for apt/pacman and then check for other distros (like gentoo)?

gsingh93:
I think this is good enough for now. We can clean up the distro checking in the future.

PR #2230, opened 2024-06-13
Start migrating to Python logging and add log_level command

gsingh93:
(Note: don't merge this until #2242 is merged. I've included that commit in here so that we know that the tests are actually passing).

Starts implementing #245.

I got tired of having to add/remove print statements when I should have just been able to set a higher/lower log level.

- Adds the `PWNDBG_LOGLEVEL` environment variable to set the initial log level (defaults to `logging.WARNING`, I might change this to `logging.INFO`)
- Adds a custom `Formatter` to the root logger, so that log messages look exactly like they do now when you use `print(message.warn())` or `print(message.error())`.
- Convert some `print` statements to proper logging calls in `commands/__init__.py`
- Add the `log_level` developer command to change the log level

Here's an example where with the default log level, warning and error messages are shown, but I can change the log level and only have errors shown:
image

### DIFF
--- a/gdbinit.py
+++ b/gdbinit.py
@@ -30,9 +30,10 @@ def hash_file(file_path: str | Path) -> str:
 
 
 def run_poetry_install(poetry_path: os.PathLike[str], dev: bool = False) -> Tuple[str, str, int]:
-    command: List[str | os.PathLike[str]] = [poetry_path, "install"]
+    command: List[str] = [str(poetry_path), "install"]
     if dev:
         command.extend(("--with", "dev"))
+    logging.debug(f"Updating deps with command: {' '.join(command)}")
     result = subprocess.run(command, capture_output=True, text=True)
     return result.stdout.strip(), result.stderr.strip(), result.returncode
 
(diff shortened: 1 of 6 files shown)


gdbinit.py
```diff
@@ -61,6 +62,40 @@
 import pwndbg  # noqa: F401
 import pwndbg.profiling
 
+
+def nop(s):
+    return s
+
+
+class ColorFormatter(logging.Formatter):
```
gsingh93: I'll move this to another file

pwndbg/commands/dev.py
```diff
@@ -65,3 +66,20 @@ def dev_dump_instruction(address=None, force_emulate=False, no_emulate=False) ->
...
+    type=str,
+    nargs="?",
+    choices=["debug", "info", "warning", "error", "critical"],
+    default="warning",
+    help="The log level to set.",
+)
+
+
+@pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.DEV)
+def log_level(level: str) -> None:
```
disconnect3d, line 81: Should this be a command or a config parameter?

disconnect3d (approved):
Lgtm though:
- maybe we shall use `pwnlib.log.getLogger(name)`? probably bad idea
- shall `log_level` be a command or a config parameter?

pwndbg/commands/dev.py
gsingh93: A command is the most straight forward way to do this right now. We need to call the `setLevel` method of the root logger to change the log level, while config parameters don't really call any functions.

gsingh93:
Rebased. Also addressed the question about `pwnlib.log.getLogger(name)` on discord.

PR #2238, opened 2024-06-19
Using full PATH length to install poetry

### LINKED ISSUE #2234: Runtime Error with Poetry
### Description

Pwndbg installs successfully but fails to run in gdb, producing the following error 

```
  File "/opt/pwndbg/gdbinit.py", line 83, in <module>
    update_deps(__file__)
  File "/opt/pwndbg/gdbinit.py", line 60, in update_deps
    stdout, stderr, returncode = run_poetry_install(dev=dev_mode)
  File "/opt/pwndbg/gdbinit.py", line 40, in run_poetry_install
    result = subprocess.run(command, capture_output=True, text=True)
  File "/usr/lib/python3.10/subprocess.py", line 503, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/usr/lib/python3.10/subprocess.py", line 971, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.10/subprocess.py", line 1863, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'poetry'
```

It appears to stem from code in commit 180102bf176596e99c4dd79a40918e1a911114a8.

The problem can be fixed by running `pip install poetry`. 

### Steps to reproduce

A minimal dockerfile

```# syntax=docker/dockerfile:1
FROM ubuntu:latest

# Necessary for pwndbg
ENV LANG en_US.UTF-8
ENV LC_CTYPE C.UTF-8

RUN apt-get update

# Install dev requirements
RUN apt-get install -y tmux git
RUN echo "set auto-load safe-path /" >> /root/.gdbinit

# Install pwndbg
RUN apt-get install -y python3 python3-pip
WORKDIR /opt
RUN git clone https://github.com/pwndbg/pwndbg
WORKDIR /opt/pwndbg
RUN ./setup.sh

ENTRYPOINT [ "/bin/bash" ]
```

Running `gdb` from the command line will trigger the error.

### My setup

```
(gdb) show version
GNU gdb (Ubuntu 12.1-0ubuntu1~22.04) 12.1
Copyright (C) 2022 Free Software Foundation, Inc.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "aarch64-linux-gnu".
<https://www.gnu.org/software/gdb/bugs/>.

For help, type "help".
Type "apropos word" to search for commands related to "word".
```

```
root@bdb668b3229b:/# python3 -c "import sys; print(sys.version)"
3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0]
```

```
root@bdb668b3229b:/opt/pwndbg# git log -1 --pretty=oneline
180102bf176596e99c4dd79a40918e1a911114a8 (HEAD -> dev, origin/dev, origin/HEAD) Auto-update dependencies after a git pull (#2203)
```

B1N4RY-P4R45173:
Fixes #2234. I have updated the `gdbinit.py` to use full PATH and also added a function to check if poetry exits on the system.

### DIFF
--- a/gdbinit.py
+++ b/gdbinit.py
@@ -23,13 +23,6 @@ if environ.get("PWNDBG_PROFILE") == "1":
 venv_path = os.environ.get("PWNDBG_VENV_PATH")
 
 
-def check_poetry():
-    if shutil.which("poetry") is None:
-        print("Poetry is not installed or not found in your PATH.")
-        print("Please run './setup.sh' to install Poetry")
-        sys.exit(1)
-
-
 def calculate_hash(file_path):
     with open(file_path, "rb") as f:
         file_hash = hashlib.sha256()
@@ -42,11 +35,18 @@ def calculate_hash(file_path):
 
 
 def run_poetry_install(dev=False):
-    check_poetry()  # Check if poetry executable exists in PATH
     poetry_path = shutil.which("poetry")
     if poetry_path is None:
-        print("Poetry was not found on the $PATH. Please ensure it is installed and on the path, or run `./setup.sh` to update Python dependencies.")
-        sys.exit(1)
+        # Additional check: Look for poetry at $HOME/.local/bin/poetry
+        home_poetry_path = os.path.expanduser("~/.local/bin/poetry")
+        if os.path.exists(home_poetry_path):
+            print(f"Found Poetry at {home_poetry_path}")
+            poetry_path = home_poetry_path
+        else:
+            print(
+                "Poetry was not found on the $PATH. Please ensure it is installed and on the path, or run `./setup.sh` to update Python dependencies."
+            )
+            return 1
 
     command = [poetry_path, "install"]
     if dev:
--- a/gdbinit.py
+++ b/gdbinit.py
@@ -46,7 +46,7 @@ def run_poetry_install(dev=False):
             print(
                 "Poetry was not found on the $PATH. Please ensure it is installed and on the path, or run `./setup.sh` to update Python dependencies."
             )
-            return 1
+            return "", "", 1
 
     command = [poetry_path, "install"]
     if dev:
--- a/gdbinit.py
+++ b/gdbinit.py
@@ -45,7 +45,7 @@ def run_poetry_install(dev=False):
     check_poetry()  # Check if poetry executable exists in PATH
     poetry_path = shutil.which("poetry")
     if poetry_path is None:
-        print("Poetry is not installed. Please install it and try again.")
+        print("Poetry was not found on the $PATH. Please ensure it is installed and on the path, or run `./setup.sh` to update Python dependencies.")
         sys.exit(1)
 
     command = [poetry_path, "install"]
--- a/gdbinit.py
+++ b/gdbinit.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import cProfile
 import hashlib
 import os
+import shutil
 import site
 import subprocess
 import sys
@@ -22,6 +23,13 @@ if environ.get("PWNDBG_PROFILE") == "1":
 venv_path = os.environ.get("PWNDBG_VENV_PATH")
 
 
+def check_poetry():
+    if shutil.which("poetry") is None:
+        print("Poetry is not installed or not found in your PATH.")
+        print("Please run './setup.sh' to install Poetry")
+        sys.exit(1)
+
+
 def calculate_hash(file_path):
     with open(file_path, "rb") as f:
         file_hash = hashlib.sha256()
@@ -34,7 +42,13 @@ def calculate_hash(file_path):
 
 
 def run_poetry_install(dev=False):
-    command = ["poetry", "install"]
+    check_poetry()  # Check if poetry executable exists in PATH
+    poetry_path = shutil.which("poetry")
+    if poetry_path is None:
+        print("Poetry is not installed. Please install it and try again.")
+        sys.exit(1)
+
+    command = [poetry_path, "install"]
     if dev:
         command.extend(("--with", "dev"))
     result = subprocess.run(command, capture_output=True, text=True)
@@ -80,8 +94,6 @@ if venv_path != "PWNDBG_PLEASE_SKIP_VENV" and not path.exists(
         print(f"Cannot find Pwndbg virtualenv directory: {venv_path}: please re-run setup.sh")
         sys.exit(1)
 
-    update_deps(__file__)
-
     site_pkgs_path = glob(os.path.join(venv_path, "lib/*/site-packages"))[0]
 
     # add virtualenv's site-packages to sys.path and run .pth files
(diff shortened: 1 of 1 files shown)


gdbinit.py
```diff
@@ -34,7 +42,13 @@ def calculate_hash(file_path):
 
 
 def run_poetry_install(dev=False):
-    command = ["poetry", "install"]
+    check_poetry()  # Check if poetry executable exists in PATH
```
gsingh93: `check_poetry` seems to do essentially what you've added below. I think we can remove it.

gdbinit.py
```diff
@@ -22,6 +23,13 @@
 venv_path = os.environ.get("PWNDBG_VENV_PATH")
 
 
+def check_poetry():
+    if shutil.which("poetry") is None:
+        print("Poetry is not installed or not found in your PATH.")
+        print("Please run './setup.sh' to install Poetry")
+        sys.exit(1)
```
B1N4RY-P4R45173:
```suggestion
        
```

gdbinit.py
```diff
@@ -34,7 +42,13 @@ def calculate_hash(file_path):
 
 
 def run_poetry_install(dev=False):
-    command = ["poetry", "install"]
+    check_poetry()  # Check if poetry executable exists in PATH
+    poetry_path = shutil.which("poetry")
+    if poetry_path is None:
+        print("Poetry is not installed. Please install it and try again.")
+        sys.exit(1)
```
gsingh93: Let's not exit, let's just return. The message we print is useful so the user knows there's a configuration issue, but the issue isn't a fatal error so we should continue.

gdbinit.py
B1N4RY-P4R45173: Removing this might fix the docker build. But it would ruin the whole point of having a function to check if poetry is installed.

gdbinit.py
```diff
@@ -34,7 +42,13 @@ def calculate_hash(file_path):
 
 
 def run_poetry_install(dev=False):
-    command = ["poetry", "install"]
+    check_poetry()  # Check if poetry executable exists in PATH
+    poetry_path = shutil.which("poetry")
+    if poetry_path is None:
+        print("Poetry is not installed. Please install it and try again.")
```
gsingh93:
```suggestion
        print("Poetry was not found on the $PATH. Please ensure it is installed and on the path, or run `./setup.sh` to update Python dependencies.")
```

gsingh93: Just wanted to confirm, if we hit this code path where poetry is not installed or on the path, and then the user manually runs `./setup.sh`, this message should go away until the next time the `poetry.lock` file changes? This is what I want to happen, and I think it'll happen based on reading the code, but if you could test it and confirm that would be great 🙂

gdbinit.py
```diff
@@ -34,7 +42,13 @@ def calculate_hash(file_path):
 
 
 def run_poetry_install(dev=False):
-    command = ["poetry", "install"]
+    check_poetry()  # Check if poetry executable exists in PATH
+    poetry_path = shutil.which("poetry")
```
gsingh93, line 38: If we don't find `poetry` with `shutil.which`, can you add an additional check to see if it's at `$HOME/.local/bin/poetry`?

B1N4RY-P4R45173: I mean I could add it, if you need. But i have tested this locally `shutil.which`  checks all the directories for the binaries of poetry. So i don't think we need to add it? please let me know if there is something wrong with this.

gdbinit.py
B1N4RY-P4R45173: okay got it.

gdbinit.py
B1N4RY-P4R45173: Okay sure.

gdbinit.py
B1N4RY-P4R45173: The above message would only pop if the poetry was not found on the system. If the `poetry.lock` changes, it won't pop the above message unless the user has uninstalled poetry.

gdbinit.py
gsingh93: `shutil.which` only works if `~/.local/bin` is on the `$PATH`. On your system (and many systems) that's the case. But on a lot of systems that's not the case (like for our Docker images). So this additional check would allow the dependency update flow to work on those systems as well.

B1N4RY-P4R45173:
I have added an additional check condition to check if `poetry` exists on `$HOME/.local/bin/poetry`. Please let me know if there are any more changes to be made.

gdbinit.py
```diff
@@ -96,6 +108,8 @@ def update_deps(file_path):
     bin_path = os.path.join(venv_path, "bin")
     os.environ["PATH"] = bin_path + os.pathsep + os.environ.get("PATH", "")
 
+    update_deps(__file__)
```
gsingh93, line 111: FYI, we can leave this in the old location, there's no need to move this here.

gdbinit.py
```diff
@@ -34,7 +35,20 @@ def calculate_hash(file_path):
...
+        # Additional check: Look for poetry at $HOME/.local/bin/poetry
+        home_poetry_path = os.path.expanduser("~/.local/bin/poetry")
+        if os.path.exists(home_poetry_path):
+            print(f"Found Poetry at {home_poetry_path}")
+            poetry_path = home_poetry_path
+        else:
+            print(
+                "Poetry was not found on the $PATH. Please ensure it is installed and on the path, or run `./setup.sh` to update Python dependencies."
+            )
+            return 1
```
gsingh93:
```suggestion
            return "", "", 1
```
Since below you're returning three things, you need to return three things here too.

gsingh93 (commented):
Make sure you test this with poetry uninstalled (or just temporary move `~/.local/bin/poetry` somewhere else for testing and move it back when you're done testing). I think it will throw an exception, see my last comment for an explanation.

(also `rm .venv/poetry.lock.hash` before every test, otherwise you'll never hit this code path)

B1N4RY-P4R45173:
> Make sure you test this with poetry uninstalled (or just temporary move `~/.local/bin/poetry` somewhere else for testing and move it back when you're done testing). I think it will throw an exception, see my last comment for an explanation.

okay, sure thanks for helping with testing process I was about to ask about it.

> (also `rm .venv/poetry.lock.hash` before every test, otherwise you'll never hit this code path)

okay, thankyou I will check and let you know

gdbinit.py
B1N4RY-P4R45173: Okay thanks for letting me know, I thought it was to be moved as per changes in https://github.com/nikoshell/pwndbg/commit/0cbdb3df09abf463d56797eeec3288362e35fd6e

B1N4RY-P4R45173:
> Make sure you test this with poetry uninstalled (or just temporary move `~/.local/bin/poetry` somewhere else for testing and move it back when you're done testing). I think it will throw an exception, see my last comment for an explanation.
> 
> (also `rm .venv/poetry.lock.hash` before every test, otherwise you'll never hit this code path)

sorry for the delay, but everything seems fine 🙃

PR #2239, opened 2024-06-20
Cleanup gdbinit.py

gsingh93:
I just wanted to change one thing and I ended up rewriting everything 😬 

- `gdbinit.py` is now checked with `mypy`
- Switch all uses of `os.path` to `pathlib.Path`
- Reduce reliance on global variables
- Reduce the amount of top-level code by moving it into functions
- Add some comments
- Other minor refactoring/cleanup

### DIFF
--- a/gdbinit.py
+++ b/gdbinit.py
@@ -9,21 +9,21 @@ import subprocess
 import sys
 import time
 from glob import glob
-from os import environ
-from os import path
+from pathlib import Path
+from typing import List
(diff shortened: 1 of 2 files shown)


gsingh93:
I checked each individual Github Workflow output to make sure the tests were actually passing, instead of silently failing and showing up as passing on GH. I also tested locally with various combinations of `poetry` and `.venv/poetry.lock.hash` being available.

PR #2240, opened 2024-06-20
Fix caching for pwndbg.gdblib.elf functions

### LINKED ISSUE #2237: Incorrect caching on `start` command
### Description

`start` caches the entry breakpoint for too long (I think this is an issue with the caching on `pwndbg.gdblib.elf.entry()` specifically), so if you enable ASLR then `start` twice, the second time won't reset the breakpoint.

### Steps to reproduce

Pick a PIE executable, then run `set disable-randomization off`, then `start`, then `start` again. If you run `memoize` to disable caching, the issue no longer persists.

### My setup

```
Gdb:      14.2
Python:   3.12.4 (main, Jun  7 2024, 06:33:07) [GCC 14.1.1 20240522]
Pwndbg:   2024.02.14 build: 180102b
Capstone: 5.0.1280
Unicorn:  2.0.1
```

Aplet123:
Fixes #2237

Changes two things:
1. Functions in `pwndbg.gdblib.elf` that used to cache on only `start` now cache on both `start` and `objfile`, since changing object files should reload ELF information.
2. Implemented priority event handlers that run before other handlers, so that cache clears will happen before other handlers on that event.

### DIFF
--- a/pwndbg/gdblib/events.py
+++ b/pwndbg/gdblib/events.py
@@ -62,10 +62,6 @@ class StartEvent:
         self.running = True
 
         for function in self.registered:
-            if debug:
-                sys.stdout.write(
-                    "{!r} {}.{}\n".format("start", function.__module__, function.__name__)
-                )
             function()
 
     def on_exited(self) -> None:
--- a/pwndbg/gdblib/events.py
+++ b/pwndbg/gdblib/events.py
@@ -89,6 +89,10 @@ registered: Dict[Any, List[Callable[..., Any]]] = {
     gdb.events.register_changed: [],
 }
 
+# Registered events are wrapped and aren't directly connected to GDB
+# This is a map from event to the actual handler connected to GDB
+connected = {}
+
 
 # When performing remote debugging, gdbserver is very noisy about which
 # objects are loaded.  This greatly slows down the debugging session.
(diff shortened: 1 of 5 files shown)


pwndbg/gdblib/events.py
```diff
@@ -97,7 +100,9 @@ def on_stop(self) -> None:
 objfile_cache: Dict[str, Set[str]] = {}
 
 
-def connect(func: Callable[P, T], event_handler: Any, name: str = "") -> Callable[P, T]:
+def connect(
+    func: Callable[P, T], event_handler: Any, name: str = "", priority: bool = False
```
disconnect3d: @gsingh93 thoughts on this? I believe we need it so that clearing cache is triggered as very first event, but I'm not sure I like a boolean value for it. I mean, this eventually means calling multiple 'priority' events will end up with the same issue as we had before.

gsingh93: Yea, `priority` is usually an enum or an int.

Also, my general thought about adding priority to these events is that we need to really think about the right way to do this, otherwise this is going to turn into a mess and get completely rewritten at some point when it becomes hard to maintain, or doesn't solve the problem well in the long run. And I don't have the time to think through it. But I guess if it's fixing an actual bug and we don't have an alternative solution it makes sense to merge it? Up to you.

Aplet123: I can switch it to an enum priority. Are there any other concerns about the implementation?

disconnect3d: Let's merge this as of now. I don't think we need to make it more complex with int/enum for now :).

disconnect3d: Oh, we already made a switch. I guess that is fine.

PR #2242, opened 2024-06-20
Exit with non-zero code from gdbinit.py if an exception occurs

### LINKED ISSUE #2218: gdb-tests and qemu-tests pass if an exception is thrown in gdbinit.py
I haven't dug into the details of why this happens, but I noticed in https://github.com/pwndbg/pwndbg/pull/2203 that tests were passing even though an exception was being thrown in gdbinit.py. The exception was never shown to the user in the CI logs.

We should fix this for both the gdb-tests and qemu-tests. Also, I think we should ensure that if no tests run, the test fails. When no tests run, we see the following output:
```
Tests passed or skipped: 0
Tests failed: 0
```

gsingh93:
If an exception occurs when sourcing `gdbinit.py` with `source gdbinit.py` or with `--command` when running GDB, GDB does not exit. This is particularly a problem in our testing infrastructure, which relies on a non-zero exit code to determine if an error occurred.

In this PR I essentially wrapped all of `gdbinit.py` in a try/catch, and if an exception occurs we call `sys.exit(1)`, forcing the GDB return code to be non-zero. Note this doesn't apply to GDB 15 (which is used on Ubuntu 24.04), we'll need to find a separate solution for that.

Note that this should only affect exceptions that occur when sourcing `gdbinit.py`. Exceptions that happen later, like when running commands, don't get caught here and don't result in an exit (as expected).

Fixes #2218.

### DIFF
--- a/gdbinit.py
+++ b/gdbinit.py
@@ -159,6 +159,8 @@ def main() -> None:
         pwndbg.profiling.profiler.start()
 
 
+# We wrap everything in try/except so that we can exit GDB with an error code
+# This is used by tests to check if gdbinit.py failed
 try:
     main()
 
--- a/gdbinit.py
+++ b/gdbinit.py
@@ -8,6 +8,7 @@ import site
 import subprocess
 import sys
 import time
+import traceback
 from glob import glob
 from pathlib import Path
 from typing import List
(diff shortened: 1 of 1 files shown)


gdbinit.py
```diff
@@ -158,8 +159,13 @@ def main() -> None:
         pwndbg.profiling.profiler.start()
 
 
-main()
+try:
```
disconnect3d, line 164:
```suggestion
# We wrap everything in try/except so that we can exit GDB with an error code
# This is used by tests to check if gdbinit.py failed
try:
```

PR #2243, opened 2024-06-20
Move GDB initialization and basic startup functionality to a new Debugger-agnostic API

mbrla0:
This is a split up of #2222, which adds the basis of a new Debugger-agnostic API under `pwndbg/dbg`, containing some basic functionality needed for Pwndbg startup, and moves initialization under GDB to that new API.

### DIFF
--- a/gdbinit.py
+++ b/gdbinit.py
@@ -59,7 +59,11 @@ gdb.execute("set charset UTF-8", to_string=True)
 environ["PWNLIB_NOTERM"] = "1"
 
 import pwndbg  # noqa: F401
+import pwndbg.dbg.gdb
 import pwndbg.profiling
+
+pwndbg.dbg = pwndbg.dbg_mod.gdb.GDB()
+pwndbg.dbg.setup()
 
 pwndbg.profiling.init(_profiler, _start_time)
 if environ.get("PWNDBG_PROFILE") == "1":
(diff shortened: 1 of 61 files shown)


pwndbg/dbg/gdb.py
```diff
@@ -0,0 +1,118 @@
...
+        comments.init()
+
+        from pwndbg.gdblib import config_mod
+
+        config_mod.init_params()
+
+        pwndbg.gdblib.prompt.show_hint()
+
+    @override
+    def addrsz(self, address):
```
gsingh93: Can you make sure to add type hints in the all the child class methods as well?

gdbinit.py
```diff
@@ -150,6 +150,12 @@ def main() -> None:
     os.environ["PWNLIB_NOTERM"] = "1"
 
     import pwndbg  # noqa: F811
+    import pwndbg.dbg.gdb
+    import pwndbg.profiling
```
gsingh93: Looks like this is getting imported twice now, let's remove the one below.

pwndbg/dbg/gdb.py
mbrla0: Sure thing!

gdbinit.py
mbrla0: I remove the duplicate in this snippet, mostly because I want to minimize the number of modules that get imported before `pwndbg.dbg` is set to a real value.

PR #2244, opened 2024-06-20
Move main configuration object from `pwndbg.gdblib.config` to `pwndbg.config`

mbrla0:
This is a split up of #2222, which adds a new `config` object, identical to the one currently in `pwndbg.gdblib.config`, to `pwndbg/__init__.py`. This PR also changes existing references to the old object in the code to the new object.

This PR also introduces a debugger-agnostic version of the `gdb.PARAM_*` family of values, under `pwndbg.lib.config.PARAM_*`.

### DIFF
--- a/pwndbg/__init__.py
+++ b/pwndbg/__init__.py
@@ -5,8 +5,9 @@ import signal
 import gdb
 
 # isort: off
-from pwndbg.config import config as config
-import pwndbg.config as config_mod
+import pwndbg.lib.config
+
+config: pwndbg.lib.config.Config = pwndbg.lib.config.Config()
 # isort: on
 
 import pwndbg.color
--- a/pwndbg/__init__.py
+++ b/pwndbg/__init__.py
@@ -4,6 +4,11 @@ import signal
 
 import gdb
 
+# isort: off
+from pwndbg.config import config as config
+import pwndbg.config as config_mod
+# isort: on
+
 import pwndbg.color
 import pwndbg.commands
 import pwndbg.gdblib
@@ -81,6 +86,10 @@ from pwndbg.commands import comments
 
 comments.init()
 
-from pwndbg.gdblib import config_mod
+from pwndbg.gdblib import config_mod as gdblib_config_mod
 
-config_mod.init_params()
+gdblib_config_mod.init_params()
+
+from pwndbg.gdblib.prompt import show_hint
+
+show_hint()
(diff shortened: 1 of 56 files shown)


pwndbg/__init__.py
```diff
@@ -81,6 +87,10 @@
...
 
-from pwndbg.gdblib import config_mod
+from pwndbg.gdblib import config_mod as gdblib_config_mod
+
+gdblib_config_mod.init_params()
+
+from pwndbg.gdblib.prompt import show_hint
 
-config_mod.init_params()
+show_hint()
```
gsingh93, line 96: I think you can remove these lines and leave them in the other PR. They seem to be the ones breaking the unit test anyway.

pwndbg/gdblib/shellcode.py
```diff
@@ -13,6 +13,11 @@
 import pwnlib.shellcraft
 
 import pwndbg
+import pwndbg.gdblib.arch
+import pwndbg.gdblib.memory
+import pwndbg.gdblib.prompt
+import pwndbg.gdblib.regs
+import pwndbg.gdblib.vmmap
```
gsingh93, line 20: I'm wondering why it was necessary to add these?

pwndbg/lldblib/__init__.py
```diff
@@ -0,0 +1,9 @@
+from __future__ import annotations
```
gsingh93: Does this need to be in this PR or can we move it to the other one?

gsingh93 (commented):
I reviewed everything, this looks good, but I think you can go ahead and fix the test (which should just be related to the `show_hint` change you can move to the following PR, and two or three type issues. Once that's done this is good to merge.

pwndbg/__init__.py
mbrla0: This has a lot to do with the order of imports sometimes causing `$ida` not to show up on the startup prompt. I did originally intend to have this change be present in only one of the PRs, but my local tests started failing for that when I removed it.

Maybe something changed when I merged the upstream changes that fixed it, but, I don't think it hurts to leave them here, just in case.

pwndbg/gdblib/shellcode.py
mbrla0: To placate `mypy`, though what exactly the error it was giving me was, I can't recall.

pwndbg/lldblib/__init__.py
mbrla0: D'oh. Forgot to remove that one when I was splitting the commit into two. Thanks for pointing it out, I'll take it out.

PR #2248, opened 2024-06-21
Call type instructions

OBarronCS:
This PR adds a "call-like" attribute to instructions. We previously relied on the capstone constant CS_GRP_CALL to identify call instructions, but this misses a handful of instructions that behave like calls, like a handful of "branch-and-link" instructions in MIPS that capstone doesn't label as calls. This makes the stepping through MIPS more clean, and makes things like nextcall work better on MIPS.


OBarronCS:
This is closed in favor of https://github.com/pwndbg/pwndbg/pull/2261

Issue #13, opened 2015-05-17
Honor delay slots in disassembly splits


zachriggle:
MIPS, SuperH, SPARC all have delay slots.
## Wrong

```
    0x767cc014    ori    $t5, $a0, 0xfeff
    0x767cc018    b      0x767cc044

    0x767cc01c    ori    $t7, $a0, 0xff7d
    0x767cc020    addiu  $v1, $v1, 8
```
## Correct

```
    0x767cc014    ori    $t5, $a0, 0xfeff
    0x767cc018    b      0x767cc044
    0x767cc01c    ori    $t7, $a0, 0xff7d

    0x767cc020    addiu  $v1, $v1, 8
```

disconnect3d:
Let me leave it here: https://en.wikipedia.org/wiki/Delay_slot

disconnect3d:
This can be tested via:

```
sudo apt install -y gcc-mips-linux-gnu
echo int main() {} > main.c
mips-linux-gnu-gcc main.c
qemu-mips -g 1234 -L /usr/mips-linux-gnu/ ./a.out
```
And in another console:
```
gdb-multiarch ./a.out
target remote :1234
```

And then stepping through the program. I believe the bug is still there.

cc: @OBarronCS

disconnect3d:
@OBarronCS Was all of this implemented? Can we close this?

OBarronCS:
Yes! This has been implemented.

disconnect3d:
Closing then! Thanks!

PR #2249, opened 2024-06-21
MIPS delay slot support

OBarronCS:
This PR enhances the behavior of the disassembly of MIPS in regards to delay slots. 

The MIPS architecture has delay slots, which mean that the instruction directly after a branch instruction is always executed, and the effect of the branch (if any) is applied after this. Now with the disassembly splits (the newlines that we put after branch instructions), the delay slots are grouped with the branch instruction to show that they are logically grouped.

Before:
mips_before

After:
mips_after

Asciinema showing it in action!
https://asciinema.org/a/ev3KkITm79sFJeK8DZThZL0DJ

Note that not all MIPS branch instructions have a delay slot. A group of instructions called the "branch-likely" instructions do not execute the branch delay slot instruction if the branch is taken, which effectively means the pipeline is stalled and it behaves like you would expect on x86 or Arm. The code handles these instructions as well.

Another thing to note is the behavior of the Unicorn engine. When the emulator is paused in a delay slot and is continued afterwards, it loses the context of the branch, effectively throwing away the branch decision made in the previous instruction. It always proceeds like the branch was not taken. This is a known limitation that doesn't seem to have a built-in workaround - https://github.com/unicorn-engine/unicorn/issues/1630

To workaround this, we disable the Unicorn emulator when we encounter a branch that has a delay slot.

This is related to the following issue: https://github.com/pwndbg/pwndbg/issues/13#issuecomment-2081476799

These changes will be later extended to apply to SPARC - another architecture that pwndbg supports that also has delay slots.


OBarronCS:
Closed in favor of https://github.com/pwndbg/pwndbg/pull/2262

Issue #1039, opened 2022-07-19
aarch64 miss x30 register


junknet:
pwndbg version 2022.01.05-1
OS: Linux junknet-manjaro 5.15.53-1-MANJARO #1 SMP PREEMPT Thu Jul 7 16:56:55 UTC 2022 x86_64 GNU/Linux
![2022-07-19_12-26](https://user-images.githubusercontent.com/55597646/179664887-d42b3d08-3462-42d7-a503-8d3df7fef612.png)
![2022-07-19_12-27](https://user-images.githubusercontent.com/55597646/179664912-96b87500-d254-4b46-896f-f9c1df44d28b.png)

disconnect3d:
Btw what's the exact CPU? Is that X30 some kind of special register?

This can be fixed easily by adding 'x30' here: https://github.com/pwndbg/pwndbg/blob/8cc0ab8b48c8a4d020c4b9215355cdc6f30eb0f3/pwndbg/regs.py#L106-L118

junknet:
> 
CPU: Snapdragon 855
X29 == fp
X30 == lr 
X31 == sp

junknet:
> 

CPU: Snapdragon 855
X29 == fp
X30 == lr
X31 == sp

disconnect3d:
@junknet can you just add it to the regs spec? Also are all other ARM64 having this reg?

junknet:
> @junknet can you just add it to the regs spec? Also are all other ARM64 having this reg?

Yes, this is the register used by the AARCH64 instruction set (RET).
This is my modification and effect drawing.
![2022-07-30_16-28](https://user-images.githubusercontent.com/55597646/181898649-151a440d-2dad-4143-93b2-4d68cf47c411.png)
![2022-07-30_16-28_1](https://user-images.githubusercontent.com/55597646/181898753-67748012-f5bf-4b33-a0f4-5710d220927d.png)

disconnect3d:
@junknet Sure, let's add X30 to the list.

But, does it make sense to display X29 separately if it is the same register as SP?

disconnect3d:
Added in https://github.com/pwndbg/pwndbg/commit/d2b2523695a19b85487b91659668b7a94e84dc6c

Shall we close this issue? WDYT about X29?

disconnect3d:
We could maybe add a TIP informing about it here: https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/tips.py#L3-L9 :D

[...]

PR #2251, opened 2024-06-23
Make retaddr default shown in context by default

OBarronCS:
This PR proposes setting the `show-retaddr-reg` config setting to `True` by default. 

Most non-x86 architectures that Pwndbg supports use a "branch-and-link" mechanism for calling functions, meaning instead of pushing the return address to the stack, it is put into a register (and the function that is called is responsible for pushing it onto the stack in the case that it is not a leaf function). Arm and PowerPC use the "BL" registers, MIPS and RISCV uses the "ra" register, and SPARC has one reserved as well.

Currently, Pwndbg opts not to display this register by default in the context or when using the `regs` command, and instead hides it behind a configuration setting. 

I found the absence of the register confusing when first jumping into these non-x86 architectures and trying to find how functions are returning. I would see the register mentioned in Architecture manuals, but not see them in Pwndbg. Especially in the context of pwn, knowing where the return address is important, and only when I dug into the code did I find that the register was hidden behind a configuration setting.

As we move to have first-class support for non-x86 architectures, I'd propose to turn this setting on by default.

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -549,7 +549,7 @@ def regs(regs=[]) -> None:
 
 
 pwndbg.config.add_param("show-flags", False, "whether to show flags registers")
-pwndbg.config.add_param("show-retaddr-reg", False, "whether to show return address register")
+pwndbg.config.add_param("show-retaddr-reg", True, "whether to show return address register")
 
 
 def get_regs(regs: List[str] = None):
--- a/pwndbg/lib/regs.py
+++ b/pwndbg/lib/regs.py
@@ -248,7 +248,6 @@ aarch64 = RegisterSet(
         "x27",
         "x28",
         "x29",
-        "x30",
     ),
     misc=(
         "w0",
@@ -576,7 +575,6 @@ riscv = RegisterSet(
     stack="sp",
     retaddr=("ra",),
     gpr=(
-        "ra",
         "gp",
         "tp",
         "t0",


disconnect3d:
@OBarronCS Can you show how context displays before and after this on different archs in screenshots?

OBarronCS:
Here are screenshots of the setting disable then enabled on RISC-V, MIPS, and AArch64. In all cases, the `retaddr` register (`RA` on MIPS and RISC-V and `LR` on AArch64) is directly before the PC - it's the second to last register displayed.


RISC-V before:
riscv_no

RISC-V after:
risc_yes

MIPS before:
mips_no

MIPS after:
mips_yes

AArch64 before:
aarch_no

AAarch64 after:
aarch_yes

disconnect3d:
@OBarronCS I always thought we don't display RA/LR registers bcoz they were aliases to other registers.

Like here: https://developer.arm.com/documentation/dui0801/a/Overview-of-AArch64-state/Predeclared-core-register-names-in-AArch64-state

> LR - Link register. This is a synonym for X30.

With the proposed change here, we will display X30 or LR twice. Does this make sense?

OBarronCS:
Hmm. Looks like the way the registers are defined in regs.py aarch64 that LR will indeed be displayed twice - this is an exception, as all the other arches have the link register as distinct. On arm (32-bit), we don't have this duplicate.

You're right - it doesn't make much sense to display LR and x30 twice. x30 is defined as the link register in the Procedure Call Standard for the Arm Architecture (good references here: https://github.com/ARM-software/abi-aa/releases) so in sane binaries it shouldn't be used in other uses - but in theory it can be used like any other general purpose register (although the same is the case for r14-LR on arm, and likely on all the other arches here). I'd be partial to removing x30 from the `gpr` list on aarch64 - but perhaps we elicit feedback from others on how to handle this case.

disconnect3d:
Fwiw https://github.com/pwndbg/pwndbg/issues/1039

I think we can safely _not_ display duplicated regs.

disconnect3d:
@OBarronCS anything else needs to be done here? Do we not display duplicates? Or does this need to be fixed?

OBarronCS:
This latest small commit removes the displaying of duplicate regs. In only changes the register definitions of AArch64 and RISCV, which had the link register listed as a `GPR` register. On AArch64, x30 = LR, and on RISCV `ra` was previously listed as a GPR.

PR #2252, opened 2024-06-24
Suppress unintended output in setup.sh

### LINKED ISSUE #2136: setup.sh should not show grep error messages
```
root@3a7eb74dfe9f:/pwndbg# ./setup.sh
root
grep: /root/.gdbinit: No such file or directory
Hit:1 http://security.ubuntu.com/ubuntu noble-security InRelease
Hit:2 http://archive.ubuntu.com/ubuntu noble InRelease
Hit:3 http://archive.ubuntu.com/ubuntu noble-updates InRelease
...
```

We should not show things like `grep: /root/.gdbinit: No such file or directory`.

Aplet123:
Fixes #2136

Also stops `root` from being printed when `sudo` doesn't exist and the script is being run as root.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -4,7 +4,7 @@ set -e
 # If we are a root in a container and `sudo` doesn't exist
 # lets overwrite it with a function that just executes things passed to sudo
 # (yeah it won't work for sudo executed with flags)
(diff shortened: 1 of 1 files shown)


gsingh93:
Nice, I was not aware of the -`s` flag. Thanks!

PR #2254, opened 2024-06-24
Move addition of custom commands to the Debugger-agnostic API

mbrla0:

This PR adds support for custom command addition to the Debugger-agnostic API, and makes it so that commands created using `pwndbg.commands.Command` use it for command registration.

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -190,22 +190,13 @@ class Command:
         if not session:
             # No interactive session we can draw a history from.
             return False
-        lines = session.history()
+        last_line = session.history(1)
 
         # No history
-        if not lines:
+        if not last_line:
             return False
 
-        last_line = lines[-1]
-        number_str, command = last_line.split(maxsplit=1)
-        try:
-            number = int(number_str)
-        except ValueError:
-            # In rare cases GDB will output a warning after executing `show commands`
-            # (i.e. "warning: (Internal error: pc 0x0 in read in CU, but not in
-            # symtab.)").
-            return False
-
+        number, command = last_line[-1]
         # A new command was entered by the user
         if number not in Command.history:
             Command.history[number] = command
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -88,17 +88,6 @@ GDB_BUILTIN_COMMANDS = list_current_commands()
 pwndbg_is_reloading = getattr(gdb, "pwndbg_is_reloading", False)
 
 
-def fallback_lex_args(command_line: str) -> List[str]:
-    """
-    Lexes the given command line into a list of arguments, according to the
-    conventions of the debugger being used and of the interactive session.
-
-    This is the fallback version, for use when `pwndbg.dbg.session()` is not
-    available.
-    """
-    return command_line.split()
-
-
 class Command:
     """Generic command wrapper"""
 
@@ -153,11 +142,7 @@ class Command:
             A ``(tuple, dict)``, in the form of ``*args, **kwargs``.
             The contents of the tuple/dict are undefined.
         """
-        session = pwndbg.dbg.session()
-        if not session:
-            return fallback_lex_args(argument), {}
-        else:
-            return session.lex_args(argument), {}
+        return pwndbg.dbg.lex_args(argument), {}
 
     def invoke(self, argument: str, from_tty: bool) -> None:
         """Invoke the command with an argument string"""
@@ -186,11 +171,7 @@ class Command:
         if not from_tty:
             return False
 
-        session = pwndbg.dbg.session()
-        if not session:
-            # No interactive session we can draw a history from.
-            return False
-        last_line = session.history(1)
+        last_line = pwndbg.dbg.history(1)
 
         # No history
         if not last_line:
@@ -558,11 +539,7 @@ class _ArgparsedCommand(Command):
         )
 
     def split_args(self, argument: str):
-        session = pwndbg.dbg.session()
-        if not session:
-            argv = fallback_lex_args(argument)
-        else:
-            argv = session.lex_args(argument)
+        argv = pwndbg.dbg.lex_args(argument)
         return (), vars(self.parser.parse_args(argv))
 
 
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -416,7 +416,7 @@ def _try2run_heap_command(function: Callable[P, T], *a: P.args, **kw: P.kwargs) 
             w(
                 f"You can try to determine the libc symbols addresses manually and set them appropriately. For this, see the `heap_config` command output and set the config for `{err.symbol}`."
             )
-        if pwndbg.gdblib.config.exception_verbose or pwndbg.gdblib.config.exception_debugger:
+        if pwndbg.config.exception_verbose or pwndbg.config.exception_debugger:
             raise err
 
         pwndbg.exception.inform_verbose_and_debug()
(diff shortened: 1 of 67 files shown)


pwndbg/dbg/gdb.py
```diff
@@ -22,6 +58,8 @@ def setup(self):
         load_gdblib()
         load_commands()
 
+        from pwndbg.gdblib import prompt
```
gsingh93, line 55: Why did you move the import in here?

pwndbg/dbg/__init__.py
```diff
@@ -5,14 +5,47 @@
...
 
 
+class Session:
+    """
+    Interactive debugger session. Handles things like commands and history.
+    """
+
+    def history(self) -> List[str]:
+        """
+        The command history of this interactive session.
```
gsingh93: This comment makes it seem like you get the whole history, but GDB will only give you the last 10 commands. It also sounds like each element is the history command, but the format is `<index> <command>`, not even mentioning whitespace. How does LLDB do this? We should either clarify in this comment what the user is going to get back (regardless of debugger), or update the implementation in `GDBSession` to match the expected format.

pwndbg/dbg/__init__.py
```diff
@@ -30,6 +63,22 @@ def setup(self, *args: Any) -> None:
         """
         raise NotImplementedError()
 
+    def session(self) -> Session | None:
```
gsingh93: Why would this ever return `None`?

pwndbg/dbg/gdb.py
```diff
@@ -13,7 +15,41 @@
...
+    def invoke(self, args: str, from_tty: bool) -> None:
+        self.handler(self.debugger, args, from_tty)
+
+
+class GDBCommandHandle(pwndbg.dbg_mod.CommandHandle):
+    def __init__(self, command: gdb.Command):
+        self.command = command
+
+    def remove(self) -> None:
+        # GDB doesn't support command removal?
```
gsingh93: Pretty sure you're correct it doesn't, we can remove the question mark. BTW, we should file a bug for this if we haven't already (check with @disconnect3d).

pwndbg/dbg/__init__.py
```diff
@@ -5,14 +5,47 @@
...
+
+    def lex_args(self, command_line: str) -> List[str]:
+        """
+        Lexes the given command line into a list of arguments, according to the
+        conventions of the debugger being used and of the interactive session.
+        """
+        raise NotImplementedError()
+
+
+class CommandHandle:
```
gsingh93, line 15: Why do we need this class? Even if the debugger supports removing commands, it makes more sense to me to have a `remove_command` method in the debugger class itself.

pwndbg/dbg/gdb.py
disconnect3d: I believe we didn't.

pwndbg/dbg/gdb.py
mbrla0: Importing `pwndbg.gdblib.prompt` triggers a chain of imports that ends with importing `pwndbg.commands.comments`, which uses `pwndbg.command.ArgparsedCommand`.

Before this PR, this wouldn't really be a problem, but this PR changes the way commands are added, such that it depends on the `pwndbg.dbg` object to be properly set up, as adding new commands relies on calling `pwndbg.dbg.add_command()`.

Because importing `pwndbg.dbg.gdb` is part of the process of setting up `pwndbg.dbg` under GDB, we can't really import anything that depends on it already being set up at the root of the module. It's fine importing it in `setup()`, though, as it only gets called after the object has been set up. In fact, the `setup()` method exists precisely because we want to do most of our setting up after `pwndbg.dbg` has been set to something other than `None`.

pwndbg/dbg/__init__.py
mbrla0: That's a fair point. I had originally intended to just do as little as needed to move this out, while also not breaking any existing code, and leave sorting this out to a future PR. But we might as well do it now.

LLDB can show the full history, in a format that is close, but not compatible for the purposes of `pwndbg.commands`. So I feel it would make sense to both change this method so that it returns the last N (defaults to 10) entries in the history, as a list of tuples <index, command>, and change the code in `GDBSession` and the code in `check_repeated` to reflect that.

pwndbg/dbg/__init__.py
mbrla0: Depends on what _exactly_ we want the session to reflect. LLDB has the notion of an ExecutionContext, which contains information about the currently selected target, process, frame, and thread. It might make sense to have it be `None` in the cases one is not available. It also might not, if we decide that it makes sense to keep methods in there that don't depend on any of these being available (such as the history method itself).

I feel like it shouldn't hurt to leave that there until we're sure about what we want. Maybe we should add a comment explaining that?

pwndbg/dbg/gdb.py
gsingh93: Ah, right. Another way of thinking about it is we shouldn't do any gdblib imports until after `load_gdblib` is called. Can you add a comment about this here? It's something non-obvious we should call out.

pwndbg/dbg/__init__.py
gsingh93: It's hard for me to say for sure this early on, but my gut feeling is it would be better to remove this layer of abstraction and if the caller requests something that depends on a non-existent ExecutionContext, we return `None` for those specific function calls. i.e. `dbg.process()` would return `None` if there was no execution context.

Our current usage of `Session` is only `history`, which means we're already using for non-execution context related things. I think it's better to remove it unless we have a concrete need for it. Once we have a concrete need for it, it should be clear at that point exactly what its purpose should be, what methods it should have etc.

mbrla0: That makes sense. I'll take it our, then.

pwndbg/dbg/__init__.py
mbrla0: It makes it cleaner for command removal under LLDB. Because of some quirks of LLDB, we need to store a global object in the root of the module, having a class makes it so that we don't have to store the name for it in a dictionary in the `Debugger` class.

pwndbg/commands/__init__.py
```diff
@@ -112,7 +112,10 @@ def __init__(
         if command_name is None:
             command_name = function.__name__
 
-        super().__init__(command_name, gdb.COMMAND_USER, gdb.COMPLETE_EXPRESSION, prefix=prefix)
```
gsingh93, line 115: I guess `prefix` is unused now? Do we want it to be used or should we remove it? I did a quick grep and did not see anyone using it in the codebase, but now that I know it exists it seems useful.

(I'm also surprised the linter didn't flag this unused variable...)

gsingh93 (commented):
LGTM, just a minor question about the usage of `prefix`.

mbrla0:
@gsingh93 anything missing?

gsingh93:
@mbrla0 did you see my comment above?

pwndbg/commands/__init__.py
mbrla0: Yeah, I did the same check when I was making that change, and it seems that it was never really used. The only place `Command` was ever instanced was through `ArgparsedCommand`, and it just has it be always `False`. I didn't remove it mostly because I failed to realize that call was the only place where it got used.

And, personally, while I think it might seem pretty useful at a first glance, every command in Pwndbg where that could get used - like in `track-got`, for instance - already uses `argparse.ArgumentParser` to achieve the same effect, so I'd say its probably for the best that we avoid it, since using `ArgumentParser` for that frees us from having to deal with subcommands in the debugger-agnostic API.

mbrla0:
> @mbrla0 did you see my comment above?

Sorry, I'd missed it

PR #2255, opened 2024-06-24
Move argument value parsing of custom commands to Debugger-agsnostic API

mbrla0:

Currently, the `pwndbg.commands` modules directly uses `gdb.Value` and `gdb.parse_and_eval` to parse the arguments passed to our custom commands. Obviously, we can't have that if we hope to use this module under LLDB - and we do.

So, this PR adds analogues of `Type` and `Value` classes to the debugger-agnostic API, as well as as an equivalent of `gdb.parse_and_eval`, named `evaluate_expression`, after LLDB's own `EvaluateExpression` method.

The methods in `Type` and `Value` mirror those of their GDB counterparts, if you were to take away every method that it not currently being used in Pwndbg. The main reason for this is that it makes it easier for us to completely change the way argument parsing works for commands based on `pwndbg.commands.Command`, but in a way that is transparent to all the existing code in Pwndbg, that relies on these methods being present, but that don't otherwise care what class implements them.


pwndbg/dbg/__init__.py
```diff
@@ -12,6 +13,169 @@
...
+    POINTER = 1
+    ARRAY = 2
+    STRUCT = 3
+    TYPEDEF = 4
+    UNION = 5
+    INT = 6
+    ENUM = 7
+
+
+class TypeField:
```
gsingh93, line 30: nit: I personally like `Field` as a name more than `TypeField`, but don't feel too strongly about it.

pwndbg/dbg/__init__.py
```diff
@@ -12,6 +13,169 @@
...
+    # functionality out of the debugger API.
+    #
+    # TODO: Move to single, common string function.
+    def string(self) -> str:
+        """
+        If this value is a string, then this method converts it to a Python string.
+        """
+        raise NotImplementedError()
+
+    # This is a GDB imeplementation detail.
```
gsingh93:
```suggestion
    # This is a GDB implementation detail.
```

pwndbg/dbg/__init__.py
```diff
@@ -12,6 +13,169 @@
...
+        raise NotImplementedError()
+
+    def __int__(self) -> int:
+        """
+        Converts this value to an integer, if possible.
+        """
+        raise NotImplementedError()
+
+    # Because casting is still sloppy (i.e. it acceps `gdb.Type` objects) in
+    # some places, we have to allow it here for lints to pass.
```
gsingh93:
```suggestion
    # Because casting is still sloppy (i.e. it accepts `gdb.Type` objects) in
    # some places, we have to allow `Any` here for lints to pass.
```

pwndbg/dbg/__init__.py
gsingh93, line 30: For my own understanding, this will eventually be something that GDB and LLDB will subclass right? We just don't need to do it at the moment for things to work?

gsingh93 (approved):
LGTM, just let me know when the typos are fixed.

pwndbg/dbg/__init__.py
mbrla0: Interestingly, it doesn't look like it's going to be strictly necessary to subclass it later on

From my work on the later changes, LLDB can fill most of them exactly, and two of them (the ones relating to enum fields and inherited members) I haven't implemented yet, but apparently can be implemented with very little effort

mbrla0: LLDB has the concept of children values. I can't confirm it right now, but they seem to be [much like fields (see the registers example)](https://lldb.llvm.org/python_api/lldb.SBValue.html#lldb.SBValue.children), but for values, rather than types

I feel this can muddy the waters a little depending on how we want to treat them in the future, so I felt like specifying these fields specifically belong to a type would be clearer - in the sense that it communicates that they don't have an innate value, unlike ones from a value

I feel like a sensible choice if we end up exposing those would be to call them ValueFields, but, then again, we could just copy LLDB and call them "children"

That being said, I do recognize it's mostly a matter of taste, and I don't feel very strongly about the name we're going with here, either, so if you still feel it makes sense to change it to just Field, I'll do it

mbrla0:
@gsingh93 I've fixed the typos. Can you check if there's anything you'd like to add to my other replies?

pwndbg/dbg/__init__.py
gsingh93: Your reasoning makes sense so lets keep it for now, at least until we have more clarity about children values.

PR #2258, opened 2024-06-25
Move tests from lib/memory.py to separate unit test

Aplet123:
Seems like a 2 year old TODO fell through the cracks

Also, the current bithack implementation only works for powers of 2 but is undocumented. Should the implementation be changed to be more general or should the documentation be updated?

### DIFF
--- a/tests/unit-tests/test_memory.py
+++ b/tests/unit-tests/test_memory.py
@@ -1,0 +1,32 @@
+from __future__ import annotations
+
+import sys
+from unittest.mock import MagicMock
+
+# Replace `pwndbg.commands` module with a mock to prevent import errors, as well
+# as the `load_commands` function
+module_name = "pwndbg.commands"
+module = MagicMock(__name__=module_name, load_commands=lambda: None)
+sys.modules[module_name] = module
+
+# Load the mocks for the `gdb` and `gdblib` modules
+import mocks.gdb
+import mocks.gdblib  # noqa: F401
+
+# We must import the function under test after all the mocks are imported
+from pwndbg.lib.memory import round_down, round_up
+
+
+def test_basic():
+    assert round_down(0xDEADBEEF, 0x1000) == 0xDEADB000
+    assert round_up(0xDEADBEEF, 0x1000) == 0xDEADC000
+
+
+def test_many():
+    for n in range(0x100):
+        for i in range(8):
+            alignment = 1 << i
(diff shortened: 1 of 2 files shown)


tests/unit-tests/test_memory.py
```diff
@@ -0,0 +1,33 @@
+from __future__ import annotations
+
+import sys
+from unittest.mock import MagicMock
+
+# Replace `pwndbg.commands` module with a mock to prevent import errors, as well
+# as the `load_commands` function
+module_name = "pwndbg.commands"
+module = MagicMock(__name__=module_name, load_commands=lambda: None)
+sys.modules[module_name] = module
```
disconnect3d, line 10: We do run tests in single sessions currently so this works fine but I believe this would not work if multiple test files would be executed in a single session, since this mock is never reverted.

disconnect3d:
> Also, the current bithack implementation only works for powers of 2 but is undocumented. Should the implementation be changed to be more general or should the documentation be updated?

What do we mean by current bithack implementation? Do u mean the rounding impl? (haven't looked at it)

Aplet123:
Yeah, the rounding impl uses a bitmask instead of modulo which won't behave properly for non-powers of 2.

tests/unit-tests/test_memory.py
gsingh93: I think this is actually a good thing. In fact, we should probably do something similar to `tests/unit-tests/mocks/gdblib.py` and import it with `mocks.commands` like below (but this doesn't need to be done in this PR).

That being said, I'm wondering why we only need to mock this out now? Why are the other tests working? What error shows up in this test if we don't have this mock?

gsingh93 (approved):
Thanks for the test! This LGTM, although I am curious about the what issue `pwndbg.commands` was causing here that wasn't happening with the other tests.

tests/unit-tests/test_memory.py
Aplet123: I copied this chunk from the other tests since both the other unit test files have it, but commenting it out doesn't actually seem to break anything, so maybe that can be removed/replaced with an actual mock?

gsingh93: Ah got it. You can leave it as is for now, but in the future we should really consolidate or remove this so it's less boilerplate when writing a test.

PR #2259, opened 2024-06-26
Manually resolve AArch64 conditional branches

OBarronCS:
This PR enhances the way we resolve conditional branches in AArch64 during instruction enhancement.

For all branch-type instructions in AArch64, we can now determine whether or not the branch is taken and resolve it's target without emulation.  

Note that all of these screenshots have emulation disabled

A `ret` instruction before.
aarch64_ret_before

After:
aarch64_ret_afterpng

A conditional branch before 
aarch64_b_before

After (note the green check mark indicating that we are taking the branch):
aarch64_b_after

### DIFF
--- a/pwndbg/gdblib/disasm/aarch64.py
+++ b/pwndbg/gdblib/disasm/aarch64.py
@@ -12,7 +12,45 @@ import pwndbg.gdblib.disasm.arch
 import pwndbg.gdblib.memory
 import pwndbg.gdblib.regs
 from pwndbg.emu.emulator import Emulator
-from pwndbg.gdblib.disasm.instruction import PwndbgInstruction
+from pwndbg.gdblib.disasm.instruction import InstructionCondition, PwndbgInstruction, boolean_to_instruction_condition
+from pwndbg.lib.regs import BitFlags
+
+
+
+def resolve_condition(condition: int, cpsr: int) -> InstructionCondition:
+    """
+    Given a condition and the NZCV flag bits, determine when the condition is satisfied
+    
+    The condition is a Capstone constant
+    """
+
+    n = (cpsr >> 31) & 1
+    z = (cpsr >> 30) & 1
+    c = (cpsr >> 29) & 1
+    v = (cpsr >> 28) & 1
+
+    condition = {
+        ARM64_CC_INVALID: True, # Capstone uses this code for the 'B' instruction, the unconditional branch
+        ARM64_CC_EQ: z == 1,
+        ARM64_CC_NE: z == 0,
+        ARM64_CC_HS: c == 1,
+        ARM64_CC_LO: c == 0,
+        ARM64_CC_MI: n == 1,
+        ARM64_CC_PL: n == 0,
+        ARM64_CC_VS: v == 1,
+        ARM64_CC_VC: v == 0,
+        ARM64_CC_HI: c == 1 and z == 0,
+        ARM64_CC_LS: not (c == 1 and z == 0),
+        ARM64_CC_GE: n == v,
+        ARM64_CC_LT: n != v,
+        ARM64_CC_GT: z == 0 and n == v,
+        ARM64_CC_LE: not (z == 0 and n == v),
+        ARM64_CC_AL: True,
+        ARM64_CC_NV: True
+    }.get(condition, False)
+
+    return InstructionCondition.TRUE if condition else InstructionCondition.FALSE
+
 
 
 class DisassemblyAssistant(pwndbg.gdblib.disasm.arch.DisassemblyAssistant):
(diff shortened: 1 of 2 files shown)


pwndbg/gdblib/disasm/aarch64.py
```diff
@@ -12,7 +12,45 @@
...
+        ARM64_CC_VC: v == 0,
+        ARM64_CC_HI: c == 1 and z == 0,
+        ARM64_CC_LS: not (c == 1 and z == 0),
+        ARM64_CC_GE: n == v,
+        ARM64_CC_LT: n != v,
+        ARM64_CC_GT: z == 0 and n == v,
+        ARM64_CC_LE: not (z == 0 and n == v),
+        ARM64_CC_AL: True,
+        ARM64_CC_NV: True,
+    }.get(condition, False)
```
disconnect3d, line 51: Fwiw Python will compute all the conditions here and construct a dict each time this is called :(

pwndbg/gdblib/disasm/aarch64.py
```diff
@@ -64,6 +102,61 @@ def generic_register_destination(self, instruction, emu: Emulator) -> None:
...
             instruction.annotation = f"{left.str} => {super()._telescope_format_list(telescope_addresses, TELESCOPE_DEPTH, emu)}"
 
+    @override
+    def _condition(
+        self, instruction: PwndbgInstruction, emu: Emulator
+    ) -> pwndbg.gdblib.disasm.arch.InstructionCondition:
+        # In ARM64, only branches have the conditional code in the instruction,
+        # as opposed to ARM32 which allows most instructions to be conditional
+        if instruction.id == ARM64_INS_B:
+            flags = super()._read_register_name(instruction, "cpsr", emu)
```
disconnect3d, line 112: Why not `self._read_register_name`?

pwndbg/gdblib/disasm/instruction.py
```diff
@@ -333,6 +337,12 @@ def __repr__(self) -> str:
         Can change PC: {self.can_change_instruction_pointer}
         Syscall: {self.syscall if self.syscall is not None else ""} {self.syscall_name if self.syscall_name is not None else "N/A"}"""
 
+        # Hacky, but this is just for debugging
+        if hasattr(self.cs_insn, "cc"):
+            info += f"\n\tARM condition code: {self.cs_insn.cc}"
```
disconnect3d, line 342: Shall we remove this?

disconnect3d: Probably not since its repr anyway?

pwndbg/gdblib/disasm/instruction.py
```diff
@@ -81,6 +81,10 @@ class InstructionCondition(Enum):
     UNDETERMINED = 3
 
 
+def boolean_to_instruction_condition(condition: bool) -> InstructionCondition:
+    return InstructionCondition.TRUE if condition else InstructionCondition.FALSE
```
disconnect3d, line 85: Honestly, I'm not sure I like this enum overall. I understand we want to encode three values but it seems like `True, False, None` would do the same job.

disconnect3d:
Merging this, but please check the comments here.

pwndbg/gdblib/disasm/aarch64.py
disconnect3d: Could be nice to refactor this to `if else ..` stuff.

OBarronCS: From now on I can implement this type of pattern as a `IF/ELIF` chain - eventually when we move to Python 3.10 we can move to the nicer `match/case` syntax.

Alternatively, this could be defined as a global dictionary of lambdas with each function taking the same parameters, then we lookup the function and call it so we only resolve the one condition we really care about.

Do you think this second option might be cleaner?

pwndbg/gdblib/disasm/instruction.py
OBarronCS: I think treating a `bool` type as three distinct values is just asking for bugs (if x vs is x if not None). I also have plans for this enum to add a fourth value - `InstructionCondition.UNCONDITIONAL`, allowing for more preciseness on the type of instruction we are dealing with. Currently, `UNDETERMINED` is very vague.

pwndbg/gdblib/disasm/instruction.py
OBarronCS: Although it's hacky, I think this is the best that we can currently do (I really do not want to subclass PwndbgInstruction just for a debug statement in a particular architecture). This print is only shown using the `dev-dump-instruction` command which is invaluable for debugging

pwndbg/gdblib/disasm/instruction.py
gsingh93: > I think treating a bool type as three distinct values is just asking for bugs (if x vs is x if not None)

Yes, but this is why we almost never use the former syntax (it sometimes sneaks in when I can't get to a code review, I'll check if we can add a lint for it).

But if you're adding more variants later, then yea, an enum makes more sense.

PR #2261, opened 2024-06-26
Call like instructions

OBarronCS:
This PR adds a "call-like" attribute to instructions. We previously relied on the capstone constant CS_GRP_CALL to identify call instructions, but this misses a handful of instructions that behave like calls, like a handful of "branch-and-link" instructions in MIPS that capstone doesn't label as calls. This makes the stepping through MIPS more clean, and makes things like nextcall work better on MIPS.

### DIFF
--- a/pwndbg/gdblib/next.py
+++ b/pwndbg/gdblib/next.py
@@ -145,7 +145,7 @@ def break_next_call(symbol_regex=None):
             break
 
         # continue if not a call
-        if capstone.CS_GRP_CALL not in ins.groups:
+        if not ins.call_like:
             continue
 
         # return call if we:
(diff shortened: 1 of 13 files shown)


pwndbg/gdblib/next.py
```diff
@@ -104,7 +102,7 @@ def next_matching_until_branch(address=None, mnemonic=None, op_str=None):
         if mnemonic_match and op_str_match:
             return ins
 
-        if set(ins.groups) & jumps:
+        if ins.groups_set & ALL_JUMP_GROUPS:
```
disconnect3d, line 105: Considering this is called like 3 times in this file I wonder if we should make a property out of it or cache it...

OBarronCS: I think that caching it as a property would be a good call. I'll look to add some caching decorators on the getter functions of the PwndbgInstruction when we start hammering down on the performance.

PR #2265, opened 2024-06-26
Underline WX memory instead of RWX memory

### LINKED ISSUE #2263: Underline addresses that have WX instead of RWX
I believe the legend should be `WX` for underline and we should underline WX no matter if read is there or not. Perhaps we should change the legend to "WX" too :).

image

In here, the RIP address should be underlined:
image

Aplet123:
Closes #2263

### DIFF
--- a/pwndbg/color/memory.py
+++ b/pwndbg/color/memory.py
@@ -105,7 +105,8 @@ def legend():
             c.heap("HEAP"),
             c.code("CODE"),
             c.data("DATA"),
-            c.wx("WX"),
+            # WX segments will also be marked as code, so do 2 formatters here
+            c.wx(c.code("WX")),
             c.rodata("RODATA"),
         )
     )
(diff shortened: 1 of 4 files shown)


gsingh93:
This LGTM, but could you make sure to include a screenshot in PRs for any UI changes?

Aplet123:
Sure.

Before:
image

After:
image

Also unrelated to this PR but the `xinfo` command seems to have a couple of bugs (I'm getting `Exception occurred: xinfo: 'p_vaddr' (<class 'KeyError'>)` after some of the dumping, and also the underline includes whitespace which is very ugly)

gsingh93:
Thanks, the screenshots look good.

> Also unrelated to this PR but the xinfo command seems to have a couple of bugs (I'm getting Exception occurred: xinfo: 'p_vaddr' (<class 'KeyError'>) after some of the dumping

Can you file an issue for it?

I'll merge this once the tests finish running.

PR #2268, opened 2024-06-29
AArch64 conditional select statement annotations

OBarronCS:
This PR adds annotations to the `conditional select` category of AArch64 instructions.

These instructions conditional execute an action, largely setting the destination register to some mutation of a source register (https://developer.arm.com/documentation/102374/0102/Program-flow---conditional-select-instructions).

Now, in addition to a green check mark indicating the action is taken, there is text to indicate the result value of the destination operand.

If emulation is disabled, it will attempt to determine the correct action that instruction takes manually (effectively emulating the instruction in the Python code).

Example:
CSEL instruction (with emulation):
conditional_select_emulate

It also works without emulation:
conditional_select_no_emulate

There are also small changes that were suggested in #2259

### DIFF
--- a/pwndbg/gdblib/disasm/aarch64.py
+++ b/pwndbg/gdblib/disasm/aarch64.py
@@ -7,6 +7,7 @@ from capstone import *  # noqa: F403
 from capstone.arm64 import *  # noqa: F403
 from typing_extensions import override
 
+import pwndbg.color.memory as MemoryColor
 import pwndbg.gdblib.arch
 import pwndbg.gdblib.disasm.arch
 import pwndbg.gdblib.memory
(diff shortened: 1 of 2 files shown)


OBarronCS:
This is ready in terms of feature additions. I kept it a draft because I dislike the way I am currently detecting the bit width of the registers in the code (Capstone does not provide a mechanism in AArch64 to determine the bit-width of a register, and the best I could think of was to determine the width of the register based on it's name).

Determining the bit-width of a register will probably become a recurring theme, perhaps we create a global datastructure that maps all register names to it's width?

```
REGISTER_WIDTHS_ARM64 = {
    "w1":4,
    "x1":8
    ....
}
```
And have maps for each architecture?

OBarronCS:
Closed in favor of #2368

PR #2273, opened 2024-07-01
gdblib/stacks.py: fix bug with stack exploration

disconnect3d:
This PR fixes a bug with stack exploration: we did not return it from `stack.get()`!

Additionally, it refactors the `is_executable` function to return early and do not cache the `nx` variable.


disconnect3d:
Debugging a binary run via qemu user: `qemu-x86_64 -g 1234 /bin/ls`

Before this PR:
image

After this PR:
image

Ofc this is still far from ideal. The exploration is still not great, compared to the normal binary run and stopped at the same time:
image

Issue #2247, opened 2024-06-21
Tests always pass on Ubuntu 24.04


gsingh93:
Our tests rely on `sys.exit` in Python exiting the GDB process with the specified return code. This behavior worked in GDB 12.1, but is different in GDB 15 which is used on Ubuntu 24.04. On GDB 15, the process will not exit. This means the tests status on CI will always pass, even if tests are actually failing, as the return code is always zero.

I asked about this on the GDB mailing list here: https://sourceware.org/pipermail/gdb/2024-June/051417.html

Some other details can be found in #2218.

gsingh93:
From the email thread, we can try `os._exit(1)`. And on discord it was suggested to use `gdb.execute(f"quit {return_code}")`

disconnect3d:
To be clear, those are the lines we will have to change for GDB 15:

https://github.com/pwndbg/pwndbg/blob/dc9c87254aae436147a577290007b0a0b3a8210a/gdbinit.py#L176-L178

We should contact GDB devs if they can persist the old behavior and if not, we should change the above lines accordingly.

gsingh93:
Actually this is the primary line that will need to be changed: https://github.com/pwndbg/pwndbg/blob/dev/tests%2Fgdb-tests%2Fpytests_launcher.py#L32

But we should change the location you mentioned as well.

According to the mailing list thread, this was not intended and we should file a bug. But getting this fixed will take time, so we should make this change anyway.

gsingh93:
I filed a GDB bug here: https://sourceware.org/bugzilla/show_bug.cgi?id=31946

In any case, let's go with the `os._exit(1)` solution for now. And if the bug ever gets fixed we can go back to the original `sys.exit`.

PR #2275, opened 2024-07-01
Qemu user test structure

OBarronCS:
This PR refactors the testing files to reduce code duplication and introduces a new qemu-user space testing mechanism. This will allow us to more easily create cross-architecture tests.

Previously, both the `gdb-tests` and `qemu-tests` folders had identical (or off by one word) `pytest_collect.py` and `pytest_launcher.py`. The file has been changed to take a parameter (which was previously a hardcoded path), and the file moved up one directory.

The `gdb-tests/tests.py` was also moved up one folder, as I contains logic that is re-used for qemu-user tests - slight modifications were made to allow us to specify the use of `gdb` vs `gdb-multiarch`, and to choose which testing suite to execute, and test for open ports for qemu to use.

`./tests.sh` still executes gdb tests like before. `./tests.sh -t qemu-user` executes the new tests.

The new tests work using a new Pytest fixture that takes in a string which represents assembly code, compiles it (with the help of pwntools), and then launches `qemu-aarch -g {PORT}` with the binary, then attaches gdb to that localhost port (it is running gdb-multiarch).

I opted for this (as opposed to full system emulation) for ease of use - I find it handy for short assembly snippets to be placed right next to the tests that use them, making it easy to iterate and understand what's going on. Future changes could incorporate full qemu-system testing for each arch.

This is a draft PR because there are still some things to work out:
- Running the tests in the Docker container fails for some unknown reason - the `vmmap` is showing as empty when gdb connects to qemu (can be tested by running the tests with --pdb and running gdb.execute("vmmap")). I can't reproduce this error outside of the docker container's testing environment. If I create a bash session in the container, and manually invoke qemu and manually attach to it, it works. A mystery to be solved.
- The top level tests.sh runs the `unit-tests` tests everytime it's invoked, meaning for both user and qemu-user tests, which is extraneous. We'll have to figure out how we want to invoke each of these distinct test suites.

There are some old qemu-user tests which compile a .C file - I kept these around (and moved them to an `old` directory temporarily) while we figure out what to do with them.


OBarronCS:
Upon doing another test, I found that after the test (in the docker container) failed and I could run commands in the Pdb debugger, I had to first run "context", and then the mappings would be populated, and "vmmap" would then print the mappings. I don't need to do this when running the tests on my host.

OBarronCS:
Another bug: locally I have aarch64-linux-gnu-gcc-13, which when compiling keeps the symbol names in the assembly. The symbols names are not kept in the version of gcc in the container (which is what pwntools `make_elf_from_assembly()` is using)

OBarronCS:
I updated the container that runs the qemu tests to 24.04, and now they run correctly. I guess the version of qemu in 22.04 was too old. I also added 24.04 to the Docker compose. To run these test, run:
```sh
docker compose run -T ubuntu24.04 ./tests.sh -t qemu-user
```

tests/qemu-tests/test_qemu.sh
```diff
@@ -36,7 +36,7 @@ test_arch() {
         -ex "file ./binaries/reference-binary.${arch}.out" \
         -ex 'py import coverage;coverage.process_startup()' \
         -ex "target remote :1234" \
-        -ex "source ./tests/user/test_${arch}.py"
+        -ex "source ./tests/user/old/test_${arch}.py"
```
gsingh93: Is this the correct path?

.github/workflows/tests.yml
```diff
@@ -63,7 +63,7 @@ jobs:
       uses: codecov/codecov-action@v3
 
   qemu-tests:
-    runs-on: [ubuntu-22.04]
+    runs-on: [ubuntu-24.04]
```
gsingh93, line 67: Be careful running tests on Ubuntu 24.04, if they're using our pytest_launcher code that uses `sys.exit` they will always pass (#2247).

tests/qemu-tests/test_qemu.sh
OBarronCS: Good catch - latest push integrates this older implementation of qemu-user testing into a new Pytest fixture, so this directory is now gone.

OBarronCS:
Latest push integrates the prior implementation of qemu-user testing (two files, test_aarch64.py and test_riscv64.py) into a Pytest fixture (`qemu_start_binary`). So now, for cross arch qemu-user tests, there are two ways of creating a test:
1. With assembly defined in Python and compiled in the `qemu_assemby_run` fixture
2. Using a .C file that is compiled (in the user/binaries folder), then passing the path to the file to the `qemu_start_binary` fixture (akin to the `start_binary` used in gdb-tests.

The old `test_qemu.sh` was deleted (as it's functionality was integrated into the tests.py file, and the folder structure of qemu-tests slightly altered to mimic gdb-tests.

The Makefile was edited to be slightly more generic (allowing more than one .C file per arch)

tests/tests.py
```diff
@@ -55,14 +95,14 @@ def getTestsList(collect_only: bool, test_name_filter: str, gdbinit_path: str) -
         exit(0)
 
     # Extract the test names from the output using regex
-    pattern = re.compile(r"tests/.*::.*")
+    pattern = re.compile(rf"{test_dir_path}.*::.*")
```
OBarronCS, line 98: The output of pytests_collect.py that we care about puts the name of each test on a new line starting with:

`Test: `

Perhaps we change this regex to search for something that starts with this?

.github/workflows/tests.yml
OBarronCS: Is this the case just on the GitHub CI?

On the latest push, when I run `docker compose run -T ubuntu24.04 ./tests.sh -t qemu-user` locally, and intentionally make a test break (add assert False or something), then it does turn up red and the test shows as failing.

OBarronCS:
One small quirk to work out:

On the aarch64 test on the .C file, it looks like ASLR is enabled. The address of main changes every time the program executes - this is likely related to having a newer version of qemu/aarch64 compiler given the Ubuntu 24.04 image.

image

.github/workflows/tests.yml
gsingh93: You'll see the message that the test failed, but the exit code of the process is likely still zero due to the bug I linked above. This means it won't show up as an error on GH CI. (feel free to test it)

gsingh93:
> On the aarch64 test on the .C file, it looks like ASLR is enabled

You can try `setarch` with the `-R` flag, but no clue if it works with a binary run inside qemu-user.

Otherwise you can just call `personality(ADDR_NO_RANDOMIZE)` from within the C file.

.github/workflows/tests.yml
OBarronCS: I see - I can make fix proposed in #2247 in this PR.

OBarronCS:
`setarch -R` did the trick.

Last thing to figure out: when to invoke the pytest unit tests. These are currently always invoked alongside the top level ./tests.sh, which mean are run regardless of if we pass `-t qemu-user` or not as a flag. How do we want to determine when each test gets run?

tests/qemu-tests/conftest.py
```diff
@@ -0,0 +1,109 @@
...
+
+        global _start_binary_called
+        # if _start_binary_called:
+        #     raise Exception('Starting more than one binary is not supported in pwndbg tests.')
+
+        _start_binary_called = True
+
+    yield _start_binary
+
+    qemu.terminate()
```
disconnect3d:
```suggestion
    qemu.kill()
```

Shall we `use qemu.kill()` instead?

Per the [Popen.terminate()](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.terminate) and kill docs, one sends `SIGTERM` and another sends `SIGKILL` and per `man 7 signal`:

> The signals SIGKILL and SIGSTOP cannot be caught, blocked, or ignored.

So if we want to have 100% reliability on plz kill that qemu process then we should probably use `.kill()` here :).

OBarronCS:
Latest commit fixes #2247 with by replacing `sys.exit()` in the tests and gdbinit.py with `os._exit()`.

.github/workflows/tests.yml
OBarronCS: Fixed.

tests/qemu-tests/tests/user/test_aarch64_reference.py
```diff
@@ -1,15 +1,17 @@
...
 
-try:
+REFERENCE_BINARY = user.binaries.get("reference-binary.aarch64.out")
+
+
+def test_aarch64_reference(qemu_start_binary):
+    qemu_start_binary(REFERENCE_BINARY, "aarch64")
     gdb.execute("break break_here")
-    assert pwndbg.gdblib.symbol.address("main") == 0x5500000A1C
+    # assert pwndbg.gdblib.symbol.address("main") == 0x7FFFF7A14A1C
```
OBarronCS: Commented this test out as is was flaky (even with setarch -R). If `break_here` (line above) is a symbol that is found, main should be found as well.

OBarronCS:
I renamed `qemu-user` tests to `cross-arch` because I think it's more descriptive to the purpose of the tests. I made a (temporary?) change to `./tests.sh` so that it doesn't invoke the unit tests when running cross arch tests.

.github/workflows/tests.yml
```diff
@@ -79,6 +79,10 @@ jobs:
         ./setup.sh
         ./setup-dev.sh
         mkdir .cov
+  
+    - name: Run cross-architecture tests
+      run: |
+        ./tests.sh -t cross-arch
```
OBarronCS: I moved these lines above the qemu system tests because they execute significantly faster - allows faster feedback for them. I noticed that if these fail, then the subsequent kernel testing doesn't take place. Are we okay with that?

tests/qemu-tests/tests/user/test_aarch64_reference.py
gsingh93: If `break_here` is found, that means GDB works correctly, but it doesn't mean `pwndbg.gdblib.symbol.address` works correctly. At the very least we should check that it doesn't return None.

We could execute all binaries with a wrapper that calls `personality` if we really want to test without ASLR, but that's probably not needed.

OBarronCS: That makes sense. I made these checks `is not None`, and reverted the `setarch -R` change as it wasn't as stable as initially tested.

tests/qemu-tests/conftest.py
```diff
@@ -0,0 +1,105 @@
...
+        )
+
+        gdb.execute(f"target remote :{QEMU_PORT}")
+        gdb.execute("set exception-verbose on")
+
+        global _start_binary_called
+        # if _start_binary_called:
+        #     raise Exception('Starting more than one binary is not supported in pwndbg tests.')
+
+        _start_binary_called = True
```
OBarronCS, line 58: These lines (copied from gdb-tests fixtures) are likely remnants of an older testing system that ran multiple tests in the same GDB session. If this is the case, would it be best to remove them?

OBarronCS:
These last commits separates the 3 types of tests into three separate top level scripts - `./tests.sh`, `./unit-tests.sh`, and `./qemu-tests.sh`, which makes all tests pass now.

I also added other cross-compilers to the Ubuntu dev setup script, because those will be needed for compiling Arm, and mips.

OBarronCS:
Code coverage is broken for qemu tests (both qemu-user and the kernel tests) - the `./.cov` directory is empty after the tests. The paths for the environment variables that set that up should be correct.

Could this be related to the qemu tests now running 24.04 Ubuntu, instead of 22.04? EDIT: it's not this - I see the same behavior on 22.04

OBarronCS:
Code coverage fixed - it was an issue with os._exit bypassing the exit handler https://github.com/nedbat/coveragepy/issues/310

OBarronCS:
Kernel tests were failing on 24.04 - since this PR doesn't relate to those tests, I reverted kernel tests to run on 22.04, and make the qemu-user tests run in a new workflow.

tests/tests.py
disconnect3d: if that is still true and makes sense then sure

tests/qemu-tests/conftest.py
disconnect3d: Yes, lets remove them; can be in another PR since I will probably just hit merge on this if the code is good to go

tests/qemu-tests/tests/user/test_aarch64_example.py
```diff
@@ -0,0 +1,136 @@
...
+    assert future_syscall_ins.syscall_name == "exit"
+
+    gdb.execute("stepuntilasm svc")
+
+    # Both for emulation and non-emulation, ensure a syscall at current PC gets enriched
+    instructions = pwndbg.gdblib.disasm.emulate_one(), pwndbg.gdblib.disasm.no_emulate_one()
+
+    for i in instructions:
+        assert i.syscall == 93
+        assert i.syscall_name == "exit"
```
disconnect3d, line 48: Hmm eventually, we want the context display to look correctly; wouldn't it make more sense to assert on context display?

Those tests are good so far; its just that a context display check would correspond to the exact thing user expects to work properly; while current asserts may still pass if context would not display the enriched syscalls.

disconnect3d (approved):

tests/qemu-tests/tests/user/test_aarch64_example.py
OBarronCS: That's a good call - future tests will focus on what the user sees 👍

PR #2278, opened 2024-07-03
Add bitwise math helper functions

OBarronCS:
This PR adds some helper functions that perform bitwise rotations and conversion between unsigned-to-signed on numbers of a given bit-width.

In nearly every Arm instruction, you can optionally apply shift operations, such as in the following instructions:
```
add x5, x5, #1, lsl #12
str w2, [x0,x1,lsl #2]
```
The `lsl` here represents a type of shift operation to be done on a source operands before a computation. There are others, such as `ror` (rotate right), `asr` (arithmetic shift right), and `lsr` (logical shift right). These helper functions can be used to manually resolve the values after the shift operation as taken place.

I added some unit tests to test these new functions as well. I didn't find a perfect place to put these functions, so for now I placed them in a new folder in `lib`. If we have a better place for these, we can move them.

### DIFF
--- a/pwndbg/gdblib/disasm/riscv.py
+++ b/pwndbg/gdblib/disasm/riscv.py
@@ -29,10 +29,6 @@ class DisassemblyAssistant(pwndbg.gdblib.disasm.arch.DisassemblyAssistant):
         else:
             src2_unsigned = 0
 
-        if self.architecture not in {"rv32", "rv64"}:
-            # TODO: figure out what this check is for. Are there other riscv arch types in GDB?
-            raise NotImplementedError(f"architecture '{self.architecture}' not implemented")
-
         src1_signed = bit_math.to_signed(src1_unsigned, pwndbg.gdblib.arch.ptrsize * 8)
         src2_signed = bit_math.to_signed(src2_unsigned, pwndbg.gdblib.arch.ptrsize * 8)
 
--- a/pwndbg/gdblib/disasm/riscv.py
+++ b/pwndbg/gdblib/disasm/riscv.py
@@ -7,6 +7,7 @@ from typing_extensions import override
 import pwndbg.gdblib.arch
 import pwndbg.gdblib.disasm.arch
 import pwndbg.gdblib.regs
+import pwndbg.lib.disasm.helpers as bit_math
 from pwndbg.emu.emulator import Emulator
 from pwndbg.gdblib.disasm.instruction import InstructionCondition
 from pwndbg.gdblib.disasm.instruction import PwndbgInstruction
@@ -28,14 +29,12 @@ class DisassemblyAssistant(pwndbg.gdblib.disasm.arch.DisassemblyAssistant):
         else:
             src2_unsigned = 0
 
-        if self.architecture == "rv32":
-            src1_signed = src1_unsigned - ((src1_unsigned & 0x80000000) << 1)
-            src2_signed = src2_unsigned - ((src2_unsigned & 0x80000000) << 1)
-        elif self.architecture == "rv64":
-            src1_signed = src1_unsigned - ((src1_unsigned & 0x80000000_00000000) << 1)
-            src2_signed = src2_unsigned - ((src2_unsigned & 0x80000000_00000000) << 1)
-        else:
+        if self.architecture not in {"rv32", "rv64"}:
+            # TODO: figure out what this check is for. Are there other riscv arch types in GDB?
             raise NotImplementedError(f"architecture '{self.architecture}' not implemented")
+
+        src1_signed = bit_math.to_signed(src1_unsigned, pwndbg.gdblib.arch.ptrsize * 8)
+        src2_signed = bit_math.to_signed(src2_unsigned, pwndbg.gdblib.arch.ptrsize * 8)
 
         condition = {
             RISCV_INS_BEQ: src1_signed == src2_signed,
--- a/pwndbg/lib/disasm/helpers.py
+++ b/pwndbg/lib/disasm/helpers.py
@@ -1,0 +1,48 @@
+from __future__ import annotations
+
+
+def to_signed(unsigned: int, bit_width: int):
+    """
+    Returns the signed number associated with the two's-complement binary representation of `unsigned`
+    """
+    extract_bit = 1 << (bit_width - 1)
+    return unsigned - ((unsigned & extract_bit) << 1)
+
+
+def logical_shift_left(n: int, shift_amt: int, bit_width: int):
+    return (n << shift_amt) & ((1 << bit_width) - 1)
+
+
+def logical_shift_right(n: int, shift_amt: int, bit_width: int):
+    """
+    `n` is truncated to the width of `bit_width` before the operation takes place.
+    """
+    n = n & ((1 << bit_width) - 1)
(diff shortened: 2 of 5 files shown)


pwndbg/gdblib/disasm/riscv.py
```diff
@@ -28,15 +29,13 @@ def _is_condition_taken(
...
 
-        if self.architecture == "rv32":
-            src1_signed = src1_unsigned - ((src1_unsigned & 0x80000000) << 1)
-            src2_signed = src2_unsigned - ((src2_unsigned & 0x80000000) << 1)
-        elif self.architecture == "rv64":
-            src1_signed = src1_unsigned - ((src1_unsigned & 0x80000000_00000000) << 1)
-            src2_signed = src2_unsigned - ((src2_unsigned & 0x80000000_00000000) << 1)
-        else:
+        if self.architecture not in {"rv32", "rv64"}:
+            # TODO: figure out what this check is for. Are there other riscv arch types in GDB?
```
disconnect3d: cc: @patryk4815 may know.

pwndbg/gdblib/disasm/riscv.py
```diff
@@ -28,15 +29,13 @@ def _is_condition_taken(
...
             raise NotImplementedError(f"architecture '{self.architecture}' not implemented")
 
+        src1_signed = bit_math.to_signed(src1_unsigned, pwndbg.gdblib.arch.ptrsize * 8)
+        src2_signed = bit_math.to_signed(src2_unsigned, pwndbg.gdblib.arch.ptrsize * 8)
```
disconnect3d, line 33: Can u move the `ptrsize * 8` calculation to a variable?

Also, we probably should have sth like `pwndbg.gdblib.arch.ptrbytes`, heh:
```
(.venv) root@pwndbg:~/pwndbg# rg 'ptrsize \* 8'
pwndbg/gdblib/memory.py
272:        size = pwndbg.gdblib.arch.ptrsize * 8

pwndbg/gdblib/onegadget.py
139:                bits = pwndbg.gdblib.arch.ptrsize * 8

pwndbg/gdblib/arch.py
89:    pwnlib.context.context.bits = ptrsize * 8

pwndbg/commands/dumpargs.py
27:            f"Detected ABI: {pwndbg.gdblib.arch.name} ({pwndbg.gdblib.arch.ptrsize * 8} bit) either doesn't pass arguments through registers or is not implemented. Maybe they are passed on the stack?"
```

disconnect3d: (In case you want to add `ptrbytes`, pls do so in a different PR)

disconnect3d (approved):
Generally LGTM though I wonder if we something like `ctypes.c_uint64(...)` would be better?

I created a lib for having C types in Python that wraps the `ctypes.c_*` types and adds operators into it: https://github.com/disconnect3d/cint which could maybe be helpful here?

tests/unit-tests/test_bit_math.py
```diff
@@ -0,0 +1,81 @@
+from __future__ import annotations
+
+import sys
+from unittest.mock import MagicMock
+
+# Replace `pwndbg.commands` module with a mock to prevent import errors, as well
+# as the `load_commands` function
+module_name = "pwndbg.commands"
+module = MagicMock(__name__=module_name, load_commands=lambda: None)
+sys.modules[module_name] = module
```
disconnect3d: Why do we need this?

pwndbg/gdblib/disasm/riscv.py
patryk4815: if gdb don't detect, he will put just `riscv` type without rv32/rv64.
```
set architecture riscv
riscv       riscv:rv32  riscv:rv64 
```

Eg at `gdb remote` start you need to force `set arch riscv:rv32` or `set arch riscv:rv64` in some cases.

OBarronCS: I see - good to know!

OBarronCS:
I'll look into the ctypes - the library you made looks really helpful! In theory ctypes would be more efficient and require less lines of code as we don't need the gymnastics of dealing with Python's infinite-width numbers.

pwndbg/gdblib/disasm/riscv.py
disconnect3d: Perhaps we should add some comment near the `self.architecture` definition to inform developers on that thing?

disconnect3d: or: tell the user here in the error we raise they have to set arch riscv:rv32 or rv64?

tests/unit-tests/test_bit_math.py
gsingh93: Everyone is just copying and pasting the preamble from the other tests 😛 We should probably do this in a file called `mocks/commands.py` or something, so the user can choose to import it if they want. But that's out of scope for this PR.

pwndbg/gdblib/disasm/riscv.py
OBarronCS: Did some more digging: it looks like our detection for the architecture will fallback to `rv64` on the side of Pwndbg if it cannot determine 32 v 64 bit.

https://github.com/pwndbg/pwndbg/blob/b74fcdde97a4cd4c6c072235040eb21ee7a42a5b/pwndbg/gdblib/arch.py#L97-L101

Some kind of alert to the user that the architecture could not be completely determined would be good.

OBarronCS: I added a comment in our arch detection function regarding this.

pwndbg/gdblib/disasm/riscv.py
OBarronCS: I'll make a note to add this when refactoring our Architecture class.

tests/unit-tests/test_bit_math.py
OBarronCS: Oop - yep, just copied them from the other files. Removed them now.

OBarronCS:
I did some playing around with c_types. To tell the truth, they didn't reduce the code complexity all that much, because we need to manually do the `rotate_right` in Python anyways, have to choose between unsigned and signed C integers, and are locked to either 8, 16, 32, or 64 bits (although I can't currently think of scenario where we would want anything other than these).

If we are okay with doing these in raw Python and not c_types, this PR is ready to go so we can use these functions to evaluate certain operands in Arm/AArch64.

pwndbg/lib/disasm/helpers.py
```diff
@@ -0,0 +1,48 @@
+from __future__ import annotations
+
+
+def to_signed(unsigned: int, bit_width: int):
+    """
+    Returns the signed number associated with the two's-complement binary representation of `unsigned`
+    """
+    extract_bit = 1 << (bit_width - 1)
+    return unsigned - ((unsigned & extract_bit) << 1)
```
disconnect3d, line 9: Just a note: only the `to_signed` function is used from here for now. I wonder if we really need other functions, but I guess/hope we do.

disconnect3d:
> If we are okay with doing these in raw Python and not c_types, this PR is ready to go so we can use these functions to evaluate certain operands in Arm/AArch64.

Lets merge it then

PR #2280, opened 2024-07-03
Expand Debugger-agnostic API and split `pwndbg.dbg.evaluate_expression`

mbrla0:

This PR expands the Debugger-agnostic API in preparation of future porting (with the `Thread` class), and in service of splitting `pwndbg.dbg.evaluate_expression` into a global, per-process version, and a local, per frame version, better reflecting the way LLDB scopes expression evaluation, and that also matches GDB's new `global_contex` parameter for `parse_and_eval`.

This PR is also the first in a series of PRs that port `ArgparsedCommand` to the Debugger-agnostic API and to LLDB.


mbrla0:
Unfortunately, it looks like the version we're using of the gdb-stubs doesn't have the `global_context` keyword for `gdb.parse_and_eval`, so `mypy` flags it as invalid.

mbrla0:
@gsingh93 The fix to the stubs got merged. How long before do you reckon it will be before they publish a new version? The newest published version is from around two months ago. Should we maybe start pulling the head of the repo directly as a dependency in pyproject, rather than stick to the last published version?

gsingh93:
> @gsingh93 The fix to the stubs got merged. How long before do you reckon it will be before they publish a new version? The newest published version is from around two months ago. Should we maybe start tracking the repo directly?

Each stub in the repo is published individually, usually within 24 hours of the last modification to that particular stub. If the last release was two months ago, that just means this is the first modification since then.

When you see the new version out, you can bump the version in the poetry.toml and run `poetry lock`.

mbrla0:
@gsingh93 Alright, everything's up to date now.

mbrla0:
@gsingh93 I think this is ready for review, btw.

pwndbg/commands/__init__.py
```diff
@@ -220,15 +224,21 @@ def fix(
     if isinstance(arg, pwndbg.dbg_mod.Value):
         return arg
 
+    frame = pwndbg.dbg.selected_frame()
+    target: pwndbg.dbg_mod.Frame | pwndbg.dbg_mod.Process = (
+        frame if frame else pwndbg.dbg.selected_inferior()
+    )
```
gsingh93, line 230: Should we add a `pwndbg.dbg.evaluate_expression` method that does this and then calls `target.evaluate_expression()`?

pwndbg/dbg/gdb.py
```diff
@@ -17,6 +17,61 @@
...
+    def __init__(self, inner: gdb.Inferior):
+        self.inner = inner
+
+    @override
+    def evaluate_expression(self, expression: str) -> pwndbg.dbg_mod.Value:
+        try:
+            return GDBValue(gdb.parse_and_eval(expression, True))
+        except TypeError:
+            # Some earlier versions of GDB don't support the global_context argument.
+            return GDBValue(gdb.parse_and_eval(expression))
```
gsingh93: Since we do this twice in this file, just with a different `global_context` value, it might be good to have a helper method in this function that does this for us, and use it here and in `GDBFrame`? I just like to consolidate as many `try/except` blocks as I can into helper functions so the rest of the code becomes easier to read.

pwndbg/dbg/gdb.py
```diff
@@ -17,6 +17,61 @@
...
+    def __init__(self, inner: gdb.Frame):
+        self.inner = inner
+
+    @override
+    def evaluate_expression(self, expression: str) -> pwndbg.dbg_mod.Value:
+        selected = gdb.selected_frame()
+        restore = False
+        if selected != self.inner:
+            self.inner.select()
+            restore = True
```
gsingh93, line 43: It would be neat if we had a Python ContextManager (https://realpython.com/python-with-statement/) for switching to a different frame (but this is out of scope for this PR).

pwndbg/commands/__init__.py
mbrla0: Given that this happens only twice in this PR, maybe it makes sense to wait and see how common the use of that `pwndbg.dbg.evaluate_expression` would actually be? Pwndbg seems to use `parse_and_eval` very rarely outside of `gdblib`, and in most of those there might be better ways to do what's being done with it, especially with the LLDB port in mind.

Plus, that construction I feel communicates quickly that it's choosing to evaluate in a narrower, then broader context, in that order. So, if it's only really gonna be used these two times, I kind of prefer the longer form.

Either way, I feel like it's fairly easy to just add that in later if it that pattern ends up spreading out. But if you feel strongly about that, I can add it in, too.

pwndbg/dbg/gdb.py
mbrla0: Yeah, that makes sense. I'll put them both in a single helper function.

pwndbg/dbg/gdb.py
mbrla0: I agree! Actually, this might be a great idea to incorporate into the next PR, which among other things has that function wrapping GDB errors in an error type of our own. Using `with` should make exiting our temporarily selected frame - regardless of success or error - pretty clean and concise.

mbrla0:
@gsingh93 anything else?

pwndbg/dbg/gdb.py
```diff
@@ -17,6 +17,64 @@
 from pwndbg.gdblib import load_gdblib
 
 
+def parse_and_eval(expression: str, global_context: bool) -> gdb.Value:
+    """
+    Same as `gdb.parse_and_eval`, but only uses `global_context` if it is
+    supported by the current version of GDB.
```
disconnect3d, line 23: Pls add info what GDB version allows for `global_context`

pwndbg/commands/__init__.py
```diff
@@ -220,15 +224,21 @@ def fix(
...
+    assert target, "Reached command expression evaluation with no frame or inferior"
+
     try:
-        parsed = pwndbg.dbg.evaluate_expression(arg)
+        parsed = target.evaluate_expression(arg)
         return parsed
```
disconnect3d: Just `return target.evaluate_expression(arg)`?

pwndbg/dbg/gdb.py
mbrla0: I've added it, now.

pwndbg/commands/__init__.py
gsingh93: No need to add it now, I thought having a generic method in `dbg` that took care of this for you without having to think about the details of frames would be useful, but let's just add that when/if it becomes necessary.

pwndbg/commands/__init__.py
mbrla0: Oh, yeah, right, should be good now.

gsingh93 (approved):
LGTM, @disconnect3d can merge if he's also finished with his review.

Issue #292, opened 2017-06-04
Show ARM/Thumb mode (Feature request)


techvoltage:
Would like to know whether I'm running in ARM/Thumb mode in gdb

disconnect3d:
Isn't it known from the instructions itself or from $cpsr (or $xpsr on some ARMs) register?

Anyway, getting the value is pretty easy - we do parse it actually just for $cpsr (so yeah, ARMs with $xpsr are not handled at all as decribed here https://github.com/pwndbg/pwndbg/issues/264) in here: https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/disasm/__init__.py#L86-L87

I would probably go with adding a `pwndbg.arch.mode` which would be `'ARM'` or `'THUMB'` for ARMs and just `None` for other archs.

The question is - where do we want to display it?

One idea would be displaying it in the `DISASM` bar, e.g. instead of:
`[──────────DISASM──────────]`

We can display:
`[──────────DISASM  - ARM MODE──────────]`

This DISASM banner can be changed easily, just change the string in here https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/commands/context.py#L154.

As this is rather an easy enhancement, would you try to add it @techvoltage? :)

techvoltage:
Now that you've specified the how and where, would be glad to work on it!

zachriggle:
If the address is odd, it's Thumb.

disconnect3d:
Just for reference: we had a PR for that in https://github.com/pwndbg/pwndbg/pull/293 but it haven't been merged finally.

disconnect3d:
Done in #2281

PR #2281, opened 2024-07-03
Arm mode in banner

OBarronCS:
This PR adds text to the context banner in Arm32 indicating whether it is currently in Arm or Thumb mode

Arm mode:
arm_mode

Thumb mode:
arm_mode_thumb

The double `arm` in arm mode is a little strange - so if there are suggestions on other ways to display this, please let me know.

Related issue: #292

### DIFF
--- a/pwndbg/gdblib/arch.py
+++ b/pwndbg/gdblib/arch.py
@@ -41,6 +41,29 @@ pwnlib_archs_mapping = {
     "rv64": "riscv64",
 }
 
+def read_thumb_bit() -> int | None:
+    """
+    Return 0 or 1, representing the status of the Thumb bit in the current Arm architecture
+    
+    Return None if the Thumb bit is not relevent to the current architecture
+    """
+    if pwndbg.gdblib.arch.current == "arm":
+        # When program initially starts, cpsr may not be readable
+        cpsr = pwndbg.gdblib.regs.cpsr
+        if cpsr is not None:
+            return (cpsr >> 5) & 1
+    elif pwndbg.gdblib.arch.current == "armcm":
+        # ARM Cortex-M procesors only suport Thumb mode. However, there is still a bit
+        # that represents the Thumb mode (which is currently architecturally defined to be 1)
+        xpsr = pwndbg.gdblib.regs.xpsr
+        if xpsr is not None:
+            return (xpsr >> 24) & 1
+    # AArch64 does not have a Thumb bit
+    return None
+
+def get_thumb_mode_string() -> Literal["arm","thumb"] | None:
+    thumb_bit = read_thumb_bit()
+    return None if thumb_bit == None else "thumb" if thumb_bit == 1 else "arm"
 
 arch = Arch("i386", typeinfo.ptrsize, "little")
 
@@ -85,5 +108,6 @@ def _get_arch(ptrsize: int):
 def update() -> None:
     arch_name, ptrsize, endian = _get_arch(typeinfo.ptrsize)
     arch.update(arch_name, ptrsize, endian)
+    arch.mode = get_thumb_mode_string()
     pwnlib.context.context.arch = pwnlib_archs_mapping[arch_name]
     pwnlib.context.context.bits = ptrsize * 8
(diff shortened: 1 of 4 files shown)


pwndbg/gdblib/arch.py
```diff
@@ -42,6 +42,32 @@
...
+def read_thumb_bit() -> int | None:
+    """
+    Return 0 or 1, representing the status of the Thumb bit in the current Arm architecture
+
+    Return None if the Thumb bit is not relevent to the current architecture
+    """
+    if pwndbg.gdblib.arch.current == "arm":
+        # When program initially starts, cpsr may not be readable
+        cpsr = pwndbg.gdblib.regs.cpsr
+        if cpsr is not None:
```
disconnect3d:
```suggestion
        if (cpsr := pwndbg.gdblib.regs.cpsr) is not None:
```

Since we support Ubuntu >= 20.04 and it uses Py >= 3.8, we can use this new and fancy assignment operator.

pwndbg/gdblib/disasm/arm.py
```diff
@@ -30,10 +30,10 @@ def _condition(self, instruction: PwndbgInstruction, emu: Emulator) -> Instructi
...
         )
 
-        N = value & (1 << 31)
-        Z = value & (1 << 30)
-        C = value & (1 << 29)
-        V = value & (1 << 28)
+        N = (value >> 31) & 1
+        Z = (value >> 30) & 1
+        C = (value >> 29) & 1
+        V = (value >> 28) & 1
```
disconnect3d, line 36: I'm okay with those changes but we can as well skip them?

Not really that important, but the `1<<31` etc. is optimized here by Python's peephole optimizer while in `(value >> 31) & 1` all the operations are computed.

This can be seen here:
image

Note that they also added specialized instructions/bytecode in Python 3.11 which means that some hot used functions may be optimized/specialized for the very-often used argument types. This is also why I am providing the `show_caches=1, adaptive=1` keyword arguments here. This may also influence the benchmark result here, but oh well, its worth using the latest tech ;). Fwiw I have given a talk on the [Python specialized bytecode here](https://www.youtube.com/watch?v=RlNM5n5C_wg&t=5034s).

pwndbg/gdblib/disasm/arm.py
```diff
@@ -57,6 +57,17 @@ def _condition(self, instruction: PwndbgInstruction, emu: Emulator) -> Instructi
...
         return InstructionCondition.TRUE if bool(cc) else InstructionCondition.FALSE
 
+    @override
+    def _resolve_target(self, instruction: PwndbgInstruction, emu: Emulator | None, call=False):
+        target = super()._resolve_target(instruction, emu, call)
+        if target is not None:
+            # On interworking branches - branches that can enable Thumb mode - the target of a jump
+            # has the least significant bit set to 1. This is not actually written to the PC
+            # and instead the CPU puts it into the Thumb mode register bit.
+            # This means we can clear the least significant bit of the target.
```
disconnect3d:
```suggestion
            # This means we can clear the least significant bit of the target.
```
Technically, I believe we have to do it.

pwndbg/lib/arch.py
```diff
@@ -11,8 +11,8 @@
 class Arch:
     def __init__(self, arch_name: str, ptrsize: int, endian: Literal["little", "big"]) -> None:
         self.update(arch_name, ptrsize, endian)
-
         self.native_endian = str(sys.byteorder)
+        self.mode: Literal["arm", "thumb"] | None = None
```
disconnect3d, line 15: Ideally, only ARM arch should have it... :p

pwndbg/gdblib/disasm/arm.py
OBarronCS: In this particular case, the functionality was incorrect with the old implementation.

There were some checks for equality between the different bit flags, and the old way of extracting the bit didn't make it 0/1, but kept the bit values in their bit positions, so some equality checks didn't pass when they should.

https://github.com/pwndbg/pwndbg/blob/5c84061262055cedb1d4b4233afeba80e7144045/pwndbg/gdblib/disasm/arm.py#L49-L52

The Python opcode optimization is very interesting, I'll have to dig deeper into that.

pwndbg/gdblib/disasm/arm.py
OBarronCS: I updated the comment to indicate this.

PR #2282, opened 2024-07-05
Add Binary Ninja integration

### LINKED ISSUE #148: Binary Ninja Integration
Currently, Pwndbg supports integration with IDA via `ida_script.py` and `pwndbg/ida.py`.

Among the interesting currently- supported functionality:

- Extraction of comments
- Extraction of symbols
- Bidirectional breakpoint setting/clearing (kind of)
- Jump-to-PC / Cursor 

This functionality should also be supported with Binary Ninja (https://binary.ninja).  It should be relatively easy to hack in support.  Later, a clean re-design would be nice to have a standardized interface to e.g. add symbols, vs the symbol-query code specifically asking IDA / BN for symbols.

Aplet123:
Closes #148

Adds Binary Ninja integration to pwndbg.

**Breaking changes:**
- Changes the `set ida-enabled on` config to `set integration-provider ida` instead (is this fine, or should I add a config var for `ida-enabled` that emits a deprecation warning)

**Technical details:**
- Installs an XML-RPC server into Binary Ninja via a plugin (`binja_script.py`) that pwndbg will connect to, similar to the IDA integration (works with personal license)
- Unifies the IDA and Binary Ninja integration under a single `pwndbg.integration.provider` interface

**How to use:**
- Copy or symlink `binja_script.py` into the Binary Ninja plugins directory
- Open the file you want integration on in Binary Ninja, then go to `Plugins > pwndbg > Start integration on current view`
- Open gdb on the file, and run `set integration-provider binja`
- Integrations will now be enabled

**Features:**
The Binary Ninja integration includes all the features of the IDA integration:
- Cursor syncing with a `bn-sync` command, or automatic syncing with the `bn-autosync` config
- Displaying the current PC with an arrow in Binary Ninja
image
- Allowing the user to set GDB breakpoints on instructions from within Binary Ninja (`Plugins > pwndbg > Toggle breakpoint here`)
image
- Fetching symbols from Binary Ninja if not present in GDB
- Displaying comments from Binary Ninja inside the disassembly
image
- Pulling decompilation from Binary Ninja, along with syntax highlighting, with IL level configurable using the `bn-il-level` setting
image
- Pulling function type signatures from Binary Ninja if not present in GDB
- A `$bn_sym` function to get the address of a symbol in Binary Ninja, e.g. `$bn_sym("main")`
image
- A `$bn_eval` function to evaluate a [Binary Ninja expression](https://api.binary.ninja/binaryninja.binaryview-module.html#binaryninja.binaryview.BinaryView.parse_expression) with all of the current registers as magic variables, as well as a `$piebase` magic variable with the executable base.
image
- A `decomp` command to decompile lines near an address, similar to what `nearpc` does
image
- A `$bn_var` function to get the address of a stack variable in the current function in Binary Ninja, e.g. `$bn_sym("var_14")`
image
- Automated labeling of stack variables based on Binary Ninja variable names
image

**Potential features:**
- [ ] Make breakpoint syncing bidirectional instead of just Binary Ninja -> GDB

**Potential features outside the scope of this PR:**
- Actually add symbols/types to GDB instead of hacking around it by querying in pwndbg
- Improve symbols on indirect jumps/calls, since currently only the information about dereferenced target is in the disassembly, when sometimes the address of the target is more relevant (e.g. GOT entries)


Aplet123:
Should be mostly done and ready for review now.

binja_script.py
```diff
@@ -0,0 +1,400 @@
...
+            lines = binaryninja.function.DisassemblyTextRenderer(func).post_process_lines(
+                orig_func.lowest_address,
+                orig_func.highest_address - orig_func.lowest_address,
+                lines,
+            )
+
+        ret = []
+        for line in lines:
+            ret.append((line.address, [(tok.text, tok.type.name) for tok in line.tokens]))
+        return ret
```
disconnect3d:
```suggestion
        return [
            (line.address, [(tok.text, tok.type.name) for tok in line.tokens])
            for line in lines
        ]
```

binja_script.py
```diff
@@ -0,0 +1,400 @@
...
+        elif level == "mlil":
+            func = func.mlil_if_available
+        elif level == "hlil":
+            func = func.hlil_if_available
+        else:
+            raise ValueError(f"Cannot decompile with level {level!r}")
+        if func is None:
+            return None
+        if level == "hlil":
+            lines = func.root.lines
```
disconnect3d, line 206: This can be moved to the above `elif level == "hlil":` case

binja_script.py
```diff
@@ -0,0 +1,400 @@
...
+        elif level == "llil":
+            func = func.llil_if_available
+        elif level == "mlil":
+            func = func.mlil_if_available
+        elif level == "hlil":
+            func = func.hlil_if_available
+        else:
+            raise ValueError(f"Cannot decompile with level {level!r}")
+        if func is None:
+            return None
```
disconnect3d, line 207: If the given IL is not available, does this mean we cannot decompile with that level and we should maybe raise a ValueError as well?

binja_script.py
```diff
@@ -0,0 +1,400 @@
...
+            return (dv.symbol.full_name, dv.address)
+        return (dv.name or f"data_{dv.address:x}", dv.address)
+
+    @should_register
+    def get_comments(self, addr: int) -> List[str]:
+        """
+        Gets a list of all comments at a specified address.
+        """
+        ret = []
+        for f in sorted(self.bv.get_functions_containing(addr), key=lambda f: f.start):
```
disconnect3d: Note: `key=lambda f: f.start` the `f` is confusing here since we do `for f in sorted(...)`, lets maybe rename one of them to `func`

binja_script.py
```diff
@@ -0,0 +1,400 @@
...
+
+
+# workaround for there to be no way to get all address tags in the python API
+def get_tag_refs(bv: binaryninja.BinaryView, ty: str) -> List[binaryninja.core.BNTagReference]:
+    tag_type = bv.get_tag_type(ty)
+    count = binaryninja.core.BNGetAllTagReferencesOfTypeCount(bv.handle, tag_type.handle)
+    ref_ptr = binaryninja.core.BNGetAllTagReferencesOfType(
+        bv.handle, tag_type.handle, ctypes.c_ulong(count)
+    )
+    return [ref_ptr[i] for i in range(count)]
```
disconnect3d:
```suggestion
    return ref_ptr[:count]
```
or `return list(ref_ptr[:count])`?

binja_script.py
Aplet123: I'm not sure the conditions where IL won't be available, but I don't think it warrants showing an error the user every single time they try to decompile anything. I'll probably change the error message to something like `{level!r} is not a recognized IL level. Supported values are: disasm, llil, mlil, hlil` to make it clearer though.

pwndbg/chain.py
```diff
@@ -147,6 +148,12 @@ def format(
         symbol = pwndbg.gdblib.symbol.get(link) or None
         if symbol:
             symbol = f"{link:#x} ({symbol})"
+        else:
+            page = pwndbg.gdblib.vmmap.find(link)
+            if page is not None and "[stack" in page.objfile:
```
disconnect3d:
```suggestion
            if page and "[stack" in page.objfile:
```

binja_script.py
Aplet123: `ref_ptr` is a ctypes pointer but it turns out slicing is supported anyways so I'll change that.

pwndbg/color/memory.py
```diff
@@ -51,6 +52,12 @@ def attempt_colorized_symbol(address: int) -> str | None:
     symbol = pwndbg.gdblib.symbol.get(address) or None
     if symbol:
         return get(address, symbol)
+    else:
+        page = pwndbg.gdblib.vmmap.find(address)
+        if page is not None and "[stack" in page.objfile:
```
disconnect3d: There is `page.is_stack()` fwiw, but it checks for `[stack]` objfile, maybe we should relax this check to look for `[stack` instead -- and then use this method here and elsewhere.

At the same time... maybe we should add `pwndbg.gdblib.vmmap.is_stack(addr)`? cc: @gsingh93  wdyt?

pwndbg/commands/binja.py
```diff
@@ -0,0 +1,78 @@
...
+import pwndbg.integration.binja
+from pwndbg.color import message
+from pwndbg.commands import CommandCategory
+
+
+@pwndbg.commands.ArgparsedCommand(
+    "Synchronize Binary Ninja's cursor with GDB.",
+    category=CommandCategory.INTEGRATIONS,
+    command_name="bn-sync",
+    aliases=["bns"],
```
disconnect3d, line 20: Maybe: bn or binja?

pwndbg/commands/context.py
```diff
@@ -747,23 +754,17 @@ def context_code(target=sys.stdout, with_banner=True, width=None):
...
 
-    # Try getting source from IDA Pro Hex-Rays Decompiler
-    if not pwndbg.ida.available():
-        return []
-
-    n = int(int(int(source_code_lines) / 2))  # int twice to make it a real int instead of inthook
-    # May be None when decompilation failed or user loaded wrong binary in IDA
-    code = pwndbg.ida.decompile_context(pwndbg.gdblib.regs.pc, n)
+    if should_decompile:
+        # May be None when decompilation failed or user loaded wrong binary in IDA
```
```suggestion
        # None when decompilation failed
```

pwndbg/integration/binja.py
```diff
@@ -0,0 +1,534 @@
...
+        decomp: List[Tuple[int, List[Tuple[str, str]]]] | None = _bn.decompile_func(
+            l2r(addr), bn_il_level.value
+        )
+        if not decomp:
+            return None
+        decomp = [
+            (r2l(addr), toks) for (addr, toks) in decomp if not all(t[0].isspace() for t in toks)
+        ]
+        ind = min(
+            [(i, x) for (i, x) in enumerate(decomp) if x[0] >= addr],
```
disconnect3d:
```suggestion
            ((i, x) for (i, x) in enumerate(decomp) if x[0] >= addr),
```

pwndbg/gdblib/regs.py
```diff
@@ -88,26 +89,29 @@ class module(ModuleType):
     last: Dict[str, int] = {}
 
     @pwndbg.lib.cache.cache_until("stop", "prompt")
-    def __getattr__(self, attr: str) -> int | None:
-        attr = attr.lstrip("$")
+    def read_reg(self, reg: str, frame: gdb.Frame | None = None) -> int | None:
```
disconnect3d, line 92: Why do we need it?

pwndbg/gdblib/regs.py
```diff
@@ -88,26 +89,29 @@ class module(ModuleType):
...
+            if reg == "pc" and pwndbg.gdblib.arch.name == "i8086":
                 if self.cs is None:
                     return None
                 value += self.cs * 16
             return int(value) & pwndbg.gdblib.arch.ptrmask
         except (ValueError, gdb.error):
             return None
 
+    def __getattr__(self, attr: str) -> int | None:
+        return self.read_reg(attr)
```
disconnect3d, line 113: Can we just do: `__getattr__ = self.read_reg`?

pwndbg/lib/cache.py
```diff
@@ -138,7 +138,13 @@ def decorator(*a: P.args, **kw: P.kwargs) -> T:
                 # Check if the value is in the cache; if we have a cache miss,
                 # we return a special singleton object `_NOT_FOUND_IN_CACHE`. This way
                 # we can also cache a result of 'None' from a function
-                cached_value: Any = cache.get(key, _NOT_FOUND_IN_CACHE)
+                no_cache = False
+                try:
+                    cached_value: Any = cache.get(key, _NOT_FOUND_IN_CACHE)
+                except TypeError:
+                    # skip caching unhashable arguments
```
disconnect3d, line 144: When does this occur? Which functions? Can we make some exception just for them, like, another decorator that would try except? The cache function is executed many many times and I'd prefer to not add additional logic if not needed

pwndbg/integration/__init__.py
```diff
@@ -0,0 +1,153 @@
...
+    global provider
+    if not provider_name.value or provider_name.value == "none":
+        provider = IntegrationProvider()
+    elif provider_name.value == "binja":
+        # do not import at start of file to avoid circular import
+        import pwndbg.integration.binja
+
+        provider = ConfigurableProvider(pwndbg.integration.binja.BinjaProvider())
+    elif provider_name.value == "ida":
+        import pwndbg.integration.ida
```
disconnect3d, line 144: Just to be sure, did we try IDA Pro? If not, we can probably remove ida.py at all, since I believe it is not working for a long time already.

Aplet123: Haven't tried it since I don't have a license, but it would be nice to get it working again (and maybe to unify Ghidra under the same API)

pwndbg/lib/cache.py
Aplet123: I need it because `read_reg` can take in an optional `gdb.Frame` which is unhashable. I could rewrite the code so that there is only extra logic in the `except` case, which should have almost no performance hit, [at least since Python 3.11](https://github.com/python/cpython/issues/84403).

pwndbg/gdblib/regs.py
Aplet123: Would have to be `__getattr__ = read_reg` but also mypy hates it and won't typecheck attribute lookups properly.

disconnect3d:
@Aplet123 pls solve conflicts :<

pwndbg/commands/__init__.py
```diff
@@ -683,82 +683,81 @@ def HexOrAddressExpr(s: str) -> int:
 
 def load_commands() -> None:
     # pylint: disable=import-outside-toplevel
-    import pwndbg.dbg
-
-    if pwndbg.dbg.is_gdblib_available():
```
disconnect3d: @Aplet123 this needs to stay, that's part of making Pwndbg debugger-agnostic and making it available for LLDB.

Can you rebase this PR over the latest dev and fix conflicts?

disconnect3d (changes requested):
Please rebase over latest dev

disconnect3d (approved):

PR #2283, opened 2024-07-05
refactor: make allocator fastbins call arena fastbins

k4lizen:
There seems to be duplicate code here, so I just cleaned it up a bit.

### DIFF
--- a/pwndbg/gdblib/heap/ptmalloc.py
+++ b/pwndbg/gdblib/heap/ptmalloc.py
@@ -758,8 +758,7 @@ class Arena:
         fd_offset = pwndbg.gdblib.arch.ptrsize * 2
         safe_lnk = pwndbg.glibc.check_safe_linking()
         result = Bins(BinType.FAST)
-        num_fastbins = 7
-        for i in range(num_fastbins):
+        for i in range(NFASTBINS):
             size += pwndbg.gdblib.arch.ptrsize * 2
             chain = pwndbg.chain.get(
                 int(self.fastbinsY[i]),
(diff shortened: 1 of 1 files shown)


pwndbg/gdblib/heap/ptmalloc.py
```diff
@@ -758,7 +758,8 @@ def fastbins(self) -> Bins:
         fd_offset = pwndbg.gdblib.arch.ptrsize * 2
         safe_lnk = pwndbg.glibc.check_safe_linking()
         result = Bins(BinType.FAST)
-        for i in range(7):
+        num_fastbins = 7
+        for i in range(num_fastbins):
```
disconnect3d: I believe we don't need that change, just a comment is probably sufficient

CptGibbon: Perhaps use the NFASTBINS constant
https://github.com/pwndbg/pwndbg/blob/21b87df06a6c93186b3438076d7c97d41de6ed82/pwndbg/gdblib/heap/ptmalloc.py#L68

CptGibbon (approved):

pwndbg/gdblib/heap/ptmalloc.py
```diff
@@ -758,8 +758,7 @@ def fastbins(self) -> Bins:
         fd_offset = pwndbg.gdblib.arch.ptrsize * 2
         safe_lnk = pwndbg.glibc.check_safe_linking()
         result = Bins(BinType.FAST)
-        num_fastbins = 7
-        for i in range(num_fastbins):
+        for i in range(NFASTBINS):
```
disconnect3d, line 761: if NFASTBINS=10 why was this 7 before in the first place? @CptGibbon

k4lizen: Because the array for fastbins in glibc is 10 pointers long, but only seven are ever used. You can probably increase the number of fastbins used via a macro / mallopt.

PR #2289, opened 2024-07-06
Improve unsorted,small,large bins corruption check

### LINKED ISSUE #1196: Heap chain corruption check is incorrect for long free lists

### LINKED ISSUE #2287: [bug] bin_at gives python exception when a bck arena ptr is corrupted
```py
bin_at():
         ....
        chain_fd = get_chain(front, fd_offset)
        chain_bk = get_chain(back, bk_offset)

        # check if bin[index] points to itself (is empty)
        if len(chain_fd) == len(chain_bk) == 2 and chain_fd[0] == chain_bk[0]:
            chain_fd = [0]
            chain_bk = [0]

        # check if corrupted
        elif chain_fd[:-1] != chain_bk[:-2][::-1] + [chain_bk[-2]]: # CRASHES HERE
            is_chain_corrupted = True

        print("fd: ", chain_fd)
        print("bk: ", chain_bk)
        return (chain_fd, chain_bk, is_chain_corrupted)
```
When the bck pointer of a {unsorted, small, large}bin gets modified to point to an invalid address `chain_bk = get_chain(back, bk_offset)` will only return one element and `chain_bk[-2]` will get list index out of range.

k4lizen:
Closes https://github.com/pwndbg/pwndbg/issues/1196 and Closes https://github.com/pwndbg/pwndbg/issues/2287.
This PR is based off of https://github.com/pwndbg/pwndbg/pull/1564.
+ Added a new heap parameter: `heap_corruption_check_limit` which states for how many chunks the corruption check is done
+ As opposed to the previous PR, every chunk has its fd/bk pointers checked instead of comparing the two lists. This results in more thorough checking when the forward and backward lists do not overlap fully.
+ `heap_corruption_check_limit` is allowed to have both a lower and a higher value than `heap-dereference-limit` which means the name "heap-dereference-limit" isn't really fitting anymore, but okay

Note:
`heap-dereference-limit` isn't really properly handled when its negative (pwndbg crashes) or zero, gdb defaults to this message when its set to zero:
```bash
pwndbg> set heap-dereference-limit 0
Set number of chunks to dereference in each bin to 'unlimited'.
``` 
but thats not actually what happens, as one chunk is printed. I set it so that `heap_corruption_check_limit` acts similarly i.e. no corruption check is done when it's set to zero. This has the unfortunate consequence that when `heap_corruption_check_limit` is set to zero, and the bin arena pointers are 0x0 (corrupted), the bin reads as "empty". I'd say this is moreso a logic error in the function that calls bin_at but whatever.

I generally tried to make it so that both parameters don't freak out when they have weird values.

### DIFF
--- a/pwndbg/gdblib/heap/__init__.py
+++ b/pwndbg/gdblib/heap/__init__.py
@@ -52,8 +52,8 @@ heap_chain_limit = add_heap_param(
     "heap-dereference-limit", 8, "number of chunks to dereference in each bin"
 )
 
-heap_corruption_check_limit = add_heap_param("heap-corruption-check-limit", 64, "lenght limit for bins which are fully checked for corruption")
-heap_corruption_check_limit += 1 # Make sure bins that are of default length are checked
+heap_corruption_check_limit = add_heap_param("heap-corruption-check-limit", 64, "amount of chunks to traverse (forwards and backwards) for the bin corruption check")
+heap_corruption_check_limit += 1 # Make sure that bins that have the default amount of chunks (64) are fully traversed
 
 resolve_heap_via_heuristic = add_heap_param(
     "resolve-heap-via-heuristic",
--- a/pwndbg/gdblib/heap/__init__.py
+++ b/pwndbg/gdblib/heap/__init__.py
@@ -52,6 +52,9 @@ heap_chain_limit = add_heap_param(
     "heap-dereference-limit", 8, "number of chunks to dereference in each bin"
 )
 
+heap_corruption_check_limit = add_heap_param("heap-corruption-check-limit", 64, "lenght limit for bins which are fully checked for corruption")
+heap_corruption_check_limit += 1 # Make sure bins that are of default length are checked
+
 resolve_heap_via_heuristic = add_heap_param(
     "resolve-heap-via-heuristic",
     "auto",
--- a/pwndbg/gdblib/heap/__init__.py
+++ b/pwndbg/gdblib/heap/__init__.py
@@ -52,8 +52,13 @@ heap_chain_limit = add_heap_param(
     "heap-dereference-limit", 8, "number of chunks to dereference in each bin"
 )
 
-heap_corruption_check_limit = add_heap_param("heap-corruption-check-limit", 64, "amount of chunks to traverse (forwards and backwards) for the bin corruption check")
-heap_corruption_check_limit += 1 # Make sure that bins that have the default amount of chunks (64) are fully traversed
+heap_corruption_check_limit = add_heap_param(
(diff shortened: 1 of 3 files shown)


CptGibbon:
Thanks for this PR, I'll review next week.

CptGibbon:
@k4lizen before I review this, do you have any tests ready to commit that could avoid this regressing?

k4lizen:
I dont

CptGibbon (changes requested):
Thanks for this PR 🙏
It looks good, appears to be much more accurate detecting bin corruption.

Just needs a regression test ensuring that bins aren't labelled as corrupt when `heap-dereference-limit` is lower than the number of chunks in a bin.

k4lizen:
Done

k4lizen:
Btw the debian CI/CD is a bit suspect: https://github.com/pwndbg/pwndbg/actions/runs/10012491583/job/27678119432?pr=2289
`Error: Unexpected error: Error: Path Validation Error: Path(s) specified in the action for caching do(es) not exist, hence no cache is being saved.`
Is that expected? It seems to be slowing down the job quite a bit.

CptGibbon (approved):

disconnect3d:
> Btw the debian CI/CD is a bit suspect: https://github.com/pwndbg/pwndbg/actions/runs/10012491583/job/27678119432?pr=2289 `Error: Unexpected error: Error: Path Validation Error: Path(s) specified in the action for caching do(es) not exist, hence no cache is being saved.` Is that expected? It seems to be slowing down the job quite a bit.

I believe not, pls open an issue for that

PR #2292, opened 2024-07-07
Arm Thumb Instruction Set support with Unicorn + Capstone

OBarronCS:
This PR adds support for Thumb-mode in Arm in the disasm view, allowing the Unicorn emulator to execute Thumb instructions, and allowing the Capstone disassembler to switch between Arm/Thumb mode while disassembling for the disasm view.

The 32-bit Arm execution state supports two architectures - Arm (A32) and Thumb (T32). Programs can freely switch between these two architectures through `interworking`, which simply means to write to the least-significant-bit of the program counter to change between Thumb and non-Thumb mode. The bit is not actually written to the PC in the hardware, and instead is directed to a bit in the CPSR flags register.

A32 mode solely consists of 4-byte instructions. Thumb (T32) mode has a completely different instruction set, which has a mix of 2 and 4 byte instructions. The Armv8-M profile (which the ARM Cortex-M series of CPU's use) executes solely in Thumb mode.

Previously, our use of the Unicorn emulator wasn't able to execute Thumb instructions, and it instead interpreted all instructions as A32 instructions. In order to use Thumb mode in Unicorn, turns out you have to set the least-significant-bit of the PC to 1 every time you start the emulator and want it to be in Thumb mode: https://github.com/unicorn-engine/unicorn/issues/391. There's a comment in the code further explaining this. This is now corrected and Thumb mode instructions are executed correctly.

To make the context display Thumb instructions correctly, we have to line up the Capstone disassembler with the emulation, so that the disassembler is aware that the emulator switched the Thumb mode and starts disassembling future instructions in the correct architecture.

The following are screenshots of a 4 instructions Arm program. Halfway through, the program set itself to Thumb mode (starting at instruction 3, we are in Thumb mode)

## Old:
We start in A32 mode, and the emulator and disassembler incorrectly interprets all future instructions as A32
old_arm_thumb_disasm_broken

The program is in Thumb mode, and the disassembler sees this, but the Unicorn emulator, unable to execute T32 instructions, still thinks we are in A32 mode, thus the instructions all being 4-bytes apart since we use the emulator PC to determine the `next instruction`. The disassembler and Unicorn are not synced.
thumb_mode_old_unicorn

## New:
We start in A32 mode, and Unicorn and Capstone are synced in the transition to T32 mode.
thumb_support_unicorn_capstone

new_thumb_mode_unicorn

### DIFF
--- a/pwndbg/emu/emulator.py
+++ b/pwndbg/emu/emulator.py
@@ -688,7 +688,7 @@ class Emulator:
         ident = self.hook_add(U.UC_HOOK_CODE, hook)
 
         pc: int = self.pc
-        if (thumb_bit := self.read_thumb_bit()):
+        if thumb_bit := self.read_thumb_bit():
             # Unicorn disregards the UC_MODE_THUMB mode passed into the constructor, and instead
             # determines Thumb mode based on the PC that is passed to the `emu_start` function
             # https://github.com/unicorn-engine/unicorn/issues/391
--- a/pwndbg/emu/emulator.py
+++ b/pwndbg/emu/emulator.py
@@ -528,7 +528,7 @@ class Emulator:
             pc = pwndbg.gdblib.regs.pc
         self.uc.reg_write(self.get_reg_enum(self.regs.pc), pc)
 
-    def read_thumb_bit(self) -> int | None:
+    def read_thumb_bit(self) -> int:
         """
         Return 0 or 1, representing the status of the Thumb bit in the current Arm architecture
 
@@ -545,7 +545,7 @@ class Emulator:
         elif self.arch == "armcm":
             if (xpsr := self.xpsr) is not None:
                 return (xpsr >> 24) & 1
-        return None
+        return 0
 
     def get_uc_mode(self):
         """
@@ -688,16 +688,16 @@ class Emulator:
         ident = self.hook_add(U.UC_HOOK_CODE, hook)
 
         pc: int = self.pc
-        if thumb_bit := self.read_thumb_bit():
-            # Unicorn appears to disregard the UC_MODE_THUMB mode passed into the constructor, and instead
-            # determines Thumb mode based on the PC that is passed to the `emu_start` function
-            # https://github.com/unicorn-engine/unicorn/issues/391
-            #
-            # Because we single-step the emulator, we always have to read the Thumb bit from the emulator
-            # and set the least significant bit of the PC to 1 if the bit is 1 in order to enable Thumb mode
-            # for the execution of the next instruction. If this `emulate_with_hook` executes multiple instructions
-            # which have Thumb mode transitions, Unicorn will internally handle them.
-            pc |= thumb_bit
+        # Unicorn appears to disregard the UC_MODE_THUMB mode passed into the constructor, and instead
+        # determines Thumb mode based on the PC that is passed to the `emu_start` function
+        # https://github.com/unicorn-engine/unicorn/issues/391
+        #
+        # Because we single-step the emulator, we always have to read the Thumb bit from the emulator
+        # and set the least significant bit of the PC to 1 if the bit is 1 in order to enable Thumb mode
+        # for the execution of the next instruction. If this `emulate_with_hook` executes multiple instructions
+        # which have Thumb mode transitions, Unicorn will internally handle them.
+        thumb_bit = self.read_thumb_bit()
+        pc |= thumb_bit
 
         try:
             self.emu_start(pc, 0, count=count)
--- a/pwndbg/emu/emulator.py
+++ b/pwndbg/emu/emulator.py
@@ -539,10 +539,10 @@ class Emulator:
 
         Mimics the `read_thumb_bit` function defined in gdblib/arch.py
         """
-        if pwndbg.gdblib.arch.current == "arm":
+        if self.arch == "arm":
             if (cpsr := self.cpsr) is not None:
                 return (cpsr >> 5) & 1
-        elif pwndbg.gdblib.arch.current == "armcm":
+        elif self.arch == "armcm":
             if (xpsr := self.xpsr) is not None:
                 return (xpsr >> 24) & 1
         return None
--- a/pwndbg/emu/emulator.py
+++ b/pwndbg/emu/emulator.py
@@ -528,6 +528,25 @@ class Emulator:
             pc = pwndbg.gdblib.regs.pc
         self.uc.reg_write(self.get_reg_enum(self.regs.pc), pc)
 
+    def read_thumb_bit(self) -> int | None:
+        """
+        Return 0 or 1, representing the status of the Thumb bit in the current Arm architecture
+
+        This reads from the emulator itself, meaning this can be read to determine a state
+        transitions between non-Thumb and Thumb mode
+
+        Return None if the Thumb bit is not relevent to the current architecture
+
+        Mimics the `read_thumb_bit` function defined in gdblib/arch.py
+        """
+        if pwndbg.gdblib.arch.current == "arm":
+            if (cpsr := self.cpsr) is not None:
+                return (cpsr >> 5) & 1
+        elif pwndbg.gdblib.arch.current == "armcm":
+            if (xpsr := self.xpsr) is not None:
+                return (xpsr >> 24) & 1
+        return None
+
     def get_uc_mode(self):
         """
         Retrieve the mode used by Unicorn for the current architecture.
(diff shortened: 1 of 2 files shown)


pwndbg/emu/emulator.py
```diff
@@ -528,6 +528,25 @@ def update_pc(self, pc=None) -> None:
...
+        Return None if the Thumb bit is not relevent to the current architecture
+
+        Mimics the `read_thumb_bit` function defined in gdblib/arch.py
+        """
+        if self.arch == "arm":
+            if (cpsr := self.cpsr) is not None:
+                return (cpsr >> 5) & 1
+        elif self.arch == "armcm":
+            if (xpsr := self.xpsr) is not None:
+                return (xpsr >> 24) & 1
```
disconnect3d, line 547: I am okay with this code but I think we should just use a correct version/register based on known architecture.

I mean, let's analyze this.

1. The `self.arch` is set in the `Emulator` object based on current arch figured out by Pwndbg: https://github.com/pwndbg/pwndbg/blob/b74fcdde97a4cd4c6c072235040eb21ee7a42a5b/pwndbg/emu/emulator.py#L184-L188
2. The `pwndbg.gdblib.arch.current` is set in the `Arch` class here: https://github.com/pwndbg/pwndbg/blob/b74fcdde97a4cd4c6c072235040eb21ee7a42a5b/pwndbg/lib/arch.py#L16-L21
3. This is updated by the `gdblib.arch.update` function: https://github.com/pwndbg/pwndbg/blob/b74fcdde97a4cd4c6c072235040eb21ee7a42a5b/pwndbg/gdblib/arch.py#L109-L111
4. Which is triggered as a hook here: https://github.com/pwndbg/pwndbg/blob/b74fcdde97a4cd4c6c072235040eb21ee7a42a5b/pwndbg/gdblib/hooks.py#L22-L26

I wonder:
- Do we really need to call it on each stop too? I bet we do it since the `stop` event may be called while none other event is called? Shouldn't we cache it if the arch did not change or sth?
- If we figure out the arch, maybe its time to do some "architecture classes" and have a `def get_thumb_bit` method defined differently for each arch? I mean, we do call such function anyway, so why not define this function differently for each arch when we know this arch already?

pwndbg/emu/emulator.py
```diff
@@ -667,8 +686,21 @@ def emu_stop(self, *a, **kw):
...
+        if thumb_bit := self.read_thumb_bit():
+            # Unicorn appears to disregard the UC_MODE_THUMB mode passed into the constructor, and instead
+            # determines Thumb mode based on the PC that is passed to the `emu_start` function
+            # https://github.com/unicorn-engine/unicorn/issues/391
+            #
+            # Because we single-step the emulator, we always have to read the Thumb bit from the emulator
+            # and set the least significant bit of the PC to 1 if the bit is 1 in order to enable Thumb mode
+            # for the execution of the next instruction. If this `emulate_with_hook` executes multiple instructions
+            # which have Thumb mode transitions, Unicorn will internally handle them.
+            pc |= thumb_bit
```
disconnect3d: Note: we could probably drop the `if` and always do the bit-or operation? I bet we can't because this may return a `None`? We could in theory always return `0` instead... heh.

pwndbg/gdblib/disasm/__init__.py
```diff
@@ -149,17 +153,19 @@ def get_disassembler_cached(arch, ptrsize: int, endian, extra=None):
     return cs
 
 
-def get_disassembler(pc):
+def get_disassembler(address):
     if pwndbg.gdblib.arch.current == "armcm":
+        thumb_mode = bool(emulated_arm_mode_cache[address])
+        if thumb_mode is None:
```
disconnect3d, line 159: It can't be `None` here since we did `bool(...)`. I believe this should be fixed @OBarronCS

pwndbg/gdblib/disasm/__init__.py
```diff
@@ -149,17 +153,19 @@ def get_disassembler_cached(arch, ptrsize: int, endian, extra=None):
...
-            (CS_MODE_MCLASS | CS_MODE_THUMB)
-            if (pwndbg.gdblib.regs.xpsr & (1 << 24))
-            else CS_MODE_MCLASS
-        )
+        extra = (CS_MODE_MCLASS | CS_MODE_THUMB) if thumb_mode else CS_MODE_MCLASS
 
     elif pwndbg.gdblib.arch.current in ("arm", "aarch64"):
-        extra = CS_MODE_THUMB if (pwndbg.gdblib.regs.cpsr & (1 << 5)) else CS_MODE_ARM
+        thumb_mode = bool(emulated_arm_mode_cache[address])
+        if thumb_mode is None:
```
disconnect3d, line 166: Seems like same issue as above - `thumb_mode` can't be `None` here?

pwndbg/gdblib/disasm/__init__.py
```diff
@@ -120,6 +120,10 @@ def clear_on_reg_mem_change() -> None:
     lambda: None
 )
 
+# Maps an address to integer 0/1, indicating the Thumb mode bit for the given address.
+# Value is None if Thumb bit irrelevent or unknown.
+emulated_arm_mode_cache: DefaultDict[int, int] = collections.defaultdict(lambda: None)
```
disconnect3d, line 125: This cache is never cleared. Is that ok?

Also can't you have a single address once executed as Thumb and once as ARM mode? Like a polyglot ARM & Thumb. I think u can?

disconnect3d (changes requested):
See specific comments

pwndbg/gdblib/disasm/__init__.py
OBarronCS: Oop -  I'll remove all the `bool`s here as we can rely on Python seeing `0` as falsy and any other integer as truthy in this context.

pwndbg/gdblib/disasm/__init__.py
OBarronCS: I think we can safely not clear this cache. In the same way the code has used `backward_cache`, the values will always be updated/overridden by more recent data when we step near the instructions again. We 'renew' the cache values for the next future instructions every time we step, so for the case of a polyglot binary, the cache will contain the values from the most recent time we stepped near the instructions in question. The cache values will always be correct in the current context (whether we are interpreting the instructions as Arm/Thumb at this moment in the polyglot binary).

They could be stale (if we step at the instructions and they are non-Thumb, then we don't step at the location for a while, and instead hopped between some breakpoints that are somewhere else, and the polyglot binary in between has switched to Thumb, and then we do `nearpc address_that_was_non_thumb_when_i_stepped_there`), but I think this is so incredibly rare that it's a better experience to just keep the cache always uncleared, so when someone does `nearpc` at some address they've been at, it will get the correct thumb/arm mode transitions.

pwndbg/gdblib/disasm/__init__.py
OBarronCS: Fixed

pwndbg/emu/emulator.py
OBarronCS: I completely agree. The codebase has a couple (5+) occurences of long `if/else` chains that check for the current architecture, and set a local variable/take some small actions based on the arch.

I've put some notes on the Pwndbg Discord regarding a new architecture class structure, where we can have functions such as `get_thumb_bit` which would reduce the amount of these `if/else` checks, and at the same time bundling in some other changes to cleanup the use of `ABI's`.

Once the architecture changes are in, we can come back and fix these occurences of `if arch == "arm"`, etc, etc.

A note on the need for updating the arch on each step:
- I can think of at least one case where it is necessary, which is the early stages of a bootloader in x86 when it changes from 16-bit mode to 32-bit instruction set.

pwndbg/emu/emulator.py
OBarronCS: Done!

OBarronCS:
The latest commit fixes the bugs and makes changes as mentioned above.

pwndbg/gdblib/disasm/__init__.py
disconnect3d: Shall we maybe clear it at least on program start or attach (probably newobjfile event)?

I mean: if we never clear it and it maps address->0|1, then having a super long GDB session where u restart an ASLRed binary 100 times will make this cache bigger and bigger, effectively leaking memory.

pwndbg/emu/emulator.py
```diff
@@ -667,8 +686,21 @@ def emu_stop(self, *a, **kw):
...
+        # Unicorn appears to disregard the UC_MODE_THUMB mode passed into the constructor, and instead
+        # determines Thumb mode based on the PC that is passed to the `emu_start` function
+        # https://github.com/unicorn-engine/unicorn/issues/391
+        #
+        # Because we single-step the emulator, we always have to read the Thumb bit from the emulator
+        # and set the least significant bit of the PC to 1 if the bit is 1 in order to enable Thumb mode
+        # for the execution of the next instruction. If this `emulate_with_hook` executes multiple instructions
+        # which have Thumb mode transitions, Unicorn will internally handle them.
+        thumb_bit = self.read_thumb_bit()
+        pc |= thumb_bit
```
disconnect3d, line 700: Why not just `pc |= self.read_thumb_bit()`?

disconnect3d (approved):

disconnect3d:
Please address any leftover comments: 1) pc |= thumb bit and 2) clearing cache, in another PR :).

PR #2297, opened 2024-07-09
Add register value retrieval to Debugger-agnostic API and extra register check to Pwndbg command handler

mbrla0:

This PR adds a new class `pwndbg.dbg_mod.Registers`, which wraps the values of the set of registers for a given frame, after a given call to the newly-added `pwndbg.dbg_mod.Frame.regs()` function. This new class contains a function that allows one to query for the value of a register, given its name.

That new class is used as part of the argument parsing logic for targets that don't support `pwndbg.gdblib.regs.fix()`, and is intended for use in later providing equivalent functionality to the `pwndbg.gdblib.regs` in a debugger-agnostic form.

To choice to have it as a separate class from `pwndbg.dbg_mod.Frame` stems from the fact that LLDB's `SBFrame.GetRegisters()` function returns all registers in the frame as a bundle. So, in order to save on calls to `GetRegisters` and avoid having to delve into internal caching - which requires events, and we don't have events in LLDB yet -, we follow the design of the LLDB API. 

This choice fits GDB pretty well, too, though it does involve an extra level of indirection if all you're doing is pulling a single register directly from the debugger-agnostic API, but that can be mitigated later by using caching in our future equivalent of `pwndbg.gdblib.regs`.

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -239,6 +239,13 @@ def fix(
         pass
 
     try:
+        # This will fail if gdblib is not available. While the next check
+        # alleviates the need for this call, it's not really equivalent, and
+        # we'll need a debugger-agnostic version of regs.fix() if we want to
+        # completely get rid of this call. We can't do that now because there's
+        # no debugger-agnostic architecture functions. Those will come later.
+        #
+        # TODO: Port architecutre functions and `pwndbg.gdblib.regs.fix` to debugger-agnostic API and remove this.
         arg = pwndbg.gdblib.regs.fix(arg)
         return target.evaluate_expression(arg)
     except Exception as e:
@@ -248,6 +255,7 @@ def fix(
     # see if that yields anything.
     if frame:
         regs = frame.regs()
+        arg = arg.strip()
         if arg.startswith("$"):
             arg = arg[1:]
         reg = regs.by_name(arg)
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -232,12 +232,12 @@ def fix(
 
     # Try to evaluate the expression in the local, or, failing that, global
     # context.
-    ex = None
     try:
         return target.evaluate_expression(arg)
     except Exception:
         pass
 
+    ex = None
     try:
         # This will fail if gdblib is not available. While the next check
         # alleviates the need for this call, it's not really equivalent, and
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -230,6 +230,9 @@ def fix(
     )
     assert target, "Reached command expression evaluation with no frame or inferior"
 
+    # Try to evaluate the expression in the local, or, failing that, global
+    # context.
+    ex = None
     try:
         return target.evaluate_expression(arg)
     except Exception:
@@ -239,6 +242,21 @@ def fix(
         arg = pwndbg.gdblib.regs.fix(arg)
         return target.evaluate_expression(arg)
     except Exception as e:
+        ex = e
+
+    # If that fails, try to treat the argument as the name of a register, and
+    # see if that yields anything.
+    if frame:
+        regs = frame.regs()
+        if arg.startswith("$"):
+            arg = arg[1:]
+        reg = regs.by_name(arg)
+        if reg:
+            return reg
+
+    # If both fail, check whether we want to print or re-raise the error we
+    # might've gotten from `evaluate_expression`.
+    if ex:
         if not quiet:
             print(e)
         if reraise:
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -258,9 +258,9 @@ def fix(
     # might've gotten from `evaluate_expression`.
     if ex:
         if not quiet:
-            print(e)
+            print(ex)
         if reraise:
-            raise e
+            raise ex
 
     if sloppy:
         return arg
--- a/pwndbg/dbg/__init__.py
+++ b/pwndbg/dbg/__init__.py
@@ -12,15 +12,18 @@ from typing import Tuple
 
 dbg: Debugger = None
 
+
 class Registers:
     """
     A handle to the register values in a frame.
     """
+
     def by_name(self, name: str) -> Value | None:
         """
         Gets the value of a register if it exists, None otherwise.
         """
         raise NotImplementedError()
+
 
 class Frame:
     def evaluate_expression(self, expression: str) -> Value:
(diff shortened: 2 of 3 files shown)


pwndbg/commands/__init__.py
```diff
@@ -230,19 +230,45 @@ def fix(
     )
     assert target, "Reached command expression evaluation with no frame or inferior"
 
+    # Try to evaluate the expression in the local, or, failing that, global
+    # context.
+    ex = None
```
gsingh93: nit: can you move this after this try/except, closer to where it's going to be used?

gsingh93 (commented):
Question: will we eventually be able to do `pwndbg.dbg.regs.rax`, for example, to access individual registers in the debugger agnostic API? I see why this new API is necessary for LLDB, but it's also less ergonomic, so I'm hoping it'll just be an implementation detail anyone writing commands won't need to worry about.

mbrla0:
> Question: will we eventually be able to do `pwndbg.dbg.regs.rax`, for example, to access individual registers in the debugger agnostic API?

Yes, that's exactly what I intend to add in the near future.

pwndbg/commands/__init__.py
mbrla0: Alright

disconnect3d:
Why don't implement it like this in the first place? I mean, that we preserve the "fetch .rax" register interface.

Also I am a bit worried that this work may conflict with future changes from @OBarronCS about using "per-architecture classes" which I guess may be related to how we fetch registers as well.

(Fwiw this may also kinda optimize how we do it? I am not sure)

mbrla0:
> Why don't implement it like this in the first place? I mean, that we preserve the "fetch .rax" register interface.

@disconnect3d If I understand you correctly, you mean implementing it like `pwndbg.dbg.rax` or something like that,  correct? The reason that's not what I went for is that I'm generally trying to keep the debugger-agnostic as thin as possible, and avoid having to do too much heavy lifting inside either the GDB or LLDB implementations.

A lot of the logic for those things in Pwndbg is implemented inside gdblib, so I generally feel like it makes more sense to, whenever possible, focus on just the primitives it needs, rather than push the responsibility to get all of those complexities right to the individual implementations of the agnostic API. And this is one of those times where can focus on the primitives.

> Also I am a bit worried that this work may conflict with future changes from @OBarronCS about using "per-architecture classes" which I guess may be related to how we fetch registers as well.

What do you worry the problem would be? As far as I can tell, their per-arch classes could probably just retrieve the actual register values using this API, no?

pwndbg/commands/__init__.py
```diff
@@ -230,19 +230,45 @@ def fix(
...
+    # If that fails, try to treat the argument as the name of a register, and
+    # see if that yields anything.
+    if frame:
+        regs = frame.regs()
+        arg = arg.strip()
+        if arg.startswith("$"):
+            arg = arg[1:]
+        reg = regs.by_name(arg)
+        if reg:
+            return reg
```
disconnect3d, line 263: I don't get it. Why do we need this?

All this logic here is about `<command> <arguments>` and the handling for each separate argument, right? If so, why do we do `arg.strip()` here at all?

Also, does this mean that llvm doesn't provide us some "pls parse the arguments and giv us debugger value objects" api?

pwndbg/dbg/__init__.py
```diff
@@ -13,6 +13,18 @@
 dbg: Debugger = None
 
 
+class Registers:
+    """
+    A handle to the register values in a frame.
+    """
+
+    def by_name(self, name: str) -> Value | None:
```
disconnect3d, line 25: Honestly, I don't like having classes with one methods which are unimpl. Is there much benefit of having a base class instead of having function per each debugger for all of this?

I mean, there is some mantra/saying that "class with just one method should be a function" and I kinda agree with it :P.

mbrla0: > Is there much benefit of having a base class instead of having function per each debugger for all of this?

Well, in this case, I believe so, yeah. LLDB pulls in all register values at once, through `SBFrame::GetRegisters()`, and personally I'd really rather us avoid having to either pull them all every single time we need a single register value, or having to add internal caching logic to the LLDB implementation to get around that.

We'll very likely want to cache the results of our own `Frame.regs()`, but I feel that should be done outside the Debugger-agnostic API implementations.

pwndbg/commands/__init__.py
mbrla0: > Why do we need this?

That check tries to temporarily alleviate the need for `regs.fix()` under LLDB. It allows us to do `<command> <regname>` where `regname` doesn't begin with a `$`, which is a subset of what `regs.fix()` allows us to do. At this point in the PR chain, `regs.fix()` is simply unavailable under LLDB. It's available in the most up-to-date branch of the port, but not here.

> If so, why do we do arg.strip() here at all?

The argument splitting logic is allowed to pass us empty spaces there, and stripping lets us match the behavior of `regs.fix()` for a single argument.

> Also, does this mean that llvm doesn't provide us some "pls parse the arguments and giv us debugger value objects" api?

It does, and the first `evaluate_expression` should work just fine with expressions in which all registers are prefixed with a `$`. The only thing that bit of code serves to do is reproduce a few of the cases `regs.fix()` would get an expression to work under GDB where it otherwise wouldn't, but under LLDB, until `regs.fix()` is available under LLDB, too.

disconnect3d: This needs to be documented in the code, otherwise its super unclear why we need it in the first place.

Also, if we have `regs.fix` in a later branch, I assume this logic is then not really needed and will be removed?

disconnect3d: We should probably have the latest change here in this PR or just merge it and go fast with newer PRs. Lets do the 2nd.

pwndbg/dbg/__init__.py
disconnect3d: Still why do we need a class at all for that?

Wouldn't it be better to just call GetRegisters on each stop and call it a day?

Or `regs.__getattr__` could have `if self._fetched_regs is None: fetch_regs(); return self.registers[regname]` or sth like that

pwndbg/commands/__init__.py
mbrla0: > This needs to be documented in the code, otherwise its super unclear why we need it in the first place.

Fair enough, I can make that clearer, then.

> Also, if we have `regs.fix` in a later branch, I assume this logic is then not really needed and will be removed?

Yes, it's a temporary solution until we get `regs.fix` working under LLDB.

disconnect3d:
@mbrla0 lets get this rolling. :)

PR #2298, opened 2024-07-09
Add an error type to the Debugger-agnostic API and move uses of the API to it

mbrla0:

This PR adds a new class `pwndbg.dbg_mod.Error`, for exceptions that happen within the debugger-agnostic, wraps instances of `gdb.error` encountered by the GDB implementation of the Debugger-agnostic API in the new error type, and moves consumers that were using both the new API and `gdb.error` to the new error type.

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -17,6 +17,7 @@ from typing import TypeVar
 from typing_extensions import ParamSpec
 
 import pwndbg.exception
+
 # These aren't available under LLDB, and we can't get rid of them until all of
 # this functionality has been ported to the Debugger API.
 #
@@ -62,8 +63,7 @@ class CommandCategory(str, Enum):
     DEV = "Developer"
 
 
-assert pwndbg.dbg.session(), "Tried to set up commands with no interactive session"
-GDB_BUILTIN_COMMANDS = pwndbg.dbg.session().commands()
+GDB_BUILTIN_COMMANDS = pwndbg.dbg.commands()
 
 # Set in `reload` command so that we can skip double checking for registration
 # of an already existing command when re-registering GDB CLI commands
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -14,19 +14,23 @@ from typing import Set
 from typing import Tuple
 from typing import TypeVar
 
-import gdb
 from typing_extensions import ParamSpec
 
 import pwndbg.exception
-import pwndbg.gdblib.heap
-import pwndbg.gdblib.kernel
-import pwndbg.gdblib.proc
-import pwndbg.gdblib.qemu
-import pwndbg.gdblib.regs
-from pwndbg.gdblib.heap.ptmalloc import DebugSymsHeap
-from pwndbg.gdblib.heap.ptmalloc import GlibcMemoryAllocator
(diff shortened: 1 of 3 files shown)


pwndbg/commands/__init__.py
```diff
@@ -153,7 +153,7 @@ def invoke(self, argument: str, from_tty: bool) -> None:
         except SystemExit:
             # Raised when the usage is printed by an ArgparsedCommand
             return
-        except (TypeError, gdb.error):
+        except (TypeError, pwndbg.dbg_mod.Error):
```
disconnect3d, line 156: Random comment: is there any reason we name it `dbg_mod` instead of just `dbg`?

pwndbg/dbg/__init__.py
```diff
@@ -13,6 +13,10 @@
 dbg: Debugger = None
 
 
+class Error(Exception):
```
disconnect3d, line 16: Is there any real benefit for doing that instead of using pure `Exception`?

pwndbg/dbg/gdb.py
```diff
@@ -30,22 +31,67 @@ def parse_and_eval(expression: str, global_context: bool) -> gdb.Value:
...
+        """
+
+        self.get_current = get_current
+        self.select = select
+        self.target = target
+        self.restore = False
+        self.current = None
+
+    @override
+    def __enter__(self) -> None:
```
disconnect3d: Wouldn't it be simpler to just use a function context manager for this? https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager

pwndbg/dbg/__init__.py
mbrla0: pwndbg often distinguishes between a debugger-API-type exception and other types of exceptions. We could forego exceptions for error handling entirely in the Debugger-agnostic API, but there's a lot of spots in pwndbg where it expects error conditions to be signaled through debugger-API-specific exceptions, so I feel it makes sense to keep it, and expose our own exception type for the places where `gdb.error` is currently being used.

pwndbg/commands/__init__.py
mbrla0: I wanted accessing the main debugger object to be concise, so I borrowed the `<name>/<name>_mod` idiom that Pwndbg uses for `pwndbg.gdblib.arch`, `pwndbg.gdblib.regs`, and `pwndbg.gdblib.config`, where `<name>` is the main object of that module, and `<name>_mod` is the module itself.

pwndbg/dbg/gdb.py
mbrla0: Yeah, probably. It just didn't occur to me at the time of writing. Do you want me to change to it?

pwndbg/dbg/gdb.py
```diff
@@ -192,7 +218,10 @@ def fetch_lazy(self) -> None:
 
     @override
     def __int__(self) -> int:
-        return int(self.inner)
+        try:
+            return int(self.inner)
+        except gdb.error as e:
+            raise pwndbg.dbg_mod.Error(e)
```
disconnect3d, line 224: When can this fail? Why was it that we did not handle this exception in the past? Same for others.

disconnect3d (approved):

PR #2303, opened 2024-07-12
CMP-like instruction annotations for AArch64, Arm

OBarronCS:
This PR adds annotation to `CMP`-like instructions (instructions that mutate flags register, often seen before branches or conditional instructions) to AArch64 and Arm.

For AArch64 there are five such instructions: `CMP, CMN, TST, CCMP`, and `CCMN`
Arm has 4 - `CMP, CMN, TST`, and `TEQ`

For the instructions, it displays the resulting value of the flags register, as well as the operation that took place.

Image for AArch64:
aarch64_cmp

Currently, it displays all the bits of the CPSR register. Would it make more sense to only display the flags that can be mutated - `N,Z,C,V`?

### DIFF
--- a/pwndbg/gdblib/disasm/aarch64.py
+++ b/pwndbg/gdblib/disasm/aarch64.py
@@ -12,8 +12,10 @@ import pwndbg.gdblib.arch
 import pwndbg.gdblib.disasm.arch
 import pwndbg.gdblib.memory
 import pwndbg.gdblib.regs
+import pwndbg.lib.disasm.helpers as bit_math
 from pwndbg.emu.emulator import Emulator
 from pwndbg.gdblib.disasm.instruction import ALL_JUMP_GROUPS
+from pwndbg.gdblib.disasm.instruction import EnhancedOperand
 from pwndbg.gdblib.disasm.instruction import InstructionCondition
 from pwndbg.gdblib.disasm.instruction import PwndbgInstruction
 from pwndbg.gdblib.disasm.instruction import boolean_to_instruction_condition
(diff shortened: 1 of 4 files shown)


disconnect3d:
> Currently, it displays all the bits of the CPSR register. Would it make more sense to only display the flags that can be mutated - `N,Z,C,V`?

I guess its probably fine to display them all?

disconnect3d:
@OBarronCS pls fix conflicts here

pwndbg/gdblib/disasm/arch.py
```diff
@@ -721,5 +722,54 @@ def _memory_string(self, instruction: PwndbgInstruction, operand: EnhancedOperan
...
+            # Some architectures have variants with more operands that modify the values before comparison.
+            if len(instruction.operands) == 2:
+                left, right = instruction.operands
+
+                if (
+                    left.before_value_resolved is not None
+                    and right.before_value_resolved is not None
+                ):
+                    print_left, print_right = pwndbg.enhance.format_small_int_pair(
+                        left.before_value_resolved, right.before_value_resolved
```
disconnect3d: We should probably assign those two vars in the `if` with `:=`?

disconnect3d (changes requested):
Fix conflicts

OBarronCS:
Working on getting this corrected - did some more testing, and there were some cases I initially didn't consider - like `cmp x5, x3, LSL #12`, and Capstone has a bit of a cumbersome API to expose the shifts and extend operations which makes generically parsing the operands not straight forward.

OBarronCS:
Latest commits fixes the cases of the operands having modifiers (shifts/extends) - there are now functions that correctly resolve registers and immediates on AArch64 and Arm.

image

pwndbg/gdblib/disasm/arch.py
OBarronCS: Done

pwndbg/gdblib/disasm/aarch64.py
```diff
@@ -132,5 +168,62 @@ def _set_annotation_string(self, instruction: PwndbgInstruction, emu: Emulator)
         # Dispatch to the correct handler
         self.annotation_handlers.get(instruction.id, lambda *a: None)(instruction, emu)
 
+    def _register_width(self, instruction: PwndbgInstruction, op: EnhancedOperand) -> int:
+        # TODO: find a better way to do this. Cache this on the operand object?
+        return 32 if instruction.cs_insn.reg_name(op.reg).upper().startswith("W") else 64
```
disconnect3d: Can the reg name be 'w' or 'W'? Aren't all lowercase or uppercase? Maybe we can skip the `.upper()` part?

Also: `s[0] == 'W'` is probably faster than `s.startswith("W")`.

pwndbg/gdblib/disasm/arch.py
```diff
@@ -755,5 +756,50 @@ def _common_generic_register_destination(
...
+        Many architectures implement near-identical `CMP`-like instructions.
+
+        It takes two values, either subtracts, adds, or does some bit operation
+        with them to set values in the flag register.
+
+        To reduce code duplication, subclasses can use this function to create an annotator for CMP-like instructions.
+        """
+        FLAG_REG_NAME_DISPLAY = flags_register_name.upper()
+
+        def handler(instruction: PwndbgInstruction, emu: Emulator):
```
disconnect3d, line 772: Note: this function is re-created each time the upper function is called. Maybe we should have it as some kind of `__privatebutnotrealy_func` :x.

disconnect3d, line 772: Note: this function is re-created each time the upper function is called. Maybe we should have it as some kind of `__privatebutnotrealy_func` :x.

pwndbg/gdblib/disasm/aarch64.py
OBarronCS: Made this change

pwndbg/gdblib/disasm/arch.py
OBarronCS: I think no matter how we do it, there's going to be an extra function created for each unique cmp-like instruction. Currently, this outer function will return the inner function with a couple variables bound in the closure (flags_register_name, char_to_separate_operands). One way or another, the final annotator function has to get these values which are unique to each different instruction type.

disconnect3d (approved):

PR #2306, opened 2024-07-14
Small fix to remote qemu branch target detection

OBarronCS:
This PR makes a small fix to our default detection of branch targets in our enhancement code, making it work with remote QEMU targets more accurately.

When debugging a remote QEMU target, the memory permissions are not always accurate - the "execute" bit, for example, is not set for the pages that are executed. In the default logic to determine branch targets, it used to check for this execute bit when detecting a target. This breaks the automatic resolution of targets with QEMU.

image

It will now bypass this check on remote targets.

### DIFF
--- a/pwndbg/gdblib/disasm/arch.py
+++ b/pwndbg/gdblib/disasm/arch.py
@@ -16,6 +16,7 @@ import pwndbg.gdblib.symbol
 import pwndbg.gdblib.typeinfo
 import pwndbg.gdblib.vmmap
 import pwndbg.lib.config
+import pwndbg.gdblib.remote
 from pwndbg.emu.emulator import Emulator
 from pwndbg.gdblib.disasm.instruction import FORWARD_JUMP_GROUP
 from pwndbg.gdblib.disasm.instruction import EnhancedOperand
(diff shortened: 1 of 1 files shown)


pwndbg/gdblib/disasm/arch.py
```diff
@@ -668,7 +669,9 @@ def _resolve_target(self, instruction: PwndbgInstruction, emu: Emulator | None,
                         addr = resolved_addr
                     else:
                         page = pwndbg.gdblib.vmmap.find(resolved_addr)
-                        if page and page.execute:
+                        # When debugging a remote QEMU target, the page permissions are not accurate.
+                        # In this case, if the candidate address is mapped at all, just go with it.
+                        if page and (pwndbg.gdblib.remote.is_remote() or page.execute):
```
disconnect3d: Which check is 'heavier'? Shall we check for `page.execute` first and only then call `is_remote()`? :)
```suggestion
                        if page and (page.execute or pwndbg.gdblib.remote.is_remote()):
```

OBarronCS: True, the check for is_remote may make a call to a GDB command (or at least does a dictionary lookup on @cache decorator- I flipped the order

PR #2309, opened 2024-07-16
Annotate all load operations Arm, AArch64, RISCV, and MIPS

OBarronCS:
This PR adds annotations to all load instructions in `Arm`, `AArch64`, `RISCV`, and `MIPS`. These instructions load some amount of bytes from memory and place them into a register.

For these instructions, the memory operand and the resulting value put into the target register is displayed. Additionally, the annotation will indicate if the instruction segfaults.

After the memory operands have been resolved, the subclasses that enhance each architecture will single out the instructions that are load instructions, and call a parent class method to do the work of the annotation. The annotation can use emulation to determine the values for future instructions, and even when emulation is disabled, it can annotate the current instruction.

For each of the architectures, I went through ISA manuals and picked out all the load operations, and found out all the different ways that memory operands are represented in the different architectures so that we can resolve the concrete memory address used using the information provided by Capstone.

RISCV:
riscv5_load

AArch64:
aarch64_cannot_dereference

Arm:
arm_memory
When emulation off, the function can still annotate the current instruction.
ARM_memory_emulate_off

The API makes it simple to add these annotations to future architectures that Pwndbg will support - simple fill in a function to resolve the memory operand, single out the load instructions, and call a parent function.

This PR has some dependencies on other open pull-requests, so it is kept as a draft initially.


OBarronCS:
Latest commits adds ability to read signed values from memory and sign extend them to the target register width in the annotations. Signed-load operations read some amount of bytes from memory (say 4), and then sign extend the result to the size of the target register. This works with and without emulation.

In this example in 64-bit RISCV, we loading a word (32-bits) and the instruction sign extends to 64-bits (the size of the target register).
signed_loads

disconnect3d:
@OBarronCS pls fix conflicts that arised :<

pwndbg/gdblib/disasm/arm.py
```diff
@@ -116,10 +158,82 @@ def _memory_string(self, instruction: PwndbgInstruction, op: EnhancedOperand) ->
...
+        # All memory operands have `base` defined
+        base = self._read_register(instruction, op.mem.base, emu)
+        if base is None:
+            return None
+
+        if op.mem.base == ARM_REG_PC:
+            # The PC as the base register is a special case - it will align the address to a word (32-bit) boundary
+            # Explanation: https://stackoverflow.com/a/29588678
+            # See "Operation" at the bottom of https://developer.arm.com/documentation/ddi0597/2024-03/Base-Instructions/LDR--literal---Load-Register--literal--
+            base = align_down(4, base)
```
OBarronCS, line 216: This is very sneaky Arm behavior that was fun to track down 😀

pwndbg/gdblib/disasm/aarch64.py
```diff
@@ -108,6 +160,27 @@ def __init__(self, architecture: str) -> None:
             ARM64_INS_CCMN: self._common_cmp_annotator_builder("cpsr", ""),
         }
 
+    @override
+    def _set_annotation_string(self, instruction: PwndbgInstruction, emu: Emulator) -> None:
+        # Dispatch to the correct handler
+        if instruction.id in AARCH64_SINGLE_LOAD_INSTRUCTIONS:
+            # TODO: replace the 8 with actual read size from other PR
```
disconnect3d: So this will be done in another PR?

pwndbg/gdblib/disasm/aarch64.py
```diff
@@ -164,9 +237,61 @@ def _resolve_target(self, instruction: PwndbgInstruction, emu: Emulator | None,
...
+        target = 0
+
+        # All memory operands have `base` defined
+        base = self._read_register(instruction, op.mem.base, emu)
+        if base is None:
+            return None
+        target += base
+
+        # Add displacement (zero by default)
+        target += op.mem.disp
```
disconnect3d: Just `target = base + op.mem.disp`?

pwndbg/gdblib/disasm/arm.py
```diff
@@ -116,10 +158,82 @@ def _memory_string(self, instruction: PwndbgInstruction, op: EnhancedOperand) ->
...
+
+        target += base
+
+        # Add displacement (zero by default)
+        target += op.mem.disp
```
disconnect3d: Just
```suggestion
        target = base + op.mem.disp
```

(+ remove previous target assignments)

pwndbg/gdblib/disasm/mips.py
```diff
@@ -152,5 +171,25 @@ def _resolve_target(self, instruction: PwndbgInstruction, emu: Emulator | None,
...
+        """
+        target = op.mem.disp
+
+        if op.mem.base != 0:
+            base = self._read_register(instruction, op.mem.base, emu)
+            if base is None:
+                return None
+            target += base
+
+        return target
```
disconnect3d: `return base` above and `return op.mem.disp` here?

disconnect3d: Tho the current code is also fine

disconnect3d (commented):
Minor comments

pwndbg/gdblib/disasm/aarch64.py
OBarronCS: Good catch - this was a small change I forgot to fill in after a rebase. Fixed it.

pwndbg/gdblib/disasm/mips.py
OBarronCS: On a second look - I think base is always valid (non-zero), so I changed this as the riscv _parse_memory to simplify them.

Issue #2162, opened 2024-05-10
RecursionError: maximum recursion depth exceeded while calling a Python object


PlatyPew:
### Description

When using [qiling](https://github.com/qilingframework/qiling) with the gdb flag, attaching to the process with `target remote :9999` causes a maximum recursion depth exceeded error. Works fine with vanilla gdb.

```
Set whether to print a full stacktrace for exceptions raised in Pwndbg commands to 'on'.
Traceback (most recent call last):
...
...
  File "/usr/lib/python3.11/site-packages/pwndbg/gdblib/vmmap.py", line 204, in explore
    flags |= 1 if not pwndbg.gdblib.stack.is_executable() else 0
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/pwndbg/lib/cache.py", line 141, in decorator
    value = func(*a, **kw)
            ^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/pwndbg/gdblib/stack.py", line 71, in is_executable
    ehdr = pwndbg.gdblib.elf.exe()
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/pwndbg/gdblib/proc.py", line 149, in wrapper
    return func(*a, **kw)
           ^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/pwndbg/lib/cache.py", line 141, in decorator
    value = func(*a, **kw)
            ^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/pwndbg/gdblib/elf.py", line 224, in exe
    return load(e)
           ^^^^^^^
  File "/usr/lib/python3.11/site-packages/pwndbg/gdblib/elf.py", line 262, in load
    return get_ehdr(pointer)[1]
           ^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/pwndbg/gdblib/elf.py", line 292, in get_ehdr
    vmmap = pwndbg.gdblib.vmmap.find(pointer)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/pwndbg/lib/cache.py", line 141, in decorator
    value = func(*a, **kw)
            ^^^^^^^^^^^^^^
[...]

Issue #2132, opened 2024-04-26
[bug]: Python Exception: Recursion Error


k4lizen:
### Description
pwndbg hitting recursion limit while gdb works fine.

 CTF challenge zip: [the_wilderness.zip](https://github.com/pwndbg/pwndbg/files/15133514/the_wilderness.zip)
Im using an older version of the sde than the challenge intended: https://www.intel.com/content/www/us/en/download/684897/784319/intel-software-development-emulator.html

### Steps to reproduce

`./sde-external-<version>-lin/sde64 -debug -no-follow-child -cet -cet_output_file /dev/null -- ./build/the_wilderness`
The program will give the command to run in gdb in a new window, it will look like: `target remote :40395`. Works as intended in gdb, while pwndbg goes crazy:
![image](https://github.com/pwndbg/pwndbg/assets/124312252/c8eac522-631b-41c1-921f-6ffe5dd56e74)

### My setup

Gdb:      13.2
Python:   3.11.8 (main, Feb  7 2024, 21:52:08) [GCC 13.2.0]
Pwndbg:   2024.02.14 build: da817b23
Capstone: 5.0.1280
Unicorn:  2.0.1

### related
My be related to https://github.com/pwndbg/pwndbg/issues/2110 though the behaviour seems a bit different, and `set follow-fork-mode parent` does not seem to have any effect.

disconnect3d:
Fixed in https://github.com/pwndbg/pwndbg/pull/2311

Issue #2110, opened 2024-04-04
RecursionError in is_executable()


lucasballr:
### Description

I am auditing an old version of nginx. I installed the .deb package of pwndbg v2024.02.14 in an ubuntu 16.04 container. Running the program works just fine, but when it forks for a connection, I get a recursion error in pwndbg.

### Steps to reproduce

**In a Dockerfile:**
FROM ubuntu:16.04 
...
RUN wget http://nginx.org/download/nginx-1.3.9.tar.gz \
    && tar -xzvf nginx-1.3.9.tar.gz \
    && rm nginx-1.3.9.tar.gz
RUN wget https://github.com/pwndbg/pwndbg/releases/download/2024.02.14/pwndbg_2024.02.14_amd64.deb
RUN apt install ./pwndbg_2024.02.14_amd64.deb

`pwndbg --args /usr/local/nginx/sbin/nginx -g 'daemon off;'`

run and connect with a basic connection.
If you break during the parsing of that http connection, the error will occur.

### My setup

Gdb:      13.2
Python:   3.11.5 (main, Aug 24 2023, 12:23:19) [GCC 12.3.0]
Pwndbg:   2024.02.14
Capstone: 5.0.1280
Unicorn:  2.0.1

disconnect3d:
@lucasballr can you show the error?

disconnect3d:
Also please provide full steps to reproduce the issue. The dockerfile doesn't build nginx etc. :/

disconnect3d:
I was able to build and run this nginx using the following additional commands:
```
apt install gcc wget tar make curl
./configure --without-http_rewrite_module --without-http_gzip_module
make -j2
make install
```

Idk where to break, I just hit ctrl+c and could reproduce the issue:
```
pwndbg> set exception-verbose on
Set whether to print a full stacktrace for exceptions raised in Pwndbg commands to 'on'.
pwndbg> set exception-debugger on                                               
Set whether to debug exceptions raised in Pwndbg commands to 'on'.
pwndbg> r                                                                       
Starting program: /usr/local/nginx/sbin/nginx -g daemon\ off\;
warning: Error disabling address space randomization: Operation not permitted
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[Attaching after Thread 0x79a7304c2700 (LWP 4908) fork to child process 4911]
[New inferior 2 (process 4911)]
[...]

PR #2311, opened 2024-07-19
Memory exploration bug fixes and minor improvements

### LINKED ISSUE #2264: Fix the stack.is_executable issues
In cases when Pwndbg can't fetch memory mappings it crashes. This is due to the fact that `stack.is_executable` is called that under the hood requires the vmmap information. The crash can be observed as a recursive call exception and was reported in the past in:
* https://github.com/pwndbg/pwndbg/issues/2110
* https://github.com/pwndbg/pwndbg/issues/2162
* https://github.com/pwndbg/pwndbg/issues/2132

The `stack.is_executable` was only needed for vmmap exploration feature which is implemented as `vmmap.explore` function: https://github.com/pwndbg/pwndbg/blob/0ab077ffdebf87dae017f58e188daaf7c23fa81b/pwndbg/gdblib/vmmap.py#L181-L215

In the past we had an automatic exploration of registers feature, which is currently disabled here (since the event trigger is commented out):
https://github.com/pwndbg/pwndbg/blob/0ab077ffdebf87dae017f58e188daaf7c23fa81b/pwndbg/gdblib/vmmap.py#L218-L222

However, we still have the `stack.is_executable` registered to trigger at a stop event in order to figure out if the NX bit is set in the binary that we are debugging or not. This is done by getting the entrypoint address from [ELF auxiliary vector information](https://articles.manugarg.com/aboutelfauxiliaryvectors) and then trying to find the corresponding memory map via the `vmmap.find` function, in order to find the memory map start address, to parse it as an ELF header and the ELF segments to get the NX status.

Now, the problem is that the `vmmap.find` function will fall back into the `explore` function, which will call into the `stack.is_executable` function to fetch the executable status:
https://github.com/pwndbg/pwndbg/blob/0ab077ffdebf87dae017f58e188daaf7c23fa81b/pwndbg/gdblib/vmmap.py#L167-L178

This is actually also wrong since this assumes that the executable bit is set for any memory mapping if the stack is executable. This was true for Linux kernel < 5.8 due to their "read implies exec" logic in regards to the NX bit set in the process that was being run. However, this was changed in >=5.8 so now enabling executable stack (not enabling NX) makes only the stack memory page being executable.

Here is more or less the stacktrace for the whole flow:
- The `stack.is_executable` is setup to trigger on a stop event
- it [calls `elf.exe`](https://github.com/pwndbg/pwndbg/blob/0ab077ffdebf87dae017f58e188daaf7c23fa81b/pwndbg/gdblib/stack.py#L70-L74) to get the ELF header (EHDR) information
- the [`elf.exe` gets entrypoint address through the `entry()` call and passes it to `load(e)`](https://github.com/pwndbg/pwndbg/blob/0ab077ffdebf87dae017f58e188daaf7c23fa81b/pwndbg/gdblib/elf.py#L244-L251)
- the [`load(e)` calls `get_ehdr`](https://github.com/pwndbg/pwndbg/blob/0ab077ffdebf87dae017f58e188daaf7c23fa81b/pwndbg/gdblib/elf.py#L289-L290)
- the [`get_ehdr` calls `vmmap.find`](https://github.com/pwndbg/pwndbg/blob/0ab077ffdebf87dae017f58e188daaf7c23fa81b/pwndbg/gdblib/elf.py#L302-L320)
- ...which calls [`explore`](https://github.com/pwndbg/pwndbg/blob/0ab077ffdebf87dae017f58e188daaf7c23fa81b/pwndbg/gdblib/vmmap.py#L168-L178)
- the `explore()` calls [`stack.is_executable()`](https://github.com/pwndbg/pwndbg/blob/0ab077ffdebf87dae017f58e188daaf7c23fa81b/pwndbg/gdblib/vmmap.py#L207C23-L207C58)
- which recurses itself forever.

I believe that we have to:
1. Stop calling the `stack.is_executable` function on a stop event
2. Ensure some UX when memory pages are not present: if we don't do it already, we should warn the user "hey we did not find memory mappings, you can add them via `vmmap_add`"
3. Either remove the memory mapping exploration feature completely, or, reimplement it.

I would vote for reimplementing the memory mapping exploration and then:
1. Fixing the documentation (since some places probably say we do it already)
2. Adding a `vmmap_explore` command
3. Fixing `explore` so it only calls `stack.is_executable` when it tries to determine the stack memory page executable status
4. Probably moving `stack.is_executable` to the same module where `explore`
5. Refactoring/reimplementing the `get_ehdr` so it does not call `vmmap.find` and instead maybe iterates over couple of memory pages (up to some sane limit) and trying to detect the ELF header?

In the future this should probably account for or/and allow specifying memory ranges which are not accessible (and may crash debugger) but that's something we need generally for our enhancer.

Aplet123:
Closes #2264 and fixes the related issues #2110, #2162, and #2132

Disables exploration when trying to find the ELF header, which fixes the infinite recursion error.

Adds an `auto-explore-pages` config option with values `no`/`warn`/`yes` to disable automatic exploration, disable exploration but print a warning on a page that's missing mappings, or enable automatic exploration.

Adds a `vmmap_explore` command to explicitly explore a page.

Improves the heuristics for exploration guessing if a page is executable beyond just checking the NX bit.

### DIFF
--- a/pwndbg/commands/vmmap.py
+++ b/pwndbg/commands/vmmap.py
@@ -376,7 +376,7 @@ parser.add_argument(
 
 @pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.MEMORY)
 @pwndbg.commands.OnlyWhenRunning
-def vmmap_explore(address: int | str) -> None:
+def vmmap_explore(address: int) -> None:
     if not isinstance(address, int):
         print("Address is not a valid integer.")
         return
--- a/pwndbg/commands/vmmap.py
+++ b/pwndbg/commands/vmmap.py
@@ -366,3 +366,23 @@ def vmmap_load(filename) -> None:
     for page in pages:
         pwndbg.gdblib.vmmap.add_custom_page(page)
         print("%r added" % page)
+
+
+parser = argparse.ArgumentParser(description="Explore a page, trying to guess permissions.")
+parser.add_argument(
+    "address", type=pwndbg.commands.sloppy_gdb_parse, help="Address of the page to explore"
+)
+
+
+@pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.MEMORY)
+@pwndbg.commands.OnlyWhenRunning
+def vmmap_explore(address: int | str) -> None:
+    if not isinstance(address, int):
+        print("Address is not a valid integer.")
+        return
+    page = pwndbg.gdblib.vmmap.explore(address)
+    if page is None:
+        print("Exploration failed. Maybe the address isn't readable?")
+        return
+    print_vmmap_table_header()
+    print(page)
--- a/pwndbg/gdblib/vmmap.py
+++ b/pwndbg/gdblib/vmmap.py
@@ -248,7 +248,8 @@ def explore(address_maybe: int) -> pwndbg.lib.memory.Page | None:
     if not flags:
         return None
 
-    flags |= 2 if pwndbg.gdblib.memory.poke(address_maybe) else 0
+    if pwndbg.gdblib.memory.poke(address_maybe):
+        flags |= 2
     # It's really hard to check for executability, so we just make some guesses:
     # If it's in the same page as the stack pointer, try to check the NX bit
     # If it's in the same page as the instruction pointer, assume it's executable
@@ -260,7 +261,7 @@ def explore(address_maybe: int) -> pwndbg.lib.memory.Page | None:
         address_maybe == pwndbg.lib.memory.page_align(pwndbg.gdblib.regs.sp)
         and pwndbg.gdblib.stack.is_executable()
     ):
-        flags |= 1 if pwndbg.gdblib.stack.is_executable() else 0
+        flags |= 1
 
     page = find_boundaries(address_maybe)
     page.objfile = "<explored>"
--- a/pwndbg/gdblib/vmmap.py
+++ b/pwndbg/gdblib/vmmap.py
@@ -66,6 +66,14 @@ Note that the page-tables method will require the QEMU kernel process to be on t
     enum_sequence=["page-tables", "monitor", "none"],
 )
 
+auto_explore = pwndbg.config.add_param(
+    "auto-explore-pages",
+    "yes",
+    "whether to try to infer page permissions when memory maps missing (can cause errors)",
+    param_class=pwndbg.lib.config.PARAM_ENUM,
+    enum_sequence=["yes", "warn", "no"],
+)
+
 
 @pwndbg.lib.cache.cache_until("objfile", "start")
 def is_corefile() -> bool:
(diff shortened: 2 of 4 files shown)


pwndbg/gdblib/vmmap.py
```diff
@@ -204,7 +249,18 @@ def explore(address_maybe: int) -> pwndbg.lib.memory.Page | None:
...
+    # If it's in the same page as the instruction pointer, assume it's executable
+    # Otherwise, just say it's not executable
+    if address_maybe == pwndbg.lib.memory.page_align(pwndbg.gdblib.regs.pc):
+        flags |= 1
+    # TODO: could maybe make this check look at the stacks in pwndbg.gdblib.stack.get() but that might have issues
+    elif (
+        address_maybe == pwndbg.lib.memory.page_align(pwndbg.gdblib.regs.sp)
+        and pwndbg.gdblib.stack.is_executable()
+    ):
+        flags |= 1 if pwndbg.gdblib.stack.is_executable() else 0
```
disconnect3d: This could be just `flags |= 1` since we already checked for `stack.is_executable()` in the `elif` condition.

Tho maybe this should be just `if addr == aligned(pc) or (addr == aligned(sp) and stack.is_executable()): flags |= 1`?

pwndbg/gdblib/vmmap.py
```diff
@@ -204,7 +249,18 @@ def explore(address_maybe: int) -> pwndbg.lib.memory.Page | None:
         return None
 
     flags |= 2 if pwndbg.gdblib.memory.poke(address_maybe) else 0
```
disconnect3d: Fwiw this is kinda stupid, should be:

```suggestion
    if pwndbg.gdblib.memory.poke(address_maybe):
        flags |= 2
```

pwndbg/commands/vmmap.py
```diff
@@ -366,3 +366,23 @@ def vmmap_load(filename) -> None:
...
+
+parser = argparse.ArgumentParser(description="Explore a page, trying to guess permissions.")
+parser.add_argument(
+    "address", type=pwndbg.commands.sloppy_gdb_parse, help="Address of the page to explore"
+)
+
+
+@pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.MEMORY)
+@pwndbg.commands.OnlyWhenRunning
+def vmmap_explore(address: int | str) -> None:
```
disconnect3d: We have a check for int only later on

```suggestion
def vmmap_explore(address: int) -> None:
```

pwndbg/gdblib/vmmap.py
```diff
@@ -87,10 +95,11 @@ def is_corefile() -> bool:
...
 
 @pwndbg.lib.cache.cache_until("start", "stop")
-def get() -> Tuple[pwndbg.lib.memory.Page, ...]:
+def get_known_maps() -> Tuple[pwndbg.lib.memory.Page, ...] | None:
     """
-    Returns a tuple of `Page` objects representing the memory mappings of the
-    target, sorted by virtual address ascending.
+    Similar to `vmmap.get()`, except only returns maps in cases where
+    the mappings are known, like if it's a coredump, or if process
+    mappings are available.
```
disconnect3d, line 102: The mappings are also known if we do fetch them via gdb-pt-dump, isn't it?

PR #2317, opened 2024-07-23
Add support for fs/gs register in Unicorn and annotations

OBarronCS:
This PR adds support for `fs`/`gs` segment registers in Unicorn and in x86 annotations. In `x86-64`, Unicorn is now able to emulate instructions that reference fs/gs segments, and annotations display the results of these instructions. There is also support for annotations in 32-bit x86.

The fix is small: we have a property of our `regs` object which will fetch the value of `fsbase` (or `gsbase`). Unicorn, however, called the constant `FS_BASE`, so we need to just add an explicit mapping so our Pwndbg `FSBASE` is mapped to the Unicorn X86 `FS_BASE` constant (same with `gsbase`). 

In `x86-64`, the `fs` register, when seen in assembly, effectively uses the value of the `fsbase` register for any address computation. The Unicorn engine follows this, so by setting the `fsbase` register accordingly, all memory accesses work correctly.

In 32-bit mode, however, these `fsbase/gsbase` registers do not exist, and instead the segment address is stored in memory in the `GDT`. In order to make fs/gs use work Unicorn, we would need to setup the GDT correctly in Unicorn, which would be significantly more difficult in most cases (we could perhaps create a fake GDT - probably more effort than is worth and [creates other complications](https://github.com/unicorn-engine/unicorn/issues/877)). 32-bit access to `fs` and `gs` still creates full annotations for the current instruction without emulation since the actual process is able to access `fs` and `gs` - it's just Unicorn that cannot.

The most visible change this provides is making it easy to see stack cookies in annotations.

fs_move

sub_fs_base

32-bit x86 instruction - works without emulation for the current instruction
32_bit_x86_current-instruction

Thanks to @lebr0nli for finding that Unicorn will now function with `fsbase/gsbase` and for [initial PoC code](https://github.com/pwndbg/pwndbg/pull/2001#discussion_r1509273898).

Related issue: #113

### DIFF
--- a/pwndbg/gdblib/disasm/x86.py
+++ b/pwndbg/gdblib/disasm/x86.py
@@ -330,12 +330,16 @@ class DisassemblyAssistant(pwndbg.gdblib.disasm.arch.DisassemblyAssistant):
         # Get memory address (Ex: lea    rax, [rip + 0xd55], this would return $rip+0xd55. Does not dereference)
         target = 0
 
-        # There doesn't appear to be a good way to read from segmented
-        # addresses within GDB.
         if op.mem.segment != 0:
-            return None
-
-        if op.mem.base != 0:
+            if op.mem.segment == X86_REG_FS:
+                target = pwndbg.gdblib.regs.fsbase
+            elif op.mem.segment == X86_REG_GS:
+                target = pwndbg.gdblib.regs.gsbase
+            else:
+                return None
+        # Both a segment and base cannot be in use
+        # A Capstone bug sometimes sets base to RIP if segment is set
+        elif op.mem.base != 0:
             base = self._read_register(instruction, op.mem.base, emu)
             if base is None:
                 return None
--- a/pwndbg/gdblib/disasm/x86.py
+++ b/pwndbg/gdblib/disasm/x86.py
@@ -332,9 +332,11 @@ class DisassemblyAssistant(pwndbg.gdblib.disasm.arch.DisassemblyAssistant):
 
         if op.mem.segment != 0:
             if op.mem.segment == X86_REG_FS:
-                target = pwndbg.gdblib.regs.fsbase
+                if (target := pwndbg.gdblib.regs.fsbase) is None:
+                    return None
             elif op.mem.segment == X86_REG_GS:
-                target = pwndbg.gdblib.regs.gsbase
+                if (target := pwndbg.gdblib.regs.gsbase) is None:
+                    return None
             else:
                 return None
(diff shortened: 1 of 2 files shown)


disconnect3d:
Awesome!

pwndbg/gdblib/disasm/x86.py
```diff
@@ -330,12 +330,18 @@ def _parse_memory(self, instruction: PwndbgInstruction, op: EnhancedOperand, emu
...
+            if op.mem.segment == X86_REG_FS:
+                if (target := pwndbg.gdblib.regs.fsbase) is None:
+                    return None
+            elif op.mem.segment == X86_REG_GS:
+                if (target := pwndbg.gdblib.regs.gsbase) is None:
+                    return None
+            else:
+                return None
+        # Both a segment and base cannot be in use
+        # A Capstone bug sometimes sets base to RIP if segment is set
```
disconnect3d, line 343: What's with this bug? :p

Is it documented anywhere or sth? can we link?

What do we do in this case?

pwndbg/gdblib/disasm/x86.py
```diff
@@ -330,12 +330,18 @@ def _parse_memory(self, instruction: PwndbgInstruction, op: EnhancedOperand, emu
         # Get memory address (Ex: lea    rax, [rip + 0xd55], this would return $rip+0xd55. Does not dereference)
         target = 0
 
-        # There doesn't appear to be a good way to read from segmented
-        # addresses within GDB.
         if op.mem.segment != 0:
```
disconnect3d, line 333: Given that we are doing a lot of accesses into `op.mem` shall we maybe assign `op.mem` to `m` or sth?

pwndbg/gdblib/disasm/x86.py
OBarronCS: I'll look to make a Capstone report of it.

This shows the capstone disassembly (using `cstool`) of `mov r9, qword ptr fs:[0x10]`. For some reason, the rip register is shown as base. I just happened to randomly catch this while stepping through a binary - I didn't see any issues on the Capstone GitHub, but I can make one.

image

OBarronCS: Oh the plot thickens. Totally glossed over the `rip` in the memory operand... This comment in the code can be removed....

I used `Keystone` to assemble `mov r9, qword ptr fs:[0x10]`, and looks like it added the `rip` into it. Very strange. Maybe a keystone bug.

PR #2321, opened 2024-07-26
Hexdump: fixed computation of `hexdump.offset`

cfreal:
This fixes two bugs related to the offset that gets updated when one calls `hexdump` repeatedly, and adds typing to the definition of the `hexdump()` function.

## Bug 1

When the first hexdump skips a few lines, the repeat call to hexdump will display a wrong offset (+30 in the example).

Screenshot from 2024-07-26 11-41-49

Now, it is fine:

Screenshot from 2024-07-26 11-44-35

## Bug 2

Before, when dumping an unaligned number of bytes, offset is off:

Screenshot from 2024-07-26 11-40-35

Now, it is fine:

Screenshot from 2024-07-26 11-39-08

I included a make-do test (repetition does not seem to be handled properly by the test suite).

Thanks for the amazing tool.
Charles

### DIFF
--- a/pwndbg/commands/hexdump.py
+++ b/pwndbg/commands/hexdump.py
@@ -86,7 +86,7 @@ def hexdump(address, count=pwndbg.config.hexdump_bytes) -> None:
 
     # TODO: What if arch endian is big, and use_big_endian is false?
     flip_group_endianness = (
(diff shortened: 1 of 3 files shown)


disconnect3d:
Awesome, thanks!

Issue #2325, opened 2024-07-28
Recolor pwndbg prompt when debuggee process is alive


Costinteo:
**Is your feature request related to a problem? Please describe.**
I really enjoyed gef's feature of coloring the prompt green when the debuggee is alive.

**Describe the solution you'd like**
I want to have the "pwndbg>" prompt color up when the process that's being debugged is alive. It's just a few lines changed around the prompt and colors.

**Describe alternatives you've considered**
The alternative (or even an addition) could be a different prompt that doesn't rely on colors to convey the state of the debuggee. Like with a symbol.

**Additional context**
I'm currently implementing this in a PR. Opening this to refer to it through the PR.

PR #2326, opened 2024-07-28
Add prompt recolor based on proc alive state

Costinteo:
Simple PR implementing this feature idea https://github.com/pwndbg/pwndbg/issues/2325

The code is small enough to be self explanatory. I add a new color in the config for the "alive_prompt" and then on state change (transitions from alive to dead and vice-versa) I set the prompt again. All in the prompt hook.

### DIFF
--- a/pwndbg/color/message.py
+++ b/pwndbg/color/message.py
@@ -42,6 +42,9 @@ config_signal_color = theme.add_color_param(
 config_prompt_color: pwndbg.lib.config.Parameter = theme.add_color_param(
     "prompt-color", "bold,red", "prompt color"
 )
+config_prompt_alive_color: pwndbg.lib.config.Parameter = theme.add_color_param(
+    "prompt-alive-color", "bold,green", "prompt alive color"
+)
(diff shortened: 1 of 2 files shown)


disconnect3d:
@Costinteo can u include screenshots before and after?

Can we maybe a testcase for this behavior too?

Costinteo:
@disconnect3d I'd love to add a test case. However I'm having difficulties in doing so. 

I was initially thinking of checking the prompt from inside a newly-added gdb test and comparing it with an expected prompt (with colors added through escape sequences). However, the testing system currently implemented runs all tests with colors disabled, so I can't check escape sequences. I was also thinking of adding a list of tests that the test runner checks against before disabling colors... but it seems a bit awkward for just one test. What do you think?

About screenshots, here is how it looked before:



Here is how it looks after my PR:

disconnect3d:
Regarding testing with colors: can the test just do `set disable-colors on` and then `set disable-colors off` at the end? Would that solve it?

Issue #2327, opened 2024-07-28
Remote memory exploration firing mem_changed event


OBarronCS:
While remote debugging, information on memory memory mappings is generally limited. To determine memory permissions, Pwndbg will attempt to write a single byte to memory to see if the page is writable. This action inadvertently sets off the `pwndbg.gdblib.events.mem_changed` event, firing all listeners. 

During the `poke` call which attempts to write a byte to memory, we should disable all registered mem_changed events, so that they are not called.

https://github.com/pwndbg/pwndbg/blob/985654040479ce46cc8765afe44e1afbcf38532e/pwndbg/gdblib/memory.py#L164-L168

disconnect3d:
This has been fixed in #2328. Thanks!

PR #2328, opened 2024-07-29
Add ability to pause events

OBarronCS:
This PR adds the ability to temporarily pause GDB event handlers (#2327), and pauses events while Pwndbg is exploring memory for page permissions. 

This makes stepping through remote processes with Qemu much smoother, as an event handler listening for `mem_changed` clears the disasm instruction cache, and without pausing the event firing, the cache would seemingly randomly get cleared.

### DIFF
--- a/pwndbg/gdblib/events.py
+++ b/pwndbg/gdblib/events.py
@@ -7,6 +7,7 @@ by using a decorator.
 from __future__ import annotations
 
 import sys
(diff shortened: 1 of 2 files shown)


disconnect3d:
Ideally this should unregister the event for the time being, but this is good enough for now. (iff this is possible)

PR #2329, opened 2024-07-30
Go runtime value dumping

Aplet123:
Allows for dumping of Go types at runtime, including slices, maps, and arrays, and also can parse runtime reflection types to better dump values, which are useful for dumping boxed interface values or for getting the type of a heap allocation.

Adds a `go-dump` command:
image

image

image

image

Adds a `go-type` command:
image

image

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -709,6 +709,7 @@ def load_commands() -> None:
         import pwndbg.commands.elf
         import pwndbg.commands.flags
         import pwndbg.commands.ghidra
+        import pwndbg.commands.godbg
         import pwndbg.commands.got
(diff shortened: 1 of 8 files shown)


Aplet123:
I just found out this doesn't work with recursive types (e.g. a struct that contains a pointer to the same struct), so I'm working on a fix right now.

Aplet123:
Seems like something is up with parsing struct field offsets in old compiler versions. Looking into it.

Found it, old Go versions store a bitflag in the LSB of the struct field offset: https://github.com/golang/go/blob/release-branch.go1.18/src/reflect/type.go#L435C2-L435C61

Aplet123:
Another issue with how type names are stored in old versions, I'll look into this tomorrow.

PR #2330, opened 2024-07-30
Disintagle `pwndbg/commands/__init__.py` from GDB

mbrla0:
This PR is part of the GSoC 2024 LLDB Port.

This PR adds two commands to the Debugger-agnostic API. One that lists the currently available commands in the debugger, and another that lets us test for the availability of `gdb` and `gdblib`. It uses these two changes to control imports and uses of these two modules - and modules that depend on them - in `pwndbg/commands/__init__.py`, allowing it to be imported outside of GDB.

Importantly, the changes in this PR let us use `pwndbg.commands.ArgparsedCommand` outside of GDB.

The implementation of these interfaces under LLDB will come in the next PR, as soon as this one is accepted, I felt it made sense to split these two changes.


pwndbg/commands/__init__.py
```diff
@@ -14,19 +14,24 @@
...
-import pwndbg.gdblib.regs
-from pwndbg.gdblib.heap.ptmalloc import DebugSymsHeap
-from pwndbg.gdblib.heap.ptmalloc import GlibcMemoryAllocator
-from pwndbg.gdblib.heap.ptmalloc import HeuristicHeap
-from pwndbg.gdblib.heap.ptmalloc import SymbolUnresolvableError
+
+# These aren't available under LLDB, and we can't get rid of them until all of
+# this functionality has been ported to the Debugger API.
+#
+# TODO: Replace these with uses of the Debugger API.
```
disconnect3d, line 24: What does it mean that it has to be replaced with Debugger API?

Is there going to be a method for applying debugger-specific imports or registering commands?

pwndbg/commands/__init__.py
```diff
@@ -58,36 +63,16 @@ class CommandCategory(str, Enum):
...
-
-
-GDB_BUILTIN_COMMANDS = list_current_commands()
+GDB_BUILTIN_COMMANDS = pwndbg.dbg.commands()
 
 # Set in `reload` command so that we can skip double checking for registration
 # of an already existing command when re-registering GDB CLI commands
 # (there is no way to unregister a command in GDB 12.x)
-pwndbg_is_reloading = getattr(gdb, "pwndbg_is_reloading", False)
+pwndbg_is_reloading = False
```
disconnect3d, line 71: Shall we update the comment above to reflect lldb state somehow too?

pwndbg/commands/__init__.py
```diff
@@ -58,36 +63,16 @@ class CommandCategory(str, Enum):
...
+if pwndbg.dbg.is_gdblib_available():
```
disconnect3d, line 72: Random thought: technically we don't have to re-compute `is_gdblib_available` ever, so it could be a `IS_GDBLIB_AVAILABLE` or `IS_GDB` or sth, but idk, method is also fine, this is not any hot code or anything.

pwndbg/commands/__init__.py
```diff
@@ -689,79 +683,82 @@ def HexOrAddressExpr(s: str) -> int:
...
-    import pwndbg.commands.telescope
-    import pwndbg.commands.tips
-    import pwndbg.commands.tls
-    import pwndbg.commands.valist
-    import pwndbg.commands.version
-    import pwndbg.commands.vmmap
-    import pwndbg.commands.windbg
-    import pwndbg.commands.xinfo
-    import pwndbg.commands.xor
+    import pwndbg.dbg
```
disconnect3d, line 686: This loads all debugger-agnostic commands, right? If so, add a comment about it
```suggestion
    # Load all debugger-agnostic commands
    import pwndbg.dbg
```

pwndbg/dbg/__init__.py
```diff
@@ -172,6 +172,15 @@ def address(self) -> Value | None:
...
 
+    # is_optimized_out is kind of a janky piece of API, honestly. It makes it
+    # so that one's ability to call all other methods in this class is often
+    # conditional on it being false, and it effectively splits the type into
+    # two.
+    #
+    # There's only _one_ part of Pwndbg that uses it, and I really feel like we
+    # should handle variables that have been optimized out some other way.
+    #
+    # TODO: Remove uses of is_optimized_out from plist and get rid of this.
```
disconnect3d, line 183: Do we know that other way we are going to handle it?

pwndbg/dbg/__init__.py
```diff
@@ -316,6 +331,14 @@ def add_command(
     # removed or replaced as the porting work continues.
     #
 
+    # We'd like to be able to gate some imports off during porting. This aids in
+    # that.
+    def is_gdblib_available(self) -> bool:
```
disconnect3d, line 336: Can it become available? If not, maybe `is_gdblib` or just `is_gdb`, idk.

Consider changing it in another short PR if u decide to change it

pwndbg/dbg/gdb.py
```diff
@@ -452,10 +453,37 @@ def selected_frame(self) -> pwndbg.dbg_mod.Frame | None:
             pass
         return None
 
+    def commands(self):
+        current_pagination = gdb.execute("show pagination", to_string=True)
```
disconnect3d, line 457: Fwiw
```
(.venv) root@pwndbg:~/pwndbg# rg 'set pagination'
pwndbg/dbg/gdb.py
232:        set pagination off
```

But sure, user can change it

pwndbg/commands/__init__.py
mbrla0: > What does it mean that it has to be replaced with Debugger API?

`OnlyWithResolvedHeapSyms` need these bits to work. So, if we want access to it under LLDB, we need to port them to the debugger-agnostic APIs. The plan is to replace these imports with the generic imports once they're all available.

> Is there going to be a method for applying debugger-specific imports or registering commands?

I'm not sure what you mean by debugger-specific imports besides what we're already doing with the `pwndbg.dbg.is_gdblib_available()` checks. If it makes sense for us to distinguish between more than just GDB v. everything else, that should be easy enough to add, but there are no LLDB specific imports as of the latest changes in the port branch.

Debugger-specific command registration can be done by just importing debugger-specific commands in the debugger's `Debugger.setup()` method. `pset` under LLDB does just that. I don't expect there to be enough of those that it warrants a separate mechanism.

pwndbg/commands/__init__.py
mbrla0: Reloading is something that can only be done for GDB, and that I feel will only be available under GDB for the foreseeable future. I guess we could add something like `Reloading isn't available outside of GDB, so just assume we're not reloading unless we're running under GDB`?

pwndbg/commands/__init__.py
mbrla0: Actually, I think it makes sense to have it be a constant, as it does get used every now and then in the hot path going forward.

pwndbg/dbg/__init__.py
mbrla0: I personally feel like returning some other type is a little more elegant here. Something like

```python3
class OptimizedOutValue:
    pass
    
OPTIMIZED_OUT = OptimizedOutValue()

# ...

if value == OPTIMIZED_OUT:
    # etc.
```

But that might just be my being used to Rust making me feel like this is clearer in lieu of using a Rust-style enum, which Python doesn't have.

pwndbg/dbg/__init__.py
mbrla0: > Can it become available?

Never

It didn't occur to me that that name would imply that it could when I chose it

pwndbg/dbg/gdb.py
mbrla0: I just copied the implementation from `pwndbg/commands/__init__.py`, felt like I didn't have much of a reason to change it

Issue #2301, opened 2024-07-12
attachp: attach to process based on partial process name


disconnect3d:
It would be nice if the `attachp` command worked with not full process names.

It should first go over and see if there is a matching process name and only then look for processes based on partial match.

```
pwndbg> attachp qemu-system
Process qemu-system not found

pwndbg> !ps auxf | grep qemu
root        3608  0.0  0.0  12504  3584 pts/3    S+   22:22   0:00  |       |               \_ bash -c ps auxf | grep qemu
root        3610  0.0  0.0  11744  2688 pts/3    S+   22:22   0:00  |       |                   \_ grep qemu
1001        3588 40.4  1.6 3575592 258200 pts/0  Sl+  22:22   0:08      \_ /home/user/qemu-system-x86_64 -L /home/user/bios -kernel /hom

pwndbg> attachp qemu-system-x86_64
Attaching to 3588
[New LWP 3589]
```

disconnect3d:
Implemented in https://github.com/pwndbg/pwndbg/pull/2331

disconnect3d:
@wwwXOX hmm it turns out that the current implementation is a bit broken: doing `attachp whatever qemu` will make us attach to a `qemu-system-x86_64`. Not sure if that was the intention but I think this is counter-intuitive for users -- as it was for me ;).

wwwXOX:
Hi, @disconnect3d ,thank you for pointing out this issue. I understand that the current behavior might be counter-intuitive. Just to clarify, the intention behind this change was to support searching for multiple commands at once, as discussed in this [comment](https://github.com/pwndbg/pwndbg/pull/2331#issuecomment-2263286101).  

For example:

> Now supports searching for multiple commands at once, e.g., ['bash', 'example-sys'].
```
#### Multi-command search example:
pwndbg> attachp "bash example-sys"
Multiple processes found. Current resolution method is "ask". Run the command `config attachp-resolution-method` to see more informations.
      pid  user    elapsed    command
--  -----  ------  ---------  -----------------------
 1      1  root    17:22      /bin/bash
 2     23  root    16:58      bash
 3     46  root    16:51      bash
 4    389  root    16:08      ./example-system-x86_64
which process to attach?(1-4)
```

In the the command `attachp whatever qemu`, could you please share what you would expect it to match? For example, would the following cases align with your expectations?

- whatever-sys-qemu-x64 (O)
- whatever-qemu-x64 (O)
- ever-qemu-x64 (X)
[...]

PR #2331, opened 2024-07-30
feat: enhance attachp command to support partial process name matching.

wwwXOX:
The PR enhances the `attachp` command to support partial process name matching. 

With this enhancement, users can specify partial process names now, and the command will attempt to find matching processes based on those partial names.

## Changes
- Added functionality to check for partial process name matches using the output of the `ps aux` command.
- Updated the logic to filter processes based on partial matches from the provided target list.

## Issue
#2301

## Other
This is my first PR to an open source project. Please let me know if there are any other changes needed or if you have any questions.

### DIFF
--- a/pwndbg/commands/attachp.py
+++ b/pwndbg/commands/attachp.py
@@ -83,6 +83,24 @@ def attachp(no_truncate, target) -> None:
                 pids = []
 
             if not pids:
+                try:
+                    command = "ps aux | awk '{print $2, $11}'"
+                    ps_output = check_output(command, shell=True, universal_newlines=True)
+                except FileNotFoundError:
+                    print(message.error("Error: did not find `ps` command"))
+                    return
+                except CalledProcessError:
+                    pids = []
+
+                target_list = target.split()
+                for line in ps_output.strip().split("\n")[1:]:
+                    process_info = line.split()
+                    pid = process_info[0]
+                    command = " ".join(process_info[1:])
+                    if any(part in command for part in target_list) and len(pid) > 1:
+                        pids.append(pid)
+
+            if not pids:
                 print(message.error(f"Process {target} not found"))
                 return
 
--- a/pwndbg/commands/attachp.py
+++ b/pwndbg/commands/attachp.py
@@ -84,20 +84,21 @@ def attachp(no_truncate, target) -> None:
 
             if not pids:
                 try:
-                    command = "ps aux | awk '{print $2, $11}'"
-                    ps_output = check_output(command, shell=True, universal_newlines=True)
+                    ps_output = check_output(["ps", "-eo", "pid,args"], universal_newlines=True)
                 except FileNotFoundError:
                     print(message.error("Error: did not find `ps` command"))
                     return
                 except CalledProcessError:
                     pids = []
 
-                target_list = target.split()
+                target_list = [part for part in target.split() if len(part) >= 2]
                 for line in ps_output.strip().split("\n")[1:]:
                     process_info = line.split()
+                    if len(process_info) <= 1:
+                        continue
                     pid = process_info[0]
-                    command = " ".join(process_info[1:])
-                    if any(part in command for part in target_list) and len(pid) > 1:
+                    command = process_info[1]
+                    if any(part in command for part in target_list):
                         pids.append(pid)
 
             if not pids:


pwndbg/commands/attachp.py
```diff
@@ -82,6 +82,24 @@ def attachp(no_truncate, target) -> None:
             except CalledProcessError:
                 pids = []
 
+            if not pids:
+                try:
+                    command = "ps aux | awk '{print $2, $11}'"
```
disconnect3d: Ideally, we don't want to rely on `ask` program existence on the user system. Can we just invoke `["ps", "aux"]` and then filter out the result in Python code?

disconnect3d: (Also I am not sure if this works correctly if process name has spaces)

wwwXOX: Hi @disconnect3d ,
How about using `ps -eo pid,args` instead?

```bash
root ➜ /pwndbg $ ps aux
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root         1  0.0  0.0   7632  4296 pts/0    Ss   01:52   0:00 /bin/bash
root     10450  0.0  0.0   7632  4408 pts/1    Ss   06:29   0:00 bash
root     10655  0.0  0.0   2772   976 pts/1    S+   06:30   0:00 ./example-system-x86_64
root     10875  0.0  0.0  10068  1564 pts/0    R+   06:44   0:00 ps aux

root ➜ /pwndbg $ ps -eo pid,args
  PID COMMAND
    1 /bin/bash
10450 bash
10655 ./example-system-x86_64
10885 ps -eo pid,args
```

For example, with PID 10885, I would only capture `ps`, not `ps -eo pid,args`. 
Do you think this approach is acceptable?

disconnect3d:
Yeah, ps is fine.

pwndbg/commands/attachp.py
disconnect3d: Using`ps` is fine. I think capturing and matching the args is also fine in general: users may like this too.

wwwXOX:
> Yeah, ps is fine.

Hi @disconnect3d 

Thank you for the feedback, I have made the following changes:

1. Restricted the command length to be greater than 2 characters. This avoids capturing single character commands which can cause readability issues. Commands such as `ps`, `ls`, `nc`, etc., are now properly handled.

```
### `ps aux` list example
pwndbg> !ps aux
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root         1  0.0  0.0   7632  4152 pts/0    ts+  09:24   0:00 /bin/bash
root        23  0.0  0.0   7632  4216 pts/1    Ss   09:24   0:00 bash
root        46  0.0  0.0   7632  4400 pts/2    Ss   09:24   0:00 bash
root       389  0.0  0.0   2772   956 pts/2    S+   09:25   0:00 ./example-system-x86_64
root      1951  0.0  0.0   2888   968 pts/1    S+   09:39   0:00 sh -c ps aux
root      1952  0.0  0.0  10068  1556 pts/1    R+   09:39   0:00 ps aux


### Execution examples:

#### Before the fix:
pwndbg> attachp b
Multiple processes found. Current resolution method is "ask". Run the command `config attachp-resolution-method` to see more informations.
      pid  user    elapsed    command
--  -----  ------  ---------  ---------
 1      1  root    03:13      /bin/bash
 2     23  root    02:49      bash
 3     46  root    02:42      bash


#### After the fix:
pwndbg> attachp b
pids: []
Process b not found
```

2. Now supports searching for multiple commands at once, e.g., `['bash', 'example-sys']`.
```
#### Multi-command search example:
pwndbg> attachp "bash example-sys"
Multiple processes found. Current resolution method is "ask". Run the command `config attachp-resolution-method` to see more informations.
      pid  user    elapsed    command
--  -----  ------  ---------  -----------------------
 1      1  root    17:22      /bin/bash
 2     23  root    16:58      bash
 3     46  root    16:51      bash
 4    389  root    16:08      ./example-system-x86_64
which process to attach?(1-4)
```

disconnect3d:
Lol the CI failed here weirdly:
```
Poetry not found. Installing Poetry...
curl: (28) Failed to connect to install.python-poetry.org port 443 after 218922 ms: Connection timed out
Creating virtualenv in path: ./.venv
./setup.sh: line 201: poetry: command not found
```
...restarting :).

disconnect3d:
@wwwXOX Thanks. One last thing, we should probably change the description of the commant to account for partial name matches too.

```
    description="""Attaches to a given pid, process name or device file.
This command wraps the original GDB `attach` command to add the ability
to debug a process with given name. In such case the process identifier is
fetched via the `pidof <name>` command.
```

wwwXOX:
Hi @disconnect3d 
Thanks for your feedback. I have updated the description to account for partial name matches. I have created a new PR for this change. Could you please review it?

Here is the link to the new PR:  #2336

PR #2333, opened 2024-07-31
Add test for prompt recolor on process state change

Costinteo:
@disconnect3d I'm back with the test for prompt color changes. The ``set disable-colors off`` command worked great, thanks for that!

I have a few issues with the way I wrote the test, however. I want your opinions on them:
1. I copy the ``set_prompt`` function and rewrite it to return the expected prompt as a string in ``prepare_prompt``. This is a problem, because if the ``set_prompt`` function changes, this might go out of sync and the test will have to change (extra work). Solving this problem will mean changing the underlying ``set_prompt`` function and then reusing it here. At least that's what I am thinking of.
2. After starting the binary, I have to manually call the ``prompt_hook`` that runs the ``set_prompt`` on process state change. I am not exactly sure why the prompt hook doesn't run in normal tests after each command? Is there a better way to make this happen?

### DIFF
--- a/tests/gdb-tests/tests/test_prompt_recolor.py
+++ b/tests/gdb-tests/tests/test_prompt_recolor.py
@@ -1,0 +1,57 @@
+from __future__ import annotations
+
+import gdb
+
+import pwndbg
+import pwndbg.color.message
+import pwndbg.gdblib.proc
+import tests
+
+BINARY = tests.binaries.get("reference-binary.out")
+
+
+def prepare_prompt(is_proc_alive):
+    prompt = "pwndbg> "
+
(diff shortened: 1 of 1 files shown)


disconnect3d:
Hey,

1) in theory we could just assert that the prompts are different or if they
contain specific alive vs non alive bytes, but that’s probably fine.

It’s also ok that changing prompt will require fixing the test. That’s what
it’s for.

2) honestly I have no idea. I guess we can leave it as it is. It’s better
to have such a test then none :)

Btw instead of lots of str.replace calls to change single bytes u can use
str.translate, unless we change more than single bytes then it can’t be
used. (See
https://pythonexamples.org/python-string-translate/)

I will merge this when I get to my pc (probably tomorrow morning).

PR #2334, opened 2024-07-31
LLDB: Implement new interfaces and finish bringup of `ArgparsedCommand`

mbrla0:

This PR implements the interfaces added to the Debugger-agnostic API on LLDB, and allows `ArgparsedCommand` to be used under LLDB, with a simple test command (`test2`) that demos this functionality, and that gets removed down the line.

### DIFF
--- a/pwndbg/dbg/lldb.py
+++ b/pwndbg/dbg/lldb.py
@@ -1,6 +1,10 @@
 from __future__ import annotations
 
+import random
+import sys
 from typing import Any
+from typing import Callable
+from typing import List
 from typing import Tuple
 
 import lldb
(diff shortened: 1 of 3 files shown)


pwndbg/dbg/lldb.py
```diff
@@ -9,22 +13,363 @@
...
+    # invalid.
+    #
+    # Obviously, this is a rather bad solution, and breaks if the version of
+    # LLDB we're running under is not in English, or if this message gets
+    # changed in the future.
+    #
+    # LLDB does internally have a way to distinguish the invalid expression
+    # case from the optimized-out one, through lldb::ExpressionResults, but
+    # there does not seem to be a way to wrangle one out of
+    # EvaluateExpression.
```
disconnect3d, line 73: What does this mean that it has a way to distinguish it internally? Is it in its C/C++ code and that is not exposed into Python API? Can we link that code here and inform on that better here?

pwndbg/dbg/lldb.py
```diff
@@ -9,22 +13,363 @@
...
+            raise pwndbg.dbg_mod.Error("could not dereference value")
+
+        return LLDBValue(deref)
+
+    @override
+    def string(self) -> str:
+        addr = self.inner.unsigned
+        error = lldb.SBError()
+
+        # Read strings up to 4GB.
```
disconnect3d, line 177: Add a docstring documenting this.

pwndbg/dbg/lldb.py
```diff
@@ -9,22 +13,363 @@
...
+        last_str = None
+        buf = 256
+        for i in range(8, 33):
```
disconnect3d: Kinda unknown why `8, 33` - probably to account for reading exactly 4GB or sth? idk, but needs either a comment, or just a better range, or idk

pwndbg/dbg/lldb.py
```diff
@@ -9,22 +13,363 @@
...
+            last_str = s
+
+            buf *= 2
+
+        return last_str
+
+    @override
+    def fetch_lazy(self) -> None:
+        # Not needed under LLDB.
+        pass
```
disconnect3d, line 195: Shall this raise an error/exception? Why not needed? Document if really not needed pls.

pwndbg/dbg/lldb.py
```diff
@@ -9,22 +13,363 @@
...
+        return self.inner.signed
+
+    @override
+    def cast(self, type: pwndbg.dbg_mod.Type | Any) -> pwndbg.dbg_mod.Value:
+        assert isinstance(type, LLDBType)
+        t: LLDBType = type
+
+        return LLDBValue(self.inner.cast(t.inner))
+
+        return LLDBValue(self.inner.Cast(type.inner))
```
disconnect3d: Something is wrong here, the second `return` is dead code.

pwndbg/dbg/lldb.py
```diff
@@ -9,22 +13,363 @@
...
+
+        return LLDBValue(value)
+
+
+def map_type_code(type: lldb.SBType) -> pwndbg.dbg_mod.TypeCode:
+    """
+    Determines the type code of a given LLDB SBType.
+    """
+    c = type.GetTypeCode()
+    f = type.GetTypeFlags()
```
disconnect3d: move ` f = ...` after the ifs with TypeCode.

pwndbg/dbg/lldb.py
```diff
@@ -9,22 +13,363 @@
...
+
+        # LLDB is very particular with the object paths it will accept. It is at
+        # its happiest when its pulling objects straight off the module that was
+        # first imported with `command script import`, so, we install the class
+        # we've just created as a global value in its dictionary.
+        rand = round(random.random() * 0xFFFFFFFF) // 1
+        rand = f"{rand:08x}"
+
+        name = f"__{rand}_LLDB_COMMAND_{command_name}"
+        print(f"adding command {command_name}, under the path {self.module}.{name}")
```
disconnect3d, line 285: I don't really get the name randomization here. Aren't all module+command combinations we create unique? Why do we need the random part at all? What is a counter-example for this?

disconnect3d: (This is not clear from the comment too)

pwndbg/dbg/lldb.py
```diff
@@ -9,22 +13,363 @@
...
+
+    @override
+    def add_command(
+        self, command_name: str, handler: Callable[[pwndbg.dbg_mod.Debugger, str, bool], None]
+    ) -> pwndbg.dbg_mod.CommandHandle:
+        debugger = self
+
+        class CommandHandler:
+            def __init__(self, debugger, _):
+                pass
```
disconnect3d, line 271: I wonder if this should maybe inherit from some base class or sth.

disconnect3d: Just to be sure, we define the class here and not in global scope bcoz we need to "register it" in sys.modules?

pwndbg/dbg/lldb.py
mbrla0: It's part of the private API, and LLDB can distinguish between these cases using it. I've poked around for it and it really seems the public API doesn't have access to it at all.

pwndbg/dbg/lldb.py
mbrla0: It covers the range 2^8 (the `256` in `buf`) to 2^32 (the 4GB in the comment), inclusive, since we double `buf` after every iteration to check for larger strings. It doesn't have to be `range(8, 33)`, but I think that's a little clearer in intent than `range(25)`. Maybe a `# log2(256) = 8, log2(4GB) = 32` or `# range(log2(256), log2(4GB) + 1)` comment could clear that up?

pwndbg/dbg/lldb.py
mbrla0: LLDB doesn't have lazy values like GDB does. So it's only there because there are parts of Pwndbg that explicitly call `fetch_lazy()` on values. The explanation for that is written over the prototype of that method in `pwndbg/dbg/__init__.py`, should I copy it here?

pwndbg/dbg/lldb.py
mbrla0: Oh, yeah, right, that's probably a mistake of mine from when I did the rebase. Thanks for catching that!

pwndbg/dbg/lldb.py
mbrla0: Yep, LLDB takes classes for commands, so we have to create a new class for every new command that we want to register. I think it's important to point out that we'd have to create a new class regardless of where we were to put it. 

The fact that we have to add it to the dictionary of globals of `lldbinit` is due to another limitation with LLDB, which needs us to have our classes under globally-accessible names when we add new commands.

pwndbg/dbg/lldb.py
mbrla0: > Aren't all module+command combinations we create unique?

I'm not really sure, so, seeing as this is not hot code, I went with the safer option of making sure that, no matter what we did, these names would be very unlikely to collide. If you're sure that they're supposed to be unique, I can just take the randomization out.

pwndbg/dbg/lldb.py
mbrla0: Alright

pwndbg/dbg/lldb.py
mbrla0: Documenting what, exactly? Is the docstring for the base method not good because of the 4GB limit?

pwndbg/dbg/lldb.py
mbrla0: > Can we link that code here and inform on that better here?

I guess we could link the three instances of `EvaluateExpression`? I'm not sure what could be clearer.

pwndbg/dbg/lldb.py
disconnect3d: I was thinking we should maybe document the 4GB limit, but idk, its probably fine if its in base method or if we dont do it, likely nobody will notice anyway ;)

pwndbg/dbg/lldb.py
disconnect3d: Pls add a comment mentioning that here when we do it

pwndbg/dbg/lldb.py
disconnect3d: Don't we have a system in place to prevent duplicates? e.g. if some1 registers command A twice, it should fail  anyway.

I think the random can be removed

PR #2337, opened 2024-08-06
Emulator Arm SP register fix

OBarronCS:
This PR makes a fix to emulation in Arm mode.

Previously, the stack pointer register - `SP` - would always mysteriously contain the value of the zero, despite us explicitly setting the value in the emulator. After some debugging, I found that the stack pointer value in the emulator resets to 0 when we copy the process flag `CPSR` register to the emulator.  I made an issue on the Unicorn Engine to track this: https://github.com/unicorn-engine/unicorn/issues/1984

This resulted in the emulator silently segfaulting very frequently (on all pushes, pops, and many load instructions which reference the stack pointer). 

The fix to this is to simply first write the value of the CPSR register to the emulator, and then our stack pointer. This involves reordering the initialization order.

### DIFF
--- a/pwndbg/emu/emulator.py
+++ b/pwndbg/emu/emulator.py
@@ -222,9 +222,10 @@ class Emulator:
 
         # Initialize the register state
         for reg in (
-            list(self.regs.retaddr)
+            list(self.regs.flags)
+            + list(self.regs.retaddr)
             + list(self.regs.misc)
-            + list(self.regs.common)  # this includes the flags register
+            + list(self.regs.common_no_flag)  # this includes the flags register
         ):
             enum = self.get_reg_enum(reg)
 
--- a/pwndbg/emu/emulator.py
+++ b/pwndbg/emu/emulator.py
@@ -220,28 +220,9 @@ class Emulator:
         # (address_successfully_executed, size_of_instruction)
         self.last_single_step_result = InstructionExecutedResult(None, None)
 
-        # The specific order of this list is very important:
-        # Due to the behavior of Arm in the Unicorn engine,
-        # we must write the flags register after PC, and the stack pointer after the flags register.
-        # Otherwise, the values will be clobbered
-        # https://github.com/pwndbg/pwndbg/pull/2337
-        raw_list = (
-            [self.regs.pc]
-            + list(self.regs.flags)
-            + [self.regs.stack, self.regs.frame]
-            + list(self.regs.retaddr)
-            + list(self.regs.misc)
-            + list(self.regs.gpr)
-        )
-
-        # Get rid of duplicates and "None" values - some registers, like .frame are sometimes non-existent
-        reg_list = []
-        for x in raw_list:
-            if x and x not in reg_list:
-                reg_list.append(x)
 
         # Initialize the register state
-        for reg in reg_list:
+        for reg in self.regs.emulated_regs_order:
             enum = self.get_reg_enum(reg)
 
             if not reg:
--- a/pwndbg/emu/emulator.py
+++ b/pwndbg/emu/emulator.py
@@ -220,7 +220,6 @@ class Emulator:
         # (address_successfully_executed, size_of_instruction)
         self.last_single_step_result = InstructionExecutedResult(None, None)
 
-
         # Initialize the register state
         for reg in self.regs.emulated_regs_order:
             enum = self.get_reg_enum(reg)
--- a/pwndbg/emu/emulator.py
+++ b/pwndbg/emu/emulator.py
@@ -132,7 +132,7 @@ arch_to_reg_const_map = {
 ) = (0, 1, 2, 4, 8, 16, 32, 64, 128)
 
 DEBUG = NO_DEBUG
-# DEBUG = -1 # ALL
+DEBUG = -1 # ALL
 # DEBUG = DEBUG_EXECUTING | DEBUG_MEM_MAP | DEBUG_MEM_READ
 
 if DEBUG != NO_DEBUG:
@@ -225,7 +225,7 @@ class Emulator:
             list(self.regs.flags)
             + list(self.regs.retaddr)
             + list(self.regs.misc)
-            + list(self.regs.common_no_flag)  # this includes the flags register
+            + list(self.regs.common_no_flag)
         ):
             enum = self.get_reg_enum(reg)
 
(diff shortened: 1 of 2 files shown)


pwndbg/lib/regs.py
```diff
@@ -75,10 +75,12 @@ def __init__(
...
         # In 'common', we don't want to lose the ordering of:
-        self.common = []
+        self.common_no_flag = []
         for reg in gpr + (frame, stack, pc) + tuple(flags):
-            if reg and reg not in self.common:
-                self.common.append(reg)
+            if reg and reg not in self.common_no_flag:
+                self.common_no_flag.append(reg)
+
+        self.common = self.common_no_flag + list(flags)
```
OBarronCS: This was the fix that involved the least code changes I could think of - I'm all ears for any other way of separating all the relevant registers (while removing duplicates) from the flag register.

OBarronCS:
This fix causes another issue due to how Unicorn enables Thumb mode. Whenever PC is written to Unicorn (emu.reg_write), Unicorn implicitly overwrites the Thumb bit of the CPSR register with the bottom bit of PC. CPSR is initially written with the correct value (T bit enabled), then we write PC (which is 2-byte aligned, bottom bit always 0), which disables T bit. This wasn't an issue before because CPSR was the last thing we copied, so it would always have the same value as the process (Thumb bit not zeroed).

Two possible solutions:
1. Make writing PC in the initialization list a special case, checking the Thumb bit of the outer process and copying it to the least significant bit (this would purely act to preserve the T bit of the CPSR that has already been copied into the emulator, working around how Thumb-mode works in Unicorn)
2.  Write the CPSR register first at the start and then at the end. I found that if you write CPSR a second time (first before writing SP, than after writing SP), it will no longer zero out the SP register.


This script showcases the Thumb bit behavior of Unicorn
```python
import unicorn as U

uc = U.Uc(U.UC_ARCH_ARM, U.UC_MODE_ARM)

# Write CPSR with T bit enabled
uc.reg_write(U.arm_const.UC_ARM_REG_CPSR,0x30)
print(uc.reg_read(U.arm_const.UC_ARM_REG_CPSR))

# Now, write a value with bottom bit zero (all instructions are at least two-byte aligned)
# Add "| 1" to the address to write thumb bit
uc.reg_write(U.arm_const.UC_ARM_REG_PC,0x40810bc8)

# CPSR has T bit disabled
print(uc.reg_read(U.arm_const.UC_ARM_REG_CPSR))
``` 

This shows that option 2 work:
```python
#!/usr/bin/env python3

import unicorn as U

uc = U.Uc(U.UC_ARCH_ARM, U.UC_MODE_ARM)

print("First time:")

uc.reg_write(U.arm_const.UC_ARM_REG_SP,100)
# This returns 100, as expected
print(uc.reg_read(U.arm_const.UC_ARM_REG_SP))

# Now, write something to the CPSR register
uc.reg_write(U.arm_const.UC_ARM_REG_CPSR,0x30)

# Reading from stack pointer returns 0, instead of 100
print(uc.reg_read(U.arm_const.UC_ARM_REG_SP))

print("Second time:")
# Writing CPSR a second time
uc.reg_write(U.arm_const.UC_ARM_REG_SP,100)
print(uc.reg_read(U.arm_const.UC_ARM_REG_SP))


# Now, write something to the CPSR register
uc.reg_write(U.arm_const.UC_ARM_REG_CPSR,0x30)

# Reading from stack pointer now returns 100
print(uc.reg_read(U.arm_const.UC_ARM_REG_SP))
```

OBarronCS:
The core issue is caused by Unicorn starting in Arm supervisor mode, which is specified by bits in the CPSR register: https://github.com/unicorn-engine/unicorn/issues/1984#issuecomment-2270323998. There are banked version of SP and LR, meaning this bug applies to both SP and LR (https://developer.arm.com/documentation/dui0801/l/Overview-of-AArch32-state/Registers-in-AArch32-state).

OBarronCS:
Last commits fix all issues - we have to first write the `pc` register, then `cpsr,` then the stack and lr and any other registers. This results in all values being as they were written once the register initialization loop is done.

pwndbg/emu/emulator.py
```diff
@@ -220,12 +220,28 @@ def __init__(self) -> None:
...
+            + list(self.regs.gpr)
+        )
+
+        # Get rid of duplicates and "None" values - some registers, like .frame are sometimes non-existent
+        reg_list = []
+        for x in raw_list:
+            if x and x not in reg_list:
+                reg_list.append(x)
+
+        # Initialize the register state
```
OBarronCS, line 223: These lines mimic the register initialization loop in pwndbg/lib/regs.py.

We are computing this every time the emulator is instantiated - should we instead make a variable like `emulated_registers` that are in this order in the RegisterSet object?

pwndbg/emu/emulator.py
```diff
@@ -220,12 +220,28 @@ def __init__(self) -> None:
...
+        raw_list = (
+            [self.regs.pc]
+            + list(self.regs.flags)
+            + [self.regs.stack, self.regs.frame]
+            + list(self.regs.retaddr)
             + list(self.regs.misc)
-            + list(self.regs.common)  # this includes the flags register
-        ):
+            + list(self.regs.gpr)
+        )
```
disconnect3d: I don't remember but how often is this called? If often, then since this order is always the same, shall we maybe have a `regs.emulator_regs_order` field in each regs struct instead of re-computing this each time here?

pwndbg/emu/emulator.py
disconnect3d: Haha, commented on that. Generally, yes. Pls do this in another PR :)

disconnect3d: (or in this PR if it isn't merged yet)

OBarronCS: Great! Last commits do this

disconnect3d:
@OBarronCS lint violation:

```
+ vermin -vvv --no-tips -t=3.8- --eval-annotations --violations pwndbg tests binja_script.py gdbinit.py ida_script.py lldbinit.py
Detecting python files..
Analyzing 287 files using 4 processes..
!2, 3.9      /home/runner/work/pwndbg/pwndbg/pwndbg/lib/regs.py
  L88: builtin generic type annotation (list[..]) requires !2, 3.9

```

:<

OBarronCS:
Oop - fixed that 👍

PR #2342, opened 2024-08-06
vis_heap_chunks: Interpret count as addr if it's big enough

### LINKED ISSUE #2341: Interpret the first argument to vis_heap_chunks as an address if it's large enough
Oftentimes I want to look at some particular chunk and have to do `vis 1 <addr>` instead of just `vis <addr>`.

The count argument should be interpreted as an address if an address isn't provided and count is large enough.

It might be best if the count and addr positional arguments swapped places. If you're not printing the whole heap with `vis` or `vis -a` you likely want to see some particular chunk (and it's surrounding chunks) and trying to do that with just `<count>` sucks since a) you have to snipe the correct value and b) clutter your output with chunks you don't care about. So the commands I use most are `vis -a` and `vis 1 <addr>` (with a reduced `max-visualize-chunk-size` to minimize clutter as well).

The downside would be that you couldn't `vis 1` (for the `perthread_tcache_struct`) or `vis <num>` if you want more chunks to get a general overview of the state of the heap; but this could be alleviated with a flag that says "take the first chunk from the start of the heap" or making addr=0 or something mean that.

However I'm not sure how other people use it and it was probably put in that order for a reason so it feels unlikely that feature request would be accepted.

**Additional context**
Nearpc has this type of behaviour, as discussed here: https://github.com/pwndbg/pwndbg/issues/2332#issuecomment-2262766170

k4lizen:
Closes https://github.com/pwndbg/pwndbg/issues/2341

### DIFF
--- a/docs/commands/heap/vis_heap_chunks.md
+++ b/docs/commands/heap/vis_heap_chunks.md
@@ -21,7 +21,7 @@ usage: vis_heap_chunks [-h] [--beyond_top] [--no_truncate] [--all_chunks] [count
 
 |Positional Argument|Help|
 | :--- | :--- |
-|`count`|Number of chunks to visualize. (default: %(default)s)|
+|`count`|Number of chunks to visualize. If the value is big enough and addr isn't provided, this is interpreted as addr instead. (default: %(default)s)|
 |`addr`|Address of the first chunk.|
 
 ## Optional Arguments
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -944,7 +944,7 @@ group.add_argument(
     nargs="?",
     type=lambda n: max(int(n, 0), 1),
     default=pwndbg.config.default_visualize_chunk_number,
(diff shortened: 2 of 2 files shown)


k4lizen:
A. this is failing tests

k4lizen:
Ah no I guess its something to do with my local setup, it passes the CI/CD.
```
E         
E         Full diff:
E           [
E         +     'pwndbg will try to resolve the heap symbols via heuristic now since we '
E         +     'cannot resolve the heap via the debug symbols.',
E         +     'This might not work in all cases. Use `help set '
E         +     'resolve-heap-via-heuristic` for more details.',
E         +     '',
E               '',
E               '0x555555559000\t0x0000000000000000\t0x0000000000000291\t................',
E               '0x555555559010\t0x0000000000000000\t0x0000000000000000\t................',
E               '0x555555559020\t0x0000000000000000\t0x0000000000000000\t................',

```

PR #2345, opened 2024-08-07
Port the event system to the Debugger-agnostic API and move modules external to `gdblib` to it

mbrla0:

This PR adds an event interface, similar to the one currently exposed by `pwndbg.gdblib.events`, to the Debugger-agnostic API. This new event interface exposes the same event types as the `gdblib` interface - except for those that are not used outside `gdblib`, like `thread` and `before_prompt` - though one under a different name, better matching the terminology used in LLDB.

This PR also moves all uses of the `pwndbg.gdblib.events` interface in modules outside of `gdblib` to the new interface. To the best of my ability, I believe that behavior should be exactly the same between the two interfaces under GDB, and the tests seem to confirm that, but there might be something I'm missing, still.

Additionally, the LLDB implementation does not support this new interface and won't until the Pwndbg LLDB REPL gets merged. This is, to make a very long story short, due to the fact that LLDB does not have a robust way for us to listen to any events related to the lifetime of the process, unless we drive the lifetime and I/O of the process ourselves. I'll elaborate more on this issue when it comes time to submit the REPL PR.

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -23,7 +23,6 @@ import pwndbg.color.syntax_highlight as H
 import pwndbg.commands
 import pwndbg.commands.telescope
 import pwndbg.gdblib.disasm
-import pwndbg.gdblib.events
 import pwndbg.gdblib.heap_tracking
 import pwndbg.gdblib.nearpc
 import pwndbg.gdblib.regs
@@ -363,7 +362,7 @@ def context_ghidra(target=sys.stdout, with_banner=True, width=None):
         return banner + [message.error(e)]
 
 
-# @pwndbg.gdblib.events.stop
+# @pwndbg.dbg.event_handler(EventType.STOP)
 
 parser = argparse.ArgumentParser(
     description="Print out the current register, instruction, and stack context."
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -362,7 +362,6 @@ def context_ghidra(target=sys.stdout, with_banner=True, width=None):
         return banner + [message.error(e)]
 
 
-
 parser = argparse.ArgumentParser(
     description="Print out the current register, instruction, and stack context."
 )
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -362,7 +362,6 @@ def context_ghidra(target=sys.stdout, with_banner=True, width=None):
         return banner + [message.error(e)]
 
 
-# @pwndbg.dbg.event_handler(EventType.STOP)
 
 parser = argparse.ArgumentParser(
     description="Print out the current register, instruction, and stack context."
--- a/pwndbg/dbg/__init__.py
+++ b/pwndbg/dbg/__init__.py
@@ -274,8 +274,8 @@ class EventType(Enum):
           process execution to continue after it had been previously suspended.
         - `NEW_MODULE`: This event is fired when a new application module has
           been encountered by the debugger. This usually happens when a new
-          application module is into the memory space of the process being
-          debugged.
+          application module is loaded into the memory space of the process being
+          debugged. In GDB terminology, these are called `objfile`s.
     """
 
     START = 0
--- a/pwndbg/dbg/__init__.py
+++ b/pwndbg/dbg/__init__.py
@@ -9,8 +9,11 @@ from typing import Any
 from typing import Callable
 from typing import List
 from typing import Tuple
+from typing import TypeVar
 
 dbg: Debugger = None
+
+T = TypeVar("T")
 
 
 class Error(Exception):
@@ -251,6 +254,39 @@ class CommandHandle:
         raise NotImplementedError()
 
 
+class EventType(Enum):
+    """
+    Events that can be listened for and reacted to in a debugger.
+
+    The events types listed here are defined as follows:
+        - `START`: This event is fired some time between the creation of or
+          attachment to the process to be debugged, and the start of its
+          execution.
+        - `STOP`: This event is fired after execution of the process has been
+          suspended, but before control is returned to the user for interactive
+          debugging.
+        - `EXIT`: This event is fired after the process being debugged has been
+          detached from or has finished executing.
+        - `MEMORY_CHANGED`: This event is fired when the user interactively makes
+          changes to the memory of the process being debugged.
+        - `REGISTER_CHANGED`: Like `MEMORY_CHANGED`, but for registers.
+        - `CONTINUE`: This event is fired after the user has requested for
+          process execution to continue after it had been previously suspended.
+        - `NEW_MODULE`: This event is fired when a new application module has
+          been encountered by the debugger. This usually happens when a new
+          application module is into the memory space of the process being
+          debugged.
+    """
+
+    START = 0
+    STOP = 1
+    EXIT = 2
+    MEMORY_CHANGED = 3
+    REGISTER_CHANGED = 4
+    CONTINUE = 5
+    NEW_MODULE = 6
+
+
 class Debugger:
     """
     The base class representing a debugger.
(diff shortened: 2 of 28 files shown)


pwndbg/commands/context.py
```diff
@@ -363,7 +362,7 @@ def context_ghidra(target=sys.stdout, with_banner=True, width=None):
         return banner + [message.error(e)]
 
 
-# @pwndbg.gdblib.events.stop
+# @pwndbg.dbg.event_handler(EventType.STOP)
```
```suggestion
```

pwndbg/dbg/__init__.py
```diff
@@ -251,6 +254,39 @@ def remove(self) -> None:
         raise NotImplementedError()
 
 
+class EventType(Enum):
+    """
+    Events that can be listened for and reacted to in a debugger.
+
+    The events types listed here are defined as follows:
+        - `START`: This event is fired some time between the creation of or
+          attachment to the process to be debugged, and the start of its
```
disconnect3d, line 263: Are we 100%  sure it is also called on attach? It is very important that we get those docs correct

pwndbg/dbg/__init__.py
```diff
@@ -251,6 +254,39 @@ def remove(self) -> None:
...
+          execution.
+        - `STOP`: This event is fired after execution of the process has been
+          suspended, but before control is returned to the user for interactive
+          debugging.
+        - `EXIT`: This event is fired after the process being debugged has been
+          detached from or has finished executing.
```
disconnect3d, line 269: Same here: 100%  sure about detached?

pwndbg/dbg/__init__.py
```diff
@@ -251,6 +254,39 @@ def remove(self) -> None:
...
+        - `MEMORY_CHANGED`: This event is fired when the user interactively makes
+          changes to the memory of the process being debugged.
+        - `REGISTER_CHANGED`: Like `MEMORY_CHANGED`, but for registers.
+        - `CONTINUE`: This event is fired after the user has requested for
+          process execution to continue after it had been previously suspended.
+        - `NEW_MODULE`: This event is fired when a new application module has
+          been encountered by the debugger. This usually happens when a new
+          application module is into the memory space of the process being
```
disconnect3d:
```suggestion
          application module is loaded into the memory space of the process being
```

Also: why 'module' and not 'objfile'? Is that naming from llvm? Maybe we should say `new module (objfile)` in the description here

pwndbg/dbg/gdb.py
```diff
@@ -14,9 +14,24 @@
...
+# We pass the responsibility of event handling to gdblib.
+GDBLIB_EVENT_MAPPING = {
+    pwndbg.dbg_mod.EventType.EXIT: pwndbg.gdblib.events.exit,
+    pwndbg.dbg_mod.EventType.CONTINUE: pwndbg.gdblib.events.cont,
+    pwndbg.dbg_mod.EventType.START: pwndbg.gdblib.events.start,
+    pwndbg.dbg_mod.EventType.STOP: pwndbg.gdblib.events.stop,
+    pwndbg.dbg_mod.EventType.NEW_MODULE: pwndbg.gdblib.events.new_objfile,
+    pwndbg.dbg_mod.EventType.MEMORY_CHANGED: pwndbg.gdblib.events.mem_changed,
+    pwndbg.dbg_mod.EventType.REGISTER_CHANGED: pwndbg.gdblib.events.reg_changed,
+}
```
disconnect3d: Random thought: yet another misdirection :(

In theory we could have just pwndbg.dbg.events.X defined based on the debugger? heh

pwndbg/dbg/gdb.py
```diff
@@ -484,6 +496,18 @@ def selected_inferior(self) -> pwndbg.dbg_mod.Process | None:
...
+    def has_event_type(self, ty: pwndbg.dbg_mod.EventType) -> bool:
+        # Currently GDB supports all event types.
+        return True
+
+    @override
+    def event_handler(
+        self, ty: pwndbg.dbg_mod.EventType
+    ) -> Callable[[Callable[..., T]], Callable[..., T]]:
+        # Just call into the gdblib handler.
+        return GDBLIB_EVENT_MAPPING[ty]
```
disconnect3d: It would be better to do if elif else imho instead of introducing a new dict object unless it has another purpose

pwndbg/gdblib/arch.py
```diff
@@ -82,6 +82,10 @@ def _get_arch(ptrsize: int):
     else:
         endian = "big"
 
+    # Importing requires that `pwndbg.dbg` already be set up, so we have to do
+    # it here, rather then on the top level.
+    import pwndbg.gdblib.proc
```
disconnect3d, line 87: Why is there this requirement?

pwndbg/dbg/__init__.py
mbrla0: On GDB, it's called the first time `new_objfile` fires and there's a selected thread. The GDB tests expect [newly attached-to processes to have their symbol tables read](https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=gdb/testsuite/gdb.base/attach.exp;h=2b655e69f8badf31b7f66ec49f881737ccd7bb60;hb=HEAD), which implies that at least the objfile for the main executable will always be created. And experimentally I can confirm that it does seem to always get created. Meaning that we get a `start` event, in the same way that we would if were starting a process.

On LLDB, we fire that event ourselves, so we can guarantee that's the case.

pwndbg/dbg/__init__.py
mbrla0: The GDB docs explicitly [mention the detach case for the exit event](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Events-In-Python.html). Additionally, since we know from the previous comment that `start` does get fired on `attach`, Pwndbg already depends on the `exit` event being fired on detach to work correctly with attach/detach. Just in case, I've also confirmed experimentally that `exit` does get fired on `detach`.

As for LLDB, it's the same as the previous comment. We'll be driving the events ourselves, so we can pretty easily guarantee it gets fired on detach.

pwndbg/dbg/__init__.py
mbrla0: Yeap, that's what LLDB calls them. Generally, the Debugger-agnostic API sticks with LLDB whenever GDB and LLDB disagree on what to call something. But yeah, the clarification there is a good idea.

mbrla0: I've added "In GDB terminology, these are called `objfile`s." to the end of line for `NEW_MODULE`. I think that should clear things up?

pwndbg/gdblib/arch.py
mbrla0: `pwndbg.dbg.gdb` needs `pwndbg.gdblib`, which imports `pwndbg.gdlib.arch` on the top level, which needs `pwndbg.gdblib.proc`, which needs `pwndbg.gdblib.qemu`, which sets up an event handler. Since setting up an event handler now needs `pwndbg.dbg` to be set, we can't have that import on the top level.

pwndbg/dbg/gdb.py
mbrla0: > In theory we could have just pwndbg.dbg.events.X defined based on the debugger?

Yeah, we could, in the same way we could do `pwndbg.dbg.IS_GDB` to optimize `pwndbg.dbg.is_gdblib_available()`. Though considering there's only about ~50 uses of this function, all during initialization, I feel optimizing this in terms of time might not be all that crucial right now? We could leave that for later, since the interfaces can be renamed with a simple `sed` later on anyway

pwndbg/dbg/gdb.py
mbrla0: Yeah, now that I think of it, I don't think the supported event list is ever gonna change. When I fist picked the dictionary I'd intended the dictionary for a possible case where the types of supported events could change over startup. But now with it implemented I don't think it makes sense to keep it. Let me change it.

PR #2346, opened 2024-08-07
Small fix to annotations crash

OBarronCS:
This small PR fixes a crash in annotations in specific scenarios.

The control flow previously didn't account for a scenario where we generate an annotation for an instruction that also changes the program counter. Most instructions explicitly change the program counter (jump, ret), but in Arm, some instructions like `add` or `sub`, can have the PC as the target, and the codeflow didn't consider this, leading to a crash in these cases. If there's a jump target, it takes precedence. 

This would cause a crash once we have `sub` annotations on Arm.
```asm
.section .text
.global _start

_start:
ldr     R1, =_target
SUB PC, R1, #1

.THUMB
_target:
add r3,r4,r5
```


disconnect3d:
@OBarronCS shall we add a testcase for tihs?

PR #2347, opened 2024-08-11
Added a configuration option to disable auto update of dependencies after a git pull

### LINKED ISSUE #2235: Add a configuration option to disable auto dependency updates
We recently added a feature where Python dependencies are automatically updated when launching GDB. In the majority of cases this is best to leave on by default, but we should add a configuration option in case anyone wants to disable this functionality in some specific cases.

The thing is that pwndbg's normal configuration infra isn't loaded yet, so we can't use it in `gdbinit.py`. Maybe we just use an environment variable?

B1N4RY-P4R45173:
Fixes #2235 

Updated the requests module to 2.32.2

Added an environment variable to enable/disable auto update of dependencies after a git pull.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -46,13 +46,13 @@ wheel = ">=0.23.0,<1.0"
 
 [[package]]
 name = "babel"
-version = "2.14.0"
+version = "2.16.0"
 description = "Internationalization utilities"
 optional = false
-python-versions = ">=3.7"
+python-versions = ">=3.8"
 files = [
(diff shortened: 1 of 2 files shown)


B1N4RY-P4R45173:
There were a lot of tests failing when I pushed the updated `gdbinit.py`(added a if-else condition based on env variable) and `setup.sh`(setting the env variable for giving to option to enable/disable auto update). So I thought of upgrading the requests module first and then adding necessary changes to check changes in which file are causing that error. There are still two tests failing after only upgrading the request module. And one of them is a lint issue, I don't understand the reason for its failure.

B1N4RY-P4R45173:
Upgrading of the requests module will be done in a seperate PR

B1N4RY-P4R45173:
Upgrading of the `requests` library has been addressed in Pull Request #2348.

Issue #1967, opened 2023-12-20
Feature request: search for pointers to rwx memory


disconnect3d:
It would be nice to have a way to find pointers to memory pages with certain permissions, like RWX, either on a memory range or on a memory page name.

This feature was requested by one of our users.

Ideally, this should probably be a flag/implementation change into one of `p2p`, `probeleak` or `leakfind` commands, so that we don't create another command that does a similar thing.

disconnect3d:
The `probeleak <start-addr> <count> --flags rwx` will do this.

disconnect3d:
It could be nice to have `probeleak stack --flags rwx`.

disconnect3d:
Ideally we probably want to have such interface[s] for some of the commands like `search`, `probeleak`, etc:
```
command [address] [count=0x100]
command [address<0x10000]                 ==> count=address, address=some_default_register
command [memory mapping name filter]      ==> use single or multiple memory pages that match name?
command [memory mapping name:rwx]         ==> additionally filter name with perms?
command [addr1:addr2]                     ==> only perform the command on the given memory range
```

PR #2349, opened 2024-08-11
search pointer feature for #1967

0xhebi:

### DIFF
--- a/pwndbg/commands/probeleak.py
+++ b/pwndbg/commands/probeleak.py
@@ -21,9 +21,7 @@ def find_module(addr, max_distance):
 
     if not pages:
         if max_distance != 0:
-            mod_filter = (
-                lambda page: page.start - max_distance <= addr < page.end + max_distance
-            )
+            mod_filter = lambda page: page.start - max_distance <= addr < page.end + max_distance
             pages = list(filter(mod_filter, pwndbg.gdblib.vmmap.get()))
 
         if not pages:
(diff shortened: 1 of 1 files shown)


pwndbg/commands/probeleak.py
```diff
@@ -71,30 +73,54 @@ def flags_str2int(flags_s):
...
+    default=None,
+    help="Search for pointers to pages with specific permissions [e.g., rwx]",
+)
+
+parser.add_argument(
+    "--page-name",
+    type=str,
+    default=None,
+    help="Search for pointers to pages with a specific name",
+)
```
disconnect3d, line 99: The `--page-name` does the same thing as `--point-to`. We should have only one argument like this.

The description `--point-to` should probably be fixed

pwndbg/commands/probeleak.py
```diff
@@ -71,30 +73,54 @@ def flags_str2int(flags_s):
...
     type=str,
     default=None,
     help="flags of the page that you want the pointers point to. [e.g. rwx]",
 )
+parser.add_argument(
+    "--permissions",
+    type=str,
+    default=None,
+    help="Search for pointers to pages with specific permissions [e.g., rwx]",
+)
```
disconnect3d, line 92: This flags seems the same as `--flags`?

disconnect3d:
Seems like the whole of this PR/issue was already implemented in  `probeleak` as `--point-to <pagename partial match>` and `--flags <flags>`

Issue #2350, opened 2024-08-11
The GDB help is broken for commands


disconnect3d:
@mbrla0 I believe the debugger-agnostic command API broke the `help <command>` available from GDB:

```
pwndbg> help nearpc
This command is not documented.
```

The `--help` from argparse works though:
```
pwndbg> nearpc --help
usage: nearpc [-h] [-e] [pc] [lines]

Disassemble near a specified address.

positional arguments:
  pc             Address to disassemble near. If this is the only argument and the
                 value provided is small enough, it is interpreted as lines instead.
  lines          Number of lines to show on either side of the address.

options:
  -h, --help     show this help message and exit
  -e, --emulate  Whether to emulate instructions to find the next ones or just linearly
                 disassemble. (default: False)
```

But yeah, we do want to register the help strings into GDB interface.

mbrla0:
I failed to realize GDB used the _docstring_ of all things to determine the help string. #2351 should fix it.

disconnect3d:
Fixed in #2351

PR #2351, opened 2024-08-11
Fix missing help strings for Pwndbg commands in GDB

mbrla0:
As mentioned in #2350, moving command addition to the Debugger-agnostic API broke the GDB `help` command when used to get help for Pwndbg commands.

This PR adds a `doc` argument to the debugger-agnostic interface for command addition, which takes the documentation for the command, if one exists. It makes it so that the GDB implementation takes that value and sets it as the doc string (`__doc__`) for the newly created command object, which GDB then [takes and uses for the `help` command](https://sourceware.org/gdb/current/onlinedocs/gdb.html/CLI-Commands-In-Python.html#CLI-Commands-In-Python).

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -90,11 +90,13 @@ class Command:
         is_alias: bool = False,
         aliases: List[str] = [],
         category: CommandCategory = CommandCategory.MISC,
+        doc: str | None = None,
     ) -> None:
         self.is_alias = is_alias
         self.aliases = aliases
         self.category = category
         self.shell = shell
+        self.doc = doc
 
         if command_name is None:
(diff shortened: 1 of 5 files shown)


mbrla0:
Added a simple test that checks for the issue this PR fixes

pwndbg/commands/__init__.py
```diff
@@ -547,14 +549,15 @@ def __init__(
         file = io.StringIO()
```
disconnect3d, line 549: Btw I believe we should explicitly close the `io.StringIO` via a context manager:
```py
with io.StringIO() as file:
    ...
```

instead of not doing it at all.

disconnect3d (approved):

PR #2352, opened 2024-08-12
Expand Debugger-agnostic API in preparation for initial command ports

mbrla0:

This PR expands the Debugger-agnostic API to include support for querying properties of the architecture, looking up the name of a symbol at a given address, and querying the virtual memory map, all in the context of a given inferior. All of these functionalities are being added in preparation for an initial batch of commands that will get ported to the Debugger-agnostic API, and to LLDB.


pwndbg/dbg/__init__.py
```diff
@@ -55,6 +86,30 @@ def bottom_frame(self) -> Frame:
...
+        """
+        Returns whether this memory map was generated from a QEMU target.
+        """
+        raise NotImplementedError()
+
+    def has_reliable_perms(self) -> bool:
+        """
+        Returns whether the permissions in this memory map are reliable.
+        """
+        raise NotImplementedError()
```
disconnect3d, line 104: Do we actually ever use `reliable_perms` or plan to use it anywhere?

pwndbg/dbg/gdb.py
```diff
@@ -111,6 +136,25 @@ def bottom_frame(self) -> pwndbg.dbg_mod.Frame:
         return GDBFrame(value)
 
 
+class GDBMemoryMap(pwndbg.dbg_mod.MemoryMap):
+    def __init__(self, reliable_perms: bool, qemu: bool, pages: Sequence[pwndbg.lib.memory.Page]):
```
disconnect3d, line 140: Isn't it better to make the `reliable_perms` a 3rd argument with a default of `True` or `False`?

pwndbg/dbg/lldb.py
```diff
@@ -206,6 +230,24 @@ def cast(self, type: pwndbg.dbg_mod.Type | Any) -> pwndbg.dbg_mod.Value:
         return LLDBValue(self.inner.Cast(t.inner))
 
 
+class LLDBMemoryMap(pwndbg.dbg_mod.MemoryMap):
+    def __init__(self, pages: List[pwndbg.lib.memory.Page]):
+        self.pages = pages
+
+    @override
+    def is_qemu(self) -> bool:
+        # Figure a way to detect QEMU later.
```
disconnect3d:
```suggestion
        # TODO/FIXME: Figure a way to detect QEMU later.
```

pwndbg/dbg/lldb.py
```diff
@@ -221,6 +263,100 @@ def evaluate_expression(self, expression: str) -> pwndbg.dbg_mod.Value:
...
+            if region.IsReadable():
+                perms |= os.R_OK
+            if region.IsWritable():
+                perms |= os.W_OK
+            if region.IsExecutable():
+                perms |= os.X_OK
+
+            # LLDB doesn't actually tell us the offset of the mapped file, just
+            # whether it is mapped or not.
+            offset = 0 if not region.IsMapped() else 1
```
disconnect3d: Is it ok we set offset=1 if its mapped?

pwndbg/dbg/lldb.py
```diff
@@ -221,6 +263,100 @@ def evaluate_expression(self, expression: str) -> pwndbg.dbg_mod.Value:
...
+
+            objfile = region.GetName()
+            if objfile is None:
+                # LLDB will sometimes give us overlapping ranges with no name.
+                # For now, we ignore them, since GDB does not show them.
+                continue
```
disconnect3d: Please move it earlier: we don't need to process perms or offset if we are just continuing?

pwndbg/dbg/lldb.py
```diff
@@ -221,6 +263,100 @@ def evaluate_expression(self, expression: str) -> pwndbg.dbg_mod.Value:
...
+            )
+
+        return LLDBMemoryMap(pages)
+
+    @override
+    def symbol_name_at_address(self, address: int) -> str | None:
+        addr = lldb.SBAddress(address, self.target)
+        ctx = self.target.ResolveSymbolContextForAddress(addr, lldb.eSymbolContextSymbol)
+
+        if not ctx.IsValid():
```
disconnect3d:
```suggestion
        if not ctx.IsValid() or not ctx.symbol.IsValid():
```

pwndbg/dbg/lldb.py
```diff
@@ -221,6 +263,100 @@ def evaluate_expression(self, expression: str) -> pwndbg.dbg_mod.Value:
...
+            return None
+
+        return ctx.symbol.name
+
+    @override
+    def arch(self) -> pwndbg.dbg_mod.Arch:
+        endian0 = self.process.GetByteOrder()
+        endian1 = self.target.GetByteOrder()
+
+        if endian0 == lldb.eByteOrderInvalid:
```
disconnect3d: What? Why? pls document when this happens etc.

pwndbg/lib/memory.py
```diff
@@ -134,6 +132,11 @@ def permstr(self) -> str:
         )
 
     def __str__(self) -> str:
+        # This module requires GDB, so it causes import failures in unit tests.
+        # This will stop being a problem as soon as this module gets ported to
+        # aglib.arch, but, for now we have to add this as a quick stopgap.
+        import pwndbg.gdblib.arch
```
disconnect3d, line 138: I don't like this but ok

disconnect3d (commented):
See comments

pwndbg/dbg/__init__.py
mbrla0: We do. There's a spot in the `vmmap` command that uses it for a warning. It's only really used there, but that was a way to convey that bit of information to it without having it rely on explicit GDB version checks after it gets ported.

pwndbg/dbg/gdb.py
mbrla0: Maybe? Though, we always set it, anyway, so I don't see how much of a benefit that would bring?

pwndbg/dbg/lldb.py
mbrla0: No, that's a mistake on my part. I totally forgot that that got changed to zero in a later commit, and didn't get backported. Lemme change it now.

pwndbg/lib/memory.py
mbrla0: This should get removed in the next one or two PRs.

pwndbg/dbg/lldb.py
mbrla0: Yeah, that makes sense. Will do.

pwndbg/dbg/lldb.py
mbrla0: It's a fallback for when the process doesn't have a sensible architecture, but the target might. I've added a comment explaining it.

pwndbg/dbg/lldb.py
mbrla0: Done

mbrla0:
Anything else, @disconnect3d?

disconnect3d (approved):

PR #2354, opened 2024-08-13
Make breakpoint tags more persistent in Binary Ninja integration

Aplet123:
Currently, both breakpoint and current PC icons are set as auto tags, which aren't saved to the bndb and get removed on reanalysis (which can happen when a function name changes).

This PR changes breakpoint tags to be user tags, which are saved and don't get automatically removed, which should be desirable behavior for decompiler-set breakpoints.

### DIFF
--- a/binja_script.py
+++ b/binja_script.py
@@ -40,20 +40,27 @@ def get_widest_func(bv: binaryninja.BinaryView, addr: int) -> binaryninja.Functi
 
 
 # workaround for BinaryView.add_tag not supporting auto tags
-def add_auto_data_tag(bv: binaryninja.BinaryView, addr: int, name: str, desc: str) -> None:
-    tag = binaryninja.core.BNCreateTag(bv.get_tag_type(name).handle, desc)
-    binaryninja.core.BNAddTag(bv.handle, tag, False)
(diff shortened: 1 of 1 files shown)


binja_script.py
```diff
@@ -40,20 +40,27 @@ def get_widest_func(bv: binaryninja.BinaryView, addr: int) -> binaryninja.Functi
 
 
 # workaround for BinaryView.add_tag not supporting auto tags
-def add_auto_data_tag(bv: binaryninja.BinaryView, addr: int, name: str, desc: str) -> None:
-    tag = binaryninja.core.BNCreateTag(bv.get_tag_type(name).handle, desc)
-    binaryninja.core.BNAddTag(bv.handle, tag, False)
-    binaryninja.core.BNAddAutoDataTag(bv.handle, addr, tag)
+def add_data_tag(
```
disconnect3d, line 43: Is the comment above still valid btw?

disconnect3d: May be worth to detail the problem a bit more or why this is a workaround/how it fixes it etc

disconnect3d:
Merged, but pls see comment

binja_script.py
Aplet123: Should still be valid, as the PC tag still uses auto tags.

PR #2357, opened 2024-08-14
Fix Register Read Bug

OBarronCS:
This PR fixes a bug in the reading of registers in GDB as used in `pwndbg.commands.fix` that causes a crash in `cyclic`, `dt`, and `bp` on certain inputs.

When reading a register from a frame in GDB, it can raise a "gdb.error", and through testing, a "ValueError" as well when the input is a random string. This PR simply adds the ValueError to the list of exceptions to catch.

Fix is currently used by 3 functions - `cyclic`, `dt`, and `bp`. 

Note that there is one other place in the codebase where `frame.read_register` is called: 

https://github.com/pwndbg/pwndbg/blob/707db877f3b27b099d1794cc55278d422832aee5/pwndbg/gdblib/regs.py#L35-L42

This case doesn't catch `gdb.error` - should this be added here as well? cc @Aplet123

### DIFF
--- a/pwndbg/dbg/gdb.py
+++ b/pwndbg/dbg/gdb.py
@@ -29,7 +29,7 @@ class GDBRegisters(pwndbg.dbg_mod.Registers):
     def by_name(self, name: str) -> pwndbg.dbg_mod.Value | None:
         try:
             return GDBValue(self.frame.inner.read_register(name))
-        except gdb.error:
+        except (gdb.error, ValueError):
             # GDB throws an exception if the name is unknown, we just return
             # None when that is the case.
             pass


disconnect3d:
or cc: @mbrla0

Aplet123:
Can `gdb_get_register` even return `None`? Seems to me like it should always return a register value or error out. Maybe error handling should just be left for the caller? The function can still throw `ValueError` as-is because if a register doesn't exist the call in the except will error.

disconnect3d:
I think the "reg is None" case comes from the past (previous and current implementation):
```
pwndbg> pi pwndbg.gdblib.regs.rax
140737353867632
pwndbg> pi pwndbg.gdblib.regs.some_nonexistent_regs is None
True
```

Also the `regs.<reg>` will return `None` anytime a process is not alive.

But I agree with what @Aplet123 wrote. We should rather raise an exception in such a case and leave it up for the caller?

mbrla0:
> We should rather raise an exception in such a case and leave it up for the caller?

Ultimately we have to pick some way to communicate that there is no register under a given name. GDB is very exception-happy when it comes to these things, but ultimately I picked returning `None` for `by_name`, since it tracks better with how `pwndbg.gdblib.regs` does it, and also how LLDB does it. So, in this case, a return value of `None` only means there is no register by the given name, so the same semantics of calling `.get()` on a dictionary, for example.

Is there a good reason for us to raise an exception there over just returning `None`, and letting whoever is directly consuming the Debugger-agnostic API handle the case where a register name isn't right the way they see best? Maybe we should more finely separate GDB exceptions that indicate things other than a given register not existing, if those come from `gdb.Frame.read_register`?

PR #2355, opened 2024-08-13
fix nix + charset + bump dependencies

patryk4815:
Fixes:
- fix charset startup 
- fix nix'ed version

Updates:
- bump nix & python dependencies
- gdb-15.1
- python-3.12.4

Before was:
- gdb-13.2
- python-3.11.5

PR #2359, opened 2024-08-14
Fix mypy errors reported with version 1.11.1

peace-maker:
Fix typing errors noticed by bumping the mypy version in #2355

### DIFF
--- a/pwndbg/commands/argv.py
+++ b/pwndbg/commands/argv.py
@@ -71,11 +71,11 @@ class argv_function(gdb.Function):
     def __init__(self) -> None:
         super().__init__("argv")
 
-    def invoke(self, number=0):
-        number = int(number)
+    def invoke(self, number_value: gdb.Value = gdb.Value(0), *args: gdb.Value) -> gdb.Value:
+        number = int(number_value)
 
         if number > pwndbg.gdblib.argv.argc:
-            return 0
+            return gdb.Value(0)
 
         ppchar = pwndbg.gdblib.typeinfo.pchar.pointer()
         value = gdb.Value(pwndbg.gdblib.argv.argv)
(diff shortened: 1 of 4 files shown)


pwndbg/commands/argv.py
```diff
@@ -94,11 +94,11 @@ class envp_function(gdb.Function):
     def __init__(self) -> None:
         super().__init__("envp")
 
-    def invoke(self, number=0):
-        number = int(number)
+    def invoke(self, number_value: gdb.Value = gdb.Value(0), *args: gdb.Value) -> gdb.Value:
```
disconnect3d, line 97: I wonder if we should just skip linting this line, heh...

pwndbg/commands/argv.py
```diff
@@ -94,11 +94,11 @@ class envp_function(gdb.Function):
...
+        number = int(number_value)
 
         if number > pwndbg.gdblib.argv.envc:
-            return pwndbg.gdblib.typeinfo.void
+            return pwndbg.gdblib.typeinfo.void.optimized_out()
```
disconnect3d, line 101: What does this optimized_out do here?

pwndbg/commands/argv.py
```diff
@@ -132,8 +132,11 @@ class environ_function(gdb.Function):
     def __init__(self) -> None:
         super().__init__("environ")
 
-    def invoke(self, name):
-        name = name.string() + "="
+    def invoke(self, name_value: gdb.Value = gdb.Value(""), *args: gdb.Value) -> gdb.Value:
```
disconnect3d, line 135: Tbh I didn't even know we had such funcs. Is this sth like `print $environ("sth")` and `print $argv(0)` etc? nice

pwndbg/gdblib/argv.py
```diff
@@ -36,6 +36,9 @@ def update() -> None:
     ptrsize = pwndbg.gdblib.arch.ptrsize
     ptrbits = 8 * ptrsize
 
+    if sp is None:
+        return
```
disconnect3d: Isn't this case handled by the try except below too?

I believe the sp should never be None here at this point and this check is redundant... if the linter cries, we should supress it

pwndbg/commands/argv.py
peace-maker: It turns the Type into a Value and tells the caller that there is no information about the actual value of the Value :D

The functions expects a `gdb.Value` return instead of a `gdb.Type` one.

pwndbg/gdblib/argv.py
peace-maker: This appears to have been cried at from pyright and mypy ignores it, so I'll just revert.

Issue #2271, opened 2024-07-01
The xuntil command should accept address expressions


disconnect3d:
TL;DR: `xuntil $rip+4` or `xuntil main` should work.

disconnect3d:
The problem here is that the `target` argument in the `xuntil` implementation is specified as a `str` type.
It is defined in the code as shown below.

Initially I thought this was done by accident but I feel like it was done to be able to set breakpoints based on string, which is stupid. We can resolve the argument ourselves (by using an `int` type) and by doing that, we will support more use cases.

https://github.com/pwndbg/pwndbg/blob/12d8c3960f3be6a7a087745be6c321a9a390bbb2/pwndbg/commands/peda.py#L22-L49

If it would be specified as `int`, this would be fixed. This is because we handle the `int` type specially in the `ArgparsedCommand` and we use `fix_int_reraise` instead:
https://github.com/pwndbg/pwndbg/blob/12d8c3960f3be6a7a087745be6c321a9a390bbb2/pwndbg/commands/__init__.py#L594-L595

This `fix_int_reraise` is then evaluating the GDB expression and return an int for it.

disconnect3d:
Whoever takes this issue to fix it, please also refactor this `target` argument a bit: we do not need to do `int(target, 0)` if we get it as an `int` type.

disconnect3d:
This was fixed in https://github.com/pwndbg/pwndbg/pull/2360

PR #2360, opened 2024-08-14
resolving address expression for #2271

0xhebi:
This will as well check if process is running , in case it's not we will execute `starti` and rebase the target since there was the problem of not being able to breakpoint

### DIFF
--- a/pwndbg/commands/peda.py
+++ b/pwndbg/commands/peda.py
@@ -19,9 +19,7 @@ def getfile() -> None:
     print(repr(pwndbg.auxv.get().AT_EXECFN))
 
 
-parser = argparse.ArgumentParser(
-    description="Continue execution until an address or function."
-)
+parser = argparse.ArgumentParser(description="Continue execution until an address or expression.")
 parser.add_argument("target", type=int, help="Location to stop execution at")
 
 
--- a/pwndbg/commands/peda.py
+++ b/pwndbg/commands/peda.py
@@ -22,18 +22,14 @@ def getfile() -> None:
 parser = argparse.ArgumentParser(
     description="Continue execution until an address or function."
 )
-parser.add_argument("target", type=int, help="Address or function to stop execution at")
+parser.add_argument("target", type=int, help="Location to stop execution at")
 
 
 @pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.NEXT)
 def xuntil(target) -> None:
-    running = pwndbg.gdblib.proc.alive
-    if not running:
-        print(message.notice("Process is not running. Starting process..."))
-        gdb.execute("starti", from_tty=False, to_string=True)
-        target = pwndbg.gdblib.functions.rebase(target)
     try:
         addr = target
+
         if not pwndbg.gdblib.memory.peek(addr):
             print(message.error("Invalid address %#x" % addr))
             return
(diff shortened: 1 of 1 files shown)


pwndbg/commands/peda.py
```diff
@@ -19,15 +19,21 @@ def getfile() -> None:
...
+parser.add_argument("target", type=int, help="Address or function to stop execution at")
 
 
 @pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.NEXT)
 def xuntil(target) -> None:
+    running = pwndbg.gdblib.proc.alive
+    if not running:
+        print(message.notice("Process is not running. Starting process..."))
+        gdb.execute("starti", from_tty=False, to_string=True)
+        target = pwndbg.gdblib.functions.rebase(target)
```
disconnect3d: We don't want to use `gdblib` in `commands/*` since this will work only in GDB and not in LLDB

pwndbg/commands/peda.py
```diff
@@ -19,15 +19,21 @@ def getfile() -> None:
...
+)
+parser.add_argument("target", type=int, help="Address or function to stop execution at")
 
 
 @pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.NEXT)
 def xuntil(target) -> None:
+    running = pwndbg.gdblib.proc.alive
+    if not running:
+        print(message.notice("Process is not running. Starting process..."))
+        gdb.execute("starti", from_tty=False, to_string=True)
```
disconnect3d: Ideally, we want debugger agnostic way to do this if possible

pwndbg/commands/peda.py
```diff
@@ -19,15 +19,21 @@ def getfile() -> None:
     print(repr(pwndbg.auxv.get().AT_EXECFN))
 
 
-parser = argparse.ArgumentParser(description="Continue execution until an address or function.")
-parser.add_argument("target", type=str, help="Address or function to stop execution at")
+parser = argparse.ArgumentParser(
+    description="Continue execution until an address or function."
+)
+parser.add_argument("target", type=int, help="Address or function to stop execution at")
```
disconnect3d:
```suggestion
parser.add_argument("target", type=int, help="Location (address, symbol or expression) to stop execution at")
```

pwndbg/commands/peda.py
```diff
@@ -19,15 +19,21 @@ def getfile() -> None:
...
 
 
 @pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.NEXT)
 def xuntil(target) -> None:
+    running = pwndbg.gdblib.proc.alive
+    if not running:
```
disconnect3d:
```suggestion
    if not pwndbg.gdblib.proc.alive:
```
No need to introduce a variable

pwndbg/commands/peda.py
disconnect3d: This will not work if the process died and we restart it with `xuntil main` - at this point the symbol will already be rebased and we will rebase it again

pwndbg/commands/peda.py
disconnect3d: But don't worry about it for now since its not there yet; from discord:
image

disconnect3d:
This seems to allow us to distinguish if GDB resolved the base address and would rebase symbols:

image

(note: I use `memoize` to disable caching since it seems the result of `.files()` was cached from previous call -- seems this should be fixed!)

PR #2367, opened 2024-08-16
New Pwndbg Architecture

OBarronCS:
This PR implements a new class based mechanism to represent architecture in Pwndbg.

There are various different modules around the codebase that relate to inspecting the process based on the current architecture that we detect (is it arm? or x86? In that case, use this set of registers, and this ABI and SyscallABI).

Additionally, spread through the codebase are long `if/else` (`if pwndbg.gdblib.arch.current == "some_arch_name": do this`), that do some minor architecture specific logic - often just setting a local variable, such as fetching the relevent Capstone/Unicorn constants to pass to the external API.

This proposes an API to both simplify and clean up the code. A parent class defines the API - the various attributes and methods - and provides sane defaults. Subclasses for specific architectures can optionally override these defaults to implement logic.

The current architecture is represented at run-time as an instance of a class. This instance is seen globally throughout the codebase.

### DIFF
--- a/pwndbg/gdblib/arch.py
+++ b/pwndbg/gdblib/arch.py
@@ -1,6 +1,18 @@
 from __future__ import annotations
 
 from typing import Literal
+from typing_extensions import override
+
+import struct
+from typing import Literal
+from typing import Dict
+from typing import Tuple
+from typing import TypeVar
+from typing import Type
+
+
+from pwndbg.lib.abi import ABI, SyscallABI,DEFAULT_ABIS,SYSCALL_ABIS,SIGRETURN_ABIS
+from pwndbg.lib.regs import RegisterSet
 
 import gdb
 import pwnlib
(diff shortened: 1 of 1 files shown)


OBarronCS:
Current road block has to do with how the codebase will access this PwndbgArchitecture object.

Previously, the `Arch` object has been an instance of the `Arch` class - and it has been the same instance through the entire runtime of the program. This allows other files to `import pwndbg.gdblib.arch`, and then do `pwndbg.gdblib.arch.current` to access the `current` value. This works because the import initially gets the instance of the arch object, and the instance itself does not change. The gdblib architecture handler will set the values of this instance (arch.current = "arm") as architecture changes are detected, but the instance stays the same so all the imports are never stale.

In the new approach, when the class changes (a new architecture is detected), the instance itself must be replaced by an instance of the new class (say `ArchArch`). All of the imports around the codebase will always refer to the initial architecture object, and not the new one. We need this imported reference to always refer to the latest instance of the class.

disconnect3d:
> Current road block has to do with how the codebase will access this PwndbgArchitecture object.
> 
> Previously, the `Arch` object has been an instance of the `Arch` class - and it has been the same instance through the entire runtime of the program. This allows other files to `import pwndbg.gdblib.arch`, and then do `pwndbg.gdblib.arch.current` to access the `current` value. This works because the import initially gets the instance of the arch object, and the instance itself does not change. The gdblib architecture handler will set the values of this instance (arch.current = "arm") as architecture changes are detected, but the instance stays the same so all the imports are never stale.
> 
> In the new approach, when the class changes (a new architecture is detected), the instance itself must be replaced by an instance of the new class (say `ArchArch`). All of the imports around the codebase will always refer to the initial architecture object, and not the new one. We need this imported reference to always refer to the latest instance of the class.

I guess it should be sth like: `arch.current` ;).

The `arch` should be a module and `current` should be a global variable in a module that will be overwritten with architecture object when a different architecture is detected/set.

We just need to be sure that we never do `from arch import current` since this will copy the reference to the global scope of another script/module and so if the `arch.current` will change, such imported `current` will still refer to the old instance.

pwndbg/gdblib/arch.py
```diff
@@ -66,12 +73,154 @@ def get_thumb_mode_string() -> Literal["arm", "thumb"] | None:
...
+]
+EndianType = Literal["little", "big"]
+
+FMT_LITTLE_ENDIAN = {1: "B", 2: "<H", 4: "<I", 8: "<Q"}
+FMT_BIG_ENDIAN = {1: "B", 2: ">H", 4: ">I", 8: ">Q"}
+
+
+class PwndbgArchitecture:
+    # Registry of all instances
+    arch_registry: Dict[ArchType, PwndbgArchitecture] = {}
```
disconnect3d, line 104: Do we really need this registry? When? What for?

pwndbg/gdblib/arch.py
```diff
@@ -66,12 +73,154 @@ def get_thumb_mode_string() -> Literal["arm", "thumb"] | None:
...
+        return PwndbgArchitecture.arch_registry[name]
+
+    def __init__(self, name: ArchType) -> None:
+        """
+        Calling the constructor will register the class with global list of PwndbgArchitectures
+        """
+        if name is not None:
+            self.arch_registry[name] = self
+
+        self.name: ArchType = name
```
disconnect3d, line 121: Can we add a `__slots__` and move all fields of such class there?

I belivee that this object should never get new fields attached to it, so it probably makes sense to use `__slots__` here, especially that it can also be accessed many times.

pwndbg/gdblib/arch.py
```diff
@@ -66,12 +73,154 @@ def get_thumb_mode_string() -> Literal["arm", "thumb"] | None:
...
+        self.current = self.name
```
disconnect3d, line 122: Eventually, this field should probably be removed in favor of just `.name`, but it can stay for now if it makes it simpler.

pwndbg/gdblib/arch.py
```diff
@@ -66,12 +73,154 @@ def get_thumb_mode_string() -> Literal["arm", "thumb"] | None:
...
+    def unpack(self, data: bytes) -> int:
+        return struct.unpack(self.fmt, data)[0]
+
+    def pack_size(self, integer: int, size: int) -> bytes:
+        return struct.pack(self.fmts[size], integer & self.ptrmask)
+
+    def unpack_size(self, data: bytes, size: int) -> int:
+        return struct.unpack(self.fmts[size], data)[0]
+
+    def read_thumb_bit(self) -> Literal[0, 1]:
```
disconnect3d, line 176: Is this needed for non-arm archs?

disconnect3d:
@OBarronCS ping

OBarronCS:
This is still on my radar  - I think it would be best to wait to push changes to the Architecture setup, as it is referenced in 70+ files in the codebase, and ongoing LLDB work is making structural changes that should take precedence and will inform the implementation.

patryk4815:
@OBarronCS I was thinking about adding `platform` and `environment` info to `pwndbg.aglib.arch` object. (related: https://github.com/pwndbg/pwndbg/issues/2545 )
Currently it is mess a lot of if statements: is_remote, is_linux, is_qemu, is_android, is_qemuuser etc,

What I see in `platform` (more like enum or class, not string):
freebsd / netbsd / openbsd / ios / tvos / macosx / linux / windows / android / baremetal

What I see in `environment` (more like enum or class, not string):
qemu / baremetal(?) / qemuuser / remote(?)

patryk4815:
Arch should have more architectures? Like:
- arm64e ?  or function like `pwndbg.aglib.arch.current.is_aarch64().is_e()`
- sparcv9 or function like `pwndbg.aglib.arch.current.is_sparc().is_v9()`
- mips32r6  or function like `pwndbg.aglib.arch.current.is_mips().is_r6()`
- i8086 
- i686
- armcm like: `pwndbg.aglib.arch.current.is_arm().read_thumb_bit()`
🤔 

And in few places we have this ugly code:
```py
    elif pwndbg.aglib.arch.current == "sparc":
        if "v9" in gdb.newest_frame().architecture().name():
            extra = CS_MODE_V9
        else:
            # The ptrsize base modes cause capstone.CsError: Invalid mode (CS_ERR_MODE)
            extra = 0
```
```py
    elif (
        pwndbg.aglib.arch.current == "mips"
        and "isa32r6" in gdb.newest_frame().architecture().name()
    ):
        extra = CS_MODE_MIPS32R6
```
```py
        elif (
            arch == "mips"
            and "isa32r6" in gdb.newest_frame().architecture().name()
        ):
            mode |= U.UC_MODE_MIPS32R6
```

Maybe unicorn / capstone `modes` should be handled inside `pwndbg/aglib/arch.py`
Like: `pwndbg.aglib.arch.get_capstone_mode()`
Like: `pwndbg.aglib.arch.get_unicorn_mode()`
🤔 

Another my disliked code (it should be `platform` dependent)
Maybe it should be like: `pwndbg.aglib.arch.get_arch_by_instruction(instr) -> returns Arch obj`

https://github.com/pwndbg/pwndbg/blob/30f7ea08dcf90152db8ddb24f94616a2b10a443b/pwndbg/aglib/disasm/x86.py#L383-L401

disconnect3d:
@OBarronCS ping, what's the state of it?

OBarronCS:
Lots of merge conflicts/changes in the codebase since this was initially pushed, for simplicity I opened a new PR #2771

PR #2369, opened 2024-08-16
Enhance canary in telescope function

OBarronCS:
This PR adds text to indicate a canary value whenever it is encountered in the telescope function.

canary
reg_canary

### DIFF
--- a/pwndbg/enhance.py
+++ b/pwndbg/enhance.py
@@ -44,6 +44,14 @@ def format_small_int_pair(first: int, second: int) -> Tuple[str, str]:
 
 
 def int_str(value: int) -> str:
+    """
+    This function attempts to enhance the value of an int.
+    First, it checks if the integer is a stack canary.
+    Then, it attempts to interpret the bytes as a short strings
+    """
+    if value == pwndbg.commands.canary.canary_value()[0]:
+        return f"{value} (canary)"
+
     retval = format_small_int(value)
 
     # Try to unpack the value as a string
--- a/pwndbg/enhance.py
+++ b/pwndbg/enhance.py
@@ -49,10 +49,11 @@ def int_str(value: int) -> str:
     First, it checks if the integer is a stack canary.
     Then, it attempts to interpret the bytes as a short strings
     """
-    if value == pwndbg.commands.canary.canary_value()[0]:
-        return f"{value} (canary)"
 
     retval = format_small_int(value)
+
+    if value == pwndbg.commands.canary.canary_value()[0]:
+        return f"{retval} (canary)"
 
     # Try to unpack the value as a string
     packed = pwndbg.gdblib.arch.pack(int(value))


pwndbg/enhance.py
```diff
@@ -44,6 +44,14 @@ def format_small_int_pair(first: int, second: int) -> Tuple[str, str]:
 
 
 def int_str(value: int) -> str:
```
OBarronCS, line 46: Do we want to rename this function given it's expanded role?

pwndbg/enhance.py
```diff
@@ -44,6 +44,14 @@ def format_small_int_pair(first: int, second: int) -> Tuple[str, str]:
...
+    """
+    This function attempts to enhance the value of an int.
+    First, it checks if the integer is a stack canary.
+    Then, it attempts to interpret the bytes as a short strings
+    """
+    if value == pwndbg.commands.canary.canary_value()[0]:
```
OBarronCS: The canary_value() function is currently not cached. We should probably cache it, because this get called VERY often, and each time it will read from memory.

OBarronCS: It just needs be cached in a smart way - wait for the memory page where the canary is to be loaded in the first place, then assume it won't change during the program execution?

disconnect3d:
Wait, why is the RAX on the screenshot not in hex but in decimal? :D

pwndbg/enhance.py
disconnect3d: The problem is: it may change ;D.

disconnect3d: Btw this calls a function from `pwndbg.commands.*` - those should likely almost never be called externally... We should probably move that canary function to some lib counterpart.

OBarronCS:
Oop - had too much tunnel vision on the `(canary)` hahaha

OBarronCS:
Fixed: image

pwndbg/enhance.py
disconnect3d: This should also be gated somehow so that this is never called when:
1) debugging linux kernel
2) debugging embedded devices

etc.

OBarronCS: :0 didn't know the stack canary could change during the invocation of the program. Does the value pointed at by AT_RANDOM change? Or is there another way it changes?

disconnect3d:
Few things here:

1. We call a function from `pwndbg.commands.*` - we should probably move that canary function to some lib counterpart so that we don't require this part of code to import from a command
2. This needs to be gated somehow: we don't want to call this code ever (and ideally, don't even perform any check) when we debug 1) linux kernel; 2) embedded device; 3) any other target where reading memory is not cheap, heh.

OBarronCS:
Given the performance concerns, I'm closing this for now. If anyone thinks of a clean way to implement this that addresses the comments above, please feel free to take this up.

PR #2371, opened 2024-08-16
Improve attachp: fix partial match, add --user and --all

disconnect3d:
This commit fixes the attachp command so it has a better partial matching logic.

It also adds `--user <uid|username>` and `--all` options.

### DIFF
--- a/pwndbg/commands/attachp.py
+++ b/pwndbg/commands/attachp.py
@@ -52,11 +52,22 @@ Original GDB attach command help:
 )
 
 parser.add_argument("--no-truncate", action="store_true", help="dont truncate command args")
-parser.add_argument("target", type=str, help="pid, process name or device file to attach to")
+parser.add_argument("--user", type=str, default=None, help="username or uid to filter by")
+parser.add_argument(
+    "-a",
+    "--all",
+    action="store_true",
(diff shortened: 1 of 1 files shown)


pwndbg/commands/attachp.py
```diff
@@ -76,32 +87,48 @@ def attachp(no_truncate, target) -> None:
             resolved_target = target
 
         else:
+            # Note: we can't use `ps -C <target>` because this does not accept process names with spaces
+            # so target='a b' would actually match process names 'a' and 'b' here
+
```
disconnect3d:
```suggestion
# so instead, we will filter by process name or full cmdline later on
```

pwndbg/commands/attachp.py
```diff
@@ -76,32 +87,48 @@ def attachp(no_truncate, target) -> None:
             resolved_target = target
 
         else:
+            # Note: we can't use `ps -C <target>` because this does not accept process names with spaces
+            # so target='a b' would actually match process names 'a' and 'b' here
+# so instead, we will filter by process name or full cmdline later on
```
disconnect3d:
```suggestion
            # so instead, we will filter by process name or full cmdline later on
```

PR #2374, opened 2024-08-17
Annotations Tests

OBarronCS:
This PR adds dozens of tests relating to our new cross-arch annotations feature. In sum, Pwndbg now has first-class support for RISC-V, ARM, AArch64, and the MIPS architecture as their disassembly views have been implemented to match the x86 experience.

Many of the tests focus on edge cases/interesting behavior that I came across during the development of the different arches annotations. Our use of the Unicorn emulator (and syncing it with Capstone so that they are on the same instruction at all times) can break in multiple subtle ways depending on the arch, which can manifest in very-hard-to-debug ways, so the goal here is to cover enough scenarios so that we can confidently make changes without breaking something that a user might only catch a long ways in the future.

Keys things that are tested are branching and determining jump targets correctly, caching issues, and the proper display of instruction annotations. The large difficulties in these annotations is ensuring that we are properly managing the Unicorn Emulator and the Capstone disassembler such that they are synced on instruction address and type, that Unicorn is in the correct state, and that our manual assessment of things like jumps targets and conditions are aligned with the direction that Unicorn is emulating. If we can prevent a desync between Unicorn/Capstone (or prevent Unicorn from crashing or having incorrect register values) and we are accurately determining branch targets and conditionals so that they line up with reality, the instruction annotations (for things like arithmetic operations) will just work.

The tests have comments explaining the purpose, and a possible root cause and links to PRs in case it breaks in a certain way.


tests/pytests_launcher.py
```diff
@@ -32,6 +32,9 @@
 
 # We must call this to ensure the code coverage file writes get flushed
 # https://github.com/nedbat/coveragepy/issues/310
-coverage._atexit()
+if (cov := coverage.Coverage.current()) is not None:
+    cov.stop()
+    cov.save()
```
disconnect3d, line 37: so this instead of coverate._atexit? do we need to probably update the comment? idk

pwndbg/gdblib/disasm/arm.py
```diff
@@ -49,9 +49,9 @@
     ARM_INS_STRBT: 1,
     ARM_INS_STRHT: 2,
     ARM_INS_STRT: 4,
-    ARM_INS_STREXB: 1,
-    ARM_INS_STREXH: 2,
-    ARM_INS_STREX: 4,
+    # ARM_INS_STREXB: 1,
+    # ARM_INS_STREXH: 2,
+    # ARM_INS_STREX: 4,
```
disconnect3d: Why?

OBarronCS: Fix this in the latest commits and added a test for it - the operands that were accessed to get the source and destination for these were incorrect.

tests/pytests_launcher.py
OBarronCS: I updated the comment to be more clear 👍

disconnect3d:
Re-running all jobs here to see if recent fixes fixed this.

pwndbg/gdblib/disasm/__init__.py
```diff
@@ -416,7 +416,7 @@ def near(
         while insn is not None and len(insns) < instructions:
             if DEBUG_ENHANCEMENT:
                 print(f"Got instruction from cache, addr={cached:#x}")
-            if insn.jump_like and insn.split == SplitType.NO_SPLIT:
+            if insn.jump_like and insn.split == SplitType.NO_SPLIT and not insn.causes_branch_delay:
```
disconnect3d, line 419: Nitpick: https://docs.python.org/3/howto/enum.html#comparisons

We should probably compare enumerations via `is` instead of `==`? Unless `insn.split` is just an int. If it is an int, then `is` will work here due to small numbers being singletons in Python, but overall `x is y` for `x,  y: int` is not correct.

pwndbg/gdblib/disasm/__init__.py
```diff
@@ -454,8 +454,11 @@ def near(
                 emulated_arm_mode_cache[emu.pc] = emu.read_thumb_bit()
 
         # Handle visual splits in the disasm view
-        # The second check here handles instructions like x86 `REP` that repeat the instruction
-        if insn.has_jump_target or insn.next == insn.address:
+        # We create splits in 3 conditions:
```
disconnect3d, line 457: What exactly are splits? xD when we put an empty line to show there is a jump/nonlinear pc change or sth?

pwndbg/gdblib/disasm/arm.py
```diff
@@ -61,16 +64,42 @@
...
+def first_op_is_pc(i: PwndbgInstruction) -> bool:
+    return i.operands[0].reg == ARM_REG_PC
+
+
+def ops_contain_pc(i: PwndbgInstruction) -> bool:
+    for op in i.operands:
+        if op.type == CS_OP_REG:
+            if op.reg == ARM_REG_PC:
+                return True
+    return False
```
disconnect3d, line 90: The ifs can be merged? :)
```py
def ops_contain_pc(i: PwndbgInstruction) -> bool:
    for op in i.operands:
        if op.type == CS_OP_REG and op.reg == ARM_REG_PC:
            return True
    return False
```

pwndbg/gdblib/disasm/arm.py
```diff
@@ -61,16 +64,42 @@
...
+
+
+ARM_CAN_WRITE_TO_PC: Dict[int, Callable[[PwndbgInstruction], bool]] = {
+    ARM_INS_ADD: first_op_is_pc,
+    ARM_INS_SUB: first_op_is_pc,
+    ARM_INS_SUBS: first_op_is_pc,
+    ARM_INS_MOV: first_op_is_pc,
+    ARM_INS_LDR: first_op_is_pc,
+    ARM_INS_POP: ops_contain_pc,
+    ARM_INS_LDM: ops_contain_pc,
```
disconnect3d, line 100: Random thought: we could document instructions here somehow for easier understanding for future readers or link them? Idk, just loud thinking; we don't necessarily have to do it

pwndbg/gdblib/disasm/arm.py
```diff
@@ -117,6 +146,17 @@ def _set_annotation_string(self, instruction: PwndbgInstruction, emu: Emulator)
                 ARM_SINGLE_STORE_INSTRUCTIONS[instruction.id],
                 instruction.operands[1].str,
             )
+        elif instruction.id in ARM_EXCLUSIVE_STORE_INSTRUCTIONS:
```
disconnect3d, line 149: Btw, could be done in another PR, but shall we do sth like:
```
if (ins_id := instruction.id) in ARM_SINGLE_LOAD_INSTRUCTIONS: ...
elif ins_id in ...: ...
# etc
```
?

Also, that's probably wishful thinking but it would be nice to order those if/elifs via the frequency in which those instructions appear in disassembly :D

pwndbg/gdblib/disasm/instruction.py
```diff
@@ -248,6 +247,28 @@ def __init__(self, cs_insn: CsInsn | None) -> None:
...
+        Most of the type, we depend on Capstone groups to check for jump instructions,
+        but sometimes these are lacking, such as in the case of general-purpose instructions
+        where the PC is the destination register, such as Arm `add`, `sub`, `ldr`, and `pop` instructions.
+
+        In these cases, we want to forcefully state that this instruction mutates the PC, so we set this attribute to True.
+
+        This helps in two cases:
+        1. Disassembly splits
+        2. Instructions like `stepuntilasm` work better, as they detect these as branches to stop at.
+        """
```
disconnect3d, line 262: Btw I just noticed that we document fields under them? I think we should document them above (just make sure that its not the function docstring 😓)

pwndbg/gdblib/disasm/instruction.py
```diff
@@ -248,6 +247,28 @@ def __init__(self, cs_insn: CsInsn | None) -> None:
         None if we don't have a determination (most cases)
         """
 
+        self.declare_is_unconditional_jump: bool = False
+        """
+        This field is used to declare that this instruction is an unconditional jump.
+        Most of the type, we depend on Capstone groups to check for jump instructions,
```
disconnect3d, line 253: "Most of the type"? Shall this be one of:
- "Most of the time"
- "For most of instruction types"
- "For most of instructions"
?

disconnect3d (approved):

Issue #2232, opened 2024-06-13
Gdb tab auto complete broken by `import readline`


PandaWill:
### Description
Gdb tab autocomplete is broken. It appears that `environ["PWNLIB_NOTERM"]=1` is the reason why.

It starts working if I remove this line: https://github.com/pwndbg/pwndbg/blob/bf091df52781de81b7d4662fc65ce18d11aeecb8/gdbinit.py#L59

### Steps to reproduce
Open Gdb, type `st<TAB>`

We expect it to show the options of "start", "step", etc. We get a literal tab.

### My setup
pwndbg> version
Gdb 12.1
Python: 3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0]
Pwndbg: 2024.02.14 build: ffc09e57
Capstone: 5.0.1280
Unicorn: 2.0.1

B1N4RY-P4R45173:
I am still very new to this project. But I have noticed two things:
1) There are multiple commands starting with `st` so it cannot autocomplete. I am assuming u are referring to autocomplete suggestion?? I noticed that pressing the tab again displays all suggestions. Maybe a bug?? idk....

![image](https://github.com/pwndbg/pwndbg/assets/77494118/2b9674d1-77ea-4172-8d26-295e8b447173)

2) If you were actually referring to autocomplete then try `sto` and hitting tab key.  It should complete it to the word `stop`. Check if that works.

PandaWill:
Sorry, yes that's what I mean suggestions. I don't get the suggestions or autocomplete though.

Unless I comment out the `environ["PWNLIB_NOTERM"]=1` line (or set it to zero).

https://github.com/pwndbg/pwndbg/blob/bf091df52781de81b7d4662fc65ce18d11aeecb8/gdbinit.py#L59

B1N4RY-P4R45173:
I tried changing `PWNLIB_NOTERM` to `0`. It didn't seem to invoke any issues at the moment but according to

https://github.com/pwndbg/pwndbg/blob/bf091df52781de81b7d4662fc65ce18d11aeecb8/docs/env_vars.md?plain=1#L17

U might face issues with decencies that use pwntools. And did it work fine before https://github.com/pwndbg/pwndbg/commit/ffc09e57f3b870f99f52815b983ef678bf69636f ?

disconnect3d:
@PandaWill The autocompletion works for me no matter of the `PWNLIB_NOTERM` setting.

[...]

PR #2375, opened 2024-08-17
Test to ensure readline is not imported

PandaWill:
Importing CPython readline breaks GDB tab autocomplete. This tests ensures we never accidentally do that.

See [#2232](https://github.com/pwndbg/pwndbg/issues/2232) for details

Apologies it took me so long to get around to writing this test!


disconnect3d:
Thanks!

PR #2380, opened 2024-08-19
Added a configuration option to disable auto dependency updates

### LINKED ISSUE #2235: Add a configuration option to disable auto dependency updates

B1N4RY-P4R45173:
Closes #2235

### DIFF
--- a/gdbinit.py
+++ b/gdbinit.py
@@ -167,8 +167,9 @@ def main() -> None:
             print(f"Cannot find Pwndbg virtualenv directory: {venv_path}. Please re-run setup.sh")
             sys.stdout.flush()
             os._exit(1)
-
-        update_deps(src_root, venv_path)
+        no_auto_update = os.getenv("PWNDBG_NO_AUTOUPDATE")
+        if no_auto_update is None:
+            update_deps(src_root, venv_path)
         fixup_paths(src_root, venv_path)
 
     # Force UTF-8 encoding (to_string=True to skip output appearing to the user)
--- a/setup.sh
+++ b/setup.sh
@@ -226,4 +226,6 @@ if [ -z "$UPDATE_MODE" ]; then
     # Load Pwndbg into GDB on every launch.
     echo "source $PWD/gdbinit.py" >> ~/.gdbinit
     echo "[*] Added 'source $PWD/gdbinit.py' to ~/.gdbinit so that Pwndbg will be loaded on every launch of GDB."
+    echo "Please set the PWNDBG_NO_AUTOUPDATE environment variable to any value to disable the automatic updating of dependencies when Pwndbg is loaded."
+
 fi


B1N4RY-P4R45173:
Okay, it took me some time to realize why some of these tests were failing earlier. There were primarily two reasons:

1) The GitHub tests raise errors when interactive commands such as `read` are passed even if we pass a default answer with timeout.

2) Setting up permanent env variables also seems to be a tricky task.

I was trying to set a permanent env variable to store the user's choice about auto updates, so we can check it from `gdbinit.py` and run appropriate command if update is needed.

One alternative could be to add a file marking the user's choice only if he chooses to turn off the auto update of dependencies. And when `gdbinit.py` is run, if the file (the one used for marking) exists, we can skip or else we need to update the dependencies

B1N4RY-P4R45173:
I think we might also need slight changes in the workflow files of these tests and the docker files

B1N4RY-P4R45173:
@disconnect3d I need some help with this

gdbinit.py
```diff
@@ -168,7 +168,12 @@ def main() -> None:
             sys.stdout.flush()
             os._exit(1)
 
-        update_deps(src_root, venv_path)
+        # Check for the environment variable to disable auto-update
+        auto_update_disabled = os.getenv("PWNDBG_AUTO_UPDATE", "true").lower() == "false"
```
disconnect3d: Let's change this to `PWNDBG_NO_AUTOUPDATE` and if it is defined with any value, skip updating deps.

setup.sh
```diff
@@ -200,8 +200,59 @@ ${PYTHON} -m venv -- ${PWNDBG_VENV_PATH}
...
+            ;;
+        *)
+            echo "Unsupported shell. You will need to manually add your shell config file."
+            return 1
+            ;;
+    esac
+}
+
+# Function to persist the environment variable in the appropriate shell configuration file
+persist_env_var() {
```
disconnect3d: We dont't want any of this. Let's just inform user on setup they can disable auto update checks with a given envvar and let the user to add it or not by themselves.

disconnect3d: I.e. we don't want to handle sh/bash/zsh/fish/bazilion other shells; the users can do it on their own

disconnect3d (changes requested):
As discussed on Discord, lets simplify this

gdbinit.py
B1N4RY-P4R45173: Okay I will change it asap.

setup.sh
B1N4RY-P4R45173: Sure I got it. Will complete it by today.

disconnect3d (approved):

disconnect3d:
Thanks!

PR #2382, opened 2024-08-19
Port Pwndbg to LLDB

mbrla0:
This PR is part of the GSoC LLDB port.

What it says on the tin. This PR is the first in a series of three PRs that bring upstream Pwndbg up to date with the LLDB port. Ideally, we'd merge all of these in smaller chunks, but work ended up progressing at a much faster pace on the branch than got upstreamed. So, as previously discussed in the Pwndbg Discord, we'll be doing this in larger chunks.

This PR introduces expansions to the Debugger-agnostic API relating to register writes, breakpoints and watchpoints, debugger-backed disassembly, source and line queries in a frame, reading and writing of memory in the process being debugged, memory searches, gdb-remote packets, downloading of remote files, creation of debugger values (ala `gdb.Value`), symbol and type queries, module queries, and more.

This PR also introduces the Pwndbg CLI, under `pwndbg/dbg/lldb/repl`. The native LLDB CLI is [quite](https://discourse.llvm.org/t/understanding-debugger-launch-events-sequence/39717/2) [problematic](https://github.com/llvm/llvm-project/blob/3b5b5c1ec4a3095ab096dd780e84d7ab81f3d7ff/lldb/source/Utility/Broadcaster.cpp#L409) for [plugins](https://github.com/llvm/llvm-project/blob/3b5b5c1ec4a3095ab096dd780e84d7ab81f3d7ff/lldb/source/Commands/CommandObjectCommands.cpp#L439) [like](https://github.com/llvm/llvm-project/blob/3b5b5c1ec4a3095ab096dd780e84d7ab81f3d7ff/lldb/source/Commands/CommandObjectCommands.cpp#L439) Pwndbg, in that it does not allow plugins loaded with `command script import` to really observe pretty much any of the events in the process being debugged, or control its execution in any truly useful way. So, to get around that, Pwndbg in LLDB uses LLDB mostly as a library, and drives program execution itself. Stuff like readline, command completions, process I/O and lifetime management are all done by us. This doesn't mean we don't accept regular LLDB commands, however. Most LLDB commands will work exactly the same in `pwndbg-lldb` as they do in LLDB proper, with a few exceptions in process and target creation/attach commands like `process launch` and `target create`, as we need to implement them ourselves. `pwndbg/dbg/lldb/repl/__init__.py` goes into more detail about why we need to do this to have an event system at all in LLDB. This also means that LLDB Pwndbg has a new dependency: `gnureadline`.

This PR also _vastly_ expands `aglib`. It now covers much of what `gdblib` does, and has replaced it for modules like `heap`, `disasm`, and `dynamic`. I hope that, after this gets merged, new code targets it and `dbg`, instead of `gdblib`, whenever possible, and we begin to phase out the user-facing parts of modules in `gdblib` as their functionality gets completely ported to `aglib`, and the remaining commands get ported to `aglib` and `dbg`. Ideally, new features should be split into code can be shared between LLDB and GDB, which goes in `aglib`, and code that is entirely debugger-specific, which goes in `dbg`.

Because `pwndbg.gdblib.{disasm,dynamic,heap}` got moved to `pwndbg.aglib`, this PR also includes updates to all the references to these modules. I hate that this had to be included, but splitting those would require even more work and lots of temporary additions to be removed by a later PR to avoid breaking Pwndbg.

This PR also ports `pwndbg.arguments`, `pwndbg.auxv`, `pwndbg.chain`, `pwndbg.color`, `pwndbg.enhance`, `pwndbg.glibc`, `pwndbg.hexdump`, parts of `pwndbg.lib`, and `pwndbg.search` to the new debugger-agnostic interfaces. These are needed fairly widely across Pwndbg, and not including them in this PR would break it. A few commands have also been changed to keep them from breaking.

As for what's missing in the port, `dbg` is still missing a way to do process execution control[^1], and there are a few modules still missing from `aglib` that are in `gdblib`: `argv`, `dt`, `got`, `heap_tracking`, `kernel`, `next`, `onegadget`, and `shellcode`. These should all be fairly easy to port in the future, since most of them are only blocked on us missing process execution control in `dbg`.

This PR is also missing both the Nix entries for LLDB, and the command ports. These are ready to go, but I didn't want this PR to be even longer than it already is, and splitting those off is actually doable.

[^1]: Stuff like resuming until a breakpoint is hit and single stepping. It's currently missing from `dbg` mostly due to the fact that it has some very odd design requirements stemming from the way we handle process lifetimes in LLDB.


pwndbg/dbg/lldb/__init__.py
```diff
@@ -304,6 +698,314 @@ def vmmap(self) -> pwndbg.dbg_mod.MemoryMap:
...
+        # At this point, we're in a bit of a pickle. LLDB doesn't give us enough
+        # information to find out what the last address it can read from is. For
+        # all we know, it could be any address in the range (address, address+size),
+        # so we have to get creative.
+        #
+        # First, try to derive that information from the mmap.
+        first_page = None
+        last_page = None
+        vmmap_size = 0
+        for page in self.vmmap().ranges():
```
patryk4815: lldb support `GetMemoryRegionContainingAddress` do you think it is worth using it here?

pwndbg/dbg/lldb/__init__.py
```diff
@@ -314,6 +1016,114 @@ def symbol_name_at_address(self, address: int) -> str | None:
...
+
+            import pwndbg.aglib.memory
+
+            if pwndbg.aglib.memory.peek(candidate):
+                # Take a guess that we hit the right TLS block and return what
+                # we got.
+                return candidate
+
+        print(
+            "warning: symbol '{sym.GetName()}' might be a TLS symbol, but it could not be resolved"
```
patryk4815:
```suggestion
            f"warning: symbol '{sym.GetName()}' might be a TLS symbol, but it could not be resolved"
```

pwndbg/dbg/lldb/__init__.py
mbrla0: `GetMemoryRegionContainingAddress` afaik doesn't give us the index of the memory region in the `SBMemoryRegionInfoList`, just the memory region itself. Since we have to scan a range of memory regions starting at the one where the address is, it's not that useful to us here.

pwndbg/aglib/file.py
```diff
@@ -0,0 +1,178 @@
...
+            return ""
+
+    #
+    # Hurray unexposed packets!
+    #
+    # The 'vFile:readlink:' packet does exactly what it sounds like,
+    # but there is no API exposed to do this and there is also no
+    # command exposed... so we have to send the packet manually.
+    #
+    cmd = "vFile:readlink:%s"
```
patryk4815, line 138: this code should go to dbg/gdb.py probably.
 Or it is working on both lldb/gdb?

mbrla0: LLDB can handle this path gracefully. It doesn't seem to implement that packet, but it fails in a way we can handle. If it ever implements it, we get support for it almost for free[^1].

[^1]: I say almost because we'd have to change a few of the strings we're expecting in the following code, and fix the formatting for `pwndbg.dbg.Process.send_remote()`, but it should work. This is mostly due to the fact that this is the only place that I know of in Pwndbg that actually relies on the format of the response being consistent.

pwndbg/aglib/proc.py
```diff
@@ -0,0 +1,186 @@
...
+            `pwndbg.gdblib.file.get_proc_exe_file()`
+            (This will call `pwndbg.gdblib.file.get_file(pwndbg.gdblib.proc.exe, try_local_path=True)`)
+        """
+
+        return pwndbg.dbg.selected_inferior().main_module_name()
+
+    @property
+    @pwndbg.lib.cache.cache_until("start", "stop")
+    def binary_base_addr(self) -> int:
+        return self.binary_vmmap[0].start
```
patryk4815, line 97: Should we return here `None` if index `0` not exists?

mbrla0: Well, while I think it would make sense to do that from a general API design standpoint, I don't think it's a great idea to do that in this PR specifically.

Here's the thing. That piece of code was lifted straight off `pwndbg/gdblib/proc.py`, so that's a potential bug that we share with `gdblib`, meaning that all code that currently exists in Pwndbg, including the one being ported to `aglib` in this PR, will probably not behave gracefully at all if we returned `None` there for that case, anyway.

I agree that we should fix that, but having the return `None` case would kick the can of gracefully handling the missing vmmap case down to every part of Pwndbg that calls into that function, and so I don't think this PR is the place to do it.

pwndbg/commands/heap.py
```diff
@@ -7,60 +7,58 @@
...
+import pwndbg.aglib.heap
+import pwndbg.aglib.memory
+import pwndbg.aglib.proc
+import pwndbg.aglib.typeinfo
+import pwndbg.aglib.vmmap
 import pwndbg.chain
 import pwndbg.color.context as C
 import pwndbg.color.memory as M
 import pwndbg.commands
-import pwndbg.gdblib.heap
```
patryk4815, line 18: There are more `pwndbg.gdblib` in this file that can be refactored

mbrla0: Yes, and those are coming in the next PR. This PR limits itself to changing only as much as needed in the commands to keep Pwndbg from breaking, because I didn't want the scope to be even larger.

pwndbg/glibc.py
```diff
@@ -17,18 +17,19 @@
...
+import pwndbg.aglib.file
+import pwndbg.aglib.memory
+import pwndbg.aglib.proc
 import pwndbg.lib.cache
 import pwndbg.lib.config
 import pwndbg.search
 from pwndbg.color import message
 
+if pwndbg.dbg.is_gdblib_available():
+    import pwndbg.aglib.heap
```
patryk4815: is there reason why `import pwndbg.aglib.heap` is in this if?

mbrla0: Oh, yeah, that's something I'd overlooked. That module was ported before `pwndbg.gdblib.heap` was ported to `aglib`, and I forgot to change it after the port. Should be good now.

pwndbg/aglib/arch.py
```diff
@@ -1,8 +1,18 @@
...
+
+import pwnlib
+
 import pwndbg
 from pwndbg.lib.arch import Arch
 
+# We will optimize this module in the future, by having it work in the same
+# way the `gdblib` version of it works, and that will come at the same
+# time this module gets expanded to have the full feature set of its `gdlib`
+# coutnerpart. For now, though, this should be good enough.
```
disconnect3d:
```suggestion
# counterpart. For now, though, this should be good enough.
```

pwndbg/aglib/arch.py
```diff
@@ -1,8 +1,18 @@
...
+from typing import Literal
+
+import pwnlib
+
 import pwndbg
 from pwndbg.lib.arch import Arch
 
+# We will optimize this module in the future, by having it work in the same
+# way the `gdblib` version of it works, and that will come at the same
+# time this module gets expanded to have the full feature set of its `gdlib`
```
disconnect3d:
```suggestion
# time this module gets expanded to have the full feature set of its `gdblib`
```

mbrla0: Actually, it already works like the `gdblib` version. That comment is outdated. I'll just take it out.

disconnect3d:
@mbrla0 can you resolve conflicts?

mbrla0:
> @mbrla0 can you resolve conflicts?

I'm on it. Just making sure nothing broke.

pwndbg/aglib/dynamic.py
```diff
@@ -40,14 +37,7 @@ def _r_debug():
...
 
-    try:
-        address = gdb.execute("output/x &_r_debug", to_string=True)
-        address = int(address, 0)
-
-        return address
-    except gdb.error:
-        # Symbol is most likely unavailable.
-        return None
+    return pwndbg.dbg.selected_inferior().symbol_address_from_name("_r_debug")
```
disconnect3d, line 40: As I said on Discord, we want to optimize this in the future so that one doesn't have to call selected inferior

pwndbg/aglib/dynamic.py
```diff
@@ -68,7 +58,7 @@ def is_dynamic() -> bool:
 R_DEBUG_LINK_MAP_CHANGED_LISTENERS: Set[Callable[..., Any]] = set()
 
 
-class RDebugLinkMapChangedHook(pwndbg.gdblib.bpoint.BreakpointEvent):
+def r_debug_link_map_changed_hook() -> Callable[[pwndbg.dbg_mod.StopPoint], bool]:
```
disconnect3d, line 61: Honestly, I prefer BreakpointEvent from a StopPoint but w/e for now

pwndbg/aglib/heap/heap.py
```diff
@@ -2,13 +2,17 @@
...
     """Heap abstraction layer."""
 
-    def breakpoint(self, event: str) -> gdb.Breakpoint:
+    # This function isn't actually implemented anywhere. It originally returned
+    # `gdb.Breakpoint`, but, in order to facilitate the port to aglib, that
+    # type association was removed. It should be put back as soon as the
+    # Debugger-agnostic API gains the ability to set breakpoints.
+    #
+    # TODO: Change `Any` to the Debugger-agnostic breakpoint type when it gets created
+    def breakpoint(self, event: str) -> Any:
```
disconnect3d, line 15: Shall we remove this func if its not implemented anywhere? What is its purpose? Why does a heap MemoryAllocator class needs a bp function?

pwndbg/aglib/memory.py
```diff
@@ -0,0 +1,359 @@
...
+import pwndbg.lib.memory
+from pwndbg.lib.memory import PAGE_SIZE
+
+GdbDict = Dict[str, Union["GdbDict", int]]
+
+
+MMAP_MIN_ADDR = 0x8000
+
+
+def read(addr: int, count: int, partial: bool = False) -> bytearray:
```
disconnect3d, line 19: This adds an unnecessary call atm with no benefit.

Ideally this should just be `read = gdb_or_lldb_read_memory_func` since their API is the same anyway?

pwndbg/aglib/nearpc.py
```diff
@@ -0,0 +1,355 @@
...
+import pwndbg.ui
+from pwndbg.aglib.disasm.instruction import SplitType
+from pwndbg.color import ColorConfig
+from pwndbg.color import ColorParamSpec
+from pwndbg.color import message
+
+# We prefix all of our configuration options with a special `aglib-` prefix if
+# we detect that gdblib is present. Pwndbg doesn't allow multiple settings under
+# the same name, so we have to add a unique prefix until we move all of Pwndbg
+# to this module, rather than the gdblib one.
```
disconnect3d, line 29: Not sure I like it :/

pwndbg/aglib/nearpc.py
```diff
@@ -0,0 +1,355 @@
...
+#
+# TODO: Port the rest of Pwndbg to this module and get rid of `pwndbg.gdblib.nearpc`.
```
disconnect3d, line 31: We need to make sure that all TODOs like this are actually resolved later on

pwndbg/aglib/regs.py
```diff
@@ -0,0 +1,296 @@
...
+) -> pwndbg.dbg_mod.Value | None:
+    if frame is None:
+        frame = pwndbg.dbg.selected_frame()
+    assert (
+        frame is not None
+    ), "pwndbg.dbg.selected_frame() should never return None when marked with @OnlyWhenRunning"
+
+    regs = regs_in_frame(frame)
+
+    return regs.by_name(name) or regs.by_name(name.upper())
```
disconnect3d, line 50: If register value is 0, what does the `bool(pwndbg.dbg_mod.Value)` return?

If it is false, then we are effectively fetching the same value again unnecessarily.

https://github.com/pwndbg/pwndbg/blob/27ccbbaceb5b559b3e07bddfc9616b7da50bab07/pwndbg/gdblib/regs.py#L35-L42

pwndbg/aglib/nearpc.py
mbrla0: I don't like it, either, but is there a better way to resolve that before we move everything over from `pwndbg.gdblib.nearpc`?

pwndbg/aglib/heap/heap.py
mbrla0: I don't know. I don't really know what the intent was behind that method, originally. I don't think there's any reason for us to keep it, personally.

pwndbg/aglib/regs.py
```diff
@@ -0,0 +1,296 @@
...
+import pwndbg.lib.cache
+from pwndbg.dbg import EventType
+from pwndbg.lib.regs import BitFlags
+from pwndbg.lib.regs import RegisterSet
+from pwndbg.lib.regs import reg_sets
+
+
+@pwndbg.lib.cache.cache_until("stop", "prompt")
+def regs_in_frame(frame: pwndbg.dbg_mod.Frame) -> pwndbg.dbg_mod.Registers:
+    return frame.regs()
```
disconnect3d, line 35: Not sure this needs to be cached but maybe

pwndbg/aglib/regs.py
```diff
@@ -0,0 +1,296 @@
...
+    match = re.search(rf'{name.split("_")[0]}=\s+([\da-fA-F]+)\s+([\da-fA-F]+)', out)
+
+    if match:
+        base = int(match.group(1), 16)
+        limit = int(match.group(2), 16)
+
+        if name.endswith("LIMIT"):
+            return limit
+        else:
+            return base
```
disconnect3d, line 66: Random comment: we should probably add some comment here to explain wth is that LIMIT stuff

pwndbg/aglib/regs.py
```diff
@@ -0,0 +1,296 @@
...
+gsbase: int
+current: RegisterSet
+fix: Callable[[str], str]
+items: Callable[[], Generator[Tuple[str, Any], None, None]]
+previous: Dict[str, int]
+last: Dict[str, int]
+pc: int | None
+
+
+class module(ModuleType):
```
disconnect3d, line 96: (Random comment: This whole module class design with global attrs is weird. Should probably move off from that in future)

pwndbg/aglib/remote.py
```diff
@@ -0,0 +1,54 @@
+"""
+Information about whether the debuggee is local (under GDB) or remote
```
disconnect3d, line 2: Shall we say, under GDB or LLDB?

pwndbg/aglib/regs.py
mbrla0: `pwndbg.dbg_mod.Value` implements neither `__bool__` or `__len__`, so, `bool()` of it [will be `True` for all valid instances](https://docs.python.org/3.8/reference/datamodel.html#object.__bool__).

pwndbg/aglib/memory.py
mbrla0: It would probably have to look something like this, but it would work:

```py
read: Callable[..., bytearray] = None

@pwndbg.dbg.event_handler(EventType.START)
def on_start():
    global read
    read = pwndbg.dbg.selected_inferior().read_memory

@pwndbg.dbg.event_handler(EventType.EXIT)
def on_exit():
    global read
    read = None
```

pwndbg/dbg/gdb.py
```diff
@@ -112,17 +87,79 @@ def __init__(self, inner: gdb.Frame):
 
     @override
     def evaluate_expression(self, expression: str) -> pwndbg.dbg_mod.Value:
+        from pwndbg.gdblib.scheduler import lock_scheduler
+
+        with lock_scheduler():
```
disconnect3d, line 92: The previous code doesn't seem to have this, why are we adding it here?

pwndbg/aglib/remote.py
mbrla0: Yes. There are probably a few of these comments remaining in `aglib` that reference GDB specifically, but that should specify both of them, or just use the more generic "debugger".

pwndbg/dbg/lldb/repl/__init__.py
```diff
@@ -0,0 +1,683 @@
...
+process_launch_ap.add_argument("-c", "--shell")
+process_launch_ap.add_argument("-e", "--stderr")
+process_launch_ap.add_argument("-i", "--stdin")
+process_launch_ap.add_argument("-k", "--structured-data-key")
+process_launch_ap.add_argument("-n", "--no-stdio")
+process_launch_ap.add_argument("-o", "--stdout")
+process_launch_ap.add_argument("-s", "--stop-at-entry", action="store_true")
+process_launch_ap.add_argument("-t", "--tty")
+process_launch_ap.add_argument("-v", "--structured-data-value")
+process_launch_ap.add_argument("-w", "--working-dir")
```
disconnect3d, line 469: All above could be a for loop but w/e

pwndbg/dbg/lldb/repl/io.py
```diff
@@ -0,0 +1,495 @@
...
+        pass
+
+    @override
+    def on_process_start(self, proc: lldb.SBProcess) -> None:
+        # Once we have `pwndbg.gdblib.shellcode` functioning, we could try to
+        # attempt a "coup" of the controlling TTY for the process, here, so we
+        # get to have the PTY we set up in this class as the main controller for
+        # this process.
+        #
+        # TODO: Replace controlling PTY of the proecss once it is set up.
```
disconnect3d:
```suggestion
        # TODO: Replace controlling PTY of the process once it is set up.
```

pwndbg/lib/cache.py
```diff
@@ -132,7 +145,7 @@ def inner(func: Callable[P, T]) -> Callable[P, T]:
 
         @wraps(func)
         def decorator(*a: P.args, **kw: P.kwargs) -> T:
-            if IS_CACHING:
+            if IS_CACHING and not any((IS_CACHING_DISABLED_FOR[e] for e in event_names)):
```
disconnect3d, line 148: We really don't want this check here; if we need this ability on the long run, this should be changed somehow

disconnect3d (commented):
LGTM. Added some comments/questions here and there.

We should wait with merging it till we hit a new release (I'll try to do it tomorrow).

pwndbg/dbg/gdb.py
mbrla0: Well, mostly so we can also cover the cases that previously needed `parse_and_eval_with_scheduler_locking`. I think there's probably a place for us to maybe add a `allow_resume_threads` keyword argument, but LLDB will most likely just ignore that, seeing as things like the Clang-based expression evaluator just uses a [thread plan on the thread that owns the frame](https://github.com/llvm/llvm-project/blob/1601879f5d690595889a5537c0049b62df4657c7/lldb/source/Expression/ExpressionParser.cpp#L62). Maybe that argument could be an optional optimization? As in: allow other threads to resume if it's slightly faster for Pwndbg, but with no guarantees that the debugger and the `dbg` implementation will actually honor it.

pwndbg/lib/cache.py
mbrla0: I wonder where else we could put it, or what other way we could achieve this, since this is fairly important, seeing as without it, caches like `thread` on LLDB would be the same as `forever`.

pwndbg/aglib/dynamic.py
mbrla0: The name `StopPoint` in the Debugger-agnostic API comes from the fact that it's generic over both breakpoints and watchpoints. We could just use the name `Breakpoint` for both, though I do lean towards not overloading terms, but ultimately this is something that can be changed fairly easily

pwndbg/aglib/dynamic.py
mbrla0: Maybe something like `pwndbg.dbg.inferior`? We always know when a new valid inferior gets created and when it gets destroyed, anyway, in both LLDB and GDB, so that wouldn't be much of an issue.

Though the same can't be said for either `selected_thread()` or `selected_frame()` under LLDB. They might be harder to do correctly as values, given that sometimes - but not always - we have to pull them from the current inferior - see `_any_thread` and `_any_bottommost_frame`.

pwndbg/aglib/regs.py
mbrla0: Caching that lets us avoid calling `SBFrame::GetRegisters` (which pulls in all register values at once) every time we want the value of a single register.

disconnect3d:
@mbrla0 damn, this now has a few conflicts that need to be resolved :/

mbrla0:
@disconnect3d Resolved

pwndbg/aglib/heap/heap.py
disconnect3d: Moved to https://github.com/pwndbg/pwndbg/issues/2404

pwndbg/dbg/gdb.py
disconnect3d: My point is, scheduler locking can cause deadlocks sometimes, so its better if the user control it and we only use it if its really needed.

pwndbg/lib/cache.py
disconnect3d: I mean... probably change the impl for those that should just call the result instead of having an additional if condition here?

pwndbg/dbg/gdb.py
mbrla0: Okay, fair enough. I wasn't aware of the issues with deadlocking. I'll add a PR to have this be conditionally enabled. We'll have to expose scheduler locking in the Debugger-agnostic API, but it's worth it, given the possible consequences of not doing it.

PR #2384, opened 2024-08-20
Heap free bins indication in telescope

OBarronCS:
This PR adds an indication that a pointer is referencing a free chunk in any heap pointers displayed in any use of the `chain` function (used in telescope, in printing registers (`regs`), and in annotations).

image

Due to possible performance concerns, it is gated behind a config setting.

### DIFF
--- a/pwndbg/chain.py
+++ b/pwndbg/chain.py
@@ -110,6 +110,7 @@ def format(
     hard_end: int = 0,
     safe_linking: bool = False,
     enhance_string_len: int | None = None,
+    display_heap_free_bins: bool = True
 ) -> str:
     """
     Recursively dereferences an address into string representation, or convert the list representation
@@ -143,7 +144,7 @@ def format(
     arrow_right = c.arrow(f" {config_arrow_right} ")
 
     # Colorize the chain
-    rest = [M.get_address_and_symbol(link) for link in chain]
+    rest = [M.get_address_and_symbol(link, display_heap_free_bins) for link in chain]
 
     # If the dereference limit is zero, skip any enhancements.
     if limit == 0:
--- a/pwndbg/chain.py
+++ b/pwndbg/chain.py
@@ -110,7 +110,7 @@ def format(
     hard_end: int = 0,
     safe_linking: bool = False,
     enhance_string_len: int | None = None,
-    display_heap_free_bins: bool = True
+    display_heap_free_bins: bool = True,
 ) -> str:
     """
     Recursively dereferences an address into string representation, or convert the list representation
--- a/pwndbg/color/memory.py
+++ b/pwndbg/color/memory.py
@@ -30,7 +30,7 @@ c = ColorConfig(
 pwndbg.config.add_param(
     "telescope-free-heap-bins",
     False,
-    "wheteher or not to resolve heap addresses to bin names while telescoping",
+    "whether or not to resolve heap addresses to bin names while telescoping",
 )
 
 
--- a/pwndbg/color/memory.py
+++ b/pwndbg/color/memory.py
@@ -34,14 +34,14 @@ pwndbg.config.add_param(
 )
 
 
-def get_address_and_symbol(address: int) -> str:
+def get_address_and_symbol(address: int, display_heap_free_bins=True) -> str:
     """
     Convert and colorize address 0x7ffff7fcecd0 to string `0x7ffff7fcecd0 (_dl_fini)`
     If no symbol exists for the address, return colorized address
     """
 
     # Attempt to resolve it as a free chunk
-    if pwndbg.config.telescope_free_heap_bins:
+    if display_heap_free_bins and pwndbg.config.telescope_free_heap_bins:
         if (page := pwndbg.gdblib.vmmap.find(address)) is not None and "[heap" in page.objfile:
             if (
                 free_bin_name := pwndbg.gdblib.heap.heap_freebin_address_lookup(address)
--- a/pwndbg/color/memory.py
+++ b/pwndbg/color/memory.py
@@ -46,7 +46,7 @@ def get_address_and_symbol(address: int) -> str:
             if (
                 free_bin_name := pwndbg.gdblib.heap.heap_freebin_address_lookup(address)
             ) is not None:
-                return get(address, f"{address} ({free_bin_name} free chunk)")
+                return get(address, f"{hex(address)} ({free_bin_name} free chunk)")
 
     symbol = pwndbg.gdblib.symbol.get(address) or None
     if symbol:
--- a/pwndbg/color/memory.py
+++ b/pwndbg/color/memory.py
@@ -34,19 +34,19 @@ pwndbg.config.add_param(
 )
 
 
-
-
 def get_address_and_symbol(address: int) -> str:
     """
     Convert and colorize address 0x7ffff7fcecd0 to string `0x7ffff7fcecd0 (_dl_fini)`
     If no symbol exists for the address, return colorized address
     """
 
-    # Attempt resolve it as a free chunk
+    # Attempt to resolve it as a free chunk
     if pwndbg.config.telescope_free_heap_bins:
         if (page := pwndbg.gdblib.vmmap.find(address)) is not None and "[heap" in page.objfile:
-            if (free_bin_name := pwndbg.gdblib.heap.heap_freebin_address_lookup(address)) is not None:
-                return get(address,f"{address} ({free_bin_name} free chunk)")
+            if (
+                free_bin_name := pwndbg.gdblib.heap.heap_freebin_address_lookup(address)
+            ) is not None:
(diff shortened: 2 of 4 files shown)


OBarronCS:
Perhaps there is a better place for the lookup function to exist - presently, I placed it in `gdblib/heap/__init__.py`

pwndbg/color/memory.py
```diff
@@ -26,12 +27,27 @@
...
-def get_address_and_symbol(address: int) -> str:
+def get_address_and_symbol(address: int, display_heap_free_bins=True) -> str:
     """
     Convert and colorize address 0x7ffff7fcecd0 to string `0x7ffff7fcecd0 (_dl_fini)`
     If no symbol exists for the address, return colorized address
     """
+
+    # Attempt to resolve it as a free chunk
+    if display_heap_free_bins and pwndbg.config.telescope_free_heap_bins:
+        if (page := pwndbg.gdblib.vmmap.find(address)) is not None and "[heap" in page.objfile:
```
CptGibbon, line 45: I understand this is for efficiency, i.e. don't call `heap_freebin_address_lookup()` if `address` isn't in a mapping labelled "[heap]".

As I recall, pwndbg only marks the GLIBC sbrk heap with "[heap]", so `heap_freebin_address_lookup()` may not run when there are other heaps present.

pwndbg/gdblib/heap/__init__.py
```diff
@@ -124,3 +124,48 @@ def resolve_heap(is_first_run: bool = False) -> None:
             )
     else:
         current = pwndbg.gdblib.heap.ptmalloc.DebugSymsHeap()
+
+
+def heap_freebin_address_lookup(addr: int) -> str | None:
```
CptGibbon, line 129: Since this is GLIBC malloc-specific, perhaps it could reside in "pwndbg/gdblib/heap/ptmalloc.py"

pwndbg/gdblib/heap/__init__.py
```diff
@@ -124,3 +124,48 @@ def resolve_heap(is_first_run: bool = False) -> None:
...
+    except Exception:
+        return None
+
+    if heap.arena is None:
+        return None
+
+    for chunk in heap:
+        if addr in chunk:
+            arena = chunk.arena
+            if arena:
```
CptGibbon, line 151: I don't think `chunk.arena` will ever be different to `heap.arena` under normal circumstances.

In fact I believe `Chunk` does `self.heap.arena` internally.

CptGibbon (changes requested):
Cool feature, thanks for the work 🙏
I've commented a couple of things to consider.

pwndbg/gdblib/heap/__init__.py
```diff
@@ -124,3 +124,48 @@ def resolve_heap(is_first_run: bool = False) -> None:
...
+                if chunk.is_top_chunk:
+                    return "top chunk"
+
+            if not chunk.is_top_chunk and arena:
+                bins_list = [
+                    allocator.fastbins(arena.address),
```
CptGibbon, line 157: I know this is duplicated from the `vis-heap-chunks` command, but I feel like Martin Fowler/Don Roberts rule of three applies.

pwndbg/color/memory.py
disconnect3d: "when there are other heaps present" - in other words, iirc, this would not work with multiple heaps that can be there when multiple threads are in play.

pwndbg/chain.py
```diff
@@ -143,7 +144,7 @@ def format(
     arrow_right = c.arrow(f" {config_arrow_right} ")
 
     # Colorize the chain
-    rest = [M.get_address_and_symbol(link) for link in chain]
+    rest = [M.get_address_and_symbol(link, display_heap_free_bins) for link in chain]
```
disconnect3d, line 147: Should we maybe handle such custom detecting/resolving of "nice things to enhance" during telescope/chain resolution in a more generic manner?

I mean, lets assume we want to enhance "errno values" one day, would we also add an additional argument here? :/

Perhaps there should be a way to register/enable/disable such different/additional handlers for formatting/telescope stuff? Not sure

disconnect3d: (I'm okay with merging this as is, as long as we resolve other comments, but perhaps this is a good time to do it properly to handle future stuff now?)

pwndbg/color/memory.py
CptGibbon: > this would not work with multiple heaps that can be there when multiple threads are in play.

Exactly.

GLIBC malloc may also create multiple heaps in a single-threaded environment if sbrk fails due to memory pressure.

OBarronCS:
Thanks for the review and suggestions - I'll be making these changes 😊

What are your thoughts on potentially adding a number that indicates the index in the corresponding free bin that a pointer is in? 0 for the head of the linked list, 1 for the next, ... etc etc

pwndbg/chain.py
OBarronCS: Yes - that's a good point. While making this function parameter, I also had the thought "well, what if one day we want to display floating point numbers in this enhance function too"... All the different arguments will add up, especially since we need to pass them through two functions.

Ideas:
- Some sort of detection for each number - check if it's in the heap area, or check if it resolves to a floating point number that's reasonable (which misses cases, and for things like errno, this gets harder as we need to remember we are from the `RAX` register)
- Pass all the additional info in a single struct instead of separate parameters
- set_telescope_handler(ENUM, True/False) - enum to set a value globally that telescope checks?
   -  or maybe a function callback instead of enum - it takes a list of functions that are chained like a plugin system, all are called to possibly enhance the value (or return None if deemed irrelevent)

disconnect3d: > this gets harder as we need to remember we are from the RAX register

We could maybe pass in some context around (?) tho I am not sure how practical that is in general. I am not 100% bought or enhancing the errno yet, but if we find a nice solution we could do that.

> set_telescope_handler(ENUM, True/False) - enum to set a value globally that telescope checks?

Could also be sth like `set telescope-enhance x,y,z` or sth.

OBarronCS: `set telescope-enhance free-bins` and `set telescope-enhance free-bins floating errno` to enable globally sounds good to me. We still need the caller to the telescope function to indicate that "in this specific case, don't do free-bins (regardless of global setting)".

Perhaps a context object would be good - a optionally `None` parameter that provides additional data relevant to the current telescope. Could contain "register: RAX", or "heap: True", "floating: True". This would avoid having to add a new parameter each time we want some new logic. I'll consider this and a `enhancer callback function` option and see which one makes most sense.

CptGibbon:
> What are your thoughts on potentially adding a number that indicates the index in the corresponding free bin that a pointer is in? 0 for the head of the linked list, 1 for the next, ... etc etc

I like it in theory, but it could be confusing in practice.

e.g. malloc searches singly-linked lists forward, but doubly-linked lists "backward".
What is the first chunk in a doubly-linked list? The one at `fd` or `bk`?

If the numbering is explicit and consistent in its direction then it could be useful e.g. always follow `fd` and ensure the user knows this.

disconnect3d:
@OBarronCS any update here? (there are also merge conflicts :<)

k4lizen:
> I like it in theory, but it could be confusing in practice.

> e.g. malloc searches singly-linked lists forward, but doubly-linked lists "backward".
> What is the first chunk in a doubly-linked list? The one at fd or bk?

> If the numbering is explicit and consistent in its direction then it could be useful e.g. always follow fd and ensure the user knows this.

image

fwiw there is already annotation (the `unsortedbin[all][0]` stuff) that follows a clear rule (fd is 0), maybe that could be used?

CptGibbon:
> fwiw there is already annotation (the `unsortedbin[all][0]` stuff) that follows a clear rule (fd is 0), maybe that could be used?

Yes that seems sensible 👍

disconnect3d:
What's the current state of it? Seems there are conflicts that need to be resolved.

I'm all good with this feature as long as it doesn't make some big perf hit (and it probably doesnt?)

pwndbg/chain.py
```diff
@@ -143,7 +144,7 @@ def format(
...
 
     # If the dereference limit is zero, skip any enhancements.
     if limit == 0:
```
disconnect3d, line 150: Do `rest` have only 1 item at this point or can it have more? If it has more, we should then compute the `rest = [...]` after this check and here just do `return something(chain[0])` ?

OBarronCS: That makes sense to me - would make this function a bit more readable.

OBarronCS:
I'm going to close this for now - thanks for all the comments. To implement this in the future, it would be great for the fix to involve determining a good way to pass or detect more context and configuration into the chain function so different types of values can be enhanced in a more fine-grained manner.

Issue #2365, opened 2024-08-15
 xinfo fails on remote qemu


OBarronCS:
Using the `xinfo` command while debugging a qemu process causes a crash. While determining information on the page in memory, it attempts to read a file which does not exist on the local machine.

![image](https://github.com/user-attachments/assets/b7b3aa56-d4d4-44af-8447-02f8380e1819)

To reproduce, connect to qemu and use `xinfo` on any address.

disconnect3d:
(Can this be somehow related to the removed assert in get_file?)

disconnect3d:
Fixed by https://github.com/pwndbg/pwndbg/pull/2386 thx to @OBarronCS

PR #2386, opened 2024-08-21
Fix vmmap on remote targets

### LINKED ISSUE #2385: VMMAP problems on remote
This issue details some issues and potential fixes to problems with detection of memory mappings when debugging a remote QEMU process.

QEMU 8.1 added support for `info proc mappings`, returning a list of memory regions, which Pwndbg uses:

https://github.com/pwndbg/pwndbg/blob/d59ae31360bec1de672008178be310a61438b60e/pwndbg/gdblib/vmmap.py#L112-L114

However, memory permissions on `vmmap` don't line up with those sent by the `info proc mappings` command. 

image

This is because our call to `info_proc_maps` has `parse_flags=False` as a default parameter. avoiding a codepath that reads the page permissions.  Additionally, the code that parses the string output of info proc mappings assumes that everything in the `objfile` column has text, which is not True for qemu, and needs to be updated.

https://github.com/pwndbg/pwndbg/blob/d59ae31360bec1de672008178be310a61438b60e/pwndbg/gdblib/vmmap.py#L437

On local processes, we use another, more accurate method https://github.com/pwndbg/pwndbg/blob/d59ae31360bec1de672008178be310a61438b60e/pwndbg/gdblib/vmmap.py#L466

Another consideration is that I've found that `info proc mappings` misses some mappings when debugging remote processes. The current code will assume that the mappings are complete, and return immediately.

https://github.com/pwndbg/pwndbg/blob/d59ae31360bec1de672008178be310a61438b60e/pwndbg/gdblib/vmmap.py#L123-L130

This means that manually "explored" pages never get returned on a call to vmmap.get. Whenever a pointer is telescoped in various parts of the codebase, it will call `explore()` with the address, adding it to a global list of explored pages:

https://github.com/pwndbg/pwndbg/blob/d59ae31360bec1de672008178be310a61438b60e/pwndbg/gdblib/vmmap.py#L266-L272

These pages are never combined with the `info_proc_mappings`, however. This causes us to not return them in vmmap, and also to continously re-explore them, since we don't remember that we've seen them before.

These lines (at the bottom of get()) could be combined with the output of the proc mappings, to properly get both our explored pages and the ones provided by `info proc mappings`.

https://github.com/pwndbg/pwndbg/blob/d59ae31360bec1de672008178be310a61438b60e/pwndbg/gdblib/vmmap.py#L183-L186

Finally, whenever we explore a new page we need to clear the cache of vmmap.get so it includes pages that have been explored in the current step.
https://github.com/pwndbg/pwndbg/blob/d59ae31360bec1de672008178be310a61438b60e/pwndbg/gdblib/vmmap.py#L122-L123

OBarronCS:
This PR fixes our parsing of `info proc mappings`, resulting in accurate vmmap's on remote qemu targets.

Previously, permissions were absent (always `---p`) and some pages were missing as well due to the parser function missing some cases. The parser didn't consider the case that the objfile column could be empty, leading the code to throw an exception in the try catch block and return None.

Old vmmap of a debugging a process in qemu:
image

Note that the name of the objfile for some pages was the permissions bits. This was due to the parser assuming the objfile field was non-empty.

New:
image

This fixes #2385. While making this change, I found the root cause of #2365 - we assumed that the `.objfile` of Page objects were local files, and didn't check if we could really open them. I added a check for file existence, so xinfo no longer crashes on remote targets.

### DIFF
--- a/pwndbg/gdblib/vmmap.py
+++ b/pwndbg/gdblib/vmmap.py
@@ -310,7 +310,7 @@ def coredump_maps() -> Tuple[pwndbg.lib.memory.Page, ...]:
     Parses `info proc mappings` and `maintenance info sections`
     and tries to make sense out of the result :)
     """
-    pages = list(info_proc_maps())
+    pages = list(info_proc_maps(parse_flags=False))
 
     started_sections = False
     for line in gdb.execute("maintenance info sections", to_string=True).splitlines():
@@ -392,10 +392,19 @@ def coredump_maps() -> Tuple[pwndbg.lib.memory.Page, ...]:
     return tuple(pages)
 
 
-def parse_info_proc_mappings_line(line: str, parse_flags: bool) -> Optional[pwndbg.lib.memory.Page]:
+def parse_info_proc_mappings_line(
+    line: str, perms_available: bool, parse_flags: bool
+) -> Optional[pwndbg.lib.memory.Page]:
     """
     Parse a line from `info proc mappings` and return a pwndbg.lib.memory.Page
     object if the line is valid.
+
+    Example lines:
+        0x4c3000           0x4c5000     0x2000    0xc2000  rw-p   /root/hello_world/main
+        0x4c5000           0x4cb000     0x6000        0x0  rw-p
+
+    The objfile column might be empty, and the permissions column is only present in GDB versions >= 12.1
+    https://github.com/bminor/binutils-gdb/commit/29ef4c0699e1b46d41ade00ae07a54f979ea21cc
 
     Args:
         line: A line from `info proc mappings`.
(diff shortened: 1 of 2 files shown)


pwndbg/gdblib/vmmap.py
```diff
@@ -453,9 +480,12 @@ def info_proc_maps(parse_flags=False) -> Tuple[pwndbg.lib.memory.Page, ...]:
         # On qemu user emulation, we may get: gdb.error: Not supported on this target.
         info_proc_mappings = []
 
+    # See if "Perms" is in the header line
+    perms_available = "Perms" in info_proc_mappings[3]
```
OBarronCS, line 484: Looking at the source code for `info proc maps` - https://github.com/bminor/binutils-gdb/blob/28c3bf55f0f9aca8619c6d01be34a02a887c5577/gdb/linux-tdep.c#L814, I made sure that it's always the fourth line of output.

pwndbg/gdblib/vmmap.py
```diff
@@ -392,31 +392,43 @@ def coredump_maps() -> Tuple[pwndbg.lib.memory.Page, ...]:
...
+        size_str = split_line[2]
+        offset_str = split_line[3]
+
+        if perms_available:
+            perm = split_line[4]
+            # The objfile column may be empty.
+            objfile = split_line[5] if len(split_line) > 5 else ""
         else:
-            start_str, _end, size_str, offset_str, perm, objfile = split_line
+            perm = "rwxp"
```
OBarronCS, line 429: A comment on #2051 (https://github.com/pwndbg/pwndbg/pull/2051#discussion_r1510297318) suggests that we change this to "----" (or maybe "---p"). This becomes relevant because another other function in this file that parses vmmaps for core dumps first calls this function to get data from `info proc mappings`, and then enriches it with data from `maintenance info sections`. The `parse_flags` boolean was made to prevent this function from setting the perm flags in that context.

disconnect3d:
Awesome, thanks!

PR #2388, opened 2024-08-23
Register annotation handlers for Arm cortex mode

OBarronCS:
This PR hooks up Arm Cortex mode - which we call `armcm` in Pwndbg - with the Unicorn emulator and our Arm annotation handlers. Now, when we debug a process detected as `armcm`, we get instruction annotations.

Put simply, Cortex-M effectively runs a subset of the instruction set that Cortex-A can. Cortex-M processors can only run Thumb mode instructions (which we added support for in #2292), which Cortex-A can switch to dynamically, so all our code written for normal Arm applies to `armcm` as well. The key difference that concerns us is a difference in the flags register.

### DIFF
--- a/pwndbg/gdblib/disasm/arm.py
+++ b/pwndbg/gdblib/disasm/arm.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from typing import Callable
 from typing import Dict
+from typing import Literal
 
 from capstone import *  # noqa: F403
 from capstone.arm import *  # noqa: F403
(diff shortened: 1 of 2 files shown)


pwndbg/gdblib/disasm/arm.py
```diff
@@ -291,4 +291,6 @@ def _parse_register(
         return target
 
 
-assistant = DisassemblyAssistant("arm")
+# Register the assistant for both ARM A-profile and ARM M-profile
+assistant = DisassemblyAssistant("arm", "cpsr")
+assistant = DisassemblyAssistant("armcm", "xpsr")
```
disconnect3d, line 296: Is this fine that we overwrite the variable here?

OBarronCS: Oh yeah - that variable is unused

PR #2390, opened 2024-08-26
Add support for AOSC OS

xtexx:
This adds support for [AOSC OS](https://github.com/AOSC-Dev/).

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -61,6 +61,15 @@ install_emerge() {
     sudo emerge --oneshot --deep --newuse --changed-use --changed-deps dev-lang/python dev-debug/gdb
 }
 
+install_oma() {
+    sudo oma refresh || true
+    sudo oma install -y gdb gdbserver python-3 glib make glibc-dbg curl
+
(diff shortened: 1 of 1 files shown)


setup.sh
```diff
@@ -61,6 +61,15 @@ install_emerge() {
     sudo emerge --oneshot --deep --newuse --changed-use --changed-deps dev-lang/python dev-debug/gdb
 }
 
+install_oma() {
+    sudo oma refresh || true
+    sudo oma install -y gdb gdbserver python-2 python-3 glib make glibc-dbg curl
```
disconnect3d: Why does this install python-2? What for?

xtexx: I saw it is installed for openSUSE (:

disconnect3d: I don't think we need it at all, I believe it can be removed. Can you see and try? (unless python-2 is used by sth else on ur system). Anyway, I'd remove it

xtexx: resolved

PR #2405, opened 2024-09-02
add gdt command

disconnect3d:
Yet another useful command when debugging Linux kernel ;)

image

See also:
* https://wiki.osdev.org/Global_Descriptor_Table
* https://wiki.osdev.org/GDT_Tutorial

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -708,6 +708,7 @@ def load_commands() -> None:
         import pwndbg.commands.dumpargs
         import pwndbg.commands.elf
         import pwndbg.commands.flags
+        import pwndbg.commands.gdt
         import pwndbg.commands.ghidra
         import pwndbg.commands.godbg
         import pwndbg.commands.got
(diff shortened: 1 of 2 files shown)


patryk4815:
every new command should be added in https://github.com/pwndbg/pwndbg/tree/dev/docs/commands ?

pwndbg/commands/__init__.py
```diff
@@ -708,6 +708,7 @@ def load_commands() -> None:
         import pwndbg.commands.dumpargs
         import pwndbg.commands.elf
         import pwndbg.commands.flags
+        import pwndbg.commands.gdt
```
mbrla0, line 711: Just a small question. You don't seem to use `gdb` or `gdblib` at all in the command. Couldn't this import go outside the `if` for GDB-only commands?

disconnect3d: I believe it should. I just didn't know where to put it. Please do it when u port some other commands to aglib

disconnect3d:
> every new command should be added in https://github.com/pwndbg/pwndbg/tree/dev/docs/commands ?

Yeah we need to update docs somehow. Need to figure out how to do it again, ugh.

Issue #2406, opened 2024-09-02
Heap tcache command doesn't produce proper display


itaysnir:
### Description

Upon running the `tcache` command, an invalid representation of `pwndbg.dbg.gdb.GDBValue` is being displayed:

```bash
pwndbg> tcache
tcache is pointing to: 0x129f010 for thread 1
{
  counts = <pwndbg.dbg.gdb.GDBValue object at 0x7f893c3db210>,
  entries = <pwndbg.dbg.gdb.GDBValue object at 0x7f893c3db250>,
}
```

### Steps to reproduce

1. Run a binary
2. Issue `tcache`

### My setup

```bash
pwndbg> version
Gdb:      14.2
Python:   3.11.9 (main, Apr  2 2024, 08:25:04) [GCC 13.2.0]
Pwndbg:   2024.08.29 build: c2bf902
Capstone: 5.0.1280
Unicorn:  2.0.1
```

P.S.
I'm deeply sorry in case there's no actual bug here. I did some pwndbg code auditing, and it seems like `GDBValue` doesn't implements `__str__` nor `__repr__` (and so is its parent class), which may be the issue.

Debugging journal:
1. By debugging `CStruct2GDB.__str__()`, defined within `pwndbg/aglib/heap/structs.py` , it does seems like the problem is missing `__str__` and `__repr__` methods.

2. By reading `GDBValue.string()`, defined within `pwndbg/dbg/gdb.py`, I thought it would serve as a candidate to replace the above stated methods. However, it doesn't seem to return a valid string.

Solution to the bug (will open PR soon, in case this really is a bug):
I've added the following method for `GDBValue`, which fixed the issue (would probably need to implement the parent class's "pure virtual" method of raising an exception + adding `@override` decorator here):

```python
def __repr__(self):
    return self.inner.__repr__()

def __str__(self):
    return self.inner.__str__()
```

Indeed, now the output is displayed properly:
```bash
pwndbg> tcache
tcache is pointing to: 0x1227010 for thread 1
{
  counts = {0, 0, 0, 0, 0, 1, 0 <repeats 58 times>},
  entries = {0x0, 0x0, 0x0, 0x0, 0x0, 0x12272a0, 0x0 <repeats 58 times>},
}
```

CptGibbon:
Probably worth speaking to @lebr0nli about this

disconnect3d:
Fixed in @2408. Needs further refactoring, but yeah, its fixed already

PR #2408, opened 2024-09-02
Add __repr__ and __str__ methods to GDBValue

itaysnir:
Fixes issue #2406

### DIFF
--- a/pwndbg/dbg/__init__.py
+++ b/pwndbg/dbg/__init__.py
@@ -720,6 +720,12 @@ class Value:
         """
         raise NotImplementedError()
 
+    def __repr__(self):
+        raise NotImplementedError()
+
+    def __str__(self):
+        raise NotImplementedError()
+
 
 class CommandHandle:
     """
--- a/pwndbg/dbg/gdb.py
+++ b/pwndbg/dbg/gdb.py
@@ -938,6 +938,14 @@ class GDBValue(pwndbg.dbg_mod.Value):
     def __getitem__(self, key: str | int) -> pwndbg.dbg_mod.Value:
         return GDBValue(self.inner[key])
 
+    @override
+    def __repr__(self):
+        return self.inner.__repr__()
+
+    @override
+    def __str__(self):
+        return self.inner.__str__()
+
 
 class GDB(pwndbg.dbg_mod.Debugger):
     @override


mbrla0:
I actually have a few reservations with this PR as it stands. 

Firstly, when adding things to the Debugger-agnostic API, I personally think it's important to also add an implementation for LLDB. I don't think we want our implementations of the API to be out of sync with each other.

Secondly, the way LLDB and GDB stringify values is very much not consistent between them, and you're choosing to use the format GDB spits out verbatim. This might not be that big of a deal, since I'm not sure we would ever have a reason to want them to be consistent, but, if we ever do, we'd have to either define a standard way to stringify values, or emulate whatever GDB does on the LLDB side, which is a considerable amount of work.

Thirdly, `SBValue` on the LLDB side doesn't have proper useful stringification that we can easily (and quickly) tap into, as far as I know, the closest it has is a `__repr__` implementation that isn't exactly friendly-looking or good to read in situations like the one in #2406. How would we have the output of `__str__` be useful there? Would we have to handle different types of values differently in the LLDB backend, all manually?

In fact, when I was writing the Debugger-agnostic API, I chose not to have `__str__` in `pwndbg.dbg.Value` because of the last two issues.

Personally, to fix #2406 I would go for changing the `__str__` method in `pwndbg.aglib.heap.structs.CStruct2GDB` instead, since that's where the problem is coming from, and we know that class only uses `int`'s anyway, and `pwndbg.dbg.Value` has an `__int__` implementation. But I feel like @disconnect3d should probably weigh in on this.

CptGibbon:
IMO an acceptable stopgap while pwndbg moves away from using debugger-specific classes like `GDBValue`

mbrla0:
> IMO an acceptable stopgap while pwndbg moves away from using debugger-specific classes like `GDBValue`

I don't know if I understand what you mean? We're not moving away from `GDBValue` at all. It's the value class in the GDB imeplmentation of the the Debugger-agnostic API. The one that implements all of the parts of Value that have to talk to GDB - like querying types, doing casts, doing derefs, etc. - when we're running under GDB.

itaysnir:
> s

Thank you for your comment. 
Would wait for @disconnect3d  reply, and I'd implement your suggested approach.

CptGibbon:
> I don't know if I understand what you mean? We're not moving away from `GDBValue` at all. It's the value class in the GDB imeplmentation of the the Debugger-agnostic API. The one that implements all of the parts of Value that have to talk to GDB - like querying types, doing casts, doing derefs, etc. - when we're running under GDB.

It's okay, I know what a GDBValue is.
We're moving away from them pervading every part of pwndbg, e.g. printing values directly from them in a pwndbg command. See #2082 for an example.

This particular case is to do with the cstruct layer for the heap code anyway.

Ping @lebr0nli about the above

mbrla0:
> > > IMO an acceptable stopgap while pwndbg moves away from using debugger-specific classes like `GDBValue`
> > 
> > 
> > I don't know if I understand what you mean? We're not moving away from `GDBValue` at all. It's the value class in the GDB imeplmentation of the the Debugger-agnostic API. The one that implements all of the parts of Value that have to talk to GDB - like querying types, doing casts, doing derefs, etc. - when we're running under GDB.
> 
> It's okay, I know what a GDBValue is. We're moving away from them pervading every part of pwndbg, e.g. printing values directly from them in a pwndbg command. This particular case is to do with the cstuct layer for the heap code anyway.
> 
> Ping @lebr0nli about the above

You're thinking of a `gdb.Value`, `GDBValue`s are something else. `GDBValue`s are implementations of `pwndbg.dbg.Value`, which is the thing we're moving to, and away from `gdb.Value`.

disconnect3d:
Ugh, that sounds really complex for something that looks very simple at first glance :).

I will accept this PR as a quickfix, to make things on dev work as expected, but it seems that @mbrla0 solution is better?

Can you @mbrla0 or @itaysnir revert this PR changes in another PR and implement what @mbrla0 suggested? Can we also add a testcase to confirm that the expected behavior work as expected after the change suggested by @mbrla0 ?

itaysnir:
> Ugh, that sounds really complex for something that looks very simple at first glance :).
> 
> I will accept this PR as a quickfix, to make things on dev work as expected, but it seems that @mbrla0 solution is better?
> 
> Can you @mbrla0 or @itaysnir revert this PR changes in another PR and implement what @mbrla0 suggested? Can we also add a testcase to confirm that the expected behavior work as expected after the change suggested by @mbrla0 ?

Sure, starting to work on this.

PR #2409, opened 2024-09-02
Add GDB scheduler-locking control to the Debugger-agnostic API

mbrla0:
As pointed out in #2382, always enabling scheduler locking when evaluating expressions can lead to deadlocks in GDB, so we want it to be off by default and let callers enable it only when absolutely required. This PR adds support for that in the Debugger-agnostic API.

It is important to note that, even though this is a fairly important thing to control, and thus I think it deserves a spot in the Debugger-agnostic API, scheduler locking is still a GDB concept, and that LLDB does [its own thing](https://github.com/llvm/llvm-project/blob/1601879f5d690595889a5537c0049b62df4657c7/lldb/source/Expression/ExpressionParser.cpp#L62) with expression evaluation, and does not have a direct equivalent to it, as far as I know. So adding this means we're leaking a GDB concept into a debugger-agnostic interface. This PR tries to minimize the impact of that by clearly documenting that `lock_scheduler` is a GDB-only parameter, and that other implementations should ignore it.

### DIFF
--- a/pwndbg/aglib/tls.py
+++ b/pwndbg/aglib/tls.py
@@ -16,7 +16,11 @@ def __call_pthread_self() -> int:
     if pwndbg.dbg.selected_inferior().symbol_address_from_name("pthread_self") is None:
         return 0
     try:
-        return int(pwndbg.dbg.selected_frame().evaluate_expression("(void *)pthread_self()"))
+        return int(
+            pwndbg.dbg.selected_frame().evaluate_expression(
+                "(void *)pthread_self()", lock_scheduler=True
+            )
+        )
     except pwndbg.dbg_mod.Error:
         return 0
 
--- a/pwndbg/dbg/gdb.py
+++ b/pwndbg/dbg/gdb.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import re
 import signal
+from contextlib import nullcontext
 from typing import Any
 from typing import Generator
 from typing import List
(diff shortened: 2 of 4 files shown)


pwndbg/aglib/tls.py
```diff
@@ -16,7 +16,11 @@ def __call_pthread_self() -> int:
     if pwndbg.dbg.selected_inferior().symbol_address_from_name("pthread_self") is None:
         return 0
     try:
-        return int(pwndbg.dbg.selected_frame().evaluate_expression("(void *)pthread_self()"))
```
disconnect3d, line 19: Did we initially lock scheduler here before implementing LLDB stuff?

If not, shall we really do it?

pwndbg/dbg/gdb.py
```diff
@@ -86,10 +87,12 @@ def __init__(self, inner: gdb.Frame):
...
     @override
-    def evaluate_expression(self, expression: str) -> pwndbg.dbg_mod.Value:
-        from pwndbg.gdblib.scheduler import lock_scheduler
+    def evaluate_expression(
+        self, expression: str, lock_scheduler: bool = False
+    ) -> pwndbg.dbg_mod.Value:
+        from pwndbg.gdblib.scheduler import lock_scheduler as do_lock_scheduler
 
-        with lock_scheduler():
+        with do_lock_scheduler() if lock_scheduler else nullcontext():
```
disconnect3d, line 95: Some problem I see with it is that if u execute:

```
evaluate_expression(a, lock=True)
evaluate_expression(b, lock=True)
```
you probably want to lock the scheduler once and only once for all the execution.

Also, if the evaluate_expression would somehow trigger another `lock()` then maybe there could be some issue? idk.

pwndbg/aglib/tls.py
mbrla0: The `gdblib` file that comes from [uses scheduler locking](https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/gdblib/tls.py#L23).

disconnect3d: Ok, this makes sense

pwndbg/dbg/gdb.py
disconnect3d: OTOH, they could then use lock scheduler outside? thats probably fine as it is now

PR #2410, opened 2024-09-02
Update Nix files for LLDB Pwndbg

mbrla0:
This PR updates our Nix files for LLDB Pwndbg support, under a new package, named `pwndbg-lldb`. It also updates `pwndbg-lldb.py` to handle different Pwndbg locations. We pick version 19 of LLDB in our flake, because when running under earlier versions we [hit an LLDB bug specific to Python >=3.12](https://github.com/llvm/llvm-project/issues/70453), which is the version of Python in the flake.

Feedback on my Nix is also especially appreciated, as I'm by no means a Nix guru.

### DIFF
--- a/nix/pwndbg.nix
+++ b/nix/pwndbg.nix
@@ -75,16 +75,12 @@ let
       cp -r lldbinit.py pwndbg $out/share/pwndbg
       cp pwndbg-lldb.py $out/bin/${pwndbgName}
 
-      # patchShebangs isn't working, so we do it manually.
-      sed -i "1 d" $out/bin/${pwndbgName}
-      sed -i "1 i #!${pkgs.lib.makeBinPath [ python3 ]}/python3" $out/bin/${pwndbgName}
-
       ${fix_init_script { target = "$out/bin/${pwndbgName}"; line = "4"; } }
 
       touch $out/share/pwndbg/.skip-venv
       wrapProgram $out/bin/${pwndbgName} \
         --prefix PATH : ${ pkgs.lib.makeBinPath [ lldb ] } \
-        '' + (pkgs.lib.optionalString !pkgs.stdenv.isDarwin ''
+        '' + (pkgs.lib.optionalString (!pkgs.stdenv.isDarwin) ''
         --set LLDB_DEBUGSERVER_PATH ${ pkgs.lib.makeBinPath [ lldb ] }/lldb-server \
         '') + ''
         --set PWNDBG_LLDBINIT_DIR $out/share/pwndbg
--- a/nix/pwndbg.nix
+++ b/nix/pwndbg.nix
@@ -83,8 +83,10 @@ let
       touch $out/share/pwndbg/.skip-venv
       wrapProgram $out/bin/${pwndbgName} \
         --prefix PATH : ${ pkgs.lib.makeBinPath [ lldb ] } \
-        --set PWNDBG_LLDBINIT_DIR $out/share/pwndbg \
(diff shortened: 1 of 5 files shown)


nix/pwndbg.nix
```diff
@@ -35,33 +40,61 @@ let
...
       "pwndbg/.*"
+    ] ++ (if isLLDB then [
+      "lldbinit.py"
+      "pwndbg-lldb.py"
+    ] else [
       "gdbinit.py"
-    ];
+    ]));
 
     nativeBuildInputs = [ pkgs.makeWrapper ];
```
patryk4815, line 59:
```suggestion
    nativeBuildInputs = [ pkgs.makeWrapper ];
    buildInputs = [ pyEnv ];
```

nix/pwndbg.nix
```diff
@@ -35,33 +40,61 @@ let
...
+    in (if isLLDB then ''
+      mkdir -p $out/share/pwndbg
+      mkdir -p $out/bin
+
+      cp -r lldbinit.py pwndbg $out/share/pwndbg
+      cp pwndbg-lldb.py $out/bin/${pwndbgName}
+
+      # patchShebangs isn't working, so we do it manually.
+      sed -i "1 d" $out/bin/${pwndbgName}
+      sed -i "1 i #!${pkgs.lib.makeBinPath [ python3 ]}/python3" $out/bin/${pwndbgName}
```
patryk4815:
```suggestion
```
Not needed when we have `buildInputs`

nix/pwndbg.nix
```diff
@@ -35,33 +40,61 @@ let
...
+
+      ${fix_init_script { target = "$out/bin/${pwndbgName}"; line = "4"; } }
+
+      touch $out/share/pwndbg/.skip-venv
+      wrapProgram $out/bin/${pwndbgName} \
+        --prefix PATH : ${ pkgs.lib.makeBinPath [ lldb ] } \
+        --set PWNDBG_LLDBINIT_DIR $out/share/pwndbg \
+        --set LLDB_DEBUGSERVER_PATH ${ pkgs.lib.makeBinPath [ lldb ] }/lldb-server
```
patryk4815:
```suggestion
      wrapProgram $out/bin/${pwndbgName} \
        --prefix PATH : ${ pkgs.lib.makeBinPath [ lldb ] } \
        '' + (pkgs.lib.optionalString !pkgs.stdenv.isDarwin ''
        --set LLDB_DEBUGSERVER_PATH ${ pkgs.lib.makeBinPath [ lldb ] }/lldb-server \
        '') + ''
        --set PWNDBG_LLDBINIT_DIR $out/share/pwndbg
```
- for darwin `LLDB_DEBUGSERVER_PATH` env must be unset
- for linux would be nice to use `set-default` it allows overriding 

it looks little ugly

patryk4815 (changes requested):

nix/pwndbg.nix
mbrla0: Yeah, it does look ugly for sure. Also, we probably don't want to let Linux users override it, since, in general, we probably want the debug server to always match the LLDB package we use. Darwin is an exception, because we have no other choice. Overriding the variable should probably be done by customizing the Nix files or overriding the derivation.

patryk4815: Ok 👍 Edited proposal

nix/pwndbg.nix
mbrla0: Removed it.

disconnect3d:
Waiting for @patryk4815 approval

patryk4815 (approved):
lgtm :)

PR #2416, opened 2024-09-03
Port a handful of commands to the Debugger-agnostic interfaces

mbrla0:
This PR ports a fairly large selection of commands to the Debugger-agnostic interfaces in Pwndbg (`dbg` and `aglib`), and allows them to run in both LLDB and GDB. This is the last PR that was split off #2382 to make it smaller and easier to merge, and it should bring upstream Pwndbg fully up to date with the features in the LLDB port branch. Personally, and like with #2382, I don't like how big this change is, either, but I couldn't come up with a way to split these that wouldn't have us dealing with right around 28 separate PRs. If you'd like me to split this off in some way, though, please tell me about it.

Aside from that, there are a two caveats to this PR:
- Not all commands in the port have tests of their own, and I'm not entirely familiar with how they work. While `aglib` is designed to behave the same as `gdblib` from the POV of these modules, I can't say that there can't be a bug somewhere in `aglib` that makes these commands break under some circumstances. I've tested them with a few inputs before and after the port, and also between GDB and LLDB, and they _look_ okay to me. These commands are: `leakfind`, `p2p`, `sigreturn`, `spray`, `valist`, `xinfo`, and `xor`.
- The `search` command has a bug, which it's had forever, as far as I can tell. It passes an `str` value to `pwndbg.search.search` straight through when `-t bytes` is used (or when no type is given, since that is the default), despite the fact that `pwndbg.search.search` expects a pattern that has a type of `bytes`. When `pwndbg.search` was GDB-only that didn't actually lead to any issues, because GDB [is more than happy to take a string in `gdb.Frame.search_memory`](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Inferiors-In-Python.html#Inferiors-In-Python), but now the Debugger-agnostic API actually expects and enforces that the value be a `bytearray`. So, because of that pre-existing bug, this port actually breaks `search` under both GDB and LLDB when `-t bytes` is used. Also, this issue never got caught in linting, despite being a typing issue, because the code in `pwndbg.commands.search` isn't typed. We should probably do something about `-t bytes`.

### DIFF
--- a/pwndbg/commands/elf.py
+++ b/pwndbg/commands/elf.py
@@ -99,7 +99,7 @@ def get_symbols_in_region(start, end, filter_text=""):
     addr = start
     while addr < end:
         name = pwndbg.dbg.selected_inferior().symbol_name_at_address(addr)
-        if name and name != "" and "+" not in name and filter_text in name:
+        if name and "+" not in name and filter_text in name:
             symbols.append((name, addr))
         addr += ptr_size
 
--- a/pwndbg/commands/elf.py
+++ b/pwndbg/commands/elf.py
@@ -12,7 +12,7 @@ from pwndbg.commands import CommandCategory
 )
 @pwndbg.commands.OnlyWithFile
 def elfsections() -> None:
-    local_path = pwndbg.gdblib.file.get_proc_exe_file()
+    local_path = pwndbg.aglib.file.get_proc_exe_file()
 
     with open(local_path, "rb") as f:
         elffile = ELFFile(f)
@@ -51,7 +51,7 @@ def plt() -> None:
 
 
 def get_section_bounds(section_name):
-    local_path = pwndbg.gdblib.file.get_proc_exe_file()
+    local_path = pwndbg.aglib.file.get_proc_exe_file()
 
     with open(local_path, "rb") as f:
         elffile = ELFFile(f)
@@ -74,8 +74,8 @@ def print_symbols_in_section(section_name, filter_text="") -> None:
         return
 
     # If we started the binary and it has PIE, rebase it
-    if pwndbg.gdblib.proc.alive:
-        bin_base_addr = pwndbg.gdblib.proc.binary_base_addr
+    if pwndbg.aglib.proc.alive:
+        bin_base_addr = pwndbg.aglib.proc.binary_base_addr
 
         # Rebase the start and end addresses if needed
         if start < bin_base_addr:
@@ -95,11 +95,11 @@ def print_symbols_in_section(section_name, filter_text="") -> None:
 
 def get_symbols_in_region(start, end, filter_text=""):
     symbols = []
-    ptr_size = pwndbg.gdblib.typeinfo.pvoid.sizeof
+    ptr_size = pwndbg.aglib.typeinfo.pvoid.sizeof
     addr = start
     while addr < end:
-        name = pwndbg.gdblib.symbol.get(addr, gdb_only=True)
-        if name != "" and "+" not in name and filter_text in name:
+        name = pwndbg.dbg.selected_inferior().symbol_name_at_address(addr)
+        if name and name != "" and "+" not in name and filter_text in name:
             symbols.append((name, addr))
         addr += ptr_size
 
--- a/pwndbg/commands/hexdump.py
+++ b/pwndbg/commands/hexdump.py
@@ -2,13 +2,11 @@ from __future__ import annotations
 
 import argparse
 
-import gdb
-
 import pwndbg
+import pwndbg.aglib.arch
+import pwndbg.aglib.memory
+import pwndbg.aglib.regs
 import pwndbg.commands
-import pwndbg.gdblib.arch
-import pwndbg.gdblib.memory
-import pwndbg.gdblib.regs
 import pwndbg.hexdump
 from pwndbg.color import message
 from pwndbg.commands import CommandCategory
@@ -29,18 +27,17 @@ pwndbg.config.add_param(
 )
 
 
-def address_or_module_name(s):
-    gdbval_or_str = pwndbg.commands.sloppy_gdb_parse(s)
-    if isinstance(gdbval_or_str, str):
-        module_name = gdbval_or_str
-        pages = list(filter(lambda page: module_name in page.objfile, pwndbg.gdblib.vmmap.get()))
+def address_or_module_name(s) -> int:
+    addr_or_str: int | str = pwndbg.commands.sloppy_gdb_parse(s)
+    if isinstance(addr_or_str, str):
+        module_name = addr_or_str
+        pages = list(filter(lambda page: module_name in page.objfile, pwndbg.aglib.vmmap.get()))
         if pages:
             return pages[0].vaddr
         else:
             raise argparse.ArgumentTypeError(f"Could not find pages for module {module_name}")
-    elif isinstance(gdbval_or_str, (int, gdb.Value)):
-        addr = gdbval_or_str
-        return addr
+    elif isinstance(addr_or_str, int):
+        return addr_or_str
     else:
         raise argparse.ArgumentTypeError("Unknown hexdump argument type.")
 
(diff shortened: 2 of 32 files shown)


disconnect3d:
Please open a bug report about `search -t bytes`

pwndbg/commands/elf.py
```diff
@@ -95,11 +95,11 @@ def print_symbols_in_section(section_name, filter_text="") -> None:
...
 def get_symbols_in_region(start, end, filter_text=""):
     symbols = []
-    ptr_size = pwndbg.gdblib.typeinfo.pvoid.sizeof
+    ptr_size = pwndbg.aglib.typeinfo.pvoid.sizeof
     addr = start
     while addr < end:
-        name = pwndbg.gdblib.symbol.get(addr, gdb_only=True)
-        if name != "" and "+" not in name and filter_text in name:
+        name = pwndbg.dbg.selected_inferior().symbol_name_at_address(addr)
+        if name and name != "" and "+" not in name and filter_text in name:
```
disconnect3d: I believe that the `if name` part does the same as if: `if name is not None and len(name)>0` so the `name != ""` is redundant now.
```suggestion
        if name and "+" not in name and filter_text in name:
```

pwndbg/commands/hexdump.py
```diff
@@ -29,18 +27,17 @@
...
+        pages = list(filter(lambda page: module_name in page.objfile, pwndbg.aglib.vmmap.get()))
         if pages:
             return pages[0].vaddr
         else:
             raise argparse.ArgumentTypeError(f"Could not find pages for module {module_name}")
-    elif isinstance(gdbval_or_str, (int, gdb.Value)):
-        addr = gdbval_or_str
-        return addr
+    elif isinstance(addr_or_str, int):
+        return addr_or_str
```
disconnect3d, line 40: Do I understand correctly that now the `sloppy_gdb_parse` returns `int` or `str` and never a gdb.value? as in the value is resolved to int already there?

disconnect3d: Random thought: the `sloppy_gdb_parse` name is not so great..

pwndbg/commands/pie.py
```diff
@@ -98,30 +99,30 @@ def piebase(offset=None, module=None) -> None:
...
-        # (as we do need the actual path that is in vmmap, not the file itself)
-        module = get_exe_name()
-    addr = translate_addr(offset, module)
-
-    if addr is not None:
-        spec = "*%#x" % (addr)
-        gdb.Breakpoint(spec)
-    else:
-        print(message.error("Could not determine rebased breakpoint address on current target"))
+if pwndbg.dbg.is_gdblib_available():
```
disconnect3d, line 102: Random thought: Wouldnt it be better/easier to just not import the command file if its not available at all?

Or is it that there could be other deps to it? I think there should never be such deps :X

pwndbg/commands/vmmap.py
```diff
@@ -54,12 +55,12 @@ def print_vmmap_gaps_table_header() -> None:
...
-        f"{'End':>{2 + 2 * pwndbg.gdblib.arch.ptrsize}} "
+        f"{'Start':>{2 + 2 * pwndbg.aglib.arch.ptrsize}} "
+        f"{'End':>{2 + 2 * pwndbg.aglib.arch.ptrsize}} "
         f"{'Perm':>4} "
         f"{'Size':>8} "
         f"{'Note':>9} "
-        f"{'Accumulated Size':>{2 + 2 * pwndbg.gdblib.arch.ptrsize}}"
+        f"{'Accumulated Size':>{2 + 2 * pwndbg.aglib.arch.ptrsize}}"
     )
     print(header)
```
disconnect3d, line 65: This should really be just `print ( ( ...) )` instead of `print(header)`. U can refactor it here or in another PR, whatever

pwndbg/commands/windbg.py
```diff
@@ -8,17 +8,28 @@
...
 
+if pwndbg.dbg.is_gdblib_available():
+    import gdb
+
+
+def enhex(size, value):
+    value = value & ((1 << 8 * size) - 1)
+    x = "%x" % abs(value)
+    x = x.rjust(size * 2, "0")
+    return x
```
disconnect3d, line 26: Just `return x.rjust(size * 2, "0")`

pwndbg/commands/windbg.py
```diff
@@ -8,17 +8,28 @@
...
+
+
+# `pwndbg.hexdump` imports `enhex` from this module, so we have to import it
+# after it's been defined in order to avoid circular import errors.
+import pwndbg.hexdump
```
disconnect3d, line 31: I don't quite understand this. It seems still to be a circular import and it should be the same/similar imports before. Why is it an issue now and wasn't before?

pwndbg/commands/xinfo.py
```diff
@@ -30,9 +30,9 @@ def xinfo_stack(page, addr) -> None:
     # If it's a stack address, print offsets to top and bottom of stack, as
     # well as offsets to current stack and base pointer (if used by debuggee)
 
-    sp = pwndbg.gdblib.regs.sp
-    frame = pwndbg.gdblib.regs[pwndbg.gdblib.regs.frame]
-    frame_mapping = pwndbg.gdblib.vmmap.find(frame)
+    sp = pwndbg.aglib.regs.sp
+    frame = pwndbg.aglib.regs[pwndbg.aglib.regs.frame]
+    frame_mapping = pwndbg.aglib.vmmap.find(frame)
```
disconnect3d, line 35: Should probably move the two variables below the 3 prints but w/e.

disconnect3d (approved):
LGTM, except maybe: https://github.com/pwndbg/pwndbg/pull/2416#discussion_r1743713099

The rest are minor/not so important refactorings

pwndbg/commands/windbg.py
mbrla0: > It seems still to be a circular import and it should be the same/similar imports before.

It still is, but, to the best of my knowledge, by importing it after `enhex` has been defined, we allow [the module code execution step of the import](https://docs.python.org/3/reference/import.html#loading) to assign a value to `enhex` before the circular import happens, and thus `pwndbg.hexdump` always sees it defined. The issue I was getting before I moved the circular import statement down was that `enhex` wasn't defined in `pwndbg.commands.windbg`, which was correct, because it very much hadn't been by that point in execution order.

> Why is it an issue now and wasn't before?

I can't say for sure, because I didn't actually spend any time digging into what import specifically caused it to break. But the thing is, circular imports are very sensitive to the order in which you first import things, and the LLDB port refactors shuffled around the order of many of the import statements in Pwndbg. My best guess is that the original relative import order for `pwndbg.commands.windbg` and `pwndbg.hexdump` got reversed somewhere along the port, and that's what caused it to break.

mbrla0: I think the original order might've had `pwndbg.hexdump` come first, and `pwndbg.commands.windbg` come later. That would prevent the `import pwndbg.hexdump` statement in `pwndbg.commands.windbg` from triggering the code execution step for `pwndbg.hexdump` before it had defined `enhex`, which is what caused our ciruclar import error.

Moving the import down also means either relative import order works now.

pwndbg/commands/pie.py
mbrla0: Well, the problem is that `breakrva` is the only command in that module that can't run without GDB. `piebase` can. If we want to avoid importing entire command modules, we would probably have to split the commands that can run and those that can't into separate modules, whenever an existing module has both.

pwndbg/commands/hexdump.py
mbrla0: > Do I understand correctly that now the `sloppy_gdb_parse` returns `int` or `str` and never a gdb.value? as in the value is resolved to int already there?

Yeap. Though, the funny thing is that I actually didn't change that, [it was like that even before I started working on the port](https://github.com/pwndbg/pwndbg/blob/db320b6b973d7684ebbbb8f1987e6d48b55dda78/pwndbg/commands/__init__.py).

PR #2418, opened 2024-09-04
Better solution for #2406 bug, as proposed by #2408

itaysnir:
Hello, adding a solution, as proposed within #2408. 

Regarding the test - I'm not sure what's the adequate place for it, as this isn't really a bug within the `tcache` command itself (which seem to have no tests), but within `CStruct2GDB`. 
What do you suggest?

### DIFF
--- a/pwndbg/aglib/heap/structs.py
+++ b/pwndbg/aglib/heap/structs.py
@@ -161,7 +161,7 @@ class CStruct2GDB:
         """
         output = "{\n"
         for f in self._c_struct._fields_:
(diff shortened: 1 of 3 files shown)


disconnect3d:
Btw shouldn't we revert the recent change with str/repr with this?

Regarding tests: its fine to test tcache command, its also fine to test CStruct2GDB (ideally we need both ;D)

disconnect3d:
@itaysnir ping: this does not revert the 2408 pr changes atm (I see a commit for this, but this commit must have been reverted by a next one ;/)

disconnect3d:
Ping @itaysnir

PR #2419, opened 2024-09-04
sigreturn command: improve help string

disconnect3d:
Inform on the default address in `help sigreturn`


disconnect3d:
https://github.com/pwndbg/pwndbg/blob/df8a9d4dc6155aaefd2a84d8a7a44e90745c8d27/pwndbg/commands/telescope.py#L88-L90 hmmmm maybe it should be done like this?

disconnect3d:
Yeah this should be done as default arg

PR #2422, opened 2024-09-10
Fix deference-limit changes not being reflected due to an early cast to int()

cfreal:
The `dereference-limit` param was cast to `int` when given as a default argument in the prototype of `pwndbg.chain.{get,format}`, which meant that if its value got updated,  the change would not get reflected in the output of (for instance) `plist`.

Had to add a few chain links in `linked-lists.c` to build a meaningful test.

### DIFF
--- a/tests/gdb-tests/tests/test_command_plist.py
+++ b/tests/gdb-tests/tests/test_command_plist.py
@@ -16,6 +16,65 @@ def startup(start_binary):
     gdb.execute("run")
     gdb.execute("up")
 
+def test_command_plist_dereference_limit_change_has_impact_on_plist(start_binary):
+    """
+    Tests the plist command with different dereference limits
+    """
+    startup(start_binary)
+    gdb.execute("set dereference-limit 5")
+    expected_out = re.compile(
+        """\
+0[xX][0-9a-fA-F]+ <node_a>: {\\s*
+  value = 0,\\s*
+  next = 0[xX][0-9a-fA-F]+ <node_b>\\s*
+}\\s*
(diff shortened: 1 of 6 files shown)


tests/gdb-tests/tests/test_command_plist.py
```diff
@@ -9,14 +9,74 @@
...
 
+def test_command_plist_dereference_limit_change_has_impact_on_plist(start_binary):
+    """
+    Tests the plist command with different dereference limits
+    """
+    startup(start_binary)
+    gdb.execute("set dereference-limit 5")
+    expected_out = re.compile(
+        """\
+0[xX][0-9a-fA-F]+ <node_a>: {\\s*
```
disconnect3d, line 28: I wonder if this would be more readable if we formatted it with `f"..."` and do sth like `{identifier} {whitespaces}` etc.

But this is good enough anyway.

disconnect3d (approved):
Thanks a lot for this! I wonder if we have more issues like this elsewhere and if we can prevent it somehow (maybe some custom lint or semgrep rules for CI? huh!).

tests/gdb-tests/tests/test_command_plist.py
cfreal: I was thinking the same, but ATM no time - and TBH no will - to implement this.

cfreal:
> Thanks a lot for this! I wonder if we have more issues like this elsewhere and if we can prevent it somehow (maybe some custom lint or semgrep rules for CI? huh!).

I agree. For what it's worth, a quick grep for `: ([a-z]+) = \1\(` yields 0 matches, so it might not be such an urgent matter.

PR #2423, opened 2024-09-10
Added support for --offset and --count for plist

cfreal:
`plist` now only displays `--count` elements (defaults to `dereference-limit`) starting from element `--offset` (defaults to zero). This is useful when you have huge linked lists and you're only interested in part of it.

Note that it changes the default behaviour: before, due to the implementation of the function, it displayed 7 elements (head, next, and 5 more from another `chain()` call). Now, it only displays 5.

_Note: Will put the commit history in order by creating a branch for next PRs._

### DIFF
--- a/pwndbg/commands/plist.py
+++ b/pwndbg/commands/plist.py
@@ -209,7 +209,8 @@ def plist(
 
     if count is None:
         count = pwndbg.config.dereference_limit
-    if count == 0:
+    elif count <= 0:
+        print("count <= 0: not displaying any elements")
         return
 
     # We suport being passed either a pointer to the first structure or the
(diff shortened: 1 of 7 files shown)


cfreal:
Note: should probably implement a `positive_int` type for command arguments.

pwndbg/commands/plist.py
```diff
@@ -182,6 +207,11 @@ def plist(path, next, sentinel, inner_name, field_name) -> None:
         print(message.error(f"{path} has been optimized out"))
         return
 
+    if count is None:
+        count = pwndbg.config.dereference_limit
+    if count == 0:
```
disconnect3d:
```suggestion
    elif count <= 0:
        print("count <= 0: not displaying any elements")
```

cfreal: We need to keep `if` in case `dereference_limit` is zero.

cfreal:
Note: tests failed because of a timeout. Need to re-run.

PR #2431, opened 2024-09-13
fix for issue #2430

### LINKED ISSUE #2430: BUG: pwndbg.gdblib.proc.binary_base_addr doesn't work on remote target
### Description

currently using `binary_base_addr` or  `binary_vmmap` doesn't work for remote targets, this breaks the binary ninja plugin

pwndbg/gdblib/proc.py
```python
      @property
      def binary_base_addr(self) -> int:
          return self.binary_vmmap[0].start

      @property
      def binary_vmmap(self) -> Tuple[pwndbg.lib.memory.Page, ...]:
          import pwndbg.gdblib.vmmap

          return tuple(p for p in pwndbg.gdblib.vmmap.get() if p.objfile == self.exe)
```

This happens because `self.exe` in `binary_vmmap` returns the absolute path, which of course won't correlate for the remote.

I propose either scanning only for the file base name (only if we have a remote target, currently running this patch locally) or finding another way to distinguish the base address.

### Steps to reproduce

1. simple setup with gdbserver and target remote
2. connecting to binja works (`set integration-provider binja`), but `bn-sync` crashes due to reasons stated above

### My setup

i used my plugin [vagd](https://github.com/gfelber/vagd) and found this bug.
1. install with `pip install vagd`
2. create template for abitrary binary `vagd template -e <BINARY>`
3. execute with gdb `./exploit.py GDB`
4. connecting to binja works (`set integration-provider binja`), but `bn-sync` crashes due to reasons stated above

gfelber:
fixes #2430, that compares base file names instead of absolute paths for `pwndbg.gdblib.proc.binary_vmmap`, if the target is run on remote.

### DIFF
--- a/pwndbg/aglib/proc.py
+++ b/pwndbg/aglib/proc.py
@@ -75,18 +75,15 @@ class module(ModuleType):
         return pwndbg.dbg.selected_inferior().alive()
 
     @property
+    @pwndbg.lib.cache.cache_until("objfile")
     def exe(self) -> str | None:
         """
-        Returns the debugged file name.
+        Returns the executed file path.
 
-        On remote targets, this may be prefixed with "target:" string.
-        See this by executing those in two terminals:
-        1. gdbserver 127.0.0.1:1234 /bin/ls
-        2. gdb -ex "target remote :1234" -ex "pi pwndbg.gdblib.proc.exe"
+        On remote targets, this path most definitly won't exist locally.
 
-        If you need to process the debugged file use:
-            `pwndbg.gdblib.file.get_proc_exe_file()`
-            (This will call `pwndbg.gdblib.file.get_file(pwndbg.gdblib.proc.exe, try_local_path=True)`)
+        If you need the locally referenced file use:
+            `gdb.current_process().filename`
         """
 
         return pwndbg.dbg.selected_inferior().main_module_name()
--- a/pwndbg/aglib/proc.py
+++ b/pwndbg/aglib/proc.py
@@ -80,7 +80,7 @@ class module(ModuleType):
         """
         Returns the executed file path.
 
-        On remote targets, this path most definitly won't exist locally.
+        On remote targets, this path may not exist locally.
 
         If you need the locally referenced file use:
             `gdb.current_process().filename`
--- a/pwndbg/dbg/gdb.py
+++ b/pwndbg/dbg/gdb.py
@@ -738,7 +738,11 @@ class GDBProcess(pwndbg.dbg_mod.Process):
     def main_module_name(self) -> str | None:
         # Can GDB ever return a different value here from what we'd get with
         # `info files`, give or take a "remote:"?
-        return gdb.current_progspace().filename
+        if not self.alive():
+            return gdb.current_progspace().filename
+
+        exe = gdb.execute("info proc exe", to_string=True)
+        return exe[exe.find("exe = '") + 7 : exe.rfind("'")]
 
     @override
     def main_module_entry(self) -> int | None:
--- a/pwndbg/gdblib/proc.py
+++ b/pwndbg/gdblib/proc.py
@@ -117,7 +117,7 @@ class module(ModuleType):
         If you need the locally referenced file use:
             `gdb.current_process().filename`
 
-        info proc exe is the only command to actually get the executed file.
+        info proc exe is the only command to actually get the executed file path.
         The gdb `file` command overwrites all internal references, this includes:
         + `filename`
         + `executable_filename`
--- a/pwndbg/gdblib/proc.py
+++ b/pwndbg/gdblib/proc.py
@@ -107,20 +107,34 @@ class module(ModuleType):
         return "It stopped with signal " in gdb.execute("info program", to_string=True)
 
     @property
+    @pwndbg.lib.cache.cache_until("objfile")
     def exe(self) -> str | None:
         """
-        Returns the debugged file name.
+        Returns the executed file path.
 
-        On remote targets, this may be prefixed with "target:" string.
-        See this by executing those in two terminals:
-        1. gdbserver 127.0.0.1:1234 /bin/ls
-        2. gdb -ex "target remote :1234" -ex "pi pwndbg.gdblib.proc.exe"
+        On remote targets, this path most definitly won't exist locally.
 
-        If you need to process the debugged file use:
-            `pwndbg.gdblib.file.get_proc_exe_file()`
-            (This will call `pwndbg.gdblib.file.get_file(pwndbg.gdblib.proc.exe, try_local_path=True)`)
+        If you need the locally referenced file use:
+            `gdb.current_process().filename`
+
+        info proc exe is the only command to actually get the executed file.
+        The gdb `file` command overwrites all internal references, this includes:
+        + `filename`
+        + `executable_filename`
+        + `symbol_file`
+        + `objfiles`
+        `run` executes the current file
+
+        If you find a better solution please create a PR <3.
+
+        Also refer to pwngdb.dbg.gdb.main_module_name
         """
-        return gdb.current_progspace().filename
+
+        if not self.alive:
+            return gdb.current_progspace().filename
+
+        exe = gdb.execute("info proc exe", to_string=True)
+        return exe[exe.find("exe = '") + 7 : exe.rfind("'")]
 
     @property
     @pwndbg.lib.cache.cache_until("start", "stop")
--- a/pwndbg/gdblib/proc.py
+++ b/pwndbg/gdblib/proc.py
@@ -112,7 +112,7 @@ class module(ModuleType):
         """
         Returns the executed file path.
 
-        On remote targets, this path most definitly won't exist locally.
+        On remote targets, this path may not exist locally.
 
         If you need the locally referenced file use:
             `gdb.current_process().filename`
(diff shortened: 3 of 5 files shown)


Aplet123:
This seems janky to me, I think it's preferable to pull the name from `pwndbg.aglib.proc.exe` instead and do an exact comparison.

gfelber:
is the new solution preferred? 
Also, sry for the commit spam i don't have a linter in my nvim setup

Aplet123:
Have you tested the new code? I think it might not work, as it doesn't remove the `target:` prefix from the exe name, and still gets the basename of the objfile.

gfelber:
~~it worked on my setup (explained in issue)~~

gfelber:
nvmd it didn't work

gfelber:
pwndbg.aglib.proc.exe still returns the absolute path (and not `target:` prefix),
do you want this comparison
```python
                if os.path.basename(p.objfile) == os.path.basename(pwndbg.aglib.proc.exe)
```

gfelber:
these are my gdb settings
```
 ['/usr/bin/gdb', '-ex', 'set sysroot /PATH/TO/SYSROOT/./.vagd/sysroot/', '-ex', 'file -readnow ./binary', '-q', '-x', '/tmp/pwnlib-gdbscript-f6on
dmql.gdb']
```
note that I set file to my local path (so i don't need to redundantly download it from the remote.
Not sure if this creates a misunderstanding, but might be interesting information

Aplet123:
I think `p.objfile == pwndbg.aglib.proc.exe.removeprefix("target:")` should work.

gfelber:
no sadly not, pwndbg.aglib.proc.exe returns the absolute path on my local system, not the remote

Aplet123:
Can you send the result of `info proc mappings` and `pi pwndbg.aglib.proc.exe`?

gfelber:
`info proc mappings`:
```
Mapped address spaces:

          Start Addr           End Addr       Size     Offset  Perms  objfile
            0x400000           0x401000     0x1000        0x0  r--p   /REMOTE/PATH/binary
           0x8000000          0x8001000     0x1000     0x1000  r-xp   /REMOTE/PATH/binary
           0x8001000          0x8002000     0x1000     0x2000  rw-p   /REMOTE/PATH/binary
          0xffff0000         0xfffff000     0xf000        0x0  rw-p   
      0x7ffff7ff9000     0x7ffff7ffd000     0x4000        0x0  r--p   [vvar]
      0x7ffff7ffd000     0x7ffff7fff000     0x2000        0x0  r-xp   [vdso]
      0x7ffffffde000     0x7ffffffff000    0x21000        0x0  rw-p   [stack]
```
`pi pwndbg.aglib.proc.exe`
```
/LOCAL/PATH/binary
```

gfelber:
> these are my gdb settings
> 
> ```
>  ['/usr/bin/gdb', '-ex', 'set sysroot /PATH/TO/SYSROOT/./.vagd/sysroot/', '-ex', 'file -readnow ./binary', '-q', '-x', '/tmp/pwnlib-gdbscript-f6on
> dmql.gdb']
> ```
> 
> note that I set file to my local path (so i don't need to redundantly download it from the remote. Not sure if this creates a misunderstanding, but might be interesting information

this might be the reason `pi pwndbg.aglib.proc.exe` returns the local path

gfelber:
yeah if you specify the file locally using args `gdb ./binary` or the file command in gdb `file ./binary` pwndbg.aglib.proc.exe get's overwritten with the local path

gfelber:
also fixed a soft lock issue if `set integration-provider binja` is executed before the binary is run

disconnect3d:
@Aplet123 Wdyt?

pwndbg/aglib/proc.py
```diff
@@ -75,18 +75,15 @@ def alive(self) -> bool:
...
     def exe(self) -> str | None:
         """
-        Returns the debugged file name.
+        Returns the executed file path.
 
-        On remote targets, this may be prefixed with "target:" string.
-        See this by executing those in two terminals:
-        1. gdbserver 127.0.0.1:1234 /bin/ls
-        2. gdb -ex "target remote :1234" -ex "pi pwndbg.gdblib.proc.exe"
+        On remote targets, this path most definitly won't exist locally.
```
disconnect3d:
```suggestion
        On remote targets, this path may not exist locally.
```

pwndbg/dbg/gdb.py
```diff
@@ -738,7 +738,11 @@ def module_section_locations(self) -> List[Tuple[int, int, str, str]]:
     def main_module_name(self) -> str | None:
         # Can GDB ever return a different value here from what we'd get with
         # `info files`, give or take a "remote:"?
-        return gdb.current_progspace().filename
+        if not self.alive():
+            return gdb.current_progspace().filename
+
+        exe = gdb.execute("info proc exe", to_string=True)
```
disconnect3d, line 744: Fwiw, just for information, it seems GDB fetches this via `/proc/$pid/exe`:

https://github.com/bminor/binutils-gdb/blob/gdb-15.1-release/gdb/infcmd.c#L3024-L3028

https://github.com/bminor/binutils-gdb/blob/gdb-15.1-release/gdb/infcmd.c#L2960-L2972


https://github.com/bminor/binutils-gdb/blob/gdb-15.1-release/gdb/linux-tdep.c#L810-L821

https://github.com/bminor/binutils-gdb/blob/gdb-15.1-release/gdb/linux-tdep.c#L882-L891

pwndbg/gdblib/proc.py
```diff
@@ -107,20 +107,34 @@ def stopped_with_signal(self) -> bool:
...
     def exe(self) -> str | None:
         """
-        Returns the debugged file name.
+        Returns the executed file path.
 
-        On remote targets, this may be prefixed with "target:" string.
-        See this by executing those in two terminals:
-        1. gdbserver 127.0.0.1:1234 /bin/ls
-        2. gdb -ex "target remote :1234" -ex "pi pwndbg.gdblib.proc.exe"
+        On remote targets, this path most definitly won't exist locally.
```
disconnect3d:
```suggestion
        On remote targets, this path may not exist locally.
```

pwndbg/gdblib/proc.py
```diff
@@ -107,20 +107,34 @@ def stopped_with_signal(self) -> bool:
...
-        2. gdb -ex "target remote :1234" -ex "pi pwndbg.gdblib.proc.exe"
+        On remote targets, this path may not exist locally.
 
-        If you need to process the debugged file use:
-            `pwndbg.gdblib.file.get_proc_exe_file()`
-            (This will call `pwndbg.gdblib.file.get_file(pwndbg.gdblib.proc.exe, try_local_path=True)`)
+        If you need the locally referenced file use:
+            `gdb.current_process().filename`
+
+        info proc exe is the only command to actually get the executed file.
```
disconnect3d:
```suggestion
        info proc exe is the only command to actually get the executed file path.
```

pwndbg/dbg/gdb.py
gfelber: make sense, but creating a custom implementation that simulates this behavior (for remote and local) would be worse imo (way to complex for what it's worth)
therefore i would believe that `info proc exe`, is still the preferred solution

patryk4815:
should we write tests to check this behavior? (`exe`)

disconnect3d:
Ideally, yes pls

Aplet123:
> @Aplet123 Wdyt?

LGTM, although I wonder what the observed deadlock is with enabling binja integration before starting the binary is, as I can't seem to reproduce it.

gfelber:
weird can't seem to reproduce the binja integration softlock bug again either, maybe it was some issue that occurred while I messed with the code base.

I just noticed that the pwndbg event handler trigger an update that resulted in calling `gdblib.proc.get_base_adddress` while the process wasn't running, causing an error and preventing gdb from any state changes (because they triggered the event handler again, which caused another crash)

gfelber:
I think i recreate the bug (or found another one), you get the same issue if you setup the integration and then stop it inside binary ninja, it doesn't seem to actually softlock gdb, but it breaks the context view (and generally speaking all pwndbg commands), you can't even disable it afterwards with `set integration-provider none`

gfelber:
yeah this is not the issue i had yesterday, but an issue nonetheless

gfelber:
ok implemented fixes for the "new" issue that should at least prevent the pwndbg soft lock issue. 
Also kept the `gdblib.proc.alive` requirement for events that actually need it to be alive, not sure if this is needed, but it makes sense to check it and safely return, to prevent potential pwndbg soft locks.

pwndbg/integration/binja.py
```diff
@@ -153,13 +153,24 @@ def init_bn_rpc_client() -> None:
...
 
 def with_bn(fallback: K = None) -> Callable[[Callable[P, T]], Callable[P, T | K]]:
+    global _bn
+
     def decorator(func: Callable[P, T]) -> Callable[P, T | K]:
+        global _bn
+
         @functools.wraps(func)
         def wrapper(*args: P.args, **kwargs: P.kwargs) -> T | K:
+            global _bn
```
disconnect3d, line 163: Does this really require global x3? I think only this one is needed?

disconnect3d (approved):

pwndbg/integration/binja.py
gfelber: i tried but it didn't work without it, just python stuff i guess

Issue #2404, opened 2024-09-01
Remove or document MemoryAllocator.breakpoint


disconnect3d:
https://github.com/pwndbg/pwndbg/blob/26ba400dbdeafa09fce1f77fb93d9940d7db00ba/pwndbg/gdblib/heap/heap.py#L11-L20

As in https://github.com/pwndbg/pwndbg/pull/2382#discussion_r1733653835 : is it sth we need? Seems like sth to be removed.

cc: @CptGibbon

CptGibbon:
Not sure about the origin of this, I assume someone wanted an easy way to breakpoint common memory allocator operations 🤷‍♂️

disconnect3d:
Let's just remove this. PRs welcome if someone wants to help :)

disconnect3d:
Done in https://github.com/pwndbg/pwndbg/pull/2433 ; the class can be further refactored in the future

PR #2433, opened 2024-09-15
Solving issue #2404 

AadishJ:
Removed breakpoint from heap as it is a redundant function and no longer required


disconnect3d:
Fwiw there's more than this that can be refactored there if u look on how the class is used, what are its methods and so on, but this PR is good anyway. Thanks!

PR #2438, opened 2024-09-17
Fix attachp tests when ptrace_scope is missing

peace-maker:
Assume we can attach to any process when ptrace_scope is not available. This is the case in WSL2.

### DIFF
--- a/tests/gdb-tests/tests/test_attachp.py
+++ b/tests/gdb-tests/tests/test_attachp.py
@@ -5,6 +5,7 @@ import os
 import re
 import subprocess
 import tempfile
+from pathlib import Path
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


tests/gdb-tests/tests/test_attachp.py
```diff
@@ -16,10 +17,15 @@
...
-        if len(result) >= 1 and result[0] == "0":
-            can_attach = True
+    ptrace_scope = Path("/proc/sys/kernel/yama/ptrace_scope")
+    if ptrace_scope.exists():
+        with ptrace_scope.open() as f:
+            result = f.read()
+            if len(result) >= 1 and result[0] == "0":
+                can_attach = True
+    else:
+        # If the file doesn't exist, assume we can attach
```
disconnect3d, line 27:
```suggestion
        # If the file doesn't exist, assume we can attach
        # This is the case e.g. for running tests under WSL2
```

PR #2439, opened 2024-09-17
Add history of context output

### LINKED ISSUE #1899: Save history of previous context output
You would save some scrolling if you could reprint the context of a previous stop. Caching the displayed data everytime context is used would allow to review old state (without actually rewinding the inferior of course).

My major motivation for this is the [splitmind layout](https://github.com/pwndbg/pwndbg/blob/dev/FEATURES.md#splitting--layouting-context) which organizes the output neatly, but loses the ability to review the output of a previous step. Looking for solutions there's [hyperpwn](https://github.com/bet4it/hyperpwn) which implements this based on a special terminal emulator. But having shortcuts like `ctrl+shift+pageup/down` to walk the history would be cool to have integrated in pwndbg without requiring a special terminal program.

@disconnect3d

peace-maker:
Every context section is cached individually to allow to display older output again. You can scroll through the old context output using `contextprev` and `contextnext`. This allows to "scroll" in TUI mode. When you cause a new stop using e.g. `si` while browsing the context history, the history position is reset and the new context is displayed right away. You can control or disable the history size using `set context-history-size 50` which is a scrollback-memory tradeoff. I don't think memory is an issue here though for a few strings.

The history position is displayed next to the memory legend:

grafik

The history is kept over multiple runs of a binary and you can display old contexts after the program exited. You can skip multiple entries by passing an argument to `contextprev`. `contextprev 10` would go back 10 entries into the past. `contextsearch something` allows to search the history for something and select the entry in the history matching that.

There is no hotkey key combination like `ctrl+shift+pageup/down` yet and you have to use `ctxp`/`ctxn` or the buttons in the TUI to scroll through the history.

Fixes #1899

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -587,15 +587,15 @@ def context(subcontext=None) -> None:
     if len(args) == 0:
         args = config_context_sections.split()
 
-    def get_history_legend():
-        if selected_history_index is not None:
+    sections = []
+    if args:
+        if selected_history_index is None:
+            sections.append(("legend", lambda *args, **kwargs: [M.legend()]))
+        else:
             longest_history = max(len(h) for h in context_history.values())
-            return f" (history {selected_history_index + 1}/{longest_history})"
-        return ""
+            history_status = f" (history {selected_history_index + 1}/{longest_history})"
+            sections.append(("legend", lambda *args, **kwargs: [M.legend() + history_status]))
 
-    sections = (
-        [("legend", lambda *args, **kwargs: [M.legend() + get_history_legend()])] if args else []
-    )
     sections += [(arg, context_sections.get(arg[0], None)) for arg in args]
 
     result = defaultdict(list)
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -2,14 +2,20 @@ from __future__ import annotations
 
 import argparse
 import ast
+import functools
+import logging
 import os
 import sys
 from collections import defaultdict
 from typing import Any
+from typing import Callable
 from typing import DefaultDict
 from typing import Dict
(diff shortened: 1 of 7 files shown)


pwndbg/commands/context.py
```diff
@@ -377,21 +569,33 @@ def context_ghidra(target=sys.stdout, with_banner=True, width=None):
...
+
     if subcontext is None:
         subcontext = []
     args = subcontext
 
     if len(args) == 0:
         args = config_context_sections.split()
 
-    sections = [("legend", lambda *args, **kwargs: [M.legend()])] if args else []
+    def get_history_legend():
```
disconnect3d: Why a function at all?

disconnect3d: I mean, its called just once and in a condition; we can just do an if and assign proper value or not.

(note that we build the function in a function, which is done on each function call; meh)

pwndbg/gdblib/tui/control.py
```diff
@@ -0,0 +1,43 @@
+from __future__ import annotations
+
+import gdb
+
+
+class ControlTUIWindow:
+    _tui_window: "gdb.TuiWindow"
+    _button_text: str = "[←]  [→]"
+    _button_spans = {"contextprev": (0, 2), "contextnext": (5, 7)}
```
disconnect3d, line 13: What is 0, 2? Can we maybe add a comment about it?

pwndbg/commands/context.py
peace-maker: You're right, I was still in the mindset of capturing loop variables from the TUI window PR.

disconnect3d:
(Needs fixing some conflicts :<)

PR #2442, opened 2024-09-18
Add `context [section(s)] [--on|--off]` to temporarily disable sub-sections

### LINKED ISSUE #1479: Add some commands for selectively turn specific sub-context off/on
e.g. If stack-context not desirable for me in some cases, and it occupy a mass of terminal space, I need to turn it off.
`ctx stack off` cmd better that current `set context-sections regs disasm code ghidra backtrace expression` (which don't set the stack in config)

Sketchy idea is that add a global variable dict (str -> bool) for filter and memorize all on/off.
Or ugly but simply `gdb.execute("set context-sections...")` in context funtion every time when encounter a off/on cmd.(Seems maybe not need extra modification beyond context function).

peace-maker:
Allow to temporarily disable sub-sections in the context output without touching the `context-sections` config parameter.

I went with `ctx [section(s)] --on` instead of `ctx [section(s)] on` suggested in #1479, since the subsection vararg would have to be awkwardly checked for sections "on" and "off" afaict and this was easier to document.

Fixes #1479

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -96,7 +96,7 @@ config_max_threads_display = pwndbg.config.add_param(
 
 # Storing output configuration per section
 outputs: Dict[str, str] = {}
-output_settings = {}
+output_settings: DefaultDict[str, Dict[str, Any]] = defaultdict(dict)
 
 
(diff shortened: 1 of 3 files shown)


disconnect3d:
Awesome, thanks!

PR #2443, opened 2024-09-19
Fix display of gdb functions documentation

### LINKED ISSUE #2424: Document added functions (base, rebase, bn_eval, bn_sym, bn_var etc)
TL;DR: Those functions should all be documented so that GDB's `apropos <...>` or `help function <funcname>` shows their help:
```
pwndbg> apropos This function is not documented
function base -- This function is not documented.
function bn_eval -- This function is not documented.
function bn_sym -- This function is not documented.
function bn_var -- This function is not documented.
function ida -- This function is not documented.
function rebase -- This function is not documented.

pwndbg> help function rebase
This function is not documented.
```

peace-maker:
The __doc__ is read in the gdb.Function constructor, so copy it from the function to the class in the decorator before calling the parent constructor.

`apropos` displays the first line of the docstring only, so move it up in the `bn_*` commands.

Fixes #2424

### DIFF
--- a/pwndbg/commands/binja.py
+++ b/pwndbg/commands/binja.py
@@ -30,9 +30,7 @@ def bn_sync(*args) -> None:
 @pwndbg.gdblib.functions.GdbFunction()
 @pwndbg.integration.binja.with_bn()
 def bn_sym(name_val: gdb.Value) -> int | None:
-    """
-    Lookup a symbol's address by name from Binary Ninja.
-    """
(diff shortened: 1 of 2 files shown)


disconnect3d:
@peace-maker hm.. we should add a test like:
```py
assert gdb.execute('apropos This function is not documented') == ''
```
:D

peace-maker:
Yes and one for empty docs too somehow 😄

PR #2444, opened 2024-09-19
Fix terminal width to 80 columns in tests

peace-maker:
Set a `PWNDBG_IN_TEST`environment variable when running gdb in tests. Use the dimensions in `LINES` and `COLUMNS` when looking up the window size when that envvar is set.

This makes context output always be 80 columns wide which allows to compare to hardcoded output.

### DIFF
--- a/pwndbg/ui.py
+++ b/pwndbg/ui.py
@@ -73,6 +73,8 @@ def get_window_size(target=sys.stdin):
     rows, cols = get_cmd_window_size()
     if rows is not None and cols is not None:
         return rows, cols
+    elif os.environ.get("PWNDBG_IN_TEST") is not None:
+        return fallback
     try:
         # get terminal size and force ret buffer len of 4 bytes for safe unpacking by passing equally long arg
         rows, cols = struct.unpack("hh", fcntl.ioctl(target.fileno(), termios.TIOCGWINSZ, b"1234"))
--- a/tests/gdb-tests/conftest.py
+++ b/tests/gdb-tests/conftest.py
@@ -4,6 +4,8 @@ This file should consist of global test fixtures.
 
 from __future__ import annotations
 
+import os
+
 import gdb
 import pytest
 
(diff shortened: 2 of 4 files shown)


tests/gdb-tests/conftest.py
```diff
@@ -17,8 +19,12 @@ def start_binary():
     """
 
     def _start_binary(path, *args):
+        os.environ["PWNDBG_IN_TEST"] = "1"
         gdb.execute("file " + path)
         gdb.execute("set exception-verbose on")
+        gdb.execute("set width 80")
+        gdb.execute("set env COLUMNS 80")
```
disconnect3d: Isn't this setting the env not for gdb but for debugged process?

pwndbg/ui.py
```diff
@@ -73,6 +73,8 @@ def get_window_size(target=sys.stdin):
     rows, cols = get_cmd_window_size()
     if rows is not None and cols is not None:
         return rows, cols
+    elif os.environ.get("PWNDBG_IN_TEST") is not None:
+        return fallback
```
disconnect3d, line 77: random thought: fallback should never change, right? nobody should ever do `os.environ[a]=b`? can we compute it once at the top of the file?

and then probably run the tests with `LINES=x COLUMNS=y ./tests.sh`?

peace-maker: I'm not sure if they are updated ever, is this an optimization? But I don't like the idea of having to pass environment variables to `tests.sh`. Running `./tests.sh` should just work.

disconnect3d: isn't tests.sh launching an external python code or script? ;D if so then it could insert the env :D.

disconnect3d: But yeah idk how often is get_window_size called, if often then we should probably optimize it if possible, heh

disconnect3d: anyway good for now

peace-maker: I didn't look at the tests.sh script, so you're probably right. We'd have to make sure gdb doesn't overwrite the COLUMNS during startup.

PR #2449, opened 2024-09-20
Make annotation reg/memory assignment symbol configurable

OBarronCS:
This PR makes the symbol that denotes a register/memory assignment in annotations configurable. Previously, the symbol was always `=>`.

For example, with `set annotation-memory-separator <=`, you would get the following
image

It introduces two config values:

```
annotation-memory-separator
annotation-register-separator
```

Currently, I kept the default the same as before, mostly due to my own personal preference and becoming accustomed with it. However, we should explore other possibilities before merging:

- Make memory operands use `:=` by default
- Make memory operands `<=` by default
- Make registers use `<=` by default as well
- Or just a `=` sign, or even a fancy Unicode character.

After we decide, we might also have to go back through the tests and make them reflect the changes.


disconnect3d:
I think we should follow what is in telescope and do `<-`? and just call it a day: we probably don't need too many configuration options

OBarronCS:
Latest commits takes this into account - removes the config and makes store operations use a left pointing arrow.

pwndbg/aglib/disasm/arch.py
```diff
@@ -126,6 +126,21 @@ def syntax_highlight(ins):
 }
 
 
+def register_assign(left: str, right: str) -> str:
+    return f"{left} => {right}"
+
+
+def memory_assign(left: str, right: str) -> str:
+    return f"{left} <= {right}"
```
disconnect3d, line 134: we introduce lots of function calls with those and funccalls are expensive in py, but maybe it doesnt hurt

OBarronCS: I get that - worst case, these functions are called once per instruction shown on the screen.

Issue #2448, opened 2024-09-19
Symbol not shown for chain of code pointer on x86_64


peace-maker:
### Description

The disassembly view / `u` command annotates the instructions and shows the symbol to an address. The `tele` / chain code doesn't.

![grafik](https://github.com/user-attachments/assets/ca21e53e-ba25-4eb6-b218-4127e8c46632)

![grafik](https://github.com/user-attachments/assets/f7e8e1b7-8d48-48a3-ac9f-7dce34bf80e2)

I'd expect to see the symbol in that single-instruction disassembly too.

### Steps to reproduce

- `gdb tests/gdb-tests/tests/binaries/reference-binary.out`, `start` and step to the `puts` call
- Observe symbol in `disasm` context section while seeing the raw address in hex in the `regs` section next to RIP
- `tele $rip` will show the address too

### My setup

Gdb:      12.1
Python:   3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0]
Pwndbg:   2024.08.29 build: 15a69272
Capstone: 5.0.1280
Unicorn:  2.0.1

PR #2450, opened 2024-09-20
Resolve instruction operand symbols in register chain

OBarronCS:
This fixes a small bug where we weren't replacing constant operands with their corresponding symbols when the `chain` function detected an instruction at the end of a pointer chain.

Example before:
image
Note that the `RIP` register points to a `call` instruction, but the constant operand is not replaced with the symbol, like it is in the disasm view.

Example After:
image


OBarronCS:
This fixes #2448

PR #2451, opened 2024-09-23
bump: capstone=5.0.3 & unicorn=2.1.1

patryk4815:
~~- todo check capstone tests~~

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -5,11 +5,11 @@
         "systems": "systems"
       },
       "locked": {
-        "lastModified": 1710146030,
-        "narHash": "sha256-SZ5L6eA7HJ/nmkzGG7/ISclqe6oZdOZTNoesiInkXPQ=",
+        "lastModified": 1726560853,
+        "narHash": "sha256-X6rJYSESBVr3hBoH0WbKE5KvhPU5bloyZ2L4K60/fPQ=",
         "owner": "numtide",
         "repo": "flake-utils",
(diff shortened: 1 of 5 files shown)


OBarronCS:
The failed archlinux tests is resulting from these lines of code: 

https://github.com/unicorn-engine/unicorn/blob/f164769a9a973a3e981f279ed7aa90459ae68545/bindings/python/unicorn/unicorn_py3/unicorn.py#L502-L511

Testing it, the binding versions are 2 and 1, but the uc_versions are 2 and 0.

It may be due to the Unicorn package on Arch not being updated yet - https://archlinux.org/packages/extra/x86_64/unicorn/

When you `pacman -Syu python-unicorn` (which the install script does) it installs the 2.0.1 versions 
image

But the binding library version in the .venv is 2.1.0. This mismatch causes the failures.

patryk4815:
@OBarronCS are you sure? We are only installing everything from `poetry install` 🤔 why archlinux would install wrong version

patryk4815:
Ah I see `pacman -S --noconfirm --needed git gdb python python-capstone python-unicorn` in https://github.com/pwndbg/pwndbg/blob/dev/setup.sh#L80
We should't use 'random' repo dependencies 😕

patryk4815:
Archlinux fixed, right now only `nix release` failing

patryk4815:
Nix fixed

patryk4815:
Ready for code review / merge

PR #2454, opened 2024-09-25
Remove `gdblib.arch` in favor of `aglib.arch`

mbrla0:
This PR removes `pwndbg.gdlib.arch` in favor of its `aglib` counterpart, as part of the effort to deduplicate the `gdblib` code that was previously ported to the Debugger-agnostic API, as outlined in #2366. 

The current `aglib.arch` module should be mostly identical in behavior to the `gdblib` version  with the exception that, before the first `START` event fires, `arch.ptrsize` will _always_ be equal to `4`, as opposed to being whatever GDB reports for `gdb.lookup_type("void").pointer().sizeof` (usually `4`). However, I don't think there's any code in Pwndbg that expects `arch.ptrsize` to be correct before the first `START` event is fired[^1], so this is likely good enough.

[^1]: Plus, any code that assumes that `gdblib.arch.ptrsize` is correct before the first `START` is wrong anyway, since currently `gdblib.arch.ptrsize` is equal to `4` before the first `START` event, at least in x86_64. We end up just repeating that behavior in `aglib`.


disconnect3d:
I'll review this hopefully next week. Currently, -ENOTIME due to https://2024finals.justctf.team/ prep ;).

disconnect3d (approved):

Issue #2455, opened 2024-09-25
search function - simple text


kerszl:
Greets

### Description

For some time now, the search function has not been working properly.

When I enter a search with a given string, I get an error. Searching by the string, of course ending it with zero, only finds the end of the string, not the beginning. Searching by bytes works, but it's a bit annoying

Error:
```
pwndbg> search What
Searching for value: 'What'
'search': Search memory for byte sequences, strings, pointers, and integer values.

By default search results are cached. If you want to cache all results, but only print a subset, use --trunc-out. If you want to cache only a subset of results, and print the results immediately, use --limit. The latter is specially useful if you're searching a huge section of memory.
Exception occurred: search: string argument without an encoding (<class 'TypeError'>)
```
OK:
```
pwndbg> search -x 57686174
Searching for value: b'What'
chk             0x555555556030 'What do you think the flag is?'
chk             0x555555557030 'What do you think the flag is?'
```
Ok, but I don't always expect it
```
pwndbg> search -t string What
Searching for value: b'What\x00'
```
### My setup

Gdb:      12.1
Python:   3.10.12 (main, Sep 11 2024, 15:47:36) [GCC 11.4.0]
Pwndbg:   2024.08.29 build: c2bf902e
Capstone: 5.0.1280
Unicorn:  2.0.1

disconnect3d:
cc: @mbrla0 I bet this bug was introduced due to lldb port refactorings? :/

mbrla0:
> cc: @mbrla0 I bet this bug was introduced due to lldb port refactorings? :/

This issue is the same as #2420. The LLDB refactorings caused `search -t bytes` to break, because now `pwndbg.search.search` actually enforces the `bytes` typing in its signature. We have to fix this in the search command itself, but I'm not familiar enough with what `search -t bytes` is supposed to do to fix it myself without changing its semantics.

Like, we could just do something like `value.encode("ascii")` in `pwndbg/commands/search.py` if we're dealing with `-t binary`, and that could be a fairly quick fix. I could do that later, if you want, and think that's good enough for a solution.

mbrla0:
> Like, we could just do something like `value.encode("ascii")` in `pwndbg/commands/search.py` if we're dealing with `-t binary`, and that could be a fairly quick fix. I could do that later, if you want, and think that's good enough for a solution.

Turns out figuring out the old behavior exactly was much more straightforward than I thought, and we can just emulate it wholesale. I've put the fix in a PR.

[...]

Issue #2420, opened 2024-09-04
`search -t bytes` calls `pwndbg.search.search` with a string


mbrla0:
### Description

In the making of #2416, I found a bug in the `search` command, which causes it to pass an `str`-typed object as the `searchfor` parameter in `pwndbg.search.search`, which expects a `bytes` object. As this part of `pwndbg.commands.search` is not typed, our linting step never caught this.

Prior to #2382 this bug did not actually lead to any issues, as `pwndbg.search.search` would end up only passing the `searchfor` parameter on to `gdb.Inferior.search_memory`, which [accepts `str` objects](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Inferiors-In-Python.html#Inferiors-In-Python). The Debugger-agnostic API, however, does not, and, since  #2382 got merged, `pwndbg.search.search` now relies on the assumption that the type of its `searchfor` parameter is correct, where it didn't before, leading to `search -t bytes` being broken.

This is the full traceback:
```
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /nix/store/ja81p25pjbgyqikbzmkx2qi4kin6x4wn-pwndbg/share/pwndbg/pwndbg/commands/__init__.py:187  │
│ in __call__                                                                                      │
│                                                                                                  │
│   184 │                                                                                          │
│   185 │   def __call__(self, *args: Any, **kwargs: Any) -> str | None:                           │
│   186 │   │   try:                                                                               │
│ ❱ 187 │   │   │   return self.function(*args, **kwargs)                                          │
│   188 │   │   except TypeError:                                                                  │
│   189 │   │   │   print(f"{self.function.__name__.strip()!r}: {self.function.__doc__.strip()}"   │
│   190 │   │   │   pwndbg.exception.handle(self.function.__name__)                                │
│                                                                                                  │
│ /nix/store/ja81p25pjbgyqikbzmkx2qi4kin6x4wn-pwndbg/share/pwndbg/pwndbg/commands/__init__.py:378  │
│ in _OnlyWhenRunning                                                                              │
│                                                                                                  │
│   375 │   def _OnlyWhenRunning(*a: P.args, **kw: P.kwargs) -> Optional[T]:                       │
│   376 │   │   # TODO: Properly support OnlyWhenRunning without `gdblib`.                         │
│   377 │   │   if pwndbg.aglib.proc.alive:                                                        │
[...]

PR #2458, opened 2024-10-02
Encode strings passed to the `search` command when `-t bytes` is used

mbrla0:
Should fix both #2420 and #2455.

As outlined in #2420, the LLDB port broke the `search` command when `-t bytes` is used, even under GDB. This PR should restore the same behavior we had before the refactorings got merged, by encoding the strings we get as UTF-8 before passing them on to `pwndbg.search.search`. This mimics  the way `str` objects passed to `gdb.Inferior.search_memory` get treated, as detailed in the comment right above the fix.


mbrla0:
Ideally we'd add a test for this too? Dunno why there wasn't ever one to begin with.

disconnect3d:
Yeah, pls add a test!

PR #2464, opened 2024-10-06
Fix #2463 - memory operands with segment registers + offset register

### LINKED ISSUE #2463: wrong annotation of instructions with segment registers
### Description
image
```bash
pwndbg> p/x $fs_base
$6 = 0x7ffff7ff5700
pwndbg> x/gx 0x7ffff7ff5700
0x7ffff7ff5700:	0x00007ffff7ff5700
pwndbg> p/x $rbp
$7 = 0xffffffffffffffc0
pwndbg> p/x $fs_base + $rbp
$8 = 0x7ffff7ff56c0
pwndbg> x/gx 0x7ffff7ff56c0
0x7ffff7ff56c0:	0x0000000000000000
pwndbg> 
```
The `mov    rbx, qword ptr fs:[rbp]` instruction puts the value at address `0x7ffff7ff56c0` into rbx, which is also obvious from the annotation of the test instruction (rbx is zero). However the annotation of the mov instruction itself implies that the `mov` puts the value of address `0x7ffff7ff5700` into rbx which is not what is happening, that is just the value of fs.

### Steps to reproduce

Should be simple to reproduce, possibly with stack canaries as well, I'll provide a more detailed guide later if it proves troublesome.

### My setup
Gdb:      15.1
Python:   3.12.6
Pwndbg:   2024.08.29 build: d6a39fc9
Capstone: 5.0.1280
Unicorn:  2.0.1

OBarronCS:
This fixes #2463 and makes correct memory address calculations when the operand uses a segment register and the offset is also a register. Previously, it assumed that the offset value was always a constant, which is typically seen in referencing stack canaries.

### DIFF
--- a/pwndbg/aglib/disasm/x86.py
+++ b/pwndbg/aglib/disasm/x86.py
@@ -270,21 +270,22 @@ class DisassemblyAssistant(pwndbg.aglib.disasm.arch.DisassemblyAssistant):
         # Get memory address (Ex: lea    rax, [rip + 0xd55], this would return $rip+0xd55. Does not dereference)
         if op.mem.segment != 0:
             if op.mem.segment == X86_REG_FS:
-                if (base := pwndbg.aglib.regs.fsbase) is None:
+                if (seg_base := pwndbg.aglib.regs.fsbase) is None:
                     return None
             elif op.mem.segment == X86_REG_GS:
-                if (base := pwndbg.aglib.regs.gsbase) is None:
+                if (seg_base := pwndbg.aglib.regs.gsbase) is None:
                     return None
             else:
                 return None
+        else:
+            seg_base = 0
 
-        # Both a segment and base cannot be in use
-        elif op.mem.base != 0:
-            base = self._read_register(instruction, op.mem.base, emu)
-            if base is None:
+        if op.mem.base != 0:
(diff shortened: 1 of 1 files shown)


k4lizen:
Maybe you could add support for the other segment registers as well (in another PR?)? So CS, ES, DS and SS are missing.
```bash
~> objdump -d -M intel vmlinux | rg "ds:"
ffffffff8100e694:	f3 48 a5             	rep movs QWORD PTR es:[rdi],QWORD PTR ds:[rsi]
[snip, many results]
~> objdump -d -M intel vmlinux | rg "es:"
ffffffff8100155d:	f3 48 ab             	rep stos QWORD PTR es:[rdi],rax
[snip, many results]
~> objdump -d -M intel vmlinux | rg "ss:"
~> objdump -d -M intel vmlinux | rg "cs:"
~> 
```
DS and ES at least seem to be widely used in the linux kernel. There are even some instructions with both of them (see output of first command).

Edit: This would involve querying the GDT I suppose since only FS and GS have `_base` variants as MSRs. Not sure on the details, would be quite niche.

k4lizen:
Also they should be interpreted differently if the CPU is in real mode, see https://wiki.osdev.org/Segmentation . But that's its own can of worms

disconnect3d:
May be worth to add a test case for this as well if possible

PR #2469, opened 2024-10-09
Add support for controlling execution to the Debugger-agnostic API

mbrla0:
This PR adds support for process execution control (currently continuing until a given breakpoint is hit, and single stepping) to the Debugger-agnostic API.

This functionality is implemented in terms of coroutines (`async def`) and awaitables, with the LLDB and GDB implementations being responsible for driving to completion all coroutines passed to them through the newly-added `Process.dispatch_execution_controller` method, with help from their respective implementations of `ExecutionController`, at some point in time between the method call and the command ending.

Using `async` has a couple of benefits due to both how we handle process lifetimes in LLDB, and how LLDB itself responds to process execution control commands issued during command processing. For the former, using `async` allows us to both keep all control of the process event loop in `pwndbg.dbg.lldb.repl`, and reap the benefits of cleaner command processing. For the latter, using coroutines allows us to easily and intuitively defer the execution control procedure until after command processing by LLDB is over, as issuing execution control commands during command processing breaks our event loop.

When it comes to GDB, its implementation runs the execution control procedure immediately, and pretty much foregoes all of the `async` piping. This solution does incur a little extra cost in terms of runtime over calling things like `gdb.execute("continue")` directly, but I do believe it is worth paying as it makes a much cleaner interface when one takes into account it has to work for both GDB and LLDB.

### DIFF
--- a/pwndbg/dbg/gdb.py
+++ b/pwndbg/dbg/gdb.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import re
 from contextlib import nullcontext
 from typing import Any
-from typing import Awaitable
 from typing import Coroutine
 from typing import Generator
 from typing import List
@@ -831,22 +830,14 @@ class GDBProcess(pwndbg.dbg_mod.Process):
                 break
 
 
-async def empty_awaitable() -> None:
-    """
-    Do nothing. We return this in our execution controller.
-    """
-    pass
-
-
 class GDBExecutionController(pwndbg.dbg_mod.ExecutionController):
     @override
-    def single_step(self) -> Awaitable[None]:
+    async def single_step(self):
         gdb.execute("si")
-        return empty_awaitable()
 
-    def cont(self, until: pwndbg.dbg_mod.StopPoint) -> Awaitable[None]:
+    @override
+    async def cont(self, until: pwndbg.dbg_mod.StopPoint):
         gdb.execute("continue")
-        return empty_awaitable()
 
 
 # Like in LLDB, we only need a single instance of the execution controller.
--- a/pwndbg/dbg/gdb.py
+++ b/pwndbg/dbg/gdb.py
@@ -3,6 +3,8 @@ from __future__ import annotations
 import re
 from contextlib import nullcontext
 from typing import Any
+from typing import Awaitable
+from typing import Coroutine
 from typing import Generator
 from typing import List
 from typing import Literal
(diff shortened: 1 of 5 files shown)


pwndbg/dbg/gdb.py
```diff
@@ -806,6 +808,50 @@ def is_dynamically_linked(self) -> bool:
...
+    pass
+
+
+class GDBExecutionController(pwndbg.dbg_mod.ExecutionController):
+    @override
+    def single_step(self) -> Awaitable[None]:
+        gdb.execute("si")
+        return empty_awaitable()
+
+    def cont(self, until: pwndbg.dbg_mod.StopPoint) -> Awaitable[None]:
```
disconnect3d: Lack of override?

also do we really need empty_awaitable? Can't we have those methods being async itself, or is that some problem?

pwndbg/dbg/lldb/__init__.py
```diff
@@ -586,6 +588,66 @@ def set_enabled(self, enabled: bool) -> None:
...
+
+class LLDBExecutionController(pwndbg.dbg_mod.ExecutionController):
+    @override
+    def single_step(self) -> Awaitable[None]:
+        return OneShotAwaitable(YieldSingleStep())
+
+    @override
+    def cont(self, target: pwndbg.dbg_mod.StopPoint) -> Awaitable[None]:
+        assert isinstance(target, LLDBStopPoint)
+        t: LLDBStopPoint = target
```
disconnect3d: Just pass `target`? Can't we handle this with typing better?

pwndbg/dbg/lldb/__init__.py
```diff
@@ -586,6 +588,66 @@ def set_enabled(self, enabled: bool) -> None:
...
+class YieldSingleStep:
+    """
+    Moves execution of the process being debugged forward by one instruction.
+
+    This class is part of the execution controller system, so it is intented to
+    be yielded by the async function with access to an execution controller, and
+    caught and hanlded by the event loop in the LLDB Pwndbg CLI.
+    """
+
+    pass
```
disconnect3d, line 629: This is todo, right? Same for YieldContinue?

pwndbg/dbg/gdb.py
mbrla0: > Lack of override?

Yeap. I forgot to add it.

> also do we really need empty_awaitable? Can't we have those methods being async itself, or is that some problem?

It didn't occur to me that I could just do that and still have MyPy accept them with the `@override`, but apparently it's completely fine with that. So, great!

pwndbg/dbg/lldb/__init__.py
mbrla0: I was gonna say "it's probably either the assignment or `typing.cast()`, which is even worse", but apparently MyPy is smart enough now that we don't need either of them, and just having the assertion is good enough for it to catch on. I'm gonna take the assignment out.

pwndbg/dbg/lldb/__init__.py
mbrla0: Not really, we just yield them up to the event loop in the Pwndbg CLI, which then distinguishes them by type, and handles them. The way we do it, having two classes is hardly any different from having a single class, since we'd still need a way to distinguish between the continue and single step cases.

Issue #2461, opened 2024-10-04
Current Python version (3.8.10) is not allowed by the project (^3.10).


c0o1sy1z3:
### Description

I can't setup pwndbg.
It showed " Current Python version (3.8.10) is not allowed by the project (^3.10). Please change python executable via the "env use" command." while running setup.sh

### Steps to reproduce

proxychains4 git clone https://github.com/pwndbg/pwndbg.git
cd pwndbg    
proxychains4 ./setup.sh    

### My setup

GNU gdb (Ubuntu 9.2-0ubuntu1~20.04.2) 9.2
Python 3.10.14 (main, Apr  6 2024, 18:45:05) [GCC 9.4.0]
pwndbg git:(dev)

disconnect3d:
Hi, can you use a release version? If not, you can build one with nix as well.

@patryk4815 I guess we dropped support for 3.8 a bit too fast..

disconnect3d:
Alternatively, please revert the Poetry.lock change with requirement of Py 3.8 from https://github.com/pwndbg/pwndbg/pull/2421/ and send a PR with that. I mean:
<img width="1116" alt="image" src="https://github.com/user-attachments/assets/540f1560-8021-45de-a599-8d53c515674f">

Pls check if that works for you before sending the PR.

patryk4815:
I think for old system you should use old pwndbg `2024.08.29`
Or we can mention this in `readme.md` 
++Add message in `setup.sh` that <ubuntu22.04 is not supported and add good message there 

Right now we have only message:
> Pwndbg is supported on Ubuntu 22.04, and 24.04 with GDB 12.1 and later. We do not test on any older versions of Ubuntu, so pwndbg may not work on these versions (for Ubuntu 18.04 use the [2023.07.17: ubuntu18.04-final release](https://github.com/pwndbg/pwndbg/releases/tag/2023.07.17)).

disconnect3d:
I merged the message update, but I am still not sure if its ok that we dropped Py 3.8 if someone still rely on its being LTS.

We can of course tell ppl to patch poetry.lock and build it, or, use a release, but they may want to use dev with latest features.

disconnect3d:
I'm happy to merge back support for 3.8 if anyone wants it. Otherwise, please use the release version or build your own all-in package with Nix.

Closing this for now.

PR #2470, opened 2024-10-09
Add message for older python version

patryk4815:
https://github.com/pwndbg/pwndbg/issues/2461

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -152,11 +152,6 @@ if linux; then
             ;;
         "gentoo")
             install_emerge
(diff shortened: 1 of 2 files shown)


setup.sh
```diff
@@ -152,11 +152,6 @@ if linux; then
             ;;
         "gentoo")
             install_emerge
-            if ! hash sudo 2> /dev/null && whoami | grep root; then
```
patryk4815, line 155: this was redundant, it is checked at first line of this file

Issue #2467, opened 2024-10-08
`ctx disasm` throws exception when attached to remote Windows gdbserver


sudoBash418:
### Description

`pwndbg` throws a `ValueError: Bad register` exception when running `ctx disasm`.
This seems to happen whenever I'm connected to a Windows `gdbserver` - I saw the same behaviour with Wine-Staging 9.10 and a Windows 10 VM.

This issue does not occur with the `2024.02.14` release.

<details>
<summary>Traceback</summary>

```
Traceback (most recent call last):
  File "/tmp/pwndbg/pwndbg/gdblib/regs.py", line 40, in gdb_get_register
    return frame.read_register(name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: Bad register

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/tmp/pwndbg/pwndbg/commands/__init__.py", line 187, in __call__
    return self.function(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/pwndbg/pwndbg/commands/context.py", line 641, in context
    func(
  File "/tmp/pwndbg/pwndbg/commands/context.py", line 312, in _serve_context_history
    current_output = function(*a, **kw)
                     ^^^^^^^^^^^^^^^^^^
  File "/tmp/pwndbg/pwndbg/commands/context.py", line 866, in context_disasm
    result = pwndbg.aglib.nearpc.nearpc(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/pwndbg/pwndbg/aglib/nearpc.py", line 142, in nearpc
    instructions, index_of_pc = pwndbg.aglib.disasm.near(
                                ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/pwndbg/pwndbg/aglib/disasm/__init__.py", line 387, in near
    emu = pwndbg.emu.emulator.Emulator()
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/pwndbg/pwndbg/emu/emulator.py", line 251, in __init__
    value = getattr(pwndbg.gdblib.regs, reg)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/pwndbg/pwndbg/lib/cache.py", line 162, in decorator
    value = func(*a, **kw)
[...]

PR #2478, opened 2024-10-16
Patch _fs_gs_helper when debugging windows remotely

AndersFelde:
I am using mingw64's gdbserver.exe to debug a windows PE remotely from Linux. Pwndbg will crash because the `fs_base` and `gs_base` registers does not exist when debugging in this way. By moving the `is_remote` check earlier (before it looks up the registers and crashes) pwndbg seems to work fine.

### DIFF
--- a/pwndbg/gdblib/regs.py
+++ b/pwndbg/gdblib/regs.py
@@ -9,7 +9,13 @@ import ctypes
 import re
 import sys
 from types import ModuleType
-from typing import Any, Callable, Dict, Generator, List, Tuple, cast
+from typing import Any
+from typing import Callable
+from typing import Dict
+from typing import Generator
+from typing import List
+from typing import Tuple
+from typing import cast
 
 import gdb
 
@@ -21,7 +27,9 @@ import pwndbg.gdblib.remote
 import pwndbg.gdblib.typeinfo
 import pwndbg.lib.cache
 from pwndbg.dbg import EventType
-from pwndbg.lib.regs import BitFlags, RegisterSet, reg_sets
+from pwndbg.lib.regs import BitFlags
+from pwndbg.lib.regs import RegisterSet
+from pwndbg.lib.regs import reg_sets
 
 
 @pwndbg.gdblib.proc.OnlyWhenRunning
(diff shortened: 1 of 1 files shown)


disconnect3d:
Hi, this is not a good solution: it will make this not work on linux gdbserver where it should work.

See https://github.com/pwndbg/pwndbg/issues/2467 for more info. What we should do is to check if fsbase/gsbase registers are there and fetch them only if they are there :).

disconnect3d:
@AndersFelde Can you fix this PR so we check the availability of the register?

AndersFelde:
@disconnect3d updated now. Is that a better approach, or do you want me to add try/except to the call inside _fs_gs_helper function?

disconnect3d:
@AndersFelde Instead of doing a try except, where we may miss one or another exception now (and miss bugs in the future) we should check if fsbase/gsbase registers are available before trying to fetch them.

AndersFelde:
@disconnect3d what do you think now? I added the `regsiter_exists` function. It works when i test it on linux and windows

disconnect3d (approved):

pwndbg/gdblib/regs.py
```diff
@@ -246,6 +255,9 @@ def _fs_gs_helper(self, regname: str, which: int) -> int:
         """Supports fetching based on segmented addressing, a la fs:[0x30].
         Requires ptrace'ing the child directory if i386."""
 
+        if not register_exists(regname):
```
disconnect3d: Hmm I think this should be only executed for x86-64 case, so within the if that is in line 261.

disconnect3d: This is probably why the tests on CI fail after this change 😅

disconnect3d (changes requested):
See comment - we need to move the check into the if. Also, pls fix lint issues (u cna run `./lint.sh -f` locally to autofix them)

AndersFelde:
@disconnect3d i've rewritten the check, and moved it inside the if-statement. When i run the tests locally with docker they all pass, but i see that some of them are not passing with github CI. I am also not able to see how the failing tests are related to my change at all?

disconnect3d:
@AndersFelde the test failures on CI are unrelated. They were uncovered by a recent change that made it so that XFAIL are accounted as test failures. I fixed like 6 of them but those 2 were left. I hope to find some time in upcomming days to fix them.. :)

disconnect3d (approved):

patryk4815:
@disconnect3d Please remember changing something in `gdblib` must be ported also to `aglib`
Right now we have differences, and this bug will reopen after removing gdblib xD
image

PR #2480, opened 2024-10-22
Port the `next` command family to the Debugger-agnostic API

mbrla0:
This PR ports the `next` family of commands and the `pwndbg.gdblib.next` module to the Debugger-agnostic async execution control interface introduced in #2469, allowing them to run under both GDG and LLDB.

Due to how the execution control interface is implemented, these commands should retain their exact pre-port behavior when run under GDB, and exhibit near-identical behavior when run under LLDB.

Unrelated to this PR, but that I still feel deserves to be noted is that some of the `next` commands display the context manually, which leads to the context being displayed twice under both GDB and LLDB. Is there any reason for that?

### DIFF
--- a/pwndbg/dbg/lldb/__init__.py
+++ b/pwndbg/dbg/lldb/__init__.py
@@ -683,6 +683,12 @@ class LLDBProcess(pwndbg.dbg_mod.Process):
         )
 
     @override
+    def stopped_with_signal(self) -> bool:
+        return self.process.GetState() == lldb.eStateStopped and any(
+            (thread.GetStopReason() == lldb.eStopReasonSignal for thread in self.process.threads)
+        )
+
+    @override
     def evaluate_expression(self, expression: str) -> pwndbg.dbg_mod.Value:
         value = self.target.EvaluateExpression(expression)
         opt_out = _is_optimized_out(value)
--- a/pwndbg/aglib/__init__.py
+++ b/pwndbg/aglib/__init__.py
@@ -22,6 +22,7 @@ def load_aglib():
     import pwndbg.aglib.heap
     import pwndbg.aglib.memory
     import pwndbg.aglib.nearpc
+    import pwndbg.aglib.next
     import pwndbg.aglib.proc
     import pwndbg.aglib.qemu
     import pwndbg.aglib.regs as regs_mod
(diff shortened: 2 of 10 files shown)


mbrla0:
The Ubuntu 24.04 tests specifically keep getting cancelled because they take too long. I wonder if there's something about Ubuntu 24.04 that makes one of the next commands take too long to execute.

image

disconnect3d (approved):

disconnect3d:
I'll try to run those tests on bare metal and see if it works.

mbrla0:
Okay, I can confirm that in Ubuntu 24.04, at least in Docker, this PR causes a _significant_ performance regression in `test_command_nextproginstr`. Interestingly, though, this only seems to happen in Ubuntu 24.04, and not in either Ubuntu 22.04 or Arch. Also very interesting to note is that `test_command_untilasm_x64`, which also uses the new functionality, doesn't have the same regression.

Left: Arch
Right: Ubuntu 24.04
image

disconnect3d:
@mbrla0 hmm we should see why is that :D

mbrla0:
Alright! Found the issue.

I forgot the `bp.remove()` in `break_next_branch` on `pwndbg/aglib/next.py`, meaning the breakpoints for `break_next_branch` weren't getting removed, and GDB was struggling to keep up with all of them. So, good thing we hit that performance regression, in fact, otherwise I wouldn't have noticed my mistake there.

Fixing that brings the test time down from ~2.5m to about 34.5s, which is in line with the current upstream version, which in my machine takes about 36s.

mbrla0:
Also, I'm really glad to see the debugger-agnostic version being just as fast as the original in that test, which is quite lengthy, at about 4500 breakpoints.

pwndbg/dbg/lldb/__init__.py
```diff
@@ -682,6 +682,12 @@ def alive(self) -> bool:
             and self.process.GetState() != lldb.eStateDetached
         )
 
+    @override
+    def stopped_with_signal(self) -> bool:
+        return self.process.GetState() == lldb.eStateStopped and any(
+            (thread.GetStopReason() == lldb.eStopReasonSignal for thread in self.process.threads)
```
disconnect3d, line 688: Hmm... Can't we get the current stopped thread stop reason instead of checking all of them in LLDB?

PR #2481, opened 2024-10-23
Fix writing to PC under LLDB

mbrla0:
The regular register writing flow for LLDB, which relies on the`register write` command, partially fails when we try to write to the PC. While the write does go through and the PC does change, in at least some instances[^1] we immediately lose access to our instance of `SBFrame`. The abstractions in the Debugger-agnostic API are not designed to handle the frame object becoming invalid upon a register write, and re-validating it would require quite a bit of legwork on the LLDB implementation side, and so this is a scenario we would very much like to avoid.

This PR circumvents that issue by adding a check to `LLDBFrame.reg_write` that looks for writes to the PC and calls the `SBFrame::SetPC` method, instead of going through the more general `register write` command. This allows us to keep our access to the original `SBFrame` object, without having to deal with the complications of having to re-validate our frame.

[^1]: `register write` causes frames obtained with `_any_bottommost_frame` to become invalid upon completion, at least in LLDB 19.

### DIFF
--- a/pwndbg/dbg/lldb/__init__.py
+++ b/pwndbg/dbg/lldb/__init__.py
@@ -24,6 +24,7 @@ import pwndbg
 import pwndbg.lib.memory
 from pwndbg.aglib import load_aglib
 from pwndbg.dbg import selection
+from pwndbg.lib.regs import reg_sets
 
 T = TypeVar("T")
 
@@ -95,6 +96,7 @@ class LLDBFrame(pwndbg.dbg_mod.Frame):
     proc: LLDBProcess
 
     def __init__(self, inner: lldb.SBFrame, proc: LLDBProcess):
+        assert inner.IsValid()
         self.inner = inner
         self.proc = proc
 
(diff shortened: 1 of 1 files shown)


pwndbg/dbg/lldb/__init__.py
```diff
@@ -119,6 +121,12 @@ def reg_write(self, name: str, val: int) -> bool:
         if val < 0:
             raise RuntimeError("Tried to write a register with a negative value")
 
+        # Writing to the PC using the normal register write flow causes the
+        # inner object to be automatically invalidated by LLDB, so we have to
+        # handle jumps manually using SBFrame::SetPC.
+        if name.lower() == reg_sets[pwndbg.aglib.arch.name].pc:
```
disconnect3d, line 127: is it okay that we are using lower only here and not below when we do rename_register?

pwndbg/dbg/lldb/__init__.py
```diff
@@ -119,6 +121,12 @@ def reg_write(self, name: str, val: int) -> bool:
         if val < 0:
             raise RuntimeError("Tried to write a register with a negative value")
 
+        # Writing to the PC using the normal register write flow causes the
+        # inner object to be automatically invalidated by LLDB, so we have to
+        # handle jumps manually using SBFrame::SetPC.
```
disconnect3d, line 126: You mean `SBFrame.SetPC` right? :p

PR #2482, opened 2024-10-25
Update Anthropic API

jkub6:
Trying to use an Anthropic AI model has given me numerous errors. This updates the API to the Messages API compatible with modern models.


disconnect3d:
@jkub6 pls fix lint issues

PR #2487, opened 2024-11-02
Port mmap and mprotect to the Debugger-agnostic API

mbrla0:
This PR ports the `mmap` and `mprotect` commands to the Debugger-agnostic API. Aditionally, it ports the `shellcode` module, required by both commands, to `aglib`.

It also adds `ExecutionController.continue` cancellation checks to GDB, which allows us to implement the breakpoint ignore behavior present in `shellcode` in a less invasive way, as having cancellation support allows us to skip through unexpected breakpoints as they happen, whereas before we had to disable all breakpoints before running the `shellcode` command.

### DIFF
--- a/pwndbg/aglib/shellcode.py
+++ b/pwndbg/aglib/shellcode.py
@@ -102,7 +102,7 @@ async def exec_shellcode(
     page = pwndbg.aglib.vmmap.find(starting_address)
     assert page is not None
 
-    clearance = page.end - len(blob) - 1
+    clearance = page.end - starting_address - len(blob) - 1
     if clearance < 0:
         # The page isn't large enough to hold our shellcode.
         raise RuntimeError(
--- a/pwndbg/aglib/shellcode.py
+++ b/pwndbg/aglib/shellcode.py
@@ -1,0 +1,163 @@
+"""
+Shellcode
+
+This module implements functionality that allows for the execution of a small
+amount of code in the context of the inferior.
+
+"""
+
+from __future__ import annotations
+
+from asyncio import CancelledError
+
+import pwnlib.asm
+import pwnlib.shellcraft
+
+import pwndbg
+import pwndbg.aglib.arch
+import pwndbg.aglib.memory
+import pwndbg.aglib.regs
+import pwndbg.aglib.vmmap
+from pwndbg.dbg import BreakpointLocation
+from pwndbg.dbg import ExecutionController
+
+if pwndbg.dbg.is_gdblib_available():
+    import pwndbg.gdblib.prompt
+
+
+def _get_syscall_return_value():
+    """
+    Reads the value corresponding to the return value of a syscall that has
+    just returned.
+    """
+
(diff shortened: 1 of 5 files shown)


mbrla0:
Weird that `codecov` thinks these aren't being tested, when both `test_mmap.py` and `test_mprotect.py` exist. Maybe that's because we never call any of the code in those directly from any of the tests, only through `gdb.execute()`?

pwndbg/aglib/shellcode.py
```diff
@@ -0,0 +1,163 @@
...
+    # Make sure the blob fits in the rest of the space we have in this page.
+    #
+    # NOTE: Technically, we could actually use anything from the whole page to
+    # all of the pages currently mapped as executable for this. There is no
+    # technical limitation stopping us from doing that, but seeing as doing it
+    # is harder to make sure it works correctly, we don't (for now, at least).
+    page = pwndbg.aglib.vmmap.find(starting_address)
+    assert page is not None
+
+    clearance = page.end - len(blob) - 1
```
disconnect3d: Shouldn't this be:
```suggestion
    clearance = page.end - starting_address - len(blob) - 1
```

pwndbg/aglib/shellcode.py
```diff
@@ -0,0 +1,163 @@
...
+    # The continue we use here will trigger an event that would get the context
+    # prompt to show, regardless of the circumstances. We don't want that, so
+    # we preserve the state of the context skip.
+    if pwndbg.dbg.is_gdblib_available():
+        would_skip_context = pwndbg.gdblib.prompt.context_shown
+
+    # Execute.
+    target_address = starting_address + len(blob)
+    bp = pwndbg.dbg.selected_inferior().break_at(BreakpointLocation(target_address), internal=True)
+    while True:
```
disconnect3d, line 126: What is this logic exactly doing? Shouldn't we need to continue until one breakpoint is hit? Why do we need a loop?

How do we know which breakpoint was hit here and if this was the one we were waiting for?

mbrla0: > Why do we need a loop?
> How do we know which breakpoint was hit here and if this was the one we were waiting for?

`ExecutionController.cont` raises `CancellationError` if we hit any breakpoints other than the one passed to it as an argument. We use the loop to retry `ec.cont` if it gets cancelled (i.e. it stops before we hit `bp`) and `disable_breakpoints=True`. This is meant to implement the case where `disable_breakpoints=True` with a behavior that's close to the `gdblib` version of this module[^1].

When `disable_breakpoints=False`, the loop should only ever execute once. Either it hits the breakpoint we expect, or the execution controller gets cancelled and the command finishes running (GDB) or the Pwndbg CLI returns control to the user (LLDB).

[^1]: The one caveat to this being that right now `ExecutionController` doesn't distinguish between breakpoints and other types of stops, and will lump them all together as "cancellations". That can be fixed fairly easily later by just including that information along with the `CancellationError`.

PR #2488, opened 2024-11-02
Add better handling for exceptions originating in execution controllers under LLDB

mbrla0:
Right now, exceptions thrown from inside execution controller code (coroutines started with `dispatch_execution_controller`) will cause the LLDB Pwndbg CLI to terminate, and kill the debugging session. This is inconsistent with the behavior of GDB Pwndbg, which merely shows a pretty stack trace, but otherwise lets the debugging session continue.

This PR aligns the behavior of LLDB Pwndbg with the behavior of GDB Pwndbg on that front, with the addition of a warning message. I feel the message is warranted, as while errors in execution controllers might not be fatal, they certainly leave the inferior in a rather inconsistent state. And while this isn't any different than how a failure in these commands would leave inferior state in GDB, I think it doesn't hurt to be explicit about it.

### DIFF
--- a/pwndbg/dbg/lldb/repl/__init__.py
+++ b/pwndbg/dbg/lldb/repl/__init__.py
@@ -351,13 +351,28 @@ def run(startup: List[str] | None = None, debug: bool = False) -> None:
 
         # At this point, the last command might've queued up some execution
         # control procedures for us to chew on. Run them now.
+        coroutine_fail_warn = False
         for process, coroutine in dbg.controllers:
             assert driver.has_process()
             assert driver.process.GetUniqueID() == process.process.GetUniqueID()
 
-            driver.run_coroutine(coroutine)
+            try:
+                driver.run_coroutine(coroutine)
+            except Exception:
+                # We treat exceptions coming from the execution controllers the
+                # same way we treat exceptions coming from commands.
+                pwndbg.exception.handle()
+                coroutine_fail_warn = True
 
         dbg.controllers.clear()
+
+        if coroutine_fail_warn:
+            print(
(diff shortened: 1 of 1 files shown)


pwndbg/dbg/lldb/repl/__init__.py
```diff
@@ -351,14 +351,29 @@ def handle_sigint(_sig, _frame):
...
             assert driver.process.GetUniqueID() == process.process.GetUniqueID()
 
-            driver.run_coroutine(coroutine)
+            try:
+                driver.run_coroutine(coroutine)
+            except Exception:
+                # We treat exceptions coming from the execution controllers the
+                # same way we treat exceptions coming from commands.
+                pwndbg.exception.handle()
+                coroutine_fail_warn = True
```
disconnect3d, line 365: Should this break the `for process, coroutine in dbg.controllers:` loop?

disconnect3d:
Added a small comment, but otherwise LGTM.

pwndbg/dbg/lldb/repl/__init__.py
mbrla0: Well, I don't think there's a straight up right answer about what we should do there. On the one hand breaking there would lead to a single easier to read error, but on the other it will stop all the other accumulated ECs from executing, which might obfuscate other errors or leave things in an even more inconsistent state.

Currently, though, there is no case in Pwndbg where breaking there would make a difference, as no commands ever schedule more than one EC. But I think this is something we should revisit in the future, for sure, and by then it should be clearer what the best approach is.

PR #2493, opened 2024-11-04
setup.sh: fix Arch install (add missing libxxhash required by gdb)

disconnect3d:
Our Arch Linux CI started to fail recently with the following error:

> 4.562 gdb: error while loading shared libraries: libxxhash.so.0: cannot open shared object file: No such file or directory

It seems this library should be a dependency of GDB on Arch, but for some reason it is not? This commit/PR adds this library to be installed with Pacman manually/explicitly.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -77,7 +77,7 @@ install_pacman() {
     if [[ "$answer" == "y" ]]; then
         sudo pacman -Syu || true
     fi
-    sudo pacman -S --noconfirm --needed git gdb python which debuginfod curl
+    sudo pacman -S --noconfirm --needed git gdb python which debuginfod curl xxhash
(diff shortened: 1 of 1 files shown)


disconnect3d:
cc: @anthraxx

disconnect3d:
Confirmed with @anthraxx that this is a packaging bug in Arch Linux: the GDB started to depend on libxxhash which is not in the GDB package depends array (needs to be added there).

They will fix hopefully today.

disconnect3d:
I'll skip merging this for now.

disconnect3d:
Fwiw, in case anyone wonders how is libxxhash used by GDB: https://github.com/bminor/binutils-gdb/commit/2299dfd4ba96c6852db862f6ec1b96880ecd6c0c

PR #2504, opened 2024-11-09
fix: Don't report uncompressed kallsyms until actually detected

### LINKED ISSUE #2503: uncompressed kallsyms incorrectly listed as detected when it's not
While doing some more testing for the uncompressed kallsyms change, I ran into a bug when both the compressed token table can't be found (due to not having set `echo 0 | sudo tee /proc/sys/kernel/yama/ptrace_scope`) and the kernel also not being patched with uncompressed:

```
pwndbg> klookup cpuidle_idle_call
Detected Uncompressed Kallsyms
Failed to find kallsyms
Exception occurred: klookup: 'Kallsyms' object has no attribute 'end_of_kallsyms_names_uncompressed' (<class 'AttributeError'>)
```

It shouldn't say it detected Uncompressed Kallsyms and also shouldn't subsequently lookup the result if it doesn't find the uncompressed.

I have a fix for this that I'll send a PR for.

fidgetingbits:
Fixes #2503 

@jetchirag

### DIFF
--- a/pwndbg/gdblib/kernel/kallsyms.py
+++ b/pwndbg/gdblib/kernel/kallsyms.py
@@ -53,11 +53,13 @@ class Kallsyms:
         self.is_big_endian = None
 
         self.token_table = self.find_token_table()
-        self.is_uncompressed = self.token_table is None
+        self.is_uncompressed = False
 
-        if self.is_uncompressed:
+        if self.token_table is None:
+            if not self.find_names_uncompressed():
+                return
             print(M.info("Detected Uncompressed Kallsyms"))
-            self.find_names_uncompressed()
+            self.is_uncompressed = True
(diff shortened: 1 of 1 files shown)


jetchirag:
@fidgetingbits Were you able to test this?

Since `find_names_uncompressed` does not return anything on successful execution, the if condition is failing all the time.

disconnect3d:
@jetchirag nice catch. I merged ur PR since it makes sense to me. @fidgetingbits pls double check it if u can :)

fidgetingbits:
thanks, nice catch. I actually had tested a slightly different version of this PR, but it originally wasn't relying on the return, and then I switched to the return last minute after I saw the return None, and unfortunately didn't retest.

Issue #2105, opened 2024-03-30
Implement strings command


disconnect3d:
It would be nice to have `strings [<page[s]...>]` command that would act similarly as `strings <binary>` in CLI, but would search strings in the debugged process memory.

B1N4RY-P4R45173:
Hello @disconnect3d 

I am thinking of getting the page size with get_memory_page_size() and then for each page 
```
address = int(page, 16)
data = pwndbg.memory.read(address, page_size)
print(f"\nStrings found in memory at {hex(address)} (min_length={min_length}):")
extract_strings(data, min_length)
```

disconnect3d:
You can get virtual memory maps via `pwndbg.aglib.vmmap.get()`, then iterate over the returned tuple of pages and do `pwndbg.aglib.memory.read(page.start, page.size)` and process the returned data to return strings.

Now the command interface could be as follows:
```
strings      # dump all strings with at least N bytes long (N = default = 4?)
strings [pagename filter] [--save-as=<file>]
```

The `[pagename filter]` can be similar as in `search` command.

disconnect3d:
Done in #2505! Thanks to @kapiw04  :)

PR #2505, opened 2024-11-09
Strings command

kapiw04:
Solves Issue #2105 

Added a `string` command.
```
pwndbg> strings -h
usage: strings [-h] [-n N] [--save-as SAVE_AS] [mapping_name]

Extracts and displays ASCII strings from all readable memory pages of the debugged process.

positional arguments:
  mapping_name       Mapping to search [e.g. libc]

options:
  -h, --help         show this help message and exit
  -n N               Minimum length of ASCII strings to include (default: 4)
  --save-as SAVE_AS  Sets the filename for the output of this command [e.g. --save-as='out.txt']
  ```

### CODE REVIEW

File: pwndbg/commands/strings.py [hunk 1]
```diff
@@ -0,0 +1,69 @@
...
+import pwndbg.commands
+import pwndbg.aglib.memory
+from pwndbg.commands import CommandCategory
+
+import re
+
+from pwndbg.lib.memory import Page 
+
+if pwndbg.dbg.is_gdblib_available():
+    import gdb # type: ignore
```
disconnect3d: Is this needed? (cc: @mbrla0)

kapiw04: Actually no, I accidentally left it. I fixed this unused import and some of trailing whitespaces in the newest [commit](https://github.com/pwndbg/pwndbg/pull/2505/commits/cb3fcbe6e297ff1fba5d0702647ad99ae4105781).

pwndbg/commands/strings.py
```diff
@@ -0,0 +1,66 @@
...
+parser = argparse.ArgumentParser(description="Extracts and displays ASCII strings from all readable memory pages of the debugged process.")
+
+parser.add_argument(
+    "-n",                 type=int,            default=4,    help="Minimum length of ASCII strings to include")
+parser.add_argument(
+    "mapping_name",       type=str, nargs="?", default=None, help="Mapping to search [e.g. libc]")
+parser.add_argument(
+    "--save-as",          type=str,            default=None, help="Sets the filename for the output of this command [e.g. --save-as='out.txt']")
+
+@pwndbg.commands.ArgparsedCommand(parser, command_name="strings", category=CommandCategory.LINUX)
```
disconnect3d:
```suggestion
@pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.LINUX)
```
Unless I am mistaken this is only needed when command name differs from function name (e.g. `a_b` vs `a-b`). I believe `ArgparsedCommand` will automagically set command name to the `func.__name__`.

pwndbg/commands/strings.py
```diff
@@ -0,0 +1,66 @@
...
+        readable_pages = [m for m in readable_pages if mapping_name in m.objfile]
+
+    if save_as:
+        # create new file if one does not exist
+        if not os.path.exists(save_as):
+            with open(save_as, 'w'):
+                pass
+        else:
+            # Remove content from a file
+            open(save_as, 'w').close()
```
disconnect3d: I think the lines 37-44 can be removed and line 65 can just do `with open(save_as, "w") as f:`

pwndbg/commands/strings.py
```diff
@@ -0,0 +1,66 @@
...
+        strings: List[bytes] = re.findall(rb'[ -~]{%d,}' % N, data)
+        decoded_strings: List[str] = [s.decode('ascii', errors='ignore') for s in strings]
+
+        if not save_as:
+            for string in decoded_strings:
+                print(string)
+            continue
+
+        with open(save_as, "a") as f:
+           f.writelines(string + "\n" for string in decoded_strings)
```
disconnect3d: Huh, TIL about this function
image

Normally, I'd do `f.write('\n'.join(decoded_strings))` though this computes the super long string and here you pass in a generator that will be used to write into the file. I guess that's probably better then the join one.

pwndbg/commands/strings.py
```diff
@@ -0,0 +1,66 @@
...
+        # create new file if one does not exist
+        if not os.path.exists(save_as):
+            with open(save_as, 'w'):
+                pass
+        else:
+            # Remove content from a file
+            open(save_as, 'w').close()
+
+    for page in readable_pages:
+        count = page.end - page.start
```
disconnect3d: image
U can just use `page.memsz` instead
(don't ask me why did we name it this way, I think we should refactor it to `page.size` lol)

pwndbg/commands/strings.py
```diff
@@ -0,0 +1,66 @@
...
+import re
+
+from pwndbg.lib.memory import Page
+
+parser = argparse.ArgumentParser(description="Extracts and displays ASCII strings from all readable memory pages of the debugged process.")
+
+parser.add_argument(
+    "-n",                 type=int,            default=4,    help="Minimum length of ASCII strings to include")
+parser.add_argument(
+    "mapping_name",       type=str, nargs="?", default=None, help="Mapping to search [e.g. libc]")
```
disconnect3d: Ideally, we probably want to accept multiple mappings here, so one can execute it as `libc stack heap` or `libc libstd libm` etc.

disconnect3d (changes requested):
Generally lgtm, but see comments for minor improvements :)

pwndbg/commands/strings.py
kapiw04: I figured it's more memory efficient to use `f.writelines()` :)

disconnect3d (approved):

pwndbg/commands/strings.py
```diff
@@ -0,0 +1,51 @@
...
+parser.add_argument(
+    "page_names", type=str, nargs="*", default=[], help="Mapping to search [e.g. libc].\nCan be used with multiple mappings [e.g libc heap stack]")
+parser.add_argument(
+    "--save-as", type=str, default=None, help="Sets the filename for the output of this command [e.g. --save-as='out.txt']")
+
+@pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.LINUX)
+@pwndbg.commands.OnlyWhenRunning
+def strings(n: int=4, page_names: List[str]=[], save_as: str=None):
+    # Extract pages with PROT_READ permission
+    readable_pages: List[Page] = [page for page in pwndbg.aglib.vmmap.get() if page.read]
```
disconnect3d, line 37: I guess this could just be a generator expression instead of list comprehension, or, even inlined in line 28, but its good to go for now.

disconnect3d:
Applied few fixes here, gonna merge this if it passes CI now

disconnect3d:
The CI failures are irrelevant (seems that some of our Go tests are flaky :().

Thanks for contributing!

Issue #2506, opened 2024-11-09
Addressify or unhexlify little endian function and command


disconnect3d:
It would be nice to have a function such that we pass it a `00 70 75 c1  cd ef 59 00` data and it returns a pointer like `0x0059efcdc1757000`.

This could be helpful when an address is copied right from a hexdump that can be printed by Pwntools' `DEBUG` log prints etc.

This should probably be both a debugger function like `$addressify(...)` such that it can be used with commands as an argument, like `telescope $addressify('00 70 75 c1  cd ef 59 00')` and a command itself, so one can decode such data and just get the output (which GDB saves in `$x` later on).

I am not sure what is the best name for such command/function. Some name ideas from chatgpt:
<img width="631" alt="image" src="https://github.com/user-attachments/assets/465836e6-561b-4389-b590-6072712b2ae7">

disconnect3d:
We could also have a command like `unhexdump` where u pass in a hexdump and it prints out all valid pointers from it along with their offsets on the hexdump? :D

disconnect3d:
OTOH:
- in pwntools script one can just do `u64(data[x:y])` and if they need multiple dumps like that, its easy to do with a loop
- in gdb, knowing an address, one could just do `x/xg $rdi+offset` or sth.

I am wondering if we really need this feature right now. Would appreciate any other feedback/comments here.

disconnect3d:
cc: @gsingh93 @CptGibbon @OBarronCS wdyt?

fidgetingbits:
fwiw I think it seems reasonable to have a command for the cli. Certainly something I've wanted in the past more than once. 

The u64() and x/xg approaches don't apply  so well to when copy and pasting from hexdump outputs where manually calculating the address/offset might be slower and/or contextually not obvious.

As for name I'd probably prefer hex2ptr or similar in that it's more self descriptive. It's also (barely) more ergonomic to construct by voice, as it's made up of real words.

CptGibbon:
This looks like it's already underway, but my two cents is that if you're copy/pasting addresses your workflow could use improvement.

That said I ignore my own advice and copy/paste addresses all the time so 🤷‍♂️ probably a useful feature.

disconnect3d:
Done in #2516

PR #2516, opened 2024-11-11
addressify #2506

0xhebi:
@disconnect3d Maybe something like this for start?


pwndbg/commands/addressify.py
```diff
@@ -0,0 +1,54 @@
...
+
+from typing import Any
+
+import gdb
+
+import pwndbg.color.message as M
+import pwndbg.commands
+
+
+class AddressifyFunction(gdb.Function):
```
jetchirag: btw, you can use existing decorator which extends gdb function. See https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/gdblib/functions.py

disconnect3d: Also pls move it to another file where we define functions - the commands/ dir is just for commands implementations

pwndbg/commands/addressify.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+AddressifyFunction()
+
+
+@pwndbg.commands.Command
+@pwndbg.commands.OnlyWhenRunning
+def addressify(*args: Any) -> None:
+    # Pwndbg command to convert hex string to little-endian address and print the result.
+    if len(args) == 1 and args[0] in ("-h", "--help"):
```
disconnect3d: This should be handled by `ArgparsedCommand`; we don't want any command to be implemented with the `Command` decorator fwiw

pwndbg/commands/addressify.py
```diff
@@ -0,0 +1,54 @@
...
+
+    combined_args = "".join(args).replace(" ", "")
+    result = addressify_common(combined_args)
+    print(M.success(f"{hex(result)}"))
+
+
+def addressify_common(arg: str) -> int:
+    arg = "".join(filter(str.isalnum, arg))
+    if len(arg) % 2 != 0:
+        raise gdb.GdbError("Hex string must contain an even number of characters.")
```
disconnect3d: Overall, we don't want to raise `GdbError`s ever ourselves as:
1) we want commands to be debugger-agnostic on the long term
2) we do `except gdb.Error` here and there (probably a bit less now or wrapped with our own debugger-agnostic exception?) and this is only to catch errors thrown by gdb itself

disconnect3d (changes requested):
See comments; also, pls add some tests for this :)

pwndbg/commands/addressify.py
```diff
@@ -0,0 +1,54 @@
...
+        print("Converts a space-separated hex string to a little-endian address.")
+        print("Example: addressify 00 70 75 c1 cd ef 59 00")
+        return
+
+    combined_args = "".join(args).replace(" ", "")
+    result = addressify_common(combined_args)
+    print(M.success(f"{hex(result)}"))
+
+
+def addressify_common(arg: str) -> int:
```
disconnect3d: The common function can be moved somewhere to pwndbg/lib

0xhebi:
applied proposed changes and added simple unit-test

pwndbg/commands/addressify.py
```diff
@@ -0,0 +1,28 @@
...
+parser = argparse.ArgumentParser(
+    prog="addressify",
+    description="Converts a space-separated hex string to a little-endian address.",
+)
+parser.add_argument(
+    "hex_string", type=str, help="Hexadecimal string to convert (e.g., '00 70 75 c1 cd ef 59 00')."
+)
+
+
+@pwndbg.commands.ArgparsedCommand(parser, command_name="addressify", category=CommandCategory.MISC)
```
disconnect3d:
```suggestion
@pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.MISC)
```
I believe the command name is deduced from the name of the function

pwndbg/commands/addressify.py
```diff
@@ -0,0 +1,28 @@
...
+import argparse
+
+import pwndbg.color.message as M
+import pwndbg.commands
+from pwndbg.commands import CommandCategory
+from pwndbg.lib.common import addressify_common
+
+# Define an argument parser for the command
+parser = argparse.ArgumentParser(
+    prog="addressify",
```
disconnect3d:
```suggestion
```

pwndbg/commands/addressify.py
```diff
@@ -0,0 +1,28 @@
...
+    description="Converts a space-separated hex string to a little-endian address.",
+)
+parser.add_argument(
+    "hex_string", type=str, help="Hexadecimal string to convert (e.g., '00 70 75 c1 cd ef 59 00')."
+)
+
+
+@pwndbg.commands.ArgparsedCommand(parser, command_name="addressify", category=CommandCategory.MISC)
+def addressify(hex_string) -> None:
+    """Pwndbg command to convert hex string to little-endian address and print the result."""
```
disconnect3d: The docstring is not needed here, I believe we auto-fill it in with the description from ArgparsedCommand, at least unless I am mistaken... ;-)

pwndbg/commands/addressify.py
```diff
@@ -0,0 +1,28 @@
...
+    prog="addressify",
+    description="Converts a space-separated hex string to a little-endian address.",
+)
+parser.add_argument(
+    "hex_string", type=str, help="Hexadecimal string to convert (e.g., '00 70 75 c1 cd ef 59 00')."
+)
+
+
+@pwndbg.commands.ArgparsedCommand(parser, command_name="addressify", category=CommandCategory.MISC)
+def addressify(hex_string) -> None:
```
disconnect3d:
```suggestion
def hex2ptr(hex_string) -> None:
```
Let's rename this for now as per @fidgetingbits suggestion

pwndbg/gdblib/functions.py
```diff
@@ -67,3 +68,17 @@ def base(name_pattern: gdb.Value | str) -> int:
         if name in p.objfile:
             return p.vaddr
     raise ValueError(f"No mapping named {name}")
+
+
+@GdbFunction(only_when_running=True)
+def addressify(hex_string: gdb.Value | str) -> int:
```
disconnect3d:
```suggestion
def hex2ptr(hex_string: gdb.Value | str) -> int:
```

How is `gdb.Value` accepted here? what does it do?

pwndbg/gdblib/functions.py
```diff
@@ -67,3 +68,17 @@ def base(name_pattern: gdb.Value | str) -> int:
...
+    """
+    Converts a hex string to a little-endian address and returns the address.
+    Example usage: $addressify("00 70 75 c1 cd ef 59 00")
```
disconnect3d:
```suggestion
    Example usage: $hex2ptr("00 70 75 c1 cd ef 59 00")
```

pwndbg/lib/common.py
```diff
@@ -0,0 +1,16 @@
+from __future__ import annotations
+
+
+# common functions
+def addressify_common(arg: str) -> int:
```
disconnect3d:
```suggestion
def addressify_common(arg: str) -> int:
```

Maybe: unhex? unhexlify? or just hex2ptr, heh (its like binascii.unhexlify, but better, lol)

tests/unit-tests/test_addressify.py
```diff
@@ -0,0 +1,24 @@
+from __future__ import annotations
+
+from pwndbg.lib.common import addressify_common
+
+
+def test_addressify_common_valid_hex():
+    assert addressify_common("00 70 75 c1 cd ef 59 00") == 0x59EFCDC1757000
```
disconnect3d: rename pls

disconnect3d (changes requested):
Added a few suggestions/comments.

In the future I'd also like to see the command to:
1) by default convert/decode to the endianess of the architecture that we currently debug
2) have a flag to decode with given endianess like `--little` `--big` for little and big endian or `--endian [big|little]`?
3) have a flag to `--telescope` the address, or, telescope it by default?
4) support more formats, like: `007075c1cdef5900` or `00 70 75 c1   cd ef 59 00` or when a full or part of `hexdump` output is passed; perhaps this should have a flag like `--long` for long strings, not sure

All of those can be implemented in future PRs, they don't have to be done in this one

pwndbg/gdblib/functions.py
0xhebi: II think when you call `p $hex2ptr("00 70 ...")`  GDB treats "00 70..." as GDB expression rather than a raw string, so expression gets evaluated to `gdb.Value` object

tests/unit-tests/test_hex2ptr.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def test_hex2ptr_common_valid_hex():
+    assert hex2ptr_common("00 70 75 c1 cd ef 59 00") == 0x59EFCDC1757000
+    assert hex2ptr_common("12 34 56 78") == 0x78563412
+
+
+def test_hex2ptr_common_invalid_hex():
+    try:
+        hex2ptr_common("12345")
```
disconnect3d, line 16: https://stackoverflow.com/a/29855337

pwndbg/commands/hex2ptr.py
```diff
@@ -0,0 +1,26 @@
...
+# Define an argument parser for the command
+parser = argparse.ArgumentParser(
+    description="Converts a space-separated hex string to a little-endian address.",
+)
+parser.add_argument(
+    "hex_string", type=str, help="Hexadecimal string to convert (e.g., '00 70 75 c1 cd ef 59 00')."
+)
+
+
+@pwndbg.commands.ArgparsedCommand(parser, command_name="hex2ptr", category=CommandCategory.MISC)
```
disconnect3d:
```suggestion
@pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.MISC)
```

disconnect3d (approved):

pwndbg/commands/hex2ptr.py
```diff
@@ -0,0 +1,26 @@
...
+    description="Converts a space-separated hex string to a little-endian address.",
+)
+parser.add_argument(
+    "hex_string", type=str, help="Hexadecimal string to convert (e.g., '00 70 75 c1 cd ef 59 00')."
+)
+
+
+@pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.MISC)
+def hex2ptr(hex_string) -> None:
+    combined_args = hex_string.replace(" ", "")
```
patryk4815, line 21: just thinking why we have here replace, but not in:
```
@GdbFunction(only_when_running=True)
def hex2ptr(hex_string: gdb.Value | str) -> int:
```
?

0xhebi: ah, I think I had it locally but accidentally deleted it when I was debugging gdb.Value and cleaning up all prints
Do I have to file new issue or I can PR straight away?

patryk4815: yes, new pr

0xhebi:
#2506 referenced this, so can be closed @disconnect3d

PR #2519, opened 2024-11-13
Port a lot of tests to aglib

patryk4815:
Changes:
- remove gdblib where can in tests
- proper jemalloc install in nix
- add new workflow for checking lldb build successfuly
- ~~add new workflow for tests running under nix~~

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -3,11 +3,11 @@ from __future__ import annotations
 import argparse
 import concurrent.futures
 import os
-import random
 import re
 import subprocess
 import sys
 import time
+from pathlib import Path
 from subprocess import CompletedProcess
 from typing import List
 from typing import Tuple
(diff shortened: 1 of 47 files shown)


tests/tests.py
```diff
@@ -271,18 +297,23 @@ def parse_args():
...
+        make_binaries(test_dir_path)
+    else:
+        raise NotImplementedError(args.type)
 
-    test_dir_path = TEST_FOLDER_NAME[args.type]
-    tests_list = getTestsList(
+    tests_list = get_tests_list(
         args.collect_only, args.test_name_filter, gdb_binary, gdbinit_path, test_dir_path
     )
-    ports = open_ports(len(tests_list))
```
disconnect3d, line 287: Why was open_ports removed?

disconnect3d:
Btw this probably means we should name it now... a `aglib-tests` or `debugger-tests` instead of `gdb-tests`? ;)

tests/tests.py
patryk4815: new function for malloc random port introduced `reserve_port` :)

PR #2525, opened 2024-11-13
Remove `pset` as a command and replace it with the `set` LLDB Pwndbg CLI command override

### LINKED ISSUE #2523: Enable `pset` Command Before Binary Execution
### Description: 
Currently, the `pset exception-debugger on` command only works after starting the binary. This limits flexibility, especially when setting configurations ahead of time is necessary, for instance, in automated debugging workflows or when specific debugging settings are needed right from the start.

### Proposed Solution
Enable the `pset` command to apply configurations, like exception-debugger on, before starting the binary. 
This would allow users to configure debugging parameters in advance, ensuring the setup is in place as soon as the binary begins executing.

image

mbrla0:
This PR replaces the `pset` LLDB command with a `set` command override implemented in the LLDB Pwndbg CLI.

For clarity, in this context, a command override is a command that LLDB Pwndbg knows about, but that it doesn't forward to LLDB, instead doing its own processing on it. Other commands in this category include `continue` and `process launch`.

The reason that we choose to implement it this way is that doing so allows us to both execute it before the process starts, and also that it allows us to use the same name `set` as is used in the GDB version of Pwndbg to change Pwndbg settings, as LLDB doesn't allow us to use that name in proper commands.

This should fix #2523, and also remove the need for us to change any of the help strings that have the set command in them when running under LLDB.

### DIFF
--- a/pwndbg/dbg/lldb/pset.py
+++ b/pwndbg/dbg/lldb/pset.py
@@ -8,25 +8,27 @@ import pwndbg.commands
 import pwndbg.lib.config as cfg
 
 
-def pset(name: str, value: str):
+def pset(name: str, value: str) -> bool:
     """
     Parses and sets a Pwndbg configuration value.
     """
     name = name.replace("-", "_")
     if name not in pwndbg.config.params:
         print(message.error(f"Unknown setting '{name}'"))
-        return
+        return False
 
     param = pwndbg.config.params[name]
     try:
         new_value = parse_value(param, value)
     except InvalidParse as e:
         print(message.error(f"Invalid value '{value}' for setting '{name}': {e.message}"))
-        return
+        return False
 
     param.value = new_value
     for trigger in pwndbg.config.triggers[param.name]:
         trigger()
+
+    return True
 
 
 class InvalidParse(Exception):
--- a/pwndbg/dbg/lldb/pset.py
+++ b/pwndbg/dbg/lldb/pset.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import argparse
 from typing import Any
 
 import pwndbg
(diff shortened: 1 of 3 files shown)


disconnect3d:
How does this work? What does the override exactly do? What is the syntax to set a real LLDB variable and a Pwndbg variable after this PR?

See: 
image

mbrla0:
> How does this work? What does the override exactly do?

For Pwndbg to work properly with LLDB, we have to override a few LLDB commands and handle them manually. We do that as part of the Pwndbg CLI, as laid out in #2382.

While the commands we _have_ to do that for all have to do with process lifetime management, it's not restricted to just lifetime commands. This PR uses the same mechanism to implement our own settings command as `set`, which LLDB wouldn't normally let us do. This lets us keep Pwndbg settings commands consistent between LLDB and GDB.

> What is the syntax to set a real LLDB variable and a Pwndbg variable after this PR?

Pwndbg settings would be changed with `set`, as in GDB, and LLDB settings would be changed with the fully spelled-out `settings` command. `set` in LLDB is usually an alias for `settings`. I've added a warning that should catch most cases of Pwndb `set` being used to try to change LLDB settings.

image

Worst comes to worse we can always just rename it back to `pset` with a one-line change, if the `set`/`settings` distinction proves too confusing.

pwndbg/dbg/lldb/pset.py
```diff
@@ -1,49 +1,40 @@
...
         trigger()
 
+    return True
+
 
 class InvalidParse(Exception):
-    pass
+    def __init__(self, message: str):
+        self.message = message
+        super().__init__()
```
disconnect3d: Shouldn't this whole init be:
```suggestion
        super().__init__(message)
```

PR #2527, opened 2024-11-14
fix(jemalloc): More gracefully handle cmd failure due to missing types and symbols

fidgetingbits:
Similar to #2519 fixing jemalloc_heap test by catching exceptions, this adds a bit more robust handling of expected failure cases in the other jemalloc commands, like if required symbols or types aren't found, rather than just throwing exceptions.

### DIFF
--- a/pwndbg/commands/heap.py
+++ b/pwndbg/commands/heap.py
@@ -1588,17 +1588,21 @@ def jemalloc_find_extent(addr) -> None:
 
     addr = int(addr)
 
-    rtree = jemalloc.RTree.get_rtree()
-    extent = rtree.lookup_hard(addr)
-    if extent is None:
-        print(message.error("ERROR: Extent not found"))
-        return
-    # print pointer address first, then extent address then extent information
-    print(f"Pointer Address: {hex(addr)}")
-    print(f"Extent Address: {hex(extent.extent_address)}")
-    print()
+    try:
+        rtree = jemalloc.RTree.get_rtree()
+        extent = rtree.lookup_hard(addr)
+        if extent is None:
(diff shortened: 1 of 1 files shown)


pwndbg/commands/heap.py
```diff
@@ -1641,17 +1650,15 @@ def jemalloc_heap() -> None:
...
-
-    extents = rtree.extents
-    if len(extents) == 0:
-        print(message.warn("No extents found"))
+        extents = rtree.extents
+        if len(extents) == 0:
+            print(message.warn("No extents found"))
+            return
+        for extent in extents:
+            # TODO: refactor so not create copies
```
disconnect3d, line 1658: wdym by copies here?

disconnect3d (changes requested):
pls change the `def jemalloc_find_extent(addr) -> None:` to `def jemalloc_find_extent(addr) -> bool:` to fix lint (mypy) issues

pwndbg/commands/heap.py
fidgetingbits: fwiw this is me refactoring code to add try except only. Original code (including that comment) is by @jetchirag and I haven't reviewed that part closely yet, so not sure.

fidgetingbits:
> pls change the `def jemalloc_find_extent(addr) -> None:` to `def jemalloc_find_extent(addr) -> bool:` to fix lint (mypy) issues

not part of my changes, but will do. fwiw likely worth changing lint.sh to run mypy outside of ci to more easily catch these during dev?

fidgetingbits:
Fixed

disconnect3d (approved):

PR #2530, opened 2024-11-15
hex2ptr - naming and hex replace

0xhebi:
@patryk4815 here

### DIFF
--- a/pwndbg/gdblib/functions.py
+++ b/pwndbg/gdblib/functions.py
@@ -79,5 +79,6 @@ def hex2ptr(hex_string: gdb.Value | str) -> int:
     if isinstance(hex_string, gdb.Value):
         hex_string = hex_string.string()
 
-    address = hex2ptr_common(hex_string)
-    return address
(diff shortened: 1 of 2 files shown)


pwndbg/gdblib/functions.py
```diff
@@ -79,5 +79,6 @@ def hex2ptr(hex_string: gdb.Value | str) -> int:
     if isinstance(hex_string, gdb.Value):
         hex_string = hex_string.string()
 
-    address = hex2ptr_common(hex_string)
-    return address
+    hex_string = hex_string.replace(" ", "")
+    pointer = hex2ptr_common(hex_string)
```
disconnect3d, line 83: Random thought: maybe we want to handle the space removal inside of hex2prt_common? idk

disconnect3d: Merging as i think its good for now; for future if we start to allow other formats (hexdump) we should probably have it all in common func

PR #2533, opened 2024-11-16
Add checks to avoid dereferencing negative values

OBarronCS:
During developing changes for #2483, issues related to dereferencing negative values during calls to `vmmap.find` and colorizing addresses appeared due to changing the type of error that was allowed in the `try/expect` clauses (`gdb.MemoryError`) around GDB/LLDB memory read operations.

This PR adds a sanity check to make sure the address is non-negative when making these calls. I added this check to the top of functions that make these memory calls (vmmap.find). Additionally, to reduce the depth of the call stack in many causes (there's something multiple intermediate functions before the address is passed to a memory read), this check was added to top of the some other functions that logically shouldn't make calls with negative addresses.

### DIFF
--- a/pwndbg/aglib/disasm/arch.py
+++ b/pwndbg/aglib/disasm/arch.py
@@ -316,7 +316,9 @@ class DisassemblyAssistant:
                 # Don't mask immediates - some computations depend on their signed values
                 if op.type is not CS_OP_IMM:
                     op.before_value &= pwndbg.aglib.arch.ptrmask
-                op.symbol = MemoryColor.attempt_colorized_symbol(op.before_value)
+                
+                if op.before_value >= 0:
+                    op.symbol = MemoryColor.attempt_colorized_symbol(op.before_value)
 
                 op.before_value_resolved = self._resolve_used_value(
                     op.before_value, instruction, op, emu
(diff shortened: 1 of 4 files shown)


disconnect3d (commented):
Shouldn't we do `assert address >= 0` (or `assert address >= 0x10000` at start to find more bugs; though this is not great because address can start at 0) instead and patch all places where we incorrectly pass in addresses?

disconnect3d:
@OBarronCS ping!

disconnect3d:
This now conflicts with a PR I merged from @patryk4815 

Should we close this? Should we port anything from here that is left and undone?

patryk4815:
@disconnect3d I did only fix places what cause crash in lldb

OBarronCS:
The fixes in #2544 should encompass all the cases we found bugs in, which is trying to dereference negative immediate operands. 

Some of the changes in this PR add checks to avoid potentially negative dereferences that could arise from other sources by placing the `addr >= 0` check at the leaf functions that read memory. However, given that recent bugs have come from the immediate operands and we don't have concrete examples of negative dereferences causing issues from other sources, I think we can close this PR.

disconnect3d:
Ok, lets close it then :)

PR #2539, opened 2024-11-16
Fix aglib reg_write to correctly handle writing to the "pc" register

### LINKED ISSUE #2537: Investigate regression in tests `test_next_command_doesnt_freeze_crashed_binary[nextproginstr]`
Investigate regression in tests `test_next_command_doesnt_freeze_crashed_binary[nextproginstr]`
After PR https://github.com/pwndbg/pwndbg/pull/2519 this tests is running 200x slower 

image

patryk4815:
Fixes https://github.com/pwndbg/pwndbg/issues/2537

@mbrla0 please look :)

### DIFF
--- a/pwndbg/aglib/regs.py
+++ b/pwndbg/aglib/regs.py
@@ -127,7 +127,8 @@ class module(ModuleType):
         if attr in ("last", "previous"):
             super().__setattr__(attr, val)
         else:
-            pwndbg.dbg.selected_frame().reg_write(attr, int(val))
+            if not pwndbg.dbg.selected_frame().reg_write(attr, int(val)):
+                raise RuntimeError(f"Attempted to write to a non-existent register '{attr}'")
 
     @pwndbg.lib.cache.cache_until("stop", "prompt")
     def __getitem__(self, item: Any) -> int | None:
--- a/pwndbg/aglib/regs.py
+++ b/pwndbg/aglib/regs.py
@@ -13,7 +13,9 @@ from typing import Any
 from typing import Callable
 from typing import Dict
 from typing import Generator
+from typing import Iterator
 from typing import List
+from typing import Set
 from typing import Tuple
 from typing import cast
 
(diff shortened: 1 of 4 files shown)


pwndbg/aglib/regs.py
```diff
@@ -125,7 +125,8 @@ def __setattr__(self, attr: str, val: Any) -> None:
         if attr in ("last", "previous"):
             super().__setattr__(attr, val)
         else:
-            pwndbg.dbg.selected_frame().reg_write(attr, int(val))
+            if not pwndbg.dbg.selected_frame().reg_write(attr, int(val)):
```
patryk4815, line 132: Or inside here we should "remap" if attr == `pc` to correct real register?
Maybe `reg_write` should only accept real registers?

And not modify `gdb.py` / `lldb/__init__.py`

patryk4815: And what about `sp` register? Like we have in `fix()` method:
```py
    def fix(self, expression: str) -> str:
        for regname in set(self.all + ["sp", "pc"]):
            expression = re.sub(rf"\$?\b{regname}\b", r"$" + regname, expression)
        return expression
```

mbrla0: > Maybe reg_write should only accept real registers?

Yeah, maybe, and have `__setattr__` handle the register aliases.

> Or inside here we should "remap" if attr == pc to correct real register?

Personally, I think mapping `pc` to `regs_sets[...].pc` and still doing the sanity check against the registers listed in the reg sets is a cleaner solution, because that means we only ever have to deal with real register names in the Debugger API. Buuuut, both GDB and LLDB accept `pc` as a register name just fine, so it's not like this won't work, if you think it's better this way.

> And what about sp register?

Yeap, this will probably need an exception for `sp`, too..

disconnect3d:
@mbrla0 Thoughts? Shall we merge this or should we change sth here>?

disconnect3d:
@patryk4815 needs fixing conflicts. Shall we also handle SP here too?

disconnect3d:
Let's fix it with the changes here. If we have ideas for better solution, pls open a GH issue about it (could also mention in a comment or sth).

patryk4815:
@disconnect3d please review again

PR #2546, opened 2024-11-19
fix(nix): add missing one_gadget package for devshell

fidgetingbits:
I had to test the onegadget command due to some bold text related to #2157, and noticed the tool didn't exist in the devshell, so added.


patryk4815 (approved):
👍

PR #2547, opened 2024-11-19
Port emulator to aglib api

patryk4815:
done

### DIFF
--- a/pwndbg/aglib/disasm/__init__.py
+++ b/pwndbg/aglib/disasm/__init__.py
@@ -7,11 +7,10 @@ from __future__ import annotations
 
 import collections
 import re
-from typing import Any
 from typing import DefaultDict
 from typing import List
+from typing import Set
 from typing import Tuple
-from typing import Union
 
 import capstone
 from capstone import *  # noqa: F403
@@ -20,6 +19,7 @@ import pwndbg
 import pwndbg.aglib.arch
 import pwndbg.aglib.disasm.arch
 import pwndbg.aglib.memory
+import pwndbg.emu.emulator
 import pwndbg.lib.cache
 from pwndbg.aglib.disasm.arch import DEBUG_ENHANCEMENT
 from pwndbg.aglib.disasm.instruction import ALL_JUMP_GROUPS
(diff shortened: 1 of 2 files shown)


pwndbg/aglib/disasm/__init__.py
```diff
@@ -308,8 +302,16 @@ def can_run_first_emulate() -> bool:
     try:
         from mmap import mmap, MAP_ANON, MAP_PRIVATE, PROT_EXEC, PROT_READ, PROT_WRITE  # isort:skip
 
+        access_mode = PROT_WRITE | PROT_READ | PROT_EXEC
+        if sys.platform == "darwin":
+            # On macOS (Darwin), creating mmap with 'rwx' permissions is blocked
+            # by System Integrity Protection (SIP), which prevents execution of code
+            # from writable and readable memory regions. Therefore, we restrict
+            # the access mode to 'rw' (read and write only).
+            access_mode = PROT_WRITE | PROT_READ
```
disconnect3d, line 311: How does UnicornEngine works on macOS/darwin then? Does it not allocate an RWX memory page on macOS? How do they JIT the code then?

pwndbg/emu/emulator.py
```diff
@@ -574,16 +577,19 @@ def get_uc_mode(self):
...
 
         elif arch in ("arm", "aarch64"):
-            mode |= U.UC_MODE_THUMB if (pwndbg.gdblib.regs.cpsr & (1 << 5)) else U.UC_MODE_ARM
+            mode |= U.UC_MODE_THUMB if (pwndbg.aglib.regs.cpsr & (1 << 5)) else U.UC_MODE_ARM
 
-        elif arch == "mips" and "isa32r6" in gdb.newest_frame().architecture().name():
+        elif (
+            arch == "mips"
+            and pwndbg.dbg.is_gdblib_available()
+            and "isa32r6" in gdb.newest_frame().architecture().name()
```
disconnect3d, line 590: Why gdb check here? does lldb not support mips?

pwndbg/aglib/disasm/__init__.py
patryk4815: I think Unicorn works fine. If you need executable page, first you create `rw`, then change it with mprotect to `rx`.
Only `wx` is not permitted.

image

pwndbg/emu/emulator.py
patryk4815: We'll address the `gdb.newest_frame().architecture().name()` calls later as they appear in many places. For now, this is a 'best effort' implementation 🤔

patryk4815: We just here don't support `MIPS32R6` arch for lldb

PR #2542, opened 2024-11-17
Port a lot of commands into debugger agnostic api

patryk4815:
Port a lot of commands into debug-agnostic api.
Some of them are just partially ported.

Additional changes:
- fix https://github.com/pwndbg/pwndbg/issues/2534 crash

New available from lldb/aglib:
- auxv ✅
- procinfo ✅
- memoize ✅
- pwndbg ✅
- tls (❌ aarch64,  x86_64 ✅)
- errno ✅
- rizin - ❓ don't know how to test
- radare2 - ❓ don't know how to test
- linkmap ✅
- gdt - ❓ don't know how to test
- plist ✅
- asm ✅
- aslr ✅

PR #2554, opened 2024-11-19
Port others to aglib

patryk4815:
others commands from https://github.com/pwndbg/pwndbg/pull/2542

### DIFF
--- a/pwndbg/commands/integration.py
+++ b/pwndbg/commands/integration.py
@@ -2,12 +2,14 @@ from __future__ import annotations
 
 import argparse
 
+import pwndbg.aglib.regs
 import pwndbg.commands
-import pwndbg.gdblib.events
-import pwndbg.gdblib.functions
-import pwndbg.gdblib.regs
+import pwndbg.dbg
 import pwndbg.integration
-import pwndbg.integration.binja
+
+if pwndbg.dbg.is_gdblib_available():
+    import pwndbg.integration.binja
+
 from pwndbg.commands import CommandCategory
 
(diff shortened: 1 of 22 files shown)


pwndbg/commands/integration.py
```diff
@@ -2,12 +2,14 @@
...
 import pwndbg.commands
-import pwndbg.gdblib.events
-import pwndbg.gdblib.functions
-import pwndbg.gdblib.regs
+import pwndbg.dbg
 import pwndbg.integration
-import pwndbg.integration.binja
+
+if pwndbg.dbg.is_gdblib_available():
+    import pwndbg.integration.binja
```
disconnect3d, line 11: what happens here under lldb? does the command[s] work etc?

pwndbg/commands/memoize.py
```diff
@@ -3,6 +3,7 @@
 import argparse
 
 import pwndbg.commands
+import pwndbg.lib.cache
```
disconnect3d, line 6: weird we add this import ( ? )

pwndbg/commands/tls.py
```diff
@@ -84,6 +87,8 @@ def threads(num_threads, respect_config) -> None:
     headers = ["global_num", "name", "status", "pc", "symbol"]
     bold_green = lambda text: pwndbg.color.bold(pwndbg.color.green(text))
 
+    import gdb
```
disconnect3d, line 90: Is this needed? Is this ok? shouldn't this be protected with some if gdblib is available?

PR #2557, opened 2024-11-20
Add nix tests

patryk4815:
Changes:
- make our tests compatible with portable-pwndbg or nix-pwndbg
- fix terminal width to 80 columns in qemu-tests - same as PR https://github.com/pwndbg/pwndbg/pull/2444
- add workflow with nix-pwndbg + ubuntu + running tests
- bump poetry.lock - for better py3.13 compatibility


.github/workflows/tests.yml
```diff
@@ -15,6 +15,63 @@ on:
...
+    timeout-minutes: 20
+    steps:
+    - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9  # @v3
+    - uses: cachix/install-nix-action@08dcb3a5e62fa31e2da3d490afc4176ef55ecd72  # @v30
+      with:
+        nix_path: nixpkgs=channel:nixos-unstable
+
+    - name: Install dependencies
+      run: |
+        ./setup-dev.sh --install-only
```
fidgetingbits, line 36: You could enter the `nix develop` environment instead to also ensure all the tests are working from nix-installed dependencies. 

Assuming `nix develop` there is https://github.com/DeterminateSystems/magic-nix-cache which I believe would help with caching things like custom jemalloc derivation, etc.

patryk4815: Unfortunately if we run tests from `nix develop` they don't work well because `gcc` (from nixpkgs) overwrites libc.so to `/nix/store/..libc.so`, and we have a lot of tests that would need to be fixed because of that.

patryk4815: In next step we could try to fix these tests. Right now I wanted to confirm that the version we are preparing for release passes the tests.

PR #2559, opened 2024-11-21
Fix hi command by casting GDBValue

### LINKED ISSUE #2558: The hi command just doesn't work
### Description

The command doesn't work.

```
pwndbg/pwndbg/aglib/heap/ptmalloc.py:468 in __contains__ 
│    465 │   │   │   next_size_field_address = next._gdbValue[self.__match_renamed_field("size")]  │
│    466 │   │   │   end_address = next_size_field_address if next.prev_inuse else next.address    │
│    467 │   │                                                                                     │
│ ❱  468 │   │   return start_address <= addr < end_address  # type: ignore[operator]              │
│    469                                                                                           │
│    470                                                                                           │
│    471 class Heap:                                                                               │
TypeError: '<=' not supported between instances of 'GDBValue' and 'int'
```

### Steps to reproduce

Run the command on any heap chunk.
```
~> gdb /bin/cat
pwndbg> run
<ctrl-c>
pwndbg> heap
pwndbg> hi <any printed address>
<crash>
```

### My setup

Gdb:      15.2
Python:   3.12.7 (main)
Pwndbg:   2024.08.29 build: a2fa68f7
Capstone: 5.0.1280
Unicorn:  2.1.0

k4lizen:
Closes https://github.com/pwndbg/pwndbg/issues/2558.

### DIFF
--- a/pwndbg/aglib/heap/ptmalloc.py
+++ b/pwndbg/aglib/heap/ptmalloc.py
@@ -452,9 +452,9 @@ class Chunk:
 
     def __contains__(self, addr: int) -> bool:
         """
-        This allow us to avoid extra constructions like 'if strart_addr <= ptr <= end_addr', etc.
+        This allow us to avoid extra constructions like 'if start_addr <= ptr < end_addr', etc.
         """
-        size_field_address = self._gdbValue[self.__match_renamed_field("size")].address
+        size_field_address = int(self._gdbValue[self.__match_renamed_field("size")].address)
         start_address = size_field_address if self.prev_inuse else self.address
 
         next = self.next_chunk()
@@ -462,10 +462,12 @@ class Chunk:
         if next is None:
             end_address = size_field_address + self.real_size
         else:
-            next_size_field_address = next._gdbValue[self.__match_renamed_field("size")].address
+            next_size_field_address = int(
+                next._gdbValue[self.__match_renamed_field("size")].address
+            )
             end_address = next_size_field_address if next.prev_inuse else next.address
 
-        return start_address <= addr < end_address  # type: ignore[operator]
+        return start_address <= addr < end_address
 
 
 class Heap:


k4lizen:
It is a bit unfortunate that every usage of `self._gdbValue` in ptmalloc has to be surrounded by `int()` and/or that this doesn't return an int:
```python
class GDBValue(pwndbg.dbg_mod.Value):
    def __init__(self, inner: gdb.Value):
        self.inner = inner

    @property
    @override
    def address(self) -> pwndbg.dbg_mod.Value | None:
        return GDBValue(self.inner.address)
```

disconnect3d:
cc: @mbrla0 wondering if we can fix it anyway better?

mbrla0:
> cc: @mbrla0 wondering if we can fix it anyway better?

We _can_ do it, but there are trade-offs.

We could change the return type of `address` to `int`, but in general the parts of Pwndbg that use `pwndbg.dbg.Value` still expect it to behave like `gdb.Value`. `gdb.Value.address` returns a `gdb.Value`, so `pwndbg.dbg.Value.address` returns a `pwndbg.dbg.Value`. Not every part of Pwndbg that uses the Value type is covered by `mypy`, as we saw with #2558, so a bug introduced by changing the return type of `address` wouldn't be immediately visible, necessarily.

We could improve the ergonomics of `Value` by adding operations like `<=` and such, but supporting that means we'd also have to decide what to do when the value doesn't have an integer type, or when it makes no sense to compare the types of the elements at the left and the right of the operator. Even if just to throw a somewhat decent exception, it'd still be best for us to do _something_ clearer than just using whatever results from `int(lhs) <= int(rhs)` in GDB and LLDB.

Having to do `int(value)` may be ugly, but it is very explicit in intent, and doesn't come with having to possibly manually check for uses like `value.address.type`, like the former, or with edge cases can that sneak in subtle complexities and jank into the Debugger API, like with the latter.

k4lizen:
> Having to do int(value) may be ugly, but it is very explicit in intent, and doesn't come with having to possibly manually check for uses like value.address.type, like the former, or with edge cases can that sneak in subtle complexities and jank into the Debugger API, like with the latter.

Yeah. To be honest, the fact that `int(value)` has to be done is not as much of a problem as the fact that it isn't clear to the users of the API (e.g. the implementor of the `__contains__`  function) that `int(value)` has to be done. However! It should have been, as this does actually get caught by mypy:
```
+ mypy pwndbg gdbinit.py lldbinit.py pwndbg-lldb.py
pwndbg/aglib/heap/ptmalloc.py: note: In member "__contains__" of class "Chunk":
pwndbg/aglib/heap/ptmalloc.py:468: error: Unsupported operand types for >= ("int" and "object")  [operator]
            return start_address <= addr < end_address
                   ^
pwndbg/aglib/heap/ptmalloc.py:468: error: Unsupported operand types for < ("int" and "Value")  [operator]
            return start_address <= addr < end_address
                                           ^~~~~~~~~~~
Found 2 errors in 1 file (checked 233 source files)
```
the reason it didn't get caught was the `# type: ignore[operator]` line.

k4lizen:
Also if you're curious why `start_address` was `object` and not `Value` it's because of 
```python
start_address = size_field_address if self.prev_inuse else self.address
#                             ^ Value
#                                                                 ^ int(self._gdbValue.address)
```

Issue #2543, opened 2024-11-17
Inconsistent behavior between GDB/LLDB in `break_at` function


patryk4815:
The behavior of the `break_at` function in LLDB is inconsistent with its counterpart in GDB. 
Specifically, the`one_shot` argument is not implemented in LLDB, resulting in unexpected behavior when attempting to create breakpoints that should automatically disable after being hit once.

<img width="768" alt="image" src="https://github.com/user-attachments/assets/02bd9250-3200-4e73-8e73-f96cf11e5026">

mbrla0:
Yeah, I'm aware of that. But we don't actually use `one_shot` anywhere, so we should just get rid of it. It doesn't work properly in GDB either. Plus, all uses of `break_at` in Pwndbg that could benefit from it already do `await ec.cont(bp); bp.remove()` anyway. Implementing it is too much effort for basically no gain.

mbrla0:
Hm, no, nevermind, it's not that simple. Give me a bit to think about it.

mbrla0:
Okay, I've had more time to look at it, and, yeah, my initial judgement seems to have been right. As far as I can tell, all of our existing uses of GDB's `Breakpoint(temporary=True)` can be translated to an `ec.cont()` followed by a `bp.remove()` when porting it to the Debugger API. So, we can just remove `one_shot`.

patryk4815:
Lets remove then 👍

disconnect3d:
Done #2562

PR #2562, opened 2024-11-21
Remove `one_shot` from `Process.break_at`

mbrla0:
As noted in #2543, the `one_shot` parameter in the Debugger-agnostic breakpoint interface is unused by Pwndbg. This PR removes it. Users can use an `await ExecutionController.cont()` followed by a `StopPoint.remove()` instead, which achieves the same effect in all of the use cases we currently have in Pwndbg.


patryk4815 (approved):
👍

PR #2566, opened 2024-11-22
Fix the lexing of arguments in LLDB to match the behavior of gdb

patryk4815:
Right now commands have issue:
```python
def rop(grep: str | None, argument: List[str]) -> None:
    print(f'grep = {repr(grep)}')
    print(f'argument = {repr(argument)}')
    
# pwndbg-lldb> rop --grep 'pop rdi'
# grep = "'pop"
# argument = ["rdi'"]
```

Gdb:
```python
>>> input_string = 'arg1 "arg two" \'arg three\' arg\\ four --arg="1 2"'
>>> gdb.string_to_argv(input_string)
['arg1', 'arg two', 'arg three', 'arg four', '--arg=1 2']
```

Shlex split:
```python
>>> input_string = 'arg1 "arg two" \'arg three\' arg\\ four --arg="1 2"'
>>> shlex.split(input_string)
['arg1', 'arg two', 'arg three', 'arg four', '--arg=1 2']
```

### DIFF
--- a/pwndbg/dbg/lldb/__init__.py
+++ b/pwndbg/dbg/lldb/__init__.py
@@ -4,6 +4,7 @@ import bisect
 import collections
 import os
 import random
+import shlex
 import sys
 from typing import Any
 from typing import Awaitable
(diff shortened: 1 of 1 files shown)


disconnect3d:
@mbrla0 let us know if that is fine pls, I'll merge this as for now

mbrla0:
> @mbrla0 let us know if that is fine pls, I'll merge this as for now

Yeap, this should be perfectly fine.

PR #2567, opened 2024-11-23
Port to aglib: entry

patryk4815:
Changes:
- add decorator `@OnlyWithDbg` to make `gdb/lldb` only commands
- port `entry`

### DIFF
--- a/pwndbg/commands/start.py
+++ b/pwndbg/commands/start.py
@@ -9,25 +9,29 @@ import argparse
 from argparse import RawTextHelpFormatter
 from shlex import quote
 
-import gdb
-
 import pwndbg
+import pwndbg.aglib.elf
+import pwndbg.aglib.proc
+import pwndbg.color.message as M
 import pwndbg.commands
-import pwndbg.gdblib.elf
-import pwndbg.gdblib.symbol
+import pwndbg.dbg
 from pwndbg.commands import CommandCategory
-from pwndbg.dbg import EventType
+from pwndbg.dbg import BreakpointLocation
 
-break_on_first_instruction = False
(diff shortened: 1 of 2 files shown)


disconnect3d:
@mbrla0 can I ask you to review this PR? :)

mbrla0:
Alright. I'll do it in the afternoon.

pwndbg/commands/start.py
```diff
@@ -9,25 +9,33 @@
...
 import pwndbg.commands
-import pwndbg.gdblib.elf
-import pwndbg.gdblib.symbol
+import pwndbg.dbg
 from pwndbg.commands import CommandCategory
-from pwndbg.dbg import EventType
+from pwndbg.dbg import BreakpointLocation
+
+if pwndbg.dbg.is_gdblib_available():
+    import gdb
```
mbrla0, line 22: I don't really love that with `@OnlyWithDbg` we still need to do the whole song and dance of gating off imports depending on the debugger, but It works well enough, and I think it's a good enough solution, all things considered, so I guess it's fine?

mbrla0:
Alrighty, looks okay to me, aside from my small gripe with import gating.

pwndbg/commands/start.py
disconnect3d: Why do we actually need import gdb there?

disconnect3d: Perhaps we can do `if gdb := pwndbg.dbg.get_gdb()`?

patryk4815: I tried do `get_gdb` but `mypy` complain

Issue #2576, opened 2024-11-26
gdbserver bug: Remote 'g' packet reply is too long


patryk4815:
Reproduction for all issues with gdbserver:
- https://github.com/pwndbg/pwndbg/issues/2319
- https://github.com/pwndbg/pwndbg/issues/2573
- https://github.com/pwndbg/pwndbg/issues/1754
- PR?: https://github.com/pwndbg/pwndbg/pull/1755
- https://github.com/pwndbg/pwndbg/issues/619

`bug.py`:
```py
import gdb

RED = "\x1b[31m"
GREEN = "\x1b[32m"
NORMAL = "\x1b[0m"

def event_handler(event):
    try:
        print(GREEN+f'START EVENT: {event.__dict__}'+NORMAL)
        gdb.execute("show language")
    except Exception as e:
        print(RED+f'ERROR: {event.__dict__} {e}'+NORMAL)

gdb.events.new_objfile.connect(event_handler)
```

execve.c:
```c
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>

int main() {
    char *args[] = { "ls", NULL };
    char *env[] = { NULL };
    execve("/bin/ls", args, env);
    return 0;
}
```

How to reproduce:
```console
gcc -o execve execve.c -g -O0
gdbserver 0:1337 ./execve
gdb -ex 'source bug.py' -ex 'target remote 0:1337' -ex 'c'
```

What i see:
```console
Remote debugging using 0:1337
Reading /work/execve from remote target...
warning: File transfers from remote targets can be slow. Use "set sysroot" to access files locally instead.
Reading /work/execve from remote target...
Reading symbols from target:/work/execve...
######## START EVENT: {'new_objfile': <gdb.Objfile filename=target:/work/execve>}
The current source language is "auto; currently c".
Reading /lib64/ld-linux-x86-64.so.2 from remote target...
Reading /lib64/ld-linux-x86-64.so.2 from remote target...
Reading symbols from target:/lib64/ld-linux-x86-64.so.2...
Reading /lib64/3e1b6cb0eebc08cf3ff812eae8a51b4efd684e.debug from remote target...
Reading /lib64/.debug/3e1b6cb0eebc08cf3ff812eae8a51b4efd684e.debug from remote target...
(No debugging symbols found in target:/lib64/ld-linux-x86-64.so.2)
[...]

PR #2577, opened 2024-11-26
Add workaround for gdbserver weird bug

patryk4815:
https://github.com/pwndbg/pwndbg/issues/2576

Disadvantage:  
- ~~Events may not be processed if a single invalid event occurs and no additional events are triggered afterward.~~ Fixed

There are two bugs in gdbserver/GDB:  
1. In certain GDB events (e.g., `new_objfile`, `clear_objfiles`, `executable_changed`), GDB commands cannot be executed when used with gdbserver and a binary performing an `execve` syscall.  
2. Using `gdb.post_event` causes issues with the `gdb.newest_frame` function, resulting in the error: `gdb.error: Selected thread is running.` This error is not true(sic!)

image

### DIFF
--- a/pwndbg/gdblib/events.py
+++ b/pwndbg/gdblib/events.py
@@ -8,12 +8,14 @@ from __future__ import annotations
 
 import sys
 from collections import defaultdict
+from collections import deque
 from enum import Enum
 from enum import auto
 from functools import partial
 from functools import wraps
 from typing import Any
 from typing import Callable
+from typing import Deque
 from typing import Dict
 from typing import List
 from typing import Set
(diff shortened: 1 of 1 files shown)


patryk4815:
@disconnect3d Ready for code review

pwndbg/gdblib/events.py
```diff
@@ -75,6 +77,69 @@ def on_stop(self) -> None:
...
+    Workaround to fix bug in gdbserver: https://github.com/pwndbg/pwndbg/issues/2576
+    """
+    queued_invalid_events: Deque[Callable[..., Any]] = deque()
+
+    def _loop_until_thread_ok():
+        if not queued_invalid_events:
+            return
+
+        if not _is_safe_event_thread():
+            gdb.post_event(_DelayedEventHandler(_loop_until_thread_ok))
```
disconnect3d: Shouldn't this be `gdb.post_event(event_handler)`?

Also the `_DelayedEventHandler` class is redundant - it does not add anything to the code iiuc.

pwndbg/gdblib/events.py
```diff
@@ -163,6 +228,10 @@ def caller(*a: P.args, **kw: P.kwargs) -> None:
     registered[event_handler].setdefault(priority, []).append(caller)
     if event_handler not in connected:
         handle = partial(invoke_event, event_handler)
+
+        if event_handler == gdb.events.new_objfile:
```
disconnect3d, line 224: I bet this can be `event_handler is gdb.events.new_objfile` but idk, maybe not?

pwndbg/gdblib/events.py
```diff
@@ -75,6 +77,69 @@ def on_stop(self) -> None:
...
+        if not _is_safe_event_thread():
+            gdb.post_event(_DelayedEventHandler(_loop_until_thread_ok))
+            return
+
+        while queued_invalid_events:
+            queued_invalid_events.popleft()()
+
+    @wraps(event_handler)
+    def _inner_handler(*a: P.args, **kw: P.kwargs):
+        if not _is_safe_event_packet():
```
disconnect3d: We should probably flip the if/else here, since the case when it is a safe event packet is what will be usually triggered

pwndbg/gdblib/events.py
patryk4815: don't know :)

PR #2580, opened 2024-11-27
Cleanup unused cache for objfile

patryk4815:
Disconnect3d tested it with code from https://github.com/pwndbg/pwndbg/issues/2576
```py
import gdb

RED = "\x1b[31m"
GREEN = "\x1b[32m"
NORMAL = "\x1b[0m"

found = []

def event_handler(event):
    #breakpoint()
    print(GREEN + f'objfile for={event.new_objfile.filename}' + NORMAL)
    if event.new_objfile.filename in found:
        print(RED + f'DUPLICATE for {event.new_objfile.filename}' + NORMAL)
    found.append(event.new_objfile.filename)

gdb.events.new_objfile.connect(event_handler)
```

### DIFF
--- a/pwndbg/gdblib/events.py
+++ b/pwndbg/gdblib/events.py
@@ -16,7 +16,6 @@ from typing import Any
 from typing import Callable
 from typing import Dict
 from typing import List
-from typing import Set
 from typing import TypeVar
 
 import gdb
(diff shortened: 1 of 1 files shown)


disconnect3d:
I'm all hands for this. The `objfile_cache` code comes from 9 years ago and it seems this may have been a workaround for a gdb/gdbserver bug. I didn't check an old GDB version (since it would be hard to run against it too) but it seems this code is redundant and could introduce real bugs into events that should be re-fired for the same objfile being loaded again.

It could be that the idea was to not re-do the same event for the same objfile being loaded twice when sth like `target extended-remote ...` was used on a `gdbserver`. But in practice a specific event handler should implement its own cache like this it is desired.

PR #2585, opened 2024-11-27
Fix #2549: block config.<name> assignments

### LINKED ISSUE #2549: Improve pwndbg.config
In few places/tests we have eg.:
```
pwndbg.config.hexdump_group_width = -1
```
Documentation mention directly setting value:
https://github.com/pwndbg/pwndbg/blob/0caf175699a1d2f627512d3dbe46674611cc181e/DEVELOPING.md?plain=1#L92

But it is invalid usage, it should be probably `pwndbg.config.hexdump_group_width.value = -1`
It is a bit confusing because then we compare this value `pwndbg.config.hexdump_group_width == 1337`  (without `.value` attribute)

Can we somehow prevent usage like `pwndbg.config.hexdump_group_width = -1` and raise error?

disconnect3d:
This commit blocks `config.<param> = <value>` assignments as they should be done via `config.<param>.value = <value>` instead.

Fixes #2549

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -89,7 +89,7 @@ Here's a small snippet of the actual test:
 ```python
 def test_hexdump(start_binary):
     start_binary(BINARY)
-    pwndbg.config.hexdump_group_width = -1
+    pwndbg.config.hexdump_group_width.value = -1
 
     gdb.execute("set hexdump-byte-separator")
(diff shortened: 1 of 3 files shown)


patryk4815:
@disconnect3d one tests is failing, and please fix also documentation 
https://github.com/pwndbg/pwndbg/blob/4fa42857eca094b968d21145914f3634a910d626/DEVELOPING.md?plain=1#L92

Issue #354, opened 2017-10-09
Android: Currently Broken


zachriggle:
Attempting to debug an Android 8.0 phone with GDB 8.0 (on macOS) causes GDB to abort.

zachriggle:
It looks like at least part of this is due to `pwndbg.android.is_android()` not being cached, and eventually we will invoke `pwndbg.android.sysroot` to set the sysroot, which causes `is_android()` to return `False`.

zachriggle:
Apparently the issue is `follow-fork-mode child`.

When a fork occurs, the following is printed to `stdout`:

```
Detaching from process <pid>
Aborted
```

zachriggle:
May be related to https://sourceware.org/bugzilla/show_bug.cgi?id=13584

gsingh93:
According to https://source.android.com/docs/core/tests/debug/gdb, GDB support is deprecated and will be removed soon (in favor of lldb), so maybe we should just close this.

gsingh93:
Closing this, if anyone wants to debug Android they should use LLDB, which unfortunately we currently don't support.

PR #2592, opened 2024-12-01
Cleanup gdblib

patryk4815:
Removed:
- gdblib.android
- gdblib.ctypes
- gdblib.elf
- gdblib.proc
- gdblib.regs
- gdblib.file
- gdblib.remote
- gdblib.memory
- gdblib.nearpc
- gdblib.next
- gdblib.net
- gdblib.qemu
- gdblib.stack
- gdblib.strings
- gdblib.tls
- gdblib.typeinfo


patryk4815:
@mbrla0 do you want do a review 🙃?

pwndbg/gdblib/android.py
mbrla0, line 1: I'm not sure it's safe for us to get rid of this file. While `is_android` is unused, the sysroot check might stil be important for GDB, so we should move it somewhere else, instead of getting rid of it altogether.

Maybe making `pwndbg.dbg.gdb` into a folder and moving these hooks to a `pwndbg.dbg.gdb.hooks` module could work. And that module could also get the rest of the hooks in `pwndbg.gdblib.hooks`.

mbrla0:
> @mbrla0 do you want do a review 🙃?

I do, but I've been a bit busy with university and my end-of-semester tests, so I can't promise I'll be able to do much today. Tomorrow should be okay, though.

pwndbg/gdblib/memory.py
mbrla0, line 1: I'm all for finally removing `gdblib.memory`, but, just to check: Are we sure the `aglib` and `gdblib` versions didn't get out of sync in the meantime?

pwndbg/gdblib/android.py
patryk4815: `sysroot` is inside `is_android` check, so I think it is safe to cleanup 🤔 Later we should add proper lldb-android support?

mbrla0: > `sysroot` is inside `is_android` check

Wait are you sure? It looks to me like it's triggering GDB-wide sidde effects in respose to an event..

pwndbg/gdblib/memory.py
patryk4815: image
I only see in `poke` func, we have events, but in aglib, we don't have. Rest function are in sync

mbrla0:
Honestly, I'm really pleasantly surprised the transition hasn't broken anything major as far as `mypy` can see! Of course, that was the intent from the start, the way I envisioned `aglib` to work, but in my heart of hearts I was still dreading it might.

Are any of the modules affected by this change covered in a blanket `mypy` ignore, though? We might want to give those a more thorough look. Doesn't look like any of them are, but there might still be single-line overrides I've missed.

patryk4815:
In few places I added 'wrapper' for now (later we have to fix that)
```py
def dbg_value_to_gdb(d: pwndbg.dbg_mod.Value) -> gdb.Value:
    from pwndbg.dbg.gdb import GDBValue

    assert isinstance(d, GDBValue)
    return d.inner

def dbg_frame_to_gdb(d: pwndbg.dbg_mod.Frame) -> gdb.Frame:
    # TODO: fix later to aglib
    from pwndbg.dbg.gdb import GDBFrame

    assert isinstance(d, GDBFrame)
    return d.inner


def gdb_frame_to_dbg(d: gdb.Frame | None) -> pwndbg.dbg_mod.Frame | None:
    # TODO: fix later to aglib
    from pwndbg.dbg.gdb import GDBFrame

    if d is None:
        return None

    return GDBFrame(d)
```

pwndbg/gdblib/memory.py
mbrla0: Actually, that is likely to be an important change. We don't support pausing events in the Debugger API yet, but I _can_ add support for it pretty quick. Should let you add equivalents to those lines in `aglib`, in this PR.

pwndbg/gdblib/android.py
patryk4815: This event is only triggered only for `is_android` right? 
I can revert that if you want, but I don't know if anyone cares about Android at this point, since we have poor gdb-support for it anyway :/

mbrla0: I think it's need in order to support Android at all? I'm not sure who's responsible for the Android bits of Pwndbg, I think they could enlighten the matter. Anyway, yeah, as far as I know we don't support Android all that well anyway, so this probably isn't that big of a deal, but it would still be nice to preserve that behavior, just in case, imo.

patryk4815: https://github.com/pwndbg/pwndbg/issues/354#issuecomment-1552214541

pwndbg/gdblib/memory.py
mbrla0: See https://github.com/pwndbg/pwndbg/pull/2593.

patryk4815: Fixed

pwndbg/commands/binder.py
```diff
@@ -74,7 +75,7 @@ def __exit__(self, exc_type, exc_value, exc_tb):
 
 
 # TODO: merge with for_each_entry?
-def for_each_hlist_entry(head, typename, field):
+def for_each_hlist_entry(head: pwndbg.dbg_mod.Value, typename, field) -> Iterator[gdb.Value]:
     addr = head["first"]
     while addr != 0:
```
patryk4815: todo check, cast `addr` to int?

patryk4815: fixed

pwndbg/aglib/proc.py
```diff
@@ -29,7 +29,7 @@
 tid: int
 thread_id: int
 alive: bool
-thread_is_stopped: bool
+# thread_is_stopped: bool
```
disconnect3d:
```suggestion
```

pwndbg/aglib/memory.py
```diff
@@ -115,9 +116,13 @@ def poke(address: int) -> bool:
     if c is None:
         return False
     try:
+        pwndbg.dbg.suspend_events(EventType.MEMORY_CHANGED)
```
disconnect3d, line 123:
```suggestion
        # Suspending mem_changed event during poke speeds up things when vmmaps are explored
        # (e.g. when stepping through remote processes run with QEMU)
        # The suspension prevents the clearing of the disasm instruction cache
        # by `aglib.disasm.clear_on_reg_mem_change`
        pwndbg.dbg.suspend_events(EventType.MEMORY_CHANGED)
```

pwndbg/dbg/gdb.py
```diff
@@ -367,12 +366,12 @@ def vmmap(self) -> pwndbg.dbg_mod.MemoryMap:
...
 
         pages = pwndbg.gdblib.vmmap.get()
-        qemu = pwndbg.gdblib.qemu.is_qemu() and not pwndbg.gdblib.qemu.exec_file_supported()
+        qemu = pwndbg.aglib.qemu.is_qemu() and not pwndbg.aglib.qemu.exec_file_supported()
 
         # Only GDB versions >=12 report permission info in info proc mappings.
         # On older versions, we fallback on "rwx".
         # See https://github.com/bminor/binutils-gdb/commit/29ef4c0699e1b46d41ade00ae07a54f979ea21cc
-        reliable_perms = not (pwndbg.gdblib.qemu.is_qemu_usermode() and gdb_version[0] < 12)
+        reliable_perms = not (pwndbg.aglib.qemu.is_qemu_usermode() and gdb_version[0] < 12)
```
disconnect3d, line 374: > Pwndbg is supported on Ubuntu 22.04, and 24.04 with GDB 12.1 and later

We should make issue for this and fix it later on

disconnect3d (approved):
The `dbg_frame_to_gdb, gdb_frame_to_dbg` are repeated here and there but i guess its temporary, so lgtm

pwndbg/gdblib/memory.py
disconnect3d: Fwiw we discussed this with @patryk4815 and we should probably disable the events even higher, during the memory exploration code, but this is fine here as well

PR #2595, opened 2024-12-04
add vmmap -C option

ksen-lin:
Do the same as with `-C` (`--context`) in `grep`.

p.s. `lint.sh` suggested to update `vmmap()` args in multiline


disconnect3d:
Can you show a screen how this works?

ksen-lin:
sure

ctx

disconnect3d:
Thanks!

PR #2596, opened 2024-12-04
Fixes gdb function help text not showing up

k4lizen:
Before:
image

After:
image

### DIFF
--- a/pwndbg/commands/argv.py
+++ b/pwndbg/commands/argv.py
@@ -76,9 +76,7 @@ def envp(name=None):
 
 
 class argv_function(gdb.Function):
-    """
-    Evaluate argv on the supplied value.
-    """
+    """Evaluate argv on the supplied value."""
 
     def __init__(self) -> None:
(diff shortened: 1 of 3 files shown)


k4lizen:
Lint is enforcing wacky indentation on `hex2ptr` and `bn_eval`:
image

patryk4815:
🤔 Maybe we should strip leading spaces from every line of `self.__doc__`
```py
class _GdbFunction(gdb.Function):
    def __init__(self, func: Callable[..., Any], only_when_running: bool) -> None:
        self.name = func.__name__
        self.func = func
        self.only_when_running = only_when_running
        self.__doc__ = func.__doc__

```

disconnect3d:
Thanks!

Regarding indents/spaces - would have to look at it, but we have multiline docs in commands as well; tho there we use some Argparse magic for that.

disconnect3d:
Anyway, I'd be happy to accept/merge a fix for those indents

PR #2599, opened 2024-12-04
Make setup.sh return non-zero for all errors

koalajoe23:
When building a customized Docker image containing pwndbg using ubuntu:20.04 as base, the build process does not fail even when setup.sh exits because Ubuntu 20.04 isn't supported anymore.

I think setup.sh should make the Docker image build process fail by emitting non-zero return values.


disconnect3d:
This makes a lot of sense; thanks!

Issue #2608, opened 2024-12-07
hexdump should not crash gdb due to OOM when too huge count is passed


disconnect3d:
Well... we should probably check if `count` is sane like <=1GB?

```
pwndbg> hexdump 500 0x5993af715eee
/build/gdb-Dh0pdX/gdb-12.1/gdb/utils.c:712: internal-error: virtual memory exhausted: can't allocate 98490838507246 bytes.
A problem internal to GDB has been detected,
further debugging may prove unreliable.
```

johnnywong0:
Hi, I’d like to try working on this task and would like some input if possible as well.

I’ve added a size limit of 1 gb for the hexdump to prevent crashes when trying to read a large amount of memory. This limit is set to 1 GB (MAX_HEXDUMP_SIZE) for now, and the read function ensures that any read request exceeding this size will raise a ValueError. I also added a test function in the test_hexdump.py to handle large counts gracefully and ensure the program doesn’t crash: 

`MAX_HEXDUMP_SIZE = 1 * 1024 * 1024 * 1024  # 1 GB

def read(addr: int, count: int, partial: bool = False) -> bytearray:
    """read(addr, count, partial=False) -> bytearray

    Read memory from the program being debugged.

    Arguments:
        addr(int): Address to read
        count(int): Number of bytes to read
        partial(bool): Whether less than ``count`` bytes can be returned

    Returns:
        :class:`bytearray`: The memory at the specified address,
        or ``None``.
    """
    if count > MAX_HEXDUMP_SIZE:
        raise ValueError(
            f"Request size too large: {count} bytes, max allowed is {MAX_HEXDUMP_SIZE} bytes"
        )
    return pwndbg.dbg.selected_inferior().read_memory(address=addr, size=count, partial=partial)`

`
def test_hexdump_large_count_handling(start_binary):
    start_binary(BINARY)
    sp = pwndbg.aglib.regs.rsp

    # Test with a large count value that could lead to OOM
    large_count = 2**31

    # Ensure the hexdump doesn't crash
    try:
        result = gdb.execute(f"hexdump $rsp {large_count}", to_string=True)
        print("Hexdump executed successfully with large count.")
        # Optionally assert that the result is reasonable
        assert result.startswith(
            f"+0000 0x{sp:x}"
        )  # Check that the result starts with correct address
    except gdb.error as e:
        # Ensure it raises an expected error or gracefully handles large counts
[...]

PR #2618, opened 2024-12-10
Fix crash in hexdump with large counts

johnnywong0:
Description:

This is an attempt to fix the issue mentioned in [Issue #2608](https://github.com/pwndbg/pwndbg/issues/2608).

Changes:
I added a check to make sure the count value is within reasonable bounds (currently set to 1 GB) before proceeding with the hexdump.
I also added a test to try and verify that the hexdump function handles larger count values without crashing.
Testing: The test should check that large count values don’t cause crashes, but I’m not totally sure it covers everything.

### DIFF
--- a/pwndbg/aglib/memory.py
+++ b/pwndbg/aglib/memory.py
@@ -16,6 +16,7 @@ GdbDict = Dict[str, Union["GdbDict", int]]
 
 
 MMAP_MIN_ADDR = 0x8000
+MAX_HEXDUMP_SIZE = 1 * 1024 * 1024 * 1024  # 1 GB
 
 
 def read(addr: int, count: int, partial: bool = False) -> bytearray:
@@ -32,6 +33,10 @@ def read(addr: int, count: int, partial: bool = False) -> bytearray:
         :class:`bytearray`: The memory at the specified address,
         or ``None``.
     """
+    if count > MAX_HEXDUMP_SIZE:
+        raise ValueError(
+            f"Request size too large: {count} bytes, max allowed is {MAX_HEXDUMP_SIZE} bytes"
+        )
     return pwndbg.dbg.selected_inferior().read_memory(address=addr, size=count, partial=partial)
 
 
--- a/pwndbg/aglib/memory.py
+++ b/pwndbg/aglib/memory.py
@@ -16,7 +16,6 @@ GdbDict = Dict[str, Union["GdbDict", int]]
 
 
 MMAP_MIN_ADDR = 0x8000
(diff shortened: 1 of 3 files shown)


tests/gdb-tests/tests/test_hexdump.py
```diff
@@ -103,3 +103,21 @@ def test_hexdump_saved_address_and_offset(start_binary):
     assert out1 == out2
     assert pwndbg.commands.hexdump.hexdump.last_address == sp + SIZE
     assert pwndbg.commands.hexdump.hexdump.offset == SIZE
+
+def test_hexdump_large_count_handling(start_binary):
+    start_binary(BINARY)
+    sp = pwndbg.aglib.regs.rsp
```
patryk4815:
```suggestion
    sp = pwndbg.aglib.regs.sp
```

patryk4815: `rsp` works only in x86 cpu
`sp` is alias on all architectures

tests/gdb-tests/tests/test_hexdump.py
```diff
@@ -103,3 +103,21 @@ def test_hexdump_saved_address_and_offset(start_binary):
...
+
+    # Test with a large count value that could lead to OOM
+    large_count = 2**31
+
+    # Ensure the hexdump doesn't crash
+    try:
+        result = gdb.execute(f"hexdump $rsp {large_count}", to_string=True)
```
patryk4815:
```suggestion
        result = gdb.execute(f"hexdump $sp {large_count}", to_string=True)
```

patryk4815:
Thank you for your contribution❤️!  It looks like there is a linting issue. You can fix it by running: `./lint.sh --fix`

pwndbg/aglib/memory.py
```diff
@@ -32,6 +33,10 @@ def read(addr: int, count: int, partial: bool = False) -> bytearray:
         :class:`bytearray`: The memory at the specified address,
         or ``None``.
     """
+    if count > MAX_HEXDUMP_SIZE:
+        raise ValueError(
+            f"Request size too large: {count} bytes, max allowed is {MAX_HEXDUMP_SIZE} bytes"
+        )
```
patryk4815: Please move these checks into the` hexdump.py` command:
- Set `MAX_HEXDUMP_SIZE` to `5`MB. Use print with an error color to indicate when the size exceeds the limit.
- Add a validation check for the `addr` parameter using: 
```
if not pwndbg.aglib.memory.peek(addr):
    print(message.error("Invalid address %#x" % addr))
    return
```

Issue #2621, opened 2024-12-13
Add hijackfd command to modify the file descriptor of a process


disconnect3d:
Add hijackfd command, similar to the one from gef: https://gef-legacy.readthedocs.io/en/latest/commands/hijack-fd/

esseciego:
I'd like to work on this issue!

patryk4815:
@esseciego  the feature is already implemented :( - https://github.com/pwndbg/pwndbg/pull/2623
but it doesn't have unit tests. If you'd like to add them – feel free, that would be awesome! :)

esseciego:
@patryk4815 O whoops. Sure I'll add the unit tests!

esseciego:
@patryk4815 Would I need to create gdb-tests, unit-tests, or both?

patryk4815:
@esseciego only gdb-tests would be enough

woopy098:
Hi @patryk4815 has the unit test been implemented yet? I'd like to contribute to it!

patryk4815:
@woopy098 Hi. unit tests are not implemented. Don't know if @esseciego is still working?

woopy098:
@patryk4815 i can get on with that then.

ghost:
is this issue solved ?

patryk4815:
Hey @gooddev-cpu, unit tests are still not implemented.

disconnect3d:
Hi, as it was already said, this feature has been implemented in #2623 but we still need tests for it.

@gooddev-cpu @woopy098 or @esseciego, if you want to contribute then go ahead and develop the tests and send a pull request. You may also find https://github.com/pwndbg/pwndbg/blob/dev/DEVELOPING.md helpful for developing Pwndbg.

I am going to close this issue since the main feature has been implemented. We don't have tests for many things but we would always appreciate adding them.

Gaurav23-bit:
is it solved?

patryk4815:
> is it solved?

yes

SanthoshChevuri-444:
@SanthoshChevuri-777 I'd like to work on this issue!

PR #2623, opened 2024-12-14
Add hijackfd command to modify the file descriptor of a process

patryk4815:
Feature: https://github.com/pwndbg/pwndbg/issues/2621

### DIFF
--- a/pwndbg/aglib/shellcode.py
+++ b/pwndbg/aglib/shellcode.py
@@ -8,6 +8,7 @@ amount of code in the context of the inferior.
 
 from __future__ import annotations
 
+import contextlib
 from asyncio import CancelledError
 
 import pwnlib.asm
@@ -32,10 +33,11 @@ def _get_syscall_return_value():
     """
 
     register_set = pwndbg.lib.regs.reg_sets[pwndbg.aglib.arch.current]
+    # FIXME: `retval` is syscall abi? or sysv abi?
     return pwndbg.aglib.regs[register_set.retval]
 
 
-def exec_syscall(
+async def exec_syscall(
     ec: ExecutionController,
     syscall,
     arg0=None,
@@ -44,7 +46,6 @@ def exec_syscall(
     arg3=None,
     arg4=None,
     arg5=None,
-    arg6=None,
     disable_breakpoints=False,
 ):
     """
@@ -56,17 +57,18 @@ def exec_syscall(
     syscall_bin = pwnlib.asm.asm(syscall_asm)
 
     # Run the syscall and pass its return value onward to the caller.
-    return exec_shellcode(
+    async with exec_shellcode(
         ec,
         syscall_bin,
         restore_context=True,
-        capture=_get_syscall_return_value,
         disable_breakpoints=disable_breakpoints,
-    )
+    ):
+        return _get_syscall_return_value()
 
 
+@contextlib.asynccontextmanager
 async def exec_shellcode(
-    ec: ExecutionController, blob, restore_context=True, capture=None, disable_breakpoints=False
+    ec: ExecutionController, blob, restore_context=True, disable_breakpoints=False
 ):
     """
     Tries executing the given blob of machine code in the current context of the
@@ -146,18 +148,15 @@ async def exec_shellcode(
     # Make sure we're in the right place.
     assert pwndbg.aglib.regs.pc == target_address
 
-    # Give the caller a chance to collect information from the environment
-    # before any of the context gets restored.
-    captured = None
-    if capture is not None:
-        captured = capture()
-
-    # Restore the code and the program counter and, if requested, the rest of
-    # the registers.
-    pwndbg.aglib.memory.write(starting_address, existing_code)
-    setattr(pwndbg.aglib.regs, register_set.pc, starting_address)
-    if restore_context:
-        for reg, val in registers.items():
-            setattr(pwndbg.aglib.regs, reg, val)
-
-    return captured
+    try:
+        # Give the caller a chance to collect information from the environment
+        # before any of the context gets restored.
+        yield
+    finally:
+        # Restore the code and the program counter and, if requested, the rest of
+        # the registers.
+        pwndbg.aglib.memory.write(starting_address, existing_code)
+        setattr(pwndbg.aglib.regs, register_set.pc, starting_address)
+        if restore_context:
+            for reg, val in registers.items():
+                setattr(pwndbg.aglib.regs, reg, val)
--- a/pwndbg/commands/hijack_fd.py
+++ b/pwndbg/commands/hijack_fd.py
@@ -245,7 +245,7 @@ PARSED_FILE_ARG = Tuple[Optional[ParsedSocket], Optional[str]]
 
 def parse_file_or_socket(s: str) -> PARSED_FILE_ARG:
     # is file
-    if s.startswith("/"):
+    if s.startswith("/") or s.startswith("./"):
         return None, s
     return parse_socket(s), None
 
--- a/pwndbg/commands/hijack_fd.py
+++ b/pwndbg/commands/hijack_fd.py
@@ -280,6 +280,9 @@ def hijack_fd(fdnum: int, newfile: PARSED_FILE_ARG) -> None:
 
     async def ctrl(ec: pwndbg.dbg_mod.ExecutionController):
         async with exec_shellcode_with_stack(ec, asm_bin, stack_size):
-            print("Operation succeeded. Errors are not captured.\nYou can verify this with `procinfo` if the file descriptor has been replaced.")
+            print(
+                "Operation succeeded. Errors are not captured.\n"
+                "You can verify this with `procinfo` if the file descriptor has been replaced."
+            )
 
     pwndbg.dbg.selected_inferior().dispatch_execution_controller(ctrl)
--- a/pwndbg/commands/hijack_fd.py
+++ b/pwndbg/commands/hijack_fd.py
@@ -7,9 +7,13 @@ from typing import Literal
 from typing import NamedTuple
 from typing import Optional
 from typing import Tuple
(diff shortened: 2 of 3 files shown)


pwndbg/aglib/shellcode.py
```diff
@@ -32,10 +33,11 @@ def _get_syscall_return_value():
     """
 
     register_set = pwndbg.lib.regs.reg_sets[pwndbg.aglib.arch.current]
+    # FIXME: `retval` is syscall abi? or sysv abi?
     return pwndbg.aglib.regs[register_set.retval]
```
disconnect3d, line 37: Isn't this both? Or does this depend on architecture?

pwndbg/aglib/shellcode.py
```diff
@@ -56,17 +57,18 @@ def exec_syscall(
...
         disable_breakpoints=disable_breakpoints,
-    )
+    ):
+        return _get_syscall_return_value()
 
 
+@contextlib.asynccontextmanager
 async def exec_shellcode(
-    ec: ExecutionController, blob, restore_context=True, capture=None, disable_breakpoints=False
+    ec: ExecutionController, blob, restore_context=True, disable_breakpoints=False
```
disconnect3d, line 71: Why did we have `capture` in the first place? Why do we remove it now?

pwndbg/commands/hijack_fd.py
```diff
@@ -0,0 +1,283 @@
...
+import pwndbg.lib.abi
+import pwndbg.lib.memory
+import pwndbg.lib.regs
+from pwndbg.commands import CommandCategory
+
+
+class ShellcodeRegs(TypedDict):
+    newfd: str
+    syscall_ret: str
+    stack: str
```
disconnect3d, line 30: Why not just a dataclass? They don't need a dict and are likely faster?

pwndbg/commands/hijack_fd.py
```diff
@@ -0,0 +1,283 @@
...
+            ), f"stack is probably corrupted size_current=f{stack_diff_size} size_max_want={stack_size}"
+
+            yield
+    finally:
+        pwndbg.aglib.memory.write(stack_start, original_stack)
+
+
+parser = argparse.ArgumentParser(
+    formatter_class=argparse.RawTextHelpFormatter,
+    description="""Replace file descriptors of a debugged process.
```
disconnect3d:
```suggestion
    description="""Replace a file descriptor of a debugged process.
```

pwndbg/commands/hijack_fd.py
```diff
@@ -0,0 +1,283 @@
...
+)
+
+parser.add_argument(
+    "fdnum",
+    help="File descriptor (FD) number to be replaced with the specified new socket or file.",
+    type=int,
+)
+
+
+class ParsedSocket(NamedTuple):
```
disconnect3d, line 168: https://stackoverflow.com/questions/51671699/data-classes-vs-typing-namedtuple-primary-use-cases 🤯

pwndbg/commands/hijack_fd.py
```diff
@@ -0,0 +1,283 @@
...
+    port: int
+
+
+def parse_socket(url: str) -> ParsedSocket:
+    if "://" in url:
+        parsed = urlparse(url)
+    else:
+        parsed = ParseResult("", url, "", "", "", "")
+
+    scheme_info = parsed.scheme.split("+", 1)
```
disconnect3d, line 188: What is the "+"? Why don't we have any example with it?

pwndbg/commands/hijack_fd.py
```diff
@@ -0,0 +1,283 @@
...
+        raise argparse.ArgumentTypeError("Domain or IP is required")
+
+    port = parsed.port
+    if not port:
+        raise argparse.ArgumentTypeError("Port is required")
+
+    protocol_ordered = [
+        ("ipv4", socket.AF_INET),
+        ("ipv6", socket.AF_INET6),
+    ]
```
disconnect3d: nitpick: its better to use tuple of tuples as it will all be treated as a const:
```py
In [11]: def foo():
    ...:     a = [(1,2), (3,4)]
    ...:

In [12]: foo.__code__.co_consts
Out[12]: (None, (1, 2), (3, 4))

In [13]: def foo():
    ...:     a = ((1,2), (3,4))
    ...:

In [14]: foo.__code__.co_consts
Out[14]: (None, ((1, 2), (3, 4)))
```

pwndbg/commands/hijack_fd.py
```diff
@@ -0,0 +1,283 @@
...
+
+    found_ip_protocol: Literal["ipv4", "ipv6"] | None = None
+    address_ipv4_or_ipv6: str = ""
+    for family_name, family_const in protocol_ordered:
+        if selected_ip_protocol and selected_ip_protocol != family_name:
+            continue
+
+        try:
+            ips = socket.getaddrinfo(domain_or_ip, None, family_const)
```
disconnect3d, line 220: might be nice to add a comment what we do here (does it make sense to use addrinfo if we have an ip?)

pwndbg/commands/hijack_fd.py
```diff
@@ -0,0 +1,283 @@
...
+    if filename:
+        stack_size, asm_bin = asm_replace_file(fdnum, filename)
+    elif socket_data:
+        stack_size, asm_bin = asm_replace_socket(fdnum, socket_data)
+    else:
+        assert False
+
+    async def ctrl(ec: pwndbg.dbg_mod.ExecutionController):
+        async with exec_shellcode_with_stack(ec, asm_bin, stack_size):
+            print("Operation probably succeeded. Errors are not captured.")
```
disconnect3d: Probably? Why probably?:)

pwndbg/commands/hijack_fd.py
```diff
@@ -0,0 +1,283 @@
...
+    syscall_abi = pwndbg.lib.abi.ABI.syscall()
+    return s + (syscall_abi.arg_alignment - (s % syscall_abi.arg_alignment))
+
+
+def asm_replace_file(replace_fd: int, filename: str) -> Tuple[int, str]:
+    filename = filename.encode() + b"\x00"
+
+    from pwnlib import asm
+    from pwnlib import constants
+    from pwnlib import shellcraft
```
disconnect3d: Why here and not at top?

pwndbg/commands/hijack_fd.py
```diff
@@ -0,0 +1,283 @@
...
+
+    regs = get_shellcode_regs()
+    stack_size = stack_size_alignment(len(filename))
+
+    open_asm = (
+        shellcraft.syscall("SYS_open", regs["stack"], "O_CREAT|O_RDWR", 0o666)
```
disconnect3d: Why not `shellcraft.open`?

```py
In [2]: shellcraft.open?
Signature: shellcraft.open(file=0, oflag=0, mode=0)
Docstring:
open(file, oflag, mode) -> str

Invokes the syscall open.

See 'man 2 open' for more information.

Arguments:
    file(char*): file
    oflag(int): oflag
    mode(int): mode
Returns:
    int
File:      ~/.venv/pwn/lib/python3.9/site-packages/pwnlib/shellcraft/internal.py
Type:      function
```

disconnect3d: Also is there a reason we use `sp` register as file argument?

pwndbg/aglib/shellcode.py
patryk4815: Don't know why we used `capture`. In my opinion, using python `with` statement would be cleaner and more readable

pwndbg/commands/hijack_fd.py
patryk4815: we don't capture errors inside shell code, when syscall fail

pwndbg/commands/hijack_fd.py
patryk4815: I prefer to use directly syscall + `push_str`, because  `shellcraft.open` is maybe wrapper in some architectures, so they do different things with stack. And In some architectures eg: freebsd `shellcraft.open` is not implemented.

Best approach might be to implement `replace_fd` directly within `pwntools` and then make it accessible as `shellcraft.replace_fd`. 
However, we still have to determin how much stack this shellcode will consume. 
We could emulate it using `Unicorn` to calculate the maximum stack size required.

pwndbg/commands/hijack_fd.py
patryk4815: yeah, it resolves domain to ip, and if you put just `ip` then you know ip version ;) You don't need implement separate method like: `if is_ipv4 / is_ipv_6` statements

pwndbg/commands/hijack_fd.py
patryk4815: I put examples here:
```py
parser.add_argument(
    "newfile",
    help="""Specify a file or a socket.
For files, the filename must start with `/` (e.g., `/etc/passwd`).
For sockets, the following formats are allowed:
- `127.0.0.1:80` (default is TCP)
- `tcp://[::1]:80`
- `udp://example.com:80`
- `tcp+ipv6://example.com:80`
    """,
    type=parse_file_or_socket,
)
```

pwndbg/commands/hijack_fd.py
patryk4815: chatgpt generated this 😄 this wasn't my idea

pwndbg/commands/hijack_fd.py
patryk4815: maybe NamedTuple then :)

pwndbg/commands/hijack_fd.py
disconnect3d: That'd be neat. In the long term we should probably move those impls outside of command code too

disconnect3d (approved):

PR #2625, opened 2024-12-15
Improve pwndbg-lldb REPL

patryk4815:
Changes:
- pwndbg-lldb now utilizes a command history file.
- command history will be read from and stored in the `~/.pwndbg_history`
- remember last command
- fix prompt arrow up glitches


pwndbg/dbg/lldb/repl/__init__.py
```diff
@@ -402,29 +403,6 @@ def handle_sigint(_sig, _frame):
             break
 
 
-def make_pty() -> Tuple[str, int]:
```
patryk4815, line 405: cleanup, it was not used

PR #2626, opened 2024-12-15
Fix tls for aarch64 and arm

patryk4815:


pwndbg/aglib/tls.py
```diff
@@ -43,10 +43,13 @@ def find_address_with_pthread_self() -> int:
...
 
-    if pwndbg.aglib.arch.current == "arm":
-        # 0x4c0 is sizeof(struct pthread)
-        # TODO: we might need to adjust the value if the size of struct pthread is changed in the future.
-        result += 0x4C0
+    if pwndbg.aglib.arch.current in ("arm", "aarch64"):
+        pthread_type = pwndbg.aglib.typeinfo.load("struct pthread")
+        if pthread_type is None:
+            # Type 'pthread' not found
+            return 0
```
patryk4815, line 58: maybe it should fallback to const value, when there are no debug info?
Like:
```py
const_size = 0x4C0
if pthread_type is None:
    size = const_size
else:
    size = pthread_type.sizeof
    
if size != const_size:
     print("STATISTIC: Please report this to pwndbg that const_size is wrong for your LIBC version")
     
result += size
```

disconnect3d: I don't get it, what do u mean a constant value? Why such value? Where does it come from?

patryk4815: `const_size = 0x4C0` -> 0x4C0 is `sizeof(struct pthread)`. But when you don't have debug info `pwndbg.aglib.typeinfo.load("struct pthread")` will return None, but it could fallback to our hardcoded size. The hardcoded size might be wrong anyway It depends on libc version

patryk4815:
```
pwndbg> x/s &__libc_version
0xfce9d579b680 <__nptl_version>:	"2.40"
pwndbg> p __libc_version
'__libc_version' has unknown type; cast it to its declared type
```

PR #2640, opened 2024-12-18
Add next part of adding kernel support for LLDB

patryk4815:
Changes:
- make gdb only commands available to lldb
- fix lldb issues with pointers 
- improve `klookup` now it search like grep

### DIFF
--- a/pwndbg/aglib/symbol.py
+++ b/pwndbg/aglib/symbol.py
@@ -106,6 +106,9 @@ def resolve_addr(addr: int) -> str | None:
     Resolution is performed in the following order:
     - Global scope symbols.
     """
+    if addr < 0:
+        return None
+
     symbol_name = pwndbg.dbg.selected_inferior().symbol_name_at_address(addr)
     if symbol_name:
         return symbol_name
--- a/pwndbg/aglib/symbol.py
+++ b/pwndbg/aglib/symbol.py
@@ -106,9 +106,6 @@ def resolve_addr(addr: int) -> str | None:
     Resolution is performed in the following order:
     - Global scope symbols.
     """
-    if addr < 0:
-        return None
-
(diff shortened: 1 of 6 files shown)


pwndbg/aglib/symbol.py
```diff
@@ -106,6 +106,9 @@ def resolve_addr(addr: int) -> str | None:
     Resolution is performed in the following order:
     - Global scope symbols.
     """
+    if addr < 0:
+        return None
```
disconnect3d: Do we need this? Just assert maybe or remove this check?

patryk4815: I can remove this from this PR. Recently someone is sending here negative addresses :/

pwndbg/commands/klookup.py
```diff
@@ -12,21 +12,35 @@
 parser.add_argument("symbol", type=str, help="Address or symbol name to lookup")
 
 
+def parse_to_addr(v: str) -> int:
+    if v.startswith("0x"):
+        return int(v[2:], 16)
+    try:
+        return int(v, 16)
```
disconnect3d: We don't need the first if:
```py
In [1]: int('0x1234', 16)
Out[1]: 4660

In [2]: int('1234', 16)
Out[2]: 4660
```

disconnect3d: This also means that `klookup <some int in base 10>` will be interpreter as base 16, is that ok?

```py
In [3]: 0xffff1234112
Out[3]: 17591936696594

In [4]: int('17591936696594', 16)
Out[4]: 6571889286276500
```

pwndbg/aglib/symbol.py
patryk4815: deleted from this PR

pwndbg/commands/klookup.py
patryk4815: changed to int(x, 0)

PR #2643, opened 2024-12-21
Port argv

patryk4815:
Screenshot 2024-12-21 at 05 58 30


pwndbg/aglib/argv.py
```diff
@@ -0,0 +1,105 @@
...
+
+#: Pointer to envp on the stack
+envp_ptr: int = None
+
+#: Total number of environment variables
+envc_numbers: int = None
+
+
+@pwndbg.dbg.event_handler(EventType.START)
+def update() -> None:
```
disconnect3d, line 24: If we can, then we should not do this as an event handler and instead fetch those values from stack memory only if argv/envp etc is used.

I mean: most of the time the user does not need those at all?

pwndbg/aglib/remote.py
```diff
@@ -12,43 +12,3 @@
 @pwndbg.lib.cache.cache_until("objfile", "start")
 def is_remote() -> bool:
     return pwndbg.dbg.selected_inferior().is_remote()
-
-
-def is_debug_probe() -> bool:
```
disconnect3d, line 17: What happened to this?

disconnect3d: Ugh I see that this is unused. Seems that we maybe did some bug removing it :/

patryk4815: previously it was implemented here https://github.com/pwndbg/pwndbg/pull/1402
logic to prevent downloading? symbols from remote debug probe

Issue #2544, opened 2024-11-18
LLDB stepsyscall crash debugger 


patryk4815:
```python
$ ./pwndbg/bin/pwndbg-lldb /bin/sh
pwndbg-lldb> run -s
pwndbg-lldb> stepsyscall
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /opt/tmp/pwndbg/lib/python3.12/site-packages/pwndbg/dbg/lldb/repl/__init__.py:387 in run         │
│                                                                                                  │
│   384 │   │   │   assert driver.process.GetUniqueID() == process.process.GetUniqueID()           │
│   385 │   │   │                                                                                  │
│   386 │   │   │   try:                                                                           │
│ ❱ 387 │   │   │   │   driver.run_coroutine(coroutine)                                            │
│   388 │   │   │   except Exception:                                                              │
│   389 │   │   │   │   # We treat exceptions coming from the execution controllers the            │
│   390 │   │   │   │   # same way we treat exceptions coming from commands.                       │
│                                                                                                  │
│ /opt/tmp/pwndbg/lib/python3.12/site-packages/pwndbg/dbg/lldb/repl/proc.py:268 in run_coroutine   │
│                                                                                                  │
│   265 │   │   while True:                                                                        │
│   266 │   │   │   try:                                                                           │
│   267 │   │   │   │   if exception is None:                                                      │
│ ❱ 268 │   │   │   │   │   step = coroutine.send(None)                                            │
│   269 │   │   │   │   else:                                                                      │
│   270 │   │   │   │   │   step = coroutine.throw(exception)                                      │
│   271 │   │   │   │   │   # The coroutine has caught the exception. Continue running             │
│                                                                                                  │
│ /opt/tmp/pwndbg/lib/python3.12/site-packages/pwndbg/commands/next.py:164 in _stepsyscall         │
│                                                                                                  │
│   161 │                                                                                          │
│   162 │   while (                                                                                │
│   163 │   │   pwndbg.aglib.proc.alive                                                            │
│ ❱ 164 │   │   and not (await pwndbg.aglib.next.break_next_interrupt(ec))                         │
│   165 │   │   and (await pwndbg.aglib.next.break_next_branch(ec))                                │
│   166 │   ):                                                                                     │
│   167 │   │   # Here we are e.g. on a CALL instruction (temporarily breakpointed by `break_nex   │
│                                                                                                  │
[...]

PR #2645, opened 2024-12-22
fix negative address in some places 

patryk4815:
Fixes:
- https://github.com/pwndbg/pwndbg/issues/2544

### DIFF
--- a/pwndbg/chain.py
+++ b/pwndbg/chain.py
@@ -54,6 +54,7 @@ def get(
     """
     if address is None:
         return None
+    assert address >= 0, "address must be positive"
 
     limit = int(limit)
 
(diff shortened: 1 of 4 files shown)


pwndbg/chain.py
```diff
@@ -138,7 +139,7 @@ def format(
     arrow_right = c.arrow(f" {config_arrow_right} ")
 
     # Colorize the chain
-    rest = [M.get_address_and_symbol(link) for link in chain]
+    rest = [M.get_address_and_symbol(addr) if addr >= 0 else "" for addr in chain]
```
disconnect3d, line 142: Lmao if `limit == 0` we will still colorize the whole chain even though we return `rest[0]`

Issue #1953, opened 2023-12-13
pwngdb no break when debug on chip


AnduinBrian:
### Description
Hello, after use openocd to create a local gdb server, i try to attach but pwngdb cant break, it just hang. I try again with normal-gdb, it work.

### Steps to reproduce
I use ST-link and stm32, so i start the server with command 
```
openocd -f /usr/share/openocd/scripts/interface/stlink.cfg -f /usr/share/openocd/scripts/target/stm32f1x.cfg
```
I start gdb-multiarch, and type those command:
```
file firmware.elf                           <- this is my firmware i want to debug
b * loop                                    <- place some bp
target extended-remote localhost:3333       <- openocd port
```
GDB just hang there, i cant interact or do anything with the console.

![image](https://github.com/pwndbg/pwndbg/assets/36282595/63197489-d756-43df-85ba-a3b9f430d5e4)

But when i try with normal-gdb, i can work with it

![image](https://github.com/pwndbg/pwndbg/assets/36282595/a85f188b-c24e-402d-af24-32657cad22b4)

disconnect3d:
Eh, I don't have such hardware to test this myself. We invoke some GDB hooks when e.g. a library is loaded, on program start or when a stop occurs.

Most likely, it is related to us trying to figure out memory mappings and our aggresive reading of parts of memory to provide great experience.

This can be debugged to some degree with `set debug-events on`. I'd be happy to merge improvements that would make chip/embedded debugging more pleasant.

nikvoid:
This is output with `debug-events` and `verbose-exceptions` enabled. Pwndbg just freezes waiting for some events?

```
pwndbg> ctx
// Just stuck until ctrl+C
^CQuit
'before_prompt' pwndbg.lib.cache.clear ()
'stop' pwndbg.lib.cache.clear ()
'stop' pwndbg.dbg.gdb._bpwp_process_deferred ()
'stop' pwndbg.aglib.regs.update_last ()
'stop' pwndbg.aglib.stack.is_executable ()
// Stuck again
^CTraceback (most recent call last):
  File "/home/nv/pwndbg/pwndbg/gdblib/prompt.py", line 95, in prompt_hook
    pwndbg.gdblib.events.after_reload(start=cur is None)
  File "/home/nv/pwndbg/pwndbg/gdblib/events.py", line 365, in after_reload
    invoke_event(gdb.events.stop)
[...]

PR #2646, opened 2024-12-22
Disable stack and vmmap exploration

patryk4815:
Fixes:
- https://github.com/pwndbg/pwndbg/issues/1953

Changes:
- add new option `set auto-explore-stack yes/warn/no` (default: warn)
- add new command `stack_explore`
- Disable vmmap auto exploration, `auto-explore-pages` now defaults to `warn` 

image
image

Additional changes:
- port vmmap custom pages to aglib / lldb
- `pwndbg.aglib.file.get_file`: Avoid falling back to the local path when the remote does not support vFile.
- `vmmap_load`: Now stores the section name in the `objfile` attribute.
image

### DIFF
--- a/pwndbg/aglib/elf.py
+++ b/pwndbg/aglib/elf.py
@@ -32,14 +32,12 @@ import pwndbg.aglib.proc
 import pwndbg.aglib.qemu
 import pwndbg.aglib.symbol
 import pwndbg.aglib.vmmap
+import pwndbg.auxv
 import pwndbg.lib.cache
 import pwndbg.lib.elftypes
 import pwndbg.lib.memory
 from pwndbg.color import message
 from pwndbg.dbg import EventType
-
-if pwndbg.dbg.is_gdblib_available():
-    import pwndbg.auxv
 
 # ELF constants
 PF_X, PF_W, PF_R = 1, 2, 4
@@ -255,10 +253,9 @@ def entry() -> int:
     """
     Return the address of the entry point for the main executable.
     """
-    if pwndbg.dbg.is_gdblib_available():
-        entry = pwndbg.auxv.get().AT_ENTRY
-        if entry:
-            return entry
+    entry = pwndbg.auxv.get().AT_ENTRY
+    if entry:
+        return entry
 
     inf = pwndbg.dbg.selected_inferior()
     entry = inf.main_module_entry()
--- a/pwndbg/aglib/kernel/vmmap.py
+++ b/pwndbg/aglib/kernel/vmmap.py
@@ -138,9 +138,9 @@ def kernel_vmmap_via_page_tables() -> Tuple[pwndbg.lib.memory.Page, ...]:
     arch = pwndbg.aglib.arch.current
     if arch == "aarch64":
         arch_backend = PT_Aarch64_Backend(machine_backend)
-    elif arch == "i386" in arch:
+    elif arch == "i386":
         arch_backend = PT_x86_64_Backend(machine_backend)
-    elif arch == "x86-64" in arch:
+    elif arch == "x86-64":
         arch_backend = PT_x86_64_Backend(machine_backend)
     elif arch == "rv64":
         arch_backend = PT_RiscV64_Backend(machine_backend)
@@ -269,3 +269,41 @@ def kernel_vmmap_via_monitor_info_mem() -> Tuple[pwndbg.lib.memory.Page, ...]:
         pages.append(pwndbg.lib.memory.Page(start, size, flags, 0, "<qemu>"))
 
     return tuple(pages)
+
+
+kernel_vmmap_mode = pwndbg.config.add_param(
+    "kernel-vmmap",
+    "page-tables",
+    "the method to get vmmap information when debugging via QEMU kernel",
+    help_docstring="""\
+kernel-vmmap can be:
+page-tables    - read /proc/$qemu-pid/mem to parse kernel page tables to render vmmap
+monitor        - use QEMU's `monitor info mem` to render vmmap
+none           - disable vmmap rendering; useful if rendering is particularly slow
+
+Note that the page-tables method will require the QEMU kernel process to be on the same machine and within the same PID namespace. Running QEMU kernel and GDB in different Docker containers will not work. Consider running both containers with --pid=host (meaning they will see and so be able to interact with all processes on the machine).
+""",
+    param_class=pwndbg.lib.config.PARAM_ENUM,
+    enum_sequence=["page-tables", "monitor", "none"],
+)
+
+
+def kernel_vmmap() -> Tuple[pwndbg.lib.memory.Page, ...]:
+    if not pwndbg.aglib.qemu.is_qemu_kernel():
+        return ()
+
+    if pwndbg.aglib.arch.current not in (
+        "i386",
+        "x86-64",
+        "aarch64",
+        "rv32",
+        "rv64",
+    ):
+        return ()
+
+    if kernel_vmmap_mode == "page-tables":
+        return kernel_vmmap_via_page_tables()
+    elif kernel_vmmap_mode == "monitor":
+        return kernel_vmmap_via_monitor_info_mem()
+
+    return ()
--- a/pwndbg/aglib/stack.py
+++ b/pwndbg/aglib/stack.py
@@ -15,9 +15,19 @@ import pwndbg.aglib.elf
 import pwndbg.aglib.memory
 import pwndbg.aglib.regs
 import pwndbg.aglib.vmmap
+import pwndbg.color.message as M
 import pwndbg.lib.cache
+import pwndbg.lib.config
 import pwndbg.lib.memory
 from pwndbg.dbg import EventType
+
+auto_explore = pwndbg.config.add_param(
+    "auto-explore-stack",
+    "warn",
+    "Enable or disable stack exploration; it may be really slow.",
+    param_class=pwndbg.lib.config.PARAM_ENUM,
+    enum_sequence=["warn", "yes", "no"],
+)
 
 
 def find(address: int):
@@ -120,6 +130,19 @@ def _fetch_via_exploration() -> Dict[int, pwndbg.lib.memory.Page]:
     An alternative to this is dumping this functionality completely and this
     will be decided hopefully after a next release.
     """
+    if auto_explore.value == "warn":
+        print(
+            M.warn(
+                "Warning: All methods to detect STACK have failed.\n"
+                "You can explore STACK using exploration, but it may be very slow.\n"
+                "To explicitly explore, use the command: `stack_explore`\n"
+                "Alternatively, enable it by default with: `set auto-explore-stack yes`"
+            )
+        )
+        return {}
+    elif auto_explore.value == "no":
+        return {}
+
     stacks: Dict[int, pwndbg.lib.memory.Page] = {}
 
     for thread in pwndbg.dbg.selected_inferior().threads():
--- a/pwndbg/aglib/stack.py
+++ b/pwndbg/aglib/stack.py
@@ -109,9 +109,10 @@ def _fetch_via_vmmap() -> Dict[int, pwndbg.lib.memory.Page]:
                 page = p
                 break
 
-        if page:
-            stacks[thread.index()] = page
+        if not page:
+            # page not found, should we explore `sp` register?
             continue
+        stacks[thread.index()] = page
 
     return stacks
 
--- a/pwndbg/aglib/stack.py
+++ b/pwndbg/aglib/stack.py
@@ -19,7 +19,9 @@ import pwndbg.color.message as M
 import pwndbg.lib.cache
 import pwndbg.lib.config
 import pwndbg.lib.memory
-from pwndbg.dbg import EventType
+
+if pwndbg.dbg.is_gdblib_available():
+    import pwndbg.gdblib.vmmap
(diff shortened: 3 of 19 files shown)


patryk4815:
rebased

disconnect3d:
Btw we have `vmmap_explore`:
```
pwndbg> pwndbg vmmap
Memory Commands [Aliases]         Description
--------------------------------  --------------------------------------------
vmmap [lm, address, vprot, libs]  Print virtual memory map pages.
vmmap_add                         Add virtual memory map page.
vmmap_clear                       Clear the vmmap cache.
vmmap_explore                     Explore a page, trying to guess permissions.
vmmap_load                        Load virtual memory map pages from ELF file.
```

Shall this do the stack exploration itself? Maybe `stack_explore` should be `vmmap_explore --stack` (only) or sth?

alufers:
Can confirm this PR prevents slow memory scanning on embedded mcus.

pwndbg/aglib/elf.py
```diff
@@ -255,10 +253,9 @@ def entry() -> int:
     """
     Return the address of the entry point for the main executable.
     """
-    if pwndbg.dbg.is_gdblib_available():
-        entry = pwndbg.auxv.get().AT_ENTRY
-        if entry:
-            return entry
+    entry = pwndbg.auxv.get().AT_ENTRY
+    if entry:
```
disconnect3d, line 257: Nitpick, but this should probably be `if entry is not None:` because if `AT_ENTRY` were to be 0, we won't return it properly? :D 

This is 100% edge case and very unlikely to ever happen... One would need a linux with `vm.mmap_min_addr=0` sysctl setting as well as a really custom binary with entrypoint set to address 0. Honestly, I don't think this is ever possible, so maybe let's leave this as it is ;).

pwndbg/aglib/kernel/vmmap.py
```diff
@@ -138,9 +138,9 @@ def kernel_vmmap_via_page_tables() -> Tuple[pwndbg.lib.memory.Page, ...]:
     arch = pwndbg.aglib.arch.current
     if arch == "aarch64":
         arch_backend = PT_Aarch64_Backend(machine_backend)
-    elif arch == "i386" in arch:
+    elif arch == "i386":
         arch_backend = PT_x86_64_Backend(machine_backend)
-    elif arch == "x86-64" in arch:
+    elif arch == "x86-64":
```
disconnect3d, line 143: Ah yeah, the ulmighty/funny weird Python pattern... :-)

disconnect3d: I'm really surprised that no linters detected this weirdness.

pwndbg/dbg/lldb/__init__.py
```diff
@@ -1089,15 +1105,25 @@ def send_remote(self, packet: str) -> bytes:
...
-                    continue
-                return ret.encode()
+        try:
+            idx = response.index("\nresponse: ")
+        except ValueError:
+            # Packets not implemented return empty
+            return b""
+
+        try:
+            is_err = response.index("\nresponse: \nerror: ") > -1
```
disconnect3d: We already have an index of `"\nresponse: "`. Wouldn't it be better to do such a test instead?:

```suggestion
            is_err = response[idx:].startswith("\nerror: ")
```

pwndbg/dbg/lldb/__init__.py
```diff
@@ -1089,15 +1105,25 @@ def send_remote(self, packet: str) -> bytes:
...
+        except ValueError:
+            is_err = False
 
-        # Packets not implemented return empty
-        return b""
+        if is_err:
+            # Packets not implemented return empty
+            return b""
+
+        out = response[idx + len("\nresponse: ") :].encode()
```
disconnect3d: Out of curiosity, why do we `.encode()`, or, does `self.dbg._execute_lldb_command` internally do call `.decode()`? 

I mean, we should probably not do decode -> encode if this is happening.

patryk4815: we should fill issue/feature request in llvm. Basically this `process plugin packet send` is buggy for \x7f + characters 

image

pwndbg/aglib/stack.py
```diff
@@ -120,26 +127,50 @@ def _fetch_via_exploration() -> Dict[int, pwndbg.lib.memory.Page]:
...
+        if page_found:
+            # Skip further exploration of stacks that have already been scanned.
+            stacks[thread_idx] = page_found
+            continue
 
-        start = sp_low
         stop = find_upper_stack_boundary(sp)
         page = pwndbg.lib.memory.Page(
-            start, stop - start, 6 if not is_executable() else 7, 0, f"[stack:{thread.index()}]"
+            start, stop - start, 6 if not is_executable() else 7, 0, "[stack]"
```
disconnect3d: Is this fine that we name this `[stack]` instead of `[stack:<tid>]`? What if there are multiple threads?

patryk4815: In different places we have eg. `pwndbg.lib.memory.Page`:
image

pwndbg/aglib/vmmap_custom.py
```diff
@@ -0,0 +1,156 @@
...
+
+def get_custom_pages() -> Tuple[pwndbg.lib.memory.Page, ...]:
+    """
+    Returns a tuple of `Page` objects representing the memory mappings of the
+    target, sorted by virtual address ascending.
+    """
+    pages: List[pwndbg.lib.memory.Page] = []
+    pages.extend(explored_pages)
+    pages.extend(custom_pages)
+    pages.sort()
```
disconnect3d, line 47: We can do:
```py
pages: List[pwndbg.lib.memory.Page] = [*explored_pages, *custom_pages]
pages.sort()
return tuple(pages)
```

disconnect3d: Could also be: `return tuple(sorted(explored_pages+custom_pages))` hmm...

patryk4815: done

pwndbg/aglib/stack.py
```diff
@@ -85,23 +96,19 @@ def _fetch_via_vmmap() -> Dict[int, pwndbg.lib.memory.Page]:
...
-        for p in pages:
-            if sp in p:
-                page = p
-                break
-
-        if page:
-            stacks[thread.index()] = page
+        page = next((page for page in pages if sp in page), None)
+        if not page:
+            # page not found, should we explore `sp` register?
```
disconnect3d: Is that a question to be decided now, or when? ;)

pwndbg/aglib/stack.py
patryk4815: its hard to say whether this info `f"[stack:{thread.index()}]"` is useful?

pwndbg/aglib/stack.py
```diff
@@ -120,26 +127,50 @@ def _fetch_via_exploration() -> Dict[int, pwndbg.lib.memory.Page]:
...
     will be decided hopefully after a next release.
     """
-    stacks: Dict[int, pwndbg.lib.memory.Page] = {}
+    if auto_explore.value == "warn":
+        print(
+            M.warn(
+                "Warning: All methods to detect STACK have failed.\n"
+                "You can explore STACK using exploration, but it may be very slow.\n"
+                "To explicitly explore, use the command: `stack_explore`\n"
+                "Alternatively, enable it by default with: `set auto-explore-stack yes`"
```
disconnect3d, line 137: Isn't this a bit weird that `fetch_via_exploration` does print this warning? Seems like maybe the code that uses it should do that?

Unless there are many places for it? I don't know huh.

pwndbg/aglib/stack.py
disconnect3d: Maybe it should be `self.objfile.startswith("[stack")` eh..

disconnect3d: I think this is to color the main thread stack generally speaking, but its not great considering that multiple stacks can exist...

pwndbg/aglib/vmmap_custom.py
```diff
@@ -0,0 +1,154 @@
...
+    elif auto_explore.value == "no":
+        return None
+
+    address_maybe = pwndbg.lib.memory.page_align(address_maybe)
+    flags = get_memory_flags(address_maybe)
+    if flags is None:
+        return None
+
+    page = find_boundaries(address_maybe)
+    page.objfile = "<explored>"
```
disconnect3d: I wonder if this should say sth like `<explored:0x123>` where 0x123 is like idk, first X nibbles of an address

pwndbg/aglib/vmmap_custom.py
```diff
@@ -0,0 +1,154 @@
...
+def get_memory_flags(address_maybe: int) -> int | None:
+    flags = 4 if pwndbg.aglib.memory.peek(address_maybe) else 0
+    if not flags:
+        return None
+
+    if pwndbg.aglib.memory.poke(address_maybe):
+        flags |= 2
+    # It's really hard to check for executability, so we just make some guesses:
+    # If it's in the same page as the stack pointer, try to check the NX bit
+    # If it's in the same page as the instruction pointer, assume it's executable
```
disconnect3d: Please reorder this line with the above one, since we first check for pc register and then for sp+NX bit

pwndbg/aglib/vmmap_custom.py
```diff
@@ -0,0 +1,154 @@
...
+
+def find_boundaries(addr: int, name: str = "", min: int = 0) -> pwndbg.lib.memory.Page:
+    """
+    Given a single address, find all contiguous pages
+    which are mapped.
+    """
+    start = pwndbg.aglib.memory.find_lower_boundary(addr)
+    end = pwndbg.aglib.memory.find_upper_boundary(addr)
+
+    start = max(start, min)
```
disconnect3d, line 154: Technically if both start,end are below min, then we end up with start>end

pwndbg/auxv.py
```diff
@@ -140,23 +139,12 @@ def explore_stack_auxv() -> AUXV | None:
...
-        except pwndbg.dbg_mod.Error:
-            pass
-
-    return auxv
+    return walk_stack2()
 
 
 def walk_stack2(offset: int = 0) -> AUXV:
+    # FIXME: This function doesn't work with Go (Golang) binaries, as Golang has two stacks.
+    # NOTE: This function is intended to work only with real binaries, not those emulated under qemu-user.
```
disconnect3d, line 147: Why? Can we elaborate on this and say which qemu-user version do we mean?

pwndbg/commands/vmmap.py
```diff
@@ -284,118 +284,135 @@ def vmmap(
...
-    )
+parser = argparse.ArgumentParser(description="Add virtual memory map page.")
+parser.add_argument("start", type=int, help="Starting virtual address")
+parser.add_argument("size", type=int, help="Size of the address space, in bytes")
+parser.add_argument(
+    "flags",
+    nargs="?",
+    type=str,
+    default="",
+    help="Flags set by the ELF file, see PF_X, PF_R, PF_W",
```
disconnect3d: How does one see `PF_X` etc? :p

We should provide an example what we expect here as valid flag

disconnect3d: Maybe: `Flags set by the ELF file (r - read, w - write, x - executable)` if that is valid? idk

pwndbg/dbg/lldb/__init__.py
```diff
@@ -1089,15 +1105,22 @@ def send_remote(self, packet: str) -> bytes:
...
+            return b""
 
-        # Packets not implemented return empty
-        return b""
+        is_err = response[idx + size :].startswith("\nerror: ")
+        if is_err:
+            # Packets not implemented return empty
+            return b""
+
+        out = response[idx + size :].encode()
```
disconnect3d: Hmm... we can probably just do:
```py
        try:
            idx = response.index("\nresponse: ")
        except ValueError:
            # Packets not implemented return empty
            return b""
        
        out = response[idx+11:]  # len("\nresponse: ") == 11
        if out.startswith("\nerror: "):
            return b""  # packets not implemented, return empty

        if out[-1] == "\n":
            out = out[:-1]

        return out.encode()
```

disconnect3d (commented):
See comments. Also thanks @alufers for checking this out!

pwndbg/auxv.py
patryk4815: We should drop "hacks" for exploring weird memory mapping created by qemu-user.
qemu-user is using procfs currently for auxv, so we will not fallback to this code

pwndbg/aglib/stack.py
patryk4815: done

pwndbg/aglib/vmmap_custom.py
patryk4815: done

pwndbg/aglib/vmmap_custom.py
patryk4815: done

pwndbg/commands/vmmap.py
patryk4815: done

pwndbg/aglib/vmmap_custom.py
patryk4815: done

PR #2661, opened 2025-01-03
load symbols from header files  #1973

### LINKED ISSUE #1973: cymbol: load symbols from header files
The `cymbol ...` command should support adding symbols from a header file. That should make it easier to use it in scripts etc.

0xRavenspar:
Summary
This PR adds the ability to load symbols directly from header files using the -f flag in the cymbol command. Alongside this enhancement, the necessary tests were implemented to ensure proper functionality, and the documentation was updated to reflect the new feature.

Closes #1973 

`cymbol -f path/to/header.h`

added `test_cymbol_header_file` test to tests/gdb-tests/tests/test_cymbol.py

updated docs/commands/cymbol.md to reflect the -f flag in the documentation.


pwndbg/commands/cymbol.py
```diff
@@ -153,6 +153,48 @@ def add_custom_structure(custom_structure_name: str) -> None:
     load_custom_structure.__wrapped__(custom_structure_name, pwndbg_custom_structure_path)
 
 
+def add_structure_from_header(header_file: str, custom_structure_name: str = None) -> None:
+    # Resolve the header file's relative path to an absolute path
+    header_file = os.path.abspath(header_file)
```
disconnect3d: Why do we need abs path and `os.path.exists`? Can't we just attempt to open the file and have an except case if it fails? (this here makes a time of check vs time of use situation; we don't really care, but yeah)

tests/gdb-tests/tests/test_cymbol.py
```diff
@@ -77,3 +78,49 @@ def test_cymbol(start_binary):
...
+
+def create_temp_header_file(content: str) -> str:
+    """Create a temporary header file with the given content."""
+    with tempfile.NamedTemporaryFile(delete=False, suffix=".h") as tmp_file:
+        tmp_file.write(content.encode())
+        return tmp_file.name
+
+
+def test_cymbol_header_file(start_binary):
+    start_binary(REFERENCE_BINARY)
```
disconnect3d, line 91: Cool test. Can we add another one when there are multiple structures in one header file?

Also does this work if I use `#include <stdint.h>` and use types like `uint16_t` etc? May be nice to add a test like this

pwndbg/commands/cymbol.py
```diff
@@ -153,6 +153,48 @@ def add_custom_structure(custom_structure_name: str) -> None:
...
+            content = src.read().strip()
+            if not content:
+                print(message.notice("Header file is empty, skipping..."))
+                return
+            f.write(content)
+    except (IOError, OSError) as e:
+        print(message.error(f"Failed to process header file: {e}"))
+        return
+
+    load_custom_structure.__wrapped__(custom_structure_name, pwndbg_custom_structure_path)
```
disconnect3d, line 189:
```suggestion
    # Avoid checking for file existance. Call the decorator wrapper directly.
    load_custom_structure.__wrapped__(custom_structure_name, pwndbg_custom_structure_path)
```

disconnect3d (commented):
Hi, thanks for the PR! This looks good; I've added small comments here and there. Ideally, would be nice to add 1-2 more test cases :)

pwndbg/commands/cymbol.py
0xRavenspar: Yes i realised the issue this causes. I have removed it in the latest commit

tests/gdb-tests/tests/test_cymbol.py
0xRavenspar: yes i have added tests to check for multiple structures now. Can you please review them. Yes using `#include <stdint.h>` works! I have the tests for it as well

pwndbg/commands/cymbol.py
0xRavenspar: fixed this 👍🏻

pwndbg/commands/cymbol.py
```diff
@@ -153,6 +153,46 @@ def add_custom_structure(custom_structure_name: str) -> None:
     load_custom_structure.__wrapped__(custom_structure_name, pwndbg_custom_structure_path)
 
 
+def add_structure_from_header(header_file: str, custom_structure_name: str = None) -> None:
+    if not os.path.exists(header_file):
```
disconnect3d: We could still just leave it for `with open(header_file, "r") ...` - no need to test for this here I think?

pwndbg/commands/__init__.py
```diff
@@ -489,7 +489,7 @@ def OnlyWithResolvedHeapSyms(function: Callable[P, T]) -> Callable[P, T | None]:
     @functools.wraps(function)
     def _OnlyWithResolvedHeapSyms(*a: P.args, **kw: P.kwargs) -> T | None:
         e = log.error
-        w = log.warn
+        w = log.warning
```
disconnect3d, line 467: This change seems irrelevant, do we need it?

disconnect3d (changes requested):
Added some last comments

pwndbg/commands/__init__.py
0xRavenspar: `log.warn` is deprecated and `log.warning` is the suggested change, the linter wouldn't allow me to push the code without resolving this. Hence why i had to make the change

pwndbg/commands/cymbol.py
0xRavenspar: This would allow us to gracefully print an error msg to the user, hence that's why i thought it would be required. Otherwise the only error msg  would be `Failed to process header file` which would be very confusing to debug. Should i remove it or what changes would you suggest?

0xRavenspar:
@disconnect3d hi i have made some changes. Can you review it please

disconnect3d (approved):

PR #2664, opened 2025-01-08
spelling

rasmus11423:


patryk4815:
@rasmus11423 this ida_script works at all?

disconnect3d:
I don't think so. We should probably just remove the script.

PR #2670, opened 2025-01-14
Speed up pygments lexer selection

richyliu:
The pygments lexer lookup process is quite slow (see: pygments/pygments#2787) because it needs to go through every single lexer to select a match. I improve upon this lookup process by selecting a lexer match from the built-in lexers based on the lexer filename pattern. I fall back to the `guess_lexer_for_filename` function if there are multiple matches or no built-in lexer matches.

This dramatically speeds up time of the first `context` when there is debug source code that needs to be highlighted (3-4x speedup, from 0.8s to 0.2s on my machine). The only chance that a lexer can be selected wrong is when a plugin lexer has the same filename pattern as a built-in lexer (in which case the built-in lexer will always be used, whereas `guess_lexer_for_filename` will select a better lexer based on its `.analyse_text()` score).

### DIFF
--- a/pwndbg/color/syntax_highlight.py
+++ b/pwndbg/color/syntax_highlight.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import fnmatch
 import os.path
 import re
 from typing import Any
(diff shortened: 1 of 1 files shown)


disconnect3d:
Thanks! Can you also show here in a comment what's the best way to profile this in the future?

richyliu:
To profile from within pwndbg, I use the `profile` command from #2671 as follows
```
gdb --batch --quiet -ex "b break_here" -ex "profiler start" -ex "run" -ex "ctx" -ex "profiler stop" tests/gdb-tests/tests/binaries/telescope_binary.out
```
The profiler outputs the time elapsed automatically. The profiling results can also be analyzed with the indicated script.

The binary used here doesn't matter much as long as it has debug symbols and associated code file (which `ctx` tries to syntax highlight, which is the path being optimized here). Note that the binary used here is from the tests which can be built by running `./tests.sh`.

The `--batch` flag tells GDB to quit after executing the provided commands. Note that the explicit `ctx` is necessary because although `ctx` is triggered on a breakpoint, for some reason that doesn't happen when in `--batch` mode.

You can also profile the entire GDB run
```sh
sudo chrt -f 99 /usr/bin/time -v gdb --batch --quiet -ex "b break_here" -ex "run" -ex "ctx" tests/gdb-tests/tests/binaries/telescope_binary.out
```
The `sudo chrt -f 99` part sets the scheduling priority so that the Linux scheduler avoids task switching (so more accurate wall clock time).

PR #2675, opened 2025-01-17
Fix terminal size calculation

### LINKED ISSUE #2673: get_window_size() is innacurate when output is split into multiple panes/windows with contextoutput()
### Description

This function takes the width from the debugger pane by default
```python
def get_window_size(target=sys.stdin):
    fallback = (int(os.environ.get("LINES", 20)), int(os.environ.get("COLUMNS", 80)))
    if not target.isatty():
        return fallback
    rows, cols = get_cmd_window_size()        # !!!! this gets called usually
    if rows is not None and cols is not None:
        return rows, cols
    elif os.environ.get("PWNDBG_IN_TEST") is not None:
        return fallback
    try:
        # get terminal size and force ret buffer len of 4 bytes for safe unpacking by passing equally long arg
        rows, cols = struct.unpack("hh", fcntl.ioctl(target.fileno(), termios.TIOCGWINSZ, b"1234"))
    except Exception:
        rows, cols = fallback
    return rows, cols
```
Which is wrong when not all panes are that size:
image
And so the `[ EXPRESSIONS ]` and `[ BACKTRACE ]` banners are too big.

If we change the 
```python
```
line to
```python
    rows, cols = None, None
```
then 
```python
rows, cols = struct.unpack("hh", fcntl.ioctl(target.fileno(), termios.TIOCGWINSZ, b"1234"))
````
gets used and works correctly. Though I assume that is less efficient - not sure if there is some better solution (is it possible to cache pane size until it gets resized?).

### Steps to reproduce

Use something like this:  https://blog.jcfg.re/posts/pwndbg-tmux/ (https://github.com/joaogodinho/pwnmux/blob/main/gdbinit.py). In this case change `contextoutput(section, p[1], True, 'top', False)` to `contextoutput(section, p[1], True, 'top', None)` so banners get generated.

### My setup
```
Gdb:      15.2
Python:   3.13.1
Pwndbg:   2024.08.29 build: 2546eef4
Capstone: 5.0.1280
Unicorn:  2.1.0
```

### PR DESCRIPTION
**Pull request by k4lizen**
Fixes https://github.com/pwndbg/pwndbg/issues/2673 .

> Can we somehow detect that the user decided to use panels and then do it properly (and less efficiently, oh well)?

This PR does this, essentially.

### DIFF
--- a/pwndbg/ui.py
+++ b/pwndbg/ui.py
@@ -70,10 +70,12 @@ def get_window_size(target=sys.stdin):
     fallback = (int(os.environ.get("LINES", 20)), int(os.environ.get("COLUMNS", 80)))
     if not target.isatty():
         return fallback
-    rows, cols = get_cmd_window_size()
-    if rows is not None and cols is not None:
-        return rows, cols
-    elif os.environ.get("PWNDBG_IN_TEST") is not None:
+    if target == sys.stdin:
+        # We can ask the debugger for the window size
+        rows, cols = get_cmd_window_size()
+        if rows is not None and cols is not None:
+            return rows, cols
+    if os.environ.get("PWNDBG_IN_TEST") is not None:
         return fallback
     try:
         # get terminal size and force ret buffer len of 4 bytes for safe unpacking by passing equally long arg
--- a/pwndbg/ui.py
+++ b/pwndbg/ui.py
@@ -39,10 +39,10 @@ def check_title_position() -> None:
         title_position.revert_default()
 
 
-def banner(title, target=sys.stdin, width=None, extra=""):
+def banner(title, target=sys.stdout, width=None, extra=""):
     title = title.upper()
-    if width is None:  # auto width. In case of stdout, it's better to use stdin (b/c GdbOutputFile)
-        _height, width = get_window_size(target=target if target != sys.stdout else sys.stdin)
+    if width is None:
+        _height, width = get_window_size(target)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pwndbg/ui.py [hunk 1]
```diff
@@ -70,10 +70,12 @@ def get_window_size(target=sys.stdin):
     fallback = (int(os.environ.get("LINES", 20)), int(os.environ.get("COLUMNS", 80)))
     if not target.isatty():
         return fallback
-    rows, cols = get_cmd_window_size()
-    if rows is not None and cols is not None:
-        return rows, cols
-    elif os.environ.get("PWNDBG_IN_TEST") is not None:
+    if target == sys.stdin:
```
patryk4815:
```suggestion
    if target in (sys.stdout, sys.stdin):
```

patryk4815:
I think this PR is good, but in pwndbg we only passing there `sys.stdin` because of bug/workaround for gdb.

https://github.com/pwndbg/pwndbg/blob/2546eef4ab71c33f2e4e15a59339cc681a00921d/pwndbg/ui.py#L42-L46

Workaround for `sys.stdout` was fixed in https://github.com/pwndbg/pwndbg/pull/2459 - soo we can safely drop `if target != sys.stdout`
And also default value for `target` parameter should be `target=sys.stdout`, not `target=sys.stdin` in functions:
- `def banner(`
- `def get_window_size(`

PR #2684, opened 2025-01-20
Add support for attaching to processes to LLDB Pwndbg

mbrla0:
This PR plugs one of the gaps in LLDB that we still have, which is the lack of support for attaching to processes. It adds code to support it both to the Pwndbg CLI and the process driver.

It's important to note that the code here isn't entirely correct. Our handling of the process lifecycle in LLDB has a pretty big blind spot when it comes to remote processes, that the code in this PR doesn't work to fix. As this is a problem that also affects our implementation of `process launch`, though, I feel it's better to just add support for `attach` and deal with it later, than to try and fix both things at once.

### DIFF
--- a/pwndbg/dbg/lldb/repl/__init__.py
+++ b/pwndbg/dbg/lldb/repl/__init__.py
@@ -308,9 +308,7 @@ def run(startup: List[str] | None = None, debug: bool = False) -> None:
                 continue
             if len(bits) > 1 and bits[1].startswith("a") and "attach".startswith(bits[1]):
                 # This is `process attach`.
-                #
-                # TODO: Implement process attach.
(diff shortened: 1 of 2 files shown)


mbrla0:
Also @patryk4815, I'd really appreciate it if you could help test this, too. Testing locally, it works fine, but I can't get vanilla LLDB to properly attach in some circumstances - namely, attaching to processes by name and waiting for new ones to launch, so I couldn't test all of the functionality here.

disconnect3d:
Just to be sure: what's the blind spot?

mbrla0:
> Just to be sure: what's the blind spot?

I'm in the process of opening an issue about it. I'll describe it in more detail there.

disconnect3d (approved):

Issue #2690, opened 2025-01-21
Basic LoongArch64 support


xtexx:
Recently, when debugging a LoongArch64 ELF, I found that it seems to be unsupported.
```
Traceback (most recent call last):
  File "/opt/pwndbg/pwndbg/gdblib/prompt.py", line 95, in prompt_hook
    pwndbg.gdblib.events.after_reload(fire_start=cur is None)
  File "/opt/pwndbg/pwndbg/gdblib/events.py", line 365, in after_reload
    invoke_event(gdb.events.stop)
  File "/opt/pwndbg/pwndbg/gdblib/events.py", line 360, in invoke_event
    f(*args, **kwargs)
  File "/opt/pwndbg/pwndbg/gdblib/events.py", line 292, in caller
    raise e
  File "/opt/pwndbg/pwndbg/gdblib/events.py", line 287, in caller
    func()
  File "/opt/pwndbg/pwndbg/gdblib/hooks.py", line 44, in update_arch
    arch_mod_aglib.update()
  File "/opt/pwndbg/pwndbg/aglib/arch.py", line 71, in update
    pwnlib.context.context.arch = pwnlib_archs_mapping[a.name]
KeyError: 'Loongarch64'
```

The support for LA64 has already been in GDB upstream for a while, are there any plans to support it?

btw `bugreport --run-browser` also failed
```
pwndbg> bugreport --run-browser
Traceback (most recent call last):
  File "/opt/pwndbg/pwndbg/commands/__init__.py", line 183, in __call__
    return self.function(*args, **kwargs)
  File "/opt/pwndbg/pwndbg/commands/version.py", line 226, in bugreport
    cmd_no = int(cmd_no)
ValueError: invalid literal for int() with base 10: 'set'
```

disconnect3d:
Fwiw based on https://en.opensuse.org/Portal:LoongArch :

> LoongArch is a new 64-bit architecture developed by the Chinese company Loongson Technology. It is based on the MIPS architecture with some modifications and extensions reminiscent to the RISC-V architecture.


Since I made this research, this arch is not supported by the libs we depend on:
* Capstone (https://github.com/capstone-engine/capstone/issues/2065, https://github.com/capstone-engine/capstone/pull/1949#issuecomment-1537793918)
* Unicorn Engine (https://github.com/unicorn-engine/unicorn/pull/1903)
* Pwntools (obviously, from the error)

[...]

Issue #2250, opened 2024-06-22
Support multiple calling conventions per architecture


OBarronCS:
Currently, for each architecture Pwndbg supports, there is a single assumed calling convention for that architecture. These are defined in https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/lib/abi.py. The codebase currently appears to assume the ABI that is most common for workstation/non-embedded uses.

It would be great to support multiple ABI's per architecture. For example, 64-bit MIPS has three different calling conventions, `o32`, `n32`, and `n64` - each which use different amounts of arguments for syscalls/functions, different stack alignments, and even have a different register sets. The calling convention is decided at compile time. 

The following table showing the registers that are present for each of these MIPS ABI's is from the book "See MIPS Run"
![image](https://github.com/pwndbg/pwndbg/assets/55004530/2be3d1b7-08cd-41f9-8849-3f74842421ef)

Note that the o32 and n32/n64 have different register names  - and I verified manually that GDB does indeed use the register names in the screenshot above for each of these ABI's. 

The same concept applies to most other architectures and their syscall ABI's. For example, x86 has the x86_64 ABI, x32, and IA-32, and Arm and PowerPC both have multiple different syscalls ABIs. They take different number of arguments, size of pointers, and have different syscall numbers. Here's a great reference for the different ABIs - https://syscalls.mebeim.net/?table=x86/64/x64/v6.6. Some other architectures have a specific ABI that is more prevalent in embedded environments. 

To allow for different calling conventions, we would need to modify both the abi.py file and the register set definitions
https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/lib/regs.py. We would need to query the ABI itself and use it to decide which register set, ABI, and SyscallABI to use. 
 

For syscall numbers, we pull those from Pwntools, which itself pulls them from header files in dietlibc (https://www.fefe.de/dietlibc/), which at first glance appears to lack support for a wide range of syscall ABI's.

OBarronCS:
Related issue: #1097

plusls:
we can get abi by `show mips abi`
```
pwndbg> show mips abi
The MIPS ABI is set automatically (currently "n32").
```

PR #2691, opened 2025-01-22
Add initial support for Loongarch64

OBarronCS:
This PR adds initial support for the Loongarch64 architecture. Previously, attempting to debug a program in this architecture caused a crash.

There's a new register set definition for Loongarch64 based on the register definitions seen here - https://docs.kernel.org/arch/loongarch/introduction.html

The latest stable release of the Capstone Engine doesn't support Loongarch64. However, using GDB's internal mechanism to disassemble instructions, we can still show some form of disassembly (without any annotations or emulation). I added a new class `ManualPwndbgInstruction` that has the same interface as `PwndbgInstruction`, and gets its data from the GDB API.

I'm not fully familiar with the internals of the new LLDB support, so there may be something special that needs to be implemented and tested for LLDB. I added some GitHub comments on some things to consider/look at.

image

### DIFF
--- a/pwndbg/aglib/arch.py
+++ b/pwndbg/aglib/arch.py
@@ -7,6 +7,7 @@ import pwnlib
 import pwndbg
 from pwndbg.lib.arch import Arch
 
+# List of architectures - used when determining GDB arch, and in OnlyWithArch
 ARCHS = (
     "x86-64",
     "i386",
@@ -19,10 +20,11 @@ ARCHS = (
     "riscv:rv32",
     "riscv:rv64",
     "riscv",
+    "loongarch64",
 )
 
 
-# mapping between gdb and pwntools arch names
+# mapping between pwndbg and pwntools arch names
 pwnlib_archs_mapping = {
     "x86-64": "amd64",
     "i386": "i386",
@@ -35,6 +37,7 @@ pwnlib_archs_mapping = {
     "armcm": "thumb",
     "rv32": "riscv32",
     "rv64": "riscv64",
+    "loongarch64": "none",
 }
 
 
--- a/pwndbg/aglib/disasm/instruction.py
+++ b/pwndbg/aglib/disasm/instruction.py
@@ -40,6 +40,10 @@ from capstone.sparc import SPARC_INS_JMP
 from capstone.sparc import SPARC_INS_JMPL
 from capstone.x86 import X86_INS_JMP
 from capstone.x86 import X86Op
+from typing_extensions import override
+
+import pwndbg.dbg
+from pwndbg.dbg import DisassembledInstruction
 
 # Architecture specific instructions that mutate the instruction pointer unconditionally
 # The Capstone RET and CALL groups are also used to filter CALL and RET types when we check for unconditional jumps,
@@ -119,17 +123,12 @@ CAPSTONE_ARCH_MAPPING_STRING = {
 # in the disasm view output (see 'pwndbg.color.disasm.instruction()'), as well as for
 # Pwndbg commands like "nextcall" that need to know the instructions target to set breakpoints
 class PwndbgInstruction:
-    def __init__(self, cs_insn: CsInsn | None) -> None:
+    def __init__(self, cs_insn: CsInsn) -> None:
         self.cs_insn: CsInsn = cs_insn
         """
         The underlying Capstone instruction, if present.
         Ideally, only the enhancement code will access the 'cs_insn' property
         """
-
-        # None if Capstone don't support the arch being disassembled
-        # See "make_simple_instruction" function
-        if cs_insn is None:
-            return
 
         self.address: int = cs_insn.address
 
@@ -555,29 +554,81 @@ class EnhancedOperand:
         return f"[{info}]"
 
 
-def make_simple_instruction(address: int) -> PwndbgInstruction:
-    """
-    Instantiate a PwndbgInstruction for an architecture that Capstone/pwndbg doesn't support (as defined in the CapstoneArch structure)
-    """
-    ins = pwndbg.dbg.selected_inferior().disasm(address)
-    asm = ins["asm"].split(maxsplit=1)
+# Represents a disassembled instruction when Capstone does not support the given architecture
+class ManualPwndbgInstruction(PwndbgInstruction):
+    def __init__(self, address: int) -> None:
+        ins: DisassembledInstruction = pwndbg.dbg.selected_inferior().disasm(address)
+        asm = ins["asm"].split(maxsplit=1)
 
-    pwn_ins = PwndbgInstruction(None)
-    pwn_ins.address = address
-    pwn_ins.size = ins["length"]
+        # ERROR: some of the functions depends on this attribute existing
+        # Give this object all the attributes it should have (defined in constructor)
+        self.address = address
+        self.size = ins["length"]
 
-    pwn_ins.mnemonic = asm[0].strip()
-    pwn_ins.op_str = asm[1].strip() if len(asm) > 1 else ""
+        self.mnemonic = asm[0].strip()
+        self.op_str = asm[1].strip() if len(asm) > 1 else ""
+        self.groups = set()
 
-    pwn_ins.next = address + pwn_ins.size
-    pwn_ins.target = pwn_ins.next
+        # Set Capstone ID to -1
+        self.id = -1
 
-    pwn_ins.groups = []
+        self.operands = []
 
-    pwn_ins.condition = InstructionCondition.UNDETERMINED
+        self.asm_string = "%-06s %s" % (self.mnemonic, self.op_str)
 
-    pwn_ins.annotation = None
+        self.next = address + self.size
+        self.target = self.next
+        self.target_const = None
 
-    pwn_ins.operands = []
+        self.condition = InstructionCondition.UNDETERMINED
 
-    return pwn_ins
+        self.declare_conditional = None
+        self.declare_is_unconditional_jump = False
+        self.force_unconditional_jump_target = False
+
+        self.annotation = None
+
+        self.annotation_padding = None
+
+        self.syscall = None
+        self.causes_branch_delay = False
+
+        self.split = SplitType.NO_SPLIT
+
+        self.emulated = False
+
+    @property
+    def bytes(self) -> bytearray:
+        return bytearray()
+
+    @property
+    def call_like(self) -> bool:
+        return False
+
+    @property
+    def jump_like(self) -> bool:
+        return False
+
+    @property
+    def has_jump_target(self) -> bool:
+        return False
+
+    @property
+    def is_conditional_jump(self) -> bool:
+        return False
+
+    @property
+    def is_unconditional_jump(self) -> bool:
+        return False
+
+    @property
+    def is_conditional_jump_taken(self) -> bool:
+        return False
+
+    @override
+    def op_find(self, op_type: int, position: int) -> EnhancedOperand:
+        return None
+
+    @override
+    def op_count(self, op_type: int) -> int:
+        return 0
--- a/pwndbg/aglib/disasm/instruction.py
+++ b/pwndbg/aglib/disasm/instruction.py
@@ -126,7 +126,7 @@ class PwndbgInstruction:
     def __init__(self, cs_insn: CsInsn) -> None:
         self.cs_insn: CsInsn = cs_insn
         """
-        The underlying Capstone instruction, if present.
+        The underlying Capstone instruction object.
         Ideally, only the enhancement code will access the 'cs_insn' property
         """
 
--- a/pwndbg/aglib/disasm/instruction.py
+++ b/pwndbg/aglib/disasm/instruction.py
@@ -5,6 +5,7 @@ from collections import defaultdict
 from enum import Enum
 from typing import Dict
 from typing import List
+from typing import Protocol
 from typing import Set
 
 # Reverse lookup tables for debug printing
@@ -119,15 +120,69 @@ CAPSTONE_ARCH_MAPPING_STRING = {
 }
 
 
+# Interface for enhanced instructions - there are two implementations defined in this file
+class PwndbgInstruction(Protocol):
+    cs_insn: CsInsn
+    address: int
+    size: int
+    mnemonic: str
+    op_str: str
+    groups: Set[int]
+    id: int
+    operands: List[EnhancedOperand]
+    asm_string: str
+    next: int
+    target: int
+    target_string: str | None
+    target_const: bool | None
+    condition: InstructionCondition
+    declare_conditional: bool | None
+    declare_is_unconditional_jump: bool
+    force_unconditional_jump_target: bool
+    annotation: str | None
+    annotation_padding: int | None
+    syscall: int | None
+    syscall_name: str | None
+    causes_branch_delay: bool
+    split: SplitType
+    emulated: bool
+
+    @property
+    def call_like(self) -> bool: ...
+
+    @property
+    def jump_like(self) -> bool: ...
+
+    @property
+    def has_jump_target(self) -> bool: ...
+
+    @property
+    def is_conditional_jump(self) -> bool: ...
+
+    @property
+    def is_unconditional_jump(self) -> bool: ...
+
+    @property
+    def is_conditional_jump_taken(self) -> bool: ...
+
+    @property
+    def bytes(self) -> bytearray: ...
+
+    def op_find(self, op_type: int, position: int) -> EnhancedOperand: ...
+
+    def op_count(self, op_type: int) -> int: ...
+
+
 # This class is used to provide context to an instructions execution, used both
 # in the disasm view output (see 'pwndbg.color.disasm.instruction()'), as well as for
 # Pwndbg commands like "nextcall" that need to know the instructions target to set breakpoints
-class PwndbgInstruction:
+# The information in this class is backed by metadata from Capstone
+class PwndbgInstructionImpl(PwndbgInstruction):
     def __init__(self, cs_insn: CsInsn) -> None:
         self.cs_insn: CsInsn = cs_insn
         """
         The underlying Capstone instruction object.
-        Ideally, only the enhancement code will access the 'cs_insn' property
+        Only the enhancement code should access the 'cs_insn' property
         """
 
         self.address: int = cs_insn.address
@@ -172,7 +227,7 @@ class PwndbgInstruction:
         # in pwndbg.aglib.disasm.arch.py
         # ***********
 
-        self.asm_string: str = "%-06s %s" % (self.mnemonic, self.op_str)
+        self.asm_string: str = f"{self.mnemonic:<6} {self.op_str}"
         """
         The full string representing the instruction - `mov    rdi, rsp` with appropriate padding.
 
@@ -554,14 +609,25 @@ class EnhancedOperand:
         return f"[{info}]"
 
 
-# Represents a disassembled instruction when Capstone does not support the given architecture
+# Represents a disassembled instruction
+# Conforms to the PwndbgInstruction interface
 class ManualPwndbgInstruction(PwndbgInstruction):
     def __init__(self, address: int) -> None:
+        """
+        This class provides an implementation of PwndbgInstruction for cases where the architecture
+        at hand is not supported by the Capstone disassembler. The backing information is sourced from
+        GDB/LLDB's built-in disassemblers.
+
+        Instances of this class do not go through the 'enhancement' process due to lacking important information provided by Capstone.
+        As a result of this, some of the methods raise NotImplementedError, because if they are called it indicates a bug elsewhere in the codebase.
+        """
         ins: DisassembledInstruction = pwndbg.dbg.selected_inferior().disasm(address)
         asm = ins["asm"].split(maxsplit=1)
 
-        # ERROR: some of the functions depends on this attribute existing
-        # Give this object all the attributes it should have (defined in constructor)
+        # The enhancement code assumes this value exists.
+        # However, a ManualPwndbgInstruction should never be used in the enhancement code.
+        self.cs_insn: CsInsn = None
+
         self.address = address
         self.size = ins["length"]
 
@@ -574,10 +640,11 @@ class ManualPwndbgInstruction(PwndbgInstruction):
 
         self.operands = []
 
-        self.asm_string = "%-06s %s" % (self.mnemonic, self.op_str)
+        self.asm_string = f"{self.mnemonic:<6} {self.op_str}"
 
         self.next = address + self.size
         self.target = self.next
+        self.target_string = None
         self.target_const = None
 
         self.condition = InstructionCondition.UNDETERMINED
@@ -591,6 +658,8 @@ class ManualPwndbgInstruction(PwndbgInstruction):
         self.annotation_padding = None
 
         self.syscall = None
+        self.syscall_name = None
+
         self.causes_branch_delay = False
 
         self.split = SplitType.NO_SPLIT
@@ -599,6 +668,9 @@ class ManualPwndbgInstruction(PwndbgInstruction):
 
     @property
     def bytes(self) -> bytearray:
+        # GDB simply doesn't provide us with the raw bytes.
+        # However, it is important that this returns a valid bytearray,
+        # since the disasm code indexes this for nearpc-num-opcode-bytes.
         return bytearray()
 
     @property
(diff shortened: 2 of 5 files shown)


OBarronCS:
References #2690

pwndbg/aglib/arch.py
```diff
@@ -7,6 +7,7 @@
 import pwndbg
 from pwndbg.lib.arch import Arch
 
+# List of architectures - used when determining GDB arch, and in OnlyWithArch
```
OBarronCS, line 10: A grep on this over the codebase shows it is used in 3 locations. Perhaps we should better determine what this list defines, especially regarding the duplicates with `riscv`:
```
    "riscv:rv32",
    "riscv:rv64",
    "riscv",
```
Currently, it is used to check that the output of gdb's `show architecture` is an architecture pwndbg understands, and in a function called `OnlyWithArch`

Maybe this could be used to make an authoritative list of all `pwndbg architecture names` - if so, we might want to remove the riscv duplicates.

Another thing worth looking into, is that `OnlyWithArch` (which references this) looks like it might be defined twice:
https://github.com/pwndbg/pwndbg/blob/53bc84fe86bd8b634d133ba02412ebf54508c07f/pwndbg/aglib/proc.py#L43
and
https://github.com/pwndbg/pwndbg/blob/53bc84fe86bd8b634d133ba02412ebf54508c07f/pwndbg/commands/__init__.py#L335

pwndbg/aglib/arch.py
```diff
@@ -35,6 +37,7 @@
     "armcm": "thumb",
     "rv32": "riscv32",
     "rv64": "riscv64",
+    "loongarch64": "none",
```
OBarronCS, line 40: pwntools allows the `arch` value to be the string `none`. This has the side-effect that some pwndbg commands that depend on pwnlib, such as `pwnlib.asm`, would fail to work. However, I believe these cases are somewhat rare and are always called through a user-invoked command.

pwndbg/aglib/disasm/instruction.py
```diff
@@ -555,29 +554,81 @@ def __repr__(self) -> str:
         return f"[{info}]"
 
 
-def make_simple_instruction(address: int) -> PwndbgInstruction:
```
OBarronCS, line 558: Upon testing, this function in conjunction with PwndbgInstruction never worked. PwndbgInstruction is deeply coupled with the underlying Capstone object, as many of its methods depend on it. I extracted this into a separate `ManualPwndbgInstruction` which implements the same interface as PwndbgInstruction to avoid any crashes.

pwndbg/dbg/gdb/__init__.py
```diff
@@ -678,6 +678,8 @@ def arch(self) -> pwndbg.dbg_mod.Arch:
             arch = gdb.execute("show architecture", to_string=True).strip()
             not_exactly_arch = True
 
+        arch = arch.lower()
```
OBarronCS, line 681: GDB returns `Loongarch64`, so this makes it lowercase.

patryk4815:
LLDB has weird behavior, probably not supported in LLDB 19.0.5 yet?
https://github.com/llvm/llvm-project/issues/112693 - here all issues are resolved, maybe LLDB 20 will support 🤔 

LLDB from latest `main` branch working fine!

pwndbg/aglib/arch.py
xtexx: https://github.com/Gallopsled/pwntools/pull/2529

patryk4815:
LLDB from latest `main` branch working fine!

we didn't implement disasm for LLDB
```console
pwndbg-lldb> b main
Breakpoint 1: where = hello`main, address = 0x0000000120001940
pwndbg-lldb> c
Warning: Avoided exploring possible address 0xffffa24ef790.
You can explicitly explore it with `vmmap_explore 0xffffa24ef000`
Unknown register: 't0'
Unknown register: 't1'
Unknown register: 't2'
Unknown register: 't3'
Unknown register: 't4'
Unknown register: 't5'
Unknown register: 't6'
Unknown register: 't7'
Unknown register: 't8'
Unknown register: 's0'
Unknown register: 's1'
Unknown register: 's2'
Unknown register: 's3'
Unknown register: 's4'
Unknown register: 's5'
Unknown register: 's6'
Unknown register: 's7'
Unknown register: 's8'
Warning: Avoided exploring possible address 0xffffa236d1d0.
You can explicitly explore it with `vmmap_explore 0xffffa236d000`
Warning: Avoided exploring possible address 0x120001940.
You can explicitly explore it with `vmmap_explore 0x120001000`
│ /nix/store/yy3xsn1xrfwaknran67cglj34blhr3fx-pwndbg-lldb/share/pwndbg/pwndbg/commands/__init__.py │
│ :183 in __call__                                                                                 │
│   180 │                                                                                          │
│   181 │   def __call__(self, *args: Any, **kwargs: Any) -> str | None:                           │
│   182 │   │   try:                                                                               │
│ ❱ 183 │   │   │   return self.function(*args, **kwargs)                                          │
│   184 │   │   except TypeError:                                                                  │
│   185 │   │   │   print(f"{self.function.__name__.strip()!r}: {self.function.__doc__.strip()}"   │
│   186 │   │   │   pwndbg.exception.handle(self.function.__name__)                                │
│ /nix/store/yy3xsn1xrfwaknran67cglj34blhr3fx-pwndbg-lldb/share/pwndbg/pwndbg/commands/context.py: │
│ 658 in context                                                                                   │
│    655 │   │   │   │   result_settings[target].update(settings)                                  │
│    656 │   │   │   │   with target as out:                                                       │
│    657 │   │   │   │   │   result[target].extend(                                                │
│ ❱  658 │   │   │   │   │   │   func(                                                             │
│    659 │   │   │   │   │   │   │   target=out,                                                   │
│    660 │   │   │   │   │   │   │   width=settings.get("width", None),                            │
│    661 │   │   │   │   │   │   │   with_banner=settings.get("banner_top", True),                 │
│ /nix/store/yy3xsn1xrfwaknran67cglj34blhr3fx-pwndbg-lldb/share/pwndbg/pwndbg/commands/context.py: │
│ 327 in _serve_context_history                                                                    │
│    324 │   │   │   ):                                                                            │
│    325 │   │   │   │   current_output = context_history[section_name][-1]                        │
│    326 │   │   │   else:                                                                         │
│ ❱  327 │   │   │   │   current_output = function(*a, **kw)                                       │
│    328 │   │   │   if (                                                                          │
│    329 │   │   │   │   len(context_history[section_name]) == 0                                   │
│    330 │   │   │   │   or context_history[section_name][-1] != current_output                    │
│ /nix/store/yy3xsn1xrfwaknran67cglj34blhr3fx-pwndbg-lldb/share/pwndbg/pwndbg/commands/context.py: │
│ 897 in context_disasm                                                                            │
│    894 │   if cs is not None and cs.syntax != syntax:                                            │
│    895 │   │   pwndbg.lib.cache.clear_caches()                                                   │
│    896 │                                                                                         │
│ ❱  897 │   result = try_emulate_if_bug_disable(                                                  │
│    898 │   │   lambda: pwndbg.aglib.nearpc.nearpc(                                               │
│    899 │   │   │   lines=disasm_lines // 2,                                                      │
│    900 │   │   │   emulate=bool(not pwndbg.config.emulate == "off"),                             │
│ /nix/store/yy3xsn1xrfwaknran67cglj34blhr3fx-pwndbg-lldb/share/pwndbg/pwndbg/commands/context.py: │
│ 873 in try_emulate_if_bug_disable                                                                │
│    870                                                                                           │
│    871 def try_emulate_if_bug_disable(handler: Callable[[], T]) -> T:                            │
│    872 │   try:                                                                                  │
│ ❱  873 │   │   return handler()                                                                  │
│    874 │   except U.UcError as e:                                                                │
│    875 │   │   print(                                                                            │
│    876 │   │   │   message.warn(                                                                 │
│ /nix/store/yy3xsn1xrfwaknran67cglj34blhr3fx-pwndbg-lldb/share/pwndbg/pwndbg/commands/context.py: │
│ 898 in <lambda>                                                                                  │
│    895 │   │   pwndbg.lib.cache.clear_caches()                                                   │
│    896 │                                                                                         │
│    897 │   result = try_emulate_if_bug_disable(                                                  │
│ ❱  898 │   │   lambda: pwndbg.aglib.nearpc.nearpc(                                               │
│    899 │   │   │   lines=disasm_lines // 2,                                                      │
│    900 │   │   │   emulate=bool(not pwndbg.config.emulate == "off"),                             │
│    901 │   │   │   use_cache=True,                                                               │
│ /nix/store/yy3xsn1xrfwaknran67cglj34blhr3fx-pwndbg-lldb/share/pwndbg/pwndbg/aglib/nearpc.py:143  │
│ in nearpc                                                                                        │
│   140 │   #         for line in symtab.linetable():                                              │
│   141 │   #             pc_to_linenos[line.pc].append(line.line)                                 │
│   142 │                                                                                          │
│ ❱ 143 │   instructions, index_of_pc = pwndbg.aglib.disasm.near(                                  │
│   144 │   │   pc, lines, emulate=emulate, show_prev_insns=not repeat, use_cache=use_cache, lin   │
│   145 │   )                                                                                      │
│   146                                                                                            │
│ /nix/store/yy3xsn1xrfwaknran67cglj34blhr3fx-pwndbg-lldb/share/pwndbg/pwndbg/aglib/disasm/__init_ │
│ _.py:397 in near                                                                                 │
│   394 │   │   │   │   raise                                                                      │
│   395 │                                                                                          │
│   396 │   # Start at the current instruction using emulation if available.                       │
│ ❱ 397 │   current = one(address, emu, put_cache=True)                                            │
│   398 │                                                                                          │
│   399 │   if DEBUG_ENHANCEMENT:                                                                  │
│   400 │   │   if emu and not emu.last_step_succeeded:                                            │
│ /nix/store/yy3xsn1xrfwaknran67cglj34blhr3fx-pwndbg-lldb/share/pwndbg/pwndbg/aglib/disasm/__init_ │
│ _.py:244 in one                                                                                  │
│   241 │   │   return None                                                                        │
│   242 │                                                                                          │
│   243 │   # A for loop in case this returns an empty list                                        │
│ ❱ 244 │   for insn in get(address, 1, emu, enhance=enhance, from_cache=from_cache, put_cache=p   │
│   245 │   │   if put_backward_cache:                                                             │
│   246 │   │   │   backward_cache[insn.next] = insn.address                                       │
│   247 │   │   return insn                                                                        │
│ /nix/store/yy3xsn1xrfwaknran67cglj34blhr3fx-pwndbg-lldb/share/pwndbg/pwndbg/aglib/disasm/__init_ │
│ _.py:279 in get                                                                                  │
│   276 │                                                                                          │
│   277 │   retval: List[PwndbgInstruction] = []                                                   │
│   278 │   for _ in range(instructions):                                                          │
│ ❱ 279 │   │   i = get_one_instruction(                                                           │
│   280 │   │   │   address, emu, enhance=enhance, from_cache=from_cache, put_cache=put_cache      │
│   281 │   │   )                                                                                  │
│   282 │   │   if i is None:                                                                      │
│ /nix/store/yy3xsn1xrfwaknran67cglj34blhr3fx-pwndbg-lldb/share/pwndbg/pwndbg/aglib/disasm/__init_ │
│ _.py:208 in get_one_instruction                                                                  │
│   205 │   │   │   return cached                                                                  │
│   206 │                                                                                          │
│   207 │   if pwndbg.aglib.arch.current not in CapstoneArch:                                      │
│ ❱ 208 │   │   return ManualPwndbgInstruction(address)                                            │
│   209 │                                                                                          │
│   210 │   md = get_disassembler(address)                                                         │
│   211 │   size = VariableInstructionSizeMax.get(pwndbg.aglib.arch.current, 4)                    │
│ /nix/store/yy3xsn1xrfwaknran67cglj34blhr3fx-pwndbg-lldb/share/pwndbg/pwndbg/aglib/disasm/instruc │
│ tion.py:560 in __init__                                                                          │
│   557 # Represents a disassembled instruction when Capstone does not support the given archite   │
│   558 class ManualPwndbgInstruction(PwndbgInstruction):                                          │
│   559 │   def __init__(self, address: int) -> None:                                              │
│ ❱ 560 │   │   ins: DisassembledInstruction = pwndbg.dbg.selected_inferior().disasm(address)      │
│   561 │   │   asm = ins["asm"].split(maxsplit=1)                                                 │
│   562 │   │                                                                                      │
│   563 │   │   # ERROR: some of the functions depends on this attribute existing                  │
│ /nix/store/yy3xsn1xrfwaknran67cglj34blhr3fx-pwndbg-lldb/share/pwndbg/pwndbg/dbg/__init__.py:537  │
│ in disasm                                                                                        │
│    534 │   │   space of the running process, or `None` if there's no valid instruction           │
│    535 │   │   at that address.                                                                  │
│    536 │   │   """                                                                               │
│ ❱  537 │   │   raise NotImplementedError()                                                       │
│    538 │                                                                                         │
│    539 │   # We probably want to expose a better module interface in the future, but,            │
│    540 │   # for now, this is good enough.                                                       │
NotImplementedError
> /nix/store/yy3xsn1xrfwaknran67cglj34blhr3fx-pwndbg-lldb/share/pwndbg/pwndbg/dbg/__init__.py(537)disasm()
-> raise NotImplementedError()
(Pdb)
```

pwndbg/lib/regs.py
```diff
@@ -622,6 +622,54 @@ def __iter__(self) -> Iterator[str]:
     retval="a0",
 )
 
+# https://docs.kernel.org/arch/loongarch/introduction.html
+loongarch64 = RegisterSet(
+    pc="pc",
+    stack="sp",
+    frame="fp",
+    retaddr=("ra",),
+    gpr=(
```
patryk4815, line 631: ~~LLDB use different registers 🤔~~
 `s[0-9]` registers and `t[0-9]` missing aliases :(

image


image
 

image

OBarronCS:
Very cool! 

On the register names:

GDB appears follows the naming conventions in this table in https://docs.kernel.org/arch/loongarch/introduction.html

It understands that `a0` is an alias for `r4` (and in GDB disassembly, the alias is the one that appears)
image

Looks like LLDB perhaps doesn't have this alias for all of these registers built into it (maybe added in LLDB 20?) We could fix with an extra LLDB register mapping struct for a layer of indirection? Or simply another register set for LLDB?

xtexx:
It seems that LLDB have these aliases defined since LLVM 16: https://github.com/llvm/llvm-project/blob/ed34590c1acb300abe51d2686148c62e8c517993/lldb/source/Plugins/Process/Utility/RegisterInfos_loongarch64.h#L92-L112
Maybe something is broken in LLDB?

pwndbg/aglib/arch.py
xtexx: I used a shorter architecture name `loong64` in my PR to pwntools. This name is also widely used and using it avoids having to reformat a large number of pwntools code. So if the pwntools PR get merged in the future, we may need to change this to `loong64`.

Update: per suggestions from pwntools, the pwnlib architecture name is now `loongarch64`.

patryk4815:
Reported to upstream - https://github.com/llvm/llvm-project/issues/123903
@OBarronCS upstream confirm this is issue with lldb+qemu-user, on real linux+lldb reports good registers

pwndbg/aglib/disasm/instruction.py
```diff
@@ -119,18 +123,13 @@ class SplitType(Enum):
 # in the disasm view output (see 'pwndbg.color.disasm.instruction()'), as well as for
 # Pwndbg commands like "nextcall" that need to know the instructions target to set breakpoints
 class PwndbgInstruction:
-    def __init__(self, cs_insn: CsInsn | None) -> None:
+    def __init__(self, cs_insn: CsInsn) -> None:
         self.cs_insn: CsInsn = cs_insn
         """
         The underlying Capstone instruction, if present.
```
disconnect3d: Please update the docstring. ~Iiuc previously we could have a `None` instruction here and now we drop this, right?~ I just read the comment below and iiuc this case was never working and happening ;).

```suggestion
        The underlying Capstone instruction object.
```

pwndbg/aglib/disasm/instruction.py
```diff
@@ -555,29 +554,81 @@ def __repr__(self) -> str:
...
 
-def make_simple_instruction(address: int) -> PwndbgInstruction:
-    """
-    Instantiate a PwndbgInstruction for an architecture that Capstone/pwndbg doesn't support (as defined in the CapstoneArch structure)
-    """
-    ins = pwndbg.dbg.selected_inferior().disasm(address)
-    asm = ins["asm"].split(maxsplit=1)
+# Represents a disassembled instruction when Capstone does not support the given architecture
+class ManualPwndbgInstruction(PwndbgInstruction):
+    def __init__(self, address: int) -> None:
```
disconnect3d, line 615: Is this okay that we never call `super().__init__(...)`? Seems like it may make; please document this.

pwndbg/aglib/disasm/instruction.py
```diff
@@ -555,29 +554,81 @@ def __repr__(self) -> str:
...
+        ins: DisassembledInstruction = pwndbg.dbg.selected_inferior().disasm(address)
+        asm = ins["asm"].split(maxsplit=1)
+
+        # ERROR: some of the functions depends on this attribute existing
+        # Give this object all the attributes it should have (defined in constructor)
+        self.address = address
+        self.size = ins["length"]
+
+        self.mnemonic = asm[0].strip()
+        self.op_str = asm[1].strip() if len(asm) > 1 else ""
```
disconnect3d, line 635: can `ins["asm"]` be splitted once and then not splitted here maybe?

pwndbg/aglib/disasm/instruction.py
```diff
@@ -555,29 +554,81 @@ def __repr__(self) -> str:
...
+        self.groups = set()
+
+        # Set Capstone ID to -1
+        self.id = -1
+
+        self.operands = []
+
+        self.asm_string = "%-06s %s" % (self.mnemonic, self.op_str)
```
disconnect3d: Use f"" if possible

pwndbg/aglib/disasm/instruction.py
```diff
@@ -555,29 +554,81 @@ def __repr__(self) -> str:
...
+    def is_conditional_jump_taken(self) -> bool:
+        return False
+
+    @override
+    def op_find(self, op_type: int, position: int) -> EnhancedOperand:
+        return None
+
+    @override
+    def op_count(self, op_type: int) -> int:
+        return 0
```
disconnect3d: This ain't true, is it?

pwndbg/aglib/disasm/instruction.py
disconnect3d: Additionally, can we add a unit test that will ensure `ManualPwndbgInstruction` has the same set of fields as `PwndbgInstruction`? I mean, since we are not calling `super().__init__()` this is not ensured and if someone changes one, they should probably change the other one?

pwndbg/aglib/disasm/instruction.py
patryk4815: "asm" returns eg: `mov    r4, #0` so `strip` must be called

pwndbg/aglib/arch.py
OBarronCS: Awesome! `loongarch64` is great. I made an issue, #2703, to consolidate our architecture name list - we'll add loongarch64 to it

pwndbg/aglib/disasm/instruction.py
OBarronCS: Good catch - fixing this.

pwndbg/aglib/disasm/instruction.py
OBarronCS: Right - I think we can determine this based on the `asm` string the dissembler returns.

pwndbg/lib/regs.py
OBarronCS: This does speak to a larger question:

In the future, we may want to add a way to switch between register sets, and select it based on more than just the architecture name.

For example. in #2250, we found that MIPS actually changes it's register names based on the ABI in use. In this case, we would need to change the register set to support the correct one for the given debugge/architecture/abi combination

OBarronCS:
> Reported to upstream - [llvm/llvm-project#123903](https://github.com/llvm/llvm-project/issues/123903) @OBarronCS upstream confirm this is issue with lldb+qemu-user, on real linux+lldb reports good registers

Very interesting...

Looks like a fix is coming in a future LLDB version: https://github.com/llvm/llvm-project/pull/124059

In the meanwhile, a possible solution is to have an extra map/layer of indirection like you mentioned earlier so that a Pwndbg user can do "reg s0", then on LLDB this is mapped to register r23 which all versions of LLBD will return. Ubuntu 24.04 is on LLDB 18.

pwndbg/aglib/disasm/instruction.py
disconnect3d: Either this, or, at least document it properly I guess (tho, maybe this should raise some exception or sth if it returns wrong number; not sure what code uses it in the end)

disconnect3d:
> In the meanwhile, a possible solution is to have an extra map/layer of indirection like you mentioned earlier so that a Pwndbg user can do "reg s0", then on LLDB this is mapped to register r23 which all versions of LLBD will return. Ubuntu 24.04 is on LLDB 18.

I think we can skip the layer of indirection if this is going to be fixed in upcomming LLDB release. Sure, users would have to use the working register name until they update LLDB but we can also make a Pwndbg release with a fixed LLDB version when it is ready.

Fwiw afaiu we do not support LLDB 18 as per the [compatibility table (which says we support LLDB 19+)](https://github.com/pwndbg/pwndbg?tab=readme-ov-file#compatibility-table).

pwndbg/aglib/disasm/instruction.py
OBarronCS: On further look - documenting it is the solution. The only code that uses this is the annotation creation code, which will only ever be called on instructions disassembled by Capstone. Without all of the metadata returned by Capstone, the enhancement process can't do much.

disconnect3d: Maybe `raise NotImplementedError` then

patryk4815:
> > Reported to upstream - [llvm/llvm-project#123903](https://github.com/llvm/llvm-project/issues/123903) @OBarronCS upstream confirm this is issue with lldb+qemu-user, on real linux+lldb reports good registers
> 
> Very interesting...
> 
> Looks like a fix is coming in a future LLDB version: [llvm/llvm-project#124059](https://github.com/llvm/llvm-project/pull/124059)
> 
> In the meanwhile, a possible solution is to have an extra map/layer of indirection like you mentioned earlier so that a Pwndbg user can do "reg s0", then on LLDB this is mapped to register r23 which all versions of LLBD will return. Ubuntu 24.04 is on LLDB 18.

lets skip this part

pwndbg/aglib/disasm/instruction.py
xtexx: I also wonder whether this should return the real operand count or the op count in the assembly (:
For example, both GCC and LLVM support a `move` instruction on LoongArch64. `move $t0, $t1` will be the same as `and.d $t0, $t1, %0` (or `add.d`, I don't remember well). And the disassemblers use the short form as much as possible. But for real CPUs, there are three operands encoded.

OBarronCS: @disconnect3d I went with the `raise NotImplementedError` solution.

@xtexChooser That is a really great point. The Capstone disassembler (in the configuration we use it) has the behavior you describe where it aliases instructions and attempts to simplify them and remove operands that are "zero" to attempt to bring them to the shortest form possible. Due to this, we have been returning this 'short form' info.

pwndbg/aglib/disasm/instruction.py
OBarronCS: I agree with these issues - the latest commits fix them by using Python `Protocols` as a form of static type-safe interface, and the PwndbgInstruction backed by Capstone and the one backed by the information returned by the debugger implement the interface.

OBarronCS:
The latest commit makes the suggested fixes, and uses Python `Protocols` to define a PwndbgInstruction interface which the two implementations conform to (the capstone backed one, and the debugger-disassembler backed one). Python Protocols don't have any runtime overhead - they are meant to provide static "duck typing" checks, intended to be used to define interfaces. Our `./lint.sh` script will throw an error if the implementations do not conform to the interface. This ensures that the two implementations are always in sync to avoid run-time crashes.

As discussed above, I kept the registers untouched.

disconnect3d (approved):

PR #2693, opened 2025-01-22
implement LLDB disasm func

patryk4815:
implement LLDB disasm func, pwndbg fallback there when capstone is not available


patryk4815:
@OBarronCS disasm for lldb, is it ok?

It working for loongarch64:
image

OBarronCS:
This looks perfect

PR #2697, opened 2025-01-22
Removing redundant function OnlyWithArch from `pwndbg/commands/__init__.py`

### LINKED ISSUE #2692: OnlyWithArch is duplicated
As pointed in https://github.com/pwndbg/pwndbg/pull/2691/files#r1924611246 the `OnlyWithArch` is defined twice:
https://github.com/pwndbg/pwndbg/blob/53bc84fe86bd8b634d133ba02412ebf54508c07f/pwndbg/aglib/proc.py#L43
and
https://github.com/pwndbg/pwndbg/blob/53bc84fe86bd8b634d133ba02412ebf54508c07f/pwndbg/commands/__init__.py#L335

We should refactor this and keep just one? ;)

AadishJ:
Resolves https://github.com/pwndbg/pwndbg/issues/2692

This PR removes the function from init.py and a few other dependent files is changed to call that function from pwndbg.aglib.proc instead of pwndbg.commands

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -332,31 +332,6 @@ def OnlyWhenUserspace(function: Callable[P, T]) -> Callable[P, Optional[T]]:
     return _OnlyWhenUserspace
 
 
-def OnlyWithArch(arch_names: List[str]) -> Callable[[Callable[P, T]], Callable[P, Optional[T]]]:
-    """Decorates function to work only with the specified archictectures."""
-    for arch in arch_names:
-        if arch not in pwndbg.aglib.arch_mod.ARCHS:
-            raise ValueError(
-                f"OnlyWithArch used with unsupported arch={arch}. Must be one of {', '.join(arch_names)}"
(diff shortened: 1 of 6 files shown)


OBarronCS:
It looks like the tests might be failing on a circular import.`pwndbg.commands` imports `pwndbg.aglib.regs`, and this adds an import in the other direction as well.

A solution that could work is to switch the file that is used - instead, keep OnlyWithArch in `pwndbg,aglib.proc`, and remove it from commands.

The codebase overall has minimal imports from `pwndbg.command`, ideally to keep it someone isolated, as the commands/ folder contains logic on user-invoked commands. This OnlyWithArch might more naturally exist in `pwndbg.aglib.proc`, and fix the circular import.

AadishJ:
@OBarronCS I have made the changes and I think the issue has been fixed. Please check :)

OBarronCS:
Nice! This looks great :)

disconnect3d:
Thanks!

PR #2705, opened 2025-01-24
lldb: add support for attaching via `attach <pid|name>`

### LINKED ISSUE #2700: pwndbg-lldb on macOS: can't continue or see regs after attach
Tested on macOS. I did `pwndbg-lldb <binary>` then `attach <binary>`. Then the `continue` doesn't work. Luckily, `process continue` works:

```
pwndbg-lldb> continue
error: no process
pwndbg-lldb> process continue
Process 5698 resuming
```

However, then my process crashes and I can't see its regs:
```
pwndbg-lldb> process status
Process 5698 stopped
* thread #1, stop reason = EXC_BAD_ACCESS (code=1, address=0x5f)
...
pwndbg-lldb> ctx
Unknown register: 'eax'
Unknown register: 'ebx'
Unknown register: 'ecx'
Unknown register: 'edx'
Unknown register: 'edi'
Unknown register: 'esi'
Unknown register: 'ebp'
Unknown register: 'esp'
Unknown register: 'eip'
Exception occurred: context: int() argument must be a string, a bytes-like object or a real number, not 'NoneType' (<class 'TypeError'>)
```

disconnect3d:
This commit:
- adds support for the `attach <pid|name>` command so that it can attach to a pid or full process name similarly to as in LLDB
- internally, it moves part of `process_attach` to a new `_attach_with_info` internal function
- it adds an `attach` method, so that both `process_attach` and `attach` use the `_attach_with_info`.

Please note that the `attach` LLDB command is a bit confusing. It may seem it is an alias for `process attach` as in here:

```
(lldb) attach -n -w htop
error: 'process attach' doesn't take any arguments.
```

However, in practice it is not. It is an alias for `_regexp-attach`:

```
(lldb) help attach
Attach to process by ID or name.  Expects 'raw' input (see 'help raw-input'.)

Syntax: _regexp-attach <pid> | <process-name>

'attach' is an abbreviation for '_regexp-attach'
```

...which has its own problems:
1) it does not perform any regexp match in practice. passing `hto.*`,
   `hto?` or `hto[p]` will not attach to `htop`. One must pass the full process name like `htop`.
2) it can work without arguments, which is stupid and we should not
   support it?:

```
(lldb) _regexp-attach
There is a running process, detach from it and attach?: [Y/n] y
Process 56358 detached
Process 56358 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGSTOP
    frame #0: 0x00007ff8118bedd2 libsystem_kernel.dylib`__select + 10
libsystem_kernel.dylib`__select:
->  0x7ff8118bedd2 <+10>: jae    0x7ff8118beddc ; <+20>
    0x7ff8118bedd4 <+12>: movq   %rax, %rdi
    0x7ff8118bedd7 <+15>: jmp    0x7ff8118b737b ; cerror
    0x7ff8118beddc <+20>: retq
```

This PR fixes #2700.

### DIFF
--- a/pwndbg/dbg/lldb/repl/__init__.py
+++ b/pwndbg/dbg/lldb/repl/__init__.py
@@ -316,6 +316,13 @@ def run(startup: List[str] | None = None, debug: bool = False) -> None:
                 continue
             # We don't care about other process commands..
 
+        if (bits[0].startswith("at") and "attach".startswith(bits[0])) or \
+           (bits[0].startswith("_regexp-a") and "_regexp-attach".startswith(bits[0])):
+            # `attach` is an alias for `_regexp-attach`
+            # (it is NOT an alias for `process attach` even if it may seem so!)
+            attach(driver, relay, bits[1:], dbg)
+            continue
+
         if bits[0].startswith("ta") and "target".startswith(bits[0]):
             if len(bits) > 1 and bits[1].startswith("c") and "create".startswith(bits[1]):
                 # This is `target create`
@@ -679,6 +686,45 @@ process_attach_unsupported = [
 ]
 
 
+def _attach_with_info(driver: ProcessDriver, relay: EventRelay, dbg: LLDB, info: lldb.SBAttachInfo, cont=False):
+    """
+    Attaches to a process based on SBAttachInfo information
+    """
+    targets = dbg.debugger.GetNumTargets()
+    assert targets < 2
+    if targets == 0:
+        print(message.error("error: no target, create one using the 'target create' command"))
+        return
+
+    # TODO/FIXME: This should ask:
+    # 'There is a running process, detach from it and attach?: [Y/n]'
+    if driver.has_process():
+        print(message.error("error: a process is already being debugged"))
+        return
+
+    io_driver = get_io_driver()
+
+    result = driver.attach(
+        dbg.debugger.GetTargetAtIndex(0),
+        io_driver,
+        info,
+    )
+
+    if not result.success:
+        print(message.error(f"Could not attach to process: {result.description}"))
+        return
+
+    # Continue execution if the user has requested it.
+    if cont:
+        # Same logic applies here as in `process_launch`.
+        relay._set_ignore_resumed(1)
+        driver.cont()
+    else:
+        # Same logic applies here as in `process_launch`.
+        dbg._trigger_event(EventType.STOP)
+
+
+
 def process_attach(driver: ProcessDriver, relay: EventRelay, args: List[str], dbg: LLDB) -> None:
     """
     Attaches to a process with the given arguments.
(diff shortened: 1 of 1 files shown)


pwndbg/dbg/lldb/repl/__init__.py
```diff
@@ -679,60 +687,96 @@ def process_launch(driver: ProcessDriver, relay: EventRelay, args: List[str], db
...
+    # The first two arguments - executable name and wait_for_launch - don't
+    # matter, we set them later. The third one is required, as it tells LLDB the
+    # attach should be asynchronous.
+    info = lldb.SBAttachInfo(None, False, True)
+
+    if args.name is not None:
+        info.SetExecutable(args.name)
+    if args.pid is not None:
+        info.SetProcessID(args.pid)
+    info.SetWaitForLaunch(args.waitfor)
```
disconnect3d, line 747: @mbrla0 Is there a reason we set it here instead of in the constructor 2nd argument?

I left it as it was in the original code, but I was wondering :p.

pwndbg/dbg/lldb/repl/__init__.py
```diff
@@ -679,60 +687,96 @@ def process_launch(driver: ProcessDriver, relay: EventRelay, args: List[str], db
...
+    assert len(args) == 1, "Expected 1 argument: <pid> or <name>"
+    arg = args[0]
+
+    # exec name - None, we set it later (or pid)
+    # wait_for_launch - False, since we don't wait
+    # third arg - tell LLDB to attach asynchronously
+    info = lldb.SBAttachInfo(None, False, True)
+
+    # Argument is pid
+    if arg.isdigit():
```
disconnect3d, line 774: This check is nice bcoz `try: int(...)` would allow some incorrect values:
```
In [1]: '-1'.isdigit()
Out[1]: False

In [2]: '11'.isdigit()
Out[2]: True

In [3]: '+11'.isdigit()
Out[3]: False

In [4]: '1.1'.isdigit()
Out[4]: False
```

This one here only allows for `000` which is invalid pid, but whatever:
```
In [5]: '000'.isdigit()
Out[5]: True
```

patryk4815 (approved):

pwndbg/dbg/lldb/repl/__init__.py
mbrla0: Just felt it read better, is all.

pwndbg/dbg/lldb/repl/__init__.py
```diff
@@ -679,60 +687,96 @@ def process_launch(driver: ProcessDriver, relay: EventRelay, args: List[str], db
...
+    _attach_with_info(driver, relay, dbg, info, cont=do_continue)
+
+
+def attach(driver: ProcessDriver, relay: EventRelay, args: List[str], dbg: LLDB) -> None:
+    """
+    Attaches to a process with the given name or pid based on regex match.
+    Used for `_regexp-attach <pid|name>` (alias for `attach <pid|name>`)
+    Note: for some reason, `attach` does not really take a regex for process name.
+    """
+    assert len(args) == 1, "Expected 1 argument: <pid> or <name>"
```
mbrla0: Are you sure an assertion here is the way to go? Doesn't look like we check for the number of arguments anywhere before this, and having `AssertError` be raised here will bring Pwndbg down.

Why not do it the same way user errors are handled in the other commands?
```suggestion
    if len(args) != 1:
        print(message.error("Expected 1 argument: <pid> or <name>"))
        return
```

pwndbg/dbg/lldb/repl/__init__.py
```diff
@@ -679,60 +687,96 @@ def process_launch(driver: ProcessDriver, relay: EventRelay, args: List[str], db
...
-        return
-
     targets = dbg.debugger.GetNumTargets()
     assert targets < 2
     if targets == 0:
         print(message.error("error: no target, create one using the 'target create' command"))
         return
 
+    # TODO/FIXME: This should ask:
+    # 'There is a running process, detach from it and attach?: [Y/n]'
```
mbrla0, line 703: Suggestion: instead implement an optional general kill/detach mechanism for all commands that can spawn or attach to a process new, since we do this same check on all of them.

PR #2709, opened 2025-01-26
Fix the behavior of the `detach` command in LLDB

mbrla0:
Currently the LLDB Pwndbg CLI foregoes handling process events after commands when LLDB reports command result statuses that imply the process state has not changed. However, a few commands will cause LLDB to report status codes that we understand to indicate no change in process status, despite the fact that they do, in fact, change the state of the process. This means that after these commands complete we incorrectly assume that the process state hasn't changed, even thought it might have.

One concrete example of this is the `detach` command, which causes the end of the lifetime of the process from the point of view of LLDB but that reports `eReturnStatusSuccessFinishResult` as its return status code. Because Pwndbg assumes commands with that code can't change process state, it ends up missing the event that would cause it to realize the process got detached, and it thinks the process is alive for much longer than it actually is.

This PR addresses this issue by forcing the process driver to always try to handle process state change events, regardless of whether we think one is likely to result from the command or not. It makes it so that when we're sure a state change will happen as part of a command, we use the normal event handling flow, and when we're not sure, we force it to handle as many events as are immediately available before we move on.

### DIFF
--- a/pwndbg/dbg/lldb/repl/proc.py
+++ b/pwndbg/dbg/lldb/repl/proc.py
@@ -233,16 +233,20 @@ class ProcessDriver:
                 sys.stdout.buffer.write(out.encode(sys.stdout.encoding, errors="backslashreplace"))
                 print()
 
+            if self.debug:
+                print(f"[-] ProcessDriver: LLDB Command Status: {ret.GetStatus():#x}")
+
             # Only call _run_until_next_stop() if the command started the process.
+            start_expected = True
             s = ret.GetStatus()
             if s == lldb.eReturnStatusFailed:
-                return
+                start_expected = False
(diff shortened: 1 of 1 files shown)


pwndbg/dbg/lldb/repl/proc.py
```diff
@@ -253,7 +257,16 @@ def run_lldb_command(self, command: str) -> None:
             #
             # TODO/FIXME: Find a way to trigger the continued event before the process is resumed in LLDB
```
disconnect3d, line 258: Is this todo/fixme still needed/valid?

mbrla0: Yes

PR #2712, opened 2025-01-28
Pwndbg architecture name list

OBarronCS:
This PR creates a single "point-of-truth" list for the names of architectures that Pwndbg supports.

See comments for rationale on specific changes.

### DIFF
--- a/pwndbg/aglib/arch.py
+++ b/pwndbg/aglib/arch.py
@@ -5,40 +5,8 @@ from typing import Literal
 import pwnlib
 
 import pwndbg
+from pwndbg.lib.arch import PWNLIB_ARCH_MAPPINGS
 from pwndbg.lib.arch import Arch
-
-# List of architectures - used when determining GDB arch, and in OnlyWithArch
-ARCHS = (
-    "x86-64",
-    "i386",
-    "aarch64",
-    "mips",
-    "powerpc",
-    "sparc",
-    "arm",
-    "armcm",
-    "riscv:rv32",
-    "riscv:rv64",
-    "riscv",
-    "loongarch64",
-)
-
-
-# mapping between pwndbg and pwntools arch names
-pwnlib_archs_mapping = {
-    "x86-64": "amd64",
-    "i386": "i386",
-    "aarch64": "aarch64",
-    "mips": "mips",
-    "powerpc": "powerpc",
-    "sparc": "sparc",
-    "arm": "arm",
-    "iwmmxt": "arm",
-    "armcm": "thumb",
-    "rv32": "riscv32",
-    "rv64": "riscv64",
-    "loongarch64": "none",
-}
 
 
 def read_thumb_bit() -> int | None:
@@ -71,7 +39,7 @@ arch: Arch = Arch("i386", 4, "little")
 def update() -> None:
     a = pwndbg.dbg.selected_inferior().arch()
 
-    pwnlib.context.context.arch = pwnlib_archs_mapping[a.name]
+    pwnlib.context.context.arch = PWNLIB_ARCH_MAPPINGS[a.name]
     pwnlib.context.context.bits = a.ptrsize * 8
 
     arch.update(a.name, a.ptrsize, a.endian)
--- a/pwndbg/aglib/arch.py
+++ b/pwndbg/aglib/arch.py
@@ -39,7 +39,7 @@ arch: Arch = Arch("i386", 4, "little")
 def update() -> None:
     a = pwndbg.dbg.selected_inferior().arch()
 
-    pwnlib.context.context.arch = PWNLIB_ARCH_MAPPINGS[a.name]
+    pwnlib.context.context.arch = PWNLIB_ARCH_MAPPINGS.get(a.name, "none")
     pwnlib.context.context.bits = a.ptrsize * 8
 
     arch.update(a.name, a.ptrsize, a.endian)
--- a/pwndbg/dbg/gdb/__init__.py
+++ b/pwndbg/dbg/gdb/__init__.py
@@ -29,6 +29,7 @@ from pwndbg.aglib import load_aglib
 from pwndbg.dbg import selection
 from pwndbg.gdblib import gdb_version
 from pwndbg.gdblib import load_gdblib
+from pwndbg.lib.arch import PWNDBG_SUPPORTED_ARCHITECTURES_TYPE
 from pwndbg.lib.memory import PAGE_MASK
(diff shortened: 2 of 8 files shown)


pwndbg/aglib/arch.py
```diff
@@ -5,41 +5,9 @@
...
-    "mips",
-    "powerpc",
-    "sparc",
-    "arm",
-    "armcm",
-    "riscv:rv32",
-    "riscv:rv64",
-    "riscv",
-    "loongarch64",
-)
```
OBarronCS, line 24: These names are specific to GDB (see `riscv:32`, `riscv:64`), so it was moved to the GDB specific file.

pwndbg/dbg/gdb/__init__.py
```diff
@@ -35,6 +35,25 @@
 T = TypeVar("T")
 
 
+# List of supported architectures that GDB recognizes
+# These strings to converted to the Pwndbg-specific name for the architecture
+gdb_architecture_name_fixup_list = (
+    "x86-64",
+    "i386",
+    "i8086",
```
OBarronCS, line 44: This `i8086` mapping is new. On the early stages of an x86 bootloader, the first ~20 instructions are in i8086 mode before the CPU converts to a 32-bit mode. Now Pwndbg doesn't crash in these cases.

pwndbg/lib/arch.py
```diff
@@ -2,8 +2,49 @@
...
+    "rv32",
+    "rv64",
+    "sparc",
+    "powerpc",
+    "loongarch64",
+]
+
+PWNDBG_SUPPORTED_ARCHITECTURES: list[PWNDBG_SUPPORTED_ARCHITECTURES_TYPE] = list(
+    typing.get_args(PWNDBG_SUPPORTED_ARCHITECTURES_TYPE)
+)
```
OBarronCS: The `Literal` above will provide "type-safety" (red underline in IDE) / auto-complete on the architecture name. Then the elements are extracted into a list that can be used during runtime.

pwndbg/lib/arch.py
```diff
@@ -2,8 +2,49 @@
...
+
+# mapping between pwndbg and pwntools arch names
+PWNLIB_ARCH_MAPPINGS = {
+    "x86-64": "amd64",
+    "i386": "i386",
+    "i8086": "none",
```
OBarronCS, line 33: New mapping to prevent crash in i8086 mode.

pwndbg/lib/arch.py
disconnect3d: Omg I need to revisit this as I don't understand it :D

pwndbg/lib/arch.py
patryk4815: do we need this mapping at all? maybe lets use it as:
`PWNLIB_ARCH_MAPPINGS.get(archname, "none")` ?

OBarronCS: This is a good call - I made this change.

pwndbg/dbg/gdb/__init__.py
```diff
@@ -35,6 +35,25 @@
...
+    "sparc",
+    "arm",
+    "armcm",
+    "riscv:rv32",
+    "riscv:rv64",
+    "riscv",
+    "loongarch64",
+)
+
+
```
patryk4815:
```suggestion
gdb_architecture_name_fixup_list = (
    "x86-64",
    "i386",
    "i8086",
    "aarch64",
    "mips",
    "powerpc",
    "sparc",
    "arm",
    "riscv:rv32",
    "riscv:rv64",
    "riscv",
    "loongarch64",
)
```
removed `armcm`, gdb don't have this arch type

pwndbg/lib/arch.py
```diff
@@ -2,8 +2,49 @@
...
+    "mips": "mips",
+    "aarch64": "aarch64",
+    "arm": "arm",
+    "iwmmxt": "arm",
+    "armcm": "thumb",
```
patryk4815:
```suggestion
    "arm": "arm",
    "armcm": "thumb",
```
removed `iwmmxt` (probably it is already handled by `arm` name)

pwndbg/lib/arch.py
```diff
@@ -2,8 +2,49 @@
...
+PWNDBG_SUPPORTED_ARCHITECTURES_TYPE = Literal[
+    "x86-64",
+    "i386",
+    "i8086",
+    "mips",
+    "aarch64",
+    "arm",
+    "armcm",
+    "iwmmxt",
+    "rv32",
```
patryk4815:
```suggestion
    "armcm",
    "rv32",
```
removed `iwmmxt` (probably it is already handled by `arm` name)

pwndbg/lib/arch.py
```diff
@@ -2,8 +2,49 @@
...
+    "rv64",
+    "sparc",
+    "powerpc",
+    "loongarch64",
+]
+
+PWNDBG_SUPPORTED_ARCHITECTURES: list[PWNDBG_SUPPORTED_ARCHITECTURES_TYPE] = list(
```
patryk4815, line 25: Can we add also this type checking here?
https://github.com/pwndbg/pwndbg/blob/e99d4e7f0b9419c3929bceabe0a624faac4866ab/pwndbg/lib/regs.py#L673
And remove `iwmmxt` register sets

patryk4815:
Only ARM architecture in GDB is handled differently then other architectures (others have proper prefix eg: "i386:x86-64", "riscv:rv64").  
So `xscale`, `iwmmxt`, and `iwmmxt2` must be manually mapped to `arm`.
image

Or we have to replace `show architecture` with this func:
```
In [3]: gdb.selected_inferior().architecture()
Out[3]: <gdb.Architecture arch_name=arm printable_name=armv8-m.main>
# Or if proc is alive:
In [4]: gdb.newest_frame().architecture()
```
(sic!) only repr show `arch_name=arm`, so we have to parse repr manually

patryk4815 (approved):

disconnect3d:
Ping @OBarronCS  see comments

pwndbg/lib/arch.py
OBarronCS: Done!

OBarronCS:
I removed the definitions of `iwmmxt` (which were just duplicates of `arm`) as noted in the above comments, and I added an `if` check inside the GDB arch detection function to map `iwmmxt`,`iwmmxt2`, and `xscale` to `arm`. I also added the new support arch type literal to a couple places for more mypy type safety.

PR #2715, opened 2025-01-29
fix f string not working on older versions of python

purupuddu:
f strings containing backslashes are not supported in python 3.10.12, this is an easy fix

### DIFF
--- a/pwndbg/commands/version.py
+++ b/pwndbg/commands/version.py
@@ -51,9 +51,10 @@ def debugger_version():
 
 
 def all_versions():
+    py_version = sys.version.replace("\n", " ")
     return (
         f"Pwndbg:   {pwndbg.__version__}",
-        f"Python:   {sys.version.replace('\n', ' ')}",
+        f"Python:   {py_version}",
         debugger_version(),
         f"Capstone: {module_version('capstone')}",
         f"Unicorn:  {module_version('unicorn')}",


patryk4815:
Thanks!
🤔 `vermin` should catch that

patryk4815:
https://github.com/netromdk/vermin/issues/264 yes it reported here

podd0:
It looks like a different issue. That's an fstring containing more single quotes, this one contains a \n. Anyways just trying `f"Python:   {sys.version.replace('\n', ' ')}"` on a python3.10 shell gives the error
image

patryk4815:
@podd0 It is same issue (they don't handle new f-string changes):

> Vermin doesn't catch changes introduced in Python 3.12 to f-strings. Specifically, f-string nesting. See https://docs.python.org/3/whatsnew/3.12.html#pep-701-syntactic-formalization-of-f-strings

image

purupuddu:
> @podd0 It is same issue (they don't handle new f-string changes):
> 
> > Vermin doesn't catch changes introduced in Python 3.12 to f-strings. Specifically, f-string nesting. See https://docs.python.org/3/whatsnew/3.12.html#pep-701-syntactic-formalization-of-f-strings
> 
> image
> 
Yeah, i understood after, thanks

PR #2723, opened 2025-02-04
Redirect user to vis_heap_chunks help when chunks are ommited

k4lizen:
I left it as `vis` in the message because it's very nice to know about the shorthand.

image


disconnect3d:
Can you fix the related test - `test_vis_heap_chunk_command` - that break now? :)

Diff here:
image

disconnect3d:
Ping @k4lizen

k4lizen:
oh yeah sure, i somehow missed the notif

PR #2730, opened 2025-02-13
Properly decode binary data sent via GDB commands

OBarronCS:
This PR makes a fix to the decoding of binary data that GDB stub's return via file read commands, such as `vFile:pread`.

GDB defines the data encoding here: 

https://sourceware.org/gdb/current/onlinedocs/gdb.html/Overview.html#:~:text=The%20binary%20data%20representation%20uses%207d

Some bytes -  `#`,`$`,`*`,`}` - have are escaped with an "}". 

For reference, here is the Qemu GDB stub function that encodes this:

https://github.com/qemu/qemu/blob/de278e54aefed143526174335f8286f7437d20be/gdbstub/gdbstub.c#L184-L200

pread handler defined here:
https://github.com/qemu/qemu/blob/de278e54aefed143526174335f8286f7437d20be/gdbstub/user-target.c#L340

This meant that previously, file fetches with `pwndbg.aglib.file.get_file` would return corrupted files due to missing this un-escaping step.

### DIFF
--- a/pwndbg/aglib/file.py
+++ b/pwndbg/aglib/file.py
@@ -185,7 +185,7 @@ def vfile_readlink(pathname: str | bytes) -> bytes:
 
     parts = response[1:].split(b";", 1)
     # bytes_read = int(parts[0], 16)
-    target = parts[1]
+    target = gdb_memtox_inverse(parts[1])
     return target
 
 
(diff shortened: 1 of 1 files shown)


patryk4815:
https://lldb.llvm.org/resources/lldbgdbremote.html#vfile-pread
hmm, I need to check source code, no info about extra characters

patryk4815:
https://github.com/llvm/llvm-project/blob/eef02053459a0eb3076b0db76555f9836ac90b07/lldb/source/Plugins/Process/gdb-remote/GDBRemoteCommunication.cpp#L756-L767
lldb is doing same stuff, nice catch 👍

PR #2731, opened 2025-02-13
bump unicorn to 2.1.2

### LINKED ISSUE #2588: Update Unicorn Engine to > 2.1.1
When UE is updated to > 2.1.1 we should revert the if/else for arm64 bug that we applied a workaround for in here: https://github.com/pwndbg/pwndbg/issues/2548

patryk4815:
https://github.com/pwndbg/pwndbg/issues/2588#issuecomment-2657029032


wtdcode:
Maybe bump 2.1.3 now =)

Issue #2734, opened 2025-02-13
Expand bitflag arguments with their constant names


OBarronCS:
Many GLIBC functions take arguments as a bitwise OR of multiple "flags".

For example, take the `mmap` function:
![Image](https://github.com/user-attachments/assets/c5a7341d-89e1-430d-8338-3f6581c336ad)

In this example, the `flags` argument is the bitwise OR of a handful of constants - `MAP_PRIVATE | MAP_FIXED | MAP_ANONYMOUS | MAP_POPULATE` - and same goes for `prot`. It would be really great if we could extract those bit flags and display their constant names alongside the number. The same goes for many GLIBC functions, such as open, seccomp, mprotect, etc.

We could hardcode the flag constants (maybe in our `pwndbg/lib/functions_data.py` file?) with the assumption that they are stable in the Linux kernel + GLIBC syscall interface functions, and display them in the context when appropriate.

disconnect3d:
Would also be nice to find some better place for function information 😅 
(func names, args, types, flags etc)

disconnect3d:
Initial support was implemented in #2740 with mmap bitflags and we have https://github.com/pwndbg/pwndbg/pull/2755 opened to fetch the bitflag values from Pwntools constants.

What flag constants are we missing? Where can we take them from?

disconnect3d:
Moved this to https://github.com/pwndbg/pwndbg/issues/2761 and https://github.com/pwndbg/pwndbg/issues/2762

PR #2740, opened 2025-02-15
Allow dumpargs to format bit flags

LorenVS:
If flags are supplied for a function in `functions_data.py`, the `dumpargs` command can pretty-print the combination of flags used in the call.

This is some preliminary work on #2734. I wanted to get some initial feedback before looking at how to populate the correct flag values for all the different calls that are currently present in `functions_data.py`.


tests/gdb-tests/tests/test_commands_dumpargs.py
```diff
@@ -0,0 +1,40 @@
...
+    gdb.execute("step")
+    # Stop on mmap(...)
+    gdb.execute("nextcall")
+
+    out = gdb.execute("dumpargs", to_string=True).splitlines()
+    assert len(out) == 6
+    assert out[0] == '        addr:      0xcafe0000'
+    assert out[1] == '        len:       0x1000'
+    assert out[2] == '        prot:      1'
+    assert out[3] == '        flags:     MAP_PRIVATE | MAP_ANONYMOUS | MAP_FIXED'
```
disconnect3d: Maybe: `flags:     0x32 (MAP_PRIVATE|MAP_ANONYMOUS|MAP_FIXED)`?

LorenVS: Agreed, it makes complete sense to retain the original integer value as well. I updated the format to this.

pwndbg/commands/dumpargs.py
```diff
@@ -38,7 +38,7 @@ def call_args() -> List[str]:
 
     for arg, value in pwndbg.arguments.get(pwndbg.aglib.disasm.one()):
         code = arg.type != "char"
-        pretty = pwndbg.chain.format(value, code=code)
+        pretty = pwndbg.chain.format(value, code=code, flags=arg.flags)
```
disconnect3d: Hmm... does it make sense to call `chain.format` at all if flags are there? If I understand correctly, when we pass a list of values, we will never pass flags; when we pass flags, we will never want to do a "chain dereference".

Maybe this should be sth like:
```py
pretty = pwndbg.chain.format(value, code=code) if not args.flags else format_flags(value, flags)
```

(The `format_flags` should probably be put in some module instead of here)

pwndbg/chain.py
```diff
@@ -181,3 +186,37 @@ def format(
...
+    if value != 0:
+        flag_names.append(hex(value))
+
+    # The final format includes the original value as hex,
+    # any matched flags, and the left-over unmatched portion
+    # of the integer.
+    #
+    # For example:
+    # 0x03 (FLAG_2|0x01)
+    return hex(original_value) + " (" + '|'.join(flag_names) + ")"
```
disconnect3d: This should be:
```suggestion
    return f"{original_value:#x} ({'|'.join(flag_names)})"
```

pwndbg/chain.py
```diff
@@ -181,3 +186,37 @@ def format(
         return enhanced
 
     return arrow_right.join(rest) + arrow_left + enhanced
+
+def _format_flags_value(flags: List[Flag], value: int):
```
disconnect3d: Ideally, we should add some tests for this function

pwndbg/lib/functions_data.py
```diff
@@ -28479,7 +28480,13 @@ def load_functions():
             Argument(type="void", derefcnt=1, name="addr"),
             Argument(type="size_t", derefcnt=0, name="len"),
             Argument(type="int", derefcnt=0, name="prot"),
-            Argument(type="int", derefcnt=0, name="flags"),
+            Argument(type="int", derefcnt=0, name="flags", flags=[
```
disconnect3d, line 28483: Should always be a tuple instead of a list

disconnect3d:
Pls fix linter issues: https://github.com/pwndbg/pwndbg/actions/runs/13348834380/job/37282697043?pr=2740 and see comments :)

I like this feature a lot!

pwndbg/commands/dumpargs.py
LorenVS: I had a hard time determining whether this logic belonged in `chain.format` or elsewhere. I switched to your suggestion. I didn't see a great place to put `format_flags_argument`, so I moved it into `functions.py`.

pwndbg/chain.py
LorenVS: Done

pwndbg/lib/functions_data.py
LorenVS: I switched the type everywhere to `tuple[Flag, ...]`. I think that is what you're asking for?

LorenVS:
Sorry for the lint failures, I had a difficult time getting them up and running locally at first. I think I have it figured out now.

disconnect3d (approved):
Awesome, thanks!

OBarronCS:
This is awesome!

Do we happen to have knowledge as to how the `functions_data.py` file was initially populated? It looks like an auto-generated file. Some of the argument names for the functions don't match exactly with what manpages show, for example.

OBarronCS:
Just thought of something to consider in the future as well:

Looks like these flag constants actually differ based on the target os/arch pair:
(searching pwntools codebase for `#define MAP_ANONYMOUS` for example - it pulls these constants from linux sourcecode)
image

We could potentially dynamically pull them - look into the pwntools `constgrep` tool for some details.

patryk4815:
> We could potentially dynamically pull them - look into the pwntools `constgrep` tool for some details.

eg using:
```py
# ./result/bin/pwndbg
pwndbg> ipi
In [2]: from pwnlib import constants
In [3]: constants.MAP_ANONYMOUS
Out[3]: Constant('MAP_ANONYMOUS', 0x20)

# ./result/bin/pwndbg --init-eval-command='set architecture mips'
The target architecture is set to "mips".
pwndbg> ipi

In [1]: from pwnlib import constants
In [2]: constants.MAP_ANONYMOUS
Out[2]: Constant('MAP_ANONYMOUS', 0x800)
```

LorenVS:
It looks like two commits from this PR (3fa6409c4a6590e8fc25bfd744aa2ec009fd9565 and 0a235d0f0015266a5188674c2be9cb8926900661) were not merged. The merged version contains the pre-review-comments implementation.

LorenVS:
Thanks for making me aware of `pwnlib`, that makes the follow-up work here trivial. I sent out PR #2755 to address the os/arch issues.

PR #2743, opened 2025-02-15
Fix #2674: context out when splitting window

### LINKED ISSUE #2674: banner not printed on new line when contextoutput is used
### Description

image

### Steps to reproduce

Use something like this: https://blog.jcfg.re/posts/pwndbg-tmux/ (https://github.com/joaogodinho/pwnmux/blob/main/gdbinit.py). In this case change `contextoutput(section, p[1], True, 'top', False)` to `contextoutput(section, p[1], False, 'top', False)` so context isn't cleared.

### My setup
```
Gdb:      15.2
Python:   3.13.1
Pwndbg:   2024.08.29 build: 2546eef4
Capstone: 5.0.1280
Unicorn:  2.1.0
```

disconnect3d:
This commit fixes #2674. A bug in `commands.context` implementation where we printed out the banner without a newline when context was forwarded to separate terminals.

This commit also refactors the `context` function a bit to simplify it.

As a result, when the context is supposed to not print anything the `history_handle_unchanged_contents` will not be called.

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -674,9 +674,7 @@ def context(subcontext=None, enabled=None) -> None:
         with target as out:
             if result_settings[target].get("clearing", config_clear_screen) and lines:
                 clear_screen(out)
-            out.write("\n".join(lines))
-            if out is sys.stdout:
-                out.write("\n")
+            out.writelines(line + "\n" for line in lines)
             out.flush()
 
 


disconnect3d:
@peace-maker can you confirm it is fine that `history_handle_unchanged_contents` is not called when context sections are empty?

disconnect3d:
@k4lizen can you double check if this fixes your issue? In my tests it does so, but I'd appreciate another look.

k4lizen:
It does fix it for me, great thanks!

peace-maker:
It is fine.

k4lizen:
Oof actually this causes a new issue:
Each panel/section has an extraneous new line at the end that simply takes up space.

image
image

mb I didn't notice this before

disconnect3d:
@k4lizen then i guess we should do print this conditionally, but maybe for certain sections? or: maybe the backtrace section does not print a newline while others do? huh

k4lizen:
@disconnect3d All sections print the trailing newline, the issue just isn't apparent on backtrace because there is a banner (`--[EXPRESSIONS]--`) on that new line.

k4lizen:
Changing the line to
```python
            out.writelines("\n" + line for line in lines)
```
seems to fix the issue for the multi-window setup, but messes it up for the normal setup.

disconnect3d:
@k4lizen shall we maybe do:
```py
# <some cool comment explaining wth is going on here>
if out is sys.stdout:
    lines = ("\n" + line for line in lines)
else:
    lines = (line + "\n" for line in lines)

out.writelines(lines)
```

PR #2745, opened 2025-02-16
Cross architecture sanity check tests 

OBarronCS:
This draft PR is a start to adding more tests in anticipation of Capstone V6, which comes with a huge list of fixes and improvements, new architectures, and also large breaking API changes that affect Pwndbg's use of it. Particularly, there is going be many, sometimes subtle, changes in the way instruction metadata is exposed across different architectures + instructions. 

This adds a sanity check test to different architectures to detect crashes which would arise from changes in the API. A program is compiled into `aarch64`, `arm`, `mips`, `mips64`, and `riscv64` and the test will step through each instruction in the program and run "context" to invoke the annotation/emulation code. If the program crashes, it means the annotation code made an incorrect assumption on retrieving data from an instruction.

~~Running the tests locally took about 250 seconds (I'm not sure if this is the sort of test we need to be running every time).~~ They now run in ~60 seconds.

A couple things I came across when making this:
1. Ubuntu 24.04 LTS `gdb-multiarch` package doesn't currently include loongarch support compiled in. This has been added to new ubuntu versions (24.10 and 25.04). See <https://launchpad.net/ubuntu/+source/gdb> and the two different versions under "Noble Numbat". The "proposed" version for 24.04 (backported from new ubuntu releases) mentions adding it in the changelog.

This means that using a 24.04 container won't work for `loongarch` tests (which is getting added to both capstone + unicorn in upcoming releases).

2. The Ubuntu `gcc-riscv64-linux-gnu` package is not compiled with `--enable-multilibs`, which would allow using the compiler to create RISCV32 programs. We would have to compile the toolchain ourselves to enable this for the testing CI/CD.

### DIFF
--- a/setup-dev.sh
+++ b/setup-dev.sh
@@ -150,6 +150,20 @@ install_apt() {
         sudo apt install shfmt
     fi
 
+    # Setup symlinks so QEMU can find cross-arch libraries
+    # This is where QEMU expects to find libraries. Run `qemu-ARCH -h` to find this.
+    # The gcc-* packages installed earlier included these library files.
+    sudo mkdir -p /usr/gnemul/
+    sudo ln -s /usr/aarch64-linux-gnu /usr/gnemul/qemu-aarch64
+    sudo ln -s /usr/arm-linux-gnueabihf /usr/gnemul/qemu-arm
+    sudo ln -s /usr/mips-linux-gnu /usr/gnemul/qemu-mips
+    sudo ln -s /usr/mips64-linux-gnuabi64/ /usr/gnemul/qemu-mips64
+    sudo ln -s /usr/riscv64-linux-gnu/ /usr/gnemul/qemu-riscv64
+    # ln -s /usr/powerpc-linux-gnu/ /usr/gnemul/qemu-ppc
+    # ln -s /usr/powerpc64-linux-gnu/ /usr/gnemul/qemu-ppc64
+    # ln -s /usr/sparc64-linux-gnu/ /usr/gnemul/qemu-sparc64
+    # ln -s /usr/loongarch64-linux-gnu/ /usr/gnemul/qemu-loongarch64
+
     command -v go &> /dev/null || sudo apt-get install -y golang
 
     download_zig_binary
--- a/setup-dev.sh
+++ b/setup-dev.sh
@@ -150,20 +150,6 @@ install_apt() {
         sudo apt install shfmt
     fi
 
-    # # Setup symlinks so QEMU can find cross-arch libraries
-    # # This is where QEMU expects to find libraries. Run `qemu-ARCH -h` to find this.
-    # # The gcc-* packages installed earlier included these library files.
-    # sudo mkdir -p /usr/gnemul/
-    # sudo ln -s /usr/aarch64-linux-gnu /usr/gnemul/qemu-aarch64
-    # sudo ln -s /usr/arm-linux-gnueabihf /usr/gnemul/qemu-arm
-    # sudo ln -s /usr/mips-linux-gnu /usr/gnemul/qemu-mips
-    # sudo ln -s /usr/mips64-linux-gnuabi64/ /usr/gnemul/qemu-mips64
-    # sudo ln -s /usr/riscv64-linux-gnu/ /usr/gnemul/qemu-riscv64
-    # # ln -s /usr/powerpc-linux-gnu/ /usr/gnemul/qemu-ppc
-    # # ln -s /usr/powerpc64-linux-gnu/ /usr/gnemul/qemu-ppc64
-    # # ln -s /usr/sparc64-linux-gnu/ /usr/gnemul/qemu-sparc64
-    # # ln -s /usr/loongarch64-linux-gnu/ /usr/gnemul/qemu-loongarch64
-
     command -v go &> /dev/null || sudo apt-get install -y golang
 
     download_zig_binary
--- a/setup-dev.sh
+++ b/setup-dev.sh
@@ -150,19 +150,19 @@ install_apt() {
         sudo apt install shfmt
     fi
 
-    # Setup symlinks so QEMU can find cross-arch libraries
-    # This is where QEMU expects to find libraries. Run `qemu-ARCH -h` to find this.
-    # The gcc-* packages installed earlier included these library files.
-    sudo mkdir -p /usr/gnemul/
-    sudo ln -s /usr/aarch64-linux-gnu /usr/gnemul/qemu-aarch64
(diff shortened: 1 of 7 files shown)


tests/qemu-tests/conftest.py
```diff
@@ -71,6 +71,12 @@ def _start_binary(asm: str, arch: str, *args):
     qemu.kill()
 
 
+QEMU_CORRECTION_MAP = {
+    "mips": ("mips", "/etc/qemu-binfmt/mips/"),
```
patryk4815: 🤔 when it is needed? `/etc/qemu-binfmt`

patryk4815: as I know ubuntu don't create `/etc/qemu-binfmt/`, so it must be manually created?

disconnect3d: We should probably add a comment describing what those paths are/how to get them etc

patryk4815: It think it depend if qemu-user is compiled with `--interp-prefix=/etc/qemu-binfmt/%M`
But I don't know if ubuntu is applying that option.

OBarronCS: Tests are currently failing because these library files don't exist - I'll add some symlinks to make them work.

On second thought, this QEMU_CORRECTION_MAP is unnecessary and limiting, and I'll change this so the test has to explicitly set the endianness so there is no ambiguity. 

I initially had it because pwntool's internally defaults to little-endian, while GCC defaults to big endian for MIPS, and this mismatch caused the wrong version of qemu to be used (`qemu-mipsel` vs `qemu-mips`. 


Also - this something that took me a while to find out - doing `qemu-archname -h` will print out the location where the library files are found (`QEMU_LD_PREFIX = ????` in the output). Whenever I have run it, it seems to be in `/usr/gnemul/`, instead of `/etc/qemu-binfmt` like many guides would indicate.

patryk4815: `QEMU_LD_PREFIX` is same like `-L` option maybe?
 `lldb` is using internally `QEMU_LD_PREFIX` env - https://github.com/llvm/llvm-project/blob/a7a02083acf39b9f1ad7edec3b7e344afc6cac49/lldb/source/Plugins/Platform/QemuUser/PlatformQemuUser.cpp#L204

patryk4815: `/usr/gnemul/qemu-%M` is default path for `--interp-prefix`  - https://github.com/qemu/qemu/blob/495de0fd82d8bb2d7035f82d9869cfeb48de2f9e/meson_options.txt#L28

patryk4815: > instead of /etc/qemu-binfmt like many guides would indicate.

`/etc/qemu-binfmt` <- I think this was valid few years ago, only for ubuntu, but was removed at some point

OBarronCS: I agree - `QEMU_LD_PREFIX` is the same is `-L`. I added a commit that creates symlinks to the /etc/gnemul/ files that qemu will search for.

patryk4815: Until `debian:12` qemu use `/etc/qemu-binfmt/%M` (Since debian:13 qemu will use /usr/gnemul/qemu-%M), every other system what I tested they use `/usr/gnemul/qemu-%M`
```
root@3baa425ae87a:/# qemu-aarch64-static --help | grep 'QEMU_LD_PREFIX'
-L path              QEMU_LD_PREFIX    set the elf interpreter prefix to 'path'
QEMU_LD_PREFIX  = /etc/qemu-binfmt/aarch64
```

setup-dev.sh
```diff
@@ -145,6 +145,20 @@ install_apt() {
...
+    sudo ln -s /usr/aarch64-linux-gnu /usr/gnemul/qemu-aarch64
+    sudo ln -s /usr/arm-linux-gnueabihf /usr/gnemul/qemu-arm
+    sudo ln -s /usr/mips-linux-gnu /usr/gnemul/qemu-mips
+    sudo ln -s /usr/mips64-linux-gnuabi64/ /usr/gnemul/qemu-mips64
+    sudo ln -s /usr/riscv64-linux-gnu/ /usr/gnemul/qemu-riscv64
+    # ln -s /usr/powerpc-linux-gnu/ /usr/gnemul/qemu-ppc
+    # ln -s /usr/powerpc64-linux-gnu/ /usr/gnemul/qemu-ppc64
+    # ln -s /usr/sparc64-linux-gnu/ /usr/gnemul/qemu-sparc64
+    # ln -s /usr/loongarch64-linux-gnu/ /usr/gnemul/qemu-loongarch64
+
```
patryk4815: I'm not in favor of adding more "root"-specific code 🤔
Can't we just compile the binaries with the correct RPATH/RUNPATH instead? - then we can remove `"-L"` mapping in python

OBarronCS: I agree it's not ideal to add more uses of `sudo` here.

Another possibility is to create a new "cross-arch tests" Docker container, given the possibility that we'll need to manually compile a risc-v cross-compiler for 32-bit uses and to get the correct version of gdb for loongarch64 tests.

OBarronCS: Latest commits compile many of the binaries with -static. However, two of the pre-existing tests appeared to depend on libc being dynamically linked, so I added a map of these dynamic library locations in the Python code so we can look it up while avoiding the symlinks.

OBarronCS:
~~I added a check for an environment variable which is needed to run these tests (they take 5 minutes for me locally, and ran in 10+ minutes on CI/CD). I think they are not needed to run and take resources all the time, but when deemed necessary we can run them locally to have more confidence that a change to the disasm code does not introduce a crash.~~. EDIT: not relevant anymore, the tests run in ~60 seconds.

tests/qemu-tests/conftest.py
```diff
@@ -72,6 +73,20 @@ def _start_binary(asm: str, arch: str, *args):
...
+    "aarch64": "/usr/aarch64-linux-gnu",
+    "arm": "/usr/arm-linux-gnueabihf",
+    "mips": "/usr/mips-linux-gnu",
+    "mips64": "/usr/mips64-linux-gnuabi64/",
+    "riscv64": "/usr/riscv64-linux-gnu/",
+    "loongarch64": "/usr/loongarch64-linux-gnu/",
+    "ppc": "/usr/powerpc-linux-gnu/",
+    "ppc64": "/usr/powerpc64-linux-gnu/",
+    "sparc64": "/usr/sparc64-linux-gnu/",
+}
```
OBarronCS, line 87: These are file paths specific to the Ubuntu installs of cross-arch libc's. The tests run in Ubuntu, so this works for CI/CD testing purposes.

PR #2755, opened 2025-02-19
Switch to using `pwnlib` constants for flags.

LorenVS:
PR #2740 added initial support for formatting flag arguments to `dumpargs`. However, the flag values were hard-coded and might not have been correct on certain platforms and architectures. @patryk4815 and @OBarronCS made me aware of `pwnlib`, which already exposes these constants through a Python library, and which is already used elsewhere in this project.

Switching the flags to use `pwnlib` constants directly avoids some repetition, and also makes this code more portable.

### DIFF
--- a/pwndbg/lib/functions.py
+++ b/pwndbg/lib/functions.py
@@ -4,6 +4,9 @@ from collections.abc import Mapping
 from typing import List
 from typing import NamedTuple
 
+from pwnlib import constants
+from pwnlib.constants.constant import Constant
+
 
 class Function(NamedTuple):
     type: str
@@ -16,12 +19,7 @@ class Argument(NamedTuple):
     type: str
     derefcnt: int
     name: str
-    flags: tuple[Flag, ...] | None = None
-
-
-class Flag(NamedTuple):
-    value: int
-    name: str
+    flags: tuple[Constant, ...] | None = None
 
 
 class LazyFunctions(Mapping[str, Function]):
@@ -46,7 +44,16 @@ class LazyFunctions(Mapping[str, Function]):
 functions = LazyFunctions()
 
 
-def format_flags_argument(flags: tuple[Flag, ...], value: int):
+def resolve_constants(*names: str) -> tuple[Constant, ...]:
+    resolved: List[Constant] = []
+    for name in names:
+        constant = getattr(constants, name, None)
+        if constant is not None:
+            resolved.append(constant)
+    return tuple(resolved)
+
+
+def format_flags_argument(flags: tuple[Constant, ...], value: int):
     original_value: int = value
     flag_names: List[str] = []
 
@@ -57,9 +64,9 @@ def format_flags_argument(flags: tuple[Flag, ...], value: int):
     # descending popcount, this loop will output more specific
     # flags.
     for flag in flags:
-        if (value & flag.value) == flag.value:
-            flag_names.append(flag.name)
-            value = value & ~flag.value
+        if (value & int(flag)) == int(flag):
+            flag_names.append(str(flag))
+            value = value & ~int(flag)
 
     # If none of the known flags matched the value, just
     # format the value as a normal hex integer.
--- a/pwndbg/lib/functions_data.py
+++ b/pwndbg/lib/functions_data.py
@@ -1,8 +1,8 @@
 from __future__ import annotations
 
 from pwndbg.lib.functions import Argument
-from pwndbg.lib.functions import Flag
 from pwndbg.lib.functions import Function
+from pwndbg.lib.functions import resolve_constants
 
 
 def load_functions():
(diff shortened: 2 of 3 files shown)


OBarronCS:
This is a great feature!

I drafted some ideas you might want to consider as well:

```python

# Perhaps some sort of function like this that can retrieve the concrete values of flags at runtime?
# This would enable the code to work across different architecture targets
def get_flags_for_function_and_arg(flag_names: list[str]) -> list[Flag]:
    flag_values = []

    for name in flag_names:
        # Dynamically lookup the constant at runtime
        # Under the hood, pwnlib will get the correct constant for the currently set architecture
        value = getattr(constants,name,None)
        # pwntools may not have this constant value defined
        if value is None:
            continue
        flag_values.append(Flag(value,name))

    return flag_values


# Some initial work on getting some flag names - these could be embedded in the functions_data.py?
# These flags are manually retrieved from the Linux Kernel headers
# - https://elixir.bootlin.com/linux/v6.13.3/source/tools/include/uapi/asm-generic
# - https://elixir.bootlin.com/linux/v6.13.3/source/include/uapi/asm-generic
# And reading man pages to find more - some only exist for certain arches.
stuff = {
    "mmap":{
        "flags":[
            "MAP_SHARED",
            "MAP_PRIVATE",
            "MAP_SHARED_VALIDATE",
            "MAP_FIXED",
            "MAP_ANONYMOUS",
            "MAP_POPULATE",
            "MAP_NONBLOCK",
            "MAP_STACK",
            "MAP_HUGETLB",
            "MAP_SYNC",
            "MAP_FIXED_NOREPLACE",
            "MAP_UNINITIALIZED",
            "MAP_GROWSDOWN",
            "MAP_LOCKED",
            "MAP_NORESERVE"
        ],
        # https://elixir.bootlin.com/linux/v6.13.3/source/tools/include/uapi/asm-generic/mman-common.h
        "prot":[
            "PROT_READ",
            "PROT_WRITE",
            "PROT_EXEC",
            "PROT_NONE",
        ]
    },
    "mprotect":{
        "prof":[
            "PROT_READ",
            "PROT_WRITE",
            "PROT_EXEC",
            "PROT_NONE",
            "PROT_SEM",
            "PROT_SAO",
            "PROT_GROWSUP",
            "PROT_GROWSDOWN"
        ]
    }
}
```

It would also be fantastic to have this printed out not only in `dumpargs`, but also in the context display, which gets the arguments to print from this function:
https://github.com/pwndbg/pwndbg/blob/0d7a135085847ffe1ca92a54c6b3bd384d482b0a/pwndbg/arguments.py#L156-L159

Looks like there is some duplication here and in dumpargs - potentially we could refactor and combine the two for consistency (also this function enhances arguments with the name "fd" for file descriptor, and we could extend this to other words like "in_fd", "out_fd", "fildes", "newfd", "oldfd" which are all also references to fd.)


I also encountered an edge case that could be considered - if the constant value is 0x0. For example, `PROT_NONE` for mmap's `prot` arg stands for `0`, and is a legit value as well, although not strictly a bit flag it could be nice to display.

LorenVS:
I moved this to a draft, since the risc-v tests were failing due to a missing constant. I was already adopting an approach similar to what you outlined to resolve this.

disconnect3d:
Are we sure we want to fetch this dynamically from pwnlib if we know the values up front? If the values do not change (e.g. per arch change) we probably don't want to do more work than needed (more flags = more calls = longer import/call time etc)

LorenVS:
Is your concern about the initialization time for the data in `functions_data.py` ? If that is the case, we could just leave the flags as `str`s on the function data and only call `resolve_flags` at `dumpargs`-time.

disconnect3d:
@LorenVS Yeah - instead of calling func for each flag we could just hardcode all of them. Can we somehow check if any of those flags differ for different architectures or something? If not, why not hardcoding them instead of fetching?

I'm fine with fetching on demand (when resolving func args) or hardcoding; just don't want to do more work during import/load time if possible.

OBarronCS:
It appears that different architectures do indeed define these constants differently - here is a search for MAP_ANONYMOUS, for example, in pwntools (which scrapes from Linux sourcecode)

image

pwndbg/lib/functions.py
```diff
@@ -57,9 +64,9 @@ def format_flags_argument(flags: tuple[Flag, ...], value: int):
     # descending popcount, this loop will output more specific
     # flags.
     for flag in flags:
-        if (value & flag.value) == flag.value:
-            flag_names.append(flag.name)
-            value = value & ~flag.value
+        if (value & int(flag)) == int(flag):
+            flag_names.append(str(flag))
+            value = value & ~int(flag)
```
disconnect3d, line 69: shall we maybe move int(flag) to resolve_constants? do we ever need pwnlib's Constant object?

disconnect3d:
@LorenVS Given that different OS may set this differently I believe we want to call `resolve_constants` (or flags or whatever) at runtime instead of in `functions_data.py`. In other words, the functions_data should probably only keep name of the flags and we should fetch their values at runtime.

disconnect3d:
@LorenVS ping, any chance u can update this?

disconnect3d:
@LorenVS ping?

PR #2758, opened 2025-02-23
Reduce context shaking when quickly stepping

k4lizen:
by reserving empty lines after the command prompt

If the whole context + reserved lines will fit in the window, we reserve, otherwise we let the user see as much of the context as possible.

https://github.com/user-attachments/assets/61939381-e428-44b9-8658-93f2c21839bf

Works on both gdb and lldb.


k4lizen:
@mbrla0

k4lizen:
Also if there are some other context sections that fail this assertion
`            assert(all("\n" not in line for line in lines))`
that I missed, lmk.

k4lizen:
I guess I fix the tests by just appending the control characters at the end of expected ouput..?

mbrla0:
> @mbrla0

I like this idea, but I do have one concern about, specifically, the way the code picks the number of lines that should be reserved.

Context displays are automatically triggered in Pwndbg at the first prompt hook after an `EventType.CONTINUE` (GDB) or an `EventType.STOP` (LLDB)[^1]. Meaning that any action that can fire either of these events, or lead to either of these events being fired will set up a context display on the next call to the prompt hook. It follows, then, that the number of lines we have to reserve may actually change between different actions, and `pwndbg.dbg.debugger_gen_lines` does not seem to account for that. 

One way - not necessarily the best way, but still fine - to get around that is, instead of a fixed number of lines, what if it were to measure the "number of lines printed by the debugger between the last CONTINUE and STOP event pair"? In LLDB, we can easily measure that, as all non-Pwndbg output has to go through the Pwndbg CLI.

Ultimately, though, this is a nitpick, as the best-guess approach for `debugger_gen_lines` will only ever be as bad as doing nothing, and when it works, which should happen most often, it helps a lot. Even if stays best-guess, though, I think it would be nice for us to at least document this bit of nuance in the docstring (and maybe add a TODO to actually handle it dynamically in the future).

[^1]: I've listed both event types for completeness' sake, but you don't need to worry about the event types being different between GDB and LLDB. From the POV of the prompt hook, they're pretty much the same.

k4lizen:
@mbrla0 
Okay I see what you're saying, but here are some things to consider:
> It follows, then, that the number of lines we have to reserve may actually change between different actions, and pwndbg.dbg.debugger_gen_lines does not seem to account for that.

This is true, for instance in gdb if you `continue` then Ctrl-C, you will get something like this:
```
pwndbg> c
<some program output>
Program received signal SIGINT, Interrupt.
0x00007ffff7e63cde in pselect64_syscall (nfds=1, readfds=0x7fffffffcec0, writefds=0x0, exceptfds=0x0, timeout=<optimized out>, sigmask=0x7ffff7f97de0 <_rl_orig_sigset>)
    at ../sysdeps/unix/sysv/linux/pselect.c:34
34	 return SYSCALL_CANCEL (pselect6_time64, nfds, readfds, writefds, exceptfds,
<the context>
pwndbg>
```
And most of these lines, including program output and the 3-4 lines gdb outputs, we don't really control and they remain unaccounted for. However, this doesn't matter in the context of terminal shakyness, as the screens visual state already changes significantly when you do this program continue-interrupt, rather than the consistency you expect from stepping.

> One way - not necessarily the best way, but still fine - to get around that is, instead of a fixed number of lines, what if it were to measure the "number of lines printed by the debugger between the last CONTINUE and STOP event pair"? In LLDB, we can easily measure that, as all non-Pwndbg output has to go through the Pwndbg CLI.

Note that the amount of lines we are reserving, is the amount that we predict will be used on the *next* event. In that sense, the amount of lines used by the debugger in the previous event is a poor predictor, an example of this being the above `c`+Ctrl-C event. It is definitely impossible to write a predictor that will always be correct. Any approach will ultimately be guess-based, though it might be possible to add a better predictor than the current one (though I wouldn't know how). The approach you propose would make the predictor most often be wrong twice when the `c`+Ctrl-C thing happens, while the current approach would most often only be wrong once.

Also, reserving too many lines could be poor UX as it would mean lots of empty space on the screen.

Possibly, `debugger_gen_lines` is a good solution for this PR, but not usable for something that wants similar information but has a different goal.

In any case, I can edit the docstring to mention this stuff if you want.

mbrla0:
> Note that the amount of lines we are reserving, is the amount that we predict will be used on the _next_ event

Oh, I get it now. Well, in that case, it makes sense, though, on an unrelated note, maybe there should be an option to control whether we reserve lines or not?

> In any case, I can edit the docstring to mention this stuff if you want.

That'd be great, thanks!

k4lizen:
Should be good now

pwndbg/commands/context.py
```diff
@@ -91,6 +91,66 @@ def clear_screen(out=sys.stdout) -> None:
...
+            # We always write to stdout since that is where the prompt will be
+            # ANSII codes
+            # \x1b[#S  scroll the terminal up by # lines
+            # \x1b[#A  move the cursor up by # lines
+            sys.stdout.write(f"\x1b[{needed_empty}S\x1b[{needed_empty}A")
+    else:
+        assert config_reserve_lines == "always"
+        # Since we are clearing the screen, leftover_lines describes actual empty lines.
+        leftover_lines = max(leftover_lines, 0)
+        if leftover_lines < needed_empty:
```
disconnect3d, line 149: Can this be just 
```suggestion
        if 0 < leftover_lines < needed_empty:
```

disconnect3d: (and without the call to max etc)

pwndbg/commands/context.py
```diff
@@ -670,12 +737,19 @@ def context(subcontext=None, enabled=None) -> None:
             with target as out:
                 res.append(pwndbg.ui.banner("", target=out, width=settings.get("width", None)))
 
+    cmd_lines = 0
```
disconnect3d, line 740: More complexity to the `context` output 😅 ! ;D

Heh, this func is a bit bloated, I started refactoring it recently but did not finish it ;<

pwndbg/commands/context.py
```diff
@@ -670,12 +737,19 @@ def context(subcontext=None, enabled=None) -> None:
...
             if result_settings[target].get("clearing", config_clear_screen) and lines:
                 clear_screen(out)
             out.writelines(line + "\n" for line in lines)
             out.flush()
+        # Assuming the command window (where the prompt is printed) is always stdout
+        if isinstance(target, StdOutput):
+            assert all("\n" not in line for line in lines)
+            cmd_lines += len(lines)
+
+    reserve_lines_maybe(cmd_lines)
```
disconnect3d, line 752: This writes to stdout, but is stdout always used e.g. in a split configuration?

pwndbg/commands/context.py
k4lizen: Well no, then it wouldn't enter the `if` when `leftover_lines`  is negative (the context doesn't fit on the screen), but it should.

pwndbg/commands/context.py
k4lizen: To my knowledge pwndbg only provides a way to redirect context output, we want to write to the stream where the prompt is, the prompt is not a context so it will always be in stdout (which is where gdb puts it, and where we put it for lldb).

If one uses a split configuration they will not experience this type of shaking at all, so they should disable this feature anyways.

Issue #2685, opened 2025-01-20
Handling of remote process lifecycles in LLDB is wrong


mbrla0:
As alluded to in #2684, our handling of the process lifecycle in `pwndbg.dbg.lldb.repl` has quite a significant blind spot, when it comes to remote processes.

Currently, when connecting to a `gdb-remote` server, we assume that it must already be attached to a live process at the time of connection. In other words, we expect that connecting to a remote server will do two things:
1. Report an active process to LLDB, which in turn gives us enough information to build a live instance of `Process`.
2. Fire a stop event, that we use to know when to return control to the user.

These expectations are fine for - and came from - QEMU, as our code that handles remote process lifecycle was entirely written based on what it does.

However, in practice LLDB also allows users to connect to a remote server that has no active process and from there either launch a new process or attach to an existing process in the context of the machine hosting the server. An example of a platform on which this is the norm is iOS, wherein you first attach to the debug server running on the phone, and only after that do you launch your application from inside LLDB using `process launch`. Currently, trying to use Pwndbg on iOS will cause it to hang after a call to `process connect`, as Pwndbg waits for a stop event that will never come.

Fixing this isn't really hard, and I've successfully done it locally as part of bringing up a proof of concept of using Pwndbg on iOS applications, but it's still important to document it, in the context of the PR I mentioned at the start of this issue.

disconnect3d:
Just to be sure. Is this about when we attach to processes via e.g. `attach <pid|name>` or is this only about something like `target remote :1234` in LLDB (I don't even know how to do it, heh) since you said "when connecting to gdb-remote"?

patryk4815:
probably only when you use `process connect` (there is no gdb equivalent)

disconnect3d:
@mbrla0 
> Fixing this isn't really hard, and I've successfully done it locally as part of bringing up a proof of concept of using Pwndbg on iOS applications, but it's still important to document it, in the context of the PR I mentioned at the start of this issue.

Do you have some PoC code for this?

mbrla0:
> @mbrla0 
> > Fixing this isn't really hard, and I've successfully done it locally as part of bringing up a proof of concept of using Pwndbg on iOS applications, but it's still important to document it, in the context of the PR I mentioned at the start of this issue.
> 
> Do you have some PoC code for this?

Need to rebase it, still, but, yeah, I do. If you want me to put it up as a draft, I can. Maybe I can try doing the rebase real quick soon, too.

[...]

PR #2763, opened 2025-02-28
Fix remote connection lifetime handling in LLDB

mbrla0:
As I've mentioned in #2685, lifetime handling in LLDB Pwndbg is partially wrong, and that's currently blocking iOS support in Pwndbg. This PR should fix that, by having the process driver properly distinguish between the cases in which we're connected and have a live process from the ones in which we're just connected to a remote debugging service.

It's marked as a draft mostly because these changes needed to be rebased, and, since I'm on vacation, I haven't had the time to test the rebase very thoroughly yet. All things considered, though, this is a pretty small change.

This PR should be enough to get LLDB Pwndbg to connect to iOS. The procedure should be the same as the one needed to connect from the regular LLDB CLI to an iOS device.


disconnect3d:
Hm I think this can be merged without a rebase, but I'll wait till u come back from vacation

mbrla0:
> Hm I think this can be merged without a rebase, but I'll wait till u come back from vacation

I meant to say I rebased it before I opened the PR. Just haven't thoroughly tested it post-rebase.

Issue #2654, opened 2024-12-29
pwdbg code/regs window messed up with unwanted red color background.


codernavi18:
### Description
When using `pwndbg_code` mode in TUI, the code window (and sometimes regs window) has random red color backgrounds which deteriorates readbility. Screenshot attached.

### Steps to reproduce
As soon as I switch to TUI mode (`ctrl-x a`) and use `tui layout pwndbg_code` , I started seeing random red colored backgrounds in my code window. 

### My setup
I am running latest (master) pwndbg on Ubuntu 24 LTS. I am connected to the Ubuntu 24 via SSH from my Mac iTerm2. I have attached pwndbg to qemu running aarch64 u-boot target.

<img width="1716" alt="image" src="https://github.com/user-attachments/assets/7afd6b26-ee66-4caa-974b-e79775913391" />

codernavi18:
I also noticed that if I use `tui layout pwndbg` then this unwanted red background starts happening in the command window.

disconnect3d:
cc: @peace-maker do you maybe know what is going on in here?

peace-maker:
Is there any output with red background in the normal mode or is this solely made up in the TUI view? Scrolling and text truncation have to be implemented ourselves in the TUI sections.

The way the ANSI escape codes are handled right now is to collect all ANSI escape codes before and after the part of the output line that is in view of the TUI window and add them around the displayed text. The idea was to keep all codes around, including colour reset codes that might not be visible in the current TUI window viewport, to prevent something like what you're seeing from happening. It looks like there is another edge case to consider.

The remaining escape codes are appended without any non-escape code inbetween. Maybe your terminal requires something special?

codernavi18:
The issue is only in the TUI mode. The normal mode is all fine.

@peace-maker upon your suggestions I changed my terminal. Earlier I was doing ssh from my Mac running iTerm2 to my Ubuntu 24 LTS machine. This time I directly logged into the Ubuntu machine and used native GNOME terminal "Version 3.52.0 for GNOME 46". I see same behaviour there as well. 

Attaching the screenshot. The code window in the background can be seen messed up with red color.

<img width="1025" alt="image" src="https://github.com/user-attachments/assets/73165b1f-7bb5-4246-9d71-fab1829938ef" />

peace-maker:
Hm, to be clear: there is no output with red background at all in the CLI mode? Is there any output with other colored background?

[...]

PR #2764, opened 2025-02-28
Fix TUI context sections display after #2743

peace-maker:
We're now using `writelines` to output the context data which wasn't implemented
for the CallOutput redirection layer. Add a smoke test for that output redirection.

Refs #2654

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -204,6 +204,9 @@ class CallOutput:
     def write(self, data) -> None:
         self.func(data)
 
+    def writelines(self, lines_iterable) -> None:
+        self.func("".join(lines_iterable))
+
     def flush(self):
         try:
             return self.func.flush()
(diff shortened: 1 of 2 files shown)


pwndbg/commands/context.py
```diff
@@ -204,6 +204,9 @@ def __eq__(self, other):
     def write(self, data) -> None:
         self.func(data)
 
+    def writelines(self, lines_iterable) -> None:
+        self.func("".join(lines_iterable))
```
disconnect3d, line 208: Wouldnt it be better to call func multiple times? Tbh no idea, this is probably better, huh

peace-maker: I tried that first but there is no way to know when a new list of outputs is written or if the current write is part of the same output.

The TUI windows are cleared before the new output is written from a call to `write` to always show the output at the same position in the window. This caused only the last line from `writelines` to be displayed. We'd need to add some other event or heuristic to clear the output buffer between context displays if we'd want to allow incremental writes with the `CallOutput` redirection method. And just writing all lines at once was the easier solution to this bug.

PR #2766, opened 2025-03-04
Upgrade to Capstone V6

OBarronCS:
This PR updates Pwndbg to work with Capstone V6.

Capstone V6 comes with many improvements and breaking API changes - majors ones are mentioned here: <https://github.com/capstone-engine/capstone/blob/next/docs/cs_v6_release_guide.md>
Many architectures have been updated to add support for newer instructions and fix bugs, making the disassembly more accurate.

A big API change is that Capstone now makes a distinction between an "alias" (the simplified version of an instruction you give an assembler) and the real underlying instruction. This change resulted in a bunch of new instruction ID's that include `_ALIAS_` in the name, and also resulted in Capstone removing some instruction that were really aliases.

For Pwndbg, this meant modifying a bunch of lists of instructions id's to include the new and corrected id's.

Other notable API changes include renaming ARM64 to AARCH64 and changes in the encoding of memory operands in Arm and AARCH64. Other, more subtle bugs arose that our Pwndbg cross architecture tests caught:

Capstone bugs/unexpected changes in behavior reported along the way:
- https://github.com/capstone-engine/capstone/issues/2634
- https://github.com/capstone-engine/capstone/issues/2632
- https://github.com/capstone-engine/capstone/issues/2631
- https://github.com/capstone-engine/capstone/issues/2630

Capstone is currently on a pre-release v6 package, but with the fixes to the issues noted above already merged to the Capstone `next` branch, I believe with the next Capstone package pre-release version we can start using it.

Currently, tests pass when compiling Capstone with the `next` branch locally.

While making the changes, I also bumped into some bugs in AArch64/Arm and added a handful of new tests. Some interesting changes are noted in GitHub comments.

The intention of this PR is to maintain functionality and set the stage to take advantage of new Capstone V6 architectures such as Loongarch64.


pwndbg/lib/regs.py
```diff
@@ -87,11 +98,12 @@ def __init__(
...
         # Otherwise, the values will be clobbered
         # https://github.com/pwndbg/pwndbg/pull/2337
-        self.emulated_regs_order: List[str] = []
+        self.emulated_regs_order: List[EmulatedRegister] = []
 
         for reg in [pc] + list(flags) + [stack, frame] + list(retaddr) + list(misc) + list(gpr):
             if reg and reg not in self.emulated_regs_order:
-                self.emulated_regs_order.append(reg)
+                emu_reg = EmulatedRegister(reg, True if reg in flags else False)
+                self.emulated_regs_order.append(emu_reg)
```
OBarronCS, line 106: Changes related to this list came from a sneaky bug in our use of AArch64 + Unicorn.

We have an optimization where we don't make a call to write a register to the emulator when the register value is zero, with the assumption that registers are always zero by default. This is not the case for the CPSR register in AArch64, which has the `Z` bit enabled by default. If we have a CMP instruction that disables the Z bit (and all other bits happen to be zero as well), then previously we would not set the emulator CPSR to 0, which caused mispredictions in the direction of the branch.

This change forces us to write 0 to the register to correct the behavior.

tests/qemu-tests/tests/user/test_arm.py
```diff
@@ -48,14 +48,14 @@ def test_arm_simple_branch(qemu_assembly_run):
         " ► 0x10000000 <_start>         mov    r2, #5       R2 => 5\n"
         "   0x10000004 <_start+4>       mov    r1, #0xa     R1 => 0xa\n"
         "   0x10000008 <_start+8>       cmp    r0, r1       0x0 - 0xa     CPSR => 0x80000010 [ N z c v q j t e a i f ]\n"
-        "   0x1000000c <_start+12>    ✔ bne    #not_equal                  <not_equal>\n"
```
OBarronCS, line 51: Previously this # before branch/variable names was incorrect, and fixed with changes in how Capstone exposes constant values in AArch64 + arm

tests/unit-tests/test_bit_math.py
```diff
@@ -15,6 +15,9 @@ def test_to_signed():
     assert bit_math.to_signed(0xFFFF_FFFF, 32) == -1
     assert bit_math.to_signed(0x8000_0000, 32) == -(2**31)
 
+    # Ignore higher bits
+    assert bit_math.to_signed(0xF_FF, 8) == -1
```
OBarronCS, line 19: Fixed bug that propagated to incorrect sign extensions in AArch64 sign extend operations (things like SXTB).

patryk4815:
> This change resulted in a bunch of new instruction ID's that include _ALIAS_ in the name

Can we just use `instruction.unalias_id` and then we can skip(?) adding a lot of duplicated `_ALIAS_` 🤔

OBarronCS:
Unfortunately, we do need to include all the `_ALIAS_` instructions - it's best to think of them as completely separate instructions.
- As an example, the MIPS `nop` instruction is really an alias for `sll $0, $0, 0`. We need to use the alias to expose to the user/the enhancement code that this instruction should be processed as a nop, not an `sll`. This same logic extends to dozens of other instructions across all the architectures with aliases, where sometimes what we think of an instruction really only exists as an alias.
- Capstone exposes the `details` of the instruction in terms of the alias (details referring to the operand data, where the alias might have a simplified set of operands). By using the alias_id, we align the logical instruction with the metadata that is exposed.

OBarronCS:
With the release of Capstone v6.0.0-Alpha4 (https://github.com/capstone-engine/capstone/releases/tag/6.0.0-Alpha4), this PR is ready to go.

patryk4815:
@OBarronCS  Please apply to fix release/nix :)
```
diff --git a/nix/pyenv.nix b/nix/pyenv.nix
index 577e3485..4fd08ea1 100644
--- a/nix/pyenv.nix
+++ b/nix/pyenv.nix
@@ -162,37 +162,10 @@ let
       }:
       prev.capstone.overrideAttrs (
         old:
-        lib.optionalAttrs ((isBuildSource old) && stdenv.hostPlatform.isDarwin) {
+        lib.optionalAttrs (isBuildSource old) {
           nativeBuildInputs = old.nativeBuildInputs ++ [
             cmake
-            fixDarwinDylibNames
           ];
-
-          preBuild = ''
-            sed -i 's/^IS_APPLE := .*$/IS_APPLE := 1/' ./src/Makefile
-
-            substituteInPlace ./setup.py \
-                --replace-fail "import sys" "import sys; sys.argv.extend(('--plat-name', 'any'))" || true
-          '';
-
-          # See: https://github.com/capstone-engine/capstone/issues/2621
-          postPatch = (
-            let
-              gitSrc = fetchFromGitHub {
-                owner = "capstone-engine";
-                repo = "capstone";
-                rev = old.version;
-                hash = "sha256-VGqqrixg7LaqRWTAEBzpC+gUTchncz3Oa2pSq8GLskI=";
-              };
-            in
-            ''
-              cp ${gitSrc}/capstone.pc.in src/
-              cp ${gitSrc}/capstone-config.cmake.in src/
-              cp ${gitSrc}/cmake_uninstall.cmake.in src/
-              cp ${gitSrc}/CPackConfig.txt src/
-              cp ${gitSrc}/CPackConfig.cmake src/
-            ''
-          );
         }
       )
     ) { };
```

patryk4815 (approved):

disconnect3d (approved):

PR #2767, opened 2025-03-04
Don't print redundant empty line in split context setup

k4lizen:
See the issue here: https://github.com/pwndbg/pwndbg/pull/2743#issuecomment-2671596820

With this:
image


k4lizen:
Something interesting is that if you just always do (in non split setup)
```python
            out.writelines("\n" + line for line in lines)
```
it messes up gdb:
image
but doesn't mess up lldb:
image
I guess because they use different readline libs?

In any case, this patch works on both.

PR #2770, opened 2025-03-05
Native Apple Silicon GDB :)

patryk4815:
Add portable native GDB multi-arch for Apple Silicon :)

Still it is not possible to debug aarch64-darwin binaries via GDB.
But GDB is useful as a frontend for remote debugging.


disconnect3d:
Awesome! <3

PR #2771, opened 2025-03-05
PwndbgArchitecture

OBarronCS:
This PR introduces a class based mechanism to refer to information related to the current architecture being debugged, as well as other related context, with the goal of simplifying and cleaning up code.

See original PR and discussion here: #2367

### DIFF
--- a/pwndbg/aglib/__init__.py
+++ b/pwndbg/aglib/__init__.py
@@ -2,12 +2,13 @@ from __future__ import annotations
 
 from pwndbg.aglib import arch as arch_mod
 from pwndbg.aglib.arch import PwndbgArchitecture
+from pwndbg.aglib.arch import get_pwndbg_architecture
 
 # from pwndbg.aglib.arch import arch as arch
 
 regs = None
 
-arch: PwndbgArchitecture = PwndbgArchitecture.get_arch("i386")
+arch: PwndbgArchitecture = get_pwndbg_architecture("i386")
 
 
 def load_aglib():
--- a/pwndbg/aglib/__init__.py
+++ b/pwndbg/aglib/__init__.py
@@ -1,14 +1,14 @@
 from __future__ import annotations
 
 from pwndbg.aglib import arch as arch_mod
-# from pwndbg.aglib.arch import arch as arch
-
 from pwndbg.aglib.arch import PwndbgArchitecture
+
+# from pwndbg.aglib.arch import arch as arch
 
 regs = None
 
-# arch: PwndbgArchitecture = None
 arch: PwndbgArchitecture = PwndbgArchitecture.get_arch("i386")
+
 
 def load_aglib():
     import pwndbg.aglib.argv
@@ -38,6 +38,7 @@ def load_aglib():
     regs_: regs_mod.module = regs_mod
     global regs
     regs = regs_
+
 
 def set_arch(pwndbg_arch: PwndbgArchitecture):
     global arch
--- a/pwndbg/aglib/__init__.py
+++ b/pwndbg/aglib/__init__.py
@@ -4,8 +4,6 @@ from pwndbg.aglib import arch as arch_mod
 from pwndbg.aglib.arch import PwndbgArchitecture
 from pwndbg.aglib.arch import get_pwndbg_architecture
 
-# from pwndbg.aglib.arch import arch as arch
-
 regs = None
 
 arch: PwndbgArchitecture = get_pwndbg_architecture("i386")
--- a/pwndbg/aglib/__init__.py
+++ b/pwndbg/aglib/__init__.py
@@ -1,10 +1,14 @@
 from __future__ import annotations
 
 from pwndbg.aglib import arch as arch_mod
-from pwndbg.aglib.arch import arch as arch
+# from pwndbg.aglib.arch import arch as arch
+
+from pwndbg.aglib.arch import PwndbgArchitecture
 
 regs = None
 
+# arch: PwndbgArchitecture = None
+arch: PwndbgArchitecture = PwndbgArchitecture.get_arch("i386")
 
 def load_aglib():
     import pwndbg.aglib.argv
@@ -34,3 +38,7 @@ def load_aglib():
     regs_: regs_mod.module = regs_mod
     global regs
     regs = regs_
+
+def set_arch(pwndbg_arch: PwndbgArchitecture):
+    global arch
+    arch = pwndbg_arch
--- a/pwndbg/aglib/arch.py
+++ b/pwndbg/aglib/arch.py
@@ -31,9 +31,6 @@ from typing_extensions import override
 import pwndbg
 import pwndbg.aglib
 from pwndbg.aglib import typeinfo
-
-# Uncommenting this create circular import
-# from pwndbg.aglib.disasm import emulated_arm_mode_cache
 from pwndbg.lib.abi import ABI
 from pwndbg.lib.abi import DEFAULT_ABIS
 from pwndbg.lib.abi import SIGRETURN_ABIS
--- a/pwndbg/aglib/arch.py
+++ b/pwndbg/aglib/arch.py
@@ -119,10 +119,9 @@ class PwndbgArchitecture(ArchDefinition):
         self.ptrbits: int = self.ptrsize * 8
         self.ptrmask: int = (1 << self.ptrbits) - 1
 
-        # TODO - allow these to be overriden by other means
-
         default_abi_identifer = (self.ptrbits, self.name, "linux")
 
+        # TODO - allow these to be selected by some other means
         self.abi = DEFAULT_ABIS.get(default_abi_identifer)
         self.syscall_abi = SYSCALL_ABIS.get(default_abi_identifer)
         self.sigreturn_abi = SIGRETURN_ABIS.get(default_abi_identifer)
--- a/pwndbg/aglib/arch.py
+++ b/pwndbg/aglib/arch.py
@@ -71,12 +71,13 @@ class PwndbgArchitecture(ArchDefinition):
     """
 
     name: PWNDBG_SUPPORTED_ARCHITECTURES_TYPE
-    name_raw: str  # The raw string returned by the debugger
+    # The raw string returned by the debugger. Example: "mips:isa64r6"
+    name_raw: str
     endian: EndianType
     ptrsize: int
     ptrbits: int
     ptrmask: int
-    abi: ABI | None
+    function_abi: ABI | None
     syscall_abi: SyscallABI | None
     sigreturn_abi: SyscallABI | None
     platform: Platform
@@ -118,8 +119,7 @@ class PwndbgArchitecture(ArchDefinition):
 
         default_abi_identifer = (self.ptrbits, self.name, "linux")
 
-        # TODO - allow these to be selected by some other means
-        self.abi = DEFAULT_ABIS.get(default_abi_identifer)
+        self.function_abi = DEFAULT_ABIS.get(default_abi_identifer)
         self.syscall_abi = SYSCALL_ABIS.get(default_abi_identifer)
         self.sigreturn_abi = SIGRETURN_ABIS.get(default_abi_identifer)
 
--- a/pwndbg/aglib/arch.py
+++ b/pwndbg/aglib/arch.py
@@ -256,7 +256,7 @@ class SparcArch(PwndbgArchitecture):
         super().__init__("sparc")
 
     def get_capstone_constants(self, address: int) -> Tuple[int, int]:
-        if pwndbg.dbg.is_gdblib_available() and "v9" in self.name_raw:
+        if "v9" in self.name_raw:
             mode = CS_MODE_V9
         else:
             # The ptrsize base modes cause capstone.CsError: Invalid mode (CS_ERR_MODE)
--- a/pwndbg/aglib/arch.py
+++ b/pwndbg/aglib/arch.py
@@ -168,6 +168,7 @@ class AMD64Arch(PwndbgArchitecture):
     def __init__(self) -> None:
         super().__init__("x86-64")
 
+    @override
     def get_capstone_constants(self, address: int) -> Tuple[int, int]:
         return (CS_ARCH_X86, CS_MODE_64)
 
@@ -182,6 +183,7 @@ class i386Arch(PwndbgArchitecture):
     def __init__(self) -> None:
         super().__init__("i386")
 
+    @override
     def get_capstone_constants(self, address: int) -> Tuple[int, int]:
         return (CS_ARCH_X86, CS_MODE_32)
 
@@ -196,6 +198,7 @@ class i8086Arch(PwndbgArchitecture):
     def __init__(self) -> None:
         super().__init__("i8086")
 
+    @override
     def get_capstone_constants(self, address: int) -> Tuple[int, int]:
         return (CS_ARCH_X86, CS_MODE_16)
 
@@ -256,6 +259,7 @@ class AArch64Arch(PwndbgArchitecture):
     def __init__(self) -> None:
         super().__init__("aarch64")
 
+    @override
     def get_capstone_constants(self, address: int) -> Tuple[int, int]:
         return (CS_ARCH_AARCH64, CS_MODE_ARM)
 
@@ -266,8 +270,8 @@ class PowerPCArch(PwndbgArchitecture):
     def __init__(self) -> None:
         super().__init__("powerpc")
 
+    @override
     def get_capstone_constants(self, address: int) -> Tuple[int, int]:
-        # TODO - what about 32-bit powerpc?
         return (CS_ARCH_PPC, CS_MODE_64)
 
 
@@ -277,6 +281,7 @@ class SparcArch(PwndbgArchitecture):
     def __init__(self) -> None:
         super().__init__("sparc")
 
+    @override
     def get_capstone_constants(self, address: int) -> Tuple[int, int]:
         mode = CS_MODE_V9 if self.ptrsize == 8 else 0
         return (CS_ARCH_SPARC, mode)
@@ -288,6 +293,7 @@ class RISCV32Arch(PwndbgArchitecture):
     def __init__(self) -> None:
         super().__init__("rv32")
 
+    @override
     def get_capstone_constants(self, address: int) -> Tuple[int, int]:
         return (CS_ARCH_RISCV, CS_MODE_RISCV32 | CS_MODE_RISCVC)
 
@@ -298,6 +304,7 @@ class RISCV64Arch(PwndbgArchitecture):
     def __init__(self) -> None:
         super().__init__("rv64")
 
+    @override
     def get_capstone_constants(self, address: int) -> Tuple[int, int]:
         return (CS_ARCH_RISCV, CS_MODE_RISCV64 | CS_MODE_RISCVC)
 
@@ -308,6 +315,7 @@ class MipsArch(PwndbgArchitecture):
     def __init__(self) -> None:
         super().__init__("mips")
 
+    @override
     def get_capstone_constants(self, address: int) -> Tuple[int, int]:
         extra = 0
         for attribute in self.attributes:
@@ -329,6 +337,7 @@ class Loongarch64Arch(PwndbgArchitecture):
     def __init__(self) -> None:
         super().__init__("loongarch64")
 
+    @override
     def get_capstone_constants(self, address: int) -> Tuple[int, int]:
         return (CS_ARCH_LOONGARCH, CS_MODE_LOONGARCH64)
 
@@ -339,6 +348,7 @@ class S390xArch(PwndbgArchitecture):
     def __init__(self) -> None:
         super().__init__("s390x")
 
+    @override
     def get_capstone_constants(self, address: int) -> Tuple[int, int]:
         return (CS_ARCH_SYSTEMZ, 0)
 
--- a/pwndbg/aglib/arch.py
+++ b/pwndbg/aglib/arch.py
@@ -201,7 +201,7 @@ class ArmArch(PwndbgArchitecture):
     def read_thumb_bit(self) -> Literal[0, 1]:
         # When program initially starts, cpsr may not be readable
         if (cpsr := pwndbg.aglib.regs.cpsr) is not None:
-            return (cpsr >> 5) & 1 # type: ignore[return-value]
+            return (cpsr >> 5) & 1  # type: ignore[return-value]
 
         return 0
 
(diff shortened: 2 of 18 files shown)


OBarronCS:
Regarding @patryk4815's notes here: https://github.com/pwndbg/pwndbg/pull/2367#issuecomment-2488956588, referring to statements like

```python
pwndbg.aglib.arch.current.is_aarch64().is_e()
pwndbg.aglib.arch.current.is_sparc().is_v9()
pwndbg.aglib.arch.current.is_mips().is_r6()
pwndbg.aglib.arch.current.is_arm().read_thumb_bit()
```
 I think this is a good idea and is a little cleaner than other methods of manually checking these sorts of things. Should we limit these sorts of lookups to boolean values? Otherwise, we'd need to somehow indicate the case where the response is not relevant, perhaps responding with `None`?

pwndbg/aglib/__init__.py
```diff
@@ -1,10 +1,14 @@
...
 from pwndbg.aglib import arch as arch_mod
-from pwndbg.aglib.arch import arch as arch
+from pwndbg.aglib.arch import PwndbgArchitecture
+
+# from pwndbg.aglib.arch import arch as arch
 
 regs = None
 
+arch: PwndbgArchitecture = PwndbgArchitecture.get_arch("i386")
+
```
OBarronCS: Contrary to my understanding of Python imports, this seems to work great - the imported value is not stale anywhere that does `import pwndbg.aglib.arch`.

pwndbg/aglib/arch.py
```diff
@@ -1,39 +1,317 @@
...
+    """
+
+    name: PWNDBG_SUPPORTED_ARCHITECTURES_TYPE
+    endian: EndianType
+    ptrsize: int
+    ptrbits: int
+    ptrmask: int
+    abi: ABI | None
+    syscall_abi: SyscallABI | None
+    sigreturn_abi: SyscallABI | None
```
OBarronCS, line 93: Attaching ABI's to this object makes it simpler for us to change them when necessary (such as when debugging on macos).

pwndbg/aglib/arch.py
```diff
@@ -42,4 +320,8 @@ def update() -> None:
     pwnlib.context.context.arch = PWNLIB_ARCH_MAPPINGS.get(a.name, "none")
     pwnlib.context.context.bits = a.ptrsize * 8
 
-    arch.update(a.name, a.ptrsize, a.endian)
+    if a.name != pwndbg.aglib.arch.name:
+        pwndbg_arch = PwndbgArchitecture.get_arch(a.name)
+        pwndbg.aglib.set_arch(pwndbg_arch)
```
OBarronCS: Once we detect that the architecture has changed, get the correct class and set the global variable to the instance.

pwndbg/lib/arch.py
```diff
@@ -46,7 +46,7 @@
 FMT_LITTLE_ENDIAN = {1: "B", 2: "<H", 4: "<I", 8: "<Q"}
 FMT_BIG_ENDIAN = {1: "B", 2: ">H", 4: ">I", 8: ">Q"}
 
-
+# Unused
```
OBarronCS: All the functionality of Arch is in PwndbgArchitecture

pwndbg/aglib/arch.py
```diff
@@ -1,39 +1,317 @@
...
+from capstone import CS_MODE_THUMB
+from capstone import CS_MODE_V9
+from typing_extensions import override
 
 import pwndbg
+import pwndbg.aglib
+from pwndbg.aglib import typeinfo
+
+# Uncommenting this create circular import
+# from pwndbg.aglib.disasm import emulated_arm_mode_cache
```
OBarronCS: Some trickiness with circular imports

pwndbg/aglib/arch.py
```diff
@@ -1,39 +1,317 @@
 from __future__ import annotations
 
+import struct
+from typing import Dict
 from typing import Literal
+from typing import Tuple
 
+import gdb
```
patryk4815:
```suggestion
if pwndbg.dbg.is_gdblib_available():
    import gdb
```
fixes lldb

OBarronCS: Newest changes avoids the use of `gdb` in the arch file

pwndbg/aglib/arch.py
```diff
@@ -1,39 +1,319 @@
...
+
+        # Pointer size in bytes
+        self.ptrsize: int = arch_definition.ptrsize
+        self.ptrbits: int = self.ptrsize * 8
+        self.ptrmask: int = (1 << self.ptrbits) - 1
+
+        default_abi_identifer = (self.ptrbits, self.name, "linux")
+
+        # TODO - allow these to be selected by some other means
+        self.abi = DEFAULT_ABIS.get(default_abi_identifer)
```
OBarronCS: A decision we need to make is what happens for "default" values here? Right now, it returns None (same behavior as before), which was historically been the source of crashes because there are a handful of functions assume this is a valid ABI object.

We could create some "dummy" object, or at all the callsights check that this value is not "None"?

OBarronCS:
Made a lot more changes, putting this into a more final state.

Given the scope of changes, it might be wise to get an initial version in which preserves functionality. Then, future PR's can add on to this to get more of the benefits we want using this new way of representing architecture.

pwndbg/lib/arch.py
```diff
@@ -43,49 +44,20 @@
 }
 
 
-FMT_LITTLE_ENDIAN = {1: "B", 2: "<H", 4: "<I", 8: "<Q"}
-FMT_BIG_ENDIAN = {1: "B", 2: ">H", 4: ">I", 8: ">Q"}
+# The platform being debugged
+class Platform(Enum):
+    LINUX = (auto(),)
+    MACOSX = (auto(),)
```
OBarronCS: Would we prefer `DARWIN` here? One example use for "platform" is to select the ABI.

pwndbg/aglib/arch.py
```diff
@@ -1,39 +1,319 @@
...
+        return (CS_ARCH_PPC, CS_MODE_64)
 
 
-arch: Arch = Arch("i386", 4, "little")
+class SparcArch(PwndbgArchitecture):
+    def __init__(self) -> None:
+        super().__init__("sparc")
+
+    def get_capstone_constants(self, address: int) -> Tuple[int, int]:
+        if pwndbg.dbg.is_gdblib_available() and "v9" in self.name_raw:
```
patryk4815:
```suggestion
        if "v9" in self.name_raw:
```

`pwndbg.dbg.is_gdblib_available()` is not needed here after removing gdb

pwndbg/aglib/arch.py
```diff
@@ -1,39 +1,319 @@
...
+        return (CS_ARCH_RISCV, CS_MODE_RISCV64 | CS_MODE_RISCVC)
+
+
+class MipsArch(PwndbgArchitecture):
+    def __init__(self) -> None:
+        super().__init__("mips")
+
+    def get_capstone_constants(self, address: int) -> Tuple[int, int]:
+        # TODO - Capstone v6 increased the number of MIPS constants
+        if pwndbg.dbg.is_gdblib_available() and "isa32r6" in self.name_raw:
```
patryk4815: `pwndbg.dbg.is_gdblib_available() is not needed here after removing gdb`
```suggestion
        if "isa32r6" in self.name_raw:
```

pwndbg/aglib/__init__.py
disconnect3d: What if one does `from pwndbg.aglib import arch`? then it would be copy and be problematic?

OBarronCS: Yes, I believe so - ideally there is a linting rule we can use to forbid importing in that way.

disconnect3d:
> Regarding @patryk4815's notes here: [#2367 (comment)](https://github.com/pwndbg/pwndbg/pull/2367#issuecomment-2488956588), referring to statements like
> 
> ```python
> pwndbg.aglib.arch.current.is_aarch64().is_e()
> pwndbg.aglib.arch.current.is_sparc().is_v9()
> pwndbg.aglib.arch.current.is_mips().is_r6()
> pwndbg.aglib.arch.current.is_arm().read_thumb_bit()
> ```
> 
> I think this is a good idea and is a little cleaner than other methods of manually checking these sorts of things. Should we limit these sorts of lookups to boolean values? Otherwise, we'd need to somehow indicate the case where the response is not relevant, perhaps responding with `None`?

I think its a bit weird to have a `is_something()` return an object that has additional `is_sth_else`.

Would u ever call `is_e()` if you are in a different arch than aarch64? Same for other cases?
If not, then u can just do `current.is_e()` and assume you are in aarch64 that has this method and not bother to have it in base class ( ? ? ? ).

pwndbg/aglib/arch.py
```diff
@@ -1,39 +1,319 @@
...
+from capstone import CS_MODE_ARM
+from capstone import CS_MODE_MCLASS
+from capstone import CS_MODE_MIPS32
+from capstone import CS_MODE_MIPS32R6
+from capstone import CS_MODE_MIPS64
+from capstone import CS_MODE_RISCV32
+from capstone import CS_MODE_RISCV64
+from capstone import CS_MODE_RISCVC
+from capstone import CS_MODE_THUMB
+from capstone import CS_MODE_V9
```
disconnect3d, line 31: can we collapse those to: `from capstone import (a,b,c,d)`?

pwndbg/aglib/arch.py
```diff
@@ -1,39 +1,319 @@
...
+from typing_extensions import override
 
 import pwndbg
+import pwndbg.aglib
+from pwndbg.aglib import typeinfo
+
+# Uncommenting this create circular import
+# from pwndbg.aglib.disasm import emulated_arm_mode_cache
```
disconnect3d: So why leaving the comment here? :)

OBarronCS: Good point - removing this, was meant to be temporary

pwndbg/aglib/arch.py
OBarronCS: Unfortunately it appears that our linting rules do not like this, `./lint.sh -f` automatically expands each import into it's own line

OBarronCS:
> I think its a bit weird to have a `is_something()` return an object that has additional `is_sth_else`.
> 
> Would u ever call `is_e()` if you are in a different arch than aarch64? Same for other cases? If not, then u can just do `current.is_e()` and assume you are in aarch64 that has this method and not bother to have it in base class ( ? ? ? ).

I agree - let's not implement this sort of `.is_x().is_sth_else()` pattern.

On the matter of not having the definition in the base class, this would introduce the additional annoyance of a "compile/lint time" error - with our linting rules, it would not be allowed, since the linting would check on the parent `PwndbgArchitecture` type and the method/attribute would not be found. Auto-complete would also be missing. For the sake of this "type safety", I think the next best solution is to include the method definitions in the parent class (which either throw errors or return some default value depending on the context) and have the child class that actually uses it override it (ex: read_thumb_bit())`

pwndbg/aglib/arch.py
disconnect3d: We shall change it in the future.

pwndbg/aglib/arch.py
```diff
@@ -1,39 +1,314 @@
...
+        """
+        While debugging a process, certain aspects of the architecture can change.
+
+        For example:
+        - Some architectures can change endianness dynamically.
+        """
+        self.name_raw = arch_definition.name_raw
+        self.platform = arch_definition.platform
+
+        self.endian: EndianType = arch_definition.endian
```
disconnect3d, line 124: Shouldn't such changes be done only in child classes and the base should just keep things as is?

pwndbg/aglib/arch.py
disconnect3d: Probably have a default or handle the None :x

pwndbg/aglib/arch.py
```diff
@@ -1,39 +1,314 @@
...
+class SparcArch(PwndbgArchitecture):
+    def __init__(self) -> None:
+        super().__init__("sparc")
+
+    def get_capstone_constants(self, address: int) -> Tuple[int, int]:
+        if "v9" in self.name_raw:
+            mode = CS_MODE_V9
+        else:
+            # The ptrsize base modes cause capstone.CsError: Invalid mode (CS_ERR_MODE)
+            mode = 0
```
disconnect3d: 1) This could be just:
```py
mode = CS_MODE_V9 if "v9" in self.name_raw else 0
```

2) Can we give an example of name_raw with sparc v9 in a comment?

3) The comment about ptrsize is really unclear to me. Can you extend it and elaborate a bit more?

pwndbg/aglib/arch.py
```diff
@@ -1,39 +1,314 @@
...
+
+        return (CS_ARCH_SPARC, mode)
+
+
+class RISCV32Arch(PwndbgArchitecture):
+    def __init__(self) -> None:
+        super().__init__("rv32")
+
+    def get_capstone_constants(self, address: int) -> Tuple[int, int]:
+        return (CS_ARCH_RISCV, CS_MODE_RISCV32 | CS_MODE_RISCVC)
```
disconnect3d, line 299:
```suggestion
        # We enable the RISC-V compressed ISA extension by default (CS_MODE_RISCVC)
        return (CS_ARCH_RISCV, CS_MODE_RISCV32 | CS_MODE_RISCVC)
```

pwndbg/aglib/arch.py
```diff
@@ -1,39 +1,314 @@
...
+
+
+class RISCV64Arch(PwndbgArchitecture):
+    def __init__(self) -> None:
+        super().__init__("rv64")
+
+    def get_capstone_constants(self, address: int) -> Tuple[int, int]:
+        return (CS_ARCH_RISCV, CS_MODE_RISCV64 | CS_MODE_RISCVC)
```
disconnect3d, line 310:
```suggestion
        # We enable the RISC-V compressed ISA extension by default (CS_MODE_RISCVC)
        return (CS_ARCH_RISCV, CS_MODE_RISCV64 | CS_MODE_RISCVC)
```

pwndbg/aglib/arch.py
disconnect3d: Actually, is there a reason why we do this? Can we somehow detect if this is there or not?

pwndbg/aglib/arch.py
```diff
@@ -1,39 +1,314 @@
...
+    name_raw: str  # The raw string returned by the debugger
+    endian: EndianType
+    ptrsize: int
+    ptrbits: int
+    ptrmask: int
+    abi: ABI | None
+    syscall_abi: SyscallABI | None
+    sigreturn_abi: SyscallABI | None
+    platform: Platform
+    # environment
```
disconnect3d: Not sure what's this comment about, to be removed?

pwndbg/aglib/arch.py
```diff
@@ -1,39 +1,314 @@
...
+        return (CS_ARCH_RISCV, CS_MODE_RISCV64 | CS_MODE_RISCVC)
+
+
+class MipsArch(PwndbgArchitecture):
+    def __init__(self) -> None:
+        super().__init__("mips")
+
+    def get_capstone_constants(self, address: int) -> Tuple[int, int]:
+        # TODO - Capstone v6 increased the number of MIPS constants
+        if "isa32r6" in self.name_raw:
```
disconnect3d: Can we provide a comment with a full name_raw example?

pwndbg/aglib/arch.py
```diff
@@ -1,39 +1,314 @@
...
+i386Arch()
+i8086Arch()
+ArmArch()
+ArmCortexArch()
+AArch64Arch()
+PowerPCArch()
+SparcArch()
+RISCV32Arch()
+RISCV64Arch()
+MipsArch()
```
disconnect3d: I don't like how this gets registered. We instantiate an object which seems to die immediately, but in practice it doesn't because it gets added to a static field. We could alternatively have all of this as a global list and have a function instead of static field for it.

It's not like I would die for changing it, but still.

pwndbg/aglib/arch.py
patryk4815: https://github.com/pwndbg/pwndbg/pull/2367#issuecomment-2487085647

pwndbg/aglib/disasm/__init__.py
```diff
@@ -124,19 +108,17 @@ def clear_on_reg_mem_change() -> None:
...
+        arch, mode = pwndbg.aglib.arch.get_capstone_constants(address)
 
-    mode |= CapstoneEndian[endian]
-
-    flavor = pwndbg.dbg.x86_disassembly_flavor()
+    mode |= CapstoneEndian[pwndbg.aglib.arch.endian]
 
     cs = Cs(arch, mode)
+
+    flavor = pwndbg.dbg.x86_disassembly_flavor()
```
disconnect3d: Hmm, Isn't it weird that we get a x86 flavor while we may not be in x86-64?

pwndbg/aglib/disasm/__init__.py
```diff
@@ -205,10 +143,11 @@ def get_one_instruction(
         if cached is not None:
             return cached
 
-    if pwndbg.aglib.arch.name not in CapstoneArch:
+    cs_info = pwndbg.aglib.arch.get_capstone_constants(address)
+    if None in cs_info:
```
disconnect3d: Is there a case when we get `(X, None)` here instead of `(None, None)`?

Should we maybe change the `get_capstone_constant` to return `None | cs_info` instead?

disconnect3d (changes requested):
I like this! I have added a few comments here and there :)

pwndbg/aglib/arch.py
disconnect3d: Makes sense, but this is about... platform? then why not comment it above platform or on the same line?

pwndbg/aglib/arch.py
OBarronCS: The comment is outdated and doesn't apply/isn't correct, so I'm removing it.

The "v9" in self.name_raw check was to detect additional attributes of the architecture, `gdb` detects sparc as  "sparc:v9" or "sparc:v8", for example, with the v9, v8 being a variant. This was a hacky way to check for this "variant", as this logic was only used in two places in the code (used as well for a mips to check for "isa32r6").

I think a better solution would be to encode this sort of information in a more reliable way, maybe some sort of list of ArchitectureAttribute object or enums, so one could do 
```python
if ArchAttribute.SPARC_V9 in self.arch_attributes:
````
or something like that.

File: pwndbg/aglib/arch.py (hunk 15 above)
Comment (OBarronCS) [reply pwndbg/aglib/arch.py:299]: I believe the compressed instructions are an extension that add encodings which don't collide with the existing riscv instructions -- capstone should really internally have this enabled all the time, since the compressed instructions are very common. There is a register (`misa`) that can be checked for this (but it may not always be implemented). I think it's simpler to always have it enabled.

File: pwndbg/aglib/arch.py (hunk 17 above)
Comment (OBarronCS) [reply pwndbg/aglib/arch.py]: Forgot to remove this comment.

File: pwndbg/aglib/arch.py (hunk 19 above)
Comment (OBarronCS) [reply pwndbg/aglib/arch.py]: I agree, not very clean. I created a function that takes these objects as a parameter to register them so that is it more clear how they are being used.

File: pwndbg/aglib/disasm/__init__.py (hunk 20 above)
Comment (OBarronCS) [reply pwndbg/aglib/disasm/__init__.py]: That function should be renamed to just "disassembly_flavor()` - in all other arches, gdb returns "intel" as the disassembly flavour, while x86 also allows "att".

File: pwndbg/aglib/disasm/__init__.py (hunk 21 above)
Comment (OBarronCS) [reply pwndbg/aglib/disasm/__init__.py]: I made this change

File: pwndbg/aglib/arch.py (hunk 7 above)
Comment (OBarronCS) [reply pwndbg/aglib/arch.py]: Done - all places that refer to these check if it's None and return early/raise error where appropriate.

File: pwndbg/aglib/arch.py (hunk 13 above)
Comment (OBarronCS) [reply pwndbg/aglib/arch.py:124]: For some architectures (as the way we categorize them), many of the things that this `update()` method sets indeed are locked in. However, in many cases, many of these the attributes can change and having this parent function that updates all of them is very convenient.

As an example, the endianness of some architectures can be little/big and some can even change dynamically at runtime. Rather than go to each arch and define it explicitly or determine if it changes dynamically, it's simpler to just have this parent function set it at runtime.

As another case, we define one class for Mips, which encapsulates all types of mips, including 32-bit and 64-bit, so we need to detect and set the ptrsize at runtime, which this function allows for.

File: pwndbg/aglib/arch.py [hunk 22]
```diff
@@ -42,4 +319,14 @@ def update() -> None:
...
 
-    arch.update(a.name, a.ptrsize, a.endian)
+    if a.name != pwndbg.aglib.arch.name:
+        pwndbg_arch = get_pwndbg_architecture(a.name)
+        if pwndbg_arch is None:
+            raise pwndbg.dbg_mod.Error(
+                f"Unsupported architecture: {a.name}. "
+                f"It may be that Pwndbg is not correctly categorizing the architecture. "
+                f"Please file a bug report. "
+            )
```
OBarronCS, line 393: Added a handler to detect unrecognized architectures - previously it would error out on a key error when fetching registers.

pwndbg/aglib/arch.py
OBarronCS: I added a comment with an example

OBarronCS:
I went through and made changes to address the comments in the review

OBarronCS:
Interestingly, tests running on ubuntu 22.04 currently fail because they are on Python 3.10 which has a strange behavior in Protocol `__init__` methods which was fixed in later Python versions - https://github.com/python/cpython/issues/88970

Will have to remove the use of Protocol class in this case I suppose.

patryk4815:
> Python 3.10 Support Ends in 1 year and 6 months (31 Oct 2026) 

So we have to wait soo long to drop py3.10 :(

pwndbg/aglib/arch.py
OBarronCS: I removed name_raw (it would only work with gdb) and replaced it with a new ArchAttribute class so we can give the arch some more information about arch / ISA variants.

pwndbg/aglib/arch.py
OBarronCS: In this specific case, turns out the V9 refers to 64-bit sparc, so I replaced this with a check on the ptrsize.

disconnect3d:
@OBarronCS pls fix conflicts

disconnect3d (approved):

PR #2777, opened 2025-03-08
Simplify local testing inside Docker environment

OBarronCS:
This PR makes a small change to the Dockerfile+Docker compose file to make it easier to run the tests locally.

Many of the tests are dependent on the environment (libc version dependencies as well as the entire cross-architecture test suite relying on cross compilers being installed), and this makes it a little easier to locally run the tests in the same environment as CI/CD.

You can run:
```sh
docker compose run --rm --build ubuntu24.04-mount ./qemu-tests.sh
```
The ./pwndbg directory is mounted, and by adding a new "stage" to the Dockerfile, we can avoid a rebuild of the image when changing files, making it easier to run the tests.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,3 @@
-version: "3.8"
 services:
   base: &base-spec
     build: .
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -28,6 +27,16 @@ services:
       args:
         image: ubuntu:24.04
 
+  ubuntu24.04-mount:
+    <<: *base-spec
+    build:
+      context: .
+      target: base
```
OBarronCS, line 34: We target the "base" stage - without specifying, it would target the result of the last stage, "full", which copies the ./pwndbg directory upon detecting any changes. Since ./pwndbg is mounted, this is unnecessary and adds build time.

PR #2778, opened 2025-03-11
Add test for mixed case handling in hex2ptr_common

saturnines:
Added a test case to check that hex2ptr_mmon handles hex strings with mixed uppercase and lowercase letters.


disconnect3d:
This is neat, thanks!

PR #2779, opened 2025-03-14
Add command for obtaining libc version and sources

### LINKED ISSUE #2775: Command for obtaining libc version and sources
It would be nice to have a command that would figure out the libc version used by the binary (during remote debugging) and that would give us a link to its sources if possible?

Bryan-atu:
Adds check for libc version and sources.

### DIFF
--- a/pwndbg/commands/libcinfo.py
+++ b/pwndbg/commands/libcinfo.py
@@ -1,9 +1,14 @@
-import gdb
+from __future__ import annotations
+
 import os
-import pwndbg.commands
-import pwndbg.aglib.vmmap
 import re
 import subprocess
+
+import gdb
+
+import pwndbg.aglib.vmmap
+import pwndbg.commands
+
 
 @pwndbg.commands.Command
 @pwndbg.commands.OnlyWhenLocal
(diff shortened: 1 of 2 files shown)


pwndbg/commands/libcinfo.py
```diff
@@ -0,0 +1,56 @@
...
+
+        try:
+            # Run the `vmmap` command and capture output
+            vmmap_output = gdb.execute("vmmap", to_string=True)
+            libc_path = None
+            real_libc_path = None
+
+            # Search for the libc entry in vmmap output
+            for line in vmmap_output.splitlines():
+                if "libc" in line and ".so" in line:
```
CptGibbon: Does this catch names like libcrypt.so etc. ?

pwndbg/commands/libcinfo.py
```diff
@@ -0,0 +1,56 @@
...
+import subprocess
+import os
+
+@pwndbg.commands.Command
+def libcinfo():
+        """Identifies the libc version used by the binary and provides a link to its sources."""
+
+        try:
+            # Run the `vmmap` command and capture output
+            vmmap_output = gdb.execute("vmmap", to_string=True)
```
patryk4815: please use `pwndbg.aglib.vmmap.get()`

pwndbg/commands/libcinfo.py
```diff
@@ -0,0 +1,56 @@
+import pwndbg.commands
+import gdb
+import re
+import subprocess
+import os
+
+@pwndbg.commands.Command
```
patryk4815, line 13: this command is intended to work only with local debugging, not remote one?
so `@pwndbg.commands.OnlyWhenLocal` is needed

pwndbg/commands/libcinfo.py
```diff
@@ -0,0 +1,56 @@
...
+            
+            print(f"[+] libc found at: {libc_path}")
+
+            # Resolve the real path in case it's a symlink
+            real_libc_path = os.path.realpath(libc_path)
+            real_libc_path = real_libc_path.replace('\x1b[0m', '')
+            print(f"[+] Resolved libc to: {real_libc_path}")
+            
+            try:
+                result = subprocess.run(["/usr/bin/strings", real_libc_path], capture_output=True, text=True, check=True)
```
patryk4815: I'm not a fan of calling subprocess. 
@disconnect3d do you have better idea? Maybe we can use `pwndbg.search.search` and search only in `.TEXT` section?

CptGibbon:
Welcome Bryan 👋
Thanks for this PR, I see patryk has already commented.
Hopefully we can get this merged soon since it's a standalone command that doesn't affect anything else.

pwndbg/commands/libcinfo.py
k4lizen: pwndbg also has a `strings` command, though maybe there is some benefit in search the file instead of the loaded process?

pwndbg/commands/libcinfo.py
Bryan-atu: added @pwndbg.commands.OnlyWhenLocal

pwndbg/commands/libcinfo.py
Bryan-atu: updated to use pwndbg.aglib.vmmap.get()

pwndbg/commands/libcinfo.py
Bryan-atu: updated to use better string matching:
`if "libc.so" in mapping.objfile:`

pwndbg/commands/libcinfo.py
Bryan-atu: removed reliance on subprocess

pwndbg/commands/libcinfo.py
```diff
@@ -0,0 +1,67 @@
...
+
+        try:
+            mappings = pwndbg.aglib.vmmap.get()
+            libc_path = None
+            real_libc_path = None
+
+            for mapping in mappings:
+                if "libc.so" in mapping.objfile:
+                    libc_path = mapping.objfile
+                    break
```
disconnect3d: This can be just:
```py
libc_path = next((m.objfile for m in mappings if "libc.so" in m.objfile), None)
```

patryk4815:
@disconnect3d In https://github.com/pwndbg/pwndbg/issues/2775 you asked for:
> It would be nice to have a command that would figure out the libc version used by the binary (during remote debugging) and that would give us a link to its sources if possible?

I mean **(during remote debugging)**, but this PR only adds this for local debugging

disconnect3d:
I guess for remote debug we could search memory for the libc pattern version and work off that?

Btw pls fix lints

disconnect3d:
Ping @Bryan-atu

disconnect3d:
I've made some modifications but this is still meh. We probably don't want to read the libc file, we can just read the version string from memory and we can probably use our strings functionality for that?

disconnect3d:
Closing this. I will reopen another PR with a better implementation of this.

disconnect3d:
Here is a better PR for that command: https://github.com/pwndbg/pwndbg/pull/2842. It uses the `pwndbg.glibc.get_version()` to obtain the glibc version and it adds a testcase for the newly added command.

PR #2785, opened 2025-03-20
Add programatic controls to the LLDB Pwndbg CLI

mbrla0:
This PR makes it so that the execution flow of the LLDB Pwndbg CLI can be controlled by the caller of `pwndbg.dbg.lldb.repl.run`, much in the same way `ExecutionController` allows the execution of an inferior to be controlled, but at the level of the Pwndbg CLI itself.

It adds a new class `PwndbgController`, in `pwndbg.dbg.lldb.repl.run`, which exposes three functions, one that prompts a single command interactively from the user, one that runs a command automatically to completion, and one that does the same as the former, but also captures its output. The last two are akin to GDB's `gdb.execute` function.

This PR is primarily intended to be a precursor to adding tests to LLDB Pwndbg, and so it exists to provide a means for it to automatically control the execution of commands, much like the GDB Pwndbg tests do. The way it is designed, however, does leave open the possibility of, maybe down the line, considering allowing LLDB Pwndbg to be used as a library by external Python code.

### DIFF
--- a/pwndbg-lldb.py
+++ b/pwndbg-lldb.py
@@ -112,7 +112,7 @@ if __name__ == "__main__":
         print("[-] Launcher: Entering Pwndbg CLI")
 
     def drive(startup: List[str] | None):
-        async def drive(pwndbg: PwndbgController):
+        async def drive(c: PwndbgController):
             if startup is not None:
                 for line in startup:
                     await pwndbg.execute_and_capture(line)
--- a/pwndbg-lldb.py
+++ b/pwndbg-lldb.py
@@ -115,7 +115,7 @@ if __name__ == "__main__":
         async def drive(c: PwndbgController):
             if startup is not None:
                 for line in startup:
-                    await pwndbg.execute_and_capture(line)
+                    await pwndbg.execute(line)
 
             while True:
                 await pwndbg.interactive()
--- a/pwndbg-lldb.py
+++ b/pwndbg-lldb.py
@@ -105,12 +105,24 @@ if __name__ == "__main__":
     if len(sys.argv) == 2:
         target = sys.argv[1]
 
+    from pwndbg.dbg.lldb.repl import PwndbgController
     from pwndbg.dbg.lldb.repl import run as run_repl
 
     if debug:
         print("[-] Launcher: Entering Pwndbg CLI")
 
-    run_repl([f"target create '{target}'"] if target else None, debug=debug)
+    def drive(startup: List[str] | None):
+        async def drive(pwndbg: PwndbgController):
+            if startup is not None:
+                for line in startup:
+                    await pwndbg.execute_and_capture(line)
+
+            while True:
+                await pwndbg.interactive()
+
+        return drive
+
+    run_repl(drive([f"target create '{target}'"] if target else None), debug=debug)
 
     # Dispose of our debugger and terminate LLDB.
     lldb.SBDebugger.Destroy(debugger)
--- a/pwndbg-lldb.py
+++ b/pwndbg-lldb.py
@@ -115,10 +115,10 @@ if __name__ == "__main__":
         async def drive(c: PwndbgController):
             if startup is not None:
                 for line in startup:
-                    await pwndbg.execute(line)
+                    await c.execute(line)
 
             while True:
-                await pwndbg.interactive()
+                await c.interactive()
 
         return drive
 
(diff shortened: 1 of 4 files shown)


pwndbg/dbg/lldb/repl/__init__.py
```diff
@@ -205,222 +270,296 @@ def handle_sigint(_sig, _frame):
...
-                continue
+            if not action._prompt_silent:
+                print(f"{PROMPT}{action._command}")
 
-            print(
-                message.error(
-                    f"The '{name}' command is not supported. Use the 'lldb' command to enter LLDB mode and try again."
-                )
-            )
+            output = BytesIO()
```
patryk4815: can we just do?
```suggestion
            output = sys.stdout.buffer if action._capture else BytesIO()
```

patryk4815: Or we can implement `action._capture_file` / `output_file` ? and default it should be `sys.stdout.buffer`?

mbrla0: MyPy yells at us if we do, since apparently `sys.stdout.buffer` doesn't inherit from `BufferedIOBase` 🥲

Of course, we can also just ignore typing, entirely, and go with `Any` in this case. I'd be perfectly fine with that. I just chose to do it this way cuz I figured in most cases it wouldn't really make much of a difference either way.

pwndbg-lldb.py
```diff
@@ -105,12 +105,24 @@ def find_lldb_python_path() -> str:
...
 
+    from pwndbg.dbg.lldb.repl import PwndbgController
     from pwndbg.dbg.lldb.repl import run as run_repl
 
     if debug:
         print("[-] Launcher: Entering Pwndbg CLI")
 
-    run_repl([f"target create '{target}'"] if target else None, debug=debug)
+    def drive(startup: List[str] | None):
+        async def drive(pwndbg: PwndbgController):
```
patryk4815:
```suggestion
        async def drive(c: PwndbgController):
```
let's not conflict with `pwndbg` import name 🤔

mbrla0: Fair enough

pwndbg/dbg/lldb/repl/__init__.py
patryk4815: > MyPy yells at us if we do

can we just do `Union[BufferedIOBase, stdout_buffer_type]` ?

patryk4815: 🤔 And I'm think about this new `typing.Protocol` just for `write` method

mbrla0: If I remember correctly, `stdout_buffer_type` is `Any`

patryk4815: @mbrla0 `typing.BinaryIO` works fine with both `sys.stdout.buffer` and io.BytesIO

mbrla0: Well, fair enough, then

mbrla0: Should be better now

patryk4815: 👍

pwndbg-lldb.py
```diff
@@ -105,12 +105,24 @@ def find_lldb_python_path() -> str:
...
+    def drive(startup: List[str] | None):
+        async def drive(c: PwndbgController):
+            if startup is not None:
+                for line in startup:
+                    await pwndbg.execute_and_capture(line)
+
+            while True:
+                await pwndbg.interactive()
+
+        return drive
```
patryk4815:
```suggestion
    def drive(startup: List[str] | None):
        async def drive(c: PwndbgController):
            if startup is not None:
                for line in startup:
                    await c.execute(line)

            while True:
                await c.interactive()

        return drive
```
just to fix other `pwndbg` names, that previously I missed

patryk4815:
From my perspective it looks good 👍

pwndbg-lldb.py
mbrla0: Ended up making this outdated, so I fixed them manually.

disconnect3d:
@patryk4815 merge?

disconnect3d (approved):

PR #2795, opened 2025-03-25
GDB: Allow partial read of memory that wraps around address space boundary

### LINKED ISSUE #2792: pwndbg will sometimes call gdb.selected_inferior().read_memory with a range that wraps around
### Description
Set #2791 for how I was able to force pwndbg to see this gdb bug

Some cases get pwndbg to try to read from an address that 'wraps around' memory, which apparently GDB does not like.

### Steps to reproduce
On a system where apparently reading at 0 and 0xffffffff works (ARM cortex-m0 in my case)
```
pwndbg> pi gdb.selected_inferior().read_memory(0,1)
<memory at 0x113d75d80>
pwndbg> pi gdb.selected_inferior().read_memory(0xffffffff,1)
<memory at 0x113d75f00>
pwndbg> pi gdb.selected_inferior().read_memory(0xffffffff,2)
Python Exception <class 'gdb.MemoryError'>: Cannot access memory at address 0x0
```

### My setup
```
pwndbg> version
Pwndbg:   2025.02.19
Python:   3.12.8 (main, Dec  3 2024, 18:42:41) [Clang 19.1.5 ]
GDB:      15.2
Capstone: 5.0.5
Unicorn:  2.1.2
Pwnlib:   4.14.0```

### PR DESCRIPTION
**Pull request by OBarronCS**
This fixes #2792, which is a crash under a very specific scenario when reading memory.

The crash occurs when the very top-page in memory is memory mapped (in 32-bit, this would be `0xFFFF_F000` to `0xFFFF_FFFF`) and the memory at the first page in memory, address starting at `0`, is not mapped. Then, we attempt to do a memory read that crosses the boundary back to 0, such as a 2-byte read from `0xFFFF_FFFF`.

The GDB memory read handler missed this edge case (when partial=True was passed to the function). Now, if this scenario occurs, we will do a partial read until the max address (0xFFFF_FFFF).

### DIFF
--- a/pwndbg/dbg/gdb/__init__.py
+++ b/pwndbg/dbg/gdb/__init__.py
@@ -451,6 +451,14 @@ class GDBProcess(pwndbg.dbg_mod.Process):
             else:
                 stop_addr = int(message.split()[-1], 0)
 
+            # Handle case of memory read that wraps around the memory space back to 0, where high memory was readable but memory at 0 was not.
+            # Example: 2-byte read at 0xFFFF_FFFF in a 32-bit address space.
+            # GDB returns error: "Cannot access memory at address 0x0"
+            if stop_addr == 0 and stop_addr < addr:
+                # We could read from the top-portion of memory, but not after wrapping around
+                # Because we are doing a partial read, read until the max address
+                return self.read_memory(addr, pwndbg.aglib.arch.ptrmask - addr)
+
             if stop_addr != addr:
                 return self.read_memory(addr, stop_addr - addr)
 
--- a/pwndbg/dbg/gdb/__init__.py
+++ b/pwndbg/dbg/gdb/__init__.py
@@ -457,7 +457,7 @@ class GDBProcess(pwndbg.dbg_mod.Process):
             if stop_addr == 0 and stop_addr < addr:
                 # We could read from the top-portion of memory, but not after wrapping around
                 # Because we are doing a partial read, read until the max address
-                return self.read_memory(addr, pwndbg.aglib.arch.ptrmask - addr)
+                return self.read_memory(addr, pwndbg.aglib.arch.ptrmask - addr + 1)
 
             if stop_addr != addr:
                 return self.read_memory(addr, stop_addr - addr)

### CODE REVIEW

File: pwndbg/dbg/gdb/__init__.py [hunk 1]
```diff
@@ -451,6 +451,14 @@ def read_memory(self, address: int, size: int, partial: bool = False) -> bytearr
             else:
                 stop_addr = int(message.split()[-1], 0)
```
OBarronCS, line 452: The bug starts here - GDB returns error: "Cannot access memory at address 0x0", and stop_addr is set to 0. Previously, the code below didn't account for the chance that stop_addr would wrap around to below `addr`

disconnect3d:
I guess we can't add a testcase for this easily since high addresses cannot be mapped in userspace?

OBarronCS:
I made some attempts to create a test:

Testing this with the following setup with `qemu-i386`

```sh
sudo sysctl -w vm.mmap_min_addr=0
gcc -m32 read.c -o read
qemu-i386 -g 1234 ./read
```

read.c file
```c
#define _GNU_SOURCE
#include <stdio.h>
#include <stdlib.h>
#include <sys/mman.h>
#include <fcntl.h>
#include <unistd.h>
#include <stdint.h>
#include <errno.h>

#define PAGE_SIZE 4096
// #define MAX_ADDR ((void*)(0xFFFFFFFF - PAGE_SIZE + 1))
#define MAX_ADDR ((void*)(0xFFFFF000))

// First, allow zero-address to be mapped with
// sudo sysctl -w vm.mmap_min_addr=0
int main() {
    void* zero_addr = NULL;

    void *zero_page = mmap(zero_addr, 1, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_FIXED | MAP_ANONYMOUS, -1, 0);
    if (zero_page == MAP_FAILED) {
        perror("mmap 0");
        printf("%d\n",errno);
        return 1;
    }

    printf("Zero-page pointer - %x\n", zero_page);
    *((uint32_t*)zero_page) = -1;
    printf("Read from 0: 0x%X\n", *((uint32_t*)zero_page));


    void *max_page = mmap(MAX_ADDR, 1, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_FIXED | MAP_ANONYMOUS, -1, 0);
    if (max_page == MAP_FAILED) {
        perror("mmap max");
        printf("%d\n",errno);
        return 1;
    }

    printf("Max page pointer - %p\n", max_page);

    void* read_ptr = ((char *)max_page) + 0xfff;
    printf("Read pointer - %x\n", read_ptr);

    printf("One byte read: 0x%X\n", *((char*)read_ptr));
    printf("4 byte read: 0x%X\n", *((uint32_t*)read_ptr));

    return 0;
}
```

Some interesting findings:

Memory reads that read across the boundary appears to segfaults with qemu-user (even with first and last page mapped). Creating a test to check the behavior on a real CPU is a bit more difficult (due to https://stackoverflow.com/questions/47712502/why-cant-i-mmapmap-fixed-the-highest-virtual-page-in-a-32-bit-linux-process-o, this can't be done on x86_64 cpu with default linux setup).

In this case, the memory address returned in the error is the `addr` that is read
```
pwndbg> pi gdb.selected_inferior().read_memory(0xffff_ffff,2)
gdb.MemoryError: Cannot access memory at address 0xffffffff
Error occurred in Python: Cannot access memory at address 0xffffffff
```

However, I have yet to validate this behavior with qemu-system.

OBarronCS:
~~tl;dr - I think we actually can test our Pwndbg functionality with `qemu-i386`.~~ Partial reads would no longer crash, and return everything from addr to MAX_ADDRESS. EDIT: testing this behavior with qemu seems to not work, see comment below.

What is still interesting is figuring out the real behavior of CPU's - will they segfault on these reads in the case that the page at address=0 is mapped? Qemu-user does crash in this case.

OBarronCS:
While creating a test, I made a new realization on this bug - it doesn't apply to qemu due to the nature of the memory reads errors that are returned by qemu.

There's a comment in the code on the current behavior of qemu: 
image

This means the error for a cross-boundary read at `0xffffffff` read returns `Error occurred in Python: Cannot access memory at address 0xffffffff`. The code that follows will naturally actually do the partial read correctly.

However, without qemu, the error would be `Error occurred in Python: Cannot access memory at address 0x0`.

In other words, creating this test would be difficult indeed as it would require mapping the high address, which as you said is difficult in normal conditions on linux.

disconnect3d:
Thanks!

PR #2803, opened 2025-03-27
feat(commands/hexdump): Add configurable size limit

### LINKED ISSUE #2608: hexdump should not crash gdb due to OOM when too huge count is passed

Gaurav23V:
This PR prevents GDB from crashing due to OOM errors when `hexdump` is called with an excessively large count.

It introduces a check before attempting the memory read, comparing the requested count against a new configurable limit. If the count exceeds the limit, a `ValueError` is raised with an informative message guiding the user on how to adjust the limit.

- Adds a new configuration parameter: `hexdump-limit-mb`
  - Default value: 10 (MB)
  - Setting to 0 disables the limit check.
- Includes tests to verify the limit check, error message, and configurability.

Fixes #2608

### DIFF
--- a/pwndbg/commands/hexdump.py
+++ b/pwndbg/commands/hexdump.py
@@ -8,6 +8,7 @@ import pwndbg.aglib.memory
 import pwndbg.aglib.regs
 import pwndbg.commands
 import pwndbg.hexdump
+import pwndbg.exception
 from pwndbg.color import message
 from pwndbg.commands import CommandCategory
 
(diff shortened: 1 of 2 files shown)


pwndbg/commands/hexdump.py
```diff
@@ -25,6 +25,16 @@
...
     "See also hexdump-highlight-group-lsb.",
 )
+pwndbg.config.add_param(
+    "hexdump-limit-mb",
+    10,
+    "the maximum size in megabytes (MB) `hexdump` will read",
+    help_docstring="""Set the maximum size in megabytes (MB) that the `hexdump` command will attempt to read at once.
+    Prevents GDB crashes due to excessive memory allocation requests.
+    Set to 0 for unlimited (use with caution).""",
+    param_class=3,
```
disconnect3d: What is param_class?

Gaurav23V: The value 3 corresponds to PARAM_ZUINTEGER, which means "Unsigned integer, allows 0". This is the correct type for hexdump-limit-mb because the limit should be a non-negative number of megabytes, and we specifically want to allow 0 as a valid value to signify "unlimited".

I initially encountered a mypy type-checking error when trying to import and use the PARAM_ZUINTEGER constant directly (it was incorrectly inferred as the wrong type). As a temporary workaround during debugging, I used the literal value 3. I mistakenly forgot to revert this to the proper method, which involves importing the constant correctly, before submitting the PR.

Gaurav23V:
Updated the code to use the correct import from pwndbg.lib.config import PARAM_ZUINTEGER and the PARAM_ZUINTEGER constant for param_class. 
If there is something else which is not appropriate please let me know.

disconnect3d:
Thanks!

PR #2804, opened 2025-03-27
Add special handling for `version` in LLDB Pwndbg

### LINKED ISSUE #2800: [LLDB] error: cannot add command: can't replace builtin command
pwndbg version: pwndbg-lldb_2025.02.19_x86_64-portable

When lunch ./pwndbg/bin/pwndbg-lldb, there have an error:
`error: cannot add command: can't replace builtin command`

I add a print in `pwndbg-lldb_2025.02.19_x86_64-portable:1808 the function add_command` ,
Found the error becasue command `version`,
hope fix it.

mbrla0:
Fixes: #2800

`version` is another Pwndbg command that has to alias an LLDB command. This PR adds it to the LLDB Pwndbg CLI, in the same way we handle `set`. This PR also introduces a name component to `pwndbg.dbg.Debugger`, as a means to distinguish between debuggers that's more specific than `is_gdblib_available`, and uses it to gate off installing `version` as a regular command when running under LLDB.

Currently, doing these overrides manually for Pwndbg commands is probably fine, because there's all in all like two of them. But if there end up being more commands like this, we should find a better, more structured way to do it.

### DIFF
--- a/pwndbg/commands/version.py
+++ b/pwndbg/commands/version.py
@@ -100,7 +100,7 @@ def version_impl() -> None:
     print("\n".join(map(message.system, all_versions())))
 
 
-if not pwndbg.dbg.name() == "lldb":
+if pwndbg.dbg.name() != "lldb":
     # In LLDB, this command is implemented as part of the Pwndbg CLI.
     @pwndbg.commands.ArgparsedCommand(
         "Displays Pwndbg and its important deps versions.", category=CommandCategory.PWNDBG
--- a/pwndbg/commands/version.py
+++ b/pwndbg/commands/version.py
@@ -93,11 +93,20 @@ def get_terminal_size():
     return f"Terminal width: {width_info}, height: {height_info}\n"
 
 
-@pwndbg.commands.ArgparsedCommand(
-    "Displays Pwndbg and its important deps versions.", category=CommandCategory.PWNDBG
-)
-def version() -> None:
+def version_impl() -> None:
+    """
+    Implementation of the `version` command.
+    """
     print("\n".join(map(message.system, all_versions())))
+
+
+if not pwndbg.dbg.name() == "lldb":
+    # In LLDB, this command is implemented as part of the Pwndbg CLI.
+    @pwndbg.commands.ArgparsedCommand(
+        "Displays Pwndbg and its important deps versions.", category=CommandCategory.PWNDBG
+    )
+    def version() -> None:
+        version_impl()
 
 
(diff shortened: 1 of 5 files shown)


pwndbg/commands/version.py
```diff
@@ -93,13 +93,22 @@ def get_terminal_size():
...
-)
-def version() -> None:
+def version_impl() -> None:
+    """
+    Implementation of the `version` command.
+    """
     print("\n".join(map(message.system, all_versions())))
 
 
+if not pwndbg.dbg.name() == "lldb":
```
patryk4815:
```suggestion
if pwndbg.dbg.name() != "lldb":
```
why not just? anyway lgtm I think

mbrla0: That was a total brain fart wow

pwndbg/commands/version.py
```diff
@@ -93,13 +93,22 @@ def get_terminal_size():
...
-)
-def version() -> None:
+def version_impl() -> None:
+    """
+    Implementation of the `version` command.
+    """
     print("\n".join(map(message.system, all_versions())))
 
 
+if pwndbg.dbg.name() != "lldb":
```
disconnect3d, line 103: Its ugly but it works :<

pwndbg/dbg/__init__.py
```diff
@@ -1134,6 +1134,12 @@ def supports_breakpoint_creation_during_stop_handler(self) -> bool:
     # removed or replaced as the porting work continues.
     #
 
+    def name(self) -> Literal["lldb", "gdb"]:
+        """
+        The name of the current debugger.
+        """
+        raise NotImplementedError()
```
disconnect3d, line 1141: Wait, dont we have some way to distinguish gdb/lldb already across the codebase?

cc: @patryk4815

patryk4815: we can detect only with `pwndbg.dbg.is_gdblib_available()`, I recommended adding `name()` in some other PR 🤔

disconnect3d: And comparing strings ;<

That should be an enum or constant :p

disconnect3d: ok, actually, strings are rather fine , we use them in many places.

But as @patryk4815 pointed out to me in some DMs this should not be an 'if'. We can add this to `ArgparsedCommand` as `exclude_debuggers=["lldb"]` kwarg?

Issue #2794, opened 2025-03-25
[LLDB] Some commands cannot be canceled


patryk4815:
Eg. it is not possible to cancel `stepsyscall` with ctrl+c

<img width="696" alt="Image" src="https://github.com/user-attachments/assets/d1e1ec5d-aa8c-4a50-9b6c-160340965aea" />

disconnect3d:
@mbrla0 can you provide any input here? do u know what's the cause? is it fixable? I guess it should be, tho it probably depends on how do we perform a "step"? :D

mbrla0:
> @mbrla0 can you provide any input here? do u know what's the cause? is it fixable? I guess it should be, tho it probably depends on how do we perform a "step"? :D

Oh, yeah, it's absolutely fixable! We could just have ^C raise a CancelledError during evaluation of the continues and steps in the execution controller. That's already how we handle unexpected breakpoints, so it's just a matter of adding an extra way to cancel controllers.

mbrla0:
@patryk4815 try it with #2805

disconnect3d:
#2805 merged. This still needs another fix - with removal of the BPs.

disconnect3d:
@mbrla0 Are you going to work on removing the breakpoints?

mbrla0:
> [@mbrla0](https://github.com/mbrla0) Are you going to work on removing the breakpoints?

Sure thing!

mbrla0:
@patryk4815 try https://github.com/pwndbg/pwndbg/pull/2820, it should fix us leaving breakpoints behind ;p

disconnect3d:
I think this is fixed now. Please check @patryk4815 an reopen if not.

PR #2805, opened 2025-03-27
Add Ctrl+C cancellation support to `ProcessController.run_coroutine`

mbrla0:
This PR is intended to add full cancellation support for `ExecutionController`-based commands.

Currently, Pwndbg only supports cancellation of those commands through LLDB's `SBProcess::SendAsyncInterrupt`, and, while that is usually fine if the execution controller mostly uses `ExecutionController.continue`, #2794 has shown that it is not sufficient for commands such as `stepsyscall`, which use `ExecutionController.single_step`, instead.

This PR adds a mechanism to `ProcessController` that allows it to be notified about external cancellation requests, and code that polls this mechanism during the `run_coroutine` loop, as well as code in the Pwndbg CLI that triggers it on Ctrl+C.

It is important to note that this PR chooses to not change the fact that `ProcessController.run_coroutine` allows execution controllers to override cancellation requests. If an execution controller chooses to catch `CancelledError`, `run_coroutine` will still not stop, even if the cancellation source is the user. This is a choice that is easy to change, however, if it's more desirable that user-requested cancellations be mandatory.

### DIFF
--- a/pwndbg/dbg/lldb/repl/__init__.py
+++ b/pwndbg/dbg/lldb/repl/__init__.py
@@ -261,6 +261,7 @@ def run(
     # Set ourselves up to respond to SIGINT by interrupting the process if it is
     # running, and doing nothing otherwise.
     def handle_sigint(_sig, _frame):
+        driver.cancel()
         if driver.has_process():
             driver.interrupt()
             print()
--- a/pwndbg/dbg/lldb/repl/proc.py
+++ b/pwndbg/dbg/lldb/repl/proc.py
@@ -71,6 +71,7 @@ class ProcessDriver:
     listener: lldb.SBListener
     debug: bool
     eh: EventHandler
+    cancellation_requested: bool
 
     def __init__(self, event_handler: EventHandler, debug=False):
         self.io = None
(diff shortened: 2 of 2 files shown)


patryk4815:
`stepsyscall` leave a lot of breakpoints, so after `ctrl+c` I need remove all breakpoints
(I think we can resolve this breakpoint cleanup in separate PR)

image

mbrla0:
> `stepsyscall` leave a lot of breakpoints, so after `ctrl+c` I need remove all breakpoints (I think we can resolve this breakpoint cleanup in separate PR)
> image

Yeah, that looks very much like a bug

pwndbg/dbg/lldb/repl/__init__.py
```diff
@@ -261,6 +261,7 @@ def run(
     # Set ourselves up to respond to SIGINT by interrupting the process if it is
     # running, and doing nothing otherwise.
     def handle_sigint(_sig, _frame):
+        driver.cancel()
```
disconnect3d, line 264: Can we add a comment here what this does?

disconnect3d:
Pls address the bps in another PR

PR #2807, opened 2025-03-28
Express context saved to the stack + Arm Cortex M Exception return address resolution

OBarronCS:
A common pattern in context switching is for registers to be pushed to the stack so that they are saved and can later be restored when switching back to the original context. This occurs commonly in syscalls and on exceptions.

This PR adds some code allowing us to express and examine these stack frames that get pushed to the stack, and uses the definition of a Arm-M exception stack frame to allow us to correctly predict the return address of an exception handler.

When an exception occurs in ARM-M, a stack frame with some select register is pushed to the stack and the instruction pointer jumps to the exception handler. The stack frame is defined here: The stack frame that has been saved is defined here: https://developer.arm.com/documentation/107706/0100/Exceptions-and-interrupts-overview/Stack-frames

Returning from an exception handler occurs in a special way. The "return address" contains a special value, called `EXC_RETURN` (typically the value 0xFFFF_FFFF9). Upon branching to this special value, the processors knows to restore some registers that we previously saved to the stack. https://developer.arm.com/documentation/107706/0100/Exceptions-and-interrupts-overview/EXC-RETURN

There is now some code that checks for this condition, allowing us to disassemble along the correct path when returning from an Arm-M exception.


pwndbg/commands/saved_context.py
```diff
@@ -0,0 +1,80 @@
...
+            desc = pwndbg.chain.format(value)
+
+            print_value(f"{regname} {desc}", address + stack_offset, print_address)
+
+
+def create_saved_context_printer(
+    command_name: str, desc: str, context: SavedContext, arches: list[str]
+):
+    parser = argparse.ArgumentParser(description=desc)
+
```
OBarronCS: The idea here is that we might want commands to print the context for various different frame types, and this automates the creation of these commands.

If preferred, another way we could do this is instead of having a separate command for each frame type, to have a single command and a switch/argument that selects the frame type.

pwndbg/commands/saved_context.py
```diff
@@ -0,0 +1,80 @@
...
+    @pwndbg.commands.ArgparsedCommand(parser, command_name=command_name)
+    @pwndbg.commands.OnlyWhenRunning
+    @pwndbg.aglib.proc.OnlyWithArch(arches)
+    def saved_context(address: int = None, print_address=False) -> None:
+        print_saved_context(context, address, print_address)
+
+
+create_saved_context_printer(
+    "saved_arm_exception_context",
```
OBarronCS: If anyone can think of a better name, let's change it. This was the best I could think of currently.

pwndbg/aglib/saved_context.py
```diff
@@ -0,0 +1,54 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+import pwndbg.aglib.memory
+
+
+@dataclass
+class SavedContext:
```
disconnect3d: Can we add docstring what's that for?

disconnect3d:
Hmm this also adds a new command, but I am not sure I understand it -- is this all only about ARM exception handling stuff?

If so, maybe a different command name, sth related to arm? I am not sure but generally speaking I think we would like to have some commands to save regs/stack memory/other memory in some "stash" and be able to revert the current state to the saved one. The "save-context" command or sth could be a nice name for this, but maybe there would be a better/different one.

pwndbg/aglib/saved_context.py
OBarronCS: Done - I also renamed it "SavedRegisterFrame".

OBarronCS:
I renamed the class to `SavedRegisterFrame` to be more descriptive. I changed the command to be `dump-register-frame` (I'm open to other names as well), which takes as argument a string representing the "type" of register frame (future types, for example, could be the register saved on an `x86 pusha instruction`, or the register frame saved on the kernel stack upon syscall entry)

image

pwndbg/aglib/saved_register_frames.py
```diff
@@ -0,0 +1,55 @@
...
+    """
+
+    # List of (offset, register name), sorted from smallest to largest offset
+    frame_layout: list[tuple[int, str]]
+    offsets: dict[str, int]
+
+    def __init__(self, register_offsets: dict[str, int]):
+        self.offsets = register_offsets
+
+        self.frame_layout = sorted([(y, x) for (x, y) in list(register_offsets.items())])
```
disconnect3d:
```suggestion
        self.frame_layout = sorted(((y, x) for (x, y) in register_offsets.items()))
```

pwndbg/aglib/saved_register_frames.py
```diff
@@ -0,0 +1,55 @@
...
+    def read_saved_register(self, reg: str, sp: int = None) -> int | None:
+        if sp is None:
+            sp = pwndbg.aglib.regs.sp
+
+        ptr_size = pwndbg.aglib.arch.ptrsize
+
+        address = sp + self.offsets[reg]
+
+        try:
+            mem = pwndbg.aglib.memory.read(address, ptr_size)
```
disconnect3d:
```suggestion
            mem = pwndbg.aglib.memory.read(address, pwndbg.aglib.arch.ptrsize)
```

The `ptr_size` line above can be removed too

disconnect3d: We can inline the `address` here too?

pwndbg/aglib/saved_register_frames.py
```diff
@@ -0,0 +1,55 @@
...
+        except pwndbg.dbg_mod.Error:
+            return None
+
+        value = pwndbg.aglib.arch.unpack(mem)
```
disconnect3d: Just return pwndbg...unpack?

pwndbg/aglib/saved_register_frames.py
```diff
@@ -0,0 +1,55 @@
...
+
+        return value
+
+
+ARM_CORTEX_M_EXCEPTION_STACK_FRAME_OFFSETS = {
```
disconnect3d, line 38: Can we link to some documentation how did we get this from?

pwndbg/commands/saved_register_frames.py
```diff
@@ -0,0 +1,80 @@
...
+
+        if reg in pwndbg.aglib.regs.flags:  # eflags or cpsr
+            reg_flags = pwndbg.aglib.regs.flags[reg]
+            desc = C.format_flags(value, reg_flags)
+
+            print_value(f"{regname} {desc}", address + stack_offset, print_address)
+        else:
+            desc = pwndbg.chain.format(value)
+
+            print_value(f"{regname} {desc}", address + stack_offset, print_address)
```
disconnect3d: This print can be outside of `if/else` branch and done just once

pwndbg/commands/saved_register_frames.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+VALID_FRAME_TYPES = {"armcm-exception": ARM_CORTEX_M_EXCEPTION_STACK}
+
+parser = argparse.ArgumentParser(
+    description="Display the registers saved to memory for a certain frame type"
+)
+
+parser.add_argument(
+    "frame_type", choices=list(VALID_FRAME_TYPES), type=str, help="The type of frame to print"
```
disconnect3d:
```suggestion
    "frame_type", choices=tuple(VALID_FRAME_TYPES), type=str, help="The type of frame to print"
```

pwndbg/commands/saved_register_frames.py
```diff
@@ -0,0 +1,80 @@
...
+    action="store_true",
+    default=False,
+    help="Show addresses of frame values",
+)
+
+
+@pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.MEMORY)
+@pwndbg.commands.OnlyWhenRunning
+def dump_register_frame(frame_type: str, address: int = None, print_address=False) -> None:
+    register_frame = VALID_FRAME_TYPES[frame_type]
```
disconnect3d: Nitpick: maybe:
```py
register_frame = VALID_FRAME_TYPES.get(frame_type)
if register_frame is None:
    print(f"Invalid frame type: {frame_type} (valid: {','.join(VALID_FRAME_TYPES.keys()))")
    return
```

disconnect3d (changes requested):
See comments

pwndbg/aglib/saved_register_frames.py
OBarronCS: Done.

pwndbg/aglib/saved_register_frames.py
OBarronCS: I added a link to the arm docs.

pwndbg/commands/saved_register_frames.py
OBarronCS: Added this.

disconnect3d (approved):

PR #2820, opened 2025-03-31
Use context management for temporary breakpoints

mbrla0:
This PR implements the context manager protocol for `StopPoint`, allowing breakpoints and watchpoints to be automatically removed at the end of a `when` block, and replaces instances of:

```python
await ec.cont(bp)
bp.remove()
```

With:

```python
with bp:
    await ec.cont(bp)
```

This makes it so that we always delete our temporary breakpoints, even in the event of a cancellation, and should keep us from leaving spurious breakpoints after commands like `stepsyscall` in LLDB.


patryk4815:
thanks

PR #2821, opened 2025-03-31
attachp: add --exact for exact command name match

### LINKED ISSUE #2816: attachp: retry until exact match
It is sometimes annoying that we match partial command line as a "last resort":
Image

It would be great to have some flag along with "--retry" to only attach to the exact matched process name so we don't have to repeat the same command mutliple times as on the screenshot.

disconnect3d:
Fixes #2816. This can now be done e.g. with `attach --retry --exact ls`

To test it, run `ls -lah /*` so ls is running long enough for GDB to attach to it

### DIFF
--- a/pwndbg/commands/attachp.py
+++ b/pwndbg/commands/attachp.py
@@ -128,6 +128,7 @@ def find_pids(target, user, exact, all):
     else:
         return pids_exact_match_cmd or pids_partial_match_cmd or pids_partial_match_args
 
+
 @pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.START)
 def attachp(target, no_truncate, retry, exact, all, user=None) -> None:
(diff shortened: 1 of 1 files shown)


pwndbg/commands/attachp.py
```diff
@@ -113,14 +119,17 @@ def find_pids(target, user, all):
...
         return pids_exact_match_cmd + pids_partial_match_cmd + pids_partial_match_args
-
-    return pids_exact_match_cmd or pids_partial_match_cmd or pids_partial_match_args
-
+    elif exact:
+        return pids_exact_match_cmd
+    elif all:
+        return pids_exact_match_cmd + pids_partial_match_cmd + pids_partial_match_args
+    else:
+        return pids_exact_match_cmd or pids_partial_match_cmd or pids_partial_match_args
```
disconnect3d, line 130:
```suggestion


```

PR #2822, opened 2025-04-01
Prefer relative paths for vmmap output

### LINKED ISSUE #2802: Strip cwd from objfile on vmmap command
The challenge binaries are often lived in sub directories of user HOME.
I organize them by this format: `/home/user/ctfs/<year>/<event>/pwn/<chall name>/`.
Sometimes the binary filenames are long so it is displayed as a long absolute path in vmmap.
I also use tmux to split the windows into two vertical panes. So the long paths are wrap-around.
Hence the output in my cases are not nice.

It would be nice if we could prefer relative paths (to the gdb's cwd) over the absolute ones.
So instead of `/home/user/ctfs/<year>/<event>/pwn/<chall name>/<binary name>`, we would just
have `<binary name>` when gdb has `/home/user/ctfs/<year>/<event>/pwn/<chall name>/` as its cwd. 

N/A

**Additional context**
I have a local patch in for git fork. The patch is less than 5 lines. It's just changed how Page `__str__()` implemented.
I'd love to upstream it if the maintainers agree. 
Image

tesuji:
Closes #2802

This also adds a new configuration `vmmap-prefer-relpaths` (True by default) to control whether to fallback to use absolute paths.

Tested locally:
```gdb
pwndbg> show vmmap-prefer-relpaths 
Show relative paths by default in vmmap is 'on'.
pwndbg> pipe vmmap  | head -n5
             Start                End Perm     Size Offset File (set vmmap_prefer_relpaths on)
          0x400000           0x401000 r--p     1000      0 binary
          0x401000           0x402000 r-xp     1000   1000 binary
          0x402000           0x403000 r--p     1000   2000 binary
pwndbg> set vmmap-prefer-relpaths off
Set show relative paths by default in vmmap to 'off'.
pwndbg> pipe vmmap  | head -n5
             Start                End Perm     Size Offset File (set vmmap_prefer_relpaths off)
          0x400000           0x401000 r--p     1000      0 /home/lzutao/ctfs/2025/swamp/pwn/buffer/binary
          0x401000           0x402000 r-xp     1000   1000 /home/lzutao/ctfs/2025/swamp/pwn/buffer/binary
          0x402000           0x403000 r--p     1000   2000 /home/lzutao/ctfs/2025/swamp/pwn/buffer/binary
```

### DIFF
--- a/pwndbg/lib/memory.py
+++ b/pwndbg/lib/memory.py
@@ -141,8 +141,11 @@ class Page:
 
     def __str__(self) -> str:
         from os.path import relpath
-        rel = relpath(self.objfile)
-        objfile = self.objfile if rel.startswith('../../') else rel
+        if pwndbg.config.vmmap_prefer_relpaths:
+            rel = relpath(self.objfile)
+            objfile = self.objfile if rel.startswith('../../') else rel
+        else:
+            objfile = self.objfile
         return f"{self.vaddr:#{2 + 2 * pwndbg.aglib.arch.ptrsize}x} {self.vaddr + self.memsz:#{2 + 2 * pwndbg.aglib.arch.ptrsize}x} {self.permstr} {self.memsz:8x} {self.offset:6x} {objfile or ''}"
 
     def __repr__(self) -> str:
--- a/pwndbg/lib/memory.py
+++ b/pwndbg/lib/memory.py
@@ -5,6 +5,7 @@ Reading, writing, and describing memory.
 from __future__ import annotations
 
 import os
+from os.path import relpath
 
 import pwndbg.aglib.arch
 
@@ -140,13 +141,14 @@ class Page:
         )
 
     def __str__(self) -> str:
-        from os.path import relpath
         if pwndbg.config.vmmap_prefer_relpaths:
             rel = relpath(self.objfile)
-            objfile = self.objfile if rel.startswith('../../') else rel
+            # Files from the /usr (for example libc) should use the absolute paths
+            objfile = self.objfile if rel.startswith("../../") else rel
         else:
             objfile = self.objfile
-        return f"{self.vaddr:#{2 + 2 * pwndbg.aglib.arch.ptrsize}x} {self.vaddr + self.memsz:#{2 + 2 * pwndbg.aglib.arch.ptrsize}x} {self.permstr} {self.memsz:8x} {self.offset:6x} {objfile or ''}"
+        width = 2 + 2 * pwndbg.aglib.arch.ptrsize
+        return f"{self.vaddr:#{width}x} {self.vaddr + self.memsz:#{width}x} {self.permstr} {self.memsz:8x} {self.offset:6x} {objfile or ''}"
 
     def __repr__(self) -> str:
         return f"{self.__class__.__name__}({self.__str__()!r})"
--- a/pwndbg/lib/memory.py
+++ b/pwndbg/lib/memory.py
@@ -140,7 +140,10 @@ class Page:
         )
 
     def __str__(self) -> str:
-        return f"{self.vaddr:#{2 + 2 * pwndbg.aglib.arch.ptrsize}x} {self.vaddr + self.memsz:#{2 + 2 * pwndbg.aglib.arch.ptrsize}x} {self.permstr} {self.memsz:8x} {self.offset:6x} {self.objfile or ''}"
+        from os.path import relpath
+        rel = relpath(self.objfile)
+        objfile = self.objfile if rel.startswith('../../') else rel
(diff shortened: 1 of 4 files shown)


disconnect3d:
Pls fix lint issues:
```diff
--- pwndbg/aglib/vmmap.py
+++ pwndbg/aglib/vmmap.py
@@ -14,6 +14,7 @@
     param_class=pwndbg.lib.config.PARAM_BOOLEAN,
 )
 
+
 @pwndbg.lib.cache.cache_until("start", "stop")
 def get() -> Tuple[pwndbg.lib.memory.Page, ...]:
     return tuple(pwndbg.dbg.selected_inferior().vmmap().ranges())

--- pwndbg/lib/memory.py
+++ pwndbg/lib/memory.py
@@ -141,9 +141,10 @@
 
     def __str__(self) -> str:
         from os.path import relpath
+
         if pwndbg.config.vmmap_prefer_relpaths:
             rel = relpath(self.objfile)
-            objfile = self.objfile if rel.startswith('../../') else rel
+            objfile = self.objfile if rel.startswith("../../") else rel
         else:
             objfile = self.objfile
         return f"{self.vaddr:#{2 + 2 * pwndbg.aglib.arch.ptrsize}x} {self.vaddr + self.memsz:#{2 + 2 * pwndbg.aglib.arch.ptrsize}x} {self.permstr} {self.memsz:8x} {self.offset:6x} {objfile or ''}"
```

pwndbg/lib/memory.py
```diff
@@ -140,7 +140,13 @@ def permstr(self) -> str:
         )
 
     def __str__(self) -> str:
-        return f"{self.vaddr:#{2 + 2 * pwndbg.aglib.arch.ptrsize}x} {self.vaddr + self.memsz:#{2 + 2 * pwndbg.aglib.arch.ptrsize}x} {self.permstr} {self.memsz:8x} {self.offset:6x} {self.objfile or ''}"
+        from os.path import relpath
```
disconnect3d: Please move the import to the top of the file

pwndbg/lib/memory.py
```diff
@@ -140,7 +140,13 @@ def permstr(self) -> str:
...
+        if pwndbg.config.vmmap_prefer_relpaths:
+            rel = relpath(self.objfile)
+            objfile = self.objfile if rel.startswith('../../') else rel
```
disconnect3d: What's the logic/reasoning here with `../../`? What are the cases for it etc?

pwndbg/lib/memory.py
```diff
@@ -140,7 +140,13 @@ def permstr(self) -> str:
...
+        else:
+            objfile = self.objfile
+        return f"{self.vaddr:#{2 + 2 * pwndbg.aglib.arch.ptrsize}x} {self.vaddr + self.memsz:#{2 + 2 * pwndbg.aglib.arch.ptrsize}x} {self.permstr} {self.memsz:8x} {self.offset:6x} {objfile or ''}"
```
disconnect3d: If we are here, since we are computing `2 + 2 * pwndbg.aglib.arch.ptrsize` twice, can we put this on a variable above and refer that variable here?

pwndbg/lib/memory.py
disconnect3d: Or/and why not just:

```py
objfile = relpath(self.objfile) if pwndbg.config.vmmap_prefer_relpaths else self.objfile
```

tesuji: Python's `relpath` will return relative paths even for system files (like /usr). So we have longer paths (with no clear benefits) over the absolute paths.

Should I document it in the code ?

disconnect3d:
```py
In [8]: os.path.relpath('/usr/lib/a/b/c/libc.so')
Out[8]: '../../../../usr/lib/a/b/c/libc.so'
```

Oh.. that's what you mean.

What about changing the implementation then to this?:
```py
prefix = os.path.commonprefix((os.getcwd(), self.objfile))
objfile = self.objfile.lstrip(prefix) if prefix != "/" else self.objfile
```

Two notes:
* `lstrip` is probably not best bcoz we may strip prefix twice in theory? probably doesn't matter
* shall we use `os.path.commonprefix` or `os.path.commonpath`? ugh...

tesuji: I like `os.path.relpath` since it is the most simple option. However it's your final call. I don't mind much.

disconnect3d: I'm ok with either solution. If we go with `relpath` then we need to document it appropriately, since it is not so obvious why we do the `"../../"` check.

disconnect3d:
Thanks!

PR #2823, opened 2025-04-01
Dev setup - avoid redownloading/rebuilding jemalloc on setup s…

yadunand-kamath:
Added the following into the `install_jemalloc()` function in `setup-dev.sh`:

- Added a check for existing jemalloc tarball and checksum
- Added a check for existing jemalloc extracted source code
- Added a check to see if jemalloc is already installed

Screenshot below shows a demo with the fix.
pwndbg_setup-dev_fix_demo

### DIFF
--- a/setup-dev.sh
+++ b/setup-dev.sh
@@ -262,7 +262,7 @@ install_jemalloc() {
         echo "Extracting jemalloc..."
         tar -C /tmp -xf "${JEMALLOC_TAR_PATH}"
     fi
- 
+
     # Check if jemalloc is already installed
     if ! command -v jemalloc-config &> /dev/null; then
         echo "Jemalloc not found in system. Configuring, building, and installing..."
--- a/setup-dev.sh
+++ b/setup-dev.sh
@@ -224,26 +224,60 @@ install_jemalloc() {
     # Install jemalloc version 5.3.0
     JEMALLOC_TAR_URL="https://github.com/jemalloc/jemalloc/releases/download/5.3.0/jemalloc-5.3.0.tar.bz2"
     JEMALLOC_TAR_SHA256="2db82d1e7119df3e71b7640219b6dfe84789bc0537983c3b7ac4f7189aecfeaa"
-    curl --location --output /tmp/jemalloc-5.3.0.tar.bz2 "${JEMALLOC_TAR_URL}"
-    ACTUAL_SHA256=$(sha256sum /tmp/jemalloc-5.3.0.tar.bz2 | cut -d' ' -f1)
-    if [ "${ACTUAL_SHA256}" != "${JEMALLOC_TAR_SHA256}" ]; then
-        echo "Jemalloc binary checksum mismatch"
-        echo "Expected: ${JEMALLOC_TAR_SHA256}"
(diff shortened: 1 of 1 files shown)


disconnect3d:
Hi,

Please fix lint issues:
```diff
--- ./setup-dev.sh.orig
+++ ./setup-dev.sh
@@ -264,7 +264,7 @@
     fi
 
     # Check if jemalloc is already installed
-    if ! command -v jemalloc-config &>/dev/null; then
+    if ! command -v jemalloc-config &> /dev/null; then
         echo "Jemalloc not found in system. Configuring, building, and installing..."
         pushd "${JEMALLOC_EXTRACT_PATH}"
         ./configure
@@ -277,7 +277,7 @@
 
     echo "Jemalloc installation complete."
 
-	 # TODO: autoconf needs to be installed with script as well?
+    # TODO: autoconf needs to be installed with script as well?
 }
```

Also, can you show this in action, what happens when u run it etc?

yadunand-kamath:
I don't know why lint keeps failing. I ran `lint.sh` locally and it ran successfully. Please advise.

pwndbg-lint

disconnect3d:
Your lint says "shfmt not installed, skipping". The lint issues come from shfmt. I showed the diff that needs to be applied in order to fix the linting issues in my previous reply.

setup-dev.sh
```diff
@@ -224,26 +224,60 @@ install_jemalloc() {
...
     JEMALLOC_TAR_URL="https://github.com/jemalloc/jemalloc/releases/download/5.3.0/jemalloc-5.3.0.tar.bz2"
     JEMALLOC_TAR_SHA256="2db82d1e7119df3e71b7640219b6dfe84789bc0537983c3b7ac4f7189aecfeaa"
-    curl --location --output /tmp/jemalloc-5.3.0.tar.bz2 "${JEMALLOC_TAR_URL}"
-    ACTUAL_SHA256=$(sha256sum /tmp/jemalloc-5.3.0.tar.bz2 | cut -d' ' -f1)
-    if [ "${ACTUAL_SHA256}" != "${JEMALLOC_TAR_SHA256}" ]; then
-        echo "Jemalloc binary checksum mismatch"
-        echo "Expected: ${JEMALLOC_TAR_SHA256}"
-        echo "Actual: ${ACTUAL_SHA256}"
-        exit 1
+    JEMALLOC_TAR_PATH="/tmp/jemalloc-5.3.0.tar.bz2"
```
OBarronCS, line 227: The `/tmp` directory gets deleted after a reboot of the device, which will result the redownload of jemalloc again. Is there a different way to check if jemalloc has already been `./configure`d, `make`d, and installed?

disconnect3d: Currently we also use /tmp/. Idk what's the best path, perhaps ./.venv/ ?

I think we can change this path etc in another PR if everything else makes sense in this one. Did u maybe check this one?

disconnect3d: I think this PR looks good. I am going to merge it. Lets see if it causes any issues etc. for the /tmp/ path if someone wants to change it - go ahead.

PR #2825, opened 2025-04-02
Fix telescope command to properly handle C++ function symbols with sp…

### LINKED ISSUE #2824: Telescope can't parse expressions with symbols that contain spaces without extra escaping
### Description

C++ function symbols contain spaces a lot of the time so in order to reference them in an expression, you can do something like
```
pwndbg> p/x &'Circle::calculateArea() const'
$2 = 0x1660
```

This unfortunately does not work the same way for the `telescope` command (and likely most other pwndbg commands that take in gdb expressions)

```
pwndbg> telescope &'Circle::calculateArea() const'
usage: telescope [-h] [-r] [-f] [-i] [address] [count]
telescope: error: argument address: debugger couldn't resolve argument '&Circle::calculateArea() const': A syntax error in expression, near the end of `&Circle::calculateArea() const'.
```

a workaround is to double quote the expression
```
pwndbg> telescope "&'Circle::calculateArea() const'"
00:0000│  0x555555555660 (Circle::calculateArea() const) ◂— movsd xmm1, qword ptr [rdi + 0x28]
01:0008│  0x555555555668 (Circle::calculateArea() const+8) ◂— add eax, 0x9b3
02:0010│  0x555555555670 (Circle::calculateArea() const+16) ◂— sal edx, 0xf
03:0018│  0x555555555678 ◂— nop dword ptr [rax + rax]
04:0020│  0x555555555680 (Circle::display() const) ◂— push r15
05:0028│  0x555555555688 (Circle::display() const+8) ◂— mov r14, qword ptr [rip + 0x2951]
06:0030│  0x555555555690 (Circle::display() const+16) ◂— lea esi, [rip + 0xa08]
07:0038│  0x555555555698 (Circle::display() const+24) ◂— add byte ptr [rax], al
```

This actually might be non-trivial to fix if we want to match gdb behavior. Since telescope can take additional arguments after the expression, the code uses `gdb.string_to_argv` to lex the arguments. Unfortunately, this removes the inner single quotes required for the expression to parse.

```python
(Pdb) p command_line
"&'Circle::calculateArea() const' 0x100"
(Pdb) p gdb.string_to_argv(command_line)
['&Circle::calculateArea() const', '0x100']
```

```python
pwndbg> p/x '&Circle::calculateArea() const'
No symbol "&Circle::calculateArea() const" in current context.
pwndbg> p/x &'Circle::calculateArea() const'
$2 = 0x1660
```
### Steps to reproduce

Compile the following C++ sample code with debug symbols.

```c++
#include <iostream>
#include <string>

// Base class
class Shape {
protected:
    std::string name;
    
public:
    // Constructor
    Shape(const std::string& shapeName) : name(shapeName) {}
    
    // Virtual method to be overridden
    virtual double calculateArea() const {
        return 0.0;
    }
    
    // Virtual method to display information
    virtual void display() const {
        std::cout << "Shape: " << name << std::endl;
    }
    
    // Virtual destructor
    virtual ~Shape() {
        std::cout << "Shape destructor called for " << name << std::endl;
    }
};

// Derived class: Circle
class Circle : public Shape {
private:
    double radius;
    
public:
    // Constructor
    Circle(const std::string& circleName, double r) 
        : Shape(circleName), radius(r) {}
    
    // Override calculateArea
    double calculateArea() const override {
        return 3.14159 * radius * radius;
    }
    
    // Override display
    void display() const override {
        std::cout << "Circle: " << name << std::endl;
        std::cout << "Radius: " << radius << std::endl;
        std::cout << "Area: " << calculateArea() << std::endl;
    }
    
    // Destructor
    ~Circle() override {
        std::cout << "Circle destructor called for " << name << std::endl;
    }
};

// Another derived class: Rectangle
class Rectangle : public Shape {
private:
    double width;
    double height;
    
public:
    // Constructor
    Rectangle(const std::string& rectName, double w, double h)
        : Shape(rectName), width(w), height(h) {}
    
    // Override calculateArea
    double calculateArea() const override {
        return width * height;
    }
    
    // Override display
    void display() const override {
        std::cout << "Rectangle: " << name << std::endl;
        std::cout << "Width: " << width << ", Height: " << height << std::endl;
        std::cout << "Area: " << calculateArea() << std::endl;
    }
    
    // Destructor
    ~Rectangle() override {
        std::cout << "Rectangle destructor called for " << name << std::endl;
    }
};

int main() {
    // Create objects dynamically
    Shape* shapes[3];
    shapes[0] = new Shape("Generic Shape");
    shapes[1] = new Circle("My Circle", 5.0);
    shapes[2] = new Rectangle("My Rectangle", 4.0, 6.0);
    
    // Using polymorphism to call overridden methods
    std::cout << "Displaying all shapes:" << std::endl;
    for (int i = 0; i < 3; i++) {
        shapes[i]->display();
        std::cout << "--------------------" << std::endl;
    }
    
    // Clean up memory
    for (int i = 0; i < 3; i++) {
        delete shapes[i];
    }
    
    return 0;
}
```

### My setup

```
pwndbg> version
Pwndbg:   2025.02.19
Python:   3.12.9 (main, Feb  4 2025, 14:38:38) [GCC 14.2.1 20241116]
GDB:      16.2
Capstone: 5.0.5
Unicorn:  2.1.3
Pwnlib:   4.14.0
pwndbg> 
```

shyam-ramani:
fixes #2824

### DIFF
--- a/pwndbg/chain.py
+++ b/pwndbg/chain.py
@@ -4,6 +4,7 @@ from typing import List
 
 import pwndbg.aglib.arch
 import pwndbg.aglib.memory
+import pwndbg.aglib.qemu
 import pwndbg.aglib.typeinfo
 import pwndbg.aglib.vmmap
 import pwndbg.color.memory as M
@@ -71,11 +72,11 @@ def get(
         try:
             address = address + offset
 
-            # Avoid redundant dereferences in bare metal mode by checking
-            # if address is in any of vmmap pages
-            if not pwndbg.dbg.selected_inferior().is_linux() and not pwndbg.aglib.vmmap.find(
-                address
-            ):
+            # Only attempt to dereference if we know the address is readable.
+            # If we have no mapping information but we're in usermode, we can try to dereference
+            # and let memory errors be caught in the exception handler.
+            # Otherwise, verify the address is mapped before proceeding.
+            if not pwndbg.aglib.vmmap.find(address) and not pwndbg.aglib.qemu.is_usermode():
                 break
 
             next_address = int(
--- a/pwndbg/commands/telescope.py
+++ b/pwndbg/commands/telescope.py
@@ -93,7 +93,11 @@ parser.add_argument(
 
 
 parser.add_argument(
-    "address", nargs="?", default="$sp", type=pwndbg.commands.AddressExpr, help="The address to telescope at."
+    "address",
+    nargs="?",
+    default="$sp",
+    type=pwndbg.commands.AddressExpr,
+    help="The address to telescope at.",
 )
 
 parser.add_argument(
--- a/pwndbg/commands/telescope.py
+++ b/pwndbg/commands/telescope.py
@@ -93,16 +93,18 @@ parser.add_argument(
 
 
 parser.add_argument(
-    "address",
-    nargs="?",
-    default="$sp",
-    type=pwndbg.commands.AddressExpr,
-    help="The address to telescope at.",
+    "address", 
+    nargs="?", 
+    default="$sp", 
+    type=pwndbg.commands.AddressExpr, 
+    help="The address to telescope at."
 )
 
 parser.add_argument(
     "count", nargs="?", default=telescope_lines, type=int, help="The number of lines to show."
 )
+
+
 
 
 @pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.MEMORY)
--- a/pwndbg/commands/telescope.py
+++ b/pwndbg/commands/telescope.py
@@ -93,7 +93,7 @@ parser.add_argument(
 
 
 parser.add_argument(
-    "address", nargs="?", default="$sp", type=int, help="The address to telescope at."
+    "address", nargs="?", default="$sp", type=pwndbg.commands.AddressExpr, help="The address to telescope at."
 )
 
 parser.add_argument(
--- a/pwndbg/commands/telescope.py
+++ b/pwndbg/commands/telescope.py
@@ -93,11 +93,11 @@ parser.add_argument(
 
 
 parser.add_argument(
-    "address", 
-    nargs="?", 
-    default="$sp", 
-    type=pwndbg.commands.AddressExpr, 
-    help="The address to telescope at."
+    "address",
+    nargs="?",
+    default="$sp",
+    type=pwndbg.commands.AddressExpr,
+    help="The address to telescope at.",
 )
 
 parser.add_argument(


Ninja3047:
this is not quite right if you want to use the "&" operator to get the address of a function with a space

```
pwndbg> p/x &'Circle::display() const'
$4 = 0x555555555680
pwndbg> telescope &'Circle::display() const'
usage: telescope [-h] [-r] [-f] [-i] [address] [count]
telescope: error: argument address: Incorrect address (or GDB expression): &Circle::display() const
```

shyam-ramani:
I think this work better.

pwndbg/chain.py
```diff
@@ -71,11 +72,11 @@ def get(
...
-            # Avoid redundant dereferences in bare metal mode by checking
-            # if address is in any of vmmap pages
-            if not pwndbg.dbg.selected_inferior().is_linux() and not pwndbg.aglib.vmmap.find(
-                address
-            ):
+            # Only attempt to dereference if we know the address is readable.
+            # If we have no mapping information but we're in usermode, we can try to dereference
+            # and let memory errors be caught in the exception handler.
+            # Otherwise, verify the address is mapped before proceeding.
+            if not pwndbg.aglib.vmmap.find(address) and not pwndbg.aglib.qemu.is_usermode():
```
disconnect3d, line 79: I don't exactly understand how this change relates to handling C++ symbols.

disconnect3d:
@Ninja3047 This being said the main change here is `int`->`AddressExpr` for parsing the argument type.

Is it that `AddressExpr` is maybe incorrect too? :/

Fwiw it should really give examples what we mean by "address expression" in its docstring...

disconnect3d:
> I think this work better.

You think, but... we still do not know. Can you show it with examples, ideally by adding tests for it?

Ninja3047:
> @Ninja3047 This being said the main change here is `int`->`AddressExpr` for parsing the argument type.
> 
> Is it that `AddressExpr` is maybe incorrect too? :/
> 
> Fwiw it should really give examples what we mean by "address expression" in its docstring...

The current approach cannot work since the quotes get stripped out by `gdb.string_to_argv` before they even reach argparse.

The idea I had was to add the quotes back using something like `shlex.quote` but that does not work either if you have anything else around it (like the & to get the address of it) since the quotes would be in the wrong spot.

This is actually quite non trivial to fix since you would likely need to do some funky hacks make it so argument lexing is gdb expression aware.

TomJGooding:
I wouldn't waste too much time on this pull request. I'm afraid this is likely LLM-generated slop based on their PRs in other projects.

disconnect3d:
@Ninja3047 Just to be sure: is it better that we keep this an 'int' type or is 'AddressExpr' a bit better?

Ninja3047:
> @Ninja3047 Just to be sure: is it better that we keep this an 'int' type or is 'AddressExpr' a bit better?

i'm not really sure tbh lol, i took a look at the pwndbg argument parsing code and it's a of a mess with lots of duplicate code (for example `AddressExpr` uses https://github.com/pwndbg/pwndbg/blob/c39f171c4d35efba73e88da97165e45f94c74e10/pwndbg/commands/__init__.py#L646-L669 under the hood, but `int` eventually funnels into https://github.com/pwndbg/pwndbg/blob/c39f171c4d35efba73e88da97165e45f94c74e10/pwndbg/commands/__init__.py#L204-L271 which is mostly the same except that `sloppy_gdb_parse` also checks if the type of the value is a function and if it is, it gets the address of it, and `fix` will try to turn the value into register?? Neither of them will resolve the original issue since the real issue here is that the quotes are stripped here https://github.com/pwndbg/pwndbg/blob/c39f171c4d35efba73e88da97165e45f94c74e10/pwndbg/dbg/gdb/__init__.py#L1455-L1456

like something that _would_ technically work is if you just slap a `shlex.quote(s)` like somewhere here https://github.com/pwndbg/pwndbg/blob/c39f171c4d35efba73e88da97165e45f94c74e10/pwndbg/commands/__init__.py#L664-L667 so that it gets requoted again before getting passed to `target.evaluate_expression(s)` which feels kind of icky but then it would mean something like
```
pwndbg> telescope 'Circle::display() const'
00:0000│  0x555555555680 (Circle::display() const) ◂— push r15
01:0008│  0x555555555688 (Circle::display() const+8) ◂— mov r14, qword ptr [rip + 0x2951]
02:0010│  0x555555555690 (Circle::display() const+16) ◂— lea esi, [rip + 0xa08]
03:0018│  0x555555555698 (Circle::display() const+24) ◂— add byte ptr [rax], al
04:0020│  0x5555555556a0 (Circle::display() const+32) ◂— stc 
05:0028│  0x5555555556a8 (Circle::display() const+40) ◂— mov edx, dword ptr [rbx + 0x10]
06:0030│  0x5555555556b0 (Circle::display() const+48) ◂— stc 
07:0038│  0x5555555556b8 (Circle::display() const+56) ◂— call 0x55555e09e20a
```
would work but it's kind of not really correct since you really want to be doing
```
pwndbg> telescope &'Circle::display() const'
usage: telescope [-h] [-r] [-f] [-i] [address] [count]
telescope: error: argument address: Incorrect address (or GDB expression): '&Circle::display() const'
```
but it'll make dealing with complicated C++ symbols slightly easier

disconnect3d:
@Ninja3047 Eh. I feel we should make some kind of "cleanup" with all of these. If you would have time to do that, I'd appreciate that a lot...

I am going to close this PR for now, since the author is not responding.

PR #2826, opened 2025-04-02
Set cyclic default size to 4

### LINKED ISSUE #2812: Is cyclic --lookup_value bugged?
Why does this take ages and never finishes?:
```
pwndbg> cyclic -l waagxaag
Finding cyclic pattern of 8 bytes: b'waagxaag' (hex: 0x7761616778616167)
```

While this works instantly?:
```
pwndbg> ipi
In [3]: cyclic(0x800).find(b'waagxaag')
Out[3]: 688
```

Arusekk:
If the value is too large to fit in a 4-byte type, a pointer type is tried and the result is cut to fit. This is consistent with Pwntools behaviour, and follows the principle of least surprise.

Ref. Gallopsled/pwntools#2573
Fixes #2812

### DIFF
--- a/pwndbg/commands/cyclic.py
+++ b/pwndbg/commands/cyclic.py
@@ -26,8 +26,9 @@ parser.add_argument(
     "-n",
     "--length",
     metavar="length",
+    default=4,
     type=int,
-    help="Size of the unique subsequences (defaults to the pointer size for the current arch)",
+    help="Size of the unique subsequences",
 )
 
 
@@ -69,7 +70,11 @@ def cyclic_cmd(alphabet, length: Optional[int], lookup, count=100, filename="") 
         lookup = pwndbg.commands.fix(lookup, sloppy=True)
 
         if isinstance(lookup, (pwndbg.dbg_mod.Value, int)):
-            lookup = int(lookup).to_bytes(length, pwndbg.aglib.arch.endian)
+            try:
+                lookup = int(lookup).to_bytes(length, pwndbg.aglib.arch.endian)
+            except OverflowError:
+                lookup = int(lookup).to_bytes(pwndbg.aglib.arch.ptrsize, pwndbg.aglib.arch.endian)
+                lookup = lookup[:length]
         elif isinstance(lookup, str):
             lookup = bytes(lookup, "utf-8")
 


k4lizen:
imho the pwntools behaviour is the unintuitive one, i wonder what others think though

Arusekk:
The reason is that often even on 64-bit archs many values are 32-bit integers, so in case some field is clobbered or only half of it is visible, the 'half-leak' is ambiguous when n=8.

k4lizen:
Right, and then you can specify `-n 4` if you want. The stack in pwndbg's UI is shown in pointer granularity (so you can double click a value -> ctrl+shift-copy to quickly select a pointer, and you would have to precisely drag to select a 4-byte value in 64-bit), and important values you could be targetting with cycle (like RBP, RIP and pointers) are pointer-sized.

idk maybe I'm biased because I remember being bitten by this behaviour before, though to be fair it was mostly about the fact that pwntools and pwndbg didn't have parity rather than the default `-n` value

disconnect3d:
@Arusekk At the same time on x64 you will likely copy  8-byte value from a register/stack/memory to do `cyclic -l <value>` and I am not sure how will that work if we set this default size to 4.

Imho it would be better to change the default in Pwntools or make it dependant on the `context.arch` pointer size 🤐.

disconnect3d:
Heh, I wrote the same thing as @k4lizen. But it would be great if you could reconsider our arguments to change the default in Pwntools instead of Pwndbg. That would make it much much more convenient IMHO.

disconnect3d:
Closing this for now. I believe we want to keep the current behavior. On x86-64 targets u want 8-byte values because that's what u will copy from registers/memory by default. The users should set `context.cyclic_size=8` in their pwntools scripts or do `cyclic(size, n=8)` when generating...

One thing we could maybe improve is to detect a cyclic sequence that may take super long to compute and if we find such, we could warn user that maybe they generated it incorrectly? Though I am not sure if that's doable.

Arusekk:
I think my suggestion actually improves the behaviour of 64-bit architectures. The extra logic is there to truncate long values preserving endianness, and outputting the offset containing the first 4 bytes of the copied value.

PR #2827, opened 2025-04-02
Refactor Docs

k4lizen:
This PR will track refactoring docs generation/build/mkdocs.yml settings.

Current changes:
+ The generated Source page looks better
image

+ The commands page looks better
image

+ The navigation is autogenerated so we don't have to remember to add a line to `mkdocs.yml` when writing docs
+ The folders are restructured so they make more sense for the docs page rather than mirror the source layout


mkdocs.yml
```diff
@@ -1,80 +1,130 @@
...
+              show_signature_annotations: true
+              show_source: false
+              show_symbol_type_heading: true
+              show_symbol_type_toc: true
+              signature_crossrefs: true
+              summary: true
+              unwrap_annotated: true
+  # Generate the whole Source page magically (hooks into mkdocstrings)
+  - api-autonav:
+      modules: ['pwndbg']
```
patryk4815, line 133: maybe because of that `--only-group` don't work

k4lizen: I can confirm that disabling api-autonav allows `--only-group` to work, but I'm not sure why. It happens even when I set `allow_inspection: false` which should prevent any module loading, so to my understanding it should just be statically analyzing the files.

disconnect3d:
(Fwiw if that makes sense, we can also merge the docs refactorings in parts)

k4lizen:
I think this is mostly it relating to what the docs look like (I'll handle other stuff in other PRs), so it should be ready for review.

Can test the PR with (it takes a bit to spin up):
```
uv run --group docs mkdocs serve
```

Feedback welcome obviously :p

disconnect3d:
Merging this. Let's see this on prod and we will give feedback based on the result on Discord, then we can have more PRs improving this further.

PR #2833, opened 2025-04-05
Check if the address is valid when using hexdump

### LINKED ISSUE #2832: Hexdump doesn't show an adress is unmapped
This seems like a bug, although it doesn't error out, huh.

```
pwndbg> hexdump 0xdeadbeef
+0000 0xdeadbeef

pwndbg> vmmap 0xdeadbeef
There are no mappings for specified address or module.
```

Pwndbg version:
```
pwndbg> version
Pwndbg:   2025.02.19 build: 97ea040b
Python:   3.11.6 (main, Apr 10 2024, 17:26:07) [GCC 13.2.0]
GDB:      14.0.50.20230907-git
Capstone: 5.0.3
Unicorn:  2.1.0
Pwnlib:   4.14.0
```

dkvhr:
I used the same approach telescope uses to check if the address is valid by peeking at it.
There is also a change at the test that checks an empty hexdump

Closes #2832

### DIFF
--- a/pwndbg/commands/hexdump.py
+++ b/pwndbg/commands/hexdump.py
@@ -77,6 +77,9 @@ def hexdump(address, count=pwndbg.config.hexdump_bytes) -> None:
         hexdump.offset = 0
 
     address = int(address)
+    if not pwndbg.aglib.memory.peek(address):
+        print("Could not read memory at specified address")
+        return
     if address > pwndbg.aglib.arch.ptrmask:
         new_address = address & pwndbg.aglib.arch.ptrmask
         print(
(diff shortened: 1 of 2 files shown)


patryk4815 (approved):

pwndbg/commands/hexdump.py
```diff
@@ -77,6 +77,9 @@ def hexdump(address, count=pwndbg.config.hexdump_bytes) -> None:
         hexdump.offset = 0
 
     address = int(address)
+    if not pwndbg.aglib.memory.peek(address):
```
disconnect3d, line 80: Shouldn't we move this after the next if condition?

disconnect3d: Pls send another PR moving this further if u don't mind

dkvhr: sure! I will send another one. Thanks for the review :)

Issue #2834, opened 2025-04-05
Not possible to disable `vmmap_prefer_relpaths` on remote debugging


patryk4815:
1. Not possible to change `set vmmap_prefer_relpaths off` on remote debugging `target remote 127.0.0.1:1234`
2. why we are adding `../` into absolute paths?
 
<img width="1213" alt="Image" src="https://github.com/user-attachments/assets/c3fb21d2-84e3-42c9-ae7f-531420ce203e" />

Related with changes in https://github.com/pwndbg/pwndbg/pull/2822

@tesuji

tesuji:
1. I cannot reproduce it locally (see attached image). Was you attaching to qemu gdbserver stub?
2. The code doesn't add/append `../`. It's python `os.path.relpath` behavior when your cwd is `/root`, which is one layer lower than the absolute file path. Perhaps we could get the shortest path instead of comparing with `../../`. I'll prepare a patch later. 

![Image](https://github.com/user-attachments/assets/a62a20ea-905d-44d1-bb02-d4b6b0a0e81f)

patryk4815:
@tesuji you are using `set vmmap-prefer-relpaths` not as hints vmmap `set vmmap_prefer_relpaths` :D 

<img width="203" alt="Image" src="https://github.com/user-attachments/assets/a983587a-b4f8-414f-86d4-3e79be4bbc3d" />

disconnect3d:
Huh, the hint is incorrect 😅

tesuji:
That's my bad :D I was under impression that gdb could use `_` interchanges with `-`. Thanks for spotting it.
I've submitted a fix on #2835 .
Edit: Btw, I don't type the config name manually but I reply on gdb's tab-completion so I don't spot it sooner.

PR #2835, opened 2025-04-06
vmmap: Prefer shortest paths in relpath config

tesuji:
cc #2834 . I'm not sure this patch fixes that issue since I haven't been able to reproduce it yet.

### DIFF
--- a/pwndbg/lib/memory.py
+++ b/pwndbg/lib/memory.py
@@ -143,9 +143,8 @@ class Page:
     def __str__(self) -> str:
         if pwndbg.config.vmmap_prefer_relpaths:
             rel = relpath(self.objfile)
-            # Prefer absolute paths for system files (for example libc) since
-            # `os.path.relpath` returns longer relative paths with no clear benefits.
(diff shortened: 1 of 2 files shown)


pwndbg/lib/memory.py
```diff
@@ -143,9 +143,8 @@ def permstr(self) -> str:
     def __str__(self) -> str:
         if pwndbg.config.vmmap_prefer_relpaths:
             rel = relpath(self.objfile)
-            # Prefer absolute paths for system files (for example libc) since
-            # `os.path.relpath` returns longer relative paths with no clear benefits.
-            objfile = self.objfile if rel.startswith("../../") else rel
+            # Keep the origin path when relative paths are longer than absolute ones.
+            objfile = self.objfile if len(rel) >= len(self.objfile) else rel
```
disconnect3d:
```suggestion
            objfile = self.objfile if len(rel) > len(self.objfile) else rel
```

disconnect3d (approved):
@patryk4815 pls confirm this works and merge it

Issue #2118, opened 2024-04-18
Command for reading/writing to MSR registers


disconnect3d:
TL;DR: I want us to have a `msr` command for reading/writing to model specific registers.

This should work by assembling and executing rdmsr/wrmsr instrucitons in the kernel and should only work/make sense in a qemu kernel context, since userspace can't rdmsr/wdmsr.

gsingh93:
Just leaving this here for reference since it'll help with the ARM implementation for this: https://github.com/androidoffsec/art-kernel-toolkit/pull/5

disconnect3d:
Fwiw bata24/GEF has sth implemented for this and I used it in the past, but I believe it can be improved.

This may also be helpful: https://gist.github.com/disconnect3d/4fa1972f5b3148bf17995406490c0e70, at least to play with MSRs (skip the exploit part, take the read/write functions or/and constants).

p4zuu:
> Fwiw bata24/GEF has sth implemented for this and I used it in the past, but I believe it can be improved.

bata24/gef seems to implement this by executing `rdmsr`/`wrmsr` instructions in the guest context.
In https://github.com/bata24/gef/blob/dev/gef.py:
```python
    def read_msr(self, const):
        codes = [b"\x0f\x32"] # rdmsr
        if is_x86_64():
            regs = {"$rcx": const}
        else:
            regs = {"$ecx": const}
        ret = ExecAsm(codes, regs=regs).exec_code()

        if ret is None:
            return None

        if is_x86_64():
            edx = ret["reg"]["$rdx"] & 0xffffffff
            eax = ret["reg"]["$rax"] & 0xffffffff
        else:
            edx = ret["reg"]["$edx"]
            eax = ret["reg"]["$eax"]
        return ((edx << 32) | eax) & 0xffffffffffffffff

    def write_msr(self, const, value):
        codes = [b"\x0f\x30"] # wrmsr
        if is_x86_64():
            regs = {"$rcx": const, "$rdx": value >> 32, "$rax": value & 0xffffffff}
        else:
            regs = {"$ecx": const, "$edx": value >> 32, "$eax": value & 0xffffffff}
        ret = ExecAsm(codes, regs=regs).exec_code()
        return bool(ret)
```

Is there already a way in pwndbg/gdb to easily execute custom code in guest to implement msr read/write (for x86 at least)?

disconnect3d:
@p4zuu Yes. This is how `mprotect` command is implemented.

[...]

PR #2836, opened 2025-04-06
Add `msr` command

p4zuu:
As suggested in #2118, this adds a `msr` command to read and write to MSR in a QEMU context only.
This also add a `--list' option to list both the name and value for commons MSRs (this list is non-exhaustive, we could add or remove some entries as we want). This PR adds MSR support for x86 and x86_64 only.

I also would like to add MSR name auto-completion when trying something like `msr -n MSR_<tab>`, but that's maybe for another PR.


pwndbg/commands/msr.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+def msr_list(arch: str):
+    msrs = COMMON_MSRS[arch]
+    for msr, value in msrs.items():
+        print(f"{msr}\t{hex(value)}")
+
+
+@pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.KERNEL, command_name="msr")
+@pwndbg.commands.OnlyWhenQemuKernel
```
disconnect3d, line 142: Nitpick: If I do bare metal debugging of linux kernel through e.g. [kgdboe](https://github.com/sysprogs/kgdboe) then I may also want to use this command.

Maybe we just need some warning and yes/no question within the command if its not qemu kernel to ask the user if they want to proceed?

Also, this can be "OnlyWithArch x86/x86-64" or something for now.

pwndbg/commands/msr.py
```diff
@@ -0,0 +1,146 @@
...
+    eax = write_value & 0xFFFFFFFF
+    edx = write_value >> 32
+    if arch == "x86-64":
+        setattr(pwndbg.aglib.regs, "rcx", msr)
+        setattr(pwndbg.aglib.regs, "rax", eax)
+        setattr(pwndbg.aglib.regs, "rdx", edx)
+    else:
+        setattr(pwndbg.aglib.regs, "ecx", msr)
+        setattr(pwndbg.aglib.regs, "eax", eax)
+        setattr(pwndbg.aglib.regs, "edx", edx)
```
disconnect3d: This is a bit ugly, if `pwndbg.aglib.regs.rax=123` does not work, then can we do `pwndbg.aglib.regs["reg"] = value`? Or is it not supported?

disconnect3d:
Can u show some example outputs in text or screenshots for this?

pwndbg/commands/msr.py
patryk4815: most of commands we have guarded with `OnlyWhenQemuKernel` :(

disconnect3d: Lets leave OnlyWhenQemuKernel here for now.

p4zuu:
> Can u show some example outputs in text or screenshots for this?

Sure, here is how it looks while debugging an x86_64 Linux kernel in qemu:
image

Please feel free to also give feedback on UI/naming :)

pwndbg/commands/msr.py
p4zuu: `pwndbg.aglib.regs.rax=123` actually works, thanks (still new to the codebase sorry).
Moreover, we should be able to avoid this difference between x86 and x86_64. On both arch, eax, ecx and edx are used (x86_64 doesn't use the full registers size), but `e{a,c,d}x` don't exit in `pwndbg.aglib.regs` on x86_64, I still need to figure out how to use 32 bit registers

patryk4815: > but e{a,c,d}x don't exit in pwndbg.aglib.regs

this looks like bug, they should exists

disconnect3d:
image


1) Can we make this have `=` between name and value? This helps copying it into scripts if needed

2) ideally, would be nice to have the same indentation for value (some fancy f-string formatting needed)

3) can we add `--range=1,10` arg that would allow us to dump msr values for a range of ints?

disconnect3d:
Btw any chance you can add some qemu kernel test for this command?

pwndbg/commands/msr.py
p4zuu: It's quite strange, I have this exception when trying to assign to `ecx` on a x86_64 kernel (`pwndbg.aglib.regs.ecx = msr`):
```
pwndbg>  msr 0
Exception occurred: msr: Attempted to write to a non-existent register 'ecx' (<class 'RuntimeError'>)
```

However, I can read it from pwndbg.aglib.regs `eax = int(pwndbg.aglib.regs["eax"])` works fine.
And pwndbg.aglib.regs["eax"] doesn't support assignment, that's why I used `setattr()`

pwndbg/commands/msr.py
```diff
@@ -0,0 +1,146 @@
...
+
+def msr_list(arch: str):
+    msrs = COMMON_MSRS[arch]
+    for msr, value in msrs.items():
+        print(f"{msr}\t{hex(value)}")
+
+
+@pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.KERNEL, command_name="msr")
+@pwndbg.commands.OnlyWhenQemuKernel
+def msr(msr=Optional[int], name=Optional[str], write=Optional[int], list_msr=None) -> None:
```
disconnect3d: What are those default arguments? This is not how types are annotated in Python.

pwndbg/commands/msr.py
```diff
@@ -0,0 +1,146 @@
...
+import pwnlib.asm
+
+import pwndbg.aglib.file
+from pwndbg.commands import CommandCategory
+
+parser = argparse.ArgumentParser(
+    description="""
+Read or write to Model Specifc Register (MSR)
+""",
+)
```
disconnect3d: Please move the argument parser just before the command (and the helper functions above), so its closer to look up the argument parser when reading the command code.

pwndbg/commands/msr.py
```diff
@@ -0,0 +1,146 @@
...
+parser.add_argument("msr", help="MSR value (as integer)", type=int, nargs="?", default=None)
+parser.add_argument(
+    "-n",
+    "--name",
+    metavar="name",
+    help="MSR litteral value (eg. MSR_STAR)",
```
disconnect3d:
```suggestion
    help="MSR literal value (eg. MSR_STAR)",
```

pwndbg/commands/msr.py
```diff
@@ -0,0 +1,146 @@
...
+        print(f"{msr}\t{hex(value)}")
+
+
+@pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.KERNEL, command_name="msr")
+@pwndbg.commands.OnlyWhenQemuKernel
+def msr(msr=Optional[int], name=Optional[str], write=Optional[int], list_msr=None) -> None:
+    arch = pwndbg.aglib.arch.name
+
+    if msr is not None and name:
+        print("use MSR litteral or integer value, not both")
```
disconnect3d: 1) Typo "litteral
```suggestion
        print("Error: pass in a MSR literal (like MSR_EFER) or integer value, not both")
        return
```

disconnect3d (dismissed):
Honestly I feel like the logic here is a bit overly complex.

Why do we need the --name argument at all? Why not just pass an int or string as one argument?

This should be similar to sysctl CLI arguments. What about interface like this?
```
msr -l|--list              # list known MSRs
msr -l|--list 0-1000       # list MSRs from 0 to 1000
msr MSR_EFER               # get just MSR_EFER value
msr 0x123                  # get value for MSR 0x123
msr 0x123 -w 123           # set MSR 0x123 to 123
```

pwndbg/commands/msr.py
OBarronCS: The amd64 register set is defined here: https://github.com/pwndbg/pwndbg/blob/261fac85438d98842f567fd58888d4e55fd5be0a/pwndbg/lib/regs.py#L302-L304

Setattr, in this case, is dynamically adding the attribute to the object, which is undesired.

The reason we can "fetch" the register is that the fetch ignores the register list (and directly gets the register value from the underlying debugger), while setting it has a check that it is in the register list:

https://github.com/pwndbg/pwndbg/blob/261fac85438d98842f567fd58888d4e55fd5be0a/pwndbg/dbg/gdb/__init__.py#L166-L168

p4zuu:
> Why do we need the --name argument at all? Why not just pass an int or string as one argument?

This was just to differentiate the 2 cases, but that makes sense to merge the 2 into the same argument that can be an int or a string. I'll rework that.
 
> msr -l|--list 0-1000       # list MSRs from 0 to 1000

Agree with these but I'm confused about this one, you mean reading MSRs from 0 to 100? I'm not sure about using the same flag for listing the MSRs names and values but also for reading a MSR range. I'm fine with allowing reading a range of MSRs but I would just use a different flag to avoid confusing between when we simply list and when we actually read the MSRs :)

pwndbg/commands/msr.py
p4zuu: Well, yep, late coding I guess. Thanks for catching it

disconnect3d:
Yes. We need a way to read a range of MSRs because that is super useful.

disconnect3d:
@p4zuu ping :)

p4zuu:
> @p4zuu ping :)

Quite busy rn, sorry for the late update. I reworked the PR with the range, I still need to add qemu-system tests, hopeful will be able to push it this evening :)

pwndbg/commands/msr.py
p4zuu: Thanks for the precision @OBarronCS. So if I understand correctly, there is no way to specifically write to ecx on amd64?

patryk4815: you can use `mov ecx, 0xsomevalue` in asm right?

p4zuu:
New version:
- merged into a single commit for simplicity's sake
- removed the `-n` flag, now a msr can be an int (as dec or hex) or the literal name (eg, `msr 3221225602` == `msr 0xc0000082` == `msr MSR_LSTAR`)
- added a range flag to `read` to a `start-end` msr range, where `end` is included
- added qemu-system msr read and write tests

pwndbg/commands/msr.py
p4zuu: Outdated now since we removed -n flag

pwndbg/commands/msr.py
p4zuu: Done

pwndbg/commands/msr.py
p4zuu: Fixed now

pwndbg/commands/msr.py
p4zuu: This error message is outdated now, but I kept the "Error:" prefix for all handled error

pwndbg/commands/msr.py
p4zuu: I guess so. I guess I could add this mov in the shellcode prior the `rdmsr` or `wrmsr` instruction

p4zuu: @patryk4815 just pushed a last version inlining `mov` instructions in the shellcodes

pwndbg/commands/msr.py
```diff
@@ -0,0 +1,181 @@
...
+    "--range",
+    dest="msr_range",
+    help="dash separated range of MSRs to read (eg. --range=1-10 where 10 is included)",
+    type=str,
+    nargs="?",
+    default=None,
+)
+
+
+@pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.KERNEL, command_name="msr")
```
patryk4815:
```suggestion
@pwndbg.commands.ArgparsedCommand(parser, category=CommandCategory.KERNEL)
```
command_name is default like your function name

pwndbg/commands/msr.py
```diff
@@ -0,0 +1,181 @@
...
+        try:
+            msr_val = int(msr, 16)
+        except ValueError:
+            if msr not in COMMON_MSRS[arch]:
+                print(f"unknown MSR {msr} on {arch}")
+                return None
+
+            msr_val = COMMON_MSRS[arch][msr]
+
+    return msr_val
```
patryk4815:
```suggestion
def parse_msr(msr: str, arch: str) -> Optional[int]:
    # first try to parse MSR name, then as int/hex
    if (val := COMMON_MSRS.get(arch, {}).get(msr)) is not None:
        return val
    
    try:
        return int(msr, 0)
    except ValueError:
        print(f"unknown MSR {msr} on {arch}")
        return None
```
@disconnect3d do you like this suggestion 🤔?

patryk4815 (approved):
From a code perspective, it looks good to me, @disconnect3d can you check if it's fine for you?

p4zuu:
The CI caught a bug in the qemu-tests, sometimes `test_command_msr_read` hangs (maybe test_command_msr_write as well), sometimes it doesn't (at least on my machine), and it's not always for the same kernel... I need to debug this

patryk4815:
> The CI caught a bug in the qemu-tests, sometimes `test_command_msr_read` hangs (maybe test_command_msr_write as well), sometimes it doesn't (at least on my machine), and it's not always for the same kernel... I need to debug this

image
do we have "bug" in `exec_shellcode` func, because this func don't recover all registers 🤔?

tests/qemu-tests/tests/system/test_commands_kernel.py
```diff
@@ -86,3 +86,25 @@ def get_slab_object_address():
         if len(matches) > 0:
             return (matches[0], cache_name)
     raise ValueError("Could not find any slab objects")
+
+
+def test_command_msr_read():
+    if pwndbg.aglib.arch.name not in ["x86", "x86-64"]:
+        return
```
disconnect3d: See https://docs.pytest.org/en/stable/reference/reference.html#pytest-skip
```suggestion
       pytest.skip("Unsupported architecture: msr tests only work on x86 and x86-64")
```

disconnect3d: This may require importing pytest if its not imported before

pwndbg/commands/msr.py
```diff
@@ -0,0 +1,176 @@
...
+def msr(
+    msr: Optional[str] = None,
+    write: Optional[int] = None,
+    list_msr=False,
+    msr_range: Optional[str] = None,
+) -> None:
+    arch = pwndbg.aglib.arch.name
+
+    if msr:
+        msr_val = parse_msr(msr, arch)
```
disconnect3d, line 152: Shall we maybe do `parse_msr(msr.upper(), arch)` to allow lowercase inputs?

disconnect3d:
> > The CI caught a bug in the qemu-tests, sometimes `test_command_msr_read` hangs (maybe test_command_msr_write as well), sometimes it doesn't (at least on my machine), and it's not always for the same kernel... I need to debug this
> 
> image do we have "bug" in `exec_shellcode` func, because this func don't recover all registers 🤔?

Yeah I believe this function is not looking which registers gets clobbered and always assuming the same set of registers being clobbered. So we have to change it accordingly so it accepts a list of clobbered registers and bring their value back too.

pwndbg/commands/msr.py
p4zuu: Yep, good idea

p4zuu: I think it's a bit better to turn to uppercase inside the `parse_msr()` function:
```python
if (val := COMMON_MSRS.get(arch, {}).get(msr.upper())) is not None:
```
With this, we don't have to repeat `.upper()` for all calls. It also doesn't interfere with the `0x` prefix if a msr hex value is submitted (even tough `0X` prefix seems to be also understood by `int(msr, 0)` )

disconnect3d:
@p4zuu Btw just a random note: you don't have to amend your git commit for the git commit history to look nice etc. since we usually squash commits for a given PR when we do a merge (unless the commit history is really nice and we don't forget to check for it, heh).

Pushing further commits that address code review issues allows us to review those more easily and when you do a force push, we have to go over all the changes again. Just saying for future :).

p4zuu:
> @p4zuu Btw just a random note: you don't have to amend your git commit for the git commit history to look nice etc. since we usually squash commits for a given PR when we do a merge (unless the commit history is really nice and we don't forget to check for it, heh).
> 
> Pushing further commits that address code review issues allows us to review those more easily and when you do a force push, we have to go over all the changes again. Just saying for future :).

Got it, thanks for letting me know, and sorry for the harder change tracking then :)

p4zuu:
> Yeah I believe this function is not looking which registers gets clobbered and always assuming the same set of registers being clobbered. So we have to change it accordingly so it accepts a list of clobbered registers and bring their value back too.

I'm a bit puzzled about this. Afaiu, registers changed to exectue `rdmsr` and `wrmsr` instructions are restored by default by `exec_shellcode()`. Btw, the potential hang happening in qemu-system tests never happened in manual testing, even though it happens quite often in qemu tests (also locally)

p4zuu:
> I'm a bit puzzled about this. Afaiu, registers changed to exectue `rdmsr` and `wrmsr` instructions are restored by default by `exec_shellcode()`. Btw, the potential hang happening in qemu-system tests never happened in manual testing, even though it happens quite often in qemu tests (also locally)

The random but frequent hangs in tests are actually hangs in the debugged kernel itself. Maybe `exec_shellcode()` still messes around with the vm state, but I don't get how yet. Also, I don't understand why I never get these with manual testing

p4zuu:
> The random but frequent hangs in tests are actually hangs in the debugged kernel itself. Maybe `exec_shellcode()` still messes around with the vm state, but I don't get how yet. Also, I don't understand why I never get these with manual testing

@disconnect3d @patryk4815 do you have some hints here? It seems indeed that `exec_shellcode()` messes around with the vm state, but according to the code, the registers changed to perform `rdmsr` and `wrmsr` are in the restored register set. Also flags are not touched by both instructions. I'm not familiar with async Python so maybe I did something wrong and some routines are not waited long enough, or something

bata24:
`-enable-kvm` ?

patryk4815:
@p4zuu you code looks correct, right now I'm on vacation, after few days I can look into that

p4zuu:
> `-enable-kvm` ?

@bata24 seems that the issue doesn't appear when running on kvm. I can't figure an always-failing test (since it's a bit random for now) for which I would have a success, but running a big bunch of consecutive tests (I mean msr read tests) on kvm works. Without kvm, I got a failing test pretty soon

disconnect3d:
@patryk4815 please look again and merge if all fine

patryk4815:
PR rebased, will be merged when tests are successful

docs/commands/kernel/msr.md
```diff
@@ -0,0 +1,35 @@
...
+
+|Short|Long|Help|
+| :--- | :--- | :--- |
+|-h|--help|show this help message and exit|
+|-w|--write|value to write in MSR|
+|-l|--list|list common MSRs for the current arch|
+|-r|--range|dash separated range of MSRs to read (eg. --range=1-10 where 10 is included)|
+
+<!-- END OF AUTOGENERATED PART. Do not modify this line or the line below, they mark the end of the auto-generated part of the file. If you want to extend the documentation in a way which cannot easily be done by adding to the command help description, write below the following line. -->
+<!-- ------------\>8---- ----\>8---- ----\>8------------ -->
```
patryk4815:
````suggestion
# msr


```text
usage: msr [-h] [-w [write_value]] [-l] [-r [MSR_RANGE]] [msr]

```

Read or write to Model Specific Register (MSR)
### Positional arguments

|Positional Argument|Help|
| :--- | :--- |
|msr|MSR value or name|

### Optional arguments

|Short|Long|Help|
| :--- | :--- | :--- |
|-h|--help|show this help message and exit|
|-w|--write|value to write in MSR|
|-l|--list|list common MSRs for the current arch|
|-r|--range|dash separated range of MSRs to read (eg. --range=1-10 where 10 is included)|


<!-- END OF AUTOGENERATED PART. Do not modify this line or the line below, they mark the end of the auto-generated part of the file. If you want to extend the documentation in a way which cannot easily be done by adding to the command help description, write below the following line. -->
<!-- ------------\>8---- ----\>8---- ----\>8------------ -->
````

p4zuu:
> PR rebased, will be merged when tests are successful

I'm afraid the qemu-tests suite is still failing (even tough qemu-tests in tests-using-nix seem ok)

PR #2838, opened 2025-04-06
Moving the check condition of hexdump invalid addresses

dkvhr:
Move the check of invalid addresses to after the `if address > pwndbg.aglib.arch.ptrmask:` check

This is a fix of PR #2833


disconnect3d:
Thanks!

PR #2840, opened 2025-04-06
Fix read_memory

patryk4815:
Before:
image

After:
```python
pwndbg> ipi

In [1]: pwndbg.dbg.selected_inferior().read_memory(0x1000000000000-0xff, 0xff, True)
Out[1]: bytearray(b'yr900dz34md3p-gdb-15.2/lib -L/nix/store/ibk04bmmyf4rrj139ppyr900dz34md3p-gdb-15.2/lib\x00nativeBuildInputs=\x00__CF_USER_TEXT_ENCODING=0x1F5:0x1D:0x2A\x00name=shell\x00depsHostHostPropagated=\x00/nix/store/4q0wb123xfksdvq23fz3d1qqmm6ncd6p-hello-2.12.1/bin/hello\x00\x00\x00\x00\x00\x00\x00\x00\x00')

In [2]: pwndbg.dbg.selected_inferior().read_memory(0x1000000000000-0x10, 0xff, True)
_is_readable 0x1000000000070 False
_is_readable 0x1000000000030 False
_is_readable 0x1000000000010 False
_is_readable 0x1000000000000 False
_is_readable 0xfffffffffff8 True
_is_readable 0xfffffffffffc True
_is_readable 0xfffffffffffe True
_is_readable 0xffffffffffff True
Out[2]: bytearray(b'n/hello\x00\x00\x00\x00\x00\x00\x00\x00\x00')

In [3]: pwndbg.dbg.selected_inferior().read_memory(0x1000000000000-0x2, 0xff, True)
_is_readable 0x100000000007e False
_is_readable 0x100000000003e False
_is_readable 0x100000000001e False
_is_readable 0x100000000000e False
_is_readable 0x1000000000006 False
_is_readable 0x1000000000002 False
_is_readable 0x1000000000000 False
_is_readable 0xffffffffffff True
Out[3]: bytearray(b'\x00\x00')

In [4]: pwndbg.dbg.selected_inferior().read_memory(0x1000000000000-0x1, 0xff, True)
_is_readable 0x100000000007f False
_is_readable 0x100000000003f False
_is_readable 0x100000000001f False
_is_readable 0x100000000000f False
_is_readable 0x1000000000007 False
_is_readable 0x1000000000003 False
_is_readable 0x1000000000001 False
_is_readable 0x1000000000000 False
_is_readable 0xffffffffffff True
Out[4]: bytearray(b'\x00')

In [5]: pwndbg.dbg.selected_inferior().read_memory(0x1000000000000-0x0, 0xff, True)
MemoryError                               Traceback (most recent call last)
File /nix/store/w50mwxpn0ay1yi1928bgidqpkys5xq56-pwndbg/share/pwndbg/pwndbg/dbg/gdb/__init__.py:463, in GDBProcess.read_memory(self, address, size, partial)
    462 try:
--> 463     result = gdb.selected_inferior().read_memory(addr, count)
    464     return bytearray(result)

MemoryError: Cannot access memory at address 0x1000000000000

Error                                     Traceback (most recent call last)
Cell In[2], line 1
----> 1 pwndbg.dbg.selected_inferior().read_memory(0x1000000000000-0x0, 0x3, True)

File /nix/store/w50mwxpn0ay1yi1928bgidqpkys5xq56-pwndbg/share/pwndbg/pwndbg/dbg/gdb/__init__.py:496, in GDBProcess.read_memory(self, address, size, partial)
    493     if stop_addr > addr:
    494         return self.read_memory(addr, stop_addr - addr)
--> 496 raise pwndbg.dbg_mod.Error(e)

Error: Cannot access memory at address 0x1000000000000
```

### DIFF
--- a/pwndbg/dbg/gdb/__init__.py
+++ b/pwndbg/dbg/gdb/__init__.py
@@ -430,54 +430,70 @@ class GDBProcess(pwndbg.dbg_mod.Process):
         pages.sort()
         return GDBMemoryMap(qemu, pages)
 
+    def _find_memory_last_readable(self, start: int, count: int) -> int:
+        end = start + count
+        result = -1
+
+        def _is_readable(addr) -> bool:
+            try:
+                self.read_memory(addr, 1)
+                return True
+            except pwndbg.dbg_mod.Error:
+                return False
+
+        if not _is_readable(start):
+            return result
+
+        while start <= end:
+            mid = (start + end + 1) // 2
+            if _is_readable(mid):
+                result = mid
+                start = mid + 1
+            else:
+                end = mid - 1
+
+        return result
+
     @override
     def read_memory(self, address: int, size: int, partial: bool = False) -> bytearray:
-        result = b""
         count = max(int(size), 0)
         addr = address
 
         try:
             result = gdb.selected_inferior().read_memory(addr, count)
+            return bytearray(result)
         except gdb.error as e:
             if not partial:
                 raise pwndbg.dbg_mod.Error(e)
 
-            message = str(e)
-
-            stop_addr = addr
-            match = re.search(r"Memory at address (\w+) unavailable\.", message)
-            if match:
-                stop_addr = int(match.group(1), 0)
+            if pwndbg.aglib.remote.is_remote():
+                # GDB remote debugging will return the start address as the failed
+                # read address. Try moving back a few pages at a time.
+                stop_addr = self._find_memory_last_readable(addr, count)
+                if stop_addr > 0:
+                    return self.read_memory(addr, stop_addr - addr + 1)
             else:
-                stop_addr = int(message.split()[-1], 0)
+                message = str(e)
 
-            # Handle case of memory read that wraps around the memory space back to 0, where high memory was readable but memory at 0 was not.
-            # Example: 2-byte read at 0xFFFF_FFFF in a 32-bit address space.
-            # GDB returns error: "Cannot access memory at address 0x0"
-            if stop_addr == 0 and stop_addr < addr:
-                # We could read from the top-portion of memory, but not after wrapping around
-                # Because we are doing a partial read, read until the max address
-                return self.read_memory(addr, pwndbg.aglib.arch.ptrmask - addr + 1)
+                stop_addr = addr
+                match = re.search(r"Memory at address (\w+) unavailable\.", message)
+                if match:
+                    stop_addr = int(match.group(1), 0)
+                else:
+                    stop_addr = int(message.split()[-1], 0)
 
-            if stop_addr != addr:
-                return self.read_memory(addr, stop_addr - addr)
+                # Handle case of memory read that wraps around the memory space back to 0, where high memory was readable but memory at 0 was not.
+                # Example: 2-byte read at 0xFFFF_FFFF in a 32-bit address space.
+                # GDB returns error: "Cannot access memory at address 0x0"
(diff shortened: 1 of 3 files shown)


pwndbg/dbg/gdb/__init__.py
```diff
@@ -430,54 +430,70 @@ def vmmap(self) -> pwndbg.dbg_mod.MemoryMap:
         pages.sort()
         return GDBMemoryMap(qemu, pages)
 
+    def _find_memory_last_readable(self, start: int, count: int) -> int:
+        end = start + count
+        result = -1
+
+        def _is_readable(addr) -> bool:
```
disconnect3d: Te function object is constructed on each execution of this function. Since it does not use any closure, pls put it outside of _find_memory_last_readable, or inline its logic below.

pwndbg/dbg/gdb/__init__.py
```diff
@@ -430,54 +430,70 @@ def vmmap(self) -> pwndbg.dbg_mod.MemoryMap:
...
                 raise pwndbg.dbg_mod.Error(e)
 
-            message = str(e)
-
-            stop_addr = addr
-            match = re.search(r"Memory at address (\w+) unavailable\.", message)
-            if match:
-                stop_addr = int(match.group(1), 0)
+            if pwndbg.aglib.remote.is_remote():
+                # GDB remote debugging will return the start address as the failed
```
disconnect3d: 1) Lets first handle local debugging instead of remote first
2) Lets expand on the comment
3) Add link to the GDB bug if its already reported. Report if not.
4) Add a general comment to the except part such as:
> Memory read failed - likely bcoz part of the memory is unmapped


Then a comment for remote:
> On remote debugging, the error message contains start address instead of stop address. This is a GDB bug tracked by <LINK>
> To work around this, we perform a binary search to find the proper stop address that does not fail

disconnect3d: Also, any chance we can add a test for this behavior?

disconnect3d:
~~Don't we have an infinite loop here if someone calls read_memory on unmapped address?~~

EDIT: No, we don't due to partial=True. but we can still use just gdb.read_mem~~ here

image

pwndbg/dbg/gdb/__init__.py
```diff
@@ -430,54 +430,74 @@ def vmmap(self) -> pwndbg.dbg_mod.MemoryMap:
...
+                    return self.read_memory(addr, stop_addr - addr + 1)
 
-            # QEMU will return the start address as the failed
-            # read address.  Try moving back a few pages at a time.
-            stop_addr = addr + count
-
-            # Move the stop address down to the previous page boundary
-            stop_addr &= PAGE_MASK
-            while stop_addr > addr:
-                result = self.read_memory(addr, stop_addr - addr)
```
OBarronCS, line 472: From what I gather, this bug resulted from read_memory() at one point returning a None object on failed memory reads, but then later it was changed to throw an error, and this was not updated to reflect that?

OBarronCS (approved):

pwndbg/dbg/gdb/__init__.py
patryk4815: Wow, I didn't know that previously it was `None` and changed to exception

OBarronCS: Just speculation - otherwise, the original implementation wouldn't have worked.

PR #2842, opened 2025-04-07
add libcinfo command

disconnect3d:
This command shows libc version of currently debugged binary and link to its sources.

Currently it only supports glibc and shows link to raw/pure gnu.org glibc sources. Meaning, it does not handle custom/patched glibc versions that may be there on Ubuntu etc.

In the future this command could support other libcs (like musl used by Alpine) or/and display the binary build_id and maybe a link to libcdb.


patryk4815:
👍 btw, we should consider adding https://github.com/niklasb/libc-database
https://github.com/niklasb/libc-database/tree/master/searchengine

We can get `build-id` from objfiles():
image

image

patryk4815 (approved):

PR #2844, opened 2025-04-07
Build docs in strict mode

k4lizen:
Since we are adhering to google style docstrings, we don't need to specify type parameters in the docstring if they are type annotated. Also, mkdocstrings doesn't like recognize this `param(type)` syntax and currently only accepts the more commonly used `param (type)` (with a space) (see https://github.com/mkdocstrings/griffe/issues/375).

### DIFF
--- a/pwndbg/emu/emulator.py
+++ b/pwndbg/emu/emulator.py
@@ -733,16 +733,14 @@ class Emulator:
         debug(DEBUG_MEM_READ, "uc.mem_read(*%r, **%r)", (a, kw))
         return self.uc.mem_read(*a, **kw)
 
-    def until_jump(self, pc=None):
+    def until_jump(self, pc: int = None):
         """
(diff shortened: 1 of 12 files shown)


pwndbg/emu/emulator.py
```diff
@@ -733,16 +733,14 @@ def mem_read(self, *a, **kw):
...
+    def until_jump(self, pc: int = None):
         """
         Emulates instructions starting at the specified address until the
         program counter is set to an address which does not linearly follow
         the previously-emulated instruction.
 
         Arguments:
-            pc(int): Address to start at.  If `None`, uses the current instruction.
-            types(list,set): List of instruction groups to stop at.
-                By default, it stops at all jumps, calls, and returns.
```
k4lizen, line 745: this parameter just isn't in the function :?

PR #2846, opened 2025-04-08
Implement autogenerating docs

k4lizen:
Okay here is the idea. Every file in `docs/commands/` is structured like:
```
<autogenerated part>
<marker>
<hand written part>
``` 
The autogenerated part includes the usage, description and parameters, extracted from the source (specifically from the argparse object). Then, running `./scripts/generate_docs.sh` will update all those `<autogenerated part>`s. The `<marker>` is used by the script and is invisible to the user viewing a doc. 

Running `./scripts/verify_docs.sh` checks if the autogenerated parts are up to date on all files, this gets run in CI when building docs. If someone updates a command help text in the source, this check will fail and tell them to run `./scripts/generate_docs.sh` which they can then commit and push, so the docs will always be up to date.

A special file that gets autogenerated is `docs/commands/index.md` which lists all the commands.

Other notes:
I made `category` a mandatory argument for commands, before they defaulted to misc, but I want this to be a more conscious decision.

The usage text is wrapped at 80 columns:
image
by patching `shutil.get_terminal_size` in `scripts/_gen_command_docs.py`. I don't know of a better way to do this (I tried soo many), and it needs to be consistent so different people don't get different doc generation (and fail CI etc..). Anyways this doesn't affect pwndbg usage, just wanted to note.

Also, it would be good to figure out any potential changes to the markers now because changing them later will be a pain. Right now we can just `rm -r docs/commands && ./scripts/generate_docs.sh` for any refactors but that won't be viable after we have added hand written parts to the commands. We could put the hand written parts in a seperate folder altogether and have the script combine them? That would give us a bit more flexibility but it could be annoying when writing the file or viewing it from github.


docs/commands/kernel/knft_list_chains.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Description
+
+
+Dump netfilter chains form a specific table
+## Usage:
+
+
+```bash
+usage: knft_list_chains [-h] [--nsid NSID] [table_family] [table_name]
```
patryk4815, line 16: command with underscore `knft_list_chains` is not valid.
Only with hyphen should be here? `knft-list-chains` 
Please fix this in autogen in next PR :)

docs/commands/linux_libc_elf/onegadget.md
```diff
@@ -0,0 +1,34 @@
...
+
+
+
+
+# onegadget
+
+## Description
+
+
+Show onegadget
```
disconnect3d, line 11: This needs better description lol. Pls improve in another PR

pwndbg/commands/branch.py
```diff
@@ -55,7 +58,9 @@ def break_if_taken(branch) -> None:
 )
 
 
-@pwndbg.commands.ArgparsedCommand(parser, command_name="break-if-not-taken")
+@pwndbg.commands.ArgparsedCommand(
+    parser, command_name="break-if-not-taken", category=CommandCategory.MISC
```
disconnect3d, line 62: Probably should be in BREAKPOINT category?

disconnect3d: Same for the other one, pls do in another PR

disconnect3d (approved):
Left few comments but pls address them in a next PR

docs/commands/kernel/knft_list_chains.md
k4lizen: image

what?

k4lizen: I can change the name of the command if you want, but it seems it's consistent with the usage

patryk4815: 🤔 aaa maybe!

PR #2829, opened 2025-04-02
Add a command to dump the TLS

dkvhr:
Read the memory from fs/gs register, get the first index, add it to an offset and then display its contents. I called the command `dumptls`, as it is not specifically related to the fs register.

This is related to this issue: #2814

PR #2847, opened 2025-04-08
Dump the TLS address by default

### LINKED ISSUE #2814: Command request: dump fs at offset (x86-64)
```
pwndbg> p/x $fs:[0x30]
A syntax error in expression, near `:[0x30]'.
```

We can have a command that would:
1) save current state
2) assemble a `mov rax, qword ptr fs:[IDX]` instruction at $RIP
3) execute it
4) revert state
5) print the FS:[0x30] value

maybe: `dumpfs`? or `dump-fs-reg`? We need a good name for it (cc: @patryk4815 @OBarronCS @CptGibbon )

dkvhr:
I am creating a new PR as I messed up the previous one: #2829 

This dumps by default the TLS address with `dt tcbhead_t <tls_address>` if there is a `tcbhead_t`. If there is not, use telescope to dump it.
It will print by default until the `__glibc_unused2` on x86-64. If there isn't one, it will print just 10 lines.
There is an argument `-a` to print the full output, which is told to the user when it runs and the output is truncated.

It also changes the TLS tests to adapt to the new changes. I plan on adding more specific tests related to x86-64 and i386 in the near future.

Closes #2814

### DIFF
--- a/pwndbg/commands/tls.py
+++ b/pwndbg/commands/tls.py
@@ -53,7 +53,6 @@ def tls(pthread_self=False, all: bool = False) -> None:
         # Trying to `dt tcbhead_t <tls_base>` it by default
         # If there is no tcbhead_t type, we can just use telescope.
         typename = "tcbhead_t"
-        address = tls_base
         # dt doesn't seem to return anything in case it succeeds.
         output = str(pwndbg.aglib.dt.dt("tcbhead_t", addr=tls_base))
 
--- a/pwndbg/commands/tls.py
+++ b/pwndbg/commands/tls.py
@@ -50,9 +50,8 @@ def tls(pthread_self=False, all: bool = False) -> None:
         print(message.success("TLS is located at:"))
         print(message.notice(pwndbg.aglib.vmmap.find(tls_base)))
 
-        # Trying to `dt tcbhead_t <tls_base>` it by default
-        # If there is no tcbhead_t type, we can just use telescope.
-        # dt doesn't seem to return anything in case it succeeds.
+        # Displaying `dt tcbhead_t <tls_base>` if possible
+        # If not, we will dump the tls with telescope
         output = str(pwndbg.aglib.dt.dt("tcbhead_t", addr=tls_base))
 
         print(message.success("Dumping the address:"))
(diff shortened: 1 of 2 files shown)


pwndbg/commands/tls.py
```diff
@@ -46,6 +49,49 @@ def tls(pthread_self=False) -> None:
         print(message.success("Thread Local Storage (TLS) base: %#x" % tls_base))
         print(message.success("TLS is located at:"))
         print(message.notice(pwndbg.aglib.vmmap.find(tls_base)))
+
+        # Trying to `dt tcbhead_t <tls_base>` it by default
+        # If there is no tcbhead_t type, we can just use telescope.
+        typename = "tcbhead_t"
+        address = tls_base
+        # dt doesn't seem to return anything in case it succeeds.
+        output = str(pwndbg.aglib.dt.dt(typename, addr=address))
```
disconnect3d: just inline the typename literal here

disconnect3d:
```suggestion
        output = str(pwndbg.aglib.dt.dt("tcbhead_t", addr=tls_base))
```

pwndbg/commands/tls.py
```diff
@@ -46,6 +49,49 @@ def tls(pthread_self=False) -> None:
         print(message.success("Thread Local Storage (TLS) base: %#x" % tls_base))
         print(message.success("TLS is located at:"))
         print(message.notice(pwndbg.aglib.vmmap.find(tls_base)))
+
+        # Trying to `dt tcbhead_t <tls_base>` it by default
+        # If there is no tcbhead_t type, we can just use telescope.
+        typename = "tcbhead_t"
+        address = tls_base
```
disconnect3d:
```suggestion
```

pwndbg/commands/tls.py
```diff
@@ -46,6 +49,48 @@ def tls(pthread_self=False) -> None:
         print(message.success("Thread Local Storage (TLS) base: %#x" % tls_base))
         print(message.success("TLS is located at:"))
         print(message.notice(pwndbg.aglib.vmmap.find(tls_base)))
+
+        # Trying to `dt tcbhead_t <tls_base>` it by default
+        # If there is no tcbhead_t type, we can just use telescope.
+        typename = "tcbhead_t"
```
disconnect3d:
```suggestion
```

PR #2850, opened 2025-04-08
fix: adjust to `rr`'s vFile reply

RocketMaDev:
`rr replay` use `"F-01,2"` to indicate a vFile error while pwndbg detects `"F-1,"`. Patch the code to process some cases like `rr`.

As this piece of code is run when displaying context, it's critical to fix it. pwndbg tries to cache `/proc/$tid/maps` which utilize vFile to grab it.

### DIFF
--- a/pwndbg/aglib/file.py
+++ b/pwndbg/aglib/file.py
@@ -173,8 +173,8 @@ def _vfile_check_response(response: bytes):
     if response.startswith(b"F") and b"," in response:
         resp_status, resp_data = response[1:].split(b",", 1)  # Skip "F"
         if int(resp_status) != -1:
-            raise NotImplementedError(f"Unknown response status {resp_status}")
-        errno = int(resp_data.decode(), 10 if is_vfile_qemu_user_bug() else 16)
+            raise NotImplementedError(f"Unknown response status {resp_status!r}")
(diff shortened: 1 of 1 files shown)


pwndbg/aglib/file.py
```diff
@@ -170,8 +170,11 @@ def is_vfile_qemu_user_bug() -> bool:
 def _vfile_check_response(response: bytes):
     if len(response) == 0:
         raise NotImplementedError("Not supported")
-    if response.startswith(b"F-1,"):
-        errno = int(response[4:].decode(), 10 if is_vfile_qemu_user_bug() else 16)
+    if response.startswith(b"F") and b"," in response:
+        resp_status, resp_data = response[1:].split(b",", 1)  # Skip "F"
```
patryk4815: `resp_data` may return `errno;attachment`, we need `split(';', 1)[0]` here

RocketMaDev: Do you mean `resp_status`?

patryk4815: `resp_data`

patryk4815: `‘F result [, errno] [; attachment]’`

PR #2854, opened 2025-04-10
Setup github pages for doc versioning

k4lizen:
preview: https://hazyclimb.dev/pwndbg/

Currently the version selector in the header only shows dev. When we release more versions they will be added there. I'm not sure we can add past versions retroactively.


disconnect3d:
Wtf?

k4lizen:
Ah it's because the other PR was merged, which updated the docs, so mkdocs amends the one gh-pages commit and so histories diverged from my branch.

I'll fix it tomorrow

patryk4815:
merged manually

PR #2855, opened 2025-04-10
Fix compatibility issues with Windows

patryk4815:
Fixed:
- readline 
- ui terminal
- disasm mmap detection for windows (it doesn't work on linux/macOS when lazy allocation is enabled in kernel `vm.overcommit_memory=1`) 
image

### DIFF
--- a/pwndbg/aglib/disasm/__init__.py
+++ b/pwndbg/aglib/disasm/__init__.py
@@ -300,19 +300,9 @@ def can_run_first_emulate() -> bool:
     first_time_emulate = False
 
     try:
-        from mmap import mmap, MAP_ANON, MAP_PRIVATE, PROT_EXEC, PROT_READ, PROT_WRITE  # isort:skip
+        from mmap import mmap
 
-        access_mode = PROT_WRITE | PROT_READ | PROT_EXEC
-        if sys.platform == "darwin":
-            # On macOS (Darwin), creating mmap with 'rwx' permissions is blocked
-            # by System Integrity Protection (SIP), which prevents execution of code
-            # from writable and readable memory regions. Therefore, we restrict
-            # the access mode to 'rw' (read and write only).
-            access_mode = PROT_WRITE | PROT_READ
-
-        mm = mmap(  # novm
-            -1, 1024 * 1024 * 1024, MAP_PRIVATE | MAP_ANON, access_mode
-        )
+        mm = mmap(-1, 1024 * 1024 * 1024)
         mm.close()
     except OSError:
         print(
--- a/pwndbg/dbg/lldb/repl/readline.py
+++ b/pwndbg/dbg/lldb/repl/readline.py
@@ -6,22 +6,22 @@ Mostly concerns itself with argument completion.
 
 from __future__ import annotations
 
-import sys
 import contextlib
 import functools
 import os.path
+import sys
 from typing import Callable
 from typing import ParamSpec
 from typing import TypeVar
 
-if sys.platform == 'win32':
-    import pyreadline3
+if sys.platform == "win32":
     import readline
 
     # pyreadline3 don't implement `set_completion_display_matches_hook`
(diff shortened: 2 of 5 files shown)


pwndbg/aglib/disasm/__init__.py
```diff
@@ -300,19 +300,9 @@ def can_run_first_emulate() -> bool:
...
-            # from writable and readable memory regions. Therefore, we restrict
-            # the access mode to 'rw' (read and write only).
-            access_mode = PROT_WRITE | PROT_READ
-
-        mm = mmap(  # novm
-            -1, 1024 * 1024 * 1024, MAP_PRIVATE | MAP_ANON, access_mode
-        )
+        from mmap import mmap
+
+        mm = mmap(-1, 1024 * 1024 * 1024)
```
disconnect3d, line 305: Note: we will now not attempt RWX alloc on linux which may fail on some secured kernels (like grsecurity) but that is probably fine.

pwndbg/dbg/lldb/repl/readline.py
```diff
@@ -9,11 +9,19 @@
...
+import sys
 from typing import Callable
 from typing import ParamSpec
 from typing import TypeVar
 
-import gnureadline as readline
+if sys.platform == "win32":
+    import readline
+
+    # pyreadline3 don't implement `set_completion_display_matches_hook`
```
disconnect3d:
```suggestion
    # pyreadline3 doesn't implement `set_completion_display_matches_hook`
```

pwndbg/dbg/lldb/repl/readline.py
```diff
@@ -9,11 +9,19 @@
...
 from typing import Callable
 from typing import ParamSpec
 from typing import TypeVar
 
-import gnureadline as readline
+if sys.platform == "win32":
+    import readline
+
+    # pyreadline3 doesn't implement `set_completion_display_matches_hook`
+    readline.set_completion_display_matches_hook = lambda *args: None
```
disconnect3d: Shall we make a check for that with some warning in case windows implement that one day?

pwndbg/dbg/lldb/repl/readline.py
```diff
@@ -9,11 +9,19 @@
 import contextlib
 import functools
 import os.path
+import sys
 from typing import Callable
 from typing import ParamSpec
 from typing import TypeVar
 
-import gnureadline as readline
+if sys.platform == "win32":
```
disconnect3d: Can we do `if sys.platform != "win32": gnureadline; else: windows`?

pwndbg/dbg/lldb/repl/readline.py
patryk4815: fixed

pwndbg/dbg/lldb/repl/readline.py
patryk4815: fixed

disconnect3d (approved):

disconnect3d:
Regarding Windows, what exactly do those changes allow us for? Running GDB and LLDB under Windows? Or just LLDB? Or is it something else? Asking just so we have this as information here. Also pls add this info to the commit log when hitting "Squash and merge".

patryk4815:
These changes primarily fix compatibility issues that were preventing `import pwndbg` from working on Windows. 
The main improvement is that Pwndbg (as library) can now be imported successfully on Windows without raising exceptions.

There are still a few blockers — for example, `os.set_blocking` in `pwndbg/dbg/lldb/repl/io.py`, which is unavailable on Windows. That will be addressed in a next PR.

Also worth noting: *we do not plan to support GDB on Windows*, as it has very limited support — particularly lacking `.pdb` debug symbol support, which makes it impractical for real-world usage.

The long-term goal is to support pwndbg with lldb.exe on Windows. 
This PR lays the groundwork for that.

PR #2856, opened 2025-04-10
Cleanup configurable parameters

k4lizen:
+ Short descriptions cant be more than 70 characters anymore. Config names can't be more than 32 characters anymore.
+ Default values are printed with the help.
+ For enums and autobool, the valid values are printed with the help
+ Replaced the `repr(v.value)` thingy with a bit more robust function for visualizing parameter values

Before:
image
image
image

After:
image
image
image

### DIFF
--- a/pwndbg/aglib/disasm/arch.py
+++ b/pwndbg/aglib/disasm/arch.py
@@ -35,9 +35,7 @@ if TYPE_CHECKING:
 pwndbg.config.add_param(
     "emulate",
     "on",
-    """
-Unicorn emulation of code from the current PC register
-""",
+    "unicorn emulation of code from the current PC register",
     help_docstring="""\
 emulate can be:
 off             - no emulation is performed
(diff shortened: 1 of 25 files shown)


k4lizen:
@disconnect3d note that this PR introduces a breaking change because the telescope parameter `telescope-skip-repeating-val-minimum` is renamed to `telescope-skip-repeating-val-min` (to bring it to 32 character limit, where some other parameters also lie, this was previously the longest parameter), we should probably point it out in the next release.

disconnect3d:
@k4lizen Maybe it should be: "telescope-skip-repeating-val" and we should skip repeating or not, instead of minimum? ;D idk

k4lizen:
@disconnect3d That does exist
```
telescope-skip-repeating-val       whether to skip repeating values of the telescope command                on 
telescope-skip-repeating-val-min   minimum amount of repeated values before skipping lines                  3 
```

disconnect3d:
Ok, that makes sense 😓

PR #2857, opened 2025-04-11
Enforce - (instead of _) in command names

k4lizen:


disconnect3d:
This needs fixing tests 😓

CptGibbon:
I think this is a big change from an established user's point of view. Perhaps I missed the conversation that lead to this decision but there's no description on this PR so could you elucidate?

k4lizen:
@CptGibbon See https://discord.com/channels/843809097920413717/843822664841166848/1359976744160989369

The rationale is that there are currently some commands with `-` and some with `_` so to unify the user experience this forces them to one of those.

CptGibbon:
> @CptGibbon See https://discord.com/channels/843809097920413717/843822664841166848/1359976744160989369

I only see one comment from yourself suggesting that it's easier to type - because you don't have to hold shift 😅

I agree, but doesn't Python disallow hyphens in names for the obvious reason? So all the commands use dashes while their matching functions use underscores? Not a big deal I suppose but I still think it's an odd change since it could (briefly) confuse existing users, but for what benefit?

Then again I do hate pressing shift 😁

k4lizen:
> So all the commands use dashes while their matching functions use underscores?

I do agree this is a bit unfortunate, didn't realize it before writing up the PR.

CptGibbon:
Hey it's your call I'm just thinking out loud.
I don't think it's common for folks to invoke pwndbg commands from a script so perhaps not an issue there.

disconnect3d:
We had this discussion on Discord. I definetely think we need to make this consistent. I also feel like using '-' is easier than '_' and people use commands from Pwndbg CLI much more often than commands as functions. And if they do, they need to use '-'. They can also use 'gdb.execute(...)' heh.

k4lizen:
I agree, just to have it written down though, one other thing that is unfortunate with `-` is that `malloc_chunk` the command had an underscore because it had the same name as the `struct malloc_chunk` struct in glibc. It's not a big deal, just mentioning it.

PR #2866, opened 2025-04-12
Add step to set latest as default docs version

Mahdiglm:
This PR adds a step to the docs-release.yml workflow to set the default documentation version to 'latest' after a new release is deployed. This ensures visitors always see the latest stable release documentation by default, rather than the development version. Fixes the issue mentioned in gh-pages branch documentation versioning.

### DIFF
--- a/.github/workflows/docs-release.yml
+++ b/.github/workflows/docs-release.yml
@@ -37,3 +37,7 @@ jobs:
             --remote origin \
             ${{ github.ref_name }} \
             latest
+
+      - name: Set latest as default version
+        run: |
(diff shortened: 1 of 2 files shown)


k4lizen:
Why though? The command only needs to be run once, not after every release is published

Mahdiglm:
> Why though? The command only needs to be run once, not after every release is published

You're right. we don't need to run this command after every release. The default version only needs to be set once.
The issue mentioned that "after the next release someone should update the default version" so this was meant to be a one-time fix rather than part of the workflow. I misunderstood and thought we should automate it for every future release.

PR #2868, opened 2025-04-13
hint users about libthread_db configuration if arena not found

tesuji:
Today I encountered an error when trying to view heap chunks while switching between (p)threads. I used `pathelf` to set runpath to my cwd. And pwndbg (or vanilla gdb) cannot use my libthread_db that exists on my cwd. So a little hint about this gdb configuration might help people in the future. 
image
image

Note: gef, however, still is [able to view/get heap](https://github.com/pwndbg/pwndbg/pull/2868#issuecomment-2800068340) chunks while it doesn't use libthread_db. I haven't dug deep to understand what gef does.

### DIFF
--- a/pwndbg/commands/ptmalloc2.py
+++ b/pwndbg/commands/ptmalloc2.py
@@ -144,6 +144,12 @@ def print_no_arena_found_error(tid=None) -> None:
             f"No arena found for thread {message.hint(tid)} (the thread hasn't performed any allocations)."
         )
     )
+    print(
+        message.hint(
+            "Use `set libthread-db-search-path <path>` if the path contains a compatible libthread_db.\n"
+            "Then rerun the program."
+        )
+    )
 
 
 def print_no_tcache_bins_found_error(tid: int | None = None) -> None:
--- a/pwndbg/commands/ptmalloc2.py
+++ b/pwndbg/commands/ptmalloc2.py
@@ -144,12 +144,13 @@ def print_no_arena_found_error(tid=None) -> None:
             f"No arena found for thread {message.hint(tid)} (the thread hasn't performed any allocations)."
         )
     )
-    print(
-        message.hint(
-            "Use `set libthread-db-search-path <path>` if the path contains a compatible libthread_db.\n"
-            "Then rerun the program."
+    if pwndbg.dbg.name() == pwndbg.dbg_mod.DebuggerType.GDB:
+        print(
+            message.hint(
+                "Use `set libthread-db-search-path <path>` if the path contains a compatible libthread_db.\n"
+                "Then rerun the program."
+            )
         )
-    )
 
 
 def print_no_tcache_bins_found_error(tid: int | None = None) -> None:


pwndbg/commands/ptmalloc2.py
```diff
@@ -144,6 +144,12 @@ def print_no_arena_found_error(tid=None) -> None:
             f"No arena found for thread {message.hint(tid)} (the thread hasn't performed any allocations)."
         )
     )
+    print(
```
disconnect3d: Since this is GDB-specific we need to guard it with:
```py
if pwndbg.dbg.name() == pwndbg.dbg_mod.DebuggerType.GDB:
```

pwndbg/commands/ptmalloc2.py
```diff
@@ -144,6 +144,12 @@ def print_no_arena_found_error(tid=None) -> None:
...
+        message.hint(
+            "Use `set libthread-db-search-path <path>` if the path contains a compatible libthread_db.\n"
+            "Then rerun the program."
```
disconnect3d: We can leave this as-is, but I wonder: shall we inform users what libthread-db is? I only discovered it today myself.

Shall we maybe also link to https://sourceware.org/gdb/current/onlinedocs/gdb.html/Threads.html#:~:text=set%20libthread%2Ddb%2Dsearch%2Dpath%20%5Bpath%5D ? Is this any helpful? Ugh!

disconnect3d:
What was the output before ur patch?

cc: @lebr0nli do you maybe know why gef works here without doing this libthread_db stuff and our code doesn't?

pwndbg/commands/ptmalloc2.py
tesuji: At that moment the `heap` commands and friends doesn't work. Users could google the configuration for more info. I was trying to be concise in the printing message.

disconnect3d:
@tesuji any chance u can add the gdb-specific check now? That way we will include this change in the latest release

tesuji:
> What was the output before ur patch?

Here is the screenshot:
image
Also the output on gef:
image

disconnect3d:
Thanks!

lebr0nli:
> cc: @lebr0nli do you maybe know why gef works here without doing this libthread_db stuff and our code doesn't?

I think the reason gef works here is because, if the arena address is not specified, the `heap chunks` command defaults to printing the chunks of first arena (`main_arena`), regardless of which thread we're debugging:
https://github.com/hugsy/gef/blob/63304c722c4ae29c3756da768ed65f1457287f06/gef.py#L6687-L6691

As for why `pwndbg` shows no result in this case, my guess is that if `libthread_db` isn't set up correctly, gdb fails to resolve the `thread_arena` symbol. However, our code doesn't raise an error when that happens and instead thought `thread_arena` is `NULL`.
^ But this is just my guess, I'm not sure how to reproduce the same issue locally.

@tesuji Any chance you could share the binaries from that CTF? Their CTFd seems not allowed me to check the challenge after the CTF ends.

tesuji:
@lebr0nli Your guess seems reasonable. GEF fallbacks to use `main_arena` when libthread_db not found, even not in the main thread. In the screenshot below, you can see that the heap addresses are the same between threads.
image

Anyway here's the binary attachment:
[NeonPulse.zip](https://github.com/user-attachments/files/19732445/NeonPulse.zip)

lebr0nli:
> Anyway here's the binary attachment: [NeonPulse.zip](https://github.com/user-attachments/files/19732445/NeonPulse.zip)

Thanks! I was able to reproduce the bug using the provided binary.

gdb indeed can't resolve the address of thread variable without proper `libthread_db`, even with the debug symbols of libc:

image

PR #2871, opened 2025-04-13
Fix "set glibc 2.31" which got broken recently

disconnect3d:
I haven't tracked when it got broken, but we were setting a tuple for a string parameter.

It did not work on 2025.02.19 but worked on 2025.01.20.

I also added a testcase for this behavior now.


disconnect3d:
Thanks to @patryk4815 who found it. This got broken in 304e83182b432a70738a58f5b6bff92c0df10311 but this commit was actually a fix for another issue where we did not set the GDB parameter value but only the internal Pwndbg value. Ugh!

Anyway, this PR/commit here still makes sense!

PR #2873, opened 2025-04-14
Add basic support for s390x

intrigus-lgtm:
Basic support for s390x.

Test binary: [composer-s390x.zip](https://github.com/user-attachments/files/19742263/composer-s390x.zip)

### DIFF
--- a/pwndbg/emu/emulator.py
+++ b/pwndbg/emu/emulator.py
@@ -127,6 +127,10 @@ arch_to_reg_const_map = {
     "s390x": create_reg_to_const_map(arch_to_UC_consts["s390x"]),
 }
 
+# Architectures for which we want to enable virtual TLB mode
+enable_virtual_tlb = {
+    "s390x": True,
+}
 
 # combine the flags with | operator. -1 for all
 (
@@ -226,6 +230,10 @@ class Emulator:
         debug(DEBUG_INIT, "# Instantiating Unicorn for %s", self.arch)
         debug(DEBUG_INIT, "uc = U.Uc(%r, %r)", (arch_to_UC[self.arch], self.uc_mode))
         self.uc = U.Uc(arch_to_UC[self.arch], self.uc_mode)
+
+        if enable_virtual_tlb[self.arch]:
+            debug(DEBUG_INIT, "# Setting TLB mode to virtual")
+            self.uc.ctl_set_tlb_mode(U.UC_TLB_VIRTUAL)
 
         self.regs: pwndbg.lib.regs.RegisterSet = pwndbg.aglib.regs.current
 
--- a/pwndbg/emu/emulator.py
+++ b/pwndbg/emu/emulator.py
@@ -233,7 +233,7 @@ class Emulator:
 
         if enable_virtual_tlb.get(self.arch, False):
             debug(DEBUG_INIT, "# Setting TLB mode to virtual")
-            self.uc.ctl_set_tlb_mode(U.UC_TLB_VIRTUAL)
+            self.uc.ctl_set_tlb_mode(U.UC_TLB_VIRTUAL)  # type: ignore[attr-defined]
 
         self.regs: pwndbg.lib.regs.RegisterSet = pwndbg.aglib.regs.current
 
--- a/pwndbg/emu/emulator.py
+++ b/pwndbg/emu/emulator.py
@@ -231,7 +231,7 @@ class Emulator:
         debug(DEBUG_INIT, "uc = U.Uc(%r, %r)", (arch_to_UC[self.arch], self.uc_mode))
         self.uc = U.Uc(arch_to_UC[self.arch], self.uc_mode)
 
-        if enable_virtual_tlb[self.arch]:
+        if enable_virtual_tlb.get(self.arch, False):
             debug(DEBUG_INIT, "# Setting TLB mode to virtual")
             self.uc.ctl_set_tlb_mode(U.UC_TLB_VIRTUAL)
 
--- a/pwndbg/emu/emulator.py
+++ b/pwndbg/emu/emulator.py
@@ -91,6 +91,7 @@ arch_to_UC = {
     # 'powerpc': U.UC_ARCH_PPC,
     "rv32": U.UC_ARCH_RISCV,
     "rv64": U.UC_ARCH_RISCV,
+    "s390x": U.UC_ARCH_S390X,
 }
 
 # Architecture specific maps: Map<"UC_*_REG_*",constant>
@@ -104,6 +105,7 @@ arch_to_UC_consts = {
     "aarch64": parse_consts(U.arm64_const),
     "rv32": parse_consts(U.riscv_const),
     "rv64": parse_consts(U.riscv_const),
+    "s390x": parse_consts(U.s390x_const),
 }
 
 # Architecture specific maps: Map<reg_name, Unicorn constant>
@@ -122,6 +124,7 @@ arch_to_reg_const_map = {
     ),
     "rv32": create_reg_to_const_map(arch_to_UC_consts["rv32"]),
     "rv64": create_reg_to_const_map(arch_to_UC_consts["rv64"]),
+    "s390x": create_reg_to_const_map(arch_to_UC_consts["s390x"]),
 }
 
 
@@ -592,6 +595,8 @@ class Emulator:
             and "isa32r6" in gdb.newest_frame().architecture().name()
         ):
             mode |= U.UC_MODE_MIPS32R6
+        elif (arch == "s390x"):
+            pass # fails with invalid mode error otherwise
         else:
             mode |= {4: U.UC_MODE_32, 8: U.UC_MODE_64}[pwndbg.aglib.arch.ptrsize]
 
--- a/pwndbg/emu/emulator.py
+++ b/pwndbg/emu/emulator.py
@@ -603,8 +603,8 @@ class Emulator:
             and "isa32r6" in gdb.newest_frame().architecture().name()
         ):
             mode |= U.UC_MODE_MIPS32R6
-        elif (arch == "s390x"):
-            pass # fails with invalid mode error otherwise
+        elif arch == "s390x":
+            pass  # fails with invalid mode error otherwise
         else:
             mode |= {4: U.UC_MODE_32, 8: U.UC_MODE_64}[pwndbg.aglib.arch.ptrsize]
 
--- a/pwndbg/aglib/disasm/__init__.py
+++ b/pwndbg/aglib/disasm/__init__.py
@@ -46,6 +46,7 @@ CapstoneArch = {
     "sparc": CS_ARCH_SPARC,
     "rv32": CS_ARCH_RISCV,
     "rv64": CS_ARCH_RISCV,
+    "s390x": CS_ARCH_SYSZ,
 }
 
 CapstoneEndian = {
@@ -69,6 +70,7 @@ VariableInstructionSizeMax = {
     "mips": 8,
     "rv32": 22,
     "rv64": 22,
+    "s390x": 6,
 }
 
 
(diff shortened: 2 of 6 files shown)


intrigus-lgtm:
Well, I should actually implement unicorn :P

Looking better now, but still not sure whether that's 100% correct.
No indication as to where the instruction will branch to.
grafik
Only on the instruction itself it will show :thinking: 
grafik

OBarronCS:
Nice! I tested this locally - looks like Unicorn is having some trouble executing s390x code. Immediately after starting execution, it is receiving an interrupt (uncomment the line `DEBUG=-1` in the emulator.py),  indicating that some sort of additional setup is necessary to get it running. In the past when I have encountered this, it often means some specific register needs to be set with a good value (like a flags register that changes the CPU behavior).

A good place to start is creating a minimal example of Unicorn running this architecture so it's easier to play around with getting the correct setup.

Here's an example of a simple file that tests MIPS64 in Unicorn that can be used as a base.
```python
#!/usr/bin/env python3
from unicorn import *
from unicorn.mips_const import *

print(unicorn.uc_version())

emu = Uc(UC_ARCH_MIPS, UC_MODE_64 | UC_MODE_BIG_ENDIAN)
# emu.ctl_set_tlb_mode(UC_TLB_VIRTUAL)

# daddu  $gp, $gp, $ra
code = b"\x03\x9f\xe0\x2d" * 2

ADDRESS =  0x1000
ADDRESS = 0x555555d56370 & ~0xFFF

emu.mem_map(ADDRESS, 4 * 1024)
emu.mem_write(ADDRESS, code)

# emu.reg_write(UC_MIPS_REG_CP0_STATUS, 0b000)
#emu.reg_write(UC_MIPS_REG_PC, ADDRESS)
#emu.reg_write(UC_MIPS_REG_GP, 0x1)
#emu.reg_write(UC_MIPS_REG_RA, 0x1)

try:
    emu.emu_start(ADDRESS, 0, count=2)
except Exception as e:
    print(e)

print(f"PC: {hex(emu.reg_read(UC_MIPS_REG_PC))}")
```

patryk4815:
@OBarronCS unicorn support for s390x was "added/tested" today/yesterday 🤔 

```
Lazymio — 07:19
https://github.com/unicorn-engine/unicorn/issues/2124#issuecomment-2800495316

I port unicorn to s390x 🙂
```

OBarronCS:
> @OBarronCS unicorn support for s390x was "added/tested" today/yesterday 🤔
> 
> ```
> Lazymio — 07:19
> https://github.com/unicorn-engine/unicorn/issues/2124#issuecomment-2800495316
> 
> I port unicorn to s390x 🙂
> ```

Oh nice! This might take some while to get into the latest Python pip package release - but testing it locally with the python bindings from the latest dev branch from Unicorn installed would hopefully work, in this case.

intrigus-lgtm:
> @OBarronCS unicorn support for s390x was "added/tested" today/yesterday 🤔
> 
> ```
> Lazymio — 07:19
> https://github.com/unicorn-engine/unicorn/issues/2124#issuecomment-2800495316
> 
> I port unicorn to s390x 🙂
> ```

I _think_ that refers to being able to *run* unicorn on s390x.
Instead of being able to run s390x on x64.

OBarronCS:
I may have found the source of Unicorn not being able to start:

We are writing the 'UC_S390X_REG_PSWM' register to the emulator on startup - if we don't write to this register, it takes on a default value and works fine. However, since we are explicitly writing it when instantiating the emulator, it is taking on some undesired value causing the emulation to not be able to execute instructions. When I was running it, it writes `0x705200180000000` into the register. When it is just 0, it runs just fine.
```python
#!/usr/bin/env python3
from unicorn import *
from unicorn.s390x_const import *

print(unicorn.uc_version())

emu = Uc(UC_ARCH_S390X, UC_MODE_BIG_ENDIAN)

# lghi   %r3, 0
code = b"\xa7\x39\x00\x00" * 2

ADDRESS =  0x1000
# ADDRESS = 0x555555d56370 & ~0xFFF

emu.mem_map(ADDRESS, 4 * 1024)
emu.mem_write(ADDRESS, code)

emu.reg_write(UC_S390X_REG_PC,ADDRESS)
emu.reg_write(UC_S390X_REG_R3, 0x1)
# emu.reg_write(UC_S390X_REG_PSWM, 0x705200180000000)

try:
    emu.emu_start(ADDRESS, 0, count=2)
except Exception as e:
    print(e)


print(f"PC: {hex(emu.reg_read(UC_S390X_REG_PC))}")
print(f"PWSM: {hex(emu.reg_read(UC_S390X_REG_PSWM))}")
```

OBarronCS:
Based on this document - https://www.kernel.org/doc/Documentation/s390/Debugging390.txt - that register changes a lot of the CPU behavior.

OBarronCS:
EDIT: nvm, still debugging the source of the bug

However, this alone doesn't fix it for me - testing further with the script above, if the `ADDRESS` is above some threshold, the emulator throws an exception.....  Looks like addresses above `0x7fffffff` cause an exception?

OBarronCS:
From what I see, there are two issues at play:
- UC_S390X_REG_PSWM changing the CPU behavior. Setting it to `0x705200180000000` (the actual value while debugging the binary linked in the initial PR comment), causes a CPU exception when trying to execute instructions.
- When the register is kept at it's default value, accesses to addresses higher than `0x7fffffff` cause a CPU exception.

patryk4815:
For GDB this works fine.
But for LLDB I have something wrong with registers... 
image
in `register read -a` values looks good
image
^ wrong values wtf

image

```
In [11]: hex(pwndbg.aglib.regs.r14)
Out[11]: '0x4a554341b87f0000'

In [20]: value = r.by_name('r14')

In [21]: hex(int(value))
Out[21]: '0x7fb84143554a'

In [22]: hex(int(value.cast(pwndbg.aglib.typeinfo.unsigned[8])))
Out[22]: '0x4a554341b87f0000'

In [23]: pwndbg.aglib.typeinfo.unsigned[8].inner
Out[23]: unsigned long long
```

Looks like it is bug in LLDB?
Why pointers are byte-reversed?

patryk4815:
Created lldb issue - https://github.com/llvm/llvm-project/issues/135707

wtdcode:
> From what I see, there are two issues at play:
> 
> * UC_S390X_REG_PSWM changing the CPU behavior. Setting it to `0x705200180000000` (the actual value while debugging the binary linked in the initial PR comment), causes a CPU exception when trying to execute instructions.
> * When the register is kept at it's default value, accesses to addresses higher than `0x7fffffff` cause a CPU exception.

Consider `uc.ctl_set_tlb_mode(UC_TLB_VIRTUAL)`

OBarronCS:
> > From what I see, there are two issues at play:
> > 
> > * UC_S390X_REG_PSWM changing the CPU behavior. Setting it to `0x705200180000000` (the actual value while debugging the binary linked in the initial PR comment), causes a CPU exception when trying to execute instructions.
> > * When the register is kept at it's default value, accesses to addresses higher than `0x7fffffff` cause a CPU exception.
> 
> Consider `uc.ctl_set_tlb_mode(UC_TLB_VIRTUAL)`

Setting this works:
- With UC_TLB_VIRTUAL enabled + PSWM copied from the process to the emulator, (`0x705200180000000`), all memory accesses start to work.
- We do need to copy the PSWM register into the process, since certain bits in it enable 64-bit addressing.

@intrigus-lgtm To make emulation work in this PR, let's declare a way to enable this virtual TLB setting for specific arches right after instantiating the Unicorn emulator.
```python
if enable_virtual_tlb[current_arch]:
    uc.ctl_set_tlb_mode(UC_TLB_VIRTUAL)
```

intrigus-lgtm:
Will do.

wtdcode:
> > > From what I see, there are two issues at play:
> > > 
> > > * UC_S390X_REG_PSWM changing the CPU behavior. Setting it to `0x705200180000000` (the actual value while debugging the binary linked in the initial PR comment), causes a CPU exception when trying to execute instructions.
> > > * When the register is kept at it's default value, accesses to addresses higher than `0x7fffffff` cause a CPU exception.
> > 
> > 
> > Consider `uc.ctl_set_tlb_mode(UC_TLB_VIRTUAL)`
> 
> Setting this works:
> 
> * With UC_TLB_VIRTUAL enabled + PSWM copied from the process to the emulator, (`0x705200180000000`), all memory accesses start to work.
> * We do need to copy the PSWM register into the process, since certain bits in it enable 64-bit addressing.
> 
> @intrigus-lgtm To make emulation work in this PR, let's declare a way to enable this virtual TLB setting for specific arches right after instantiating the Unicorn emulator.
> 
> ```python
> if enable_virtual_tlb[current_arch]:
>     uc.ctl_set_tlb_mode(UC_TLB_VIRTUAL)
> ```

I don't have a very detailed look at how pwndbg uses unicorn. However, unless pwndbg plans to setup MMU or something similar, virtual TLB mode is more promising as a default setup, though I have no idea if this introduces big breakages.

intrigus-lgtm:
@wtdcode unicorn is complaining:
> uc.reg_read(23 aka f6)
WARNING: Your register accessing on id 23 is deprecated and will get UC_ERR_ARG in the future release (2.2.0) because the accessing is either no-op or not defined. If you believe the register should be implemented or there is a bug, please submit an issue to https://github.com/unicorn-engine/unicorn. Set UC_IGNORE_REG_BREAK=1 to ignore this warning.

Are floating-point registers simply not implemented?
(Looking at https://github.com/unicorn-engine/unicorn/blob/f8c6db950420d2498700245269d0b647697c5666/qemu/target/s390x/unicorn.c#L57-L85 that seems to be the case?)

wtdcode:
> @wtdcode unicorn is complaining:
> 
> > uc.reg_read(23 aka f6)
> > WARNING: Your register accessing on id 23 is deprecated and will get UC_ERR_ARG in the future release (2.2.0) because the accessing is either no-op or not defined. If you believe the register should be implemented or there is a bug, please submit an issue to https://github.com/unicorn-engine/unicorn. Set UC_IGNORE_REG_BREAK=1 to ignore this warning.
> 
> Are floating-point registers simply not implemented? (Looking at https://github.com/unicorn-engine/unicorn/blob/f8c6db950420d2498700245269d0b647697c5666/qemu/target/s390x/unicorn.c#L57-L85 that seems to be the case?)

Yeah, definitely a bug. Could you post this to unicorn issue too? I will address it recently.

intrigus-lgtm:
Done: https://github.com/unicorn-engine/unicorn/issues/2166

wtdcode:
I assure you this will be in 2.2.0

wtdcode:
https://github.com/unicorn-engine/unicorn/pull/2167 has initial implementation.

intrigus-lgtm:
I will fix the failing tests tomorrow.

OBarronCS:
Awesome! Emulation working great! 
image

pwndbg/emu/emulator.py
```diff
@@ -224,6 +231,10 @@ def __init__(self) -> None:
         debug(DEBUG_INIT, "uc = U.Uc(%r, %r)", (arch_to_UC[self.arch], self.uc_mode))
         self.uc = U.Uc(arch_to_UC[self.arch], self.uc_mode)
 
+        if enable_virtual_tlb[self.arch]:
```
patryk4815:
```suggestion
        if enable_virtual_tlb.get(self.arch, False):
```

pwndbg/emu/emulator.py
```diff
@@ -592,6 +603,8 @@ def get_uc_mode(self):
             and "isa32r6" in gdb.newest_frame().architecture().name()
         ):
             mode |= U.UC_MODE_MIPS32R6
+        elif (arch == "s390x"):
+            pass # fails with invalid mode error otherwise
```
patryk4815:
```suggestion
        elif arch == "s390x":
            pass  # fails with invalid mode error otherwise
```

pwndbg/emu/emulator.py
```diff
@@ -224,6 +231,10 @@ def __init__(self) -> None:
         debug(DEBUG_INIT, "uc = U.Uc(%r, %r)", (arch_to_UC[self.arch], self.uc_mode))
         self.uc = U.Uc(arch_to_UC[self.arch], self.uc_mode)
 
+        if enable_virtual_tlb.get(self.arch, False):
+            debug(DEBUG_INIT, "# Setting TLB mode to virtual")
+            self.uc.ctl_set_tlb_mode(U.UC_TLB_VIRTUAL)
```
patryk4815:
```suggestion
            self.uc.ctl_set_tlb_mode(U.UC_TLB_VIRTUAL)  # type: ignore[attr-defined]
```

intrigus-lgtm:
Thank you @patryk4815 for fixing the remaining problems.

PR #2876, opened 2025-04-15
Handle the case where GDB fails to resolve certain heap-related symbols without proper `libthread_db`

lebr0nli:
gdb might fail to resolve the address of thread variable without proper `libthread_db`, even with the debug symbols of libc:

image

And this might cause some wrong result which can be reproduce by the case mentioned in #2868:

> 433133397-f5f545d9-e6cb-473c-ae91-60b60ab4d28f

We should consider fallback to the heuristic instead of showing the wrong result (`... hasn't performed any allocations`):

image
image

And I moved the hint about `libthread-db-search-path` to the result of `help set resolve-heap-via-heuristic`:

image

But the output of the `help` is kinda bloated(?), I'm not sure if it's a good idea. Let me know if there's a cleaner or more appropriate approach.


patryk4815:
Btw, in lldb we had to manually implement resolving TLS symbols  

image

k4lizen:
> But the output of the help is kinda bloated(?), I'm not sure if it's a good idea. Let me know if there's a cleaner or more appropriate approach.

I don't think it's a problem if there's too much text. Maybe you could format it like this:
````
In addition, even you have the debug symbols of libc, you might still see the following warning
when debugging a multi-threaded program:
```
warning: Unable to find libthread_db matching inferior's thread library, thread
debugging will not be available.
```
You'll need to ensure that the correct `libthread_db.so` is loaded. To do this, set the search path
using: `set libthread-db-search-path <path having correct libthread_db.so>`. Then, restart your
program to enable proper thread debugging.
````
And the backticks will render nicely in markdown in docs.

Maybe we want to add some sort of logic in the future to ensure some line-wrapping when displaying the help, rather than hoping the author of the help hit the right width.

lebr0nli:
> Maybe you could format it like this:

Thanks for the feedback! I've updated the help text to include line-wrapping and used backticks for proper markdown formatting.

disconnect3d (approved):

PR #2879, opened 2025-04-16
Small Refactor to DisassemblyAssistant

OBarronCS:
This is a small refactor to how DisassemblyAssistants are created and referenced. Previously, these assistants were global singleton objects. This made it hard to share context and information between instructions that were disassembled back-to-back in a single logical pass. It is now easier to share this sort of information by adding attributes to the DisassemblyAssistant. No new features are added to take advantage of this in this PR.


pwndbg/aglib/disasm/aarch64.py
```diff
@@ -469,4 +470,4 @@ def _parse_register(
         return target
 
 
-assistant = DisassemblyAssistant("aarch64")
+register_disassembly_assistant("aarch64", lambda: DisassemblyAssistant("aarch64"))
```
OBarronCS: Would call this a factory pattern(?) - call the lambda function to create a new assistant for this architecture, as the constructors parameters differ and are specific to each architecture's implementation.

pwndbg/aglib/disasm/arch.py
```diff
@@ -134,17 +135,39 @@ def memory_or_register_assign(left: str, right: str, mem_assign: bool) -> str:
     return memory_assign(left, right) if mem_assign else register_assign(left, right)
 
 
+# A registry of functions that create disassembly assistants for each architecture
+ALL_DISASSEMBLY_ASSISTANTS: Dict[str, Callable[[], DisassemblyAssistant]] = {}
```
disconnect3d: since the value is a callable anyway, can't we just add all in here instead of having the `register_*` function + calling it in separate files?

pwndbg/aglib/disasm/arch.py
```diff
@@ -134,17 +135,39 @@ def memory_or_register_assign(left: str, right: str, mem_assign: bool) -> str:
...
+
+
+def register_disassembly_assistant(
+    arch: PWNDBG_SUPPORTED_ARCHITECTURES_TYPE, assistant: Callable[[], DisassemblyAssistant]
+):
+    ALL_DISASSEMBLY_ASSISTANTS[arch] = assistant
```
disconnect3d: If we keep this func, the `assistant` can be renamed to `create_assistant` or `new_assistant` or `get_assistant`?

pwndbg/aglib/disasm/arch.py
OBarronCS: Very good point - just ran into this issue today as the `register_*` function is called as a side-effect of importing each individual file that implements the DisassemblyAssistant, and if you forgot to add the module to a global list of modules to import it doesn't register it.

I'll make this change.

OBarronCS:
I updated it so all the assistants are registered in a single list as mentioned in a comment.

In order to do, it required moving around a file to avoid circular imports (`disasm/__init__.py` was otherwise imported when not wanted when doing `import pwndbg.aglib.disasm.something)`, so `disasm/__init__.py` was moved to `disassembly.py`

Don't know whats happening with the lint errors 
image

pwndbg/aglib/disasm/disassembly.py
```diff
@@ -0,0 +1,566 @@
...
+    #
+    # This helps with infinite loops and RET sleds.
+
+    while insns and len(insns) > 2 and insns[-3].address == insns[-2].address == insns[-1].address:
+        del insns[-1]
+
+    return (insns, index_of_current_instruction)
+
+
+ALL_DISASSEMBLY_ASSISTANTS: Dict[
```
OBarronCS, line 541: New contents in this file starts here. This file was renamed from __init__.py to avoid circular import issues.

pwndbg/aglib/disasm/disassembly.py
```diff
@@ -0,0 +1,566 @@
...
+    PWNDBG_SUPPORTED_ARCHITECTURES_TYPE, Callable[[], DisassemblyAssistant]
+] = {
+    "aarch64": lambda: pwndbg.aglib.disasm.aarch64.AArch64DisassemblyAssistant("aarch64"),
+    "i386": lambda: pwndbg.aglib.disasm.x86.X86DisassemblyAssistant("i386"),
```
disconnect3d, line 545: Fwiw some of those assistants don't need to take in an arch name argument

disconnect3d (approved):

PR #2885, opened 2025-04-16
Add capstone for loongarch64

patryk4815:


pwndbg/aglib/disasm/__init__.py
```diff
@@ -71,6 +72,7 @@
     "rv32": 22,
     "rv64": 22,
     "s390x": 6,
+    "loongarch64": 4,
```
patryk4815: It should be `8`?

OBarronCS: I think 4 is right - that's the maximum length of instructions in Loongarch64 (unless there is some extensions I'm not aware of). https://blog.xen0n.name/en/posts/tinkering/loongarch-faq/#:~:text=has%2032%2Dbit%20fixed%2Dlength%20instruction%20words.

OBarronCS:
Awesome! Now the disassembly for Loongarch looks great + branches are followed

image

Issue #2883, opened 2025-04-16
binary ninja integration OverflowError


k4lizen:
### Description

I was testing the binary ninja integration and got this exception:
```
pwndbg> start
pwndbg> start
(yes, twice)
pwndbg> set integration-provider binja
Pwndbg successfully connected to Binary Ninja (4.2.6455 Personal) xmlrpc: http://127.0.0.1:31337
Set which provider to use for integration features to 'binja'.
pwndbg> bns
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ ~/pwndbg/pwndbg/commands/__init__.py:195 in __call__                      │
│                                                                                                  │
│   192 │                                                                                          │
│   193 │   def __call__(self, *args: Any, **kwargs: Any) -> str | None:                           │
│   194 │   │   try:                                                                               │
│ ❱ 195 │   │   │   return self.function(*args, **kwargs)                                          │
│   196 │   │   except TypeError:                                                                  │
│   197 │   │   │   print(f"{self.function.__name__.strip()!r}: {self.function.__doc__.strip()}"   │
│   198 │   │   │   pwndbg.exception.handle(self.function.__name__)                                │
│                                                                                                  │
│ ~/pwndbg/pwndbg/commands/__init__.py:407 in _OnlyWhenRunning              │
│                                                                                                  │
│   404 │   def _OnlyWhenRunning(*a: P.args, **kw: P.kwargs) -> Optional[T]:                       │
│   405 │   │   # TODO: Properly support OnlyWhenRunning without `gdblib`.                         │
│   406 │   │   if pwndbg.aglib.proc.alive:                                                        │
│ ❱ 407 │   │   │   return function(*a, **kw)                                                      │
│   408 │   │   else:                                                                              │
│   409 │   │   │   log.error(f"{func_name(function)}: The program is not being run.")             │
│   410 │   │   │   return None                                                                    │
│                                                                                                  │
│ ~/pwndbg/pwndbg/commands/binja.py:20 in bn_sync                           │
│                                                                                                  │
│   17 │   """                                                                                     │
│   18 │   Synchronize Binary Ninja's cursor with GDB                                              │
│   19 │   """                                                                                     │
│ ❱ 20 │   pwndbg.integration.binja.navigate_to(pwndbg.aglib.regs.pc)                              │
│   21                                                                                             │
│                                                                                                  │
│ ~/pwndbg/pwndbg/integration/binja.py:169 in wrapper                       │
[...]

PR #2896, opened 2025-04-17
Make Binary Ninja integration not dependent on IDA integration to work

Aplet123:
The current binja integration depends on the IDA integration to inject a custom int marshaller into the xmlrpc client, which can cause the binja integration to error if the IDA integration doesn't get run. This PR makes it so both integrations will inject a large `MAXINT` and `MININT` so that they're not dependent on each other.

### DIFF
--- a/pwndbg/integration/binja.py
+++ b/pwndbg/integration/binja.py
@@ -48,6 +48,9 @@ from pwndbg.dbg import StopPoint
 from pwndbg.lib.functions import Argument
 from pwndbg.lib.functions import Function
 
+xmlrpc.client.MAXINT = 10**100  # type: ignore[misc]
+xmlrpc.client.MININT = -(10**100)  # type: ignore[misc]
+
 bn_rpc_host = pwndbg.config.add_param(
     "bn-rpc-host", "127.0.0.1", "Binary Ninja XML-RPC server host"
 )
--- a/pwndbg/integration/binja.py
+++ b/pwndbg/integration/binja.py
@@ -189,17 +189,11 @@ def can_connect() -> bool:
 
 
 def l2r(addr: int) -> int:
-    exe = pwndbg.aglib.elf.exe()
-    if not exe:
-        raise Exception("Can't find EXE base")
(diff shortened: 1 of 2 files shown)


Aplet123:
Partially fixes #2883

pwndbg/integration/binja.py
```diff
@@ -189,17 +192,11 @@ def can_connect() -> bool:
 
 
 def l2r(addr: int) -> int:
-    exe = pwndbg.aglib.elf.exe()
```
patryk4815, line 192: why removed?

Aplet123: The call to `.exe()` used to be there because I used it to get the exe base, but since the code uses `pwndbg.aglib.proc.binary_base_addr` now there's no point in the extra function call.

pwndbg/integration/ida.py
```diff
@@ -42,7 +42,8 @@
 ida_rpc_port = pwndbg.config.add_param("ida-rpc-port", 31337, "ida xmlrpc server port")
 ida_timeout = pwndbg.config.add_param("ida-timeout", 2, "time to wait for ida xmlrpc in seconds")
 
-xmlrpc.client.Marshaller.dispatch[int] = lambda _, v, w: w("<value><i8>%d</i8></value>" % v)
+xmlrpc.client.MAXINT = 10**100  # type: ignore[misc]
+xmlrpc.client.MININT = -(10**100)  # type: ignore[misc]
```
patryk4815: can we move it into `init_ida_rpc_client` func?

pwndbg/integration/binja.py
```diff
@@ -48,6 +48,9 @@
 from pwndbg.lib.functions import Argument
 from pwndbg.lib.functions import Function
 
+xmlrpc.client.MAXINT = 10**100  # type: ignore[misc]
+xmlrpc.client.MININT = -(10**100)  # type: ignore[misc]
```
patryk4815: can we move it into `init_bn_rpc_client`?

k4lizen:
Thanks! This does seem to fix the issue for me :p

PR #2898, opened 2025-04-18
Context display fix involving unknown syscall ABI's + MIPS Disassembly CapstoneV6 Hotfix

OBarronCS:
1. Adds a try/catch block to a place that fetches the syscall ABI, which might return a KeyError - this results in a crash when encountering a syscall instruction in the disasm context view for an arch where the syscall ABI is not defined.
- This is fixed in #2771, but before that is merged I wanted to get a fix in

2. We just updated to Capstone V6, which changed how to specify the disassembler mode for MIPS - now, we have to be much more precise about the specific ABI version being used. Adds a quick check to select the correct constant to pass to Capstone to correctly disassemble instructions.

### DIFF
--- a/pwndbg/aglib/disasm/disassembly.py
+++ b/pwndbg/aglib/disasm/disassembly.py
@@ -499,6 +499,11 @@ def near(
                     emu.valid = False
 
                 split_insn = one(insn.address + insn.size, None, put_cache=True)
+
+                # There might not be a valid instruction at the branch delay slot
+                if split_insn is None:
+                    break
+
                 insns.append(split_insn)
 
                 # Manually make the backtracing cache correct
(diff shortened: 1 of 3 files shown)


pwndbg/aglib/disasm/disassembly.py
```diff
@@ -181,12 +181,35 @@ def get_disassembler(address):
...
+        if pwndbg.dbg.is_gdblib_available():
+            # Example: "mips:isa64r2"
+            raw_arch_name = gdb.newest_frame().architecture().name()
+
+            if "isa32r2" in raw_arch_name:
+                extra = CS_MODE_MIPS32R2
+            elif "isa32r3" in raw_arch_name:
+                extra = CS_MODE_MIPS32R3
+            elif "isa32r5" in raw_arch_name:
+                extra = CS_MODE_MIPS32R5
```
OBarronCS, line 194: Please excuse this ugly code block - with #2771 this will be much nicer, just wanted to get something in to correctly instantiate the disassembler.

PR #2900, opened 2025-04-18
Improve convenience function helps and autogenerate documentation

k4lizen:
See the docs in gdb with `help function function_name`, and check out the site with `./scripts/docs-live.sh`.

Moved gsbase and fsbase to gdblib/functions.py


k4lizen:
Can anyone reproduce the CI gdb crash on their machine?

k4lizen:
Okay I reproduced it on an ubuntu machine, now need to figure out what is going on.

k4lizen:
Okay so the bug is (I guess) in gdb. When python exits eventually this gets called (via glibc `__run_exit_handlers`):
```c++
gdbpy_tui_window_maker::~gdbpy_tui_window_maker ()
{
  /* Remove this gdbpy_tui_window_maker from the global list.  */
  if (is_linked ())
    m_window_maker_list.erase (m_window_maker_list.iterator_to (*this));

  if (m_constr != nullptr)
    {
      gdbpy_enter enter_py;
      m_constr.reset (nullptr);
    }
}
```
In the gdb version that ubuntu 24.04 has, `m_constr` isn't null so the gdbpy_enter constructor is ran, which results in:
```
gdbpy_enter::gdbpy_enter
set_active_ext_lang
check_quit_flag ()
gdbpy_check_quit_flag
gdbpy_gil::gdbpy_gil
```
which invokes `PyGILState_Ensure () from /lib/x86_64-linux-gnu/libpython3.12.so.1.0`
where libpython gets to this point: https://github.com/python/cpython/blob/2b47f46d7dc30d27b2486991fea4acd83553294b/Python/pystate.c#L2832

except the `runtime` variable is null so we segfault because of a nullptr dereference.

Now, this has been fixed in this commit:
```
commit dfba4847f6a22eef700917e2718ea847d9f5cc9d
Author: Tom de Vries <tdevries@suse.de>
Date:   Wed Sep 25 19:29:57 2024 +0200
 
    [gdb/python] Make sure python sys.exit makes gdb exit
    
    With gdb 15.1, python sys.exit no longer makes gdb exit:
    ...
    $ gdb -q -batch -ex "python sys.exit(2)" -ex "print 123"; echo $?
    Python Exception <class 'SystemExit'>: 2
    Error occurred in Python: 2
    $1 = 123
    0
    ...
    
    This is a change in behaviour since commit a207f6b3a38 ("Rewrite "python"
    command exception handling"), first available in gdb 15.1.
    
    This patch reverts to the old behaviour by handling PyExc_SystemExit in
    gdbpy_handle_exception, such what we have instead:
    ...
    $ gdb -q -batch -ex "python sys.exit(2)" -ex "print 123"; echo $?
    2
    ...
    
    Tested on x86_64-linux, with python 3.6 and 3.13.
    
    Tested-By: Guinevere Larsen <blarsen@redhat.com>
    Approved-By: Tom Tromey <tom@tromey.com>
    
    PR python/31946
    Bug: https://sourceware.org/bugzilla/show_bug.cgi?id=31946
```
Except it's more of a workaround, the segfault isn't mentioned anywhere and it was probably never noticed. It is quite possible that there is some buggy code still somewhere (not sure if gdb or cpython is at fault, maybe both), except it's unreachable because of this patch.

This commit has been cherry-picked to the `gdb-15-branch` so the problem should at least not appear in gdb 15, however ubuntu 24.04 (LTS btw) doesn't actually use the `gdb-15-branch` branch to package gdb, so the problem exists in the `ubuntu-latest` runner. Btw, `ubuntu-latest` is I think not "latest" but "latest LTS", since the actual latest interim release is 25.04.

k4lizen:
Also, there is a bug in the verifier script, since it is exiting (which is needed to fail the CI) because it detects a difference in docs it generates and the ones in the branch, when there shouldn't be any. I can see this difference on ubuntu but not on artix so I guess some dependency is doing something different so I'll figure that out.

After that is fixed, gdb only has a potential to segfault when the docs are actually not updated, so at least it is consistent - CI will never fail when it was supposed to pass. But I will probably look to file a bug (?) report to ubuntu to update their gdb package.

patryk4815:
@k4lizen it is ready to merge? This GDB segfault is irrelevant, just we have to see few lines above it. (VERIFICATION FAILED)
image

k4lizen:
Not yet, I will be ready when verification doesn't fail in the first place. Should probably be done tonight.

patryk4815:
@k4lizen 
image

k4lizen:
LMAO. okay, good to know

k4lizen:
Mann
```python
def my_func():
    """
    AAAA
    BBBB
    """
    pass

print(repr(my_func.__doc__))
```
Older (ubuntu) versions of python give:
`'\n    AAAA\n    BBBB\n'`
while newer give:
`'\nAAAA\nBBBB\n'`

k4lizen:
Should be good now.

I left one instance of `os._exit()` in the code since it's around deadlock prevention stuff, so just in case.

disconnect3d (approved):

PR #2919, opened 2025-04-24
Remove shell commands and cleanup command parsing

k4lizen:
Breaking: Removed shell commands (like `ls`, `unhex` etc.), they have been deprecated for some time and were the only thing depending on non-argparsed command logic.

Otherwise, there should be no functional changes, the code should just be much more readable.

Like, before it was doing `functools.update_wrapper(self, function)` to update the Command classes \_\_doc\_\_ and \_\_name\_\_ to match the command function. But the command function's doc is not the autorithative description, it's the one passed by argparse, so there was a `function.__doc__ = doc` just for that reason. Oh and also obviously a command argument may not have the same name as the function so `self.__name__ = command_name`...

Renamed the classes, I'm not particularly fond of the new names but "ArgparsedCommand" doesn't fit since now all commands are "argparsed".

The `Command` class (previously `ArgparsedCommand`)  now only sets up the `CommandObj` class (previously `Command`), which performs all input sanitization and initialization (including what `_ArgparsedCommand` used to do).


pwndbg/commands/__init__.py
```diff
@@ -71,8 +71,21 @@ class CommandCategory(str, Enum):
...
-class Command:
-    """Generic command wrapper"""
+class InvalidDebuggerError(Exception):
+    """
+    Raised when a command is called in a debugger for which
+    it is disallowed.
+    """
+
+    def __init__(self, message: str):
+        super().__init__(message)
```
disconnect3d: Is this even needed? Can't you just do `class A(Exception): pass`?

patryk4815:
Commands: "constgrep", "disasm", "pwn", "unhex"
are removed without deprecation warning

k4lizen:
@patryk4815 that is true but they are deprecated on the pwntools side, if you run the command youll see a deprecation notice

patryk4815:
No idea if we should change name `ArgparsedCommand ` to `Command`, because there are other project that use that decorator, eg:
- https://github.com/scwuaptx/Pwngdb/tree/master/pwndbg
- https://github.com/two-heart/dbgtools/tree/main/dbgtools/commands
- https://github.com/wes4m/vheap/blob/master/vheap.py

@disconnect3d ^

patryk4815:
Or we should provide "compat script for backward compatibility layer" for out-of-tree users.
Like out-of-tree users can do: 
```py
import pwndbg.compat
pwndbg.compat.set_compatibility_with("2025.01.29")
```
🤔 

```py
def deprecated(func):
    @functools.wraps(func)
    def new_func(*args, **kwargs):
        warnings.warn(
            f"Function {func.__name__} is deprecated and will be removed in the 2025.12 pwndbg release cycle.",
            category=DeprecationWarning,
            stacklevel=2
        )
        return func(*args, **kwargs)
    return new_func

@deprecated
def ArgparsedCommand(*args, **kwargs):
    return pwndbg.commads.Command(*args, **kwargs)
```
May we should consider doing this is future?

I mean, there have already been other PRs that introduced breaking changes, like changes in function names and import paths, or moving `gdblib` -> `aglib`

pwndbg/commands/__init__.py
```diff
@@ -88,82 +101,133 @@ class Command:
...
+    def initialize_parser(self):
+        # Set parser.prog so the help is generated properly.
+        self.parser.prog = self.command_name
 
-        self.repeat = False
+        # Generate command help
+        file = io.StringIO()
+        self.parser.print_help(file)
+        file.seek(0)
+        self.help_str = file.read()
```
disconnect3d: With `io.StringIO()` its best to do: `with io.StringIO() as file: ...` to close the file eventually.

But... we probably don't need to do all that. Here is an excerpt from IPython session:
```py
In [15]: a=ArgumentParser()

In [16]: a.print_help??
Signature: a.print_help(file=None)
Docstring: <no docstring>
Source:
    def print_help(self, file=None):
        if file is None:
            file = _sys.stdout
        self._print_message(self.format_help(), file)
File:      /usr/local/Cellar/python@3.12/3.12.7/Frameworks/Python.framework/Versions/3.12/lib/python3.12/argparse.py
Type:      method
```

We can see that the `ArgumentParser.print_help(file)` method does `self.format_help()` under the hood. It seems we should be able to use just that instead of all the logic here?

```py
In [17]: a.format_help()
Out[17]: 'usage: ipython3 [-h]\n\noptions:\n  -h, --help  show this help message and exit\n'
```

pwndbg/commands/__init__.py
```diff
@@ -88,82 +101,133 @@ class Command:
...
+        # FIXME: this is weird
+        for action in self.parser._actions:
+            if isinstance(action, argparse._SubParsersAction):
+                action.type = str
+            if action.dest == "help":
+                continue
+            if action.type is int:
+                action.type = fix_int_reraise_arg
+            if action.type is None:
+                action.type = fix_reraise_arg
```
disconnect3d, line 184: Should the `action.dest == "help"` be the first if condition here?

Also shouldn't some of those be `elif`?

pwndbg/commands/__init__.py
```diff
@@ -88,82 +101,133 @@ class Command:
...
+            and self.parser.description.strip()
+            and "A command must contain a description."
+        )
+        self.description = self.parser.description.strip()
 
-        This is only used by pwndbg/commands/shell.py which is deprecated.
-        Usually _ArgparsedCommand.split_args is called.
+        # We want to run all integer and otherwise-unspecified arguments
+        # through fix() so that GDB parses it.
+        # FIXME: this is weird
```
disconnect3d, line 175: Just to be sure is this logic what was already there before? Or is it all new now?

pwndbg/commands/__init__.py
```diff
@@ -88,82 +101,133 @@ class Command:
...
+            # argparse complained about incorrect usage or printed
+            # help and exited. Either way the appropriate message
+            # is already printed and we shouldn't call the function.
+            return
+
         try:
             self.repeat = self.check_repeated(argument, from_tty)
-            self(*args, **kwargs)
+            # Call this object, goes to __call__
+            self(**kwargs)
```
disconnect3d: For what is worth... not that it matters anyhow, but `self.__call__` is 2x faster in latest Python if called lots of times. When its called just once, it can be 25% faster :D.

...its all nanoseconds, so nothing we care about, but yeah :P.

```py
➜  bin i
Python 3.12.7 (main, Oct  1 2024, 02:05:46) [Clang 15.0.0 (clang-1500.3.9.4)]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.22.1 -- An enhanced Interactive Python. Type '?' for help.
Imported: os, sys, json, ctypes, re, subprocess, time, collections, binascii, string, dis, zlib, random, math
Shortcuts: dt, md5, sha256, b64d, b64e, r(ead), w(rite), j(json.loads), cint

In [1]: class A:
   ...:     def foo(self):
   ...:         self()
   ...:     def __call__(self):
   ...:         pass
   ...:

In [2]: class B:
   ...:     def foo(self):
   ...:         self.__call__()
   ...:     def __call__(self):
   ...:         pass
   ...:

In [3]: a, b = A(), B()

In [4]: afoo, bfoo = a.foo, b.foo

In [5]: %timeit afoo()
100 ns ± 1.17 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [6]: %timeit bfoo()
55.1 ns ± 1.51 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [7]: dis.dis(afoo, show_caches=1, adaptive=1)
  2           0 RESUME                   0

  3           2 PUSH_NULL
              4 LOAD_FAST                0 (self)
              6 CALL                     0
              8 CACHE                    0 (counter: 42204)
             10 CACHE                    0 (func_version: 0)
             12 CACHE                    0
             14 POP_TOP
             16 RETURN_CONST             0 (None)

In [8]: dis.dis(bfoo, show_caches=1, adaptive=1)
  2           0 RESUME                   0

  3           2 LOAD_FAST                0 (self)
              4 LOAD_ATTR_METHOD_WITH_VALUES     1 (NULL|self + __call__)
              6 CACHE                    0 (counter: 832)
              8 CACHE                    0 (version: 133861)
             10 CACHE                    0
             12 CACHE                    0 (keys_version: 554)
             14 CACHE                    0
             16 CACHE                    0 (descr: 4415530432)
             18 CACHE                    0
             20 CACHE                    0
             22 CACHE                    0
             24 CALL_PY_EXACT_ARGS       0
             26 CACHE                    0 (counter: 832)
             28 CACHE                    0 (func_version: 17461)
             30 CACHE                    0
             32 POP_TOP
             34 RETURN_CONST             0 (None)

In [9]:
```

And when no hot-loop is involved:
```py
In [1]: class A:
   ...:     def foo(self):
   ...:         self()
   ...:     def __call__(self):
   ...:         pass
   ...:

In [2]: class B:
   ...:     def foo(self):
   ...:         self.__call__()
   ...:     def __call__(self):
   ...:         pass
   ...:

In [3]: a, b = A(), B()

In [4]: afoo, bfoo = a.foo, b.foo

In [5]: dis.dis(afoo, show_caches=1, adaptive=1)
  2           0 RESUME                   0

  3           2 PUSH_NULL
              4 LOAD_FAST                0 (self)
              6 CALL                     0
              8 CACHE                    0 (counter: 17)
             10 CACHE                    0 (func_version: 0)
             12 CACHE                    0
             14 POP_TOP
             16 RETURN_CONST             0 (None)

In [6]: dis.dis(bfoo, show_caches=1, adaptive=1)
  2           0 RESUME                   0

  3           2 LOAD_FAST                0 (self)
              4 LOAD_ATTR                1 (NULL|self + __call__)
              6 CACHE                    0 (counter: 17)
              8 CACHE                    0 (version: 0)
             10 CACHE                    0
             12 CACHE                    0 (keys_version: 0)
             14 CACHE                    0
             16 CACHE                    0 (descr: 0)
             18 CACHE                    0
             20 CACHE                    0
             22 CACHE                    0
             24 CALL                     0
             26 CACHE                    0 (counter: 17)
             28 CACHE                    0 (func_version: 0)
             30 CACHE                    0
             32 POP_TOP
             34 RETURN_CONST             0 (None)

In [7]: %timeit?

In [8]: %timeit -n1 afoo()
The slowest run took 6.95 times longer than the fastest. This could mean that an intermediate result is being cached.
460 ns ± 393 ns per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [9]: %timeit -n1 bfoo()
The slowest run took 7.65 times longer than the fastest. This could mean that an intermediate result is being cached.
384 ns ± 339 ns per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [10]: %timeit -n1 bfoo()
305 ns ± 201 ns per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [11]: %timeit -n1 afoo()
The slowest run took 6.64 times longer than the fastest. This could mean that an intermediate result is being cached.
427 ns ± 364 ns per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

pwndbg/commands/__init__.py
```diff
@@ -194,13 +258,93 @@ def __call__(self, *args: Any, **kwargs: Any) -> str | None:
...
+            )
+        # Since CommandObj has __call__ defined, an instance of it is a
+        # callable object (which essentially decorates the function).
+        return CommandObj(
+            function,
+            self.parser,
+            self.command_name,
+            self.category,
+            False,
+            self.aliases,
```
disconnect3d, line 343: Random thought: since we pass aliases there anyway, wouldn't it be better for the `CommandObj` to call `self.register_command(command_name)` for command name + each of aliases instead of us re-creating the whole object for each alias?

pwndbg/commands/misc.py
```diff
@@ -96,29 +92,17 @@ def errno_(err) -> None:
...
-@pwndbg.commands.ArgparsedCommand(parser, command_name="pwndbg", category=CommandCategory.PWNDBG)
-def pwndbg_(filter_pattern, shell, all_, category_, list_categories) -> None:
+@pwndbg.commands.Command(parser, command_name="pwndbg", category=CommandCategory.PWNDBG)
+def pwndbg_(filter_pattern, category_, list_categories) -> None:
     if list_categories:
         for category in CommandCategory:
             print(C.bold(C.green(f"{category.value}")))
         return
 
-    if all_:
```
disconnect3d, line 108: Just for future: ideally this should be in another PR, but we can leave it in this one.

pwndbg/commands/misc.py
```diff
@@ -142,37 +126,29 @@ def pwndbg_(filter_pattern, shell, all_, category_, list_categories) -> None:
         print()
 
 
-def list_and_filter_commands(filter_str, pwndbg_cmds=True, shell_cmds=False):
+def list_and_filter_commands(filter_str):
     sorted_commands = list(pwndbg.commands.commands)
-    sorted_commands.sort(key=lambda x: x.__name__)
+    sorted_commands.sort(key=lambda x: x.command_name)
```
disconnect3d: Should be just:
```py
sorted_commands = sorted(
    pwndbg.commands.commands, key=lambda c: c.command_name
)
```

pwndbg/commands/misc.py
```diff
@@ -142,37 +126,29 @@ def pwndbg_(filter_pattern, shell, all_, category_, list_categories) -> None:
...
+        desc = c.description
 
-        if docs:
-            docs = docs.strip()
-        if docs:
-            docs = docs.splitlines()[0]
+        if desc:
+            desc = desc.strip()
+        if desc:
+            desc = desc.splitlines()[0]
```
disconnect3d: Could probably be jut one if?
```py
if desc:
    desc = desc.strip().splitlines()[0]
```

disconnect3d (commented):
See comments

disconnect3d:
> No idea if we should change name `ArgparsedCommand ` to `Command`, because there are other project that use that decorator, eg:
> 
> * https://github.com/scwuaptx/Pwngdb/tree/master/pwndbg
> * https://github.com/two-heart/dbgtools/tree/main/dbgtools/commands
> * https://github.com/wes4m/vheap/blob/master/vheap.py
> 
> @disconnect3d ^

There are couple of solutions:
1. Just change it and inform those projects that we made this change.
2. Provide `ArgparsedCommand = Command`, heh. Alternatively, inherit from it and provide a one-time deprecation warning. And remove this completely in half a year or a year.
3. Stay with `ArgparsedCommand`

disconnect3d:
(Also this now has conflicts due to some other PR that was merged, sorry ;<)

pwndbg/commands/shell.py
```diff
@@ -1,94 +0,0 @@
...
-shellcmds = list(filter(which, shellcmd_names))
-
-
-def register_shell_function(cmd, deprecated=False) -> None:
-    def handler(*a) -> None:
-        if os.fork() == 0:
-            os.execvp(cmd, (cmd,) + a)
-        os.wait()
-        print(
-            f"This command is deprecated in Pwndbg. Please use the GDB's built-in syntax for running shell commands instead: !{cmd} <args>"
```
OBarronCS, line 77: I discovered the `!` feature of GDB through this message. Perhaps we could add something similar to the "tip of the day" feature.

Something like
```
Use "!" before a command to run it in the system shell, like `!ls` or `!bash`
```

disconnect3d: Pls add a tip in another PR

pwndbg/commands/__init__.py
k4lizen: oh yup that makes sense, those lines are from before so I didn't think to question them

pwndbg/commands/__init__.py
k4lizen: it was already there

pwndbg/commands/__init__.py
k4lizen: huh okay interesting. I also recently read a post about `if some_list:` being faster than `if len(some_list) != 0:`. I wonder if these cases could be optimized by cpython or if there's some theoritical problem/difference at hand.

disconnect3d: There is: `len` is a global lookup ... it can be monkey patched etc. 💩

k4lizen: maybe cpython can check/prove that it wasn't? but yeah, seems non-trivial from that standpoint

disconnect3d: I bet they may have a bytecode specialization for the case of `len() != 0` for the `POP_JUMP_IF_FALSE` opcode (like: `POP_JUMP_IF_FALSE_LIST_OR_TUPLE`), but they don't have it right now - or at least not in Py 3.12:

image

disconnect3d: On that topic, if u haven't heard about Python specialized bytecode, I gave a talk about it in the past ([video](https://www.youtube.com/watch?v=RlNM5n5C_wg&t=5034s), [slides](https://docs.google.com/presentation/d/13ZiJPzQrNVC5azJPlryfrPtHulCvohts0rZJFepTmis))

k4lizen: hehe okay I'll give it a watch later

k4lizen:
> (Also this now has conflicts due to some other PR that was merged, sorry ;<)

no worries, I expected it

k4lizen:
I think that now `cmd -h` and `help cmd` have the exact same output in all cases, we could add a test for that.

k4lizen:
btw the `self.parser.formatter_class = ArgumentDefaultsHelpFormatter` line I added overrode some command's settings and messed up their formatting:
```
pwndbg> gdt -h
usage: gdt [-h] address [count]

Decode X86-64 GDT entries at address See also: * https://wiki.osdev.org/Global_Descriptor_Table *
https://wiki.osdev.org/GDT_Tutorial Note: In 64-bit mode, the Base and Limit values are ignored, each descriptor
covers the entire linear address space regardless of what they are set to.

positional arguments:
  address     x86-64 GDTR base address (e.g. read from sgdt instruction from [16:79] bits)
  count       Number of entries to dump (should be (GDTR.size+1)/8) (default: 8)

options:
```
but it should be good now (as in, it's the same as before this PR):
```
pwndbg> gdt -h
usage: gdt [-h] address [count]

Decode X86-64 GDT entries at address

See also:
* https://wiki.osdev.org/Global_Descriptor_Table
* https://wiki.osdev.org/GDT_Tutorial

Note:
In 64-bit mode, the Base and Limit values are ignored, each descriptor covers the entire linear address space regardless of what they are set to.

positional arguments:
  address     x86-64 GDTR base address (e.g. read from sgdt instruction from [16:79] bits)
  count       Number of entries to dump (should be (GDTR.size+1)/8) (default: 8)

options:
```

disconnect3d (approved):

PR #2922, opened 2025-04-25
Fix ARM IT blocks in disassembly

OBarronCS:
Arm Thumb2 has the concept of `IT Blocks`, which are a sequence of up to 4 instructions which each run conditionally based on a single initial condition defined in an IT instruction. This was a feature introduced to Thumb mode to allow for more efficient execution of conditional instructions, because Thumb mode instructions generally cannot be made conditional (while normal Arm instructions can).

There's a good overview of these instructions here - https://azeria-labs.com/arm-conditional-execution-and-branching-part-6/#:~:text=CONDITIONAL%20EXECUTION%20IN%20THUMB

Official docs: https://developer.arm.com/documentation/dui0473/m/arm-and-thumb-instructions/it

Once the Unicorn engine starts executing an IT block, it doesn't stop until all the instructions are executed, akin to it's behavior in MIPS branch delay slots. https://github.com/unicorn-engine/unicorn/issues/1630#:~:text=In%20short%2C%20it%20is%20similar%20to%20ARM%20that%20the%20condition%20might%20be%20lost%20if%20we%20stop%20inside%20a%20delay%20slot. The solution here is to disable the emulator in these blocks, since it skips over the instructions in the block that we still want to display.

### DIFF
--- a/pwndbg/aglib/disasm/arm.py
+++ b/pwndbg/aglib/disasm/arm.py
@@ -243,12 +243,11 @@ class ArmDisassemblyAssistant(pwndbg.aglib.disasm.arch.DisassemblyAssistant):
 
         # Disable Unicorn while in IT instruction blocks since Unicorn cannot be paused in it.
         flags_value = pwndbg.aglib.regs[self.flags_reg]
-        if flags_value is not None:
-            it_state = itstate_from_cpsr(flags_value)
+        it_state = itstate_from_cpsr(flags_value)
 
-            if instruction.id == ARM_INS_IT or it_state != 0:
-                if emu:
-                    emu.valid = False
+        if instruction.id == ARM_INS_IT or it_state != 0:
+            if emu:
+                emu.valid = False
 
     @override
     def _condition(self, instruction: PwndbgInstruction, emu: Emulator) -> InstructionCondition:
--- a/pwndbg/aglib/disasm/arm.py
+++ b/pwndbg/aglib/disasm/arm.py
@@ -246,9 +246,8 @@ class ArmDisassemblyAssistant(pwndbg.aglib.disasm.arch.DisassemblyAssistant):
         flags_value = pwndbg.aglib.regs[self.flags_reg]
         it_state = itstate_from_cpsr(flags_value)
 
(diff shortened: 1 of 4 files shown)


OBarronCS:
Before (it treated the IT instruction as a jump due the emulator not pausing where expected):
image


After:
image

pwndbg/aglib/disasm/arm.py
```diff
@@ -124,6 +124,29 @@
 }
 
 
+def itstate_from_cpsr(cpsr_value: int) -> int:
+    """
+    The ITSTATE bits are spread across 3 sections of Arm flags register to a total of 8 bits.
```
disconnect3d, line 130:
```suggestion
    ITSTATE == If-Then execution state bits for the Thumb IT instruction
    The ITSTATE bits are spread across 3 sections of Arm flags register to a total of 8 bits.
```

pwndbg/aglib/disasm/arm.py
```diff
@@ -218,6 +241,14 @@ def _prepare(
             # https://github.com/capstone-engine/capstone/issues/2630
             instruction.groups.remove(CS_GRP_CALL)
 
+        # Disable Unicorn while in IT instruction blocks since Unicorn cannot be paused in it.
+        flags_value = pwndbg.aglib.regs[self.flags_reg]
+        it_state = itstate_from_cpsr(flags_value)
+
+        if instruction.id == ARM_INS_IT or it_state != 0:
+            if emu:
```
disconnect3d: Maybe just:
```py
if (instruction.id == ARM_INS_IT or it_state != 0) and emu:
```
?

pwndbg/aglib/disasm/disassembly.py
```diff
@@ -94,7 +94,7 @@ def clear_on_reg_mem_change() -> None:
 
 # Maps an address to integer 0/1, indicating the Thumb mode bit for the given address.
 # Value is None if Thumb bit irrelevent or unknown.
```
disconnect3d:
```suggestion
# Value is None if Thumb bit is irrelevent or unknown.
```

disconnect3d (approved):

Issue #2903, opened 2025-04-18
invalid symbols not properly detected in integration providers


k4lizen:
When you pass an invalid symbol to ida:
```
pwndbg> p/x $ida("invalid_thingy")
$1 = 0x555555553fff
pwndbg> vmmap bash
►   0x555555554000     0x55555556f000 r--p    1b000      0 /usr/bin/bash
[...]
```
It returns `-1` which gets rebased onto the executable, and thus the user thinks the symbol exists when in actuallity it doesn't and the returned address is invalid.

When you pass an invalid symbol to binja:
```
pwndbg> p/x $bn_var("main")
TypeError: Could not convert Python object: None.
Error while executing Python code.
```
This error message is pretty bad and could be improved.

wtfashwin:
hey @disconnect3d I wanted to work on this issue, can you kindly assign it to me.

patryk4815:
@wtfashwin Hi, pwndbg is an open-source project - there's no need to assign issues. Just go ahead and submit your code or a create PR ;)

disconnect3d:
Fixed in https://github.com/pwndbg/pwndbg/pull/2926

k4lizen:
I think this should be reopened because the binary ninja error has not been touched/fixed and tbh the error message for ida could be improved.

k4lizen:
Fixed by https://github.com/pwndbg/pwndbg/pull/3401 since the symbols are now in the debugger rather than doing expression evaluation in the decompiler. And the debugger will print out a clear error message for this.

PR #2923, opened 2025-04-26
Fix: [Issue #2903] added safe_symbol function

wtfashwin:
Added a safe_symbol function


patryk4815:
@wtfashwin please send screen after this change, what does IDA returns when we use:
```
pwndbg> p/x $ida("invalid_thingy")
???
```

disconnect3d:
Hi, did you actually run Pwndbg and synchronized it with IDA Pro tool and you found that issue and fixed it that way?

As @patryk4815 mentioned, can you show us a screenshot of that?

PR #2927, opened 2025-04-27
Update register definitions + manually emulated register set

OBarronCS:
This PR updates the register set definitions to allow us to associate more information with registers, such as their size and their subregisters (`rax` contains `eax`, `ax`, `ah`, and `al`).

Using this new information, there is also a new class dubbed `PsuedoEmulatedRegisterFile`, which functions as a register set where we can manually track registers values by writing (`.write_register("eax", 0xFFFF_FFFF)`) and reading (`.read_register("al") == 0xFF`) to allow for some manual dynamic/static analysis without the need for a full emulator (or in cases where we cannot emulate). Hopefully the tests illustrate the functionality.

The class is not yet used for any new feature in this PR, but sets the stage for it's use in enhancing the disassembly view.

### DIFF
--- a/pwndbg/lib/regs.py
+++ b/pwndbg/lib/regs.py
@@ -80,7 +80,7 @@ class RegisterSet:
     #: Reg objects containing information on each register
     reg_definitions: Dict[str, Reg]
 
-    #: Map of register name to the full register it resides in. Example mapping: "eax" -> "rax"
+    #: Map of register name to the full register it resides in. Example mapping: "eax" -> Reg("rax")
     full_register_lookup: Dict[str, Reg]
 
     def __init__(
--- a/pwndbg/lib/regs.py
+++ b/pwndbg/lib/regs.py
@@ -5,6 +5,7 @@ standardized interface to registers like "sp" and "pc".
 
 from __future__ import annotations
 
+from collections import defaultdict
 from dataclasses import dataclass
 from typing import Dict
 from typing import Iterator
@@ -14,19 +15,32 @@ from typing import Set
 from typing import Tuple
 from typing import Union
 
+import pwndbg.lib.disasm.helpers as bit_math
 from pwndbg.lib.arch import PWNDBG_SUPPORTED_ARCHITECTURES_TYPE
 
 BitFlags = OrderedDict[str, Union[int, Tuple[int, int]]]
 
 
 @dataclass
-class EmulatedRegister:
+class UnicornRegisterWrites:
     """
     Represent a register to write to the Unicorn emulator.
     """
 
     name: str
     force_write: bool
+
+
+@dataclass
+class Reg:
+    name: str
+    size: int | None = None
+    """Register width in bytes. None if the register size is arch.ptrsize"""
+    offset: int = 0
+    """Relevant for subregisters - the offset of this register in the main register"""
+    zero_extend_writes: bool = False
+    """Upon writing a value to this subregister, are the higher bits of the full register zeroed out?"""
+    subregisters: tuple[Reg, ...] = ()
 
 
 class RegisterSet:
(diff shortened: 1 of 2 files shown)


pwndbg/lib/regs.py
```diff
@@ -63,33 +77,49 @@ class RegisterSet:
...
         self.args = args
         self.retval = retval
 
+        self.reg_definitions = {}
+        self.full_register_lookup = {}
+        for reg in gpr + (stack, frame, pc) + retaddr:
+            if reg:
+                self.reg_definitions[reg.name] = reg
+                self.full_register_lookup[reg.name] = reg
+                for subregister in reg.subregisters:
```
OBarronCS, line 119: Consideration: subregisters should probably be added to the `.all` set?

It would fix the issue described here https://github.com/pwndbg/pwndbg/pull/2836#discussion_r2032188909, as previously the "misc" category of registers for amd64 was missing ecx/edx, so attempting to write with `pwndbg.aglib.regs.ecx = 1` failed because it didn't exist in the `.all` set.

OBarronCS: Ok yeah this is necessary, some tests rely on these subregisters being writable

pwndbg/lib/regs.py
```diff
@@ -63,33 +77,52 @@ class RegisterSet:
...
+                    self.reg_definitions[subregister.name] = subregister
+                    self.full_register_lookup[subregister.name] = reg
+                    all_sub_registers.append(subregister.name)
+
         # In 'common', we don't want to lose the ordering of:
         self.common = []
-        for reg in gpr + (frame, stack, pc) + tuple(flags):
+        for reg in self.gpr + (self.frame, self.stack, self.pc) + tuple(self.flags):
```
disconnect3d: Why adding selfs here?

Btw above we do: `gpr + (stack, frame, pc) + retaddr`

I guess we could do:
`some_regs = gpr + (stack, frame, pc)`
and use it here and above :p

disconnect3d: Or actually, use `itertools.chain` or something? We don't need to add lists of registers if we don't care about the final product and we only want to iterate over them

pwndbg/lib/regs.py
```diff
@@ -98,14 +131,28 @@ def __init__(
...
+        self.emulated_regs_order: List[UnicornRegisterWrite] = []
 
-        for reg in [pc] + list(flags) + [stack, frame] + list(retaddr) + list(misc) + list(gpr):
+        for reg in (
+            [self.pc]
+            + list(self.flags)
+            + [self.stack, self.frame]
+            + list(self.retaddr)
+            + list(self.misc)
+            + list(self.gpr)
```
disconnect3d: Same issue as above, not great we create/add lists just to iterate over

pwndbg/lib/regs.py
OBarronCS: With the new `Reg` definition, the constructor parameters take many of the values as type `Reg` or `tuple[Reg]` when they used be `str`, so local variable `gpr: Tuple[Reg,...]`. 

But the attributes of the object with the same name expect to still be `str`, and are used as such in various places around the codebase, such as
https://github.com/pwndbg/pwndbg/blob/3a783d0637027829fb9c0e13b60c7ed55ab49539/pwndbg/aglib/regs.py#L154-L156

and

https://github.com/pwndbg/pwndbg/blob/3a783d0637027829fb9c0e13b60c7ed55ab49539/pwndbg/aglib/regs.py#L162-L164


Not ideal to have `typeof(gpr) != typeof(self.gpr)` in this `__init__` function, but I didn't want to change the name of the parameters while still keeping the attributes of the RegisterSet as strs.

pwndbg/lib/regs.py
OBarronCS: I made this change

pwndbg/lib/regs.py
OBarronCS: In other words - currently in this PR the use of the `Reg` type is completely internal and doesn't leave this lib file.

disconnect3d (approved):

PR #2930, opened 2025-04-27
Run all mips32 tests as little and big-endian

OBarronCS:
This PR makes all the MIPS tests compile and run both little and big-endian, where previously they were running as big-endian only. Adds gcc-mipsel cross-compiler to setup-dev.sh.


patryk4815:
https://github.com/rust-lang/rust/pull/34910#issuecomment-233694689

```
@alexcrichton I did a survey of the OpenWRT musl releases and 
all the mips (8 devices) and mipsel (12 devices) releases are using the soft float ABI.

So, yes. We should change the mipsel-musl target to soft float as well.
```

Just to context, how many devices use mipsel or mips

PR #2931, opened 2025-04-27
Cleanup sys.exit, os._exit, flush stdout, fix tests

patryk4815:
In python:
sys.exit - is doing flush stdout/stdout + raise SystemExit exception
os._exit - syscall exit, without flushing buffers

Because of bug in GDB, we are patching sys.exit with os._exit


gdbinit.py
```diff
@@ -180,7 +179,19 @@ def rewire_exit():
         # a segfault. See:
         # https://github.com/pwndbg/pwndbg/pull/2900#issuecomment-2825456636
         # https://sourceware.org/bugzilla/show_bug.cgi?id=31946
-        sys.exit = os._exit
+        def _patched_exit(exit_code):
+            # argparse requires a SystemExit exception, otherwise our CLI commands will exit incorrectly on invalid arguments
+            stack_list = traceback.extract_stack(limit=2)
+            if len(stack_list) == 2:
+                p = stack_list[0]
+                if p.filename.endswith("/argparse.py"):
```
disconnect3d, line 187: Uh, I bet this is bug prone. Shall we maybe inherit from `ArgumentParser` and override its `exit` method?

See https://peps.python.org/pep-0389/#discussion-sys-stderr-and-sys-exit

Or: https://github.com/python/cpython/blob/995b1a72f20e4b0bc44b6471d40c2c368d74efb1/Lib/argparse.py#L2656-L2662

disconnect3d (approved):
Lgtm, but see my comment about ArgumentParser, maybe we can do better for argparse stuff?

gdbinit.py
patryk4815: > ArgumentParser and override its exit method?

yes, I tried that, but it require replacing all ArgumentParser constructors, add `add_subparsers`

k4lizen: > yes, I tried that, but it require replacing all ArgumentParser constructors

i don't think it does, you don't have to subclass ArgumentParser you can just do `self.parser.exit = whatever`

patryk4815: easy way it just to create global variable eg:
```
try:
  pwndbg.started_parsing += 1
  args = parser.parse_args()
finally:
  pwndbg.started_parsing -= 1
```

and then
```
if pwndbg.started_parsing > 0:
   raise SystemExit()
```

disconnect3d: > it require replacing all ArgumentParser constructors

Yes, we would have to use our own parser class instead of ArgumentParser.

But... we could always do monkey patching like `ArgumentParser.exit = mocked_implementation` :P 

---

I don't quite get the `started_parsing > 0: raise SystemError()` approach. How does this fix the issue?

patryk4815: @k4lizen You need because we are using `add_subparsers` and it is not possible to override with `self.parser.exit`

disconnect3d: Anyway, lets merge this for now since this fixes an actual issue.

If u think its worth it, please add an issue to cleanup this part^

PR #2932, opened 2025-04-27
Adding an extra x64 register set for qemu-system kernel pwning

jxuanli:
## Added the following to x64 register set when displaying `context` during a `qemu-system-x86_64` session
- Control registers: CR0, CR3, CR4 
  - potentially interesting flag bits for CR0 and CR4
- MSRs: gsbase, fsbase, efer
  -  similar to how the `gsbase` and `fsbase` commands are implemented, the register names should be `gs_base` and `fs_base` instead of `gsbase` and `fsbase`)
  - `NXE` flag for `efer`
- Segment registers
  - compact display 
  - useful when returning from kernel land 
- `AC` flag for `eflags`
  - useful for temporarily disabling SMAP

## Internal changes
- Replaced the `BitFlags` alias with a backward compatible `class BitFlags`
- Extended `class BitFlags` and added `class SegmentRegisters` and `class AddressingRegister` for more flexible `context` outputs
  - Register values for those classes are assigned in `pwndbg/aglib/regs.py` 
- Added a sanity check to ensure the newly added features are present in releases

## demo
https://github.com/user-attachments/assets/11312b55-6b89-4191-a7a9-3831cb1fb1eb


pwndbg/lib/regs.py
```diff
@@ -339,8 +451,8 @@ def __iter__(self) -> Iterator[str]:
         "es",
         "fs",
         "gs",
-        "fsbase",
```
patryk4815, line 681: just for context, we have wrapper (ptrace) for `fsbase` and `gsbase` register in `aglib/regs.py`:
image

Im happy to remove that wrapper at all, but it will remove fsbase/gsbase for x86-32bit
@disconnect3d any idea?

jxuanli: The reason I changed them was:

Using the original naming

image

gives:

image

 so doesn't seem it is using the wrapper, also:
 
 
image

pwndbg/lib/regs.py
```diff
@@ -307,10 +384,45 @@ def __iter__(self) -> Iterator[str]:
...
+        "cr4": BitFlags(
+            [
+                ("UMIP", 11),
+                ("FSGSBASE", 16),
+                ("SMEP", 20),
+                ("SMAP", 21),
+                ("PKE", 22),
+                ("CET", 23),
+                ("PKS", 24),
+            ]
```
disconnect3d, line 670: All lists here (and above and below) should be tuples since we won't modify the flags length at runtime etc.

pwndbg/lib/regs.py
disconnect3d: I never seen this error with unknown register, when does this occur? Is it that we never displayed those regs or what?

disconnect3d (commented):
Hi,

Thanks for sending this PR. I like this in general. If I understand correctly, we are mostly moving the logic of rendering context registers from `get_regs` inline code to context register classes?

I have got some comments and questions:

1. This may collide with @OBarronCS https://github.com/pwndbg/pwndbg/pull/2927 work, so I'd like @OBarronCS to review this too.
2. I don't like the fact that we are creating a bunch of new classes, lots of which have just a single field and method. (Its usually expensive in Python, and I'd prefer context display to be faster than slower if possible, heh). I am not saying no, but lets discuss it (?).
3. Does this work when one passes registers to `get_regs`? I think this happens when we do e.g. `regs rax rbx` command.
4. How does the context display look like when one does `set show-compact-regs on`? (eh, we should rename this to `context-compact-regs` probably...)
5. May be worth to add some comments/docstrings here and there to inform on the purpose of certain classes etc.

jxuanli:
@disconnect3d 

> we are mostly moving the logic of rendering context registers from get_regs inline code to context register classes?

Yes, my reason is to have a more flexible way of rendering registers. At the moment, we only have two different ways of rendering register context -- we are only rendering `BitFlags` differently from the rest of simple registers (denoted by just a string). However, if we want render registers in other ways, more code needs to be inlined. As shown in the demo I attached in the PR description, it may make sense to render the segment registers (CC, CS, etc) differently than GPRs (RAX, RBX, etc.). If more flag registers are added, more checks need to performed as well to check if the register should be rendered as a flag or not.

I guess ideally this should be handled in a more "non-switch-on-type" manner. One idea is to have some register classes to have a `context` function. I believe @OBarronCS 's PR can help with this effort but we might want to discuss it in more details as how we want to do it as getting register values cannot be done within `pwndbg/lib/regs.py` because this would result in cyclic imports.

> classes are usually expensive in Python

Thanks for pointing that out! Though I would say that given the maximum number of registers across different archs, this overhead may not impact the performance too much.

> Does this work when one passes registers to `get_regs`? 

I believe so. As shown in the attached demo, it runs `regs rax rbx` successfully.

> How does the context display look like when one does `set show-compact-regs on`?

Looks fine in my demo.

I'll attach it here as well:
.

https://github.com/user-attachments/assets/95c77a3d-782e-40a4-a342-3de3245ef516

pwndbg/commands/context.py
```diff
@@ -281,6 +284,104 @@ def isatty(self):
             return False
 
 
+class RegisterContext:
+    reg: str
```
OBarronCS: Possible comment:
```
An instance of this class contains the functionality to pretty-print a single register value.

For example, ....
```

pwndbg/commands/context.py
```diff
@@ -281,6 +284,104 @@ def isatty(self):
...
+        self.flags = flags
+
+    def desc(self):
+        val = self.value()
+        if val is None:
+            return None
+        return C.format_flags(val, self.flags, pwndbg.aglib.regs.last.get(self.reg, 0))
+
+
+class AddressingRegisterContext(RegisterContext):
```
OBarronCS: Looks like the .is_virtual is not checked anywhere - what would the the .is_virtual boolean here provide?

pwndbg/commands/context.py
```diff
@@ -281,6 +284,104 @@ def isatty(self):
...
+        change_marker = f"{C.config_register_changed_marker}"
+        m = (
+            " " * len(change_marker)
+            if self.reg not in changed
+            else C.register_changed(change_marker)
+        )
+        return f"{m}{regname} {hex(self.value())}   "
+
+
+class RegistersContext:
```
OBarronCS: This and `RegisterContext` have a very similar name - could they be differentiated/commented so someone new to these classes could immediately understand the use of the class. Looking through the code here, it's not clear to me where this is used.

pwndbg/commands/context.py
```diff
@@ -899,6 +1000,23 @@ def regs(regs=[]) -> None:
 
 
 def get_regs(regs: List[str] = None):
+    regs: List[str | Tuple[str | None, BitFlags | AddressingRegister | Tuple[str, ...]]] = regs
+
+    def select_context(reg) -> RegisterContext | RegistersContext | None:
+        if reg is None:
+            return None
+        elif isinstance(reg, str):
+            return RegisterContext(reg)
```
OBarronCS: I understand the utility of this function - however it might be difficult to discover/maintain. A more direct mapping, where we can immediately jump from "reg" -> "how to pretty print" might be easier to maintain, but would require some crafty planning, as you said, to avoid circular imports.

pwndbg/lib/regs.py
```diff
@@ -16,7 +16,69 @@
 
 from pwndbg.lib.arch import PWNDBG_SUPPORTED_ARCHITECTURES_TYPE
 
-BitFlags = OrderedDict[str, Union[int, Tuple[int, int]]]
+
+class BitFlags:
```
OBarronCS, line 23: Do we use this BitFlags class differently from OrderedDict - is this aimed for future changes?

pwndbg/lib/regs.py
OBarronCS: GDB internally calls these variables `fs_base`/`gs_base`, which means a direct debugger register lookup now works.

image

This means the lookup is going straight through to the process, and looks like it is not using the `helper` functions noted above. Looks like the helper function does extra logic to determine the effective value of fs/gs on 32-bit, where more complex logic is needed.

In this particular case, because it's x64, it works - but if we aim to have this logic work in the general case (so the code could extend to work 32-bit contexts), perhaps we could try to call these wrapper functions when fetching the value, so this would work in 32-bit? This would mean instead of `.aglib.regs["fsbase"]`, to do `.aglib.regs.fsbase`

OBarronCS:
Awesome stuff! I made some comments - mostly questions I had to understand the code better.

Noticed one side effect in the display - registers that do not exist will print "None":
image

Previously, this didn't display "None".
image


> I guess ideally this should be handled in a more "non-switch-on-type" manner. One idea is to have some register classes to have a context function. I believe @OBarronCS 's PR can help with this effort but we might want to discuss it in more details as how we want to do it as getting register values cannot be done within pwndbg/lib/regs.py because this would result in cyclic imports.

Completely agree with this, and I'll give some thought as to a clean way to do this. If I'm thinking about it correctly, this boils down to a mapping of "register name" to "how we want to display it in the context".

Perhaps, in line with changes to register definitions in #2927, we could apply some sort of tag to the registers, or create callbacks functions in that file which take in the needed context (I think just the previous and current register value is needed?) and returns the corresponding strings to represent the register? I'll give it some more thought.

jxuanli:
Another idea for handling the register context display is to have the functions in `pwndbg/aglib/regs.py` assgin the values because we have the necessary register classes now with this PR and #2927. Though if we go with that, the return types of the properties in `pwndbg/aglib/regs.py` might need to be changed. (?) We could potentially handle `Reg` the same way I handled `BitFlags`, i.e., change it such that every string operation on `Reg` is still valid and equivalent to performing the string op on the register name, etc.

jxuanli:
fwiw, my previous comment works and the changes should be much more cleaner now. Now it's just waiting for #2927 , and I believe I can adapt it once it's merged. It should be similar to how `AddressingRegister` is implemented now

disconnect3d:
@jxuanli We merged #2927 so let us know when we can revisit this PR :)

jxuanli:
After giving it some thoughts, I decided to follow #2927 and keep `Reg` internal to `pwndbg/lib/regs.py` for now. Although some refactoring will make the code for rendering context slightly cleaner, too many functions expect the representations of **simple** registers are strings in `class RegisterSet`. Therefore, the cost of refactoring may outweigh its return at the moment. Also, there are a few other features I have in mind that I wish to add first. However, if we think that it would be worth it, I can later open a PR that is dedicated to refactoring the relevant code, perhaps using the `reg_definitions` in `class RegisterSet`.

OBarronCS:
Great! I'll have time tomorrow to take a deeper look.

disconnect3d:
@OBarronCS did u have chance to look at it? :)

pwndbg/aglib/regs.py
```diff
@@ -167,9 +169,43 @@ def frame(self) -> str | None:
...
+                    control.update(reg, self[reg], self.last.get(reg, 0))
+                else:
+                    control.update(self[reg])
+        msrs = kernel_regs.msrs
+        if msrs is not None:
+            for reg, msr in msrs.items():
+                if isinstance(msr, BitFlags):
+                    msr.update(reg, self[reg], self.last.get(reg, 0))
+                else:
+                    msr.update(self[reg])
```
OBarronCS: I feel like if we are going with an object oriented approach, we should be able to benefit from polymorphism/a uniform API, so that all the .update functions have the same parameters so we don't need to do any runtime `isinstance` checks.

pwndbg/commands/context.py
```diff
@@ -896,10 +897,31 @@ def regs(regs=[]) -> None:
 
 
 def get_regs(regs: List[str] = None):
+    regs: List[Any] = regs  # ei, not dealing with types here is probably fine
+
+    def context(reg):
```
OBarronCS: Maybe we should define this function at the top-level of this file so it doesn't get redefined on each run?

pwndbg/commands/context.py
```diff
@@ -912,42 +934,33 @@ def get_regs(regs: List[str] = None):
...
+            if flags is not None:
+                regs += flags.values()
+        if pwndbg.aglib.qemu.is_qemu_kernel() and pwndbg.aglib.regs.kernel is not None:
+            if pwndbg.aglib.regs.kernel.segments is not None:
+                regs.append(pwndbg.aglib.regs.kernel.segments)
 
     for reg in regs:
         if reg is None:
             continue
+        if not isinstance(reg, str):
```
OBarronCS, line 1018: In what cases is a reg not a string? When it is one of the new classes defined in this PR?

pwndbg/lib/regs.py
```diff
@@ -16,10 +16,170 @@
...
+        if self.reg in changed:
+            regname = C.register_changed(regname)
+        change_marker = f"{C.config_register_changed_marker}"
+        m = (
+            " " * len(change_marker)
+            if self.reg not in changed
+            else C.register_changed(change_marker)
+        )
+        import pwndbg.chain
+
```
OBarronCS: I see the need for this import here (would otherwise get circular imports). Could we possibly explore other patterns of matching functionality to these register definitions, things like "Visitor Pattern" or dependency injection

OBarronCS: ^^ which would allow us to define the printing logic outside of this file while still being able to invoke it from within this file. This is just my own tendencies in coding, though.

pwndbg/lib/regs.py
```diff
@@ -16,10 +16,170 @@
...
+
+class SegmentRegisters:
+    """
+    Represents the x86 segment register set
+    """
+
+    regs: List[str]
+    idx_map: Dict[str, int]
+    values: List[int]
```
OBarronCS: Are these keeping copies of the global register set?

pwndbg/lib/regs.py
```diff
@@ -129,6 +294,16 @@ def __init__(
             if regname and regname not in self.common:
                 self.common.append(regname)
 
+        # pwndbg.aglib.qemu.is_qemu_kernel() results in error here
+        # because pwndbg is only partially initialized at this point
+        if self.kernel is not None:
+            controls = self.kernel.controls
+            segments = self.kernel.segments
+            msrs = self.kernel.msrs
+            for regname in tuple(controls) + tuple(segments.regs) + tuple(msrs):
```
OBarronCS: In this case, we could use `itertools.chain(...)` to iterate a series of iterables.

pwndbg/aglib/regs.py
```diff
@@ -167,9 +169,43 @@ def frame(self) -> str | None:
...
+
+        segments = kernel_regs.segments
+        if isinstance(segments, SegmentRegisters):
```
OBarronCS: Ideally, we wouldn't need any runtime checks of types.

OBarronCS:
Added some comments, mostly regarding the use of the new classes. Just wondering, for future maintainability, if it would be possible to avoid runtime `isinstance` checks, and have the classes have uniform method APIs (perhaps enforced using Protocol's (https://typing.python.org/en/latest/spec/protocol.html) or subclassing?)

pwndbg/lib/regs.py
disconnect3d: How would this look like with such a pattern?

jxuanli:
Thanks for the suggestions! The visitor pattern worked particularly well and i refactored everything into `class RegisterContext`. Many of the comments are addressed by this. In particular, it resolved the issues related to the use of `isinstance` and `import pwndbg.chain`. I also responded to the comments that are still relevant.

pwndbg/lib/regs.py
OBarronCS: Looks like this: https://github.com/pwndbg/pwndbg/blob/fda88c6da643f2faa73059d4fd77693ebd434939/pwndbg/lib/regs.py#L60-L61 (thanks @jxuanli!)

OBarronCS:
Thanks! I think this is in great shape!

pwndbg/lib/regs.py
```diff
@@ -19,7 +19,107 @@
...
+        return iter(self.flags)
+
+    def __len__(self):
+        return len(self.flags)
+
+    def __repr__(self):
+        return f"BitFlags({self.flags})"
+
+    def update(self, regname: str):
+        self.reg = regname
```
OBarronCS: Quick question: this is called in `get_regs` for the BitFlags and the AddressingRegister type. Just wondering if this is necessary - is the reg name not known at initialization time?

jxuanli:
should we merge this PR? or are there more things need to be done

disconnect3d:
@OBarronCS wdyt? can we merge? :)

pwndbg/lib/regs.py
```diff
@@ -19,7 +19,107 @@
 import pwndbg.lib.disasm.helpers as bit_math
 from pwndbg.lib.arch import PWNDBG_SUPPORTED_ARCHITECTURES_TYPE
 
-BitFlags = OrderedDict[str, Union[int, Tuple[int, int]]]
+
+class BitFlags:
+    # this should be backwards compatibility
+    reg: str  # this is intentionally uninitialized -- arm uses the same self.flags structuture for different registers
```
OBarronCS: Can there be an example in the comment of what this means in arm?

pwndbg/lib/regs.py
```diff
@@ -19,7 +19,107 @@
 import pwndbg.lib.disasm.helpers as bit_math
 from pwndbg.lib.arch import PWNDBG_SUPPORTED_ARCHITECTURES_TYPE
 
-BitFlags = OrderedDict[str, Union[int, Tuple[int, int]]]
+
+class BitFlags:
+    # this should be backwards compatibility
```
OBarronCS:
```suggestion
```

pwndbg/commands/context.py
```diff
@@ -895,8 +896,72 @@ def regs(regs=[]) -> None:
...
+        val = self.get_register_value(reg)
+        if val is None:
+            return None
+        desc = pwndbg.chain.format(val)
+        prefix = self.get_prefix(reg)
+        return f"{prefix} {desc}"
+
+
 def get_regs(regs: List[str] = None):
+    regs: List[Any] = regs  # ei, not dealing with types here is probably fine
```
OBarronCS:
```suggestion
    regs: List[Any] = regs
```

pwndbg/lib/regs.py
```diff
@@ -19,7 +19,107 @@
...
+
+
+class AddressingRegister:
+    """
+    Represents a register that is used to store an address, e.g. cr3, gsbase, fsbase
+    """
+
+    reg: str
+    value: int
+    is_virtual: bool  # inicating if the address is a virtual address
```
OBarronCS:
```suggestion
    is_virtual: bool
```
What does virtual mean here?

pwndbg/lib/regs.py
```diff
@@ -129,6 +234,16 @@ def __init__(
             if regname and regname not in self.common:
                 self.common.append(regname)
 
+        # pwndbg.aglib.qemu.is_qemu_kernel() results in error here
+        # because pwndbg is only partially initialized at this point
```
OBarronCS:
```suggestion
```

OBarronCS:
I did notice one new performance penalty that should be fixed:

This `.changed` access calls a "getter" function 

https://github.com/pwndbg/pwndbg/blob/d8e09c71a26c1b54386e8d0a0c8e78cfdcec01f7/pwndbg/commands/context.py#L901

Which is defined here:
https://github.com/pwndbg/pwndbg/blob/d8e09c71a26c1b54386e8d0a0c8e78cfdcec01f7/pwndbg/aglib/regs.py#L207-L212

We now run this fetch for every single register, instead of once, which in my local tests add non-insignificant overhead (the function call now accounts for 15% of the time spent running doing the entire register display logic)

jxuanli:
@OBarronCS Thanks for pointing is out. I think the performance issue should have been fixed, could you check it again and let me know if there are other issues?

OBarronCS:
Tested it - performance issue is fixed. Thanks for the work on this - I think this is ready to go.

disconnect3d:
Thanks!

PR #2935, opened 2025-04-29
Switch to Zig for cross-architecture compiling

OBarronCS:
This PR moves to using `zig cc` for cross-compiling for the cross-arch tests.

Zig has some great benefits, a main one being that it is self-contained and internally includes libc source code that it can statically/dynamically link while compiling, making it very simple to cross compile programs. With Zig, we can also specify a specific version of glibc to link to.

Removing the dependency on gcc cross compilers also makes it easier for developers to run the tests locally on any Linux distribution, where previously it relied on gcc+libc package versions on Ubuntu 24.04. Likewise, it will be easier and more lightweight to add support for tests for new architectures.

LLVM (which Zig uses under-the-hood) loads the programs into memory into different places compared to gcc, so the tests had to be refreshed to reflect the new addresses. I manually went through each test to ensure everything was in order.


tests/qemu-tests/tests/user/test_aarch64.py
```diff
@@ -15,6 +15,15 @@
 mov x0, 0
 mov x8, 93
 svc 0
+nop
+nop
+nop
+nop
+nop
+nop
+nop
```
OBarronCS, line 36: Interesting thing I encountered - the `svc 0` above is often the last instruction to be executed in these AArch64 tests and are placed in memory after all the other instructions. The bytes after it in memory (to fill the rest of the page) are typically filled with 0's. I noticed that with Zig, which should be entirely self-contained, compiling the program on different distros (ubuntu 24.04, arch) on the same version (zig 0.13.0) would result in a couple of the bytes after svc being slightly different, resulting in the disassembly outputting slightly different instructions, like `udf #0` or `udf #23` depending on the source distro.

To make this problem go away, I added the nops so that the disassembled instructions are consistent.

tests/qemu-tests/tests/user/test_aarch64.py
```diff
@@ -558,7 +567,7 @@ def test_aarch64_memory_operands(qemu_assembly_run):
 ADD X0, X1, W2, UXTB
 
 ADD X0, X1, X2, ASR 2
-MOV X0, X1, ROR 2
```
OBarronCS, line 561: Note: GCC supports this as an alias to orr, but it's not supported in clang.

tests/tests.py
```diff
@@ -299,20 +299,22 @@ def main():
...
-        if args.type == "cross-arch":
-            gdb_binary = "gdb-multiarch"
-        gdb_path = shutil.which(gdb_binary)
+        if (
+            args.type == "cross-arch"
+            and (gdb_multiarch := shutil.which("gdb-multiarch")) is not None
+        ):
+            gdb_path = gdb_multiarch
+        else:
+            gdb_path = shutil.which("gdb")
```
OBarronCS: This could use some work/suggestions. Ubuntu requires gdb-multiarch, while in other distros the `gdb` package is compiled with multiarch support out-of-the-box. This falls back to gdb in case gdb-multiarch isn't found. assuming it includes multiarch debugging. Perhaps some sort of detection for "this gdb binary supports debugging these arches" could come in handy here.

tests/qemu-tests/tests/user/binaries/Makefile
```diff
@@ -1,27 +1,36 @@
...
-# CFLAGS.loongarch64 = $(ALL_FLAGS)
+CFLAGS.aarch64 = $(ALL_FLAGS) --target=aarch64-linux-musl
+CFLAGS.arm = $(ALL_FLAGS) --target=arm-linux-musleabihf
+# CFLAGS.riscv32 = $(ALL_FLAGS) --target=riscv32-linux-musl
+CFLAGS.riscv64 = $(ALL_FLAGS) --target=riscv64-linux-musl
+CFLAGS.mips32 = $(ALL_FLAGS) --target=mips-linux-musl # Big-endian MIPS
+CFLAGS.mipsel32 = $(ALL_FLAGS) --target=mipsel-linux-musl # Little-endian MIPS
+CFLAGS.mips64 = $(ALL_FLAGS) --target=mips64-linux-musl
+# CFLAGS.loongarch64 = $(ALL_FLAGS) --target=loongarch64-linux-musl
+# CFLAGS.s390x = $(ALL_FLAGS) --target=s390x-linux-musl -mcpu=z13
```
OBarronCS, line 31: The commented arches (1) we currently don't have tests for, (2) need some more figuring out if musl/any libc is supported for them yet (at least loongarch64 doesn't have musl support at least in 0.13.0, some comments here https://redirect.github.com/ziglang/zig-bootstrap). In any case, future work can handle implementing tests for these.

tests/tests.py
patryk4815:
```
gdb -q -ex "py import os; archs = ['i386', 'aarch64', 'arm', 'mips', 'riscv', 'sparc']; len([arch for arch in archs if arch in gdb.architecture_names()]) == len(archs) if os._exit(3) else os._exit(2)" -ex 'quit'
```
+ subprocess 
+ supported_multiarch = exitcode == 3

tests/qemu-tests/tests/user/test_mips.py
```diff
@@ -5,6 +5,13 @@
 
 import pwndbg.color
 
+# Zig requires the start symbol to be __start instead of _start for MIPS
+MIPS_PREAMBLE = """
```
patryk4815, line 9: maybe this should be move into `qemu_assembly_run`?

setup-dev.sh
```diff
@@ -140,16 +140,9 @@ install_apt() {
         gdb \
         gdb-multiarch \
         parallel \
-        iproute2 \
```
disconnect3d, line 143: Why this?

tests/qemu-tests/conftest.py
```diff
@@ -28,33 +67,67 @@ def qemu_assembly_run():
...
-        context.clear()
-        context.arch = arch
+        if arch not in COMPILATION_TARGETS or arch not in COMPILE_AND_RUN_INFO:
+            raise Exception(f"Unknown compilation target: {arch}")
+
+        zig_target, extra_cli_args, qemu_suffix = COMPILE_AND_RUN_INFO[arch]
+
+        # Declare _start symbol at start of the assembly if it's not already declared
+        if "_start:" not in asm:
+            asm = ".text\n.globl _start\n_start:\n" + asm
```
disconnect3d: Can't we just require this in asm code instead of putting it here?

tests/qemu-tests/tests/user/test_aarch64.py
disconnect3d: May be worth adding a comment above mentioning this ;/

tests/qemu-tests/tests/user/test_aarch64.py
```diff
@@ -54,23 +63,23 @@ def test_aarch64_branch_enhancement(qemu_assembly_run):
...
-        "   0x1000002c               udf    #0\n"
+        "   0x101012c <end>         mov    x0, #0        X0 => 0\n"
+        "   0x1010130 <end+4>       mov    x8, #0x5d     X8 => 0x5d\n"
+        "   0x1010134 <end+8>       svc    #0\n"
+        "   0x1010138 <end+12>      nop    \n"
+        "   0x101013c <end+16>      nop    \n"
+        "   0x1010140 <end+20>      nop    \n"
+        "   0x1010144 <end+24>      nop    \n"
+        "   0x1010148 <end+28>      nop    \n"
+        "   0x101014c <end+32>      nop    \n"
```
disconnect3d, line 95: Ideally we could do some regex matching here and e.g. ensure that the offsets between instructions are correct, or, match only the base address, heh

patryk4815:
```
pytest.context_equal(
  expected_base=0x10011b8, 
  expected_str, 
  tested_base=pwndbg.aglib.regs.pc, 
  tested_str,
)
```
maybe something like this we should create, just for context
but in next PR / cleanup PR

tests/tests.py
OBarronCS: Awesome! Thanks - I added this.

setup-dev.sh
OBarronCS: At some point, the `tests.py` used the `ss` tool provided by `iproute2` to get a list of open ports for qemu to use, but that was changed a while ago to use a native python method. Removed it will getting rid of the other dependencies - I should have noted that.

tests/qemu-tests/conftest.py
OBarronCS: Yes - I'll inline it.

tests/qemu-tests/tests/user/test_aarch64.py
OBarronCS: Definitely should simplify these - although this was a good chance to manually double check all the tests again (and make sure the move from gcc -> zig didn't radically change the compiled binaries).

One considerations for this context equality check is that many of the assembly snippets are checking that branch targets/memory access addresses are correct. These are displayed in the annotations of instructions and not just as instruction addresses, which needs to be accounted for.

tests/qemu-tests/tests/user/test_mips.py
OBarronCS: Ended up in-lining the preambles for all the other architectures like we do in MIPS, instead of adding it adhoc

patryk4815 (approved):

disconnect3d (approved):

PR #2937, opened 2025-04-29
Fix `procinfo` for cross-namespace (root → netns) process attach

patryk4815:
This change adds support for displaying the correct `procinfo` when attaching from the root namespace to a process in a network namespace.

Before: 
image

After:
image


patryk4815:
Ready for CR

PR #2938, opened 2025-04-29
Fix default parameter value handling

k4lizen:
Unfortunately `argparse.ArgumentDefaultsHelpFormatter` kind of sucks since it shows `None` for toggle-able options and such:
image

Though the solution we had before https://github.com/pwndbg/pwndbg/pull/2919 didn't show `None`:
```
            if isinstance(action, argparse._SubParsersAction):
                action.type = str
            if action.dest == "help":
                continue
            if action.type is int:
                action.type = fix_int_reraise_arg
            if action.type is None:
                action.type = fix_reraise_arg
            if action.default is not None:                                # these two 
                action.help += " (default: %(default)s)"          # lines
```
I'm not sure if there's a benefit to that approach rather than subclassing a formatter_class (which is what I did in that and this PR), maybe it is more resistant to python internal API breakage?

In any case I added some more checks (leveraging the `argparse.ArgumentDefaultsHelpFormatter` code) so it is a bit more robust.
Previously:
```
module      Module to choose as base. Defaults to the target executable. (default: )
```
```
options:
  -h, --help         show this help message and exit
  --beyond_top, -b   Attempt to keep printing beyond the top chunk. (default: False)
  --no_truncate, -n  Display all the chunk contents (Ignore the `max-visualize-chunk-size` configuration). (default: False)
  --all_chunks, -a    Display all chunks (Ignore the default-visualize-chunk-number configuration). (default: False)
```
Now:
```
module      Module to choose as base. Defaults to the target executable. (default: '')
```
```
options:
  -h, --help         show this help message and exit
  --beyond_top, -b   Attempt to keep printing beyond the top chunk.
  --no_truncate, -n  Display all the chunk contents (Ignore the `max-visualize-chunk-size` configuration).
  --all_chunks, -a   Display all chunks (Ignore the default-visualize-chunk-number configuration).
```
Strings are shown in quotes. False booleans are usually trigger parameters so they aren't shown. 
## Docs generation
The docs generation in regards to default values was borked at times (like `--hex` from below is missing the default, which isn't intended from looking at the code, and before the afformentioned PR the help also showed: `Target is a hex-encoded (for bytes/strings) (default: %(default)s)` because the default val substitution didn't happen):
image

image

I removed the Defaults column since it is often mostly filled with None/False (the defaults are now in the param help, just like in the CLI). Defaults now show in positional params also. Removed the code highlight on parameter names.
image
image


scripts/_gen_command_docs.py
```diff
@@ -48,18 +44,21 @@ def category_to_folder_name(category) -> str:
...
     Returns:
         (A dictionary that maps the filenames of .md files to the corresponding
-        command's ArgumentParser objects, A dictionary that maps a category name
+        CommandObj objects, A dictionary that maps a category name
         to a list of filenames for commands that belong to the category).
     """
     filename_to_source = {}
     category_to_filename = {}
 
+    # FIXME: If an lldb-only command is added this (or pwndbg.commands) will need to be
```
disconnect3d, line 59: Can we somehow ensure we dont forget about this?

disconnect3d (approved):

docs/commands/context/contextoutput.md
```diff
@@ -20,17 +20,17 @@ usage: contextoutput [-h] section path clearing [banner] [width]
...
 |Positional Argument|Help|
 | :--- | :--- |
-|`section`|The section which is to be configured. ('regs', 'disasm', 'code', 'stack', 'backtrace', 'ghidra', 'args', 'threads', 'heap_tracker', 'expressions', and/or 'last_signal')|
-|`path`|The path to which the output is written|
-|`clearing`|Indicates weather to clear the output|
-|`banner`|Where a banner should be placed: both, top , bottom, none|
-|`width`|Sets a fixed width (used for banner). Set to None for auto|
+|section|The section which is to be configured. ('regs', 'disasm', 'code', 'stack', 'backtrace', 'ghidra', 'args', 'threads', 'heap_tracker', 'expressions', and/or 'last_signal')|
+|path|The path to which the output is written|
+|clearing|Indicates weather to clear the output|
```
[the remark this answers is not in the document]
disconnect3d: @k4lizen ugh i think this is wrong place to make this change?

scripts/_gen_command_docs.py
k4lizen: Maybe it's best I just look to solve it now.

k4lizen: hmm but even if I refactored, conceptually, if the doc gen script is run inside a debugger, that debugger won't have access to the other debugger's API, and importing a file that defines a command for only one of them (like how `pwndbg.command.ai` does `import gdb`, so even loading the command inside lldb is impossible) doesn't work.

k4lizen: ooo maybe I could replace the
```python
import gdb
```
with
```python
    import gdb
```
in the command files.

k4lizen: ~~hmm doesn't quite work~~
```python
    import gdb
    from pwndbg.gdblib.functions import GdbFunction
else:
    class GdbFunction:
        def __init__(self):
            pass
        def __call__(self, fn):
            return fn
```
This does work, but it's poor developer experience tbh. Maybe the idea should be to run the doc generator from all supported debuggers. I'll open an issue about it so we don't forget.

PR #2940, opened 2025-04-30
CI: add uploading `nix develop` to cachix

patryk4815:


nix/devshell.nix
```diff
@@ -76,19 +76,6 @@ in
       }
       ++ [
         jemalloc-static
-        # from qemu-tests.sh
```
patryk4815, line 79: not needed since PR: https://github.com/pwndbg/pwndbg/pull/2935

PR #2943, opened 2025-05-01
empty context-sections

marywinter3:
When context sections are set empty, ctx command gives a warning.

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -702,14 +702,13 @@ def context(subcontext=None, enabled=None) -> None:
         subcontext = []
     args = subcontext
 
-    if len(args) == 0:
-        args = config_context_sections.split()
-
     # Inform when sections set to be empty
     sections = pwndbg.config.context_sections.split()
     if not sections:
         print(message.warn("Sections set to be empty."))
-        return
+    
+    if len(args) == 0:
+        args = config_context_sections.split()
 
     sections = []
     if args:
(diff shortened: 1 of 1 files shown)


pwndbg/commands/context.py
```diff
@@ -702,6 +702,11 @@ def context(subcontext=None, enabled=None) -> None:
         subcontext = []
     args = subcontext
 
+    # Inform when sections set to be empty
+    sections = pwndbg.config.context_sections.split()
```
disconnect3d, line 706: What's the purpose of split here if we are looking if its empty anyway?

peace-maker:
The warning should only be printed when the command is executed manually. I think this would print it too when context is printed automatically on stop?

k4lizen:
Doesn't this kind of go against the logic here?
https://github.com/pwndbg/pwndbg/blob/001926ebc0fc6b9c6a87833ec45346dd496d91f3/pwndbg/commands/context.py#L177-L207

peace-maker:
We could deprecate that logic of allowing empty context-sections now that there is `ctx --off` and `ctx --on`. It might be useful to print the warning when the user explicitly calls the context command while it's disabled to remind them of the way to enable it again. It should somehow only be for interactive commands the user triggered manually, not from scripts though.

disconnect3d:
Random comment: `ctx --on|--off` could really be `ctx on|off`, heh. I agree with the warning as @peace-maker says, since we want to not break things too much or too fast :(.

disconnect3d:
Closing this for now; pls send a new PR when its ready/tested etc

PR #2751, opened 2025-02-18
feat(commands/canary): add TLS canary address display and improve architecture support

kgand:
Display the actual Thread Local Storage (TLS) address where the stack canary is stored,
eliminating the need for manual searching. This addresses and fixes issue https://github.com/pwndbg/pwndbg/issues/2511.

Key changes:
- Add architecture-specific TLS canary offsets with source references
- Implement robust TLS base address detection with fallback mechanisms
- Add support for x86_64 (fs:0x28), i386 (gs:0x14), and aarch64 (tpidr_el0+0x28)
- Add value verification to detect potential TLS address calculation issues
- Improve error handling and user feedback for unsupported architectures

The command now shows:
- AT_RANDOM location (seed value source)
- TLS address where canary is stored
- Actual canary value
- Stack locations where canary is copied

Example output:
pwndbg> canary
AT_RANDOM = 0x7fff05b46609 # points to global canary seed value
TLS Canary = 0x7fff05b46628 # address where canary is stored
Canary    = 0x989538cda22ba700 (may be incorrect on != glibc)

Closes https://github.com/pwndbg/pwndbg/issues/2511

PR #2944, opened 2025-05-03
add TLS canary address display

### LINKED ISSUE #2511: canary command should show the TLS canary address
```
pwndbg> canary
AT_RANDOM = 0x7fff05b46609 # points to (not masked) global canary value
Canary    = 0x989538cda22ba700 (may be incorrect on != glibc)
Thread 1: Found valid canaries.
```

The "points to (not masked) global canary value" is somewhat a lie, because this is pointer to `AT_RANDOM` in AUXV from where the global canary value is derived and stored in TLS (its first byte is zeroed). So yeah, we should clarify this and try to display the TLS address too.

Without this ppl have to do `search -8 0x989538cda22ba700` to find global canary address in TLS...

Er3X:
Reopen #2751 with fixes, as original author (@kgand) appears inactive

Fix:
- Corrected usage of `@pwndbg.commands.Command(...)` instead of `@pwndbg.commands.ArgparsedCommand(...)`
- Linting

Closes #2511


pwndbg/commands/canary.py
```diff
@@ -14,8 +15,25 @@
 
 DEFAULT_NUM_CANARIES_TO_DISPLAY = 1
 
+# Architecture-specific TLS canary offsets
+# These offsets are from the TLS base to the canary
+# References:
+# - x86_64: fs:0x28 (https://elixir.bootlin.com/glibc/latest/source/sysdeps/x86_64/nptl/tls.h)
```
disconnect3d: Can we change the links to not be 'latest' but to latest current version? This way we know when it worked last time if it ever changes

pwndbg/commands/canary.py
```diff
@@ -14,8 +15,25 @@
...
+# - aarch64: tpidr_el0 + 0x28 (https://elixir.bootlin.com/glibc/latest/source/sysdeps/aarch64/nptl/tls.h)
+TLS_CANARY_OFFSETS = {
+    "x86-64": 0x28,
+    "i386": 0x14,
+    "aarch64": 0x28,
+}
+
 
 def canary_value():
+    """Get the global canary value from AT_RANDOM.
```
disconnect3d:
```suggestion
    """Get the global canary value from AT_RANDOM with its last byte masked (as glibc does)
```

pwndbg/commands/canary.py
```diff
@@ -40,15 +88,30 @@ def canary_value():
...
     global_canary, at_random = canary_value()
 
     if global_canary is None or at_random is None:
         print(message.error("Couldn't find AT_RANDOM - can't display canary."))
         return
 
-    print(
-        message.notice("AT_RANDOM = %#x # points to (not masked) global canary value" % at_random)
-    )
+    print(message.notice("AT_RANDOM = %#x # points to global canary seed value" % at_random))
```
disconnect3d: Can we use f'string-formatting'?

disconnect3d:
Hi, thanks for the PR! I've added a couple of comments.

When we are on it, can you also add a testcase for this, at least for x86-64?

Did u also validate that this works properly on other architectures?

disconnect3d:
Hi, there are two docs check fixes to be done:
```
WARNING -  griffe: pwndbg/commands/canary.py:35: No type or annotation for returned value 'tuple'
WARNING -  griffe: pwndbg/commands/canary.py:59: No type or annotation for returned value 'int'
```

disconnect3d:
@k4lizen can u help here and suggest changes for the docs issue?

k4lizen:
yeah, add (return value) type annotations to the `find_tls_canary_addr` and `canary_value` functions. griffe sees that the functions are supposed to return something from the docstring, but it is complaining that it doesn't know the type of the thing (it can be specified either in the docstring or in the function type annotation, but it's cleaner to specify with function type annotation)

tests/gdb-tests/tests/test_command_canary.py
```diff
@@ -0,0 +1,48 @@
...
+    Tests the canary command for x86-64 and i386 architectures
+    """
+    start_binary(binary)
+
+    gdb.execute("break main")
+    gdb.execute("run")
+    gdb.execute("stepi")
+
+    register = getattr(pwndbg.aglib.regs, reg_name)
+    canary_value, at_random = pwndbg.commands.canary.canary_value()
```
disconnect3d, line 34: Fwiw shouldn't this test actually do sth like `output = gdb.execute("canary", to_string=True)` and assert on that?

disconnect3d:
Fwiw this is how it looks now:
image

Random comment: we could color the value on the right with our normal/standard coloring of values so it is colored according to the legend. The text could also not be colored at all. Just a nit.

PR #2945, opened 2025-05-03
PLT command - search all .plt.* sections

OBarronCS:
Binaries compiled with the `-fcf-protection` emit a `.plt.sec` section, which functions identically to `.plt` but the instructions include control flow integrity tracking, such as the `endbr64` instruction in x86.

Example:
image

Based on this blog post, looks like the new section was introduced due to the plt stub size increasing with the new instructions, so the structure no longer fit in the old format
https://maskray.me/blog/2021-09-19-all-about-procedure-linkage-table#:~:text=add%20a%20new%20section%20.plt.sec

In this PR, the `plt` command now searches the `.plt.sec` section in addition to the `.plt` section.

image

If no .plt.sec section is found but .plt is found, it doesn't print about .plt.sec: 
image

If neither are found, it prints this:
image

### DIFF
--- a/pwndbg/commands/elf.py
+++ b/pwndbg/commands/elf.py
@@ -47,12 +47,12 @@ def gotplt() -> None:
 
 PLT_SECTION_NAMES = (".plt", ".plt.sec", ".plt.got", ".plt.bnd")
 
+
 @pwndbg.commands.Command(
     "Prints any symbols found in the .plt section if it exists.", category=CommandCategory.LINUX
 )
 @pwndbg.commands.OnlyWithFile
 def plt() -> None:
-
     local_path = pwndbg.aglib.file.get_proc_exe_file()
 
     bin_base_addr = 0
@@ -68,7 +68,7 @@ def plt() -> None:
 
         for section_name in PLT_SECTION_NAMES:
             section = elffile.get_section_by_name(section_name)
-            
+
             if section:
                 start = section["sh_addr"]
                 size = section["sh_size"]
@@ -97,7 +97,7 @@ def plt() -> None:
                     print(hex(int(addr)) + ": " + symbol)
 
     if not found:
-        print(message.error(f"No .plt.* sections found"))
+        print(message.error("No .plt.* sections found"))
 
 
 def get_section_bounds(section_name: str):
@@ -116,14 +116,12 @@ def get_section_bounds(section_name: str):
         return (start, start + size)
 
 
-def print_symbols_in_section(section_name, filter_text=""):
-    """
-    Return True if any symbols were found
-    """
+def print_symbols_in_section(section_name, filter_text="") -> None:
     start, end = get_section_bounds(section_name)
 
     if start is None:
         print(message.error(f"Could not find section {section_name}"))
+        return
 
     # If we started the binary and it has PIE, rebase it
     if pwndbg.aglib.proc.alive:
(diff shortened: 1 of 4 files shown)


OBarronCS:
Looks like there are four sections like this in total  (based on GDB code that handles PLT stuff)-  https://github.com/bminor/binutils-gdb/blob/38d726a24c1a85abdb606e7ab6cefad17872aad7/bfd/elf64-x86-64.c#L5775-L5780

`.plt`
`.plt.got`           # which is different from .got.plt
`.plt.sec`
`.plt.bnd`          # old version of .plt.sec it appears - https://ecos.sourceware.org/ml/binutils/2017-05/msg00139.html

That link has the reason for the name `plt.sec` - "Rename .plt.bnd to .plt.sec to indicate that this is used as the second
PLT section."

All which handle plt stuff but with different levels of lazy-binding/different types of instructions

OBarronCS:
image

OBarronCS:
Updated the output so if no sections are found, it looks like this:
image

But if the sections are found / symbols are found, they are printed
image

disconnect3d:
> image

One minor comment: the sections display could probably be ordered by their start/end address?

pwndbg/commands/elf.py
```diff
@@ -42,15 +45,60 @@ def gotplt() -> None:
     print_symbols_in_section(".got.plt", "@got.plt")
```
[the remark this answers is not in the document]
disconnect3d: Good idea to add, maybe with links?

OBarronCS:
That's a good call - made this change so the sections are ordered from low-to-high address when printing.

image

pwndbg/commands/elf.py
```diff
@@ -42,15 +45,71 @@ def gotplt() -> None:
...
+
+                sections_found.append((section_name, start, end))
+
+    # Sort by the start address so we print from lowest to highest
+    sections_found.sort(key=lambda x: x[1])
+
+    for section_name, start, end in sections_found:
+        symbols = get_symbols_in_region(start, end, "@plt")
+
+        print(message.notice(f"Section {section_name} {start:#x}-{end:#x}:"))
```
k4lizen:
```suggestion
        print(message.notice(f"Section {section_name} {start:#x} - {end:#x}:"))
```
This way a user can double click an address to select it.

pwndbg/commands/elf.py
```diff
@@ -42,15 +45,71 @@ def gotplt() -> None:
...
 
 
+# These are all the section names associated with PLTs.
+# .plt.sec and .plt.bnd are associated with control flow transfer integrity.
+# These are derived from this list that GDB recognizes: https://github.com/bminor/binutils-gdb/blob/38d726a24c1a85abdb606e7ab6cefad17872aad7/bfd/elf64-x86-64.c#L5775-L5780
+PLT_SECTION_NAMES = (".plt", ".plt.sec", ".plt.got", ".plt.bnd")
+
+
 @pwndbg.commands.Command(
     "Prints any symbols found in the .plt section if it exists.", category=CommandCategory.LINUX
```
k4lizen: Maybe this could be fixed up a bit since it doesn't only look at .plt anymore.

OBarronCS: I updated the comment to be more generic 👍

Issue #2887, opened 2025-04-17
Missing handling of /usr/lib/debug in portable release


patryk4815:
User will see:
<img width="458" alt="Image" src="https://github.com/user-attachments/assets/42fe3b39-d303-4d10-8b21-47779b490f31" />

After adding undocumented env `export NIX_DEBUG_INFO_DIRS=/usr/lib/debug` pwndbg-gdb starts reading `/usr/lib/debug` directory.
<img width="567" alt="Image" src="https://github.com/user-attachments/assets/7c3b22ef-82a8-4b27-82c8-a22f22068ac1" />

Btw we have similar code for handling `/usr/lib/debug` inside pwndbg:
https://github.com/pwndbg/pwndbg/blob/411ffd93700e7be3c3da67e796e72991e9dca834/pwndbg/dbg/gdb/debug_sym.py#L4
This solve only problem for `file /binary` after pwndbg dbg session has been started, not for passing binary using gdb args (`gdb /binary`).

Todo:
- remove `debug_sym.py` at all? - or maybe somehow we can force gdb first load pwndbg and then loads binary?
- embed env `NIX_DEBUG_INFO_DIRS=/usr/lib/debug` into portable gdb 
- fix weird message about `UTF-32` conversion

disconnect3d:
What is /usr/lib/debug? Do I understand that without it the portable gdb cannot find a main function in the binary u debug that has debug symbols? Why, huh?

patryk4815:
Our portable gdb use different path for searching debug-file-directory - it is inherit from nixpkgs.

Eg. `/usr/lib/debug` is where Ubuntu stores debug symbols for stripped binaries.

When you install something like `apt install hello`, it comes without debug info. 
To get it, you install `apt install hello-dbgsym`, which puts the symbols in `/usr/lib/debug`.

On Ubuntu, GDB is compiled with `/usr/lib/debug` set as the default `debug-file-directory`. 
Other distros might use different paths, but most also use /usr/lib/debug for debug symbols.

disconnect3d:
Fixed in https://github.com/pwndbg/pwndbg/pull/2947

Issue #2774, opened 2025-03-06
Agnostic .pwndbginit for configuration


patryk4815:
Continuation from - https://github.com/pwndbg/pwndbg/issues/2225#issuecomment-2403437242

Currently, when using our portable build, users cannot add pwndbg-configuration options to `.gdbinit`

A possible solution could be to introduce a `.pwndbginit` file where users can set their options after Pwndbg is loaded.

This is open for discussion regarding the best approach.

Other solution?
```
gdb -nx \
  --init-eval-command='source ~/.gdbinit-before' \
  --command=/home/pwndbg/gdbinit.py \
  --eval-command='source ~/.gdbinit-after'
```

Example problems:
- fixed in https://github.com/pwndbg/pwndbg/pull/2912 -  double loading `source /other/pwndbg/gdbinit.py` one added in `.gdbinit` and another loading from portable pwndbg `--command=/portable/pwndbg/gdbinit.py` 
- loading out of tree custom pwndbg commands/script, eg: `source /out-of-tree/pwndbg-foo-command.py`
<img width="688" alt="Image" src="https://github.com/user-attachments/assets/54a80ad3-a454-4c67-94e9-acc5c4c751ef" />

- loading pwndbg configuration, eg: `set show-tips off`
<img width="700" alt="Image" src="https://github.com/user-attachments/assets/e0c02cff-f28f-44a3-a2eb-d30151d86e6f" />

b1nhack:
This issue exists in pwndbg installed via Nix. After the Nix build, an executable script called `pwndbg` is output. This script calls gdb and passes the following argument: `--command=/nix/store/xxxx-pwndbg/share/pwndbg/gdbinit.py`. This is the root cause of the problem. Pwndbg always loads after `.gdbinit`. I completely understand the pwndbg team's intention to help users quickly install without issues, but perhaps we should also output the gdb executable after the Nix build, so users can choose to manually load pwndbg in `.gdbinit` and avoid this issue. In fact, I installed an additional gdb via Nix to resolve the problem.

disconnect3d:
Yup, that makes sense and we are aware of this problem. As stated in the issue, an alternative solution could be to load `.pwndbginit` file instead of `.gdbinit`. This could solve the described issue, but at the same time it could introduce two other issues:
1) How would users find out about this file? It makes it harder to know about it etc.
2) Should it work in both GDB and LLDB? Or should we have a separate one for each one? etc.

@b1nhack what do you think about this?

[...]

PR #2947, opened 2025-05-04
Fix .gdbinit for portable pwndbg

patryk4815:
**Closes**: 
- #2774 
- #2887 

**Changes**:
- GDB now starts with the `--nx` option to prevent automatic loading of any `.gdbinit` files.
- On startup, we manually load `.gdbinit` files following logic similar to GDB's native behavior.
- Replaced the `--command=gdbinit.py` option with `--init-command=gdbinit.py`, since `init-command` runs before the inferior is started.

**Before:**
- GDB would automatically load `.gdbinit` (both home and local) **before** pwndbg was initialized.

**After:**
- GDB does **not** load any `.gdbinit` files (due to `--nx`).
- Instead, pwndbg implements its own logic to load `.gdbinit` files in the same order and manner as GDB.

**Implications:**
- This change allows `.gdbinit` files (e.g., user configurations) to be loaded **after** pwndbg is fully initialized.
- It also enables disabling or overriding default pwndbg settings (like `set follow-fork-mode child`) **only in pwndbg-portable mode** 

**Question:**
Should we allow users to override default settings (e.g., `set follow-fork-mode child`)?
Previously, this was not possible because GDB loaded `.gdbinit` before pwndbg.  
Now, with pwndbg handling `.gdbinit` loading (in portable mode), such overrides are possible — but only in pwndbg-portable

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -75,8 +75,6 @@ or
 ./pwndbg/bin/pwndbg-lldb
 ```
 depending on which version you installed. You may add the appropriate file to your shell's PATH.
-!!! warning ".gdbinit doesn't work for portable release"
-    If you're running `./pwndbg/bin/pwndbg` from the portable release, it is a known limitation that pwndbg settings in your `.gdbinit` won't work (see [issue #2774](https://github.com/pwndbg/pwndbg/issues/2774)). Also, make sure not to source pwndbg in your gdbinit as it already happens automatically for portable releases (if this is the first time you're installing pwndbg, you don't need to worry about this).
 
 ## Installing from source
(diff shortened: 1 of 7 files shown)


disconnect3d:
> Should we allow users to override default settings (e.g., set follow-fork-mode child)?

Yes, of course! The user should be able to set any configs they want, even in portable mode.

docs/setup.md
```diff
@@ -75,8 +75,6 @@ or
 ./pwndbg/bin/pwndbg-lldb
 ```
 depending on which version you installed. You may add the appropriate file to your shell's PATH.
-!!! warning ".gdbinit doesn't work for portable release"
```
disconnect3d, line 78: This is an important note we need to write about when releasing this! I mean that this now works :D

disconnect3d (approved):

k4lizen:
this broke doc generation xd
image

is this behaviour intended? I can confirm the exit is happening in `_load_gdbinit()`

PR #2960, opened 2025-05-07
Show aliases in command help and docs

k4lizen:
## Before
image
image
image
## After
image
image
image


k4lizen:
wut
image

k4lizen:
We could also try to get a more gdb-like output:
image
where the aliases are printed in the first line, though that would require overriding the argparse format_help function.

disconnect3d:
> We could also try to get a more gdb-like output (...) where the aliases are printed in the first line, though that would require overriding the argparse format_help function.

Both current and this works for me. One nitpick - maybe remove the dot at the end of alias list?

disconnect3d (approved):

PR #2962, opened 2025-05-07
Don't print all pwndbg function on startup

k4lizen:
This is kind of opinionated so let me know what you think, but it seems to me this print was created when there weren't as many functions. The line is kind of long and easily wraps the terminal when the width is small, so this makes it shorter.

Before:
image
After:
image

### DIFF
--- a/pwndbg/gdblib/prompt.py
+++ b/pwndbg/gdblib/prompt.py
@@ -55,9 +55,9 @@ last_alive_state = False
 def show_hint() -> None:
     hint_lines = (
         f"loaded {len(pwndbg.commands.commands)} pwndbg commands."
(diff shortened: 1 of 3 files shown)


k4lizen:
wth
image
xddd?

disconnect3d (approved):

disconnect3d:
Btw shall `help function` mark pwndbg functions somehow? like `[pwndbg]` in their description? Idk

k4lizen:
@disconnect3d 

since 'help function' is a GDB thing i'm not sure we can easily control the output. We could prepend each description with "[pwndbg] " and get something like this
image

but then
image

printing the help would also print it. which, you know, might be okay?

PR #2963, opened 2025-05-08
Manually propagate register values across instructions while disassembling

OBarronCS:
This PR adds the ability to manually keep track of register values at the moment of each instruction's execution during disassembly.

Typically, the Unicorn emulator is used for this - when we need the value of a register at a future point in time when annotating a future instruction, we read from the emulator. However, there are contexts where we can't do this, such as after a syscall (we can't emulate syscalls), or when disassembling at locations that are not at the current process instruction pointer (we can't emulate at arbitrary places, because we need to copy the current processor state to instantiate the emulator.)

Instead, this PR allows us to manually keep track of register values as they are moved from register-to-register, mainly in the form of `mov`-like instructions. For example, in a case we have `mov rax, 1`, we can assume for all the instructions after this that `rax` has the value `1`, assuming it is not clobbered by another instruction. This allows the enhancement process to read from `rax` in future instructions.

### Examples

A visible example of this is showing syscall names in cases where Unicorn emulation cannot be used to determine the syscall number (we disable the emulator once we hit a syscall / function call)

Before: note that only the first syscall name is shown, after which Unicorn emulation is disabled and we can no longer reason about register values:
image

After: there are constant values `mov`ed into `rax`, and we can propagate the value in this register into the subsequent instructions. This means that when it comes time to enhance each syscall, we can read `rax` to determine the syscall number:
image

Example: disassembling glibc clone function, no emulation. Note that `syscall` names are shown
image

Another example of disassembly with `set emulate off`

Before:

image

After:
image

In the first instruction, we know the constant value written to `rsi`. In the following two `mov` instructions, this value is propagated to `rax` and `rdi`.

In the fourth instruction, there is an `xor edx, edx`. Since writes to edx zero-extend to `rdx`, in the following instruction we know what value is `mov`ed to r9.

The last two `mov` instructions highlight how we can move a value to a register (`eax`), and later on read from the subregisters (`al`) correctly.


pwndbg/lib/regs.py
```diff
@@ -166,7 +166,7 @@ def __iter__(self) -> Iterator[str]:
         yield from self.all
 
 
-class PsuedoEmulatedRegisterFile:
+class PseudoEmulatedRegisterFile:
```
OBarronCS, line 291: Fixed a typo: "Psuedo" -> "Pseudo"

pwndbg/aglib/disasm/arch.py
```diff
@@ -513,9 +551,7 @@ def _telescope(
                 if page and not page.write:
                     try:
                         address = int(
-                            pwndbg.aglib.memory.get_typed_pointer_value(
-                                pwndbg.aglib.typeinfo.ppvoid, address
-                            )
+                            pwndbg.aglib.memory.get_typed_pointer_value(size_type, address)
```
OBarronCS, line 560: This codepath was not respecting the `read_size` parameter previously.

k4lizen:
does this PR make this kinda incorrect?
https://pwndbg.re/pwndbg/dev/contributing/improving-annotations/#when-to-use-emulation-reasoning-about-process-state
https://github.com/pwndbg/pwndbg/blob/e83dee14c24f5f4a7bd66d2e2d214f04debb4be8/docs/contributing/improving-annotations.md?plain=1#L38-L59

disconnect3d:
@OBarronCS ping :D can we finalize this one way or another? :P

OBarronCS:
Yes - I made the last changes for correctness and did more testing locally on x86. It's ready to go! It will currently be enabled for x86.

PR #2964, opened 2025-05-08
Argument display fixes

OBarronCS:
1. If we detect the argument as a file descriptor, we enhance it with the underlying filename. Previously, it looked for the parameter name `fd` - this adds a couple other names that refer to file descriptors as found in this syscall guide: https://syscalls.mebeim.net/?table=x86/64/x64/v6.6

image

2. Show bitflag argument constant names in context (previously only shown in `dumpargs`) - referring to the `flags` argument in this example
image

### DIFF
--- a/pwndbg/arguments.py
+++ b/pwndbg/arguments.py
@@ -159,6 +159,8 @@ def arguments(abi: pwndbg.lib.abi.ABI | None = None):
         yield argname(i, abi), argument(i, abi)
 
 
+# When an argument is named one of these in Linux syscalls/glibc, it refers to a file descriptor
+# Search for strings containing "fd" in https://chromium.googlesource.com/chromiumos/docs/+/master/constants/syscalls.md
 FILE_DESCRIPTOR_ARG_NAMES = {
     "fd",
     "in_fd",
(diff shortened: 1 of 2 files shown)


pwndbg/arguments.py
```diff
@@ -158,14 +159,36 @@ def arguments(abi: pwndbg.lib.abi.ABI | None = None):
...
-        pretty = pwndbg.chain.format(value, code=code)
+        pretty = (
+            pwndbg.chain.format(value, code=code)
+            if not arg.flags
+            else format_flags_argument(arg.flags, value)
+        )
 
         # Enhance args display
-        if arg.name == "fd" and isinstance(value, int):
+        if arg.name in FILE_DESCRIPTOR_ARG_NAMES and isinstance(value, int):
```
disconnect3d, line 191: On a long run such enhancing should probably be done on a per-function basis instead of here ( ? ).

Maybe we shall create an issue for that as 'code cleanup' idk...

We could attach such 'enhancers' or display helpers to the function objects or whatever (?). Loud thinking here.

disconnect3d:
Fwiw may be nice to add a testcase for this if u could implement one

PR #2966, opened 2025-05-09
Enable line-too-long lint

k4lizen:
Currently we ignore the `E501` [line-too-long](https://docs.astral.sh/ruff/rules/line-too-long/)  linting rule.

That is kinda bad since that is one of the more important linting rules in a codebase. This happened because when we were switching to ruff for linting/formatting the rationale went like: "We have a lot of important lint rules that we fail, and we fail them in many places. Over time we will fix them, and new code will be conformant, so we will enable them later".

However in practice, if your IDE/editor doesn't yell at you, and our CI doesn't yell at you, you will inevitably write non-conformant code since you don't have all the lint rules memorized/on your mind all the time. As a result, the rules that we currently ignore have not been getting better, but rather the situation has been getting worse and worse.

If we look at E501 in particular, there are 735 violations in the current dev branch, and this won't get any better.

So I wanted to tackle this. However, unlike with `clang-format`, there isn't a good python tool that formats code so it conforms to a specified line length.
[ruff doesn't do it](https://github.com/astral-sh/ruff/issues/6936)
[black kinda does it](https://black.readthedocs.io/en/stable/the_black_code_style/future_style.html#improved-string-processing):
1. it is an unsable feature
2. there isn't an option in black to *only* do line-length formatting, and running black on the codebase causes ~8k LOC changes.
3. it doesn't even work, it only formats some lines (less than half at a certain point that I tried)

[autopep8 is bad at it](https://pypi.org/project/autopep8/) - it only touches some lines, gets borked at format strings. it's also an experimental feature
[pydocstringformatter](https://pydocstringformatter.readthedocs.io/en/latest/) - doesn't even format the whole docstring much less align everything
[docformatter](https://pypi.org/project/docformatter/) - same story

In general there is nothing that handles comments, strings and docstrings properly, or even just one of those. At least from what I tested / could find.

In the end, I used some mix-match of the available python tooling and reflowed a lot (most) of the lines manually. Yes it was painful.


k4lizen:
The text in some docs now has some new-lines, but it shouldn't be reflected on the site since it's markdown.

For review take note of pyproject.toml .

mkdocs.yml
```diff
@@ -129,9 +129,9 @@ plugins:
                 warn_unknown_params: true
 
   # Generate the whole Source page magically (hooks into mkdocstrings)
-  - api-autonav:
-      modules: ['pwndbg']
-      nav_section_title: "Source"
+  # - api-autonav:
+  #     modules: ['pwndbg']
+  #     nav_section_title: "Source"
```
disconnect3d: Is this ok?

k4lizen: oh whooops

giampaolo:
Nice write up. FWIW black's string_processing option works pretty good. Its the only reason why I still use black, which I use in conjunction with ruff (linter, not formatter).

disconnect3d:
I think we will skip it for now? :/

I am not super in favor of limiting long lines and its partially done already by some of the lint heh, it also got a bit stale since we didn't merge early :(.

Issue #2970, opened 2025-05-10
Improve performance of string colorizing


OBarronCS:
When we want to apply a color to a string, we have functions like this that we call:

https://github.com/pwndbg/pwndbg/blob/98b7d6c5e22e0c89e62f62e402911ada21707ae1/pwndbg/color/context.py#L49-L50

On a single context printout, there can be around ~300 calls to the `generateColorFunction` call, which dynamically creates a function. This adds up to a performance hit. 

One solution is to do:
```python
highlight = generateColorFunction(config.highlight_color)(x)
```

However, this would mean that changes to `set highlight-color` would not apply at runtime. Maybe this is okay? Or, these functions can be changed to not generate a function just to call it immediately?

Although not directly related, while fixing this, it may be natural to fix this: https://github.com/pwndbg/pwndbg/issues/2874

patryk4815:
`set highlight-color` can implement `@trigger` and override global `highlight_func` variable?

OBarronCS:
> `set highlight-color` can implement `@trigger` and override global `highlight_func` variable?

Nice! This works locally:
```python
highlight = generateColorFunction(config.highlight_color)

@config.trigger(config_highlight_color)
def test():
    global highlight
    highlight = generateColorFunction(config.highlight_color)
```
There are, although, +50 occurrences of these function creations, which adds a lot of boilerplate.

disconnect3d:
We can probably apply triggers dynamically can't we?

OBarronCS:
Awesome - I dynamically add a trigger to update the functions in #2977

OBarronCS:
One round of optimizations applied in #2977. See https://github.com/pwndbg/pwndbg/pull/2977#discussion_r2091978493 for further optimization ideas.

disconnect3d:
What is the current timing/profiling for string colorizing?

I did some test and compiled Pygments with Cython and got like 2x speed up on some artificial benchmark:

![Image](https://github.com/user-attachments/assets/8789edde-e04b-481e-9a48-48af6156b1c5)

I believe it was on this script: [message (1).txt](https://github.com/user-attachments/files/20512306/message.1.txt)

I may continue on this and release a Cythonized version of Pygments and we could use it instead? Or, we could cythonize this in release builds. Not sure.

PR #2977, opened 2025-05-15
Cache string coloring functions

OBarronCS:
Address #2970 by caching the result of `generateColorFunction`, and only generating a new function when setting the parameter (`set context-register-color ...`). On my machine, I found that this reduces the time to print the context by ~0.5 milliseconds on average.

### DIFF
--- a/pwndbg/color/__init__.py
+++ b/pwndbg/color/__init__.py
@@ -164,7 +164,7 @@ class ColorParamSpec(NamedTuple):
 class ColorConfig:
     def __init__(self, namespace: str, params: List[ColorParamSpec]) -> None:
         self._namespace = namespace
-        self._params: Dict[str, Parameter] = {}
+        self._params: Dict[str, theme.ColorParameter] = {}
         for param in params:
             self._params[param.name] = theme.add_color_param(
                 f"{self._namespace}-{param.name}-color", param.default, param.doc
@@ -173,7 +173,7 @@ class ColorConfig:
     def __getattr__(self, attr: str) -> Callable[[str], str]:
         param_name = attr.replace("_", "-")
         if param_name in self._params:
-            return generateColorFunction(self._params[param_name])
+            return self._params[param_name].color_function
 
         raise AttributeError(f"ColorConfig object for {self._namespace} has no attribute '{attr}'")
 
--- a/pwndbg/color/context.py
+++ b/pwndbg/color/context.py
@@ -2,8 +2,6 @@ from __future__ import annotations
 
 from typing import List
 
-from pwndbg import config
-from pwndbg.color import generateColorFunction
 from pwndbg.color import theme
 from pwndbg.lib.regs import BitFlags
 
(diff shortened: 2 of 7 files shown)


pwndbg/color/theme.py
```diff
@@ -32,5 +41,11 @@ def add_param(
     )
 
 
-def add_color_param(name: str, default: Any, set_show_doc: str) -> Parameter:
-    return config.add_param_obj(ColorParameter(name, default, set_show_doc, scope=Scope.theme))
+def add_color_param(name: str, default: Any, set_show_doc: str) -> ColorParameter:
+    color_parameter = ColorParameter(name, default, set_show_doc, scope=Scope.theme)
+
+    config.triggers[name].append(lambda: color_parameter.update_color_function())
```
disconnect3d:
```suggestion
    config.triggers[name].append(color_parameter.update_color_function)
```
Shouldn't this be like this?

pwndbg/color/__init__.py
```diff
@@ -173,7 +173,7 @@ def __init__(self, namespace: str, params: List[ColorParamSpec]) -> None:
     def __getattr__(self, attr: str) -> Callable[[str], str]:
         param_name = attr.replace("_", "-")
```
disconnect3d, line 174: Loud thinking: Do we even have attr with "-" here?

pwndbg/color/context.py
```diff
@@ -43,51 +41,51 @@
 
 
 def prefix(x: object) -> str:
-    return generateColorFunction(config.code_prefix_color)(x)
+    return config_prefix_color.color_function(x)
```
disconnect3d, line 44: I am wondering, what is the call trace here? Let's assume "prefix" does "red" color.

I think the call trace is:

prefix(x) -> config_prefix_color.color_function(x) -> wrapper (from generateColorFunctionInner) which does old(new(x)) - that is red(str(x)) -> colorize(x, RED)

Whereas it should really end up being just: prefix(x) -> return f"{RED}{x}{END}"

Ugh.

I am wondering if we can do it better, maybe like this...:
```py
In [20]: def prefix(x: object) -> str:
    ...:     return f"A{x}B"
    ...:

In [21]: prefix.__code__.co_consts
Out[21]: (None, 'A', 'B')

In [22]: prefix(1337)
Out[22]: 'A1337B'

In [23]: prefix.__code__ = prefix.__code__.replace(co_consts=(None, '<red>', '</red>'))

In [24]: prefix(1337)
Out[24]: '<red>1337</red>'
```

Just to be clear, I am not saying we have to do any of it in this PR or at all, but on a long term, this is probably the optimal way of doing all this bullshit with updating color functions AND keep them performant - aka - with no stupid redundant calls?

disconnect3d: Then each color change would be just sth like this?:

`given_coloring_func.__code__ = stub_coloring_func.__code__.replace(co_consts=(None, COLOR_CODE, END_COLOR_CODE))`

pwndbg/color/theme.py
```diff
@@ -32,5 +41,11 @@ def add_param(
...
+
+    config.add_param_obj(color_parameter)
```
disconnect3d, line 49: Cant' we do `return config.add_param_obj(...)` like previously?

OBarronCS: add_param_obj has typehint to return the parent class `Parameter`, so this avoids the type error as this function returns a ColorParameter.

I played around with getting a generic type working, so that the function would accept subclasses of Parameter and returns the same type. Something like:

```python
class A:
    T = TypeVar("T", bound=Parameter)
    def add_param_obj[T](self, param: T) -> T: 
          ....
          return param
```

But mypy and my code editor have different levels of understanding this syntax, causing strange behavior (one of them errors (red underline) when the other doesn't). I think it's simpler to return the object explicitly to avoid this.

pwndbg/color/__init__.py
OBarronCS: There's a couple cases this occurs, like this:
https://github.com/pwndbg/pwndbg/blob/c56cc4c033e57593e88ae006cba312a41e7a90d2/pwndbg/commands/context.py#L67

On access, we do `object.frame_label`.

pwndbg/color/context.py
OBarronCS: That is really cool! This would be very interesting - the long chain of functions just to add a color to a string is not ideal at all.

Some of the functions have multiple "colors" applied, like red and bold, which manifests as nested functions like `bold(red(x))` (and in theory one could apply as many colors/effects as they want, but this can probably be extended to work in those cases - and if not, 99% of color functions apply just one color, so this would work great)

Issue #2974, opened 2025-05-12
Better method of detecting embedded target


OBarronCS:
Pwndbg's telescoping logic largely relies on the existance of virtual memory pages which allows us to provide colors to pointers and to know when a value is not a pointer.

Processes on embedded systems/bare metal, however, typically have flat memory models, which requires us to handle telescoping slightly differently:

1. Fine-grained paging doesn't exist, so all memory is "valid" and often initialized to zero, cause every value to appear to be a pointer to 0.
2. On embedded systems, it's not uncommon for MMIO regions to exist where
    memory reads might mutate the hardware/process state.

If we determine we are on an embedded system, we avoid making stray memory dereferences to protect against these cases and provide nicer output.

https://github.com/pwndbg/pwndbg/blob/da297308b603caee01044d291f99b6b8f194c494/pwndbg/chain.py#L76

Currently, Pwndbg uses `pwndbg.gdb.selected_inferior().is_linux()` as a stand-in for "we are on an embedded system", around which we put these checks to avoid incorrect memory dereferences. 

https://github.com/pwndbg/pwndbg/blob/da297308b603caee01044d291f99b6b8f194c494/pwndbg/dbg/gdb/__init__.py#L854-L860

`show osabi` is an unreliable indicator, of this and the output is dependent on the version of GDB in use (depending on how GDB is compiled, "GNU/Linux" may not be a valid OS ABI).

More accurately determining if the debugging target is an embedded system will fix incorrect telescoping and fix issues with memory reads in sensitive MMIO regions.

See a previous issue related to this: https://github.com/pwndbg/pwndbg/pull/385

This will fix: #2806

PR #2978, opened 2025-05-15
Add cache to `is_linux` call

OBarronCS:
Due to it's placement in code that dereferences memory,  `pwndbg.dbg.selected_inferior().is_linux()` is being called nearly 100 times during a single context printout. Previously, the value was not cached, resulting in performance hit (~1.5 ms) when printing the context. This is now cached.


pwndbg/aglib/baremetal.py
```diff
@@ -0,0 +1,22 @@
+from __future__ import annotations
+
+import pwndbg
+import pwndbg.lib.cache
+
+
+@pwndbg.lib.cache.cache_until("start", "objfile")
+def is_baremetal() -> bool:
```
OBarronCS: I made this function to set the groundwork for #2974 - also, our @ cache decorator doesn't work with class methods, so I couldn't add it directly to `pwndbg.dbg.Process.is_linux()` method.

patryk4815:
I think this should be config option, eg: `set platform baremetal/vmcore-linux/kgdb-linux/qemu-system-linux/userspace-linux` 
Default it could be `auto` option, with detection on first connection.
Right now we are not supporting `kgdb` well or `vmcore`, only qemu-system :/

Or also we should have different option `set is-pagefault-supported on/off/auto`?

pwndbg/aglib/baremetal.py
patryk4815: maybe this should be better name `is_pagefault_supported`?

OBarronCS: That's a way better name - making this change. I also moved it to `pwndbg.aglib.memory.py`, because that makes a bit more sense in this context.

pwndbg/aglib/memory.py
```diff
@@ -414,3 +414,20 @@ def resolve_renamed_struct_field(struct_name: str, possible_field_names: Set[str
...
     raise ValueError(f"Field name did not match any of {possible_field_names}.")
+
+
+@pwndbg.lib.cache.cache_until("start", "objfile")
+def is_pagefault_supported() -> bool:
+    """
+    This function should be called before stray memory dereferences to protect against the following situations:
+
+    1. On embedded systems, it's not uncommon for MMIO regions to exist where memory reads might mutate the hardware/process state.
+    2. On baremetal/embedded, paging doesn't always exist, so all memory is "valid" (and often initialized to zero) - this makes everything value appear to be a pointer.
```
disconnect3d:
```suggestion
    2. On baremetal/embedded, paging doesn't always exist, so all memory is "valid" (and often initialized to zero) - this makes every value appear to be a pointer.
```

pwndbg/aglib/memory.py
```diff
@@ -414,3 +414,20 @@ def resolve_renamed_struct_field(struct_name: str, possible_field_names: Set[str
...
+
+    As such, we disable dereferencing by default for bare metal targets.
+
+    See more discussion here: https://github.com/pwndbg/pwndbg/pull/385
+    """
+
+    # TODO: use a better detection method
+    return pwndbg.dbg.selected_inferior().is_linux()
```
disconnect3d, line 433:
```
dc@jctf:~/pwndbg$ rg 'def is_linux' -A40
pwndbg/dbg/gdb/__init__.py
841:    def is_linux(self) -> bool:
842-        # Detect current ABI of client side by 'show osabi'
843-        #
844-        # Examples of strings returned by `show osabi`:
845-        # 'The current OS ABI is "auto" (currently "GNU/Linux").\nThe default OS ABI is "GNU/Linux".\n'
846-        # 'The current OS ABI is "GNU/Linux".\nThe default OS ABI is "GNU/Linux".\n'
847-        # 'El actual SO ABI es «auto» (actualmente «GNU/Linux»).\nEl SO ABI predeterminado es «GNU/Linux».\n'
848-        # 'The current OS ABI is "auto" (currently "none")'
849-        #
850-        # As you can see, there might be GDBs with different language versions
851-        # and so we have to support it there too.
852-        # Lets assume and hope that `current osabi` is returned in first line in all languages...
853-        abi = gdb.execute("show osabi", to_string=True).split("\n")[0]
854-
855-        # Currently we support those osabis:
856-        # 'GNU/Linux': linux
857-        # 'none': bare metal
858-
859-        return "GNU/Linux" in abi
```

This should really be fetched once and never ever until starting a process again or attaching to another target.

Do we really want to have this as a function too? Maybe this should be just a field that is never changed by the user and just used read across the code?

pwndbg/chain.py
```diff
@@ -59,6 +59,9 @@ def get(
     limit = int(limit)
 
     result = [address] if include_start else []
+
+    is_pagefault_supported = pwndbg.aglib.memory.is_pagefault_supported()
```
disconnect3d, line 63: If we care about perf, this is a func call anyway and a cache lookup... maybe it should be just `pwndbg.dbg.selected_inferior().is_pagefault_supported`?

At the same time, ideally this should be just `something_not_a_call_or_property.is_pagefault_supported` (aka one attr lookup on an object instead of `selected_inferior()` call ;().

pwndbg/aglib/memory.py
patryk4815: sometime I change this during debugging with:
`set osabi none`

disconnect3d:
Going to merge this but I think we can do this even more robust/faster - see comments and let me know if that makes sense.

pwndbg/chain.py
OBarronCS: Definitely agree with being wary of unnecessary use of @ cache for things that really should be static, like whether or not the process supports page faults.

pwndbg/aglib/memory.py
OBarronCS: Seen similar cases like this function:
https://github.com/pwndbg/pwndbg/blob/2f60a3a4a93eb7d1e0a9617e22941cc8a08d70ba/pwndbg/dbg/gdb/__init__.py#L1707
(and many other functions on our `Debugger` classes) which are not cached, but used in other functions that themselves are cached. For some of these, perhaps we can make use of `set parameter=` triggers to set object attributes with the provided setting (like `global_process_settings_or_something.is_linux`), instead of using function calls that have the @ cache decorator?

PR #2979, opened 2025-05-16
A couple disassembly performance optimizations

OBarronCS:
Small tweaks to avoid some unnecessary function calls related to disassembly. See comments for specific details. Changes save ~1.5 ms per context call.

### DIFF
--- a/pwndbg/aglib/disasm/disassembly.py
+++ b/pwndbg/aglib/disasm/disassembly.py
@@ -18,6 +18,7 @@ from capstone import *  # noqa: F403
 
 import pwndbg
 import pwndbg.aglib.arch
+import pwndbg.aglib.disasm.arch
 import pwndbg.aglib.disasm.aarch64
 import pwndbg.aglib.disasm.arm
 import pwndbg.aglib.disasm.disassembly
@@ -98,11 +99,8 @@ emulated_arm_mode_cache: DefaultDict[int, int | None] = collections.defaultdict(
 
 
 @pwndbg.lib.cache.cache_until("objfile")
-def get_disassembler(address: int, cs_info: Tuple[int, int] = None):
-    if cs_info is not None:
-        arch, mode = cs_info
-    else:
-        arch, mode = pwndbg.aglib.arch.get_capstone_constants(address)
+def get_disassembler(cs_info: Tuple[int, int]):
+    arch, mode = cs_info
 
     mode |= CapstoneEndian[pwndbg.aglib.arch.endian]
 
@@ -136,9 +134,12 @@ def get_one_instruction(
 
     cs_info = pwndbg.aglib.arch.get_capstone_constants(address)
     if cs_info is None:
-        return ManualPwndbgInstruction(address)
+        instr = ManualPwndbgInstruction(address)
+        if enhance:
+            pwndbg.aglib.disasm.arch.basic_enhance(instr)
+        return instr
 
-    md = get_disassembler(address, cs_info)
+    md = get_disassembler(cs_info)
     data = pwndbg.aglib.memory.read(address, pwndbg.aglib.arch.max_instruction_size, partial=True)
     for ins in md.disasm(bytes(data), address, 1):
         pwn_ins: PwndbgInstruction = PwndbgInstructionImpl(ins)
--- a/pwndbg/aglib/disasm/disassembly.py
+++ b/pwndbg/aglib/disasm/disassembly.py
@@ -18,8 +18,8 @@ from capstone import *  # noqa: F403
 
 import pwndbg
 import pwndbg.aglib.arch
-import pwndbg.aglib.disasm.arch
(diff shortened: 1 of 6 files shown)


pwndbg/aglib/disasm/disassembly.py
```diff
@@ -98,11 +99,8 @@ def clear_on_reg_mem_change() -> None:
 
 
 @pwndbg.lib.cache.cache_until("objfile")
-def get_disassembler(address: int, cs_info: Tuple[int, int] = None):
-    if cs_info is not None:
-        arch, mode = cs_info
-    else:
-        arch, mode = pwndbg.aglib.arch.get_capstone_constants(address)
+def get_disassembler(cs_info: Tuple[int, int]):
```
OBarronCS, line 102: Change this function so it no longer takes address - this was messing with the cache behavior, resulting in creating the Capstone object multiple times.

pwndbg/emu/emulator.py
```diff
@@ -844,22 +844,23 @@ def single_step(self, pc=None) -> Tuple[int, int]:
 
         pc = pc or self.pc
 
-        insn = pwndbg.aglib.disasm.disassembly.one_raw(pc)
+        if check_instruction or DEBUG & DEBUG_EXECUTING:
```
OBarronCS, line 847: No need to check that this instruction is valid by default - in current usage, this function wouldn't even get called on an invalid instruction.

pwndbg/enhance.py
```diff
@@ -112,8 +112,9 @@ def enhance(
         rwx = exe = False
 
     if exe:
-        pwndbg_instr = pwndbg.aglib.disasm.disassembly.one(value)
+        pwndbg_instr = pwndbg.aglib.disasm.disassembly.one_raw(value)
```
OBarronCS, line 115: `one()` goes through the entire enhancement process, which does work unrelated to displaying an instruction while telescoping. 
image
In telescoping, we only care about syntax highlighting and replacing constants with their symbols.

OBarronCS:
Interesting - this somehow improves the disassembly of ARM IT blocks - the instructions in the IT block now have their condition attached to the instruction (`moveq`, `movne`).
image
Now that the Capstone object is persistent, capstone somehow now relates those instructions to the `IT` instruction before it? The Thumb2 instructions in the IT block, to my understanding, don't have the condition encoded into them - they are reliant on the IT instruction.

OBarronCS:
Very strange behavior:

When stepped on the IT block, all future instructions now have the condition codes in the assembly (even the nops, which are not in the IT block)
image

Stepping one down, however:
image

But `nearpc` only makes two of the instructions have the condition?? 
image

And disabling the regs section changes the output of disassembly (which could disassemble it when telescoping for PC reg)
image

Some internal state is being tracked in the Capstone disassembler object?

OBarronCS:
Made an issue on https://github.com/capstone-engine/capstone/issues/2702 related to this

OBarronCS:
Fixed the ARM IT block tests to reflect the new behavior. In short, before this PR the instructions were always missing their conditional suffix. After this PR, the disassembly is not correct in all cases in the context view (conditional suffixes are missing sometimes, sometimes too many instructions have the conditional suffix) but it is correct when disassembling with nearpc or emulate outside of the context view.

Making the disassembly completely correct in the context view would take larger changes that should be done in a different PR, but the change in behavior (accidentally) encountered here is a step in a better direction for IT instruction blocks.

pwndbg/aglib/disasm/disassembly.py
```diff
@@ -98,11 +99,8 @@ def clear_on_reg_mem_change() -> None:
...
+    arch, mode = cs_info
 
     mode |= CapstoneEndian[pwndbg.aglib.arch.endian]
```
disconnect3d, line 105: Btw why do we need to do this? Shouldn't the proper endian be already in place? Can we add a comment here?

pwndbg/aglib/disasm/disassembly.py
```diff
@@ -98,11 +99,8 @@ def clear_on_reg_mem_change() -> None:
 
 
 @pwndbg.lib.cache.cache_until("objfile")
-def get_disassembler(address: int, cs_info: Tuple[int, int] = None):
-    if cs_info is not None:
-        arch, mode = cs_info
-    else:
-        arch, mode = pwndbg.aglib.arch.get_capstone_constants(address)
+def get_disassembler(cs_info: Tuple[int, int]):
+    arch, mode = cs_info
```
disconnect3d, line 103: Random comment: since we unpack it anyway, maybe its better to just accept/pass `arch, mode`? Idk, just saying, maybe its better as it is.

pwndbg/emu/emulator.py
disconnect3d: Are u saying this func is never called? If so, why do we even have it here?

disconnect3d (approved):

PR #2980, opened 2025-05-16
Adding full buddy allocator support when debugging x86-64 linux kernels

jxuanli:
## Provide full support for buddy allocator when debugging x86-64 linux kernels
- achieves the original `pcplist`'s intended functionalities
  - `buddydump -p`
  - supports old and new kernel versions
- supports the debugging of shared free lists
- cross checked with `/proc/buddyinfo` and `/proc/proc/pagetypeinfo` and verified its correctness
- consistent with my understanding of how the page allocator works
  - confirmed with some examples

## Internal changes
- added `page_offset`, `page_to_physmap` to `class ArchOps`
- implemented `pwndbg.dbg_mod.__len__(self)`
- improved kernel address handling when KASLR is enabled
- abstracted `class IndentContextManager` to a more generic file
- added a test

## Demo
image
image
image


tests/qemu-tests/tests/system/test_commands_kernel.py
```diff
@@ -74,12 +74,12 @@ def test_command_slab_contains():
     assert f"{addr} @ {slab_cache}" in res
 
 
+@pytest.mark.skipif(
+    pwndbg.aglib.arch.name not in ["x86", "x86-64"],
+    reason="function page_offset is only implemented for x86",
+)
```
disconnect3d, line 80: For future this should probably be done in a separate PR, but its fine here

disconnect3d:
I am going to merge this since this makes sense to me.

I haven't really double checked if it works, but I believe/hope it does especially that there is a test (even though it doesn't really check for the exact values etc).

One note though, when we truncate the output like here:
image

Shall this maybe show how one can not truncate it?

jxuanli:
> I am going to merge this since this makes sense to me.
> 
> I haven't really double checked if it works, but I believe/hope it does especially that there is a test (even though it doesn't really check for the exact values etc).
> 
> One note though, when we truncate the output like here: image
> 
> Shall this maybe show how one can not truncate it?

my main concern was that, some of the lists occasionally contain more than 0x700 elements, and if we need to find an address that got truncated due to the length of the list, we can always do `buddydump -f <address>`, this will find it even if the address got truncated

PR #2981, opened 2025-05-16
Debugger agnostic documentation generation

### LINKED ISSUE #2955: Debugger agnostic documentation generation
Currently, the docs are generated by starting up gdb+pwndbg, then inspecting the python objects defined in pwndbg and extracting the necessary information to generate the documentation markdown files. Currently if a command is registered to only be for some debugger, its CommandObj won't be instantiated and it won't be added to the pwndbg.commands.commands list. In fact the corresponding file for that command won't even be loaded, this is controlled here:
https://github.com/pwndbg/pwndbg/blob/001926ebc0fc6b9c6a87833ec45346dd496d91f3/pwndbg/commands/__init__.py#L782-L876

 Thus, if an lldb-only command is added, its documentation won't be generated.

There are two ways to fix this that I can think of:
1. Make all command( files)s loadable by all debuggers; just not callable by all debuggers (or at least gdb). Thus we could generate the docs by only invoking gdb.
2. Run doc generation once for each debugger and create all the docs each debugger sees.

Regarding option one:
It's possible but a pain. Doing something like this in command files:
```python
    import gdb
    from pwndbg.gdblib.functions import GdbFunction
else:
    class GdbFunction:
        def __init__(self):
            pass
        def __call__(self, fn):
            return fn
```
Would allow lldb to load gdb-only commands (we would prevent them from ever being invoked). The same could be done for lldb-only commands being seen in gdb. That being said this is a bit of a pain for the developer. Going with this option would allow `load_commands()` to be replaced with code that does "import all `.py` files in `pwndbg/commands/` except `__init__.py`" which would be neat I guess since it would prevent people from forgetting to add their command to `load_commands()`. 

Regarding option two:
Maybe this is the better option but there are some kinks to figure out. Like, currently, if you rename a command, the doc gen script sees an old/extraneous markdown file which it can safely delete (`docs/commands/old_name.md`), and generate a new file (`docs/commands/new_name.md`) because it knows/assumes that it can see all commands that exist. But if debuggers only see a subset of the commands, they cannot confidently delete some file they don't understand because it might belong to another debugger. Also if there is a discrepency between the docs of a command in two different debuggers (maybe the command is defined differently/to do different things? (I would hope this wouldn't happen)), this could also be troublesome.

k4lizen:
Closes https://github.com/pwndbg/pwndbg/issues/2955 . I went with option two from that issue.

I renamed the doc scripts to use `-` instead of `_` to be consistent with the other scripts. All the documentation should be the exact same except for:
+ the aforementioned naming switch
+ the commands index gets a bug fix (telescope was duplicated)
+ the functions are sorted alphabetically (like everything else)

See the `./scripts/_docs/README.md` file for an overview of how it works.

I'll handle the matter of mismatching debugger information in another PR.


k4lizen:
hm need to figure out a way to get lldb running in CI, ~~iirc the version on ubuntu is too old~~

k4lizen:
image
awa

disconnect3d (approved):

disconnect3d:
Pls fix conflicts ;/

k4lizen:
btw I guess we effectively renamed the pcplist command to buddydump. it might be prudent to provide an alias with a deprecation notice for the time being

disconnect3d:
> btw I guess we effectively renamed the pcplist command to buddydump. it might be prudent to provide an alias with a deprecation notice for the time being

Hmm... lets discuss this on Discord.

Issue #2985, opened 2025-05-18
Install uv as an external tool


k4lizen:
Currently in our setup.sh we setup the virtual environment, then install uv *inside it*. When we want to use uv in scripts we activate the venv.

This workflow is kind of backwards. uv is a tool for managing python packages and virtual environments. It should be the one to setup the venv. We don't need to `source ./.venv/bin/activate` ever since `uv run ..` handles all that. This has personally caused me lots of confusion since general advice and our documentation mentions activating the venv and things not working without doing that.

I think we should install uv system-wide in the `./setup.sh` script if the user doesn't already have it installed - this would also mean people wouldn't have two uv installations because of us (one that they might have regardless of pwndbg, and one in the pwndbg venv). This is analogous to us already installing python/pip system-wide in the script. The uv we invoke would then be updated when the system package manager updates it (I think currently it is never updated) if the user already had it installed.

Then we just use `uv run ...` in the scripts (like tests.sh, generate-docs.sh etc.) so the end user doesn't have to worry about any of that, and recommend the use of `uv run ...` for stuff like running lldb (though tbh that should also probably be pulled out into a script).

If we do end up doing this, I'm not sure how it interacts with nix. I'm guessing the portable builds would be unaffected by this change, maybe they would be smaller?

disconnect3d:
Just for clarity, we have been discussing this here and there (on Discord, on some GH issues) and I am still wondering what is the best course of action here.

I think we will end up keeping uv installed in venv to not trash/corrupt/whateva the user environment, as the user shouldn't care about uv at all IMHO.

disconnect3d:
Ok. Let's keep UV in venv and not pollute users environment.

k4lizen:
just to have it written down: this would only affect people who want to install pwndbg from source, rather than through our packages or their package manager - i.e. mostly developers

PR #2986, opened 2025-05-18
More robust uv usage

k4lizen:
This PR fixes up some uv invocations which were previously borked. For instance we didn't specify `--group dev` in `lint.sh` even though the dependancy is very much needed (for gdb types). Running the common scripts should now be more straightforward for people.

All relevant python/command invocations go through the venv / uv now.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -99,14 +99,13 @@ jobs:
         echo 'GDB py:'
         gdb --batch --quiet --nx --nh --ex 'py import sys; print(sys.version)'
         echo 'Installed py:'
-        ./.venv/bin/python -V
+        uvx python -V
         echo 'Installed packages:'
-        ./.venv/bin/python -m pip freeze
+        uv pip freeze
 
     # We set `kernel.yama.ptrace_scope=0` for `attachp` command tests
     - name: Run tests
       run: |
-        source .venv/bin/activate
         mkdir .cov
         sudo sysctl -w kernel.yama.ptrace_scope=0
         ./tests.sh --cov
@@ -115,8 +114,8 @@ jobs:
     - name: Process coverage data
       if: matrix.os == 'ubuntu-22.04'
       run: |
-        ./.venv/bin/coverage combine
-        ./.venv/bin/coverage xml
+        uvx coverage combine
+        uvx coverage xml
 
     - name: "Upload coverage to Codecov"
       if: matrix.os == 'ubuntu-22.04'
@@ -146,8 +145,8 @@ jobs:
 
     - name: Process coverage data
       run: |
-        ./.venv/bin/coverage combine
-        ./.venv/bin/coverage xml
+        uvx coverage combine
+        uvx coverage xml
 
     - name: Upload coverage to Codecov
       uses: codecov/codecov-action@v3
@@ -190,8 +189,8 @@ jobs:
 
     - name: Process coverage data
       run: |
-        ./.venv/bin/coverage combine
-        ./.venv/bin/coverage xml
+        uvx coverage combine
+        uvx coverage xml
 
     - name: Upload coverage to Codecov
       uses: codecov/codecov-action@v3
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,4 +1,4 @@
-name: Unit tests
+name: Tests
 on:
   push:
     branches:
@@ -25,6 +25,7 @@ jobs:
     timeout-minutes: 40
     env:
       TMPDIR: /tmp
+      PWNDBG_VENV_PATH: PWNDBG_PLEASE_SKIP_VENV
     steps:
     - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9  # @v3
     - uses: cachix/install-nix-action@08dcb3a5e62fa31e2da3d490afc4176ef55ecd72  # @v30
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -89,7 +89,7 @@ jobs:
         path: ~/.cache/pip
         key: ${{ matrix.os }}-cache-pip
 
-    - name: Install dependencies
+    - name: Setup pwndbg
       run: |
         ./setup.sh
         ./setup-dev.sh
@@ -99,9 +99,9 @@ jobs:
         echo 'GDB py:'
         gdb --batch --quiet --nx --nh --ex 'py import sys; print(sys.version)'
         echo 'Installed py:'
-        uvx python -V
+        ./.venv/bin/python -V
         echo 'Installed packages:'
-        uv pip freeze
+        ./.venv/bin/pip freeze
 
     # We set `kernel.yama.ptrace_scope=0` for `attachp` command tests
     - name: Run tests
@@ -114,8 +114,8 @@ jobs:
     - name: Process coverage data
       if: matrix.os == 'ubuntu-22.04'
       run: |
-        uv run --group tests coverage combine
-        uv run --group tests coverage xml
+        ./.venv/bin/uv run --group tests coverage combine
+        ./.venv/bin/uv run --group tests coverage xml
 
     - name: "Upload coverage to Codecov"
       if: matrix.os == 'ubuntu-22.04'
@@ -133,7 +133,7 @@ jobs:
         path: ~/.cache/pip
         key: ${{ matrix.os }}-cache-pip
 
-    - name: Install dependencies
+    - name: Setup pwndbg
       run: |
         ./setup.sh
         ./setup-dev.sh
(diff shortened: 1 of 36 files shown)


k4lizen:
@patryk4815 if https://github.com/pwndbg/pwndbg/issues/2985 sounds sensible to you, could you take a look at the nix stuff?

.github/workflows/tests.yml
```diff
@@ -99,14 +99,13 @@ jobs:
         echo 'GDB py:'
         gdb --batch --quiet --nx --nh --ex 'py import sys; print(sys.version)'
         echo 'Installed py:'
-        ./.venv/bin/python -V
+        uvx python -V
```
disconnect3d: is this typo?

disconnect3d (approved):

setup.sh
```diff
@@ -206,20 +206,36 @@ if [[ -z "$is_supported" ]]; then
...
 if [[ -z "${PWNDBG_VENV_PATH}" ]]; then
     PWNDBG_VENV_PATH="./.venv"
 fi
-echo "Creating virtualenv in path: ${PWNDBG_VENV_PATH}"
-
-${PYTHON} -m venv -- ${PWNDBG_VENV_PATH}
-source ${PWNDBG_VENV_PATH}/bin/activate
 
-# Install uv
-pip install uv
```
disconnect3d, line 219: Why this change? Isn't this best to keep uv in the venv to not clash with any system/user-wide things?

k4lizen: I opened an issue to discuss it here: https://github.com/pwndbg/pwndbg/issues/2985 , but after thinking about it some more I am relatively fine with keeping uv in the venv as long as we make our scripts robust so they never have to be invoked *with* uv. But if we expect users to actively invoke uv while contributing I think it is better to install it system-wide in setup.sh because it is kind of backwards to have to activate the venv which uv manages to use uv.

k4lizen: ~~i'll try to refactor to keep uv in the venv~~

Since this is possible (https://discord.com/channels/843809097920413717/871339601149706270/1374406587564949514):
image

I think that might be the best strategy. lmk what you think

disconnect3d: > as long as we make our scripts robust so they never have to be invoked with uv

I mean, the scripts can either activate venv first or execute the uv path from venv etc.?

k4lizen: yeah yeah. it's not a problem to implement that. the most egregious counter-example now is that there isn't a script for starting lldb so it requires invoking uv.

k4lizen: i'll fix the PR to install uv in venv

k4lizen:
@patryk4815 could you help me out with the nix stuff? The build failures are probably because I added uv to our deps (which makes sense since it is, and I need to do that for some workflows to work).

k4lizen:
Should be good now. Extracted some common functionality to `./scripts/common.sh`.

A thing that needed to be handled is that the venv is not always in the same place (we use `/venv` in docker). Also sometimes the venv may not even exist (if it is set to PWNDBG_PLEASE_SKIP_VENV).

nix users may need to manually set `PWNDBG_VENV_PATH=PWNDBG_PLEASE_SKIP_VENV` if they want to run scripts

k4lizen:
A funny thing to mention is that in the lint workflow, we don't need to do the whole setup.sh so we don't do it to save time. However, we do need to setup the venv for the lint dependancies.

So now in the lint workflow we
1. install uv
2. using that uv we create the venv and install all python deps
2.1 that installs uv inside the venv as well, since we run all scripts based on the uv inside the venv rather than the system installed uv
3. the ./lint.sh script performs the lint by using the uv installed inside the venv

kinda wacky but it makes sense I suppose

gdbinit.py
```diff
@@ -44,7 +44,9 @@ def hash_file(file_path: str | Path) -> str:
 def run_uv_install(
     binary_path: os.PathLike[str], src_root: Path, dev: bool = False
 ) -> Tuple[str, str, int]:
-    command: List[str] = [str(binary_path), "sync", "--extra", "gdb"]
+    # We don't want to quietly uninstall dependancies by just specifying
+    # `--extra gdb` so we will be conservative and pull all extras in.
+    command: List[str] = [str(binary_path), "sync", "--all-extras"]
```
disconnect3d, line 49: Does this mean we install also dev or docs deps just because we do an update?

gdbinit.py
```diff
@@ -83,7 +85,7 @@ def update_deps(src_root: Path, venv_path: Path) -> None:
     if current_hash == stored_hash:
         return
 
-    print("Detected outdated Pwndbg dependencies (uv.lock). Updating.")
+    print("Detected outdated pwndbg dependencies (uv.lock). Updating.")
```
disconnect3d:
```suggestion
    print("Detected outdated Pwndbg dependencies (uv.lock). Updating.")
```

gdbinit.py
k4lizen: nope, only these:
```
[project.optional-dependencies]
lldb = [
    # The LLDB REPL requires readline.
    'gnureadline>=8.2.10,<9; sys_platform != "win32"',
    'pyreadline3>=3.5.4,<4; sys_platform == "win32"'
]
gdb = []
```

k4lizen: mind the
```python
    if dev:
        command.extend(("--all-groups",))
```
if you don't have the dev marker you won't get dev stuff

disconnect3d:
This needs conflicts resolution @k4lizen

pyproject.toml
```diff
@@ -16,6 +16,10 @@ dependencies = [
     "pyelftools>=0.29,<0.30",
     "pygments>=2.18.0,<3",
     "psutil>=6.1.1,<7",
+    # for the (unused) inform_unmet_dependencies function
+    "setuptools>=80.8.0",
```
patryk4815: setuptools was removed https://github.com/pwndbg/pwndbg/pull/3017

patryk4815:
This fix is required to get `nix develop` work as before:

```
diff --git a/nix/devshell.nix b/nix/devshell.nix
index e79421bc..674d31fa 100644
--- a/nix/devshell.nix
+++ b/nix/devshell.nix
@@ -94,6 +94,7 @@ in
       ];
     shellHook = ''
       export PWNDBG_NO_AUTOUPDATE=1
+      export PWNDBG_NO_UV=1
       export PWNDBG_VENV_PATH="${pyEnv}"
       export ZIGPATH="${pkgs.lib.getBin pkgs.zig_0_13}/bin/"
       export REPO_ROOT=$(git rev-parse --show-toplevel)
diff --git a/scripts/common.sh b/scripts/common.sh
index e02cb1d3..bc14e351 100644
--- a/scripts/common.sh
+++ b/scripts/common.sh
@@ -8,7 +8,7 @@ if [[ -z "${PWNDBG_VENV_PATH}" ]]; then
     PWNDBG_VENV_PATH="${PWNDBG_ABS_PATH}/.venv"
 fi
 
-if [[ "$PWNDBG_VENV_PATH" == "PWNDBG_PLEASE_SKIP_VENV" ]]; then
+if [[ "$PWNDBG_VENV_PATH" == "PWNDBG_PLEASE_SKIP_VENV" || "$PWNDBG_NO_UV" == "1" ]]; then
     # We are using the dependencies as installed on the system
     # so we shouldn't use uv (and can't, since it's not installed).
     UV=""
```

disconnect3d:
@k4lizen let us know when u apply the nix fix + remove setuptools dep

lldbinit.py
```diff
@@ -30,7 +30,9 @@ def hash_file(file_path: str | Path) -> str:
 def run_uv_install(
     binary_path: os.PathLike[str], src_root: Path, dev: bool = False
 ) -> Tuple[str, str, int]:
-    command: List[str] = [str(binary_path), "sync", "--extra", "lldb"]
+    # We don't want to quietly uninstall dependencies by just specifying
+    # `--extra lldb` so we will be conservative and pull all extras in.
+    command: List[str] = [str(binary_path), "sync", "--all-extras"]
     if dev:
         command.extend(("--all-groups",))
```
disconnect3d: Fwiw this should be just `command.append("--all-groups")` huh?

lldbinit.py
```diff
@@ -64,7 +66,7 @@ def update_deps(src_root: Path, venv_path: Path) -> None:
     if current_hash == stored_hash:
         return
 
-    print("Detected outdated Pwndbg dependencies (uv.lock). Updating.")
+    print("Detected outdated pwndbg dependencies (uv.lock). Updating.")
```
disconnect3d:
```suggestion
    print("Detected outdated Pwndbg dependencies (uv.lock). Updating.")
```

k4lizen:
@disconnect3d ^^^ my commit disabled auto-merge :(

PR #2987, opened 2025-05-19
TUI: fix gdb version crash

patryk4815:
image

```
Fish — 18:42
Has anyone else had issues with "-1" appended on RPM version names? I worked around it but unless I did something strange I imagine others have this problem.

File "/home//Git/pwndbg/pwndbg/gdblib/tui/context.py", line 203, in ContextTUIWindow
    if tuple(map(int, gdb.VERSION.split(".")[:2])) < (16, 3):
       ~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: invalid literal for int() with base 10: '3-1'
```

### DIFF
--- a/pwndbg/gdblib/tui/context.py
+++ b/pwndbg/gdblib/tui/context.py
@@ -13,6 +13,7 @@ from pwndbg.commands.context import context
 from pwndbg.commands.context import context_sections
 from pwndbg.commands.context import contextoutput
 from pwndbg.commands.context import resetcontextoutput
+from pwndbg.gdblib import gdb_version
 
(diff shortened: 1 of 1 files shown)


k4lizen:
I guess this isn't an issue here
```
def rewire_exit():
    major_ver = int(gdb.VERSION.split(".")[0])
    if major_ver <= 15:
        # On certain verions of gdb (used on ubuntu 24.04) using sys.exit() can cause
```
since distros will probably only append dashes after the minor ver number?

Issue #2669, opened 2025-01-13
`slab list` seems to be incorrect


tesuji:
### Description

Debugging kernel version 6.7.9, pwndbg `slab list`  and libslub `sblist` disagrees with each other.
But manually parsing `/proc/slabinfo` of the vm confirms that libslub is more correct.
```gdb
pwndbg> pipe slab list | head
Name                       # Objects    Size    Obj Size    # inuse    order
-----------------------  -----------  ------  ----------  ---------  -------
kheap_obj                         16     512         472        472        1
9p-fcall-cache                     1  131096      131096     131096        6
pwndbg> sblist
name                    objs inuse slabs size obj_size objs_per_slab pages_per_slab
kheap_obj                  0     0     0  512      472            16              2
9p-fcall-cache             0     0     0 131096   131096             1             64
```

```console
> sudo cat /proc/slabinfo | head
slabinfo - version: 2.1
# name            <active_objs> <num_objs> <objsize> <objperslab> <pagesperslab> : tunables <limit> <batchcount> <sharedfactor> : slabdata <active_slabs> <num_slabs> <sharedavail>
kheap_obj              0      0    512   16    2 : tunables    0    0    0 : slabdata      0      0      0
```

### Steps to reproduce

Quite cumbersome to setup. Might need QEMU to load the kernel
and use gdb 'target remote 0:1234' to remotely debug the kernel.

### My setup

```
pwndbg> version
Gdb:      15.1
Python:   3.12.4 (main, Jun  6 2024, 18:26:44) [GCC 13.3.0]
Pwndbg:   2024.08.29
Capstone: 5.0.1280
Unicorn:  2.0.1
```

cc https://github.com/pwndbg/pwndbg/issues/1831

disconnect3d:
cc: @fidgetingbits any chance you can check this out?

jxuanli:
this and #1831 can be closed as well

PR #2988, opened 2025-05-20
Improving `slab` commands

jxuanli:
This PR depends on / is developed on top of #2980 (this PR uses a class modified by #2980), so I will open this for review after it is merged.

## Fixed issues such as
- `slab info -v` results in an infinite loop when the linked list is cyclic
- `slab list` now print less inaccurate `inuse`
  - the kernel does not track full slabs so the value does not include those (unless `CONFIG_SLUB_DEBUG` is enabled, which is not for most kernel builds)
  - but it is consistent with `libslub` now
  - addressing #2669 
- `slab contains $rax` crashes

## Added features
- additional metadata are displayed
  - for new kernel versions as well
- supporting filter by `--cpu`, `--node`, `--active`, `--partial`, etc
- added index display to indicate the location of a freed object in the linked list
- added cyclic linked list detection (happens during exploit dev)

## Demo
image
image
as shown below, `[0x00]->[0x01]->[0x02]->[0x03]->[0x00]->...` forms a cyclic linked list. With this PR, this is correctly detected and handled gracefully.
image


k4lizen:
fyi I think you'll want to `git rebase -i dev` and drop the commits from the buddydump PR

jxuanli:
> fyi I think you'll want to `git rebase -i dev` and drop the commits from the buddydump PR

thanks! finally figured out how to do this properly after experimenting for a few hours

disconnect3d (approved):

Issue #2958, opened 2025-05-07
Rebase addresses in elfsections command


OBarronCS:
The `elfsections` command prints out sections of the binary that are loaded into memory.

![Image](https://github.com/user-attachments/assets/4b16ae15-b313-4cc2-993d-3cbe6dd2c4ff)

It would be great to add a commandline flag to optionally rebase these addresses so that it displays the actual memory locations that they are placed in the address space due to PIE/ASLR.

malik-aditya-r:
Hi, If it is still available I would like to work on it.

disconnect3d:
Hi @malik-aditya-r! Then just work on it and send a pull request when you implement it (and ideally tests for it) and test it locally.

PR #2990, opened 2025-05-21
FEAT: Rebase addresses in elfsections command #2958

Emanuele19:
This PR adds support for displaying relocated section addresses in the elf command, as requested in issue [#2958](https://github.com/pwndbg/pwndbg/issues/2958).

This is actually the first time I contribute to an open-source project, I hope I correctly followed the guidelines. Here is what I've done:
Changes:
1. Show relocated address ranges when the binary is PIE and currently running
2. Relocation base address is extracted via `pwndbg.gdblib.vmmap`
3. Output format updated with an additional RELOCATED column
4. Includes tests to validate against `info files` section addresses
Testing:
1. Verified manually with PIE and non-PIE binaries
2. Added automated test using pytest, passes via ./tests.sh

### DIFF
--- a/pwndbg/commands/elf.py
+++ b/pwndbg/commands/elf.py
@@ -9,13 +9,24 @@ import pwndbg.commands
 from pwndbg.color import message
 from pwndbg.commands import CommandCategory
 
+from pwnlib.elf import ELF # for PIE check
+import pwndbg.aglib.vmmap # for address mapping
 
 @pwndbg.commands.Command(
     "Prints the section mappings contained in the ELF header.", category=CommandCategory.LINUX
 )
 @pwndbg.commands.OnlyWithFile
-def elfsections() -> None:
+def elfsections() -> None:    
     local_path = pwndbg.aglib.file.get_proc_exe_file()
+
+    # IF PIE is set and binary is in execution, save relocation base address.
+    pie_relocation = False
+    if pwndbg.aglib.proc.alive and ELF(local_path).pie:
+        for mapping in pwndbg.aglib.vmmap.get():
+            if mapping.objfile == local_path:
+                pie_relocation = True
+                base_addr = mapping.start
+                break
 
     with open(local_path, "rb") as f:
         elffile = ELFFile(f)
@@ -32,8 +43,17 @@ def elfsections() -> None:
 
         sections.sort()
 
+        # Header
+        print('ADDRESS', end='\t\t')
+        if pie_relocation:
+            print('RELOCATED', end='\t\t\t')
+        print(' SECTION')
+
         for start, end, name in sections:
-            print(f"{start:#x} - {end:#x} ", name)
+            print(f"{start:#x} - {end:#x}", end='\t')
+            if pie_relocation:
+                print(f"{start + base_addr:#x} - {end + base_addr:#x}", end='\t')
+            print(name)
 
 
 @pwndbg.commands.Command(
--- a/pwndbg/commands/elf.py
+++ b/pwndbg/commands/elf.py
@@ -4,19 +4,19 @@ from typing import List
 from typing import Tuple
 
 from elftools.elf.elffile import ELFFile
+from pwnlib.elf import ELF  # for PIE check
 
+import pwndbg.aglib.vmmap  # for address mapping
 import pwndbg.commands
 from pwndbg.color import message
 from pwndbg.commands import CommandCategory
 
-from pwnlib.elf import ELF # for PIE check
-import pwndbg.aglib.vmmap # for address mapping
 
 @pwndbg.commands.Command(
     "Prints the section mappings contained in the ELF header.", category=CommandCategory.LINUX
 )
 @pwndbg.commands.OnlyWithFile
-def elfsections() -> None:    
+def elfsections() -> None:
     local_path = pwndbg.aglib.file.get_proc_exe_file()
 
     # IF PIE is set and binary is in execution, save relocation base address.
@@ -44,15 +44,15 @@ def elfsections() -> None:
         sections.sort()
 
         # Header
-        print('ADDRESS', end='\t\t')
+        print("ADDRESS", end="\t\t")
         if pie_relocation:
-            print('RELOCATED', end='\t\t\t')
-        print(' SECTION')
+            print("RELOCATED", end="\t\t\t")
+        print(" SECTION")
 
         for start, end, name in sections:
-            print(f"{start:#x} - {end:#x}", end='\t')
+            print(f"{start:#x} - {end:#x}", end="\t")
             if pie_relocation:
-                print(f"{start + base_addr:#x} - {end + base_addr:#x}", end='\t')
+                print(f"{start + base_addr:#x} - {end + base_addr:#x}", end="\t")
             print(name)
 
 
--- a/pwndbg/commands/elf.py
+++ b/pwndbg/commands/elf.py
@@ -5,6 +5,7 @@ from typing import Tuple
 
 from elftools.elf.elffile import ELFFile
 from pwnlib.elf import ELF  # for PIE check
+from tabulate import tabulate
 
 import pwndbg.aglib.vmmap  # for address mapping
 import pwndbg.commands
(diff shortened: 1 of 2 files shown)


pwndbg/commands/elf.py
```diff
@@ -17,6 +19,15 @@
...
     local_path = pwndbg.aglib.file.get_proc_exe_file()
 
+    # IF PIE is set and binary is in execution, save relocation base address.
+    pie_relocation = False
+    if pwndbg.aglib.proc.alive and ELF(local_path).pie:
+        for mapping in pwndbg.aglib.vmmap.get():
+            if mapping.objfile == local_path:
+                pie_relocation = True
+                base_addr = mapping.start
+                break
```
disconnect3d: Please change this to:
```py
base_address = next((p.start for p in pwndbg.aglib.vmmap.get() if p.objfile == pwndbg.aglib.proc.exe), 0)
```

And then do: `if base_address != 0:` - show relocated.

pwndbg/commands/elf.py
```diff
@@ -32,8 +43,17 @@ def elfsections() -> None:
 
         sections.sort()
 
+        # Header
+        print("ADDRESS", end="\t\t")
+        if pie_relocation:
+            print("RELOCATED", end="\t\t\t")
```
disconnect3d: Can you show how this looks like?

Shall we maybe use the `tabulate` library we use in different areas of Pwndbg instead of tabulating things manually? Can you check if this would be better?

This would be something like:
```py
from tabulate import tabulate
print(tabulate(table, headers))
```

See how it is used in pwndbg/command/tls.py

tests/gdb-tests/tests/test_commands_elf.py
```diff
@@ -260,3 +260,62 @@ def test_command_got_for_target_binary_and_loaded_library(binary_name):
...
+    ),
+)
+def test_command_elf_with_relocation_address_if_pie(binary_name, is_pie):
+    binary = tests.binaries.get(binary_name)
+    gdb.execute(f"file {binary}")
+
+    try:
+        gdb.execute("starti")
+    except gdb.error:
+        pytest.skip("Test not supported on this platform.")
```
disconnect3d, line 280: Hmm... does anyone know why do we do this in the following way in different files like onegadget/heap_bins/tls? cc: @k4lizen @patryk4815 @CptGibbon

disconnect3d:
Hey, thanks for the PR! Please see comments :)

Emanuele19:
@disconnect3d thank you for the suggestions, I'm going to implement those and I'm going to investigate on the try-catch block, I thought it was standard procedure.

Edit: just fixed with your suggestions. Regarding the try-catch block I'm documenting [here](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Exception-Handling.html#Exception-Handling), but I can't see an obvious reason for the message `"Test not supported on this platform."`

tests/gdb-tests/tests/test_commands_elf.py
k4lizen: not sure

disconnect3d:
> @disconnect3d thank you for the suggestions, I'm going to implement those and I'm going to investigate on the try-catch block, I thought it was standard procedure.
> 
> Edit: just fixed with your suggestions. Regarding the try-catch block I'm documenting [here](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Exception-Handling.html#Exception-Handling), but I can't see an obvious reason for the message `"Test not supported on this platform."`


Great! Please don't worry about the try-catch block. Let's leave it as it is, we can investigate this in the future.

tests/gdb-tests/tests/test_commands_elf.py
disconnect3d: Nvm lets not worry about it here.

tests/gdb-tests/tests/test_commands_elf.py
```diff
@@ -260,3 +260,61 @@ def test_command_got_for_target_binary_and_loaded_library(binary_name):
...
+
+    elf_output = gdb.execute("elf", to_string=True).splitlines()
+
+    assert ("RELOCATED" in elf_output[0]) == is_pie
+
+    # Check if every address returned is correct
+    if is_pie:
```
disconnect3d, line 287: Hmm, we don't do any testing when is_pie is False, is that ok?

Emanuele19: In theory, the code I wrote just adds logic for the case where pie is detected. The only thing that has been changed as a whole is the header that is now printed as table with the module you've suggested.
Maybe it is worth mensioning that I have used the testing binaries already found in the test directory but I have also locally tested on some binaries of mine.
Btw, the last binary of the test case should be compiled as PIE but when running checksec I find that it is, I haven't recompiled it because I could have broken othet tests, should I raise an issue for that?

disconnect3d: > Btw, the last binary of the test case should be compiled as PIE but when running checksec I find that it is, I haven't recompiled it because I could have broken othet tests, should I raise an issue for that?

Wdym?

Emanuele19: > last binary of the test case should be compiled as PIE

Typo: shouldn't

> > Btw, the last binary of the test case should be compiled as PIE but when running checksec I find that it is, I haven't recompiled it because I could have broken othet tests, should I raise an issue for that?
> 
> Wdym?

Basically, the test `/pwndbg/tests/gdb-tests/tests/test_commands_elf.py` declares `NOPIE_I386_BINARY_WITH_PLT = "reference_bin_nopie.i386.out"` that should be NOPIE as the name said. However checksec says that PIE is enabled. This binary is used at least in another test case.

pwndbg/commands/elf.py
```diff
@@ -17,23 +20,37 @@
 def elfsections() -> None:
     local_path = pwndbg.aglib.file.get_proc_exe_file()
 
+    # If PIE is set and binary is in execution, get the relocation base address
+    base_addr = 0
+    if pwndbg.aglib.proc.alive and ELF(local_path).pie:
+        base_addr = next(
+            (p.start for p in pwndbg.aglib.vmmap.get() if p.objfile == pwndbg.aglib.proc.exe),
```
disconnect3d, line 27: Btw it turns out there is even a simpler way to fetch base addr:
https://github.com/pwndbg/pwndbg/pull/2998/files#diff-ae21269239ab947b05978bb844ccf121eabe0d71e1e25a027d781b942ff83e1fR25

tests/gdb-tests/tests/test_commands_elf.py
disconnect3d:
```
dc@jctf:~/pwndbg$ rg reference_bin_nopie -C5

172:reference_bin_nopie.out: reference-binary.c
173:	@echo "[+] Building reference_bin_nopie.out"
174:	${ZIGCC} -fno-pie -o reference_bin_nopie.out reference-binary.c
175-
176:reference_bin_nopie.i386.out: reference-binary.c
177:	@echo "[+] Building reference_bin_nopie.i386.out"
178:	${ZIGCC} -fpie -target x86-linux-gnu -o reference_bin_nopie.i386.out reference-binary.c
```

There seems to be a bug here: `-fpie` is passed for `reference_bin_nopie.i386.out` instead of `-fno-pie`.

disconnect3d: But anyway, my point is that ideally the test should assert no-pie values too, at least to see if the output matches the format?

pwndbg/commands/elf.py
Emanuele19: Considering that implementation, should I close mine? I thought I were the only one working on it

disconnect3d: So did I. Honestly it is very rare that two people pick the same thing to implement/work on. I'm sorry that this happened here.

I feel like we can proceed with https://github.com/pwndbg/pwndbg/pull/2999 as it advanced further, so let's close this here. Feel free to help with code reviews there if you want! :)

Btw if u have time, any chance u can fix the PIE (`-fno-pie` flag) for the `reference_bin_nopie.i386.out` binary in a separate PR?

Emanuele19: > I'm sorry that this happened here.

Not a problem, just bad luck lol.

> Btw if u have time, any chance u can fix the PIE (-fno-pie flag) for the reference_bin_nopie.i386.out binary in a separate PR?

Sure, I'm going to do that as soon as I can.
Anyway, thank you for reviewing the PR. See you in the next one :) .

disconnect3d:
Closing in favor of https://github.com/pwndbg/pwndbg/pull/2999

PR #2993, opened 2025-05-22
Order the documentation navigation

k4lizen:
Before:
image
After:
image

### Side note
Adding the doc dep with `uv add --group docs mkdocs-awesome-nav` updated all the other stuff as well? So the uv.lock was updated.

And I'm guessing some mypy stuff got updated as well because the lint started complaining:
```
Uninstalled 1 package in 0.16ms                                                                                               
Installed 1 package in 0.32ms                                                                                                 
pwndbg/exception.py:51: error: Library stubs not installed for "pkg_resources"  [import-untyped]                              
        import pkg_resources                                                                                                  
    ^                                                                                                                         
pwndbg/exception.py:51: note: Hint: "python3 -m pip install types-setuptools"                                                 
pwndbg/exception.py:51: note: (or run "mypy --install-types" to install all missing stub packages)                            
pwndbg/exception.py:51: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports                     
Found 1 error in 1 file (checked 233 source files)                      
```
But actually unlike what the hint says pkg_resources is not available in types-setuptools (https://github.com/python/typeshed/issues/12398) because it became typed so the types are available in the same package that provides the pkg_resources import. However that package is setuptools, which we don't have imported at all. So probably, if the code ever reached that import line it would have crashed, but it never did because that function (inform_unmet_dependencies) is unused. So I added the setuptools package to our deps for now, but we may take it out in the future if we decide to remove that unused code.


pyproject.toml
```diff
@@ -16,6 +16,8 @@ dependencies = [
     "pyelftools>=0.29,<0.30",
     "pygments>=2.18.0,<3",
     "psutil>=6.1.1,<7",
+    # for the (unused) inform_unmet_dependencies function
+    "setuptools>=80.8.0",
```
disconnect3d, line 20: Can you elaborate more on that? Why is that needed here and not in dev or docs deps?

k4lizen: we could put it in the lint group and have no issues, however, in a literal sense it is a direct dependancy. the setuptools package provides the `import pkg_resources` which is used here:
https://github.com/pwndbg/pwndbg/blob/f49a43aaa2e0f2208ccdd0389f91ec2cca6e232d/pwndbg/exception.py#L47-L52

the only reason users aren't actively crashing because of this is because this function is dead code. If I put this import at the top of the file I get `ModuleNotFoundError: No module named 'pkg_resources'` the moment I open gdb (unless we put setuptools as a direct dependancy).

disconnect3d: Huh... maybe we should remove this function if its dead code? Can be done in separate PR too..

PR #2996, opened 2025-05-22
Show differences in debuggers in the documentation

k4lizen:
Debuggers can disagree on what some command / parameter / function looks like. If we do not count the differences of something existing in one debugger and not existing in the other, the only disagreement between LLDB and GDB currently is for the `resolve-heap-via-heuristic` parameter.

This PR handles this issue by providing a tab to the users that they can click to specify which debugger's version they want to see. This is only shown when there are differences between debuggers. This is what it looks like for the  `resolve-heap-via-heuristic` parameter.
image
image
(the screenshot is zoomed-out)

When the user selects a debugger, that selection is persisted for other places on the site as well (so if they're using LLDB they won't have to click the LLDB tab everywhere it's available).

This is what the tabs would look like for differing commands if we had any:
image

Furthermore, when not all debuggers support some command / parameter, now there is a note below the title which describes in which debuggers it is available. Example:
image

Functions won't show this since they are only available in gdb. All our parameters are available in both debuggers so there isn't an example like that, but if this is changed it will work for parameters also.


disconnect3d:
@k4lizen obviously i changed the 'gdb'->'GDB' in wrong place; pls fix :(

PR #3002, opened 2025-05-22
Show skip reason in skipped tests

disconnect3d:

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -177,7 +177,7 @@ class TestStats:
             # Serial mode does not capture stdout, so it's not possible to check the result
             return
 
-        test_status = "FAIL"
+        test_status, reason = "FAIL", ""
         if process.returncode == 0:
             result = re.search(
                 r"(\x1b\[3.m(PASSED|FAILED|SKIPPED|XPASS|XFAIL)\x1b\[0m)",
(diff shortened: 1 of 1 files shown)


disconnect3d:
Eh, this needs to be done only for skipped tests:
image

disconnect3d:
I don't quite understand what is going in here. Why does this show up like that:

> gdb-tests/tests/heap/test_try_free.py::test_try_free_corrupted_unsorted_chunks SKIPPED 1.79s ays. The heap_bug.c file has the function:corrupted_unsorted_chunks())

Locally I got it once, then I tried some debugging and now I am always getting the expected one:

> gdb-tests/tests/heap/test_try_free.py::test_try_free_corrupted_unsorted_chunks SKIPPED 1.04s  (Needs review. In the heap.py on the line 972 the condition is true always. The heap_bug.c file has thefunction: corrupted_unsorted_chunks())

disconnect3d:
image

Wth? This is really just pytest truncating the output? Oo

disconnect3d:
Considering the output, I feel this is good enough for now. It turns out the truncation comes from a bug in pytest: https://github.com/pytest-dev/pytest/issues/11146#issuecomment-2902277950

Although not nice, I am okay with this for now.

Issue #3003, opened 2025-05-22
[bug] [question] instruction enhancement caching weirdness


k4lizen:
@OBarronCS 

I'm confused about this:
https://github.com/pwndbg/pwndbg/blob/e83dee14c24f5f4a7bd66d2e2d214f04debb4be8/pwndbg/aglib/disasm/disassembly.py#L55-L57

Does it makes sense? What if someone is in a loop and puts a conditional breakpoint a few instructions forward to let the loop iterate a few times wouldn't the annotation be incorrect?

However, I tried to create a reproducer ~~and didn't succeed :(~~ See comment.

source:
```c
int main (){
  int awa = 0;
  awa += 20;
  awa += 20;
  awa += 20;
  awa += 20;
  awa += 20;
  awa += 20;
  awa += 20;
  awa += 20;
  awa += 20;
  awa += 20;

  for (int i = 0; i < 100; ++i) {
    int x = i;
    int y = x + 4;
    y += 0x44;
    x += 0x1337;
  }

  return 0;
}
```
Compiled with `gcc -g main.c -o main`

Expected:
```
pwndbg> start
pwndbg> b *0x0000555555555158 if i == 50
pwndbg> continue

 ► 0x555555555158 <main+63>    mov    dword ptr [rbp - 8], eax         [0x7fffffffe188] <= 0x32
   0x55555555515b <main+66>    mov    eax, dword ptr [rbp - 8]         EAX, [0x7fffffffe188] => 0x32
   0x55555555515e <main+69>    add    eax, 4                           EAX => 54 (0x32 + 0x4)
   0x555555555161 <main+72>    mov    dword ptr [rbp - 4], eax         [0x7fffffffe18c] <= 0x36
   0x555555555164 <main+75>    add    dword ptr [rbp - 4], 0x44        [0x7fffffffe18c] <= 122 (0x36 + 0x44)
   0x555555555168 <main+79>    add    dword ptr [rbp - 8], 0x1337      [0x7fffffffe188] <= 0x1369 (0x32 + 0x1337)
   0x55555555516f <main+86>    add    dword ptr [rbp - 0x10], 1        [0x7fffffffe180] <= 51 (0x32 + 0x1)
```
With cond bp:
```
pwndbg> start
pwndbg> b *0x0000555555555155
pwndbg> continue

 ► 0x555555555155 <main+60>    mov    eax, dword ptr [rbp - 0x10]      EAX, [0x7fffffffe180] => 0
[...]

PR #3004, opened 2025-05-22
Tidy up some development documentation

k4lizen:
`2-dev-notes.md` and `improving-annotations.md` are both pretty good, i'd leave them as is - these are just a couple of fixups on formatting and such.

There might be some changes needed to `improving-annotations.md` after https://github.com/pwndbg/pwndbg/pull/2963 is merged and https://github.com/pwndbg/pwndbg/issues/3003 is figured out, but we get there when we get there.

### DIFF
--- a/docs/contributing/improving-annotations.md
+++ b/docs/contributing/improving-annotations.md
@@ -158,7 +158,7 @@ Taken the raw instruction bytes and pass them to `cstool` to see the information
 cstool -d mips 0x0400000c
 ```
 
-The number of operands may not match the visual appearance. You might also check the instruction groups, and verify that an instruction that we might consider a `call` has the Capstone `call` group. Capstone is not 100% correct in every single case in all architectures, so it's good to verify. Report a bug to Capstone if there appears to be an error, and in the meanwhile we can create a fix in pwndbg to work around the current behavior.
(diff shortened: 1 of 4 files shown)


docs/contributing/improving-annotations.md
```diff
@@ -150,21 +150,22 @@ mov qword ptr [rsp], rsi at 0x55555555706c (size=4) (arch: x86)
...
+2\. Use the Capstone disassembler to verify the number of operands the instruction groups.
 
 Taken the raw instruction bytes and pass them to `cstool` to see the information that we are working with:
 
 ```sh
 cstool -d mips 0x0400000c
 ```
 
-The number of operands may not match the visual appearance. You might also check the instruction groups, and verify that an instruction that we might consider a `call` has the Capstone `call` group. Capstone is not 100% correct in every single case in all architectures, so it's good to verify. Report a bug to Capstone if there appears to be an error, and in the meanwhile we can create a fix in Pwndbg to work around the current behavior.
+The number of operands may not match the visual appearance. You might also check the instruction groups, and verify that an instruction that we might consider a `call` has the Capstone `call` group. Capstone is not 100% correct in every single case in all architectures, so it's good to verify. Report a bug to Capstone if there appears to be an error, and in the meanwhile we can create a fix in pwndbg to work around the current behavior.
```
disconnect3d:
```suggestion
The number of operands may not match the visual appearance. You might also check the instruction groups, and verify that an instruction that we might consider a `call` has the Capstone `call` group. Capstone is not 100% correct in every single case in all architectures, so it's good to verify. Report a bug to Capstone if there appears to be an error, and in the meanwhile we can create a fix in Pwndbg to work around the current behavior.
```

PR #3008, opened 2025-05-23
FIX: reference_bin_nopie.i386 compiled with pie instead of no-pie

Emanuele19:
The binary `reference_bin_nopie.i386` was mistakenly compiled with `-fpie` instead of `-fno-pie`.
The change didn't break any test.


disconnect3d:
Thanks!

PR #3018, opened 2025-05-26
Fix regression with `lldb` after changes in Value.__len__

patryk4815:
image


jxuanli:
would #2997 have resolved the issue?

patryk4815:
@jxuanli tested with latest `dev`, different crash occur.
We should not implement `__len__` in any case I think
image

PR #3019, opened 2025-05-26
Add ability to suspend all events via SUSPEND_ALL

patryk4815:

### DIFF
--- a/pwndbg/dbg/lldb/__init__.py
+++ b/pwndbg/dbg/lldb/__init__.py
@@ -1986,7 +1986,7 @@ class LLDB(pwndbg.dbg_mod.Debugger):
         if ty not in self.event_handlers:
             # No one cares about this event type.
             return
-        if self.suspended_events[ty]:
+        if self.suspended_events[ty] or self.suspended_events[pwndbg.dbg_mod.EventType.SUSPEND_ALL]:
             # This event has been suspended.
             return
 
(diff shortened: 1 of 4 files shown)


pwndbg/dbg/lldb/__init__.py
```diff
@@ -1986,17 +1986,17 @@ def _trigger_event(self, ty: pwndbg.dbg_mod.EventType) -> None:
...
 
         for handler in self.event_handlers[ty]:
             try:
                 handler()
             except Exception as e:
-                import pwndbg.exception
+                from pwndbg.exception import handle as pwndbg_exception
 
-                pwndbg.exception.handle()
+                pwndbg_exception()
```
disconnect3d, line 1999: Why this change?

disconnect3d, line 1999: Why this change?

disconnect3d (approved):
I have some mixed feelings here: we are adding one more lookup to each event call:
```py
if self.suspended_events[ty] or self.suspended_events[pwndbg.dbg_mod.EventType.SUSPEND_ALL]:
```

There are 2-3 other options we could do instead:
1. We could set a global var and check that var - no dict lookup then, but still a variable lookup
2. In LLDB we have `if ty not in self.event_handlers: return` case, which could be used here: the suspend all could swap `self.event_handlers` with an empty one until the events are unsuspended? But we do not have an equivalent in GDB. In GDB we would have to somehow delete events from some list or something I believe.


I am going to accept it for now, but we can move this to a global

pwndbg/dbg/lldb/__init__.py
disconnect3d: Without this change this happens: 
image

PR #3020, opened 2025-05-26
Improving `vmmap` output when debugging kernel

jxuanli:
## Improving `vmmap`

Demos are based off the challenge `The Jumps` from `NahamCon CTF 2025` (but with KASLR enabled for demo purposes.)

Currently, when running `qemu-system`, the `vmmap` command outputs:
image

The `File` column contains only entries with the format `[pt_xxxxx...]` that are not useful. This PR improves the `vmmap` output when debug symbols are present.

## Demo

image

image


patryk4815:
each page must have unique name, otherwise some commands may not work? 🤔 @disconnect3d do you remember what?

@jxuanli can you check with `set kernel-vmmap monitor`? I'm going to upstream some pt-dump changes into qemu - later pt-dump can be removed xD

jxuanli:
@patryk4815 its certainly trickier with the current `set kernel-vmmap monitor` (`monitor info mem`). i don't think it is useable yet without some very heavy coalescing on page gaps (or i guess we can omit printing out the `%esp stacks`, but that would not work if debug sysmbols are not present since I used kernel symbols to identify different regions). It also doesn't show if a page is executable -- one option is to find its PTE entry for every page but that's not very efficient and may not always work if debug symbols are not present. but I guess i should look into how `pt-dump` is implemented to perhaps get some ideas.

it might be more viable with your changes but we would then need to support older versions of qemu.

disconnect3d:
Wow, this PR is awesome!

> each page must have unique name, otherwise some commands may not work? 🤔 @disconnect3d do you remember what?

This is about `search` command and few others (I think `probeleak` or `leakfind`?) where u want to target a specific page or pages. That's why we did `[pt_<part of address>]` initially.

Shall we maybe name those `userland{0,1,2,...}` or `userland_400`, `userland_401` etc for this reason?

pwndbg/aglib/kernel/vmmap.py
```diff
@@ -20,11 +20,131 @@
...
+    KERNELLAND = "kernel [.text]"
+    KERNELRO = "kernel [.rodata]"
+    KERNELBSS = "kernel [.bss]"
+    KERNELDRIVER = "kernel [loadable driver]"
+
+    def __init__(self, pages: List[pwndbg.lib.memory.Page]):
+        self.pages = pages
+        if not pwndbg.aglib.kernel.has_debug_syms():
+            return
+        # https://www.kernel.org/doc/Documentation/x86/x86_64/mm.txt
```
disconnect3d: Please add a comment with kernel version (or versions)

pwndbg/aglib/vmmap.py
```diff
@@ -34,3 +34,40 @@ def find(address: int | pwndbg.dbg_mod.Value | None) -> pwndbg.lib.memory.Page |
...
+    if arch_name == "x86-64":
+        address = pwndbg.aglib.kernel.get_idt_entries()[0].offset
+    elif arch_name == "aarch64":
+        address = pwndbg.aglib.regs.vbar
+    else:
+        return None
+
+    mappings = pages
+    for mapping in mappings:
+        # TODO: Check alignment
```
disconnect3d: What about this?

tests/qemu-tests/tests/system/test_commands_kernel.py
```diff
@@ -128,6 +128,22 @@ def test_command_msr_write():
     gdb.execute(f"msr MSR_LSTAR -w {prev_msr_lstar}")
 
 
+@pytest.mark.skipif(not pwndbg.aglib.kernel.has_debug_syms(), reason="test requires debug symbols")
+def test_command_kernel_vmmap():
+    res = gdb.execute("vmmap", to_string=True)
+    assert any(
+        key in res
+        for key in [
```
disconnect3d: Use a tuple

pwndbg/color/memory.py
```diff
@@ -88,10 +88,12 @@ def get(
         color = normal
     elif "[stack" in page.objfile:
         color = c.stack
-    elif "[heap" in page.objfile:
-        color = c.heap
     elif page.execute:
         color = c.code
+    elif not page.write:
+        color = c.rodata
+    elif any(keyword in page.objfile for keyword in ["[heap", "physmap", "vmalloc"]):
```
disconnect3d:
```suggestion
    elif any(keyword in page.objfile for keyword in ("[heap", "physmap", "vmalloc")):
```

pwndbg/aglib/kernel/vmmap.py
```diff
@@ -20,11 +20,131 @@
...
+        physmap = pwndbg.aglib.symbol.lookup_symbol_value("page_offset_base")
+        vmalloc = pwndbg.aglib.symbol.lookup_symbol_value("vmalloc_base")
+        vmemmap = pwndbg.aglib.symbol.lookup_symbol_value("vmemmap_base")
+        self.kbase = kbase = pwndbg.aglib.vmmap.find_kbase(pages)
+        if kbase is None:
+            return
+        self.sections = [
```
disconnect3d: Can we use a tuple here?

patryk4815:
> It also doesn't show if a page is executable

@jxuanli executable for x86-64 is fixed on my forked qemu https://github.com/patryk4815/qemu/commit/0621a6b9130ca21e1a89bbe3a2ab65a5e92756b8
`monitor info mem` right now is only supported in qemu for x86, x86-64 and riscv64

About `x` flag on upstream qemu: 
- riscv64 has `x` flags implemented (but our pwndbg code just omit it..)
- x86-32 - cpu don't support NX flag at all
- x86-64 - `x` flag is not implemented, I need send my patch https://github.com/patryk4815/qemu/commit/0621a6b9130ca21e1a89bbe3a2ab65a5e92756b8

jxuanli:
> This is about search command and few others (I think probeleak or leakfind?) where u want to target a specific page or pages. That's why we did [pt_<part of address>] initially.

how is userland handling this? iirc some of the userand pages have the same names (libc for example)

jxuanli:
@patryk4815 if your patch is merged in main branch then that prob works yea but we also need to support older versions of qemu?

pwndbg/color/memory.py
jxuanli: done

pwndbg/aglib/kernel/vmmap.py
jxuanli: fixed

tests/qemu-tests/tests/system/test_commands_kernel.py
jxuanli: done

pwndbg/aglib/vmmap.py
jxuanli: i think the mappings which are pages are by definition aligned

patryk4815:
@jxuanli yes, in pwndbg we will support older qemu version, qemu version can be detected by using `monitor info version`
btw do you see other bugs in `monitor info mem` that need be fixed?

pwndbg/aglib/kernel/vmmap.py
jxuanli: i believe this mapping has been pretty much the same for most recent kernels (v5.x, v6.x at least), but there might be outliers for different (non-x64) kernel builds, i ll add a comment once i figured that out

jxuanli:
> btw do you see other bugs in monitor info mem that need be fixed?

~~no~~

jxuanli:
update: i have addressed some of the above comments. ~~In addition, I have decided to not include `%esp fixup stacks` in the outputs of `vmmap` because those are not useful in my experience  and will cluster up the display especially with `monitor info mem` (@patryk4815 can you try it, i don't think its necessarily a bug in qemu but rather `pt-dump`).~~ it turned out that i was wrong and it is used in some cases. i will have another PR soon on some more page helpers which will help the effort in supporting older versions of qemu when `kernel-vmmap` is set to `monitor`

jxuanli:
btw when debugging a userland program, `vmmap` prints sections with the same name.
image

jxuanli:
@patryk4815 per our prior discussion on the excessively many `%esp fixup stacks` being displayed by `mem info`, it appears that `pt-dump` was buggy. A pagewalk using the command from #3051 shows that those pages displayed from `mem info` are indeed present even though they are not shown by `pt-dump`.
image

jxuanli:
this branch is somewhat ready, though it depends on #3051.

it handles both modes of `kernel-vmmap` although we need to discuss more about how we should handle the excessive amount of entries printed by `info mem`. The handling of `info mem` could also benefit from #3065 if we decide to omit most of the `%esp fixup stacks`.

disconnect3d:
@jxuanli pls fix conflicts

jxuanli:
@disconnect3d fixed and also did some fixing so that the output for aarch64 is sane  (updated the test to assert some of those changes)

PR #3023, opened 2025-05-26
Optimize page lookups

OBarronCS:
Instead of looping through all pages to find the page that an address belongs to, this PR creates a lookup table in the MemoryMap object which is used to make the vmmap.find() function faster.


OBarronCS:
I changed the strategy to use a binary search - I realized that building the lookup table itself has a non-negligible performance impact. In cases where you have huge mappings, such as those added with vmmap-add, or when debugging linux kernel such as in #3020 where you have large mappings, building the lookup table takes a long time, and it significantly outweighs the performance gains during page lookup. Local benchmarks verified this.

A binary search provides similar speed gains.

pwndbg/aglib/vmmap.py
```diff
@@ -15,9 +16,14 @@
 )
 
 
+@pwndbg.lib.cache.cache_until("start", "stop")
+def get_memory_map() -> MemoryMap:
+    return pwndbg.dbg.selected_inferior().vmmap()
```
disconnect3d, line 21: Do we really need yet another function? Shouldn't we just cache `pwndbg.dbg.selected_inferior().vmmap` instead?

disconnect3d (approved):

PR #3029, opened 2025-05-26
Fix install script for busybox

patryk4815:
image

### DIFF
--- a/docs/install.sh
+++ b/docs/install.sh
@@ -33,7 +33,7 @@ for cmd in wget tar xz uname mktemp rm mkdir ln grep; do
 done
 
 if [ -n "$missing" ]; then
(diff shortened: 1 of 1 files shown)


docs/install.sh
```diff
@@ -194,7 +194,15 @@ URL="https://github.com/pwndbg/pwndbg/releases/download/${VERSION}/${FILE}"
 trap "rm -rf $TEMP_DIR" EXIT
 
 echoinfo "Downloading... ${URL}"
-wget -q --show-progress "$URL" -O "$TEMP_DIR/$FILE" || {
+
+# 'wget' on BusyBox don't support progress options
+if wget --help 2>&1 | grep -qi 'busybox'; then
```
disconnect3d, line 199: should probably check link instead, but w/e

PR #3038, opened 2025-05-28
Additional fix to instruction.condition

OBarronCS:
Additional fix to #3030 so x86 call instructions are handled correctly, not incorrectly labeled as conditional instructions.


pwndbg/aglib/disasm/arm.py
```diff
@@ -264,12 +264,11 @@ def _condition(self, instruction: PwndbgInstruction, emu: Emulator) -> Instructi
                 instruction.declare_conditional = False
             return InstructionCondition.UNDETERMINED
 
-        # We can't reason about anything except the current instruction
-        if instruction.address != pwndbg.aglib.regs.pc:
```
disconnect3d, line 268: So previously we only reasoned about current instruction?

Is this ok we removed this condition?

PR #3039, opened 2025-05-28
Add total vs cumulative time to tests + disable some tests on CI for speedup

disconnect3d:


disconnect3d:
Current output looks like this:
image

PR #3041, opened 2025-05-28
Cleanup testing scripts a bit

k4lizen:
I was very confused when I was first trying to understand the testing infra so I renamed some stuff to hopefully make it more clear to people looking at it for the first time. The most relevant rename is `./qemu-tests.sh` -> `./crossarch-tests.sh` (because the script, despite its name, doesn't run all qemu tests), others: `images -> kimages`, `tests.sh -> system-tests.sh` (because they only run qemu-system stuff). Also, added a script that runs the kernel tests.

We are missing the iproute2 dependancy for the ubuntu docker to be able to run the kernel tests, so I added it to `setup-dev.sh`. The other distributions are still missing packages, for instance pacman is missing the qemu stuff.

### DIFF
--- a/setup-dev.sh
+++ b/setup-dev.sh
@@ -144,7 +144,8 @@ install_apt() {
         parallel \
         qemu-system-x86 \
         qemu-system-arm \
-        qemu-user
+        qemu-user \
+        iproute2
 
     # Some tests require i386 libc/ld, eg: test_smallbins_sizes_32bit_big
     if uname -m | grep -q x86_64; then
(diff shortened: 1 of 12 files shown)


setup-dev.sh
```diff
@@ -198,6 +199,8 @@ EOF
         gdb \
         parallel
 
+    # FIXME: add the necessary deps for testing
```
disconnect3d, line 202: :eyes:

k4lizen: arch doesn't have all the deps necessary for testing, e.g. qemu stuff. felt a bit out of place for this PR though since tbh setup-dev.sh has lots of wacky behaviour that should be fixed

Issue #3044, opened 2025-05-29
extra new line in gdb tui legend


k4lizen:
### Description

![Image](https://github.com/user-attachments/assets/2b4d7a0b-a16c-4f81-a481-5e9636d37c99)

### Steps to reproduce

```
gdb /bin/sh
pwndbg> start
pwndbg> tui enable
pwndbg> layout pwndbg
pwndbg> ni
```

### My setup
Pwndbg:   2025.05.28 build: 2f0005ff
Python:   3.13.3
GDB:      16.3
Capstone: 6.0.0
Unicorn:  2.1.3
Pwnlib:   4.14.0

disconnect3d:
@peace-maker can you take a look at this?

peace-maker:
Do you use HiDPI or some other high resolution? It's about the TUI layout algorithm and the pwndbg layout definition 
![Image](https://github.com/user-attachments/assets/84a89af2-0804-4cf6-ae55-dc4cfe98ee0d)

Changing the weight of the legend row to "0" works for me when zooming out
![Image](https://github.com/user-attachments/assets/903ec311-9281-4444-a4d3-2ce24bb25dd9)

```
tui new-layout pwndbg {-horizontal  {   { -horizontal    { pwndbg_disasm 1 } 2    {      { -horizontal pwndbg_legend 8 pwndbg_control 2 } 0 pwndbg_regs 6 pwndbg_stack 6    } 3   } 7 cmd 3  } 3 { pwndbg_backtrace 2 pwndbg_threads 1 pwndbg_expressions 2 } 1 } 1 status 1
```

k4lizen:
> Do you use HiDPI or some other high resolution?

nope, resolution is 1920x1080

I might have unusually *high* font size though?

PR #3053, opened 2025-05-30
Various docs fixups round 2

k4lizen:
+ Added short intro to configuration page.
+ Moves notes and examples from various command descriptions to the `Notes:` and `Examples:` sections.
+ Update FEATURES page
   + updated leakfind images
   + add section for probeleak, track-heap, track-got, `search -4`, attachp
+ make track-heap show chunk sizes in hex
+ added comment to the `profile` command
+ added docstrings for each top-level python module

### side-notes
+ I didn't update the procinfo images because it nicely shows android debugging which is mentioned in the text right above it. I don't have the setup for that rn, I can update it to something else, though the android image is a nice example.
+ Didn't update the GDB TUI image because of https://github.com/pwndbg/pwndbg/issues/3044 . 
+ Didn't update the `Splitting / Layouting Context` image because I think it looks nice there, but a bit wacky for me locally. I can use kittydbg for the screenie though it seems `context-clear-screen on` doesn't work for some reason?

@disconnect3d I can update these^^^ if you think the consistency is important, lmk.

re:
> malloc-chunk: should probably say that it prints a glibc heap chunk, maybe link to some info about glibc heap internals? or heap attacks? not sure

Should we add that to other glibc heap commands as well? I mean the command is in the `GLibc ptmalloc2 Heap Commands` category, it should be relatively obvious that it's only for glibc. i've never used this command btw xd so if we want to put some heap exploitation resources somewhere maybe this is not the best place


disconnect3d:
<3




> ### side-notes

Yeah, lets update those.

> * I didn't update the procinfo images because it nicely shows android debugging which is mentioned in the text right above it. I don't have the setup for that rn, I can update it to something else, though the android image is a nice example.

Regarding procinfo, a good example is attaching (`sudo -E gdb` -> `attachp <procname>`) to some process like firefox or other that uses files, UDS or netlink etc.

> * Didn't update the GDB TUI image because of [extra new line in gdb tui legend #3044](https://github.com/pwndbg/pwndbg/issues/3044) .

Fixed in https://github.com/pwndbg/pwndbg/pull/3056

> * Didn't update the `Splitting / Layouting Context` image because I think it looks nice there, but a bit wacky for me locally. I can use kittydbg for the screenie though it seems `context-clear-screen on` doesn't work for some reason?

This one looks not that old, so its probably ok, heh.


> > malloc-chunk: should probably say that it prints a glibc heap chunk, maybe link to some info about glibc heap internals? or heap attacks? not sure
> 
> Should we add that to other glibc heap commands as well? I mean the command is in the `GLibc ptmalloc2 Heap Commands` category, it should be relatively obvious that it's only for glibc. i've never used this command btw xd so if we want to put some heap exploitation resources somewhere maybe this is not the best place

If we do it for that one, we should probably do for others too. I also haven't really used that command, maybe it is not that useful, I don't know :/.

pwndbg/__init__.py
```diff
@@ -14,4 +17,10 @@
 from pwndbg.dbg import dbg as dbg
 
 __version__ = pwndbg.lib.version.__version__
+"""
+Pwndbg version.
+"""
 version = __version__
+"""
+Pwndbg version.
+"""
```
disconnect3d, line 26: Heh, those comments kinda suck :<

docs/commands/linux_libc_elf/got.md
```diff
@@ -29,5 +23,13 @@ Examples:
 |-a|--all|Process all libs/obfjiles including the target executable.|
 |-r|--show-readonly|Also display read-only entries (which are filtered out by default).|
 
+### Examples
+```text
+got
+got puts
+got -p libc
+got -a
```
disconnect3d, line 31: Is this useful at all not showing output? Also not sure what -p -a means, i mean, i know its described above, heh...

disconnect3d (approved):

pwndbg/__init__.py
k4lizen: I don't disagree! But otherwise there is an empty space here :<
image

image

By the way do we actually need both variables? I don't see the reason by just grepping through the source? Maybe it is used by some python tooling we use?

Maybe at least formatting the comments like this makes them not so intrusive
image

PR #3055, opened 2025-05-31
Show demo gif on docs front page

k4lizen:
image

-----

image


k4lizen:
The cheatsheet is inaccurate since we changed `_` to `-` in commands btw.

PR #3057, opened 2025-05-31
Make it clear to contributors LLDB is needed for doc generation

k4lizen:
Since https://github.com/pwndbg/pwndbg/pull/2981 you need to have LLDB installed to do doc verification and generation since the debuggers may disagree on the commands they see.

We *could* make it so that the doc generation still runs but simply doesn't use LLDB (would require a slight refactor but nothing too major), and prints a warning about this. This will give correct output for now since LLDB doesn't see any more information than GDB does. But in the future contributors might fail the CI check because of this. The current solution is simple and I think it might be a good thing that will coax contributors to test their changes with LLDB as well.

### DIFF
--- a/docs/contributing/index.md
+++ b/docs/contributing/index.md
@@ -82,7 +82,7 @@ The `./docs/commands`, `./docs/functions`, and `./docs/configuration` folders ar
 ```{.bash .copy}
 ./scripts/generate-docs.sh
 ```
-to update the documentation. Commit these changes in a separate commit.
+to update the documentation. You need to have a supported version of GDB *and* [LLDB installed](setup-pwndbg-dev.md#running-with-lldb) for this to work. Commit these changes in a separate commit.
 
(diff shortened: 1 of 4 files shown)


disconnect3d:
I'm a bit afraid that the LLDB requirement may cause less contributions on the long term... but lets keep it for now

k4lizen:
> We could make it so that the doc generation still runs but simply doesn't use LLDB

I tried to implement this but idk what I was tripping, the script won't know to make this without LLDB installed
image

if it was only a matter of "this command is registered in this debugger" I would have the PR ready for that, but unfortunately, it is not :<

k4lizen:
I guess I could just do "if LLDB is not available, don't even try to generate config docs" but man does that feel flaky

CptGibbon:
Could document generation be done by the CI?

k4lizen:
@CptGibbon I think so yes: https://github.com/marketplace/actions/add-commit
And yeah I agree that might be the best way forward.

PR #3069, opened 2025-06-02
Run `run -s` automatically for `entry`

### LINKED ISSUE #3058: The `entry` command does not work in pwndbg-lldb
TL;DR: 
```
$ pwndbg-lldb /bin/ls
pwndbg: loaded 155 pwndbg commands commands.
------- tip of the day (some of these don't work in LLDB yet!) (disable with set show-tips off) -------
Use the telescope command to dereference a given address/pointer multiple times (if the dereferenced value is a valid ptr; see config telescope to configure its behavior)
pwndbg-lldb> target create '/bin/ls'
Current executable set to '/bin/ls' (x86_64)
pwndbg-lldb> entry
Pwndbg commands require a target binary to be selected
pwndbg-lldb> file /bin/ls
Current executable set to '/bin/ls' (x86_64).
pwndbg-lldb> entry
Pwndbg commands require a target binary to be selected
pwndbg-lldb>
```

mbrla0:
Fixes #3058

This PR makes it so that a process is automatically launched if needed when `entry` is called in LLDB, to better match the behavior of that command in GDB. This is achieved by using an override that launches the process.

I'm not particularly fond of using an override for this, personally. Ideally, we'd only ever use them for things LLDB absolutely wouldn't let us do otherwise, and I think the better solution for `entry` would be to launch using `ProcessController` from inside the command itself. Trouble with that is that it would require a bit of a rework of the ProcessDriver and of command handling under LLDB, which I feel is too much to support a single command, when achieving the same outcome with an override is so short and simple.

We should probably revisit this later, though, after we get rid of the whole "Pwndbg commands require an active inferior" deal.

### DIFF
--- a/pwndbg/dbg/lldb/repl/__init__.py
+++ b/pwndbg/dbg/lldb/repl/__init__.py
@@ -514,6 +514,26 @@ def exec_repl_command(
         run_ipython_shell()
         return True
 
+    if (
+        bits[0] == pwndbg.commands.start.entry.command_name
+        or bits[0] in pwndbg.commands.start.entry.aliases
+    ):
+        # 'entry' is actually a Pwndbg command. For convenience, we launch the
+        # process on its behalf, before letting it run.
+        #
+        # In the LLDB back-end, there is no proper mechanism to make a process
+        # start from inside of a command, as there is in GDB. Ideally, we'd
+        # rework `ProcessDriver` so that it lets us do that with an execution
+        # controller, but that is quite a bit of work to fix a single command,
+        # when using an override is enough to achieve the same goal.
+        #
+        # In any case, we should consider doing it if this proves to be too janky.
+        if not driver.has_process():
+            process_launch(driver, relay, ["-s"], dbg)
+
+        # This intentionally falls through. We want LLDB to do the rest of the
+        # work of processing 'entry'.
+
     # The command hasn't matched any of our filtered commands, just let LLDB
     # handle it normally. Either in the context of the process, if we have
     # one, or just in a general context.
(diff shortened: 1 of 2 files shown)


pwndbg/dbg/lldb/repl/__init__.py
```diff
@@ -514,6 +514,26 @@ def exec_repl_command(
...
+        #
+        # In the LLDB back-end, there is no proper mechanism to make a process
+        # start from inside of a command, as there is in GDB. Ideally, we'd
+        # rework `ProcessDriver` so that it lets us do that with an execution
+        # controller, but that is quite a bit of work to fix a single command,
+        # when using an override is enough to achieve the same goal.
+        #
+        # In any case, we should consider doing it if this proves to be too janky.
+        if not driver.has_process():
+            process_launch(driver, relay, ["-s"], dbg)
```
disconnect3d, line 532: This fixes the issue, but shouldn't this really be in the entry command implementation? it has a TODO, so if we fix it here, we should update its comment at least.

disconnect3d, line 532: This fixes the issue, but shouldn't this really be in the entry command implementation? it has a TODO, so if we fix it here, we should update its comment at least.

mbrla0: > [...] but shouldn't this really be in the entry command implementation?

Yes! But, as Pwndbg is right now, there is just no way to launch processes from inside commands. We'd need support from the `ExecutionController`/`ProcessDriver` runtime to achieve that[^1], plus a few changes to how command arguments are parsed when there is no active inferior.

[^1]: GDB also has to contend with the same problem in `gdb.execute`, but it handles it automagically™ for us.

Maybe we'll be able to revisit that once LLDB testing is implemented, as that will need some form of an async runtime that can handle launching processes, and it might be a good idea to reuse `ExecutionController` in some form for it.

> [...] so if we fix it here, we should update its comment at least.

Fair enough, I'll change it to reflect that. For some reason that eluded me.

PR #3071, opened 2025-06-03
Make lifetime management in `ProcessDriver` more explicit

mbrla0:
Refactor process lifetime management in `ProcessDriver` some more, to further work on the issues pointed to by #2685.

This PR breaks off repeated logic in `attach` and `launch`, while also making the distinction between remote and local operations clearer. Additionally, it adds types for operation results in the process driver that make result conditions and what they entail more explicit, as well as allowing for checking them with `match` expressions.

This PR also adds code that allows the process driver and the code around it to handle early exit conditions[^1] correctly.

[^1]: An early exit happens when LLDB sends us an event containing an exit status before we have the chance to encounter even a single `eStatusStopped`.

### DIFF
--- a/pwndbg/dbg/lldb/repl/__init__.py
+++ b/pwndbg/dbg/lldb/repl/__init__.py
@@ -1150,20 +1150,17 @@ def process_connect(driver: ProcessDriver, relay: EventRelay, args: List[str], d
     io_driver = get_io_driver()
     result = driver.connect(auto.target, io_driver, args.remoteurl, "gdb-remote")
 
-    error = False
     match result:
         case LaunchResultError(what, disconnected):
             print_error(f"could not connect to remote: {what.description}")
             if disconnected:
                 print_warn("disconnected")
-            error = True
+            auto.close()
+            return
         case LaunchResultEarlyExit():
             print_warn("remote exited early")
-            error = True
-
-    if error:
-        auto.close()
-        return
+            auto.close()
+            return
 
     # Tell the debugger that the process was suspended, if there is a process.
     if driver.has_process():
--- a/pwndbg/dbg/lldb/repl/__init__.py
+++ b/pwndbg/dbg/lldb/repl/__init__.py
@@ -1027,8 +1027,6 @@ def _attach_with_info(
         print_error("a process is already being debugged")
         return
 
-    io_driver = get_io_driver()
-
     auto = AutoTarget(dbg)
     if not auto:
         print_error(f"could not create empty target for attaching: {auto.error.description}")
--- a/pwndbg/dbg/lldb/repl/__init__.py
+++ b/pwndbg/dbg/lldb/repl/__init__.py
@@ -68,6 +68,10 @@ from pwndbg.dbg.lldb.pset import pset
 from pwndbg.dbg.lldb.repl.io import IODriver
 from pwndbg.dbg.lldb.repl.io import get_io_driver
 from pwndbg.dbg.lldb.repl.proc import EventHandler
+from pwndbg.dbg.lldb.repl.proc import LaunchResultConnected
+from pwndbg.dbg.lldb.repl.proc import LaunchResultEarlyExit
+from pwndbg.dbg.lldb.repl.proc import LaunchResultError
+from pwndbg.dbg.lldb.repl.proc import LaunchResultSuccess
 from pwndbg.dbg.lldb.repl.proc import ProcessDriver
 from pwndbg.dbg.lldb.repl.readline import PROMPT
 from pwndbg.dbg.lldb.repl.readline import enable_readline
@@ -957,9 +961,15 @@ def process_launch(driver: ProcessDriver, relay: EventRelay, args: List[str], db
         os.getcwd(),
     )
 
-    if not result.success:
-        print_error(f"could not launch process: {result.description}")
-        return
+    match result:
+        case LaunchResultError(what, disconnected):
+            print_error(f"could not launch process: {what.description}")
+            if disconnected:
+                print_warn("disconnected")
+            return
+        case LaunchResultEarlyExit():
+            print_warn("process exited early")
+            return
 
     # Continue execution if the user hasn't requested for a stop at the entry
     # point of the process. And handle necessary events.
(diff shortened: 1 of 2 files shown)


pwndbg/dbg/lldb/repl/proc.py
```diff
@@ -435,130 +530,180 @@ def run_coroutine(self, coroutine: Coroutine[Any, Any, None]) -> bool:
...
-    def launch(
-        self, target: lldb.SBTarget, io: IODriver, env: List[str], args: List[str], working_dir: str
-    ) -> lldb.SBError:
+    def _prepare_listener_for(self, target: lldb.SBTarget):
         """
-        Launches the process and handles startup events. Always stops on first
-        opportunity, and returns immediately after the process has stopped.
+        Prepares the internal event listener for the given target.
+        """
+        self.listener = lldb.SBListener("pwndbg.dbg.lldb.repl.proc.ProcessDriver")
```
disconnect3d, line 537: Ugh what's that? Why do we pass here a string? Is this a class lldb will instantiate or what?

mbrla0: That's just a label for the logs, makes it easy to tell who the listener belongs to

disconnect3d:
Pls fix conflict

pwndbg/dbg/lldb/repl/__init__.py
```diff
@@ -1030,14 +1038,20 @@ def _attach_with_info(
...
         info,
     )
 
-    if not result.success:
-        print_error(f"could not attach to process: {result.description}")
-        auto.close()
-        return
+    match result:
+        case LaunchResultError(what, disconnected):
+            print_error(f"could not attatch to process: {what.description}")
```
disconnect3d:
```suggestion
            print_error(f"could not attach to process: {what.description}")
```

pwndbg/dbg/lldb/repl/proc.py
```diff
@@ -588,34 +726,39 @@ def connect(self, target: lldb.SBTarget, io: IODriver, url: str, plugin: str) ->
...
-            ):
-                # This indicates that on this implementation, connecting isn't
-                # enough to have the process launch or attach, and that we have
-                # to do that manually.
-                break
-
-        return error
+            match result:
+                case _PollResultStopped():
+                    # The process has startarted. We can fire off the created event
```
disconnect3d:
```suggestion
                    # The process has started. We can fire off the created event
```

pwndbg/dbg/lldb/repl/proc.py
```diff
@@ -435,130 +530,180 @@ def run_coroutine(self, coroutine: Coroutine[Any, Any, None]) -> bool:
...
+        return error
 
-        if not error.success:
-            # Undo any initialization Launch() might've done.
-            self.process = None
-            self.listener = None
-            return error
+    def _attach_remote(self, pid: int) -> lldb.SBError:
+        """
+        Attatch to a process in a remote debugserver.
```
disconnect3d:
```suggestion
        Attach to a process in a remote debugserver.
```

pwndbg/dbg/lldb/repl/__init__.py
```diff
@@ -1134,10 +1148,20 @@ def process_connect(driver: ProcessDriver, relay: EventRelay, args: List[str], d
...
+        case LaunchResultError(what, disconnected):
+            print_error(f"could not connect to remote: {what.description}")
+            if disconnected:
+                print_warn("disconnected")
+            error = True
+        case LaunchResultEarlyExit():
+            print_warn("remote exited early")
+            error = True
+
+    if error:
```
disconnect3d: Lines 1044-1054 have very similar logic, but they do `auto.close(); return` within the match cases instead.

The two differs in the error messages... shall we maybe put that in a helper function and just call it with proper err msgs?

Or just have it do the same logic here and there but consistently?

pwndbg/dbg/lldb/repl/proc.py
```diff
@@ -60,6 +59,93 @@ def modules_loaded(self):
         pass
 
 
+class _PollResult:
```
disconnect3d, line 62: Honestly that's a lot of result-like classes. I have mixed feelings about that, but idk...

What would be the alternative here?

Also, why do those start with `_` and the `Launch*` classes do not?

pwndbg/dbg/lldb/repl/proc.py
```diff
@@ -435,130 +530,180 @@ def run_coroutine(self, coroutine: Coroutine[Any, Any, None]) -> bool:
...
+        args: List[str],
+        working_dir: str | None,
+    ) -> LaunchResult:
+        """
+        Launches the process and handles startup events. Always stops on first
+        opportunity, and returns immediately after the process has stopped.
+
+        Fires the created() event.
+        """
+        if self.has_connection():
```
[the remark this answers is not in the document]
disconnect3d: Please resolve this comment if this doesn't make sense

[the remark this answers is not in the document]
mbrla0: > Please resolve this comment if this doesn't make sense

Context on why it doesn't make sense: The listener is created as part of establishing the connection, not only does recreating it makes no sense, it would break things. The assertion only triggers if we somehow fail to prepare the listener during the connection, but still have a valid process. But then again, catching this sort of mistake is exactly what the assertion is there for.

pwndbg/dbg/lldb/repl/proc.py
mbrla0: > Honestly that's a lot of result-like classes. I have mixed feelings about that, but idk...
>
> What would be the alternative here?

Well, the alternative would probably be to use `enum.Enum`, but those can't have data associated with their variants in an ADT-like way, so we'd end up having to return a tuple of `(Status, Any | None)` or something else that's basically just as hacky.

We could get rid of the base class and do something like `ResultType = ClassA | ClassB | ClassC | ...`, but ultimately there's probably no getting away from using classes here if we want to keep things legible.

> Also, why do those start with `_` and the `Launch*` classes do not?

Those are meant to be private to the module. Doesn't make sense to expose conditions like `TimedOut` or the exact LLDB event causing a stop to other modules as a launch result.

pwndbg/dbg/lldb/repl/__init__.py
mbrla0: I've changed them so they're the same now.

PR #3074, opened 2025-06-03
fix bug for att flavour

### LINKED ISSUE #3073: disasm annotations and emulation for non-x86 architectures with at&t syntax is broken
### Description

Because pictures speak louder than words, this is a riscv64 example of the issue:

Image

With att syntax set, the annotation tries to dereference the opposite register - s6, instead of the stack pointer. The same happens on aarch64 (and, as far as I can tell, all the other architectures but I haven't tested it):

Image

There's also a related issue with emulation on riscv64, sometimes I can't view the context and I get this instead:

Image

I'm guessing the bug is because of how capstone handles att syntax, I was able to fix both issues with

```diff
diff --git a/pwndbg/aglib/disasm/disassembly.py b/pwndbg/aglib/disasm/disassembly.py
index dfce409d..e648383b 100644
--- a/pwndbg/aglib/disasm/disassembly.py
+++ b/pwndbg/aglib/disasm/disassembly.py
@@ -108,7 +108,11 @@ def get_disassembler(cs_info: Tuple[int, int]):
 
     flavor = pwndbg.dbg.x86_disassembly_flavor()
     try:
-        cs.syntax = CapstoneSyntax[flavor]
+        # On other architectures than x86 and x86-64 the arguments aren't swapped
+        if pwndbg.aglib.arch.name in ("i386", "x86-64"):
+            cs.syntax = CapstoneSyntax[flavor]
+        else:
+            cs.syntax = CapstoneSyntax["intel"]
     except CsError:
         pass
     cs.detail = True
```

### Steps to reproduce

Compile `int main() {}` and step few times

### My setup

```
pwndbg> version
Pwndbg:   2025.05.30 build: d380c99
Python:   3.13.3 (main, Apr  9 2025, 07:44:25) [GCC 14.2.1 20250207]
GDB:      16.3
Capstone: 6.0.0
Unicorn:  2.1.3
Pwnlib:   4.14.0
```

poniponiponiponiponiponiponiponiponi:
fix #3073

### DIFF
--- a/pwndbg/aglib/disasm/disassembly.py
+++ b/pwndbg/aglib/disasm/disassembly.py
@@ -108,7 +108,11 @@ def get_disassembler(cs_info: Tuple[int, int]):
 
     flavor = pwndbg.dbg.x86_disassembly_flavor()
     try:
-        cs.syntax = CapstoneSyntax[flavor]
+        # On other architectures than x86 and x86-64 the arguments aren't swapped
+        if pwndbg.aglib.arch.name in ("i386", "x86-64"):
+            cs.syntax = CapstoneSyntax[flavor]
+        else:
+            cs.syntax = CapstoneSyntax["intel"]
     except CsError:
         pass
     cs.detail = True


pwndbg/aglib/disasm/disassembly.py
```diff
@@ -108,7 +108,11 @@ def get_disassembler(cs_info: Tuple[int, int]):
 
     flavor = pwndbg.dbg.x86_disassembly_flavor()
     try:
-        cs.syntax = CapstoneSyntax[flavor]
+        # On other architectures than x86 and x86-64 the arguments aren't swapped
+        if pwndbg.aglib.arch.name in ("i386", "x86-64"):
```
OBarronCS, line 112: I added a fix in #3077 to fix this in a slightly different location (the place in pwndbg code where the operands are reversed).

OBarronCS:
@poniponiponiponiponiponiponiponiponi, thanks for catching this bug and making this PR! While debugging this, I ended up implementing a fix in a separate PR (https://github.com/pwndbg/pwndbg/pull/3077) which handles another case (arch=i8086) and places the fix in a slightly different location, closer to the site where the bug takes effect.

Issue #3072, opened 2025-06-03
Support fzf completion in lldb like GEP


zhouzq-thu:
Support [fzf](https://github.com/junegunn/fzf) completion in lldb like [GEP](https://github.com/lebr0nli/GEP)

## All commands complete
![Image](https://github.com/user-attachments/assets/79e74c98-2560-4188-afed-d4e575402f15)

## breakpoint subcommands completion
![Image](https://github.com/user-attachments/assets/6f4ee53b-2c80-43bc-af51-1b9a58591a85)

## breakpoint completion
 ![Image](https://github.com/user-attachments/assets/233f47d4-cc58-4e99-a2f5-31a338aabf7f)

k4lizen:
omg that looks so cool

CptGibbon:
Aren't these just screenshots of GEP by @lebr0nli which is explicitly compatible with pwndbg?

CptGibbon:
Aren't these just screenshots of GEP by @lebr0nli?
Since GEP is explicitly compatible with pwndbg perhaps this should be a GEP issue?

zhouzq-thu:
@CptGibbon, please try the pull request.

CptGibbon:
Ah the ol' write the PR before the issue technique

k4lizen:
maybe there is also the option of packaging GEP and bundling it with pwndbg, in case we or the author doesn't want pwndbg to annex the GEP codebase

lebr0nli:
> maybe there is also the option of packaging GEP and bundling it with pwndbg, in case we or the author doesn't want pwndbg to annex the GEP codebase

It would be super cool if pwndbg were bundled with GEP, and I'd be happy to help make that happen if we decided to do so :)

Here are a few thoughts on how we might approach this:
- Since I want GEP to also support GEF and other use cases, I'll probably continue maintaining it as a separate codebase. So for the long term matintenance, rather than moving GEP's logic directly into pwndbg, maybe it's better to make part of GEP into an importable Python module and use it similarly to how we use [gdb-pt-dump](https://github.com/martinradev/gdb-pt-dump). (But if it turns out that implementing it directly in pwndbg is the better approach, I'm happy to help maintain it.)
[...]

PR #3075, opened 2025-06-03
Implement fzf completion in lldb like GEP

zhouzq-thu:
Implement the feature proposed in #3072.


CptGibbon:
@lebr0nli your thoughts on this?
Perhaps you'd like the MIT license included inline?

pwndbg/dbg/lldb/repl/fuzzy.py
```diff
@@ -0,0 +1,307 @@
...
+    proposals = [matches.GetStringAtIndex(i) for i in range(1, matches.GetSize())]
+    return [i for i in proposals]
+
+
+def safe_get_help_docs(dbg: LLDB, command: str) -> str | None:
+    """
+    A wrapper for gdb.execute('help <command>', to_string=True), but return None if gdb raise an exception.
+    """
+    try:
+        return dbg._execute_lldb_command("help %s" % command).strip()
```
mbrla0: Is there no other way to do this? (Perhaps `SBCommandInterpreter::HandleCompletionWithDescriptions`)

`_execute_lldb_command` is an implementation detail and so _really_ not meant to be used outside the classes in `pwndbg/dbg/lldb/__init__.py`. It's not an LLDB version of `gdb.execute`, it will mess Pwndbg up if you use it wrong, and there are no guarantees on what constitutes "using it wrong".

Seeing as this does not touch anything process-related, it is unlikely to break anything, but I would still strongly prefer this be done another way, if possible.

pwndbg/dbg/lldb/repl/__init__.py
```diff
@@ -308,12 +317,17 @@ def handle_sigint(_sig, _frame):
...
+                    line = session.prompt(message=PROMPT)
+                else:
+                    line = input(PROMPT)
                 # If the input is empty (i.e., 'Enter'), use the previous command
                 if line:
                     last_command = line
                 else:
                     line = last_command
+            except KeyboardInterrupt:
+                continue
```
mbrla0: Did you encounter any instances of `KeyboardInterrupt`? We handle SIGINT ourselves, so normally I don't think we would reach that part of the code.

pwndbg/dbg/lldb/repl/fuzzy.py
```diff
@@ -0,0 +1,307 @@
...
+            functools.partial(
+                fzf_tab_autocomplete,
+                dbg=dbg,
+                preview=preview,
+                fifo_in=fifo_in,
+                fifo_out=fifo_out,
+            )
+        )
+    else:
+        print("Install fzf for better experience with pwndbg")
```
mbrla0: I don't get what the need for this message is. Plus, it seems this function is never called unless `HAZ_FZF` is true, anyway. Maybe take this out?
```suggestion
    # key binding for fzf history search
    bindings.add("c-r")(fzf_reverse_search)
    # key binding for fzf tab completion
    fifo_in, fifo_out = create_preview_fifos()
    preview = "echo {n} > %s\ncat %s" % (fifo_in, fifo_out)
    bindings.add("c-i")(
        functools.partial(
            fzf_tab_autocomplete,
            dbg=dbg,
            preview=preview,
            fifo_in=fifo_in,
            fifo_out=fifo_out,
        )
    )
```

pwndbg/dbg/lldb/repl/__init__.py
zhouzq-thu: This handle the exception raised from `session.prompt`.

pwndbg/dbg/lldb/repl/fuzzy.py
zhouzq-thu: Use `SBCommandInterpreter::HandleCommand` to implement.

pwndbg/dbg/lldb/repl/fuzzy.py
zhouzq-thu: Accepted.

k4lizen:
@mbrla0 could this be done in a debugger-agnostic way? it seems GEP does support gdb

disconnect3d:
> @lebr0nli your thoughts on this? Perhaps you'd like the MIT license included inline?

Just to be sure we have this there:
```
# Licensed under the MIT License
# Modified from https://github.com/lebr0nli/GEP
# Copyright (c) 2022 Alan Li
# Copyright (c) 2025 Zhi-Qiang Zhou
```
But we can include the full license text if you want.

lebr0nli:
> > @lebr0nli your thoughts on this? Perhaps you'd like the MIT license included inline?
> 
> Just to be sure we have this there:
> 
> ```
> # Licensed under the MIT License
> # Modified from https://github.com/lebr0nli/GEP
> # Copyright (c) 2022 Alan Li
> # Copyright (c) 2025 Zhi-Qiang Zhou
> ```
> 
> But we can include the full license text if you want.

Thanks for the ping!
Cool, I haven't tried migrating my code to LLDB.
I haven't fully reviewed the PR, but since the abstractions and implementations are very similar to GEP, it might have the same bugs. (See also [this comment](https://github.com/pwndbg/pwndbg/issues/3072#issuecomment-2939873469).)
I'm a heavy GDB user but not very familiar with LLDB, so it would be great if someone who uses LLDB a lot could check whether these features introduce any issues there.
Regarding the license, yeah, I'm good with the current one.

mbrla0:
> @mbrla0 could this be done in a debugger-agnostic way? it seems GEP does support gdb

We don't fully control the CLI loop in GDB like we do in LLDB, so I wouldn't really count on it being possible, at least not without some major hacking.

But I also don't think it's so out there of a possibility that it's not worthy of some investigation.

pwndbg/dbg/lldb/repl/fuzzy.py
```diff
@@ -0,0 +1,307 @@
...
+    proposals = [matches.GetStringAtIndex(i) for i in range(1, matches.GetSize())]
+    return [i for i in proposals]
+
+
+def safe_get_help_docs(dbg: LLDB, command: str) -> str | None:
+    interp: lldb.SBCommandInterpreter = dbg.debugger.GetCommandInterpreter()
+    add_to_history = False
+    result = lldb.SBCommandReturnObject()
+    interp.HandleCommand(f"help {command}", result, add_to_history)
+    return result.GetOutput()
```
mbrla0, line 74: While this is better in that it doesn't call that function, I still would like to know if there really is no way to do this without invoking an LLDB command.

pwndbg/dbg/lldb/repl/__init__.py
mbrla0: We handle ^C ourselves as part of how we handle inferior interrupts and operation cancellations. If you're getting any `KeyboardInterrupts` at all in that part of the code that might point to a deeper issue with how this is being integrated into Pwndbg.

pwndbg/dbg/lldb/repl/fuzzy.py
zhouzq-thu: As far as I know, the [CommandObjectHelp.cpp](https://github.com/llvm/llvm-project/blob/main/lldb/source/Commands/CommandObjectHelp.cpp) is used to generate help docs, however the class is not exposed.

disconnect3d:
Pls fix lint.

@mbrla0 or @lebr0nli pls let us know when u feel this can be merged

pwndbg/dbg/lldb/repl/fuzzy.py
mbrla0: What about `HandleCompletionWithDescriptions`?

zhouzq-thu: The `HandleCompletionWithDescriptions` only generate a brief description:

Screenshot 2025-06-05 at 05 44 02

However, with help:

Screenshot 2025-06-05 at 05 46 09

zhouzq-thu:
> Pls fix lint.
> 
> @mbrla0 or @lebr0nli pls let us know when u feel this can be merged

Fixed.

pwndbg/dbg/lldb/repl/__init__.py
```diff
@@ -64,16 +64,23 @@
 from pwndbg.dbg.lldb import OneShotAwaitable
 from pwndbg.dbg.lldb.pset import pget
 from pwndbg.dbg.lldb.pset import pset
+from pwndbg.dbg.lldb.repl.fuzzy import HAS_FZF
```
lebr0nli: If we're only using `pwndbg.dbg.lldb.repl.fuzzy` when `fzf` is present, then perhaps we could define the `HAS_FZF` directly in `pwndbg/dbg/lldb/repl/__init__.py` instead of importing it with `from pwndbg.dbg.lldb.repl.fuzzy import HAS_FZF`? That way, future developers won't need to trace or modify the definition across different files. It possibly better for performance too (since we don't need to import the entire `pwndbg.dbg.lldb.repl.fuzzy` if we don't need it).

Note that `pwndbg.dbg.lldb.repl.fuzzy` itself doesn't depend on `HAS_FZF` actually. I originally defined `HAS_FZF` in GEP because GEP also supports the scenario where fzf is not available. So, I believe you can safely remove the `if not HAS_FZF` checks from `pwndbg.dbg.lldb.repl.fuzzy` as well.

zhouzq-thu: Yeap.

disconnect3d:
@lebr0nli can we merge this?

lebr0nli:
> @lebr0nli can we merge this?

Yeah, I think we can merge it now.

pwndbg/dbg/lldb/repl/fuzzy.py
mbrla0: Alrighty, then. I think resorting to the command in this case is justified.

mbrla0:
> Pls fix lint.
> 
> @mbrla0 or @lebr0nli pls let us know when u feel this can be merged

All-clear for merging on my end.

PR #3077, opened 2025-06-04
Fix at&t disassembly-flavor setting breaking non-x86 architectures

### LINKED ISSUE #3073: disasm annotations and emulation for non-x86 architectures with at&t syntax is broken

OBarronCS:
Fix #3073

The implementation didn't take into account that the GDB setting `set disassembly-flavor` could be set to `att` while debugging a non-x86 architecture.

The PR also adds a test for this.

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -1092,6 +1092,7 @@ def context_disasm(target=sys.stdout, with_banner=True, width=None):
     # The `None` case happens when the cache was not filled yet (see e.g. #881)
     if cs is not None and cs.syntax != syntax:
         pwndbg.lib.cache.clear_caches()
+        pwndbg.aglib.disasm.disassembly.computed_instruction_cache.clear()
 
     result = try_emulate_if_bug_disable(
         lambda: pwndbg.aglib.nearpc.nearpc(
--- a/pwndbg/aglib/disasm/instruction.py
+++ b/pwndbg/aglib/disasm/instruction.py
@@ -249,7 +249,7 @@ class PwndbgInstructionImpl(PwndbgInstruction):
         # For ease, for x86 we will assume Intel syntax (destination operand first).
         # However, Capstone will disassemble using the `set disassembly-flavor` preference,
         # and the order of operands are read left to right into the .operands array. So we flip operand order if AT&T
-        if self.cs_insn._cs.syntax == CS_OPT_SYNTAX_ATT:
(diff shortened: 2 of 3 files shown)


pwndbg/commands/context.py
```diff
@@ -1092,6 +1092,7 @@ def context_disasm(target=sys.stdout, with_banner=True, width=None):
     # The `None` case happens when the cache was not filled yet (see e.g. #881)
     if cs is not None and cs.syntax != syntax:
         pwndbg.lib.cache.clear_caches()
+        pwndbg.aglib.disasm.disassembly.computed_instruction_cache.clear()
```
OBarronCS, line 1095: Make it so changing disassembly flavor mid-debugging doesn't result in dangling instructions with the previous syntax style.

disconnect3d: Shouldn't .clear_caches() do this already?

PR #3079, opened 2025-06-04
Skip test_command_procinfo.py if unable to bind to address

AC01010:
Patches test_command_procinfo.py to skip test if unable to bind to address; encountering failed test in WSL2 because of lack of IPv6 support.


pwndbg/commands/__init__.py
```diff
@@ -906,6 +906,7 @@ def load_commands() -> None:
     import pwndbg.commands.kchecksec
     import pwndbg.commands.kcmdline
     import pwndbg.commands.kconfig
+    import pwndbg.commands.kdmesg
```
disconnect3d: This change should not belong to this PR, same for the kdmesg.py file

PR #3081, opened 2025-06-05
Improve argument parsing and launch handling in LLDB Pwndbg

### LINKED ISSUE #2918: Support additional command line arg for lldb
The current pwndbg lldb python repl only supports passing the target as a single argument. This is annoying when I want to attach to an existing process using the command line.

It should support at least the following:
```
--attach-name <name> Tells the debugger to attach to a process with the given name
--attach-pid <pid>   Tells the debugger to attach to a process with the given pid.
--wait-for           Tells the debugger to wait for a process with the given pid or name to launch before attaching.
```
There are a lot more flags that also would be nice to have but I think those are the most useful.

mbrla0:
Fixes #2918

Moves `pwndbg-lldb.py` to `ArgumentParser` and implements more startup options. Adds support for raw arguments with `--` and shell expansion in `process launch` and brings the behavior of `run` in line with LLDB.

It is important to note that, after this PR is merged, `run` will not accept `process launch` flags anymore. This is intended, and matches the behavior of the LLDB CLI. Having `run` accept those flags was an uninentional deviation in Pwndbg.

### DIFF
--- a/pwndbg-lldb.py
+++ b/pwndbg-lldb.py
@@ -2,11 +2,29 @@
 
 from __future__ import annotations
 
+import argparse
 import os
 import re
 import subprocess
 import sys
 from typing import List
+
+PARSER = argparse.ArgumentParser(prog="pwndbg-lldb")
+PARSER.add_argument("-v", "--verbose", action="store_true", help="Enable debug output")
+PARSER.add_argument("target", nargs="?")
+parser_attach = PARSER.add_mutually_exclusive_group()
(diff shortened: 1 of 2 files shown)


mbrla0:
Still missing launch arguments, so still a draft. I'll implement them tomorrow.

disconnect3d:
Pls fix lint

mbrla0:
Alrighty. I think this one's ready for review.

PR #3082, opened 2025-06-05
Revert "Improving `vmmap` output when debugging kernel"

patryk4815:
Reverts pwndbg/pwndbg#3020

reverting due to untested on other archs (eg aarch64)

aarch64:
image

x86-64 (qemu is started with -S, so kernel is not started yet, but it should be still allowed to debug at this stage..):
image


jxuanli:
thanks for catching those edge cases, those should have been fixed although i believe the x64 issue would also exist for the current `dev` (as you can expect somewhat when the kernel's mem hasn't yet been stabalized)

patryk4815:
@jxuanli current dev don't crash context 

image

PR #3087, opened 2025-06-05
mallocng: print information about slot, group, meta at given address

k4lizen:
Dump information about a slot given the start of it's user memory:
image
image

Similarly for group and meta:
image
image


CptGibbon:
Will this be split into smaller PR's? At over 1000 lines, this could take some time to review.

k4lizen:
@CptGibbon I kinda tried with https://github.com/pwndbg/pwndbg/pull/3096 but :S 

I generally rebased commits so they should be mostly incremental, it should be no problem to split this into multiple PRs. I would also like to get at least some of this reviewed soonish in case any major changes are requested. Currently the biggest blocker for me opening this for review is what to do with 3096.

Lmk what you think.

btw pretty much the whole of aglib/heap/mallocng.py is "read struct at memory address" boilerplate so the LOC is a bit misleading

CptGibbon:
> @CptGibbon I kinda tried with https://github.com/pwndbg/pwndbg/pull/3096 but :S 

@disconnect3d is there anything holding back #3096

CptGibbon:
@k4lizen could mallocng-explain easily be moved into its own PR? That looks like an easy merge.

k4lizen:
https://github.com/pwndbg/pwndbg/pull/3104

disconnect3d:
> > @CptGibbon I kinda tried with #3096 but :S
> 
> @disconnect3d is there anything holding back #3096

I was hesitant to add it in its current version, but I'm okay to add it if its needed here. Merged.

disconnect3d:
@CptGibbon Shall we have a branch 'malloc-ng' and merge things to it instead?

disconnect3d (commented):
Btw is mallocng only used within musl? If yes, shall we call commands 'musl-*' maybe since mallocng is a bit long? just loud thinking

k4lizen:
> Btw is mallocng only used within musl? If yes, shall we call commands 'musl-*' maybe since mallocng is a bit long? just loud thinking

To my knowledge yes, however the musl sources have a folder for the older version of the allocator. Mallocng was merged around ~5 years ago I think. In practice I doubt this will be a problem though. 

All `mallocng-*` commands are aliased as `ng-*` so command length shouldn't be a problem. I would even consider having them aliased as `ng*` or (something else that doesn't use a dash) since if you're debugging the allocator you would probably be running the commands often and the dash requires a reach on the keyboard; but couldn't think of something that feels good.

k4lizen:
I'd like to take out the `add mallocng-explain command` commit from here but can't since it introduces the MUSL command category. If that (https://github.com/pwndbg/pwndbg/pull/3104) is merged I'll rebase this onto dev and can open it for review

k4lizen:
Changed it from `ng-uslot` to `ng-slotu` because I feel it rolls nicer. I'll also be adding `ng-slots` (**slot** **s**tart) which finds a slot given its internal start address rather than user start address. But that'll be in next PR.

CptGibbon:
> I would even consider having them aliased as `ng*`

A thought about this that could be a later addition;

Perhaps after detecting (or the user specifying) a memory allocator, some common heap commands e.g. `heap` could do something different based on the detected allocator?

That way user-friendly commands like `heap` and `vis[-heap-chunks]` aren't GLIBC malloc specific.

pwndbg/commands/mallocng.py
```diff
@@ -144,3 +151,322 @@ def mallocng_explain() -> None:
...
+
+@pwndbg.commands.Command(
+    parser,
+    category=CommandCategory.MUSL,
+    aliases=["ng-meta"],
+)
+@pwndbg.commands.OnlyWhenRunning
+def mallocng_meta(address: int) -> None:
+    if not memory.is_readable_address(address):
+        print(message.error(f"Address {hex(address)} not readable."))
```
disconnect3d:
```suggestion
        print(message.error(f"Address {address:#x} not readable."))
```

pwndbg/commands/mallocng.py
```diff
@@ -144,3 +151,322 @@ def mallocng_explain() -> None:
...
+
+@pwndbg.commands.Command(
+    parser,
+    category=CommandCategory.MUSL,
+    aliases=["ng-group"],
+)
+@pwndbg.commands.OnlyWhenRunning
+def mallocng_group(address: int) -> None:
+    if not memory.is_readable_address(address):
+        print(message.error(f"Address {hex(address)} not readable."))
```
disconnect3d:
```suggestion
        print(message.error(f"Address {address:#x} not readable."))
```

pwndbg/commands/mallocng.py
```diff
@@ -144,3 +151,322 @@ def mallocng_explain() -> None:
...
+        return
+
+    print(dump_group(group), end="")
+
+    try:
+        meta = group.meta
+        meta.preload()
+        print(dump_meta(meta), end="")
+    except pwndbg.dbg_mod.Error as e:
+        print(message.error(f"Failed loading meta: {str(e)}"))
```
disconnect3d: I believe the str() is not needed?

pwndbg/lib/pretty_print.py
```diff
@@ -61,24 +61,51 @@ def add(self, prop_group: List[Property]) -> None:
...
+        max_value_len = 16
 
         for prop in prop_group:
             self.text += self.indent_level * self.indent_size * " "
+
             colored_name = self.name_color_func(prop.name) + ":"
-            self.text += colored_name.ljust(max_name_len + 1, " ")
+            self.text += color.ljust_colored(colored_name, max_name_len + 1)
+
             self.text += self.padding * " "
```
disconnect3d: Could probably be sth like:

```py
prefix = self.indent_level * self.indent_size * " "
suffix = ":" + self.padding * " "

for prop in prop_group:
    self.text += prefix + color.ljust_colored(self.name_color_func(prop.name), max_name_len+1) + suffix

    ...
```

Fwiw max_name_len should maybe have +1 included (or be renamed, idk)?:P

pwndbg/lib/pretty_print.py
```diff
@@ -61,24 +61,51 @@ def add(self, prop_group: List[Property]) -> None:
...
+                # list of strings, we want each one under the other
+                assert isinstance(prop.extra, list)
+                assert len(prop.extra) > 1
+
+                self.text += "  " + prop.extra[0]
+                for i in range(1, len(prop.extra)):
+                    self.text += "\n"
+                    self.text += self.indent_level * self.indent_size * " "
+                    self.text += (max_name_len + 1) * " "
+                    self.text += self.padding * " "
```
disconnect3d: We're computing things like `self.padding * " "` many times :P

disconnect3d (commented):
I'm ok to merge this in its current version but I left some comments too

k4lizen:
done

btw `ng-slotu` currently only works on allocated slots. i'm looking to add support for free and avail slots in a future PR.

Issue #3090, opened 2025-06-07
heap tracker: color or mark pointers?


disconnect3d:
It would be awesome if heap tracker (`track-heap enable`) printer for malloc/free etc would color pointers uniquely.

This would help to see when a given pointer is allocated and freed as currently, we only see output like this:

```
[*] malloc(                  16) -> 0x55555557ca00, 32 bytes real size
[*] free(0x55555557c9e0)
[*] malloc(                  32) -> 0x55555557ca20, 48 bytes real size
[*] free(0x55555557ca00)
[*] free(0x55555557ca20)
```

And it may be hard to catch visually which malloc corresponds to which free.

kojish2018:
I’m interested in helping out with this. Is this issue still open for contributions? if so, let me know!

varchasvakhare2022:
I’d like to work on this. Is it still open?

disconnect3d:
Hi, I hid both of your comments for spam. Please do not ask if an issue is open or if you can be assigned.  The issues are open when they are open. Sure, there are mistakes when we forget about something, but then you can just figure it out yourself if a feature is implemented or not -- and then let us know if an issue needs to be closed etc.

Otherwise, if you want to help us and contribute, just implement the thing and send us a pull request.

At the same time, please, do your homework and don't send us code generated by LLM you don't understand or haven't even tested yourself.

hiratinspace:
Hi! I'm a student in the CodePath AI301 program and I'd like to work on this issue. I'll start by reproducing the problem locally this week. Any pointers on where to start in the codebase would be appreciated! Thanks

PR #3094, opened 2025-06-07
assign and print id for malloc/free pairs (track-heap)

syttpz:
This PR addresses the "mark pointers" part of #3090, helping visually correlate malloc/free pairs.

Local test output: 

```
[*] malloc(32) -> 0x4052a0, 0x30 bytes real size (id=1)
[*] malloc(64) -> 0x4052d0, 0x50 bytes real size (id=2)
[*] malloc(1024) -> 0x405320, 0x410 bytes real size (id=3)
[*] free(0x4052a0) (id=1)
[*] free(0x4052d0) (id=2)
[Inferior 1 (process 136) exited normally]
```

### DIFF
--- a/pwndbg/gdblib/ptmalloc2_tracking.py
+++ b/pwndbg/gdblib/ptmalloc2_tracking.py
@@ -199,6 +199,9 @@ class Tracker:
         self.free_watchpoints: Dict[int, FreeChunkWatchpoint] = {}
         self.memory_management_calls: Dict[int, bool] = {}
 
+        self.alloc_ids: Dict[int, int] = {}
+        self.next_id: int = 1
+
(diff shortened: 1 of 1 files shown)


disconnect3d:
I am not sure about it.

The IDs make some sense, but we want the line to be as easily parseable by human as possible. In:
> [*] malloc(32) -> 0x4052a0, 0x30 bytes real size (id=1)

This is way too hard to parse. There are at least 4 numbers to consider: allocation size, returned address, real size and id.

And there are no colors.

Also consider that we may want to display the allocation site/origin somehow in the future.

syttpz:
I see, I'll take some time to rethink the format and explore ways to make it cleaner and easier to parse. Appreciate the feedback.

PR #3096, opened 2025-06-08
Facility for formatting command output

k4lizen:
I want to apply some formatting on the command outputs for https://github.com/pwndbg/pwndbg/pull/3087 . I would like to not repeat code, so I factored out the formatting logic to `pwndbg/lib`. 

#### Code
```python
    pp = PropertyPrinter()
    pp.start_section("slab")
    pp.set_padding(7)
    pp.add(
        [
            Property(name="group", value=slot.group.addr, is_addr=True),
            Property(name="meta", value=slot.meta.addr, is_addr=True),
        ]
    )
    pp.end_section()

    pp.start_section("general")
    pp.set_padding(2)
    pp.add(
        [
            Property(name="start", value=slot.start, is_addr=True),
            Property(name="end", value=slot.end, is_addr=True),
            Property(name="stride", value=slot.meta.stride),
            Property(name="user start", value=slot.p, is_addr=True),
            Property(name="user size", value=slot.user_size),
        ]
    )
    pp.end_section()

    pp.start_section("in-band")
    pp.set_padding(4)
    pp.add(
        [
            Property(name="offset", value=slot.offset),
            Property(name="index", value=slot.idx),
            Property(name="reserved", value=slot.reserved),
        ]
    )
    pp.end_section()

    pp.print()
```
#### Output
image

### DIFF
--- a/pwndbg/lib/pretty_print.py
+++ b/pwndbg/lib/pretty_print.py
@@ -1,0 +1,143 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any
+from typing import Callable
+from typing import List
+from typing import Optional
+
+import pwndbg.color as color
+
+
+@dataclass
+class Property:
+    """
+    A (property name, property value) pair
+    with optional extra information.
+
+    Used by the PropertyPrinter.
+    """
+
+    name: str
+    value: Any
+    extra: str = ""
+    is_addr: bool = False
+    use_hex: bool = True
+
+
+class PropertyPrinter:
+    """
+    When you have (property name, property value) pairs
+    that you want to print, each on a new line.
+    """
+
+    def __init__(
+        self,
+        *,
+        name_color_func: Optional[Callable[[str], str]] = None,
+        value_color_func: Optional[Callable[[str], str]] = None,
+        section_color_func: Optional[Callable[[str], str]] = None,
+        indent_size: int = 2,
+    ):
+        self.name_color_func = name_color_func
+        if self.name_color_func is None:
+            self.name_color_func = color.bold
+
+        self.value_color_func = value_color_func
+        if self.value_color_func is None:
+            self.value_color_func = color.yellow
+
+        self.section_color_func = section_color_func
+        if self.section_color_func is None:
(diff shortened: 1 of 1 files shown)


disconnect3d:
Random thought: isn't this lots of unnecessary calls - start_section, padding, end_section - does it make sense to ever call any of this twice? probably not?

Wouldnt such interface be better?

```py
print_props((
    Section("slab", padding=7),
    Property(name="group", value=slot.group.addr, is_addr=True),
    Property(name="meta", value=slot.meta.addr, is_addr=True),

    Section("general", padding=2),
    Property(name="start", value=slot.start, is_addr=True),
    Property(name="end", value=slot.end, is_addr=True),
    Property(name="stride", value=slot.meta.stride),
    Property(name="user start", value=slot.p, is_addr=True),
    Property(name="user size", value=slot.user_size),
))
```

k4lizen:
> does it make sense to ever call any of this twice? probably not?

yes if you want subsections for example, you can do
```python
    pp.start_section("outersec")
    pp.set_padding(7)
    pp.start_section("innersec1")
    pp.add([
            Property(name="awa", value=nya),
     ])
    pp.end_section()
    pp.start_section("innersec2")
    pp.add([
            Property(name="awa2", value=nya2),
     ])
    pp.end_section()
    pp.end_section()
```
and this can be useful, kernel commands have lots of nested sections for instance 

btw: the main reason I even have the `set_padding` calls is to make sure the values are aligned between different property groups - because I think it looks nicer that way. Ideally you wouldn't have to do this and they would just be aligned by default - but you have to know the length of the longest property name to perform alignment and you don't know it ahead of time.

the approach I took with applying operations on the class is more (easily) extendible than supplying an array of objects like you did. e.g. here there is no clear way to perform subsections, indent, pure text write. but your thing is much more concise for the consumer of the API; and there would be no need to specify padding for instance since all properties all passed at once which is really neat.

I leaned on the side of flexibility with my implementation because im not 100% sure yet what I will want/need for the musl commands. If you want we can close this for now and revisit it after the picture is a bit clearer.

PR #3097, opened 2025-06-09
Add command to print dmesg

AC01010:
The kdmesg command displays the kernel ring buffer (dmesg) contents. Supports only the "new" kernel [ring buffer implementation](https://github.com/torvalds/linux/blob/19272b37aa4f83ca52bdf9c16d5d81bdd1354494/kernel/printk/printk_ringbuffer.h#L61) that is present in kernel versions 5.10+.


AC01010:
image

AC01010:
image

peace-maker:
Judging by the first screenshot there's a rounding difference for the times as seen in the x86/PAT log line.

PR #3101, opened 2025-06-11
Rework structure of testing framework

mbrla0:
This is the initial rework of the testing framework, with the intent of adding support for LLDB tests. It is mostly concerned with restructuring the current testing system to allow better integration with LLDB in a later commit, around two main concepts: **hosts** and **libraries**.

**Libraries** are logical collections of tests, grouped around the kind of workloads they execute and what debuggers can execute them, stored under `tests/library`. **Hosts** are the different debuggers that can host testing code, whose instrumentation code for testing is stored under `tests/host`.

The five libraries are stored in `tests/library/gdb`, `tests/library/lldb`, `tests/library/dbg`, and `tests/library/qemu-{user,system}`. With `gdb` holding `gdb`-only tests, `lldb` holding `lldb`-only tests, `dbg` holding tests that may run under either LLDB or GDB, and `qemu-{user,system}` simply being a renaming of the `cross-arch` and `kernel` tests. Initially `gdb` keeps the same tests it always has, and as tests get ported from it, they get moved to `dbg`.

This structure is intended to facilitate keeping the Pwndbg CI/CD pipelines working as they always have, while allowing for incremental porting of tests to LLDB.


mbrla0:
There's a lot of renaming on this one, so it looks like a much bigger change than it actually is.

mbrla0:
Also still a draft, since it's not _quite_ fully integrated into the CI/CD pipelines. I'll get to that tomorrow. Plus, some feedback.

mbrla0:
Alright, functionality seems to be catched-up with upstream

tests/library/qemu-user/conftest.py
```diff
@@ -59,6 +58,65 @@
...
+
+        # these three are necessary just to get the port into a TIME_WAIT state
+        with contextlib.closing(socket()) as s2:
+            s2.connect(sockname)
+            sock, _ = s.accept()
+            with contextlib.closing(sock):
+                return sockname[1]
+
+
+def ensure_qemu_port():
```
patryk4815, line 110: why not just pass `reserve_port()` into qemu_assembly_run/qemu_start_binary?
QEMU_PORT can be removed from globals, not used outside this file

mbrla0: > why not just pass reserve_port() into qemu_assembly_run/qemu_start_binary?

Doing it this way took the least amount of changes, and I'm trying to minimize those given how large the PR already is, at least to try to make it less of a chore to both review and make sure it won't break our tests once it's merged.

> QEMU_PORT can be removed from globals, not used outside this file

Yeah, you're right. I'll take it out

.dockerignore
```diff
@@ -13,4 +13,4 @@ tests/**/binaries/div_zero_binary/binary
 !tests/**/binaries/*.go
 
 # ignore QEMU test images (could also be tainted)
-tests/qemu-tests/images
+tests/library/qemu-*/images
```
k4lizen: This should be `/kimages` (I guess I missed it in https://github.com/pwndbg/pwndbg/pull/3041). Do we want to ignore them though? Since it takes some time to download them all does the docker really need to redownload them?

k4lizen:
You should probably update the "running tests" docs as well
https://pwndbg.re/pwndbg/dev/contributing/#running-tests

e.g. since this PR deletes the `./crossarch-tests.sh` file users will get confused.

mbrla0:
> You should probably update the "running tests" docs as well https://pwndbg.re/pwndbg/dev/contributing/#running-tests
> 
> e.g. since this PR deletes the `./crossarch-tests.sh` file users will get confused.

Done :D

docs/contributing/writing-tests.md
```diff
@@ -34,25 +36,25 @@ positional arguments:
...
+  -g {gdb,dbg,cross-arch-user}, --group {gdb,dbg,cross-arch-user}
+  -d {gdb}, --driver {gdb}
   -p, --pdb             enable pdb (Python debugger) post mortem debugger on failed tests
   -c, --cov             enable codecov
   -v, --verbose         display all test output instead of just failing test output
   -s, --serial          run tests one at a time instead of in parallel
   --nix                 run tests using built for nix environment
-  --collect-only        only show the output of test collection, don't run any tests
+  --collect-only        only show the output of test collection, don't run any tests```
 ```
```
k4lizen: you duplicated the closing \`\`\` here :p

docs/contributing/writing-tests.md
```diff
@@ -34,33 +36,33 @@ positional arguments:
...
-  -t, --type {gdb,cross-arch}
+  -g {gdb,dbg,cross-arch-user}, --group {gdb,dbg,cross-arch-user}
+  -d {gdb}, --driver {gdb}
   -p, --pdb             enable pdb (Python debugger) post mortem debugger on failed tests
   -c, --cov             enable codecov
   -v, --verbose         display all test output instead of just failing test output
   -s, --serial          run tests one at a time instead of in parallel
   --nix                 run tests using built for nix environment
-  --collect-only        only show the output of test collection, don't run any tests
+  --collect-only        only show the output of test collection, don't run any tests```
```
patryk4815:
```suggestion
  --collect-only        only show the output of test collection, don't run any tests
```

Issue #2796, opened 2025-03-26
Fix the failing tests for Arch Linux CI builds


disconnect3d:
Not sure why, but some Arch Linux tests fail and its annoying to see that build failing all the time.

We should either fix it or disable those tests for Arch.

E.g. https://github.com/pwndbg/pwndbg/actions/runs/14073424817/job/39411998030?pr=2795 :
```
Tests completed in 70 seconds

*********************************
********* TESTS SUMMARY *********
*********************************
Tests Passed: 209
Tests Skipped: 7
Tests Failed: 3

Failing tests:
- gdb-tests/tests/heap/test_heap_bins.py::test_heap_bins
- gdb-tests/tests/test_go.py::test_go_dumping_x64
- gdb-tests/tests/test_go.py::test_go_dumping_x86
```

Trace:
```
test_go_dumping_x64                                                    FAILED 1.20s

pwndbg: loaded 187 pwndbg commands and 41 shell commands. Type pwndbg [--shell | --all] [filter] for a list.
pwndbg: created $rebase, $base, $hex2ptr, $argv, $envp, $argc, $environ, $bn_sym, $bn_var, $bn_eval, $ida GDB functions (can be used with print/break)
_____________________________ test_go_dumping_x86 ______________________________

start_binary = <function start_binary.<locals>._start_binary at 0x7f743e51dc60>

    def test_go_dumping_x86(start_binary):
>       helper_test_dump(start_binary, GOSAMPLE_X86)

start_binary = <function start_binary.<locals>._start_binary at 0x7f743e51dc60>

gdb-tests/tests/test_go.py:56: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

start_binary = <function start_binary.<locals>._start_binary at 0x7f743e51dc60>
filename = '/pwndbg/tests/gdb-tests/tests/binaries/gosample.x86'

    def helper_test_dump(start_binary, filename):
        gdb.execute("set environment GOMAXPROCS=1")
        start_binary(filename)
        gdb.execute("break gosample.go:6", to_string=True)
        gdb.execute("continue")
>       first = gdb.execute("go-dump any &x", to_string=True)
E       gdb.error: Error occurred in Python: Cannot access memory at address 0x1

filename   = '/pwndbg/tests/gdb-tests/tests/binaries/gosample.x86'
[...]

PR #3104, opened 2025-06-11
Add mallocng-explain command

k4lizen:
I named the category `musl` so it fits other musl-specific but not allocator-specific commands. It would feel a bit silly to me to have a seperate category for the two.

image
image
image

The slots are a bit complex and I'm not yet 100% sure on everything about them so I'll add that explanation when I feel more confident about it.


k4lizen:
Whats up with the tests :S

patryk4815:
> Whats up with the tests :S

probably ubuntu just upgraded some package (golang), containers are not reproducible

patryk4815:
@k4lizen https://github.com/actions/runner-images/issues/12176 ;)
> Go 1.21 will be removed and the latest version, 1.24, will be added and set as the default on 2025-06-13 for Windows and Ubuntu 22 & 24 images.
> Image deployment will start on 2025-06-09 and will take 2-3 days.

Our tests just don't work with go 1.24:
- https://github.com/pwndbg/pwndbg/issues/2796

PR #3106, opened 2025-06-12
Add command to display information about kernel modules

AC01010:
The `kmod` command displays information about loaded kernel modules. One optional argument, `module_name`, can filter modules by a substring of their names if provided.

image


k4lizen:
nitpicks:
Even though thats not the behaviour of `lsmod` it might be good to show the size in hex?
The columns also look a bit cramped to my eyes, it might be better to use (at least) two spaces instead of one

pwndbg/commands/kmod.py
```diff
@@ -0,0 +1,64 @@
...
+"""
+
+from __future__ import annotations
+
+import argparse
+
+import pwndbg.commands
+from pwndbg.aglib.kernel.macros import container_of
+
+parser = argparse.ArgumentParser(description="Displays information about loaded kernel modules.")
```
disconnect3d: Shall we say "Displays the loaded Linux kernel modules."?

pwndbg/commands/kmod.py
```diff
@@ -0,0 +1,64 @@
...
+parser.add_argument(
+    "module_name", nargs="?", type=str, help="A module name substring to search for"
```
disconnect3d: Search for, or, filter for? :)

pwndbg/commands/kmod.py
```diff
@@ -0,0 +1,64 @@
...
+    try:
+        print(f"{'Address':<18} {'Name':<19} {'Size':>8} {'Used by':<8}")
+        head = pwndbg.aglib.memory.get_typed_pointer_value("struct list_head", modules_head)
+        current = int(head["next"])
+
+        # Iterate through the linked list of modules
+        while current != modules_head:
+            module = container_of(int(current), "struct module", "list")
+
+            addr = hex(int(module["mem"][0]["base"]))
```
disconnect3d: Do we need hex()? Can we just do #x formatting with f'' during print?

pwndbg/commands/kmod.py
```diff
@@ -0,0 +1,64 @@
...
+    if modules_head is None:
+        print(
+            "The modules symbol was not found. This may indicate that the symbol is not available in the current build."
+        )
+        return
+
+    print(f"Kernel modules address found at {hex(modules_head)}.\n")
+
+    try:
+        print(f"{'Address':<18} {'Name':<19} {'Size':>8} {'Used by':<8}")
```
disconnect3d: Fwiw idk but maybe we should use `tabulate(table, headers)` to print this? Pls see how it is used within the project and see if it fits, if not, then its fine as it is I guess

pwndbg/commands/kmod.py
```diff
@@ -0,0 +1,64 @@
...
+        head = pwndbg.aglib.memory.get_typed_pointer_value("struct list_head", modules_head)
+        current = int(head["next"])
+
+        # Iterate through the linked list of modules
+        while current != modules_head:
+            module = container_of(int(current), "struct module", "list")
```
disconnect3d: Can't we do:
```py
for module in for_each_entry(head, "struct module", "list"):
    ...
```

Or something like this? Does the `for_each_entry` not fit here?

pwndbg/commands/kmod.py
```diff
@@ -0,0 +1,64 @@
...
+
+            addr = hex(int(module["mem"][0]["base"]))
+            name = pwndbg.aglib.memory.string(int(module["name"].address)).decode(
+                "utf-8", errors="ignore"
+            )
+            size = sum(int(module["mem"][i]["size"]) for i in range(4))
```
disconnect3d, line 50: Why do we have this for i in range thing? May be worth adding a comment explaining it

tests/library/qemu-system/tests/test_commands_kernel.py
```diff
@@ -48,6 +48,16 @@ def test_command_kdmesg():
...
 
 
+def test_command_kmod():
+    if not pwndbg.aglib.kernel.has_debug_syms():
+        res = gdb.execute("kmod", to_string=True)
+        assert "may only be run when debugging a Linux kernel with debug" in res
+        return
+
+    res = gdb.execute("kmod", to_string=True)
+    assert "Kernel modules address found at" in res or "The modules symbol was not found." in res
```
disconnect3d, line 58: Do we have any modules that are always there? Can we assert for them and at their format?

pwndbg/commands/kmod.py
AC01010: image

Looks a bit better, thank you for the suggestion!

pwndbg/commands/kmod.py
AC01010: Huh, didn't know this was a thing, good to know!

pwndbg/commands/kmod.py
AC01010: See the enum [here](https://github.com/torvalds/linux/blob/master/include/linux/module.h#L335C1-L346C3). My understanding is that the size should only be the size of the memory allocated after load, not sections freed after init

tests/library/qemu-system/tests/test_commands_kernel.py
AC01010: Unfortunately not, unless it is loaded manually

pwndbg/commands/kmod.py
disconnect3d: Please add a comment about it

disconnect3d (approved):

Issue #171, opened 2017-03-01
Save and Diff commands


zachriggle:
Two new commands should capture the output of a command (like "tee") and compare it to the current output (like "diff").

One might, for example, do something like "save hexdump esp" and later "diff hexdump esp".

"diff" by itself should diff the last command which was saved.
"save" by itself should save the output of the last command (might be doable via checking the command history)

peace-maker:
@disconnect3d

disconnect3d:
ping: @peace-maker does context history save command outputs?

peace-maker:
No, just context section output.

k4lizen:
is it bad manners to be like "gef already has this, we could probably take inspiration from his implementation"?

<img width="633" height="664" alt="Image" src="https://github.com/user-attachments/assets/cc619505-92c5-49d7-9e61-a5a9c218bf4e" />

happens quite often xd

k4lizen:
I wish that `pwndbg> saveo` or w/e saved the output of the previous command, this could be done by permanently keeping the output of the previous command in a buffer, possibly with a max size - and if you reach the max size and run `pwndbg> saveo` you get a hint saying "Previous output cropped, you may use `pwndbg> saveo -f heap` to get the full output" or w/e

k4lizen:
> is it bad manners to be like "gef already has this, we could probably take inspiration from his implementation"?

I would also use the same command name so its easier for people to switch between the two

PR #3111, opened 2025-06-17
Save diff commands

marywinter3:
The saveoutput command allows saving the output of any GDB command, either explicitly or by reusing the last executed one. The diffoutput command compares the current output of a command to its previously saved version and shows a diff if differences are found. The tests cover correct argument handling, saved output integrity, and edge cases like missing commands or outputs.


peace-maker:
Closes #171

pwndbg/commands/diffoutput.py
```diff
@@ -0,0 +1,41 @@
...
+        if not last_command:
+            print(message.error("No previous command to diff."))
+            return
+        cmd = last_command
+    if cmd not in saved_outputs:
+        print(message.error(f"No saved output for command: '{cmd}'"))
+        return
+        
+    try:
+        current = gdb.execute(cmd, to_string=True)
```
disconnect3d: Since you use gdb directly here, this command must be conditionally imported in `__init__.py` to be used only within gdb in the `if pwndbg.dbg.is_gdblib_available():` there.

Or: maybe we have some aglib way to execute cmd under any debugger?
(Note: we do support LLDB too, so it would be good if this command worked in lldb as well)


We have sth like `pwndbg.dbg.selected_inferior().evaluate_expression` but I don't see any API to run commands hm

disconnect3d:
@marywinter3 this needs an update, see CI builds failing

pwndbg/commands/diffoutput.py
k4lizen: Process.runcmd

pwndbg/commands/saveoutput.py
```diff
@@ -0,0 +1,44 @@
...
+        if not last_command:
+            print(message.error("No previous command to save."))
+            return
+
+        cmd = last_command
+    try:
+        result = gdb.execute(cmd, to_string=True)
+        saved_outputs[cmd] = result
+        last_command = cmd
+        print(message.success(f"Output saved for command: '{cmd}'"))
```
k4lizen: it would probably also make sense to print the output of the command

tests/library/gdb/tests/test_command_saveoutput.py
```diff
@@ -0,0 +1,67 @@
...
+def test_saveoutput_uses_last_command_when_no_args(start_binary):
+    pwndbg.commands.saveoutput.saved_outputs.clear()
+    pwndbg.commands.saveoutput.last_command = None
+
+    start_binary(REFERENCE_BINARY)
+    gdb.execute("break main")
+    gdb.execute("run")
+
+    cmd = "info registers"
+    pwndbg.commands.saveoutput.saveoutput(["info", "registers"])
```
k4lizen, line 53: why run the command function instead of running the command inside the debbuger?

k4lizen (changes requested):
you need to port the command and tests to be debugger-agnostic. (for tests see `tests/library/dbg/`).

disconnect3d:
Since this is very stale, I am going to close this for now. Please reopen when ready

marywinter3:
Please reopen we have new changes @disconnect3d

disconnect3d:
@marywinter3 pls commit & push them?

marywinter3:
In our fork this test also passed after retrigger @disconnect3d 
Please run tests again I think this is ready to merge

pwndbg/commands/diffoutput.py
```diff
@@ -0,0 +1,60 @@
...
+import difflib
+
+import pwndbg.color.message as message
+import pwndbg.commands
+import pwndbg.commands.saveoutput as saveoutput_mod
+import pwndbg.dbg_mod
+from pwndbg.commands import CommandCategory
+
+saved_outputs = saveoutput_mod.saved_outputs
+last_command: str | None = None
```
peace-maker, line 13: This variable is never written to (outside of tests). Is it really required?

pwndbg/commands/saveoutput.py
```diff
@@ -0,0 +1,76 @@
...
+
+def run_debugger_command(cmd: str) -> str:
+    proc = pwndbg.dbg.selected_inferior()
+    if proc is None:
+        raise pwndbg.dbg_mod.Error("No selected inferior process.")
+
+    stdout_buffer = io.StringIO()
+    stderr_buffer = io.StringIO()
+
+    with contextlib.redirect_stdout(stdout_buffer), contextlib.redirect_stderr(stderr_buffer):
```
peace-maker, line 44: Does this work when [redirecting the `context` output](https://pwndbg.re/stable/commands/context/contextoutput/) to something else than stdout?

pwndbg/commands/saveoutput.py
```diff
@@ -0,0 +1,76 @@
...
+
+save_parser.add_argument(
+    "args",
+    nargs=argparse.REMAINDER,
+    type=str,
+    help="Command to run and save.",
+)
+
+
+def get_command_to_save(args: list[str]) -> str | None:
```
peace-maker, line 28: Can we grab the last command from the command history here? `show commands`
https://sourceware.org/gdb/current/onlinedocs/gdb.html/Command-History.html

I'm sure lldb has something similar too.

That way you can just run some command, then do `saveoutput`, and later run it again. `saveoutput` should still remember the command it saved, so `diffoutput` works without typing the command again. But you wouldn't have to type `saveoutput the actual command`.

pwndbg/commands/saveoutput.py
```diff
@@ -0,0 +1,76 @@
...
+import pwndbg.color.message as message
+import pwndbg.commands
+import pwndbg.dbg_mod
+from pwndbg.commands import CommandCategory
+
+saved_outputs: dict[str, str] = {}
+last_command: str | None = None
+
+save_parser = argparse.ArgumentParser(
+    description="Save the output of a debugger command for later diffing."
```
peace-maker, line 17: Can you add more details here, that the command is run again instead of saving the previously printed output?

pwndbg/commands/diffoutput.py
```diff
@@ -0,0 +1,60 @@
...
+
+diff_parser = argparse.ArgumentParser(
+    description="Diff current output of a debugger command against a saved snapshot."
```
peace-maker, line 16: There is documentation missing for the behavior of which command is executed when you don't specify any arguments.

pwndbg/commands/diffoutput.py
```diff
@@ -0,0 +1,60 @@
...
+        saved.splitlines(),
+        current.splitlines(),
+        fromfile="saved",
+        tofile="current",
+        lineterm="",
+    )
+    diff_text = "\n".join(diff_iter)
+
+    if diff_text:
+        print(message.notice("Differences:\n" + diff_text))
```
peace-maker, line 58: Can you post some screenshots of how this looks please? Does it work well with terminal coloring or are colors disabled while capturing to StringIO?

peace-maker (changes requested):
This has stalled for a while. I'm no authority at all but can you provide some screenshots and gdb sessions where you use this command in different ways, so one can see how you envisioned people to use it please?

The tests should be from the user's perspective instead of calling the command handlers through code I think as @k4lizen pointed out. Are both tests required? Maybe you can find some pwndbg commands to diff that have the same output in gdb and lldb?

disconnect3d:
Closing this for now. Please re-open another PR if you decide to work on this further and then start by resolving the comments from here

PR #3113, opened 2025-06-18
mallocng: add ng-find command

k4lizen:
image


k4lizen:
This PR allows us to look at and traverse actual allocator state. Up to now we only looked at local data.

pwndbg/aglib/heap/mallocng.py
```diff
@@ -303,6 +324,36 @@ def internal_offset(self) -> int:
         # We can calculate it more easily than musl does:
         return (self.p - self.start) // UNIT
 
+    def contains_group(self) -> bool:
+        """
+        Does this slot nest a group?
+        """
+        return self.reserved == 6
```
disconnect3d, line 333: Why is it '6'? can we comment on that?

pwndbg/aglib/heap/mallocng.py
```diff
@@ -527,11 +581,465 @@ def slot_size(self):
...
+
     def __init__(self, addr: int) -> None:
-        self.addr = addr
+        self.addr: int = addr
+
+        self.check: int = 0
+        self.meta_area: int = 0
+        self.nslots: int = 0
+        # Alignment offsets it by 0x4
+        self.slots: int = self.addr + 0x18
```
disconnect3d: Can we comment on why 0x18?

pwndbg/aglib/heap/mallocng.py
```diff
@@ -527,11 +581,465 @@ def slot_size(self):
...
+
+        self.load()
+
+    def load(self):
+        ptrsize = pwndbg.aglib.arch.ptrsize
+        uint64size = pwndbg.aglib.typeinfo.uint64.sizeof
+        endian = pwndbg.aglib.arch.endian
+
+        data: bytearray = memory.read(self.addr, 0x14)
```
disconnect3d: Can we comment on why 0x14?

pwndbg/aglib/heap/mallocng.py
```diff
@@ -527,11 +581,465 @@ def slot_size(self):
...
+        cur_offset += int_size()
+
+        assert cur_offset == 0x14
+
+    def at_index(self, idx: int) -> int:
+        """
+        Returns the address of the meta object located
+        at index idx.
+        """
+        return self.addr + 0x18 + idx * Meta.sizeof()
```
disconnect3d: That's at least 2nd time we use 0x18 - shall we move it to a constant?

pwndbg/aglib/heap/mallocng.py
```diff
@@ -527,11 +581,465 @@ def slot_size(self):
...
+        cur_offset += ptrsize
+        self.avail_meta_count = int.from_bytes(
+            data[cur_offset : (cur_offset + size_tsize)], endian, signed=False
+        )
+        cur_offset += size_tsize
+        self.avail_meta_area_count = int.from_bytes(
+            data[cur_offset : (cur_offset + size_tsize)], endian, signed=False
+        )
+        cur_offset += size_tsize
+        self.meta_alloc_shift = int.from_bytes(
```
disconnect3d: I wonder if we could do sth like:
```py
next_data = lambda: blabla # or def?

self.sth = next_data(size)
self.sth2 = next_data(size2)
```

pwndbg/aglib/heap/mallocng.py
```diff
@@ -527,11 +581,465 @@ def slot_size(self):
...
+        # No debug information :(
+        self.has_debug_syms = False
+
+        # We will find the __malloc_context object by searching memory for
+        # the secret.
+        # https://elixir.bootlin.com/musl/v1.2.5/source/src/malloc/mallocng/malloc.c#L50
+        # Extract the secret first.
+        # https://elixir.bootlin.com/musl/v1.2.5/source/src/malloc/mallocng/glue.h#L49
+        at_random = int(pwndbg.auxv.get()["AT_RANDOM"])
+        self.secret = memory.read_pointer_width(at_random + 8)
```
disconnect3d: Shouldn't this be `memory.read(at_random+8, ptrsize)` so that we won't have to do `.pack` later on?

Or is there any benefit for us to have it with the type returned by `read_pointer_width` saved into `self.secret`?

pwndbg/aglib/heap/mallocng.py
```diff
@@ -527,11 +581,465 @@ def slot_size(self):
...
+
+        # There are going to be multiple matches. We don't
+        # want those on the stack (actual AT_RANDOM) or heap
+        # (structures copying the secret). We want it either from the libc.so
+        # mapping (if musl is dynamically linked) or the executable's
+        # mapping (if musl is statically linked).
+        possible: List[Tuple[int, str]] = []
+        for sm in secret_matches:
+            mapping_name = pwndbg.aglib.vmmap.find(sm).objfile
+            if "[stack" in mapping_name or "[heap" in mapping_name:
```
disconnect3d: Other threads stacks may not be `[stack` iirc and we could in theory skip them?

Iirc we had sth like `stacks.get()` or sth

pwndbg/aglib/heap/mallocng.py
CptGibbon: I don't know what these hardcoded values represent, but are they valid for architectures that use different width pointers?

pwndbg/aglib/heap/mallocng.py
k4lizen: > the type returned by [...]
`def read_pointer_width(addr: int) -> int:`

but yeah fair enough

k4lizen:
i'll look to update the other size constants tomorrow

pwndbg/aglib/heap/mallocng.py
k4lizen: i thought about, but then thought about how something like that should exist in pwndbg.lib / .aglib, couldn't think of a good API, and ended up doing nothing about it XD

i'll update it to something like this:
```python
def read_uint(size: int) -> int:
    nonlocal cur_offset
    val = int.from_bytes(data[cur_offset : cur_offset + size], endian, signed=False)
    cur_offset += size
    return val
```

pwndbg/aglib/heap/mallocng.py
```diff
@@ -864,7 +864,7 @@ def set_ctx_addr(self):
         self.ctx_addr = pwndbg.aglib.symbol.lookup_symbol_addr("__malloc_context")
         if self.ctx_addr is not None:
             self.has_debug_syms = True
-            self.secret = memory.read_pointer_width(self.ctx_addr)
+            self.secret = memory.read(self.ctx_addr, 8)
```
disconnect3d: Is 8 okay? What if we run 32-bit musl? Shouldnt this be pointer width?

pwndbg/aglib/heap/mallocng.py
```diff
@@ -889,9 +890,14 @@ def set_ctx_addr(self):
         # mapping (if musl is dynamically linked) or the executable's
         # mapping (if musl is statically linked).
         possible: List[Tuple[int, str]] = []
+        thread_stacks = pwndbg.aglib.stack.get()
+
         for sm in secret_matches:
+            if any(sm in stack_page for stack_page in thread_stacks.value()):
```
disconnect3d: I think this is a bug? Should be .values()? but maybe also it should be in line 893 so we don't call values() multiple times?

disconnect3d (changes requested):
See comment - potential bug with dict .value[s]

pwndbg/aglib/heap/mallocng.py
k4lizen: wau yeah you're right, crazy my LSP (and mypy) doesn't catch that....

does your editor catch it?

pwndbg/aglib/heap/mallocng.py
k4lizen: it's actually hardcoded to be uint64_t in the source https://elixir.bootlin.com/musl/v1.2.5/source/src/malloc/mallocng/glue.h#L49

k4lizen:
should be good now :>

pwndbg/aglib/heap/mallocng.py
```diff
@@ -352,32 +405,32 @@ def preload(self) -> None:
...
-        cur_offset += ptrsize
-        self._next = pwndbg.aglib.arch.unpack(data[cur_offset : (cur_offset + ptrsize)])
-        cur_offset += ptrsize
-        self._mem = pwndbg.aglib.arch.unpack(data[cur_offset : (cur_offset + ptrsize)])
-        cur_offset += ptrsize
-        self._avail_mask = int.from_bytes(
-            data[cur_offset : (cur_offset + int_size())], endian, signed=False
-        )
-        cur_offset += int_size()
-        self._freed_mask = int.from_bytes(
```
disconnect3d, line 368: Honestly, it would be better if we fetched `int_size` once here as:
```py
int_size = pwndbg.aglib.typeinfo.sint.sizeof
```

Generally, the function is not great, heh

disconnect3d (approved):

pwndbg/aglib/heap/mallocng.py
k4lizen: > Generally, the function is not great, heh

I agree it's a bit wack but it's used in various classes in various functions (e.g. Meta stuff).

PR #3117, opened 2025-06-18
Reland "mallocng: print information about slot, group, meta at given address"

k4lizen:
See https://github.com/pwndbg/pwndbg/pull/3087 and https://github.com/pwndbg/pwndbg/pull/3115 .

This fixes the performance issues, but I'm not sure why `signed int` would be missing from kernel in the first place.


k4lizen:
@patryk4815 is this still to slow or does it look good?

PR #3119, opened 2025-06-19
Use gdb/lldb from pypi

patryk4815:
What is missing?
- after this PR, I will create fixes for portable release (it would be too much for one PR)
- maybe more docs changes is needed?

What is changes?
- **New module**: Introduced a new top-level module `pwndbginit` to encapsulate debugger initialization logic.
- **`gdbinit.py`**: Now requires a `.venv` directory to be present.  
  For regular users, nothing changes — you can still add `gdbinit.py` directly to your `~/.gdbinit`  

  This impacts only **distribution packages** (e.g. **Arch Linux** and **Gentoo**),  
  which should **no longer invoke `gdbinit.py` directly**, but instead use the script entry point defined in `pyproject.toml`:
  
  ```toml
  [project.scripts]
  pwndbg = "pwndbginit.pwndbg_gdb:main"
  pwndbg-lldb = "pwndbginit.pwndbg_lldb:main"
  ```
  This ensures proper virtual environment detection and patch setup.

- Replaced `.skip-venv` placeholder with `.pwndbg_root` empty file
- Refactored `lldbinit.py` and `gdbinit.py` to share common logic and patch management.

How can it be used:
```
uvx --from 'git+https://github.com/patryk4815/pwndbg@pypi[gdb]' pwndbg
uvx --from 'git+https://github.com/patryk4815/pwndbg@pypi[lldb]' pwndbg-lldb

uvx --from './[gdb]' pwndbg
uvx --from './[lldb]' pwndbg-lldb

uv run --extra gdb pwndbg
uv run --extra lldb pwndbg-lldb
```

Via venv:
```
source ./.venv/bin/activate
uv sync --extra gdb --extra lldb

pwndbg
pwndbg-lldb
```


k4lizen:
>  For regular users, nothing changes — you can still add gdbinit.py directly to your ~/.gdbinit

will the system installed gdb then be used instead of the venv one?

patryk4815:
> > For regular users, nothing changes — you can still add gdbinit.py directly to your ~/.gdbinit
> 
> will the system installed gdb then be used instead of the venv one?

Just to clarify:
- `source /path/to/pwndbg-git/gdbinit.py` will still work as before
- `pwndbginit/pwndbg_gdb.py` (pwndbg binary) - will work fine with `venv gdb` or `system gdb`

If you want use system gdb, then you need use system python

disconnect3d (approved):

PR #3123, opened 2025-06-21
Adding custom address markers

jxuanli:
- introduced a new `ArchPagingInfo` class to simplify accesses to important marker address for areas such as `physmap`, `vmalloc`, and `vmemmap`
- removed the reliance on kernel configurations to compute marker addresses for Aarch64 by reading the `TCR_EL1` register
- `buddydump` is now supported for debugging Aarch64 kernels as a side effect


jxuanli:
ready for review

pwndbg/aglib/kernel/__init__.py
```diff
@@ -266,10 +264,33 @@ def pfn_to_page(self, phys: int) -> int:
     def page_to_pfn(self, page: int) -> int:
         raise NotImplementedError()
 
+    @property
+    def STRUCT_PAGE_SIZE(self):
+        a = pwndbg.aglib.typeinfo.load("struct page")
+        if a is None:
+            return 0x40
```
disconnect3d: Why 0x40? Can we add a comment?

disconnect3d: Fwiw each call to this will re-load the type info; shouldn't we maybe cache this in a field?

jxuanli: added a comment and caching

PR #3132, opened 2025-06-23
mallocng: Explain what a slot looks like in ng-explain

k4lizen:
image
image


k4lizen:
I wanted to give a visual representation also but tbh it's kinda hard, and the `ng-slotu` command essentially does just that.

PR #3135, opened 2025-06-24
Improving the slab commands

jxuanli:
Improving/fixing the handling and display of `slab info` and `slab contains`

## demo
image

### DIFF
--- a/pwndbg/aglib/kernel/slab.py
+++ b/pwndbg/aglib/kernel/slab.py
@@ -172,6 +172,10 @@ class SlabCache:
         return int(self._slab_cache["size"])
 
     @property
+    def slab_size(self) -> int:
+        return 0x1000 << self.oo_order
+
+    @property
     def object_size(self) -> int:
         return int(self._slab_cache["object_size"])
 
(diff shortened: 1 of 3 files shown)


k4lizen:
awesome!

Could the `-v` option be the default? If performance is the problem maybe a config could control it?

jxuanli:
@k4lizen i removed the verbose option so it is now default -- it introduces a noticeable delay if run on its own but the handling code is not a common path. Also, the net delay averages out when `slab` commands are repeatedly invoked (possibly due to some implicit caching)

k4lizen:
nice

k4lizen (approved):

pwndbg/aglib/kernel/slab.py
```diff
@@ -91,7 +91,15 @@ def __iter__(self) -> Generator[int, None, None]:
         seen: set[int] = set()
         current_object = self.start_addr
         while current_object:
-            addr = int(current_object)
+            try:
+                addr = int(current_object)
+            except Exception:
+                print(
+                    M.warn(
+                        f"Corrupted slab freelist detected at {hex(current_object)} when length is {len(seen)}"
```
disconnect3d, line 99:
```suggestion
                        f"Corrupted slab freelist detected at {current_object:#x} when length is {len(seen)}"
```

PR #3139, opened 2025-06-26
Drop `PWNDBG_PLEASE_SKIP_VENV` option

patryk4815:
The `PWNDBG_PLEASE_SKIP_VENV` option is no longer needed — it was just a workaround for handling Nix in previous releases.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -25,7 +25,6 @@ jobs:
     timeout-minutes: 40
     env:
       TMPDIR: /tmp
-      PWNDBG_VENV_PATH: PWNDBG_PLEASE_SKIP_VENV
     steps:
     - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9  # @v3
     - uses: cachix/install-nix-action@08dcb3a5e62fa31e2da3d490afc4176ef55ecd72  # @v30
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -25,6 +25,7 @@ jobs:
     timeout-minutes: 40
     env:
       TMPDIR: /tmp
+      PWNDBG_NO_UV: 1
     steps:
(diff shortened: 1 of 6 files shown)


k4lizen:
it makes sense to me that there is no need for both that and `PWNDBG_NO_UV`

k4lizen:
btw could you explain what `.pwndbg_root` is for? this used to be `.skip-venv` right? Is it different semantically now or does it mark the same thing?

Also `nix/portable.nix` still mentions it, is that alright?
```
80:  skipVenv = pkgs.writeScript "pwndbg-skip-venv" "";
116:      "exe/.skip-venv"
157:      "exe/.skip-venv"
```

patryk4815:
`portable.nix` will be handled in different PR, there are more things to fix

patryk4815:
> it makes sense to me that there is no need for both that and `PWNDBG_NO_UV`

yes, later we can focus on PWNDBG_NO_UV

patryk4815:
> btw could you explain what `.pwndbg_root` is for?

```
     # If pwndbg is installed in `/venv/lib/pythonX.Y/site-packages/pwndbg/`,
     # the `.pwndbg_root` file will not exist because `src_root` will point to the
     # `/venv/lib/pythonX.Y/site-packages/` directory, not the original source directory
     #
     # However, if pwndbg is installed in editable mode (our recommended way), this file will exist,
     # and the condition will be False, allowing auto-update.
```
^ this should explain it. It is just for checking if `pwndbg` is copied into venv or installed using editable mode
If it is coped then we shouldn't preform auto-update - because venv is fully locked / encapsulated. 
Auto update should only happen for editable mode, when you `git clone` and after sometime did `git pull`

k4lizen:
Ahhh I see. I somehow missed that `.pwndbg_root` is in the repository

Feels a bit weird to have a variable called `src_root` not point to the root folder of the source but aight

PR #3140, opened 2025-06-26
autoupdate: Add support for upgrading editable tool

patryk4815:

### DIFF
--- a/pwndbginit/common.py
+++ b/pwndbginit/common.py
@@ -23,13 +23,20 @@ def hash_file(file_path: str | Path) -> str:
 
 
 def run_uv_install(
-    binary_path: os.PathLike[str], src_root: Path, dev: bool = False
+    binary_path: os.PathLike[str], src_root: Path, venv_path: Path, dev: bool = False
 ) -> Tuple[str, str, int]:
-    # We don't want to quietly uninstall dependencies by just specifying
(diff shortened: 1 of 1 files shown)


pwndbginit/common.py
```diff
@@ -23,13 +23,20 @@ def hash_file(file_path: str | Path) -> str:
...
-    # `--extra gdb` so we will be conservative and pull all extras in.
-    command: List[str] = [str(binary_path), "sync", "--all-extras"]
-    if dev:
-        command.append("--all-groups")
+    # Check if the package was installed using: `uv tool install --editable .[lldb,gdb]`
+    # Tools are located at: ${HOME}/.local/share/uv/tools/${TOOL_NAME}/uv-receipt.toml
+    is_tool_install = (venv_path / "uv-receipt.toml").exists()
+    if is_tool_install:
+        tool_name = venv_path.name
+        command: List[str] = [str(binary_path), "tool", "upgrade", tool_name]
```
k4lizen, line 33: Are you sure `tool upgrade` is useful since we install it with `--editable`? Are you expecting `uv tool upgrade ...` to update pwndbg dependancies? Does it actually do that?

patryk4815: yes, it actually updating dependencies

PR #3144, opened 2025-07-01
mallocng: add ng-slots command

k4lizen:
Dump information about a mallocng slot, given its start address. The previous command took user address, sometimes start address is more convenient.


disconnect3d (approved):

disconnect3d:
Btw the `read_success` var looks a bit tricky.. the code also have some nested `try: try:` blocks. Would be nice if we could simplify some of these, but idk if its possible

PR #3116, opened 2025-06-18
Supporting kernel debugging in the absence of provided type information

jxuanli:
Several kernel debugging commands rely on type information to access fields in a kernel structure. The goal of this PR is to support some of those commands when `kallsyms` is present but type information is not provided. Relevant kernel configurations are now automatically detected even if `CONFIG_IKCONFIG` is not set. Therefore, `kconfig()` now always returns an object. Function and test decorators have been changed accordingly.

See https://jxuanli.github.io/posts/linux-kernel-structure-layouts-partial-recovery/ for more details.

PR #3145, opened 2025-07-02
Improving `kconfig`

jxuanli:
- removed `requires_kconfig` function decorator
- added automatic detection for specific kernel configurations needed for #3116 
- added option to load and parse a kernel configuration file

## demo

image


pwndbg/aglib/kernel/__init__.py
```diff
@@ -130,36 +110,28 @@ def get_first_kernel_ro() -> pwndbg.lib.memory.Page | None:
...
 
-        if len(results) == 0:
-            return None
-
-        config_start = results[0] + len("IKCFG_ST")
-        config_end = list(pwndbg.search.search(b"IKCFG_ED", start=config_start))[0]
-
+        if len(results) != 0:
+            config_start = results[0] + len("IKCFG_ST")
+            config_end = list(pwndbg.search.search(b"IKCFG_ED", start=config_start))[0]
```
disconnect3d: Both this and `results` should probably do:

`next(pwndbg.search.search(...), None)` and instead of checking `len(..)` then check it as `is not None`

Doing it this way, we won't search all memory for those values as well as we won't render a Python list for the returned results.

pwndbg/lib/kernel/kconfig.py
```diff
@@ -59,3 +83,70 @@ def __contains__(self, name: object) -> bool:
...
+        return pwndbg.aglib.symbol.lookup_symbol("__kasan_cache_create") is not None
+
+    @property
+    def CONFIG_SMP(self) -> bool:
+        return pwndbg.aglib.symbol.lookup_symbol("pcpu_get_vm_areas") is not None
+
+    def update_with_file(self, file_path):
+        for line in open(file_path, "r").read().splitlines():
+            split = line.split("=")
+            if len(line) == 0 or line.startswith("#") or len(split) != 2:
```
disconnect3d:
```suggestion
            if len(line) == 0 or line[0] == "#" or len(split) != 2:
```

Doesn't matter much but yeah

pwndbg/aglib/kernel/__init__.py
jxuanli: changed

pwndbg/aglib/kernel/__init__.py
```diff
@@ -130,36 +110,28 @@ def get_first_kernel_ro() -> pwndbg.lib.memory.Page | None:
...
-        config_end = list(pwndbg.search.search(b"IKCFG_ED", start=config_start))[0]
+        result = next(pwndbg.search.search(b"IKCFG_ST", mappings=[mapping]), None)
 
+        if result is not None:
+            config_start = result + len("IKCFG_ST")
+            config_end = next(pwndbg.search.search(b"IKCFG_ED", start=config_start), None)
     if config_start is None or config_end is None:
-        return None
+        _kconfig = pwndbg.lib.kernel.kconfig.Kconfig(None)
+        return _kconfig
```
disconnect3d, line 129: Why not just `return KConfig(None)`?

disconnect3d (approved):

PR #3147, opened 2025-07-06
Fix: Make color functions respect PWNDBG_DISABLE_COLORS  #3142

### LINKED ISSUE #3142: many commands don't respect `NO_COLOR=1`
### Description

Simply, many commands don't respect `PWNDBG_DISABLE_COLORS=1`. I realized mallocng commands don't acknowledge this but many others don't as well.

Image
Image

Pretty much if anything uses these functions
https://github.com/pwndbg/pwndbg/blob/567ac6048cf345f876e3387fcfc08d515a185401/pwndbg/color/__init__.py#L52-L65
...

That by default means it won't respect `PWNDBG_DISABLE_COLORS`. I guess the "intended" way is to run them through this
https://github.com/pwndbg/pwndbg/blob/567ac6048cf345f876e3387fcfc08d515a185401/pwndbg/color/__init__.py#L185-L201

See for example, `vis` code:
https://github.com/pwndbg/pwndbg/blob/567ac6048cf345f876e3387fcfc08d515a185401/pwndbg/commands/ptmalloc2.py#L1068-L1078

But imo this is very non-obvious and counterintuitive. That being said, if we add the `PWNDBG_DISABLE_COLORS` check to all the coloring functions, that may affect performance - not sure.

### Repro
```bash
PWNDBG_DISABLE_COLORS=1 gdb /bin/sh
pwndbg> pwndbg 
```

### Versions
```
Pwndbg:   2025.05.30 build: f3e333031
Python:   3.13.3
GDB:      16.3
Capstone: 6.0.0
Unicorn:  2.1.3
Pwnlib:   4.14.0
```

JoaoBeernardoo:
This PR updates the color functions (e.g., red, green) to check for disable_colors so that they properly honor the PWNDBG_DISABLE_COLORS environment variable. Fixes #3142.

### DIFF
--- a/pwndbg/color/__init__.py
+++ b/pwndbg/color/__init__.py
@@ -10,7 +10,6 @@ from typing import Callable
 from typing import Dict
 from typing import List
 from typing import NamedTuple
-from pwndbg.color import disable_colors
 from pwndbg.lib.config import Parameter
 
 from . import theme
--- a/pwndbg/color/__init__.py
+++ b/pwndbg/color/__init__.py
@@ -10,7 +10,7 @@ from typing import Callable
 from typing import Dict
 from typing import List
 from typing import NamedTuple
-
+from pwndbg.color import disable_colors
 from pwndbg.lib.config import Parameter
 
 from . import theme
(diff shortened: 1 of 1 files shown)


disconnect3d (changes requested):
Hi, first of all, if the `x` argument passed to the color function is already a `str` we can then just return it instead of doing a `return str(x)`.

I also think that adding a branch to each color function is not great. While this may not seem important I think we do not want to make the "happy path" (aka: when colors are enabled) any slower.

Instead, we should probably change either the `colorize` function used under the hood to just return the first argument, or, swap/patch each our color function to do this by itself.

k4lizen:
> Instead, we should probably change either the colorize function used under the hood to just return the first argument, or, swap/patch each our color function to do this by itself.

you mean like a `pwndbg.color.green = nothing` or `pwndbg.color.green = green_impl` patch that is activated on load and when the config changes?

disconnect3d:
> > Instead, we should probably change either the colorize function used under the hood to just return the first argument, or, swap/patch each our color function to do this by itself.
> 
> you mean like a `pwndbg.color.green = nothing` or `pwndbg.color.green = green_impl` patch that is activated on load and when the config changes?

As long as we always call it via `color.green(..)` and not `green(..)` then yeah.

If that's not true - we should probably either change calls or come up with something more magic here.

disconnect3d:
Ping @JoaoBeernardoo

disconnect3d:
Closing due to inactivity

PR #3152, opened 2025-07-09
mallocng: implement inspection of non-allocated slots

k4lizen:
When you try to read out any slot, for freed and available slots you won't be able to reach the group and meta - thats how the allocator was designed. This PR makes it so that in those cases, we search the heap and try to find the slot while propagating the information about from which group and meta the slot is coming from.
+ Added config option that can turn of running this search automatically (if it gets slow for programs with lots of allocations).
+ Show slot allocation state in command outputs.
+ Print out as much information as possible about every slot (even completly corrupted works).
+ Moved two meta fields into `alt_values`s.
+ Got rid of set_padding in PropertyPrinter, replaced with sane defaults.

image
image
image
image
image

### DIFF
--- a/pwndbg/aglib/heap/mallocng.py
+++ b/pwndbg/aglib/heap/mallocng.py
@@ -204,12 +204,10 @@ class Slot:
         self._pn3 = pheader[5]
         # ==
 
-        # Read the group's meta pointer.
-        _ = self.meta
-
-        # To calculate footer and p header fields
+        # To calculate footer and start header fields
         # we need self.meta.stride. However we want to be able to
-        # return some information even if the meta is corrupt, so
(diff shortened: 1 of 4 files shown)


pwndbg/aglib/heap/mallocng.py
```diff
@@ -503,19 +552,51 @@ def from_start(cls, start: int) -> "Slot":
             obj = cls(p)
             obj._sn3 = sn3
         else:
+            # freed / avail slots will also go into this branch.
             p = start
             obj = cls(p)
```
disconnect3d, line 557: Fwiw the `p` variable could probably be just removed across this function

disconnect3d (approved):

PR #3153, opened 2025-07-10
Fix relative path to .zig directory in Makefiles

OBarronCS:
This PR fixes a relative path to the ZIGPATH environment variable in the Makefiles that compile programs for tests. A directory restructure in #3101 meant the paths were no longer accurate. This makes running `make` manually while in the test directory work.

The fixed assignment statement only takes effect when manually running "make" while in the directory - when running ./tests.sh, the $ZIGPATH environment variable is already set by the time it internally calls `make`, and the `ZIGPATH ?= ...` syntax means it won't override if the environment variable is already set.


k4lizen:
oh nice! I had this issue and was wondering what it was about, I didn't know about the `?=` syntax

Issue #3156, opened 2025-07-14
`$base` function shows traceback on non-existant mapping


k4lizen:
### Description

```
pwndbg> p $base("arstarst")
Traceback (most recent call last):
  File "/pwndbg/pwndbg/gdblib/functions.py", line 64, in invoke
    return self.func(*args)
           ~~~~~~~~~^^^^^^^
  File "/pwndbg/pwndbg/gdblib/functions.py", line 135, in base
    raise ValueError(f"No mapping named {name}")
ValueError: No mapping named arstarst
Error occurred in Python: No mapping named arstarst
```

I would expect something nicer looking like:
```
pwndbg> p $base("arstarst")
No mapping named "arstarst".
```

### Steps to reproduce

Try to `$base()` a non-existant mapping.

### My setup

Pwndbg:   2025.05.30 build: 958a83122
Python:   3.13.5
GDB:      16.3
Capstone: 6.0.0
Unicorn:  2.1.3
Pwnlib:   4.14.0

AalbatrossGuy:
Hey, I opened a PR following this issue. Check it out maybe?

disconnect3d:
Done in https://github.com/pwndbg/pwndbg/pull/3158. Big thanks to @AalbatrossGuy !

PR #3158, opened 2025-07-15
Fixed output message when unmatched mapping was encountered in the `base` function.

AalbatrossGuy:
Citing issue #3156, 
The function no longer shows traceback on non-existent mapping(s). It shows a simple error message.

image


pwndbg/gdblib/functions.py
AalbatrossGuy, line 1: removing `raise ValueError()` fixed the traceback output

k4lizen:
I do think thats better but I wonder if theres a way we can make the `p/x` fail cleanly instead of using `0` as the fall-back value.

AalbatrossGuy:
@k4lizen something like `raise gdb.GdbError()` could help fix that but then, it wouldn't show a _clean_ message any more. It'd contain something like 
```py
Exception in class gdb.error: No mapping named 'mylib': 
Error occurred in command: No mapping named 'mylib'

**Comment by disconnect3d** [comment]
> @k4lizen something like `raise gdb.GdbError()` could help fix that but then, it wouldn't show a _clean_ message any more. It'd contain something like
> 
> ```python
> Exception in class gdb.error: No mapping named 'mylib': 
> Error occurred in command: No mapping named 'mylib'
> ```

Does it really show both exception and an error? Doh, that's sad

**Comment by AalbatrossGuy** [comment]
@disconnect3d @k4lizen I think I found a simple way to solve that, 
Running this piece of code,
```py
a = [1, 2, 3]
try:
    print(a[4])
except Exception as err:
    print("Custom Index Error Message: ", err)
```
gives you a clean output like,
```py
Custom Index Error Message:  list index out of range
```
Are you guys trying to get a result similar to this?

k4lizen:
Yeah if the output was something like
```
pwndbg> p/x $base("mylib")
$base error: No mapping named "mylib".
pwndbg>
```
I think that'd be ideal.

Otherwise, I think what you did with reporting the message and returning zero is good. I would only add clarification that the error message is coming from `$base` to help figure out what is wrong in more complex expressions, and add to the function description that it returns zero if nothing is found.

AalbatrossGuy:
HM okay, got it!
I'll try to do something like that

AalbatrossGuy:
Hi, @k4lizen @disconnect3d, updated the output as you guys wanted. 
Screenshot_20250719_175342

k4lizen:
very nice! thanks

PR #3161, opened 2025-07-18
add relative imports into tests

patryk4815:
`qemu-user` and `qemu-system` dirs are invalid python imports, so they must be renamed with `_`


disconnect3d (approved):

mbrla0:
I think this should be alright.

mbrla0 (approved):

Issue #1965, opened 2023-12-20
Detect cyclic pattern in registers or pointers


peace-maker:
When using a cyclic pattern and stopping at some invalid memory access, you currently have to copy&paste the pattern from the registers and invoke `cyclic -l` for each of them. This is repetetive and slow if there are multiple controlled registers.

While looking through the IDA Sploiter plugin I stumbled upon the [Detecting a pattern](https://iphelix.medium.com/ida-sploiter-9604e0b8829#3c6a) section with a dialog showing an overview of every register and if it was able to match the pattern to some offset.
I'd like a `cyclic --detect` option to display all registers and, in case a register contains a pointer, memory locations that contain a cyclic pattern as well as its offset.

That same detection could be displayed in the `regs` / `context` output too.

disconnect3d:
cc: @mbrla0

armoredvortex:
Hello! I'm interested in working on this issue.
My plan is to add a --detect flag to the cyclic command that iterates through registers and memory to find pattern offsets.
Any initial pointers on where to start looking in the codebase would be much appreciated. Thanks!

k4lizen:
See `pwndbg/commands/cyclic.py` and https://pwndbg.re/pwndbg/dev/contributing/ :>

PR #3162, opened 2025-07-18
feat(cyclic): Add --detect flag in cyclic command

armoredvortex:
This PR enhances the `cyclic` command by adding a `--detect` flag.
As mentioned in #1965 .

 When `cyclic --detect` or `cyclic -d` is run, it performs a scan of all immediate values in registers or values stored at pointers for cyclic patterns and prints their offsets.
 
###  Example Output:
 
image


disconnect3d:
Hi, thanks for implementing it! Any chance you can add a testcase for this feature?:)

armoredvortex:
@disconnect3d 
> Hi, thanks for implementing it! Any chance you can add a testcase for this feature?:)

I wanted to ask for some guidance on testing.
What would be the best place for such a test?

Should I add it to the existing `tests/library/gdb/test_command_cyclic.py`, or should I create a new cross-architecture test file in the `tests/library/qemu-user/` directory?

pwndbg/commands/cyclic.py
```diff
@@ -8,10 +8,53 @@
...
+                found_patterns.append((f"{reg_name}->", value, offset))
+
+    if not found_patterns:
+        print(message.notice("No cyclic patterns found."))
+        return
+
+    print(f"{'Register':<12} {'Value':<20} {'Offset'}")
+    print(f"{'----------':<12} {'------------------':<20} {'------'}")
+    for reg, val, off in found_patterns:
+        print(f"{reg:<12} {hex(val):<20} {off}")
```
disconnect3d:
```suggestion
        print(f"{reg:<12} {val:<#20x} {off}")
```

disconnect3d:
> Should I add it to the existing `tests/library/gdb/test_command_cyclic.py`,

I think adding to the existing is fine; not sure if we need cross arch for this, probably not really

peace-maker:
Blindly running cyclic_find on any value might stall for a while. Maybe a timeout or heuristic to detect probable cyclic patterns is required.

`cyclic_find('hdjakdbj', n=8)` takes a hot minute on my phone for example.

disconnect3d:
I like @peace-maker suggestion. @ArmoredVortex can you look if we can add a timeout?

armoredvortex:
> I like @peace-maker suggestion. @ArmoredVortex can you look if we can add a timeout?

Yes, an optional argument `timeout` that accepts time in seconds.
It skips over the register if the lookup takes longer that said timeout.

Does this sound fair?
EDIT: Or is setting a default timeout value better?

disconnect3d:
Lgtm, I think 2s timeout default is fine. I haven't tested it myself but I hope the signal handler logic works fine with GDB and LLDB.

PR #3120, opened 2025-06-19
Add LLDB test driver and initial Debugger API tests

mbrla0:
This PR expands the reworked testing structure from #3101 so that it supports running tests using LLDB as well as GDB, and adds the first few Debugger-agnostic tests - that currently only check launching - that can be run under either debugger.

PR #3165, opened 2025-07-22
Port tests to debugger-agnostic test group: First Batch

mbrla0:
This PR contains the first batch (out of two) of tests ported to the Debugger API test group introduced in #3120.

### DIFF
--- a/pwndbg/dbg/lldb/__init__.py
+++ b/pwndbg/dbg/lldb/__init__.py
@@ -191,6 +191,15 @@ class LLDBFrame(pwndbg.dbg_mod.Frame):
                     # event.
                     self.proc.dbg._trigger_event(pwndbg.dbg_mod.EventType.REGISTER_CHANGED)
 
+                    # If we set the stack pointer, the inner object might have been invalidated, try
+                    # to restore it, as it should still be the selected frame.
+                    if (
+                        name in (reg_sets[pwndbg.aglib.arch.name].frame, "sp")
+                        and not self.inner.IsValid()
+                    ):
+                        self.inner = thread.GetSelectedFrame()
+                        assert self.inner.GetSP() == val
+
                     # Make sure we've caught and handled the special cases in which the inner object
                     # might be invalidated by the command.
                     assert self.inner.IsValid()
--- a/pwndbg/dbg/lldb/repl/__init__.py
+++ b/pwndbg/dbg/lldb/repl/__init__.py
@@ -46,6 +46,8 @@ import sys
 import threading
 from contextlib import contextmanager
 from io import BytesIO
+from io import TextIOBase
+from io import TextIOWrapper
 from typing import Any
 from typing import Awaitable
 from typing import BinaryIO
@@ -369,7 +371,7 @@ def run(
                 last_exc = asyncio.CancelledError()
                 continue
 
-            if not exec_repl_command(line, sys.stdout.buffer, dbg, driver, relay):
+            if not exec_repl_command(line, sys.stdout, dbg, driver, relay):
                 last_exc = asyncio.CancelledError()
                 continue
 
@@ -386,12 +388,12 @@ def run(
 
             if action._capture:
                 with BytesIO() as output:
-                    should_continue = exec_repl_command(action._command, output, dbg, driver, relay)
+                    should_continue = exec_repl_command(
+                        action._command, TextIOWrapper(output), dbg, driver, relay
+                    )
                     last_result = output.getvalue()
             else:
-                should_continue = exec_repl_command(
-                    action._command, sys.stdout.buffer, dbg, driver, relay
-                )
+                should_continue = exec_repl_command(action._command, sys.stdout, dbg, driver, relay)
 
             if not should_continue:
                 last_exc = asyncio.CancelledError()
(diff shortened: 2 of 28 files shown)


mbrla0:
Important to note that the GDB tests these are based on - and thus any repeated tests - are planned to be removed on a later PR.

tests/library/dbg/tests/test_command_branch.py
```diff
@@ -0,0 +1,48 @@
...
+
+@pwndbg_test
+@pytest.mark.parametrize("binary", [CONDBR_X64_BINARY], ids=["x86-64"])
+async def test_command_break_if_x64(ctrl: Controller, binary: str) -> None:
+    """
+    Tests the chain for a non-nested linked list
+    """
+    import pwndbg
+
+    if not pwndbg.dbg.is_gdblib_available():
```
patryk4815, line 22: better add decorator pytest skipIf ?

mbrla0: When running in LLDB, Pwndbg is only properly initialized sometime after the test starts, as part of how we handle asynchronous interactions with the Pwndbg CLI, so `skipif` would evaluate the condition way too early to be useful. We could do `pytest.skip()` instead of returning, though.

tests/host/lldb/launch_guest.py
```diff
@@ -25,23 +25,39 @@ class _LLDBController(Controller):
         def __init__(self, pc: PwndbgController):
             self.pc = pc
 
-        async def launch(self, binary: Path) -> None:
+        async def launch(self, binary: Path, args: List[str] = []) -> None:
             await self.pc.execute(f"target create {binary}")
-            await self.pc.execute("process launch -s")
+            await self.pc.execute("process launch -s -- " + " ".join(args))
```
patryk4815:
```suggestion
            await self.pc.execute("process launch -s -- " + " ".join(shlex.quote(arg) for arg in args))
```
probably this way we should handle it?

mbrla0: Maybe. Originally I chose not to do it because I felt leaving that up to the tests themselves was good enough, and, indeed, over the course of porting the tests it never came up. But if you feel it's better to add quotations automatically, I'm okay with it, too, and it might be helpful for people adding new tests in the future.

pwndbg/dbg/lldb/repl/__init__.py
```diff
@@ -400,14 +400,43 @@ def handle_sigint(_sig, _frame):
...
+        dbg.debugger.SetOutputFile(
+            lldb.SBFile.Create(output_to, borrow=True, force_io_methods=True)
+        )
+
+        return _exec_repl_command(line, output_to.buffer, dbg, driver, relay)
+    finally:
+        if stdout is not None:
+            sys.stdout = stdout
+        if lldb_out is not None:
+            dbg.debugger.SetOutputFile(lldb_out)
```
patryk4815, line 427: I know this is in one place, but moving this logic into contextmanager would be more readable eg:
```py
@contextlib.contextmanager
def ctx_output_something(dbg, output_to) -> typing.Iterator[io.BytesIO]:
    stdout = None
    lldb_out = None
    try:
        stdout = sys.stdout
        lldb_out = dbg.debugger.GetOutputFile()

        sys.stdout = output_to
        dbg.debugger.SetOutputFile(
            lldb.SBFile.Create(output_to, borrow=True, force_io_methods=True)
        )
        yield output_to.buffer
    finally:
        if stdout is not None:
            sys.stdout = stdout
        if lldb_out is not None:
            dbg.debugger.SetOutputFile(lldb_out)

def exec_repl_command(
     line: str,
     output_to,
     dbg: LLDB,
     driver: ProcessDriver,
     relay: EventRelay,
 ) -> bool:
     with ctx_output_something(dbg, output_to) as buffer:
         return _exec_repl_command(line, buffer, dbg, driver, relay)

```

tests/library/dbg/tests/test_command_branch.py
patryk4815: Ok, lets add just `pytest.skip("Not yet available outside GDB.")`

patryk4815 (approved):
Looks good to me, but I dislike that we will lose git history :/

mbrla0:
> Looks good to me, but I dislike that we will lose git history :/

Worried that we might not be able to find the exact commit if some of these cause regressions later down the line? I could break off some of these commits into their own PRs if you're worried it's going to become a problem.

pwndbg/dbg/lldb/__init__.py
```diff
@@ -191,6 +191,15 @@ def reg_write(self, name: str, val: int) -> bool:
                     # event.
                     self.proc.dbg._trigger_event(pwndbg.dbg_mod.EventType.REGISTER_CHANGED)
 
+                    # If we set the stack pointer, the inner object might have been invalidated, try
+                    # to restore it, as it should still be the selected frame.
+                    if (
+                        name in (reg_sets[pwndbg.aglib.arch.name].frame, "sp")
```
disconnect3d, line 197: Is the `"sp"` literal check needed here bcoz `frame` may be e.g. `"rsp"`? or maybe we have some alias which makes `regs.sp` always work even on different archs?

disconnect3d (approved):

tests/host/lldb/launch_guest.py
```diff
@@ -25,23 +25,39 @@ class _LLDBController(Controller):
         def __init__(self, pc: PwndbgController):
             self.pc = pc
 
-        async def launch(self, binary: Path) -> None:
+        async def launch(self, binary: Path, args: List[str] = []) -> None:
             await self.pc.execute(f"target create {binary}")
-            await self.pc.execute("process launch -s")
+            await self.pc.execute("process launch -s -- " + " ".join(shlex.quote(arg) for arg in args))
```
disconnect3d:
```suggestion
            await self.pc.execute(
                "process launch -s -- " + " ".join(shlex.quote(arg) for arg in args)
            )
```

pwndbg/dbg/lldb/__init__.py
mbrla0: It's the latter.

disconnect3d:
The shlex needs an import i think?

PR #3169, opened 2025-07-23
Cleanup the "find gcc" code

k4lizen:
Seems to be a typo. I was trying to include a header file into a 64-bit process that referenced the gcc intrinsic `__int128` type, but it's not available on `-m32` targets so the compilation was failing on me.

### DIFF
--- a/pwndbg/lib/gcc.py
+++ b/pwndbg/lib/gcc.py
@@ -11,72 +11,20 @@ import platform
 from typing import Any
 from typing import List
 
+import pwnlib
+
 from pwndbg.lib.arch import ArchDefinition
 
 printed_message = False
 
 
 def which(arch: ArchDefinition) -> List[str]:
-    gcc = _which_binutils("g++", arch)
-
-    if gcc is None:
-        global printed_message
-        if not printed_message:
-            printed_message = True
-            print("Can't find appropriate GCC, using default version")
-
-        if arch.ptrsize == 32:
-            return ["g++", "-m32"]
-        elif arch.ptrsize == 64:
-            return ["g++", "-m64"]
-        else:
-            raise ValueError(f"Unknown pointer size: {arch.ptrsize}")
+    try:
+        gcc = pwnlib.asm.which_binutils("g++")
+    except pwnlib.exception.PwnlibException as _:
+        raise ValueError("Couldn't find g++ for the current architecture.")
 
     return [gcc] + _flags(arch.name)
-
-
-def _which_binutils(util: str, arch: ArchDefinition, **kwargs: Any) -> str | None:
-    ###############################
-    # Borrowed from pwntools' code
-    ###############################
-
-    arch_name = arch.name
-
-    # Fix up binjitsu vs Debian triplet naming, and account
-    # for 'thumb' being its own binjitsu architecture.
-    arches: List[str | None] = [arch_name] + {
-        "thumb": ["arm", "armcm", "aarch64"],
-        "i386": ["x86_64", "amd64"],
-        "i686": ["x86_64", "amd64"],
-        "i386:x86-64": ["x86_64", "amd64"],
-        "amd64": ["x86_64", "i386"],
-    }.get(arch_name, [])
-
-    # If one of the candidate architectures matches the native
-    # architecture, use that as a last resort.
-    machine = platform.machine()
-    machine = "i386" if machine == "i686" else machine
-
-    if arch_name in arches:
-        arches.append(None)
-
-    for arch_name in arches:
-        # hack for homebrew-installed binutils on mac
-        for gutil in ["g" + util, util]:
-            # e.g. objdump
-            if arch_name is None:
-                pattern = gutil
-
-            # e.g. aarch64-linux-gnu-objdump
-            else:
-                pattern = f"{arch_name}*linux*-{gutil}"
-
-            for dir in os.environ["PATH"].split(":"):
-                res = sorted(glob.glob(os.path.join(dir, pattern)))
-                if res:
-                    return res[0]
-
-    return None
 
 
 def _flags(arch_name: str) -> List[str]:
--- a/pwndbg/lib/gcc.py
+++ b/pwndbg/lib/gcc.py
@@ -28,7 +28,7 @@ def which(arch: ArchDefinition) -> List[str]:
         if arch.ptrsize == 32:
             return ["g++", "-m32"]
         elif arch.ptrsize == 64:
-            return ["g++", "-m32"]
+            return ["g++", "-m64"]
         else:
             raise ValueError(f"Unknown pointer size: {arch.ptrsize}")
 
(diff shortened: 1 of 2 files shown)


k4lizen:
@jxuanli you might want to check if this affects anything re https://github.com/pwndbg/pwndbg/pull/3116

pwndbg/lib/gcc.py
```diff
@@ -28,7 +28,7 @@ def which(arch: ArchDefinition) -> List[str]:
         if arch.ptrsize == 32:
             return ["g++", "-m32"]
         elif arch.ptrsize == 64:
-            return ["g++", "-m32"]
+            return ["g++", "-m64"]
         else:
             raise ValueError(f"Unknown pointer size: {arch.ptrsize}")
```
jxuanli: I believe the conditions were not correct. The exception gets triggered when I manually set `gcc = None`. The reason being `arch.ptrsize=8` instead of 64.

jxuanli: However, in the case that `g++` cannot be found with `_which_binutils`, I'm not sure why it doesn't just immediately throws an exception instead of having a fallback. But otherwise your change looks correct to me just that the fallback path hasn't been triggered for me so far because `g++` could be found on my machine.

OBarronCS: Oop - yeah, `.ptrsize` is in bytes instead of bits (`.ptrbits` for bits).

OBarronCS:
https://github.com/pwndbg/pwndbg/blob/4022193b496167030cda1048198fb6953758f78c/pwndbg/lib/gcc.py#L38 appears to be an old version of https://github.com/gallopsled/pwntools/blob/3eb690bd38c3311b0e6ce79a45310f80c2b5222e/pwnlib/asm.py#L158. 

We might be able to replace some of these with the newer version counterparts in Pwntools:

```python
# The pwnlib context (.arch, .context, .etc) is already set correctly
pwnlib.asm.which_binutils("g++")
```

pwndbg/lib/gcc.py
k4lizen: > The reason being arch.ptrsize=8 instead of 64.

oh wow true, I tested it and it seemed to me like my patch changed something but I guess I hallucinated xd

k4lizen:
Changed to just using the pwntools function. This also implies pwntools controlling the error message, but I think it looks fine honestly:
image

patryk4815:
@k4lizen is this code wrong for big endian architecture? eg. mips + BE or LE? I think there are missing endianness flags.. :/

k4lizen:
The endianness is encoded in the architecture name right? `mips64-linux-gnu-gcc` is BE and `mipsel-linux-gnu-gcc` is LE. So I think actually it should work as long as pwntools sets `context.arch` to `mipsel` on little endian mips.

patryk4815:
This code(+pwntools) was broken from beginning xD

> it should work as long as pwntools sets context.arch to mipsel

pwntools don't support `mipsel`/`mips64` name, only `mips`

> So I think actually it should work as long as pwntools sets context.arch to mipsel on little endian mips.

no, where we use pwntools flags here? We are only using `def _flags(arch_name: str)` which don't set endianness 🤔 

> The endianness is encoded in the architecture name right? 

not in pwndbg 
image

k4lizen:
> The endianness is encoded in the architecture name right?
> > not in pwndbg

I meant when compiling. This is true for gcc as well as for zig right? As far as I understand you specify a target triple which is like `mips-linux-gnueabi` for big endian and `mipsel-linux-gnueabi` for little endian, there is no `--endianness=little` or whatever. I may be wrong but that is my understanding.

> no, where we use pwntools flags here? We are only using def _flags(arch_name: str) which don't set endianness 🤔

`pwnlib.asm.which_binutils` fetches its `context.arch` which we seem to update in `pwndbg.aglib.arch:update()`. but yeah if the endianness is not encoded in the `.arch` that means it won't work

> This code(+pwntools) was broken from beginning xD
> > it should work as long as pwntools sets context.arch to mipsel
>
> pwntools don't support mipsel/mips64 name, only mips

Ah. Welp. :P 

On the bright side, this means the kernel PR by @jxuanli will work on other arches if we fix this right? That's pretty cool.

image

disconnect3d (approved):

disconnect3d:
> `pwnlib.asm.which_binutils` fetches its `context.arch` which we seem to update in `pwndbg.aglib.arch:update()`. but yeah if the endianness is not encoded in the `.arch` that means it won't work

@k4lizen pls create an issue for that case in pwntools repo and link it here

k4lizen:
^^ https://github.com/Gallopsled/pwntools/issues/2591

Arusekk:
> I meant when compiling. This is true for gcc as well as for zig right? As far as I understand you specify a target triple which is like `mips-linux-gnueabi` for big endian and `mipsel-linux-gnueabi` for little endian, there is no `--endianness=little` or whatever. I may be wrong but that is my understanding.

Actually, there is, but every arch does it differently (from man gcc):
```
AArch64 Options (AArch64 Options) -mabi=name  -mbig-endian  -mlittle-endian [...]
ARM   Options   (ARM   Options)   [...] -mlittle-endian   -mbig-endian  -mbe8   -mbe32 [...]
C6X Options (C6X Options) -mbig-endian  -mlittle-endian  [...]
C-SKY   Options   (C-SKY   Options)  -march=arch   -mcpu=cpu  -mbig-endian   -EB   -mlittle-endian   -EL [...]
eBPF  Options  (eBPF  Options)  -mbig-endian  -mlittle-endian [...]
Options (IA-64  Options)  -mbig-endian   -mlittle-endian [...]
MCore   Options   (MCore   Options)   [...] -mlittle-endian  -mbig-endian [...]
MicroBlaze   Options   (MicroBlaze  Options)  [...] -mbig-endian  -mlittle-endian [...]
MIPS Options (MIPS Options) -EL  -EB [...]
NDS32  Options  (NDS32 Options) -mbig-endian  -mlittle-endian [...]
RISC-V  Options (RISC-V Options) [...] -mbig-endian -mlittle-endian [...]
RS/6000  and PowerPC Options (RS/6000 and PowerPC Options) [...] -mlittle -mlittle-endian -mbig -mbig-endian [...]
RX Options (RX Options)  [...] -mbig-endian-data   -mlittle-endian-data [...]
```
And, pwntools actually accounts for it, but internally in pwnlib.asm._assembler. The logic might be worth exporting.
What interface would you like the most/cleanest?

> `pwnlib.asm.which_binutils` fetches its `context.arch` which we seem to update in `pwndbg.aglib.arch:update()`. but yeah if the endianness is not encoded in the `.arch` that means it won't work

On it.

> > pwntools don't support mipsel/mips64 name, only mips
> 
> Ah. Welp. :P

That's right, and this is on purpose, I think the best discrimination criterion for Pwntools is the instruction set. For example, riscv32 includes RV32C but riscv64 including RV64C is currently distinct, but IMO RV32C is more distinct from RV32 than RV64 is. Whatever. If changing endianness does not change the instruction set, the pwntools arch stays the same.

patryk4815:
@Arusekk Right now we removed gcc and we are using zig for compiling. Just single `zig` binary is needed to handle most of architectures ;)
https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/lib/zig.py

PR #3172, opened 2025-07-24
Make ida integration work in shared libraries

k4lizen:
If you have a shared library open in IDA the integration would just fail to provide the decompilation since it was incorrectly calculating the address offset by always assuming the base is `exe = pwndbg.aglib.elf.exe()`.


pwndbg/aglib/vmmap.py
```diff
@@ -41,3 +42,34 @@ def find(address: int | pwndbg.dbg_mod.Value | None) -> pwndbg.lib.memory.Page |
...
+    """
+    Let's define a "region" as contiguous memory compromised of memory mappings
+    which all have the same object file name. Also referred to as "File (Base)" by
+    `xinfo`.
+
+    Returns:
+        The start of the memory region this address belongs to, or None if the address
+        is not mapped or a valid memory region couldn't be found.
+    """
+    address = int(address)
```
disconnect3d, line 57: Is int() needed if the arg is supposed to be int anyway? or can it be gdb.Value (our wrapper) or what?:)

pwndbg/integration/ida.py
```diff
@@ -180,18 +180,18 @@ def can_connect() -> bool:
...
 
 
 def r2l(addr: int) -> int:
-    exe = pwndbg.aglib.elf.exe()
-    if not exe:
-        raise Exception("Can't find EXE base")
-    result = (addr - base() + int(exe.address)) & pwndbg.aglib.arch.ptrmask
+    region_start = pwndbg.aglib.vmmap.addr_region_start(addr)
+    if region_start is None:
+        return 0
```
disconnect3d, line 193: is it ok that we now return 0 and we were raising exc before?

pwndbg/aglib/vmmap.py
```diff
@@ -41,3 +42,34 @@ def find(address: int | pwndbg.dbg_mod.Value | None) -> pwndbg.lib.memory.Page |
...
+    if address < 0:
+        return None
+
+    page = find(address)
```
disconnect3d: is it needed if we do vmmap.get() and get objfiles anyway?

pwndbg/integration/ida.py
k4lizen: Before, we were returning invalid pointers all the time, not even 0 or raising the exception, the calling code handles this fine. The exception was only raised when the ELF's mappings couldn't be found which is only one possible failure mode.

If I raise the exception here now this case:
```python
    # Check the region is contiguous. It can easily happen that it isn't, for instance
    # something like "[anon:libc_malloc]" may be found all over the address space.
    for i in range(len(sorted_pages) - 1):
        if sorted_pages[i].end != sorted_pages[i + 1].start:
            return None
```
will be hit quite often, which shouldn't really happen.

pwndbg/aglib/vmmap.py
k4lizen: I guess not

k4lizen: fixed

pwndbg/aglib/vmmap.py
```diff
@@ -52,24 +51,34 @@ def addr_region_start(address: int | pwndbg.dbg_mod.Value) -> int | None:
...
-    if page is None:
-        return None
+    mappings = sorted(pwndbg.aglib.vmmap.get(), key=lambda p: p.vaddr)
+    idx = -1
+    for i in range(len(mappings)):
+        if mappings[i].start <= address < mappings[i].end:
+            idx = i
+            break
+
+    if idx == -1:
```
disconnect3d, line 67: This is a perfect case for a `for: ... else:` :D

k4lizen: oh! didn't know about that! i'll keep it in mind for next time

PR #3174, opened 2025-07-24
Set frame register for aarch64

k4lizen:
It wasn't defined because of this before: https://github.com/pwndbg/pwndbg/issues/1039#issuecomment-1207312942 . But now that `LR` is shown seperately, the display isn't confusing anymore.

image

### DIFF
--- a/pwndbg/lib/regs.py
+++ b/pwndbg/lib/regs.py
@@ -598,7 +598,7 @@ armcm = RegisterSet(
 
 # AArch64 has a PSTATE register, but GDB represents it as the CPSR register
 aarch64 = RegisterSet(
-    retaddr=(Reg("lr", 8),),
+    retaddr=(Reg("lr", 8),), # x30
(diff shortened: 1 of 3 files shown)


pwndbg/lib/regs.py
```diff
@@ -647,7 +643,6 @@ def __repr__(self):
         Reg("x26", 8, subregisters=(Reg("w26", 4, zero_extend_writes=True),)),
         Reg("x27", 8, subregisters=(Reg("w27", 4, zero_extend_writes=True),)),
         Reg("x28", 8, subregisters=(Reg("w28", 4, zero_extend_writes=True),)),
```
disconnect3d, line 645:
```suggestion
        Reg("x28", 8, subregisters=(Reg("w28", 4, zero_extend_writes=True),)),
        # Note: x29 is FP (frame) and x30 is LR (retaddr) register
```

PR #3176, opened 2025-07-24
Add colorful malloc/free output for #3090

jacksonkvandyke:
In response to issue "heap tracker: color or mark pointers? #3090" the following has been added to this PR.

Enhanced Heap Tracker Visuals:
- Introduced dynamic coloring for malloc, free, and realloc events
- Each allocation or deallocation prints its associated pointer in a visually distinct color to aid heap behavior analysis and debugging.


pwndbg/gdblib/ptmalloc2_tracking.py
```diff
@@ -92,6 +95,20 @@ def is_enabled() -> bool:
...
+    color = event_colors[malloc_counter % len(event_colors)]
+    malloc_counter += 1
+    return f"{color}{ptr}\033[0m"
+
+
+def colorize_free(ptr):
+    global free_counter
+    color = event_colors[free_counter % len(event_colors)]
+    free_counter += 1
+    return f"{color}{ptr}\033[0m"
```
k4lizen: if possible you should use `pwndbg.color.whatever` instead of writing the ANSII by hand

k4lizen (commented):
Thanks! Could you shown a screenshot of this working in action?

disconnect3d:
Hi, thanks for the PR!

Looking at the code I am wondering:

> * Each allocation or deallocation prints its associated pointer in a visually distinct color to aid heap behavior analysis and debugging

Is this really true? I think that if we have:
```c
x = malloc(...);
y = malloc(...);

free(y);
free(x);
```

We will get printed colors 1, 2, 1, 2 instead of 1, 2, 2, 1.

I mean, that the freeing of a given pointer should print free in the same color as its corresponding malloc.

pwndbg/gdblib/ptmalloc2_tracking.py
jacksonkvandyke: Thank you for the feedback. I can definitely change it to use pwndbg.colors.

jacksonkvandyke:
I can see your reasoning for questioning the functionality. I will revisit the drawing board to ensure they print with the correct pointer.

jacksonkvandyke:
I have updated the functionality to reflect the above statements and made the following changes:

- Uses pwndbg.color.somecolor instead of manual ANSII
- Now instead of using 2 function, one for malloc and one for free, it now uses a single function that stores the memory pointers current color in a dictionary. This ensures that on malloc, realloc, and frees that the memory pointer will retain the same color. This is useful because it also retains the color for future malloc's that use the same memory pointer.

Please let me know if there is anything else that is desired or should be changed.

jacksonkvandyke:
I have updated the functionality to reflect the above statements and made the following changes:

- Uses pwndbg.color.somecolor instead of manual ANSII
- Now instead of using 2 functions, one for malloc and one for free, it now uses a single function that stores the memory pointers current color in a dictionary. This ensures that on malloc, realloc, and frees that the memory pointer will retain the same color. This is useful because it also retains the color for future malloc's that use the same memory pointer.

Here is an image of the new functionality:
image

Please let me know if there is anything else that is desired or should be changed.

pwndbg/gdblib/ptmalloc2_tracking.py
```diff
@@ -92,6 +93,31 @@ def is_enabled() -> bool:
...
+    This function works by assigning a color to a specific memory pointer, storing the pointer
+    in a colors array and returning the colored pointer. If a color has already been stored for
+    the given pointer, it will be returned.
+    """
+    if ptr in mem_colors:
+        color = mem_colors[ptr]
+    else:
+        color = pos_colors[len(mem_colors) % len(pos_colors)]
+        mem_colors[ptr] = color
+    return color(ptr)
```
disconnect3d:
```suggestion
    if color := mem_colors.get(ptr) is None:
        color = pos_colors[len(mem_colors) % len(pos_colors)]
        mem_colors[ptr] = color
    return color(ptr)
```

pwndbg/gdblib/ptmalloc2_tracking.py
```diff
@@ -468,10 +495,11 @@ def stop(self) -> bool:
         if requested_size == 0:
             # There's no right way to handle realloc(..., 0). C23 says it's
             # undefined behavior, and prior versions say it's implementation-
-            # defined. Either way, print a warning and do nothing.
+            # defined. Either way, print a warning and do nothing.'
```
disconnect3d:
```suggestion
            # defined. Either way, print a warning and do nothing.
```

disconnect3d:
I like the current approach, but two comments:
1. Shall we cache color functions or the result from coloring? Is there any disadventage from caching the result?
2. We probably need to clear the cache on process restart, but idk if that feature supports this well at all :/

jacksonkvandyke:
We could cache the color results. 

Another option is we could just use the pointer itself to determine the color? Meaning colorarray[somepointer%len(colorarray)] this way we don't have to store anything. This does however risk have 2 mallocs right by each other having the exact same color.

disconnect3d:
Yeah, let's cache the results, it doesn't hurt to do so. We should probably also add more colors if possible.

Regarding `colorarray[somepointer%len(colorarray)]` it probably isn't that great since depending on the allocator and object sizes the returned pointers and `% some_value` may end up returning the same color (not to mention, that we would have to do some bitshift here, since usually the returned pointers are aligned to some value).

jacksonkvandyke:
Sounds good, I should be able to get back to this PR sometime next week.

disconnect3d:
@jacksonkvandyke ping, any update here?:)

disconnect3d:
@jacksonkvandyke ping! :)

jacksonkvandyke:
@disconnect3d Sorry for the late response, last week was very busy for me.

I am still looking at a good way to cache the color results and its taking me a bit longer becuase I am still somewhat new to caching over using global variables. I will be sure to send an update on this PR once I have had some more time to brainstorm.

disconnect3d:
> @disconnect3d Sorry for the late response, last week was very busy for me.
> 
> I am still looking at a good way to cache the color results and its taking me a bit longer becuase I am still somewhat new to caching over using global variables. I will be sure to send an update on this PR once I have had some more time to brainstorm.

I think u can do sth like:

```py
colorized_heap_ptrs = {}  # or other varname idk

# in heap tracker command
global colorized_heap_ptrs
colorized_heap_ptrs = {}
# or: some_module.colorized_heap_ptrs = {}

# in function that updates it:
colorized_heap_ptrs[addr] = colored_string
```

jacksonkvandyke:
Thank you for the guidance @disconnect3d 

I have updated the colorize to store the results in the Tracker. Here is a screenshot with additional colors added as well: 
image

Please let me know if I added too many colors of if there is anything else that should be changed.

pwndbg/gdblib/ptmalloc2_tracking.py
```diff
@@ -228,6 +229,45 @@ def exit_memory_management(self) -> None:
...
+            pwndbg.color.cyan,
+            pwndbg.color.light_gray,
+            pwndbg.color.gray,
+            pwndbg.color.light_red,
+            pwndbg.color.light_green,
+            pwndbg.color.light_yellow,
+            pwndbg.color.light_blue,
+            pwndbg.color.light_purple,
+            pwndbg.color.light_cyan,
+        ]
```
disconnect3d: Make this a global tuple please

pwndbg/gdblib/ptmalloc2_tracking.py
```diff
@@ -228,6 +229,45 @@ def exit_memory_management(self) -> None:
...
+        # Check cache first
+        if ptr in self.colorized_heap_ptrs:
+            return self.colorized_heap_ptrs[ptr]
+
+        # Compute color
+        malloc_order = list(self.colorized_heap_ptrs.keys())
+        if ptr not in malloc_order:
+            malloc_order.append(ptr)
+
+        index = malloc_order.index(ptr) % len(pos_colors)
```
disconnect3d: This `malloc_order` thing is overengineered/overly complex

disconnect3d:
@jacksonkvandyke hi, i've simplified the colorize function a bit, let me know if this looks good for u :)

disconnect3d:
Thanks!

Issue #2324, opened 2024-07-27
conflict hex value display between disassembly-flavor and disasm result in Arch Linux


RocketMaDev:
### Description

I know the default disassembly-flavor for pwndbg is intel, and code should be displayed like `sub rsp, 0x10`. But it's so weird on my system that these hex values are displayed like `10h`. An example screenshot is attached below.

![屏幕截图_20240727_220718](https://github.com/user-attachments/assets/0dfb78e9-0ddb-4ebd-bd35-b48487080552)

I tried to `set disassembly-flavor intel`, nothing happened. But when I `set disassembly-flavor att`, the hex value displayed become normal like:

![屏幕截图_20240727_220901](https://github.com/user-attachments/assets/c8a9098c-2deb-44a2-9951-010e24cf049c)

Even more weirdly, I then `set disassembly-flavor intel` back, and the code was displayed just like the screenshot above, it didn't go back to `sub rsp, 10h`.

### Steps to reproduce

My `.gdbinit`:

```gdbinit
set debuginfod enabled on
source /usr/share/pwndbg/gdbinit.py
add-auto-load-safe-path /usr/lib/go/src/runtime/runtime-gdb.py
set context-sections regs disasm code backtrace threads expressions
```

Setting disassembly-flavor to intel and att some times when debugging or write it into gdbinit

### My setup

System: Arch Linux
Gdb:      15.1
Python:   3.12.4 (main, Jun  7 2024, 06:33:07) [GCC 14.1.1 20240522]
Pwndbg:   2024.02.14
Capstone: 5.0.1280
Unicorn:  2.0.1

disconnect3d:
Hi, this is ... interesting. We use capstone for disassembling code. I really wonder why this behaves as you described.

Do you maybe want to try to debug this? Otherwise I can look at it in upcomming days.

RocketMaDev:
Actually I haven't dug into capstone. I'm just a pwndbg user. I can take a look at it if you could tell me where to check the code. Or else I can wait for your update.

OBarronCS:
This is a very strange bug. Capstone is providing the hex literal, and these lines initialize the disassembler:

https://github.com/pwndbg/pwndbg/blob/0f61bff5bb290279eefb61a5f8c616c7da51b82d/pwndbg/gdblib/disasm/__init__.py#L138-L151

disconnect3d:
[...]

PR #3181, opened 2025-07-27
docs: update setup instructions for Arch Linux

code-IM-perfect:
The pwndbg package has been added to the official repositories of Arch Linux.

So, the recommended way to install should not be manually downloading and installing, instead it should be downloading the package from the official repository.

I also added the command to add `source /usr/share/pwndbg/gdbinit.py` to `~/.gdbinit` because I myself struggled to find out that this was a necessary step.\

I'm not sure if this is needed for other platforms too but I know for sure that on Arch Linux the package manager does not touch user controlled configuration, so this is necessary on Arch Linux. If that is a case on other platforms too, we can add it as a separate note on the page, but I feel it is important it be mentioned somewhere.

Build was successful with the instructions provided in the contributing guidelines.

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -38,7 +38,9 @@ apk add --allow-untrusted ./pwndbg_2025.05.30_x86_64.apk
 ```
 Arch Linux:
 ```{.bash .copy}
-pacman -U ./pwndbg-2025.05.30-1-x86_64.pkg.tar.zst
+pacman -S pwndbg
+# Make gdb load pwndbg on startup
+echo 'source /usr/share/pwndbg/gdbinit.py' >> ~/.gdbinit
 ```
 
 ## Installing pwndbg-lldb


patryk4815:
Pwndbg officially only supports its own `.pkg.tar.zst` package for pacman. 
If we were to include information about the Arch Linux repository version, it would belong in a separate `"Unofficial packages"` section, clearly marked to indicate that it is not maintained by the pwndbg team and may often be outdated, broken, or incompatible with the latest version.

To avoid confusion and ensure users follow the officially supported installation paths, we’ll be closing this PR.

code-IM-perfect:
Have there been any issues in the past because of this? Because the source for the Arch package is this repo itself so I would assume the `.pkg.tar.zst` package is what is used there as well. And like this being the norm of how projects usually work and everything.

Anyways even if the instructions are not changed, can we please mention that we need to source the `gdbinit.py` in `~/.gdbinit`, somewhere in the setup page, because that is a critical part of setup and it left me and apparently many others very confused after installation.

disconnect3d:
Hi, yes, there have been some issues with the official Arch package in the past:
- In https://github.com/pwndbg/pwndbg/issues/2324#issuecomment-2294849983, Arch moved to a newer Capstone which had a bug and wasn't used by us internally
- In https://github.com/pwndbg/pwndbg/pull/2493, the GDB package started to require libxxhash which was detected only at runtime (instead of when gdb was built, which would be detected by some CI/CD)
- The Arch Linux package didn't work initially when we started to create a Python virtual environment for Pwndbg

I wouldn't say those are necessarily Arch Linux's fault or anything, but since we couldn't keep up with various distros and test or maintain Pwndbg against different dependencies versions we decided to eventually ship Pwndbg with all dependencies included (even GDB and Python).

Fwiw @anthraxx is/was maintaining the Arch package, so he may have some more input here.

disconnect3d:
@anthraxx actually, is there any chance the Arch package could start using our release packages?

disconnect3d:
I'm going to close this for now since I don't think we will recommend this atm unless Arch starts using our release package directly.

patryk4815:
@disconnect3d I think this could be added to our documentation in a separate "Unofficial packages" section with installation instructions. ArchLinux has its own build system and won't be using our official package version.

Note that Arch Linux is a bleeding-edge distribution, so pwndbg will often be used with newer versions of dependencies than expected. This tends to result in significantly more bug reports and unexpected behavior.

There's also an unofficial pwndbg package available for Gentoo: https://github.com/gentoo/gentoo/tree/master/dev-debug/pwndbg

Xeonacid:
> @anthraxx actually, is there any chance the Arch package could start using our release packages?

Distro would like to build everything as much as possible for more controllable dependency management and to avoid supply-chain attack. :)

PR #3182, opened 2025-07-27
empty context-sections

### LINKED ISSUE #2510: ctx command: display context even when context-sections is empty?
This is to be discussed: I find it annoying that the `ctx` command displays nothing when `set context-sections ''` is empty.

It should probably display the default context at this point?

marywinter3:
When context sections are set empty, ctx command gives a warning.
Closes #2510


disconnect3d:
Pls fix lint issue[s]. 

Is this message only printed when `context` command is used, or is it also printed when context is displayed automatically when the debugger is stopped again after step/next/continue commands?

marywinter3:
> Pls fix lint issue[s].
> 
> Is this message only printed when `context` command is used, or is it also printed when context is displayed automatically when the debugger is stopped again after step/next/continue commands?

It is also printed when context is displayed automatically.

disconnect3d:
> It is also printed when context is displayed automatically.

I think we don't want this: when you set it to empty, u probably do it on purpose and don't want the warning etc.

disconnect3d:
Btw please see https://github.com/pwndbg/pwndbg/issues/2510#issuecomment-3130047874 - it probably makes more sense to:
- Add a new configuration variable `auto-context` with `on|off` value
- When user sets context sections to empty, probably warn them that this is not anymore the way to do it, and set the `auto-context off` for them, not changing the context sections value?

marywinter3:
@disconnect3d Are this tets ok?

peace-maker:
`context --off` is pretty similar and could be changed to just apply to the auto-display?

pwndbg/commands/context.py
```diff
@@ -735,6 +735,15 @@ def context(subcontext=None, enabled=None) -> None:
         subcontext = []
     args = subcontext
 
+    # Inform when auto-context is off
+    if not pwndbg.config.auto_context:
+        pwndbg.config.context_sections.value = ""
```
disconnect3d: The code here makes it so the `context` command sets empty sections and we end up not displaying them + printing a warning -- when `set auto-context off` was applied.

However, this is not what we want since:
1) It reconfigures the `context-sections` to be empty -- and we do not want to change this setting for the user
2) If user executes the `context` command by hand when `set auto-context off` is applied, the user will not see any context. And we want to display the context for them since they explicitly requested it

What we want to do is to have a trigger for  the`context_sections` change and if user sets it to empty, then warn the user that the proper way to do it is to `set auto-context off` instead and we want to apply it there for the user, at the same time not setting the `context_section` to be empty (cancelling the change of it).

We actually have a trigger for `context_sections` parameter change to validate its value in here:
https://github.com/pwndbg/pwndbg/blob/4891cceafebebbc7d5814157703187f6d4f25c2d/pwndbg/commands/context.py#L178-L208

So we should modify this code to do `pwndbg.config.auto_context.value = False` when context sections are set to be empty, and not modifying the value of sections. Though, I wonder if there is a way to revert or not accept a new param value from such a trigger -- I believe there is -- you would have to check that out.

Then, for the `auto_context` not displaying the context, we want a simple `if auto_context is False: return` instead of calling the `context` function by the code that shows a Pwndbg prompt (since this is where the automatic context display happens).

disconnect3d: For GDB that would be in https://github.com/pwndbg/pwndbg/blob/4891cceafebebbc7d5814157703187f6d4f25c2d/pwndbg/gdblib/prompt.py#L82-L99

For LLDB that would be in https://github.com/pwndbg/pwndbg/blob/4891cceafebebbc7d5814157703187f6d4f25c2d/pwndbg/dbg/lldb/hooks.py#L83-L91

disconnect3d (changes requested):
Please see my inline code comments

mmajewsk:
@disconnect3d  somehow the tests are running without the newest commit - can you retrigger them?

pwndbg/commands/context.py
```diff
@@ -177,36 +181,48 @@ def reserve_lines_maybe(cmd_lines: int) -> None:
...
-        config_context_sections.value = ""
         print(
             message.warn(
-                f"Sections set to be empty. FYI valid values are: {', '.join(valid_values)}"
+                f"To set sections to be empty use set auto-context off. FYI valid values are: {', '.join(valid_values)}"
             )
         )
+        _restore_guard = True
+        config_context_sections.value = _last_valid_sections
+        _restore_guard = False
```
disconnect3d, line 211: What is this and why is this set and unset here?

Hmm does this mean `config_context_sections.value` trigger a call to `validate_context_sections`? If so, it seems that we may be setting the config value incorrectly?

k4lizen (changes requested):
dunno what the status of this is exactly, but the tests need to be ported to `dbg/` as well.

disconnect3d:
I'm going to close this for now, but please open another PR after updating this, fixing conflicts, moving to aglib/dbg etc.

PR #3189, opened 2025-07-30
Port tests to debugger-agnostic test group: Second Batch

mbrla0:
This PR completes #3165 with the second batch (out of two) of tests ported to the Debugger API test group introduced in https://github.com/pwndbg/pwndbg/pull/3120.

Keep in mind not all of these work! The enhancements to LLDB that allow them to work have were split into a separate PR.

### DIFF
--- a/tests/binaries/host/heap_bugs.c
+++ b/tests/binaries/host/heap_bugs.c
@@ -38,7 +38,9 @@ malloc_chunk *f_real = (malloc_chunk*)(f - 2*SIZE_SZ); \
 malloc_chunk *g_real = (malloc_chunk*)(g - 2*SIZE_SZ); \
 int *tmp; \
 int tmp2, tmp3; \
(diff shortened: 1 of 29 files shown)


disconnect3d:
Hmm some tests fail on CI after those changes?

mbrla0:
> Hmm some tests fail on CI after those changes?

Some of these were expected.

Since a few of the binaries were changed, a couple of tests in the GDB suite fell out of sync. These were completely ported to the new suite, so I felt the extra effort to keep these GDB-specific but otherwise identical versions up to date probably wouldn't be worth it if they're probably gonna be scrapped soon anyway. If you want, though, I can still update them so they keep working until we fully switch.

Though I'll be honest that the "Error: pc [...] in address map, but not in symtab."-related errors caught me by surprise, and that I didn't actually expect that adding DWARF to the ASM binaries would be a big deal. Also, weirdly enough, those pass just fine locally in both the new and old suites.

mbrla0:
Decided to backport those changes to the GDB suite anyway.

Issue #3124, opened 2025-06-21
[question/discussion] how we distribute pwndbg


k4lizen:
Edit: If you're a packager with some questions, see https://pwndbg.re/pwndbg/latest/tutorials/packaging/ first. If that document doesn't answer your questions, let us/me know.

Honestly I'm a bit lost in the sauce with all the ways we distribute pwndbg and the rationale. I'll dump some thoughts / questions here. Feel free to let me know if you disagree with something or I misunderstood. Let's centralize the information a bit.

We essentially have three main use-cases distribution-wise.
1. Portable package - this is very nice on bare targets / fresh containers.
2. Package manager - People who want to use pwndbg on their machine should be able to install it with a package manager. This is really good quality of life as it makes it easy to update and uninstall. I personally know people running ancient versions of pwndbg :( .
3. From source - There will be people who want the latest features. They may also not necessarily be developers, but all developers will probably use pwndbg as installed from source.

All three of these should have one blessed way that we support that should be as streamlined and footgunless as possible. The portable package works really well, no comments there. 

## Modifying shell vars

A thing I believe we should strive for is to make a setup for 2. and 3. that doesn't recommend/require the user to modify their shell variables. We cannot really do this in a clean way programatically, and it's even a pain to write proper instructions for it since different shells can have different syntaxes (bash, fish, if we're on windows then cmd, pwsh - all different). It can be confusing for newish linux users.

https://github.com/pwndbg/pwndbg/pull/3119 puts gdb and lldb into the venv, so ubuntu contributors won't have to install lldb-19 and edit shell vars to export it to PATH which is awesome (doc gen works! yay!).

A valid exception to this is `/usr/local/bin` and `$HOME/.local/bin` being in PATH which the install script requires. I'm assuming Ubuntu has it in path by default (@patryk4815 correct me if I'm wrong) and users of more saucy distros (e.g. arch) will know to set their PATH.

### User install
The command should be short. It may seem like a small thing but I think it affects usability a lot. There is a reason `gdb`, `gef`, `lldb` are all a few letters. Even the `pwndbg-gdb` and `pwndbg-lldb` that the install script nicely exposes are a bit too long imo and I would be aliasing it to `pgdb` or `plldb` or something. Maybe we should make a poll on discord about how users would prefer it.

### From source
When pwndbg is installed from source, it should be easily invokeable from whichever folder without having to enter/specify the source folder.

[...]

PR #3195, opened 2025-08-01
Package lldb for distros

Ninja3047:


patryk4815:
there is ongoing discussion if we should remove rpm/apt/deb from pwndbg
https://github.com/pwndbg/pwndbg/issues/3124

disconnect3d:
I guess it doesn't hurt to include this for now. We can remove it later if we decide we don't want to ship distro packages and only keep the portable one.

disconnect3d:
Thanks @Ninja3047 for doing this!

PR #3196, opened 2025-08-01
update flake.lock for gdb 16.3

Ninja3047:


patryk4815:
@Ninja3047 it is already gdb 16.3 xD (it was bumped ~14d ago, and ~3days ago)

PR #3197, opened 2025-08-02
Fix: Replaced assert with explicit ValueError in generate-readme.py

talalhafizmuhammad:
Replaced the assert statement in scripts/generate-readme.py with an explicit ValueError to ensure safer validation of the first line in README.md. This makes the script more robust, especially in production or optimized environments where assertions may be disabled.

### DIFF
--- a/scripts/generate-readme.py
+++ b/scripts/generate-readme.py
@@ -22,15 +22,17 @@ gif = """
 
 def main():
     README_PATH = "README.md"
-    TARGET_PATH = "./docs/index.md"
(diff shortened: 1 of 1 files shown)


disconnect3d:
@k4lizen can u review this?

k4lizen:
i mean, not sure what's the point but I guess it's fine.

if we end up running this script with `python -O ...` then a ValueError will catch the issue while asserts will be disabled

disconnect3d:
I agree that this change is not really needed. We never run with -O. Python folks almost never use that. Lets skip that PR

PR #3200, opened 2025-08-04
pass `dbg` variable into ipython interpreter for lldb

Ninja3047:
This makes the ipython interpreter more useful since this lets you access the debugger state


patryk4815:
you can access this state using `pwndbg.dbg`, no idea why you need that

Ninja3047:
oops, didn't figure that out
thanks!

PR #3202, opened 2025-08-04
cymbol: use zig compiler

patryk4815:
image


pwndbg/lib/zig.py
```diff
@@ -0,0 +1,72 @@
...
+from typing import Dict
+from typing import List
+from typing import Literal
+from typing import Tuple
+
+from pwndbg.lib.arch import PWNDBG_SUPPORTED_ARCHITECTURES_TYPE
+from pwndbg.lib.arch import ArchDefinition
+from pwndbg.lib.arch import Platform
+
+# 'x86-64', 'little', 8)
```
disconnect3d:
```suggestion

```

pwndbg/lib/zig.py
```diff
@@ -0,0 +1,72 @@
...
+    elif arch.platform == Platform.DARWIN:
+        osabi = "macos-none"
+    else:
+        return None
+
+    arch_mapping = _arch_mapping.get((arch.name, arch.endian, arch.ptrsize), None)
+    if arch_mapping is None:
+        return None
+
+    return arch_mapping + '-' + osabi
```
disconnect3d:
```suggestion
    return f"{arch_mapping}-{osabi}"
```

disconnect3d: both versions are fine but f-string is probably faster? idk, not that it matters

pwndbg/lib/zig.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def flags(arch: ArchDefinition) -> List[str] | None:
+    try:
+        import ziglang  # type: ignore[import-untyped]
```
disconnect3d, line 59: Just to be sure: is it installed? shall it ever not be importable?

patryk4815: I think it should be optional, it not exists for all platforms

disconnect3d (approved):

Issue #3091, opened 2025-06-07
heap tracker: show offset from heap instead of full address?


disconnect3d:
Currently the `track-heap enable` shows us full addresses returned by malloc/calloc or passed to free.

I think that if the address belongs to the `[heap]` page, it may be sometimes useful to show the offset from the heap base instead, or, offset from the very first heap top?

Note that teh `[heap]` page address can be fetched as follows:
```py
heap_base = next(p for p in pwndbg.aglib.vmmap.get() if p.objfile=='[heap]').start
```

But it should also be fetched just once for the entire tracker I guess?

cc: @CptGibbon - what do you think here?

CptGibbon:
I've not used the heap-tracker feature.
Did you have a scenario in mind in which a chunk's offset from the start of its heap could be useful?

disconnect3d:
@CptGibbon Please, try using it! I am not sure if showing a chunk's offset from start of the heap is that help helpful but I guess it could be in a case where you have ASLR enabled and only care about the relative position of the allocation.

On the other hand, we can probably do achieve much more here like what i described in https://github.com/pwndbg/pwndbg/issues/3092#issuecomment-2961898854 🤔

CptGibbon:
Yes #3092 got me thinking, I'll comment there once I've collected my thoughts.

PR #3212, opened 2025-08-06
Added heap offset for #3091

jacksonkvandyke:
In response to issue "heap tracker: show offset from heap instead of full address? #3091" the following has been added to this PR.

Enhanced heap tracker print:

- Introduced new function to format each pointer with the offset from the heap
- Maintained original pointer in conjunction with new heap offset pointer for further debugging aid

Here is a screenshot with the new change:
image

### DIFF
--- a/pwndbg/gdblib/ptmalloc2_tracking.py
+++ b/pwndbg/gdblib/ptmalloc2_tracking.py
@@ -63,6 +63,7 @@ import pwndbg.aglib.symbol
 import pwndbg.aglib.typeinfo
 import pwndbg.aglib.vmmap
 import pwndbg.lib.cache
+from pwndbg.lib.cache import cache_until
 from pwndbg.color import message
 
 LIBC_NAME = "libc.so.6"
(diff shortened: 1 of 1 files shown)


CptGibbon:
Thanks for this 🙏

Perhaps a future PR could add compatibility with non-main arena heaps.

disconnect3d:
Hi, I am wondering... we are kinda duplicating the output now and making it amybe a bit more confusing?

Maybe instead of: `malloc(X) -> 0x55..1234 [heap+0x1234]` we should display it as one of the following, depending on user configuration - which could be specified with a flag when enabling heap tracker?:

```
malloc(X) -> 0x55...1234
malloc(X) -> heap+0x1234
```

Also please consider that `malloc` may return something that is not on the heap (an address can be mmaped, or, the allocator may in some cases return attacker-controlled address).

CptGibbon:
I'm keen on letting users print both pieces of information at the same time or just one using options.

jacksonkvandyke:
I will look into allowing the user to enable the different printing options with flags.

disconnect3d:
@jacksonkvandyke ping :)

disconnect3d:
@jacksonkvandyke ping

PR #3215, opened 2025-08-07
Integrate Debugger-agnostic tests into the test pipelines

mbrla0:
This PR completes[^1] the porting of test suite to LLDB by integrating it into the automated testing pipelines.

[^1]: With the small caveat that duplicated tests haven't been removed from the GDB-only suite yet, but those should be easy enough to do in the future.

It also includes fixes for a few more bugs that were uncovered, small tweaks to a few tests, and code to skip tests that are known to not play well with LLDB's feature set, but that I forgot to add in the last PR.

While the volume of code changed here is small, it bundles together quite a large number of small tweaks to different parts of Pwndbg. Nonetheless, none of these tweaks should prove to be a big deal, at least not as far as the tests can see.


tests/library/dbg/tests/__init__.py
```diff
@@ -7,14 +7,17 @@
...
 from .... import host
 from ....host import Controller
 
 BINARIES_PATH = os.environ.get("TEST_BINARIES_ROOT")
 
+T = ParamSpec("T")
 
-def pwndbg_test[**T](
+
+def pwndbg_test(
```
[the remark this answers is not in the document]
mbrla0: https://docs.python.org/3/library/typing.html#typing.ParamSpec

.github/workflows/tests.yml
```diff
@@ -112,10 +117,21 @@ jobs:
...
         sudo sysctl -w kernel.yama.ptrace_scope=0
         ./tests.sh -d gdb -g gdb
+  
+    - name: Run DBG Tests on GDB
+      run: |
+        sudo sysctl -w kernel.yama.ptrace_scope=0
+        ./tests.sh -d gdb -g dbg
+    
+    - name: Run DBG Tests on LLDB
+      if: matrix.os != 'ubuntu-22.04'
```
disconnect3d, line 131: Why this? Can we add a comment?

disconnect3d (approved):

.github/workflows/tests.yml
mbrla0: I was having trouble with a one-off issue that test in particular in 22.04, and since we don't really support it for LLDB, I figured disabling it was okay. Plus, the docker test for 22.04 is still enabled, and it passes fine.

PR #3224, opened 2025-08-11
Fix aarch64 abi

k4lizen:
bug: When arguments are shown, the fifth argument (and onwards) is taken from the stack instead of from $x4 (then $x5 etc.).
```
 ► 0x7f3e900f3d00 <caller+17>    blr    x8                          <callee(int, int, int, int, char const*)>
        arg1: <$x0>
        arg2: <$x1>
        arg3: <$x2>
        arg4: <$x3>
        arg5: <value from stack>
```
This PR fixes this bug.


patryk4815:
it was copied from https://github.com/Gallopsled/pwntools/blob/d665644ae4cfcb66a0129b5d9888b3d4dda307ae/pwnlib/abi.py#L148

PR #3233, opened 2025-08-12
Remove `cast` in pwndbg.aglib.regs

patryk4815:
This `cast` was a fix for issue https://github.com/pwndbg/pwndbg/issues/743
and implemented here https://github.com/pwndbg/pwndbg/pull/775

I have no idea why we still need it... 

Removing cast, fix LLDB issue remote debugging for qemu-user/gdbserver:
image
^ this invalid size will cause casting pointer into `0`
(this is bug in lldb)


patryk4815:
reported upstream - https://github.com/llvm/llvm-project/issues/153377

PR #3242, opened 2025-08-15
Add Config Option for Branch Jump Markers

gregbartell:
My terminal font doesn't have the default symbols for branches, so I took things into my own hands.

This commit adds two new config options:
- disasm-branch-on for branches that WILL be taken (default: '✔')
- disasm-branch-off for branches that will NOT be taken (default: '✘')

Happy to rename them if somebody's got a better name in mind for them.

### DIFF
--- a/docs/configuration/theme.md
+++ b/docs/configuration/theme.md
@@ -348,6 +348,28 @@ Color for disasm (branch/call instruction).
 
 ----------
 
+## **disasm-branch-off**
+
+
+Marker for branches that will NOT be taken.
+
+
+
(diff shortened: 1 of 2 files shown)


gregbartell:
I don't believe the failing test is related to my change. If I get some more time, I'll look into it further

patryk4815:
> I don't believe the failing test is related to my change. If I get some more time, I'll look into it further

@gregbartell not your fault, this test is failing randomly because of known bug in `exec_shellcode` (race condition)

disconnect3d:
Thanks!

PR #3244, opened 2025-08-16
Better Display of varargs

### LINKED ISSUE #3235: handle varargs better in function call annotation
This is what the function parameter list looks like for a vararg function:

Image

There are more arguments though!

Check if the second-to-last argument resembles a format string (has `%d`, `%s` and whatnot), and if it does, count the number of argument specifiers, and show that many parameters for the vararg.

Maybe if a vararg is encountered just show more parameters than the default (4).

akamikado:
- Displays all variadic arguments passed to a variadic function.
- Refactor pwndbg.arguments.get function

Closes: #3235


pwndbg/arguments.py
```diff
@@ -104,6 +105,23 @@ def get(instruction: PwndbgInstruction) -> List[Tuple[pwndbg.lib.functions.Argum
 
     if func:
         args = func.args
+        if args[-1].name == "vararg":
+            args.pop()
+            if len(args) > 0:
```
k4lizen:
```suggestion
            if len(args) > 1:
                args.pop()
```
Maybe like this?

pwndbg/arguments.py
```diff
@@ -104,6 +105,23 @@ def get(instruction: PwndbgInstruction) -> List[Tuple[pwndbg.lib.functions.Argum
...
+                format_value = pwndbg.enhance.enhance(argument(len(args) - 1, abi))
+                vararg_cnt = 0
+                try:
+                    vararg_cnt = int(format_value)
```
k4lizen: when does this happen :? (maybe add a comment?)

pwndbg/arguments.py
```diff
@@ -104,6 +105,23 @@ def get(instruction: PwndbgInstruction) -> List[Tuple[pwndbg.lib.functions.Argum
...
+                except ValueError:
+                    m = re.findall(
+                        r"%[-+ #0]?(?:[0-9]+|\*)?(?:\.(?:[0-9]+|\*))?(?:hh|h|ll|l|j|z|t|L)?[diuoxXfFeEgGaAcspn]",
```
k4lizen: missing a few non-standard specifiers that are mentioned in https://www.man7.org/linux/man-pages/man3/printf.3.html 
```suggestion
                        r"%[-+ #0]?(?:[0-9]+|\*)?(?:\.(?:[0-9]+|\*))?(?:hh|h|l|ll|q|L|j|z|Z|t)?[diuoxXfFeEgGaAcsCSpn]",
```

k4lizen (changes requested):
Thanks for the PR!

pwndbg/arguments.py
akamikado: I've kept this in case the last argument before variadic arguments start, indicates the count of variadic arguments
Do you think I should keep this or remove this part?

k4lizen: i would say if you know of a case where this is used like this then leave it, otherwise it's okay to remove

k4lizen:
btw i think if the (second to) last arg is not a string, or there are no format specifiers found inside the string, then the `vararg` shoudn't be popped. better to err on the side of caution and inform the user there is a vararg rather than omit information

akamikado:
Yeah that makes sense
I wanted to pop cause sometimes it might start displaying garbage value
I'll make that change

pwndbg/arguments.py
akamikado: I'll remove this part
It can be added back later if needed

akamikado: I have removed this part in the latest commit

k4lizen (approved):
LGTM

pwndbg/arguments.py
```diff
@@ -104,6 +105,19 @@ def get(instruction: PwndbgInstruction) -> List[Tuple[pwndbg.lib.functions.Argum
...
+                r"%[-+ #0]?(?:[0-9]+|\*)?(?:\.(?:[0-9]+|\*))?(?:hh|h|l|ll|q|L|j|z|Z|t)?[diuoxXfFeEgGaAcsCSpn]",
+                format_value,
+            )
+            vararg_cnt = len(m)
+            if vararg_cnt > 0:
+                args.pop()
+                args += [
+                    pwndbg.lib.functions.Argument("int", 0, argname(len(args) + i, abi))
+                    for i in range(vararg_cnt)
+                ]
```
disconnect3d, line 106: Since u are on it, can you make the following changes in this PR or in a next one, that would be just refactoring of this function?:
1) remove the `result = []` at the top and change the return to `return [(arg, argument(i, abi) for i, arg in enumarate(args)]`
2) the `if sym:` branch and logic, it seems it only makes sense on the `else:` branch below (aka when `if func:` doesn't pass) - can we move it into there?


Loud thinking: I wonder if `n_args_default` can ever be uninitialized ?

disconnect3d: Fwiw the `sym = pwndbg.aglib.symbol.lookup_frame_symbol(name)` logic should also be in the `else:`, however, we need to save the `name` before replacements for it: like `orig_name` or sth.

akamikado: I've made these changes in this PR itself
Lmk if you think any thing should be changed

disconnect3d (approved):

disconnect3d:
Thanks a lot!

disconnect3d:
Ideally, can u also add a testcase for this? I'll merge this atm, but if u could send another PR with a test, it would be awesome :)

PR #3248, opened 2025-08-18
LLDB: Print process status information on exit

mbrla0:
This PR makes it so that the Pwndbg LLDB CLI prints process status information on exit, along with any extra information LLDB provides us. It's kind of silly that this wasn't a thing up until now, but I guess using `-v` constantly made me forget at some point that this information was never given outside of verbose mode.

image

image

### DIFF
--- a/pwndbg/dbg/lldb/repl/__init__.py
+++ b/pwndbg/dbg/lldb/repl/__init__.py
@@ -108,7 +108,7 @@ def print_hint(msg: str, *args):
 
 def print_info(msg: str, *args):
     """
-    Print an informative message in the style of the LLDB CLI.
+    Print an information message in the style of the LLDB CLI.
     """
     print(message.info("info:"), msg, *args)
 
--- a/pwndbg/dbg/lldb/repl/__init__.py
+++ b/pwndbg/dbg/lldb/repl/__init__.py
@@ -71,12 +71,6 @@ from pwndbg.dbg.lldb.pset import pget
 from pwndbg.dbg.lldb.pset import pset
 from pwndbg.dbg.lldb.repl.io import IODriver
 from pwndbg.dbg.lldb.repl.io import get_io_driver
-from pwndbg.dbg.lldb.repl.proc import EventHandler
-from pwndbg.dbg.lldb.repl.proc import LaunchResultConnected
-from pwndbg.dbg.lldb.repl.proc import LaunchResultEarlyExit
-from pwndbg.dbg.lldb.repl.proc import LaunchResultError
-from pwndbg.dbg.lldb.repl.proc import LaunchResultSuccess
-from pwndbg.dbg.lldb.repl.proc import ProcessDriver
 from pwndbg.lib.tips import color_tip
 from pwndbg.lib.tips import get_tip_of_the_day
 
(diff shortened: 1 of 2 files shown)


pwndbg/dbg/lldb/repl/__init__.py
```diff
@@ -90,6 +84,42 @@
...
+    Print a hint message in the style of the LLDB CLI.
+    """
+    print(message.hint("hint:"), msg, *args)
+
+
+def print_info(msg: str, *args):
+    """
+    Print an informative message in the style of the LLDB CLI.
+    """
+    print(message.info("info:"), msg, *args)
```
disconnect3d, line 113: Those functions are kinda meh/boilerplate-y like :(

pwndbg/dbg/lldb/repl/__init__.py
```diff
@@ -90,6 +84,42 @@
...
+def print_hint(msg: str, *args):
+    """
+    Print a hint message in the style of the LLDB CLI.
+    """
+    print(message.hint("hint:"), msg, *args)
+
+
+def print_info(msg: str, *args):
+    """
+    Print an informative message in the style of the LLDB CLI.
```
disconnect3d:
```suggestion
    Print an information message in the style of the LLDB CLI.
```

pwndbg/dbg/lldb/repl/__init__.py
disconnect3d: if we impl it, do we want it also for gdb?

disconnect3d:
Hmm... on the screenshot it says "coz its not yet safe to do so, and will be tried later" - what does this mean? who retries it? etc.

Also: should we maybe add info such as process pid when we say it detached etc?

patryk4815:
> Hmm... on the screenshot it says "coz its not yet safe to do so, and will be tried later" - what does this mean? who retries it? etc.

I think is it direct message from LLDB, not ours

pwndbg/dbg/lldb/repl/__init__.py
```diff
@@ -71,12 +71,6 @@
 from pwndbg.dbg.lldb.pset import pset
 from pwndbg.dbg.lldb.repl.io import IODriver
 from pwndbg.dbg.lldb.repl.io import get_io_driver
-from pwndbg.dbg.lldb.repl.proc import EventHandler
```
patryk4815, line 74: 🤔 why we need move around imports?

pwndbg/dbg/lldb/repl/__init__.py
mbrla0: > Those functions are kinda meh/boilerplate-y like

Perhaps, but both the warn and error message kinds are used so often in the CLI that it made sense to split them into a function of their own. The other two aren't used as often, but at least it helps keep the messaging style consistent.

> if we impl it, do we want it also for gdb?

Not really, those are really just for the Pwndbg LLDB CLI, and they're meant to have the look and feel of regular LLDB. It's a way to make `pwndbg-lldb` feel less like it's doing everything on its own and more like it's just plugging into the regular LLDB CLI.

PR #3249, opened 2025-08-18
Add initial support for Objective-C and Mach-O on Darwin

mbrla0:
Implement support parsing of some aspects of the Objective-C and Mach-O ABIs to `aglib`.

These will serve as the basis for future PRs refining the display of the vmmap on Darwin platforms, as well as method call resolution for Objective-C calls.

### DIFF
--- a/pwndbg/aglib/macho.py
+++ b/pwndbg/aglib/macho.py
@@ -1,9 +1,160 @@
 from __future__ import annotations
 
+import struct
+from typing import Callable
 from typing import Generator
+from typing import Generic
+from typing import Tuple
+from typing import TypeVar
 
 import pwndbg
 import pwndbg.aglib.memory
+
+
+def _uleb128(ptr: int) -> Tuple[int, int]:
+    """
+    Decode a ULEB128 value at the start of the given address, and return the
+    decoded number, along with how many bytes the entire number takes.
+    """
+    acc = 0
+    i = 0
+    while True:
+        byte = pwndbg.aglib.memory.u8(ptr + i)
+
+        acc |= (byte & 0x7F) << (7 * i)
+        if byte & 0x80 == 0:
+            # This is the terminator byte.
+            break
+
+        i += 1
+
+    return acc, i + 1
+
+
+class _RawTrie:
+    """
+    This is the untyped base implementation of Trie.
+    """
+
+    def __init__(self, ptr: int):
+        self._ptr = ptr
+
+    def _walk(
+        self,
+        offset: int,
+        acc: bytes,
+        edgesel: Callable[[bytes, bytes], bool],
+        nodesel: Callable[[bytes], bool],
+    ) -> Generator[Tuple[bytes, int, int]]:
+        """
+        Walk the trie.
+
+        Allows callers to select edges for exploration and nodes for yielding
+        through the `edgesel` and `nodesel` callables.
+
+        At every edge, this function will call `edgesel` with the currently
+        accumulated name and the name associated with the edge, and will take
+        action according to the value it returns. If it returns True, that edge
+        will be explored, otherwise, the edge will be ignored.
+
+        At every node, this function will call `nodesel` with the currently
+        accumulated name. If it returns True, the node will be yielded,
+        otherwise, it will be ignored.
+
+        Yielded node information consists of a tuple of (name, ptr, length),
+        where `name` is the name of the node, `ptr` is the address of the first
+        byte of its associated data, and `length` is the length of its
+        associated data, in bytes.
+        """
+        base = self._ptr + offset
+
+        node_data_len, node_data_len_len = _uleb128(base)
+        if node_data_len != 0 and nodesel(acc):
+            # The user selected this node, stop the walk here.
+            yield acc, base + node_data_len_len, node_data_len
+
+        cursor = base + node_data_len_len + node_data_len
+
+        # The number of children is NOT a ULEB128.
+        children = pwndbg.aglib.memory.u8(cursor)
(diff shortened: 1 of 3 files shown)


pwndbg/aglib/macho.py
```diff
@@ -0,0 +1,667 @@
...
+            + (blob[19] << 24)
+            + (blob[20] << 32)
+            + (blob[21] << 40)
+            + (blob[22] << 48)
+            + (blob[23] << 56)
+        )
+
+        a %= 0x10000000000000000
+        b %= 0x10000000000000000
+        c %= 0x10000000000000000
```
disconnect3d, line 251: Do we need `a, b, c` to be u64?

Shall we maybe just use `a = ctypes.c_uint64(level)` and `a.value` for operations?

We could also use `cint` library, huh.

Otoh maybe `%=` is fine?

At the same time, `%=` is probably much slower than `&= 0xff...ff`, so maybe just and operation?

pwndbg/aglib/macho.py
```diff
@@ -0,0 +1,667 @@
...
+        return pwndbg.aglib.memory.u32(self._ptr + 0x20 + index * 4)
+
+    def _tab(self, index: int) -> int:
+        assert index & ~self.mask == 0
+        return pwndbg.aglib.memory.u8(self._ptr + 0x420 + index)
+
+    def _index_of(self, key: bytes) -> int:
+        lookup = _lookup8(key, self.salt)
+        return ((lookup >> self.shift) % 0x100000000) ^ self._scramble(
+            self._tab(lookup & self.mask)
```
disconnect3d: Since scramble and tab are only used here, shall we just do:
```py
assert index < 256
scramble = pwndbg.aglib.memory.u32(self._ptr + 0x20 + index * 4)

assert index & ~self.mask == 0
tab = pwndbg.aglib.memory.u8(self._ptr + 0x420 + index)
```
and use the values instead of calling methods?

pwndbg/aglib/macho.py
```diff
@@ -0,0 +1,667 @@
...
+    def __init__(self, addr: int):
+        self.addr = addr
+
+        # Preload a few a few values, to speed things up later.
+        if self._header_size() <= 0x1C4:
+            self._images_base = self.addr + pwndbg.aglib.memory.u32(self.addr + 0x18)
+            self.image_count = pwndbg.aglib.memory.u32(self.addr + 0x1C)
+        else:
+            self._images_base = self.addr + pwndbg.aglib.memory.u32(self.addr + 0x1C0)
+            self.image_count = pwndbg.aglib.memory.u32(self.addr + 0x1C4)
```
disconnect3d: Shall we maybe do:
```py
some_offset = 0x18 if self._header_size() <= 0x1c4 else 0x1c0
self._images_base = self.addr + pwndbg.aglib.memory.u32(self.addr + some_offset)
self.image_count = pwndbg.aglib.memory.u32(self.addr + some_offset + 4)
```
?

pwndbg/aglib/macho.py
mbrla0: > Do we need `a, b, c` to be u64?

Yes, the hash algorithm makes heavy use of u64 wraparound subtractions. Those three statements specifically aren't that important, but the ones like them in `_mix64` are essential.

> Shall we maybe just use `a = ctypes.c_uint64(level)` and `a.value` for operations? [...]

In CPython 3.13, operating on `c_uint64(_).level` is even more involved than operating on `int`'s directly. `level` is [implemented as a getset](https://github.com/python/cpython/blob/e39255e76d4b6755a44f6d4e63180136c778d2a5/Modules/_ctypes/_ctypes.c#L5538) which just [converts the value in the `c_uint64` to and from `PyLong`](https://github.com/python/cpython/blob/e39255e76d4b6755a44f6d4e63180136c778d2a5/Modules/_ctypes/cfield.c#L646), meaning we'll still be operating on `int`s, but with extra steps.
 
> Otoh maybe `%=` is fine?
> 
> At the same time, `%=` is probably much slower than `&= 0xff...ff`, so maybe just and operation?

That's sensible to assume, but looking at the implementations for [`&`](https://github.com/python/cpython/blob/e39255e76d4b6755a44f6d4e63180136c778d2a5/Objects/longobject.c#L5559) and [`%`](https://github.com/python/cpython/blob/e39255e76d4b6755a44f6d4e63180136c778d2a5/Objects/longobject.c#L4537) on `PyLong`, it really doesn't look like we'd see a world of difference switching from one to the other. Also, we wouldn't be gaining anything by keeping the divisor `n <= UINT64_MAX`, since [`PyLong`s store their values in groups of 30 bits](https://github.com/python/cpython/blob/e39255e76d4b6755a44f6d4e63180136c778d2a5/Include/cpython/longintrepr.h#L64), and both 64 and 65 bit numbers need three of those groups.

pwndbg/aglib/macho.py
mbrla0: Well, yeah, I think those warrant getting switched to variables.

Honestly, there are plenty of opportunities to change methods for instance variables in some of these classes, but since as-is all the code dealing with the Objective-C ABI is already quite hard to read, from trying to keep up with all the weird bespoke structures Apple uses, I figured I'd try going with what I feel is easiest to read over performance, if performance is otherwise good enough.

In this case, though, it absolutely makes sense to switch, since this is a decently high-volume operation, and individual memory reads in LLDB can be quite expensive.

pwndbg/aglib/macho.py
mbrla0: Yeah, that works.

PR #3250, opened 2025-08-18
Document removal of quarantine flag during setup of portable version for macOS users

### LINKED ISSUE #2689: Add note on macOS installation instructions regarding quarantine
When one download the tarball from releases on macOS (15.0.1, Intel here) with a browser and have the security settings set to `Allow applications from: App Store & Known Developers` they will get lots and lots of popups about untrusted binaries and libraries when trying to run. This is also showing the following in the CLI:

```
➜  bin ./pwndbg-lldb
Could not find the LLDB Python Path: b"dyld[46108]: Library not loaded: @executable_path/../lib/libLLVM.dylib\n  Referenced from: <2C0DE2E9-7058-3F2C-88A6-E09AC16244D2> /Users/user/pwndbg/exe/lldb\n  Reason: tried: '/Users/user/pwndbg/lib/libLLVM.dylib' (code signature in <B1323D6B-0167-340C-A432-A2496C26CCA6> '/Users/user/pwndbg/lib/libLLVM.dylib' not valid for use in process: library load disallowed by system policy), '/usr/local/lib/libLLVM.dylib' (no such file), '/usr/lib/libLLVM.dylib' (no such file, not in dyld cache)\n"
```

This is because there is a quarantine attribute added to the unpacked files when the archive was downloaded with a web browser.

This can be fixed by doing `xattr -rd com.apple.quarantine pwndbg` -- and we should probably note this in a "README" file that we would add to the archive (maybe also in GH readme, but not sure if needed there).

mbrla0:
Fixes #2689

This PR adds a section to the setup guide detailing why and how a user can remove the quarantine flag from the portable version of Pwndbg on macOS.

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -104,7 +104,7 @@ one to be run can get quite tiresome.
 In order to do this to all files at once, you may choose to run the following command, which removes
 the quarantine flag from all extracted files at once:
 
-```
+```{.bash .copy}
 xattr -rd com.apple.quarantine pwndbg
 ```
 
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -94,5 +94,21 @@ or
 ```
 depending on which version you installed. You may add the appropriate file to your shell's PATH.
 
+### Removing Quarantine Flags (macOS)
+
+When first setting up the portable version of Pwndbg in macOS, Gatekeeper will normally try to prevent
+any code in the extracted files from running until the user explicitly allows each file to be run.
+As we ship many files which would require this, the process of manually granting permission for each
+one to be run can get quite tiresome.
+
+In order to do this to all files at once, you may choose to run the following command, which removes
+the quarantine flag from all extracted files at once:
+
+```
+xattr -rd com.apple.quarantine pwndbg
+```
+
+Assuming that the files were extracted to a folder called `pwndbg`.
+
 ## Installing from source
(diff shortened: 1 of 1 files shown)


docs/setup.md
```diff
@@ -94,5 +94,21 @@ or
...
+
+When first setting up the portable version of Pwndbg in macOS, Gatekeeper will normally try to prevent
+any code in the extracted files from running until the user explicitly allows each file to be run.
+As we ship many files which would require this, the process of manually granting permission for each
+one to be run can get quite tiresome.
+
+In order to do this to all files at once, you may choose to run the following command, which removes
+the quarantine flag from all extracted files at once:
+
+```
```
k4lizen:
````suggestion
```{.bash .copy}
````
gives it syntax highlighting, and enables a button to click to copy to clipboard :>

k4lizen (approved):

patryk4815:
Can we show proper message in our pwndbg-lldb wrapper 🤔  (without docs)
if xattr something ./exe/lldb -> show message?

mbrla0:
Yeah, it's a little more involved, since we'd have to at least keep track of the paths some of the files we'd want to check. But, ultimately, we could do that too, I don't really have much of a preference either way.

disconnect3d:
> Can we show proper message in our pwndbg-lldb wrapper 🤔 (without docs) if xattr something ./exe/lldb -> show message?

Pls change the docs if we eventually merge that wrapper change :P

PR #3255, opened 2025-08-20
Refine `vmmap` output on Darwin

mbrla0:
This PR makes it so that virtual memory mapping information on Darwin platforms takes the DYLD Shared Cache into account, though a few changes to Pwndbg.

The first change is to the LLDB backend of the Debugger API. Normally, LLDB will yield separate ranges for different sections in Mach-O binaries[^1]. These sections are - as far as Pwndbg is concerned - identical and contiguous in virtual memory. So, this PR makes it so that the LLDB backend will join these mappings together into a single one, whenever possible.

[^1]: For example, it may emit one mapping for `__TEXT,__text` and another for `__TEXT,__stubs`, even when these are contiguous in virtual memory and have the same mapping flags.

The second change is to make  `pwndbg.aglib.vmmap.get_memory_map()` uses Shared Cache information to correctly populate the virtual memory map with the names and memory ranges of different system libraries in the cache, by dividing mappings that span over multiple libraries into multiple mappings, one for each library.

The third change is to have the `vmmap` command omit by default entries in the Shared Cache.

### DIFF
--- a/pwndbg/aglib/vmmap.py
+++ b/pwndbg/aglib/vmmap.py
@@ -1,12 +1,16 @@
 from __future__ import annotations
 
+import bisect
 from typing import Tuple
 
 import pwndbg
+import pwndbg.aglib.arch
 import pwndbg.aglib.vmmap_custom
 import pwndbg.lib.cache
 import pwndbg.lib.memory
 from pwndbg.dbg import MemoryMap
+from pwndbg.lib.arch import Platform
+from pwndbg.lib.memory import Page
 
 pwndbg.config.add_param(
     "vmmap-prefer-relpaths",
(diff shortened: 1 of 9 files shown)


pwndbg/aglib/vmmap.py
```diff
@@ -16,9 +20,86 @@
 )
 
 
+def _refine_memory_map(pages: MemoryMap) -> MemoryMap:
+    if (
+        pwndbg.aglib.arch.platform == Platform.DARWIN
+        and pwndbg.aglib.macho.shared_cache() is not None
+    ):
```
patryk4815:
```suggestion
def _refine_memory_map(pages: MemoryMap) -> MemoryMap:
    if not (
        pwndbg.aglib.arch.platform == Platform.DARWIN
        and pwndbg.aglib.macho.shared_cache() is not None
    ):
        return pages
```

patryk4815: I think it is better to have less nested ifs

pwndbg/aglib/vmmap.py
```diff
@@ -16,9 +20,86 @@
 )
 
 
+def _refine_memory_map(pages: MemoryMap) -> MemoryMap:
```
patryk4815, line 23: this should be implemented in `vmmap()` in `gdb/__init__` and `lldb/__init__`, not here?

mbrla0: I was honestly torn on whether to implement it as part of the back-ends or as part of `aglib`, but ultimately settled on the latter. 

I feel it makes more sense to have it there, as that operation is - in its essence - debugger-agnostic, even if in practice we'll only ever trigger it from LLDB. Contrast that with `LLDBProcess._process_vmmap_pages`, which _does_ go into the LLDB backend, but whose purpose is to help make the output of LLDB specifically look more sensible to the rest of Pwndbg.

Untimately, though, I don't have strong feelings for where it ends up going in the end, but I feel explaining my reasoning here should make things clearer.

patryk4815: yes, but calling `pwndbg.dbg.selected_inferior().vmmap()` - now return different things, we should avoid that 🤔?

patryk4815: I mean, people right now are confused what to use:
`pwndbg.dbg.selected_inferior().vmmap()` vs `pwndbg.aglib.vmmp.get()`

mbrla0: > yes, but calling `pwndbg.dbg.selected_inferior().vmmap()` - now return different things, we should avoid that 🤔?

Yes, but at the same time there were very few direct calls to `.vmmap()` on the inferior that I could find, all of which the PR changes to `aglib` calls. And I feel that is still consistent with the role of other `aglib` interfaces, that provide enhanced operations built on top of the more primitive ones offered by the Debugger API.

mbrla0: > I mean, people right now are confused what to use: `pwndbg.dbg.selected_inferior().vmmap()` vs `pwndbg.aglib.vmmp.get()`

Yes, and that is a genuine issue that we should find a way to address, I just think explaining the differences between these interfaces better should be good enough? At least, when compared to entirely unifying them.

disconnect3d:
@mbrla0 needs a conflict fix :(

patryk4815:
image

@mbrla0 can we display this as:
`0x1818d0000       0x2ae634000 ---p  12cd64000       0 [DYLD Shared Cache]`
?

mbrla0:
> @mbrla0 can we display this as: `0x1818d0000 0x2ae634000 ---p 12cd64000 0 [DYLD Shared Cache]` ?

Yeah, sure thing!

mbrla0:
Is this better?

image

disconnect3d:
> Is this better?

I'd say its fine; though maybe the size column should be longer? Or, for such big size we should say "X.Y MB"? 

But tbh its rather good enough to merge now

PR #3259, opened 2025-08-21
Add max-decimal-number config

k4lizen:
We had multiple places in code that did a
```python
        if abs(value) < 0x10:
            return "%i" % value

        return "%#x" % value
```
type of thing, and different cut-off values are being used including 10, 16 and 255. This PR centralizes this logic to one place (the pretty_printer) and puts it behind a configuration option.

Setting this config param to unlimited and 0 doesn't work until https://github.com/pwndbg/pwndbg/pull/3258 is merged.


k4lizen:
as a side-note, on one hand it seems many `pwndbg.lib*` modules try not to `import pwndbg`, but some do, so I'm not sure what's the policy/intent here.

k4lizen:
@OBarronCS so the tests are failing in `mov rax, 0xa`-type places. from what I can see the `rax, 0xa` comes from 
```python
        self.op_str: str = cs_insn.op_str
```
capstone, and we do not modify it, do I understand this correctly?

The implication being that this `max-decimal-number` config will affect stuff like enhancements but there is no way for it to affect the actual disassembly. Agreed?

If that is the case i'll document it and set the default `max-decimal-number` value to align with capstone, which would be the value 9 afaict.

OBarronCS:
Yes - we are directly printing the assembly mnemonic and operands as Capstone provides them. We would need to do a "find and replace" if we wanted this to propagate to the assembly operands. There is one place in the code where we do this sort of find-and-replace in a different context, which is to substitute constant values with their corresponding symbols:

https://github.com/pwndbg/pwndbg/blob/26db4533aa08d77c4bbc359b4760a0944e0c6b23/pwndbg/aglib/disasm/arch.py#L322-L331

Tested it a bit - Capstone does have `9` as the maximum decimal value - documenting and making this uniform is great.

k4lizen:
replacing the numbers in the string feels a bit finicky so I'll leave that out of this PR

k4lizen:
should be good for review

disconnect3d:
Conflict:(

k4lizen:
fixed the conflict and added a drive-by formatting fix from the pretty-printer PR. though tbh while it is less lines of code, personally not sure if it actually improves readability

PR #3260, opened 2025-08-21
Resolve arguments for Objective-C method calls

mbrla0:
This PR makes it so that `pwndbg.aglib.arguments.get` is able to resolve the arguments to Objective-C method calls made through `objc_msgSend`. We do that by parsing reflection information baked into the binary by a compiler that complies with the Darwin Objective-C ABI.

One important thing to note about Objective-C reflection information is that their being correct is not required for the program itself to function correctly. This means that programs can lie about that information as much as they wish, and that can pose a problem for us, since we can't entirely trust these fields.

With that in mind, by default we impose limits on both the depth of recursive types - pointers, arrays, and co. - and on the number of arguments we can resolve. We also won't try *too* hard to make sense of a signature that looks too weird.

### DIFF
--- a/pwndbg/arguments.py
+++ b/pwndbg/arguments.py
@@ -83,7 +83,17 @@ def get(instruction: PwndbgInstruction) -> List[Tuple[pwndbg.lib.functions.Argum
         name = name.replace("_chk", "")
         name = name.strip().lstrip("_")  # _malloc
 
-    func = pwndbg.lib.functions.functions.get(name, None)
+    # Try to resolve an Objective-C method call.
+    #
+    # Checking this first keeps us from resolving these as simple calls to
+    # `objc_msgSend` and functions like it, which have definitions that are
+    # rather barren of semantics in comparison.
+    func = pwndbg.aglib.objc.try_resolve_call_at_current_pc(instruction)
+
+    if func is None:
+        # If more specific call information can't be determined, use the regular
+        # function resolution flow.
+        func = pwndbg.lib.functions.functions.get(name, None)
 
     # Try to grab the data out of IDA
     if not func and target:
--- a/pwndbg/arguments.py
+++ b/pwndbg/arguments.py
@@ -29,6 +29,7 @@ import pwndbg.lib.funcparser
 import pwndbg.lib.functions
 from pwndbg.aglib.disasm.instruction import PwndbgInstruction
 from pwndbg.aglib.nearpc import c as N
+from pwndbg.lib.arch import Platform
 from pwndbg.lib.functions import format_flags_argument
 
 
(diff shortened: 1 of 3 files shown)


mbrla0:
Also, one important limitation of this PR is that, while it always indicates to the selector in the call, `enhance` doesn't always display the selector string itself. But that can be fixed in another PR.

pwndbg/arguments.py
```diff
@@ -83,7 +83,17 @@ def get(instruction: PwndbgInstruction) -> List[Tuple[pwndbg.lib.functions.Argum
         name = name.replace("_chk", "")
         name = name.strip().lstrip("_")  # _malloc
 
-    func = pwndbg.lib.functions.functions.get(name, None)
+    # Try to resolve an Objective-C method call.
+    #
+    # Checking this first keeps us from resolving these as simple calls to
+    # `objc_msgSend` and functions like it, which have definitions that are
+    # rather barren of semantics in comparison.
+    func = pwndbg.aglib.objc.try_resolve_call_at_current_pc(instruction)
```
disconnect3d: I am not sure I like this.

Can we somehow detect obj-c language? Does it show with `show language`? Maybe GDB does auto-detect it? (Does LLDB do so? :/)

patryk4815: image

disconnect3d: Alternatively, we could do this code only when running on macOS/Darwin? Does Obj-C code ever run on Linux?:D

patryk4815: yes you can run obj-c on linux

mbrla0: > yes you can run obj-c on linux

Honestly I don't think that matters much? `pwndbg.aglib.objc` is currently only concerned with the Darwin version of the ABI, which isn't exactly binary compatible with GNUStep, afaik.

disconnect3d: So, any better ideas how to handle this except of checking for lang? :(

disconnect3d: Or, we can leave it as it is for now

disconnect3d:
@mbrla0 can u fix the doc conflict?

pwndbg/arguments.py
patryk4815: we should add language checking before executing `pwndbg.aglib.objc.*` code

mbrla0:
> @mbrla0 can u fix the doc conflict?

I'm taking a little break, in Europe attending Eurofurence. As soon as I'm back home by next-ish week I'll fix it.

patryk4815 (changes requested):
@mbrla0 Can you just add `if pwndbg.aglib.arch.platform == Darwin` and/or language check == ObjC?

patryk4815:
Before:
image

After:
image

PR #3266, opened 2025-08-24
fix: handle quarantine flag during portable startup on macOS

patryk4815:
Fixes:
- https://github.com/pwndbg/pwndbg/issues/2689

image

### DIFF
--- a/nix/portable.nix
+++ b/nix/portable.nix
@@ -55,12 +55,28 @@ let
       export PATH="$dir/bin/:$PATH"
     '';
 
+  macosQuarantine = lib.optionalString pkgs.stdenv.isDarwin ''
+    libpython="libpython${python3.pythonVersion}.dylib"
+
+    if command -v xattr >/dev/null 2>&1 && command -v grep >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


patryk4815:
@mbrla0 ok?

nix/portable.nix
```diff
@@ -55,12 +55,28 @@ let
...
+  '';
+
   wrapperBinPy =
     file:
     pkgs.writeScript "pwndbg-wrapper-bin-py" ''
       #!/bin/sh
       dir="$(cd -- "$(dirname "$(dirname "$(realpath "$0")")")" >/dev/null 2>&1 ; pwd -P)"
       ${commonEnvs}
+      ${macosQuarantine}
       exec ${ldLoader} "$dir/exe/python3" "$dir/${file}" "$@"
```
disconnect3d, line 80: Shouldn't we check for quarantine only if we fail to run Pwndbg?

disconnect3d (approved):

nix/portable.nix
patryk4815: with `exec` it is not possible to catch exit code

PR #3268, opened 2025-08-25
Added shortcuts when ipi is launched

### LINKED ISSUE #3205: Entering ipi should provide some nice aliases
On my local machine, I typically have an `alias i=ipython` along with an ipython startup script that imports useful modules and sets up some functions I use from time to time:

Image

I think it would be nice to have the same thing here, so entering `ipi` should provide us some useful aliases for often used functions like:
- memory and registers read and write
- getting vmmaps
- memory search, memory hexdump

At the same time, maybe this is a time to define some nicer user-facing API, with shorter names?

Rishikesh-Reddy:
Fixes: #3205

Added shortcuts when launching ipi command.


disconnect3d (changes requested):
Honestly, I don't like this overall: we probably need to inform better about the shortcuts OR have docstrings that can be found out with `<funcname>?` in the IPython shell.

Also, why `r` and `w` this way? Is this to accept an address as string or int? That's neat.

pwndbg/commands/ipython_interactive.py
```diff
@@ -46,7 +46,208 @@ def ipi() -> None:
...
+from pwndbg.aglib.memory import read, write
+import pwndbg.aglib.vmmap as vmmap
+from pwndbg.commands.hexdump import hexdump
+from pwndbg.commands.search import search
+
+
+@wraps(read)
+def r(addr, length):
+    '''
+    Wrapper alias for pwndbg.read with input validation.
```
disconnect3d, line 62: 1) It's pwndbg.aglib.memory.read not pwndbg.read
2) Why [functools.wraps](https://docs.python.org/3/library/functools.html#functools.update_wrapper)? We don't need the function to look like the read function, we need it either to be just the read function, or, some fancy wrapper.

pwndbg/commands/ipython_interactive.py
```diff
@@ -46,7 +46,208 @@ def ipi() -> None:
...
+        Prints warnings or errors if the memory cannot be accessed.
+
+    Notes:
+        - Results are paged based on pwndbg\'s config options:
+          * `hexdump-width`: number of bytes shown per line.
+          * `hexdump-group-width`: number of bytes per grouping.
+          * `hexdump-group-use-big-endian`: flips group display endianness.
+          * `hexdump-limit-mb`: maximum allowed dump size.
+        - `hexdump.last_address` and `hexdump.offset` are updated internally
+          to support consecutive calls.
```
disconnect3d, line 151: Seriously? Nobody will want to read such long description. Please think twice when generating things with LLMs

pwndbg/commands/ipython_interactive.py
```diff
@@ -46,7 +46,208 @@ def ipi() -> None:
...
+
+    Returns the virtual memory map for the current process.
+
+    Returns:
+        Typically a list of memory mapping entries as provided by pwndbg.vmmap.get.
+
+    Raises:
+        None.
+    '''
+    return vmmap.get()
```
disconnect3d, line 174: This is awful. You know this could be `from pwndbg.aglib.vmmap import get as vv`? lol...

disconnect3d:
Sorry, but this PR is just a waste of my time. Please don't just generate things with LLMs but think twice how people would use things.

Issue #3170, opened 2025-07-23
consider cropping the ghidra context section


k4lizen:
### Description

Currently, the ghidra decompilation we show with `ctx ghidra` shows the decompilation of the whole function. See example for /bin/sh's main:

<img width="395" height="1041" alt="Image" src="https://github.com/user-attachments/assets/f15c2997-ec92-425c-b99f-4ecbd6aac8c1" />

I can't zoom my terminal out more to show the whole thing, but imagine that, times four, and then above that your `ctx regs` is shown. Pretty unusable :D 

### Steps to reproduce

Use the [ghidra integration](https://pwndbg.re/pwndbg/dev/tutorials/decompiler-integration/ghidra-integration/) inside /bin/sh's main function.

### My setup

Pwndbg:   2025.05.30 build: c8ee6809c
Python:   3.13.5
GDB:      16.3
Capstone: 6.0.0
Unicorn:  2.1.3
Pwnlib:   4.14.0

disconnect3d:
Should be fixed by https://github.com/pwndbg/pwndbg/pull/3269 :)

PR #3269, opened 2025-08-25
Improving ghidra decompilation support

### LINKED ISSUE #3134: Supports for decompiling kernel functions
Currently, decompiling kernel functions with the `ghidra` command is not supported. Supporting it would be useful when source code of a kernel module or custom syscall is not provided.

jxuanli:
closes #3134 and #3170

## demo

image

### DIFF
--- a/pwndbg/aglib/elf.py
+++ b/pwndbg/aglib/elf.py
@@ -202,7 +202,7 @@ def get_containing_sections(elf_filepath: str, elf_loadaddr: int, vaddr: int):
     return sections
 
 
-def get_elf_base(elf_filepath: str):
+def get_vmlinux_unrand_base(elf_filepath: str):
     elf = get_elf_info(elf_filepath)
     for seg in elf.segments:
         if seg["p_type"] == "PT_LOAD":
--- a/pwndbg/aglib/elf.py
+++ b/pwndbg/aglib/elf.py
@@ -202,6 +202,14 @@ def get_containing_sections(elf_filepath: str, elf_loadaddr: int, vaddr: int):
     return sections
 
 
+def get_elf_base(elf_filepath: str):
+    elf = get_elf_info(elf_filepath)
+    for seg in elf.segments:
+        if seg["p_type"] == "PT_LOAD":
+            return seg["p_vaddr"]
+    return None
+
+
 def dump_section_by_name(
     filepath: str, section_name: str, try_local_path: bool = False
 ) -> Tuple[int, int, bytes] | None:
--- a/pwndbg/aglib/file.py
+++ b/pwndbg/aglib/file.py
@@ -90,7 +90,7 @@ def get_file(path: str, try_local_path: bool = False) -> str:
         path = path[7:]  # len('target:') == 7
 
     local_path = path
-    if not pwndbg.aglib.remote.is_remote():
+    if not pwndbg.aglib.remote.is_remote() or pwndbg.aglib.qemu.is_qemu_kernel():
         if not os.path.exists(local_path):
             raise OSError(f"File '{local_path}' does not exist", errno.ENOENT)
 
--- a/pwndbg/ghidra.py
+++ b/pwndbg/ghidra.py
@@ -17,7 +17,6 @@ import pwndbg.rizin
 if pwndbg.dbg.is_gdblib_available():
     import pwndbg.gdblib.symbol
 
-from pwndbg.color import message
 
 decompiler = pwndbg.config.add_param(
     "decompiler",
@@ -26,18 +25,6 @@ decompiler = pwndbg.config.add_param(
     param_class=pwndbg.lib.config.PARAM_ENUM,
     enum_sequence=["radare2", "rizin"],
 )
-
-
-@pwndbg.config.trigger(decompiler)
-def set_decompiler() -> None:
-    if decompiler.value in ["radare2", "rizin"]:
-        return
-    print(
-        message.warn(
-            f"Invalid decompiler: `{decompiler.value}`, please select from radare2 or rizin"
-        )
-    )
-    decompiler.revert_default()
 
 
 def decompile(func=None):
--- a/pwndbg/ghidra.py
+++ b/pwndbg/ghidra.py
@@ -49,6 +49,7 @@ def decompile(func=None):
 
     Raises Exception if any fatal error occurs.
     """
+    func_specified = func is not None
     try:
         if decompiler == "radare2":
             r = pwndbg.radare2.r2pipe()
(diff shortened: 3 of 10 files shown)


pwndbg/aglib/elf.py
```diff
@@ -202,6 +202,14 @@ def get_containing_sections(elf_filepath: str, elf_loadaddr: int, vaddr: int):
     return sections
 
 
+def get_elf_base(elf_filepath: str):
+    elf = get_elf_info(elf_filepath)
+    for seg in elf.segments:
+        if seg["p_type"] == "PT_LOAD":
+            return seg["p_vaddr"]
```
disconnect3d, line 209: This should look for segment with offset=0, right?

disconnect3d, line 209: Shouldn't this maybe check for offset=0 as well?

U can have a binary where first PT_LOAD is not offset=0, like this:

```asm
BITS 64
  org 0x400000

ehdr:           ; Elf64_Ehdr
  db 0x7f, "ELF", 2, 1, 1, 0 ; e_ident
  times 8 db 0
  dw  2         ; e_type
  dw  0x3e      ; e_machine
  dd  1         ; e_version
  dq  _start    ; e_entry
  dq  phdr - $$ ; e_phoff
  dq  0         ; e_shoff
  dd  0         ; e_flags
  dw  ehdrsize  ; e_ehsize
  dw  phdrsize  ; e_phentsize
  dw  2         ; e_phnum (changed from 1 to 2)
  dw  0         ; e_shentsize
  dw  0         ; e_shnum
  dw  0         ; e_shstrndx
ehdrsize  equ  $ - ehdr

phdr:           ; First Elf64_Phdr (new one with offset=1)
  dd  1         ; p_type
  dd  5         ; p_flags
  dq  1         ; p_offset (changed from 0 to 1)
  dq  $$+1      ; p_vaddr (adjusted for offset=1)
  dq  $$+1      ; p_paddr (adjusted for offset=1)
  dq  filesize-1; p_filesz (adjusted for offset=1)
  dq  filesize-1; p_memsz (adjusted for offset=1)
  dq  0x1000    ; p_align
phdrsize  equ  $ - phdr

phdr2:          ; Second Elf64_Phdr (original one with offset=0)
  dd  1         ; p_type
  dd  5         ; p_flags
  dq  0         ; p_offset (original offset=0)
  dq  $$        ; p_vaddr
  dq  $$        ; p_paddr
  dq  filesize  ; p_filesz
  dq  filesize  ; p_memsz
  dq  0x1000    ; p_align

_start:
  mov rax, 1
  mov rdi, 1
  mov rsi, message
  mov rdx, 13
  syscall
  mov rax, 60
  xor rdi, rdi
  syscall

message: db "hello, world", 0xa

filesize  equ  $ - $$
```

The binary still works:
```
$ nasm -f bin hello.asm -o hello.out && chmod +x hello.out && ./hello.out
hello, world
```

And its sections are:
```
$ readelf --segments hello.out

Elf file type is EXEC (Executable file)
Entry point 0x4000b0
There are 2 program headers, starting at offset 64

Program Headers:
  Type           Offset             VirtAddr           PhysAddr
                 FileSiz            MemSiz              Flags  Align
  LOAD           0x0000000000000001 0x0000000000400001 0x0000000000400001
                 0x00000000000000e1 0x00000000000000e1  R E    0x1000
  LOAD           0x0000000000000000 0x0000000000400000 0x0000000000400000
                 0x00000000000000e2 0x00000000000000e2  R E    0x1000
```

disconnect3d: On the other hand: can we have a binary with a LOAD segment with offset != 0 that still works? I am not sure, but probably yes? lol

pwndbg/ghidra.py
```diff
@@ -19,25 +19,25 @@
...
+    if decompiler.value in ["radare2", "rizin"]:
         return
     print(
         message.warn(
-            f"Invalid r2decompiler: `{r2decompiler.value}`, please select from radare2 or rizin"
+            f"Invalid decompiler: `{decompiler.value}`, please select from radare2 or rizin"
         )
     )
-    r2decompiler.revert_default()
+    decompiler.revert_default()
```
disconnect3d: Btw is this trigger needed at all considering that we have a PARAM_ENUM and `enum_sequence=["radare2", "rizin"],`? Isn't this config param providing such handling by itself? (I don't know on top of my head, but I believe it rather should :P)

jxuanli: removed it

pwndbg/aglib/elf.py
jxuanli: good point, this is a bit tricky. I believe my implementation is true for standard kernel images but I m not sure about userland elfs. Since the function is only currently used when debugging kernels, I renamed the function to be used for vmlinux only

pwndbg/aglib/elf.py
```diff
@@ -202,6 +202,14 @@ def get_containing_sections(elf_filepath: str, elf_loadaddr: int, vaddr: int):
     return sections
 
 
+def get_vmlinux_unrand_base(elf_filepath: str):
+    elf = get_elf_info(elf_filepath)
```
disconnect3d, line 206:
```suggestion
    # This should work fine for any standard vmlinux output
    # but may not work for *any* ELF file as it may have the first
    # PT_LOAD with offset != 0
    # See: https://github.com/pwndbg/pwndbg/pull/3269#discussion_r2299313293
    elf = get_elf_info(elf_filepath)
```

disconnect3d: Lol I can't commit this suggestion :D

disconnect3d (approved):

pwndbg/aglib/file.py
```diff
@@ -90,7 +90,7 @@ def get_file(path: str, try_local_path: bool = False) -> str:
         path = path[7:]  # len('target:') == 7
 
     local_path = path
-    if not pwndbg.aglib.remote.is_remote():
```
patryk4815, line 93: not a fun, this code is also wrong for embedded debugging

patryk4815: and dislike idea with:
```
if not pwndbg.aglib.remote.is_remote() or pwndbg.aglib.qemu.is_qemu_kernel()
```
in alot of places, more like it should be (not only for qemu)
```
if not is_remote() or is_kernel() or is_embeded() 
```
or  this should be implemented for checking if target can support it

patryk4815: for example `info proc exe` just returns `target not supported` it all cases kernel/embeded/etc

disconnect3d:
@patryk4815 shall we revert this? or can we fix it in another PR?

PR #3278, opened 2025-08-27
Fix NO_COLOR environment variable support

### LINKED ISSUE #3142: many commands don't respect `NO_COLOR=1`

disconnect3d:
Implement proper NO_COLOR support by adding a nocolor function and configuration trigger that dynamically replaces the colorize function's behavior when colors are disabled. This approach avoids performance overhead by modifying function behavior at runtime rather than adding conditional checks to each colorize call.

Fixes #3142

### DIFF
--- a/pwndbg/color/__init__.py
+++ b/pwndbg/color/__init__.py
@@ -133,6 +133,10 @@ def colorize(x: str, color: str) -> str:
     return color + terminateWith(str(x), color) + NORMAL
 
 
+def nocolor(x: str, color: str) -> str:
+    return x
+
+
 # Taken from https://stackoverflow.com/a/14693789
 ansi_escape_8bit = re.compile(
     r"(?:\x1B[@-Z\\-_]|[\x80-\x9A\x9C-\x9F]|(?:\x1B\[|\x9B)[0-?]*[ -/]*[@-~])"
@@ -148,6 +152,17 @@ disable_colors = theme.add_param(
     bool(os.environ.get("NO_COLOR")),
     "whether to color the output or not",
 )
+
+
+@pwndbg.config.trigger(disable_colors)
+def _disable_colors_trigger():
+    if disable_colors:
+        if not hasattr(colorize, 'original_code'):
+            colorize.original_code = colorize.__code__
+        colorize.__code__ = nocolor.__code__
+    else:
+        if hasattr(colorize, 'original_code'):
+            colorize.__code__ = colorize.original_code
 
 
 def generateColorFunctionInner(
--- a/pwndbg/color/__init__.py
+++ b/pwndbg/color/__init__.py
@@ -11,6 +11,7 @@ from typing import Dict
 from typing import List
 from typing import NamedTuple
 
+import pwndbg
 from pwndbg.lib.config import Parameter
 
 from . import theme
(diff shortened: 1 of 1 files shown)


disconnect3d:
@k4lizen wdyt about this?

k4lizen:
> @k4lizen wdyt about this?

tbh feels like crimes against humanity. this could probably be solved more "conventionally" with some polymorphism like
```python
class AbstractColorizer:
    def colorize(self, text: str, color: str) -> str:
        raise NotImplementedError()


class Colorizer(AbstractColorizer):
    def colorize(self, text: str, color: str) -> str:
        # actual function implementation
       pass

class NopColorizer(AbstractColorizer):
    def colorize(self, text: str, color: str) -> str:
        return text
```
then have one global object like
```python
colorizer = Colorizer()
```
and
```python
@pwndbg.config.trigger(disable_colors)
def _disable_colors_trigger():
    global colorizer    
    if disable_colors:
        colorizer =  NopColorizer()
    else:
        colorizer = Colorizer()
```
and keep the external API the same with
```python
def colorize(x: str, color: str) -> str:
    return colorizer.colorize(x, color)
```
although I will admit this solution is more verbose.

If you end up going with the `__code__` replacement approach, i think it would be nice to add a comment about this to the `def colorize()` function itself.

By the way, if the environment variable NO_COLOR is false, then disable_colors will be initialized to `False`, but will `_disable_colors_trigger` actually be triggered? edit: i meant `initialized`, `True`

patryk4815:
> By the way, if the environment variable NO_COLOR is false, then disable_colors will be initialized to False, but will _disable_colors_trigger actually be triggered?

probably not trigger

disconnect3d:
@k4lizen re: abstraction - yeah, but the abstraction means more calls and more calls means slower speed of execution. With running this thousands of times, this adds up etc... :(

I don't think that matters usually, but I believe it is better to keep it KISS or keep it simple fast :P

PR #3279, opened 2025-08-28
Include ziglang in portable release

patryk4815:


disconnect3d:
This is needed for `cymbol` command etc.

PR #3280, opened 2025-08-28
Fix cachedir on Windows

magnified103:
This PR fixes the `cachedir` function on Windows. Previously, it defaulted to `C:\Windows\System32`, which made it unusable. With this fix, `LOCALAPPDATA` is correctly used to store application-specific data.

Tested with:
```py
Python 3.11.6 (tags/v3.11.6:8b6ee5b, Oct  2 2023, 14:57:12) [MSC v.1935 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import pwndbg.lib.tempfile
>>> pwndbg.lib.tempfile.cachedir("onegadget")
'C:\\Users\\xxx\\AppData\\Local\\pwndbg\\onegadget'
>>>
```

### DIFF
--- a/pwndbg/lib/tempfile.py
+++ b/pwndbg/lib/tempfile.py
@@ -31,7 +31,7 @@ def cachedir(namespace: str | None = None) -> str:
         cachehome = os.getenv("LOCALAPPDATA")
     else:
         cachehome = os.getenv("XDG_CACHE_HOME") or os.path.join(os.getenv("HOME", ""), ".cache")
-    cachedir = os.path.join(cachehome, "pwndbg")
+    cachedir = os.path.join(cachehome or tempfile.gettempdir(), "pwndbg")
     if namespace:
         cachedir = os.path.join(cachedir, namespace)
     os.makedirs(cachedir, exist_ok=True)
--- a/pwndbg/lib/tempfile.py
+++ b/pwndbg/lib/tempfile.py
@@ -5,6 +5,7 @@ Common helper and cache for pwndbg tempdir
 from __future__ import annotations
 
 import os
+import sys
 import tempfile
 
 import pwndbg.lib.cache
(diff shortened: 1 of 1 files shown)


pwndbg/lib/tempfile.py
```diff
@@ -22,11 +23,14 @@ def tempdir() -> str:
...
+    based on XDG_CACHE_HOME or ~/.cache or LOCALAPPDATA (Windows)
 
     Optionally creates a sub namespace inside the pwndbg cache folder.
     """
-    cachehome = os.getenv("XDG_CACHE_HOME") or os.path.join(os.getenv("HOME", ""), ".cache")
+    if sys.platform == "win32":
+        cachehome = os.getenv("LOCALAPPDATA")
+    else:
+        cachehome = os.getenv("XDG_CACHE_HOME") or os.path.join(os.getenv("HOME", ""), ".cache")
     cachedir = os.path.join(cachehome, "pwndbg")
```
patryk4815:
```suggestion
    cachedir = os.path.join(cachehome or tempfile.gettempdir(), "pwndbg")
```
when all envs are unset,  `cachehome` will return None, better add fallback to `tempfile.gettempdir()` at end

pwndbg/lib/tempfile.py
```diff
@@ -22,12 +23,15 @@ def tempdir() -> str:
...
+    based on XDG_CACHE_HOME or ~/.cache or LOCALAPPDATA (Windows)
 
     Optionally creates a sub namespace inside the pwndbg cache folder.
     """
-    cachehome = os.getenv("XDG_CACHE_HOME") or os.path.join(os.getenv("HOME", ""), ".cache")
-    cachedir = os.path.join(cachehome, "pwndbg")
+    if sys.platform == "win32":
+        cachehome = os.getenv("LOCALAPPDATA")
+    else:
+        cachehome = os.getenv("XDG_CACHE_HOME") or os.path.join(os.getenv("HOME", ""), ".cache")
```
patryk4815, line 33: ech, 
`cachehome = os.path.join(os.getenv("HOME", ""), ".cache")`
will return `/.cache/`? but when no envs provided it should use `tempfile.gettempdir()` 🤔

magnified103: I believe that not having either of those env vars is a bigger problem in itself, so it should just use `os.getenv("HOME")` and crash itself.

disconnect3d: Why? Why is it a problem to use a tempdir when home is not there?

disconnect3d: Hmm I think I get the problem now.

```py
        cachehome = os.getenv("XDG_CACHE_HOME") or os.path.join(os.getenv("HOME", ""), ".cache")
    cachedir = os.path.join(cachehome or tempfile.gettempdir(), "pwndbg")
```

If `HOME` is not there the `cachehome` is `.cache` and we never get the temp dir, so this is a bug now?

disconnect3d:
I'm going to merge this since the issue we are mentioning is not related to the added Windows handling here.

Issue #3277, opened 2025-08-27
Unset Pwndbg-breaking environment variables in the Pwndbg wrapper script


disconnect3d:
Related: https://github.com/pwndbg/pwndbg/issues/3276#issuecomment-3229828223

<img width="768" height="337" alt="Image" src="https://github.com/user-attachments/assets/e08f99eb-7e97-4230-bdcd-d7daa3e72db7" />

disconnect3d:
Very important comment regarding this here: https://github.com/pwndbg/pwndbg/pull/3287#issuecomment-3259734476

disconnect3d:
...and not just a comment, but the whole discussion has moved to https://github.com/pwndbg/pwndbg/pull/3287#issuecomment-3259734476 so far.

For completeness and shortness, the TL;DR here is - we are unsure whether we should pursue the unsetting of envvars in the wrapper, since GDB itself do pass all (or most?) envvars into the debugged target, which can be seen here:

<img width="1662" height="672" alt="Image" src="https://github.com/user-attachments/assets/663aa358-ef16-4a6f-b802-bf76350188f1" />

disconnect3d:
We now detect and warn on unsafe variables thanks to https://github.com/pwndbg/pwndbg/pull/3287. Closing this issue.

PR #3287, opened 2025-09-03
Detect and warn on unsafe LD_LIBRARY_PATH/LD_PRELOAD

piotrzarycki:
#3277

### DIFF
--- a/nix/portable.nix
+++ b/nix/portable.nix
@@ -23,27 +23,49 @@ let
   );
   ldLoader = if pkgs.stdenv.isLinux then "\"$dir/lib/${ldName}\"" else "";
   riskEnvsCheck = ''
-      # Check for potentially problematic LD_PRELOAD LD_LIBRARY_PATH
-      if [[ "$*" != *"--quiet"* ]] && [[ "$*" != *"-q"* ]]; then
-          if [[ -n "$LD_LIBRARY_PATH" ]] || [[ -n "$LD_PRELOAD" ]]; then
-              echo
-              echo "WARNING: Potentially problematic environment variables detected!"
-              echo "Check if LD_LIBRARY_PATH or LD_PRELOAD environment variables are set."
-              echo "These may cause library loading issues with debugging tools like pwndbg."
-              echo
-              
-              if [[ -n "$LD_LIBRARY_PATH" ]]; then
-                  echo "LD_LIBRARY_PATH is set to: $LD_LIBRARY_PATH"
-              fi
-              
-              if [[ -n "$LD_PRELOAD" ]]; then
-                  echo "LD_PRELOAD is set to: $LD_PRELOAD"
-              fi
-              
-              echo
-          fi
-      fi
-    '';
+    if [[ "$*" != *"--quiet"* ]] && [[ "$*" != *"-q"* ]]; then
+        detected=0
+        platform=$(uname -s)
+
+        if [[ "$platform" == "Darwin" ]]; then
+            if [[ -n "$DYLD_LIBRARY_PATH" ]] || [[ -n "$DYLD_INSERT_LIBRARIES" ]] || \
+              [[ -n "$DYLD_FALLBACK_LIBRARY_PATH" ]] || [[ -n "$DYLD_FRAMEWORK_PATH" ]]; then
+                detected=1
+            fi
+        else
+            if [[ -n "$LD_LIBRARY_PATH" ]] || [[ -n "$LD_PRELOAD" ]]; then
+                detected=1
+            fi
+        fi
+
+        if [[ $detected -eq 1 ]]; then
+            echo
+            echo "WARNING: Potentially problematic environment variables detected!"
+            echo "These may cause library loading issues with debugging tools like pwndbg."
+            echo
+
+            if [[ "$platform" == "Darwin" ]]; then
+                if [[ -n "$DYLD_LIBRARY_PATH" ]]; then
+                    echo "DYLD_LIBRARY_PATH is set to: $DYLD_LIBRARY_PATH"
+                fi
+
+                if [[ -n "$DYLD_INSERT_LIBRARIES" ]]; then
+                    echo "DYLD_INSERT_LIBRARIES is set to: $DYLD_INSERT_LIBRARIES"
+                fi
+            else
+                if [[ -n "$LD_LIBRARY_PATH" ]]; then
+                    echo "LD_LIBRARY_PATH is set to: $LD_LIBRARY_PATH"
+                fi
+
+                if [[ -n "$LD_PRELOAD" ]]; then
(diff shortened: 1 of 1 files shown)


disconnect3d:
We discussed this internally with @patryk4815 and we think this is probably far from ideal.

With typical GDB experience, if user has any envvars set, those envvars are then passed to the debugged program...

In a similar fashion, if someone has breaking values for those and run just GDB, it may break GDB too.

So we should either not touch those and "keep broken user environment being broken" or - maybe - save those (and some other) envvars, unset them and bring them back inside of GDB.

Ninja3047:
is that how gdb normally works?
I thought you had to do `set environ` or something to actually set it for the target 
doing `LD_PRELOAD` would change the library path of GDB which is probably the wrong thing to do in any case.
saving the value and unsetting for pwndbg and forwarding it to the target might be fine but also kind of weird

piotrzarycki:
Ok, so you want to say that if a user has a custom environment setup, it's their responsibility to adjust their setup without overriding the required environment variables on pwndbg side? :)

disconnect3d:
> is that how gdb normally works? I thought you had to do `set environ` or something to actually set it for the target doing `LD_PRELOAD` would change the library path of GDB which is probably the wrong thing to do in any case. 

Yes, you can see it here:
```
$ LD_PRELOAD=lolnothinghere gdb --quiet --nx /bin/sh
ERROR: ld.so: object 'lolnothinghere' from LD_PRELOAD cannot be preloaded (cannot open shared object file): ignored.
Reading symbols from /bin/sh...

This GDB supports auto-downloading debuginfo from the following URLs:
  <https://debuginfod.ubuntu.com>
Enable debuginfod for this session? (y or [n]) n
Debuginfod has been disabled.
(No debugging symbols found in /bin/sh)
(gdb) run -c 'echo $LD_PRELOAD'
Starting program: /usr/bin/sh -c 'echo $LD_PRELOAD'
ERROR: ld.so: object 'lolnothinghere' from LD_PRELOAD cannot be preloaded (cannot open shared object file): ignored.
ERROR: ld.so: object 'lolnothinghere' from LD_PRELOAD cannot be preloaded (cannot open shared object file): ignored.
lolnothinghere
[Inferior 1 (process 182807) exited normally]
(gdb)
```

Or here a screenshot with my highlights:
image

> saving the value and unsetting for pwndbg and forwarding it to the target might be fine but also kind of weird

I totally agree and have a bit mixed feelings over this. But since our portable release versions do contain everything included they may not work properly with the given `LD_PRELOAD`/`LD_PRELOAD_PATH` environment variables set by the user while their own ("local") `gdb` may work with it, lol.

disconnect3d:
> Ok, so you want to say that if a user has a custom environment setup, it's their responsibility to adjust their setup without overriding the required environment variables on pwndbg side? :)

I am still thinking out loud and trying to decide what we should do here but I am leaning towards what you are saying :P

disconnect3d:
We could also check those envvars and warn the user "hey this may just not work due to this and that", but if the user wants to force it anyway it may be annoying... maybe `pwndbg --quiet` should skip displaying of such warning? (Do we pass this flag to `gdb` anyway? I don't know lol)

piotrzarycki:
yeah, i think notifying the user without forcing changes is fair + quiet - happy to implement that!

piotrzarycki:
@disconnect3d  implemented some nice warnings without forcing any environment changes on the pwndbg side :)

pwndbg/commands/quiet.py
```diff
@@ -0,0 +1,19 @@
...
+    Implementation of the `quiet` command.
+    """
+
+
+@pwndbg.commands.Command(
+    "Suppress Output Messages",
+    category=CommandCategory.PWNDBG,
+)
+def quiet() -> None:
+    quiet_impl()
```
disconnect3d: This whole file seems like AI generated. It is not needed at all for any of the changes regarding quiet flag ;)

pwndbginit/pwndbg_gdb.py
```diff
@@ -44,8 +44,36 @@ def prepend_venv_bin_to_path():
...
+        print("These may cause library loading issues with debugging tools like pwndbg.")
+        print()
+
+        if ld_library_path:
+            print(f"LD_LIBRARY_PATH is set to: {ld_library_path}")
+
+        if ld_preload:
+            print(f"LD_PRELOAD is set to: {ld_preload}")
+
+        print()
```
disconnect3d: If the environment variables are set to something that breaks GDB, this code will never run.

The checks - if we want to add them - should be implemented in the wrapper script.

disconnect3d (changes requested):
The checks need to be implemented in the wrapper script - if we decide to add them (cc: @patryk4815 wdyt?)

pwndbg/commands/quiet.py
piotrzarycki: nope - i used the code as reference https://github.com/pwndbg/pwndbg/blob/2195afbc3385fc3589732875472dad8b5022bd1a/pwndbg/commands/version.py#L99 -  because the docs script is complaining about an undocumented quiet flag.

pwndbginit/pwndbg_gdb.py
piotrzarycki: can you give me some advice on where the wrapper script is? :)

pwndbg/commands/quiet.py
k4lizen: could you post the error that the docs script is giving you when this file doesn't exist?

piotrzarycki: there was an issue when I tried using argparse probably, now is not working so i removed this

pwndbginit/pwndbg_gdb.py
disconnect3d: ...you modified a part of the generated wrapper script in https://github.com/pwndbg/pwndbg/pull/3287/commits/ab59b78cb0fcc36688c08906b3e8c85e24d9b642

piotrzarycki: added

disconnect3d:
cc: @patryk4815 any final thoughts on this?

patryk4815:
This wrapper is also called for macOS, maybe for macOS should be different detection DYLD_INSERT_LIBRARIES/DYLD_LIBRARY_PATH?

I think best idea is to use whitelist few environments into python/gdb/lldb.
And then only forward all envs into debugged process?

disconnect3d:
@piotrzarycki any chance u can update this accordingly?:<

piotrzarycki:
added support for DYLD_INSERT_LIBRARIES, DYLD_LIBRARY_PATH

disconnect3d:
@piotrzarycki Thanks! Sorry for taking so long time to merge it. I wasn't sure if we really want this but after some discussions, let's take it in :)

PR #3301, opened 2025-09-11
Draft: lldb core file support

patryk4815:
@mbrla0 if you can do better PR from this changes 🙏

### DIFF
--- a/pwndbg/dbg/lldb/repl/__init__.py
+++ b/pwndbg/dbg/lldb/repl/__init__.py
@@ -568,7 +568,7 @@ def _exec_repl_command(
     if bits[0].startswith("ta") and "target".startswith(bits[0]):
         if len(bits) > 1 and bits[1].startswith("c") and "create".startswith(bits[1]):
             # This is `target create`
-            target_create(bits[2:], dbg)
+            target_create(driver, bits[2:], dbg)
             return True
         if len(bits) > 1 and bits[1].startswith("de") and "delete".startswith(bits[1]):
             # This is `target delete`
@@ -903,7 +903,6 @@ target_create_ap.add_argument("-v", "--version")
 target_create_ap.add_argument("filename")
 target_create_unsupported = [
     "build",
-    "core",
     "no-dependents",
     "remote-file",
     "symfile",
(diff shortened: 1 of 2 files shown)


patryk4815:
btw in normal LLDB it is possible to load core file without target binary.
```
(lldb) target create --core "./core"
Core file '/tmp/shared/core' (loongarch64) was loaded.
```

mbrla0:
I think an important thing we need to think about is how we want to do events when handling core files.

`SBTarget::LoadCore` doesn't let us give it an `SBListener`, which means we can't get an easy way to receive events from the process object it creats. On one hand, that should be fine, as, if you inspect the logs given by LLDB, the only event I've been able to get it to generate is `state-changed` with `state = stopped`[^1]. On the other, though, it gives us two kinks that we've got to iron out:

1. Pwndbg expects some events to be generated before it can do anything useful, and
2. `_run_until_next_stop` wasn't exactly designed to work with a process that never generates events.

[^1]: `Broadcaster("lldb.process")::BroadcastEvent (event_sp = 0x55ddf41e9870 Event: broadcaster = 0x55ddf419ad98 (lldb.process), type = 0x00000001 (state-changed), data = { process = 0x55ddf419ad60 (pid = 1091826), state = stopped}, unique=false)`

It should be easy enough to deal with (1). Generating a special event is something we can just do from inside `ProcessDriver` with no regard to our lack of native LLDB events, and the rest of Pwndbg should be none-the-wiser that this is a special case.

As for (2), I feel we should tread a little more carefully. `_run_until_next_stop` is designed around processes that do emit events. That being said, though, because of an unrelated reason having to do with command handling, the existing version _should_ in most cases be able to handle a process that never emits any events, but I'm not sure how graceful that handling is.

mbrla0:
To clarify my previous comment, `_run_until_next_stop` will only handle this case correctly if `only_if_started` is given, and is `True`. Otherwise it is likely to hang.

This is fairly common for `ProcessDriver`. For example, both `ProcessDriver.cont` and `ProcessDriver.run_coroutine` both unconditionally expect the process to fire events, as `only_if_started` is `False` for them.

PR #3282, opened 2025-08-29
Fix DYLD Shared Cache

patryk4815:
This PR also include changes from: https://github.com/pwndbg/pwndbg/pull/3302

Old version of `_dyld_get_shared_cache_range` was wrong from begging, this func has required first argument.
And lldb21+ now checks if "evaluation" is in proper state:
image

Closes:
- https://github.com/pwndbg/pwndbg/issues/3281

PR #3302, opened 2025-09-11
Fix handling of missing DYLD Shared Cache

mbrla0:
During early execution, the DYLD Shared Cache may not be available. This can manifest as the `_dyld_get_shared_cache_range` being communicated to us as having been optimized out. This PR makes it so that that case is properly handled.

Additionally, this PR adds an optional list of ignored values to `pwndbg.lib.cache.cache_until`, so that caching is not performed if they are returned, and the cached function is allowed to run again, which is used by the shared cache function, to allow it to be retried later on. An alternative to this would be to cache until `"stop"`, but that would not entirely capture the fact that, once that value is found, it stays the same for the entire lifetime of the process.

### DIFF
--- a/pwndbg/aglib/macho.py
+++ b/pwndbg/aglib/macho.py
@@ -637,7 +637,7 @@ class DyldSharedCache:
         return DyldSharedCacheHashSet(ptr)
 
 
-@pwndbg.lib.cache.cache_until("exit")
+@pwndbg.lib.cache.cache_until("exit", ignore_for=[None])
 def shared_cache() -> DyldSharedCache | None:
     """
     Base address of the Darwin shared cache.
@@ -659,12 +659,16 @@ def shared_cache() -> DyldSharedCache | None:
     [1]: https://github.com/apple-oss-distributions/objc4/blob/f126469408dc82bd3f327217ae678fd0e6e3b37c/runtime/objc-opt.mm#L434
     [2]: https://github.com/apple-oss-distributions/dyld/blob/main/doc/dyld4.md#libdylddylib
     """
-    base = int(
-        pwndbg.dbg.selected_inferior().evaluate_expression(
-            "(const void*)_dyld_get_shared_cache_range()"
-        )
-    )
+    if pwndbg.aglib.symbol.lookup_symbol("_dyld_get_shared_cache_range") is None:
+        return None
 
+    base = pwndbg.dbg.selected_inferior().evaluate_expression(
+        "(const void*)_dyld_get_shared_cache_range()"
+    )
+    if base.is_optimized_out:
+        return None
+
+    base = int(base)
     if base == 0:
         return None
 
--- a/pwndbg/lib/cache.py
+++ b/pwndbg/lib/cache.py
@@ -126,8 +126,7 @@ IS_CACHING_DISABLED_FOR: Dict[str, bool] = {
     "forever": False,
 }
 
-
-def cache_until(*event_names: str) -> Callable[[Callable[P, T]], Callable[P, T]]:
+def cache_until(*event_names: str, ignore_for: List[Any] | None = None) -> Callable[[Callable[P, T]], Callable[P, T]]:
     if any(event_name not in _ALL_CACHE_EVENT_NAMES for event_name in event_names):
         raise ValueError(
             f"Unknown event name[s] passed to the `cache_until` decorator: {event_names}.\n"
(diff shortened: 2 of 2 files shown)


pwndbg/lib/cache.py
```diff
@@ -166,7 +165,9 @@ def decorator(*a: P.args, **kw: P.kwargs) -> T:
                 if isinstance(value, list):
                     print(f"Should not cache mutable types! {func.__name__}")
 
-                cache[key] = value
+                if ignore_for is None or value not in ignore_for:
+                    # Only cache if we were not requested to ignore this value.
+                    cache[key] = value
```
disconnect3d, line 170: Not sure I like it since it influences each cached call.

What about doing another decorator to call cache wrapper if ignored value is not what was returned?

pwndbg/aglib/macho.py
```diff
@@ -659,12 +659,13 @@ def shared_cache() -> DyldSharedCache | None:
     [1]: https://github.com/apple-oss-distributions/objc4/blob/f126469408dc82bd3f327217ae678fd0e6e3b37c/runtime/objc-opt.mm#L434
     [2]: https://github.com/apple-oss-distributions/dyld/blob/main/doc/dyld4.md#libdylddylib
     """
-    base = int(
-        pwndbg.dbg.selected_inferior().evaluate_expression(
-            "(const void*)_dyld_get_shared_cache_range()"
-        )
+    base = pwndbg.dbg.selected_inferior().evaluate_expression(
```
patryk4815, line 665: you should check if symbol `_dyld_get_shared_cache_range` is available

mbrla0: Fair enough. Done.

pwndbg/lib/cache.py
mbrla0: > What about doing another decorator to call cache wrapper if ignored value is not what was returned?

You mean calling `.cache.clear()` if we detect an ignored value? Well, that's okay, I guess, even if it would kinda make the two of them order-sensitive? What exactly are you worried about, putting it in the main decorator?

pwndbg/aglib/macho.py
```diff
@@ -659,12 +659,16 @@ def shared_cache() -> DyldSharedCache | None:
...
-        pwndbg.dbg.selected_inferior().evaluate_expression(
-            "(const void*)_dyld_get_shared_cache_range()"
-        )
+    if pwndbg.aglib.symbol.lookup_symbol("_dyld_get_shared_cache_range") is None:
+        return None
+
+    base = pwndbg.dbg.selected_inferior().evaluate_expression(
+        "(const void*)_dyld_get_shared_cache_range()"
     )
+    if base.is_optimized_out:
```
patryk4815, line 668: when you add check for symbol, `base.is_optimized_out` is not needed anymore 🙃

mbrla0: Honestly, that feels a little surprising, since neither `pwndbg.aglib.symbol.lookup_symbol` or `Process.lookup_symbol`, seem to document that behavior. 

Of course, an optimized-out value won't have an address, but those lookup functions return `Value`s, which does mean the Debugger API back-end is allowed to return an otherwise-valid value that is flagged as optimized-out, and that only errors out when we attempt to extract the value.

Doesn't look like our LLDB implementation ever does this, but omitting the check might be relying on implementation details, rather than on the guarantees given by the interface itself?

patryk4815: maybe when `is_opimized_out` is True, `evaluate_expression` should return None

mbrla0: > maybe when `is_opimized_out` is True, `evaluate_expression` should return None

`Value.is_optimized_out` in the Debugger API was designed to be compatible with `gdb.Value.is_optimized_out`, and part of that is letting us distinguish between bindings that can be named in an expression but that don't have a retrievable value due to having been optimized out from those that aren't even available to the expression.

I'm sure we could also signal that using `None`, but at least when the Debugger API was first written, Pwndbg did make use of `is_optimized_out`, specifically to make that distinction. I'm not sure if it still does, though.

I guess you can think of it working that way as a small quirk the Debugger API inherited from GDB.

pwndbg/aglib/macho.py
```diff
@@ -637,7 +637,7 @@ def objc_builtin_selectors(self) -> DyldSharedCacheHashSet:
         return DyldSharedCacheHashSet(ptr)
 
 
-@pwndbg.lib.cache.cache_until("exit")
```
patryk4815, line 640: I think we can use `new_objfile` here, and `ignore_for` don't need to be implemented

patryk4815:
I’m closing this PR in favor of https://github.com/pwndbg/pwndbg/pull/3282.
Your changes have been cherry-picked, and the original "Author" field has been preserved.

PR #3310, opened 2025-09-23
Add parse-seccomp command to resolve bpf filter using ceccomp or seccomp-tools

dbgbgtf1:
May add a get-secocmp commandto resolve process bpf filter later

image


pwndbg/commands/parse_seccomp.py
```diff
@@ -0,0 +1,53 @@
...
+    help="Address of sock_fprog structure in target process memory (e.g. 0xdeadbeef)",
+)
+
+
+@pwndbg.commands.Command(parser, command_name="parse-seccomp", category=CommandCategory.MISC)
+@pwndbg.commands.OnlyWhenRunning
+def parse_seccomp(addr: int) -> None:
+    """Parse a struct sock_fprog at a given address and pass filter to external tool."""
+
+    addr = int(addr) & pwndbg.aglib.arch.ptrmask
```
disconnect3d: Isn't `addr` int already?

Why do we need to do ` & ptrmask`?

pwndbg/commands/parse_seccomp.py
```diff
@@ -0,0 +1,53 @@
...
+    ptrsize = pwndbg.aglib.typeinfo.ptrsize
```
disconnect3d: redundant variable, used just once, pls move it to line 33

pwndbg/commands/parse_seccomp.py
```diff
@@ -0,0 +1,53 @@
...
+            ["ceccomp", "disasm", "--color", "always"], input=filter_bytes, capture_output=True
+        )
+        print(proc.stdout.decode())
+    elif shutil.which("seccomp-tools"):
+        proc = subprocess.run(
+            ["seccomp-tools", "disasm", "-"], input=filter_bytes, capture_output=True
+        )
+        print(proc.stdout.decode())
+    else:
+        print("install ceccomp or seccomp-tools to parse seccomp")
```
disconnect3d, line 50: Generally, we should think if we want to install it for the user, but this can stay as-is for now

cc: @patryk4815 fyi

pwndbg/commands/parse_seccomp.py
```diff
@@ -0,0 +1,53 @@
...
+    description="Parse a struct sock_fprog from memory and dump its filter"
+)
+parser.add_argument(
+    "addr",
+    type=int,
+    help="Address of sock_fprog structure in target process memory (e.g. 0xdeadbeef)",
+)
+
+
+@pwndbg.commands.Command(parser, command_name="parse-seccomp", category=CommandCategory.MISC)
```
patryk4815: I think it should go to the `CommandCategory.LINUX` category

pwndbg/commands/parse_seccomp.py
patryk4815: we can include `ceccomp` because it is C, and probably it can work on every platform

pwndbg/commands/parse_seccomp.py
RocketMaDev: On 32-bit platforms, some pointers may greater than `0x7fffffff`, so when input addr with `$edx` or other register values, it is a negative number, rather than `uint32_t` big number.

pwndbg/commands/parse_seccomp.py
dbgbgtf1: 52e27a8f6f3c8941c773a4bf86d13c4ea58f0f2d

pwndbg/commands/parse_seccomp.py
disconnect3d: Iirc we have our own Value type which should handle this already and not make the type "signed" as GDB does it. But maybe we don't?

Also, if this is really the case as you write it, then the function argument annotation here is wrong because this means that not only int is accepted here, but also a Value (DbgValue or whatever name this type we use has -- I don't remember on top of my head). Please fix the annotation if this is the case

pwndbg/commands/parse_seccomp.py
dbgbgtf1: I am a bit confused with the command category, so I will just do as you say
Solved in 370c894e3f80f41d31dc7839fbf3209ecc7d86b7

pwndbg/commands/parse_seccomp.py
```diff
@@ -0,0 +1,53 @@
...
+    if shutil.which("ceccomp"):
+        proc = subprocess.run(
+            ["ceccomp", "disasm", "--color", "always"], input=filter_bytes, capture_output=True
+        )
+        print(proc.stdout.decode())
+    elif shutil.which("seccomp-tools"):
+        proc = subprocess.run(
+            ["seccomp-tools", "disasm", "-"], input=filter_bytes, capture_output=True
+        )
+        print(proc.stdout.decode())
```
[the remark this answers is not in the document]
disconnect3d: We can skip this imho

pwndbg/commands/parse_seccomp.py
```diff
@@ -0,0 +1,53 @@
...
+
+    filter_len = pwndbg.aglib.memory.u16(addr)
+    filter_addr = pwndbg.aglib.memory.u(addr + ptrsize)
+
+    print(message.success(f"sock_fprog @ {addr:#x}"))
+    print(f"  len          = {filter_len}")
+    print(f"  filter_addr  = {filter_addr:#x}")
+
+    filter_size = filter_len * 8
+    filter_bytes = pwndbg.aglib.memory.read(filter_addr, filter_size, partial=True)
```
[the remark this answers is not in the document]
disconnect3d: This seems to make sense to handle this nicely

pwndbg/commands/parse_seccomp.py
patryk4815: if ` & pwndbg.aglib.arch.ptrmask` is required, then it is bug in pwndbg and should be fixed in different place

dbgbgtf1: I'm not sure, so I learn that from `telescope`
`pwndbg/commands/telescope.py:125`
`address = int(address) & pwndbg.aglib.arch.ptrmask`

dbgbgtf1: > if ` & pwndbg.aglib.arch.ptrmask` is required, then it is bug in pwndbg and should be fixed in different place

I agree. I miss C

disconnect3d: I just confirmed that `& ptrmask` is not needed to prevent negative numbers: an `int` is always returned here, even if we do sth like `telescope $rax`.

In other words, the `telescope` command also does not need this bit masking operation

PR #3311, opened 2025-09-26
Handling `kmem_cache` recovery edge case + improving SLUB corrupted list handling

jxuanli:
- fixed the edge when recovering `kmem_cache` for kernel builds with `CONFIG_SLAB_FREELIST_RANDOM=y` but without `CONFIG_HARDENED_USERCOPY`  or `CONFIG_KASAN`
- supported detection of unaligned and OOB kernel object
- moved corrupted SLUB list reporting to `slab info` so it does not report errors each time the list is traversed
- general refactoring


jxuanli:
hmmm my changes shouldn't have effected the failed tests

pwndbg/commands/slab.py
```diff
@@ -300,12 +331,14 @@ def slab_contains(address: str) -> None:
...
-            elif not slab.is_cpu and slab.is_partial:
-                desc = f"[partial, node {slab.node_cache.node}]"
+            if slab.is_active:
+                if slab.is_cpu:
+                    desc = f"[active, cpu {slab.cpu_cache.cpu}]"
+            else:
+                if slab.is_cpu:
+                    desc = f"[partial, cpu {slab.cpu_cache.cpu}]"
+                else:
+                    desc = f"[partial, node {slab.node_cache.node}]"
```
disconnect3d, line 341: can we just do:
```
if
elif
else
```
?

disconnect3d:
Btw can u show a screenshot with output of this?

disconnect3d:
I've added some comment[s] pls address them in another PR

jxuanli:
the changes in output only mattered to corrupted slub freelists where the type of corruption is only reported as part of the output of `slab info` (whereas previously it is reported each time the list is traversed, which was not necessary and clusters/disrupts the output), demo:

image

pwndbg/commands/slab.py
jxuanli: I think the nested ifs are clearer in this case (see the diff)

Issue #3100, opened 2025-06-10
fuzzy missprint pwndbg-lldb on arch


k4lizen:
### Description

```
~> python pwndbg-lldb.py /bin/sh
pwndbg: loaded 158 pwndbg commands commands.
------- tip of the day (some of these don't work in LLDB yet!) (disable with set show-tips off) -------
Want to display each context panel in a separate tmux window? See https://github.com/pwndbg/pwndbg/blob/dev/FEATURES.md#splitting--layouting-context
ANSI('\x1b[34mpwndbg-lldb> ')target create '/bin/sh'
Current executable set to '/bin/sh' (x86_64)
pwndbg-lldb> 
```

The `ANSI('\x1b[34mpwndbg-lldb> ')` text is ???

I guess it's coming from the fzf code. Btw is there not a config to turn it off?

### Steps to reproduce

```
python pwndbg-lldb.py /bin/sh
```

### My setup

Pwndbg:   2025.05.30 build: ac7813bb
Python:   3.13.3
LLDB:     20.1
Capstone: 6.0.0
Unicorn:  2.1.3
Pwnlib:   4.14.0

lldb version 20.1.6

disconnect3d:
Fixed by #3314 I believe - @k4lizen let us know if it doesnt

k4lizen:
tbh i can't reproduce the issue even without the fixing commit, but yeah I guess its fixed now. ill reopen it if I ever manage to reproduce it but i'm optimistic

PR #3314, opened 2025-09-29
Fix lldb fzf startup prompt

arcz:
When starting up, the `pwndbg-lldb> ` prompt was incorrectly formatted when `fzf` detected.

Before:
```
ANSI('\x1b[34mpwndbg-lldb> ')target create 'test'
Current executable set to 'test' (arm64)
```

After:
```
pwndbg-lldb> target create 'test'
Current executable set to 'test' (arm64)
```

### DIFF
--- a/pwndbg/dbg/lldb/repl/__init__.py
+++ b/pwndbg/dbg/lldb/repl/__init__.py
@@ -393,7 +393,7 @@ def run(
             last_command = action._command
 
             if not action._prompt_silent:
-                print(f"{PROMPT}{action._command}")
+                print(f"{PROMPT.value if HAS_FZF else PROMPT}{action._command}")
 
             try:
                 if action._capture:
--- a/pwndbg/dbg/lldb/repl/fuzzy.py
+++ b/pwndbg/dbg/lldb/repl/fuzzy.py
@@ -39,7 +39,7 @@ from pwndbg.dbg.lldb import LLDB
 P = ParamSpec("P")
 T = TypeVar("T")
 
-PROMPT = ANSI("\x1b[34mpwndbg-lldb> ")
+PROMPT = ANSI("\x1b[34mpwndbg-lldb>\x1b[0m ")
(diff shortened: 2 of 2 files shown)


pwndbg/dbg/lldb/repl/__init__.py
```diff
@@ -393,7 +393,7 @@ def handle_sigint(_sig, _frame):
             last_command = action._command
 
             if not action._prompt_silent:
-                print(f"{PROMPT}{action._command}")
+                print(f"{PROMPT.value if HAS_FZF else PROMPT}{action._command}")
```
disconnect3d, line 396: Do we know `HAS_FZF` value up front in fuzzy.py? If so, shouldn't we define `PROMPT = ...` based on it?

disconnect3d: Note that we also do this conditionally:
```py
HAS_FZF = shutil.which("fzf") is not None
if HAS_FZF:
    from pwndbg.dbg.lldb.repl.fuzzy import PROMPT
    from pwndbg.dbg.lldb.repl.fuzzy import get_prompt_session
    from pwndbg.dbg.lldb.repl.fuzzy import wrap_with_history
else:
    from pwndbg.dbg.lldb.repl.readline import PROMPT
    from pwndbg.dbg.lldb.repl.readline import enable_readline
    from pwndbg.dbg.lldb.repl.readline import wrap_with_history
```

huh

k4lizen:
re https://github.com/pwndbg/pwndbg/issues/3100

pwndbg/dbg/lldb/repl/__init__.py
arcz: If you look earlier in this file, there are places which depend on `HAS_FZF` in the same way so I just followed, but there is definitely a bigger refactor opportunity

disconnect3d:
@k4lizen so this fixes https://github.com/pwndbg/pwndbg/issues/3100 right?

Issue #2874, opened 2025-04-14
add color parameter validation


k4lizen:
Currently when someone wants to set a color parameter like e.g. `telescope-register-color` they have no way of knowing which parameter values are valid. Furthermore, it is possible to set invalid values. And an exception is thrown when such an invalid color is displayed (which is bad):
```
│   193 │                                                                                          │
│   194 │   for color in config.split(","):                                                        │
│   195 │   │   func_name = color.lower().replace("-", "_")                                        │
│ ❱ 196 │   │   function = generateColorFunctionInner(function, _globals[func_name])               │
│   197 │   return function                                                                        │
│   198                                                                                            │
│   199                                                                                            │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
KeyError: 'meow'
```
but also, the valid values are taken from the python builtin `globals()` function, which is very weird and allows some funky behaviours:
```
pwndbg> set telescope-register-color ansi_escape_8bit
```
```
│                                                                                                  │
│   150 │   old: Callable[[object], str], new: Callable[[str], str]                                │
│   151 ) -> Callable[[object], str]:                                                              │
│   152 │   def wrapper(text: object) -> str:                                                      │
│ ❱ 153 │   │   return new(old(text))                                                              │
│   154 │                                                                                          │
│   155 │   return wrapper                                                                         │
│   156                                                                                            │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
TypeError: 're.Pattern' object is not callable
```
This should be cleaned up. Probably something like the existing PARAM_ENUM logic.

k4lizen:
To be clear, ideally the logic should prevent you from setting a bad color value in the first place (both from the code (assert?) and as a user (clear error message)), rather than only catching the error when the color is displayed.

The enum logic is enforced by gdb though (and by us on the lldb side), and it can't be reused exactly since we allow comma-separated values like: `bold,red`.

luca-c-coll:
@k4lizen is this issue open for work? if so, may i help on this issue?

k4lizen:
@luca-c-coll yup, feel free :P

Giosvelte:
Hello! I would like to work on this issue as a first-time contributor. I plan to try a slightly different approach for color parameter validation to prevent invalid values and runtime exceptions. Thank you!

debosmita09:
[...]

PR #3315, opened 2025-09-30
Color Parameter Validation in generateColorFunction

luca-c-coll:
**Issue:** #2874 

**Description:** 

Adds validation for color names in ColorParameter.update_color_function to prevent KeyError and invalid inputs. Invalid colors are now filtered out and produce a single warning. 

**Changes:** 

- Added validation for color names before calling generateColorFunction. 
- Filtered invalid values and added warning messages. 

**Test/Lint:** 

- pytest -v test_theme_local.py passed (valid, invalid, empty cases). 
- ruff check . and isort passed.


luca-c-coll:
@k4lizen let me know if these changes work. if not i'll work on an alternative asap. thanks

k4lizen:
hello! ideally we want wrong colors to be caught when they get set by the user, rather than here. so id expect this to be handled in the configuration logic

also, you should link the github issue you are trying to tackle in the PR description since that makes it much easier to cross reference them. 

there was a PR that also tried to tackle this issue but didnt get merged, but im ironically having trouble finding it because it also didnt link to the issue

luca-c-coll:
sounds good, thank you. will make necessary adjustments.

luca-c-coll:
@k4lizen QEMU test_command_kernel_vmmap failure is unrelated to this change... allother tests pass.

k4lizen:
If I do
```
pwndbg> set prop-value-color blabstarst
pwndbg> theme
```
I get an exception. We should make these changes in a way to prevent this. Afterwards when I do
```
pwndbg> show 
Color used to highlight the value in name-value pairs is 'blabstarst'. See `help set prop-value-color` for more information.
```
I expect the value to *not* be `blabstarst`, but rather the original value (in this case `yellow`).

Also to be doubly clear, try to mimic the behaviour of setting an incorrect enum value to a configuration parameter. It doesn't make sense for the color parameters to deviate from that behaviour, right? In fact I recommend seeing how pwndbg-lldb> handles this.
image
you may grep for that error string to find it in the code

luca-c-coll:
@k4lizen i made a few changes. let me know if these changes work and if any new adjustments are needed. thanks

k4lizen:
please don't ask for rereview until you address all the comments i made (they haven't been addressed)

luca-c-coll:
hey @k4lizen, i believe i've addressed your requirements now based on the test cases i created locally (removed in latest commit). let me know if this works. thank you.

disconnect3d:
Hi, there are still some issues with this and honestly, I think it would be easier to just reimplement this from scratch and make it as easy/stupid as possible.

Issues:
1. there shouldn't be any valid/non-valid logic, it should just reject if anything is invalid
2. what is the set() function doing?
3. sys.stderr.write is not the way to print an error message
4. the VALID_COLORS array should somehow be tied to the color.py file so they can't get desynced

Please, for the sake of our sanity, don't generate code with AI/LLM but write it by yourself... the LLM-generated code is bloat/bullshit. Thanks.

I am going to close this as I believe it would be easier to implement this from scratch and not waste time on reviews that needs re-reviews and re-reviews...

PR #3318, opened 2025-10-03
Added `klookup` option for applying symbols

### LINKED ISSUE #2809: Add $klookup("symbol") function
During kernel debugging I can do:
```c
pwndbg> klookup nat_helper_ftp
0xffffffff92c337a0 = nat_helper_ftp
```

But I would like to do:
```
pwndbg> break *($klookup("nat_helper_ftp")+0x123)
```

This function should probably have a 2nd argument of index of symbol to use if there are multiple returned. This second argument should be 0?

jxuanli:
- improved `klookup` such that its output matches `cat /proc/kallsyms`
- added option for applying symbols (closes #2809) using lief

## demo

image


jxuanli:
although `b <added symbol>` works, `disass <added symbol>` does not. I'm not too sure if it is possible to make `disass` work, some searching suggests that the size of the function is needed so perhaps not.

PR #3319, opened 2025-10-03
Robust Dprintf

### LINKED ISSUE #3243: Robust dprintf
I often do: `dprintf foo, "foo(%p)\n", $rdi` which works great!

But:
- I may want to see who was the caller (which function called foo)
- I may want to see some thread identifier too

Example usage - `-d` would be depth 1 for backtrace?:
```
dp "foo(%p)" --tid -d=1
```

Example print - when foo is called by bar:

```
[TID: 2] bar > foo(0x1234)
```

ThomasSpeer252:
Closes #3243

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -909,6 +909,7 @@ def load_commands() -> None:
     import pwndbg.commands.distance
     import pwndbg.commands.dt
     import pwndbg.commands.dumpargs
+    import pwndbg.commands.dprintf
     import pwndbg.commands.elf
     import pwndbg.commands.flags
     import pwndbg.commands.gdt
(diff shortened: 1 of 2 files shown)


disconnect3d:
What is this code doing? Is this some garbage generated by LLMs? There are multiple issues with this code that I don't really want to spend my time on it:
1) You didn't even show the output from it, how one can use it, no description of PR about limitations etc.
2) There are unused arguments for the `dp` function like `arguments` ...
3) This works only in GDB: this may be ok, bcoz `dprintf` is purely a GDB function, but then the command must only be imported in GDB context (we support LLDB as well in Pwndbg)
4) The lint, dev, test CI checks do fail here

I am going to close this for now. Please reopen/send new PR if you decide to fix all of above.

PR #3320, opened 2025-10-03
Adding $heap(offset) convenience function. Issue# 2849

ordinalsoul:


disconnect3d:
1. We already have the `$base` function that can be used for this, e.g.: `$base("heap")+0x123`
2. I don't mind adding a `$heap(offset or 0)` convenience function but the proposed implementation in this PR is awful. See how `def base` is implemented in pwndbg/gdblib/functions.py and implement the `$heap` one similarly and in there.

Please open another PR if u decide to implement this properly.

PR #3324, opened 2025-10-06
Make install script more robust 

gavhu10:
I made so that downloading the files and installing pwngdb are functions that are called at the end of the script, so that no files are changed if a file download is interrupted.


patryk4815:
your changes don't fix anything

PR #3326, opened 2025-10-06
Replaced LLDBValue with lldb.value() and made changes to create_value…

Heisenberg-Vader:


Heisenberg-Vader:
I have replaced the LLDBValue with lldb.value wherever possible without altering much functionality of the class LLDBValue itself. Also made some changes to create_value function as mentioned in the issue description. I ran the `./tests.sh -d gdb -g gdb` just to make sure things have not broken. It seems to work all fine. There was one failed test case which remained from when the repository was cloned.

Heisenberg-Vader:
The changes were made only in `pwndbg/dbg/lldb/__init__.py`

patryk4815:
this PR is very weak, it is impossible to review it

PR #3330, opened 2025-10-06
remove lief from dep list

jxuanli:
as per discussion on discord, can revert back if we find out this is needed


pwndbg/commands/cymbol.py
```diff
@@ -143,6 +142,11 @@ def generate_debug_symbols(
 
 
 def create_blank_elf():
+    try:
+        import lief
+    except ImportError:
+        print(message.error("lief python package is not installed."))
```
disconnect3d:
```suggestion
        print(message.error("lief python package is not installed (this will be auto-installed with next version of Pwndbg; for now, you can install it manually in the Pwndbg venv)."))
```

disconnect3d:
GH broke here for some reason and we can't merge it atm; can u make any commit here?

PR #3345, opened 2025-10-10
Drop debug info requirement for kernel section markers

Ninja3047:
I think all of this info is actually available after this PR https://github.com/pwndbg/pwndbg/pull/3116

Removing this debug check appears to work for me for x86_64 (although i did not look too carefully to see if it was actually correct, but it looked okay). Did not check aarch64 but based on the code it should be fine.


Ninja3047:
https://github.com/pwndbg/pwndbg/pull/3346

Issue #3349, opened 2025-10-13
`klookup` fails


bata24:
### Description

`klookup` fails because it calls `(None).splitlines()`.

<img width="633" height="72" alt="Image" src="https://github.com/user-attachments/assets/ac924e57-f52b-4534-a089-533788d3a33e" />

### Steps to reproduce

```
[term1]
$ ./start.sh

[term2]
$ gdb -ex 'tar rem :1234'
pwndbg> klookup
```

### My setup

Ubuntu 25.04 amd64

```
root@Ubuntu2504:~/shared-c/qemu-buildroot-x64-6.17.0-rc3# qemu-system-x86_64 --version
QEMU emulator version 9.2.1 (Debian 1:9.2.1+ds-1ubuntu5.2)
Copyright (c) 2003-2024 Fabrice Bellard and the QEMU Project developers
root@Ubuntu2504:~/shared-c/qemu-buildroot-x64-6.17.0-rc3# 
```

```
pwndbg> version 
Pwndbg:   2025.10.10 build: 85b93a792 (Linux)
Python:   3.13.3 (main, Aug 14 2025, 11:53:40) [GCC 14.2.0]
GDB:      17.0.50.20250514-git
Capstone: 6.0.0
Unicorn:  2.1.4
Pwnlib:   4.14.1
pwndbg> 
```

kernel (vmlinuz):
https://kernel.ubuntu.com/mainline/v6.17-rc3/amd64/linux-image-unsigned-6.17.0-061700rc3-generic_6.17.0-061700rc3.202508241638_amd64.deb

```
[run.sh]
#!/bin/sh
qemu-system-x86_64 \
  -M pc \
  -m 512M \
  -cpu kvm64,+smep,+smap \
  -kernel $(ls vmlinuz-*) \
  -drive file=rootfs.ext2,if=virtio,format=raw \
  -append "rootwait root=/dev/vda console=tty1 console=ttyS0 noapic" \
  -net nic,model=virtio \
  -net user \
  -nographic \
  -s

reset
```

### Cause

It seems that `find_module_all` is not being found.

```
pwndbg> set exception-verbose on
Set print a full stacktrace for exceptions raised in Pwndbg commands to 'on'.
pwndbg> pi pwndbg.aglib.kernel.kallsyms.get()
Detected Compressed Kallsyms
Traceback (most recent call last):
  File "<string>", line 1, in <module>
[...]

PR #3350, opened 2025-10-13
fix(kernel): Prevent klookup crash when symbol is not found

Prakhar0013:
Changes made:
Added a check for None in qword_mov_reg_ripoff method
Reason:
We just need to add a simple check at the beginning of the qword_mov_reg_ripoff function to handle the case where disass is None(This could be most probable).

I have successfully ran lint tests as well as the kernel tests which came to be successful, please check if this is right fix.
Fixes klookup fails #3349.

### DIFF
--- a/pwndbg/aglib/kernel/symbol.py
+++ b/pwndbg/aglib/kernel/symbol.py
@@ -368,6 +368,8 @@ class x86_64Symbols(ArchSymbols):
         return None
 
     def qword_mov_reg_ripoff(self, disass):
+        if disass is None:
+            return None
         result = self.regex(
             "".join(disass.splitlines()),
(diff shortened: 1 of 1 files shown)


Prakhar0013:
@patryk4815 , what is the issue? The errors are associated with heap management, not because of my code. Please check if this is okay.

jxuanli:
hi thanks for the PR, this fixes #3349 but doesn't address the other `disass()` calls might return None. The proper fix is to check if the function symbol actually can be resolved. see #3352

jxuanli:
also see #3349, i made a comment on what the underlying issue was for that particular issue

Prakhar0013:
Sure, thanks for letting me know

Prakhar0013:
Got to learn something

PR #3351, opened 2025-10-13
feat(dp): add MVP dp command (basic formatting + --tid & --depth) (Fi…

kkrishnan10:


disconnect3d:
What the hell is this? It looks like some AI/LLM generated trash. It doesn't even execute the `dprintf` command at all, it just prints some info and there is so much bloat (like redundant comments) here.

Please stop using AI for writing code and start using your brain. Also, please don't contribute to projects you don't inherently understand. It seems you haven't really run the command yourself and didn't think what is its purpose.

PR #3354, opened 2025-10-15
`kbpf` 

jxuanli:
- implemented `kbpf` to print information about the linux kernel bpf progs and maps
- added detection for `CONFIG_SECURITY`
- some more refactoring

## demo

image


jxuanli:
unfortunately, the kernel stores the bpf bytecode differently than userland input after verification, therefore some insns cannot be disassembled by `capstone`. maybe the capstone maintainers could look into this but I understand that is very niche and probably have a low priority (and it doesn't seem to have an ISA for kernel's internel bpf bytecode)

Issue #3348, opened 2025-10-11
Discussion: Mandatory venv installation makes it impossible to distribute pwndbg as a distribution package


hamarituc:
In df12edc0d5cb7885060199081efa13a386484a18 the support for the `.skip-venv` was dropped. Using a dedicated virtual environment makes it impossible to distribute pwndbg system-wide through the Linux distributions package manager.

I'm the maintainer of the Gentoo. Skipping the virtualenv check was my way to go then packaging pwndbg for Gentoo. In this case all of the pwndbg dependencies are provided system-wide through the package manager instead of relying that pwndbg installs its dependencies by itself.

I would appreciate if you could provide a way of packaging pwndbg for Linux distributions. From my point of view not only Gentoo will benefit from it.

See also https://github.com/pwndbg/pwndbg/issues/3124#issuecomment-2994541044

k4lizen:
see https://github.com/pwndbg/pwndbg/pull/3119

<img width="902" height="299" alt="Image" src="https://github.com/user-attachments/assets/cd773e85-252e-43b0-9531-9e3d56ffa4c0" />

@patryk4815 anything else necessary for packaging?

disconnect3d:
@hamarituc does the script entry point defined in pyproject.toml work for your case?

hamarituc:
> [@hamarituc](https://github.com/hamarituc) does the script entry point defined in pyproject.toml work for your case?

Yes it works. Thank you very much. I just missed the relevant information in #3119. Sorry for the noise. I will push the update during the next days.

k4lizen:
@hamarituc ping me when you do (and send the link to the gentoo package), I'll then update the docs to make the pwndbg packaging changes more apparent

k4lizen:
@anthraxx pinging you here because of the arch package, idk if you know about these changes

hamarituc:
> [@hamarituc](https://github.com/hamarituc) ping me when you do (and send the link to the gentoo package), I'll then update the docs to make the pwndbg packaging changes more apparent

The version bump in Gentoo is tracked in https://github.com/gentoo/gentoo/pull/44181.

PR #3355, opened 2025-10-16
[docs] Add packaging page to docs/tutorials

k4lizen:
image

re: https://github.com/pwndbg/pwndbg/issues/3348

Tbh I wonder if packagers are actually likely to find this page on their own, but at least we have something to link to.


disconnect3d:
Shouldn't this start with some paragraph that we package it with nix and link to some docs related to it? And then be some "If you want to package Pwndbg for ..." :x

k4lizen:
> Shouldn't this start with some paragraph that we package it with nix and link to some docs related to it? And then be some "If you want to package Pwndbg for ..." :x

I wanted this page to be specifically for distro maintainers who want to package pwndbg for their distro. Maybe if this isn't clear from the docs I should clarify it? But I think for them the fact that we package it as nix is not important to them.

Issue #2750, opened 2025-02-16
enforce consisent number of lines from section output (override context-*-lines ?)


k4lizen:
**Is your feature request related to a problem? Please describe.**
The output of pwndbg shakes when stepping quickly because the sections do not always output the same number of lines, particularly egregious is the disasm section due to jmp emulation and call arguments etc.

E.g.
![Image](https://github.com/user-attachments/assets/b655dedf-aa22-485f-b2d1-e33e34a80de9)
(3 lines)
![Image](https://github.com/user-attachments/assets/7a65eff6-ed72-4a5e-a216-c42809412039)
(2 lines)
are both from `set context-disasm-lines 1`.

**Describe the solution you'd like**
A way to make the pwndbg sections output a consistent number of lines. (it is fine if some line wraps due to terminal width, this issue is not about this)

**Describe alternatives you've considered**
Not sure whether it is better to just make the `context-*-lines` config options do this, or add something else. I am inclined towards changing the behaviour of context-*-lines.

disconnect3d:
That makes sense to me, although, I feel lime there may be some hard to support edge cases.

Also, should we really care about `set context-disasm-lines 1`? Why would you set it to such small number? :P

k4lizen:
> Also, should we really care about set context-disasm-lines 1? Why would you set it to such small number? :P

Ah I just did that so the screenshot is obvious, probably noone would use such a small number

disconnect3d:
@k4lizen was this maybe fixed in the PRs that did some fixes for context screen shaking?

k4lizen:
They did fix the main issue yeah. Though I think it would still be more pleasant for the eyes if you could force the sections to always have the same size so nothing jumps around.

Though tbh I think the disasm section is the biggest offender for this, and trying to actually implement this consistently for all sections may be non trivial, so it might be a good idea to only limit this feature for the disasm section for now.

disconnect3d:
I guess there are also two issues at hand here where we do not truncate 1) symbol names and 2) telescope chains which can get too long and span over few lines.

k4lizen:
I think for disasm at least, the sauce is in the fact that the `lines` parameter of the `pwndbg.aglib.nearpc` function means "give me this many instructions" instead of "give me this many lines", so for instance

![Image](https://github.com/user-attachments/assets/574b3d7d-9651-450a-99e2-3aa5183451c7)

branch empty lines aren't taken into account. A simple truncation should fix it though.

[...]

PR #3357, opened 2025-10-16
Fix nearpc lines parameter

OBarronCS:
This PR addresses a behavior of the `nearpc` and `emulate` commands relating to how many instructions are disassembled and displayed. The `lines` parameter didn't line up with the number of instructions that were displayed.

For example, `nearpc 10` doesn't disassemble only 10 lines, instead it could disassemble up to 21 instructions forwards, depending on how many instructions before the PC we can pull from the cache. In general, `nearpc N` really meant "display `2*N + 1` total instructions on the screen". It would read up to N instructions from the backward cache, and fill the remaining with forward instructions. 
https://github.com/pwndbg/pwndbg/blob/9a6f4434adf050f0068d9d37cfd1df41e2f6bbff/pwndbg/aglib/disasm/disassembly.py#L399

This made running `nearpc addr N` confusing, as more than N instructions were disassembled.

Additionally, disassembling a lot of instructions, like `nearpc 100`, meant you were disassembling and printing a huge number of instructions before the target PC, which is likely not the desired default behavior.

# New behavior

The internal `nearpc` and disassembly functions now require the caller to be explicit about how many instructions to disassembling forwards, how many to get from the cache, or many "total" instructions you want to display.

`nearpc N` will disassemble N future instructions, including the instruction at the PC. Use `nearpc 1` to disassemble one instruction (previously, you had to do `nearpc 0`). A new setting, `nearpc-backward-lines`, dictates how many previous instructions (instructions that led to the current PC) are to be displayed.

This can be overriden with the `-r` argument, `nearpc 60 -r 20`, lets to choose how many previous instructions to show (20 in this case).

To use the "old" behavior, where you want to display a certain number of instructions to the screen, with a dynamic number of instructions after the PC being disassembled, use the new `-t/--total` CLI arg to choose the total number of instructions to display.

(all references to `nearpc` here also apply to the `emulate` command).

Old:
image

New:
image


k4lizen:
maybe related: https://github.com/pwndbg/pwndbg/issues/2750#issuecomment-2835470807

disconnect3d:
@OBarronCS pls fix conflicts

disconnect3d:
@OBarronCS seems that some tests needs fixing here?

PR #3358, opened 2025-10-17
Fix #3299 - bugs in arm conditional instruction annotations

### LINKED ISSUE #3299: missing branch mark and following call? in `bl` instruction on arm32
missing branch mark and following call? in `bl` instruction on arm32

Image

OBarronCS:
This PR fixes the incorrect disassembly output seen here: https://github.com/pwndbg/pwndbg/issues/3299#issuecomment-3285329644

Two issues addressed:
- Although the branch was not taken (there's an ✘ on the bxne), the disassembly follows the path as if it was taken. This was caused by a bug in the code in how branch conditionality was determined. 
- It will no longer show annotations for conditional instructions (`ldrbne`, `movne`, `cmpne` in the linked screenshot) when the condition fails.

I added three tests for this.

### DIFF
--- a/pwndbg/aglib/disasm/instruction.py
+++ b/pwndbg/aglib/disasm/instruction.py
@@ -23,6 +23,7 @@ from capstone.arm import ARM_INS_TBH
 from capstone.loongarch import LOONGARCH_INS_ALIAS_JR
 from capstone.loongarch import LOONGARCH_INS_B
 from capstone.loongarch import LOONGARCH_INS_BL
+from capstone.loongarch import LOONGARCH_INS_CALL36
 from capstone.loongarch import LOONGARCH_INS_JIRL
 from capstone.mips import MIPS_INS_ALIAS_B
 from capstone.mips import MIPS_INS_ALIAS_BAL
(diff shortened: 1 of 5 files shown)


pwndbg/aglib/disasm/instruction.py
```diff
@@ -44,23 +45,24 @@
...
 from capstone.x86 import X86_INS_JMP
+from capstone.x86 import X86_INS_RET
 from capstone.x86 import X86Op
 from typing_extensions import override
 
 import pwndbg.dbg
 from pwndbg.dbg import DisassembledInstruction
 
 # Architecture specific instructions that mutate the instruction pointer unconditionally
-# The Capstone RET and CALL groups are also used to filter CALL and RET types when we check for unconditional jumps,
```
OBarronCS, line 60: This logic was incorrect. For example, in arm you can have the `blne` instruction, which is a CALL instruction, yet it is conditional. Same with `bxne   lr`, which is a conditional ret instruction.

pwndbg/aglib/disasm/instruction.py
```diff
@@ -314,21 +315,6 @@ def __init__(self, cs_insn: CsInsn) -> None:
         FALSE if the instruction has a conditional action, and we know it is not taken.
         """
 
-        self.declare_conditional: bool | None = None
```
OBarronCS, line 317: This field was unnecessary and just added another attribute to track conditionality which can be determined by other attributes.

disconnect3d:
The CI issues are unrelated

disconnect3d (approved):
lgtm

Issue #3214, opened 2025-08-07
improve function call detection


k4lizen:
**Is your feature request related to a problem? Please describe.**
Not sure how common this is in general buf or me on aarch64 inside PLT entries, functions are "branched" into (with a `br`) instruction rather than called into (`bl`). The x86_64 equivalents would be jmp and call. When this happens, pwndbg doesn't show function arguments. This makes sense in the general case because we want to avoid false-positives like showing arguments on if-statement branches etc.

**Describe the solution you'd like**
That being said, we could do a check like "is the destination of this branch a function (symbol)". That way we would avoid any false-positives and nicely cover this edge case.

**Describe alternatives you've considered**
Special case PLT entries somehow in case:
1. symbol lookup is too expensive
2. this doesn't happen in other situations

k4lizen:
Okay just encountered another example:
```c
int func1() {
    return func2();
}
```
this will emit a `b` instead of a `bl`.

disconnect3d:
> That being said, we could do a check like "is the destination of this branch a function (symbol)". That way we would avoid any false-positives and nicely cover this edge case.

Oh, that's a really neat idea. I like that.

disconnect3d:
Is this just about ARM or any architecture jump vs call?

k4lizen:
just checked: also reproduces for x86_64

<img width="729" height="167" alt="Image" src="https://github.com/user-attachments/assets/9e1b5e0a-3bfd-4eae-b83f-ae6d1c142f00" />

but yeah it should be for all arches

PR #3360, opened 2025-10-18
Improve function call detection

OBarronCS:
This implements the idea suggested in #3214 to show function call arguments in more cases. If an unconditional branch goes to a symbol, we infer that it is a function call.

Currently, function call arguments are shown for instructions explicitly denoted as call instructions - x86 `CALL`, arm `bl`, etc. Function calls can be invoked other branch types, however, such as in tail recursion and in PLT stubs.

image

cc @k4lizen

### DIFF
--- a/pwndbg/arguments.py
+++ b/pwndbg/arguments.py
@@ -33,6 +33,17 @@ from pwndbg.lib.arch import Platform
 from pwndbg.lib.functions import format_flags_argument
 
 
+# Sometimes function calls are not invoked with a "call instruction",
+# such as with tail-recursion or in PLT stubs
+def fallback_check_for_function_call(instruction: PwndbgInstruction) -> bool:
+    if instruction.is_unconditional_jump and instruction.target_string:
+        # Check if it is to the start of a function
+        # Otherwise, the symbol would contain +constant
+        # Like `main+132`
+        return "+" not in instruction.target_string
+    return False
+
+
 def get(instruction: PwndbgInstruction) -> List[Tuple[pwndbg.lib.functions.Argument, int]]:
     """
     Returns an array containing the arguments to the current function,
@@ -47,7 +58,7 @@ def get(instruction: PwndbgInstruction) -> List[Tuple[pwndbg.lib.functions.Argum
     if instruction.address != pwndbg.aglib.regs.pc:
         return []
 
-    if instruction.call_like:
+    if instruction.call_like or (fallback_check_for_function_call(instruction)):
         abi = pwndbg.aglib.arch.function_abi
 
         if abi is None:


OBarronCS:
On further thought, something more is needed to rule out false-positives. When writing assembly, it's common to create labels that you jmp to (our tests do this a lot).
```asm
jmp end
nop
nop

end:
nop
nop
```
The label `end` is the symbol of the branch target. We would need to rule this case out as well.

k4lizen:
## 1
I'm playing around with this to see if I can spot the difference:
```python
def fallback_check_for_function_call(instruction: PwndbgInstruction) -> bool:
    target = instruction.target
    tarsym = pwndbg.aglib.symbol.resolve_addr(target)
    if tarsym is not None:
        print("raw: ", tarsym)
        sym = pwndbg.aglib.symbol.lookup_symbol(tarsym, type=SymbolLookupType.FUNCTION)
        if sym is not None:
            print("human: ", sym.value_to_human_readable())
            if sym.type is not None:
                print("type: ", sym.type.name_to_human_readable)
                print("is func", sym.type.code == pwndbg.dbg_mod.TypeCode.FUNC)
                print("target?", sym.type.target())
                if sym.type.target() is not None:
                    print("target?", sym.type.target().name_to_human_readable)
                    print("tar is func?", sym.type.target().code == pwndbg.dbg_mod.TypeCode.FUNC)
```
main.c:
```c
#include <fcntl.h>

int meow() {
  return 0;
}

int main() {
  meow();
  int (*fptr)() = &meow;
  fptr();
  open("./doesntexist", 0);
  open("./doesntexist", 0);
  return 0;
}
```
jmpend.asm:
```asm
section .text
global _start

_start:
    jmp some_label
    nop
    nop

some_label:
    nop
    nop

    mov     rax, 60
    xor     rdi, rdi
    syscall
```
## 2
The `jmp some_label` looks like this regardless of debug info:
```
raw:  some_label
human:  0x401004 <some_label>
type:  <text variable, no debug info> *
is func False
target? <pwndbg.dbg.gdb.GDBType object at 0x7fe6c3f74ff0>
target? <text variable, no debug info>
tar is func? True
```
You can detect both the resolver jump and the plt jump:
image
image
image
image
using the human readable type name thing. It works with debug info, without debug info, and also for stripped binary somehow (maybe it depends on debug symbols from libc/ld?).
## 3
Furthermore, you can detect the `fptr();` as a function call if you have debug symbols, but not if you dont (using this information that i'm printing, i'm disregarding that we already properly detect that due to the `call` instruction).
## 4
The current PR also probably has some false negatives like `operator +(unsigned long)` in C++, but thats absolutely fine imo.
## 5
Tested it, it does depend on symbols from libc/ld. Which is nice for CTFs since people usually do, but doesn't help me with android debugging ;;

OBarronCS:
Closing this. Once #3908 is solved, this might be easier to implement in a uniform way. @k4lizen if you want to implement the ideas you had, please do open them in a new PR, because I can see the benefits of this.

PR #3361, opened 2025-10-18
Disassembly branch visualization

OBarronCS:
This PR adds a visualization of branch sources and targets, making it easier to visually track control flow. See the screenshots below for what it looks like.

Use the `-v` flag to nearpc, `nearpc -v`, to show the visualization:

This is currently a proof of concept - everything is hardcoded and the code is messy, which will be cleaned up. I wanted to get people playing around with it first to get some feedback.

Open questions:
- How to handle cases where the are a lot (dozen+) of branch ranges that overlap?
       *  Currently, with a hardcoded maximum of ~12 overlapping branches, it breaks, with the lines no longer lining up.
- Where in the display should the visualization go? All the way to the left, or closer to the assembly mnemonics?
      *   To test it out, you can add `-w 1` argument to nearpc to show the visualization directly next to the assembly instructions.

The credit for this idea goes to `radare2` - I saw some screenshots from [the tool](https://github.com/radareorg/radare2) and thought they were a great. I use the same unicode characters for the visualization as in radare2.

image

image

For testing, you can change where the branch visualizations go - add `-w 1` to put it directly next to the instructions:

`nearpc addr count -v -w 1`

image

image

image


OBarronCS:
Backward branches have dotted lines:
image

k4lizen:
>  * Currently, with a hardcoded maximum of ~12 overlapping branches, it breaks, with the lines no longer lining up.

I think the maximum should be something like 4 or 5. Can we make it consistently not break in that case?

k4lizen:
image

`ret` shouldn't show a branch

also, I think:
1. Showing the branches next to the disassembly instructions makes it more easy to spot what is going where and why because you can see the instructions
2. Showing the branches next to the instructions makes it a bit too crowded

So I think its maybe better if they are on the left

k4lizen:
any thoughts on adding this to our disassembly context? not sure if it would play well though considering that we dont show a linear view of the memory

OBarronCS:
This now should be ready to go.

Add the `-v` flag to nearpc to display the branch visualization. 

Example:

image


The branch visualization takes 20 columns of space, and can show up to 19 overlapping jumps. If there are any more jumps, they are clumped into the left-most column.

image


> any thoughts on adding this to our disassembly context? not sure if it would play well though considering that we dont show a linear view of the memory

I think only really makes sense for linearly disassembled instructions,  as "start" and "end" of branches no longer 
make spatial sense because you could be jumping all over the place

pwndbg/aglib/nearpc.py
```diff
@@ -166,9 +455,12 @@ def nearpc(
     symbols = [f"<{sym}> " if sym else "" for sym in symbols]
 
     # Pad out all of the symbols and addresses
-    if pwndbg.config.left_pad_disasm and not repeat:
```
OBarronCS, line 169: `and not repeat` doesn't make sense here - it removes alignment of the output if you running nearpc back to back.

k4lizen:
image

Hmm would be awesome if the branches would flow nicely between repeat invocations of the command, but that feels like it might be a rabbit hole that complicates stuff a lot.

pwndbg/commands/nearpc.py
```diff
@@ -48,11 +48,25 @@
     help="Whether to emulate instructions to find the next ones or just linearly disassemble.",
 )
 
+parser.add_argument(
+    "-v",
```
k4lizen: idk if `-v` is the best argument since it usually means verbose, maybe `-b` would make more sense

k4lizen (approved):
LGTM

k4lizen:
oh yeah btw, I feel this is a pretty cool feature that will not be very discoverable. any reason not to enable it by default (for the nearpc command)?

OBarronCS:
Switched to using `-b` to enable it.

It now works better across repeated uses of nearpc:

image

If on repeated uses you come across a backwards branch that goes back to before the current region starts, it adds a special character to indicate this 
image

It's not perfect, as between nearpc disassembled regions, the symbols can be different widths, which breaks the padding. But it's much better than before, and this should only really happened when disassembling across large regions and across functions where the symbol changes.

image

k4lizen:
Very nice. Are you up for making it enabled by default? (Then I guess `-b` would disable :p)

OBarronCS:
Yes - made one last change to enable it by default (I was concerned about having a bunch of padding around for no reason)

Now it's enabled by default, and if there are no branches to display, it doesn't have the empty space:
image

Now, you add `-n` to disable it.

k4lizen (approved):

tests/binaries/host/branch_visualization.x86-64.asm
```diff
@@ -0,0 +1,27 @@
+global _start
+
+; This binary is there to test
+; emulate vs nearpc/u/pdisas commands
+; The emulate should show just jump and one nop
+; The rest should show jump and two nops
+;
+; Motivated by https://github.com/pwndbg/pwndbg/issues/315
```
k4lizen: is this comment intentional?

OBarronCS: oop good catch, copied that from another file

PR #3369, opened 2025-10-23
Bug fix: variable accidentally overwritten

OBarronCS:
A loop variable had the same name as the variable in the function, accidentally overwriting it.

A couple lines down, at https://github.com/pwndbg/pwndbg/blob/42e8af33d2411109f39078e82b20cbc7a852f478/pwndbg/color/syntax_highlight.py#L85-L86

the variable "filename" is used again. This was accidentally cobbled by the for loop variable.

This was causing some file types to not get syntax highlighting.


k4lizen:
you may try to enable the lint rule using the ruff code https://docs.astral.sh/ruff/rules/redefined-argument-from-local/

PR #3379, opened 2025-10-28
Tracing kernel memory management

jxuanli:
- added `kmem-trace` to trace the SLUB and buddy allocators for a given (arbitrary) command
- implemented `trace_ret`, `runcmd`, and `ctx_suspend_once` dbg interfaces

## demo
image
image


jxuanli:
This command works as intended for GDB. For LLDB, I think there are a few bugs unrelated to this PR that makes this command fail when using LLDB. See #3380 and #3381. 

I also believe there are a few functions, particularly for freeing pages, that I did not include in `pfree_names`, so I will mark this PR as a draft first to allow time for people to experiment with it.

jxuanli:
Another thing I'd want some inputs is that for `[PAGE/SLAB ALLOC]`, due to the way I am tracing the relevant functions' return values, the backtrace does not include the actual function name for allocating the page/object. For example, `kmem_cache_alloc` may be the actual function that allocates the `task_struct` but not shown.

pwndbg/dbg/__init__.py
```diff
@@ -638,6 +645,12 @@ def add_symbol_file(self, path, base=None):
         """
         raise NotImplementedError()
 
+    def runcmd(self, cmd):
+        """
+        Runs a debugger command
+        """
```
k4lizen, line 651: this use-case makes sense because you're running a user provided command, but maybe add a warning or something so people don't fall into the trap of doing stuff like

```
if is_gdb:
  dbg.runcmd("b *{addr}")
else:
   # lldb
  dbg.runcmd("br *{addr}")
```
or whatever

pwndbg/commands/kmem_trace.py
```diff
@@ -0,0 +1,255 @@
...
+import pwndbg.arguments
+import pwndbg.color as C
+import pwndbg.color.message as M
+import pwndbg.commands
+from pwndbg.dbg import BreakpointLocation
+
+parser = argparse.ArgumentParser(
+    description="""
+    Tracing kernel memory (SLUB and buddy) allocations and frees.
+    Unless --all is specified, only the (de)allocations triggered by the current frame will be printed.
```
k4lizen: what is a "frame"? is it a stack frame or some other kernel-specific concept?

k4lizen:
this is pretty awesome

k4lizen:
i think it would be more readable if the output was column-aligned. its fine if there are some pathological caches with really long names that aren't aligned as long as most of them are

k4lizen:
it would be cool if we could port this logic to the `track-heap` command as well, but thats out of the scope of this pr

pwndbg/commands/kmem_trace.py
jxuanli: it is a stack frame yes

k4lizen: maybe more clear wording would be "until the current function returns" :?

pwndbg/aglib/regs.py
```diff
@@ -105,6 +105,9 @@ class module(ModuleType):
 
     @pwndbg.lib.cache.cache_until("stop", "prompt")
     def read_reg(self, reg: str, frame: pwndbg.dbg_mod.Frame | None = None) -> int | None:
+        return self.read_reg_uncached(reg, frame)
```
disconnect3d, line 108: Each first register read is now one more call. Since we cache this, I feel its probably fine.

For what is worth, this could probably also be solved by doing:
```
read_reg = pwndbg.lib.cache_until(Class.read_read_reg_uncached)
```

or something like this, though I am not sure if the `Class` would work here ;)


Anyway, its all fine here as it is here.

pwndbg/commands/kmem_trace.py
disconnect3d: Can we inform the user when its worth to actually execute it? Like some use cases? Would be nice to add some documentation page with an example about it, heh

pwndbg/commands/kmem_trace.py
```diff
@@ -0,0 +1,269 @@
...
+
+
+class KmemTracepoints:
+    def __init__(self):
+        # try to capture the lowest possible level of exported functions in the (de)alloc chain
+        # for example __alloc_pages_bulk calls __alloc_pages and only __alloc_pages is included
+        # lists might not be complete
+        # try to resolve all names, if does not exist, means it is not exported for that version
+        kmalloc_names = (
+            "__kmalloc",
```
disconnect3d, line 103: Can we add a kernel version in comment from which we constructed this list?

pwndbg/commands/kmem_trace.py
```diff
@@ -0,0 +1,269 @@
...
+                sp = inf.break_at(bp, KmemTracepoints.kfree_handler, internal=True)
+                self.sps.append(sp)
+        if self.buddy_tracepoints_enabled:
+            for palloc in self.pallocs:
+                bp = BreakpointLocation(palloc)
+                sp = inf.break_at(bp, KmemTracepoints.palloc_handler, internal=True)
+                self.sps.append(sp)
+            for pfree in self.pfrees:
+                bp = BreakpointLocation(pfree)
+                sp = inf.break_at(bp, KmemTracepoints.pfree_handler, internal=True)
```
disconnect3d, line 222: I wonder if we can simplify the interface to not have to call BpLocation everywhere to call break_at huh

disconnect3d:
@jxuanli @k4lizen let me know when this is ready for review.

Btw is there some easy way to test kernel things? Like some single command I can use to reproduce kernel build and run it against Pwndbg?

I think @patryk4815 has some nix env for this too. We should add some documentation around that if possible

pwndbg/aglib/regs.py
jxuanli: I believe `pwndbg.lib.cache_until` takes in a string

pwndbg/commands/kmem_trace.py
jxuanli: I believe one of the purposes of the `BreakpointLocation` is for `break_at` to handle both `breakpoint` and `watchpoint`

jxuanli:
@disconnect3d ready for review

Issue #2659, opened 2024-12-31
heap commands: get_file called with incorrect path


sasha-999:
### Description

```
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /home/sasha/gdb/pwndbg_2025/pwndbg/commands/__init__.py:183 in __call__                          │
│                                                                                                  │
│   180 │                                                                                          │
│   181 │   def __call__(self, *args: Any, **kwargs: Any) -> str | None:                           │
│   182 │   │   try:                                                                               │
│ ❱ 183 │   │   │   return self.function(*args, **kwargs)                                          │
│   184 │   │   except TypeError:                                                                  │
│   185 │   │   │   print(f"{self.function.__name__.strip()!r}: {self.function.__doc__.strip()}"   │
│   186 │   │   │   pwndbg.exception.handle(self.function.__name__)                                │
│                                                                                                  │
│ /home/sasha/gdb/pwndbg_2025/pwndbg/commands/__init__.py:514 in _OnlyWithResolvedHeapSyms         │
│                                                                                                  │
│   511 │   │   │   ):                                                                             │
│   512 │   │   │   │   # In auto mode, if the debug symbols are not enough, we will try to use    │
│   513 │   │   │   │   heuristic_heap = HeuristicHeap()                                           │
│ ❱ 514 │   │   │   │   if heuristic_heap.can_be_resolved():                                       │
│   515 │   │   │   │   │   pwndbg.aglib.heap.current = heuristic_heap                             │
│   516 │   │   │   │   │   w(                                                                     │
│   517 │   │   │   │   │   │   "pwndbg will try to resolve the heap symbols via heuristic now s   │
│                                                                                                  │
│ /home/sasha/gdb/pwndbg_2025/pwndbg/aglib/heap/ptmalloc.py:1734 in can_be_resolved                │
│                                                                                                  │
│   1731 │   │   return self._structs_module                                                       │
│   1732 │                                                                                         │
│   1733 │   def can_be_resolved(self) -> bool:                                                    │
│ ❱ 1734 │   │   return self.struct_module is not None                                             │
│   1735 │                                                                                         │
│   1736 │   @property                                                                             │
│   1737 │   def main_arena(self) -> Arena | None:                                                 │
│                                                                                                  │
│ /home/sasha/gdb/pwndbg_2025/pwndbg/aglib/heap/ptmalloc.py:1724 in struct_module                  │
│                                                                                                  │
[...]

PR #3386, opened 2025-10-29
fix get_file() checking relative paths

mowteam:
This PR fixes issues mentioned in #2659. Delete the path prefix checks in `get_file()` since POSIX standards do not require a `./` or `../` in front of relative paths. Invalid paths should/will be caught by their respective file open operations.

### DIFF
--- a/pwndbg/aglib/file.py
+++ b/pwndbg/aglib/file.py
@@ -82,10 +82,6 @@ def get_file(path: str, try_local_path: bool = False) -> str:
         The local path to the file
     """
     has_target_prefix = path.startswith("target:")
-    has_good_prefix = path.startswith(("/", "./", "../")) or has_target_prefix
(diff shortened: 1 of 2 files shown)


mowteam:
For reference, the relative path check was originally added in commit 4490705. From the commit message and original code, it appears to be a temporary solution building on top of the original absolute path check. Since we have come full circle and allow absolute paths and any relative path, it seems fair to remove this code.

disconnect3d:
@mowteam is there any chance you could add a testcase which will show the case that this fixes, so we don't reintroduce this bug in the future (and this will also welcome us to have more tests in this area in the future)?

mowteam:
I added a test case for get(). I wasn't completely sure if I should test the actual get() function or the heap's use of it. For simplicity, I went the the former option, but let me know if I should change how I do the test.

PR #3389, opened 2025-11-05
Update one-liner install command in docs

OBarronCS:
This updates the one-liner shell command indicated in the docs to install pwndbg via curl. See comments for specified details.

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -10,7 +10,7 @@ There are multiple ways to install Pwndbg, depending on whether you want to use 
 ## Installing pwndbg-gdb
 Install via curl/sh (Linux/macOS)
 ```{.bash .copy}
-curl -qsL 'https://install.pwndbg.re' | sh -s -- -t pwndbg-gdb
+curl --proto '=https' --tlsv1.2 -LsSf 'https://install.pwndbg.re' | sh -s -- -t pwndbg-gdb
 ```
 Install via Homebrew (macOS)
 ```{.bash .copy}
@@ -68,7 +68,7 @@ command.
 
 Install via curl/sh (Linux/macOS)
 ```{.bash .copy}
-curl -qsL 'https://install.pwndbg.re' | sh -s -- -t pwndbg-lldb
+curl --proto '=https' --tlsv1.2 -LsSf 'https://install.pwndbg.re' | sh -s -- -t pwndbg-lldb
 ```
 Install via Homebrew (macOS)
 ```{.bash .copy}


docs/setup.md
```diff
@@ -10,7 +10,7 @@ There are multiple ways to install Pwndbg, depending on whether you want to use
 ## Installing pwndbg-gdb
 Install via curl/sh (Linux/macOS)
 ```{.bash .copy}
-curl -qsL 'https://install.pwndbg.re' | sh -s -- -t pwndbg-gdb
+curl --proto '=https' --tlsv1.2 -LsSf 'https://install.pwndbg.re' | sh -s -- -t pwndbg-gdb
```
OBarronCS, line 13: I believe the -q flag was used mistakenly (for `wget` it turns off the output, but for curl the -q disables a [local config file](https://man7.org/linux/man-pages/man1/curl.1.html#:~:text=%2D%2Dnegotiate.-,%2Dq%2C%20%2D%2Ddisable,-If%20used%20as)) .

-s makes output silent
-S includes output in case of download error (nice error message if the domain doesn't resolve)
-f means that if there is an HTTP error code (like the link goes to a 404 website), it fails. Otherwise, it would attempt to execute the 404 HTML page contents as a shell script

The `--proto '=https' --tlsv1.2` part is common in these one-liner installs that are meant to be copy-pasted (see [Rust](https://rust-lang.org/tools/install/#:~:text=Using%20rustup%20(Recommended)), [Ruff](https://github.com/astral-sh/ruff/releases#:~:text=Install%20ruff%200.14.3-,Install%20prebuilt%20binaries%20via%20shell%20script,-curl%20%2D%2Dproto%20%27))

patryk4815: `--proto '=https' --tlsv1.2` long 🤔

patryk4815: I think `--tlsv1.2` is not needed anymore.
> The minimum acceptable version is by default TLS v1.2 since [8.16.0](https://curl.se/ch/8.16.0.html) (unless the TLS library has a stricter rule).
commit: https://github.com/curl/curl/commit/9d8998c99408e1adf8eba629fad9f87b3235bdfa

`--proto '=https'` probably is needed, if our server redirect from https -> http 🤔 but it is issue on our end

disconnect3d: @patryk4815 pls send anothe rPR if there is anything to be changed here; I think we can safely add those flags? Merged this

OBarronCS: > I think `--tlsv1.2` is not needed anymore.
> 
> > The minimum acceptable version is by default TLS v1.2 since [8.16.0](https://curl.se/ch/8.16.0.html) (unless the TLS library has a stricter rule).
> > commit: [curl/curl@9d8998c](https://github.com/curl/curl/commit/9d8998c99408e1adf8eba629fad9f87b3235bdfa)
> 
> `--proto '=https'` probably is needed, if our server redirect from https -> http 🤔 but it is issue on our end

tslv.12 being default is great to hear - although it'll take some time to propagate to most distros. Although the --proto adds some length, I believe the main use case of these types of commands is to copy-paste them to install. Having the familiarity of seeing the same pattern of flags in pwndbg's install instructions as in different projects' `curl | bash` install instructions is worth it, in my opinion.

Issue #3391, opened 2025-11-06
TypeError in handle_kernel_pages when connecting to kernel


k4lizen:
### Description

I get this:
```
pwndbg: loaded 214 pwndbg commands. Type pwndbg [filter] for a list.
pwndbg: created 13 GDB functions (can be used with print/break). Type help function to see them.
Reading symbols from vmlinux...
0x0000000003a2b589 in ?? ()
------- tip of the day (disable with set show-tips off) -------
Use the procinfo command for better process introspection (than the GDB's info proc command)
int too big to convert
context: Print out the currently enabled context sections.

This is the text that gets printed on every stop. It can be useful
to run this command manually when you change some process/debugger
state but don't want to step/continue (e.g. after using the `down`
and `up` commands).
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /pwndbg/commands/__init__.py:370 in __call__                          │
│                                                                                                  │
│   367 │                                                                                          │
│   368 │   def __call__(self, *args: Any, **kwargs: Any) -> str | None:                           │
│   369 │   │   try:                                                                               │
│ ❱ 370 │   │   │   return self.function(*args, **kwargs)                                          │
│   371 │   │   except TypeError:                                                                  │
│   372 │   │   │   print(f"{self.command_name}: {self.description}")                              │
│   373 │   │   │   pwndbg.exception.handle(self.function.__name__)                                │
│                                                                                                  │
│ /pwndbg/commands/context.py:765 in context                            │
│                                                                                                  │
│    762 │   │   │   │   result_settings[target].update(settings)                                  │
│    763 │   │   │   │   with target as out:                                                       │
│    764 │   │   │   │   │   result[target].extend(                                                │
│ ❱  765 │   │   │   │   │   │   func(                                                             │
│    766 │   │   │   │   │   │   │   target=out,                                                   │
│    767 │   │   │   │   │   │   │   width=settings.get("width", None),                            │
│    768 │   │   │   │   │   │   │   with_banner=settings.get("banner_top", True),                 │
│                                                                                                  │
│ /pwndbg/commands/context.py:392 in _serve_context_history             │
│                                                                                                  │
│    389 │   │   │   ):                                                                            │
│    390 │   │   │   │   current_output = context_history[section_name][-1]                        │
│    391 │   │   │   else:                                                                         │
│ ❱  392 │   │   │   │   current_output = function(*a, **kw)                                       │
[...]

PR #3392, opened 2025-11-06
Cleanup kmem-trace docs a bit

k4lizen:
Also some type-checking-related stuff.


k4lizen:
I played around with using the command and was quite confused for a bit so I think this clarifies some stuff up

jxuanli:
why do we need those asserts? for example, `pwndbg.aglib.regs.retval` is just the ret register name which should always be there for supported archs.

jxuanli:
but otherwise looks good

k4lizen:
> why do we need those asserts? for example, pwndbg.aglib.regs.retval is just the ret register name which should always be there for supported archs.

if it's always there then it shouldn't return `str | None` no?

```python
    @property
    def retval(self) -> str | None:
        return reg_sets[pwndbg.aglib.arch.name].retval
```

k4lizen:
To clarify a bit, if you open the file with an LSP with typing checks you get this
image

k4lizen:
And probably the bug in https://github.com/pwndbg/pwndbg/issues/3391 is a result of not taking care of a returned None :>

docs/commands/kernel/kmem-trace.md
```diff
@@ -6,19 +6,27 @@ usage: kmem-trace [-h] [-s] [-b] [-v] [-c COMMAND] [--all]
 
 ```
 
-Tracing kernel memory (SLUB and buddy) allocations and frees.
-    Unless --all is specified, only the allocations triggered by the until the function returns will be printed.
-    This option may be helpful if you also want to trace frees scheduled with rcu or if the traced command steps out of the current function.
+Trace kernel memory (SLUB and buddy) allocations and frees.
+
+This command will execute `next` in the debugger, and print out all (de)allocations that happen until
+the command finishes. As such this makes most sense to call when the PC is on a branch. Only (de)allocations
```
disconnect3d: "on a branch" - maybe "on a function call instruction" is more clear?

docs/commands/kernel/kmem-trace.md
```diff
@@ -6,19 +6,27 @@ usage: kmem-trace [-h] [-s] [-b] [-v] [-c COMMAND] [--all]
...
-|-b|--trace-buddy|enable buddy allocator tracing|
+|-s|--trace-slab|do only slab allocator tracing|
+|-b|--trace-buddy|do only buddy allocator tracing|
 |-v|--verbose|print backtraces|
-|-c|--command|command to be traced (e.g. `n`, `nextret`) (default: 'n')|
+|-c|--command|trace during the execution of this command (default: 'next')|
 ||--all|display ALL memory allocations/frees regardless if they are triggered by the current function.|
 
+### Notes
+--all may be helpful if you also want to trace frees scheduled with rcu or if the traced command steps out of the current function.
```
disconnect3d: Shouldn't this be `--all` - with inline code highlighting?
```suggestion
The `--all` flag may be helpful if you also want to trace frees scheduled with rcu or if the traced command steps out of the current function.
```

docs/commands/kernel/kmem-trace.md
```diff
@@ -6,19 +6,27 @@ usage: kmem-trace [-h] [-s] [-b] [-v] [-c COMMAND] [--all]
...
+You may also find `-c finish` and `-c continue` useful.
```
disconnect3d: or `-c nextret` maybe, but idk

k4lizen: i can type `finish` faster so I'm biased :>

pwndbg/commands/kmem_trace.py
```diff
@@ -42,8 +56,17 @@ def __init__(self, verbose, trace_all):
...
+            if inf:
+                parent = inf.parent()
+                if parent:
+                    pc = parent.pc()
+                    symbol_name = pwndbg.aglib.symbol.resolve_addr(pc)
+                    if symbol_name:
+                        self.curr = symbol_name.split("+")[0]
+
+            if self.curr is None:
+                print(M.warn("Couldn't locate frame properly. Tracing --all."))
```
disconnect3d, line 66:
```py
        if not trace_all:
            # current frame only accounting for jumps
            if inf := pwndbg.dbg.selected_frame():
                if parent := inf.parent():
                    pc = parent.pc()
                    if symbol_name := pwndbg.aglib.symbol.resolve_addr(pc):
                        self.curr = symbol_name.split("+")[0]

            if self.curr is None:
                print(M.warn("Couldn't locate frame properly. Tracing --all."))
```

maybe walrus?:)

pwndbg/commands/kmem_trace.py
```diff
@@ -185,9 +212,12 @@ def _palloc_handler() -> bool:
 
     @staticmethod
     def palloc_handler(sp: pwndbg.dbg_mod.StopPoint) -> bool:
+        inf = pwndbg.dbg.selected_inferior()
+        assert inf
```
disconnect3d, line 213: Does this make sense? if `inf` is not there, this will crash anyway with TypeError - executing method on NoneType?

pwndbg/commands/kmem_trace.py
```diff
@@ -235,23 +266,51 @@ def get_kmem_tracepoints():
...
+    parser,
+    category=pwndbg.commands.CommandCategory.KERNEL,
+    notes="""
+--all may be helpful if you also want to trace frees scheduled with rcu or if the traced command steps out of the current function.
+You may also find `-c finish` and `-c continue` useful.
+""",
+    # FIXME: The -c option
+    #     default="next",
+    # is not portable.
+    only_debuggers={DebuggerType.GDB, DebuggerType.LLDB},
```
disconnect3d, line 276: Oh, wow? Which debugger does not support it then?:P

k4lizen: windbg

pwndbg/commands/kmem_trace.py
k4lizen: im a bit torn on this because
1. i believe there is a difference between explicitly documenting an expectation and having it implicit
2. doing this satisfies type correctness hints
3. the assert will give a worse looking backtrace to the user than a TypeError would

k4lizen: really, I wish the documentation for these functions
```python
    def selected_inferior(self) -> Process | None:
        """
        The inferior process currently being focused on in this interactive session.
        """
        raise NotImplementedError()

    def selected_thread(self) -> Thread | None:
        """
        The thread currently being focused on in this interactive session.
        """
        raise NotImplementedError()

    def selected_frame(self) -> Frame | None:
        """
        The stack frame currently being focused on in this interactive session.
        """
        raise NotImplementedError()
```
would explicitly say *when* they can return `None` so I can know if that invariant is satisfied when I'm writing some code or not.
@mbrla0 
is `selected_inferior` being non-None equivalent to `@pwndbg.commands.OnlyWhenRunning` ?

disconnect3d: We should add this docs then; also, the `x=Sth; assert x` is code that executes/its not optimized, so that's why I am a bit hesitant, but we can have it for now too..

k4lizen: i think a potentially nice idea to remedy this is to just have selected_inferior() always return `Process`, and if it can't get a valid process it throws an exception instead of returning None.

and then the few places which actually need to check that a valid inferior exists can catch that exception

this way we have both type safety and performance and less verbose code.

PR #3394, opened 2025-11-07
Scanning page tables but on steroids

jxuanli:
- achieved 25% speed up for x64 and more than 10x speed up for aarch64 compared to `gdb-pt-dump` 
  - uses `Qqemu.PhyMemMode` (work on gdb and lldb)
- on average takes less than 0.09 seconds to complete for x64 and 0.12 for aarch64
  - `gdb-pt-dump` takes ~0.12 for x64 and a few seconds for aarch64 
- checked output against `gdb-pt-dump`
- also doesn't require `ptrace` since this reads the physmem with qemu support

Note that the `gdb-pt-dump` dependency is not removed by this PR to allow time for experimentation. This PR adds the option `pt-dump` to the config `kernel-vmmap`. When set, `gdb-pt-dump` is used instead. 

The `gdb-pt-dump` dependency should be removed in a subsequent PR.


pwndbg/aglib/kernel/vmmap.py
```diff
@@ -386,20 +387,23 @@ def kernel_vmmap_via_monitor_info_mem() -> Tuple[pwndbg.lib.memory.Page, ...]:
...
-    enum_sequence=["page-tables", "monitor", "none"],
+    enum_sequence=["page-tables", "pt-dump", "monitor", "none"],
 )
 
 
 @pwndbg.lib.cache.cache_until("stop")
-def kernel_vmmap_pages() -> Tuple[pwndbg.lib.memory.Page, ...]:
+def kernel_vmmap_pages() -> Tuple[Page, ...]:
     if kernel_vmmap_mode == "page-tables":
+        return pwndbg.aglib.kernel.pagetable_scan()
```
jxuanli: this is only supported for `x86-64` and `aarch64`, should we resort to `monitor` for the other archs? tbh, only `x86-64` and `aarch64` kernel debugging is well supported by us atm

pwndbg/aglib/kernel/paging.py
```diff
@@ -156,7 +267,16 @@ def pagewalk_helper(self, target, entry) -> Tuple[PageTableLevel, ...]:
         )
         return tuple(result)
 
-    def pageentry_flags(self, level) -> BitFlags:
+    def pagetable_scan_helper(self, entry, is_kernel=None) -> List[Page]:
+        # only two possible return values: https://qemu-project.gitlab.io/qemu/system/gdb.html
+        oldval = pwndbg.dbg.selected_inferior().send_remote("qqemu.PhyMemMode").decode()
+        pwndbg.dbg.selected_inferior().send_remote("Qqemu.PhyMemMode:1")
```
jxuanli: this would occasionally mess with the gdbserver -- if the pagescan fails, the value of `PhyMemMode` is never restored so the gdbserver will keep operate under `PhyMemMode = 1` when reconnected. But not sure if there is a better way to do this. PhysMemMode is need because invoking `self.physmap` results in cyclic calls.

k4lizen:
> as it is used to display context

where is the page table crawling relevant for the context btw? for coloring pointers according to page permissions, or something more fundamental?

jxuanli:
it is used to generate address markers (search for places where `kernel_vmmap_pages` gets invoked) which is used to categorize pointers (stack, heap, etc)

pwndbg/aglib/kernel/vmmap.py
```diff
@@ -386,23 +387,32 @@ def kernel_vmmap_via_monitor_info_mem() -> Tuple[pwndbg.lib.memory.Page, ...]:
...
 
 @pwndbg.lib.cache.cache_until("stop")
-def kernel_vmmap_pages() -> Tuple[pwndbg.lib.memory.Page, ...]:
-    if kernel_vmmap_mode == "page-tables":
-        return kernel_vmmap_via_page_tables()
-    elif kernel_vmmap_mode == "monitor":
-        return kernel_vmmap_via_monitor_info_mem()
+def kernel_vmmap_pages() -> Tuple[Page, ...]:
+    match kernel_vmmap_mode:
+        case "page-tables":
```
jxuanli, line 410: this is not supported for riscv, but in general we don't really support debugging riscv kernels (the support is very limited atm), should we fallback to `monitor`? what do you think @patryk4815

jxuanli:
ready for review

pwndbg/aglib/kernel/paging.py
disconnect3d: Can we tell this to the user somehow? Shall we document it somewhere? in `vmmap`?

pwndbg/aglib/kernel/paging.py
```diff
@@ -175,7 +286,19 @@ def pagewalk_helper(self, target, entry) -> Tuple[PageTableLevel, ...]:
...
+        try:
+            scan = PageTableScan(self, is_kernel)
+            scan.scan(entry, self.paging_level)
+        except Exception:  # so that the PhyMemMode value is always restored
+            pass
+        pwndbg.dbg.selected_inferior().send_remote(f"Qqemu.PhyMemMode:{oldval}")
```
disconnect3d: Should this be in `final`? If lines 294-295 fail with an error that is not inherited from `Exception`, we will not call it 😓

pwndbg/aglib/kernel/paging.py
```diff
@@ -175,7 +286,19 @@ def pagewalk_helper(self, target, entry) -> Tuple[PageTableLevel, ...]:
...
         return tuple(result)
 
-    def pageentry_flags(self, level) -> BitFlags:
+    def pagetable_scan_helper(self, entry, is_kernel=None) -> List[Page]:
+        # only two possible return values: https://qemu-project.gitlab.io/qemu/system/gdb.html
+        oldval = pwndbg.dbg.selected_inferior().send_remote("qqemu.PhyMemMode").decode()
+        pwndbg.dbg.selected_inferior().send_remote("Qqemu.PhyMemMode:1")
+        try:
+            scan = PageTableScan(self, is_kernel)
+            scan.scan(entry, self.paging_level)
```
disconnect3d, line 269: Shouldn't `PageTableScan` be just a function that returns what we now have in `scan.result`?

pwndbg/aglib/kernel/paging.py
```diff
@@ -339,16 +465,32 @@ def handle_kernel_pages(self, pages):
                 page.objfile = "kernel [stack]"
 
     def pagewalk(self, target, entry) -> Tuple[PageTableLevel, ...]:
+        # kpti is not an issue here
```
disconnect3d: What do we mean? When would it be an issue?

pwndbg/aglib/kernel/paging.py
jxuanli: there are quite a few internal states needs to be kept track of so I thought its better to have it in a separate class

pwndbg/aglib/kernel/paging.py
jxuanli: I thought it would cause some unexpected output for aarch64 when the target pgd is set with `kcurrent`. however, i dont think that would be an issue after investigating how the TTBR registers work a bit further

fyi, I think TTBR1_EL1 is static for all tasks which makes sense

pwndbg/aglib/kernel/paging.py
jxuanli: fwiw, i moved the core implementation of pagewalk into the class as well to reuse the same cache building mechanism

pwndbg/aglib/kernel/symbol.py
```diff
@@ -275,7 +275,7 @@ def __init__(self):
         self.bpf_map_heuristic_func = "bpf_map_free_id"
         self.current_task_heuristic_func = "common_cpu_up"
 
-    def disass(self, name, lines=5):
+    def disass(self, name, lines=10):
```
disconnect3d, line 284: Shall this change be here?

pwndbg/aglib/kernel/vmmap.py
```diff
@@ -270,11 +271,11 @@ def _parser_mem_info_line_x86(line: str) -> pwndbg.lib.memory.Page | None:
...
     flags = 0
     if "r" in perm:
-        flags |= 4
+        flags |= Page.R_OK
     if "w" in perm:
-        flags |= 2
+        flags |= Page.W_OK
     if "x" in perm:
-        flags |= 1
+        flags |= Page.X_OK
```
disconnect3d, line 278: Its probably better to leave it with 4 2 1 and leave some comments --- if we want it to be slightly faster but maybe it doesn't matter.

jxuanli: I benchmarked the dereference (over millions of those dereferences) I dont think the result is any different than using constants

pwndbg/aglib/kernel/symbol.py
jxuanli: yes tho maybe the definition of `nearpc` should also be changed (default value for `lines`)

disconnect3d:
Hmm `test_x64_extra_registers_under_kernel_mode` fails after this:
> FAILED tests/library/qemu_system/tests/test_commands_kernel.py::test_x64_extra_registers_under_kernel_mode - TypeError: argument of type 'module' is not iterable

tests/library/qemu_system/tests/test_commands_kernel.py
```diff
@@ -139,12 +141,7 @@ def test_x64_extra_registers_under_kernel_mode():
         assert reg.upper() in res
     # those are the most important ones, and their presence should indicate it's working as intended
     for flag in ["smep", "smap", "wp"]:
-        assert flag in res or flag.upper() in res
-
-
-def get_slab_freelist_elements(out):
-    out = pwndbg.color.strip(out)
-    return re.findall(r"- \[0x[0-9a-fA-F\-]{2}\] (0x[0-9a-fA-F]+)", out)
+        assert flag in res or flag.upper() in re
```
k4lizen:
```suggestion
        assert flag in res or flag.upper() in res
```

disconnect3d:
```suggestion
        assert flag in res or flag.upper() in res
```

disconnect3d: > You do not have permission to apply this suggestion.

buuu :P

jxuanli:
yes sorry guys i removed one extra char when updating the tests :[

PR #3398, opened 2025-11-09
Support system Zig in addition to the one bundled in a Python package

carlsmedstad:
Add support for locating the Zig executable with the following precedence:
~1. ZIGCC environment variable - allows explicit override.~
2. ziglang module - if installed, use bundled Zig.
3. zig in PATH - fallback to system installation.

On Arch Linux we don't package the ziglang Python package. This change makes it possible for pwndbg to use the Zig executable from our zig0.14 package [0].

[0]: https://archlinux.org/packages/extra/x86_64/zig0.14/

Disclaimer: Authored with assistance from Claude Code.

### DIFF
--- a/pwndbg/lib/zig.py
+++ b/pwndbg/lib/zig.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import os
 import os.path
 import pathlib
 import subprocess
@@ -71,6 +72,25 @@ _asm_header: Dict[str, str] = {
 }
 
 
+def get_zig_executable() -> str:
+    """
+    Get the path to the zig executable.
+    Precedence: ziglang module, zig in PATH.
+    """
+    try:
+        import ziglang  # type: ignore[import-untyped]
+        return os.path.join(os.path.dirname(ziglang.__file__), "zig")
+    except ImportError:
+        pass
+
+    import shutil
+    zig_in_path = shutil.which("zig")
+    if zig_in_path:
+        return zig_in_path
+
+    raise ValueError("Python module ziglang not available and zig not found in PATH")
+
+
 def _get_zig_target(arch: ArchDefinition) -> str | None:
     if arch.platform == Platform.LINUX:
         # "gnu", "gnuabin32", "gnuabi64", "gnueabi", "gnueabihf",
@@ -90,10 +110,7 @@ def _get_zig_target(arch: ArchDefinition) -> str | None:
 
 
 def flags(arch: ArchDefinition) -> List[str]:
-    try:
-        import ziglang  # type: ignore[import-untyped]
-    except ImportError:
-        raise ValueError("Can't import ziglang")
+    zig_executable = get_zig_executable()
 
     zig_target = _get_zig_target(arch)
     if zig_target is None:
@@ -102,7 +119,7 @@ def flags(arch: ArchDefinition) -> List[str]:
         )
 
     return [
-        os.path.join(os.path.dirname(ziglang.__file__), "zig"),
+        zig_executable,
         "cc",
         "-target",
         zig_target,
@@ -120,10 +137,7 @@ def asm(arch: ArchDefinition, data: str, includes: List[pathlib.Path] | None = N
 
 
 def _asm(arch_mapping: str, data: str, includes: List[pathlib.Path] | None = None) -> bytes:
-    try:
-        import ziglang
-    except ImportError:
-        raise ValueError("Can't import ziglang")
+    zig_executable = get_zig_executable()
 
     header = _asm_header.get(arch_mapping, None)
     if header is None:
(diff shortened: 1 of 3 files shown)


pwndbg/lib/zig.py
```diff
@@ -71,6 +72,29 @@
...
+    """
+    Get the path to the zig executable.
+    Precedence: ZIGCC environment variable, ziglang module, zig in PATH.
+    """
+    zig_env = os.environ.get("ZIGCC")
+    if zig_env:
+        return zig_env
+
+    try:
+        import ziglang  # type: ignore[import-untyped]
```
patryk4815, line 87: the logic should be reversed, it should first use `import ziglang`, then it should search in system as fallback

patryk4815: I mean `os.environ.get("ZIGCC")` we shouldn't use that

pwndbg/lib/zig.py
```diff
@@ -71,6 +72,29 @@
...
+        return os.path.join(os.path.dirname(ziglang.__file__), "zig")
+    except ImportError:
+        pass
+
+    import shutil
+    zig_in_path = shutil.which("zig")
+    if zig_in_path:
+        return zig_in_path
+
+    raise ValueError("Can't find zig: ZIGCC environment variable not set, ziglang module not available, and zig not found in PATH")
```
patryk4815: probably we should add here version checking.. because zig 0.15.1 and 0.15.2 is broken, only zig 0.14.1 works fine

patryk4815: https://github.com/ziglang/zig/issues/25653

pwndbg/lib/zig.py
patryk4815: @disconnect3d do we want support `ZIGCC`? Or this should be `pwndbg.config.` option?

carlsmedstad: Didn't see your last message - removed support for it in ecf391f2d73c27eae53117b6a7d1c56263d3d027.

pwndbg/lib/zig.py
carlsmedstad: Something like this maybe ab7f2af985553316952f76f7b411011a67e7de7e?

patryk4815: @carlsmedstad looks good to me, I will add few more comments, and we can merge it

carlsmedstad: Super, squashing the fixup commit and rewriting the commit message to remove the mention of `ZIGCC`

pwndbg/lib/zig.py
```diff
@@ -71,6 +72,48 @@
...
+
+    if zig_path is None:
+        raise ValueError("Python module ziglang not available and zig not found in PATH")
+
+    try:
+        result = subprocess.run(
+            [zig_path, "version"],
+            capture_output=True,
+            text=True,
+            timeout=1,
```
patryk4815: @carlsmedstad `timeout=1` should be increased for example on MacOS (because of Gatekeeper/XProtect)

```
psondej@m4 ~ $ time /opt/homebrew/Caskroom/pwndbg-gdb/2025.10.20/pwndbg/lib/python3.13/site-packages/ziglang/zig version
0.14.1

real	0m3,320s
user	0m0,022s
sys	0m0,072s
```
On my system it took 3.3s
Maybe 15s should be enough.

pwndbg/lib/zig.py
```diff
@@ -71,6 +72,48 @@
...
+
+    if zig_path is None:
+        import shutil
+        zig_path = shutil.which("zig")
+
+    if zig_path is None:
+        raise ValueError("Python module ziglang not available and zig not found in PATH")
+
+    try:
+        result = subprocess.run(
```
patryk4815, line 97: please move this version check only for system zig, not for pypi ziglang package (it is locked)

pwndbg/lib/zig.py
```diff
@@ -71,6 +72,48 @@
 }
 
 
+ZIG_SUPPORTED_VERSION = "0.14.1"
+
+
+def get_zig_executable() -> str:
```
patryk4815, line 81: add decorator `@pwndbg.lib.cache.cache_until("forever")`

pwndbg/lib/zig.py
```diff
@@ -71,6 +72,48 @@
...
+    zig_path = None
+
+    try:
+        import ziglang  # type: ignore[import-untyped]
+        zig_path = os.path.join(os.path.dirname(ziglang.__file__), "zig")
+    except ImportError:
+        pass
+
+    if zig_path is None:
+        import shutil
```
patryk4815: move import outside func

patryk4815 (changes requested):

patryk4815:
@carlsmedstad BTW, commits don’t need to be perfect - we'll squash them when merging the PR. This makes it easier to review new commits. Please don't force-push. Thanks! 😊

carlsmedstad:
There we go 668c10263ea2b820ac558c3a4d17c42f6a765de4 :)

PR #3399, opened 2025-11-10
Apply readline import fix to unit tests

carlsmedstad:
Running `unit-tests.sh` with Python 3.13 in a freshly cloned repository on Arch Linux results in the following test failure:

```sh
FAILED tests/unit_tests/test_readline.py::test_readline_not_imported - AssertionError: assert 'readline' not in {'__future__': <module '__future__' from '/usr/lib/python...
```

I believe this is because pwndbg.exception imports pdb (ipdb is not available), and in Python 3.13, pdb imports rlcompleter, which imports readline.

To address this, apply the readline import fix from pwndbginit/gdbpatches.py to the suite via a conftest module. Related:
https://github.com/pwndbg/pwndbg/pull/2622

Disclaimer: Authored with assistance from Claude Code.

### DIFF
--- a/tests/unit_tests/conftest.py
+++ b/tests/unit_tests/conftest.py
@@ -1,0 +1,14 @@
+from __future__ import annotations
+
+import importlib.abc
+import sys
+
+
+class RemoveReadlineFinder(importlib.abc.MetaPathFinder):
+    def find_spec(self, fullname, path=None, target=None):
+        if fullname == "readline":
+            raise ImportError("readline module disabled in unit tests")
+        return None
+
+
+sys.meta_path.insert(0, RemoveReadlineFinder())


tests/unit_tests/conftest.py
```diff
@@ -0,0 +1,14 @@
+from __future__ import annotations
+
+import importlib.abc
+import sys
+
+
+class RemoveReadlineFinder(importlib.abc.MetaPathFinder):
+    def find_spec(self, fullname, path=None, target=None):
+        if fullname == "readline":
+            raise ImportError("readline module disabled in unit tests")
```
disconnect3d, line 10: Since we raise exception, is this ok and not failing due to an exception thrown?

patryk4815:
The test is correct and must remain that way, the error is somewhere else. 
Tell me how the pwndbg setup is done and how you run the tests?

TL;DR
Somehow this file was never executed:
https://github.com/pwndbg/pwndbg/blob/dev/pwndbginit/gdbpatches.py
It should be executed when starting `pwndbg` entrypoint https://github.com/pwndbg/pwndbg/blob/dev/pyproject.toml#L143

k4lizen:
i can also repro. @carlsmedstad does clearly explain the issue I suppose
> I believe this is because pwndbg.exception imports pdb (ipdb is not available), and in Python 3.13, pdb imports rlcompleter, which imports readline.

but yeah disabling the check in the tests is not the way to go. I guess we have to pin the rlcompleter version? 

although it seems a bit suspect. https://docs.python.org/3/library/rlcompleter.html#module-rlcompleter
says its for gnureadline specifically

patryk4815:
the correct fix is to remove `tests/unit_tests/test_readline.py`, `unit_tests` are not running inside `gdb` tests 

image

Somehow we have 2 tests, one for gdb, another without gdb.
This test without gdb can be removed.

patryk4815:
@carlsmedstad This will be fixed in PR https://github.com/pwndbg/pwndbg/pull/3400

carlsmedstad:
@patryk4815 Thanks!

patryk4815:
@carlsmedstad 
btw, If you want test gdb, only stable test are `cross-arch-user` for other distros:
`./tests.sh -d gdb -g cross-arch-user`

They should require only `zig` + `qemu-user` binaries

carlsmedstad:
That's very helpful, thanks! Was struggling with the other tests :)

patryk4815:
Unfortunately, we have many tests that only work on ubuntu+x86_64. 
Only `cross-arch-user` works stably on other distros :(

PR #3401, opened 2025-11-11
Reimplement decompiler integration

k4lizen:
Okay this is a chonker. 

See the docs/tutorials/decompiler-integration.md for some overview. And docs/assets/caps/decomp_integration_ex.png for an image example

#### pros
+ we don't have any decompiler-side code in Pwndbg anymore, which means we don't have to maintain that code, which is nice
+ previously, there were multiple versions of the debugger-side code depending on which decompiler we were talking to. now, the same piece of debugger code talks to all decompilers
+ we support ghidra now, and i mean like actually
+ we also support angr-management, though the user-base probably isn't that large
+ symbols actually added to the debugger (can `p my_decompiler_symbol`)
+ function vars added to the debugger (can `p $my_function_local`)
+ function-locals are in annotations and the stack!
+ proper syntax highlighting
+ in general, all decompilers are ~equally supported. before it was like binja >> ida >>>> ghidra
+ works with LLDB as well
+ the code is more maintainable now i'd say, but ig thats subjective
#### cons
+ some features that existed in binja only are now removed, can be added in the future but the PR is already getting too large
    + comments in decompilation (were present in binja, now not, were and are present in ida)
    + breakpoint syncing (did this work? dont know)
+ function signature syncing. this was a really nice feature for `call` annotations, want to add it for sure in the future but for now lets shelf it.
+ adding a new feature requires some coordination with decomp2dbg to add a new decompiler-plugin-side api.

### Breaking changes

Bunch of commands and config options removed.


pwndbg/dbg/lldb/__init__.py
```diff
@@ -1834,6 +1834,16 @@ def dispatch_execution_controller(
...
         return self.dbg._execute_lldb_command(cmd)
 
+    @override
+    def set_convenience_var(self, name: str, value: str, type: str) -> None:
+        """
+        Set a convenience variable which will be accessible with $name in the
+        debugger.
+        """
+        # https://stackoverflow.com/questions/11192511/does-lldb-have-convenience-variables-var
+        # FIXME: Is there a nicer version through the API?
```
patryk4815: no api for that

patryk4815: only way: 
```py
   global _global_new_variable_id
    _global_new_variable_id += 1
    var = f"$_pwndbg_internal_shared_cache_size{_global_new_variable_id}"

    base = pwndbg.dbg.selected_inferior().evaluate_expression(
        f"size_t {var} = 0; (const void*)_dyld_get_shared_cache_range(&{var})"
    )
 ```
 similar you can use in GDB.

Comment (k4lizen) [reply pwndbg/dbg/lldb/__init__.py]: i'm not trying to create a debugger variable with an unknowable name. i want to specifically have it so that if you are debugging a function which you decompiled has a variable with the name `my_cool_variable`, you can print it in gdb with `p $my_cool_variable`.

Comment (patryk4815) [reply pwndbg/dbg/lldb/__init__.py]: > you can print it in gdb with p $my_cool_variable.

yes, this is exactly doing that

Comment (k4lizen) [reply pwndbg/dbg/lldb/__init__.py]: oh okay wow. but, do you think that
```
self.dbg._execute_lldb_command(f"expr {type} ${name} = {value}")
```
won't work? or do you dislike that i'm using `_execute_lldb_command`?

**Comment by k4lizen** [comment]
oke performance is an issue, will need to look for some optimizations

File: pwndbg/dbg/__init__.py [hunk 2]
```diff
@@ -651,11 +659,12 @@ def runcmd(self, cmd):
         """
         raise NotImplementedError()
 
-    def set_convenience_var(self, name: str, value: str, type: str) -> None:
+    def set_convenience_var(self, name: str, value: str, type: Optional[str]) -> None:
         """
         Set a convenience variable which will be accessible with $name in the
         debugger.
         """
+        # FIXME: is this actually tied to the lifetime of the process?
```
patryk4815: for lldb "convenience_var" are lifetime of lldb session, vars can't be removed.

and when you start process (`r`) few times in same lldb session, you will get corrupted "convenience_var"'s

see: https://github.com/llvm/llvm-project/issues/84806#issuecomment-1995055683

k4lizen: the issue you linked seems to be related to variables getting corrupted due to the print command, rather than expr which is what i'm using for lldb.

> are lifetime of lldb session, vars can't be removed.

I think this is fine, having the `p $func_var` feature is very useful, even if old variables are still alive. It is a bit scuffed but I will document it so that users know what to expect (and decomp2dbg has the same issue, because they also persist in gdb).

> you will get corrupted "convenience_var"'s

hmm, i mean, if they retain old values i think this is okay, but if they get corrupted like in the issue than its a problem

k4lizen: > are lifetime of lldb session, vars can't be removed.

but I think that means that I should move this function out of `pwndbg.dbg_mod.Process` and into ` Debugger`

pwndbg/dbg/lldb/__init__.py
```diff
@@ -1834,6 +1847,19 @@ def dispatch_execution_controller(
...
+    @override
+    def set_convenience_var(self, name: str, value: str, type: Optional[str]) -> None:
+        """
+        Set a convenience variable which will be accessible with $name in the
+        debugger.
+        """
+        # https://stackoverflow.com/questions/11192511/does-lldb-have-convenience-variables-var
+        # FIXME: Is there a nicer version through the API?
+        if type is not None:
+            self.dbg._execute_lldb_command(f"expr {type} ${name} = (({type})({value}))")
```
k4lizen: m but this is a bug I guess, it will fail if the variable already exists

pwndbg/dbg/__init__.py
patryk4815: > the issue you linked seems to be related to variables getting corrupted due to the print command

issue and linked comment are two different things ;)

image

k4lizen: ohhh damn I see, yeah I can reproduce that as well. I will add a comment about it in the function

k4lizen:
oke still need the function type signature stuff, but that should be the end of any `IntegrationManager` logic. after that will need to figure out the UX of the integration managment.

k4lizen:
@patryk4815 seems nix can't build decomp2dbg?

patryk4815:
@k4lizen 
Error is: `ModuleNotFoundError: No module named 'setuptools'`
Try adding `"decomp2dbg"` into array `pkgsNeedSetuptools`

k4lizen:
image

idk what to do about this :woozy_face:

k4lizen:
There are two blockers currently:
1. The nix build
2. There are some bugs on the decomp2dbg side that should be ironed out first

k4lizen:
Some examples:
```

pwndbg> di --help
usage: decompiler-integration [-h] {connect,c,disconnect,d,sync,s,jump,j,install,decomp,list,l} ...

Control Pwndbg decompiler integration.

positional arguments:
  {connect,c,disconnect,d,sync,s,jump,j,install,decomp,list,l}
    connect (c)         Connect to the decompiler
    disconnect (d)      Disconnect from the decompiler
    sync (s)            Sync data from the decompiler
    jump (j)            Make the decompiler's cursor jump to the PC
    install             Install the decompiler plugins
    decomp              Just use the `decomp` command
    list (l)            List the variables for the current stack frame

options:

Alias: di
pwndbg> di install --help
usage: di install [-h] which ...

Install/update the decompiler plugins. You need a decompiler plugin installed to allow the decompiler to communicate back to Pwndbg. The decompiler plugins are from decomp2dbg (<3). If you
already have decomp2dbg installed, this command will overwrite that installation in order to pin the proper version that Pwndbg needs. You will still be able to use decomp2dbg outside of
Pwndbg. You should take care not to invoke `source /path/to/decomp2dbg/d2d.py` in your ~/.gdbinit because we implement the debugger-side logic independently, and it might conflict.

positional arguments:
  which
    ida       Install the IDA decompiler plugin
    binja     Install the Binary Ninja decompiler plugin
    ghidra    Install the Ghidra decompiler plugin
    angr      Install the angr-management decompiler plugin

options:
pwndbg> di install ida
Installing the IDA decompiler plugin.

decomp2dbg version: 3.12.0 (required: 3.12.0)

Source:       /home/user/code/pwndbg/.venv/lib/python3.12/site-packages/decomp2dbg/decompilers/d2d_ida
Destination:  /home/user/.idapro/plugins

Making sure destination folder exists..

Deleting old files (if they exist):
	/home/user/.idapro/plugins/d2d_ida
	/home/user/.idapro/plugins/d2d_ida.py

Creating symlinks:
	/home/user/.idapro/plugins/d2d_ida -> /home/user/code/pwndbg/.venv/lib/python3.12/site-packages/decomp2dbg/decompilers/d2d_ida/d2d_ida
	/home/user/.idapro/plugins/d2d_ida.py -> /home/user/code/pwndbg/.venv/lib/python3.12/site-packages/decomp2dbg/decompilers/d2d_ida/d2d_ida.py

The fact that symlinks are used means the decompiler plugin will be automatically
updated when the decomp2dbg python package is updated. But if the decomp2dbg
installation path gets changed, don't forget to reinstall!
(so take care if you change the folder of your Pwndbg installation)

If you want to change the plugin destination, run `set decompiler-ida-plugin-path the/new/path`.
(and put this line into your ~/.gdbinit so you don't have issues in the future)

Installed successfully. If your decompiler is already open, restart it.
pwndbg> 
```

k4lizen:
image

k4lizen:
(rebased to dev to sync with https://github.com/pwndbg/pwndbg/pull/3429)

disconnect3d:
Some conflicts needs to be resolved :<

k4lizen:
self-reminder: need to see how to handle kernel:
```
The decompiled program chall.ko doesn't seem to be loaded. We will keep an eye out for it.
```
edit: mostly handled

k4lizen:
fixed everything up in decomp2dbg that was a blocker, now only left to figure out how to satisfy nix.

btw, to be clear, I only need
1. the actual python source of decomp2dbg to be somewhere on the filesystem
2. to know where it is

I don't need any of its dependancies, nor to actually run its setup or anything. So probably the ideal solution is to somehow no-op the setup of that package, so we don't include any extra stuff we don't need in the pwndbg portable build.

maybe some magic with `decomp2dbgVersion =
    builtins.head (builtins.match "decomp2dbg[ =<>~]*\"?([0-9]+\\.[0-9]+\\.[0-9]+)\"?" (builtins.readFile pyprojectFile));` and `  decomp2dbgSrc = pkgs.fetchFromGitHub {`? dunno

disconnect3d:
cc: @patryk4815 can u help with this, how do we fetch the decomp2gdb so it is available as source code?

Otherwise, we can always just vendor it, heh.

mahaloz:
@disconnect3d, d2d author here, if the dependencies are too insane for nix I can attempt to reduce them. I would ofc prefer to keep d2d as a lib install and not vendored code if possible :).

mahaloz:
Also @k4lizen, I added a small ack in the integration code: https://github.com/k4lizen/pwndbg/pull/38

patryk4815:
@k4lizen apply this patch: https://github.com/pwndbg/pwndbg/commit/e2ccd267ab9f6cf40dac7c4ad6f2f62fba8a32cf

@mahaloz I had issue to install decomp2dbg on Apple Silicon:
image
Fixed by modifying `setup.py` to use: `sys.argv.append("macosx_11_0_arm64")`

patryk4815:
@mahaloz probably decomp2dbg should use `py3-none-any.whl`

k4lizen:
@mahaloz 

> d2d author here, if the dependencies are too insane for nix I can attempt to reduce them. I would ofc prefer to keep d2d as a lib install and not vendored code if possible :).

i am also against vendoring this, but of course reducing the deps would be awesome since it will also reduce the size of the pwndbg install :>

k4lizen:
@patryk4815 thanks! :heart_hands:

k4lizen:
@disconnect3d ready to review / merge :>

pwndbg/aglib/disasm/arch.py
```diff
@@ -736,7 +739,9 @@ def _enhance_next(
 
         if instruction.has_jump_target and instruction.target >= 0:
             # Only bother doing the symbol lookup if this is a jump
-            instruction.target_string = MemoryColor.get_address_or_symbol(instruction.target)
+            instruction.target_string = MemoryColor.get_address_or_symbol(
+                instruction.target, pwndbg.integration.manager.get_stack_var_dict_all()
```
disconnect3d, line 743: Is it ok that we do call this multiple times in different functions?

k4lizen: i don't understand the question, what would be an alternative?

also, in `dev` get_address_or_symbol does in fact reach out to integration

Issue #3406, opened 2025-11-15
Binja integration, host/port settings improperly read from environment


korabal:
### Description

To change the HOST/PORT value of the XML-RPC server in Binja, you can set environment variables. 
On macOS, the variables are not read.

### Steps to reproduce

On macOS 
in shell 

```sh 
export PWNDBG_BINJA_SERVER_HOST=192.168.139.3
{path_to_binaryninja}/binaryninja
```
In Binary ninja, load a binary, start the pwndbg integration. You get the log 
```
[pwndbg-integration] XML-RPC server listening on http://127.0.0.1:43717
```

### My setup

BinaryNinja version : 5.2.8614 Personal (66bd5453)
macOS 26.0.1
shell: zsh 5.9

korabal:
As a solution, Binary Ninja offers an API to register settings. I think I will add that to the `binja_script.py`

korabal:
Ok so I did try more tests. 
Now I have no more issues getting a Binja plugin to read environment variables. 
So I must have had an issue with my initial setup. Sorry for that. 

Since there were no bugs in the first place, I close this issue.

PR #3407, opened 2025-11-15
Binja integration, add host and port setting for the XML-RPC server in Binary Ninja

korabal:
Answer to issue #3406 . 
Add the settings for host and port with some helper description.

If the settings are left as default, will also check the environment variable values to ensure backward-compatibility. 

As far as I can read, there are no tests for this script, and I could find no way to properly test it, aside from running it by hand.

### DIFF
--- a/binja_script.py
+++ b/binja_script.py
@@ -11,8 +11,10 @@ from typing import List
 from typing import Tuple
 from xmlrpc.server import SimpleXMLRPCRequestHandler
 from xmlrpc.server import SimpleXMLRPCServer
+import json
 
 import binaryninja
+from binaryninja.settings import Settings
 from binaryninja.enums import RegisterValueType
 from binaryninja.enums import VariableSourceType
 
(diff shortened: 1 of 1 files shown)


k4lizen:
> As far as I can read, there are no tests for this script, and I could find no way to properly test it, aside from running it by hand.

yes there are no tests for this.

k4lizen:
this code looks good and I would be for merging it. though I am curious what the actual reason for
>  On macOS, the variables are not read.

is, and would like to see some warnings about this also added to the docs.

**but**, I am currently in the process of refactoring the integration sauce to use decomp2dbg which will for sure end up deleting this code if/when it goes through, so I would likely hold off on merging this and then eventually close it.

so from your POV the best bet is probably to just use this branch for your personal use until the integration refactor rolls around and then see if that fixes your issue.

korabal:
Cf the issue, the problem must have come from my side. The issue is closed.
 I see your point for decomp2dbg, so I close the MR.

PR #3416, opened 2025-11-19
Check that PRs dont introduce more mypy --strict errors

k4lizen:
Depends on https://github.com/pwndbg/pwndbg/pull/3415


k4lizen:
This is failing lint because https://github.com/pwndbg/pwndbg/pull/3415 isn't merged yet, it will pass.

.github/workflows/docker.yml
```diff
@@ -17,29 +17,29 @@ jobs:
...
+      - name: Run GDB Tests on ${{ matrix.images }}
+        run: |
+          docker compose run ${{ matrix.images }} ./tests.sh -d gdb -g gdb
 
-    - name: Run DBG Tests on LLDB on ${{ matrix.images }}
-      run: |
-        docker compose run ${{ matrix.images }} ./tests.sh -d lldb -g dbg
+      - name: Run DBG Tests on GDB on ${{ matrix.images }}
+        run: |
+          docker compose run ${{ matrix.images }} ./tests.sh -d gdb -g dbg
```
k4lizen, line 38: i don't remember adding this, but prolly makes sense if we want to move away from gdb-only tests

PR #3419, opened 2025-11-20
Enable cross-architecture instruction patching

OBarronCS:
Enable the use of the `patch` command for cross-architecture targets.

Previously, the `patch` command delegated to pwnlibs `asm` function, which searches for an appropriate `gcc` program locally that targets the given architecture, possibly a cross-architecture gcc binary. Same with `patch-list`, which used gcc under-the-hood for disassembly. If the appropriate gcc cross-compiler cannot be found, it crashes. 

Since we ship Zig with pwndbg, we can use that instead to assembly. For disassembly in `patch-list`, it uses Capstone now instead of delegating to pwnlib.

For reference, `patch-list` prints out all the patches you've made:
image

Adds a test for cross-architecture patching.

### DIFF
--- a/pwndbg/commands/patch.py
+++ b/pwndbg/commands/patch.py
@@ -90,9 +90,6 @@ def patch_list() -> None:
             [f"{x.mnemonic} {x.op_str}".strip() for x in cs.disasm(new, offset=addr)]
         )
 
-        # old_insns = disasm(old, byte=False, offset=False)
-        # new_insns = disasm(new, byte=False, offset=False)
-
         colored_addr = pwndbg.color.memory.get(addr)
 
         old_insns, new_insns = map(
--- a/pwndbg/commands/patch.py
+++ b/pwndbg/commands/patch.py
@@ -4,20 +4,19 @@ import argparse
 from typing import Dict
 from typing import Tuple
 
-from pwnlib.asm import asm
-from pwnlib.asm import disasm
-
+import pwndbg.aglib.asm
 import pwndbg.aglib.memory
 import pwndbg.color.context
 import pwndbg.color.memory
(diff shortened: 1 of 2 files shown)


OBarronCS:
I marked the test as "@pytest.mark.xfail" - the version of qemu-user in Ubuntu24.04 doesn't support GDB writing to memory.

EDIT: *doesn't support writing to memory that doesn't have write attribute on the page

k4lizen:
> the version of qemu-user in Ubuntu24.04 doesn't support GDB writing to memory.

but like no way this is true right? writing to memory is like the most basic feature. if you couldn't write to memory you wouldn't be able to set breakpoints

OBarronCS:
The GDB `write_memory` method and doing `set *(int*)pc=123` fails on the $pc address. I tested it again, and looks like it fails when writing to pages without the write attribute (which gdb normally should be able to overwrite anyways) on this version of qemu. 
image

k4lizen:
crazy.

does this only happen on aarch64?

OBarronCS:
I got the same behavior on aarch64, arm, risc-v, and mips. I believe it applies across all the architectures on `8.2.2`. On version `9.2.1` and up, you can write as expected.

disconnect3d:
Can it be QEMU regression?

Also maybe its some W^X or R^X option or sth?

OBarronCS:
I tried to find a note in the Qemu changelogs since 8.2.2 relating to adding this ability (since it works on newer versions, at least 9.2.1), but on initial search couldn't find a commit related to it

Ninja3047:
https://github.com/qemu/qemu/commit/87ab270429

PR #3421, opened 2025-11-20
Setrlimit command

### LINKED ISSUE #967: Add setrlimit or ulimit command
TL;DR: It would be nice if we could do `ulimit` or `setrlimit` within Pwndbg to set a limit for the current process.

I think it is fine if we do it by calling the underlying glibc `setrlimit` function. 

An example is shown here: https://qa-stack.pl/superuser/404239/setting-ulimit-on-a-running-process  - we just need a more convenient interface then the one below:
```
$ gdb -p $PID
...
(gdb) set $rlim = &{0ll, 0ll}
(gdb) print getrlimit(9, $rlim)
$1 = 0
(gdb) print *$rlim
$2 = {-1, -1}
(gdb) set *$rlim[0] = 1024*1024
(gdb) print setrlimit(9, $rlim)
$3 = 0
```

References:
* https://man7.org/linux/man-pages/man2/getrlimit.2.html

marywinter3:
Closes #967


tests/library/gdb/tests/test_command_setrlimit.py
```diff
@@ -0,0 +1,48 @@
...
+import pwndbg.commands.setrlimit as sr
+
+
+def test_setrlimit_unknown_resource(capfd):
+    with patch.object(sr, "_invoke_setrlimit") as mock_invoke:
+        sr.setrlimit("unknown", "1")
+        mock_invoke.assert_not_called()
+
+    out, _ = capfd.readouterr()
+    assert "Unknown resource 'unknown'" in out
```
disconnect3d: Can't we just do `gdb.execute("setrlimit unknown 1", to_string=True) == "Unknown resource 'unknown'"` or something like that instead of doing the mocks etc?

disconnect3d: I mean, instead of using [the `capfd` fixture across the tests here](https://docs.pytest.org/en/stable/reference/reference.html#std-fixture-capfd)

I think we never use that anywhere heh

marywinter3:
@disconnect3d Are tests ok now?

pwndbg/commands/setrlimit.py
```diff
@@ -0,0 +1,92 @@
...
+    type=str,
+    nargs="?",
+    default=None,
+    help="Hard limit value (integer) or 'infinite' (defaults to soft)",
+)
+
+
+def _invoke_setrlimit(num: int, soft_val: int, hard_val: int) -> str:
+    return gdb.execute(
+        f"call (int) setrlimit({num}, (struct rlimit[]){{ {{ {soft_val}, {hard_val} }} }})",
```
disconnect3d: Ideally, we should have a fallback which would use shellcode execution for this iff the function is not available etc.

pwndbg/commands/setrlimit.py
```diff
@@ -0,0 +1,92 @@
...
+        to_string=True,
+    )
+
+
+@pwndbg.commands.Command(parser, category=CommandCategory.MISC)
```
k4lizen, line 92: missing onlywhenrunning decorator.

k4lizen (changes requested):
there are a bunch of cases where this command won't work: it requires symbols, debugging information, userspace, a running process, a standard library, a posix standard library, ....

this stuff should be checked before the `call` is executed and an error message returned to the user.

also, this is gdb-only code but there is nothing intrinsic to gdb that is relevant to this command, so it should be made debugger agnostic using the `dbg/` abstraction layer

jxuanli:
hi, see #3509, we can still use the test cases in this PR but should implement a generic syscall api instead

disconnect3d:
Since this is a bit stale and I want to keep the list of opened PRs short/small if possible, I'm going to close this for now. Please reopen when ready.

marywinter3:
I'm ready to continue working with this issue.

tests/library/gdb/tests/test_command_setrlimit.py
```diff
@@ -8,30 +8,38 @@
 
 
 def test_setrlimit_unknown_resource():
+    gdb.execute("file /bin/true", to_string=True)
```
disconnect3d: Please look at other tests: we have special reference binaries in our test suite to be used. I bet we can't know for sure if /bin/true is present :D

marywinter3:
One test is missing. Should I change somethig? @disconnect3d

pwndbg/commands/setrlimit.py
```diff
@@ -0,0 +1,129 @@
...
+    except ValueError:
+        print(message.error(f"Invalid limit '{val}'"))
+        raise
+
+
+def ensure_can_call_setrlimit() -> pwndbg.dbg_mod.Process:
+    proc = pwndbg.dbg.selected_inferior()
+
+    if proc is None or not proc.alive():
+        raise pwndbg.dbg_mod.Error("No running inferior - start or attach to a process first.")
```
disconnect3d: This is redundant, it is checked by `@pwndbg.commands.OnlyWhenRunning` already.

pwndbg/commands/setrlimit.py
```diff
@@ -0,0 +1,129 @@
...
+    "nofile": 7,
+    "as": 9,
+}
+
+parser = argparse.ArgumentParser(
+    description="Set a POSIX resource limit in the debugged process.",
+)
+parser.add_argument(
+    "resource",
+    type=str,
```
disconnect3d: Shall we use `param_class=pwndbg.lib.config.PARAM_ENUM` here?

Please see other places where its used

disconnect3d (changes requested):
See comments

pwndbg/commands/setrlimit.py
marywinter3: I tried using param_class=pwndbg.lib.config.PARAM_ENUM here, but I’m getting the following error:
TypeError: _StoreAction.__init__() got an unexpected keyword argument 'param_class' 

Because of that, I used choices instead. Is this ok here?

tests/library/gdb/tests/test_command_setrlimit.py
```diff
@@ -0,0 +1,49 @@
+from __future__ import annotations
```
disconnect3d, line 1: Why is it implemented in gdb and not dbg? Does all of this not work with LLDB? Why?

tests/library/gdb/tests/test_command_setrlimit.py
```diff
@@ -0,0 +1,49 @@
...
+    gdb.execute("start", to_string=True)
+
+    called: dict[str, tuple[int, int, int]] = {}
+
+    def fake_invoke(num: int, soft_val: int, hard_val: int) -> int:
+        called["args"] = (num, soft_val, hard_val)
+        return 0
+
+    with patch.object(sr, "invoke_setrlimit", side_effect=fake_invoke):
+        out = gdb.execute("setrlimit cpu 10", to_string=True)
```
disconnect3d, line 46: can't we do the gdb.execute right away without patch?

disconnect3d (commented):
I've re-reviewed it. Please see my questions related to tests.

I'd also prefer if that would be implemented as a syscall invocation instead of a function invocation...

This could really be implemented the same way as the `mprotect` command we have - using the pwntools' library shellcraft module. They do have a way to emit a shellcode for the setrlimit syscall:
```
$ i
Python 3.12.3 (main, Mar  3 2026, 12:15:18) [GCC 13.3.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.20.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from pwn import *

In [2]: shellcraft.amd64.setrlimit
Out[2]: <function pwnlib.shellcraft.amd64.linux.syscalls.setrlimit(resource=0, rlimits=0)>

In [3]: shellcraft.amd64.setrlimit?
Signature: shellcraft.amd64.setrlimit(resource=0, rlimits=0)
Docstring:
setrlimit(resource, rlimits) -> str

Invokes the syscall setrlimit.

See 'man 2 setrlimit' for more information.

Arguments:
    resource(rlimit_resource_t): resource
    rlimits(rlimit*): rlimits
Returns:
    int
File:      /usr/lib/python3/dist-packages/pwnlib/shellcraft/internal.py
Type:      function

```

marywinter3:
@disconnect3d I rewrote the implementation to use syscall invocation.

disconnect3d:
Please fix lint issue - use `from pwndbg.color import message`

image

PR #3422, opened 2025-11-20
Update Fix vmmap indicator overwriting address prefix 

jackmisbach:
vmware_kFR2BPclQY
This is the output of his change shown above, so it does actually work. 
vmware_RhP29LscO3
I tested the original proposed change from PR3413:

text = prefix + " " + text

and confirmed it fixes the issue – the ► no longer overwrites the 0 in 0x (see first screenshot).

I then made a small follow-up change for consistency with the rest of memory.py, switching to an f-string:

text = f"{prefix} {text}"

Functionally, this is identical, but matches the existing style in this file (which already uses f-strings for address/symbol formatting). The second screenshot shows the output with this updated line.


jackmisbach:
@k4lizen Let me know if this looks good

k4lizen:
@jackmisbach thanks for the screenshot

yeah it does work in the sense that the prefix doesn't overlap the text, but the problem is that the output isn't aligned anymore. sorry for not being clear about this in the original message

jackmisbach:
Okay, yeah, that makes total sense. I was going to do that, but just wasn't sure honestly lol, thanks for the clarification!
Here is a screenshot with the updated commit. 
vmware_VNddUuLX0V
Previously, when vmmap was called with a filter, only the pages that matched the filter received a backtrace_prefix (e.g. the ► marker). That meant those lines had an extra prefix column while the other lines did not, causing the address column to shift and become misaligned.

So with the change it
Computes the configured prefix string once (prefix_str = str(pwndbg.config.backtrace_prefix)),

And, when vmmap is filtered, it assigns a “blank” prefix of the same width (" " * len(prefix_str)) to non-matching pages.

so all rows now have a prefix column of equal width: spaces for non-matches, the arrow for matches. The columns stay aligned, and the addresses remain intact for copy-paste.

pwndbg/commands/vmmap.py
k4lizen: the header is now misaligned

pwndbg/color/memory.py
k4lizen: since you changed the behaviour of this function you need to
1. check all the callsites that use `prefix=` to make sure none of them will be messed up by this change
2. change this docstring to reflect the new behaviour

k4lizen (changes requested):
thats better, but it still needs some fixes before being merged

jackmisbach:
vmware_v1tm0t3Bxi
Okay, yeah, I'm not sure how I didn't notice that, but it is aligned now, and I updated the docstrings to match the new behavior. Thanks for pointing that out.
 Also checked all of the prefix= callsites, and everything looks good.

Please let me know if you see any other issues!

k4lizen (approved):
lgtm, thanks

pwndbg/color/memory.py
```diff
@@ -109,8 +109,8 @@ def get(
         text = pwndbg.lib.pretty_print.int_to_string(address)
 
     if prefix:
-        # Replace first N characters with the provided prefix
-        text = prefix + text[len(prefix) :]
+        # Prepend the prefix and a space before the existing text
+        text = f"{prefix} {text}"
```
disconnect3d, line 113: Fwiw the check should probably be `if prefix is not None` since in theory, you may have an empty string as a prefix and you may still want the space to be there? :P 

Or: maybe the default prefix should not be `None` but an empty string?

Idk, it probably doesn't matter that much.

pwndbg/commands/vmmap.py
```diff
@@ -271,8 +271,13 @@ def vmmap(
         print_vmmap_gaps(tuple(total_pages))
         return
 
+    # Determine prefix width for alignment when showing filtered results
+    prefix_str = str(pwndbg.config.backtrace_prefix)
+    empty_prefix = " " * len(prefix_str) if filtered_pages else None
```
disconnect3d, line 276: Can we just do `len(pwndbg.config.backtrace_prefix)` instead?

disconnect3d (approved):

PR #3423, opened 2025-11-21
Fix aarch64_be disassembly

OBarronCS:
In `aarch64_be`, the endianness refers to byte order of data. However, the instructions are always little-endian.

https://developer.arm.com/documentation/102376/0200/Alignment-and-endianness/Endianness

image

This fixes the constant that we pass to Capstone, to make disassembly always little-endian for aarch64. Without this, byte order was flipped while disassembling, resulting in incorrect disassembly.

Adds a test to check for this as well.


OBarronCS:
Example of a program compiled with a bunch of nops (when compiling with aarch64 or aarch64_be, the instruction byteorder is identical), and running with qemu-aarch64_be

Before:
image


After fix:
image

PR #3424, opened 2025-11-21
drop ppc64 big-endian portable release

patryk4815:
For reference, see: [golang/go#34850](https://github.com/golang/go/issues/34850)

We will stop distributing our **pwndbg-portable** release for **ppc64 (big-endian)**.  
Maintaining ppc64 is increasingly difficult as many dependencies have dropped support, and there have been no actively supported distributions for ppc64 since around 2020.

We will continue to support the **ppc64le** portable release.

Currently, I propose supporting portable releases according to the following tiers:

### Tier 1 (most popular)
- x86_64
- aarch64

### Tier 2 (less popular)
- loongarch64
- riscv64

### Tier 3 (older or less common)
- ppc64le
- s390x
- armv7l
- i686 (x86 32-bit)

> In a few years, we may need to drop support for **i686 (x86 32-bit)** as most popular distributions have already dropped it.

### DIFF
--- a/.github/workflows/releases.yml
+++ b/.github/workflows/releases.yml
@@ -12,7 +12,6 @@ jobs:
           pwndbg-gdb-cross-arm32-tarball,
           pwndbg-gdb-cross-riscv64-tarball,
           pwndbg-gdb-cross-s390x-tarball,
-          pwndbg-gdb-cross-ppc64-tarball,
           pwndbg-gdb-cross-ppc64le-tarball,
           pwndbg-gdb-cross-x86_32-tarball,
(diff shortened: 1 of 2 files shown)


k4lizen:
i think we should be careful about dropping arches if possible, these obscures arches almost noone uses are probably a breeding ground for bugs (which you want to debug with pwndbg)

so i think at the very least we should differentiate between arches that we never supported, and stuff like ppc64be where we don't officially support it but like, you can try and it may work :)

patryk4815:
@k4lizen we drop only our portable release, still we should support ppc64be with your system gdb

Issue #3418, opened 2025-11-20
subcommand int params dont get the debugger parse


k4lizen:
### Description

Usually in pwndbg, if I define a command parameter like
```python
parser_jump.add_argument(
    "jump_addr",
    metavar="addr",
    type=int,
    nargs="?",
    default=None,
    help="Address to jump to. (default: pc)",
)
```
The `int` will be rewritten to take in debugger values, hex values and whatnot. But for subcommands this doesn't happen.

jackmisbach:
I opened a PR for this. If you'd like to review it and let me know what you think. Thanks!

jackmisbach:
This can be closed whenever, btw, fixed with PR #3432

PR #3432, opened 2025-11-25
Fix debugger-aware parsing for int arguments in subcommands

jackmisbach:
Issue #3418 Fix subcommand int parameters to accept hex values and debugger expressions.

Subcommand arguments with `type=int` only accepted decimal values. Regular command arguments worked fine, but subcommands failed:
vmware_YxH2XOexeL

Made initialize_parser() recursive to process subparser actions. When encountering a _SubParsersAction, iterate through its subparsers and apply the same type=int → fix_int_reraise_arg replacement.

vmware_VsiF04OXaz

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -228,15 +228,22 @@ class CommandObj:
         # We want to run all integer and otherwise-unspecified arguments
         # through fix() so that GDB parses it.
         # FIXME: this is weird
-        for action in self.parser._actions:
-            if isinstance(action, argparse._SubParsersAction):
-                action.type = str
-            if action.dest == "help":
-                continue
-            if action.type is int:
-                action.type = fix_int_reraise_arg
-            elif action.type is None:
-                action.type = fix_reraise_arg
+        def process_actions(actions):
+            """Recursively process actions, including subparsers"""
+            for action in actions:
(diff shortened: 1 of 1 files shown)


pwndbg/commands/__init__.py
```diff
@@ -228,15 +228,22 @@ def initialize_parser(self):
...
+                    for subparser in action.choices.values():
+                        process_actions(subparser._actions)
+                if action.dest == "help":
+                    continue
+                if action.type is int:
+                    action.type = fix_int_reraise_arg
+                elif action.type is None:
+                    action.type = fix_reraise_arg
+
+        process_actions(self.parser._actions)
```
disconnect3d, line 246: Does this order matter? As in: shall this be done last, or, maybe, first?

k4lizen (approved):
It's a nice solution for a generally hacky piece of code. If it works it works, lgtm.

Thanks!

pwndbg/commands/__init__.py
k4lizen: my 2c is that it doesn't matter, but good question

jackmisbach: it shouldn't change anything functionally, but I could add a continue after handling _SubParsersAction if you'd like

PR #3433, opened 2025-11-25
Fix unhashable cache key handling (#3428)

### LINKED ISSUE #3428: caching fails silently for unhashable types
### Description

Example:
Image

Image

If we remove the `try`:

Image

We should `assert` on this, rather than silently not caching. God knows how many places in the code aren't getting cached because of this, even though we think they are.

unknownIoT:
Fixes #3428

### Summary
This PR makes the cache fail loudly when unhashable keys are used, instead of
silently skipping caching.

### Changes
- Update `DebugCacheDict.__getitem__` to raise `AssertionError` when a
  `TypeError` occurs due to an unhashable key.
- Keep existing behaviour for `KeyError` (increment `misses` and re-raise).
- Increment `hits` only when a value is successfully retrieved.

### Testing
- ./tests.sh (local)

### DIFF
--- a/pwndbg/lib/cache.py
+++ b/pwndbg/lib/cache.py
@@ -38,22 +38,28 @@ class DebugCacheDict(UserDict):  # type: ignore[type-arg]
         self.func = func
         self.name = f'{func.__module__.split(".")[-1]}.{func.__name__}'
 
-    def __getitem__(self, key: Tuple[Any, ...]) -> Any:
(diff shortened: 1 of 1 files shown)


k4lizen:
1. i want a screenshot of pwndbg actually running after this change
2. don't delete the PR template
3. you didn't touch the code literarly shown in the issue description

PR #3437, opened 2025-11-26
Register read fix - correctly mask register using the real register size

OBarronCS:
This fixes a bug when reading the values of subregisters (`eax` within `rax` in x86, or `w0` from `x0` on AArch64), where the value of the subregister was incorrectly signed-extended to the size of the full-register it resides in.

When Pwndbg fetches a register value from the underlying debugger, it might receive a negative value. Example  in GDB (the `-1` on the right side):
image
The debugger knows the register width and treats the 0xff twos-complement, converting it to -1 which Pwndbg receives.

Previously, we would mask this value with `pwndbg.aglib.arch.ptrmask`, instead of with a mask that corresponded to the target register size, resulting in unexpected sign-extension if the subregister happened to be a negative value. This is fixed my masking with a correct bitmask. 

Before:
image

After:
image

### DIFF
--- a/pwndbg/aglib/regs.py
+++ b/pwndbg/aglib/regs.py
@@ -129,7 +129,15 @@ class module(ModuleType):
                 if self.cs is None:
                     return None
                 value += self.cs * 16
-            return int(value) & pwndbg.aglib.arch.ptrmask
+
+            # The value that the native debugger returns can be negative.
+            # We convert this to the unsigned bit representation by masking it
+            reg_definition = pwndbg.aglib.regs.current.reg_definitions.get(reg.lower())
+            if reg_definition and reg_definition.mask is not None:
+                mask = reg_definition.mask
+            else:
+                mask = pwndbg.aglib.arch.ptrmask
+            return int(value) & mask
         except (ValueError, pwndbg.dbg_mod.Error):
             return None
 
--- a/pwndbg/aglib/regs.py
+++ b/pwndbg/aglib/regs.py
@@ -216,6 +216,12 @@ class module(ModuleType):
         return reg_sets[pwndbg.aglib.arch.name].all
 
     def fix(self, expression: str) -> str:
+        """
+        This is used in CLI parsing.
+        It takes in a string with a register name, "rax", and prefixes it with
+        a $ ("$rax") so that the underlying debugger can evaluate it to resolve the value
+        """
+        expression = pwndbg.aglib.regs.current.resolve_aliases(expression)
         for regname in self.all:
             expression = re.sub(rf"\$?\b{regname}\b", r"$" + regname, expression)
         return expression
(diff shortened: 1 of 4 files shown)


OBarronCS:
Tests are failing due to the codebase using `pwndbg.aglib.regs.sp` in various places - and `sp` is being detected as the 16-bit subregister of rsp for masking. "sp" is not always aliased to the stack pointer for underlying debuggers (such as in windbg, this is not the case). I'll make another PR to address that issue first.

patryk4815:
> `pwndbg.aglib.regs.sp`

probably we should do something like: `pwndbg.aglib.regs.stack`

OBarronCS:
> > `pwndbg.aglib.regs.sp`
> 
> probably we should do something like: `pwndbg.aglib.regs.stack`

I think that is good solution. The codebase currently has a dozen or so places that currently use `pwndbg.aglib.regs.stack`, which is currently a reference to the string name of the current stack register - i.e. you can use `pwndbg.aglib.regs[pwndbg.aglib.regs.stack]` to read it. Maybe we could consider replacing .stack as a reference to value of the stack pointer, and instead use `pwndbg.aglib.regs.current.stack` to get the name when needed.

As a note, along with `pwndbg.aglib.regs.sp`, `pwndbg.aglib.regs.pc` is another "magic" way to read/write from registers that so far has worked (because underlying debuggers convert `sp` -> real stack pointer, and `pc` to read the program counter.). However. `pwndbg.aglib.regs.pc` will have a problem for the `avr` architecture (which pwndbg currently doesn't support) - the byte-level PC register that GDB exposes is named `pc2` (https://sourceware.org/legacy-ml/gdb-patches/2009-11/msg00211.html). All references to `pwndbg.aglib.regs.pc` should read from `pc2` register for avr, so that lookup needs a special setter/getter on the regs object.

OBarronCS:
An important part about this PR is that because of being more precise about register masking, there is a case in x86 that needs to be considered - doing `pwndbg.aglib.regs.read_reg("sp")` will no longer work to read the stack pointer (this is not used anywhere in the codebase). You have to use `pwndbg.aglib.regs.sp` to read it (or `pwndbg.aglib.regs.read_reg(pwndbg.aglib.regs.current.stack)`).

"sp" is the 16-bit subregister of "rsp". Previously, reading "sp" happened to work on GDB and LLDB, which internally alias "sp" to the full "rsp" register. We shouldn't assume that all underlying debuggers do this conversion. Windbg doesn't, and reading "sp" just returns the 16-bit subregister, and you need to explicitly request `rsp` to get the stack pointer.

Now, due to more precise masking, `pwndbg.aglib.regs.read_reg("sp")` would return the value of the 16-bit `sp` subregister of `rsp`.

However, if the user is requesting to read "sp" using the regs command, they likely intend to read the stack pointer, so this PR adds aliases for "sp" and "pc" to ensure that if passed to the `regs` command, they are resolved to the full registers before passing the request to the underlying debugger. This is also the case for CLI argument fixup, where the `sp` and `pc` names are resolved before passing to the debugger.

Regs command change:
Before:
image

After:

image

I added two tests for this.

disconnect3d:
@OBarronCS fix conflict pls

PR #3440, opened 2025-11-26
Replace pwndbg.aglib.regs[x] with pwndbg.aglib.regs.read_reg(x)

OBarronCS:
Currently, there are 3 ways to read registers:
1. `pwndbg.aglib.regs.x`        -  the `__getattr__` property
2. `pwndbg.aglib.regs["x"]`    - the `__getitem__` property
3. `pwndbg.aglib.regs.read_reg("x")`

This PR replaces all occurrences of  `pwndbg.aglib.regs["x"]` with `pwndbg,aglib.regs.read_reg("x")`.

A future PR will remove the first method, `__getattr__`, but it requires more than just changing object accesses, since we have some "special" properties on the regs module to override logic for fetching a specific register, such as `fs_base`, and some places in the codebase uses `getattr(pwndbg.aglib.regs,reg_name)` in order to invoke the special handlers in case they exist for the given register.

https://github.com/pwndbg/pwndbg/blob/f035e158a019a0dbed26fe2e228517c3ac41a123/pwndbg/aglib/kernel/vmmap.py#L175

https://github.com/pwndbg/pwndbg/blob/f035e158a019a0dbed26fe2e228517c3ac41a123/pwndbg/aglib/regs.py#L223

### DIFF
--- a/pwndbg/aglib/regs.py
+++ b/pwndbg/aglib/regs.py
@@ -134,14 +134,6 @@ class module(ModuleType):
             if not pwndbg.dbg.selected_frame().reg_write(attr, int(val)):
                 raise RuntimeError(f"Attempted to write to a non-existent register '{attr}'")
 
-    @pwndbg.lib.cache.cache_until("stop", "prompt")
-    def __getitem__(self, item: Any) -> int | None:
-        if not isinstance(item, str):
-            print("Unknown register type: %r" % (item))
-            return None
-
-        return self.read_reg(item)
-
     def __contains__(self, reg: str) -> bool:
         return reg_sets[pwndbg.aglib.arch.name].__contains__(reg)
 
@@ -200,7 +192,7 @@ class module(ModuleType):
 
     def items(self) -> Generator[Tuple[str, Any], None, None]:
         for regname in self.all:
-            yield regname, self[regname]
+            yield regname, self.read_reg(regname)
 
     reg_sets = reg_sets
 
@@ -208,7 +200,7 @@ class module(ModuleType):
     def changed(self) -> List[str]:
         delta: List[str] = []
         for reg, value in self.previous.items():
-            if self[reg] != value:
+            if self.read_reg(reg) != value:
                 delta.append(reg)
         return delta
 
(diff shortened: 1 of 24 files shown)


pwndbg/aglib/regs.py
```diff
@@ -134,14 +134,6 @@ def __setattr__(self, attr: str, val: Any) -> None:
...
                 raise RuntimeError(f"Attempted to write to a non-existent register '{attr}'")
 
-    @pwndbg.lib.cache.cache_until("stop", "prompt")
-    def __getitem__(self, item: Any) -> int | None:
-        if not isinstance(item, str):
-            print("Unknown register type: %r" % (item))
-            return None
-
-        return self.read_reg(item)
-
```
OBarronCS, line 144: Removes this method of fetching registers. I'm not sure where the "isinstance(item,str)" check comes from, but I was unable to get it to trigger.

Was this meant to be a check for if the item was "None"? Looking at the callsites that are edited in this PR, it looks like they don't guard against a "None" return value, indicating this check was not expected to fail, so removing it should be safe.

k4lizen: > Looking at the callsites that are edited in this PR, it looks like they don't guard against a "None" return value, indicating this check was not expected to fail,

hihihi. i wonder how many `.selected_inferior()` calls check for None :DD 

and `read_reg()` itself returns `int | None`. but anyway, i think its fine to remove this check.

pwndbg/commands/msr.py
k4lizen, line 141: random observation: this should be `OnlyWithArch` rather than checking in msr_write and msr_read

k4lizen (approved):

pwndbg/aglib/regs.py
OBarronCS: Haha, sounds good :)

disconnect3d (approved):

patryk4815:
Added label `breaking-changes` since we are removing `__getitem__` maybe someone was using that out-of-tree

PR #3441, opened 2025-11-26
Remove pwndbg.aglib.regs.__getattr__, convert uses to pwndbg.aglib.regs.read_reg

OBarronCS:
Continuing the work in #3440, this PR converts uses of `pwndbg.aglib.regs.x` to `pwndbg.aglib.regs.read_reg("x")` to unify the API used for register reads. It also removes the `__setattr__` (`pwndbg.aglib.regs.x = y`) in favor of using `pwndbg.aglib.regs.reg_write(x,y)`.

Two special getter properties are added to the object, `pc` and `sp`. This means you can still do `pwndbg.aglib.regs.pc` and `pwndbg.agilb.regs.sp`, which is very common in the codebase. By specifically handling them, we also gain the advantage of resolving the name of the actual stack pointer (like `rsp`) before forwarding the request to the underlying debugger, instead of depending on the debugger to convert `sp` -> `rsp`.

### DIFF
--- a/docs/contributing/writing-tests.md
+++ b/docs/contributing/writing-tests.md
@@ -52,7 +52,7 @@ Using a [`pytest`](https://docs.pytest.org/en/latest/) fixture at the beginning 
 GDB will attach to a [`binary`](https://github.com/pwndbg/pwndbg/tree/dev/tests/library/gdb/conftest.py)
 or connect to a [`QEMU instance`](https://github.com/pwndbg/pwndbg/tree/dev/tests/library/qemu-user/conftest.py).
 Each test runs some commands and uses Python `assert` statements to verify correctness. We can access Pwndbg
-library code like `pwndbg.aglib.regs.rsp` as well as execute GDB commands with `gdb.execute()`.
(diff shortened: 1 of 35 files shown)


OBarronCS:
I do see value in `pwndbg.aglib.regs.x` being used for fetching and setting values in quick scripting - perhaps in the future this could be put into a separate place, like `pwndbg.scripts.regs.x = 123` that has other nice shorthands for common tasks?

disconnect3d:
Random thought: `.regs.reg_write(...)` is a bit meh, should maybe be `aglib.regs_write(...)`?

I mean, ideally we probably want one unified/simple API in a single namespace for `.{mem,reg}_{read,write}`? 

Though maybe this is out of scope of this PR...

OBarronCS:
Oops - that was a typo - it's `.regs.write_reg()`. I still agree, could be cleaner, it's deeply nested. I think a future PR could find a better location.

OBarronCS:
Pushed change to remove an unnecessary helper function

patryk4815:
PR reverted, some tests are failing

PR #3442, opened 2025-11-27
Optimize and cleanup Dockerfile

jackmisbach:
161a29380b3422cde47b8b384ecf53a3

I cleaned up installing packages to remove redundancy (apt update and install were run twice for some reason before) and make it follow a template I use for apt distros that is cleaner and easier to follow imo
```
 RUN apt-get update && \
    apt-get install -y --no-install-recommends \
      packagename packagename packagename packagename \
    && apt-get clean && rm -rf /var/lib/apt/lists/*
```

I also changed all ADD commands throughout the Dockerfile to be COPY instead of ADD, as that should be the case unless they are archives or retrieved remotely, such as using wget, which in this case they aren't. just a habit I've picked up, but even if they are ADD should really never be used imo, would rather RUN curl even if that was the case, it just does unexpected things behind the scenes
https://docs.docker.com/build/building/best-practices/#add-or-copy
mkdir scripts isn't needed as well, because both the COPY and ADD commands will automatically create directories that don't already exist if you're trying to transfer a file into them

Finally, I added .git/ to the .dockerignore, which reduced build context by a few hundred MB and therefore reduced image size by about 200MB on my machine, but it'll differ by arch, and it only would get larger over time

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -14,3 +14,6 @@ tests/binaries/**/div_zero_binary/binary
 
 # ignore QEMU test images (could also be tainted)
 tests/library/qemu_system/kimages
+
+# ignore .git directory
+.git/
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,19 +22,17 @@ ENV UV_PROJECT_ENVIRONMENT=/venv
 RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && \
     echo $TZ > /etc/timezone && \
     apt-get update && \
-    apt-get install -y locales && \
-    rm -rf /var/lib/apt/lists/* && \
+    apt-get install -y --no-install-recommends \
+        locales vim && \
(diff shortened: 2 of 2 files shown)


.dockerignore
```diff
@@ -14,3 +14,6 @@ tests/binaries/**/div_zero_binary/binary
 
 # ignore QEMU test images (could also be tainted)
 tests/library/qemu_system/kimages
+
+# ignore .git directory
+.git/
```
disconnect3d, line 19: I think we wanted this in case someone wants to do local development inside of docker container. Is it anyhow problematic for us to include that inside of the container?

disconnect3d: Though idk maybe then someone should do `-v ./:/pwndbg` or sth. I haven't used that method for a long time myself

disconnect3d:
I think this makes sense (the .git ignoring too). Thanks!

Issue #3381, opened 2025-10-28
LLDB `break_at` does not work as documented


jxuanli:
In `pwndbg/dbg/__init__.py`

>         Aditionally, one may specify a stop handler function, to be run when
>         the breakpoint or whatchpoint is hit, and that determines whether
>         execution should stop. With a return value of `True` being interpreted
>         as a signal to stop, and a return value of `False` being interpreted as
>         a signal to continue execution. The extent of the actions that may be
>         taken during the stop handler is determined by the debugger.

It seems for LLDB that no matter what I make the stop handler to return, it will always be interpreted as a singal to stop.

disconnect3d:
cc: @mbrla0 can you check this out?

disconnect3d:
Should be fixed by/in #3443 

Please let us know if this is not true ;)

mbrla0:
Wait a second. I'm genuinely confused as to how GitHub still seems to imply #3443 fixes this. But it really doesn't. The fix is quite simple, yes, but I pulled it out of #3443 pretty much immediately after I opened the PR (I'd combined them by mistake)

mbrla0:
Oh, for fuck's sake. Okay, I'm never putting issue names in commit information intended for GitHub ever again.

mbrla0:
Can't seem to be able to reproduce this issue anymore on latest `dev`.

disconnect3d:
Closing this. Please reopen or/and comment if the issue still persists - ideally with steps to reproduce. Thanks!

PR #3443, opened 2025-11-27
Fix handling of `O_NONBLOCK` in the `IODriver` family

mbrla0:
From #3431:
> I'm not entirely sure what the root cause is for these errors, but seeing as these are merely informative messages for Pwndbg developers working on the LLDB backend itself, I feel just swallowing the error and informing the user of it is good enough.

Famous last words.

As it turns out, that instance of `BlockingIOError` was pointing to a much larger problem with how we do IO in pwndbg-lldb, as I would find out while testing out my fix for #3381. 

In short, the issue is that currently Pwndbg assumes that calling `os.set_blocking(sys.stdin.fileno())` will not change the blocking policy on any other file descriptors - especially those backing `sys.stdout` and `sys.stderr`. Turns out [that assumption fails](https://stackoverflow.com/questions/23865898/when-non-blocking-i-o-is-turned-on-for-stdout-is-it-correct-for-the-os-to-turn) in a lot of cases involving Unix-like systems, as - for a lot of very fun reasons I assume - blocking policy is shared among file descriptors created through `dup(2)` or `F_DUPFD`, and `STDIN`, `STDOUT` and `STDERR` are often created from a single file descriptor in that exact way.

This causes `BlockingIOError` to be raised when large amounts of I/O happened during the active periods of the `IODriver`. This is often triggered by verbose mode, but, as I later found out, can also be triggered by inferiors with large amounts of output.

This PR makes it so that `ProcessDriver` and `IODriver` (from the LLDB Pwndbg CLI) operate entirely in terms of non-blocking file descriptors, as that is the lowest common denominator behavior that still guarantees we can have non-blocking `STDIN`, and fixes `BlockingIOError` being raised in the rest of Pwndbg for large amounts of I/O.

The code changes outside `io.py` and `proc.py` are intended to shield the rest of Pwndbg from the move to non-blocking I/O in the CLI, and allow them to stay the same.

### DIFF
--- a/pwndbg/dbg/lldb/__init__.py
+++ b/pwndbg/dbg/lldb/__init__.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import bisect
 import collections
+import enum
 import os
 import random
 import re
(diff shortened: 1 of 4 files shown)


disconnect3d:
Some tests fail here: 
image

mbrla0:
Should be good now

PR #3445, opened 2025-11-28
Bump gdb from pypi + fix gdb-pt-dump dependency

patryk4815:
Our pypi (https://github.com/pwndbg/pypi-for-pwndbg) , now supports wheel for archs:
* linux (glibc 2.28>=), x86_64, aarch64, s390x, powerpc64le, i686
* linux (glibc 2.31>=), armv7l
* linux (glibc 2.36>=), loongarch64
* linux (glibc 2.39>=), riscv64
* macos, x86_64, aarch64

Before wheel was only supported on:
* linux (glibc 2.28>=), x86_64, aarch64
* macos, x86_64, aarch64

After this change with `pt`, it is possible to install pwndbg using `pip`:
```console
$ pip install 'pwndbg[gdb] @ git+https://github.com/pwndbg/pwndbg'
$ pwndbg
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ dependencies = [
     # Optional? only for pretty print traceback
     "rich>=14.1.0,<15",
     # Optional? only for qemu-system vmmap
-    "pt",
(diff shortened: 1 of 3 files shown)


jxuanli:
i have some reasonable beliefs that this is causing the kernel tests to run x2 slower by checking when the kernel tests started to timeout

Issue #3409, opened 2025-11-17
ask the debugger for stack variables when debugging information is present


k4lizen:
Currently we have decompiler integration code that retrieves stack variable names and shows them in various outputs (currently only for binja, soonish for ghidra and ida too). E.g.

<img width="571" height="120" alt="Image" src="https://github.com/user-attachments/assets/24596eda-326b-4a6b-8d00-3366a8e9edbd" />

But if I have debugging information and actually know the exact types, locations and values of the stack variables, no decompiler involved, they will not be shown.

We should add a `Process` API that provides stack variables when debugging info is present. There is already code in place to show them, we just don't retrieve them.

k4lizen:
it is exceedingly funny to me that we haven't thought of this before

disconnect3d:
Done in https://github.com/pwndbg/pwndbg/pull/3451

PR #3451, opened 2025-11-30
Retrieve debug info when present to show stack variables

jackmisbach:
Issue #3409 

asks the debugger for stack variables when debugging information is present using the GDB and LLDB APIs
debug info is prioritized over decompiler integration
variables display with offset notation where needed (`{buffer+0x8}`) 
vmware_rVjQBgG2r7

vmware_sEwXhbpMCH

+ [x] I read the contributing documentation
+ [x] I am providing a screenshot of the (new feature)


pwndbg/gdblib/stack_vars.py
```diff
@@ -0,0 +1,80 @@
+"""
+Looking up stack variable names by their addresses.
+"""
```
k4lizen: Image

This file should be in `pwndbg/dbg/gdb/stack_vars.py` (though possibly there is no need for a new file)

pwndbg/gdblib/stack_vars.py
```diff
@@ -0,0 +1,80 @@
...
+        return None
+
+    if not frame:
+        return None
+
+    return int(frame.pc())
+
+
+@pwndbg.lib.cache.cache_until("stop", "start")
+def _get_current_frame_vars(frame_pc: int) -> tuple[tuple[int, int, str], ...]:
```
k4lizen: `frame_pc` is unused no?

pwndbg/dbg/lldb/__init__.py
```diff
@@ -819,6 +819,64 @@ def evaluate_expression(self, expression: str) -> pwndbg.dbg_mod.Value:
...
+                return None
+            frame = thread.GetSelectedFrame()
+            if not frame.IsValid():
+                return None
+            return frame.GetPC()
+        except Exception:
+            return None
+
+    @pwndbg.lib.cache.cache_until("stop", "start")
+    def _get_current_frame_vars(self, frame_pc: int) -> Tuple[Tuple[int, int, str], ...]:
```
k4lizen: `frame_pc` is unused no?

pwndbg/dbg/lldb/__init__.py
```diff
@@ -819,6 +819,64 @@ def evaluate_expression(self, expression: str) -> pwndbg.dbg_mod.Value:
...
+        try:
+            thread = self.process.GetSelectedThread()
+            if not thread.IsValid():
+                return ()
+
+            frame = thread.GetSelectedFrame()
+            if not frame.IsValid():
+                return ()
```
k4lizen: I 100% support the fact that you are checking this stuff, that is good, but you are also checking it in `_get_frame_pc()` which makes me wonder if there is a way to architecture this PR so that you only need to perform these checks once

pwndbg/dbg/lldb/__init__.py
```diff
@@ -819,6 +819,64 @@ def evaluate_expression(self, expression: str) -> pwndbg.dbg_mod.Value:
...
+            if not thread.IsValid():
+                return None
+            frame = thread.GetSelectedFrame()
+            if not frame.IsValid():
+                return None
+            return frame.GetPC()
+        except Exception:
+            return None
+
+    @pwndbg.lib.cache.cache_until("stop", "start")
```
k4lizen: Image

`import pwndbg.lib.cache`

k4lizen (changes requested):
I think `_get_current_frame_vars` should be a public function (with a different name) in the `pwndbg.dbg.Frame` class. And `get_stack_var_name` should be a thing in `pwndbg.aglib` because there is no debugger-specific code that it needs after it calls `_get_current_frame_vars`.

k4lizen:
It would also be nice if you added a test for this functionality

pwndbg/gdblib/stack_vars.py
jackmisbach: fixed locally - and yeah, no new file is needed. thanks!

pwndbg/dbg/lldb/__init__.py
jackmisbach: Oh lol, thanks! fixed locally, ill push soon

pwndbg/dbg/lldb/__init__.py
jackmisbach: oh, yeah, there is duplicate checking there, fixed locally, thanks!

pwndbg/gdblib/stack_vars.py
jackmisbach: it's not called but used by the @cache_until decorator for the cache key, so that variables are invalidated when switching frames
the gdb.Frame object would represent differnt stack frames at different times

jackmisbach: I updated it though so its much cleaner, ill push here in a bit im going to look at making the test case for this like you asked though first

tests/library/gdb/tests/test_context_commands.py
```diff
@@ -583,3 +585,42 @@ def receive_output(output):
     assert "STACK" not in receive_output.context_output
 
     pwndbg.commands.context.resetcontextoutput("regs")
+
+
+def test_stack_variable_names_from_dwarf(start_binary):
```
k4lizen: no need to have the test defined twice, just the `/dbg/` test is enough because we can run it in gdb with `./tests.sh -g dbg -d gdb`

pwndbg/aglib/stack.py
```diff
@@ -189,3 +189,24 @@ def callstack() -> List[int]:
         frame = frame.parent()
 
     return addresses
+
+
+def get_stack_var_name(address: int) -> str | None:
```
k4lizen, line 195: probably makes sense to cache this

pwndbg/dbg/gdb/__init__.py
```diff
@@ -91,6 +92,36 @@ def parse_and_eval(expression: str, global_context: bool) -> gdb.Value:
         return gdb.parse_and_eval(expression)
 
 
+@pwndbg.lib.cache.cache_until("stop", "start")
+def _get_frame_stack_variables(frame: gdb.Frame, frame_pc: int) -> Tuple[Tuple[int, int, str], ...]:
```
k4lizen: i think this simply will not be cached due to https://github.com/pwndbg/pwndbg/issues/3428 (but if i'm wrong feel free to correct me) . and while I understand the logic behind the `frame_pc` parameter it feels a bit like too much black magic for my liking, and that this problem should be solved in a different way.

i think for now it would be okay to remove caching for this function, and tackle this in the future, because the problem of "i want to cache a Frame" is a thing that needs to be solved in general in the codebase but its not quite obvious how to do it.

pwndbg/dbg/lldb/__init__.py
```diff
@@ -77,6 +78,33 @@ def by_name(self, name: str) -> pwndbg.dbg_mod.Value | None:
         return None
 
 
+@pwndbg.lib.cache.cache_until("stop", "start")
+def _get_frame_stack_variables(
+    frame: lldb.SBFrame, frame_pc: int
+) -> Tuple[Tuple[int, int, str], ...]:
```
k4lizen: same comment as for the lldb function.

k4lizen (changes requested):

pwndbg/dbg/gdb/__init__.py
jackmisbach: Oh, you're right, I forgot about that issue. I've gone ahead and removed it for the time being until we can cache frames

tests/library/gdb/tests/test_context_commands.py
jackmisbach: Alright, I've gone ahead and removed it, but just curious, why are all of the other tests duplicated in that case? I believe that some of the tests in gdb then as well don't need to be there, right? Not that it matters, just wondering so I can have a better understanding for the future. This would just be the first one that is not duplicated, is that the new way it will be done?

k4lizen: > why are all of the other tests duplicated in that case?

thats a good question :)

k4lizen: > This would just be the first one that is not duplicated,

I don't think this is true, @OBarronCS removed some tests from gdb recently iirc, and `test_mallocng.py` is only in `dbg/` for instance.

> is that the new way it will be done?

yeah. only code that is gdb-only (there are some commands like that which haven't been ported), should be in the `gdb/` tests.

jackmisbach: oh yeah you're right, but sounds good will make sure to remember that! 

edit: Wait, but in test_context_commands, everything is duplicated still, no?

k4lizen: no I mean: I agree that there is a lot of duplicate stuff currently, I was just saying that it is not the case that all files are duplicated. probably we want to delete a bunch of stuff from `./gdb/` in the future.

jackmisbach: > probably we want to delete a bunch of stuff from `./gdb/` in the future.

Okay, this is just what I wanted to confirm, 

Thanks for the clarification, I'll push everything here in a second

k4lizen (approved):
lgtm. thanks!

pwndbg/aglib/stack.py
```diff
@@ -189,3 +189,25 @@ def callstack() -> List[int]:
...
+    Returns the variable name with optional offset (e.g., "buf+0x8") if the address
+    is within a variable but not at its start. Returns None if no variable is found.
+    """
+    frame = pwndbg.dbg.selected_frame()
+    if frame is None:
+        return None
+
+    variables = frame.stack_variables()
+
+    for start, end, name in variables:
```
disconnect3d:
```suggestion
    for start, end, name in frame.stack_variables():
```

pwndbg/dbg/gdb/__init__.py
```diff
@@ -91,6 +91,34 @@ def parse_and_eval(expression: str, global_context: bool) -> gdb.Value:
         return gdb.parse_and_eval(expression)
 
 
+def _get_frame_stack_variables(frame: gdb.Frame) -> Tuple[Tuple[int, int, str], ...]:
+    try:
+        block = frame.block()
+    except (gdb.error, RuntimeError):
```
disconnect3d, line 98: Can we add a comment of examples when gdb.error or RuntimeError happens?

jackmisbach:
Hey, I updated the test to match the new schema, and so idk if you want me to revert it and just do it later, or just wait to merge this PR until 3457 is merged?

disconnect3d:
image

Is this a bug?

jackmisbach:
Yeah, mb. I messed it up when refactoring it to work on all architectures last night before I went to bed. Tested it locally, and all tests passed now, so once they pass here should be good to merge and also will match Cypis's new long-term goal for tests working on all architectures and match the new naming he has for the test files, so he won't have to update anything when he updates his branch

disconnect3d:
@jackmisbach we did some test refactoring, can u fix conflict?:(

disconnect3d:
Nvm fixed conflicts here

PR #3459, opened 2025-12-02
Cause thread to continue when `stop_handler` returns `False` in LLDB

### LINKED ISSUE #3381: LLDB `break_at` does not work as documented

mbrla0:
Fix #3381

This PR makes it so that `stop_handler`s passed to LLDB stop points are actually able to decide whether execution continues or is interrupted.

As #3381 indicates, LLDB Breakpoint/Watchpoint handlers apparently cannot continue execution by merely returning a different boolean value, [despite what the reference says](https://lldb.llvm.org/use/tutorials/breakpoint-triggered-scripts.html). Interestingly, the reference [also uses `process.Continue()` elsewhere for this same purpose](https://lldb.llvm.org/use/tutorials/breakpoint-triggered-scripts.html).

In my local testing this fixes the issue, but it would be nice to get some more feedback.


k4lizen:
> nce [also uses process.Continue() elsewhere for this same purpose](https://lldb.llvm.org/use/tutorials/breakpoint-triggered-scripts.html).

maybe you pasted the wrong link here?

k4lizen:
I don't reproduce the issue with this repro:

```python
import lldb

def on_hit(frame, bp_loc, dict):
    print("Hit!")
    return False

def __lldb_init_module(debugger, _dict):
    target = debugger.GetSelectedTarget()
    if not target:
        print("No target selected.")
        return

    process = target.GetProcess()
    thread = process.GetSelectedThread()
    frame = thread.GetSelectedFrame()

    file_path = frame.GetLineEntry().GetFileSpec().fullpath
    if not file_path:
        print("Cannot determine current file.")
        return

    bp = target.BreakpointCreateByLocation(file_path, 5)
    print(f"Created breakpoint {bp.id} at {file_path}:5")

    func_path = f"{__name__}.on_hit"

    command = f"breakpoint command add -F {func_path} {bp.id}"
    debugger.HandleCommand(command)
    print(f"Registered callback: {func_path}")
```
and running:
```bash

(lldb) target create "main"
Current executable set to '/home/user/lldb-bp-hit/main' (x86_64).
(lldb) b main
Breakpoint 1: where = main`main + 4 at main.c:2:7, address = 0x000000000000111d
(lldb) r
Process 13851 launched: '/home/user/lldb-bp-hit/main' (x86_64)
Process 13851 stopped
* thread #1, name = 'main', stop reason = breakpoint 1.1
    frame #0: 0x000055555555511d main`main at main.c:2:7
   1   	int main(){
-> 2   	 int x = 12;
   3   	 x++;
   4   	 x++;
   5   	 x++;
   6   	 x++;
   7   	 x++;
(lldb) command script import script2.py
Created breakpoint 2 at /home/user/lldb-bp-hit/main.c:5
Registered callback: script2.on_hit
(lldb) c
Process 13851 resuming
Hit!
Process 13851 exited with status = 0 (0x00000000) 
(lldb)
```
The process does in fact continue.

k4lizen:
So I'm inclined to believe we are messing something up, but I'm not sure what (maybe there is an exception being raised in the bp callback?).

I'd rather avoid continuing the process like this, seems like it might come back to bite us in magic ways. (though i admit there does seem to be an explicit check for it, if i'm reading the code correctly (https://github.com/llvm/llvm-project/blob/c0f0936f5a47270d47486f6d5860b5f8e30e0e32/lldb/source/Target/StopInfo.cpp#L627))

mbrla0:
> > nce [also uses process.Continue() elsewhere for this same purpose](https://lldb.llvm.org/use/tutorials/breakpoint-triggered-scripts.html).
> 
> maybe you pasted the wrong link here?

Good catch, I did. [This is the one I meant to paste](https://lldb.llvm.org/use/tutorials/script-driven-debugging.html#the-decision-point-breakpoint-commands). (I got the one below it instead)

mbrla0:
> So I'm inclined to believe we are messing something up, but I'm not sure what (maybe there is an exception being raised in the bp callback?).

Yes, this warrants further investigation, then, I'll look into it.

> I'd rather avoid continuing the process like this, seems like it might come back to bite us in magic ways. (though i admit there does seem to be an explicit check for it, if i'm reading the code correctly (https://github.com/llvm/llvm-project/blob/c0f0936f5a47270d47486f6d5860b5f8e30e0e32/lldb/source/Target/StopInfo.cpp#L627))

I don't think that is much of a problem in and of itself. Yes, it is _beyond_ jank, but I'm fairly sure that use is reasonable from the perspective of LLDB, at least insofar as far as the official reference still explicitly uses it that way. Though, if we can avoid it, I absolutely prefer avoiding it.

disconnect3d:
Shall we merge this PR at this point or are we still waiting for something here?

mbrla0:
I'm closing this PR because I haven't been able to reproduce the original issue on latest `dev`, and haven't been for more than two weeks now, as I've already talked about in the Pwndbg Discord. If this becomes relevant again, I'll open a new one.

PR #3460, opened 2025-12-02
tests: relax jemalloc heap address regex for portability

Kaz-scr:
Problem:
- The jemalloc heap tests assumed addresses of the form `0x7ffff[0-9a-fA-F]{6,9}`.
- On my system, jemalloc maps extents at addresses like `0x7ec738c00000` with more
  hex digits, so the tests failed even though the jemalloc-heap output was valid.

Change:
- Relax the address regex used by the jemalloc heap tests to `r"0x[0-9a-fA-F]{6,16}"`.
- Regex is intentionally more general to support different virtual address layouts (ASLR / jemalloc mappings).
  
Rationale:
- The tests only need to verify that pwndbg prints a well-formed hex address, not a
  specific address range tied to a single VA layout.
- The new pattern still validates hex addresses while making the tests pass across
  different environments.

Verification:
- Locally, the following tests now pass:

  - tests/library/gdb/tests/heap/test_heap.py::test_jemalloc_heap
  - tests/library/gdb/tests/heap/test_jemalloc_find_extent
  - tests/library/gdb/tests/heap/test_heap.py::test_jemalloc_extent_info

### DIFF
--- a/tests/library/gdb/tests/heap/test_heap.py
+++ b/tests/library/gdb/tests/heap/test_heap.py
@@ -512,7 +512,9 @@ def test_heuristic_fail_gracefully(start_binary, is_multi_threaded):
 ##
 HEAP_JEMALLOC_EXTENT_INFO = get_binary("heap_jemalloc_extent_info.out")
 HEAP_JEMALLOC_HEAP = get_binary("heap_jemalloc_heap.out")
-re_match_valid_address = r"0x7ffff[0-9a-fA-F]{6,9}"
+# Relax address regex to accept different virtual address layouts (ASLR / jemalloc mappings).
+# Old pattern assumed addresses starting with 0x7ffff and a limited digit count which fails on some hosts.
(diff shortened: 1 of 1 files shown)


disconnect3d:
The previous CI tests failed due to: 
image

Now they stopped failing?

Kaz-scr:
Thanks for the quick CI run! The earlier failures were caused by the test assuming 0x7ffff... addresses, on my machine jemalloc mapped extents at 0x7ec7.... The change relaxes the pattern to r"0x[0-9a-fA-F]{6,16}" so we assert only a well formed hex address and avoid making tests weak across VA layouts/ASLR/jemalloc versions. Happy to narrow it if you prefer a different pattern or add more asserts.
-----------------------------------------------------------------------------------------------------------------------------------------------
Before:
image

image

-----------------------------------------------------------------------------------------------------------------------------------------------
After:
image

image

PR #3461, opened 2025-12-02
Add vmlinux command to load kernel syms from bzImage/vlinux

disconnect3d:
This commit implements the `vmlinux` command that loads kernel debug symbols from a `bzImage` or `vmlinux` files. It works by loading it through `vmlinux-to-elf` tool that needs to be in `$PATH` or provided as `--tool <path>` argument.

I am not sure but for some reason it doesn't work for me that the tool is in `$PATH` and I have to pass it as argument.

Then the `vmlinux` command extracts the kernel ELF from the kernel object file and then loads it at appropriate kernel base address detected by Pwndbg.

This then makes commands such as `kmod` to work correctly.

This is still WIP, but can be merged in current form.

It needs:
1. Vendoring of vmlinux-to-elf tool.
1. Caching of the resulting kernel ELF file as we currently re-create it on each call to `vmlinux` command.

---

Screenshot:
image


pwndbg/commands/vmlinux.py
```diff
@@ -0,0 +1,124 @@
...
+        if not os.access(vmlinux_tool, os.X_OK):
+            print(M.error(f"Specified tool is not executable: {vmlinux_tool}"))
+            return
+    else:
+        # GDB may not inherit the full PATH, so check common locations
+        vmlinux_tool = shutil.which("vmlinux-to-elf")
+        if not vmlinux_tool:
+            # Check common user install locations
+            common_paths = [
+                os.path.expanduser("~/.local/bin/vmlinux-to-elf"),
```
disconnect3d, line 69: that logic is going to be deleted, we have --tool for that

pwndbg/commands/vmlinux.py
```diff
@@ -0,0 +1,111 @@
...
+    "-t",
+    "--tool",
+    type=str,
+    help="Path to the vmlinux-to-elf tool (if not in PATH)",
+)
+
+
+@pwndbg.commands.Command(parser, category=CommandCategory.KERNEL)
+@pwndbg.commands.OnlyWhenQemuKernel
+@pwndbg.commands.OnlyWhenPagingEnabled
```
k4lizen, line 38: do these imply `OnlyWhenRunning`? if not, it should be there (because `pwndbg.dbg.selected_inferior()` would return None, i.e. add-symbol-file doesn't make sense)

jxuanli (approved):

pwndbg/commands/vmlinux.py
```diff
@@ -0,0 +1,111 @@
...
+def vmlinux(filepath: str, tool: str = None) -> None:
+    # Verify the input file exists
+    if not os.path.isfile(filepath):
+        print(M.error(f"File not found: {filepath}"))
+        return
+
+    # Get the kernel base address
+    base = pwndbg.aglib.kernel.arch_paginginfo().kbase
```
jxuanli:
```suggestion
    base = pwndbg.aglib.kernel.kbase()
```

jxuanli: ~~this prob wouldnt work for aarch64 from experience tho, for some reason the first 0x10000 bytes of an aarch64 kernel is not mappped~~

pwndbg/commands/vmlinux.py
```diff
@@ -0,0 +1,111 @@
...
+    if base is None:
+        print(M.error("Unable to locate the kernel base address"))
+        return
+
```
jxuanli, line 50:
```suggestion

    if pwndbg.aglib.arch.name == "aarch64":
        base -= 0x10000

```

jxuanli: hmm maybe `kbase` should be changed instead, somehow the first 0x10000 bytes for aarch64 is not mapped

disconnect3d: Yeah I guess lets move this to kbase?

Also not sure why aarch64 has this offset :x

pwndbg/commands/vmlinux.py
disconnect3d: No, it does not :(

image

k4lizen:
i dont think i fully understand the point of this command. if you need vmlinux-to-elf for this command to work anyway, why dont you just call it once like `vmlinux-to-elf bzImage` and then do `gdb <result of that command>`?

patryk4815:
> i dont think i fully understand the point of this command. if you need vmlinux-to-elf for this command to work anyway, why dont you just call it once like `vmlinux-to-elf bzImage` and then do `gdb <result of that command>`?

we need add more commands 🙃

disconnect3d:
> i dont think i fully understand the point of this command. if you need vmlinux-to-elf for this command to work anyway, why dont you just call it once like `vmlinux-to-elf bzImage` and then do `gdb <result of that command>`?

As we could ship vmlinux-to-elf and have all batteries included? :P 

If others think it makes no sense to have it, we could have it removed too.

PR #3462, opened 2025-12-03
fix: also look in $XDG_CONFIG_HOME (defaults to ~/.config for gdbinit)

c0rydoras:

before:
before

after:
after

### DIFF
--- a/pwndbg/dbg/gdb/__init__.py
+++ b/pwndbg/dbg/gdb/__init__.py
@@ -4,6 +4,7 @@ import re
 from asyncio import CancelledError
 from contextlib import contextmanager
 from contextlib import nullcontext
+from os import environ
 from pathlib import Path
 from typing import Any
 from typing import Coroutine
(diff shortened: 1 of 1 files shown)


pwndbg/dbg/gdb/__init__.py
```diff
@@ -1384,7 +1385,10 @@ def _load_gdbinit(self):
         if disable_any:
             return
 
-        home_file = Path("~/.gdbinit").expanduser().resolve()
+        config_home = environ.get("XDG_CONFIG_HOME", Path("~/.config").expanduser().resolve())
+        home_file = (Path(config_home) / "gdb" / "gdbinit").resolve()
```
disconnect3d, line 1389: ? Why shall `~/gdb/gdbinit` be the first path to check and not `~/.gdbinit`?

c0rydoras: thats how gdb does it ^^

c0rydoras: also respecting the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir/latest/)...

c0rydoras: gdb also looks in `/etc/gdb` for a `gdbinit` and loads that as well

k4lizen (approved):
You can also add the `/etc/gdb/gdbinit` lookup imo. `strace gdb` shows:
```bash
/etc/gdb/gdbinit
~/.config/gdb/gdbinit
~/.gdbinit
.gdbinit
```
but otherwise, lgtm.

c0rydoras:
/etc/gdb/gdbinit is a seperate lookup, which also gets sourced when you have a ./.gdbinit / a ~/.config/gdb/gdbinit / a $XDG_CONFIG_HOME/gdb/gdbinit etc. (imo makes more sense in a different PR)

PR #3463, opened 2025-12-04
Shortcuts

### LINKED ISSUE #2849: Add $heap(offset) convenience function
When dumping information repeatedly over multiple debug sessions of the same binary, with aslr the address changes every run. You can use `p/x $base("heap")+0x10` but that's quite some typing.

GEF has shortcut functions for common memory regions https://hugsy.github.io/gef/functions/heap/ next to `$base` (it is called `$_base` in GEF annoyingly) for `$_heap`, `$_stack` , `$_bss`, and `$_got`. I'd like to see those in pwndbg too!

marywinter3:
Closes #2849


pwndbg/gdblib/functions.py
```diff
@@ -136,6 +136,134 @@ def base(name_pattern: gdb.Value | str):
...
+    pwndbg> p/x $heap(0x10)
+    $2 = 0x555555559010
+    pwndbg> x/32bx $heap()
+    0x555555559000:	0x00	0x00	0x00	0x00	0x00	0x00	0x00	0x00
+    0x555555559008:	0x91	0x02	0x00	0x00	0x00	0x00	0x00	0x00
+    0x555555559010:	0x00	0x00	0x00	0x00	0x00	0x00	0x00	0x00
+    0x555555559018:	0x00	0x00	0x00	0x00	0x00	0x00	0x00	0x00
+    ```
+    """
+    base_addr = gdb.parse_and_eval('$base("heap")')
```
disconnect3d: Can we not involve parsing and evaluation of an expression inside of one?:D

Can we do something like:
```py
heap_base = next(p.start for p in pwndbg.aglib.vmmap.get() if p.objfile == "[heap]", None)

if heap_base is not None:
    return heap_base + int(offset)

<some error handling - raise RuntimeError? or other? not sure>
```

pwndbg/gdblib/functions.py
```diff
@@ -136,6 +136,134 @@ def base(name_pattern: gdb.Value | str):
...
+    0x7ffff7ffd000     0x7ffff7fff000 rw-p     2000   38000 /usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
+    0x7ffffffde000     0x7ffffffff000 rw-p    21000       0 [stack]
+    [...]
+    pwndbg> p/x $stack()
+    $1 = 0x7ffffffde000
+    pwndbg> x/gx $stack()
+    0x7ffffffde000:	0x0000000000000000
+    ```
+    """
+    base_addr = gdb.parse_and_eval('$base("stack")')
```
disconnect3d: Same as in `$heap()` - we want to fetch this without doing additional parsing etc.

If we really want to forward this to `$base` then we can do it without calling `gdb.parse_and_eval`.

pwndbg/gdblib/functions.py
disconnect3d: Alternatively, we could just call the `base` function directly? but its still meh? We probably want to pass in `"[heap]"` string to be more correct.

pwndbg/gdblib/functions.py
```diff
@@ -136,6 +136,134 @@ def base(name_pattern: gdb.Value | str):
...
+    return int(base_addr) + int(offset)
+
+
+def _resolve_symbol_address(candidates: list[str]) -> int | None:
```
disconnect3d, line 196: Pretty sure we have an API for that in aglib, also in gdb lib

pwndbg/gdblib/functions.py
```diff
@@ -136,6 +136,134 @@ def base(name_pattern: gdb.Value | str):
...
+    pwndbg> p/x $bss(0x10)
+    $2 = 0x555555558020
+    ```
+    """
+    bss_symbols = [
+        "__bss_start",
+        "_bss_start",
+        "__bss_start__",
+        "bss_start",
+        "_edata",
```
disconnect3d, line 227: I wonder if there is a way to get this based on `.bss` segment/section name but idk

pwndbg/gdblib/functions.py
```diff
@@ -136,6 +136,134 @@ def base(name_pattern: gdb.Value | str):
...
+    """
+    got_symbols = [
+        "_GLOBAL_OFFSET_TABLE_",
+        "__global_offset_table",
+        "__global_offset_table__",
+        "GOT",
+        "__got_start",
+    ]
+
+    base = _resolve_symbol_address(got_symbols)
```
disconnect3d, line 258: Pretty sure we have some logic to find got somewhere, we can probably reuse it here

disconnect3d (changes requested):
See comments

pwndbg/gdblib/functions.py
```diff
@@ -136,6 +137,131 @@ def base(name_pattern: gdb.Value | str):
...
+
+@GdbFunction(only_when_running=True)
+def got(offset: gdb.Value = gdb.Value(0)) -> int:
+    """
+    Return the address of the Global Offset Table (GOT), optional offset allowed.
+
+    Example:
+    ```
+    pwndbg> p/x $got()
+    $1 = 0x7ffff7e02b80
```
k4lizen, line 245: i tested this command and I get the same output as here, i.e. the wrong output.

the GOT you are looking for in most situations will be the GOT of the main binary, and this function returns the addr of the got in libc.

we likely have existing logic somewhere to find the GOT address, see `pwndbg/commands/got.py`. don't import that directly though, if you need some logic there factor it out to a pwndbg/aglib/ file.

k4lizen (changes requested):
Thanks for the PR. 

The `got` function being borked is a blocker. You also need to run `./scripts/generate_docs.sh` and commit that to pass the CI.

My opinions (which are not blockers for this PR being merged).
1. I don't see the value of passing the offset inside, why not just do `$got() + offset` inside gdb?
2. I think a convenience varaible such as `$bss` would be more convenient than a convenience function such as `$bss()`.
3. If we end up going the path of a function, I think a much more useful parameter would be a pattern-match for the mappings of an objectfile. So something like `$bss("libc")` woud work.

Note that section addresses can be obtained by running the `elf` command. This should also probably be mentioned in the descriptions.

disconnect3d:
This has also merge conflicts as of now that needs to be resolved

disconnect3d:
I'm going to close this for now. Please reopen when ready

marywinter3:
Please reopen this issue @disconnect3d

PR #3272, opened 2025-08-26
Implementing baseline glibc 2.42 support for #3264

gfelber:
* fixes #3264

* implement tcache_perthread_struct changes from https://elixir.bootlin.com/glibc/glibc-2.42/source/malloc/malloc.c#L3127

* implemented new num_slots decrement instead of counts increment: https://elixir.bootlin.com/glibc/glibc-2.42/source/malloc/malloc.c#L3129

* added warning for limited glibc 2.42 feature implementation

Issue #3438, opened 2025-11-26
command doesn't work when there’s a variable named tcache in the program”


abbadinikhilreddy55:

patryk4815:
Steps to reproduce? Or screenshots 🙏

k4lizen:
seems possible the issue was copypasted from this title?:

https://github.com/pwndbg/pwndbg/issues/2310

Issue #2613, opened 2024-12-08
gdb lookup symbol bug


patryk4815:
Gdb should allow `lookup_symbol` for all 'aliases' 
<img width="956" alt="image" src="https://github.com/user-attachments/assets/c67d040d-5a8e-43a8-9c03-901f248ea979">
<img width="582" alt="image" src="https://github.com/user-attachments/assets/d6611479-3999-4488-83c6-4fe0eff6d9cb">

Issue was mentioned:
https://sourceware.org/pipermail/gdb/2022-October/050362.html

Reported upstream: https://sourceware.org/bugzilla/show_bug.cgi?id=32474

Harshal-Patel-In:
Hi, I'm new to open source and would like to work on this issue. Could you please assign it to me?

k4lizen:
@Harshal-Patel-In you don't need to be assigned, simply work on it and submit a PR while following the documentation. try using pwndbg for a CTF to see how it works, molecon qualifiers are today. consider not using AI for the PR.

also, your PR should probably be to gdb itself?

Issue #2310, opened 2024-07-19
`bins` command doesn't work when theres a variable named `tcache` in the program


k4lizen:
### Description
When there's a variable named `tcache` in the program. The `bins` command doesn't work. `tests/gdb-tests/tests/binaries/heap_bins.out` is affected by this. Other name collisions might exist but I dont know about them. It's possible this was known about since: https://github.com/pwndbg/pwndbg/blob/dev/tests/gdb-tests/tests/binaries/heap_malloc_chunk.c#L46 seems to intentionally avoid this collision.

#### In the tests
1. `gdb tests/gdb-tests/tests/binaries/heap_bins.out`
2. After the heap is initialized (in `alloc_chunks()`) run `bins`
3. profit
```
pwndbg> bins
pwndbg will try to resolve the heap symbols via heuristic now since we cannot resolve the heap via the debug symbols.
This might not work in all cases. Use `help set resolve-heap-via-heuristic` for more details.

tcachebins
0x160 [3393]: 0
0x170 [  2]: 0
fastbins
empty
unsortedbin
empty
smallbins
empty
largebins
empty
```
Letting the `tcache_test` function finish and then `bins` is even more fun:
```
pwndbg> bins
tcachebins
0x20 [21849]: 0
0x30 [21845]: 0x31
0x40 [  5]: 0x55500000c659
0x50 [  0]: 0x5227a5a14e8dead6
0x60 [60118]: 0
0x70 [20109]: 0
0x80 [42401]: 0
0x90 [21031]: 0x31
0xa0 [  0]: 0x55500000c669
0xb0 [  0]: 0x5227a5a14e8dead6
0xf0 [  0]: 0x31
0x100 [  0]: 0x55500000c639
0x110 [  0]: 0x5227a5a14e8dead6
0x150 [  0]: 0x31
0x160 [ 49]: 0x55500000c6c9
0x170 [  0]: 0x5227a5a14e8dead6
0x1a0 [51193]: 0
0x1b0 [  0]: 0x31
0x1c0 [21840]: 0
0x1e0 [60118]: 0
0x1f0 [20109]: 0
0x200 [42401]: 0
0x210 [21031]: 0x31
0x270 [  0]: 0x31
0x2d0 [  0]: 0x31
0x2e0 [ 49]: 0
0x320 [51081]: 0
0x330 [  0]: 0x211
0x340 [21840]: 0
0x360 [60118]: 0
0x370 [20109]: 0
[...]

PR #3464, opened 2025-12-04
improved glibc 2.42 support

### LINKED ISSUE #3303: heap inspection does not work at all on arch
### Description

Run any heap command on arch and you'll get
```
AssertionError: Could not find __libc_malloc_initialized or __malloc_initialized
```

### Steps to reproduce

Debug any process on arch and try to run any heap command (e.g. `heap`, `vis`, etc..).

### My setup

Pwndbg:   2025.05.30 build: b6c62f6f2 (Linux)
Python:   3.13.7
GDB:      16.3
Capstone: 6.0.0
Unicorn:  2.1.4
Pwnlib:   4.14.1

### Additional information

My glibc has symbols, and is dynamically linked. The issue is that `__malloc_initialized` has been removed in glibc 2.42. Here it is in 2.41: https://elixir.bootlin.com/glibc/glibc-2.41/source/malloc/arena.c#L116 .

@patryk4815 says that the other one, `__libc_malloc_initialized` has been removed since 2.34, but I can't find it even in earlier versions.

In any case, both of the symbols aren't present in the latest glibc.

I assume a non-trivial amount of people that do pwn use arch, this is a pretty bad bug.

### LINKED ISSUE #3454: GLIBC 2.42, heap commands broken due to missing/changed symbols
### GLIBC 2.42, heap commands broken due to missing/changed symbols

GLIBC 2.42 introduced major changes to symbol names that break a bunch of heap commands. notably tcache commands.
They seem to be doing some major refactoring of the glibc heap code, that might break commands even further in the future.

### Steps to reproduce

Using a modern glibc, e.g. through `archlinux`  docker container and trying to look at tcachebins e.g. through `tcachebins` command.

Affected CTF challenge https://lakectf.epfl.ch/challenges#still-not-malloc-139 
and a real QUICK and BAD patch for tag `2025.10.20` that I wrote during the CTF:

```diff
diff --git a/pwndbg/aglib/heap/ptmalloc.py b/pwndbg/aglib/heap/ptmalloc.py
index b3e883d92..fbee2e3d6 100644
--- a/pwndbg/aglib/heap/ptmalloc.py
+++ b/pwndbg/aglib/heap/ptmalloc.py
@@ -1150,6 +1150,7 @@ class GlibcMemoryAllocator(pwndbg.aglib.heap.heap.MemoryAllocator, Generic[TheTy
     @property
     @pwndbg.lib.cache.cache_until("objfile", "thread")
     def multithreaded(self) -> bool:
+        return True
         """Is malloc operating within a multithreaded environment."""
         addr = pwndbg.aglib.symbol.lookup_symbol_addr("__libc_multiple_threads")
         if addr:
@@ -1260,7 +1261,10 @@ class GlibcMemoryAllocator(pwndbg.aglib.heap.heap.MemoryAllocator, Generic[TheTy
             )
             setattr(GlibcMemoryAllocator.tcachebins, "tcache_2_42_warning_issued", True)
 
-        counts = tcache["counts"]
+        try:
+          counts = tcache["counts"]
+        except Exception:
+          counts = tcache["num_slots"]
         entries = tcache["entries"]
 
         num_tcachebins = entries.type.sizeof // entries.type.target().sizeof
@@ -1275,7 +1279,7 @@ class GlibcMemoryAllocator(pwndbg.aglib.heap.heap.MemoryAllocator, Generic[TheTy
             size = self._request2size(tidx2usize(i))
             count = int(counts[i])
             if pwndbg.glibc.get_version() >= (2, 42):
-                count = pwndbg.aglib.heap.structs.TCACHE_FILL_COUNT - count
+                count = 7 - count
             chain = pwndbg.chain.get(
                 int(entries[i]),
                 offset=self.tcache_next_offset,
@@ -1573,7 +1577,7 @@ class DebugSymsHeap(GlibcMemoryAllocator[pwndbg.dbg_mod.Type, pwndbg.dbg_mod.Val
         return self._main_arena
 
     def has_tcache(self) -> bool:
-        return self.mp is not None and "tcache_bins" in self.mp.type.keys()
+        return True or self.mp is not None and "tcache_bins" in self.mp.type.keys()
 
     @property
     def thread_arena(self) -> Arena | None:
@@ -1715,11 +1719,12 @@ class DebugSymsHeap(GlibcMemoryAllocator[pwndbg.dbg_mod.Type, pwndbg.dbg_mod.Val
         return sbrk_region
 
     def is_initialized(self) -> bool:
-        addr = pwndbg.aglib.symbol.lookup_symbol_addr("__libc_malloc_initialized")
-        if addr is None:
-            addr = pwndbg.aglib.symbol.lookup_symbol_addr("__malloc_initialized")
-        assert addr is not None, "Could not find __libc_malloc_initialized or __malloc_initialized"
-        return pwndbg.aglib.memory.s32(addr) > 0
+        # addr = pwndbg.aglib.symbol.lookup_symbol_addr("__libc_malloc_initialized")
+        # if addr is None:
+        #     addr = pwndbg.aglib.symbol.lookup_symbol_addr("__malloc_initialized")
+        # assert addr is not None, "Could not find __libc_malloc_initialized or __malloc_initialized"
+        # return pwndbg.aglib.memory.s32(addr) > 0
+        return True
 
 
 class SymbolUnresolvableError(Exception):
```

gfelber:
+ update to #3272 after more code has been refactored in glibc 2.42
+ fixes #3454 
+ fixes #3303 
+ added fallback for `DebugSymsHeap.is_initialized `
+ fixed `DebugSymsHeap.has_tcache` for new `struct malloc_par` attribute `.tcache_small_bins`
+ fixed rename from tcache counts to num_slots
+ fixed heap heuristic
+ fixed try_free


disconnect3d:
...so now the code won't work with glibc heap <2.42? 

What about inheriting from the `class GlibcMemoryAllocator` with some `Glibc2_42MemoryAllocator` that would just change whatever is needed to be changed?

TL;DR: We need some abstract way of doing this I think.

Fwiw we also had some classes or structs for older glibcs I think. May be worth looking there as well, but I think they don't differ as much.

disconnect3d:
We also need tests for this.

gfelber:
had a mistake because `"counts" in tcache` isn't allowed. 
Generally speaking this should work for newer and older versions  as it simply adds additional contraints to check. 
I tested it against `ubuntu:24.04` which uses glibc 2.39 and the new `ubuntu:26.04` which uses glibc 2.42.
I didn't test against archlinux directly as getting correct debug symbols is a pain, but it should work if the debug symbols (notably `mp_`) are given.

Also I'm i doing local lint testing wrong somehow, as running:
```bash
uv run --group lint isort --check-only --diff pwndbg pwndbginit tests binja_script.py gdbinit.py ida_script.py scripts
```
doesn't throw any linting errors

gfelber:
do agree that testing would make a lot of sense. would it be possible to just add a test for `ubuntu:26.04`?

gfelber:
this is basically and update to #3272 after more code has been refactored in glibc 2.42

pwndbg/aglib/heap/ptmalloc.py
```diff
@@ -1263,7 +1263,10 @@ def tcachebins(self, tcache_addr: int | None = None) -> Bins | None:
             )
             setattr(GlibcMemoryAllocator.tcachebins, "tcache_2_42_warning_issued", True)
 
-        counts = tcache["counts"]
+        try:
+            counts = tcache["counts"]
+        except Exception:
+            counts = tcache["num_slots"]
```
k4lizen, line 1287: you should add a comment explaining that this is because of 2.42

pwndbg/aglib/heap/ptmalloc.py
```diff
@@ -1576,7 +1579,9 @@ def main_arena(self) -> Arena | None:
         return self._main_arena
 
     def has_tcache(self) -> bool:
-        return self.mp is not None and "tcache_bins" in self.mp.type.keys()
+        return self.mp is not None and any(
+            x in self.mp.type.keys() for x in ["tcache_bins", "tcache_small_bins"]
```
k4lizen, line 1604: ditto

pwndbg/aglib/heap/ptmalloc.py
```diff
@@ -1721,7 +1726,8 @@ def is_initialized(self) -> bool:
         addr = pwndbg.aglib.symbol.lookup_symbol_addr("__libc_malloc_initialized")
         if addr is None:
             addr = pwndbg.aglib.symbol.lookup_symbol_addr("__malloc_initialized")
-        assert addr is not None, "Could not find __libc_malloc_initialized or __malloc_initialized"
+        if addr is None:
+            return int(self.mp["sbrk_base"]) != 0
```
k4lizen, line 1756: ditto

k4lizen (changes requested):

pwndbg/aglib/heap/ptmalloc.py
gfelber: ok, also for some reasons for early commits of the 2.42 version you need to use `tcache["counts"]` and newer commits you need to use `tcache["num_slots"]`

patryk4815:
> do agree that testing would make a lot of sense. would it be possible to just add a test for `ubuntu:26.04`?

you can copy glibc2.42 into this folder:
image

gfelber:
so should i create a new directory 2.42 and add files similarly to 2.33 (and also update the README.md)?

gfelber:
also i guess i need to somehow add the external debug symbols?

gfelber:
yeah it looks like the 2.33 path is hard coded. So i guess this requires more changes?
```bash
$ grep glibcs -R tests
tests/binaries/host/glibcs/README.md:## glibcs used by test
tests/binaries/host/makefile:GLIBC_2_33=$(PWD)/glibcs/2.33
```
but it looks like the debug symbols are just in a hidden folder in the same directory
```bash
$ find tests -name "*2.33*"
tests/binaries/host/glibcs/2.33
tests/binaries/host/glibcs/2.33/.debug/libc-2.33.so
tests/binaries/host/glibcs/2.33/ld-2.33.so
tests/binaries/host/glibcs/2.33/libc-2.33.so
tests/binaries/host/glibcs/2.33/libpthread-2.33.so
```

gfelber:
so looked at the code and started adding tests for the new glibc 2.42 while keeping tests for 2.33.
There is currently a bunch of redundant code, but i prefer if the tests stay seperated.
for some reason `./tests.sh -d dbg -g gdb` works locally and `./tests.sh -d gdb -g gdb` not for GLIBC 2.42.
Not sure if it is due to the kernel, as it seems to expect different layout for /prof/self/map, let's see what happens if i run them through github actions

gfelber:
ok i need to also implement some changes in ptmalloc2.py

gfelber:
also considering removing the tests in /dbg/ again because they don't seems to test any changes between glibc 2.33 and glibc 2.42

gfelber:
ok all checks but `test_try_free_double_free_tcache_2_42` should pass now.
the issue is that it returns
```
General checks
Tcache checks
Using tcache_put
Fastbin checks
----------
Free should succeed!
```
but it should have a line `Will do checks for tcache double-free`.
But this might just be some difference between GLIBC 2.33 and 2.42.
I would just remove/skip it.

gfelber:
Ok the failed tests work locally. I guess I need to spin up a different kernel for testing

gfelber:
ok yeah so most heap tests are actually run against the current distro. So the easiest way to test glibc 2.42 is by adding ubuntu 26.04

gfelber:
ok it isn't possible to add a runner for ubuntu 26.04 as it isn't supported by github (yet).
https://github.com/actions/runner-images

patryk4815:
> ok it isn't possible to add a runner for ubuntu 26.04 as it isn't supported by github (yet). https://github.com/actions/runner-images

add here `ubuntu:26.04` or `fedora:43`

./.github/workflows/docker.yml
Zrzut ekranu 2025-12-5 o 19 41 09

+ modify `./docker-compose.yml`

gfelber:
So there are a bunch of issues with porting the tests to glibc 2.42 (due to major changes to previous versions).
Sadly I don't have any more time to invest here, my current changes should make the basic experience in using pwndbg with glibc 2.42 usable.
~~I started adding skipifs to mark failing tests, I can mark all failed tests with pytest.mark.skipif for GLIBC 2.41 atm so current working tests can be merged.~~

gfelber:
ok the try_free tests seemingly fail at random :(

gfelber:
especially as running with `--pdb` doesn't fail any of the tests

gfelber:
ok added `pytest-rerunfailures==16.1` and `@pytest.mark.flaky(reruns=3, reruns_delay=1)` for the try-free tests as this seemingly makes them work. Not sure what happens but seems to be some issue with the test environment?

tests/library/gdb/tests/heap/test_try_free.py
```diff
@@ -207,7 +208,140 @@ def test_try_free_corrupted_consolidate_backward(start_binary):
...
+    chunks = setup_heap(start_binary, 1, HEAP_BINARY_2_42)
+
+    result = gdb.execute(f"try-free {hex(chunks['a'])}", to_string=True)
+    assert "free(): invalid pointer -> &chunk + chunk->size > max memory" in result
+    os.remove(OUTPUT_FILE)
+
+
+@pytest.mark.flaky(reruns=3, reruns_delay=1)
+def test_try_free_invalid_misaligned_2_42(start_binary):
+    chunks = setup_heap(start_binary, 2, HEAP_BINARY_2_42)
```
disconnect3d: Why are they flaky? This must be something wrong? :c 

Also, do we need strictly separate tests? Can't we make a parametrized test that will take a parameter which would be the binary?

disconnect3d: We should definetely go with parametrized tests instead, and, make the tests not flaky

tests/binaries/host/glibcs/README.md
```diff
@@ -2,3 +2,4 @@
 |     name     | version |         pkgversion         |       docker tag/id       |                              sha256                              |
 | ------------ | ------- | -------------------------- | ------------------------- | ---------------------------------------------------------------- |
 | libc-2.33.so |  2.33   | Ubuntu GLIBC 2.33-0ubuntu5 | ubuntu:21.04/de6f83bfe0b6 | 86ca990a4719b1d4ed8f56e9c6c373e33ad8a40a85fb262cc9ac94ab67feaed0 |
+| libc-2.42.so |  2.42   | Ubuntu GLIBC 2.42-2ubuntu2 | ubuntu:26.04/a1b1375645cc | 901617b8bedb2c5063400661137388b26239253c10788a9f211d086e0c12ea7e |
```
disconnect3d: TODO for me: double check the hashes of the binaries before or right away after merge.

docker-compose.yml
```diff
@@ -38,6 +38,14 @@ services:
     volumes:
       - .:/pwndbg
 
+  ubuntu26.04:
+    <<: *base-spec
+    build:
+      context: .
+      dockerfile: Dockerfile
+      args:
+        image: ubuntu:26.04
```
disconnect3d: If we do that, I assume we also run normal tests on that. Do we then need to include the glibc 2.42 binaries ourselves in the test suite?

uv.lock
```diff
@@ -1972,6 +1974,19 @@ wheels = [
     { url = "https://files.pythonhosted.org/packages/a7/4b/8b78d126e275efa2379b1c2e09dc52cf70df16fc3b90613ef82531499d73/pytest_cov-4.1.0-py3-none-any.whl", hash = "sha256:6ba70b9e97e69fcc3fb45bfeab2d0a138fb65c4d0d6a41ef33983ad114be8c3a", size = 21949, upload-time = "2023-05-24T18:44:54.079Z" },
 ]
 
+[[package]]
+name = "pytest-rerunfailures"
```
disconnect3d: Ideally, we don't want that package to be included - we don't want flaky tests if possible.

tests/binaries/host/glibcs/2.42/ld-linux-x86-64.so.2
```diff
@@ -0,0 +1 @@
+ld-2.42.so
```
disconnect3d: This is a symbolic link, right? GitHub should mark it as such in here too not just in the tab on the left, heh...

pwndbg/commands/ptmalloc2.py
```diff
@@ -1337,7 +1337,13 @@ def finalize(errors_found: int, returned_before_error: bool) -> None:
         and "key" in allocator.tcache_entry.keys()
     ):
         tc_idx = (chunk_size_unmasked - chunk_minsize + malloc_alignment - 1) // malloc_alignment
-        if allocator.mp is not None and tc_idx < int(allocator.mp["tcache_bins"]):
+        # tcache_bins was renamed to tcache_small_bins in GLIBC 2.42
+        tcache_bins = int(
+            allocator.mp["tcache_bins"]
+            if "tcache_bins" in allocator.mp.type.keys()
+            else allocator.mp["tcache_small_bins"]
```
disconnect3d: I don't really like this way of fetching this: `x if A else Y` - I mean, imagine glibc 2.43 changing this again. This gets really bloated/unreadable that way :/.

We should have this as a method, and we should have different class for different versions?

OR: is the glibc version something we can't figure right away? (I think we can?)

(If we can't and we need to do an if, I guess the if could tell us which version it is? lol)

pwndbg/aglib/heap/ptmalloc.py
disconnect3d: Can sbrk_base be <0? :P 

if its unsigned, we could have the `val > 0` done once, heh

pwndbg/aglib/heap/ptmalloc.py
disconnect3d: Damn, that's not so nice, but I guess we have to live with it

k4lizen:
@gfelber thanks for the PR. supporting 2.42 is getting more and more important as time goes on.

I like the allocator-specific changes, but I don't like the testing changes. In particular, I don't think we should be adding/working on `/gdb/` tests, but rather moving everything to `/dbg/` (because both lldb and gdb can run them). The `@pytest.mark.flaky` is also quite spicy. Frankly this should be investigated now, because if it is not, I doubt someone will take the time to look into this in the future.

Pwndbg should have clean support for testing various glibc's, but since we have only tested for one using the Makefile (the various versions are implicitly tested through the distros we test) the weight of figuring this fell on you / this PR even though it is not really connected to glibc heap code.

So, my take is that we should take all testing changes out of this PR, test it sufficiently manually, and I will work on a seperate PR to re-enable testing on archlinux in the CI which will implicitly test this code. @disconnect3d , @patryk4815 what do you think about this?

k4lizen:
I tested the thing on arch with debug symbols and it seems good.

One glibc malloc specific thing I would like to see fixed though is `pwndbg.aglib.heap.structs.MallocPar` and `pwndbg.aglib.heap.structs.DEFAULT_MP_`. The definitions are wrong 2.42 and as a result Pwndbg throws an exception on arch linux when there are no glibc debug symbols, fixing this would probably be enough to enable clean heuristic heap usage.

gfelber:
Ok sgtm. Do you also want to take out the added ubuntu:26.04 container for tests / only remove it from the workflow?

docker-compose.yml
gfelber: the try_free tests are special as the link to locally supplied glibc 2.33 libraries.
the other tests are done with the distro library

pwndbg/commands/ptmalloc2.py
gfelber: this is a similar issue, where early commits of glibc 2.42 still had mp["tcache_bins"] and new ones changed it to mp["tcache_small_bins"].
I guess we can just ignore that edge cases

tests/binaries/host/glibcs/2.42/ld-linux-x86-64.so.2
gfelber: yes symbolic link

tests/library/gdb/tests/heap/test_try_free.py
gfelber: yeah honestly no idea why they are flaky

k4lizen:
> Do you also want to take out the added ubuntu:26.04 container for tests / only remove it from the workflow?

It doesn't pass CI so I would take it out for now.

RocketMaDev:
Will Ubuntu:26.04 lock glibc to 2.43? Since its feature freezing time is close to glibc 2.43 release.

gfelber:
ok, so I will just remove all changes relating to tests and try to fix `pwndbg.aglib.heap.structs.MallocPar` and `pwndbg.aglib.heap.structs.DEFAULT_MP_` as well as implement feedback to the none test code.
I would keep the change to the Dockerfile in though, as `/usr/share/locale/locale.alias` seems to have been removed in ubuntu 26.04 and this will prevent any issues in the future. 
I plan on implementing the changes tomorrow.

disconnect3d:
> Will Ubuntu:26.04 lock glibc to 2.43? Since its feature freezing time is close to glibc 2.43 release.

@RocketMaDev Who knows? I bet they won't make a random fast switch to 2.43, but yeah, we will see I guess.

> @gfelber thanks for the PR. supporting 2.42 is getting more and more important as time goes on.
> 
> I like the allocator-specific changes, but I don't like the testing changes. In particular, I don't think we should be adding/working on `/gdb/` tests, but rather moving everything to `/dbg/` (because both lldb and gdb can run them). The `@pytest.mark.flaky` is also quite spicy. Frankly this should be investigated now, because if it is not, I doubt someone will take the time to look into this in the future.
> 
> Pwndbg should have clean support for testing various glibc's, but since we have only tested for one using the Makefile (the various versions are implicitly tested through the distros we test) the weight of figuring this fell on you / this PR even though it is not really connected to glibc heap code.
> 
> So, my take is that we should take all testing changes out of this PR, test it sufficiently manually, and I will work on a seperate PR to re-enable testing on archlinux in the CI which will implicitly test this code. @disconnect3d , @patryk4815 what do you think about this?

@k4lizen Overall +1, but: why on Arch? Is it to test glibc 2.42? What if they move to other version? I think doing the testing through distros is fine, but Arch is probably not the best here to test 2.42 consistently (?).

Ubuntu 26.04 seems ok, as long as they are on 2.42 and won't switch to 2.43 - if they do - maybe its fine to just leave it as 2.43 if there aren't changes between the two, or, to include glibc 2.42 binary just for tests?

k4lizen:
> Overall +1, but: why on Arch? Is it to test glibc 2.42? What if they move to other version? I think doing the testing through distros is fine, but Arch is probably not the best here to test 2.42 consistently (?).

I agree. I want arch re-enabled because I'm on arch and it allows us to be up to date with various changes that break us. It will test glibc 2.42 until arch moves on to the next version.

If someone wants to enable ubuntu 26.04 or set up better testing infrastructure for differing glibc versions, that would be great (but it seems that is not trivial (e.g. there are non-heap 26.04 test failures), and as such should not be a part of this PR).

gfelber:
ok implemented the changes, I now fail the following tests locallly with glibc 2.42:

```
Time Spent   : 34.73s (cumulative: 522.46s)
Tests Passed : 234
Tests Skipped: 4
Tests Failed : 17

Failing tests:
- tests/library/gdb/tests/heap/test_heap.py::test_mp_heuristic
- tests/library/gdb/tests/heap/test_heap_bins.py::test_heap_bins
- tests/library/gdb/tests/heap/test_vis_heap_chunks.py::test_vis_heap_chunk_command
- tests/library/gdb/tests/test_attachp.py::test_attachp_command_attaches_to_procname
- tests/library/gdb/tests/test_attachp.py::test_attachp_command_attaches_to_pid
- tests/library/gdb/tests/test_attachp.py::test_attachp_command_attaches_to_procname_resolve_none
- tests/library/gdb/tests/test_attachp.py::test_attachp_command_attaches_to_procname_resolve_ask
- tests/library/gdb/tests/test_attachp.py::test_attachp_command_attaches_to_procname_resolve_oldest
- tests/library/gdb/tests/test_attachp.py::test_attachp_command_attaches_to_procname_resolve_newest
- tests/library/gdb/tests/test_attachp.py::test_attachp_command_attaches_to_procname_resolve_none_no_truncate
- tests/library/gdb/tests/test_command_dt.py::test_command_dt_works_with_no_address
- tests/library/gdb/tests/test_command_dt.py::test_command_dt_works_with_address
- tests/library/gdb/tests/test_command_vmmap.py::test_command_vmmap_on_coredump_on_crash_simple_binary[False]
- tests/library/gdb/tests/test_command_vmmap.py::test_command_vmmap_on_coredump_on_crash_simple_binary[True]
- tests/library/gdb/tests/test_commands_elf.py::test_command_elf[reference_bin_nopie.out-False]
- tests/library/gdb/tests/test_commands_elf.py::test_command_elf[reference_bin_pie.out-True]
- tests/library/gdb/tests/test_windbg.py::test_windbg_dX_commands
```
Most of the tests would need to be rewritten as they don't account for new glibc 2.42 behaviour.
Also sadly `test_mp_heuristic` still doesn't seem to work
all the failed attachp tests seems to be due to an problem with my local testing environment, so they can be ignored.

pwndbg/aglib/heap/ptmalloc.py
gfelber: sry I don't understand what you mean with that

disconnect3d: @gfelber if u look what is returned here and outside of the if condition, then iff the sbrk_base is unsigned, we could just do:

```py
val = int(self.mp["sbrk_base"]) if addr is None else pwndbg.aglib.memory.s32(addr)

return val > 0
```

disconnect3d: At the same time, this way is not great/telling why the if is there and which case is which, eh

pwndbg/aglib/heap/ptmalloc.py
```diff
@@ -1672,6 +1687,17 @@ def tcache_perthread_struct(self) -> pwndbg.dbg_mod.Type | None:
     def tcache_entry(self) -> pwndbg.dbg_mod.Type | None:
         return pwndbg.aglib.typeinfo.load("struct tcache_entry")
 
+    @property
+    @pwndbg.lib.cache.cache_until("objfile")
+    def tcache_small_bins(self) -> int | None:
+        if not self.has_tcache():
+            return None
+        if "tcache_small_bins" in self.mp.type.keys():
```
disconnect3d: Hm, weird way of getting it; I'd assume `self.mp` is dict from the return and then would just do `if key in dict` but it seems its probably not a dict since it has a `self.mp.type` field?

disconnect3d: Damn, `mp` is a property:

Image

Shall we maybe fetch it once and cache it in a local variable instead of calling `self.mp` multiple times?

gfelber: I basically copied this from `has_tcache()` function. Seems to work though

gfelber: Why isn't the property cached?

gfelber: but sure will updated and push a change

pwndbg/aglib/heap/structs.py
```diff
@@ -949,12 +949,88 @@ class c_malloc_par_2_35(Structure):
...
+
 class MallocPar(CStruct2GDB):
     """
     This class represents the malloc_par struct with interface compatible with `pwndbg.dbg_mod.Value`.
     """
 
-    if GLIBC_VERSION >= (2, 35):
+    if GLIBC_VERSION >= (2, 42):
+        _c_struct = c_malloc_par_2_42
+    elif GLIBC_VERSION >= (2, 35):
```
disconnect3d, line 1034: I wonder if we should default to the highest version if we don't detect a version or something since now we fallack to 2.12 which is meh?

gfelber: also affects tcache_perthread_struct

gfelber:
are there some remaining changes you need/want in this pull request?

pwndbg/aglib/heap/structs.py
gfelber: also if no version is detected, don't you need to specify it manually with `set glibc VERSION`

gfelber:
ok i realized why heap heuristic stopped working. for some reason glibc changes the value of [mp_.tcache_max_bytes](https://elixir.bootlin.com/glibc/glibc-2.42/source/malloc/malloc.c#L5607) through [do_set_tcache_max](https://elixir.bootlin.com/glibc/glibc-2.42/source/malloc/malloc.c#L5588), which changes it from `MAX_TCACHE_SMALL_SIZE` to `request2size(MAX_TCACHE_SMALL_SIZE) + 1`, breaking the default values in DEFAULT_MP used for heuristics. added a working fix PoC.
Actually no idea where the +1 comes from, it should be `1040`, but according to gdb it is `1041`.
Actually 1041 is hardcoded in the library (not set through a function call), so no idea where it comes from ...

RocketMaDev:
>  Actually 1041 is hardcoded in the library (not set through a function call), so no idea where it comes from ...

Arch don't build the package against release code from glibc, instead, Arch build directly from git tree. Notice that the package is named `2.42+r33+gde1fe81f4714-1`, which indicates that it includes some more commits. Look through git log, here is a message, which add `tcache_max_bytes` by one:

```
malloc/malloc.c:1980:  .tcache_max_bytes = MAX_TCACHE_SMALL_SIZE + 1,

commit ad4caba4146583fc543cd434221dec7113c03e09
Author: Wilco Dijkstra <wilco.dijkstra@arm.com>
Date:   Thu Jul 17 14:31:06 2025 +0000

    malloc: Fix MAX_TCACHE_SMALL_SIZE
    
    MAX_TCACHE_SMALL_SIZE should use chunk size since it is used after
    checked_request2size.  Increase limit of tcache_max_bytes by 1 since all
    comparisons use '<'.  As a result, the last tcache entry is now used as
    expected.
    
    Reviewed-by: DJ Delorie <dj@redhat.com>
```

So that's why you see 1041, instead of 1040, I guess. @gfelber

gfelber:
also just noticed that the tcache_perthread_struct doesn't need to be on the top of the sbrk page ... so e.g. if you allocate something outside tcache first `malloc(0x410)` it will not create a `tcache_perthread_struct` and break tcachebins commands etc. in pwndbg for glibc 2.42. This will break `thread_cache` as it expects the tcache_perthread_struct on the top of the sbrk page

gfelber:
this mainly happens because `multithreaded` is currently returning `False`, due to the missing `__libc_multiple_threads` symbol

gfelber:
they renamed it to `__libc_single_threaded`  and inverted the value ...

pwndbg/aglib/heap/ptmalloc.py
gfelber: `sbrk_base` is a ptr not a bool (as the previous values). I guess it should never be have the highest bit set (as this is usually reserved for kernel addresses).
Please let me know if you want to have the one liner instead anyways.

RocketMaDev:
I noticed that Debian released glibc 2.42 package, but somehow port the commit I mentioned above like Arch packagers do.

gfelber:
yeah ubuntu:26.04 does the same thing

gfelber:
how did aarch64 fail O_o ... debugging

gfelber:
ok wtf why is the tcache symbol wrong on aarch64 ...

gfelber:
ahhh duplicated symbols ...

gfelber:
the duplicate symbols issues needs a global fix, e.g. by only looking up symbol names from libc, seems to be also related to this issue #3438

k4lizen:
https://github.com/pwndbg/pwndbg/issues/2310

gfelber:
yeah the reason you couldn't reproduce it, was probably that is just takes the first chunk on top of the heap if the application is running in single threaded mode (aka the default until you create a new thread with pthread_create). I now prioritze the tcache symbol (if it exists on glibc with symbols) as for glibc 2.42 tcache_perthread_struct doesn't need to be on the top of heap.
But this change makes pwndbg more perceptible if duplicate symbols exist, but it still seems kinda random which symbol is prioritized, e.g. it works fine with x86-64 but it doesn't work for aarch64 (look testes above)

gfelber:
is there some way to easily distinguish if a symbol is from the glibc library?
And if no, can i somehow retrieve the next symbol (alternatively i could use the old fallback and just use the first chunk on heap).

patryk4815:
> is there some way to easily distinguish if a symbol is from the glibc library?

our symbol api can search symbols in specific objfile, eg:
`pwndbg.aglib.symbol.lookup_symbol_addr("your_symbol", objfile_endswith="/libc.so.6")`

But no idea how it behaves with splitted debuginfo eg: `libc.so.6` (stripped) + `.debug-id/<hex>/<hex>.debug`

gfelber:
seems to work even with stripped + splitted debuginfo, at least on the ubuntu:26.04 docker container

gfelber:
also added a fallback in case we have a static binary?

RocketMaDev:
What about search libc pattern in maps first? Some challenges may give you a `libc-2.23.so` or something. That does not end with `libc.so.6`.

k4lizen:
There is `pwndbg.glibc.get_libc_filename_from_info_sharedlibrary`.

gfelber:
afaik there currenly is only `objfile_endswith` as a parameter (for searching for specific libraries, so no pattern search).
but if the first "restriced" search fails it will just search for the symbol in the whole scope.

pwndbg/aglib/heap/ptmalloc.py
```diff
@@ -1157,6 +1170,9 @@ def multithreaded(self) -> bool:
         addr = pwndbg.aglib.symbol.lookup_symbol_addr("__libc_multiple_threads")
         if addr:
             return pwndbg.aglib.memory.s32(addr) > 0
+        # glibc 2.42 replaced __libc_multiple_threads with __libc_single_threaded
+        elif bool(addr := pwndbg.aglib.symbol.lookup_symbol_addr("__libc_single_threaded")):
+            return pwndbg.aglib.memory.s32(addr) == 0
```
disconnect3d: Note: we should probably make this function to ensure the symbol really lies within libc objfile?

pwndbg/aglib/heap/ptmalloc.py
```diff
@@ -1157,6 +1170,9 @@ def multithreaded(self) -> bool:
         addr = pwndbg.aglib.symbol.lookup_symbol_addr("__libc_multiple_threads")
         if addr:
             return pwndbg.aglib.memory.s32(addr) > 0
```
disconnect3d: Random thought: is s32 correct here since we do `>0` instead of `!= 0`? ;)

pwndbg/aglib/heap/ptmalloc.py
```diff
@@ -1252,8 +1268,11 @@ def tcachebins(self, tcache_addr: int | None = None) -> Bins | None:
         if tcache is None:
             return None
 
-        if pwndbg.glibc.get_version() >= (2, 42) and not hasattr(
-            GlibcMemoryAllocator.tcachebins, "tcache_2_42_warning_issued"
+        # this will break expected output during tests, so we skip it
+        if (
+            pwndbg.glibc.get_version() >= (2, 42)
+            and not not hasattr(GlibcMemoryAllocator.tcachebins, "tcache_2_42_warning_issued")
+            and os.environ.get("PWNDBG_IN_TEST") is None
```
disconnect3d, line 1275: Do this PR not implement all glibc 2.42 support?

...since if it does, we can remove the warning?

pwndbg/aglib/heap/ptmalloc.py
```diff
@@ -1598,36 +1624,41 @@ def thread_cache(self) -> pwndbg.dbg_mod.Value | None:
...
-                print(
-                    message.error(
-                        "Error fetching tcache. GDB cannot access "
-                        "thread-local variables unless you compile with -lpthread."
-                    )
-                )
-                return None
+        tcache_ptr = pwndbg.aglib.symbol.lookup_symbol_addr(
+            "tcache",
+            objfile_endswith=pwndbg.glibc.get_libc_filename_from_info_sharedlibrary,
```
disconnect3d: Are we missing a call here?
```suggestion
            objfile_endswith=pwndbg.glibc.get_libc_filename_from_info_sharedlibrary(),
```

pwndbg/aglib/heap/ptmalloc.py
```diff
@@ -1598,36 +1624,41 @@ def thread_cache(self) -> pwndbg.dbg_mod.Value | None:
...
-        return None
+        try:
+            self._thread_cache = pwndbg.aglib.memory.get_typed_pointer_value(
+                self.tcache_perthread_struct, tcache
+            )
+            self._thread_cache["entries"].fetch_lazy()
+        except Exception:
+            print(
+                message.error(
+                    "Error fetching tcache. GDB cannot access "
```
disconnect3d:
```suggestion
                    "Error fetching tcache. Cannot access "
```
Its GDB or LLDB, isn't it?

pwndbg/commands/ptmalloc2.py
```diff
@@ -1354,7 +1354,12 @@ def finalize(errors_found: int, returned_before_error: bool) -> None:
 
             # May be an array, and tc_idx may be negative, so always cast to a
             # pointer before we index into it.
-            counts = allocator.get_tcache()["counts"]
+            # counts was renamed to num_slots in newer version of GLIBC 2.42
+            tcache = allocator.get_tcache()
+            try:
+                counts = tcache["counts"]
+            except Exception:
+                counts = tcache["num_slots"]
```
disconnect3d, line 1360: We should probably default to latest and fallback to the old one, but w/e

pwndbg/aglib/heap/ptmalloc.py
gfelber: No, there is no largbin tache support (which isn't enabled by default). 
Didn't implement anything regarding that

pwndbg/aglib/heap/ptmalloc.py
gfelber: should i also add fallbacks without libc then?

pwndbg/aglib/heap/ptmalloc.py
gfelber: yeah looks like it, should be fixed in latest commit

pwndbg/aglib/heap/ptmalloc.py
disconnect3d: @gfelber there is `and not not` - is this intended - the double not?

pwndbg/aglib/heap/ptmalloc.py
gfelber: yeah i guess u32 or `!=0` is better

pwndbg/aglib/heap/ptmalloc.py
gfelber: sry, so basically you can now enable large bin tcache support (honestly no idea how it works):
https://elixir.bootlin.com/glibc/glibc-2.42/source/malloc/malloc.c#L295

gfelber: i didn't implement anything for this

pwndbg/aglib/heap/ptmalloc.py
gfelber: ~~i guess using the sbrk fallback is better~~
patched the wrong function, but the inferior threads fallback is better than looking outside the shared library for objects too

gfelber: but this affects a bunch of lookup_symbol_addr functions that should probably rather filter only for the current libc. Not sure if it makes sense to patch all these occurrences here. Also to reduce redundancy it is probably best to implement some function or function parameter for this use cases (potentially with fallback support without matching objfile name)

gfelber:
ah ok the lookup with objfile_endswith doesn't work ... it runs into a fallback for #2613, that just queries all obj files
~~i guess we run into the seperate debug file issue :(~~
from my testing this just doesn't work at all, no matter which file you specify, but it might be related to split libc & debug files

gfelber:
also it doesn't work with lldb

gfelber:
so yeah I'm reverting all the "restricted" lookup_symbol_addr with objfile_endswith= and renaming the variable in the test binary again.

gfelber:
if someone wants to fix it here are the broken functions:
https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/dbg/lldb/__init__.py#L1430
https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/dbg/gdb/__init__.py#L717

gfelber:
I don't really want to fix/implement anything else with this pull request.
The goal of adding working 2.42 GLIBC support is achieved.
So I would rather not spend any more time with it.

k4lizen (approved):
looks good and works. thank you for your effort!

PR #3470, opened 2025-12-07
Replace system readelf with pyelftools for multi-arch support

### LINKED ISSUE #3295: Improve `readelf` wrapper to support multi-arch systems
The current https://github.com/pwndbg/pwndbg/blob/dev/pwndbg/wrappers/readelf.py uses the system `readelf`, which only works for the host architecture. This causes issues on multi-arch setups when analyzing foreign binaries.

For example llvm-readelf support multi-arch, maybe zig too? 
Or maybe our code can be replace with `pyelfutils`... or other libraries lief?

carlpinto25:
This PR replaces the dependency on the system's readelf command with the pyelftools library. This addresses issue #3295 by allowing pwndbg to correctly analyze binaries from foreign architectures (e.g., ARM on x86) without relying on the host's toolchain.

image
closes #3295


tests/unit_tests/test_readelf.py
```diff
@@ -0,0 +1,49 @@
+from __future__ import annotations
+
+import os
+import sys
+from unittest.mock import MagicMock
+
+# Mock gdb before importing pwndbg
+sys.modules["gdb"] = MagicMock()
+sys.modules["gdb"].VERSION = "12.1"
```
disconnect3d: I believe this is wrong / not needed.

disconnect3d: If we do need to mock gdb I think we have something done for it already somewhere?

tests/unit_tests/test_readelf.py
```diff
@@ -0,0 +1,49 @@
...
+            assert "type" in item
+            assert "value" in item
+            assert "name" in item
+
+            # Check types
+            assert isinstance(item["offset"], int)
+            assert isinstance(item["value"], int)
+            assert isinstance(item["name"], str)
+
+    print("test_get_got_entry passed")
```
disconnect3d: This print is redundant

pwndbg/commands/got.py
```diff
@@ -141,37 +141,29 @@ def _got(path: str, accept_readonly: bool, symbol_filter: str) -> None:
...
+    for category, entries in got_entry.items():
+        for entry in entries:
+            offset = entry["offset"]
+            value = entry["value"]
+            name = entry["name"]
+
+            # Ensure types for mypy
+            assert isinstance(offset, int)
+            assert isinstance(value, int)
+            assert isinstance(name, str)
```
disconnect3d, line 140: Can't we just do sth like `offset: int = ...` instead or it won't work?

pwndbg/wrappers/readelf.py
```diff
@@ -17,18 +18,81 @@ class RelocationType(Enum):
...
-    entries: Dict[RelocationType, List[str]] = {category: [] for category in RelocationType}
-    for line in readelf_out.splitlines():
-        if not line or not line[0].isdigit() or " " not in line:
-            continue
-        category = line.split()[2]
-        # TODO/FIXME: There's a bug here, somehow the IRELATIVE relocation might point to somewhere in .data.rel.ro, which is not in .got or .got.plt
-        for c in RelocationType:
-            if c.name in category:
-                entries[c].append(line)
+def get_got_entry(local_path: str) -> Dict[RelocationType, List[Dict[str, int | str]]]:
```
disconnect3d: Why did you rewrite the logic here? Do the old logic with parsing lines had any bugs or something?

I get it we may want to return a structured output - that's fine - but - does the code do exactly the same stuff now?

tests/unit_tests/test_readelf.py
```diff
@@ -0,0 +1,49 @@
...
+            assert "offset" in item
+            assert "info" in item
+            assert "type" in item
+            assert "value" in item
+            assert "name" in item
+
+            # Check types
+            assert isinstance(item["offset"], int)
+            assert isinstance(item["value"], int)
+            assert isinstance(item["name"], str)
```
disconnect3d, line 61: Honestly, the test should really check for some values here instead of just checking if the types match

disconnect3d (changes requested):
Please see comments. I don't quite understand why did we change the `get_got_entry` parsing. I understand the want of it returning a dict, though a `NamedTuple` would be much better here, but I don't understand why did we change the parsing logic to something else.

Some questions and notes, apart from what is there in the comments:
1. Was there any deficiency or a bug with previous parsing logic?
2. The `tests/binaries/host/reference-binary` must be removed from the PR. Isn't it gitignored anyway?

tests/unit_tests/test_readelf.py
carlpinto25: Added checks for specific symbol names (puts, __libc_start_main) and GLIBC version presence

tests/unit_tests/test_readelf.py
carlpinto25: yes there is i updated the code

tests/unit_tests/test_readelf.py
carlpinto25: my bad its removed now

pwndbg/commands/got.py
carlpinto25: yes its possible i updated it

carlpinto25:
@disconnect3d The old logic had executed the system's readelf command, which is architecture-specific. When debugging an ARM binary on x86_64, the host's readelf fails to parse ARM-specific structures correctly, causing incorrect/missing GOT entries. This is literally issue #3295.

> Please see comments. I don't quite understand why did we change the `get_got_entry` parsing. I understand the want of it returning a dict, though a `NamedTuple` would be much better here, but I don't understand why did we change the parsing logic to something else.
>
Agreed it'd be cleaner. Should I change it in this PR?
> 2. The `tests/binaries/host/reference-binary` must be removed from the PR. Isn't it gitignored anyway?
>
i removed it

carlpinto25:
@disconnect3d I'm seeing conflicting mypy requirements for the pyelftools calls in 
readelf.py
:

With # type: ignore[no-untyped-call]:
mypy --strict passes
mypy fails with "unused 'type: ignore' comment"

Without the comments:
mypy passes
mypy --strict fails with "Call to untyped function"

disconnect3d:
> @disconnect3d I'm seeing conflicting mypy requirements for the pyelftools calls in readelf.py :
> 
> With # type: ignore[no-untyped-call]: mypy --strict passes mypy fails with "unused 'type: ignore' comment"
> 
> Without the comments: mypy passes mypy --strict fails with "Call to untyped function"

Please make strict happy ideally by adding additional annotations instead of 'unused type ignore' comments.

Anyway, If one works and another fails, don't worry and let us know, we are going to fix this soon (we just noticed this problem too).

carlpinto25:
@disconnect3d  Thanks for the suggestion!
>Please make strict happy ideally by adding additional annotations instead of 'unused type ignore' comments.

adding annotations didnt work,snice nothing what i tried worked i did Mypy config override since it would be cleaner as temporary fix

pyproject.toml
```diff
@@ -251,6 +251,13 @@ module = [
 ]
 ignore_missing_imports = true
 
+[[tool.mypy.overrides]]
+module = [
+    "pwndbg.wrappers.readelf",
+    "pwndbg.commands.got",
+]
+disable_error_code = ["no-untyped-call"]
```
disconnect3d, line 284: @k4lizen do we want those changes?

disconnect3d: What do the tool mypy overrides module do even ugh

k4lizen: we dont really want it but idk of a better solution so yeah its fine

jxuanli:
I'd also suggest that the readelf wrapper file should be removed since it is now technically not a wrapper anymore. It's one functions anyways and only gets invoked in one place so we should move it into commands/got.py

pyproject.toml
jxuanli: more annotations?

jxuanli: or maybe add the `#[no-typed-call]` at the end of the effected lines rather the entire file

carlpinto25:
@disconnect3d  all the tests pass now also what do you think of 

> I'd also suggest that the readelf wrapper file should be removed since it is now technically not a wrapper anymore. It's one functions anyways and only gets invoked in one place so we should move it into commands/got.py

disconnect3d:
> @disconnect3d all the tests pass now also what do you think of
> 
> > I'd also suggest that the readelf wrapper file should be removed since it is now technically not a wrapper anymore. It's one functions anyways and only gets invoked in one place so we should move it into commands/got.py

That sounds good / makes sense to me.

disconnect3d:
I will just merge this and open an issue for the moving of this to got.py

disconnect3d:
Created https://github.com/pwndbg/pwndbg/issues/3844

PR #3468, opened 2025-12-06
Make tests architecture-agnostic v4

patryk4815:
This PR introduces additional changes to architecture-agnostic test execution; after this PR, the tests can be run on aarch64 without any failing tests.

aarch64:
```
Time Spent   : 104.12s (cumulative: 402.48s)
Tests Passed : 114
Tests Skipped: 75
Tests Failed : 0
```

PR #3471, opened 2025-12-07
Remove regs getattr + setattr

OBarronCS:
This fixes the failing tests that reverted #3441. I had missed some cases of using `setattr(pwndbg,aglib.regs, name, val)`, which didn't crash as it simply added an attribute to the regs object. These are now replaced with `pwndbg.aglib.regs.write_reg(name, val)`

Recap of the PR:

This PR removes the `__getattr__` and `__setattr__` methods of the `pwndbg.aglib.regs` module, which were used to fetch and set the values of registers. To unify the API, reads now all use `pwndbg.aglib.regs.read_reg(reg_name)`, and writes use `pwndbg.aglib.regs.write_reg(reg_name, val)`

Two special getter (and setter) properties are added to the object, pc and sp. This means you can still do `pwndbg.aglib.regs.pc` and `pwndbg.agilb.regs.sp` to read from these special registers, which is very common in the codebase. By specifically handling them, we also gain the advantage of resolving the name of the actual stack pointer (like `rsp`) before forwarding the request to the underlying debugger, instead of depending on the debugger to convert `sp -> rsp`.

### DIFF
--- a/pwndbg/dbg/lldb/__init__.py
+++ b/pwndbg/dbg/lldb/__init__.py
@@ -171,10 +171,12 @@ class LLDBFrame(pwndbg.dbg_mod.Frame):
         if name not in pwndbg.aglib.regs:
             return False
 
-        # Writing to the PC using the normal register write flow causes the
+        current_reg_set = reg_sets[pwndbg.aglib.arch.name]
+
+        # Writing to the PC (or any subregister of the PC) using the normal register write flow causes the
         # inner object to be automatically invalidated by LLDB, so we have to
         # handle jumps manually using SBFrame::SetPC.
-        if name in (reg_sets[pwndbg.aglib.arch.name].pc, "pc"):
+        if name == "pc" or current_reg_set.full_register_lookup.get(name) == current_reg_set.pc:
             ret = self.inner.SetPC(val)
             self.proc.dbg._trigger_event(pwndbg.dbg_mod.EventType.REGISTER_CHANGED)
             return ret
@@ -229,12 +231,12 @@ class LLDBFrame(pwndbg.dbg_mod.Frame):
                     # event.
                     self.proc.dbg._trigger_event(pwndbg.dbg_mod.EventType.REGISTER_CHANGED)
 
-                    # If we set the stack pointer, the inner object might have been invalidated, try
+                    # If we set the stack pointer (or a subregister of the stack pointer), the inner object might have been invalidated, try
                     # to restore it, as it should still be the selected frame.
                     if (
-                        name in (reg_sets[pwndbg.aglib.arch.name].stack, "sp")
-                        and not self.inner.IsValid()
-                    ):
+                        name == "sp"
+                        or current_reg_set.full_register_lookup.get(name) == current_reg_set.stack
+                    ) and not self.inner.IsValid():
                         self.inner = thread.GetSelectedFrame()
                         assert self.inner.GetSP() == val
 
--- a/pwndbg/dbg/lldb/__init__.py
+++ b/pwndbg/dbg/lldb/__init__.py
@@ -232,7 +232,7 @@ class LLDBFrame(pwndbg.dbg_mod.Frame):
                     # If we set the stack pointer, the inner object might have been invalidated, try
                     # to restore it, as it should still be the selected frame.
                     if (
-                        name in (reg_sets[pwndbg.aglib.arch.name].frame, "sp")
+                        name in (reg_sets[pwndbg.aglib.arch.name].stack, "sp")
                         and not self.inner.IsValid()
                     ):
                         self.inner = thread.GetSelectedFrame()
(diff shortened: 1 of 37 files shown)


tests/library/dbg/tests/test_command_xor.py
```diff
@@ -34,7 +34,7 @@ async def test_command_xor_with_int(ctrl: Controller) -> None:
 
     await ctrl.launch(REFERENCE_BINARY)
 
-    before = pwndbg.aglib.regs.rsp
+    before = pwndbg.aglib.regs.read_reg("rsp")
```
OBarronCS: Ahh - these are going to have conflicts with the changes in #3468 to makes these tests architecture agnostic - the read is replaced with a read of "sp"

tests/library/gdb/tests/test_command_cyclic.py
```diff
@@ -39,8 +39,9 @@ def test_command_cyclic_register(start_binary):
     ptr_size = pwndbg.aglib.arch.ptrsize
     test_offset = 45
     pattern = cyclic(length=80, n=ptr_size)
-    pwndbg.aglib.regs.rdi = int.from_bytes(
-        pattern[test_offset : test_offset + ptr_size], pwndbg.aglib.arch.endian
+    pwndbg.aglib.regs.write_reg(
```
OBarronCS, line 42: A fair number of the edited tests under gdb/tests appear to be duplicates of tests in dbg/test. A future PR should remove these, as its adding unnecessary extra work for these PRs.

patryk4815: yes!

tests/library/dbg/tests/test_command_telescope.py
```diff
@@ -146,7 +146,7 @@ async def test_command_telescope_frame_bp_below_sp(ctrl: Controller) -> None:
     await launch_to(ctrl, TELESCOPE_BINARY, "break_here")
     await ctrl.execute("memoize")  # turn off cache
 
-    pwndbg.aglib.regs.sp = pwndbg.aglib.regs.read_reg(pwndbg.aglib.regs.frame) + 1
```
OBarronCS, line 149: I think it would be a good idea to add a @property.setter for the sp and pc values on the regs module. This pattern is used very rarely in the codebase (basically in a couple tests), however without it, it's natural to resort to `pwndbg.aglib.regs.write_reg("sp", ...)` (as I did here) which is incorrect, since you should be getting the real stack pointer name from `.regs.current.stack` to ensure we are doing the conversion from sp -> real architectural name at the pwndbg level. Having a setter property would avoid this sort of mistake

OBarronCS: Did this in latest commits

pwndbg/dbg/lldb/__init__.py
```diff
@@ -229,12 +231,12 @@ def reg_write(self, name: str, val: int) -> bool:
...
-                    # If we set the stack pointer, the inner object might have been invalidated, try
+                    # If we set the stack pointer (or a subregister of the stack pointer), the inner object might have been invalidated, try
                     # to restore it, as it should still be the selected frame.
                     if (
-                        name in (reg_sets[pwndbg.aglib.arch.name].frame, "sp")
-                        and not self.inner.IsValid()
-                    ):
+                        name == "sp"
+                        or current_reg_set.full_register_lookup.get(name) == current_reg_set.stack
+                    ) and not self.inner.IsValid():
```
OBarronCS: This logic ensures that if we explicitly write to a subregister of rsp (like esp), we still restore the frame.

pwndbg/lib/regs.py
```diff
@@ -154,47 +154,52 @@ class Reg:
 
 
 class RegisterSet:
-    #: Program counter register
     pc: str
+    """Program counter register"""
```
OBarronCS, line 158: I moved these comments so intellisense detects them

OBarronCS:
This PR is ready - the CI errors looks like transient error (jemalloc not finding mappings)

disconnect3d:
Hmm there is a `test_command_ktask` failure after those changes on arm64. Can it be that we break how some special arm64 reg is fetched or sth?

 https://github.com/pwndbg/pwndbg/actions/runs/20112499981/job/57713345207?pr=3471#step:14:441

OBarronCS:
I ran the test locally and the systems tests pass for all the images, including for the one that fails in CI: `linux-6.1.79-arm64 `. I took another look at all codepaths that access registers from these kernel commands, and outside of changing `pwndbg.aglib.reg.REGNAME` to `pwndbg.aglib.regs.read_read("REGNAME")` (and removing using getattr to get reg, which I checked doesn't try to access the special .idt or .idt_limit helper functions), I don't see where this could have failed.

For peace of mind, I think we should rerun the `tests-using-nix (ubuntu-24.04, qemu-system-tests)` test in CI to verify it wasn't a random test failure

patryk4815:
> test in CI to verify it wasn't a random test failure

yes, it was random

patryk4815 (approved):

k4lizen (approved):

PR #3474, opened 2025-12-09
Improve load_custom_structure message

### LINKED ISSUE #3472: "Symbols are loaded!" is confusing when debugging kernel
This happens on af90112ca and imho is confusing at times:

Image

This print comes from cymbol hook that loads some symbols but imho it is not great UX-wise:

Image

Like... what symbols are loaded? Types defined by me? What types? We should be more specific here and inform the user better :).

disconnect3d:
Fixes #3472. Makes the "Symbols are loaded!" message less confusing by changing it to: "Loaded custom symbols! (from {path})" so that the user can inspect what were the symbols loaded.

Before:
image

After:
image


k4lizen:
the fact that this message shows up at all is a bug (because the cymbol api should be structured differently)

this message should only show up when the user explicitly invokes cymbol, and in that case im not sure seeing the fs path is very valuable (since the cymbol command has an option to list loaded symbols and print them right?)

but yeah, doesnt hurt ig

jxuanli (approved):

k4lizen (approved):

PR #3478, opened 2025-12-10
Re-enable archlinux tests

k4lizen:
Fixed everything except for heap.

image

Probably ubuntu26.04 should be easy to add as well. Related: https://github.com/pwndbg/pwndbg/pull/3464


setup-dev.sh
```diff
@@ -141,6 +141,7 @@ EOF
         nasm \
         gcc \
         glibc-debug \
+        lib32-glibc \
```
patryk4815, line 144: this should be only x86-64 option, not for aarch64 probably

patryk4815: no idea if there is archlinux:aarch64 image on dockerhub ;/

tests/library/dbg/tests/test_command_canary.py
```diff
@@ -44,6 +46,7 @@ async def test_command_canary(ctrl: Controller, binary: str, reg_name: str, skip
         await ctrl.step_instruction()
 
     canary_value, at_random = pwndbg.commands.canary.canary_value()
+    assert at_random is not None
```
disconnect3d, line 49: Why do we need the changes in this file? The test failed on Arch but didn't on other setups? Why?

tests/library/dbg/tests/test_command_dt.py
```diff
@@ -27,10 +27,11 @@ async def test_command_dt_works_with_address(ctrl: Controller) -> None:
 
     out = await ctrl.execute_and_capture(f'dt "struct tcache_perthread_struct" {tcache_addr}')
 
+    # Accounting for differences between architectures and glibc versions (specifically 2.42)
     exp_regex = (
         "struct tcache_perthread_struct @ 0x[0-9a-f]+\n"
-        "    0x[0-9a-f]+ \\+0x0000 counts +: +.*\\{([0-9]+, [0-9]+ <repeats 63 times>|(\\s*\\[[0-9]+\\] = [0-9]){20,64}\\s*([.]+\\s*)?)\\}\n"
-        "    0x[0-9a-f]+ \\+0x[0-9a-f]{4} entries +: +.*\\{(0x[0-9a-f]+, 0x[0-9a-f]+ <repeats 63 times>|(\\s*\\[[0-9]+\\] = (0x[0-9a-f]+|NULL)){20,64}\\s*([.]+\\s*)?)\\}"
+        "    0x[0-9a-f]+ \\+0x0000 (counts|num_slots) +: +.*\\{([0-9]+, [0-9]+ <repeats (63|75) times>|(\\s*\\[[0-9]+\\] = [0-9]){20,76}\\s*([.]+\\s*)?)\\}\n"
+        "    0x[0-9a-f]+ \\+0x[0-9a-f]{4} entries +: +.*\\{(0x[0-9a-f]+, 0x[0-9a-f]+ <repeats (63|75) times>|(\\s*\\[[0-9]+\\] = (0x[0-9a-f]+|NULL)){20,76}\\s*([.]+\\s*)?)\\}"
```
disconnect3d, line 34: Heh, this is ugly, but I guess its the easiest way to do it...

setup-dev.sh
k4lizen: to clarify, lib32-glibc is needed for the 32-bit ld required to run the i386 binaries.

> this should be only x86-64 option, not for aarch64 probably

that makes sense. we currently don't have any architecture-checking code in setup-dev.sh but we could change that.

image

doesn't exist :(

tests/library/dbg/tests/test_command_canary.py
k4lizen: these changes are just code cleanup to satisfy lsp

disconnect3d: LSP? Or some linter[s] we do have?

k4lizen: it specifically fixed a few `mypy --strict` complaints. (but to be clear, i think these changes are good, which is why they are here, they are not here simply to appease the almighty linter gods)

tests/library/dbg/tests/test_commands_elf.py
```diff
@@ -77,7 +77,8 @@ async def test_command_elf(ctrl: Controller, binary_name: str, is_pie: bool) ->
         pytest.skip("TODO multiarch")
 
     out = (await ctrl.execute_and_capture("elf")).splitlines()
-    assert len(out) == 25
+    # Never versions of gcc emit an additional `.sframe` section.
```
patryk4815, line 80: Newer

PR #3479, opened 2025-12-11
Revert "Add vmlinux command to load kernel syms from bzImage/vlinux (…

k4lizen:
…#3461)"

This reverts commit a97aff4f1c3370cc31c13a93f45217c837f96fb9.

https://github.com/pwndbg/pwndbg/pull/3461#issuecomment-3634921013 :

> As we could ship vmlinux-to-elf and have all batteries included? :P
> 
> If others think it makes no sense to have it, we could have it removed too.

@patryk4815 @disconnect3d and esp. @jxuanli wdyt?

I think we should revert this because:
1. Just doing `vmlinux-to-elf bzImage` and running `gdb <output>` works.
    + It is more ergonomic, because you only pay the cost of running the tool once, and don't need to put the call in your `.gdbinit` or whatever.
2. Having the `vmlinux` command might confusing people into actually using the command, instead of the more ergonomic alternative.
3. The command is more code that needs to be maintained.


jxuanli:
Ehh I don't have a strong opinion either way but your argument makes sense to me. I'll also take this chance to (shamelessly) plug my own tool that autodetects vmlinux/bzImage and source the vmlinux in gdb so it sort of does what I would use this command for: https://github.com/jxuanli/kpinit

PR #3481, opened 2025-12-11
chore: run pyupgrade (via ruff) && enable pyupgrade rules

c0rydoras:

### DIFF
--- a/pwndbg/aglib/elf.py
+++ b/pwndbg/aglib/elf.py
@@ -13,8 +13,8 @@ import importlib
 import subprocess
 import sys
 from typing import NamedTuple
+from typing import TypeAlias
 from typing import TypeVar
-from typing import Union
 
 from elftools.elf.constants import SH_FLAGS
 from elftools.elf.elffile import ELFFile
@@ -65,8 +65,8 @@ class ELFInfo(NamedTuple):
         return self.is_pic
 
 
-Ehdr = Union[pwndbg.lib.elftypes.Elf32_Ehdr, pwndbg.lib.elftypes.Elf64_Ehdr]
-Phdr = Union[pwndbg.lib.elftypes.Elf32_Phdr, pwndbg.lib.elftypes.Elf64_Phdr]
+Ehdr: TypeAlias = pwndbg.lib.elftypes.Elf32_Ehdr | pwndbg.lib.elftypes.Elf64_Ehdr
+Phdr: TypeAlias = pwndbg.lib.elftypes.Elf32_Phdr | pwndbg.lib.elftypes.Elf64_Phdr
 
 
 @pwndbg.dbg.event_handler(EventType.START)
--- a/pwndbg/aglib/elf.py
+++ b/pwndbg/aglib/elf.py
@@ -12,10 +12,7 @@ import ctypes
 import importlib
 import subprocess
 import sys
-from typing import Dict
-from typing import List
 from typing import NamedTuple
-from typing import Tuple
 from typing import TypeVar
 from typing import Union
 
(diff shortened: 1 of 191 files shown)


pwndbg/aglib/elf.py
```diff
@@ -62,8 +62,8 @@ def is_pie(self) -> bool:
         return self.is_pic
 
 
-Ehdr = Union[pwndbg.lib.elftypes.Elf32_Ehdr, pwndbg.lib.elftypes.Elf64_Ehdr]
-Phdr = Union[pwndbg.lib.elftypes.Elf32_Phdr, pwndbg.lib.elftypes.Elf64_Phdr]
+Ehdr: TypeAlias = pwndbg.lib.elftypes.Elf32_Ehdr | pwndbg.lib.elftypes.Elf64_Ehdr
+Phdr: TypeAlias = pwndbg.lib.elftypes.Elf32_Phdr | pwndbg.lib.elftypes.Elf64_Phdr
```
k4lizen: what is `: TypeAlias`?

pyproject.toml
```diff
@@ -116,6 +116,7 @@ select = [
     "RET506", # flake8-return: superfluous-else-raise
     "RET507", # flake8-return: superfluous-else-continue
     "RET508", # flake8-return: superfluous-else-break
+    "UP",     # pyupgrade
```
k4lizen: are you sure this upgrades to the minimal python version specified in pyproject.toml as opposed to the currently installed python version? maybe i'm blind but I don't see any clarification about this in the ruff/pyupgrade docs

k4lizen (commented):
i guess `typing` was added to the python standard library before types like `list` etc. were added to the language itself?

pyproject.toml
c0rydoras: it does, otherwise this PR would look very differently (e.g. instead of `A: TypeAlias = X | Y | None`, you could do `type A = X | Y | None`)

pwndbg/aglib/elf.py
c0rydoras: https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases

c0rydoras:
`list`, `tuple` etc. used to not be generic, e.g. `list[int]` wouldn't be valid

see https://docs.python.org/3/whatsnew/3.9.html#type-hinting-generics-in-standard-collections

k4lizen (approved):
neat , thanks!

pyproject.toml
c0rydoras: e.g. if you change the python version with `requires-python = ">=3.13, <4"`

image

k4lizen:
@disconnect3d @jxuanli even though there are a bunch of PR's open now, I don't think they would be majorly affected if this was done now (at the time there were, most of the stuff open now is kind of slow (except for the CI changes, which don't touch python anyway :p)).

@c0rydoras would have to re-run the lint though

pwndbg/commands/probeleak.py
```diff
@@ -154,8 +154,8 @@ def probeleak(
             else:
                 right_text = f"({page.permstr}) {mod_name} + 0x{p - page.start:x}"
 
-            offset_text = "0x%0*x" % (off_zeros, i)
-            p_text = "0x%0*x" % (int(ptrsize * 2), p)
+            offset_text = f"0x{i:0{off_zeros}x}"
+            p_text = f"0x{p:0{int(ptrsize * 2)}x}"
```
k4lizen, line 158: this works?

k4lizen (approved):
some of the tests looked prettier the way they were previously since they were aligned, but i don't think that actually matters so LGTM

pwndbg/commands/probeleak.py
c0rydoras: yes

c0rydoras:
> some of the tests looked prettier the way they were previously since they were aligned

yea :pensive:

PR #3482, opened 2025-12-11
Show backtrace in invalid allocations in kmem-trace

k4lizen:
If I pass `-v` to `kmem-trace` I want to see the backtrace even for "invalid" allocations (currently they are "invalid" because we need to fix our code, but these allocations would also be interesting if they were "invalid" due to some kernel shenanigens).

Before:
image

After:
image


jxuanli:
I think we should instead figure out why `find_containing_slab_cache` throws an exception

k4lizen:
> I think we should instead figure out why find_containing_slab_cache throws an exception

We for sure should, but I am kind of under the impression that it will be a bit of a cat and mouse game of new kernel versions coming out and us fixing the `slab` command, and during the time it takes us to fix this stuff it would be nice to still have *something* working

jxuanli:
can you run slab contains to see if that fails, if not then its not on the kversion

jxuanli:
also it should work for all 5.x and 6.x releases to date

k4lizen:
> can you run slab contains to see if that fails, if not then its not on the kversion

it does. has failed me on various versions even, but the latest issue that I could reproduce you is not mainline kernel but the fork used for google kernelctf which implements some extra mitigations (with debugging info).

jxuanli:
are you on mitigation or cos? lts works for me, perhaps #3394 accidentally fixed some issues lol

disconnect3d:
Sounds to me as this makes sense? If you disagree @jxuanli let us know :P

PR #3485, opened 2025-12-12
Fix tests v5

patryk4815:
Changes:
- pwndbg now install without problems on python 3.14 (pydantic-core issue)
- i386 tests should run only when kernel support IA32
- enhance setup.sh/setup-dev.sh
- opt-out installing i386-glibc with setup.sh
- enhance github-actions
- fix some tests


.github/workflows/tests.yml
```diff
@@ -87,76 +87,9 @@ jobs:
         run: |
           ./kernel-tests.sh --nix
 
-  tests:
```
patryk4815, line 90: Removed, since same systems we are running with docker_x86-64

.github/workflows/lock.yml
```diff
@@ -87,7 +87,7 @@ jobs:
         run: nix develop --accept-flake-config
 
   lock_flake:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
```
patryk4815, line 90: arm runner is faster than x86-64

setup.sh
```diff
@@ -22,18 +22,13 @@ osx() {
 
 install_apt() {
     sudo apt-get update || true
-    sudo apt-get install -y git gdb gdbserver python3-dev python3-venv python3-setuptools libglib2.0-dev libc6-dbg curl
-
-    if uname -m | grep -q x86_64; then
-        sudo dpkg --add-architecture i386 || true
```
patryk4815, line 28: installing i386 with `setup.sh` should be opt-out. But it is ok to install glibc.i386 with `setup-dev.sh`

uv.lock
```diff
@@ -1755,104 +1755,135 @@ wheels = [
...
+sdist = { url = "https://files.pythonhosted.org/packages/69/44/36f1a6e523abc58ae5f928898e4aca2e0ea509b5aa6f6f392a5d882be928/pydantic-2.12.5.tar.gz", hash = "sha256:4d351024c75c0f085a9febbb665ce8c0c6ec5d30e903bdb6394b7ede26aebb49", size = 821591, upload-time = "2025-11-26T15:11:46.471Z" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/6a/c0/ec2b1c8712ca690e5d61979dee872603e92b8a32f94cc1b72d53beab008a/pydantic-2.11.7-py3-none-any.whl", hash = "sha256:dde5df002701f6de26248661f6835bbe296a47bf73990135c7d07ce741b9623b", size = 444782, upload-time = "2025-06-14T08:33:14.905Z" },
+    { url = "https://files.pythonhosted.org/packages/5a/87/b70ad306ebb6f9b585f114d0ac2137d792b48be34d732d60e597c2f8465a/pydantic-2.12.5-py3-none-any.whl", hash = "sha256:e561593fccf61e8a20fc46dfc2dfe075b8be7d0188df33f221ad1f0139180f9d", size = 463580, upload-time = "2025-11-26T15:11:44.605Z" },
 ]
 
 [[package]]
 name = "pydantic-core"
-version = "2.33.2"
+version = "2.41.5"
```
patryk4815, line 1773: Upgraded `pydantic-core` because old version was not supported on py3.14

tests/library/gdb/tests/test_symbol.py
```diff
@@ -72,4 +72,4 @@ def test_symbol_duplicated_symbols_issue_1610():
 def _get_section_addr(sect):
     result = gdb.execute("maintenance info sections", to_string=True).split("\n")
     text_line = next(line for line in result if f": {sect} " in line)
-    return int(text_line.split(" at ")[1].split(":")[0], 16)
```
patryk4815, line 75: this test was wrong from beginning 😮

.github/workflows/tests.yml
```diff
@@ -87,76 +87,9 @@ jobs:
...
-          ./setup-dev.sh
-
-      - name: Run cross-architecture tests
-        run: |
-          ./tests.sh -g cross-arch-user -d gdb
-
   qemu-system-tests:
     runs-on: [ubuntu-22.04]
-    timeout-minutes: 30
+    timeout-minutes: 40
```
k4lizen, line 92: for sure we have a bug somewhere no?

k4lizen (approved):

k4lizen:
btw why don't we move all the docker tests into the `tests.yml` file?

patryk4815:
> btw why don't we move all the docker tests into the `tests.yml` file?

yes, I was looking how to remove `tests.yml` and replace with docker.yml xD

.github/workflows/tests.yml
patryk4815: not sure, they are now 1.5x times slower

PR #3489, opened 2025-12-12
Cleanup emulator register reads

OBarronCS:
This is code cleanup in the emulator object. Remove `__getattr__` for registers reads on the emulator object. Clean up a function used for converting register string name to Unicorn register enum.

### DIFF
--- a/pwndbg/emu/emulator.py
+++ b/pwndbg/emu/emulator.py
@@ -273,7 +273,7 @@ class Emulator:
             debug(DEBUG_INIT, "# Setting TLB mode to virtual")
             self.uc.ctl_set_tlb_mode(U.UC_TLB_VIRTUAL)  # type: ignore[attr-defined]
 
-        self.regs: pwndbg.lib.regs.RegisterSet = pwndbg.aglib.regs.current
+        self.reg_set: pwndbg.lib.regs.RegisterSet = pwndbg.aglib.regs.current
 
         # Whether the emulator is allowed to emulate instructions
         # There are cases when the emulator is incorrect or we want to disable it for certain instruction types,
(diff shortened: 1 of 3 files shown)


pwndbg/emu/emulator.py
```diff
@@ -717,35 +712,15 @@ def get_reg_enum(self, reg: str) -> int | None:
...
         if e is not None:
             return e
 
-        # If we're looking for an abstract register which *is* accounted for,
-        # we can also do an indirect lookup.
-        #
-        #   'pc' ==> 'eip' ==> enum
-        #
-        if hasattr(self.regs, reg):
-            return self.get_reg_enum(getattr(self.regs, reg))
```
OBarronCS, line 739: This logic was mostly broken, as the `hasattr` would return true for all sorts of values, like "common" or "gpr".

To my awareness, this was only used to handle the conversion of "pc" -> "rip", which this PR handles with a dedicated function to fetch pc.

pwndbg/emu/emulator.py
```diff
@@ -717,35 +712,15 @@ def get_reg_enum(self, reg: str) -> int | None:
...
-
-        # If we're looking for an abstract register which does not exist on
-        # the RegisterSet objects, we need to do an indirect lookup.
-        #
-        #   'sp' ==> 'stack' ==> 'esp' ==> enum
-        #
-        elif reg == "sp":
```
OBarronCS, line 746: This codepath was not used. If an arch needs to convert "sp" to real enum, it should be added to the `arch_to_reg_const_map` map defined in this file, if the conversion is not already automatic (which occurs when SP is the real architectural name, and there exists a `UC_*_REG_SP` enum)

k4lizen (approved):
very nice

PR #3490, opened 2025-12-12
Add more type hints to register printing logic

OBarronCS:
#2932 created a more generic way to express register printing logic.

This PR adds more typing information to this code for maintainability.

### DIFF
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -1010,13 +1010,13 @@ class RegisterContext(RegisterContextProtocol):
         return f"{prefix} {desc}"
 
 
-def get_regs(in_regs: List[str] = None):
+def get_regs(in_regs: List[str | VisitableRegister | None] | None = None):
+    if in_regs is None:
+        in_regs = []
+
     regs: List[str | VisitableRegister | None] = in_regs
     result: List[str] = []
     rc = RegisterContext()
-
-    if regs is None:
-        regs = []
 
     if len(regs) == 0:
         regs += pwndbg.aglib.regs.gpr
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -1011,9 +1011,10 @@ class RegisterContext(RegisterContextProtocol):
 
 
 def get_regs(in_regs: List[str | VisitableRegister | None] | None = None):
+    # Python default parameters are instantiated once and shared across calls.
+    # Instead of a default value of [], we need to do this check so we get a fresh list each time
     if in_regs is None:
         in_regs = []
-
     regs: List[str | VisitableRegister | None] = in_regs
     result: List[str] = []
     rc = RegisterContext()
--- a/pwndbg/commands/context.py
+++ b/pwndbg/commands/context.py
@@ -18,6 +18,7 @@ from typing import TypeVar
 
 import unicorn as U
 from typing_extensions import ParamSpec
+from typing_extensions import override
 
 import pwndbg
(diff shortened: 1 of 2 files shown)


pwndbg/commands/context.py
```diff
@@ -1046,11 +1053,12 @@ def get_regs(regs: List[str] = None):
     for reg in regs:
         if reg is None:
             continue
+        # If it's a VisitableRegister which has special logic to determine what to print
         if not isinstance(reg, str):
             desc = reg.context(rc)
             if desc is not None:
                 result.append(desc)
-                continue
```
OBarronCS, line 1053: Moving `continue` back fixes a crash when the flags register name cannot be looked up in GDB (it would fallthrough to the next lines of code, which should not be possible). This would occur when debugging `sparc`, when the `psr` register is unavailable.

pwndbg/lib/regs.py
```diff
@@ -31,27 +55,25 @@ class BitFlags:
 
     def __init__(self, flags: List[Tuple[str, Union[int, Tuple[int, int]]]] = [], value=None):
```
k4lizen: Image

pwndbg/lib/regs.py
k4lizen, line 154: Image

pwndbg/commands/context.py
```diff
@@ -1003,9 +1010,9 @@ def register_context_default(self, reg):
         return f"{prefix} {desc}"
 
 
-def get_regs(regs: List[str] = None):
-    regs: List[Any] = regs
-    result = []
+def get_regs(in_regs: List[str] = None):
+    regs: List[str | VisitableRegister | None] = in_regs
+    result: List[str] = []
```
k4lizen, line 1019: Image

k4lizen (changes requested):
nice, there are a few type errors still remaining

OBarronCS:
I fixed the issues mentioned in the comments

pwndbg/commands/context.py
```diff
@@ -1003,13 +1010,13 @@ def register_context_default(self, reg):
         return f"{prefix} {desc}"
 
 
-def get_regs(regs: List[str] = None):
-    regs: List[Any] = regs
-    result = []
-    rc = RegisterContext()
+def get_regs(in_regs: List[str | VisitableRegister | None] | None = None):
+    if in_regs is None:
+        in_regs = []
```
k4lizen, line 1017:
```suggestion
def get_regs(in_regs: List[str | VisitableRegister | None] = []):
```

Maybe this makes more sense?

pwndbg/commands/context.py
```diff
@@ -1003,13 +1010,13 @@ def register_context_default(self, reg):
...
```
k4lizen:
```suggestion
```
(i missed this line in the previous suggested change)

k4lizen (approved):

pwndbg/commands/context.py
OBarronCS: Good call, making this change

OBarronCS: Oh nevermind - since we add contents to this list, and Python default parameters are always the same instance of the list, this will break (future calls don't get a fresh empty list)

k4lizen: ahh you're right, my bad!

k4lizen: could maybe add a comment about this

PR #3492, opened 2025-12-13
Fix import shenanigans 1/2

k4lizen:
This PR renames the `pwndbg/dbg` module into `pwndbg/dbg_mod` and the `pwndbg/aglib/arch` module into `pwndbg/aglib/arch_mod`. 

People who wanted to use the stuff from the previously-named `pwndbg.dbg` submodule had to do
```python
import pwndbg.dbg_mod
```
anyway, so this will not change any call sites. There were some places that did `from pwndbg.dbg` and `import pwndbg.dbg` which are both non-sensical since `pwndbg.dbg` is exported as the object, but they worked due to python resolution magic.

Similarly, lots of places in the code did
```python
import pwndbg.aglib.arch
```
even though they actually wanted to use the `arch` object and not the `arch` module.

This renames the files themselves to match the way they should be imported, so everything is clear and there is no confusion.

I also plan to do this type of refactor for pwndbg.aglib.proc and pwndbg.aglib.regs but it is more involved since I need to add some caching logic first.


k4lizen:
Also, we should document that it is not safe to do
```python
from pwndbg.aglib import arch
```
because the arch object gets replaced when the correct architecture is detected.

disconnect3d:
> Also, we should document that it is not safe to do
> 
> ```python
> from pwndbg.aglib import arch
> ```
> 
> because the arch object gets replaced when the correct architecture is detected.

This is something we should detect on CI/CD and ban. Maybe with CodeQL or Semgrep, idk

PR #3497, opened 2025-12-14
Fix import shenanigans 2/2

k4lizen:
based on: https://github.com/pwndbg/pwndbg/pull/3492

+ Renames `aglib/arch.py` into `aglib/arch_mod.py`
+ Renames `aglib/regs.py` into `aglib/regs_mod.py`, now `regs` is a normal class
+ Takes the `class module` out of `aglib/proc.py`, now all functions are global in that file.
+ Refactors `lib/` files to take out any `aglib` mentions (there were 38). (this was a bit involved)
+ Adds documentation that writes down these design decisions.
+ Takes out unnecessary imports from files.
+ Deletes test_find_fastbin_size and the lib/heap/helpers thing.
+ Optimizes lib.cache to use integer flags instead of a dict


k4lizen:
~~hmm this is not throwing the "unhashable" error for some stuff where i would expect it to~~

is fine

k4lizen:
image

this is kinda scary hmmmmm

docs/contributing/common-pitfalls.md
```diff
@@ -0,0 +1,57 @@
...
+
+Doing it this way prevents "fun" surprises, makes the code more maintanable, makes the dependancy graph cleaner and as such prevents import cycles.
+
+#### `from pwndbg.aglib import arch` doesn't work
+
+If you look at `pwndbg/aglib/__init__.py` you will see that `arch` is a None-initialized object and gets swapped out at runtime depending on which architecture we are debugging. As such, to get the correct information about the current architecture you must always do `aglib.arch.whatever()`.
+
+#### `from pwndbg.aglib import regs` doesn't work
+
+If you look at `pwndbg/aglib/__init__.py` you will see that `regs` is a None-initialized object and gets swapped out during initialization. To use it properly you must do `aglib.regs.whatever()`.
```
disconnect3d:
```suggestion
If you look at `pwndbg/aglib/__init__.py` you will see that `regs` is a None-initialized object and gets swapped out during initialization. To use it properly you must not make a copy of it via `from pwndbg.aglib import regs` and instead always access it via `aglib.regs` like: `aglib.regs.whatever()`.
```

pwndbg/aglib/__init__.py
```diff
@@ -11,17 +11,36 @@
...
+    from pwndbg.aglib.arch_mod import PwndbgArchitecture
+    from pwndbg.aglib.regs_mod import RegisterManager
+
+    # These will be set during debugger setup.
+    # Thus you can't import them with `from pwndbg.aglib import arch`.
+    arch: PwndbgArchitecture = cast(PwndbgArchitecture, None)
+    regs: RegisterManager = cast(RegisterManager, None)
+else:
+    arch = None
+    regs = None
```
disconnect3d, line 34: I don't like this at all, but I understand we may just need it for now?

Can we fix this somehow in the future?

disconnect3d: Also: wouldn't it be better to default to the default debugger architecture when no file is loaded?

Like in GDB:
```
$ gdb --nx --ex 'show architecture'
The target architecture is set to "auto" (currently "i386").
```

pwndbg/aglib/arch_mod.py
```diff
@@ -418,7 +418,8 @@ def update() -> None:
     pwnlib.context.context.endian = a.endian
     pwnlib.context.context.os = PWNLIB_PLATFORM_MAPPINGS.get(a.platform, "linux")
 
-    if a.name != pwndbg.aglib.arch.name:
+    # Will be None the first time around.
+    if pwndbg.aglib.arch is None or a.name != pwndbg.aglib.arch.name:
```
disconnect3d, line 423: Please see https://github.com/pwndbg/pwndbg/pull/3497/changes#r2634812380

pwndbg/aglib/regs_mod.py
```diff
@@ -117,16 +91,18 @@ def read_reg(self, reg: str, frame: pwndbg.dbg_mod.Frame | None = None) -> int |
...
                 return None
             value = int(value)
             if reg == "pc" and pwndbg.aglib.arch.name == "i8086":
-                if self.cs is None:
-                    return None
-                value += self.cs * 16
+                # FIXME: What is going on here?
+                # if self.cs is None:
+                #     return None
+                # value += self.cs * 16
```
disconnect3d: Why do we change the logic here? I believe this was needed for something :)

pwndbg/lib/memory.py
```diff
@@ -30,11 +30,11 @@ def round_up(address: int, align: int) -> int:
 
 
 def format_address(
-    vaddr: int, memsz: int, permstr: str, offset: int, objfile: str | None = None
+    vaddr: int, memsz: int, permstr: str, offset: int, ptrsize: int, objfile: str | None = None
 ) -> str:
     "Format the given address as a string."
 
-    width = 2 + 2 * pwndbg.aglib.arch.ptrsize
+    width = 2 + 2 * ptrsize
```
disconnect3d, line 37: Can we really have two pages at the same time with different ptrsize?

Note that we add ptrsize to Page meaning we pass it around lots of times and e.g. if we have 100 pages we now have +100 python integers around

disconnect3d (changes requested):
1. Not sure we want to add `ptrsize` to `Page` object, especially in this PR
2. This change also seems to not belog to this PR - it also most likely breaks something:
```py
            if reg == "pc" and pwndbg.aglib.arch.name == "i8086":
                # FIXME: What is going on here?
                # if self.cs is None:
                #     return None
                # value += self.cs * 16
                # pass
```

k4lizen:
> This change also seems to not belog to this PR - it also most likely breaks something:

the "what is going on here" was referring to the fact that `regs` doesn't have a `.cs` property, so it would throw an exception at runtime and so i'm not breaking anything by commenting it out

image

i'll change it to this

pwndbg/lib/memory.py
k4lizen: i agree, but how do you propose to resolve the underlying issue of this file calling into aglib?

pwndbg/aglib/__init__.py
k4lizen: > Also: wouldn't it be better to default to the default debugger architecture when no file is loaded?

the reason i didn't, is because i'm afraid it will cause subtle bugs when
```
from pwndbg.aglib import arch
```
whereas this way, it will just be None and throw an exception.

but I see what you're saying and I will rethink it.

pwndbg/aglib/regs_mod.py
```diff
@@ -31,79 +27,58 @@
...
         try:
-            value = get_register(reg, frame)
+            value = self.get_register(reg, frame)
             if value is None and reg.lower() == "xpsr":
-                value = get_register("xPSR", frame)
+                value = self.get_register("xPSR", frame)
             if value is None:
                 return None
             value = int(value)
             if reg == "pc" and pwndbg.aglib.arch.name == "i8086":
```
OBarronCS: This "pc" should also be changed to "eip", as that's the register GDB uses in this case.

k4lizen: done

pwndbg/aglib/__init__.py
k4lizen: okay so I'm not sure what architecture-specific stuff we can even do before an inferior is running, so I'm not sure if we have a use-case for a default architecture (and as I said, having it as None is more likely to catch bugs)

I agree that if we decide that we do want to set some default architecture here, that we likely want either:

1. The default architecture that the underlying debugger selects, or
2. The architecture the host is on

But in either of those cases, I feel this is a bit out of scope for this PR. So I would prefer to leave it as `= None`, but if you want I can put it back to `= get_pwndbg_architecture("i386")`.

k4lizen:
@disconnect3d 
Regarding `ptrsize` to `Page` I don't think the performance is really relevant here but I agree that it is ugly that we are passing a seemingly unrelated field to the constructor of that class (although in practice, everything that uses a Page is in fact allowed to access aglib.arch.ptrsize). There are a few options:

1. Move the whole file to aglib - this is pretty bad, lets not do this
2. Move just the page class to aglib - maybe viable, still feels bad just for one field.
3. Just use 64 bits - I don't think it would be erroneous, but I think it would be noticable for users, and I would rather not make a user-facing sacrifice for a not-that-important implementation detail. 
4. Pass the ptrsize just to the formatting function. I think this is the most reasonable alternative to the current solution, but it is some effort to do. We would have to take `__str__` out of the class and replace it with a custom `string()` function or whatever and change every call-site. It's particularly bad to work on because my editor does not show me `str(pageobject)` references when I ask for references on `__str__`, so I would rather not do it for this PR.

I will note that the problem of a lib file accessing aglib (which I solve here by passing ptrsize) is not a theoretical one, I have gotten import cycles due to this during this refactor.

disconnect3d (approved):

PR #3230, opened 2025-08-12
feat(heap): Implemented collapsible output for vis_heap_chunks

sharmaut:
This PR implements collapsible output for the vis_heap_chunks command to make the output cleaner and easier to read for large chunks with repetitive data. The core logic for the feature is completed. 
Regarding testing & linting, I was unable to get the local development environment running on my Macbook due to persistent nix: command not found errors, even after following the official setup documentation, common troubleshooting steps and reaching out on your discord server as well. I couldn't run the linter locally but I am submitting this PR with the expectation that the CI checks will fail on the linting step. I am happy to fix any and all formatting issues that the CI pipeline reports.

PR #3504, opened 2025-12-15
Add vis-skip-repeating-val config enabled by default

disconnect3d:
1. Adds a `vis-skip-repeating-val` config which is enabled by default.
2. Adds a `--no-skip` flag to `vis` command that disables the skipping of repeating values
3. Fixes a bug in `vis` where an empty line was printed at start
4. Changes `_` to `-` in other options/flags of `vis` command

This config changes the `vis` command rendering so that it collapses a heap chunk display like this, instead of repeating its lines:
```
  0x555555559000    0x0000000000000000    0x0000000000000291    ................
  0x555555559010    0x0000000000000000    0x0000000000000000    ................
      ... ↓      39 repeated lines skipped
  0x555555559290    0x0000000000000000    0x0000000000000111    ................
```

See screenshot in the post below (it includes the empty line which was removed, but it shows the skipping of repeated values).


disconnect3d:
This overrides https://github.com/pwndbg/pwndbg/pull/3230

disconnect3d:
Here is how it looks now:
image

With skip disabled:
image

k4lizen:
i haven't studied the code but just a few mentions:

1. Should the `-n` flag disable this? I think its probably fine not to
2. Regardless of chunk boundaries, if you encounter an address which is in a freelist or top you should not skip this line
3. There seems to be a bug in `vis` currently where the first line of output is empty, it would be nice if you could drive-by fix this

disconnect3d:
1. I added a `--no-skip | -s` flag for this
2. I believe the `if not lables` check covers it
3. I fixed it

I also changed the flags to use `-` instead of `_`.

PR #3507, opened 2025-12-16
Add --page filter to spray command for targeted pointer spraying

Eshaan-byte:
Adds a `--page` filter to the `spray` command that allows filtering pointers to only spray those pointing to a specific vmmap page.

## Changes
- Added `--page` argument that accepts both page names (e.g., `[heap]`, `[stack]`) and addresses (hex or decimal)
- The filter works independently or in combination with the existing `--only-funcptrs` flag
- Provides contextual feedback messages based on which filters are active

## Usage Examples
```bash
# Spray only pointers pointing to the heap
spray 0x555555558000 --page=[heap]

# Spray only pointers pointing to a specific address/page
spray 0x555555558000 --page=0x7ffff7dd5000

# Combine with --only-funcptrs to spray only function pointers pointing to heap
spray 0x555555558000 --only-funcptrs --page=[heap]

### CODE REVIEW

File: pwndbg/commands/spray.py [hunk 1]
```diff
@@ -29,22 +29,53 @@
...
+
+    # Resolve target page for filtering if --page is specified
+    target_page = None
+    if page:
+        # Try to parse as address first (hex or decimal)
+        try:
+            if page.startswith("0x"):
+                page_addr = int(page, 16)
+            else:
+                page_addr = int(page)
```
disconnect3d: `page_addr = int(page, 16) if page.startswith("0x") else int(page)`

pwndbg/commands/spray.py
```diff
@@ -29,22 +29,53 @@
...
+            target_page = pwndbg.aglib.vmmap.find(page_addr)
```
disconnect3d, line 56: this shouldnt be in try/except.

pwndbg/commands/spray.py
```diff
@@ -29,22 +29,53 @@
...
+            if target_page is None:
+                print(M.error(f"Invalid address {page}: can't find vmmap containing it"))
+                return
+        except ValueError:
+            # Not an address, treat as page name (e.g., [heap], [stack])
+            # Find the page by name
+            for p in pwndbg.aglib.vmmap.get():
+                if p.objfile == page:
+                    target_page = p
```
disconnect3d: This may allow for multiple pages. e.g. "libc" will account for multiple libc pages. We probably want to have some additional filtering here for page[idx] or some pages permissions filter ehh

disconnect3d (changes requested):
See comments

disconnect3d:
Btw for page filter, u probably want partial match, e.g. `-p=heap` should match `[heap]` page ojbfile.

Eshaan-byte:
Hi @disconnect3d! I've addressed all the review feedback:

The implementation now supports partial matching while handling the case where multiple pages match (e.g., "libc" matching multiple libc pages). When multiple matches are found, it displays all options with their address ranges and prompts the user to use a more specific page name or an address.

pwndbg/commands/spray.py
```diff
@@ -29,22 +29,64 @@
...
+                print(M.error(f"Invalid page name '{page}': can't find matching vmmap page"))
+                return
+            elif len(matching_pages) == 1:
+                target_page = matching_pages[0]
+            else:
+                # Multiple matches found, print them for user to choose
+                print(M.error(f"Multiple pages match '{page}':"))
+                for idx, p in enumerate(matching_pages):
+                    print(f"  [{idx}] {p.objfile} ({hex(p.start)}-{hex(p.end)})")
+                print(M.notice("Please specify the exact page name or use an address"))
```
disconnect3d: While I think it makes sense for the user to specify this, what if the user wants all of the pages with such a name instead of filtering an exact specific one?

disconnect3d:
@Eshaan-byte Please see my comment/question. We probably want to handle that case as well. We also need to fix conflicts from recent changes on the dev branch

Eshaan-byte:
@disconnect3d Good point! I've updated the implementation to handle the case where users want all matching pages. New behavior:
When multiple pages match (eg, --page=libc matching multiple libc pages), it now sprays pointers to any of those matching pages instead of erroring out
Displays a notice: "Found N pages matching 'X', will spray pointers to any of them"

Output messages adapt based on whether filtering a single page or multiple pages

Also resolved the merge conflicts with the latest dev branch. Changes are in the latest commit d01e26d58.

pwndbg/commands/spray.py
```diff
@@ -66,18 +108,68 @@ def spray(addr, length, value, only_funcptrs) -> None:
...
-                page = pwndbg.aglib.vmmap.find(ptr_candidate)
-                if page is not None and page.execute:
+                ptr_page = pwndbg.aglib.vmmap.find(ptr_candidate)
+
+                # Apply filters
+                should_spray = True
+
+                # Check if pointer points to executable page (if --only-funcptrs is set)
+                if only_funcptrs:
+                    should_spray = should_spray and (ptr_page is not None and ptr_page.execute)
```
disconnect3d, line 125: Why the `should_spray and` part if `should_psray` is always set to True before that?

disconnect3d: Please read the code yourself before sending it from an AI :(

disconnect3d:
@Eshaan-byte ping?

disconnect3d:
@Eshaan-byte are u going to update this or should we close it

disconnect3d:
Closing due to inactivity

Issue #2183, opened 2024-05-24
Dynamically scale context code lines based terminal height 


fidgetingbits:
At the moment context output can be manually configured to show a certain number of code lines. 

Output could be enhanced by adding an option to scale code lines dynamically given the height of the terminal. This should be still be optional behavior however.

This is a screenshot shared by a user on discord demonstrating the problem:
![ss.png](https://github.com/pwndbg/pwndbg/assets/13679876/6b021aa1-19a3-4ed1-b5d6-a14af808a9fb)

disconnect3d:
This should be fairly easy to implement and I think is a nice feature.

peace-maker:
I think it's worth it to grow other sections as well instead of only the disasm view? Maybe show more of the stack and sourcecode if debug symbols are available? That would require some weight or ratio though to control which section gets how much of the additional available vertical space. This feels like replicating the TUI `tui new-layout` layout code though.

Do you think other sections should grow too or just let the disasm section have all the space for now?

disconnect3d:
Tbh I'm not sure; I think it makes sense to play with it and see how it works if we rescale all the sections.

Also not sure if we want this since we have layout tui now, but I also havent tested the TUI myself :D

peace-maker:
You have to want to use the TUI since it's not perfect, so this is still nice to have in my opinion.

I think this only makes sense for sections that could show additional information or are truncated, but not for all like heap_tracker or regs which show a fixed size.

PR #3511, opened 2025-12-17
Fill full height of context section in TUI mode

peace-maker:
Add new `context-tui-adjust-height` config option to dynamically increase the number of lines printed in `regs`, `disasm`, `code`, and `threads` sections to fill the whole TUI window allocated for the section.

This ignores the other config values in place to limit the number of lines for each individual section and interacts weirdly with the context history, because the same entry could be added twice with longer section output when switching from CLI to TUI mode and the context was already output in CLI mode. That is a future TODO to replace the section data in the latest history entry when applicable.

grafik

grafik

This is part of #2183 but streching the disasm window in the CLI context output to use the full height isn't implemented yet.


pwndbg/gdblib/tui/context.py
```diff
@@ -15,6 +15,17 @@
 from pwndbg.commands.context import resetcontextoutput
 from pwndbg.gdblib import gdb_version
 
+tui_adjust_height = pwndbg.config.add_param(
+    "context-tui-adjust-height",
+    True,
+    "adjust height of context sections to fit the TUI",
+    help_docstring="""
+    Adjust the height of context sections to fit the TUI window, ignoring section specific limits.
```
k4lizen: other help_docstring's in the code look like this
```
    help_docstring="""
The bins are traversed both forwards and backwards.
""",
```
is the indentation here fine?

pwndbg/commands/context.py
```diff
@@ -319,10 +319,17 @@ def output(section: str):
...
+    type=int,
+    nargs="?",
+    default=None,
+    help="Sets a fixed height (used for number of lines to display before cutoff). Set to None for auto",
+)
 
 
 @pwndbg.commands.Command(parser, aliases=["ctx-out"], category=CommandCategory.CONTEXT)
-def contextoutput(section, path, clearing, banner="both", width: int = None):
+def contextoutput(section, path, clearing, banner="both", width: int = None, height: int = None):
```
k4lizen: `height: int | None = None`

pwndbg/commands/context.py
```diff
@@ -319,10 +319,17 @@ def output(section: str):
     default=None,
     help="Sets a fixed width (used for banner). Set to None for auto",
 )
+parser.add_argument(
+    "height",
+    type=int,
+    nargs="?",
+    default=None,
+    help="Sets a fixed height (used for number of lines to display before cutoff). Set to None for auto",
```
k4lizen: maybe  extend the comment a bit to describe that it ignores specific limits, extends some contexts while not others etc

pwndbg/commands/context.py
```diff
@@ -889,7 +898,7 @@ def compact_regs(regs: List[str], width=None, target=sys.stdout) -> List[str]:
 
 
 @serve_context_history
-def context_regs(target=sys.stdout, with_banner=True, width=None):
+def context_regs(target=sys.stdout, with_banner=True, width=None, height=None):
```
k4lizen: would be nice to add a type hint to the other places too

k4lizen (changes requested):
plz check the `help_docstring` thing, otherwise lgtm

pwndbg/gdblib/tui/context.py
peace-maker: Down the line the `help_docstring` is stripped, so there shouldn't be a difference.

https://github.com/pwndbg/pwndbg/blob/dfdfbd4f7bd45869d7b4b7d3e4418aede29359d2/pwndbg/lib/config.py#L241

k4lizen: strip only deletes whitespace from the end and start of the string, you have indentation effectively in the middle of the string

k4lizen: image

peace-maker: Ah, right!

disconnect3d:
docs on CI:
```
VERIFICATION FAILED. The files differ from what would be auto-generated.
Error: File docs/commands/context/contextoutput.md is too short. Expected 30 lines, got 29.
Please run ./scripts/generate-docs.sh from project root and commit the changes.
Error: Process completed with exit code 2.
```

peace-maker:
I believe the failing tests are unrelated to the changes in this branch.

k4lizen (approved):
they are.

thanks, and for the drive-by type hint fixes as well :p

PR #3514, opened 2025-12-18
Fix loong64 uv installation and make few tests optional

patryk4815:
## Makefile Updates
- Added detection for `jemalloc` via `pkg-config` .  
- Conditional compilation of targets:
  - `heap_jemalloc_*` targets are built only if `jemalloc` is available.  
  - `heap_musl_*` targets are built only if a `musl-gcc` compiler is present.  

## Python Project (`pyproject.toml`) Updates

- Restricted `ziglang` dependency to non-LoongArch64 Linux platforms. Previously, `ziglang` was unavailable on LoongArch64, preventing `uv sync` from running; now it works correctly.  
- Added custom `uv` sources pointing to a mirror (`https://mirrors.loong64.com/pypi/simple`) for two packages (`uv` and `ruff`) to ensure successful installation on LoongArch64.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ dependencies = [
     "pt @ git+https://github.com/martinradev/gdb-pt-dump@50227bda0b6332e94027f811a15879588de6d5cb",
     # Optional? 'ziglang' must be optional, as it is not available on every platform.
     # It is only used for the 'cymbol' command and some `asm` functionality.
-    "ziglang==0.14.1",
+    "ziglang==0.14.1 ; (platform_system == 'Linux' and platform_machine != 'loongarch64') or platform_system != 'Linux'",
 ]
 
 [project.optional-dependencies]
@@ -70,7 +70,7 @@ dev = [
 lint = [
     "mypy>=1.18.2,<2",
     "isort>=5.13.2,<6",
-    "ruff>=0.4.1,<0.5",
+    "ruff>=0.4.1,<0.15",
     "vermin>=1.6.0,<2",
 ]
 tests = [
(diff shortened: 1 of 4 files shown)


k4lizen:
is it possible to also upload lldb-for-pwndbg and gdb-for-pwndbg to https://mirrors.loong64.com/pypi/simple ? it would speed up `uv run` times significantly since those wheels could be cached

tests/binaries/host/makefile
```diff
@@ -14,6 +14,7 @@ else ifeq ($(HOST_ARCH),i686)
 endif
 
 ZIGCC           = uv run python3 -m ziglang cc
+PKG_CONFIG      ?=     pkg-config
```
k4lizen, line 17: hmm a system dependancy. but I guess it is already included in everything we install in `setup-dev.sh`?

pyproject.toml
```diff
@@ -30,7 +30,7 @@ dependencies = [
     "pt @ git+https://github.com/martinradev/gdb-pt-dump@50227bda0b6332e94027f811a15879588de6d5cb",
     # Optional? 'ziglang' must be optional, as it is not available on every platform.
     # It is only used for the 'cymbol' command and some `asm` functionality.
-    "ziglang==0.14.1",
+    "ziglang==0.14.1 ; (platform_system == 'Linux' and platform_machine != 'loongarch64') or platform_system != 'Linux'",
```
k4lizen, line 33: why not just `platform_machine != 'loongarch64'`?

k4lizen (approved):

tests/binaries/host/makefile
patryk4815: yes, it is included in setup-dev.sh

disconnect3d (commented):
Why are we making jemalloc or musl targets conditional?

Is it that they dont exist on certain targets (loong64?)?

I am a bit worried that making the targets compilation or/and their tests conditional we may end up with those tests never executing at some point and we not noticing it?

Or am I wrong and that would be somehow detected?

(Also I am not 100% sure if the corresponding tests are conditional or not, but I assume that is yes since we do the changes here)

patryk4815:
> we may end up with those tests never executing at some point and we not noticing it?

maybe we should add a test that will crash randomly to make people look at CI?

PR #3519, opened 2025-12-21
Clean up Dockerfile comments

k4lizen:
There are three changes:
1. Get rid of the "For testing use:" line
   + It personally confused me. I was trying to test some failing test and later realized I actually want the "For development" line. I don't really know of a use-case for the "testing" command, so I think its better to remove it to reduce confusion.
2. Changed the tags to be different
   + Not too uncommonly I want to debug a test on different distributions. If I run the commands as-is the image will be rebuilt which takes ten thousand years.
3. Use `$(pwd)` instead of backticks. The former works on bash zsh and fish, while the latter doesn't work on fish which means I have to fix-up the command every time which is annoying.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,10 @@
 # This dockerfile was created for development & testing purposes, for APT-based distro.
 #
-# Build as:             docker build -t pwndbg .
+# Build as:
+#   docker build -f Dockerfile -t pwndbg .
 #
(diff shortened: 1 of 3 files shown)


disconnect3d (approved):

disconnect3d:
I think its fine. Users should probably understand what it means when we do -v ? :f

PR #3521, opened 2025-12-21
Fix import shenanigans 3/2

k4lizen:
Based on: https://github.com/pwndbg/pwndbg/pull/3497

Pulls out some commits from there into this seperate PR.

This is mainly about adding/removing imports and the top of files. And renaming `as` stuff to be more consistent.


pwndbg/aglib/regs_mod.py
```diff
@@ -31,79 +27,57 @@
...
-items: Callable[[], Generator[Tuple[str, Any], None, None]]
-previous: Dict[str, int]
-last: Dict[str, int]
-pc: int | None
-
-
-class module(ModuleType):
-    previous: Dict[str, int] = {}
-    last: Dict[str, int] = {}
+        return None
```
disconnect3d, line 65: drop else to just return base here?

PR #3525, opened 2025-12-22
bump gdb 17.1

patryk4815:
before:
image

after:
image


k4lizen:
we need to fix the coloring issue that bata mentioned first, notice that in after the `pwndbg` is white

patryk4815:
On all x86-64 distros same tests fail: (but why not on fedora41?)
```
- tests/library/gdb/tests/test_commands_elf.py::test_command_got_for_target_binary_and_loaded_library[x86-64]
```


Turns out `info sharedlibrary` returns different values on gdb 17 then on gdb 16.
Patch - https://sourceware.org/pipermail/gdb-patches/2025-March/216381.html

gdb16:
```
In [12]: gdb.execute("info sharedlibrary")
From                To                  Syms Read   Shared Object Library
0x00007ffff7fc5090  0x00007ffff7fee315  Yes         /lib64/ld-linux-x86-64.so.2
0x00007ffff7c28700  0x00007ffff7dba9bd  Yes         /lib/x86_64-linux-gnu/libc.so.6
```

gdb17:
```
In [1]: gdb.execute("info sharedlibrary")
From                To                  Syms Read   Shared Object Library
0x00007ffff7fc3000  0x00007ffff7fff000  Yes         /lib64/ld-linux-x86-64.so.2
0x00007ffff7c00000  0x00007ffff7e29000  Yes         /lib/x86_64-linux-gnu/libc.so.6
```

disconnect3d:
@patryk4815 what about the coloring issue? can we disable the GDB color somehow?

patryk4815:
> we need to fix the coloring issue that bata mentioned first, notice that in after the `pwndbg` is white

but do we want do workaround for gdb bugs?


image

k4lizen:
maybe its not a bug and there is a simple way to fix it? we should ask them (i wanted to but forgot)!

patryk4815:
I don't see easy fix, they just mess-up ASCII colors

k4lizen:
okey then i dont think it should be a blocker for this PR

(i just wish we communicated with upstream more ;;)

patryk4815:
> okey then i dont think i should be a blocker for this PR

I can try patch-out for our gdb version, and recompile

k4lizen:
that would be pretty cool!

but plz write the bug report, i think its likely that it will be fixed quickly since its a regression and they probably know from which commit too

PR #3526, opened 2025-12-22
Use loong64 pypi for gdb/lldb

patryk4815:
@k4lizen said that `uv run` is very slow, so now using `loong64.pypi` it should use the "cache" and be faster?

Ofc course, if we use 3rd party `pypi` we have to be careful about supply chain attack, so we need to verify the hash wheel to see if it matches what we published on github

Before changes:
image

After changes:
image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,24 +38,10 @@ lldb = [
     # The LLDB REPL requires readline.
     'gnureadline>=8.2.13,<9; sys_platform != "win32"',
     'pyreadline3>=3.5.4,<4; sys_platform == "win32"',
-    "lldb-for-pwndbg==21.1.6.post1 ; (platform_system == 'Linux' and platform_machine != 'loongarch64') or platform_system != 'Linux'",
-
-    # pypi.org don't support loongarch64 yet. https://discuss.python.org/t/pypi-supports-loongarchs-wheel-what-should-we-do/26253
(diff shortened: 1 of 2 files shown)


disconnect3d:
@patryk4815 can we do the wheel hash check automagically in the PR? its fine to hardcode it too.

k4lizen:
It is better to benchmark with `uv run sleep 1`, and run it a couple of times on both branches. dev:
```
~> time uv run sleep 1
~> time uv run sleep 1
~> time uv run sleep 1
________________________________________________________
Executed in    1.76 secs      fish           external
   usr time   17.29 millis  283.00 micros   17.00 millis
   sys time   13.70 millis  744.00 micros   12.96 millis
```
This PR:
```
~> time uv run sleep 1
~> time uv run sleep 1
~❯ time uv run sleep 1

________________________________________________________
Executed in    1.02 secs      fish           external
   usr time    9.24 millis    0.00 micros    9.24 millis
   sys time   10.76 millis  708.00 micros   10.05 mill
```
so this PR speeds it up from 0.76 seconds to 0.02 seconds. But I will say, I have seen dev be as slow as 2-3 seconds rather than just 0.76 (github is really wonky these days), so this is really awesome.

Here is the reasoning btw:
image

k4lizen (approved):

PR #3532, opened 2025-12-23
Adding last_signal to config_context_sections

Acee11:

This PR makes last_signal section to be displayed by default in context. Previously, you had to add something like
`set context-sections regs last_signal disasm code ghidra stack backtrace expressions threads heap_tracker`
to .gdbinit.
It's a followup to https://github.com/pwndbg/pwndbg/pull/3531
Before:

before

After:

after

### DIFF
--- a/docs/configuration/config.md
+++ b/docs/configuration/config.md
@@ -360,7 +360,7 @@ Which context sections are displayed (controls order).
 
 
 
-**Default:** 'regs disasm code ghidra stack backtrace expressions threads heap_tracker'  
+**Default:** 'last_signal regs disasm code ghidra stack backtrace expressions threads heap_tracker'  
 
 ----------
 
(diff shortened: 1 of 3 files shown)


k4lizen:
This seems reasonable but since I don't use this context I'm not that familiar with how it works. So do you know:
1. Will the signal be displayed even if the debugee catches it (i.e. registers a handler)?
2. What if its a "normal" signal like e.g. SIGINT or whatever
3. Does GDB always stop/break at a signal? If it doesn't, is this section inaccurate?
4. If an e.g. SIGINT is caught and processed, will this context section keep displaying it even long after?

If the answers are: 1. yes 2. yes 3. (?, no) 4. no. Then I think it is reasonable to add this to the default, otherwise it may require some more thought.

btw, my terminal is often kinda zoomed in, so I would maybe prefer it if this section was at the bottom to make sure it isn't cropped out of view since it is kinda important

disconnect3d:
This now has some conflicts, please solve them :(

PR #3535, opened 2025-12-24
Rework old and implement new compact register displays

k4lizen:

Changes:
+ Changed register ordering in the compact register context so it makes more sense
+ Added two new compact register modes (see screenshots, code comments and parameter description)
+ Change the default compact register mode to one of the two new ones

## Off

image

## Before
`set show-compact-regs-columns 3` and `set show-compact-regs on`:
image

(not the exact same register state as "After" since I had to restart pwndbg to switch branches)

## After
`set show-compact-regs-columns 3` and `set show-compact-regs on`:
image
`set show-compact-regs very`
image
`set show-compact-regs hardcut`
image

### DIFF
--- a/docs/configuration/config.md
+++ b/docs/configuration/config.md
@@ -908,9 +908,17 @@ Whether glibc uses safe-linking.
 
 Whether to show a compact register view with columns.
 
+Values explained:
 
++ `off` - Disable compact registers (default). Every other option tries to make the register context use less rows by putting the registers into multiple columns.
++ `on` - If a register printout doesn't fit it will be added to the end of the register context.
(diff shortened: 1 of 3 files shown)


k4lizen:
I thought I opened an issue a long time ago about changing the ordering, but I can't find it now :x

k4lizen:
open to suggestions on the styling, unicode characters, the default compact or w/e

disconnect3d:
It seems to me like the 'hardcut' is the most reasonable option, but idk, I'm fine with all that. Having more options doesn't hurt and may help people? Hope they will indeed use it :P

PR #3542, opened 2025-12-26
Rewrite `info files` to `maintenance info target-sections`

### LINKED ISSUE #3396: onegadget doesn't work, module_section_locations is borked, info files doesn't do what we expect?

patryk4815:
## Rewrite info files to maintenance info target-sections
Fixes:
- https://github.com/pwndbg/pwndbg/issues/3396

### DIFF
--- a/pwndbg/dbg_mod/gdb/__init__.py
+++ b/pwndbg/dbg_mod/gdb/__init__.py
@@ -1045,55 +1045,9 @@ class GDBProcess(pwndbg.dbg_mod.Process):
     def module_section_locations(self) -> List[Tuple[int, int, str, str]]:
         import pwndbg.gdblib.info
 
-        # Example:
-        #
-        # 0x0000555555572f70 - 0x0000555555572f78 is .init_array
-        # 0x0000555555572f78 - 0x0000555555572f80 is .fini_array
-        # 0x0000555555572f80 - 0x0000555555573a78 is .data.rel.ro
-        # 0x0000555555573a78 - 0x0000555555573c68 is .dynamic
-        # 0x0000555555573c68 - 0x0000555555573ff8 is .got
-        # 0x0000555555574000 - 0x0000555555574278 is .data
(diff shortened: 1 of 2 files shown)


pwndbg/dbg_mod/gdb/__init__.py
```diff
@@ -1045,55 +1045,9 @@ def disasm(self, address: int) -> pwndbg.dbg_mod.DisassembledInstruction | None:
...
-            if len(div1) == 2:
-                module = div1[1].strip()
-            else:
-                module = main
-
-            section = div1[0].strip()
-
-            result.append((beg, end - beg, section, module))
+        for section in pwndbg.gdblib.info.sections():
+            result.append((section.start, section.size, section.section, section.objfile))
```
disconnect3d, line 1050: Can't we just:

```py
return [(s.start, s.size, s.section, s.objfile) for s in pwndbg.gdblib.info.sections()]
```

Or even `return tuple(...)` if we can return a tuple :p

disconnect3d: Shall this function return a `Section` objects instead of a tuples?

PR #3543, opened 2025-12-26
Fix `gdb.execute(.., to_string=True)` in `ipi` context

patryk4815:
##  Fix `gdb.execute(.., to_string=True)` in `ipi` context:

Before:
image

After:
image


disconnect3d:
Awesome. May want to add a test for it, tho a test is probably hard to write for it...

PR #3554, opened 2025-12-30
Allow using gdbinit without venv

Xeonacid:

#3119 allows using `pwndbg` command from shell with a distro-packaged pwndbg. But using `source /usr/share/pwndbg/gdbinit.py` from `.gdbinit` still fails when no venv is created.

Using the same logic as `verify_venv` in `pwndbginit.common`:
https://github.com/pwndbg/pwndbg/blob/dca241bf82d70392e4ceb783d1af9cef20081cc2/pwndbginit/common.py#L123-L126

cc @patryk4815

### DIFF
--- a/gdbinit.py
+++ b/gdbinit.py
@@ -45,16 +45,31 @@ def get_venv_path(src_root: Path):
 
 def main() -> None:
     src_root = Path(__file__).parent.resolve()
-    venv_path = get_venv_path(src_root)
-    if not venv_path.exists():
-        print(
-            f"\nCannot find Pwndbg virtualenv directory: {venv_path}. Please (re-)run setup.sh from the Pwndbg source folder.\n"
-            "(see https://pwndbg.re/dev/setup/#installing-from-source)",
-            flush=True,
-        )
-        os._exit(1)
 
-    fixup_paths(src_root, venv_path)
+    skip_venv = False
+    try:
+        # This can fail if Pwndbg is not installed to system site-packages
+        # on our regular virtualenv setup
+        from pwndbginit.common import is_system_installation
+
+        # If Pwndbg is installed by distro package manager, skip virtualenv check
+        if is_system_installation(src_root):
+            skip_venv = True
+    except ImportError:
+        pass
+
+    if not skip_venv:
+        venv_path = get_venv_path(src_root)
+        if not venv_path.exists():
+            print(
+                f"\nCannot find Pwndbg virtualenv directory: {venv_path}. Please (re-)run setup.sh from the Pwndbg source folder.\n"
+                "(see https://pwndbg.re/dev/setup/#installing-from-source)",
+                flush=True,
+            )
+            os._exit(1)
+
+        fixup_paths(src_root, venv_path)
+
     from pwndbginit.gdbinit import main_try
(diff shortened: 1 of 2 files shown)


patryk4815:
`echo "py import pwndbginit.gdbinit; pwndbginit.gdbinit.main_try()" >> ~/.gdbinit`
why not?

I disagree to support hardcoded paths in `.gdbinit`

Xeonacid:
> `echo "py import pwndbginit.gdbinit; pwndbginit.gdbinit.main_try()" >> ~/.gdbinit` why not?

This is definitely not a good experience for normal users, right?

> I disagree to support hardcoded paths in `.gdbinit`

May I know what's your concern? For a pwndbg environment setup from source, `.pwndbg_root` will exist and enforce a venv. Only for distro-packaged pwndbg, there will be no `.pwndbg_root`, thus allows running without venv.

User should be able to choose which to use, `pwndbg` command or gdbinit. Allowing skipping venv does not break anything IMO.

patryk4815:
> This is definitely not a good experience for normal users, right?

Yes! manual modification `~/.gdbinit` is really bad user experience. That's why we introduced `pwndbg` entrypoint.

If we want support `source /usr/xx/xx/gdbinit.py` - This option should be good documented and only for advance users.
I dislike hardcoding paths in .gdbinit

gdbinit.py
```diff
@@ -45,16 +45,27 @@ def get_venv_path(src_root: Path):
...
-            "(see https://pwndbg.re/dev/setup/#installing-from-source)",
-            flush=True,
-        )
-        os._exit(1)
-
-    fixup_paths(src_root, venv_path)
+
+    skip_venv = False
+    try:
+        from pwndbginit.common import skip_autoupdate
```
patryk4815: please create new func eg: `is_system_installation` or similar, don't use `skip_autoupdate`

patryk4815: `+` add comment why we skipping `venv` when `is_system_installation is True`

Xeonacid: > please create new func eg: `is_system_installation` or similar, don't use `skip_autoupdate`

~~The new function will basically have duplicated code with `skip_autoupdate`. Is this fine?~~

Not really. `skip_autoupdate` will respect `PWNDBG_NO_AUTOUPDATE` but `is_system_installation` will not.

Xeonacid:
> If we want support source /usr/xx/xx/gdbinit.py - This option should be good documented and only for advance users.
I dislike hardcoding paths in .gdbinit

Sounds like distros like Arch should add a `post_install`, which will show a notice for users.

Arch expects users to read all notice during installing or upgrading. There are many examples like [emscripten](https://gitlab.archlinux.org/archlinux/packaging/packages/emscripten/-/blob/main/emscripten.install?ref_type=heads#L1-4).

gdbinit.py
patryk4815: > The new function will basically have duplicated code with skip_autoupdate. Is this fine?

yes it is ok, to "refactor/move" only `.pwndbg_root` check

Xeonacid: Updated. Thanks for the review!

PR #3559, opened 2025-12-30
Fix stdout regex matching for GDB driver

### LINKED ISSUE #3556: Tests incorrectly marked as passed when using gdb
### Description

Using gdb driver (`-d gdb` ) while running tests incorrectly marks skipped tests as passed.

### Steps to reproduce

- Run `tests.sh` with the flag `d gdb` on a test that will be skipped (or modify code to intentionally skip tests). This can be done using `./tests.sh -s -d gdb -g dbg` on an aarch64 system.
- Enable verbose logging using `- v` and run the tests.
- Note that even though the tests are skipped (as shown in stdout), the final result is still marked as "PASSED" for every test that is skipped.

### My setup

- GNU gdb (Debian 13.1-3) 13.1
- Python 3.11.2
- Pwndbg:   2025.10.20 build: dca241bf8 (Linux)

AppleMangoOrange:

Fixes #3556

### DIFF
--- a/pwndbg/commands/integration.py
+++ b/pwndbg/commands/integration.py
@@ -176,7 +176,7 @@ def install_generic_plugin(
     print("Deleting old files (if they exist):")
     for _, dest in paths:
         print(f"\t{dest}")
-        if dest.exists(follow_symlinks=False):
+        if dest.exists() or dest.is_symlink():
             if dest.is_symlink() or not dest.is_dir():
                 # Works for regular files and symlinks.
                 # Note that `dest.is_dir()` returns True for
--- a/pwndbg/commands/integration.py
+++ b/pwndbg/commands/integration.py
@@ -176,7 +176,7 @@ def install_generic_plugin(
     print("Deleting old files (if they exist):")
     for _, dest in paths:
         print(f"\t{dest}")
-        if dest.exists() or dest.is_symlink():
+        if dest.exists(follow_symlinks=False):
             if dest.is_symlink() or not dest.is_dir():
                 # Works for regular files and symlinks.
                 # Note that `dest.is_dir()` returns True for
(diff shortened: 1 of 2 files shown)


k4lizen (changes requested):
Thanks for the PR! I think the fix should look somewhat different.

This output is coming from pytest, not from the debugger, why is there a difference between debuggers? We should make it so there is no difference between the output, rather than match the different outputs.

Also, the former `re.search` looks for this:
image

While the latter looks for this:
image

I don't think we should be matching different things.

k4lizen:
Maybe change the current matching logic to something like this?
```python
        # FIXME: pytest output for LLDB is colored even though it shouldn't be?
        entries = re.search(
            r"((\x1b\[3.m)?(PASSED|FAILED|SKIPPED|XPASS|XFAIL)(\x1b\[0m)?)( .*::.* -)?( (.*))?",
            result.stdout,
            re.MULTILINE,
        )
        if entries:
            stdout_status = entries[3]
            stdout_context = entries[6]

```
which should match for both tests in lldb and gdb. (but it is weird that lldb is throwing out colored output)

AppleMangoOrange:
I have used the expression `(?:\x1b\[3.m)?(PASSED|FAILED|SKIPPED|XPASS|XFAIL)(?:\x1b\[0m)?(?: .*::.* -)?(?: (.*))?` because the one suggested was not working for all cases.

k4lizen:
awesome, could you just add a comment in the code about some tricky cases?

AppleMangoOrange:
I've tried my best to comment about edge cases, please take a look. Also, I think this check should be removed as the process seems to always exit with code 0

```
# Determine low-granularity status from process return code.
status = TestStatus.PASSED if result.returncode == 0 else TestStatus.FAILED
```

k4lizen (approved):
LGTM, thanks!

PR #3560, opened 2025-12-30
Don't install vendored GDB and LLDB in setup.sh

Xeonacid:

setup.sh will create a venv and use gdbinit to launch pwndbg, without installing the pwndbg and pwndbg-lldb commands.

With the gdbinit approach, users will use the system GDB and LLDB, instead of the ones installed in venv.

So installing these binaries in venv is a waste of time, bandwidth and disk space for users.

Feel free to blame me if I'm wrong. :P

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -215,7 +215,7 @@ pip install uv
 
 # Install dependencies
 echo "Installing dependencies.."
-uv sync --extra gdb --extra lldb --quiet
+uv sync --quiet
 
 if [ -z "$UPDATE_MODE" ]; then
     if grep -qs '^[^#]*source.*pwndbg/gdbinit.py' ~/.gdbinit; then


patryk4815:
@k4lizen what do you think about this?

`setup.sh` is now a legacy installation method, but at the moment too much logic still lives inside it. Today, users only need `uv` to install and use pwndbg — they do not even need to run `git clone ...`.

I believe that `git clone` + `./setup.sh` should no longer be treated as a supported installation path for end users. Cloning the repository should primarily be for contributors and developers.

For most users, they should use our vendored GDB/LLDB binaries together with our `pwndbg` / `pwndbg-lldb` entrypoints. Installation could look like this:
- `uv tool install git+https://github.com/pwndbg/pwndbg[gdb,lldb]`
- Or `pip install "pwndbg[gdb,lldb] @ git+https://github.com/pwndbg/pwndbg"` (though I am not sure we should continue supporting `pip` due to lack of version locking)
- Or `curl install.pwndbg.re | sh`

For users who want to use `system-gdb` together with the `pwndbg` entrypoint, they could install like this:
- `uv tool install git+https://github.com/pwndbg/pwndbg`
- Or `pip install "pwndbg @ git+https://github.com/pwndbg/pwndbg"`
- the `pwndbg` entrypoint would then use the `system-gdb` found in `$PATH`

For users who prefer `system-gdb` + `~/.gdbinit` + `source`, they could do something like:
- `uv tool install git+https://github.com/pwndbg/pwndbg`
- Or `pip install "pwndbg @ git+https://github.com/pwndbg/pwndbg"`
- manually add the following to `~/.gdbinit`:
  `source /root/.local/share/uv/tools/pwndbg/share/gdbinit.py`
(We should probably find a better solution for this case. Because `/share/gdbinit.py` is not included yet.)
- Or we should support for example: `curl install.pwndbg.re | sh -s -- -t system-gdb-and-gdbinit`??

Xeonacid:
> I believe that git clone + ./setup.sh should no longer be treated as a supported installation path for end users. Cloning the repository should primarily be for contributors and developers.

If that's true, I think `setup.sh` can be merged into `setup-dev.sh`. 🤔

k4lizen:
> I believe that git clone + ./setup.sh should no longer be treated as a supported installation path for end users. Cloning the repository should primarily be for contributors and developers.

That's already kinda how it is no? Thats why I put
image

installing from source into the contributing section of the docs.

But I agree with what @Xeonacid is saying, currently the vendored gdb is bloat on my system since I'm using system gdb in the dev environment. But I think the path forward should actually be to use `uv tool install` in `setup.sh`, so the vendored gdb is used by default, and then we provide an alternative for contributors who want to use system gdb (like me).

But I guess the system gdb will be in a perpetual state of "not actually supported" since we have or will have patches in our version of gdb and lldb? Then it is kind of weird that we are even providing people with the option to use the system gdb no?

disconnect3d:
I think that ideally we want allow ppl to use just their own 'gdb' with pwndbg - so the old .gdbinit etc --- if possible, heh.

But on the long term, I'm ok with dropping support for it and vendoring everything even with normal installs...

At the same time, I feel like if people do git clone+setup.sh they want to be able to change pwndbg source files on the fly / maybe without all the dev setup etc.

disconnect3d:
Just to sum up, the problem is that `setup.sh` installs `.venv/bin/{gdb,lldb,pwndbg,pwndbg-lldb}` binaries.

However, the change proposed does not fix this anyway because running `gdb` (the system one) will then update deps installing both gdb and lldb binaries:

```
dc@jctf:~/pwndbg$ ls .venv/bin/{gdb,lldb}
ls: cannot access '.venv/bin/gdb': No such file or directory
ls: cannot access '.venv/bin/lldb': No such file or directory

dc@jctf:~/pwndbg$ gdb --batch --ex quit
Detected outdated Pwndbg dependencies (uv.lock). Updating.
pwndbg: loaded 212 pwndbg commands. Type pwndbg [filter] for a list.
pwndbg: created 9 GDB functions (can be used with print/break). Type help function to see them.

dc@jctf:~/pwndbg$ ls .venv/bin/{gdb,lldb}
.venv/bin/gdb  .venv/bin/lldb
```

disconnect3d:
I think we should not drop the installation of gdb and lldb binaries -- or at least not by default.

I think we should make it in the other direction: setup.sh should install both gdb and lldb - as is - by default - but also it should inform users on how to use it.

It should also not install a system GDB binary and it should only do that if someone really wants it, with some --flag?


In the long term, we can probably kill/remove setup.sh in favor of just requiring users to do `uv sync` (or `uv tool install --editable ./pwndbg` or sth)? Though, maybe we want to keep setup.sh so that we can install them pip+uv+execute uv sync..?

disconnect3d:
I'm going to close this: it doesn't fix the stated problem + we really want to use our own gdb and lldb instead of system one.

We want to show users how they can use the vendored gdb and lldb instead -- so they actually also start using lldb too!

[Or, well, we may want to inform users how they should install from a release mode too]

k4lizen:
re: https://github.com/pwndbg/pwndbg/pull/3763

Issue #3557, opened 2025-12-30
Usage of `pathlib.Path.exists(follow_symlinks=False)`


AppleMangoOrange:
### Description

The `pwndbg/commands/integration.py` file uses `dest.exists(follow_symlinks=False)`. This should not be allowed as `follow_symlinks` argument for the `pathlib.Path.exists()` method was officially introduced in Python 3.12 and the project needs to support Python 3.10.

### Steps to reproduce

- Run `./lint.sh`

```
pwndbg/commands/integration.py: note: In function "install_generic_plugin":
pwndbg/commands/integration.py:179: error: Unexpected keyword argument "follow_symlinks" for "exists" of "Path"  [call-arg]
            if dest.exists(follow_symlinks=False):
               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### My setup

Pwndbg:   2025.10.20 build: 7dadfebe9 (Linux)
Python:   3.10.12 (main, Nov  4 2025, 08:48:33) [GCC 11.4.0]

(official docker compose container)

AppleMangoOrange:
I don't understand why this wasn't caught by github actions..

k4lizen:
> I don't understand why this wasn't caught by github actions..

Has been discussed, seems its a vermin bug so we opened an issue

https://github.com/netromdk/vermin/issues/327

PR #3558, opened 2025-12-30
Replace usage of follow_symlinks in pathlib.Path.exists()

AppleMangoOrange:
+ [x] I read the contributing documentation.
+ [ ] I am providing a screenshot of the (new feature) / (fixed bug).

Fixes #3557

PR #3563, opened 2025-12-31
Run lint on the lowest supported ubuntu version

k4lizen:
Should prevent issues like https://github.com/pwndbg/pwndbg/issues/3557 in the future.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -12,7 +12,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-24.04]
+        # We perform lint on the lowest ubuntu version we support to allow mypy
+        # to catch backwards-incompatibility bugs (when vermin fails to, like #3557).
+        os: [ubuntu-22.04]
     runs-on: ${{ matrix.os }}
     timeout-minutes: 10
     steps:
--- a/scripts/common.sh
+++ b/scripts/common.sh
@@ -36,7 +36,10 @@ else
     UV_RUN="${UV} run"
     UV_RUN_TEST="${UV_RUN} --group dev --group tests --all-extras"
     UV_RUN_LINT="${UV_RUN} --group lint"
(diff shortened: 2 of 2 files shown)


k4lizen:
Ugh.
```bash
+ /home/user/code/pwndbg/.venv/bin/uv run --group lint vermin -vvv --no-tips -t=3.10- --eval-annotations --violations pwndbg pwndbginit tests gdbinit.py scripts
Detecting python files..
Analyzing 407 files using 16 processes..
Minimum required versions: 3.10
Incompatible versions:     2
+ [[ -z '' ]]
+ /home/user/code/pwndbg/.venv/bin/uv run --python 3.10.12 --group dev --group lint --group tests --extra gdb --extra lldb mypy pwndbg pwndbginit tests/host
error: Querying Python at `/home/userlocal/share/uv/python/cpython-3.10.12-linux-x86_64-gnu/bin/python3.10` failed with exit status exit status: 127

[stderr]
/home/user/.local/share/uv/python/cpython-3.10.12-linux-x86_64-gnu/bin/python3.10: error while loading shared libraries: libcrypt.so.1: cannot open shared object file: No such file or directory
```
Required me to install the `libxcrypt-compat` arch package.

k4lizen:
Okay I won't force contributors to have the same python version as CI, so we don't need libxcrypt-compat or whatever. (force pushed the fix)

k4lizen:
okay cool, mypy catches the bug now: https://github.com/pwndbg/pwndbg/actions/runs/20613698512/job/59202572271?pr=3563 .

this should be merged after https://github.com/pwndbg/pwndbg/pull/3558 .

k4lizen:
we could pin mypy to use python 3.10 version without specifying patch, but this is a bit tricky for various reasons (e.g. I don't want to install multiple python versions for all contributors, and docgen already has its thing) so while i'm not opposed to doing that in the future i think its fine to merge this as-is

patryk4815:
maybe we should add option like `--ci` only for GitHub-CI?

PR #3564, opened 2026-01-01
stepsyscall should allow to step until syscall number or until given expression is true issue #3552

### LINKED ISSUE #3552: stepsyscall should allow to step until syscall number or until given expression is true
The `stepsyscall` command should allow me to step until a given syscall number is hit (e.g. `SYS_ptrace` or `123`) or until a given expression/condition is met?

I could imagine the interface could look like:
```
stepsyscall SYS_ptrace
stepsyscall 123
stepsyscall $rax==123
stepsyscall SYS_write $rdi=0
```

Maybe like this? Ideally, would be nice if we allowed for some memory dereferences here... But the least we need is stepsyscall until a given syscall is hit.

D3LET3-57:
This PR implements the requested feature along with test coverage.

Support for loongarch64 could not be included, as pwnlib.constants.linux does not currently provide a loongarch64 module. I’ve kept the implementation structured so this can be added easily in the future.

Closes: https://github.com/pwndbg/pwndbg/issues/3552.


pwndbg/aglib/next.py
```diff
@@ -152,6 +153,91 @@ async def break_next_interrupt(
...
+                   (e.g., checking register values like $rdi==0)
+
+    Returns:
+        The instruction we stopped at, or None if process died/signaled
+    """
+    # Get the syscall register for the current architecture
+    syscall_reg = None
+    if syscall_num is not None:
+        arch_name = pwndbg.aglib.arch.name if pwndbg.aglib.arch else None
+        syscall_reg = {
```
disconnect3d: this is wrong, we should have that info somewhere in aglib/architecture information and we should fetch from there

D3LET3-57: yeah i found it. we can use `pwndbg.aglib.arch.syscall_abi`

D3LET3-57:
Can you also help me with those LLDB issues??

D3LET3-57:
in `ubuntu24.04` jemalloc tests are falling. It is same with my local machine too.  
edit : fixed the issue on my machine but not sure in docker

k4lizen:
dont worry about the jemalloc tests, they are flaky for some reason

pwndbg/commands/next.py
```diff
@@ -155,34 +158,87 @@ def nextsyscall() -> None:
...
+) -> None:
     """
     Execution controller for the `stepsyscall` command.
     """
-
-    while (
-        pwndbg.aglib.proc.alive()
-        and not (await pwndbg.aglib.next.break_next_interrupt(ec, honor_current_branch=True))
-        and (await pwndbg.aglib.next.break_next_branch(ec, including_current=True))
+    if await pwndbg.aglib.next.break_next_interrupt_filtered(
```
disconnect3d: Fwiw the old logic is slow already: do we maybe want to keep/run the old code if syscal num and condition are both none?

pwndbg/commands/next.py
```diff
@@ -155,34 +158,87 @@ def nextsyscall() -> None:
...
+        stepsyscall SYS_write $rdi==1  - Break at write syscall when fd==1 (stdout)
+        stepsyscall $rax==60         - Break when syscall number is 60 (exit)
     """
+    syscall_num = None
+    cond_callable = None
+
+    # Parse syscall argument
+    if syscall is not None:
+        # Check if it's actually a condition (starts with $ or contains operator)
+        if syscall.startswith("$") or any(op in syscall for op in ["==", "!=", ">", "<"]):
```
disconnect3d: I don't think this makes sense ;/

disconnect3d (commented):
I'm also wondering: there is `catch syscall open` in GDB etc. does this allow for conditions e.g. for register values?

D3LET3-57:
> I'm also wondering: there is `catch syscall open` in GDB etc. does this allow for conditions e.g. for register values?


https://techno-overload.blogspot.com/2014/10/advanced-debugging-tricks-using-gdb.html

i have found this article where we can use `catch` and `condition` to break the process. In my opinion `stepsyscall` can be a higher level abstraction instead of two different commands.

pwndbg/commands/next.py
D3LET3-57: > Fwiw the old logic is slow already: do we maybe want to keep/run the old code if syscal num and condition are both none?

i think old logic is fine. first we are looking for an interrupt in present branch if it doesn't exists we are stepping into next one. If you have any better logic i will try to implement it.

pwndbg/commands/next.py
D3LET3-57: well it thought this could handle this type of situation but i didnt test.
`stepsyscall $rdi==0 SYS_write`
i will look into it.

D3LET3-57:
@disconnect3d ??

pwndbg/lib/syscall.py
```diff
@@ -0,0 +1,153 @@
...
+        "armcm": linux_thumb,
+        "rv32": linux_riscv64,
+        "rv64": linux_riscv64,
+        "sparc": linux_sparc,
+        "sparc64": linux_sparc64,
+        "powerpc": linux_powerpc,
+        "powerpc64": linux_powerpc64,
+        "s390x": linux_s390x,
+        # Note: loongarch64 not available in pwnlib
+    }.get(pwndbg.aglib.arch.name)
```
disconnect3d: Image

we have this code elsewhere; it should be unified somehow.

pwndbg/lib/syscall.py
```diff
@@ -0,0 +1,153 @@
...
+        if attr_name.startswith("__NR_") and isinstance(value, Constant)
+    }
+
+    num = SYS_BY_NAME.get(syscall_name)
+    if num is None:
+        return (None, None)
+    return (int(num), name_or_num)
+
+
+def parse_condition(condition: str) -> Optional[Callable[[], bool]]:
```
disconnect3d: I think we really don't want to parse a condition ourselves. We rather want to execute the condition under the debugger that is used. Sth like `int(pwndbg.dbg.selected_inferior().evaluate_expression("$rdi == 1"))`

pwndbg/aglib/next.py
```diff
@@ -152,6 +153,78 @@ async def break_next_interrupt(
     return None
 
 
+async def break_next_interrupt_filtered(
```
disconnect3d, line 157: I thikn it would be better to do sth like is doone here: https://github.com/pwndbg/pwndbg/pull/3601/changes

that we break if a predicate is met, instead of getting syscall number or condition here and doing any kind of `if syscall_num is not None and syscall_reg:` blablabla.

disconnect3d (changes requested):
this is still far from good tbh, see comments

pwndbg/lib/syscall.py
D3LET3-57: this function can be used when user input has syscall number.

we don't need syscall name we just need syscall number so that we can check at possible interrupts.

D3LET3-57:
@disconnect3d i've made some changes please look into it.

k4lizen:
Frankly I think that this whole `next.py` system in pwndbg should probably be reworked to use the underlying `catch` commands of the debugger instead of literarly stepping a billion times. Surely the debuggers do some ptrace magic and intercept the interrupts instead of just running `stepi` like we do i.e. I think what we are doing is insanely slow.

But okay, thats out of scope for this PR. @D3LET3-57 you should merge dev into this branch and fix the lint issues. I'll review now.

docs/commands/step_next_continue/stepsyscall.md
```diff
@@ -2,13 +2,20 @@
...
 ```
 
 Breaks at the next syscall by taking branches.
 
 **Alias:** stepsc
+### Positional arguments
+
+|Positional Argument|Help|
+| :--- | :--- |
+|syscall|Syscall number (e.g., 1, 0x3c) or name (e.g., SYS_write, SYS_exit)|
```
k4lizen: for sure we don't want to force the user to type "SYS_", why not just `write`, `exit`, etc?

pwndbg/aglib/next.py
```diff
@@ -152,6 +153,58 @@ async def break_next_interrupt(
...
+    ec: pwndbg.dbg_mod.ExecutionController,
+    predicate: Callable[[], bool] | None = None,
```
k4lizen: is there a reason that the predicate is optional?

pwndbg/lib/syscall.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+def get_arch_module(arch_name: str | None) -> Any:
+    """
+    Gets the architecture module for the current architecture.
+
+    Returns None if no architecture is set (e.g., no process running) or unsupported.
+    """
+    if arch_name is None:
+        return None
```
k4lizen: why is arch_name `| None`? It would make more sense API-wise if it was just `str`.

pwndbg/commands/next.py
k4lizen: I agree with @disconnect3d that we shouldn't do this. I think its fine to just make the condition an argument with a flag `--condition` / `-c` and make our lives easier.

pwndbg/lib/syscall.py
```diff
@@ -0,0 +1,106 @@
...
+from pwnlib.constants.linux import riscv64 as linux_riscv64
+from pwnlib.constants.linux import s390x as linux_s390x
+from pwnlib.constants.linux import sparc as linux_sparc
+from pwnlib.constants.linux import sparc64 as linux_sparc64
+from pwnlib.constants.linux import thumb as linux_thumb
+
+from pwndbg.lib.regs import reg_sets
+
+
+def get_arch_module(arch_name: str | None) -> Any:
```
k4lizen: this should be called `_get_pwntools_arch_module` and not accept None

pwndbg/lib/syscall.py
```diff
@@ -0,0 +1,106 @@
...
+        "powerpc": linux_powerpc,
+        "powerpc64": linux_powerpc64,
+        "s390x": linux_s390x,
+        # Note: loongarch64 not available in pwnlib
+    }.get(arch_name)
+
+    return arch_module
+
+
+def get_syscall(name_or_num: str, arch_name: str | None) -> Tuple[Optional[int], Optional[str]]:
```
k4lizen: since this is library API, it shouldn't have loose parsing like this. split this up into two functions, one which takes an int and returns the name, and one which takes the name and returns an int

pwndbg/lib/syscall.py
```diff
@@ -0,0 +1,106 @@
...
+    """
+    Resolve the syscall into (number, name).
+
+    Accepts:
+        - Syscall number as string: "1", "60", "0x3c"
+        - Syscall name with SYS_ prefix: "SYS_write", "SYS_exit"
```
k4lizen: i don't really see why it should have the SYS_ prefix, its probably fine to just return it without that prefix

pwndbg/lib/syscall.py
```diff
@@ -0,0 +1,106 @@
...
+    # Handle name lookup
+    if name_or_num.startswith("SYS_"):
+        syscall_name = name_or_num[4:]  # "SYS_write" -> "write"
+    else:
+        return (None, None)
+
+    SYS_BY_NAME = {
+        attr_name[5:]: value  # "__NR_write" -> "write"
+        for attr_name, value in vars(arch_module).items()
+        if attr_name.startswith("__NR_") and isinstance(value, Constant)
```
k4lizen: whats up with the __NR_ thing?

pwndbg/lib/syscall.py
k4lizen: everywhere where you are taking an arch name, the type should be `PWNDBG_SUPPORTED_ARCHITECTURES_TYPE` instead of `str`

tests/library/dbg/tests/test_command_stepsyscall.py
```diff
@@ -30,7 +36,73 @@ async def test_command_nextsyscall(ctrl: Controller) -> None:
     import pwndbg.aglib.symbol
 
     await ctrl.launch(STEPSYSCALL_X64_BINARY)
+    if not pwndbg.dbg.is_gdblib_available():
+        await ctrl.step_instruction()
```
k4lizen, line 44: why? do gdb and lldb stop on different instructions? this should be clarified with a comment

tests/library/gdb/tests/test_command_stepsyscall.py
```diff
@@ -1,5 +1,7 @@
 from __future__ import annotations
```
k4lizen, line 1: no need to touch any gdb/ tests, they are deprecated in favour of dbg/. revert the changes in this file plz.

k4lizen (changes requested):
I'm generally quite satisfied with the code, just a bunch of ~small things to fix up in various places.

pwndbg/commands/next.py
D3LET3-57: ok i will add flags so that it will be easy

pwndbg/aglib/next.py
D3LET3-57: yeah my bad it shouldn't be optional even if args are not present we are returning a predicate which always gives true which is similar to normal `stepsyscall`

D3LET3-57:
@k4lizen i have added the changes please look into it.

tests/library/dbg/tests/test_command_stepsyscall.py
```diff
@@ -30,7 +38,83 @@ async def test_command_nextsyscall(ctrl: Controller) -> None:
     import pwndbg.aglib.symbol
 
     await ctrl.launch(STEPSYSCALL_X64_BINARY)
+    # LLDB launches with -s flag which causes a SIGSTOP that can interfere with
+    # stepping commands. Step once to move past the initial stop state.
+    if not pwndbg.dbg.is_gdblib_available():
```
k4lizen, line 43: i would probably only write this comment once xd

and also check "this is lldb" rather than "this is not gdb" with this if-statement

but its not a big deal

k4lizen (approved):
Alright, LGTM, thanks for the PR!

Let's wait to see if @disconnect3d has any final comments before merging, since it was his feature request.

k4lizen:
btw in the future please put a link to the original issue in the PR description instead of the PR title. Like here:

image


image

pwndbg/commands/next.py
```diff
@@ -155,34 +158,111 @@ def nextsyscall() -> None:
...
-    while (
-        pwndbg.aglib.proc.alive()
-        and not (await pwndbg.aglib.next.break_next_interrupt(ec, honor_current_branch=True))
-        and (await pwndbg.aglib.next.break_next_branch(ec, including_current=True))
-    ):
-        # Here we are e.g. on a CALL instruction (temporarily breakpointed by `break_next_branch`)
-        # We need to step so that we take this branch instead of ignoring it
-        await ec.single_step()
-        continue
+    def predicate() -> bool:
```
disconnect3d: Please note that `predicate` will be called on each step so if it does any unnecessary action, it will just be wasting time on each step, which may be called thousands or milions of time.

As a result, doing things like `if syscall_num is not None` in it, is rather unnecessary - because we can do check this once.

Same for '`if condition is not None`.

like, if we know predicate is empty, we should have no predicate etc.

disconnect3d (changes requested):
The predicate checks does a lot of unnecessary job, wasting computing time heavily; i can't accept it in current form

disconnect3d:
@D3LET3-57 Please see feedback and let us know if u have any more questions

D3LET3-57:
@disconnect3d  i tried to add respective predicates for given args

pwndbg/aglib/next.py
```diff
@@ -152,6 +153,58 @@ async def break_next_interrupt(
...
+        if branch:
+            if branch.address != pwndbg.aglib.regs.pc:
+                proc = pwndbg.dbg.selected_inferior()
+                with proc.break_at(BreakpointLocation(branch.address), internal=True) as bp:
+                    await ec.cont(bp)
+            # Step past the branch
+            await ec.single_step()
+        else:
+            # No branch found either, just single step
+            await ec.single_step()
```
disconnect3d: the `ec.single_step()` is in both if and else; why? just put it outside.

disconnect3d (changes requested):
While it may make sense to have some kind of utility function to "step until syscall + predicate", I believe the current `break_next_interrupt_filtered` is not great. It is hard to read - has duplicated code lines and seems unncessarily complex. Please see this:

Image

TL;DR: We should just reimplement `_stepsyscall` with an additional condition - the predicate check. And call either `_stepsyscall` if no predicate is there, or the conditioned version of it.

pwndbg/commands/next.py
```diff
@@ -155,34 +158,128 @@ def nextsyscall() -> None:
...
+            return
+        syscall_reg = syscall_abi.syscall_register
+
+        def check_syscall() -> bool:
+            return pwndbg.aglib.regs.read_reg(syscall_reg) == syscall_num
+
+        if condition is not None:
+
+            def check_syscall_and_condition() -> bool:
+                if pwndbg.aglib.regs.read_reg(syscall_reg) != syscall_num:
```
disconnect3d: I think that this whole function could be just:
```py
return pwndbg.aglib.regs.read_reg(syscall_reg) == syscall_num and bool(condition())
```

I am not sure about try except tho, when/why do we need it? Can we talk about it?

Because also `except Exception:` - aka "catch them all" is usually not great and we want to avoid it if possible

k4lizen:
Your code is correct but organizing it this way would make it more readable:

do it like this (@disconnect3d agrees with this)
```python
async def _stepsyscall(
    ec: pwndbg.dbg_mod.ExecutionController,
    syscall_num: int | None = None,
    condition: Callable[[], int] | None = None,
) -> None:

    predicate: Callable | None = _get_syscall_predicate(syscall_num, condition)

    if predicate is None:
        while (
            pwndbg.aglib.proc.alive()
            and not (await pwndbg.aglib.next.break_next_interrupt(ec, honor_current_branch=True))
            and (await pwndbg.aglib.next.break_next_branch(ec, including_current=True))
        ):
            # Here we are e.g. on a CALL instruction (temporarily breakpointed by `break_next_branch`)
            # We need to step so that we take this branch instead of ignoring it
            await ec.single_step()
            continue
    else:
        while (
            pwndbg.aglib.proc.alive()
            and not (await pwndbg.aglib.next.break_next_interrupt_filtered(ec, honor_current_branch=True, predicate))
            and (await pwndbg.aglib.next.break_next_branch(ec, including_current=True))
        ):
            await ec.single_step()
            continue
```
Then make `break_next_interrupt_filtered` take a non-optional `predicate`, and do the exact same thing as `break_next_interrupt` except also check the condition.

It is a bit confusing that currently `break_next_interrupt` stops at basic blocks but `break_next_interrupt_filtered` doesn't. They should both do the same (stop at basic blocks).

D3LET3-57:
@disconnect3d added those changes please look into it.

disconnect3d:
Thanks a lot for this!

PR #3565, opened 2026-01-02
Replace lief with niche-elf

k4lizen:
I wrote this library to facilitate some stuff we need: https://github.com/k4lizen/niche-elf . bata and mahaloz can also use it if they like :>

It is of similar speed to the objcopy method, i.e. is orders of magnitude times faster than lief and `klookup --apply` actually finishes in a few seconds (instead of hours (?)).

I might rewrite it to C++ with python bindings if it seems worth at some point. When we get https://github.com/mahaloz/decomp2dbg/issues/149 I plan to add an API optimized for changing a symbols name on the ELF file, so we may at some point be able to actually turn on `decompiler-autosync-syms` which would be pretty cool.


k4lizen:
I need to fix one more thing, which is that it seems that GDB actually looks at the `e_machine` ELF header which specifices the CPU architecture. I was hoping that wasn't the case, as ELF requires specifying the ptrsize anyways, but yeah some weird stuff happens when there is a mismatch there. So I guess I'll have to write a mapping like zig architecture -> ELF e_machine field.

pwndbg/commands/klookup.py
```diff
@@ -43,22 +49,36 @@ def klookup(symbol: str, apply: bool) -> None:
...
-            for sym_name, sym_type, sym_addr in syms:
-                symelf.add_symtab_symbol(symelf.export_symbol(sym_name, sym_addr))
-            symelf.write(path)
-            pwndbg.dbg.selected_inferior().add_symbol_file(path)
-            print(message.success(f"Added {len(syms)} symbols"))
-        except Exception as e:
-            print(message.error(e))
+        base: int = paging_info.kbase
+
+        _, elf_path = tempfile.mkstemp(prefix="ks-symbols-", suffix=".elf")
```
patryk4815: this file should be removed, but when?

k4lizen: ou true, I forgot about that! hmmm

I guess I could pass `-readnow` to add-symbol-file and remove it instantly. otherwise I guess we can remove it on pwndbg exit

pyproject.toml
```diff
@@ -34,8 +34,8 @@ dependencies = [
     # For installing the decompiler plugins (we don't actually use it as a python library).
     # If you bump this, also bump the pwndbg/commands/integration.py:d2d_required* variables.
     "decomp2dbg==3.14.0",
-    # Optional? Used for getting symbols into the debugger with the decompiler synchronization
-    "lief>=0.17.1 ; (platform_system == 'Linux' and (platform_machine == 'x86_64' or platform_machine == 'aarch64' or platform_machine == 'i686')) or platform_system != 'Linux'",
+    # Used for symbol syncing (e.g. klookup --apply and decompiler integration)
+    "niche-elf==0.3.4 ; platform_machine != 'loongarch64'",
```
patryk4815: `platform_machine != 'loongarch64'` can be removed, no idea why you added that xd

k4lizen: i was having some issues with dep resolution, the actual fix was that I needed to run `uv sync -n --all-extras --all-groups` to bypass the pypi cache since the package was too recently published.

btw, when do you actually need to add `platform_machine != 'loongarch64'`?

patryk4815: Only when you want use `loong64.pypi.org` for wheels - mostly for python libraries that use `C` bindings

pwndbg/commands/klookup.py
```diff
@@ -43,22 +49,36 @@ def klookup(symbol: str, apply: bool) -> None:
                 syms.append(sym)
         if len(syms) == 0:
             print(message.error(f"No symbol found for {symbol}"))
-    for sym_name, sym_type, sym_addr in syms:
-        print(message.success(f"{sym_addr:#x} {sym_type} {sym_name}"))
+
+    if not apply or symbol != "":
```
jxuanli: let's remove this outer if. for example `klookup` without any input args should display all symbols (similar to if you just `cat /proc/kallsyms` without grep). For `apply`, I think we should tell the user what symbols are added.

pwndbg/commands/klookup.py
```diff
@@ -43,22 +49,36 @@ def klookup(symbol: str, apply: bool) -> None:
...
 
-            symelf = lief.ELF.parse(path)
-            for sym_name, sym_type, sym_addr in syms:
-                symelf.add_symtab_symbol(symelf.export_symbol(sym_name, sym_addr))
-            symelf.write(path)
-            pwndbg.dbg.selected_inferior().add_symbol_file(path)
-            print(message.success(f"Added {len(syms)} symbols"))
-        except Exception as e:
-            print(message.error(e))
+        base: int = paging_info.kbase
```
jxuanli: I believe this can be None despite its annotated to be always int.

pwndbg/commands/klookup.py
jxuanli: yea we can remove it right after its added but depends on if you want to add a caching layer until objfile or something like that, tho im not exactly sure how we can be syre the objfile is vmlinux - maybe we can get the file path for functions decorated with cache_until(objfile) but im not sure how to do that

pwndbg/commands/klookup.py
k4lizen: image

fun!

pwndbg/commands/klookup.py
k4lizen: I 100% agree that we should always print the symbols to the user when they do `klookup`, the only case where its annoying to me is specifically when I run:
```bash
klookup --apply
```
with no filters. Because then I really don't care about all the symbols in the kernel (and if I do I can simply re-run `klookup`), but it completely destroys my terminal scrollback. I changed the if condition to make it more clear

k4lizen: but if you still want me to remove it lmk, i'm not that passionate about it

k4lizen:
I am pinning e_machine to x86_64 and the size of the ELF to ELFCLASS64. I tested:

+ Debugging x86_64 binary on x86_64 host.
+ Debugging aarch32 binary on x86_64 host.
+ Debugging aarch64 binary on aarch64 host.

and it all works. If there are problems in the future I will fix it then, but I'm 99% sure noone will encounter any issues. 

I sent an email to the binutils folks because they maintain the ELF parsing subsystem of GDB about it also.

jxuanli:
@k4lizen I opened some issues in the niche-elf repo. I think we should address those.

pwndbg/commands/klookup.py
```diff
@@ -43,22 +50,49 @@ def klookup(symbol: str, apply: bool) -> None:
...
-        except Exception as e:
-            print(message.error(e))
+
+        base: int | None = paging_info.kbase
+
+        if base is None:
+            # I would be suprised if this was actually possible, we managed to find kallsyms but not
+            # kbase?
+            # But anyway, passing 0 to ELFFile and no ADDR argument to add_symbol_file should still work.
+            elf: niche_elf.ELFFile = niche_elf.ELFFile(0)
```
jxuanli, line 76: we should be able to specify the target arch here. see https://github.com/pwndbg/niche-elf/issues/1

pwndbg/commands/klookup.py
```diff
@@ -43,22 +50,49 @@ def klookup(symbol: str, apply: bool) -> None:
...
+
+            if sym_type in ["T", "t", "W", None]:
+                elf.add_function(sym_name, sym_addr, bind=bind)
+            else:
+                elf.add_object(sym_name, sym_addr, bind=bind)
+
+        _, elf_path = tempfile.mkstemp(prefix="ks-symbols-", suffix=".elf")
+        elf.write(elf_path)
+
+        if base is None:
```
jxuanli: `add_symbol_file` (only implemented for gdb atm) handles the None case (feel free to add that for lldb:p)

disconnect3d (approved):
Go ahead and merge it if there will be no more comments :)

pwndbg/commands/klookup.py
k4lizen: fixed!

k4lizen:
@jxuanli I'll investigate the stuff you raised in niche-elf further (i feel proud of my own repo being relevant enough to get an issue ;p), but I don't think they are blockers for this, so I'll go ahead and merge it

PR #3561, opened 2025-12-30
Use qGDBServerVersion for QEMU version detection

nosuchfilefound:
## Summary:

  - Add a cached qGDBServerVersion query + parser for QEMU version detection.
  - Use the new packet for QEMU detection and version checks, but still keep legacy fallbacks.

  ## Tests That I Ran:
  - ./lint.sh
```
$ ./lint.sh
+ LINT_FILES='pwndbg pwndbginit tests *.py scripts'
...snip...
uv run --group dev --group lint --group tests --extra gdb --extra lldb mypy pwndbg pwndbginit tests/host
Success: no issues found in 264 source files
```
  - ./unit-tests.sh
```
$ ./unit-tests.sh
===================== test session starts ======================
platform linux -- Python 3.12.3, pytest-8.0.2, pluggy-1.6.0
rootdir: /home/rob/src/github.com/pwndbg
plugins: cov-4.1.0
collected 49 items                                             

tests/unit_tests/test_arch.py .                          [  2%]
tests/unit_tests/test_bit_math.py .....                  [ 12%]
tests/unit_tests/test_functions.py ....                  [ 20%]
tests/unit_tests/test_hex2ptr.py ...                     [ 26%]
tests/unit_tests/test_memory.py ..                       [ 30%]
tests/unit_tests/test_qemu.py ....                       [ 38%]
tests/unit_tests/test_regs.py ......                     [ 51%]
tests/unit_tests/test_which.py ...                       [ 57%]
tests/unit_tests/test_zig_asm.py ..................sss   [100%]

=========== 46 passed, 3 skipped in 60.90s (0:01:00) ===========
```

  ## Issue: 

 - Fixes #3501

  ## Checklist:

  - [x] I read the contributing documentation.
  - [x] I am providing a screenshot of the (new feature) / (fixed
    bug).
     - QEMU 8 and QEMU 10 screenshots of behavior provided in a comment below.

Issue #3501, opened 2025-12-15
pwndbg should start using `qGDBServerVersion` for qemu version detection


patryk4815:
https://github.com/pwndbg/pwndbg/issues/2813#issuecomment-3655928084

The qGDBServerVersion was added in QEMU 10.1.0 thx to https://github.com/qemu/qemu/commit/b2654598b3330aaa58ab0cec2114843bfa96ddaa

nosuchfilefound:
I came across this recently on my own when setting up a qemu environment for linuc kernel VR. Figured, I'll take a shot at implementing this feature request as it would help me a lot. Stay tuned for a merge request!

nosuchfilefound:
I added a pull request here that is going through the pipeline: https://github.com/pwndbg/pwndbg/pull/3561

nosuchfilefound:
Pipeline passed and ready for your review. :)

disconnect3d:
Completed in https://github.com/pwndbg/pwndbg/pull/3561 :) Thanks!

disconnect3d:
Reopening as #3561 code was not best.

I believe one place where we want to check for QEMU's gdbserver version is here: https://github.com/pwndbg/pwndbg/blob/8a48052d20307a7cef60b90cacd6336fbcdf5f6e/pwndbg/aglib/file.py#L152-L164

We do some not great operations (trying to open a file) on QEMU end to detect a bug we added to QEMU one day in the past. This bug (rendering numbers as `%d` instead of`%x` or vice versa, I don't remember)  has been fixed since then in a QEMU release.

Which version? I don't know, but we have to check that.

 @patryk4815 where else shall we check for version?

patryk4815:
@disconnect3d https://github.com/qemu/qemu/commit/8b647bd352505234cab2acd2422aba183a1aa1fd

disconnect3d:
Ok, so its only this exact case when we want to check for version.

nosuchfilefound:
Thank you both for the feedback, I will follow up in the MR #3574 with updated code to review today!

disconnect3d:
Fixed in https://github.com/pwndbg/pwndbg/pull/3574


Thanks @nosuchfilefound

PR #3574, opened 2026-01-02
Fix is_old_qemu_user logic for qGDBServerVersion

nosuchfilefound:
Summary:

  - Fixes comment [here](https://github.com/pwndbg/pwndbg/pull/3561#discussion_r2658421792) in MR #3561
  - Use the existence of qGDBServerVersion as “not old QEMU” in if condition since it exists only in QEMU 10.1+.
  - Keep exec_file_supported() as the fallback when qGDBServerVersion does not exist

  Tests:

  - ./lint.sh
  - ./unit-tests.sh

  Issue: Follow‑up to #3501

Screenshots available [here](https://github.com/pwndbg/pwndbg/pull/3561#issuecomment-3700240334).


pwndbg/aglib/qemu.py
```diff
@@ -98,9 +98,9 @@ def is_old_qemu_user() -> bool:
     if not is_qemu_usermode():
         return False
 
-    version = qemu_gdbserver_version()
-    if version is not None:
-        return version < (8, 1)
```
disconnect3d, line 103: @patryk4815 do we want to check against some specific version here?

patryk4815: code need to be rolled back to the previous version

disconnect3d: cc @nosuchfilefound

nosuchfilefound: Done! Just rolled it back to the previous version. Please let me know if I can take another shot at integrating this feature or any feedback/advice you have.

disconnect3d: So you reverted the whole PR/commit that was merged?

I believe we still want to use the qGDBServerVersion *somewhere*...

disconnect3d: Please see https://github.com/pwndbg/pwndbg/issues/3501#issuecomment-3712073218.

pwndbg/aglib/qemu.py
```diff
@@ -4,51 +4,16 @@
...
-    return _parse_qgdbserverversion(response)
-
 
 @pwndbg.lib.cache.cache_until("stop")
 def is_qemu() -> bool:
     inferior = pwndbg.dbg.selected_inferior()
     if not inferior.is_remote():
         return False
 
-    if qemu_gdbserver_version() is not None:
```
disconnect3d, line 49: Fwiw the previous code was also not great since we are doubling on `selected_inferior()` and `is_remote()` calls :(.

pwndbg/aglib/qemu.py
```diff
@@ -4,51 +4,16 @@
...
-def qemu_gdbserver_version() -> tuple[int, ...] | None:
-    """
-    Returns QEMU version. Works since QEMU 10.1.0
-    """
-    inferior = pwndbg.dbg.selected_inferior()
-    if not inferior.is_remote():
-        return None
-
-    response = inferior.send_remote("qGDBServerVersion")
-    return _parse_qgdbserverversion(response)
```
disconnect3d, line 40: 1) we want to keep those functions that were here

nosuchfilefound:
From reviewing the qemu source, it appears the bug fix (https://github.com/qemu/qemu/commit/8b647bd352505234cab2acd2422aba183a1aa1fd) was first introduced in qemu version  `v10.0.0-rc0`.

```bash
$ cd qemu/
$ git describe --contains 8b647bd352505234cab2acd2422aba183a1aa1fd
v10.0.0-rc0~72^2~3
```

However, the `qGDBServerVersion` packet wasn't added until `v10.1.0-rc0` (https://github.com/qemu/qemu/commit/b2654598b3330aaa58ab0cec2114843bfa96ddaa).

```bash
$ git describe --contains b2654598b3330aaa58ab0cec2114843bfa96ddaa
v10.1.0-rc0~58^2~1
```

So the version check needs to consider the following scenarios:
1. qemu version is  `>= v10.1.0-rc0`
    * Query `qGDBServerVersion`
    * `is_vfile_qemu_user_bug()` returns `False`
2. qemu version is  `>= v10.0.0-rc0` but `< v10.1.0-rc0`
    * Still need to run the hacky `vFile` check since its been fixed but `qGDBServerVersion` doesn't exist yet
    * `is_vfile_qemu_user_bug()` returns `False`
3. qemu version is `< v10.0.0-rc0`
    * Need to run the hacky `vFile` check
    * `is_vfile_qemu_user_bug()` returns `True`

So because the bug fix for the decimal/hex happened before the `qGDBServerVersion` packet was added, we don't get to replace the version logic check completely with the `qGDBServerVersion` packet check. :( I can go ahead and add this version check in anyways for the easiest case of "qemu version is  `>= v10.1.0-rc0`", but... the hacky `vFile` check I think still needs to exist.

Please let me know if I'm understanding correctly.

pwndbg/aglib/qemu.py
nosuchfilefound: I added back in the function `qemu_gdbserver_version()`, but let me know if you think `_parse_qgdbserverversion()` should be `parse_qgdbserverversion()` instead to have other code use their own `inferior` and simply use the parse function.

pwndbg/aglib/qemu.py
nosuchfilefound: Added the version check in `is_vfile_qemu_user_bug()`!

PR #3580, opened 2026-01-04
add ctx -a 

D3LET3-57:

### DIFF
--- a/tests/library/dbg/tests/test_context_commands.py
+++ b/tests/library/dbg/tests/test_context_commands.py
@@ -485,6 +485,46 @@ async def test_context_hide_sections(ctrl: Controller) -> None:
 
 
 @pwndbg_test
+async def test_context_all_sections_flag(ctrl: Controller) -> None:
+    """
+    Tests that context -a/--all shows all sections regardless of context-sections config.
+    """
+    await launch_to(ctrl, REFERENCE_BINARY, "main")
+
+    # Get default context output to establish baseline
+    await ctrl.execute("set context-sections regs disasm code stack backtrace")
+    default_out = await ctrl.execute_and_capture("context")
+
+    # Count section headers in default output
+    default_sections = default_out.count("[ ")
+
+    # Now restrict to minimal config
+    await ctrl.execute("set context-sections regs")
+    minimal_out = await ctrl.execute_and_capture("context")
+    minimal_sections = minimal_out.count("[ ")
+
+    # Minimal should have fewer sections
+    assert minimal_sections < default_sections
+
+    # Now use -a flag - should show all sections despite minimal config
+    all_out = await ctrl.execute_and_capture("context -a")
+    all_sections = all_out.count("[ ")
+
+    # -a should show at least as many as the default config
+    assert all_sections >= default_sections
+
+    # Verify specific core sections are present
+    assert "REGISTERS" in all_out
+    assert "DISASM" in all_out
+    assert "STACK" in all_out
+    assert "BACKTRACE" in all_out
+
+    # Verify --all alias works identically
+    alias_out = await ctrl.execute_and_capture("context --all")
+    assert alias_out.count("[ ") == all_sections
+
+
+@pwndbg_test
 async def test_context_history_prev_next(ctrl: Controller) -> None:
     import pwndbg
 
--- a/tests/library/dbg/tests/test_context_commands.py
+++ b/tests/library/dbg/tests/test_context_commands.py
@@ -538,7 +538,7 @@ async def test_context_all_sections_flag(ctrl: Controller) -> None:
     # Now use -a flag - should include ARGUMENTS section when displaying all sections
     all_out_after_nextcall = await ctrl.execute_and_capture("ctx -a")
     expected_all.insert(0, "ARGUMENTS")
-    
+
     all_sections_after_nextcall = extract_context_sections(all_out_after_nextcall)
     assert all_sections_after_nextcall == expected_all
 
--- a/tests/library/dbg/tests/test_context_commands.py
+++ b/tests/library/dbg/tests/test_context_commands.py
@@ -490,38 +490,43 @@ async def test_context_all_sections_flag(ctrl: Controller) -> None:
     Tests that context -a/--all shows all sections regardless of context-sections config.
     """
     await launch_to(ctrl, REFERENCE_BINARY, "main")
+    # We need to start the binary using start command
+    await ctrl.execute("start")
 
     # Get default context output to establish baseline
     await ctrl.execute("set context-sections regs disasm code stack backtrace")
     default_out = await ctrl.execute_and_capture("context")
 
-    # Count section headers in default output
-    default_sections = default_out.count("[ ")
+    for section in ["REGISTERS", "DISASM", "STACK", "BACKTRACE", "SOURCE (CODE)"]:
+        assert f"[ {section} " in default_out
+        assert "LAST SIGNAL" not in default_out
 
-    # Now restrict to minimal config
-    await ctrl.execute("set context-sections regs")
-    minimal_out = await ctrl.execute_and_capture("context")
-    minimal_sections = minimal_out.count("[ ")
+    # Now use -a flag. It should capture LAST SIGNAL 
 
-    # Minimal should have fewer sections
-    assert minimal_sections < default_sections
-
-    # Now use -a flag - should show all sections despite minimal config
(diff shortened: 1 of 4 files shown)


tests/library/dbg/tests/test_context_commands.py
```diff
@@ -484,6 +484,46 @@ async def test_context_hide_sections(ctrl: Controller) -> None:
...
+    Tests that context -a/--all shows all sections regardless of context-sections config.
+    """
+    await launch_to(ctrl, REFERENCE_BINARY, "main")
+
+    # Get default context output to establish baseline
+    await ctrl.execute("set context-sections regs disasm code stack backtrace")
+    default_out = await ctrl.execute_and_capture("context")
+
+    # Count section headers in default output
+    default_sections = default_out.count("[ ")
```
disconnect3d: That's not the best way to count sections, likely :D

disconnect3d: i.e. disasm output can have `[sth]`

disconnect3d: Can we make this more robust and ensure that each of the requested section is actually there instead?

D3LET3-57: i tried to add a better version for testing ctx -a

tests/library/dbg/tests/test_context_commands.py
```diff
@@ -484,6 +485,53 @@ async def test_context_hide_sections(ctrl: Controller) -> None:
...
+
+    for section in ["DISASM", "STACK", "BACKTRACE", "SOURCE (CODE)"]:
+        assert f"[ {section} " not in default_out
+
+    # Now use -a flag. It should capture all sections regardless of config.
+    all_out = await ctrl.execute_and_capture("context -a")
+    for section in ["REGISTERS", "DISASM", "STACK", "BACKTRACE", "SOURCE (CODE)"]:
+        assert f"[ {section} " in all_out
+
+    # Now proceed to next function call (i.e at func-with_args) have different context output and ensure arguments are present
```
disconnect3d: Is this code vibe coded or what? It is really not great. This comment says "and ensure arguments are present" which is not true...

tests/library/dbg/tests/test_context_commands.py
```diff
@@ -484,6 +485,53 @@ async def test_context_hide_sections(ctrl: Controller) -> None:
...
+    await ctrl.execute("nextcall")
+
+    default_out_after_nextcall = await ctrl.execute_and_capture("context")
+    for section in ["DISASM", "STACK", "BACKTRACE", "SOURCE (CODE)"]:
+        assert f"[ {section} " not in default_out_after_nextcall
+        assert "ARGUMENTS" not in default_out_after_nextcall
```
disconnect3d: This assert is executed on each iteration. Why this isn't in the for ... items?

tests/library/dbg/tests/test_context_commands.py
```diff
@@ -484,6 +485,53 @@ async def test_context_hide_sections(ctrl: Controller) -> None:
...
+    await ctrl.execute("set context-sections regs")
+    default_out = await ctrl.execute_and_capture("context")
+
+    for section in ["DISASM", "STACK", "BACKTRACE", "SOURCE (CODE)"]:
+        assert f"[ {section} " not in default_out
+
+    # Now use -a flag. It should capture all sections regardless of config.
+    all_out = await ctrl.execute_and_capture("context -a")
+    for section in ["REGISTERS", "DISASM", "STACK", "BACKTRACE", "SOURCE (CODE)"]:
+        assert f"[ {section} " in all_out
```
disconnect3d: This does not check for all context sections, does it ... ? Compare this to lines 516-525 ... The two asserts should be literally the same

Please make a variable like `all_sections = (a,b,c,d)` and use it to ensure all sections.

tests/library/dbg/tests/test_context_commands.py
```diff
@@ -484,6 +485,53 @@ async def test_context_hide_sections(ctrl: Controller) -> None:
...
+        "DISASM",
+        "STACK",
+        "BACKTRACE",
+        "SOURCE (CODE)",
+        "ARGUMENTS",
+    ]:
+        assert f"[ {section} " in all_out_after_nextcall
+
+    # Now use -a flag - should show all sections
+    all_out = await ctrl.execute_and_capture("context -a")
```
disconnect3d: If we want to double check alias for  the flag, we can also do that for the command too
```suggestion
    all_out = await ctrl.execute_and_capture("ctx -a")
```

tests/library/dbg/tests/test_context_commands.py
D3LET3-57: no what i meant was "[ ARGUMENT" is present in ctx -a

tests/library/dbg/tests/test_context_commands.py
```diff
@@ -484,6 +485,53 @@ async def test_context_hide_sections(ctrl: Controller) -> None:
...
+    Tests that context -a/--all shows all sections regardless of context-sections config.
+    """
+    await launch_to(ctrl, CONTEXT_ARGS_BINARY, "main")
+
+    # First, set context-sections to only regs
+    await ctrl.execute("set context-sections regs")
+    default_out = await ctrl.execute_and_capture("context")
+
+    for section in ["DISASM", "STACK", "BACKTRACE", "SOURCE (CODE)"]:
+        assert f"[ {section} " not in default_out
```
disconnect3d: Yet, we didn't even check if regs section is there...

Why don't we do something like:
- find all instances in the output lines for `─[ [A-Z() ]+ \]─` regular expression
- check if they match an exact list

Each assertion here should do that so that it is super robust and can't cause regression in the future.

disconnect3d (changes requested):
See comments

tests/library/dbg/tests/test_context_commands.py
D3LET3-57: sorry i will add it section
i had some problem in capturing LAST SIGNAL in ctx -a, due to that i had done some changes from before commit

disconnect3d: See https://github.com/pwndbg/pwndbg/pull/3580#discussion_r2664588773

tests/library/dbg/tests/test_context_commands.py
D3LET3-57: i couldnt capture all sections like theads, last signal, and heap_tracker but i checked that it works fine manually.
I am getting issue with lldb while capturing ctx for last signal

D3LET3-57:
as you requested i can apply regex and find all matches with it and add them to list and assert it. I am facing issue with last signal section it appears in gdb but not in lldb how to fix it?

disconnect3d:
> as you requested i can apply regex and find all matches with it and add them to list and assert it. I am facing issue with last signal section it appears in gdb but not in lldb how to fix it?

Yeah this section is only available in GDB:
```py
1722 if pwndbg.dbg.is_gdblib_available():
1723     gdb.events.cont.connect(save_signal)
1724     gdb.events.stop.connect(save_signal)
1725     gdb.events.exited.connect(save_signal)
```

So you have to do something like this in the test:

```py
    expected_all_sections.append(<add last signal section here>)
```

tests/library/dbg/tests/test_context_commands.py
```diff
@@ -485,51 +485,74 @@ async def test_context_hide_sections(ctrl: Controller) -> None:
...
-    for section in ["DISASM", "STACK", "BACKTRACE", "SOURCE (CODE)"]:
-        assert f"[ {section} " not in default_out_after_nextcall
-        assert "ARGUMENTS" not in default_out_after_nextcall
-
     # Now use -a flag - should show all sections including ARGUMENTS.
-    all_out_after_nextcall = await ctrl.execute_and_capture("context -a")
-    for section in [
+    all_out_after_nextcall = await ctrl.execute_and_capture("ctx -a")
+    expected_all_with_args = [
+        "ARGUMENTS",
```
disconnect3d: This only appears on a call? Huh... TIL.

Shall we maybe just use the previous `expected_all` here and do `expected_all.insert(0, "ARGUMENTS")`?

disconnect3d: fixed, good to go

D3LET3-57: Thanks!

disconnect3d (approved):

PR #3581, opened 2026-01-04
[docs] Add documentation for cross-arch testing

k4lizen:

image

...

### DIFF
--- a/docs/contributing/testing-crossarch.md
+++ b/docs/contributing/testing-crossarch.md
@@ -55,7 +55,7 @@ Now we run the image using qemu-system:
 sudo nix run github:patryk4815/kernel#vm-aarch64-linux \
     --accept-flake-config --extra-experimental-features flakes \
     --extra-experimental-features nix-command \
-    -- -i pwndbg:aarch64 --refresh
+    -- -i pwndbg:aarch64
 ```
 If you want to use more CPUs, you can pass `--cpus N`. If you have enough RAM you should also consider passing `--runtime tmpfs`. After it has started:
 ```{.bash .copy}
@@ -67,4 +67,4 @@ pwndbg
 ```
 And you are running pwndbg inside of another architecture!
 
-If you make some changes, you need to exit qemu-system and rebuild the docker image (by rerunning the previously shown command).
+If you make some changes to Pwndbg on your host machine that you want to be reflected in the guest, you need to exit qemu-system, rebuild the docker image (by rerunning the previously shown command), and rerun the nix command **with the `--refresh` flag**. If you need to communicate with the host from the guest, don't forget that QEMU exposes the host at IP address 10.0.2.2 .
--- a/docs/contributing/testing-crossarch.md
+++ b/docs/contributing/testing-crossarch.md
@@ -67,4 +67,4 @@ pwndbg
 ```
 And you are running pwndbg inside of another architecture!
 
-If you make some changes, you need to exit qemu-system and rebuild the docker image (be rerunning the previously shown command).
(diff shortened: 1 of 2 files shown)


docs/contributing/testing-crossarch.md
```diff
@@ -0,0 +1,54 @@
...
+    --extra-experimental-features nix-command \
+    -- -i debian:13 -v /your/path/to/pwndbg:/pwndbg
+```
+If you want to use more CPUs, you can pass `--cpus N`. If you have enough RAM you should also consider passing `--runtime tmpfs`. After it has started, we can:
+```{.bash .copy}
+bash
+cd /mnt
+./setup.sh
+```
+to install Pwndbg inside. Depending on your use-case, you may want to run `./setup-dev.sh` as well. You should expect this to be relatively slow. Note that this shell is transient, if you exit it, you will need to re-run the setup.
```
patryk4815: >  You should expect this to be relatively slow.

There is a faster way: build a Docker image with qemu-user and then use that image together with qemu-system.
Building the image with qemu-user is much faster than running `./setup-dev.sh` inside qemu-system.

For example:
```
docker build --platform linux/aarch64 -f ./Dockerfile.dnf -t pwndbg-local:latest .
nix run github:patryk4815/kernel#vm-aarch64-linux -- -i pwndbg-local:latest
```

k4lizen: yoo awesome, I didn't know!

docs/contributing/testing-crossarch.md
```diff
@@ -0,0 +1,54 @@
...
+
+If you just want to test a userspace binary on a different architecture, but are fine with GDB running on your host architecture, your best bet is qemu-user.
+
+Say for instance, you would like to debug the `tests/library/dbg/tests/test_command_telescope.py` test on aarch64. First we need to compile `tests/binaries/host/telescope_binary.native.c` for aarch64. The easiest way to do this is with zig:
+```{.bash .copy}
+zig cc tests/binaries/host/telescope_binary.native.c \
+    --target=aarch64-linux-musl -static \
+    -o ./tele-aarch64 
+```
+We compile it with `-static` because I don't want to install `/lib/ld-linux-aarch64.so.1` which is required to run aarch64 binaries on my x86_64 system. We use `-musl` instead of `-gnu` because glibc does not support static linking.
```
Xeonacid:
```suggestion
We compile it with `-static` because I don't want to install `/lib/ld-linux-aarch64.so.1` which is required to run aarch64 binaries on my x86_64 system. We use `-musl` instead of `-gnu` because zig does not support static linking with glibc.
```

That's a limit of zig. `aarch64-linux-gnu-gcc -static` works fine.

docs/contributing/testing-crossarch.md
Xeonacid: Anyway to use kvm accel? (Maybe a feature request for https://github.com/patryk4815/kernel :P)

docs/contributing/testing-crossarch.md
k4lizen: its not a "limit" of zig, they specifically disallow it.

https://stackoverflow.com/questions/57476533/why-is-statically-linking-glibc-discouraged

docs/contributing/testing-crossarch.md
patryk4815: > Anyway to use kvm accel?

Yes, but the intention of the project was to emulate other architectures, "kvm accel" don't work with that

k4lizen:
okay I tested this, it works, I'm satisfied and I think its mergeable

disconnect3d:
hm maybe i merged it too fast, but @patryk4815 let us know if u have any more comments xD

PR #3539, opened 2025-12-25
Enable mypy linting on all lint files

k4lizen:
Previously this mostly wasn't done because of ida_script.py and binja_script.py. Still need to figure out what the docs error is though.

Requires: https://github.com/pwndbg/pwndbg/pull/3538 since that removes those two files.
Closes: https://github.com/pwndbg/pwndbg/issues/3493

PR #3583, opened 2026-01-05
Enable mypy lint on all files

AalbatrossGuy:
Following the conversation in this [PR](https://github.com/pwndbg/pwndbg/pull/3539), the error message said that mypy was locating the same source file in two different locations, `_docs.command_docs_common` & `scripts._docs.command_docs_common`. The issue was that mypy was detecting `scripts/_docs` as a top-level package directory which made it find the same file in two different locations. The fix was to add `__init__.py` to `scripts/` and re-run the linter using `lint.sh`.


pyproject.toml
```diff
@@ -36,6 +36,7 @@ dependencies = [
     "decomp2dbg==3.14.0",
     # Optional? Used for getting symbols into the debugger with the decompiler synchronization
     "lief>=0.17.1 ; (platform_system == 'Linux' and (platform_machine == 'x86_64' or platform_machine == 'aarch64' or platform_machine == 'i686')) or platform_system != 'Linux'",
+    "mypy>=1.18.2",
```
k4lizen: you shouldn't do this, users of pwndbg don't need to have mypy installed, mypy is listed in the lint dependancy group. this is also why you have the check lock files errors

k4lizen (changes requested):

pyproject.toml
AalbatrossGuy: Ah okay, I'll change it

AalbatrossGuy: I removed that from the dependency part. Still failing? Am I missing something?

k4lizen: ye you didn't revert the `uv.lock` file change

AalbatrossGuy: Ahhh my bad. Forgot bout it sigh

.github/workflows/lint.yml
```diff
@@ -12,22 +12,20 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # We perform lint on the lowest ubuntu version we support to allow mypy
-        # to catch backwards-incompatibility bugs (when vermin fails to, like #3557).
-        os: [ubuntu-22.04]
```
k4lizen, line 17: whats up with these changes? you commited them as if they are in my PR but they arent (https://github.com/pwndbg/pwndbg/commit/3576fb9b39b50d1ab6169aef4938d87d387af235)

AalbatrossGuy: I think I pulled the whole lint.yml file from your PR. Should've pulled only the changes. My bad, will fix it asap

k4lizen: you should probably use `git cherry-pick 3576fb9` on your branch to add my commit (but you will need to have pulled my fork of pwndbg for that
```
git remote add k4lizen https://github.com/k4lizen/pwndbg
git fetch k4lizen
```)

Comment (AalbatrossGuy) [reply .github/workflows/lint.yml:17]: didn't work, hence I had to manually do it. I'll look into it. sorry for the bad PR

Comment (AalbatrossGuy) [reply .github/workflows/lint.yml:17]: you can check it out now. Fixed it

Comment (AalbatrossGuy) [reply .github/workflows/lint.yml:17]: > you should probably use `git cherry-pick 3576fb9` on your branch to add my commit (but you will need to have pulled my fork of pwndbg for that
> 
> ```
> git remote add k4lizen https://github.com/k4lizen/pwndbg
> git fetch k4lizen
> ```)
> ```

Did that, didn't work. Idk why, weird. First time encountering this problem. Never had this before

Comment (k4lizen) [reply .github/workflows/lint.yml:17]: this PR still deletes that comment ^^^ (and it shouldn't)

**Review by k4lizen — CHANGES_REQUESTED** [review]

**Comment by k4lizen** [comment]
after that is fixed I'm down to merge this. @disconnect3d The lint will inevitably fail for this PR because we are enabling `mypy --strict` on more files so there will be more errors

**Comment by AalbatrossGuy** [comment]
All's good now ig

File: lint.sh [hunk 3]
```diff
@@ -69,5 +69,6 @@ $UV_RUN_LINT vermin -vvv --no-tips -t=3.10- --eval-annotations --violations ${LI
 
 # mypy is run in a separate step on GitHub Actions
 if [[ -z "$GITHUB_ACTIONS" ]]; then
-    $UV_RUN_MYPY mypy pwndbg pwndbginit tests/host
+    # $UV_RUN_MYPY mypy pwndbg pwndbginit tests/host
```
k4lizen:
```suggestion
```

k4lizen (approved):

jxuanli:
maybe we should just run it against changed files? if we could display the diff between dev and the pr branch that would be really helpful as well i think

mypy takes ~1.5 min to run on my machine which can be a bit annoying

k4lizen:
> maybe we should just run it against changed files?

I don't think thats viable because you can cause errors in file X by touching file Y.

>  if we could display the diff between dev and the pr branch that would be really helpful as well i think

Yeah I agree. Should be trivial.

> mypy takes ~1.5 min to run on my machine which can be a bit annoying

Same, I feel it wasn't this slow before, maybe there was a perf regression? Maybe we could look to migrate to `ty`, and yeah we should look to do something about `./lint.sh -f`.

Regardless I think this work is for another PR.

PR #3585, opened 2026-01-05
Make pwndbg.dbg.selected_inferior() raise an exception instead of returning None

k4lizen:
There are not a lot of places which actually need to check this, because most of our commands only run when the process is alive. There isn't a good way to embed this assumption into the type system that I know of, so lets move the burden to catching the exception.

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -26,6 +26,7 @@ import pwndbg.aglib.heap
 import pwndbg.aglib.kernel
 import pwndbg.aglib.proc
 import pwndbg.aglib.qemu
+import pwndbg.aglib.symbol
 import pwndbg.aglib.typeinfo
 import pwndbg.color.message as message
 import pwndbg.dbg_mod
@@ -440,7 +441,9 @@ class CommandObj:
 
     def invoke(self, argument: str, from_tty: bool) -> None:
         """Invoke the command with an argument string"""
-        if not pwndbg.dbg.selected_inferior():
+        try:
+            _ = pwndbg.dbg.selected_inferior()
+        except pwndbg.dbg_mod.NoInferior:
             log.error("Pwndbg commands require a target binary to be selected")
             return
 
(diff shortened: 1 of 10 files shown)


pwndbg/commands/__init__.py
```diff
@@ -440,7 +441,9 @@ def initialize_parser(self) -> None:
 
     def invoke(self, argument: str, from_tty: bool) -> None:
         """Invoke the command with an argument string"""
-        if not pwndbg.dbg.selected_inferior():
+        try:
+            _ = pwndbg.dbg.selected_inferior()
+        except pwndbg.dbg_mod.NoInferior:
             log.error("Pwndbg commands require a target binary to be selected")
```
k4lizen, line 447: (tangent)

btw in GDB selected_inferior is always existant, but in LLDB the process actually needs to be alive for it to be existant.

So I'm not sure that this if-statement is actually good, because it is effectively enforcing @OnlyWhenRunning on every LLDB command. Also the error message looks really ugly.

k4lizen:
pyright has had my editor red all over because of this typing issue btw:
image

but it seems mypy is living its best life:
```
~❯ uv run mypy --strict pwndbg/aglib/kernel/vmmap.py
pwndbg/aglib/kernel/vmmap.py: note: In class "QemuMachine":
pwndbg/aglib/kernel/vmmap.py:94: error: Class cannot subclass "Machine" (has type "Any")  [misc]
    class QemuMachine(Machine):
                      ^~~~~~~
pwndbg/aglib/kernel/vmmap.py: note: In member "__init__" of class "QemuMachine":
pwndbg/aglib/kernel/vmmap.py:95: error: Function is missing a return type annotation  [no-untyped-def]
        def __init__(self):
        ^
pwndbg/aglib/kernel/vmmap.py:95: note: Use "-> None" if function does not return a value
pwndbg/aglib/kernel/vmmap.py:98: error: Call to untyped function "get_qemu_pid" in typed context  [no-untyped-call]
            self.pid = QemuMachine.get_qemu_pid()
                       ^~~~~~~~~~~~~~~~~~~~~~~~~~
pwndbg/aglib/kernel/vmmap.py: note: In member "__del__" of class "QemuMachine":
pwndbg/aglib/kernel/vmmap.py:101: error: Function is missing a type annotation  [no-untyped-def]
        def __del__(self):
        ^
pwndbg/aglib/kernel/vmmap.py: note: In member "get_qemu_pid" of class "QemuMachine":
pwndbg/aglib/kernel/vmmap.py:123: error: Function is missing a return type annotation  [no-untyped-def]
        def get_qemu_pid():
        ^
pwndbg/aglib/kernel/vmmap.py: note: In function "kernel_vmmap_via_page_tables":
pwndbg/aglib/kernel/vmmap.py:186: error: Call to untyped function "QemuMachine" in typed context  [no-untyped-call]
            machine_backend = QemuMachine()
                              ^~~~~~~~~~~~~
Found 6 errors in 1 file (checked 1 source file)
```
god knows what mypy is doing :S

jxuanli:
@k4lizen its just complaining things are not typed but i didnt bother to type those cuz `gdb-pt-dump` should be retired/removed in the near future (cc @patryk4815)

k4lizen:
ye I just pasted the mypy output to show that it is *not* showing the "send_monitor() is not a known attribute of None" error on the dev branch, whereas it surely should (though its not gonna matter after this is merged)

disconnect3d (approved):

PR #3592, opened 2026-01-06
Show mypy --strict diff in CI

k4lizen:
Should make it easier for contributors to spot the problem. re: https://github.com/pwndbg/pwndbg/pull/3583#issuecomment-3712791126


disconnect3d (approved):

k4lizen:
@disconnect3d i think you have to merge this since it requires bypassing branch protection

Issue #3591, opened 2026-01-06
separate lint.sh to --format and --all


k4lizen:
Currently running mypy takes a minute and a half on my machine. Sometimes I want to just run the formatting and not run mypy, so the lint.sh script should support that. If only formatting is run it should print a clear message that mypy wasn't run.

k4lizen:
related: https://github.com/pwndbg/pwndbg/pull/3583#issuecomment-3712791126

disconnect3d:
Done in https://github.com/pwndbg/pwndbg/pull/3594 !

PR #3594, opened 2026-01-07
Refactor lint.sh: Flag separation for skipping vermin/mpy, and improved error messaging

jackmisbach:
vmware_GjSAWAkuES

vmware_YfbMS2LU6H

Issue https://github.com/pwndbg/pwndbg/issues/3591

Just a few questions I want to ask before I update the docs
1. Do we want a flag to be required? Like, should --format or --all be required? As of right now, the default without any flags is just all. Would need to update the lint.yml workflow if it required a flag 
2. Should mypy be the only difference between --format and --all as I have it now, or do we want them more accurate to the naming and have --format only run ruff format and shfmt? and then --all runs mypy, vermin, and ruff check?
currently I have it as shown in the image with the only difference being mypy, so if we want to keep it that way im going to update the help command to say "Use --format for skipping mypy (type checking)" or something instead of "quick formatting-only"
Because mypy is ofc the main time consumer, followed by vermin, and then ruff check is very fast, just wondering because naming wouldn't match up then because --format also runs ruff check, a linter, and vermin an analyzer for example

### DIFF
--- a/lint.sh
+++ b/lint.sh
@@ -91,7 +91,7 @@ else
         echo "========================================="
         echo "ERROR: Formatting issues detected by ruff."
         echo "       Exiting early. All checks were NOT run."
-        echo "       Use -f or --fix-and-check to fix issues automatically."
+        echo "       Use -f to fix issues automatically."
         echo "========================================="
         exit 1
     fi
--- a/lint.sh
+++ b/lint.sh
@@ -7,32 +7,41 @@ source "$(dirname "$0")/scripts/common.sh"
 cd $PWNDBG_ABS_PATH
 
 help_and_exit() {
-    echo "Usage: ./lint.sh [-f|--fix] [--format] [--all]"
-    echo "  -f,  --fix         fix issues if possible"
-    echo "  --format           run only ruff, shfmt, and vermin checks (skip mypy)"
-    echo "  --all              run all checks including mypy (default behavior)"
+    echo "Usage: ./lint.sh [--check | -fo|--fix-only | -f|--fix-and-check]"
+    echo "  --check                 run all checks without applying fixes (default behavior)"
+    echo "  -fo, --fix-only         fix formatting only, without running checks"
+    echo "  -f,  --fix-and-check    fix formatting first, then run checks"
     echo ""
-    echo "By default, all checks are run.  Use --format for a quick formatting-only check."
+    echo "By default, all checks are run. Fixes are not applied unless specified."
     exit 1
 }
 
-FIX=0
-FORMAT_ONLY=0
+if [[ $# -gt 1 ]]; then
+    help_and_exit
+fi
+
+CHECK_ONLY=1
+FIX_ONLY=0
+FIX_AND_CHECK=0
 
 while [[ $# -gt 0 ]]; do
     case $1 in
-        -f | --fix)
-            FIX=1
+        --check)
+            CHECK_ONLY=1
+            FIX_ONLY=0
+            FIX_AND_CHECK=0
             shift
             ;;
-        --format)
-            FORMAT_ONLY=1
+        -fo | --fix-only)
+            CHECK_ONLY=0
+            FIX_ONLY=1
+            FIX_AND_CHECK=0
             shift
             ;;
-        --all)
-            # Explicitly run all checks (default behavior)
-            # do we want to require --format or --all instead?
-            FORMAT_ONLY=0
+        -f | --fix-and-check)
+            CHECK_ONLY=0
+            FIX_ONLY=0
+            FIX_AND_CHECK=1
             shift
             ;;
         *)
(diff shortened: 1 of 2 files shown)


jackmisbach:
Also, before it would run help_and_exit if two flags were specified; should we do something like this now, making them mutually exclusive? just as an edge case in case someone tried to run both... idk
```
while [[ $# -gt 0 ]]; do
    case $1 in
        -f | --fix)
            FIX=1
            shift
            ;;
        --format)
            if [[ $ALL_FLAG == 1 ]]; then
                echo "Error: --format and --all are mutually exclusive"
                help_and_exit
            fi
            FORMAT_ONLY=1
            shift
            ;;
        --all)
            if [[ $FORMAT_ONLY == 1 ]]; then
                echo "Error: --format and --all are mutually exclusive"
                help_and_exit
            fi
            ALL_FLAG=1
            shift
            ;;
        *)
            help_and_exit
            ;;
    esac
done
  ```

**Comment by k4lizen** [comment]
Okay so as you say, the only difference between format and all is mypy, which doesn't really make sense. I kind of want to support three use-cases:
1. Run all checks (what `./lint.sh` currently does).
2. Only run the formatting and don't run anything that won't modify the code (so e.g. don't run vermin here).
3. Run the formatting and all checks (what `./lint.sh -f` currently does).

On further thought, I think that probably the best way to support this is:
1. `--check` flag, that is the same as not passing any arguments to `./lint.sh`
2. `--fix-only` flag, shorthand `-fo`
3. `--fix-and-check`, shorthand `-f`

So no `--format` or `--all` flag. This way, we are "backwards-compatible" with the existing workflow, and would require minimal workflow/docs changes, if any.

**Comment by jackmisbach** [comment]
Alright, I agree with that and went ahead and made those changes.
Also added error messages that will show up if ruff or shfmt find issues without -f or -fo, and it exits early, making it more clear that there could be more issues and checks weren't finished. 
vmware_aCJjgWZ4ku
vmware_STdc37wjHX
Should I maybe say "Use -f or -fo to fix issues automatically" instead, or leave it as is?

And then, if all formatting passes, I updated the --check (default) warning message to say vermin and mypy were NOT run
vmware_54VubxHt0w

For the docs, do you want me to update them or leave them?
The only thing I would add is maybe that you can run -fo to skip mypy and vermin and run a quicker lint

File: lint.sh [hunk 1]
```diff
@@ -62,11 +114,9 @@ else
     fi
 
     $UV_RUN_LINT ruff check --output-format="${RUFF_OUTPUT_FORMAT}" ${LINT_FILES}
+    $UV_RUN_LINT vermin -vvv --no-tips -t=3.10- --eval-annotations --violations ${LINT_FILES}
 fi
 
-# Checking minimum python version
-$UV_RUN_LINT vermin -vvv --no-tips -t=3.10- --eval-annotations --violations ${LINT_FILES}
```
k4lizen, line 68: can this line stay here?

lint.sh
```diff
@@ -47,13 +67,45 @@ call_shfmt() {
...
 else
-    $UV_RUN_LINT ruff format --check --diff ${LINT_FILES}
-    call_shfmt
+    if ! $UV_RUN_LINT ruff format --check --diff ${LINT_FILES}; then
+        set +o xtrace
+        echo ""
+        echo "========================================="
+        echo "ERROR: Formatting issues detected by ruff."
+        echo "       Exiting early. All checks were NOT run."
+        echo "       Use -f or --fix-and-check to fix issues automatically."
```
k4lizen:
```suggestion
        echo "       Use -f to fix issues automatically."
```
Use `A` or `B` reads to me a bit like they are different flags, when one is just an alias for the other. I think its fine not to mention `-fo` since if you got to this code branch, you want the check to run.

k4lizen:
```suggestion
        echo "       Use -f to fix issues automatically."
```
Use `A` or `B` reads to me a bit like they are different flags, when one is just an alias for the other. I think its fine not to mention `-fo` since if you got to this code branch, you want the check to run.

lint.sh
```diff
@@ -47,13 +67,45 @@ call_shfmt() {
...
+        exit 1
+    fi
+
+    if ! call_shfmt; then
+        set +o xtrace
+        echo ""
+        echo "========================================="
+        echo "ERROR: Formatting issues detected by shfmt."
+        echo "       Exiting early. All checks were NOT run."
+        echo "       Use -f or --fix-and-check to fix issues automatically."
```
k4lizen:
```suggestion
        echo "       Use -f to fix issues automatically."
```

lint.sh
jackmisbach: Yes, I just removed it because I didn't know which vermin run to put it on because it is under fix and check, and default (check) but I'll just leave it down there on that last one

k4lizen (approved):

k4lizen:
>  The only thing I would add is maybe that you can run -fo to skip mypy and vermin and run a quicker lint

yeah you can add it to the NOTE thing here

https://pwndbg.re/dev/contributing/#linting

lint.sh
jackmisbach: yeah I agree,
and mmm yeah ur right if you did get to that code branch you want vermin and mypy, I'll make it just -f

jackmisbach:
Alright, should be all good. lmk if what I put in the docs is fine or if you want to word it differently
I put
You can also run `./lint.sh -fo` to skip vermin and mypy, allowing for a quicker lint.

k4lizen (approved):
LGTM! Thanks!

(I can't merge it rn due to the lock file stuff (which is not your fault))

PR #3600, opened 2026-01-07
lint.sh: error out on aglib import + fix ipi in gdb

disconnect3d:
1. Implements a trivial detection of aglib imports in pwndbg/lib/ which should be banned.
2. Fixes `ipi` command in GDB.
3. Improves the `ipi` help/aliases info and refactors some of them to be trivial alias for aglib.* funcs

image


lint.sh
```diff
@@ -64,6 +64,14 @@ else
     $UV_RUN_LINT ruff check --output-format="${RUFF_OUTPUT_FORMAT}" ${LINT_FILES}
 fi
 
+# Check that pwndbg/lib does not import from pwndbg.aglib
+echo "Checking for aglib usage in pwndbg/lib..."
+if grep -rE "(import .*\.aglib|from .*\.aglib)" pwndbg/lib/; then
+    echo "Error: pwndbg/lib must not import from pwndbg.aglib"
+    echo "The 'lib' module is a low-level library and cannot depend on 'aglib'"
+    exit 1
+fi
```
k4lizen, line 127: nice.

I think we should pull this out to a seperate file (like `scripts/custom-lint-rules.sh` or `scripts/custom-lint-rules.py`) so we can keep adding stuff without bloating lint.sh

k4lizen: This would also be nice:
+ Link the relevant doc: https://pwndbg.re/dev/contributing/common-pitfalls/#pwndbglib-files-should-only-access-pwndbglib
+ Add a check for pwndbg.dbg. and pwndbg.dbg_mod as well
+ Use ripgrep if it is available on the system

k4lizen: We should probably also grep for `pwndbg.aglib` in the code in general, rather than just in the import

disconnect3d: It is "used", well, referenced in tips.py ;D

PR #3602, opened 2026-01-08
Fix lockfile issues caused by pytest-cov update

jackmisbach:

Dependabot updating pytest-cov from 4.1.0 to 7.0.0 caused issues because pytest-cov no longer uses setuptools, they use hatchling. https://github.com/pytest-dev/pytest-cov/commit/fa72cf3b21c30599eb1793d3f769e1b7090e76da#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711

This fixes the issue by updating what packages need hatchling to include pytest-cov and hatch-fancy-pypi-readme. Also removes pytest-cov from requiring setuptools.

### DIFF
--- a/nix/pyenv.nix
+++ b/nix/pyenv.nix
@@ -77,7 +77,6 @@ let
     "coverage"
     "mypy-extensions"
     "pytest"
-    "pytest-cov"
     "mypy"
     "vermin"
     # decomp2dbg deps
@@ -110,6 +109,8 @@ let
     "plumbum"
     "rpyc"
     "iniconfig"
+    "pytest-cov"
+    "hatch-fancy-pypi-readme"
     # decomp2dbg deps
     "decomp2dbg"
     "filelock"
@@ -140,7 +141,7 @@ let
     final: prev:
     (genPkgsNeeded pkgsNeedSetuptools [ "setuptools" ] final prev)
     // (genPkgsNeeded pkgsNeedFlitcore [ "flit-core" ] final prev)
-    // (genPkgsNeeded pkgsNeedHatchling [ "hatchling" "hatch-vcs" ] final prev)
+    // (genPkgsNeeded pkgsNeedHatchling [ "hatchling" "hatch-vcs" "hatch-fancy-pypi-readme" ] final prev)
     // (genPkgsNeeded pkgsNeedPoetry [ "poetry-core" ] final prev);
 
   dummy = pkgs.runCommand "dummy" { } "mkdir $out";
@@ -153,11 +154,6 @@ let
 
     # ziglang is only supported on few platforms
     ziglang = prev.ziglang.override {
-      sourcePreference = "wheel";
-    };
-
-    # use wheel to avoid build system issues with pytest-cov 7.0.0
-    pytest-cov = prev.pytest-cov.override {
       sourcePreference = "wheel";
     };
 
--- a/nix/pyenv.nix
+++ b/nix/pyenv.nix
@@ -156,6 +156,11 @@ let
       sourcePreference = "wheel";
     };
 
+    # use wheel to avoid build system issues with pytest-cov 7.0.0
+    pytest-cov = prev.pytest-cov.override {
+      sourcePreference = "wheel";
+    };
+
     psutil = pkgs.callPackage (
       {
         darwin,
--- a/nix/pyenv.nix
+++ b/nix/pyenv.nix
@@ -110,7 +110,6 @@ let
     "rpyc"
     "iniconfig"
     "pytest-cov"
-    "hatch-fancy-pypi-readme"
     # decomp2dbg deps
     "decomp2dbg"
     "filelock"


nix/pyenv.nix
```diff
@@ -140,7 +141,7 @@ let
     final: prev:
     (genPkgsNeeded pkgsNeedSetuptools [ "setuptools" ] final prev)
     // (genPkgsNeeded pkgsNeedFlitcore [ "flit-core" ] final prev)
-    // (genPkgsNeeded pkgsNeedHatchling [ "hatchling" "hatch-vcs" ] final prev)
+    // (genPkgsNeeded pkgsNeedHatchling [ "hatchling" "hatch-vcs" "hatch-fancy-pypi-readme" ] final prev)
```
patryk4815, line 143: `"hatch-fancy-pypi-readme"` is needed here?

jackmisbach: Unfortunately
I tested without first and it doesn't work because pytest-cov 7.0.0 needs both

jackmisbach:
```error: Cannot build '/nix/store/knics0z30qmc2j1707p0cwavl4mylwkk-pytest-cov-7.0.0.drv'.
       Reason: builder failed with exit code 2.
       Output paths:
         /nix/store/8n5xapx5j9i05axnmw4412waa7hx6aj2-pytest-cov-7.0.0
       Last 25 log lines:
       >                             ~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
       >   File "/nix/store/06b4r1qjybmkxlcg561sakbrxdh3immp-hatchling-1.27.0/lib/python3.13/site-packages/hatchling/builders/plugin/interface.py", line 90, in build
       >     self.metadata.validate_fields()
       >     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
       >   File "/nix/store/06b4r1qjybmkxlcg561sakbrxdh3immp-hatchling-1.27.0/lib/python3.13/site-packages/hatchling/metadata/core.py", line 265, in validate_fields
       >     _ = self.version
       >         ^^^^^^^^^^^^
       >   File "/nix/store/06b4r1qjybmkxlcg561sakbrxdh3immp-hatchling-1.27.0/lib/python3.13/site-packages/hatchling/metadata/core.py", line 149, in version
       >     self._version = self._get_version()
       >                     ~~~~~~~~~~~~~~~~~^^
       >   File "/nix/store/06b4r1qjybmkxlcg561sakbrxdh3immp-hatchling-1.27.0/lib/python3.13/site-packages/hatchling/metadata/core.py", line 244, in _get_version
       >     core_metadata = self.core
       >                     ^^^^^^^^^
       >   File "/nix/store/06b4r1qjybmkxlcg561sakbrxdh3immp-hatchling-1.27.0/lib/python3.13/site-packages/hatchling/metadata/core.py", line 187, in core
       >     metadata_hooks = self.hatch.metadata.hooks
       >                      ^^^^^^^^^^^^^^^^^^^^^^^^^
       >   File "/nix/store/06b4r1qjybmkxlcg561sakbrxdh3immp-hatchling-1.27.0/lib/python3.13/site-packages/hatchling/metadata/core.py", line 1587, in hooks
       >     raise UnknownPluginError(message)
       > hatchling.plugin.exceptions.UnknownPluginError: Unknown metadata hook: fancy-pypi-readme
       >   x Failed to build `/build/pytest_cov-7.0.0`
       >   |-> The build backend returned an error
       >   `-> Call to `hatchling.build.build_wheel` failed (exit status: 1)
       >       hint: This usually indicates a problem with the package or the build
       >       environment.
       > DEBUG Released lock at `/build/.tmpmDd1Pn/.lock
 ```
 Here are the tests without it
 , and this is why 
 https://github.com/pytest-dev/pytest-cov/commit/fa72cf3b21c30599eb1793d3f769e1b7090e76da#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R2

jackmisbach:
I accidentally had "hatch-fancy-pypi-readme" in pkgsNeedHatchling and genPkgsNeeded pkgsNeedHatchling, so I removed that.

But yeah, because only pytest-cov needs hatch-fancy-pypi-readme, would it be better to have that as a separate override?
maybe do something like this? so it doesn't install it for all the other hatchling needing pkgs and just for pytest-cov
```  // (genPkgsNeeded pkgsNeedHatchlingWithFancyReadme [ "hatchling" "hatch-vcs" "hatch-fancy-pypi-readme" ] final prev)```
Or we could just install the wheel for pytest-cov, but that's only done for zig and zig has a good reason so idk abt that

disconnect3d:
I'm going to merge this for now just for the sake of CI passing but I'll leave the question about hatch-*pypi-readme to @patryk4815.

Overall, I think we should just remove pytest-cov from deps since we do not use it atm. We did use coverage gathering in the past but we 1) didn't do much with the data, 2) it slowed down CI heavily, 3) we didn't have it set properly so I think results from various runs weren't merged properly.

PR #3603, opened 2026-01-08
commands: enhance stepsyscall to support names, numbers, and expressions.

### LINKED ISSUE #3552: stepsyscall should allow to step until syscall number or until given expression is true

Gitjay11:
Closes #3552

- [yes] I read the contributing documentation.
- [yes] I am providing a screenshot of the new feature.

### Description
Enhanced the [stepsyscall] command to support flexible stopping criteria. Users can now step until a specific syscall name is hit, a syscall number is reached, or a general GDB expression/condition becomes true.

### Proposed Changes
- **Argument Parsing**: Updated the [stepsyscall] (pwndbg/pwndbg/commands/next.py:254:0-265:5) command to accept names, numbers, and expressions.
- **Improved Looping**: Implemented a robust stepping loop that uses break_next_branch (pwndbg/pwndbg/aglib/next.py:117:0-132:18) and instruction re-fetching to ensure accurate matching.
- **GDB Integration**: Implemented GDBValue.__bool__ to support correct conditional evaluation of GDB expressions in Python.

### Verification Results 
All 8 regression tests passed in a WSL Ubuntu environment.
result

### Manual Verification
The following usage patterns are now fully supported and verified: 

1. Stop by Syscall Name (stepsyscall write) The debugger skips all intermediate instructions and stops exactly at the 
write syscall.

``` pwndbg> stepsyscall write
0x00000000004000a0 in label1 ()
0x0000000000400080 in write ()
─────────────────────────────────────────────[ REGISTERS / show-flags off / show-compact-regs off ]─────────────────────────────────────────────
*RAX  1
 RBX  0
 RCX  0
*RDX  0xc
*RDI  1
*RSI  0x401000 (msg) ◂— 'hello world\n'
 R8   0
 R9   0
 R10  0
 R11  0
 R12  0
 R13  0
 R14  0
 R15  0
 RBP  0
 RSP  0x7fffffffdd18 —▸ 0x4000a5 (exit) ◂— mov eax, 0x3c
*RIP  0x400099 (syscall_write_label) ◂— syscall
──────────────────────────────────────────────────────[ DISASM / x86-64 / set emulate on ]──────────────────────────────────────────────────────
   0x400080 <write>                    mov    eax, 1           EAX => 1
   0x400085 <write+5>                  mov    edi, 1           EDI => 1
   0x40008a <write+10>                 movabs rsi, msg         RSI => 0x401000 (msg)
   0x400094 <write+20>                 mov    edx, 0xc         EDX => 0xc
 ► 0x400099 <syscall_write_label>      syscall <SYS_write>
        fd: 1 (/dev/pts/1)
        buf: 0x401000 (msg) ◂— 'hello world\n'
        n: 0xc
   0x40009b <syscall_write_label+2>    ret

   0x40009c <_start>                   nop
   0x40009d <_start+1>                 jmp    label1                      <label1>
    ↓
   0x4000a0 <label1>                   call   write                       <write>

   0x4000a5 <exit>                     mov    eax, 0x3c        EAX => 0x3c
   0x4000aa <exit+5>                   mov    edi, 0           EDI => 0
───────────────────────────────────────────────────────────────────[ STACK ]────────────────────────────────────────────────────────────────────
00:0000│ rsp 0x7fffffffdd18 —▸ 0x4000a5 (exit) ◂— mov eax, 0x3c
01:0008│     0x7fffffffdd20 ◂— 1
02:0010│     0x7fffffffdd28 —▸ 0x7fffffffdfb9 ◂— '/home/nixjay/pwndbg/tests/binaries/host/stepsyscall.x86-64.out'
03:0018│     0x7fffffffdd30 ◂— 0
04:0020│     0x7fffffffdd38 —▸ 0x7fffffffdff8 ◂— 'SHELL=/bin/bash'
05:0028│     0x7fffffffdd40 —▸ 0x7fffffffe008 ◂— 'WSL2_GUI_APPS_ENABLED=1'
06:0030│     0x7fffffffdd48 —▸ 0x7fffffffe020 ◂— 'WSL_DISTRO_NAME=Ubuntu'
07:0038│     0x7fffffffdd50 —▸ 0x7fffffffe037 ◂— 'WT_SESSION=3eec80ab-d31f-4437-be5c-40040814ec37'
─────────────────────────────────────────────────────────────────[ BACKTRACE ]──────────────────────────────────────────────────────────────────
 ► 0         0x400099 syscall_write_label
   1         0x4000a5 exit
   2              0x1 None
   3   0x7fffffffdfb9 None
   4              0x0 None 
```

2. Stop by Syscall Number (stepsyscall 60) The command correctly resolves 60 as the exit syscall on x86-64.
( I can provide the screenshot of result if it is needed.  it can be generated by : stepsyscall 60)

3. Conditional Stop (stepsyscall write $rdi == 1) Execution continues until the 
write syscall targets file descriptor 1 (stdout).
( I can provide the screenshot of result if it is needed )

4. Pure Expression Stop (stepsyscall $rax == 1) Stops at the very next syscall where RAX is 1.

I’ve aimed for a robust and performant implementation that follows the project’s existing patterns. I have also added regression tests and manually verified the logic across several common syscall scenarios. 
I’m happy to make any adjustments or improvements based on your review. Thank you for your time :)


disconnect3d:
What is this:
> GDB Integration: Implemented GDBValue.bool to support correct conditional evaluation of GDB expressions in Python.

What does not work without it?

pwndbg/commands/next.py
```diff
@@ -155,34 +156,114 @@ def nextsyscall() -> None:
     pwndbg.dbg.selected_inferior().dispatch_execution_controller(_nextsyscall)
 
 
-async def _stepsyscall(ec: pwndbg.dbg_mod.ExecutionController):
```
disconnect3d, line 158: I don't like it.

1) Why don't we leave the function as-is and have another one that will take a function like stepsyscall and a condition function and execute them in a loop until the condition is hit? Note that this would allow to implement any step/next* function with a condition that way.

2) The `arch_module={...}.get(..)` part is likely redundant and we should have some way of getting "current arch module" in the codebase already.

disconnect3d:
See also https://github.com/pwndbg/pwndbg/pull/3601#issuecomment-3724047258

Gitjay11:
> What is this:
> > GDB Integration: Implemented GDBValue.bool to support correct conditional evaluation of GDB expressions in Python. 
> What does not work without it?
@disconnect3d  

Without this implementation, conditional checks on GDB values in Python , would always evaluate to True, even for values that represent 0, NULL, or False.
This happens because GDBValue is a wrapper class. In Python, an instance of a class is considered "truthy" by default unless it implements bool (or __len__). Without the bool method, a GDBValue wrapping a pointer to 0x0 (NULL) would still be treated as True because the wrapper object itself exists.

The GDBValue.bool implementation ensures that Python respects the actual underlying GDB value:
Without it: if GDBValue(0): evaluates to True (Incorrect).
With it: if GDBValue(0): evaluates to False (Correct).

I am going to review the changes again:)
let me check the other comments and make changes. Thankyou.

disconnect3d:
> Without this implementation, conditional checks on GDB values in Python , would always evaluate to True, even for values that represent 0, NULL, or False. (...)

I know all of this, but this still realy do not explain why we are getting a GDB value in the first place...

I looked it up. So e.g. for a condition of `$rdi == 1` you do `pwndbg.dbg.selected_inferior().evaluate_expression('$rdi == 1')` and this returns a `pwndbg.dbg_mod.gdb.GDBValue` which has the problem you describe:
```
In [4]: pwndbg.aglib.regs.read_reg("$rdi")
Out[4]: 140737354130144

In [5]: x=pwndbg.dbg.selected_inferior().evaluate_expression('$rdi == 1')

In [6]: x
Out[6]: <pwndbg.dbg_mod.gdb.GDBValue at 0x729f57b8ea20>

In [7]: bool(x)
Out[7]: True

In [8]: int(x)
Out[8]: 0

In [9]: y=pwndbg.dbg.selected_inferior().evaluate_expression('$rdi == 140737354130144')

In [10]: int(y)
Out[10]: 1
```

Well, the current solution has probably two problems:
- its GDB only
- it adds a `bool()` to sth we probably don't want to have bool in; I also don't necessarily like the implementation of it you propose

What if we just do `int(val)` when we get the expression value, inline in the function, instead of going into magic of `bool()`? Re: non-trivial types - any examples? Do the proposed `bool(self.inner)` really help here?

Gitjay11:
> I know all of this, but this still realy do not explain why we are getting a GDB value in the first place...
@disconnect3d   

You're absolutely right, and I apologize for missing the core issue. The core issue is that evaluate_expression('$rdi == 1') returns a GDBValue with inconsistent bool()/int() behavior, as your example clearly shows
Your solution is perfect: to replace if not val: with if int(val) == 0: in next.py:224. This avoids the GDB-specific bool entirely.
Regarding non-trivial types - I couldn't find concrete examples where the complex bool(self.inner) fallback is actually needed. The current implementation seems overly defensive.
Your approach is cleaner, Thanks.
Should I update PR with this one-line fix?

disconnect3d:
> > I know all of this, but this still realy do not explain why we are getting a GDB value in the first place...
> > @disconnect3d
> 
> You're absolutely right, and I apologize for missing the core issue. The core issue is that evaluate_expression('$rdi == 1') returns a GDBValue with inconsistent bool()/int() behavior, as your example clearly shows Your solution is perfect: to replace if not val: with if int(val) == 0: in next.py:224. This avoids the GDB-specific bool entirely. Regarding non-trivial types - I couldn't find concrete examples where the complex bool(self.inner) fallback is actually needed. The current implementation seems overly defensive. Your approach is cleaner, Thanks. Should I update PR with this one-line fix?

"You're absolutely right!"

Can you please start writing as a normal human instead of generating even your comments with AI?

We don't have to reiterate on "core issue" and you could just quickly answer and send an updated revision.

I hope you do understand that you are wasting both your and reviewers time this way. 

> I couldn't find concrete examples where the complex bool(self.inner) fallback is actually needed.

And this is literally disappointing because you just admitted you sent us code you neither tested nor understand. Or rather: even this reply was vibe coded with your AI...

Do I really have to say "please update this PR with one-line fix"? I could already review that updated version but instead here we are.

Please stop wasting my time, write like human and review and run the code yourself before sending it.

Gitjay11:
@disconnect3d please I am really sorry. i am just being formal and yes i have reviewed the code and run it myself. i will do so for  any other required changes too , i am working on this that's why i have converted this into draft before only.

disconnect3d:
As no work has been done here since a month and there is code that I don't like, I am going to close this.

We will most likely continue to iterate on https://github.com/pwndbg/pwndbg/pull/3564/, or, rewrite it all from scratch, or just drop the idea. I don't understand why so many people started to work on this feature considering that those people likely do not need the feature by themselves... but thanks for trying to contribute anyway.

Gitjay11:
sorry, I was just waiting for some feedback. okay sure , i am looking forward to it.

PR #3604, opened 2026-01-08
lldb: implement add_symbol_file and remove_symbol_file

### LINKED ISSUE #3595: implement `add_symbol_file` and `remove_symbol_file` for lldb
need to implement for lldb:
https://github.com/pwndbg/pwndbg/blob/cf2351f92cd2941ce71524d17a5c1d8fd2da18be/pwndbg/dbg_mod/__init__.py#L692-L706

see 
https://github.com/pwndbg/pwndbg/blob/cf2351f92cd2941ce71524d17a5c1d8fd2da18be/pwndbg/dbg_mod/gdb/__init__.py#L1108-L1120

also need to remove the following two code blocks

https://github.com/pwndbg/pwndbg/blob/cf2351f92cd2941ce71524d17a5c1d8fd2da18be/pwndbg/integration/__init__.py#L564-L567

and

https://github.com/pwndbg/pwndbg/blob/cf2351f92cd2941ce71524d17a5c1d8fd2da18be/pwndbg/commands/klookup.py#L60-L63

AdityaByte:

### Description
This PR implements the  `add_symbol_file` and `remove_symbol_file` methods for the LLDB debugger backend. 

### Changes
- **`pwndbg/pwndbg/dbg_mod/__init__.py`**: 
    - Implemented `add_symbol_file` using `SBTarget.AddModule` and `SetModuleLoadAddress`.
    - Implemented `remove_symbol_file` using `SBFileSpec` and `SBTarget.RemoveModule`.
- **`pwndbg/pwndbg/integration/__init__.py`**: Removed the check blocking symbolication on LLDB.
- **`pwndbg/pwndbg/commands/klookup.py`**: Removed the check blocking `klookup` on LLDB.

### Implementation Details
- In LLDB, `AddModule` is used to introduce the symbol file to the target. 
- `SetModuleLoadAddress` is called when a `base` address is provided to ensure symbols are correctly shifted to match the memory layout of the debugged process.
- `remove_symbol_file` uses `SBFileSpec` to find the correct module handle before calling `RemoveModule`.

### Testing
- Logic verified against the [LLDB SBTarget API documentation](https://lldb.llvm.org/python_api/lldb.SBTarget.html).
- Static syntax checking performed on the modified files.
- Note: Full live integration testing in WSL was limited by a local LLDB-14 environment issue (`ModuleNotFoundError: No module named 'lldb.embedded_interpreter'`), but the code follows the established Pwndbg debugger abstraction.

Fixes #3595

### DIFF
--- a/pwndbg/dbg_mod/__init__.py
+++ b/pwndbg/dbg_mod/__init__.py
@@ -20,6 +20,7 @@ from typing import Tuple
 from typing import TypedDict
 from typing import TypeVar
 
+import lldb
 import pwndbg.lib.memory
 from pwndbg.lib.arch import ArchDefinition
 
@@ -693,7 +694,14 @@ class Process:
         """
         Adds a symbol file at base.
         """
-        raise NotImplementedError()
+        target = lldb.debugger.GetSelectedTarget()
+        # AddModule returns a SBModule object representing the file on disk
+        module = target.AddModule(path, None, None)
+
+        if module.IsValid and base is not None:
+            # Map the module to a specific address in the process memory
+            target.SetModuleLoadAddress(module, base)
+
 
     def remove_symbol_file(self, path: str) -> bool:
         """
@@ -703,7 +711,14 @@ class Process:
             True if we succeeded, False if not. If the file was never
             added or doesn't exist, that counts as failure.
         """
-        raise NotImplementedError()
+        target = lldb.debugger.GetSelectedTarget()
+        spec = lldb.SBFileSpec(path)
+        module = target.FindModule(spec)
+
+        if module.IsValid():
+            return target.RemoveModule(module)
+
+        return False
 
     def runcmd(self, cmd: str) -> str:
         """
--- a/pwndbg/commands/klookup.py
+++ b/pwndbg/commands/klookup.py
@@ -57,11 +57,6 @@ def klookup(symbol: str, apply: bool) -> None:
             print(message.success(f"{sym_addr:#x} {sym_type} {sym_name}"))
 
     if apply:
-        if pwndbg.dbg.name == pwndbg.dbg_mod.DebuggerType.LLDB:
-            print(message.error("Symbolication is not yet supported on LLDB."))
-            # Until we implement add_symbol_file for LLDB.
-            return
-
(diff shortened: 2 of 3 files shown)


mbrla0:
> Logic verified against the [LLDB SBTarget API documentation](https://lldb.llvm.org/python_api/lldb.SBTarget.html).

Am I missing something? The API documentation for SBTarget is silent on the behavior of `AddModule`/`SetModuleLoadAddress` when it comes to loading potentially overlapping modules for symbol data. If anything, the `const char* symfile` in [one of the signatures](https://lldb.llvm.org/python_api/lldb.SBTarget.html#lldb.SBTarget.AddModule) for `AddModule` and [the opening line in the docs for `SBModule`](https://lldb.llvm.org/python_api/lldb.SBModule.html) seem to point in the direction that LLDB expects binaries and their corresponding symbol data to be bundled into a single module, rather than allowing overlay-like modules for symbols.

I could be wrong, and LLDB could allow for this, but I think it's important to test this out first before going ham with `AddModule` and just assuming it does what we want.

> * Note: Full live integration testing in WSL was limited by a local LLDB-14 environment issue [...]

Could you please be more specific as to what has been tested? It's not clear to me what this implied partial live integration test means, and I think the LLDB docs are ambiguous enough in this case to warrant a more thorough investigation.

pwndbg/dbg_mod/__init__.py
mbrla0, line 1: Please don't take this the wrong way, but these changes makes no sense at all, and they almost certainly break `pwndbg-gdb`.

These belong under `pwndbg/dbg_mod/lldb`, assuming this actually does what we want it to do. This file defines the abstract interface for the Debugger API, and the `raise NotImplementedError()` statements are there for good reason.

Also, `lldb.debugger` is not how we interact with our instance of `SBDebugger`, that should be `self.debugger`, in the `pwndbg.dbg_mod.lldb.LLDB` class.

mbrla0 (changes requested):
I'm sorry, but to me this just comes across as AI slop, though I'd love to be proved wrong in that judgement. Still, I don't think we should take in this PR as is.

If you have questions of your own as to how to implement this, and want to do a v2, I probably wouldn't mind answering them.

PR #3606, opened 2026-01-09
Fix mypy errors in tests/ and scripts/ and enable it in CI

k4lizen:
Also deleted some gdb/ files.


k4lizen:
image

tests/library/dbg/tests/test_memory.py
```diff
@@ -67,25 +69,6 @@ async def test_memory_peek_poke(ctrl: Controller) -> None:
...
-    for not_parsable_as_int in (b"asdf", "asdf"):
-        with pytest.raises(ValueError):
-            pwndbg.aglib.memory.poke(not_parsable_as_int)
-    """
-
-    # Acceptable inputs (not great; maybe we should ban them?)
-    # Note: they return 0 because the address 0 is not mapped
-    assert pwndbg.aglib.memory.peek(0.0) is None
-    assert pwndbg.aglib.memory.peek("0") is None
-    assert pwndbg.aglib.memory.peek(b"0") is None
```
disconnect3d, line 87: Why remove this? The TODO says we should do one or another?

k4lizen: But whats the point? Should we add guard clauses on the start of every function that checks that the passed in type is the one thats annotated?

disconnect3d: Maybe peek should return True|False, heh :)

PR #3609, opened 2026-01-09
install.sh should support installing pwndbg on armv8l

patryk4815:
install.sh should support installing pwndbg on armv8l, it is 32bit mode on aarch64 cpu

eg:
```bash
$ setarch armv7l /bin/bash
$ uname -m
armv8l
$ curl install.pwndbg.re | sh
// should work fine now
```


docs/install.sh
```diff
@@ -194,7 +194,8 @@ case "$OS" in
             x86_64) FILE="${BINARY_NAME}_${VERSION}_x86_64-portable.tar.xz" ;;
             i686) FILE="${BINARY_NAME}_${VERSION}_x86_32-portable.tar.xz" ;;
             aarch64) FILE="${BINARY_NAME}_${VERSION}_arm64-portable.tar.xz" ;;
-            armv7*) FILE="${BINARY_NAME}_${VERSION}_armv7-portable.tar.xz" ;;
```
patryk4815, line 197: `armv7*` was allowing install pwndbg on armv7b (big-endian...) but we provide only binaries for little-endian

PR #3613, opened 2026-01-11
Add `di setpath` and improve syncing error messages

k4lizen:

## Before
image

## After
image

### DIFF
--- a/docs/commands/integrations/decompiler-integration.md
+++ b/docs/commands/integrations/decompiler-integration.md
@@ -3,7 +3,7 @@
 
 ```text
 usage: decompiler-integration [-h]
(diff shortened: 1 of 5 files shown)


k4lizen:
@disconnect3d i remember when you tried this out live you had a binary of a different name, and I feel like setting the path is more convenient than setting the base address (because it works with ASLR enabled)

disconnect3d:
Shall this be a command? Why not a config variable?

k4lizen:
because i want consistency with setbase, and setbase cant be a config because idk how to store 64 bit values with gdb parameters (i guess it could be a string and then we parse it to an int? idk)

disconnect3d:
Store as int? anyway, whatever, it can be as it is :)

disconnect3d (approved):

PR #3614, opened 2026-01-11
Enable caching for docker tests, fix archlinux docker test from failing, and cleanup/optimize dockerfiles to follow best practices

jackmisbach:

before
chrome_tOoujk4asX
after
chrome_n8iEbtq9va
5m 21s > 1m 56s

would ofc take effect only after this is merged 

only reason it takes 1m 30s - 2m ish instead of 30s is because of the --load flag so that it can be ran with docker compose run

ArchLinux Docker test
before
chrome_tIEMQhQx3a
after
chrome_X2xGQsKBJ9
fail > pass
was broken because of Arch Linux updating to Python 3.14

This also makes it run on dev, in which it didn't before, just to update the cache if any changes are made to any files that would affect the image.

This PR also adds build tools to setup-sh for pacman and dnf install portions so that wheels for Python 3.14 aren't required for all packages. Without this, currently fedora:43 and archlinux:latest will fail when running setup.sh.
Also, then with that removes the Dockerfile.dnf patch that was put in there to support gnureadline 
https://github.com/ludwigschwardt/python-gnureadline/issues/80#issuecomment-3640302277
well and jpype1

lmk thoughts

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -19,7 +19,7 @@ on:
       - "!*.md"
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.ref }}${{ github.event_name == 'pull_request' && format('-{0}', github.run_id) || '' }}
   cancel-in-progress: ${{ github.event_name == 'push' }}
 
 permissions:
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -19,6 +19,8 @@ on:
       - "!*.md"
 
 concurrency:
+  # Pushes (dev): queue sequentially (prevent cache corruption)
+  # PRs: cancel old commits (only test latest)
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,11 +1,30 @@
 name: Test Docker
+
 on:
+  push:
+    branches:
+      - dev
+    paths:
+      - 'Dockerfile*'
+      - 'setup.sh'
+      - 'setup-dev.sh'
+      - 'uv.lock'
+      - 'pyproject.toml'
+      - '.github/workflows/docker.yml'
   pull_request:
     paths:
       - "**"
       - "!mkdocs.yml"
       - "!docs/**"
       - "!*.md"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'push' }}
+
+permissions:
+  contents: read
+  packages: write
 
 jobs:
   docker_x86-64:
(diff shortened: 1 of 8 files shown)


Dockerfile.arch
```diff
@@ -19,17 +19,16 @@ ENV UV_PROJECT_ENVIRONMENT=/venv
 RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && \
     echo $TZ > /etc/timezone && \
     pacman -Syu --noconfirm && \
-    localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8 && \
-    pacman -S --noconfirm vim && \
+    localedef -i en_US -c -f UTF-8 en_US.UTF-8 && \
+    pacman -S --noconfirm vim base-devel && \
```
k4lizen: if this is needed for running tests on arch, it should be in the arch portion of setup-dev.sh

k4lizen:
looks good but I don't understand this stuff well enough to review, @patryk4815 can you take a look?

Dockerfile.arch
jackmisbach: It's already in setup-dev.sh.
However, setup.sh is run before setup-dev.sh, and therefore before it is installed, but it is needed to build packages because some don't have wheels for Python 3.14 yet.
I just followed what y'all did for Dockerfile.dnf, and that is install it before setup.sh is ran, because I agree with that, because it's not a big deal to just skip over it later especiallly because in setup-dev.sh there is already the --needed flag, it won't reinstall it or anything.

This is because Python was just updated to 3.14 in Arch (Jan 10th), which is the same reason it was added to Dockerfile.dnf, because fedora43 updated Python earlier. So maybe actually base-devel should be added to install_pacman() for setup.sh actually at least temporarily, because otherwise, does it currently fail on a fresh install? I'd presume it does because gnureadline doesn't have wheels for Python 3.14, and maybe other packages. I added it for now, just into setup.sh and removed from the Dockerfile and setup-dev.sh (or maybe it can stay duplicated in setup-dev.sh because of other packages are duplicated in setup.sh and setup-dev.sh), lmk what you think if we should do this, or just keep it in the Dockerfile.arch.

jackmisbach: So with that actualy, I decided to add gcc g++ make, ncurses-devel, and patch to install_dnf because I think both pacman and dnf distros should be able to run ./setup.sh with no problems on a fresh install of fedora:latest and archlinux:latest, for example, which you currently cannot. 
lmk if you disagree, but I don't think actually the Docker tests/Dockerfiles should have any different packages installed before (besides like vim or whatever ofc) setup.sh because I think seeing them fail is good to know, because that means on fresh installs the pwndbg install from source isn't working, because I think they should be just because it should be able to build packages even if wheels aren't updated for every single package, but maybe not.

So, to make this all simpler.

1. Can we not put any packages in Dockerfiles (besides vim, etc.) before setup.sh and simply put them in setup.sh to match the development install guide (installing from source) and fresh builds? - This is already done
2. Leave or remove duplication of packages in setup.sh and setup-dev.sh?

patryk4815:
> l@patryk4815 can you take a look?

yes, in few days

Dockerfile.arch
disconnect3d: Both 1 and 2 sounds reasonable.

.github/workflows/docker.yml
```diff
@@ -1,12 +1,31 @@
 name: Test Docker
+
 on:
+  push:
+    branches:
+      - dev
```
disconnect3d, line 6: If we build on pull_request do we also want to do it on dev push?

.github/workflows/docker.yml
```diff
@@ -1,12 +1,31 @@
...
   pull_request:
     paths:
       - "**"
       - "!mkdocs.yml"
       - "!docs/**"
       - "!*.md"
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}${{ github.event_name == 'pull_request' && format('-{0}', github.run_id) || '' }}
+  cancel-in-progress: ${{ github.event_name == 'push' }}
```
disconnect3d:
```suggestion
  # Cancel an in-progress build if a new push to the same branch happens
  cancel-in-progress: ${{ github.event_name == 'push' }}
```

Hm does this work for PRs too? Or is it like that for PRs anyway maybe?

disconnect3d (approved):
@patryk4815 pls review as well

.github/workflows/docker.yml
jackmisbach: Yes, just because it only gets cached on dev push, and I only have it running on dev push in cases where the cache would need to be updated. Although if you'd like, I could disable the tests from running on dev push so that it is closer to how it was before, and it just updates the cache on dev pushes instead of also running tests as it does on PRs, but I'm not sure that matters much.

.github/workflows/docker.yml
jackmisbach: So the way that I have it setup currently, it won't cancel in progress PR builds if there is a new commit, as well as allow them to run concurrently (the `group:` line before allows concurrent runs on PRs). Which is the current behavior, which I agree with. Because they should run concurrently to see which commit introduced issues, as well as they aren't updating the cache so it doesn't matter if they run concurrently on PRs, only on dev it matters where it could cause issues with cache if they are running concurrently and updating cache using --cache-to.

Dockerfile.arch
jackmisbach: > Both 1 and 2 sounds reasonable.

Okay, cool. I'll add in 2 to the PR later today if patryk agrees with that.

edit: added

.github/workflows/docker.yml
disconnect3d: While this makes sense in general I think its often unwanted to not cancel previous commit/push runs since you usually want to get the result out of the latest commit asap -- at least that's what I often want.

Maybe we should go with cancelling the old push runs too?

k4lizen: > you usually want to get the result out of the latest commit asap -- at least that's what I often want.

but will a PR's CI actually run faster if the previous one is cancelled?

jackmisbach: > While this makes sense in general I think its often unwanted to not cancel previous commit/push runs since you usually want to get the result out of the latest commit asap -- at least that's what I often want.

Not opposed to that at all if you're down for that.

> but will a PR's CI actually run faster if the previous one is cancelled?

Yeah, it would, just because free organizations can only have 20 concurrent jobs. 
Like when Dependabot makes 5 PRs at once, for example, it'll take quite some time for all the Docker tests to finish.

I'll make these changes and the setup-dev duplication changes after work and class in like 6-7 hours or so, or maybe I'll find time in class, although I think we are still waiting for Patryk's review, so not sure it matters much. Maybe I should've made the Arch Linux tests as a separate PR initially, a part from caching and such heh.

jackmisbach: @disconnect3d Okay so what I decided to do was 
```yaml
concurrency:
  # Pushes (dev): queue sequentially (prevent cache corruption)
  # PRs: cancel old commits (only test latest)
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```
also added comments though like ur review suggestion to make it clear what it's doing

this way it is fast on PRs, and you don't have to wait it just tests the latest commit so you can see if you fixed issues
and so that it still runs for every commit on dev, and doesn't cancel anything there, just queues as to prevent cache corruption

lmk if you agree, and then also want me to do the same thing for the "tests" workflow (well not the queueing part on dev because there isn't cache, but the cancelling on PRs), and then I could for the lockfile and lint workflows although those are already rather fast so it wouldn't matter as much as these would, although they still do take up the concurrency limit of 20, so idk lmk what you want to do there, but I think the tests workflow should at least also follow this standard as well

k4lizen:
you can seperate out the archlinux thing from this PR into another PR so we can merge that one quickly, since archlinux CI failing everywhere is kinda annoying

PR #3630, opened 2026-01-14
Add claude md

disconnect3d:
Fwiw this takes 3k tokens at launch.
image

We could also mark some files with `@some/file` and that would force Claude to load it at startup, but idk if that wouldn't fill context too much. I may play with it a bit and see.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -34,6 +34,10 @@ Pwndbg is a GDB and LLDB plugin that enhances debugging for low-level software d
 ./scripts/generate-docs.sh    # Update auto-generated docs (requires GDB + LLDB)
 ./scripts/verify-docs.sh      # Check docs match source code
 ./scripts/docs-live.sh        # Preview docs at http://127.0.0.1:8000/pwndbg/
+
+# Run custom Python code wit Pwndbg+GDB
+./.venv/bin/gdb /bin/ls --ex 'entry' --ex 'source code.py'  # if Pwndbg is sourced in ~/.gdbinit
+./.venv/bin/pwndbg /bin/ls --ex 'entry' --ex 'source code.py'  # if Pwndbg is not sourced
 ```
 
 ### Running Single Tests
@@ -150,7 +154,9 @@ The parameters can be set inside of a debugger with `set <param> <value>` or sho
 
 **Read `docs/contributing/writing-tests.md` for full details.**
 
-### Dual-Debugger Tests (PREFERRED)
+### Dual-Debugger Tests
+
+Use this for new GDB/LLDB tests and for any architecture-agnostic things.
 
 ```python
 # tests/library/dbg/tests/test_my_feature.py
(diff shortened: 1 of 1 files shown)


disconnect3d:
Hmm, if we include some additional files:
image

disconnect3d:
With Opus model:
image

CLAUDE.md
```diff
@@ -0,0 +1,252 @@
...
+- **GDB**: 12.1+
+- **LLDB**: 19+ (requires Python 3.12+)
+- `vermin` enforces Python 3.10 compatibility in lint
+
+## Common Patterns
+
+```python
+# NEVER do this since those objects are runtime-swapped
+from pwndbg.aglib import regs  # INSTEAD access regs via: pwndbg.aglib.regs.read_regs("<register>")
+from pwndbg.aglib import arch  # INSTEAD access current arch via: pwndbg.aglib.arch.current
```
patryk4815: what is `pwndbg.aglib.arch.current`?

CLAUDE.md
```diff
@@ -0,0 +1,252 @@
...
+    await ctrl.launch(BINARY)
+    result = await ctrl.execute_and_capture("my-command")
+    assert "expected" in result
+```
+
+Run with: `./tests.sh -d gdb -g dbg` (GDB) or `./tests.sh -d lldb -g dbg` (LLDB)
+
+### GDB-Only Tests
+
+Only for truly GDB-specific functionality:
```
patryk4815, line 180: only for legacy gdb tests and only for the x86_64 architecture

CLAUDE.md
```diff
@@ -0,0 +1,252 @@
...
+
+The parameters can be set inside of a debugger with `set <param> <value>` or shown with `show <param> <value>`. They are also displayed in `config`, `theme` or `heap_config` commands.
+
+## Writing Tests
+
+**CRITICAL**: Write tests in `tests/library/dbg/tests/` for dual-debugger support (both GDB and LLDB). Only write debugger-specific tests in `tests/library/gdb/tests/` as a last resort when functionality is truly debugger-specific.
+
+**Read `docs/contributing/writing-tests.md` for full details.**
+
+### Dual-Debugger Tests (PREFERRED)
```
patryk4815: only for new gdb/lldb tests and only for the "any"/agnostic architecture

CLAUDE.md
```diff
@@ -0,0 +1,252 @@
...
+
+## Linting and Type Checking
+
+**Read `docs/contributing/index.md#linting` for full details.**
+
+- Runs: ruff, shfmt, vermin (Python 3.10+ compat), mypy, custom-lint.py
+- Both `mypy` and `mypy --strict` must pass (or not increase error count)
+- Use type checker in your editor (mypy --strict, pyright, etc.)
+- Pre-push hook runs lint automatically after `./setup-dev.sh`
+
```
patryk4815, line 203: I don't think this section need be mentioned here, I think only `./lint.sh` should be mentioned without extra info about what linters?

k4lizen:
i am not convinced that adding this file will have a positive impact on the project

image

but okay, if it'll help some people out then maybe its worth it

regarding the doc itself, I don't have much experience with prompt engineering so can't review that aspect of it. otherwise, I agree with @patryk4815 's comments and don't have anything else to add

CLAUDE.md
disconnect3d: Why not? If the AI knows it and u tell it that lint fails, it will know more of it up front. Same if it runs lint.sh it will have initial knowledge what it does. 

I am not sure if all this is irrelevant or not hm

disconnect3d:
> i am not convinced that adding this file will have a positive impact on the project

It should help iterate on it with Claude faster and some people already do it (including me) so lets see how it goes...

E.g. if we advise it better on how to do testing, there is a chance that people will run & implement those better :s

PR #3632, opened 2026-01-15
#3486 followup - Adds function data to syscalls and disinguishes syscalls from function calls

disconnect3d:
Followup on #3486


k4lizen:
tests be dying

Issue #3588, opened 2026-01-06
decompiler-integraton list has some weird offsets printed


k4lizen:
### Description

<img width="893" height="152" alt="Image" src="https://github.com/user-attachments/assets/2058ac40-50b2-4ab6-96b4-f0efd0cbb5e9" />

Why are they negative?

<img width="429" height="105" alt="Image" src="https://github.com/user-attachments/assets/746eacb3-bd9f-43c4-b64e-c9caa36deccb" />

Why does this not look correct.

### Steps to reproduce

I can repro in the main function of 2025/hxp/brainfast.

k4lizen:
closed by https://github.com/pwndbg/pwndbg/pull/3633

PR #3633, opened 2026-01-15
Fix Frame.sp() in GDB

### LINKED ISSUE #3587: pwndbg.dbg_mod.Frame.sp() and .start() are borked
### Description

They return different things between LLDB and GDB, there is some duplication, bug fixes needed etc. Needs to be investigated.

k4lizen:

Fixes: https://github.com/pwndbg/pwndbg/issues/3587 and https://github.com/pwndbg/pwndbg/issues/3588

The stack offsets line up now.
image
image

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -215,7 +215,7 @@ pip install uv
 
 # Install dependencies
 echo "Installing dependencies.."
-uv sync --extra gdb --extra lldb --quiet
+uv sync --extra gdb --extra lldb
 
 if [ -z "$UPDATE_MODE" ]; then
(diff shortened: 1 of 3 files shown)


setup.sh
```diff
@@ -215,7 +215,7 @@ pip install uv
 
 # Install dependencies
 echo "Installing dependencies.."
-uv sync --extra gdb --extra lldb --quiet
```
disconnect3d, line 218: Is this a change we want to make?

k4lizen: ye, the initial reasoning for adding that was because the sync was quick and cluttered the output, but it can be very long now so its better to see it

PR #3637, opened 2026-01-16
Implement a libc abstraction layer

### LINKED ISSUE #2034: Refactor code to simplify adding another implementations of libc
Now it's impossible to use pwndbg with libc other than glibc - musl or uclibc. There are now a lot of glibc/ptmalloc-specific things that makes difficult to support another libc's

Code refactoring to add more abstractions would be great.

### LINKED ISSUE #3643: dynamically/statically linked detection does not work on LLDB.
### Description

Our LLDB code sees the `[vdso]` as a "module" so it thinks there are >1 modules loaded, and thus concludes that we aren't statically linked even when we are.

### Steps to reproduce

```
uv run pwndbg-lldb tests/binaries/host/heap_musl_static.native.out
pwndbg-lldb> starti
pwndbg-lldb> ipi
In [1]: pwndbg.dbg.selected_inferior().is_dynamically_linked()
Out[1]: True
```

### My setup
Pwndbg:   2025.10.20 build: 196036fed (Linux)
Python:   3.12.3 
LLDB:     21.1.7
Capstone: 6.0.0
Unicorn:  2.1.4
Pwnlib:   4.14.1

k4lizen:
Our codebase is too tightly coupled to glibc. This cleans up the code significantly and decouples the dependancy. Should be easy to add new libc implementations now, need to implement 4 non-trivial functions and 4 trivial functions and you're done.

Closes: https://github.com/pwndbg/pwndbg/issues/2034 .
Closes: https://github.com/pwndbg/pwndbg/issues/3643 .


k4lizen:
Idk if the tests will pass but the PR is pretty much done.

k4lizen:
I took a different approach than we usually do, instead of subclassing I used a protocol. Feels solid to me. I'm returning modules as objects here which is a bit :S , but okay if it brings issues in the future I'll refactor it to use classes. But I'm quite optimistic.

k4lizen:
I legit can't repro this test_musl dynamic failure even with our ubuntu dockerfile:
https://github.com/pwndbg/pwndbg/actions/runs/21094584172/job/60670332553?pr=3637#step:5:1411

I also want to know how tf it is possible that on fedora there is debugging information but internal symbols are missing:
https://github.com/pwndbg/pwndbg/actions/runs/21094584172/job/60670332564?pr=3637#step:5:1323

And in fact on the fedora docker container I get the opposite:
image

Which is still weird, but possible!

On arch `__polevll` is an internal symbol but on fedora its exported by default?? What??

pwndbg/libc/glibc.py
```diff
@@ -0,0 +1,170 @@
...
+        # https://elixir.bootlin.com/glibc/glibc-2.3/source/include/libc-symbols.h#L642
+        # https://elixir.bootlin.com/glibc/glibc-2.3/source/stdlib/exit.c#L84
+        # and I don't see it in other libc's.
+        if (
+            pwndbg.aglib.symbol.lookup_symbol("__GI_exit", objfile_endswith=mapping_name)
+            is not None
+        ):
+            return True
+        else:
+            return False
```
c0rydoras:
```suggestion
        return (
            pwndbg.aglib.symbol.lookup_symbol("__GI_exit", objfile_endswith=mapping_name)
            is not None
        )
```

pwndbg/libc/glibc.py
```diff
@@ -0,0 +1,170 @@
...
+        rodata: tuple[int, int, bytes] | None = pwndbg.aglib.elf.section_by_name(
+            mapping_name, ".rodata", try_local_path=True
+        )
+        if rodata is None:
+            return False
+        _, _, data = rodata
+        if b"GNU C Library" in data:
+            return True
+        else:
+            return False
```
c0rydoras:
```suggestion

        return b"GNU C Library" in data
```

pwndbg/libc/musl.py
```diff
@@ -0,0 +1,104 @@
...
+    rodata: tuple[int, int, bytes] | None = pwndbg.aglib.elf.section_by_name(
+        mapping_name, ".rodata", try_local_path=True
+    )
+    if rodata is None:
+        return False
+    _, _, data = rodata
+    if b"/tmp/tmpnam_XXXX" in data:
+        return True
+    else:
+        return False
```
c0rydoras:
```suggestion
    return b"/tmp/tmpnam_XXXX" in data
```

pwndbg/libc/musl.py
```diff
@@ -0,0 +1,104 @@
...
+            git="git://git.musl-libc.org/musl",
+        )
+    else:
+        ver_str = ".".join(map(str, ver))
+        return LibcURLs(
+            versioned_readable_source=f"https://elixir.bootlin.com/musl/v{ver_str}/source",
+            versioned_compressed_source=f"https://musl.libc.org/releases/musl-{ver_str}.tar.gz",
+            homepage="https://musl.libc.org/",
+            git="git://git.musl-libc.org/musl",
+        )
```
c0rydoras:
```suggestion
    ver_str = ".".join(map(str, ver))
    return LibcURLs(
        versioned_readable_source=f"https://elixir.bootlin.com/musl/v{ver_str}/source",
        versioned_compressed_source=f"https://musl.libc.org/releases/musl-{ver_str}.tar.gz",
        homepage="https://musl.libc.org/",
        git="git://git.musl-libc.org/musl",
    )
```

pwndbg/libc/util.py
```diff
@@ -0,0 +1,29 @@
...
+    return pwndbg.aglib.symbol.lookup_symbol("fscanf", objfile_endswith=mapping_name) is not None
+
+
+def clean_path(path: str) -> str:
+    # Why try to avoid: Path("[heap]").resolve() == PosixPath('/home/user/<cwd>/[heap]')
+    # FIXME: This is quite flaky and should be standardized in the codebase, see #3641 .
+    if not (path.startswith("target:") or path.startswith("[")):
+        return str(Path(path).resolve())
+    else:
+        return str(Path(path))
```
c0rydoras:
```suggestion
    return str(Path(path))
```

c0rydoras (changes requested):
just some general (python) stuff

pwndbg/libc/glibc.py
k4lizen: true true, I missed these changes after doing https://github.com/pwndbg/pwndbg/pull/3637/changes/29194295c94b04fbde87ad30180d55db3650f318

pwndbg/commands/onegadget.py
```diff
@@ -33,10 +33,19 @@ def onegadget(show_unsat: bool = False, no_unknown: bool = False, verbose: bool
...
+    which_libc = pwndbg.libc.which()
+    if which_libc != pwndbg.libc.LibcType.GLIBC:
+        if which_libc == pwndbg.libc.LibcType.UNKNOWN:
+            print(message.error("Could not find libc. Please ensure it's loaded."))
+        else:
+            print(
+                message.error(
+                    f"Currently active libc is {which_libc.value}, but onegadget only supports glibc."
+                )
+            )
```
c0rydoras, line 45:
```suggestion
        if which_libc == pwndbg.libc.LibcType.UNKNOWN:
            print(message.error("Could not find libc. Please ensure it's loaded."))
            return
        print(
            message.error(
                f"Currently active libc is {which_libc.value}, but onegadget only supports glibc."
            )
        )
```

k4lizen: i don't understand whats going on in this suggested change

c0rydoras: i messed up the indentation originally :see_no_evil:

it does the same as before, but without the `else`; -> a bit less nesting

k4lizen: in general I support the early termination pattern but I think here the way it is currently is more readable (because the control flow is more clear)

pwndbg/libc/util.py
k4lizen: same as below.

k4lizen:
the heap test failures are *probably* due to https://github.com/pwndbg/pwndbg/issues/3641 . the musl test failure is wack. I also get different behaviour from running `./tests.sh` inside the docker container and from running the commands manually inside the docker container which is wild.

k4lizen:
~~blocked by https://github.com/pwndbg/pwndbg/issues/3643 :S~~

fixed

k4lizen:
I somehow locally still fail `- tests/library/dbg/tests/test_musl.py::test_musl_dynamic_detection` in the fedora docker container, but it shouldn't be possible, lets see what CI says. edit: yup it passes in CI. man what is my environment

k4lizen:
@disconnect3d this is ready. the failing docker_aarch64 test is a flake

pwndbg/commands/__init__.py
```diff
@@ -937,20 +946,6 @@ def _OnlyWithResolvedHeapSyms(*a: P.args, **kw: P.kwargs) -> T | None:
...
-                    "Please set the GLIBC version you think the target binary was compiled (using `set glibc <version>` command; e.g. 2.32) and re-run this command."
-                )
-            else:
-                e(
-                    "Can't find GLIBC version required for this command to work, maybe is because GLIBC is not loaded yet."
-                )
-                w(
-                    "If you believe the GLIBC is loaded or this is a statically linked binary. "
-                    "Please set the GLIBC version you think the target binary was compiled (using `set glibc <version>` command; e.g. 2.32) and re-run this command"
-                )
```
k4lizen, line 953: i removed this branch of code since I think that it is impossible for us to fail detecting the version for glibc. if we do actually fail. `pwndbg.libc.glibc. _get_version` raises an exception, because I consider that a violation of my assumptions, and want the user to report a bug

it is obviously theoretically possible for someone to strip out the `GNU C Library` string from glibc and mess us up, but surely that will never come up in practice

disconnect3d: Surely no CTF chall will do that xD

PR #3647, opened 2026-01-19
Full support for debugging kernel tasks without typeinfo

jxuanli:
- Automatic recovery of types relevant to `struct task_struct`
- Implemented the `kstack`, `knamespace`, `ksighand` and `kseccomp` commands
- Improved the output of `ktask`, `vmmap`, `kfile`
- General refactoring (introduced `typeinfo_recovery` decorator, etc)

## demo

image

image

image

image

image

image

image


k4lizen:
you are cooking so hard bro

pwndbg/aglib/kernel/ktask.py
```diff
@@ -0,0 +1,1101 @@
...
+        ptrsize = pwndbg.aglib.arch.ptrsize
+        start = cur = pwndbg.aglib.memory.read_pointer_width(int(self.mm))
+        while cur:
+            yield cur
+            next = pwndbg.aglib.memory.read_pointer_width(cur + ptrsize * 2)
+            if next == start:
+                break
+            cur = next
+
+    def maple_tree_parse(self) -> Generator[int, None, None]:
```
k4lizen: maple tree is a generic kernel datastructure, ideally the parsing code for it shouldn't be tied to this NextVmaFinder class

pwndbg/aglib/kernel/ktask.py
```diff
@@ -0,0 +1,1101 @@
...
+        yield from __parse_node(ma_root, 1)
+
+
+def get_stack_offset(tasks: list[int]) -> int:
+    ptrsize = pwndbg.aglib.arch.ptrsize
+    for i in range(0x10):
+        for task in tasks:
+            a = pwndbg.aglib.memory.read_pointer_width(task + i * ptrsize)
+            b = pwndbg.aglib.memory.read_pointer_width(task + (i + 1) * ptrsize)
+            # for x64, the frist kernel pointer should be the stack
```
k4lizen:
```suggestion
            # for x64, the first kernel pointer should be the stack
```

pwndbg/aglib/kernel/ktask.py
```diff
@@ -0,0 +1,1101 @@
...
+    mm_offset = None
+    ptrsize = pwndbg.aglib.arch.ptrsize
+    init_mm = pwndbg.aglib.symbol.lookup_symbol_addr("init_mm")
+    for i in range(0x200):
+        off = i * ptrsize
+        val = pwndbg.aglib.memory.read_pointer_width(task + off)
+        if pwndbg.aglib.kernel.in_kmem_cache(val, "mm_struct") or (init_mm and init_mm == val):
+            mm_offset = off
+            break
+    assert mm_offset, f"cound not find the offset of task_struct->mm: (task: {hex(task)}"
```
k4lizen: no asserts on stuff that depends on user IO (like the shape of their kernel) please

pwndbg/aglib/kernel/ktask.py
```diff
@@ -0,0 +1,1101 @@
...
+        unsigned long i_ino;
+    }};
+    """
+
+
+def get_file_struct(file: int | None) -> str:
+    ptrsize = pwndbg.aglib.arch.ptrsize
+    result = ""
+    krelease = pwndbg.aglib.kernel.krelease()
+    assert krelease, "cannot find kernel version (shouldn't happen)"
```
k4lizen, line 513: why not? because the callees should already have resolved a krelease?

pwndbg/aglib/signal.py
```diff
@@ -115,10 +115,12 @@ def get_segv_pkuerr_description() -> str | None:
     31: "SIGSYS",
 }
 
+# https://man7.org/linux/man-pages/man7/signal.7.html
 PER_ARCH_SIGNAL_MAPPINGS: dict[
     pwndbg.lib.arch.PWNDBG_SUPPORTED_ARCHITECTURES_TYPE, dict[int, SIGNALS]
 ] = {
     "x86-64": X86_64_NUM_TO_SIGNAL_MAPPING,
+    "aarch64": X86_64_NUM_TO_SIGNAL_MAPPING,
```
k4lizen: lets rename the thing then xd

pwndbg/commands/kcurrent.py
```diff
@@ -2,22 +2,106 @@
...
+        regs, syscall_reg = task.pt_regs()
+        if regs:
+            # so that it doesn't warn on user addresses
+            old_val = pwndbg.config.auto_explore_pages.value
+            pwndbg.config.auto_explore_pages.value = "no"
+            try:
+                namelen = max(len(reg) for reg, _ in regs)
+                for reg, val in regs:
+                    _val = pwndbg.chain.format(val)
+                    desc = DisassemblyAssistant._syscall_name(val)
```
k4lizen: surely this function is underscored for a reason :S

pwndbg/commands/ktask.py
```diff
@@ -64,61 +118,215 @@ def mm(self):
...
+            if seccomp is None:
+                return None
+        result = []
+        cur = seccomp["filter"]
+        while int(cur):
+            result.append(cur["prog"])
+            cur = cur["prev"]
+        return result
+
+    def __str__(self) -> str:
```
k4lizen, line 241: i think its better to call the function `to_string()` or something if the visualization has non-trivial logic

k4lizen (approved):
can you post screenshots of the new commands in action?

i'm good for merging this though i'd like to see the maple tree parsing pulled out

also i remember we talked about the asserts but i forgot what conclusion we reached and why

k4lizen:
(branch needs to be updated, i cannot do it idk why)

pwndbg/commands/kcurrent.py
```diff
@@ -72,36 +234,19 @@ def kfile(pid=None, fd=None):
     "--set",
     dest="set_pid",
     action="store_true",
-    help="sets the kernel task used for supported pwndbg commands (kfile, pagewalk, vmmap), this option does not change internal mem (purely effects how certain commands behaves)",
+    help="sets the kernel task used for supported pwndbg commands (kfile, kstack, knamespace, ksighand, kseccomp, pagewalk, vmmap), this option does not change internal mem (purely effects how certain commands behaves)",
```
disconnect3d: Maybe:
* mem -> memory
* purely effects how certain commands behaves -> only affects how commands behave

~Also @k4lizen do u think we should say "pwndbg commands" here or just "commands"? On one side those are really Pwndbg commands, on another, we list those commands too...~

EDIT: Nvm, lets keep "Pwndbg commands" here.

pwndbg/commands/kcurrent.py
```diff
@@ -72,36 +234,19 @@ def kfile(pid=None, fd=None):
...
+    if kthread:
         indent.print(kthread)
-    if set_pid:
-        mm = kthread.mm
-        if not mm:
-            print(message.warn("mm not found, current kernel task not set."))
-            return
-        KCURRENT_PID = pid
-        KCURRENT_PGD = int(mm["pgd"])
+    if set_pid and kthread:
```
disconnect3d: Maybe like this?:

```py
if kthread is None:
    return

indent.print(kthread)
if set_pid:
    global _KCURRENT
    _KCURRENT = kthread
```

pwndbg/commands/ktask.py
```diff
@@ -18,30 +22,80 @@
 from pwndbg.aglib.kernel.macros import for_each_entry
 from pwndbg.lib.exception import IndentContextManager
 
-parser = argparse.ArgumentParser(description="Displays information about kernel tasks.")
-parser.add_argument("task_name", nargs="?", type=str, help="A task name to search for")
 
-indent = IndentContextManager()
+class pt_regs_x86_64(ctypes.Structure):
+    _fields_ = [
```
disconnect3d: Is there any reason `_fields_` is a `list` and not a `tuple`? If we can, lets use a `tuple`.

pwndbg/commands/ktask.py
```diff
@@ -64,61 +118,215 @@ def mm(self):
...
+                pt_regs = pt_regs_x86_64
+                sz = ctypes.sizeof(pt_regs)
+                # the name is differnet than the canonical name
+                syscall_reg = "orig_ax"
+            case "aarch64":
+                pt_regs = pt_regs_aarch64
+                sz = ctypes.sizeof(pt_regs) + 0x20
+                kversion = pwndbg.aglib.kernel.krelease()
+                if kversion and (5, 10) <= kversion < (6, 18):
+                    sz += 0x10
```
disconnect3d, line 205: Might be worth to add a comment with source for this (when added, when removed/changed)

pwndbg/commands/ktask.py
```diff
@@ -64,61 +118,215 @@ def mm(self):
...
-        prefix = f"[pid {self.pid}]"
+        pid = f"[pid {self.pid}]"
+        pid = color.blue(f"{pid:<11}")
+        cpu = "[cpu: -]"  # not scheduled on a cpu
+        for i in range(pwndbg.aglib.kernel.nproc()):
+            if pwndbg.aglib.kernel.current_task(i) == int(self.thread):
+                cpu = f"[cpu: {i}]"
+        cpulen = 7 + len(str(pwndbg.aglib.kernel.nproc() - 1))
+        cpu = color.red(f"{cpu:<{cpulen}}")
         desc = " "
```
disconnect3d: This line can be removed iiuc

pwndbg/commands/ktask.py
disconnect3d: I'm fine with both but I agree

pwndbg/commands/ktask.py
```diff
@@ -64,61 +118,215 @@ def mm(self):
...
+            prefix = " " * len(prefix)
+        prefix = color.blue(prefix)
         thread = color.blue(hex(int(self.thread)))
-        prefix = f"[pid {self.pid}]"
+        pid = f"[pid {self.pid}]"
+        pid = color.blue(f"{pid:<11}")
+        cpu = "[cpu: -]"  # not scheduled on a cpu
+        for i in range(pwndbg.aglib.kernel.nproc()):
+            if pwndbg.aglib.kernel.current_task(i) == int(self.thread):
+                cpu = f"[cpu: {i}]"
```
disconnect3d, line 253: Shall this also have a `break` from for loop?

pwndbg/commands/ktask.py
```diff
@@ -64,61 +118,215 @@ def mm(self):
...
+            for task in for_each_entry(_init_task["tasks"], "struct task_struct", "tasks"):
+                if (task := int(task)) and task not in seen and pwndbg.aglib.memory.is_kernel(task):
+                    seen.add(task)
+                    tasks.append(Ktask(task))
     except pwndbg.dbg_mod.Error as e:
-        print(message.error(f"ERROR: {e}"))
-        return None
+        print(message.error(f"ERROR (get_ktasks): {e}"))
+        return ()
     return tuple(tasks)
```
disconnect3d: Shouldn't we remove `tasks` and just convert `seen` to a tuple of task objects at the end like this, since it seems we duplicate the same collection?:

```py
return tuple(Ktask(task) for task in seen)
```

pwndbg/commands/ktask.py
```diff
@@ -64,61 +118,215 @@ def mm(self):
...
     threads = []
     for task in get_ktasks():
         for thread in task.threads:
             if task_name is not None and task_name not in thread.name:
                 continue
+            if pid is not None and pid != thread.pid:
+                continue
             threads.append(thread)
+    threads.sort(key=lambda thread: (thread.pid, thread.name))
+    indent = IndentContextManager()
```
disconnect3d, line 326: Is this needed? We didn't have this previously, what does this add/change?

pwndbg/commands/ktask.py
```diff
@@ -64,61 +118,215 @@ def mm(self):
...
 @pwndbg.commands.OnlyWhenPagingEnabled
-@pwndbg.commands.OnlyWithKernelDebugInfo
-def ktask(task_name=None) -> None:
+@pwndbg.commands.OnlyWithKernelSymbols
+def ktask(task_name: str | None = None, pid: int | None = None) -> None:
+    pwndbg.aglib.kernel.ktask.recover_ktask_typeinfo()
     threads = []
     for task in get_ktasks():
         for thread in task.threads:
             if task_name is not None and task_name not in thread.name:
```
disconnect3d, line 320: random thought: if `task_name` would be an empty string, we could skip the `is not None` check here, heh...

pwndbg/commands/ktask.py
```diff
@@ -64,61 +118,215 @@ def mm(self):
...
+                    tasks.append(Ktask(task))
     except pwndbg.dbg_mod.Error as e:
-        print(message.error(f"ERROR: {e}"))
-        return None
+        print(message.error(f"ERROR (get_ktasks): {e}"))
+        return ()
     return tuple(tasks)
 
 
+parser = argparse.ArgumentParser(description="Displays information about kernel tasks.")
```
disconnect3d, line 306: Hmm... this seems to be a bit confusing since it does `indent.print(thread)` and not `indend.print(task)` and we call it `ktask` and not `kthread`, heh, but maybe this is correct?

Shall the `description` maybe say `Displays information about kernel task threads` then?

pwndbg/commands/parse_seccomp.py
```diff
@@ -20,6 +20,22 @@
...
 
 
+def _parse_seccomp(filter_addr: int, filter_len: int) -> str:
+    filter_size = filter_len * 8
+    filter_bytes = pwndbg.aglib.memory.read(filter_addr, filter_size, partial=False)
+    if shutil.which("ceccomp"):
+        proc = subprocess.run(
+            ["ceccomp", "disasm", "--color", "always"], input=filter_bytes, capture_output=True
+        )
+        return proc.stdout.decode() or proc.stderr.decode()
```
disconnect3d, line 30: can we just do `return subprocess.check_output(...)` here instead, or, would it not work the same?

pwndbg/commands/parse_seccomp.py
```diff
@@ -20,6 +20,22 @@
...
+    if shutil.which("seccomp-tools"):
+        proc = subprocess.run(
+            ["seccomp-tools", "disasm", "-"], input=filter_bytes, capture_output=True
+        )
+        return proc.stdout.decode() or proc.stderr.decode()
+    return "install ceccomp or seccomp-tools to parse seccomp"
```
disconnect3d, line 36: Shouldn't those always be available with Pwndbg? or do we actually have this as optional dependency somehow?

cc: @patryk4815 @k4lizen

pwndbg/aglib/kernel/vmmap.py
```diff
@@ -16,77 +16,104 @@
...
+
+
+@pwndbg.lib.cache.cache_until("stop")
+def annotate(pages: pwndbg.dbg_mod.MemoryMap) -> None:
+    try:
+        pwndbg.aglib.kernel.ktask.recover_ktask_typeinfo()
+    except (pwndbg.lib.err.TypeNotRecoveredError, AttributeError):
+        return
+    handle_user_stack_and_filepaths(pages)
+    handle_offsets(pages)
```
disconnect3d: Seems like those functions are an internal API that will not be used elsewhere - `_annotate_user_stack_and_filepaths` and `_annotate_page_offsets` would probably be better names?

pwndbg/commands/vmmap.py
```diff
@@ -240,6 +240,9 @@ def vmmap(
 
     # All displayed pages, including lines after and lines before
     vmmap = pwndbg.aglib.vmmap.get_memory_map()
+    if pwndbg.aglib.qemu.is_qemu_kernel():
+        # called here so to not impact kernel_vmmap
+        pwndbg.aglib.kernel.vmmap.annotate(vmmap)
```
disconnect3d, line 245: So `get_memory_map()` will return pages without the cool naming etc. vs when one does it + `kernel.vmmap.annotate(pages)` -- wouldn't users always want the API that return vmmap pages to contain full info or sth?

pwndbg/aglib/kernel/ktask.py
```diff
@@ -0,0 +1,1101 @@
...
+                return i * ptrsize
+        return None
+
+    ptrsize = pwndbg.aglib.arch.ptrsize
+    pgd_offset = None
+    match pwndbg.aglib.arch.name:
+        case "x86-64":
+            reg = "cr3"
+        case "aarch64":
+            reg = "TTBR0_EL1"
```
disconnect3d, line 166: Didn't we have some case "this or that register depending on QEMU version" with this register maybe?

pwndbg/aglib/kernel/ktask.py
jxuanli: the caller checks it

pwndbg/commands/kcurrent.py
jxuanli: using `syscall_number_to_name` instead

pwndbg/commands/ktask.py
jxuanli: its just for pretty print

pwndbg/commands/ktask.py
jxuanli: I think it's fine as it is. the `threads` are actually a field of `task_struct` so we are indeed displaying kernel tasks

pwndbg/commands/vmmap.py
jxuanli: it's only used to annotate vmmap output when debugging qemu kernel. see the comment as to why it's not called in `get_memory_map` -- its not visible to the user unless we are doing vmmap

pwndbg/commands/parse_seccomp.py
k4lizen: i think this is okay the way it is

pwndbg/aglib/kernel/ktask.py
k4lizen: image

is a different one i suppose

PR #3648, opened 2026-01-19
[docs] Explain stacked PRs

k4lizen:
Yeah it's not really related to pwndbg but it was easy enough to write and I've seen it be relevant so why not.

image


k4lizen:
@mbrla0 you also played with stacked PRs a bunch, so lmk if you handled them differently while you were still using git

Issue #3644, opened 2026-01-19
cache (?) the jemalloc build


k4lizen:
For me, the jemalloc build takes by far the largest amount of time in building the docker image. It is genuinely annoying to have to wait so long for a thing that I'm not even testing. Can something be done about it?

We literarly pin the jemalloc version as well.
```
        JEMALLOC_TAR_URL="https://github.com/jemalloc/jemalloc/releases/download/5.3.0/jemalloc-5.3.0.tar.bz2"
        JEMALLOC_TAR_SHA256="2db82d1e7119df3e71b7640219b6dfe84789bc0537983c3b7ac4f7189aecfeaa"
        JEMALLOC_TAR_PATH="/tmp/jemalloc-5.3.0.tar.bz2"
        JEMALLOC_EXTRACT_PATH="/tmp/jemalloc-5.3.0"
```
Can we just prebuild the binaries for the dockerfile distros and fetch them or something? Maybe not for arch but at least for ubuntu, debian and fedora?

jackmisbach:
Are you opposed to me just adding to PR #3614, or a seperate one after that's merged, to push the images, and then we can make it so the default in the compose is just to pull the image automatically (for ubuntu24.04-mount), and then remove the --build flag from running tests docs, because it's only needed if you are making changes to the Dockerfile or something otherwise tests can just use the latest dev version. Takes about 3 minutes on my internet to pull the whole thing and much longer to build. This would also remove any local Dockerfile versioning issues when running tests

k4lizen:
My current workflow is running:
```bash
docker build -f Dockerfile -t pwndbg .
```
then
```bash
docker run -it --cap-add=SYS_PTRACE --security-opt seccomp=unconfined -v $(pwd):/pwndbg pwndbg bash
```
. Does it make more sense to use a different workflow for debugging test failure by using docker compose?

> and then remove the --build flag from running tests docs, because it's only needed if you are making changes to the Dockerfile or something otherwise tests can just use the latest dev version. 

I'm not sure what this means and what the implications of this are exactly. For instance, if somone does in fact make changes to the dockerfile will the tests that are run in the CI reflect that?

In any case, it should probably be a seperate PR.

jackmisbach:
> Does it make more sense to use a different workflow for debugging test failure by using docker compose?

nah I mean that doesn't really matter ur doing it the same way the compose would currently, I mean u could do 
[...]

PR #3656, opened 2026-01-20
Push docker images to allow pulling, and add jemalloc caching

jackmisbach:

On pushes to dev, all Docker images will be tagged and pushed to ghcr, allowing for pulling images. Pulling is quicker than building, in almost every case, usually, though it is significantly faster. 

As well as cache mounts for jemalloc, which cuts the `RUN ./setup-dev.sh` layer build time in half when you are building locally, in the case where you can't pull for whatever reason. This layer took a very long time to build before, so I wanted to optimize it as well as pulling so that no matter what you have to do, it is sped up.
Closes Issue - https://github.com/pwndbg/pwndbg/issues/3644


disconnect3d (approved):

jackmisbach:
should be good to review now.
@k4lizen, also, if you'd like to give your review on the comments I have in the Dockerfiles, let me know if it all makes sense to you, and you think it would be easy to understand for everyone using the dockerfiles

k4lizen:
> the comments I have in the Dockerfiles, let me know if it all makes sense to you

yeah it makes sense. one question I have though is whether we actually want all the `-mount` docker compose images. is there a benefit to doing it that way as opposed to passing `-v` when running?

Also, is there ever a situation where you don't want to mount the local folder? 

I like that the comments in the dockerfiles are comprehensive, but I wish there was also like a: "you want to run these two commands" (pull and mount) instead of just presenting the options.

This might not be necessary though if we had a `### Debugging test failures` section in the docs which explained this.

jackmisbach:
> yeah it makes sense. one question I have though is whether we actually want all the `-mount` docker compose images.

I don't think so. I was having trouble deciding between doing that for all of them or just -v, and after sleeping, I like -v $(pwd):/pwndbg better, although, I kind of want the docs to be that as well and just remove the ubuntu24.04-mount from the compose and docs then, or I suppose we could just keep that one and leave it on the docs, I don't think it matters much.
```docker compose run --rm -v $(pwd):/pwndbg ubuntu24.04 ./tests.sh -d gdb -g gdb```

> is there a benefit to doing it that way as opposed to passing `-v` when running?

There is a slight benefit if you are building locally, as you skip the final COPY layer that copies the whole pwndbg folder because you just bind mount your own anyway, that's the only reason I think I decided to just do mounts for all last night, but most people should be pulling anyways.

> Also, is there ever a situation where you don't want to mount the local folder?

I never have. Maybe if you just wanted to use the images that will be available after this PR to just quickly pull and run without any local files or local changes, and just wanted the latest dev branch pwndbg in a container for testing something or using it, idk?

> I like that the comments in the dockerfiles are comprehensive, but I wish there was also like a: "you want to run these two commands" (pull and mount) instead of just presenting the options.

I agree with that. I think if someone really wants to run without a mount or build locally, it is somewhat obvious as to how to do that as well, and if not, I can put a section in the docs. Let me know what you think about what I just put in the Dockerfile.

Unless we just wanted to keep building as default and just add pulling as a note as something you can do to make it faster and just keep as is for the most part.


edit: the only thing I'm 50/50 on is just making pull policy always and then just removing the comment to update your prebuilt image, and basically make it so it always gets updated if it's not latest, and then it's just one command, pulls and runs the latest image. I like pull policy as missing better, though, with the extra command saying how to update your prebuilt image.

k4lizen (approved):
looks good to me, thanks!

PR #3657, opened 2026-01-21
Follow branches for Sparc

OBarronCS:
Enhances disassembly while debugging Sparc programs. There are now handlers that determine whether branches are taken or not, and the disassembly flow follows them accordingly. Previously, because we didn't determine branch outcomes, the disassembly jumped all over the place.

See this in action: https://asciinema.org/a/AaGBYGuQMIYzRK3f

An interesting quirk of delay slots that came up while debugging Sparc: GDB allows you to be paused within a delay slot instruction in Sparc. This is not the case for MIPS, where you cannot put a breakpoint and cannot be paused within a delay slot - single-stepping a branch with a delay slot will skip over it. With Sparc, you can be paused within the delay slot. The disassembly code had an implicit assumption that this was not possible, which had to be fixed.

### DIFF
--- a/pwndbg/aglib/disasm/sparc.py
+++ b/pwndbg/aglib/disasm/sparc.py
@@ -1,9 +1,9 @@
 from __future__ import annotations
 
 from collections.abc import Callable
-from typing import override
 
 from capstone.sparc import *  # noqa: F403
+from typing_extensions import override
 
 import pwndbg.aglib.disasm.arch
 from pwndbg.aglib.disasm.instruction import ALL_JUMP_GROUPS
--- a/pwndbg/aglib/disasm/sparc.py
+++ b/pwndbg/aglib/disasm/sparc.py
@@ -1,6 +1,15 @@
 from __future__ import annotations
 
+from collections.abc import Callable
+from typing import override
+
 from capstone.sparc import *  # noqa: F403
+
+import pwndbg.aglib.disasm.arch
+from pwndbg.aglib.disasm.instruction import ALL_JUMP_GROUPS
+from pwndbg.aglib.disasm.instruction import InstructionCondition
+from pwndbg.aglib.disasm.instruction import PwndbgInstruction
+from pwndbg.emu.emulator import Emulator
 
 # Instruction groups for future use
 SPARC_LOAD_INSTRUCTIONS = {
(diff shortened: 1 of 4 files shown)


pwndbg/aglib/disasm/sparc.py
```diff
@@ -18,3 +27,89 @@
...
+# Key is ICC code
+# lambda parameters: (current ccr value)
+# Source of conditions: https://arcb.csc.ncsu.edu/~mueller/codeopt/codeopt00/notes/condbranch.html
+ICC_CONDITION_RESOLVERS: dict[int, Callable[[int], bool]] = {
+    SPARC_CC_ICC_NE: lambda ccr: not (ccr & CCR_Z_MASK),
+    SPARC_CC_ICC_E: lambda ccr: bool(ccr & CCR_Z_MASK),
+    SPARC_CC_ICC_G: lambda ccr: not ((ccr & CCR_N_MASK) ^ (ccr & CCR_V_MASK))
+    and not (ccr & CCR_Z_MASK),
+    SPARC_CC_ICC_LE: lambda ccr: bool((ccr & CCR_N_MASK) ^ (ccr & CCR_V_MASK))
+    or bool(ccr & CCR_Z_MASK),
```
OBarronCS, line 68: The linter makes this look a bit weird....

pwndbg/emu/emulator.py
```diff
@@ -82,7 +82,7 @@ def create_reg_to_const_map(
     "i386": U.UC_ARCH_X86,
     "x86-64": U.UC_ARCH_X86,
     "mips": U.UC_ARCH_MIPS,
-    "sparc": U.UC_ARCH_SPARC,
+    # "sparc": U.UC_ARCH_SPARC,
```
OBarronCS, line 85: I commented this out to indicate Sparc emulation is disabled. It didn't work in the first place, but while trying to enable it for this PR, I found that Unicorn segfaults sometimes when stepping the program, causing GDB to crash.

pwndbg/emu/emulator.py
```diff
@@ -649,6 +655,8 @@ def map_page(self, page) -> bool:
         page = pwndbg.lib.memory.page_align(page)
         size = pwndbg.lib.memory.PAGE_SIZE
 
+        # TODO: for Sparc, page size must be 0x2000
```
OBarronCS, line 658: Unicorn requires that Sparc pages are aligned to `0x2000`, otherwise it returns an error in this function.

patryk4815: maybe at some point we should get PAGE_SIZE from `AT_PAGESZ` auxv
but I think for now it is ok

pwndbg/aglib/disasm/sparc.py
```diff
@@ -18,3 +27,89 @@
...
+# 3 = N (Negative)
+CCR_C_MASK = 1 << 0
+CCR_V_MASK = 1 << 1
+CCR_Z_MASK = 1 << 2
+CCR_N_MASK = 1 << 3
+
+# Key is ICC code
+# lambda parameters: (current ccr value)
+# Source of conditions: https://arcb.csc.ncsu.edu/~mueller/codeopt/codeopt00/notes/condbranch.html
+ICC_CONDITION_RESOLVERS: dict[int, Callable[[int], bool]] = {
```
patryk4815, line 62: can be wrapped with `# fmt: off` .... `# fmt: on`

PR #3241, opened 2025-08-15
Detecting target file when debugging kernels

jxuanli:
- fixes `main_module_name` in `class GDBProcess` to properly detect target file when debugging kernels
 - closes #3193

PR #3658, opened 2026-01-21
Implementing `typeinfo_recovery` decorator and general refactoring

jxuanli:
- introduced `typeinfo_recovery` decorator
- fixed a pagescan bug
- general refactoring (renamed some functions, etc)

changes are from #3647. those changes are some general refactoring and can be reviewed separately. blocks #3647

### DIFF
--- a/pwndbg/aglib/disasm/arch.py
+++ b/pwndbg/aglib/disasm/arch.py
@@ -594,11 +594,13 @@ class DisassemblyAssistant:
             )
 
     @staticmethod
-    def _syscall_name(number: int, arch: str) -> str | None:
+    def _syscall_name(number: int, arch: str | None = None) -> str | None:
         """
         Given a syscall number and architecture, returns the name of the syscall.
         E.g. execve == 59 on x86-64
         """
+        if arch is None:
+            arch = pwndbg.aglib.arch.name
         arch_module = {
             "arm": linux.arm,
             "armcm": linux.arm,
--- a/pwndbg/aglib/disasm/arch.py
+++ b/pwndbg/aglib/disasm/arch.py
@@ -588,13 +588,11 @@ class DisassemblyAssistant:
         )
 
     @staticmethod
-    def _syscall_name(number: int, arch: str | None = None) -> str | None:
+    def _syscall_name(number: int, arch: str) -> str | None:
         """
         Given a syscall number and architecture, returns the name of the syscall.
         E.g. execve == 59 on x86-64
         """
-        if arch is None:
-            arch = pwndbg.aglib.arch.name
         arch_module = {
             "arm": linux.arm,
             "armcm": linux.arm,
--- a/pwndbg/aglib/kernel/__init__.py
+++ b/pwndbg/aglib/kernel/__init__.py
@@ -305,11 +305,7 @@ def get_double_linked_list(head: int, minlen: int = 0x1, maxlen: int = 0x1000) -
 
 def in_kmem_cache(val: int, name: str, strict: bool = True) -> bool:
     # name is a substr of any of the target caches' names
-    cache = None
-    try:
-        cache = pwndbg.aglib.kernel.slab.find_containing_slab_cache(val)
-    except Exception:
-        pass
+    cache = pwndbg.aglib.kernel.slab.find_containing_slab_cache(val)
     if not cache:
         return False
     if strict:
--- a/pwndbg/aglib/kernel/__init__.py
+++ b/pwndbg/aglib/kernel/__init__.py
@@ -131,7 +131,7 @@ def typeinfo_recovery(
                 if "CONFIG_RANDSTRUCT" in pwndbg.aglib.kernel.kconfig():
                     print(
                         message.warn(
-                            "please note that some structs may not be recoverable when CONFIG_RANSTRUCT=y"
+                            "please note that some structs may not be recoverable when CONFIG_RANDSTRUCT=y"
                         )
                     )
                 return False
--- a/pwndbg/aglib/kernel/__init__.py
+++ b/pwndbg/aglib/kernel/__init__.py
@@ -19,6 +19,7 @@ import pwndbg.aglib.kernel.vmmap
 import pwndbg.aglib.memory
 import pwndbg.aglib.proc
 import pwndbg.aglib.symbol
+import pwndbg.aglib.typeinfo
 import pwndbg.color.message as message
 import pwndbg.dbg_mod
 import pwndbg.lib.cache
@@ -27,12 +28,19 @@ import pwndbg.lib.memory
 import pwndbg.search
 from pwndbg.aglib.kernel.paging import ArchPagingInfo
 from pwndbg.aglib.kernel.paging import PagewalkResult
+from pwndbg.lib.regs import BitFlags
 
 _kconfig: pwndbg.aglib.kernel.kconfig_mod.Kconfig | None = None
 
 P = ParamSpec("P")
 D = TypeVar("D")
 T = TypeVar("T")
+
+
+class TypeNotRecovered(Exception):
+    def __init__(self, name: str, msg: str) -> None:
+        self.name = name
+        super().__init__(msg)
 
 
 def BIT(shift: int):
@@ -101,24 +109,26 @@ def requires_debug_info(default: D = None) -> Callable[[Callable[P, T]], Callabl
 
 def typeinfo_recovery(
     name: str, requires_kversion: bool = False, requires_kbase: bool = False
-) -> Callable[[Callable[P, str]], Callable[P, bool]]:
-    def decorator(f: Callable[P, str]) -> Callable[P, bool]:
+) -> Callable[[Callable[P, str]], Callable[P, None]]:
+    def decorator(f: Callable[P, str]) -> Callable[P, None]:
         # returns true if the type exists or has been successfully recovered
         @functools.wraps(f)
-        def func(*args: P.args, **kwargs: P.kwargs) -> bool:
-            if has_debug_info():
-                return True
-            if not has_debug_symbols():
+        def func(*args: P.args, **kwargs: P.kwargs) -> None:
+            if not pwndbg.dbg.selected_inferior().is_linux():
                 # make sure the target is linux, should we specify symbols instead?
-                return False
+                raise TypeNotRecovered(name, "target is not linux")
+            if has_debug_info():
+                return
             if pwndbg.aglib.typeinfo.lookup_types(name) is not None:
-                return True
+                return
             if requires_kversion and kversion() is None:
-                print(message.warn(f"recovering {name} failed because kversion is unavailable"))
-                return False
+                raise TypeNotRecovered(
+                    name, message.warn(f"recovering {name} failed because kversion is unavailable")
+                )
             if requires_kbase and kbase() is None:
-                print(message.warn(f"recovering {name} failed because kbase is unavailable"))
-                return False
+                raise TypeNotRecovered(
+                    name, message.warn(f"recovering {name} failed because kbase is unavailable")
+                )
             # f(*args, **kwargs)
             try:
                 result = f(*args, **kwargs)
@@ -127,15 +137,8 @@ def typeinfo_recovery(
                 pwndbg.commands.cymbol.add_structure_from_header(header_file_path, fname, True)
                 os.unlink(header_file_path)
             except Exception as e:
-                print(message.warn(f"recovering {name} failed with error:\n{e}"))
-                if "CONFIG_RANDSTRUCT" in pwndbg.aglib.kernel.kconfig():
-                    print(
-                        message.warn(
-                            "please note that some structs may not be recoverable when CONFIG_RANDSTRUCT=y"
-                        )
-                    )
-                return False
-            return True
+                raise TypeNotRecovered(name, str(e))
+            return
 
         return func
 
@@ -201,7 +204,9 @@ def kconfig() -> pwndbg.aglib.kernel.kconfig_mod.Kconfig | None:
             config_start = result + len("IKCFG_ST")
             config_end = next(pwndbg.search.search(b"IKCFG_ED", start=config_start), None)
     if (
-        not pwndbg.aglib.memory.is_kernel(config_start)
+        not config_start
+        or not config_end
+        or not pwndbg.aglib.memory.is_kernel(config_start)
         or not pwndbg.aglib.memory.is_kernel(config_end)
         or config_start >= config_end
     ):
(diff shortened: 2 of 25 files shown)


pwndbg/aglib/kernel/__init__.py
```diff
@@ -97,11 +98,52 @@ def func(*args: P.args, **kwargs: P.kwargs) -> T | D:
...
 
-@requires_debug_symbols("nr_cpu_ids", default=1)
+def typeinfo_recovery(
+    name: str, kversion: bool = False, kbase: bool = False
+) -> Callable[[Callable[P, str]], Callable[P, bool]]:
+    def decorator(f: Callable[P, str]) -> Callable[P, bool]:
+        # returns true if the type exists or has been successfully recovered
+        @functools.wraps(f)
+        def func(*args: P.args, **kwargs: P.kwargs) -> bool:
+            if has_debug_info():
```
disconnect3d: is this correct? i mean, the debug info may not have info about this type? lol

k4lizen:
I think a `typeinfo_recovery` decorator makes a lot of sense, but I think it shouldn't print messages like this. It should either:
1. Be clearly marked as a command-only decorator, and only decorate commands, and print error messages, or
2. raise an exception, and then we catch that exception in the command-exception-catching logic and print the appropriate error.

Also I dont fully grok why it changes a callable that returns a string into a callable that returns a bool, i can look into it in more detail tomorrow night

pwndbg/aglib/kernel/__init__.py
jxuanli: if the user sourced a symbol file as target and the target contains typeinfo and the target doesnt contain the type, the relevant commands will fail. but then the target is prob not linux

jxuanli:
@k4lizen im not sure which message you are referring to. but the decorator is not for commands. It's for the top-level functions that recover kernel typeinfos. If recovery failed then the command will either panic or not run. It's just meant so that the same checks don't need to be included in every such functions and prob easier in terms of error handling / early returns. If devs need to debug, just call it outside of the try block (uncomment the line right above the try block), but users don't really need to see the panic happen.

pwndbg/aglib/kernel/__init__.py
```diff
@@ -97,11 +99,53 @@ def func(*args: P.args, **kwargs: P.kwargs) -> T | D:
     return decorator
 
 
-@requires_debug_symbols("nr_cpu_ids", default=1)
+def typeinfo_recovery(
+    name: str, kversion: bool = False, kbase: bool = False
```
disconnect3d: shall we call it `require_kversion` and `require_kbase` or `need_{kversion,kbase}`?

pwndbg/aglib/kernel/__init__.py
disconnect3d: But if you have debug info and the required type, shouldn't `pwndbg.aglib.typeinfo.lookup_types(name) is not None:` already tell us if we have what we need?

I mean, isn't the check here too relaxed - as in - we may miss cases due to it?

pwndbg/aglib/kernel/__init__.py
```diff
@@ -97,11 +99,53 @@ def func(*args: P.args, **kwargs: P.kwargs) -> T | D:
...
+            # f(*args, **kwargs)
+            try:
+                result = f(*args, **kwargs)
+                header_file_path = pwndbg.commands.cymbol.create_temp_header_file(result)
+                fname = name.split()[-1] + "_structs"
+                pwndbg.commands.cymbol.add_structure_from_header(header_file_path, fname, True)
+                os.unlink(header_file_path)
+            except Exception as e:
+                print(message.warn(f"recovering {name} failed with error:\n{e}"))
+                if "CONFIG_RANSTRUCT" in pwndbg.aglib.kernel.kconfig():
```
disconnect3d: Is this correct?

Shouldn't this be `CONFIG_RANDSTRUCT`? Same in the warn print

pwndbg/aglib/kernel/__init__.py
```diff
@@ -228,6 +272,47 @@ def get_idt_entries() -> list[pwndbg.lib.kernel.structs.IDTEntry]:
...
+
+
+def in_kmem_cache(val: int, name: str, strict: bool = True) -> bool:
+    # name is a substr of any of the target caches' names
+    try:
+        cache = pwndbg.aglib.kernel.slab.find_containing_slab_cache(val)
+        if strict:
+            return name == cache.name
+        return name in cache.name
+    except Exception:
```
disconnect3d: Do we really need try except here and not e.g. in find_containing_slab_cache or it just returning a None?

If we keep try/except here, ideally it should be smallest possible, so e.g. if we get `cache.name` to be incorrect at some point e.g. due to refactoring, the try/except should not catch the `AttributeError`

pwndbg/aglib/kernel/bpf.py
```diff
@@ -217,24 +216,17 @@ def get_bpf_struct_offsets(prog_idr, map_idr) -> int:
...
-        return
+        raise AssertionError("cannot find either prog_idr or map_idr")
     xarray_pad_sz = get_bpf_struct_offsets(prog_idr, map_idr)
     if not xarray_pad_sz:
-        print(
-            message.warn(
-                "cannot find xa_head -- might be uninitialized (add a bpf prog/map first!)"
-            )
+        raise AssertionError(
+            "cannot find xa_head -- might be uninitialized (add a bpf prog/map first!)"
```
disconnect3d: coudl also just be `assert xarray_pad_sz is not None, "<the text here>"` but both are fine

pwndbg/aglib/kernel/paging.py
```diff
@@ -84,7 +92,8 @@ def scan(self, entry: int, is_kernel: bool = False) -> list[Page]:
             if is_kernel:
                 nbits = self.ptrsize * 8 - kernel_prefix_shift
                 offset += ((1 << nbits) - 1) << kernel_prefix_shift
-            if curr and offset in curr and flags == curr.flags:
+            if curr and offset == curr.end and flags == curr.flags:
```
disconnect3d, line 96: Can there be case when `offset>curr.end`? Do we work fine in that case?

pwndbg/aglib/kernel/bpf.py
k4lizen: I used to prefer `assert` because it feels more "aggresive", but doing `raise AssertionError` gives a better backtrace for bug reports (since it is caught by our exception formatter), and allows format strings. So /shrug

pwndbg/aglib/kernel/__init__.py
jxuanli: > But if you have debug info and the required type, shouldn't pwndbg.aglib.typeinfo.lookup_types(name) is not None: already tell us if we have what we need?

see #3241 why this may not work

this is actually a stricter check than just check if the type exists

pwndbg/aglib/kernel/paging.py
jxuanli: I updated the comment to state the assumption that this doesn't happen. I don't expect this to happen due to how the list is constructed albeit I wouldn't claim the page scanning logic is completely error free given the optimizations but I'm pretty confident with it.

jxuanli: the worst case scenario is that some chunks are not merged but that should be ok

k4lizen:
> im not sure which message you are referring to

image

These. My take is that we generally do not want aglib/ functions to print messages because the caller of the function may or may not want to use it in a way which warrants printing stuff to the user.

You'll see this pattern in LLDB a lot for instance.
https://github.com/llvm/llvm-project/blob/5af4c1c65a0a4500a4f9f892c3562f2fab0392b2/lldb/source/Breakpoint/BreakpointLocation.cpp#L264
With diagnostics objects being passed around and `Status` objects being returned instead of directly printing.

So, the two ways of returning a "rich" error for us are returning a custom error object instead of a bool, or raising a custom exception (e.g. `TypeNotRecovered`). Both would have the error message (e.g. `recovering {name} failed because kbase is unavailable`) attached so the one who handles the error/exception can choose to print it to the user. I think the case of `typeinfo_recovery`, most callees of that will want to abort the command that reached this point, so it makes more sense to go the exception route, and catch it here:
https://github.com/pwndbg/pwndbg/blob/0249deecba8069599b81f0d89207dd10df32b28b/pwndbg/commands/__init__.py#L499
And print the message here.

But anyways, I just think this is something we could/should do to improve the health of error handling in our code-base, I don't think it's a blocker for this PR if you disagree

pwndbg/aglib/disasm/arch.py
```diff
@@ -588,11 +588,13 @@ def _telescope_format_list(self, addresses: list[int], limit: int, emu: Emulator
         )
 
     @staticmethod
-    def _syscall_name(number: int, arch: str) -> str | None:
+    def _syscall_name(number: int, arch: str | None = None) -> str | None:
```
k4lizen: wait why? you don't use this anywhere

pwndbg/aglib/kernel/__init__.py
```diff
@@ -97,11 +99,53 @@ def func(*args: P.args, **kwargs: P.kwargs) -> T | D:
...
+                        message.warn(
+                            "please note that some structs may not be recoverable when CONFIG_RANDSTRUCT=y"
+                        )
+                    )
+                return False
+            return True
+
+        return func
+
+    return decorator
```
k4lizen, line 140: plz set up your editor 

Image

(and fix these)

pwndbg/aglib/kernel/__init__.py
```diff
@@ -228,6 +272,49 @@ def get_idt_entries() -> list[pwndbg.lib.kernel.structs.IDTEntry]:
     return entries
 
 
+def current_cpu() -> int:
+    return pwndbg.dbg.selected_thread().index() - 1
+
+
+def get_double_linked_list(head: int, minlen: int = 0x1, maxlen: int = 0x1000) -> list[int] | None:
```
k4lizen, line 281: looking at pwndbg/chain.py:get() , you might want to check `is_pagefault_supported`?

pwndbg/aglib/kernel/__init__.py
```diff
@@ -599,18 +658,20 @@ def kbase() -> int | None:
 
 
 @pwndbg.lib.cache.cache_until("stop")
-def pagewalk(addr, entry=None) -> tuple[PageTableLevel, ...]:
+def pagewalk(addr, entry: int = None, virt: bool = True) -> PagewalkResult:
+    # assumes entry is a valid physaddr (+ flags)
```
k4lizen: you should write stuff like this in the docstring so its obvious when hovering in the IDE from callsites

pwndbg/aglib/kernel/paging.py
```diff
@@ -33,10 +33,20 @@ def first_kernel_page_start() -> int:
...
+    idx: int | None = None
+    level: int | None = None
+
+
+@dataclass
+class PagewalkResult:
+    virt: int | None = None  # within physmap
+    phys: int | None = None  # physcal address
+    entry: int | None = None
+    levels: tuple[PageTableLevel, ...] = ()
```
k4lizen, line 50: very cool

pwndbg/aglib/kernel/paging.py
```diff
@@ -499,20 +524,21 @@ def handle_kernel_pages(self, pages: tuple[Page, ...]) -> None:
                     page.objfile = self.KERNELBSS
                 else:
                     page.objfile = self.KERNELRO
-            if pwndbg.aglib.regs.read_reg(pwndbg.aglib.regs.stack) in page:
-                page.objfile = "kernel [stack]"
+        for i, page in enumerate(pages):
+            if stack and stack in page:
```
k4lizen: the `if stack` can be pulled out as a perf optimization

pwndbg/aglib/kernel/symbol.py
```diff
@@ -304,6 +317,7 @@ def regex(self, s: str, pattern: str, nth: int) -> re.Match[Any] | None:
             return matches[nth]
         return None
 
+    @pwndbg.lib.cache.cache_until("stop")
```
k4lizen, line 325: "objfile" also maybe?

k4lizen (changes requested):

k4lizen:
btw why are a bunch of fields in PageTableLevel and PagewalkResult now None-able :?

pwndbg/aglib/kernel/paging.py
jxuanli: hmm idk the linter would complain otherwise

pwndbg/aglib/kernel/__init__.py
jxuanli: I'm not sure what you are referring to here specifically but `is_linux` is helpful elsewhere

pwndbg/aglib/disasm/arch.py
jxuanli: its from #3647, i just removed this change from this pr

pwndbg/aglib/kernel/paging.py
k4lizen: image
image


doesn't seem to complain for me

pwndbg/aglib/kernel/symbol.py
jxuanli: later when i get to properly handle undesired values

pwndbg/aglib/kernel/__init__.py
k4lizen: ah my bad, i thought it was a per-address(range) function, not a per-process check

k4lizen (approved):
Very nice.

We may even look to use `TypeNotRecovered` for other subsystems/allocators in the future :>

k4lizen:
As a side-note, it might make sense to change this in kmem_trace.py:
```python
        try:
            cache = pwndbg.aglib.kernel.slab.find_containing_slab_cache(objaddr)
            name = cache.name
        except Exception:
            self.add_result(message.warn(f"{prefix} invalid SLUB object @ {objaddr:#x}"))
            return
```
because it might give a misleading message in the case that `TypeNotRecovered` is raised. There is also an assert in `find_containing_slab_cache` which will be caught here even though it probably shouldn't be.

But its fine in some other PR

pwndbg/aglib/kernel/__init__.py
jxuanli: just also to note that it's intended to check if the sourced vmlinux (if any) contains typeinfo already, if so we use the actual (presumably) compiled typeinfo instead as this would always be as accurate as what can be recovered if not more. suppose that `has_debug_info` doesn't do that then it would be considered buggy.

PR #3664, opened 2026-01-22
Add more mappings to MIPS variant detection 

OBarronCS:
Detect `gs262e`, `gs464`, and `gs464e` MIPS variants and select the correct ISA so that Capstone disassembles instructions correctly.

cc @patryk4815


OBarronCS:
Since Capstone v6, for MIPS, we need to provide a fine-grained ISA version. Something that is purposefully not added here is code after this line:

https://github.com/pwndbg/pwndbg/blob/e1a613a78bf9457a938f93fa9c300a94e03ae135/pwndbg/dbg_mod/gdb/__init__.py#L965-L967

We could add code like
```
elif ptrsize == 8:
    arch_attributes.append(ArchAttribute.MIPS_ISA_64R6)
elif ptrsize == 4:
    arch_attributes.append(ArchAttribute.MIPS_ISA_32R6)
```
to make the latest versions of the ISA the default, whereas currently it defaults to to just `CS_MODE_MIPS64` or `CS_MODE_MIPS32` which uses some very old version of the ISA.

However, this might lead to situations where instead of getting a disassembly failure that we can investigate and fix, it would choose to disassemble the bytes according to the latest ISA, which might have different meaning than the intended MIPS variant.

OBarronCS:
Added a warning that is printed in the case that the observed MIPS variant returned by GDB is not in our list of known variants.

pwndbg/dbg_mod/gdb/__init__.py
```diff
@@ -959,6 +965,12 @@ def arch(self) -> ArchDefinition:
 
             if (attribute := gdb_mips_to_arch_attribute_map.get(isa)) is not None:
                 arch_attributes.append(attribute)
+            else:
+                message.warn(
+                    f"Warning: unable to detect the corrects MIPS variant from GDB architecture '{arch}'\n"
+                    "Instruction disassembly may be inaccurate\n"
+                    "Please create a PR including the output of `pi gdb.newest_frame().architecture().name()`"
```
disconnect3d:
```suggestion
                    "Please create a GitHub issue including the output of `pi gdb.newest_frame().architecture().name()`"
```

disconnect3d:
Perfect, thanks!

PR #3665, opened 2026-01-23
LLDB: Move process killing logic into `ProcessDriver`

mbrla0:
Currently, the process restarting code for both `starti` and `run` in `pwndbg-lldb` lives outside of `ProcessDriver`. This PR integrates that bit of functionality into the driver and the `pwndbg-lldb` event loop.

This PR also takes the opportunity to clarify a few unrelated bits of method semantics in `ProcessDriver`.

### DIFF
--- a/pwndbg/dbg_mod/lldb/repl/proc.py
+++ b/pwndbg/dbg_mod/lldb/repl/proc.py
@@ -185,6 +185,10 @@ class ProcessDriver:
     Drives the execution of a process, responding to its events and handling its
     I/O, and exposes a simple synchronous interface to the REPL interface.
 
+    # Signal Handler Safety
+    With few exceptions, no method in this class is safe to call during signal
+    handlers. Exceptions are always explicitly noted as such.
+
     # IODriver State Machine
     Because LLDB can make Python code from Pwndbg execute while an I/O driver is
     active, and having the I/O driver active while Pwndbg is running leads to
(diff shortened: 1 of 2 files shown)


pwndbg/dbg_mod/lldb/repl/proc.py
```diff
@@ -276,9 +280,35 @@ def is_core_file(self) -> bool:
...
+            # time this method returns, so we must drive the process to
+            # completion.
+            #
+            # By the time we start driving, there might be pending stop events
+            # that aren't necessarily the ones singnaling the termination of
+            # the inferior. We keep calling _run_until_next_stop until it sees
+            # that the process has been killed, which also clears driver state
+            # for us automatically.
+            while self.has_process():
+                self._run_until_next_stop()
```
disconnect3d, line 303: Is it ok that `run_until_next_stop` will not be called if process is not there anymore? You say that it does some cleanup, so this cleanup may not be done ( ? )

disconnect3d:
Added a comment, but its good so far so merging

pwndbg/dbg_mod/lldb/repl/proc.py
mbrla0: That cleanup in `_run_until_next_stop` is what makes `has_process` stop being true. I don't think there's a way for `process.Kill()` to do that on its own, since at most it should only invalidate the process. `_run_until_next_stop` should always run at least once there.

PR #3667, opened 2026-01-23
consistency between workflows, cancelling in-progress runs on PRs

jackmisbach:

This is to have consistency between the workflows, instead of just the docker tests, and build base image workflows, cancelling in-progress on PRs.

> Maybe we should go with cancelling the old push runs too?

@disconnect3d, would you still want that ^^? idk personally, I think it should just be done on PRs, and on push runs, it should just follow through and finish. I just think that way if somehow a test did fail, it is good for attribution sake, especially on push runs, because we are sacrificing it on PR runs already for speed, which there I think is fine. 
https://github.com/pwndbg/pwndbg/pull/3614#discussion_r2694025818

### DIFF
--- a/.github/workflows/base-image.yml
+++ b/.github/workflows/base-image.yml
@@ -16,7 +16,7 @@ on:
 
 concurrency:
   # Pushes (dev): queue sequentially (prevent cache corruption)
-  # PRs: cancel old commits (only test latest)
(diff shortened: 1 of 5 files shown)


disconnect3d:
I think so, let's see how it works and if its annoying or something, we can always revert

PR #3671, opened 2026-01-24
LLDB + Hexagon: do not immediately crash 

OBarronCS:
A quick patch to not crash when debugging Hexagon with LLDB.

Note that I had to do the following for this to work, assuming you are running Hexagon through qemu locally:
```sh
cd pwndbg
uv run pwndbg-lldb

# This is necessary for it not to crash immediately
target create --arch hexagon /absolute/path/to/hexagon/executable

# Make debugging errors easier
set exception-verbose on

gdb-remote localhost:1234
```

Capstone does not support Hexagon as a disassembly target, so we fallback to displaying what LLDB would display using it's internal disassembler (what you get with the `dis` command). Hexagon is a VLIW architecture which can bundle instructions into "packets" of instructions that can run in parallel. The disassembler doesn't appear take this into account, so the output contains a lot of "empty instructions" 

LLDB disassembler output: image

We copy this directly to our output:
image

There are some intro docs to the Hexagon architecture here: <https://github.com/programa-stic/hexag00n/blob/master/docs/intro_to_hexagon.rst>

In the current state, this is very incomplete, but allows for pwndbg to at least not crash with Hexagon.

### DIFF
--- a/pwndbg/aglib/disasm/instruction.py
+++ b/pwndbg/aglib/disasm/instruction.py
@@ -710,7 +710,7 @@ class ManualPwndbgInstruction(PwndbgInstruction):
         self.address = address
         self.size = ins["length"]
 
-        self.mnemonic = asm[0].strip()
+        self.mnemonic = asm[0].strip() if len(asm) > 0 else ""
         self.op_str = asm[1].strip() if len(asm) > 1 else ""
         self.groups = set()
 
--- a/pwndbg/aglib/disasm/instruction.py
+++ b/pwndbg/aglib/disasm/instruction.py
@@ -60,6 +60,7 @@ from capstone6pwndbg.x86 import X86_INS_RET
 from capstone6pwndbg.x86 import X86Op
 from typing_extensions import override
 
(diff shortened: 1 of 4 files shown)


pwndbg/aglib/disasm/instruction.py
```diff
@@ -701,6 +702,7 @@ def __init__(self, address: int, padding: int = 6) -> None:
         As a result of this, some of the methods raise NotImplementedError, because if they are called it indicates a bug elsewhere in the codebase.
         """
         ins: DisassembledInstruction = pwndbg.dbg.selected_inferior().disasm(address)
+        print("ins", ins)
```
disconnect3d: To remove

OBarronCS:
Just repushed with relevant changes (removed the debug print statement). This is ready to go

pwndbg/aglib/disasm/instruction.py
OBarronCS: Done

PR #3674, opened 2026-01-26
`Process.is_core_file` API

blankRiot96:
Following #3675, there is a lack of a standard API for both GDB and LLDB to determine whether a coredump file has been loaded

This is useful to know for e.g: when you want to prevent a certain command from being run on a corefile

### DIFF
--- a/pwndbg/aglib/proc.py
+++ b/pwndbg/aglib/proc.py
@@ -44,6 +44,13 @@ def alive() -> bool:
     return pwndbg.dbg.selected_inferior().alive()
 
 
+def is_core_file() -> bool:
+    """
+    Returns whether the loaded program is a corefile
+    """
+    return pwndbg.dbg.selected_inferior().is_core_file()
+
+
 def stopped_with_signal() -> bool:
     """
     Returns whether the program has stopped with a signal
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -831,16 +831,24 @@ def WarnOnKernelConfigRandstruct(function: Callable[P, T]) -> Callable[P, T | No
     return _WarnOnKernelConfigRandstruct
 
 
-def OnlyWhenRunning(function: Callable[P, T]) -> Callable[P, T | None]:
-    @functools.wraps(function)
-    def _OnlyWhenRunning(*a: P.args, **kw: P.kwargs) -> T | None:
-        # TODO: Properly support OnlyWhenRunning without `gdblib`.
-        if pwndbg.aglib.proc.alive():
-            return function(*a, **kw)
-        log.error(f"{func_name(function)}: The program is not being run.")
-        return None
+def OnlyWhenRunning(
+    func_when_decorator_kwargs: Callable[P, T] | None = None, *, allow_core: bool = True
+) -> Callable[[Callable[P, T]], Callable[P, T | None]] | Callable[P, T | None]:
+    def decorator(func: Callable[P, T]) -> Callable[P, T | None]:
+        @functools.wraps(func)
+        def _OnlyWhenRunning(*a: P.args, **kw: P.kwargs) -> T | None:
+            if pwndbg.aglib.proc.alive() and not (
+                not allow_core and pwndbg.aglib.proc.is_core_file()
+            ):
+                return func(*a, **kw)
+            log.error(f"{func_name(func)}: The program is not being run.")
+            return None
 
-    return _OnlyWhenRunning
+        return _OnlyWhenRunning
+
+    if func_when_decorator_kwargs is None:
+        return decorator
+    return decorator(func_when_decorator_kwargs)
 
 
 def OnlyWithTcache(function: Callable[P, T]) -> Callable[P, T | None]:
--- a/tests/library/dbg/tests/test_is_core_file.py
+++ b/tests/library/dbg/tests/test_is_core_file.py
@@ -1,0 +1,41 @@
+from __future__ import annotations
+
+import tempfile
+from pathlib import Path
+
+from ....host import Controller
+from . import get_binary
+from . import pwndbg_test
+
+CRASH_SIMPLE_BINARY = get_binary("crash_simple.native.out")
+
+
+@pwndbg_test
+async def test_is_core_file(ctrl: Controller) -> None:
+    """
+    Ensure Process.is_core_file() correctly distinguishes between
+    a live inferior and a loaded core file.
+    """
+    import pwndbg
+
+    await ctrl.launch(CRASH_SIMPLE_BINARY)
+
+    proc = pwndbg.dbg.selected_inferior()
+    assert proc is not None
+    assert proc.is_core_file() is False
+
+    await ctrl.cont()
+
+    with tempfile.NamedTemporaryFile(delete=False) as f:
+        core_path = Path(f.name)
+
+    try:
+        await ctrl.generate_core_file(Path(core_path))
+
+        proc = pwndbg.dbg.selected_inferior()
+        assert proc is not None
+        assert proc.is_core_file() is True
+
+    finally:
+        if core_path.exists():
+            core_path.unlink()
--- a/pwndbg/dbg_mod/__init__.py
+++ b/pwndbg/dbg_mod/__init__.py
@@ -434,6 +434,12 @@ class Process:
         """
         raise NotImplementedError()
 
+    def is_core_file(self) -> bool:
+        """
+        Returns whether this process is a coredump file.
+        """
+        raise NotImplementedError()
+
     def stopped_with_signal(self) -> bool:
         """
         Returns whether this process was stopped by a signal.
--- a/pwndbg/dbg_mod/gdb/__init__.py
+++ b/pwndbg/dbg_mod/gdb/__init__.py
@@ -604,6 +604,15 @@ class GDBProcess(pwndbg.dbg_mod.Process):
         return gdb.selected_thread() is not None
 
     @override
+    def is_core_file(self) -> bool:
+        inferior = gdb.selected_inferior()
+        return (
+            hasattr(inferior, "connection")
+            and inferior.connection is not None
+            and inferior.connection.type == "core"
+        )
+
+    @override
     def stopped_with_signal(self) -> bool:
         return "It stopped with signal " in gdb.execute("info program", to_string=True)
 
--- a/pwndbg/dbg_mod/lldb/__init__.py
+++ b/pwndbg/dbg_mod/lldb/__init__.py
@@ -983,6 +983,10 @@ class LLDBProcess(pwndbg.dbg_mod.Process):
         )
 
     @override
+    def is_core_file(self) -> bool:
+        return bool(self.process.GetCoreFile().IsValid())
+
+    @override
     def stopped_with_signal(self) -> bool:
         return self.process.GetState() == lldb.eStateStopped and any(
             thread.GetStopReason() == lldb.eStopReasonSignal for thread in self.process.threads
--- a/tests/host/__init__.py
+++ b/tests/host/__init__.py
@@ -215,6 +215,15 @@ class Controller:
         """
         raise NotImplementedError()
 
+    async def generate_core_file(self, path: Path) -> None:
+        """
+        Generate a core file at `path` and switch the debugger
+        to debugging that core file.
+
+        After this returns `pwndbg.dbg.selected_inferior().is_core_file()` must be True
+        """
+        raise NotImplementedError()
+
 
 def start(controller: Callable[[Controller], Coroutine[Any, Any, None]]) -> None:
     """
--- a/tests/host/gdb/pytests_launcher.py
+++ b/tests/host/gdb/pytests_launcher.py
@@ -64,6 +64,10 @@ class _GDBController(host.Controller):
         gdb.execute("set debug-file-directory")
         gdb.execute("set debuginfod enabled off")
 
+    async def generate_core_file(self, path: Path) -> None:
+        gdb.execute(f"generate-core-file {path}")
+        gdb.execute(f"core-file {path}")
+
 
 def _start(outer: Callable[[host.Controller], Coroutine[Any, Any, None]]) -> None:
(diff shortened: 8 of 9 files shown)


patryk4815:
that is not good, otherwise we break a lot of commands

blankRiot96:
@patryk4815 true, but `alive` *should* still return False for corefiles

so what if in those commands instead of only checking if the process is "alive" (because we should be able to use those for things like corefiles too), we also had clauses for corefiles?

blankRiot96:
Something like this?

In `aglib/proc.py`
```py
def core_loaded() -> bool:
    """
    Returns whether the loaded program is a corefile
    """
    inferior = gdb.selected_inferior()
    return (
        hasattr(inferior, "connection")
        and inferior.connection is not None
        and inferior.connection.type == "core"
    )

```

In `commands/__init__.py`
```py
def OnlyWhenRunning(function: Callable[P, T] | None = None, *, allow_core: bool = False):
    def decorator(func: Callable[P, T]) -> Callable[P, T | None]:
        @functools.wraps(func)
        def _OnlyWhenRunning(*a: P.args, **kw: P.kwargs) -> T | None:
            # TODO: Properly support OnlyWhenRunning without `gdblib`.
            if pwndbg.aglib.proc.alive() or (
                allow_core and pwndbg.aglib.proc.core_loaded()
            ):
                return func(*a, **kw)

            log.error(f"{func_name(func)}: The program is not being run.")
            return None

        return _OnlyWhenRunning

    if function is not None:
        return decorator(function)

    return decorator
```

Then we can use it for the commands that look at process state rather than change/run it like this: 
```py
@pwndbg.commands.Command(parser, category=CommandCategory.MEMORY)
@pwndbg.commands.OnlyWhenRunning(allow_core=True)
def hexdump(
```

blankRiot96:
Seems to be working
```
pwndbg> python print(pwndbg.dbg.selected_inferior().alive())
False
pwndbg> hexdump
+0000 0x7ffc0236e340  e0 e3 36 02 fc 7f 00 00  ca a1 62 a5 39 73 00 00  │..6.....│..b.9s..│
+0010 0x7ffc0236e350  90 e3 36 02 fc 7f 00 00  68 e4 36 02 fc 7f 00 00  │..6.....│h.6.....│
+0020 0x7ffc0236e360  40 c0 65 61 01 00 00 00  29 d1 65 61 02 62 00 00  │@.ea....│).ea.b..│
+0030 0x7ffc0236e370  68 e4 36 02 fc 7f 00 00  dc 3a d3 1f 33 3e 59 9a  │h.6.....│.:..3>Y.│
pwndbg> 
```

Should I go ahead with this approach?
I'll have to make `allow_core=True` for `telescope`, `hexdump`, `xinfo`, `regs`, `context`, `dumpargs`, `canary` and `got`

blankRiot96:
As per [suggestion in Discord](https://discord.com/channels/843809097920413717/843822664841166848/1465266422686158858) I will go ahead and keep `allow_core=True` by default so that it doesn't break stuff unintentionally

```py
def OnlyWhenRunning(function: Callable[P, T] | None = None, *, allow_core: bool = True):
    def decorator(func: Callable[P, T]) -> Callable[P, T | None]:
        @functools.wraps(func)
        def _OnlyWhenRunning(*a: P.args, **kw: P.kwargs) -> T | None:
            # TODO: Properly support OnlyWhenRunning without `gdblib`.
            if pwndbg.aglib.proc.alive() or (pwndbg.aglib.proc.core_loaded() and allow_core):
                return func(*a, **kw)

            log.error(f"{func_name(func)}: The program is not being run.")
            return None

        return _OnlyWhenRunning

    if function is not None:
        return decorator(function)

    return decorator
```

I will explicitly make `allow_core=False` for `errno` for now, and for any other functions upon further suggestion
```py
@pwndbg.commands.Command(parser, command_name="errno", category=CommandCategory.LINUX)
@pwndbg.commands.OnlyWhenRunning(allow_core=False)
def errno_(err) -> None:
```

disconnect3d:
Random thought without thinking too much here: `OnlyWhenRunning` could do `try: <call command> ; except: <check if corefile - mention that command may have failed bcoz of that if its corefile`.

Not sure what others think about it tho

patryk4815:
I think we should do nothing about it, just print a warning that some commands may crash when debugging coredump

We generally make some "issues" without thinking about whether they make sense or whether they can be solved well.

pwndbg/aglib/proc.py
```diff
@@ -44,6 +44,21 @@ def alive() -> bool:
     return pwndbg.dbg.selected_inferior().alive()
 
 
+def core_loaded() -> bool:
+    """
+    Returns whether the loaded program is a corefile
+    """
+
+    import gdb
```
k4lizen: cant have gdb-only code in aglib/

k4lizen (changes requested):

k4lizen:
I think a robust solution would require:

1. Looking into what LLDB does to manage this distinction because I think that codebase is sensible
2. Thinking about the appropriate abstraction or API in the context of what limitations a corefiled process has and if any other program states also have that limitation. Do we want a "has address space but isn't alive" API like `proc.is_loaded()` or do we just want something like `is_corefile()`?
3. How to detect this state
4. Thinking about what limitations there are for a corefile debugging session as opposed to a normal one, possibly taking a glance at which commands are and aren't available in lldb and gdb. Definitely expression evaluation is the major thing.
5. Making all commands with OnlyWhenRunning and all callsites of `alive()` only require the "looser" thing i.e. `is_loaded()` or `alive()` depending on what was decided in the previous step. I.e. this means keeping the behaviour the same as it is currently, but with a more sensible API.
6. Going through all of those commands and callsites and thinking about whether they require the stuff from `(4)` and annotating them approprietly. I.e. actually fixing the issue

In any case I agree with @patryk4815 and @jxuanli that the default behaviour of the users of the API should stay the same (but not necessarily the "name" of this API), and any place where a corefile-like-state is required should be done intentionally / explicitly.

Potentially we should have enum that describes what "kind" of debugging session are we in. Are we debugging qemu-user? qemu-system? windows? remote? linux? bare metal? corefile? Some stuff like this, not exactly sure what the dilineation would be but yeah. Then commands would be annotated with all the states that they do support or do not support.

Anyways, since this is some significant amount of work, I think this approach of having `@OnlyWhenRunning` accept an optional argument regarding the corefile makes sense and is probably the best "bang for our buck". And in this case you shouldn't update the behaviour of `alive()`.

k4lizen:
> We generally make some "issues" without thinking about whether they make sense or whether they can be solved well.

I did think about it! Maybe I just didn't express in the issue that this is a nontrivial problem

k4lizen:
If @disconnect3d and @patryk4815 agree with the approach you are taking (I think it is sensible) I would expect:
1. A `pwndbg.dbg.selected_inferior().is_corefile()` API
2. No changes in `alive()`
3. Adding the kwarg to @OnlyWhenRunning
4. Adding a test for this in dbg/

patryk4815:
> If @disconnect3d and @patryk4815 agree with the approach you are taking (I think it is sensible) I would expect:
> 
> 1. A `pwndbg.dbg.selected_inferior().is_corefile()` API

I think better approach would be something like: `pwndbg.aglib.arch.connection_attributes?` (enum: core file, qemu-user, etc)

---

Btw `errno` command can work without problem with core files, We can fix that commend..

blankRiot96:
So, to clarify:

- Need a `pwndbg.dbg.selected_inferior().is_corefile()` API and use that in `aglib/proc.py`.
- The `Process.is_corefile` API would have to be implemented in both `GDBProcess` and `LLDBProcess`
- Let `alive()` keep returning True for corefiles as original
- Keep the `allow_core` kwarg as it is, adjust boolean logic for `alive()` being True
- Add a test in `tests/library/dbg/tests`
- Let `errno` continue being allowed to be used with corefiles for now

Is this right?

k4lizen:
yeah. i think thats sensible. we can look into a `connection_attributes` type of thing in the future.

pwndbg/commands/__init__.py
```diff
@@ -831,16 +831,25 @@ def _WarnOnKernelConfigRandstruct(*a: P.args, **kw: P.kwargs) -> T | None:
...
-def OnlyWhenRunning(function: Callable[P, T]) -> Callable[P, T | None]:
-    @functools.wraps(function)
-    def _OnlyWhenRunning(*a: P.args, **kw: P.kwargs) -> T | None:
-        # TODO: Properly support OnlyWhenRunning without `gdblib`.
-        if pwndbg.aglib.proc.alive():
-            return function(*a, **kw)
-        log.error(f"{func_name(function)}: The program is not being run.")
-        return None
+def OnlyWhenRunning(
+    function: Callable[P, T] | None = None, *, allow_core: bool = True
```
k4lizen: why `function: Callable[P, T] | None = None`?

pwndbg/commands/__init__.py
```diff
@@ -831,16 +831,25 @@ def _WarnOnKernelConfigRandstruct(*a: P.args, **kw: P.kwargs) -> T | None:
...
+) -> Callable[[Callable[P, T]], Callable[P, T | None]] | Callable[P, T | None]:
+    def decorator(func: Callable[P, T]) -> Callable[P, T | None]:
+        @functools.wraps(func)
+        def _OnlyWhenRunning(*a: P.args, **kw: P.kwargs) -> T | None:
+            # TODO: Properly support OnlyWhenRunning without `gdblib`.
+            if pwndbg.aglib.proc.alive() and not (
+                pwndbg.aglib.proc.is_core_file() and not allow_core
```
k4lizen:
```suggestion
                not allow_core and pwndbg.aglib.proc.is_core_file() 
```
tiny perf optimization

pwndbg/commands/__init__.py
```diff
@@ -831,16 +831,25 @@ def _WarnOnKernelConfigRandstruct(*a: P.args, **kw: P.kwargs) -> T | None:
...
-            return function(*a, **kw)
-        log.error(f"{func_name(function)}: The program is not being run.")
-        return None
+def OnlyWhenRunning(
+    function: Callable[P, T] | None = None, *, allow_core: bool = True
+) -> Callable[[Callable[P, T]], Callable[P, T | None]] | Callable[P, T | None]:
+    def decorator(func: Callable[P, T]) -> Callable[P, T | None]:
+        @functools.wraps(func)
+        def _OnlyWhenRunning(*a: P.args, **kw: P.kwargs) -> T | None:
+            # TODO: Properly support OnlyWhenRunning without `gdblib`.
```
k4lizen:
```suggestion
```
you can also remove this comment, i think its outdated

tests/library/dbg/tests/test_is_core_file.py
```diff
@@ -0,0 +1,48 @@
...
+
+    try:
+        if pwndbg.dbg.is_gdblib_available():
+            await ctrl.execute(f"generate-core-file {core_path}")
+            await ctrl.execute("file")
+            await ctrl.execute(f"core-file {core_path}")
+        else:
+            await ctrl.execute(f"process save-core {core_path}")
+            await ctrl.execute("target delete")
+            await ctrl.execute(f"target create --core {core_path}")
```
k4lizen: this is kinda bad and will break if we e.g. add windbg support. better add a `generate_core_file` function to `tests/host/__init__.py`

pwndbg/commands/__init__.py
```diff
@@ -831,16 +831,25 @@ def _WarnOnKernelConfigRandstruct(*a: P.args, **kw: P.kwargs) -> T | None:
...
+                return func(*a, **kw)
+            log.error(f"{func_name(func)}: The program is not being run.")
+            return None
+
+        return _OnlyWhenRunning
 
-    return _OnlyWhenRunning
+    if function is None:
+        return decorator
+    return decorator(function)
```
k4lizen: i am getting a bit lost in the decorator sauce, why are these changes necessary?

k4lizen (commented):
Thanks! Looks pretty good, just a few comments

k4lizen:
Also change the PR description please

pwndbg/commands/__init__.py
blankRiot96:
```py
>>> # lets say we have this
>>> def when_running(f, *, allow_core=True):
...     def _when_running(*a, **kw):
...         return f(*a, **kw)
...     return _when_running
...     
>>> # this works when no arguments are passed to the decorator
>>> @when_running
... def x(): ...
... 
>>> # because this is just:
>>> x = when_running(x)
>>> # but lets say you add arguments,
>>> @when_running(allow_core=False)
... def x(): ...
... 
  File "<python-input-16>", line 1, in <module>
    @when_running(allow_core=False)
     ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
TypeError: when_running() missing 1 required positional argument: 'f'
>>> # this doesnt work because this translates to: `x = when_running(allow_core=False)(x)`
>>> # we wanted `x` to be passed like `when_running(x, allow_core=False)`
>>> # so we usually solve this by just keeping the first argument optional and having another function layer to consume `x` 
>>> def when_running(_f=None, *, allow_core=True):
...     def decorator(f):
...         def _when_running(*a, **kwargs):
...             return f(*a, **kwargs)
...         return decorator
...     if _f is None:  # when `x = when_running(allow_core=False)(x)`
...         return decorator  # so `x` gets passed to this function
...     else:  # when `x = when_running(x)` then `_f=x`
...         return decorator(_f)
...         
>>> @when_running
... def x(): ...
... 
>>> @when_running(allow_core=False)
... def x(): ...
... 
>>> 
```

pwndbg/commands/__init__.py
blankRiot96: explained the need for `function` to be optional further down, but i realize i should prob change the naming a bit

pwndbg/commands/__init__.py
k4lizen: makes sense, thanks for the explanation

k4lizen (approved):
Cool, thanks!

PR #3680, opened 2026-01-28
Fix bug with the `finally` clause

jxuanli:
- Fix bug with the `finally` clause
- Update kernel test result parsing so that tests actually fails now (I'd guess a prior PR changed the test outputs but forgot to change kernel test result parsing)
- refactoring kconfig

thanks to @k4lizen for bringing that up to me in DM

### DIFF
--- a/pwndbg/aglib/kernel/paging.py
+++ b/pwndbg/aglib/kernel/paging.py
@@ -336,6 +336,7 @@ class ArchPagingInfo:
         oldval, success = self.switch_to_phymem_mode()
         if not success or not scan:
             return PagewalkResult()
+        result = PagewalkResult()
         try:
             result = scan.walk(target, entry)
             if not virt:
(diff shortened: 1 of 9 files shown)


pwndbg/aglib/kernel/paging.py
```diff
@@ -425,7 +430,7 @@ def get_vmalloc_vmemmap_bases(self) -> tuple[int | None, int | None]:
     @pwndbg.lib.cache.cache_until("stop")
     def physmap(self) -> int:
         result: int | None = pwndbg.aglib.kernel.symbol.try_usymbol("page_offset_base")
-        if result is None:
+        if not result:
```
k4lizen, line 433: I think the idiomatic pattern is actually `result is None`, because `not result` is a "truthy" check whereas `result is None` is explicit (and probably faster)

pwndbg/aglib/kernel/paging.py
```diff
@@ -348,21 +349,25 @@ def _pagewalk(self, target: int, entry: int, virt: bool) -> PagewalkResult:
...
     def _pagescan(self, entry: int, is_kernel: bool = False) -> list[Page]:
         scan = self.pagetablescan(entry)
         oldval, success = self.switch_to_phymem_mode()
         if not success or not scan:
             return []
+        result = []
         try:
-            return scan.scan(entry, is_kernel)
+            result = scan.scan(entry, is_kernel)
+        except Exception:
```
k4lizen, line 366: what exceptions are you actually trying to catch with this though? it is hard to read what kind of faliure mode you are trying to anticipate from scan.scan when you just do `except Exception`, not to mention the fact that you are catching assertions and stuff like `TypeNotRecovered` which you should be letting bubble up the call-stack.

https://docs.astral.sh/ruff/rules/blind-except/

also catching stuff like system exit, keyboard interrupt, ....

k4lizen (approved):

k4lizen:
@jxuanli The grep should probably be the same one that we use here right?

https://github.com/pwndbg/pwndbg/blob/9b805001823b097cbca122a041fe9d5465d3b4f1/tests/host/__init__.py#L51-L56

We had some testing coloring changes and I remember there were differences between gdb and lldb, not sure what the state of that is now. I guess we don't run the kernel tests under lldb?

jxuanli:
ehh well the other are all optional so. but this is just a hot patch, later I think this should be rewritten in python

PR #3681, opened 2026-01-28
Correcting parameter name for `commands.OnlyWhenRunning` decorator

blankRiot96:
Fixing tiny naming mistake in #3674 

I apologize, in the heat of the moment my rationale was that "when `func_when_decorator_kwargs` is `<Some Default>` that means it hasn't been altered and hence we should trigger the case when the user has provided decorator kwargs"

But in this case, that `<Some Default>` is `None`. And so instead it reads like:
```py
if func_when_decorator_kwargs is None:  # reads like decorator kwargs are NOT supplied
    # do the condition for decorator kwargs IS being supplied
```

but actually, the `func_when_decorator_kwargs` parameter is only `None` when decorator kwargs ARE being assigned, so it should be called `func_when_no_decorator_kwargs` or shorter `func_when_no_kwargs`

Now, it reads like
```py
if func_when_no_kwargs is None:  # reads like kwargs ARE being supplied
    # do the condition when the kwargs ARE supplied
```

so this is more correct

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -831,7 +831,7 @@ def WarnOnKernelConfigRandstruct(function: Callable[P, T]) -> Callable[P, T | No
 
 
 def OnlyWhenRunning(
-    func_when_decorator_kwargs: Callable[P, T] | None = None, *, allow_core: bool = True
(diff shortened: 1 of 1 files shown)


k4lizen:
thanks :p

PR #3682, opened 2026-01-29
Register help subcommands to work with GDB

### LINKED ISSUE #3547: Investigate "help <cmd> <subcmd>" in GDB showing help for cmd instead of subcmd
TL;DR Our subcommands has an issue: if you `help di connect` it doesn't show the actual help of connect and instead shows help for `di` whereas `di connect --help` is correct.

It may be that we need to use the `prefix` kwarg of a GDB `Command` object to fix that, at least hopefully?

```
Function: Command.__init__ (name, command_class [, completer_class [, prefix]])

    The object initializer for Command registers the new command with GDB. This initializer is normally invoked from the subclass’ own __init__ method.

    name is the name of the command. If name consists of multiple words, then the initial words are looked for as prefix commands. In this case, if one of the prefix commands does not exist, an exception is raised.

    There is no support for multi-line commands.

    command_class should be one of the ‘COMMAND_’ constants defined below. This argument tells GDB how to categorize the new command in the help system.

    completer_class is an optional argument. If given, it should be one of the ‘COMPLETE_’ constants defined below. This argument tells GDB how to perform completion for this command. If not given, GDB will attempt to complete using the object’s complete method (see below); if no such method is found, an error will occur when completion is attempted.

    prefix is an optional argument. If True, then the new command is a prefix command; sub-commands of this command may be registered. 
```

blankRiot96:
Fixes #3547 

Previously, using `help` on subcommands would return the information of the main command. Whereas `<main> <sub>  --help` worked

In order to make `help` work with subcommands, the main command needs to be registered as a prefix command

So I've added an `is_prefix` flag to `dbg.add_command` which gets passed down the chain to `gdb.Command`'s constructor

I'm iterating over `self.parser._actions` to find the subcommand help docs and I'm adding the commands as GDB subcommands with their respective documentation
```py
subcmd_help = {}
    if isinstance(action, argparse._SubParsersAction):
        for name, subparser in action.choices.items():
            subcmd_help[name] = subparser.format_help()

for subcmd in self.subcommand_names:
    doc = subcmd_help.get(subcmd, f"See help for {self.command_name} {subcmd}")

    def _proxy_handler(
        _debugger: pwndbg.dbg_mod.Debugger,
        arguments: str,
        is_interactive: bool,
        sc: str = subcmd,
    ) -> None:
        full_args = f"{sc} {arguments}" if arguments else sc
        self.invoke(full_args, is_interactive)

    self.handles.append(
        pwndbg.dbg.add_command(
            f"{self.command_name} {subcmd}", _proxy_handler, doc, None, is_prefix=False
        )
    )
```

This works with the full command names, but the main command aliases have not been registered, and so I also register the aliases with GDB
```py
for alias in self.aliases:
    gdb.execute(f"alias {alias} = {self.command_name}")
```

```
pwndbg> help di connect
Connect to the decompiler.

The host and port to connect to are governed by the `decompiler-host`
and `decompiler-port` config variables. Try `help set decompiler-host`.

usage: decompiler-integration connect [-h]

options:
pwndbg> 

```

### Test cases
There's an issue with the test cases because by adding the subcommands as *GBD* subcommands, GDB also prints out the subcommand help information. Which creates duplication in help printing

image

For the `tests/library/gdb/tests/test_help.py` test case, I simply removed that second part of it and checked with the remaining

But there's also a failing `tests/library/gdb/tests/test_consistent_help.py` which I haven't touched for now, because the only thing to do compatible with my changes is to remove it

### Opinion on duplication
I need an opinion on the help duplication thing? I don't know if there's a way to suppress GDB subcommand help output, but if so it would probably affect builtin commands too? 
It's not my place to decide, but since this is the only way to make `help <main> <sub>` work with GDB (that is, by using prefix and subcommands within GDB) there might not be a way to avoid duplication

So if we want this issue fixed, this is the only way. Or I guess we can just leave it as it is, with `help <main> <sub>` only showing information on the main command 

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -233,20 +233,53 @@ class CommandObj:
         ) -> None:
             self.invoke(arguments, is_interactive)
 
-        # Keep a handle to the command and its aliases so we can
-        # easily remove them if necessary (not supported with GDB).
+        is_gdb = pwndbg.dbg.is_gdblib_available()
+        is_prefix = bool(self.subcommand_names)
+
         self.handles = [
-            # Tell the debugger about the command...
             pwndbg.dbg.add_command(
-                self.command_name, _handler, self.help_str, self.subcommand_names
+                self.command_name,
+                _handler,
+                self.help_str,
+                self.subcommand_names,
+                is_prefix=is_prefix,
             )
         ]
 
-        # ...and all of its aliases.
-        self.handles.extend(
-            pwndbg.dbg.add_command(alias, _handler, self.help_str, self.subcommand_names)
-            for alias in self.aliases
-        )
+        if is_gdb and is_prefix and self.subcommand_names is not None:
+            import gdb
+
+            subcmd_help = {}
+            for action in self.parser._actions:
+                if isinstance(action, argparse._SubParsersAction):
+                    for name, subparser in action.choices.items():
+                        subcmd_help[name] = subparser.format_help()
+
+            for subcmd in self.subcommand_names:
+                doc = subcmd_help.get(subcmd, f"See help for {self.command_name} {subcmd}")
+
+                def _proxy_handler(
+                    _debugger: pwndbg.dbg_mod.Debugger,
+                    arguments: str,
+                    is_interactive: bool,
+                    sc: str = subcmd,
+                ) -> None:
+                    full_args = f"{sc} {arguments}" if arguments else sc
+                    self.invoke(full_args, is_interactive)
+
+                self.handles.append(
+                    pwndbg.dbg.add_command(
+                        f"{self.command_name} {subcmd}", _proxy_handler, doc, None, is_prefix=False
+                    )
+                )
+
+            for alias in self.aliases:
+                gdb.execute(f"alias {alias} = {self.command_name}")
+        else:
+            self.handles.extend(
+                pwndbg.dbg.add_command(alias, _handler, self.help_str, self.subcommand_names)
+                for alias in self.aliases
+            )
 
         command_names.add(self.command_name)
         commands.append(self)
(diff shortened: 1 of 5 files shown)


AalbatrossGuy:
you gotta rebuild the lock file I presume considering the lock file checks failed

blankRiot96:
@AalbatrossGuy thanks for the notice

no it was because I forgot to add `is_prefix` to LLDB's API too
```py

        is_prefix=is_prefix,
        ^^^^^^^^^^^^^^^^^^^^
    )
    ^
TypeError: LLDB.add_command() got an unexpected keyword argument 'is_prefix'

```

I think I'll just take and ignore it for now?

k4lizen:
this is a bit of a conundrum
does `is_prefix` affect anything else from the perspective of gdb?

i think I am more wiling to bite the bullet on the fact that the output is duplicated than on the fact that `help cmd subcmd` doesn't work. but it feels reaaally scuffed that the output is duplicated.

A potential solution could be to just remove our argparse output and just have the gdb one, but:
1. i like the argparse one more
2. i don't know how that would play with lldb

I also potentially wouldn't mind a solution, where we don't change anything, but for commands with subcommands, a hint was printed that said that you have to use `-h` rather than `help`. But again, I would like to know if `is_prefix` affects anything else

blankRiot96:
@k4lizen no, `is_prefix` does not affect anything else in GDB and does not do anything at all in LLDB (im simply taking it and ignoring it)

if i delete the `test_consistent_help.py` testcase all the other testcases pass 

yes, i agree that if we're not liking help duplication, then at least we can print a message that `-h` should be used for `help` on subcommands

k4lizen:
Okay then could you:
1. Close this PR in its current state so we can refer back to the code and reasoning
2. Open a new PR where, for commands which have subcommands, you print this help message, something like "Hint: Use `actual_command_name subcommand --help` instead of `help actual_command_name subcommand`."

k4lizen:
Thanks for investigating this btw

blankRiot96:
Alright, thanks for the review

PR #3686, opened 2026-01-30
Fixup previous errno PR and fix lldb TLS resolution

k4lizen:
Also silence the vermin warning.


pwndbg/dbg_mod/lldb/__init__.py
patryk4815: ??

k4lizen: a good catch, my bad

PR #3687, opened 2026-01-30
Parallelize docker tests

jackmisbach:

This PR parallelizes the DBG-GDB and DBG-LLDB tests to speed up CI.

There are 2 options besides keeping as is, so I'd appreciate opinions.

1. Current state of this PR, parallelize DBG-GDB and DBG-LLDB
~12m 20s > ~8m 34s for the full test suite to run, including GDB tests
Example with a purposeful test failure: https://github.com/Monthly-Recurring-Revenue/pwndbg/actions/runs/21532428557/job/62051170529

2. Parallelize GDB, DBG-GDB, and DBG-LLDB
~12m 20s > ~7m 25s for the full test suite to run
Here, the benefits start to slow down because of CPU limitations, but it still saves an extra minute, surprisingly.
Example with a purposeful test failure: https://github.com/Monthly-Recurring-Revenue/pwndbg/actions/runs/21532428592/job/62051159246
Cons: Non-realtime, output is printed after tests are run, otherwise it is interweaved, which can be slightly confusing and not worth it imo. I'm not sure if people sit there and watch the workflow run, and want to see the tests run in real time. I don't really, but maybe some people do, and in that case, this could be a negative and not worth the extra time savings.

3. Keep as is 

Though if we choose option 2, or maybe even 1, we should have an entire summary at the end, so there is no need for scrolling up and down.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -69,20 +69,40 @@ jobs:
             -t ghcr.io/pwndbg/pwndbg:${{ matrix.image }} \
             .
 
-      - name: Run GDB Tests on ${{ matrix.image }}
-        run: docker compose run --rm ${{ matrix.image }} ./tests.sh -d gdb -g gdb
-
-      - name: Run DBG Tests (DBG-GDB & DBG-LLDB in parallel) on ${{ matrix.image }}
+      - name: Run Tests (GDB, DBG-GDB, DBG-LLDB in parallel) on ${{ matrix.image }}
         run: |
-          docker compose run --rm ${{ matrix.image }} bash -c '
+          docker compose run -T --rm ${{ matrix.image }} bash -c '
             set -euo pipefail
+            # Needed to prevent race conditions - prebuild binaries and create libpython symlink
             make -C /pwndbg/tests/binaries/host -j4 all
+            gdb --batch
+
             set +e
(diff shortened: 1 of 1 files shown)


k4lizen:
Lets go with 2. I agree that we should have a nice summary at the end.

We should also print a message like "Running generic tests on gdb, generic tests on lldb, and gdb-only tests on gdb at the same time. This should take around 8 minutes..."

jackmisbach:
Alright, everything should be good to go for option 2, parallelizing all tests. Let me know if you're fine with the combined test summary as is, or if you have any other ideas. I just put all of the test summaries at the end, basically, so you wouldn't have to scroll.
Here is an example with failed tests:
https://github.com/Monthly-Recurring-Revenue/pwndbg/actions/runs/21612220608/job/62283320480

Just as a note, this PR also adds -T and --rm flags to docker compose run
chrome_4oNDMo0x5i
Before this would've removed the warning 
chrome_uQTZHm5ABZ
Now that we are going with option 2, we only have one container running throughout the whole process, so it doesn't matter, but I think it is fine to maintain good hygiene.
chrome_DisPt4CmB1
The default is auto-detected, so I presume it was fine before, but I thought it wouldn't hurt to make this explicit.

k4lizen (approved):

k4lizen:
thanks!

PR #3692, opened 2026-02-01
Add hint to access subcommand help information

### LINKED ISSUE #3547: Investigate "help <cmd> <subcmd>" in GDB showing help for cmd instead of subcmd

blankRiot96:
Fixes #3547
See #3682 for rationale

It flew over my head when I suggested a hint for subcommand information that if it were possible to detect when the builtin `help <main> <sub>` were being used WITHOUT registering the main command as prefix command, we could have just printed the help information that way instead of a hint message 

So the only remaining way to add a hint is in every `help <main>` command in `CommandObj.register_command` for all pwndbg commands which have subcommands

image

### DIFF
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -233,6 +233,13 @@ class CommandObj:
         ) -> None:
             self.invoke(arguments, is_interactive)
 
+        if self.subcommand_names is not None and len(self.subcommand_names) > 0:
+            # In order to add `help <main> <sub>` support, the main
+            # command needs to be registered as a prefix command in
+            # GDB. Since this causes help info duplication, for now
+            # we simply show a hint to use `--help`
+            self.help_str += f"\nHint: Use `{self.command_name} <subcmd> --help` if you want to see subcommand information"
+
         # Keep a handle to the command and its aliases so we can
         # easily remove them if necessary (not supported with GDB).
         self.handles = [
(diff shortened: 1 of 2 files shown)


k4lizen:
oh I was fine with both `help di` and `di --help` showing this hint, it's cool that you figured out how to make only `help di` show it!

you should edit the test so that it checks that the two messages are the same *except* for this hint

k4lizen (approved):

k4lizen:
Thanks!

Issue #3668, opened 2026-01-23
debugger output shown for lldb test but not for gdb tests


k4lizen:
### Description

When an LLDB test fails, we get nice output like this that shows exactly all the commands ran and their output:

<img width="1127" height="926" alt="Image" src="https://github.com/user-attachments/assets/d7c634ed-cf40-461a-a9a1-ec02a29cce7a" />

But this does not happen for GDB for some reason:

<img width="1032" height="550" alt="Image" src="https://github.com/user-attachments/assets/ce293c48-125f-40ed-8bf4-a13b0629acbd" />

I wish I saw this output for GDB, as it is very helpful for test debugging

k4lizen:
hehe we don't have a "GDB-port" label :>

jackmisbach:
Added in PR https://github.com/pwndbg/pwndbg/pull/3698

PR #3698, opened 2026-02-02
Show debugger output for gdb tests

jackmisbach:

Fixes Issue: https://github.com/pwndbg/pwndbg/issues/3668

This was already handled for LLDB with the  REPL, which is why LLDB tests showed the nice output in tests, so I made the same thing occur for GDB tests.

I added two helper methods
`_gdb_execute()` to print the command that is being run
`_show_context()` to show context after stops 

I also made `execute_and_capture()` print the command being run and the output, which previously wasn't shown. It can be redundant for test_context_commands because context is then run twice; however, for almost all other tests, it isn't redundant.
LLDB doesn't currently show results of commands run with `execute_and_capture` like GDB now does, if this is something that would also be wanted lmk. You can see an example of what I mean here with my changes made and test_command_telescope set to fail purposefully: https://github.com/Monthly-Recurring-Revenue/pwndbg/actions/runs/21581242815/job/62178853948

### DIFF
--- a/pwndbg/dbg_mod/lldb/repl/__init__.py
+++ b/pwndbg/dbg_mod/lldb/repl/__init__.py
@@ -425,6 +425,7 @@ def run(
                                     action._command, output, dbg, driver, relay
                                 )
                                 last_result = output.buffer.getvalue()
+                                print(last_result.decode("utf-8", errors="surrogateescape"))
                         else:
                             should_continue = exec_repl_command(
                                 action._command, sys.stdout, dbg, driver, relay
--- a/tests/host/gdb/pytests_launcher.py
+++ b/tests/host/gdb/pytests_launcher.py
@@ -15,6 +15,21 @@ from ... import host
 
 
 class _GDBController(host.Controller):
+    def _gdb_execute(self, command: str) -> None:
+        """Execute a GDB command and print the command to output."""
+        print(f"pwndbg> {command}")
+        gdb.execute(command)
+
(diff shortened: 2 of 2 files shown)


k4lizen:
> LLDB doesn't currently show results of commands run with execute_and_capture like GDB now does, if this is something that would also be wanted lmk.

Yeah that would be awesome.

By the way, what is the difference between the testing setup and normal execution, which makes it so that in the testing `context` isn't called after a e.g. `continue`?

jackmisbach:
> By the way, what is the difference between the testing setup and normal execution, which makes it so that in the testing `context` isn't called after a e.g. `continue`?

vmware_rKdUPyOCtC

During normal use, the context command is called when stops occur via `prompt_hook()`, so just using `gdb.execute()` won't show context, which is why I was initially trying to actually just use `prompt_hook()` for the gdb tests. However, the test_cache_clear_has_priority.py test luckily found an issue because stop was called twice due to `after_reload()` being called in `prompt_hook()` `if cur != new`.

Also added the LLDB part, you can see here if you'd like to check with a failing test - 
https://github.com/Monthly-Recurring-Revenue/pwndbg/actions/runs/21609699420/job/62275232032
chrome_LDB0YWYvSo
For example, `ctrl.execute_and_capture("telescope &a")` in `test_command_telescope()` now shows the output.

k4lizen (approved):
Cool, thanks!

PR #3707, opened 2026-02-04
Docker tests - only build on PRs when Dockerfiles or setup scripts change

jackmisbach:

Just some finishing touches I wanted to make. 

Now that pulling images is available and is the default in the docs for running the Docker tests, I made the workflow no longer build on PRs unless any Dockerfile or setup scripts are changed. 

This way, they are built only on pushes (to push the latest dev image), unless needed. This also reduces time by an additional ~10%, roughly ~1m.

This PR also reduces the timeout-minutes from 30m > 15m to accurately reflect after how long something is wrong, and we should just timeout.

https://www.kenmuse.com/blog/the-many-shas-of-a-github-pull-request/#a-sha-for-every-occasion
https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#setting-an-output-parameter

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -42,7 +42,7 @@ jobs:
           - { image: fedora43, dockerfile: Dockerfile.dnf, base: fedora:43 }
 
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 15
     steps:
       - uses: actions/checkout@v6
         with:
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -45,8 +45,24 @@ jobs:
     timeout-minutes: 30
     steps:
       - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: Check for Dockerfile or setup changes
+        id: build-check
+        if: github.event_name == 'pull_request'
+        run: |
+          set -euo pipefail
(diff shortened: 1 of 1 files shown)


k4lizen:
Is a rebuild not needed on pyproject.toml / uv.lock updates?

jackmisbach:
> Is a rebuild not needed on pyproject.toml / uv.lock updates?

Nah, just because tests.sh uses common.sh which will do uv run, which calls uv sync as well, so if the image has different deps then the containers bind mount does, it'll sync to the bind mount and they'll get tested.

k4lizen (approved):
Aight, very cool, thanks :p

k4lizen:
> Nah, just because tests.sh uses common.sh which will do uv run, which calls uv sync as well, so if the image has different deps then the containers bind mount does, it'll sync to the bind mount and they'll get tested.

I was also thinking about if it would play well with someone who doesn't want to run the tests but just wants to run pwndbg, but I guess our syncing that we do when pwndbg is started should be robust.

jackmisbach:
> I was also thinking about if it would play well with someone who doesn't want to run the tests but just wants to run pwndbg, but I guess our syncing that we do when pwndbg is started should be robust.

If someone is pulling an image, it'll always have the latest deps in the image.
Then when they run pwndbg inside, it runs uv sync just like tests would but with `verify_venv()` from common.py instead of common.sh. So they will be using the deps they have locally in uv.lock if they differ. So yeah it is robust and will work all good!

Issue #3506, opened 2025-12-15
fix up vis-skip-repeating-val


k4lizen:
The idea behind the feature is very nice but the way it is currently makes it harder to parse the output than with it disabled:

<img width="766" height="500" alt="Image" src="https://github.com/user-attachments/assets/93cb5430-cd56-42ee-be49-37f2008b3ca7" />

Firstly, there is a bug where it doesn't properly override the skip caused by max-visualize-chunk-size.

Secondly, it should only perform the skip if the repeat is higher than some base value. For telescope this is controlled by `telescope-skip-repeating-val-min` (default 3), probably the default could be something like 5.

Finally, the "4 repeated lines skipped" text itself should be the same color as the chunk, because otherwise there is too much visual noise.

k4lizen:
> Firstly, there is a bug where it doesn't properly override the skip caused by max-visualize-chunk-size.

Not only is it a visual artifact, but it also actually just reports the wrong number as well

k4lizen:
for reference, I have `set max-visualize-chunk-size 0x50`, and I think this is kind of where you want to be.

<img width="840" height="779" alt="Image" src="https://github.com/user-attachments/assets/841b0a5f-6905-4692-b60d-7baee40a0344" />

you can see the first 4 ptrs of every chunk which is crucial, and a bit at the end as well (could be one line shorter imo but w/e). `vis-skip-repeating-val` should *not* collapse the whole chunk here. 

<img width="720" height="360" alt="Image" src="https://github.com/user-attachments/assets/652d92d8-b88b-43a2-a617-be0ee2da0206" />

you want to see the addresses for the start and end of the chunk because they allow for easy parsing/copying

k4lizen:
i think we shouldn't release a new pwndbg version until we fix this

PR #3708, opened 2026-02-04
Set vis-skip-repeating-val to False by default

k4lizen:
Since it's a bit clunky now (https://github.com/pwndbg/pwndbg/issues/3506) and doesn't play well with an existing arg and we want to cut a release, lets at least set it to False by default.


k4lizen:
@disconnect3d ping for visibility

Issue #2688, opened 2025-01-21
Handle macOS SIP protected binaries gracefully


disconnect3d:
When one tries to debug SIP protected binaries on SIP enabled system, Pwndbg crashes with the following error.

We should probably detect that the target binary can't be run due to SIP and handle this gracefully by printing a nice error message to the user explaining what is going on.

Below you can see how this works as of now. Triggered on macOS 15.0.1 on Intel cpu.

```
➜  bin ./pwndbg-lldb /bin/ls
pwndbg: loaded 145 pwndbg commands and 0 shell commands.
------- tip of the day (some of these don't work in LLDB yet!) (disable with set show-tips off) -------
Use the killall command to kill all specified threads (via their ids)
pwndbg-lldb> target create '/bin/ls'
Current executable set to '/bin/ls' (x86_64)
pwndbg-lldb> run -s
Exception occurred: Error: 'NoneType' object has no attribute 'regs' (<class 'AttributeError'>)
For more info invoke `set exception-verbose on` and rerun the command
or debug it by yourself with `set exception-debugger on`
Traceback (most recent call last):
  File "/Users/dc/tob/pwndbg/exe/pwndbg-lldb.py", line 113, in <module>
    run_repl([f"target create '{target}'"] if target else None, debug=debug)
  File "/Users/dc/tob/pwndbg/lib/python3.12/site-packages/pwndbg/dbg/lldb/repl/readline.py", line 82, in _wrapped
    return function(*a, **kw)
           ^^^^^^^^^^^^^^^^^^
  File "/Users/dc/tob/pwndbg/lib/python3.12/site-packages/pwndbg/dbg/lldb/repl/__init__.py", line 336, in run
    process_launch(driver, relay, bits[1:], dbg)
  File "/Users/dc/tob/pwndbg/lib/python3.12/site-packages/pwndbg/dbg/lldb/repl/__init__.py", line 630, in process_launch
    result = driver.launch(
             ^^^^^^^^^^^^^^
  File "/Users/dc/tob/pwndbg/lib/python3.12/site-packages/pwndbg/dbg/lldb/repl/proc.py", line 410, in launch
    self.eh.created()
  File "/Users/dc/tob/pwndbg/lib/python3.12/site-packages/pwndbg/dbg/lldb/repl/__init__.py", line 122, in created
    self.dbg._trigger_event(EventType.START)
[...]

PR #3710, opened 2026-02-04
Add SIP hint on LLDB launch/attach failure on macOS

jackmisbach:

finishing touches to close issue https://github.com/pwndbg/pwndbg/issues/2688

I tried to do this and thought it would be nice to get a hint, and then saw there was already an issue for this.

Screenshot 2026-02-04 at 3 31 12 PM


pwndbg/dbg_mod/lldb/repl/__init__.py
```diff
@@ -1248,6 +1252,10 @@ def _attach_with_info(
             return
         case LaunchResultEarlyExit():
             print_warn("process exited early")
+            if sys.platform == "darwin":
+                print_hint(
+                    "It may be that you are trying to attach to a protected binary on a SIP enabled system."
+                )
             auto.close()
```
disconnect3d, line 1259: if `auto` would be a context manager, we could probably have this match case and the one in process_launch to be using the same codebase (the only difference is "launch" vs "attach") but yeah, can be done in another PR or left as is.

disconnect3d (approved):

PR #3717, opened 2026-02-05
use dorny/paths-filter instead of on.paths

jackmisbach:

When only docs are changed in a PR, branch protection will sit and wait for the checks to report back, but they will never run because using on.paths won't mark the workflow as skipped, but rather just never start it, as it should, but I think that this should differ for branch protection expected checks; however, it doesn't. So we have to work around it.

chrome_7BrszAWhQC

So there are three options.
1. Keep as is and just bypass branch protection; this is what is done for docs-only PRs currently. not optimal
2. Remove !docs from on.paths, and run tests even on docs-only PRs. also not optimal

3. What is done in this PR, which I think is the best option we have, is to use the dorny/path-filter action instead of on.paths so that the workflows will start, and then be skipped, which will be considered pass and auto-merge to work fine and branch protection to pass. I don't like that it adds another job though...

I added it to the Docker tests as well. Even though it isn't included in the required branch protection tests, I wanted it to be consistent.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -4,17 +4,7 @@ on:
   push:
     branches:
       - dev
-    paths:
-      - "**"
-      - "!mkdocs.yml"
-      - "!docs/**"
-      - "!*.md"
   pull_request:
-    paths:
-      - "**"
-      - "!mkdocs.yml"
-      - "!docs/**"
-      - "!*.md"
 
 concurrency:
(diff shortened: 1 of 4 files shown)


k4lizen:
> I don't like that it adds another job though...

Yeah agreed, but I think we want this now, and can maybe look into optimizations in the future.

k4lizen (approved):
thanks!

PR #3722, opened 2026-02-08
Add cache mounts for build base image tests

jackmisbach:

When Dependabot creates PRs or dependencies are updated, the build base pwndbg image workflows can take up to 3 hours because they must build from source for many packages on all arches except amd64 and arm64, with these packages specifically taking exceptionally long. 
chrome_277gnUbdTZ

This uses 8/20 concurrent jobs, and Dependabot usually opens a couple of PRs at once, which takes up almost all concurrent job capacity for hours. This can be tedious if you just pushed a PR after, as you have to wait hours for your jobs to finally start.

Unlike the registry cache, where a change invalidates the entire layer, the same doesn't apply to cache mounts; only the changed deps will be rebuilt, as everything else will still be cached using GitHub Actions Cache.

It uses roughly 3.6GB of cache across all arches, with a limit of 10GB of free cache for public repos. So this is fine, and qemu images are our only other large cache, being 3.2GB. Leaving us at about 6.8GB with the rest being smaller caches that don't have deletion setup for stale cache, which is fine because least recently used will automatically be deleted when the limit is reached, but it may be something I clean up later for good practice with some of the other workflows.

https://docs.astral.sh/uv/guides/integration/docker/#caching
https://github.com/reproducible-containers/buildkit-cache-dance

### DIFF
--- a/.github/workflows/base-image.yml
+++ b/.github/workflows/base-image.yml
@@ -92,7 +92,7 @@ jobs:
 
       - name: Restore Docker cache mounts
         id: cache
-        uses: actions/cache@v4
+        uses: actions/cache@v5
         with:
           path: |
             uv-${{ matrix.platform_base.tag_name }}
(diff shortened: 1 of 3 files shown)


k4lizen:
cache the world


thanks!

PR #3724, opened 2026-02-08
Better tracking of sequence of instructions executed at runtime

### LINKED ISSUE #3721: branch annotation for x86_64 loop instruction is wrong
### Description

Image

Clearly in the previous iterations, we did take the loop jump, and only didn't take it in the last one. The emulation is correct but the annotation is wrong.

OBarronCS:
In the disassembly display, we disassemble both forwards and backwards from the instruction pointer to display the future and the previous instructions that have executed. In order to disassemble backwards, we need to track how we got to the current instruction.

The previous method of tracking the sequence of instructions did not handle loops well (for each address, we could only remember a single "previous" address, which breaks down in while entering loops, as you can enter the basic block in multiple ways). This PR creates a better way to track the actual sequence of instructions executed at runtime.

This fixes errors when stepping through and displaying tight loops (where the entire basic block of the loop fits on the display). Previously, when stepping into such a loop, we would lose track of where we came from. Now, we accurately track the sequence of instructions that have actually executed, and correctly disassemble backwards from the instruction pointer.

image

Fixes #3721


OBarronCS:
I added a test for stepping through tight loops as well

pwndbg/aglib/disasm/disassembly.py
```diff
@@ -193,25 +240,49 @@ def get_one_instruction(
...
+            linear_backward_cache[insn.address + insn.size] = insn.address
+
+        if put_cache:
+            computed_instruction_cache[address] = insn
+
+    return insn
+
+
 # Return None on failure to fetch an instruction
 def one(
```
k4lizen: should code that wants to disassemble something use `one` or `one_with_cache_policy`?

k4lizen (approved):

tests/library/dbg/tests/test_disassembly.py
```diff
@@ -0,0 +1,107 @@
...
+
+from ....host import Controller
+from . import get_binary
+from . import pwndbg_test
+
+LOOP_INSTRUCTION_BINARY = get_binary("loop_instruction.x86-64.out")
+
+
+@pwndbg_test
+async def test_context_disasm_loop_instruction(ctrl: Controller) -> None:
```
k4lizen, line 11: how come this test doesn't fail on our aarch64 host CI?

OBarronCS: I had a similar question, but then I saw this existing test https://github.com/pwndbg/pwndbg/blob/28208ec40953a3acb0a19bef0d09e3d6aaf185c4/tests/library/dbg/tests/test_command_stepsyscall.py#L7 which also worked.

It looks like the `launch` method checks if the compiled binary exists, and skips if not: https://github.com/pwndbg/pwndbg/blob/28208ec40953a3acb0a19bef0d09e3d6aaf185c4/tests/host/gdb/pytests_launcher.py#L33-L44

And the makefile only creates the binary if the arch matches the name of the assembly file.

pwndbg/aglib/disasm/disassembly.py
OBarronCS: The `one_with_cache_policy` should be used when disassembling many instructions back-to-back in sequence, as the caching is used to track the order of instructions and save enhanced instructions (relevant to the disasm display). I created this function because before the caching stuff was deeply nested within multiple functions calls and was unnecessary. I'll add a comment.

tests/library/dbg/tests/test_disassembly.py
k4lizen: a nice!

pwndbg/aglib/disasm/disassembly.py
OBarronCS: I ended up reverting this change, and we just now have `one()`. It was just another function that didn't really provide anything new

k4lizen (approved):

PR #3736, opened 2026-02-11
Nearpc: disassemble entire function

OBarronCS:
Allow running nearpc to disassemble entire functions at a time.

image


OBarronCS:
In GDB
image


In LLDB:

image

pwndbg/dbg_mod/lldb/__init__.py
```diff
@@ -1636,6 +1636,19 @@ def lookup_symbol(
...
+
+        func = addr.GetFunction()
+        if func.IsValid():
+            start = func.GetStartAddress().GetLoadAddress(self.target)
+            end = func.GetEndAddress().GetLoadAddress(self.target)
+
+            return start, end
+
+        return None
+
```
OBarronCS, line 1653: I am not super familiar with the LLDB API and I'm sure this is missing something, given the complexity of lookup in other functions, but these functions are what I found in the various LLDB docs.

pwndbg/dbg_mod/__init__.py
```diff
@@ -583,6 +583,8 @@ def lookup_symbol(
         Raises:
         - pwndbg.dbg_mod.Error: If no object file matching the `objfile_endswith` pattern is found.
         """
+
+    def get_function_boundaries(self, address: int) -> tuple[int, int] | None:
```
OBarronCS, line 587: I believe this was an appropriate time to add to the dbg API? The logic to get function boundaries depends on the underlying debugger.

pwndbg/commands/nearpc.py
```diff
@@ -94,6 +107,26 @@ def nearpc(
         # -t was specified
         back_lines = min(int(nearpc_backwards_lines), total - 1)
 
+    end_address = None
+    if function is not None:
+        pc = lookup_symbol_addr(function)
```
OBarronCS: TODO: like the GDB `disass` command, allow the user to pass in an address, and disassemble the entire function that contains that address.

pwndbg/dbg_mod/gdb/__init__.py
```diff
@@ -954,6 +954,18 @@ def lookup_symbol(
...
+        block = gdb.block_for_pc(address)
+
+        if block is not None:
+            while block.function is not None:
+                block = block.superblock
+
+            return block.start, block.end
+
+        return None
+
```
jxuanli, line 969: does this work in the kernel context?

pwndbg/commands/nearpc.py
OBarronCS: Done

pwndbg/dbg_mod/gdb/__init__.py
OBarronCS: I tested it using the kernels in our kernel tests, and it worked.

image

OBarronCS:
The command now follows the same behavior as GDB's built-in `disassemble` command, in that you can provide an expression, a symbol, or an address, and it will resolve the address that the input evaluates to, and disassembles the entire function in which that address resides in.

disconnect3d (approved):

pwndbg/dbg_mod/gdb/__init__.py
k4lizen: ~~no sorry, you're gonna have to gate this off for the kernel. https://github.com/pwndbg/pwndbg/pull/3730~~

ehh actually, if it doesn't touch the blocks symbols maybe we are fine. lets leave it in for now

pwndbg/aglib/symbol.py
```diff
@@ -115,3 +115,14 @@ def resolve_addr(addr: int) -> str | None:
         return symbol_name
 
     return pwndbg.dintegration.manager.symbol_at_address(addr)
+
+
+@pwndbg.lib.cache.cache_until("objfile")
+def resolve_function_boundaries(addr: int) -> tuple[int, int] | None:
+    """
+    Return the function start and end address for a function that
+    contains address `addr`.
```
k4lizen, line 124: it'd be nice to clarify whether the end address is exclusive or not

pwndbg/commands/nearpc.py
```diff
@@ -55,6 +58,14 @@
     help="Disable showing branch visualizations.",
 )
 
+parser.add_argument(
+    "-f",
+    "--function",
+    type=str,
```
k4lizen: why doesn't passing `int` here work? GDB will still parse function names etc

pwndbg/dbg_mod/gdb/__init__.py
```diff
@@ -954,6 +954,18 @@ def lookup_symbol(
             raise pwndbg.dbg_mod.Error(e)
         return None
 
+    @override
+    def get_function_boundaries(self, address: int) -> tuple[int, int] | None:
+        block = gdb.block_for_pc(address)
+
+        if block is not None:
+            while block.superblock.function is not None:
```
k4lizen: this is a bit risky, you should also be checking that block.superblock isn't None

pwndbg/dbg_mod/lldb/__init__.py
```diff
@@ -1636,6 +1636,19 @@ def lookup_symbol(
...
         return self.create_value(resolved_addr, cast_type)
 
+    @override
+    def get_function_boundaries(self, address: int) -> tuple[int, int] | None:
+        addr = self.target.ResolveLoadAddress(address)
+
+        func = addr.GetFunction()
+        if func.IsValid():
+            start = func.GetStartAddress().GetLoadAddress(self.target)
+            end = func.GetEndAddress().GetLoadAddress(self.target)
```
k4lizen: i would also likely check https://lldb.llvm.org/python_api/lldb.SBAddress.html#lldb.SBAddress.IsValid 
and `return int(start), int(end)`

pwndbg/dbg_mod/__init__.py
k4lizen: yeah I think it makes sense. I would add the same docstring from `/symbol/` though, also note on whether the end address is exclusive or not, and likely also make this function cached

pwndbg/commands/nearpc.py
OBarronCS: For some reason I was under the impression that doing raw symbol names (`execve`) wouldn't resolve correctly (that you had to do `&execve`). But this is not the case. Changing this to int - this cleans this up a lot.

pwndbg/aglib/symbol.py
OBarronCS: I added a comment here + the dbg_mod interface

pwndbg/dbg_mod/gdb/__init__.py
OBarronCS: Good call - fixed it

pwndbg/dbg_mod/lldb/__init__.py
OBarronCS: I added a check here

OBarronCS:
Thanks for the review - I made the suggested changes.

tests/library/dbg/tests/test_musl.py
```diff
@@ -13,7 +13,7 @@
 async def test_musl_dynamic_detection(ctrl: Controller) -> None:
     import pwndbg.libc
 
-    ctrl.disable_debuginfod()
+    await ctrl.disable_debuginfod()
```
OBarronCS, line 16: Unrelated but adding an await here fixed this test from failing: https://github.com/pwndbg/pwndbg/actions/runs/21959148598/job/63431577025

OBarronCS:
Current test failures are unrelated: 
image

tests/library/dbg/tests/test_musl.py
k4lizen: holyy you're a lifesaver, how did you even figure it

pwndbg/dbg_mod/lldb/__init__.py
```diff
@@ -1636,6 +1636,21 @@ def lookup_symbol(
...
+        addr = self.target.ResolveLoadAddress(address)
+
+        if not addr.IsValid():
+            return None
+
+        func = addr.GetFunction()
+        if func.IsValid():
+            start: int = func.GetStartAddress().GetLoadAddress(self.target)
+            end: int = func.GetEndAddress().GetLoadAddress(self.target)
+            return start, end
```
k4lizen, line 1650: heh I meant adding the IsValid check to start and end, because yeah lldb functions like not documenting that they can return invalid objects, but looking at the actual implementation
```cpp
SBAddress SBFunction::GetStartAddress() {
  LLDB_INSTRUMENT_VA(this);

  SBAddress addr;
  if (m_opaque_ptr)
    addr.SetAddress(m_opaque_ptr->GetAddress());
  return addr;
}

SBAddress SBFunction::GetEndAddress() {
  LLDB_INSTRUMENT_VA(this);

  SBAddress addr;
  if (m_opaque_ptr) {
    AddressRanges ranges = m_opaque_ptr->GetAddressRanges();
    if (!ranges.empty()) {
      // Return the end of the first range, use GetRanges to get all ranges.
      addr.SetAddress(ranges.front().GetBaseAddress());
      addr->Slide(ranges.front().GetByteSize());
    }
  }
  return addr;
}
```
It seems like it will always be valid (i looked at the callees as well), so in the end I think this is good

k4lizen (approved):
Very nice QoL!

We'll need to wait for @disconnect3d to merge by bypassing branch protection though

pwndbg/dbg_mod/lldb/__init__.py
OBarronCS: Ohh I see - I was looking only at `.GetLoadAddress()` (not `GetStartAddress()`) and seeing that it returns an int not an SBAddress.

disconnect3d:
This is AWESOME. Thanks for this!

jxuanli:
`get_function_boundaries` doesn't work if the target doesn't have debug info. also should the function take a string as its arg instead (based on the comments)? @OBarronCS

OBarronCS:
The `int` arg on the argparser makes the conversion happen by default - the underlying debugger will convert any symbols and expression strings into an int.

I don't know how GDB computes `block_for_pc` it under the hood. It's also failing on a program with a patched libc
image

Issue #3631, opened 2026-01-15
running pwndbg gives: GDB is compiled for python x.y, but your interpreter is x.z


k4lizen:
### Description

I recently updated my artix installation which bumped python to 3.14. I did the following:
1. Recompiled gdb for python 3.14
2. Deleted the pwndbg .venv
3. Re-ran `./setup.sh`
4. Ran `uv tool uninstall pwndbg` 
5. Ran `uv tool install .`

But still when I run:
```
~❯ pwndbg ./vuln
ERROR: GDB is compiled for Python ('libpython3.14.so.1.0', '3.14'), but your Python interpreter is ('libpython3.12.so.1.0', '3.12')
```
How? Where is that coming from?

```
~❯ uv run python -V
Python 3.14.2
```

Checking the pwndbg source:
```python
    expected = (sysconfig.get_config_var("INSTSONAME"), sysconfig.get_config_var("VERSION"))
    have = get_gdb_version(gdb_path)
    if have != expected:
        print(
            f"ERROR: GDB is compiled for Python {have}, but your Python interpreter is {expected}"
        )
        sys.exit(1)
```

But when I run a python interactive session in the terminal:
```
~❯ ipython
In [1]: import sysconfig

In [2]: sysconfig.get_config_var("INSTSONAME")
Out[2]: 'libpython3.14.so.1.0'

In [3]: sysconfig.get_config_var("VERSION")
Out[3]: '3.14'
```
So where is 3.12 coming from?

patryk4815:
@k4lizen what this commands returns?

`type pwndbg`
`cat /path/returned/from/type/command/pwndbg`

k4lizen:
```
~❯ type pwndbg
pwndbg is /home/user/.local/bin/pwndbg
user@host ~ 
~❯ cat /home/user/.local/bin/pwndbg
#!/home/user/.local/share/uv/tools/pwndbg/bin/python
# -*- coding: utf-8 -*-
import sys
from pwndbginit.pwndbg_gdb import main
if __name__ == "__main__":
    if sys.argv[0].endswith("-script.pyw"):
        sys.argv[0] = sys.argv[0][:-11]
    elif sys.argv[0].endswith(".exe"):
        sys.argv[0] = sys.argv[0][:-4]
    sys.exit(main())
```

patryk4815:
`/home/user/.local/share/uv/tools/pwndbg/bin/python` what python version is this?

k4lizen:
[...]

PR #3737, opened 2026-02-11
Add gdbinit.py into venv/share/ dir

patryk4815:
Why we want add `gdbinit.py` to ./share directory?
- To support something like (ubuntu24.04):  
```bash
apt update
apt install -y git curl gdb python3

curl -LsSf https://astral.sh/uv/install.sh | sh
source $HOME/.local/bin/env

PY_VER=$(gdb -nx --batch -iex 'py import sysconfig; print(sysconfig.get_config_var("VERSION"))')
uv tool install --python=$PY_VER --no-managed-python git+https://github.com/pwndbg/pwndbg

echo "source $(uv tool dir)/pwndbg/share/pwndbg/gdbinit.py" >> ~/.gdbinit

gdb
```
Example1:
image

Example2:
image

### DIFF
--- a/gdbinit.py
+++ b/gdbinit.py
@@ -33,6 +33,7 @@ def fixup_paths(src_root: Path, venv_path: Path):
     # sys.prefix must be changed to point to the virtual environment.
     # This is what python expect: https://docs.python.org/3/library/sys.html#sys.prefix
     sys.prefix = str(venv_path)
+    sys.exec_prefix = str(venv_path)
 
 
 def get_venv_path(src_root: Path):
--- a/gdbinit.py
+++ b/gdbinit.py
@@ -39,6 +39,16 @@ def get_venv_path(src_root: Path):
     venv_path_env = os.environ.get("PWNDBG_VENV_PATH")
     if venv_path_env:
         return Path(venv_path_env).expanduser().resolve()
+
+    # Handle case when `gdbinit.py` is running from inside venv, eg: `venv/share/pwndbg/gdbinit.py`
+    # See, example usage: https://github.com/pwndbg/pwndbg/pull/3737
+    if (
+        src_root.parent.name == "share"
+        and src_root.name == "pwndbg"
+        and (src_root / "../../pyvenv.cfg").exists()
+    ):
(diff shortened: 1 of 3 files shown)


patryk4815:
Why `--no-managed-python` is required?

Basically we shouldn’t mix system `gdb` (which is linked against the system `libpython.so`) with a Python that comes from `uv`....

`uv` installs its own Python build, compiled differently than the system one. We do playing with `PYTHONPATH`, so we easily end up with a weird hybrid where some things come from the system Python and some from the `uv` one - and that’s when random stuff like missing `_lsprof` starts happening.

So the fix is simple: don’t use the `uv`-managed Python here. Run with:
`--no-managed-python` and just use the system Python instead....

Maybe we remove that restriction, by better manipulating `PYTHONPATH` / `sys.path` - (better pickup what things should come from system, and what from venv)

image

k4lizen:
so this is specifically when someone wants to:
1. install from `git+https://github.com/pwndbg/pwndbg` and doesn't have the repo cloned
2. doesn't want to use `pwndbg` but wants to use `gdb`

Yes?

Btw in this case, if they run `pwndbg` they will get the double-loading error right?

patryk4815:
> Btw in this case, if they run `pwndbg` they will get the double-loading error right?

Yes 😿

k4lizen (approved):
Makes sense to me, I only wish we had some of this stuff explained somewhere in the docs

patryk4815:
@k4lizen Ok, I fixed all bugs. `--no-managed-python` restriction is not needed anymore

k4lizen:
heh I would test it but I still cant run `pwndbg` / `pwndbg-lldb`: https://github.com/pwndbg/pwndbg/issues/3631#issuecomment-3763610457

patryk4815:
> heh I would test it but I still cant run `pwndbg` / `pwndbg-lldb`: [#3631 (comment)](https://github.com/pwndbg/pwndbg/issues/3631#issuecomment-3763610457)

Did you ask this problem in uv-astral discord?

k4lizen:
haven't, i'll ask

PR #3738, opened 2026-02-11
show offset from heap instead of full address

### LINKED ISSUE #3091: heap tracker: show offset from heap instead of full address?

D3LET3-57:
+ [ x] I read the contributing documentation.
+ [ x] I am providing a screenshot of the (new feature) / (fixed bug).
Here are ss of the feature.

image
image

Fixes: https://github.com/pwndbg/pwndbg/issues/3091

### DIFF
--- a/pwndbg/gdblib/ptmalloc2_tracking.py
+++ b/pwndbg/gdblib/ptmalloc2_tracking.py
@@ -76,6 +76,7 @@ last_issue: str | None = None
 
 # Useful to track possbile collision errors.
 PRINT_DEBUG = False
+HEAP_OFFSET_MODE = False
 
 PTRS_COLORS = (
     pwndbg.color.red,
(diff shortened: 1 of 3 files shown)


pwndbg/gdblib/ptmalloc2_tracking.py
```diff
@@ -217,13 +218,25 @@ def __init__(self) -> None:
...
         if thread not in self.memory_management_calls:
             return False
         return self.memory_management_calls[thread]
 
+    def get_heap_region(self) -> tuple[int, int] | None:
+        """
+        Returns the base and end address of the heap region, if it can be determined.
+        """
+        for page in pwndbg.aglib.vmmap.get():
+            if page.objfile == "[heap]":
```
k4lizen: will this work if there are multiple heap arenas / multiple threads?

D3LET3-57: i didn't think about it i will try to test it.

k4lizen:
cc: @disconnect3d (since it's your feat req)

disconnect3d:
I've provided some feedback on Discord here https://discord.com/channels/843809097920413717/843822664841166848/1471132255220142240

I'll pass any further feedback here instead, after fixes are pushed

D3LET3-57:
@disconnect3d i've added requested changes
and as @k4lizen have asked it is not valid for multithread process since we need to fetch every heap arena and assign a heap base for every thread.
i am having a rough idea like this:
make a list of all heap areans start and end address and check ptr for it's respective heap and print it's offset and also heap may increase so we need to actively update our list. I think it is tough to implement...

k4lizen:
maybe you could also just look up the name of the mapping where the pointer is and use the start of that? `vmmap.addr_region_start` i think.

maybe it would be slow?

disconnect3d:
Can you check out what @k4lizen points out, please?

D3LET3-57:
image
<br>
output when we run it by default
<br>
<br>


image
image
<br>output with the feature  `track-heap enable -r` .

image
image
<br>
output with `track-heap enable`

D3LET3-57:
@k4lizen i have implemented it using `pwndbg.aglib.vmmap.addr_region_start`

k4lizen:
seems that `realloc` addresses aren't properly relativized, but otherwise lgtm

also I would maybe separate the output like `base + offset` instead of `base+offset` to make the offset easier to copy-paste

Issue #3495, opened 2025-12-13
Enable skipped tests on aarch64 by making them architecture-agnostic (x86-64 & aarch64)


patryk4815:
A large number of tests under `tests/library/dbg/tests/` are currently marked as `SKIPPED` when running the test suite on `aarch64`. These tests pass on `x86-64`, but their current implementation relies on x86-specific assumptions (e.g. registers, ABI details, instruction layout, or heap internals), which prevents them from running on aarch64.

The goal of this task is to rewrite or refactor these tests so that they are architecture-agnostic and can be executed on both x86-64 and aarch64, without unconditional `pytest.skip` based on the architecture.

### Scope of work
- The work does **not** need to be completed in a single pull request — individual tests or small groups of tests can be refactored independently and submitted as separate PRs, allowing multiple contributors to work in parallel.
- Identify and remove x86-64–specific assumptions in the affected tests (e.g. register names, instruction encodings, ABI- or heap-layout details).
- Refactor tests to rely on pwndbg architecture abstractions where available.
- Use test parameterization or architecture-aware helpers instead of unconditional `pytest.skip`.
- Replace architecture-dependent assertions with conditional assertions when behavior is expected to differ.
- Keep `pytest.skip` only for cases where a feature is genuinely unavailable on a given architecture.

### Acceptance criteria

- At least one previously skipped test is made to run successfully on aarch64 without breaking x86-64 support.
- Refactored tests are no longer marked as **SKIPPED** on aarch64.
- The test suite continues to pass on:
  - x86-64
  - aarch64
- No regressions are introduced for existing x86-64 test coverage.
- Tests remain readable, maintainable, and consistent with existing pwndbg test patterns.

### Affected tests (currently skipped on aarch64)
```
tests/library/dbg/tests/heap/test_find_fake_fast.py::test_find_fake_fast_command
tests/library/dbg/tests/heap/test_heap.py::test_global_max_fast_heuristic
tests/library/dbg/tests/heap/test_heap.py::test_jemalloc_extent_info
tests/library/dbg/tests/heap/test_heap.py::test_jemalloc_find_extent
tests/library/dbg/tests/heap/test_heap.py::test_jemalloc_heap
tests/library/dbg/tests/heap/test_heap.py::test_malloc_chunk_command
tests/library/dbg/tests/heap/test_heap.py::test_malloc_chunk_command_heuristic
tests/library/dbg/tests/heap/test_heap.py::test_malloc_chunk_dump_command
[...]

PR #3741, opened 2026-02-12
Make test_launch and test_windbg tests architecture-agnostic

carlpinto25:
issue - #3495
made test_launch and test_windbg tests architecture-agnostic


tests/library/gdb/tests/test_windbg.py
```diff
@@ -4,15 +4,12 @@
 
 import pwndbg.aglib
 import pwndbg.aglib.memory
-import pwndbg.aglib.vmmap
 
 from . import get_binary
 
-MEMORY_BINARY = get_binary("memory.x86-64.out")
+MEMORY_BINARY = get_binary("memory.native.out")
```
OBarronCS: Nice! It's great to move more tests to work across architecture.

This file (the `gdb/tests/test_windbg.py`) can be removed. For background - a while back many of the tests were copied from `gdb` (only run for GDB) to `dbg` (work in all underlying debuggers, these tests run for both GDB and LLBD) when adding support for LLDB. During the transition some of the `gdb` test files were kept intact.

Now that we know things are all working in the `dbg` version, we can delete this test file (the `dbg` version runs the test with GDB as well).

carlpinto25: oh .ive deleted it now

carlpinto25: @OBarronCS some test that arent related to this issue and  were passing before are failling now, though i cant understand why

OBarronCS:
Regarding the unrelated flaky tests: the coredump test sometimes fails for unknown reasons (the kernel version used to run the CI job possibly changes), so don't worry about that one (hopefully we can make it more stable).

image


~~One more thing: I doubled check the deleted test file, and can you make sure that all of the tests that existed in the `gdb` file have been moved to the `dbg` file? It looks like `test_windbg_eX_commands` wasn't copied over initially, same with `test_windbg_commands_x86`~~ nvm, they have been moved but were hidden in the GitHub UI.

Thanks for doing this!

OBarronCS (approved):

PR #3742, opened 2026-02-12
Fix bata24/gef size in the README

k4lizen:
@bata24


patryk4815:
I think we should add a workflow that will periodically download the bata/gef size and update it in our readme, it will save us a lot of time

disconnect3d:
> I think we should add a workflow that will periodically download the bata/gef size and update it in our readme, it will save us a lot of time

You sound like AI xd

PR #3747, opened 2026-02-14
[docs] Add docs about implementing an allocator inspector

k4lizen:
Also moves `pwndbg.aglib.heap.ptmalloc.SymbolUnresolvableError` -> `pwndbg.lib.SymbolNotRecovered`.


CptGibbon:
Appreciate you adding this.
Sound advice, if perhaps a tad daunting for a first-time contributor.
Do you prefer `SymbolNotRecoveredError` or just `SymbolNotRecovered` wrt the exception name?

docs/contributing/allocator.md
```diff
@@ -0,0 +1,206 @@
...
+
+### Use debug info if it is present
+
+See [Implementing Libc support#Debug Info](./libc-provider.md#debug-info) for rationale.
+
+The mallocng implementation unfortunately does not do this, the jemalloc implementation unfortunately only does this (i.e. does not support jemalloc without debug info), the glibc implementation handles this pretty cleanly by having one class which uses debug info and one class which uses heuristics.
+
+The SLUB implementation handles this the best, see `pwndbg/aglib/kernel/__init__.py:typeinfo_recovery()`. If debug info is present, we do nothing, otherwise we add the debug info ourselves using the `cymbol` API. From then on we can simply save assume that we "have" debug info.
+
+### Fail gracefully
```
jxuanli, line 148: this isn't specifically just for allocators

docs/contributing/allocator.md
```diff
@@ -0,0 +1,206 @@
...
+
++ 2.42: start of changelog (#3272, #3464, #3487)
+
+### kernel SLUB
+
+Tested for [???, 6.18\].
+
++ 6.18: start of changelog (#3689)
+
+CONFIG_SLAB_VIRTUAL support: mitigation-v4-6.12 with debug info > older versions with debug info > mitigation-v4-6.12 without debug info > older versions without debug info.
```
jxuanli:
```suggestion
`CONFIG_SLAB_VIRTUAL` support: until the mitigation feature gets accepted into mainline, only the latest patch (currently `mitigation-v4-6.12`) is fully supported. The `slab` commands are not guaranteed to work on older patches because the feature is experimental and may undergo significant changes.
```

jxuanli (changes requested):

docs/contributing/allocator.md
jxuanli: but I guess I should write a doc on how to contribute to the kernel subsystem at some point.

docs/contributing/allocator.md
k4lizen: thanks, do you also maybe know what the lowest version of the kernel is that we support?

oh and feel free to retroactively add entries to the changelog, like what was the difference between 6.17 and 6.18 that made it break?

k4lizen:
For reference, this is how bata24/gef's `heap -h` changelog looks like:
image

i think it's really nice

k4lizen:
>  Sound advice, if perhaps a tad daunting for a first-time contributor.

heh thats probably because the task is actually a tad daunting

I think the hardest part of writing an allocator inspector is probably actually taking the time to fully understand how the allocator works. If they cross this threshhold, it is really valuable to make the implementation robust.

>  Do you prefer SymbolNotRecoveredError or just SymbolNotRecovered wrt the exception name?

changed them to the former, I was also thinking about this

PR #3757, opened 2026-02-16
Adding support for breaking at the start of a JIT compiled eBPF prog

jxuanli:
## demo

image


pwndbg/commands/kbpf.py
```diff
@@ -184,20 +207,25 @@ def print_bpf_progs(verbose):
                             opstr = insn.op_str
                             opstr = handle_bpf_aux_reg_for_opstr(opstr, aux_regs[i])
                             indent.print(f"{indent.addr_hex(address)}\t{mnemonic}\t{opstr}")
+                if nth is not None:
+                    if pwndbg.aglib.memory.is_kernel(func):
+                        indent.print(
+                            message.info(pwndbg.dbg.selected_inferior().runcmd(f"break *{func}"))
```
jxuanli: is there a better way?

k4lizen: `pwndbg.dbg.selected_inferior.break_at(BreakpointLocation(addr))`

k4lizen: also `break *addr` doesn't work in lldb (because `b` != `break`)

k4lizen:
I think this is cool, but just to verify, it would be equivalent if the user did `kbpf` and copy pasted the func address and did `b *func_addr`?

jxuanli:
yes

pwndbg/commands/kbpf.py
jxuanli: the output of the func feels a bit strange within the context of the overall output but its fine ig

pwndbg/commands/kbpf.py
```diff
@@ -208,9 +235,9 @@ def print_bpf_maps(verbose):
                 key_size = int(bpf_array["map"]["key_size"])
                 value_size = int(bpf_array["map"]["value_size"])
                 max_entries = int(bpf_array["map"]["max_entries"])
-                bpf_array = int(bpf_array)
```
jxuanli, line 211: the linter suddenly doesn't understand that it is an int ... wtv

pwndbg/commands/kbpf.py
k4lizen: hm yeah that is unfortunate, you could bass `internal=True` and then it wouldn't show the output, but I don't think you want to do that in this case since it hides the bp from the user

k4lizen (approved):

k4lizen:
image
where "merge dev into branch" button Githubbbb?

jxuanli:
blocked by #3776

PR #3760, opened 2026-02-17
Fix curl error handling in setup-dev.sh for jemalloc download

### LINKED ISSUE #3185: Fix dev setup curl download that fail which result in incorrect checksum of binary file
We have a case in setup-dev.sh (maybe elsewhere too?) where we download a binary with curl and then check it against the expected checksum which may ocassionally fail with no error status code. This happens here: https://github.com/pwndbg/pwndbg/actions/runs/16581783014/job/46899275815#step:4:979, or in text:
```
Jemalloc not found in system. Downloading, configuring, building, and installing...
Downloading jemalloc...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:01 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:02 --:--:--     0
100 54894    0 54894    0     0  26603      0 --:--:--  0:00:02 --:--:-- 26595
Jemalloc binary checksum mismatch
Expected: 2db82d1e7119df3e71b7640219b6dfe84789bc0537983c3b7ac4f7189aecfeaa
Actual: 3590bcb90a75c32ba8b10d692d26838caedbc267a57db23931694abc9598c873
```

And here is the relevant code: https://github.com/pwndbg/pwndbg/blob/15524bcfaa5ed5c66f1e4d6414622fa5f2798019/setup-dev.sh#L205-L213

From what we looked at with @patryk4815 it doesn't seem like its a hacking attempt (where one would replace the binary returned by the github server), and instead, curl is just not returning an error when it failed to download the file?

Here is some topic that may be relevant: https://superuser.com/questions/590099/can-i-make-curl-fail-with-an-exitcode-different-than-0-if-the-http-status-code-i

hsnyus-09:

스크린샷 2026-02-17 145751

Fixes #3185

## Summary
`setup-dev.sh` downloads the jemalloc tarball with `curl` and then validates it via SHA256.
When the server returns an HTTP error page (4xx/5xx), `curl` may still exit successfully and write the error page to the output path. This results in a confusing “checksum mismatch” instead of a clear download failure.

This PR makes the jemalloc download step fail fast on HTTP errors and retries transient failures.

## Problem
- HTTP error pages can be saved as the jemalloc tarball
- The subsequent SHA256 check fails with “checksum mismatch”
- This can look like tampering, while the root cause is a failed download

## Changes
Updated both jemalloc download `curl` invocations in `setup-dev.sh` to use:
- `--fail` (`-f`): fail on HTTP errors (4xx/5xx)
- `--retry 3`: retry transient failures up to 3 times
- `--retry-connrefused`: include connection refused in retry logic

### DIFF
--- a/setup-dev.sh
+++ b/setup-dev.sh
@@ -195,7 +195,7 @@ install_jemalloc() {
             ACTUAL_SHA256=$(sha256sum "${JEMALLOC_TAR_PATH}" | cut -d' ' -f1)
             if [ "${ACTUAL_SHA256}" != "${JEMALLOC_TAR_SHA256}" ]; then
                 echo "Jemalloc tarball exists but has incorrect checksum. Re-downloading..."
-                curl --location --output "${JEMALLOC_TAR_PATH}" "${JEMALLOC_TAR_URL}"
+                curl --fail --location --retry 3 --retry-connrefused --output "${JEMALLOC_TAR_PATH}" "${JEMALLOC_TAR_URL}"
                 ACTUAL_SHA256=$(sha256sum "${JEMALLOC_TAR_PATH}" | cut -d' ' -f1)
                 if [ "${ACTUAL_SHA256}" != "${JEMALLOC_TAR_SHA256}" ]; then
                     echo "Jemalloc binary checksum mismatch after re-download."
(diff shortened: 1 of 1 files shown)


hsnyus-09:
@disconnect3d Hello, I solved the issue #3185, can you check it?

hsnyus-09:
The current test failure does not seem to be related to the content of the PR I generated

hsnyus-09:
@disconnect3d @patryk4815 copilot has completed the review. I think we have solved the issue, so may I ask you to review if there is anything more we need to revise?

disconnect3d:
Thanks!

PR #3763, opened 2026-02-17
Cleanup installation from source

k4lizen:
+ Explain to the users that they don't actually have to clone the repo
+ Partial revert of https://github.com/pwndbg/pwndbg/pull/3761
    + I realize it makes more sense to just leave decomp2dbg in the main deps, since all installations where we control the installation should actually just always install decomp2dbg, and distros dont care about pyproject.toml (probably?)
+ ~~Make `setup.sh` install pwndbg via uv instead of doing the gdb source dance~~

### DIFF
--- a/pwndbginit/common.py
+++ b/pwndbginit/common.py
@@ -30,9 +30,9 @@ def run_uv_install(
         tool_name = venv_path.name
         command: list[str] = [str(binary_path), "tool", "upgrade", tool_name]
     else:
-        # We don't want to quietly uninstall dependencies by just specifying
-        # `--extra gdb` so we will be conservative and pull all extras in.
-        command = [str(binary_path), "sync", "--all-extras"]
+        # --inexact makes it so any installed extras aren't uninstalled
+        # (it will also leave in dropped deps, but what can you do /shrug)
+        command = [str(binary_path), "sync", "--inexact"]
         if dev:
             command.append("--all-groups")
     logging.debug(f"Updating deps with command: {' '.join(command)}")
--- a/pwndbginit/common.py
+++ b/pwndbginit/common.py
@@ -30,9 +30,9 @@ def run_uv_install(
         tool_name = venv_path.name
         command: list[str] = [str(binary_path), "tool", "upgrade", tool_name]
     else:
-        # --inexact makes it so any installed extras aren't uninstalled
-        # (it will also leave in dropped deps, but what can you do /shrug)
-        command = [str(binary_path), "sync", "--inexact"]
+        # We don't want to quietly uninstall dependencies by just specifying
+        # `--extra gdb` so we will be conservative and pull all extras in.
+        command = [str(binary_path), "sync", "--all-extras"]
         if dev:
             command.append("--all-groups")
     logging.debug(f"Updating deps with command: {' '.join(command)}")
--- a/setup.sh
+++ b/setup.sh
@@ -75,12 +75,6 @@ install_pacman() {
         sudo pacman -Syu || true
     fi
     sudo pacman -S --noconfirm --needed git gdb python which debuginfod curl gcc make patch
-    if [ -z "$UPDATE_MODE" ]; then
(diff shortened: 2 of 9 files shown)


k4lizen:
On an ubuntu container:
```bash
# apt install curl git gdb -y
# curl -LsSf https://astral.sh/uv/install.sh | sh
# PY_VER=$(gdb -nx --batch -iex 'py import sysconfig; print(sysconfig.get_config_var("VERSION"))')
# uv tool install --python=$PY_VER  git+https://github.com/pwndbg/pwndbg
# pwndbg
Python Exception <class 'ModuleNotFoundError'>: No module named '_lsprof'
Error occurred in Python: No module named '_lsprof'
(gdb) 
```
I think we already saw this somewhere before?

pwndbginit/common.py
```diff
@@ -30,9 +30,9 @@ def run_uv_install(
         tool_name = venv_path.name
         command: list[str] = [str(binary_path), "tool", "upgrade", tool_name]
     else:
-        # We don't want to quietly uninstall dependencies by just specifying
-        # `--extra gdb` so we will be conservative and pull all extras in.
-        command = [str(binary_path), "sync", "--all-extras"]
+        # --inexact makes it so any installed extras aren't uninstalled
+        # (it will also leave in dropped deps, but what can you do /shrug)
+        command = [str(binary_path), "sync", "--inexact"]
```
patryk4815, line 35: hmm, nice, I never used this option

setup.sh
```diff
@@ -213,18 +202,19 @@ source ${PWNDBG_VENV_PATH}/bin/activate
...
-        # Load Pwndbg into GDB on every launch.
-        echo "source $PWD/gdbinit.py" >> ~/.gdbinit
-        echo "[*] Added 'source $PWD/gdbinit.py' to ~/.gdbinit so that Pwndbg will be loaded on every launch of GDB."
-    fi
-    echo "Please set the PWNDBG_NO_AUTOUPDATE environment variable to any value"
-    echo "to disable the automatic updating of dependencies when Pwndbg is loaded."
+uv sync
+
+# Install pwndbg and pwndbg-lldb
+uv tool install . --force --editable
```
patryk4815: setup.sh has historically only installed `.gdbinit` without the pwndbg/pwndbg-lldb binaries – and I would leave it as is; I wouldn't modify `setup.sh` before release.


a) If we want to change setup.sh, we should make it continue to install `.gdbinit`, i.e.:
```
uv sync
echo "$PWD/.venv/share/pwndbg/gdbinit.py" >> ~/.gdbinit
```
but without the pwndbg/pwndbg-lldb binaries

b) Generally, I would add the message "`echo 'This install method is deprecated; Learn about new installation methods on pwndbg.re'; exit 1`" to setup.sh at first line.

And merge setup.sh into setup-dev.sh.

patryk4815 (changes requested):
I'm ok with changing the documentation and uv.lock/py, but I wouldn't change setup.sh at this point

setup.sh
k4lizen: I think `b)` makes sense but I'm not 100% on board yet, so I'd like us to merge something like this PR and then move forward with `b)` later.

Regarding `a)`. Why do you not want to change it? People who already have pwndbg installed from source will just do `git pull`, so they won't be affected. People who have never installed pwndbg before will also not have any issues. I guess a potential issue could be people who have used pwndbg before but want to install it on a new computer? I haven't thought about that, but I'm also not convinced it's an actual issue since the `echo` messages and uv output should make it clear what is going on.

But, that being said, I'll revert this change in this PR before this release and we can discuss afterwards

patryk4815: we have in few places something like:
Do `git pull && ./setup.sh` to perform pwndbg upgrade

^ I think people are doing so

k4lizen:
okay, should be good now

I left the benign edits in setup.sh

patryk4815:
@k4lizen looks ok now

k4lizen:
@patryk4815 feel free to merge it then :p (i cant because you didn't submit an approving review)

PR #3739, opened 2026-02-11
Fix reading unmapped pages in idt

PatrickDaG:
As per the intel [instruction manual](https://www.intel.com/content/www/us/en/developer/articles/technical/intel-sdm.html) and [osdev](https://wiki.osdev.org/Interrupt_Descriptor_Table) the idt can contain a maximum of 256 entries.
For some reason sometimes the limit as declared in the idt register is higher than that. 
```
IDT=     fffffe0000000000 0000ffff
```
On an 64bit system such as mine this limit is equivalent to 512 entries.
Apparently the kernel still knows that that there cannot be any accesses to entries above 256 and does not map those pages. This in turn leads to this error:
```
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /nix/store/dfvz9wwaxw1swbmpryac2a6mfp4lp80q-pwndbg-env/lib/python3.13/site-packages/pwndbg/dbg_m │
│ od/gdb/__init__.py:687 in read_memory                                                            │
│                                                                                                  │
│    684 │   │   addr = address                                                                    │
│    685 │   │                                                                                     │
│    686 │   │   try:                                                                              │
│ ❱  687 │   │   │   result = gdb.selected_inferior().read_memory(addr, count)                     │
│    688 │   │   │   return bytearray(result)                                                      │
│    689 │   │   except gdb.error as e:                                                            │
│    690 │   │   │   if not partial:                                                               │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
MemoryError: Cannot access memory at address 0xfffffe0000008000

During handling of the above exception, another exception occurred:

╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /nix/store/dfvz9wwaxw1swbmpryac2a6mfp4lp80q-pwndbg-env/lib/python3.13/site-packages/pwndbg/comma │
│ nds/__init__.py:498 in __call__                                                                  │
│                                                                                                  │
│    495 │                                                                                         │
│    496 │   def __call__(self, *args: Any, **kwargs: Any) -> str | None:                          │
│    497 │   │   try:                                                                              │
│ ❱  498 │   │   │   return self.function(*args, **kwargs)                                         │
│    499 │   │   except TypeError:                                                                 │
│    500 │   │   │   print(f"{self.command_name}: {self.description}")                             │
│    501 │   │   │   pwndbg.exception.handle(self.function.__name__)                               │
│                                                                                                  │
[...]

PR #3768, opened 2026-02-18
Use generator for idt entries

PatrickDaG:
Switch `get_idt_entries` to return a iterator instead of a list.

This is a better fix for #3739.

I'm not sure if it is actually the best fix, I think Iterator usage here is a bit awkward and it also doesn't really fix the todo. While not used currently if we ever want to use more entries they still are read one by one. Lastly this still does not return a useful error in case idt pages are unmapped.

Another option that might be worth is to add a separate function just returning the first element, something like `get_first_idt_entry`. Or we keep the iterator and add new exception for when the address access fails to tell people what might've went wrong.


pwndbg/aglib/kernel/__init__.py
```diff
@@ -272,25 +273,21 @@ def krelease() -> tuple[int, ...] | None:
...
     size = pwndbg.aglib.arch.ptrsize * 2
-    num_entries = min((limit + 1) // size, 256)
-
-    entries = []
+    num_entries = (limit + 1) // size
 
     # TODO: read the entire IDT in one call?
     for i in range(num_entries):
         entry_addr = base + i * size
         entry = pwndbg.lib.kernel.structs.IDTEntry(pwndbg.aglib.memory.read(entry_addr, size))
```
jxuanli, line 287: we can do something like 

`pwndbg.aglib.memory.read(entry_addr, size, partial = True)` instead. This not only solves the TODO but also prevents reading unmapped pages.

jxuanli:
eventually I plan to move those IDT functions into the `class IDTEntry` because they are only relevant to x86. If you'd like, feel free to do so as well.

pwndbg/aglib/kernel/__init__.py
disconnect3d: @PatrickDaG wdyt?

disconnect3d:
I'll merge this for now as its good to read just one item from memory; since we dont use idt anywhere else (i believe) its ok to not use a partial read for now iiuc

PR #3776, opened 2026-02-19
adding a kernel symbol lookup function decorator

jxuanli:
how it should have been implemented from the start


pwndbg/aglib/kernel/symbol.py
```diff
@@ -287,31 +291,69 @@ def load_common_structs_on_load_linux() -> None:
             pwndbg.aglib.kernel.page_typeinfo_recovery_failure = e
 
 
+P = ParamSpec("P")
+
+
+def kernel_symbol_func(
+    t: str | None = None, use_symbol: bool = True, prefix: str = ""
+) -> Callable[[Callable[P, Any]], Callable[P, int | pwndbg.dbg_mod.Value | None]]:
```
k4lizen: can I get a chonky docstring cuz I'm a bit lost xd

pwndbg/aglib/kernel/symbol.py
```diff
@@ -325,89 +367,44 @@ def regex(self, s: str, pattern: str, nth: int) -> re.Match[Any] | None:
...
+
+    @kernel_symbol_func()
+    def map_idr(self) -> bool:
+        return False
+
+    @kernel_symbol_func()
+    def prog_idr(self) -> bool:
+        return False
+
+    # using symbols usually yield incorrect results
```
k4lizen, line 440: wait what when why

jxuanli: my hypothesis is that it happens when the VM uses multiple CPUs but I don't have definitive proof.

jxuanli: fwiw, I changed the semantics of the param so that symbol can be a fallback

pwndbg/aglib/kernel/symbol.py
```diff
@@ -287,31 +290,96 @@ def load_common_structs_on_load_linux() -> None:
...
+                return result
+            if prefer_symbol:
+                result = pwndbg.aglib.symbol.lookup_symbol(
+                    f.__name__ if symbol_name is None else symbol_name
+                )
+            if not isinstance(result, pwndbg.dbg_mod.Value):
+                # we use heuristics if the symbol could not be resolved by lookup_symbol
+                heuristic_func = f"{f.__name__}_heuristic_func"
+                if hasattr(self, heuristic_func):
+                    func_name = getattr(self, heuristic_func)
```
k4lizen: it took me a while to realize that the `func_name` here is the name of a kernel function that is used inside of our functions which use heuristics to resolve a kernel symbol

maybe we could rename  `func_name` to `kfunc_name` and `_heuristic_func` to `_heuristic_kfunc` to make this distinction more clear?

k4lizen: declaring the type
```suggestion
                    func_name: str = getattr(self, heuristic_func)
```
makes it a bit more clear what is going on also

pwndbg/aglib/kernel/symbol.py
```diff
@@ -287,31 +290,96 @@ def load_common_structs_on_load_linux() -> None:
...
+    A return value of another type indicates furthur lookup is needed.
+
+    Returns:
+        If the symbol could not be resolved, None is returned.
+        Otherwise, if the symbol can be resolved:
+            If a type string is specified. A pwndbg.dbg_mod.Value with the given type is returned.
+            Otherwise, an int is returned.
+    """
+
+    def decorator(f: Callable[P, Any]) -> Callable[P, int | pwndbg.dbg_mod.Value | None]:
```
k4lizen: Idk if we can easily get rid of the `Any`, but in actuality what are the expected return types here? some mix of int, pwndbg.dbg_mod.Value, None and NeedLookup ?

pwndbg/aglib/kernel/symbol.py
```diff
@@ -287,31 +290,96 @@ def load_common_structs_on_load_linux() -> None:
...
+                    if not pwndbg.aglib.symbol.lookup_symbol(func_name):
+                        return None
+                    _func: Callable[[], int | None] = getattr(self, f"_{f.__name__}")
+                    result = _func()
+            if not isinstance(result, pwndbg.dbg_mod.Value | int) and not prefer_symbol:
+                result = pwndbg.aglib.symbol.lookup_symbol(
+                    f.__name__ if symbol_name is None else symbol_name
+                )
+            if t and isinstance(result, pwndbg.dbg_mod.Value | int):
+                if t[-1] == "*":
```
k4lizen: a bit scary because stuff like function pointers don't have the `*` at the end, but rather in the middle of the type, but okay we can handle that if it ever comes up

pwndbg/aglib/kernel/symbol.py
```diff
@@ -287,31 +290,96 @@ def load_common_structs_on_load_linux() -> None:
...
+                    if not pwndbg.aglib.kernel.has_debug_info():
+                        result = pwndbg.aglib.memory.get_typed_pointer(t[:-1], result)
```
k4lizen: i don't understand why `not pwndbg.aglib.kernel.has_debug_info()` (maybe add a comment?)

pwndbg/aglib/kernel/symbol.py
```diff
@@ -287,31 +290,96 @@ def load_common_structs_on_load_linux() -> None:
...
+                else:
+                    result = pwndbg.aglib.memory.get_typed_pointer_value(t, result)
```
k4lizen: maybe you want to catch `TypeNotFoundError` here which these two functions throw? dunno but just pointing it out in case you missed it

pwndbg/aglib/kernel/symbol.py
```diff
@@ -325,89 +393,44 @@ def regex(self, s: str, pattern: str, nth: int) -> re.Match[Any] | None:
...
-        if modules:
-            return modules
-        if pwndbg.aglib.kernel.has_debug_symbols(self.modules_heuristic_func):
-            modules = self._modules()
-        return pwndbg.aglib.memory.get_typed_pointer("unsigned long", modules)
-
-    @pwndbg.lib.cache.cache_until("stop")
-    def db_list(self) -> pwndbg.dbg_mod.Value:
-        if pwndbg.aglib.kernel.krelease() >= (6, 10):
+    @kernel_symbol_func("unsigned long*")
```
k4lizen: why did we remove `@pwndbg.lib.cache.cache_until("stop")`?

pwndbg/aglib/kernel/symbol.py
```diff
@@ -287,31 +290,96 @@ def load_common_structs_on_load_linux() -> None:
...
+
+def kernel_symbol_func(
+    t: str | None = None, prefer_symbol: bool = True, symbol_name: str | None = None
+) -> Callable[[Callable[P, Any]], Callable[P, int | pwndbg.dbg_mod.Value | None]]:
+    """
+    Marks a kernel symbol lookup function.
+    @t: the type string. If the last character is `*`, a Value with a pointer type is returned.
+        The format follows the C syntax.
+    @prefer_symbol: if true, this decorator will try to resolve the actual symbol address with lookup_symbol first.
+    @symbol_name: the actual name of the symbol, if different from the function name.
```
k4lizen:
```suggestion

    Arguments:
        t: the type string. If the last character is `*`, a Value with a pointer type is returned.
        The format follows the C syntax.
         prefer_symbol: if true, this decorator will try to resolve the actual symbol address with lookup_symbol first.
         symbol_name: the actual name of the symbol, if different from the function name.
```
Formatting it like this will make it be picked up properly by the docs reference (e.g. https://pwndbg.re/dev/reference/pwndbg/color/memory/#pwndbg.color.memory.get), which doesn't matter in the grand scheme of things since noone really uses it, but its kinda nice to have the same style throughout the codebase (and we mostly do) (we "officially" use the google python docstrings style (https://google.github.io/styleguide/pyguide.html#383-functions-and-methods) (its also one of the two styles which get properly parsed by python tooling))

pwndbg/aglib/kernel/symbol.py
```diff
@@ -448,6 +471,19 @@ def qword_op_reg_memoff(
...
+    # returns the first 0x... as an int if exists
+    def dword_op_memoff_reg(
+        self, disass: str, op: str, sign: str = "-", nth: int = 0
+    ) -> int | None:
+        result = self.regex(disass, rf"{op}.*\[.*{re.escape(sign)}\s(0x[0-9a-f]{{1,8}})\]", nth)
+        if result is not None:
+            if sign == "-":
+                return (1 << 64) - int(result.group(1), 16)
+            return int(result.group(1), 16)
+        return None
```
k4lizen, line 495: theres lots of disass-related functions like this in the kernel code, they should likely be moved somewhere else at some point

pwndbg/aglib/kernel/symbol.py
```diff
@@ -287,31 +290,96 @@ def load_common_structs_on_load_linux() -> None:
...
+        If the symbol could not be resolved, None is returned.
+        Otherwise, if the symbol can be resolved:
+            If a type string is specified. A pwndbg.dbg_mod.Value with the given type is returned.
+            Otherwise, an int is returned.
+    """
+
+    def decorator(f: Callable[P, Any]) -> Callable[P, int | pwndbg.dbg_mod.Value | None]:
+        @functools.wraps(f)
+        def func(*args: P.args, **kwargs: P.kwargs) -> int | pwndbg.dbg_mod.Value | None:
+            self = args[0]
```
k4lizen, line 334: does this decorator only work on methods? this should be documented

k4lizen (commented):
quite a nice cleanup

i mostly have reasoning questions xd

pwndbg/aglib/kernel/symbol.py
jxuanli: Actually I figured only one of the symbols needs to have a type (so its more convenient later), moved the type casting to `kernel/__init__.py`

pwndbg/aglib/kernel/symbol.py
jxuanli: moved to `kernel/__init__.py`. I decided that most caching should happen at the python package level, and all functions/values that need to be accessed outside should be accessible from `__init__` (of course not true atm, but as refactoring goes on, this will eventually be true).

pwndbg/aglib/kernel/symbol.py
disconnect3d: +1

pwndbg/commands/buddydump.py
```diff
@@ -111,6 +110,10 @@ def cpu_limitcheck(cpu: str | None):
     help="The address to find in page free lists.",
 )
 
+parser.add_argument(
+    "--MIGRATE_TYPES", type=int, default=3, help="Use the specified MIGRATE_TYPES value"
```
disconnect3d, line 114: Shall we maybe extend the docstring to elaborate more on MIGRATE_TYPES - what is this, why one wants to set it, when?

disconnect3d (approved):

PR #3787, opened 2026-02-22
making gnureadline optional

### LINKED ISSUE #3786: archlinux missing gnureadline, make gnureadline optional
On Arch Linux there is an additional problem:

There is no `python-gnureadline` package in the repositories.
Because of that, users cannot install `gnureadline` through the system package manager.
At the moment `pwndbg-lldb` effectively hardcodes: `import gnureadline`

This makes pwndbg-lldb impossible to run on Arch Linux, even though the system Python itself is correctly linked against GNU Readline and works fine without gnureadline.

In other words, Arch has:
- working GNU Readline in python
- but no way to satisfy the forced gnureadline import

### Why `pwndbg-lldb` require `gnureadline`
pwndbg-lldb works best with **GNU Readline**, but some Python builds are linked against **libedit** instead (e.g. `uv` Python, custom builds, macOS).

When this happens, the REPL experience is degraded:
* `Ctrl+R` history search is broken or inconsistent
* keybindings don’t fully match GNU Readline behavior

### "Example" implementation
```python
try:
    import gnureadline as readline
except ImportError:
    import readline
    readline_backend = None
    if sys.version_info >= (3, 13):
        readline_backend = readline.backend
        
    if readline_backend != "readline": 
        print("INFO: gnureadline is not installed, for best REPL experience please install gnureadline")
```

omseervi098:
Closes: #3786 

### DIFF
--- a/pwndbg/dbg_mod/lldb/repl/readline.py
+++ b/pwndbg/dbg_mod/lldb/repl/readline.py
@@ -24,10 +24,11 @@ if sys.platform != "win32":
         if sys.version_info >= (3, 13):
             backend = readline.backend
 
-        if backend is not None and backend != "readline":
-            print(
-                "INFO: gnureadline is not installed, for best REPL experience please install gnureadline"
-            )
+        if backend != "readline":
+            info_msg = "INFO: gnureadline is not installed, for best REPL experience please install gnureadline"
+            if sys.stdout.isatty():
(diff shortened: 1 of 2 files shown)


patryk4815:
please change description in pyproject.toml about gnureadline and please make this `print` info gray when TTY colors are available. Also please don't handle `None` because it is unknown what version user use.

patryk4815:
image

PR #3788, opened 2026-02-22
GDB: show values in hex with print command

OBarronCS:
Add `set output-radix 16` to default GDB settings. Now, when printing values with `p`, the output is always hexadecimal. Previously, printing variables and registers would display them as decimal.

image


OBarronCS:
This is causing the `plist` tests to fail, as some elements of the struct are now being printed as hex instead of decimal:

image

This is caused by using GDB's `.value_to_human_readable` API, which must use the underlying config for formatting:

https://github.com/pwndbg/pwndbg/blob/d1c540cc028b5b5173df4c222e08e24f57e41c61/pwndbg/commands/plist.py#L440

Is this okay? Since these are struct elements, I tried to set `set print type hex off` to counteract the change, but it didn't change the formatting.

disconnect3d:
Idk, can we fix the tests somehow?:p

OBarronCS:
:( there is gonna be trouble with LLDB running the plist tests - getting uniform output between GDB and LLDB will be hard as their native formatters for types are different. 

image

An alternative path for this PR is to make a "helpful commands" section in the docs and put  this there as a suggestion for GDB users, instead of running it by default.

disconnect3d:
The plist tests could just check if gdb => this, else that?

disconnect3d:
or: take value as string and parse it using `int(s, 0)`:

```
In [2]: int('0x123', 0)
Out[2]: 291

In [3]: int('123', 0)
Out[3]: 123
```

OBarronCS:
I made the tests pass by fixing the regex to allow hex ints when necessary. This change currently only affects GDB, with LLDB behavior not changing.

PR #3789, opened 2026-02-22
Disassembly: indicate when current pc at breakpoint

### LINKED ISSUE #3783: make cursor red when on a breakpoint
Image

Image

I wish that it was more visually obvious that I am standing on a breakpoint in the first screenshot. This could potentially be solved by making the cursor (►) red or even having it be `b►` or `b>`.

OBarronCS:
Previously, if the current program counter was at an instruction without a breakpoint, there would be no indication of this in the display. Now, make the prefix red and add a "b" to show that this is the case.

image

Resolves #3783

### DIFF
--- a/pwndbg/aglib/nearpc.py
+++ b/pwndbg/aglib/nearpc.py
@@ -520,26 +520,33 @@ def nearpc(
     breakpoint_prefix = breakpoint_sign.ljust(len(prefix_sign) + 1)
     breakpoint_prefix = c.breakpoint(breakpoint_prefix)
 
+    current_insn_breakpoint_prefix = pwndbg.color.red(c.prefix(f"b{prefix_sign}"))
+
     # Print out each instruction
     for i, (address_str, symbol, instruction, asm) in enumerate(
         zip(addresses, symbols, instructions, assembly_strings)
     ):
-        # Show prefix only on the specified address and don't show it while in repeat-mode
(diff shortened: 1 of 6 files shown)


pwndbg/aglib/nearpc.py
```diff
@@ -520,26 +520,33 @@ def nearpc(
     breakpoint_prefix = breakpoint_sign.ljust(len(prefix_sign) + 1)
     breakpoint_prefix = c.breakpoint(breakpoint_prefix)
 
+    current_insn_breakpoint_prefix = pwndbg.color.red(c.prefix(f"b{prefix_sign}"))
```
OBarronCS: This is a hardcoded `b`, whereas the `b+` for normal breakpoints is the default value for the configuration variable `nearpc-breakpoint-prefix`.

OBarronCS: Nevermind - in line with the other prefixes, this one is now a config

tests/library/dbg/tests/test_nearpc.py
```diff
@@ -179,6 +179,7 @@ async def test_nearpc_opcode_seperator(ctrl: Controller, separator_bytes: int) -
 @pwndbg_test
 async def test_nearpc_highlight_breakpoint(ctrl: Controller) -> None:
     import pwndbg.aglib.symbol
+    import pwndbg.color
```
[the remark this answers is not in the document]
OBarronCS: done

PR #3793, opened 2026-02-22
downgrade buildkit-cache-dance action version to fix it and add error handling for delete stale cache step

jackmisbach:

makes https://github.com/pwndbg/pwndbg/pull/3722 actually work
I tested it on v3.1.2 and updated to v3.3.1 when making PR #3722, not realizing
https://github.com/reproducible-containers/buildkit-cache-dance/pull/45
would cause it not to work after 3.1.2
So for now, I shall revert it, so we can have the tests take 1-10m and not hours for every dep update.
the 400 byte buildkit caches will get overwritten with the real ~400mb caches.
I may make a PR, though, to fix the buildkit-cache-dance action itself, because it is kind of annoying.

And somehow the delete stale cache step failed one time, so I added it so it will just warn in GitHub Actions and continue, instead of failing, which I had it set to before, so that I would know if it was broken. It's not a big deal if it bugs out and doesn't remove them a time or two, as GitHub actions cache evicts based on LRU anyways, but it should still be removing them for best practice.

### DIFF
--- a/.github/workflows/base-image.yml
+++ b/.github/workflows/base-image.yml
@@ -108,14 +108,14 @@ jobs:
             --key "buildkit-${{ matrix.platform_base.tag_name }}-" \
             --json key -q '.[].key' | while read -r key; do
             echo "Deleting stale cache: $key"
-            gh cache delete "$key"
+            gh cache delete "$key" || echo "::warning::Failed to delete stale cache: $key"
           done
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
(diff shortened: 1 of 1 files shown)


jackmisbach:
actually whatever, I'll just update it to v3.3.1 and make it work for now until I make a PR there to fix what I want to with the action so it just uses the active builder by default without having to specify.

disconnect3d:
Should we merge it now?

jackmisbach:
yes it is good to merge 😁

PR #3798, opened 2026-02-26
Detect musl libc in statically linked binaries

disconnect3d:
The existing musl detection searches `.rodata` for the string `"/tmp/tmpnam_XXXX"` (from musl's `tmpnam.c`). This works for dynamically linked binaries where the full libc is loaded, but fails for statically linked ones — the linker only includes referenced code/data, so if the program never calls `tmpnam()`, that string is absent.

Add a fallback: search `.rodata` for musl's mallocng `size_classes` array, a 96-byte `uint16_t[48]` constant from `malloc.c`. This table is linked into any program that calls `malloc()` (nearly all of them), and the 96-byte pattern makes false positives essentially impossible. Both little-endian and big-endian representations are checked.

This covers musl v1.2.1+ (2020), when the mallocng allocator replaced oldmalloc. The same `size_classes` values are already used in `pwndbg/aglib/heap/mallocng.py` for heap inspection.

### DIFF
--- a/pwndbg/libc/musl.py
+++ b/pwndbg/libc/musl.py
@@ -14,6 +14,38 @@ from . import util
 from .dispatch import LibcType
 from .dispatch import LibcURLs
 
+# Precomputed byte patterns for musl's mallocng size_classes array (uint16_t[48]).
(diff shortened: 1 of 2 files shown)


k4lizen (approved):
lgtm

Issue #3745, opened 2026-02-13
allocators should be able to work with "unknown" libc with a user-set libc version


k4lizen:
### Description

Since https://github.com/pwndbg/pwndbg/pull/3637 :
When compiling `gcc main.c -static -o main -g` on a glibc system, you'll get:

<img width="480" height="244" alt="Image" src="https://github.com/user-attachments/assets/7dde9dde-cc35-476f-b110-3b8217e9854b" />

Before that was merged, support actually worked:

<img width="674" height="349" alt="Image" src="https://github.com/user-attachments/assets/b42ad4db-43f6-49e2-87ea-c2f6507e4ee8" />

### What to do

The problem is that in statically compiled binaries, stuff that isn't used isn't compiled in. So, stuff like version information (which is critically important) isn't available in the binary.

So we are going to need to ask for some information from the user, hopefully only the version (`set glibc 2.42`).

One way to take this could be to allow the user to also override the libc detection like `set libc glibc`, and then use the libc/glibc backend instead of libc/unknown. But I think this isn't a good idea because:
1. you can still use the "unknown" libc to correctly get stuff like `filepath()`, `addr()`, ...
2. You can't really trust what the glibc backend will tell you about version/debug info/symbols because you don't know what will be compiled in and what won't be.

Othewrise, if the libc is unknown we just let the allocator code figure out stuff like "do I have the symbols/structures I need". This is a bit annoying because it will have to do a bunch of type/symbol lookups which may or may not succeed, but I don't really see a better solution.

PR #3799, opened 2026-02-26
Allow heap commands to attempt resolution for unknown libc when glibc version is user-set

carlpinto25:
#3745
Full heap visualization may still fail in some static binary configurations due to issue with how the heuristic locates the mp_ struct


disconnect3d:
@carlpinto25 can u fix conflicts? we merged the glibc 2.43 support right now

pwndbg/commands/__init__.py
```diff
@@ -928,9 +929,24 @@ def _OnlyWithResolvedHeapSyms(*a: P.args, **kw: P.kwargs) -> T | None:
...
-        if pwndbg.libc.which() != pwndbg.libc.LibcType.GLIBC:
-            e(f"The currently active libc isn't glibc. It's {pwndbg.libc.which().value}.")
-            return None
+        libc_type = pwndbg.libc.which()
+        if libc_type != pwndbg.libc.LibcType.GLIBC:
+            if libc_type == pwndbg.libc.LibcType.UNKNOWN and pwndbg.libc.glibc.glibc_version:
+                # User gave us an explicit glibc version (e.g. `set glibc 2.35`).
+                # Proceed with heuristics and warn that results may be approximate.
+                w(
+                    f"The currently active libc is 'unknown' (e.g. statically-linked binary), "
```
k4lizen, line 938:
```suggestion
                    f"The currently active libc is 'unknown', "
```
this comment makes it sound like if the binary is  statically linked, we never know which libc it is using, but this is not the case.

pwndbg/libc/unknown.py
```diff
@@ -18,6 +18,13 @@ def _is_being_used() -> bool:
 
 
 def version(libc_filepath: str) -> tuple[int, ...]:
+    # If the user explicitly set a glibc version (e.g. via `set glibc 2.35`),
+    # honour it even when the libc is unknown (e.g. statically-linked binaries).
+    # This allows the heuristic heap allocator to work in those cases.
+    from pwndbg.libc.glibc import glibc_version as _user_glibc_version
+
+    if _user_glibc_version:
+        return tuple(int(i) for i in _user_glibc_version.value.split("."))
```
k4lizen, line 27: are we gonna have an import here for every libc implementation? this is not a good solution for this problem

pwndbg/aglib/heap/structs.py
```diff
@@ -26,8 +26,12 @@ def fastbin_index(size: int) -> int:
 
 # Operating under the assumption that the pwndbg/libc/ code can figure out
 # that we are using glibc with at least as good accuracy as the ptmalloc code.
```
k4lizen, line 28: you left this comment in, but you are invalidating this assumption

pwndbg/commands/__init__.py
```diff
@@ -928,9 +929,24 @@ def _OnlyWithResolvedHeapSyms(*a: P.args, **kw: P.kwargs) -> T | None:
 
         # Operating under the assumption that the pwndbg/libc/ code can figure out
         # that we are using glibc with at least as good accuracy as the ptmalloc code.
```
k4lizen, line 931: you left this comment in, but you are invalidating this assumption

k4lizen (changes requested):
when i wrote this code, it was definitely operating under a different assumption (i.e. that ptmalloc would never catch "unknown"). looking at the code itself, i think it doesn't use any glibc-specific stuff from pwndbg.libc. other than the tcache linking stuff which is fine, so something like this could work.

we cannot merge this due to the changes in pwndbg.libc.unknown.version though.

also i think i would prefer a

`set libc glibc`
`set libc-version 2.32`

UX, rather than
`set glibc 2.32`

so it is more agnostic to different libc's, but oke this is not a blocker here

disconnect3d:
Closing for now due to inactivity

Issue #3770, opened 2026-02-18
kernel backtrace unaligned


k4lizen:
<img width="734" height="78" alt="Image" src="https://github.com/user-attachments/assets/0f00626b-507c-4f4e-8889-d0ed4024e3d0" />

The names should be aligned

k4lizen:
Similarly, here:

<img width="412" height="201" alt="Image" src="https://github.com/user-attachments/assets/0b8a4de8-4071-4a7f-b230-791bb691bed8" />

The pointers should be aligned.

k4lizen:
https://github.com/pwndbg/pwndbg/pull/3785 does fix the second  issue but not the first:

<img width="416" height="147" alt="Image" src="https://github.com/user-attachments/assets/22b76a6a-daa3-4dac-8f5c-e9e30924ee36" />

PR #3800, opened 2026-02-26
Fix backtrace

carlpinto25:
closes- #3770


tests/library/dbg/tests/test_context_commands.py
```diff
@@ -780,3 +780,85 @@ async def test_cli_fixup_resolves_sp_pc_aliases(ctrl: Controller) -> None:
...
+    import pwndbg.aglib
+
+    # Use the binary designed to nest 12+ layers deep
+    DEEP_CALL_BINARY = get_binary("deep_call.native.out")
+    await ctrl.launch(DEEP_CALL_BINARY)
+
+    # We break on break_deep which provides 12+ frames of callstack.
+    # Use pwndbg's 'b' alias which works on both GDB and LLDB.
+    await ctrl.execute("b break_deep")
+    await ctrl.cont()
```
disconnect3d, line 800: Shouldn't we just do `ctrl.launch_to(..., "break_deep")` or it doesn't work here?

tests/library/dbg/tests/test_context_commands.py
```diff
@@ -780,3 +780,85 @@ async def test_cli_fixup_resolves_sp_pc_aliases(ctrl: Controller) -> None:
...
+
+    address_start_cols = []
+    for line in bt_lines:
+        addr_idx = line.find("0x")
+        if addr_idx != -1:
+            address_start_cols.append(addr_idx)
+
+    # Assert all addresses start at the same column (proving frame labels are identical length)
+    assert len(set(address_start_cols)) == 1, (
+        f"Addresses do not align vertically (frame label padding issue): {address_start_cols}"
```
disconnect3d, line 833: Honestly it would be nicer if the test showed the exact expected output and assert against that. Can we do that or does this have any problems?

Maybe we should also call the `context_backtrace` func directly instead of via a command, but its probably fine as well

disconnect3d (changes requested):
Can we test this in a more readable way (assert output)?

disconnect3d:
@carlpinto25 ping?

disconnect3d:
Closing due to inactivity

PR #3803, opened 2026-02-27
kernel/slab: fix slab list field name for pre-5.8 kernels

navytuner:
In kernels older than ~5.8, struct page uses the 'lru' field as the linked list head for node partial slabs. The 'slab_list' name was only introduced when the slab-specific fields in struct page were explicitly named (later backported into struct slab in 5.17).

NodeCache.partial_slabs hardcoded "slab_list" as the field passed to container_of(), causing offsetof() to return None on old kernels and a TypeError crash:
  unsupported operand type(s) for -: 'int' and 'NoneType'

Fix by probing at runtime whether the slab struct actually has a 'slab_list' field, falling back to 'lru' if not.

Fixes: slab list crashing on kernel 4.4.x and other pre-5.8 kernels

I attach crash/after-fix screenshot to below
Screenshot 2026-02-27 at 4 38 58 PM
Screenshot 2026-02-27 at 4 40 08 PM

+ [O] I read the contributing documentation.
+ [O] I am providing a screenshot of the (new feature) / (fixed bug).

### DIFF
--- a/pwndbg/aglib/kernel/slab.py
+++ b/pwndbg/aglib/kernel/slab.py
@@ -46,7 +46,7 @@ def slab_struct_type() -> str:
 
 
 def slab_list_field() -> str:
-    # In kernels older than ~5.8, struct page uses 'lru' instead of 'slab_list' for the node partial slab list.
+    # In kernels older than 4.18, struct page uses 'lru' instead of 'slab_list' for the node partial slab list.
     page_type = pwndbg.aglib.typeinfo.load(f"struct {slab_struct_type()}")
     if page_type is not None and page_type.offsetof("slab_list") is not None:
         return "slab_list"
--- a/pwndbg/aglib/kernel/slab.py
+++ b/pwndbg/aglib/kernel/slab.py
@@ -45,6 +45,14 @@ def slab_struct_type() -> str:
     return "page"
 
 
+def slab_list_field() -> str:
+    # In kernels older than ~5.8, struct page uses 'lru' instead of 'slab_list' for the node partial slab list.
+    page_type = pwndbg.aglib.typeinfo.load(f"struct {slab_struct_type()}")
+    if page_type is not None and page_type.offsetof("slab_list") is not None:
+        return "slab_list"
+    return "lru"
+
+
 OO_SHIFT = 16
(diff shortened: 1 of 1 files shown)


jxuanli:
thanks for the PR! But it seems to me though that `slab_list` is a valid field of `struct page` for all `5.x` releases
https://elixir.bootlin.com/linux/v5.0.21/source/include/linux/mm_types.h#L100 (a quick look also suggests it's true for `4.x` releases, though we don't really support them)

navytuner:
> thanks for the PR! But it seems to me though that `slab_list` is a valid field of `struct page` for all `5.x` releases
> 
> https://elixir.bootlin.com/linux/v5.0.21/source/include/linux/mm_types.h#L100 (a quick look also suggests it's true for `4.x` releases, though we don't really support them)

I see. Pwndbg doesn't have a plan for supporting slab command in 4.x version? It's because slab commands crash in 4.x versions. It was my mistake. It's for 4.x kernel, not 5.x.

jxuanli:
Commands are only guaranteed to work on kernel versions >= 5.0. But I'm fine with supporting older versions as well. We can merge this.

pwndbg/aglib/kernel/slab.py
```diff
@@ -45,6 +45,14 @@ def slab_struct_type() -> str:
     return "page"
 
 
+def slab_list_field() -> str:
+    # In kernels older than ~5.8, struct page uses 'lru' instead of 'slab_list' for the node partial slab list.
```
jxuanli: please update the comment and provide a link where `slab_list` isn't a valid field of `struct page`

jxuanli (approved):

pwndbg/aglib/kernel/slab.py
navytuner: > please update the comment and provide a link where `slab_list` isn't a valid field of `struct page`

https://elixir.bootlin.com/linux/v4.17/source/include/linux/mm_types.h#L78

This shows v4.17's `struct page` where `slab_list` doesn't exist yet. It was added in v4.18 by commit `4da1984edbbe`.

disconnect3d:
Awesome, thanks!

PR #3818, opened 2026-03-08
Display errno name for negative syscall return values

### LINKED ISSUE #2151: errno definition in return register output
It would be nice if the register output of the return ~~address~~ register would display errno definitions (if the value is negative):

current output:
```
*RAX  0xfffffffffffffff7
 RBX  0x7fffffffde88 —▸ 0x7fffffffe1da ◂— '/home/user/Downloads/demo'
*RCX  0x7ffff7fbc017 ◂— 0
 RDX  0
 RDI  0x7fffffffdc98 ◂— '/bin//sh'
 RSI  0
 R8   0x7ffff7e03b20 (main_arena+96) —▸ 0x55555555a3b0 ◂— 0
 R9   0x410
 R10  0x7ffff7c0abe8 ◂— 0x11002200006cb5
*R11  0x346
 R12  1
 R13  0
 R14  0x555555557dc8 (__do_global_dtors_aux_fini_array_entry) —▸ 0x555555555170 (__do_global_dtors_aux) ◂— endbr64 
 R15  0x7ffff7ffd000 (_rtld_global) —▸ 0x7ffff7ffe2e0 —▸ 0x555555554000 ◂— 0x10102464c457f
 RBP  0x7fffffffdd60 —▸ 0x7fffffffde00 —▸ 0x7fffffffde60 ◂— 0
 RSP  0x7fffffffdc98 ◂— '/bin//sh'
*RIP  0x7ffff7fbc017 ◂— 0
```
what i expect:
```
*RAX  0xfffffffffffffff7 <EBADF: fd is not a valid file descriptor> 
 RBX  0x7fffffffde88 —▸ 0x7fffffffe1da ◂— '/home/user/Downloads/demo'
*RCX  0x7ffff7fbc017 ◂— 0
 RDX  0
 RDI  0x7fffffffdc98 ◂— '/bin//sh'
 RSI  0
 R8   0x7ffff7e03b20 (main_arena+96) —▸ 0x55555555a3b0 ◂— 0
 R9   0x410
 R10  0x7ffff7c0abe8 ◂— 0x11002200006cb5
*R11  0x346
 R12  1
 R13  0
 R14  0x555555557dc8 (__do_global_dtors_aux_fini_array_entry) —▸ 0x555555555170 (__do_global_dtors_aux) ◂— endbr64 
 R15  0x7ffff7ffd000 (_rtld_global) —▸ 0x7ffff7ffe2e0 —▸ 0x555555554000 ◂— 0x10102464c457f
 RBP  0x7fffffffdd60 —▸ 0x7fffffffde00 —▸ 0x7fffffffde60 ◂— 0
 RSP  0x7fffffffdc98 ◂— '/bin//sh'
*RIP  0x7ffff7fbc017 ◂— 0
```

I think \<EBADF\> should be enough, but having an additional short description would be nice.

Br1an67:
Fixes #2151

## Summary
This PR adds errno name display for the return value register when it contains a negative value indicating a syscall error. Linux syscalls return negative errno values (-1 to -4095) on error, and this change makes these values more informative during debugging.

## Changes
- Added errno detection logic to the `RegisterContext.register_context_default()` method in `pwndbg/commands/context.py`
- When the return value register (e.g., RAX on x86-64) contains a value in the range -1 to -4095, the errno name is displayed alongside the value

## Example
Before:
```
*RAX  0xfffffffffffffff7
```

After:
```
*RAX  0xfffffffffffffff7 <EBADF>
```

## Testing
The implementation was tested by verifying the errno detection logic with various negative values:
- `0xfffffffffffffff7` (-9) → `EBADF`
- `0xffffffffffffffea` (-22) → `EINVAL`
- `0xffffffffffffffff` (-1) → `EPERM`

```
pwndbg/commands/context.py | 45 +++++++++++++++++++++++++++++++++++++++++++--
 1 file changed, 45 insertions(+), 1 deletion(-)
```


pwndbg/commands/context.py
```diff
@@ -1260,11 +1261,54 @@ def register_context_default(self, reg: str) -> str | None:
         val = self.get_register_value(reg)
         if val is None:
             return None
-        desc = ""
         desc = pwndbg.chain.format(val)
         prefix = self.get_prefix(reg)
+
+        # Check if this is the return value register and value might be a negative errno
+        retval = pwndbg.aglib.regs.current.retval
+        if retval is not None and reg.lower() == retval.lower():
```
disconnect3d, line 1269: So for each register, even if the register is not a retval register - we will check if its a retval register? Sounds like an easy solution, but not a great one.

pwndbg/commands/context.py
```diff
@@ -1260,11 +1261,54 @@ def register_context_default(self, reg: str) -> str | None:
...
+            # are -4095 to -1
+            if 0xFFFFFFFFFFFFF001 <= val <= 0xFFFFFFFFFFFFFFFF:
+                errno_val = val - (1 << 64)
+            else:
+                return None
+        elif ptrsize == 4:
+            # For 32-bit: values from 0xfffff001 to 0xffffffff
+            # are -4095 to -1
+            if 0xFFFFF001 <= val <= 0xFFFFFFFF:
+                errno_val = val - (1 << 32)
```
disconnect3d, line 1299: Those should just do sth like `ctypes.c_int64(0xFFFFFFFFFFFFFFFF).value` or read the register value as int through the debugger.

I also don't really like the if/elif for ptrsize.

pwndbg/commands/context.py
disconnect3d: Can you think/see if we can do any better here? I mean, execute such logic only for the retaddr register but not any other GPR registers. What would it take to do so?

disconnect3d:
@Br1an67 ping, please see comments

themavik (commented):
Mapping high unsigned values to Linux errno names is readable. nit: a valid pointer in the top 4k of the canonical 64-bit range could still get a bogus errno tag.

pwndbg/commands/context.py
disconnect3d: I think its ok to leave it as it is for now on, but we could still update the other part with converting values.

pwndbg/commands/context.py
disconnect3d: Alternative: since we need a negative value anyway, why don't we compute one right away?

disconnect3d:
Closing due to inactivity.

PR #3823, opened 2026-03-16
Fix shortcomings of `nearpc --function` compared to `disass`

### LINKED ISSUE #3822: Shortcomings of `nearpc --function` compared to `disass`
The recently introduced command `nearpc --function` (#3736) to disassemble functions is great but has a few shortcomings.

First, it always highlights the first instruction, not the instruction `$pc` is pointing to:

Image

This is semi-intentional since `nearpc` can also receive an explicit address as first argument, which then gets highlighted. It is further complicated by the fact that the output of `nearpc --function $pc` is identical to that of `nearpc --function main`, even though it should probably highlight the address `0x555555555156` (like `nearpc $pc` and `nearpc 0x555555555156` do) for consistency. If the given address does not fall onto an instruction boundary but instead lies *inside* the <var>i</var>-th instruction (say), then one could argue that the <var>i</var>-th instruction should be highlighted. This complexity can be avoided by mimicking GDB's `disass` command, which always highlights the instruction that `$pc` is pointing at with an arrow `=>`.

Second, `nearpc --function` requires a function name or address to be specified, whereas `disass` will default to dumping the current function.

Third, it requires debug symbols to find the function boundaries. Without debug symbols, it fails with `Error: function boundaries of '93824992235862' could not be found` in gdb and with `Pwndbg commands require a target binary to be selected` in lldb.

Ordoviz:
Fixes #3822.

This PR eliminates two differences between `nearpc --function` and GBD's builtin `disass` command.
First, it adds the nearpc-prefix (►) to the instruction at the pc of the selected frame instead of the first instruction at the function start.
Second, it allows `nearpc --function` to be run without a function argument. In that case, the function surrounding the pc of the selected frame is disassemble.

This is still missing unit tests.


pwndbg/aglib/nearpc.py
```diff
@@ -566,94 +570,18 @@ def nearpc(
...
         # In case there are duplicate instances of an instruction (tight loop),
         # ones that the instruction pointer is not at stick out a little, to indicate the repetition
-        elif not pwndbg.config.highlight_pc or instruction.address != pc or repeat:
+        elif (
+            not pwndbg.config.highlight_pc or instruction.address != address_to_highlight or repeat
+        ):
             address_str = c.address(address_str)
             symbol = c.symbol(symbol)
-        elif pwndbg.config.highlight_pc and i == index_of_pc:
+        elif pwndbg.config.highlight_pc and instruction.address == address_to_highlight:
```
OBarronCS: One tricky part here is that sometimes the disassembly display shows the same instruction multiple times, like when you are stepping through a tight loop (and emulation is enabled)

Lets say the loop is 3 instruction loop (A,B,C), but you are in the second iteration of it. The display shows the following:

A
B
C <- jmp to A instruction
A
B
C <- pc is actually here
[future instructions here]

We only want the "current" instance of instruction C to have the prefix. That's why we held track of the index in the list that that "current" pc is at.

Ordoviz: Thanks for explaining. I've decided to fix this by restoring the `pc_index` and simply requiring that the internal `nearpc` function cannot use a custom `address_to_highlight` when loops are displayed non-linearly.

Ordoviz:
I would like to get some feedback on this approach before I add unit tests and make the git history nicer.

pwndbg/commands/nearpc.py
```diff
@@ -110,9 +121,14 @@ def nearpc(
...
 
     end_address = None
+    address_to_highlight = None
     if function is not None:
         # Emulate GDB behavior of "disass" - it disassembles the entire function in which
         # the input address resides. User can input integer or string name of function, or an expression
+        address_to_highlight = int(pwndbg.dbg.selected_frame().pc())
+        if function is CURRENT_FUNCTION:
+            function = address_to_highlight
+
```
OBarronCS, line 131: Nice! This is a great quality of life improvement!

OBarronCS:
Awesome! Thanks for this, this is a really great to have. The codes looks good to me. FYI for future reference, on PRs don't worry about managing the git history - on merge it will be squashed. Force pushing and editing the history makes it harder to review because GitHub has a harder time tracking of comments corresponding to certain lines of code.

OBarronCS:
I'm testing this locally, and seeing a bug that it looks like we don't have a test for:

image

In the tight loop, the instructions bytes `e2 fb` are being highlighted for all occurrences (these bytes are enabled with `set nearpc-num-opcode-bytes 10`)

The code in the screenshot is the following:
```asm
.global _start
.section .text
_start:
.intel_syntax noprefix

nop
nop

xor eax, eax
mov ecx, 4

.loop:
    add eax, 1    
    loop .loop

    nop
    nop

    mov eax, 1     
    xor ebx, ebx
    int 0x80
```
You can manually compile it with `zig cc -target x86_64-freestanding input.S -o out`

The tests currently hardcode to disabling colors, so making a test for this might be difficult: https://github.com/pwndbg/pwndbg/blob/242a5c0ef8a4acbb24746d61169a51d63e2afdf8/tests/host/gdb/__init__.py#L71. Allowing enabling colors in tests would be a larger change, but it would nice to have this fixed and manually verify it for now.

Ordoviz:
> In the tight loop, the instructions bytes `e2 fb` are being highlighted for all occurrences (these bytes are enabled with `set nearpc-num-opcode-bytes 10`)

Good catch! This is fixed now.

OBarronCS (approved):
Awesome! This looks great

PR #3827, opened 2026-03-21
fix unsupported architecture: powerpc64le

zeguilherme1:
# fix lldb - "Unsupported architecture: powerpc64le" error

Fixes the Unsupported architecture: powerpc64le error when using pwndbg with LLDB.

### Reproduction
Compile a powerpc64le binary  
`powerpc64le-linux-gnu-gcc -g test.c -o test_ppc64le`

Then run it using pwndbg-lldb. It results in

`pwndbg.dbg_mod.Error: Unsupported architecture: powerpc64le. It may be that Pwndbg is not correctly categorizing the architecture. Please file a bug report `

error

### Cause
We can see that at at lib/arch.py we have a Python Dict called PWNLIB_ARCH_MAPPING, which originally have only the mapping `"powerpc": "powerpc"`for the powerpc, but LLDB reports the architecture as `powerpc64le`, because of this pwndbg fails to recognize the architecture.

### Fixing
At  aglib/arch_mod.py we have this Update() function, which takes the pwndbg selected arch from lldb, if we try to see the name that we are receiving we can see that we get `powerpc64le`

So i included this section in LLDBProcess at dbg_mod/lldb/__init__py
 ```
    elif arch_name in ("powerpc", "powerpc64le", "ppc64le"):
        arch_name = "powerpc" 
```
LLDBProcess converts from lldb to pwndbg.

image

LLDB recognizes and sets ptrsize to 8 bytes if we try to open a powerpc64le binary or 4 bytes if we try to use powerpc 32 bits.

And it worked!

noerror

### CODE REVIEW

**Comment by mbrla0** [comment]
Thanks for the contribution!

While this is a very reasonable idea, we shouldn't deal with LLDB specifics outside of `pwndbg/dbg_mod/lldb`. Maybe you could add something that achieves this same goal to `LLDBArch.arch()`, instead? Code there already does these LLDB to Pwndbg arch name conversions. If you're unsure where to look, it's [here](https://github.com/pwndbg/pwndbg/blob/c1a626def5a40466dc7ac372f000a9a5dbf7af48/pwndbg/dbg_mod/lldb/__init__.py#L1783) as of the time I'm writing this.

Also, are you sure `powerpc` is the correct architecture name to map `powerpc64le` to? My first guess is that `powerpc` would be the original 32-bit variant of the architecture...

**Comment by zeguilherme1** [comment]
@mbrla0 I've made some changes removing cast import from typing lib. Also, as described in the MD, pwdbg are dealing okay recognizing powerpc 32 and 64 bits.

image

image

I'll aprecciate if you can review it ;)

**Review by themavik — COMMENTED** [review]
nit: mapping powerpc64le/ppc64le to powerpc assumes the rest of pwndbg treats them the same; if anything branches on 64-bit PowerPC specifics, this alias might be too coarse.

**Comment by disconnect3d** [comment]
> nit: mapping powerpc64le/ppc64le to powerpc assumes the rest of pwndbg treats them the same; if anything branches on 64-bit PowerPC specifics, this alias might be too coarse.

Sure: do you know such a case when we do this kind of branching? If not, I'm fine with it.

**Comment by disconnect3d** [comment]
@mbrla0 can u review once again?

**Comment by mbrla0** [comment]
> nit: mapping powerpc64le/ppc64le to powerpc assumes the rest of pwndbg treats them the same; if anything branches on 64-bit PowerPC specifics, this alias might be too coarse.

@themavik On the contrary, weirdly enough.

`GDBProcess.arch()` [does the exact same thing](https://github.com/pwndbg/pwndbg/blob/8380cc58f947b3c33bb222007039df4ffbb3ce1e/pwndbg/dbg_mod/gdb/__init__.py#L1018) for PowerPC, both 64 and 32-bit variants use the same architecture name, but different pointer size and - potentially also - endianness values. As counter-intuitive as it seems at first, just renaming the architecture to `"powerpc"` is perfectly consistent with how the rest of Pwndbg works.

Additionally, the rest of the code in `LLDBProcess.arch()` takes care of making sure the pointer size and endianness are correct, regardless of the name we're picking, so that's also not a concern.

File: pwndbg/aglib/arch_mod.py [hunk 1]
```diff
@@ -423,6 +423,7 @@ def update() -> None:
     # Will be None the first time around.
     if pwndbg.aglib.arch is None or a.name != pwndbg.aglib.arch.name:
         pwndbg_arch = get_pwndbg_architecture(a.name)
+
```
mbrla0, line 426: Is there a point to this line being here?

pwndbg/dbg_mod/lldb/__init__.py
```diff
@@ -1854,6 +1854,8 @@ def _has_xpsr(thread) -> bool:
         elif arch_name == "riscv64":
             # Pwndbg use a different name for riscv64.
             arch_name = "rv64"
+        elif arch_name in ("powerpc", "powerpc64le", "ppc64le"):
+            arch_name = "powerpc"
```
mbrla0, line 1858: It would be nice to explain _why_ it is okay to do this in a comment here. Feel free to use my explanation.

mbrla0 (approved):
@disconnect3d @zeguilherme1 

I've only got minor nitpicks about style, and renaming to `"powerpc"` is consistent with what `GDBProcess` does. I think this is good to go!

OBarronCS:
> Sure: do you know such a case when we do this kind of branching? If not, I'm fine with it.

For clarity on the arch name: the arch name that the debugger exposes must to be mapped to the name that pwndbg uses, which are defined here: 

https://github.com/pwndbg/pwndbg/blob/d04a3a80c03e59d0090dd43a1c072bf0340f832d/pwndbg/lib/arch.py#L30-L44

The choice to consolidate all the powerpc variants (32-bit, 64-bit) under the single name `powerpc` (as opposed to having a separate "powerpc" and "powerpc64"), is ultimately arbitrary and the codebase takes this into account (and we basically do this because they share almost everything except pointer width, from the perspective of pwndbg),

PR #3830, opened 2026-03-23
slab: show inuse/total object count in slab_contains output

navytuner:
I made "slab contains [addr]" command display the number of inuse/total objects as well.
It would help kernel debugging because the number of inuse/total objects are displayed only in slab info command, which has very long output.

image

### DIFF
--- a/pwndbg/commands/slab.py
+++ b/pwndbg/commands/slab.py
@@ -321,7 +321,9 @@ def slab_contains(address: str) -> None:
         desc = "[inactive]"
         inuse = f"[something went wrong: {hex(addr)}]"
         slab = slab_cache.find_containing_slab(addr)
+        objcnt = ""
         if slab:
+            objcnt = f"[{slab.inuse}/{slab.object_count}]"
             if addr in slab.free_objects:
                 inuse = "free"
             elif addr in slab.objects:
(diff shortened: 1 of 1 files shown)


disconnect3d:
I think this is small enough we can just merge it. @jxuanli let me know in case this is not ok, or open a PR with a change if something needs to be changed :P

jxuanli:
I don't think this belongs here. `slab contains` is meant to display info of a particular object. The `[12/21]` could be mistaken as the index of the slot instead of inuse/total (the 12th slot out of the 21). It is also an info about the slab not the object. The actual total vs inuse is better displayed with the `slab list` command, it aggregates all (partial and active) slabs for a given cache. 
image

navytuner:
I agree with that `[12/21]` notation could be mistaken as the index of the object in the slab. However, I'm sure this would help kernel debugging when someone tries cross-cache, which requires tracking the number of in-use objects in the slab that contains vulnerable object. 

`slab list`, as you provided, doesn't display the number of in-use objects in **one** slab, it just shows the number of in-use objects in **all** the slabs of the cache(e.g. kmalloc-192). So `slab info [cache]` is the only way to display the number of objects in each slab, which has very long output if there are multiple cores, and many partial slabs.

Moreover, `slab contains [addr]` seems displaying the slab info containing the given address(object), so it would be okay if a few more slab info is displayed on `slab contains`.

navytuner:
For avoiding misperception, how about taking one more line and display like below?
`In-Use: 12/21`

This notation is same as `slab info [cache]` command.

disconnect3d:
> For avoiding misperception, how about taking one more line and display like below? `In-Use: 12/21`
> 
> This notation is same as `slab info [cache]` command.

Let's do that

PR #3833, opened 2026-03-24
slab: display object count on its own line in slab_contains output

navytuner:
- Display the inuse/total object count on a separate "In-Use:" line instead of appending it to the status line
- Remove square brackets around the count for cleaner output

Follow-up to #3830

image

### DIFF
--- a/pwndbg/commands/slab.py
+++ b/pwndbg/commands/slab.py
@@ -323,7 +323,7 @@ def slab_contains(address: str) -> None:
         slab = slab_cache.find_containing_slab(addr)
         objcnt = ""
         if slab:
-            objcnt = f"[{slab.inuse}/{slab.object_count}]"
+            objcnt = f"{slab.inuse}/{slab.object_count}"
             if addr in slab.free_objects:
                 inuse = "free"
             elif addr in slab.objects:
@@ -339,6 +339,7 @@ def slab_contains(address: str) -> None:
         else:
             inuse = "in-use"
         indent.print("slab:", message.hint(f"{hex(base)}"), desc)
-        indent.print("status:", message.hint(inuse), objcnt)
+        indent.print("status:", message.hint(inuse))
+        indent.print("In-Use:", message.hint(objcnt))
     except Exception as e:
         print(message.warn(f"address does not belong to a SLUB cache: {e}"))
--- a/pwndbg/commands/slab.py
+++ b/pwndbg/commands/slab.py
@@ -321,9 +321,8 @@ def slab_contains(address: str) -> None:
         desc = "[inactive]"
         inuse = f"[something went wrong: {hex(addr)}]"
         slab = slab_cache.find_containing_slab(addr)
-        objcnt = ""
         if slab:
-            objcnt = f"{slab.inuse}/{slab.object_count}"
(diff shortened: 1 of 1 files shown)


jxuanli:
maybe append it in the slab desc (second line), something like `[partial, node 0, 17/21 in-use]` so it is clear as to it is describing the containing slab. If you want, you can also indicate the slot index of the obj.

navytuner:
It would be better! I'll insert the info into the [] together as you suggested.

pwndbg/commands/slab.py
jxuanli:
```suggestion
                    desc = f"active, cpu {slab.cpu_cache.cpu}"
```

pwndbg/commands/slab.py
```diff
@@ -336,9 +335,10 @@ def slab_contains(address: str) -> None:
                     desc = f"[partial, cpu {slab.cpu_cache.cpu}]"
```
jxuanli:
```suggestion
                    desc = f"partial, cpu {slab.cpu_cache.cpu}"
```

pwndbg/commands/slab.py
```diff
@@ -336,9 +335,10 @@ def slab_contains(address: str) -> None:
...
                 else:
                     desc = f"[partial, node {slab.node_cache.node}]"
```
jxuanli:
```suggestion
                    desc = f"partial, node {slab.node_cache.node}"
```

pwndbg/commands/slab.py
```diff
@@ -336,9 +335,10 @@ def slab_contains(address: str) -> None:
...
+            desc = desc[:-1] + ", " + objcnt + "]"
```
jxuanli:
```suggestion
            desc = "[" + desc + ", " + objcnt + "]"
```

jxuanli: or maybe use an fstring

jxuanli: adding `]` just to have it later ignored seems a bit strange

navytuner:
Used a separate location variable to avoid overloading desc for both slab placement and the final description string.

Also, since an active slab is always associated with a CPU cache (is_active=True implies is_cpu=True), I removed the redundant is_cpu check under the is_active branch. 

The fallback label for slabs not found in partial/cpu lists is now [full] instead of the misleading [inactive].

disconnect3d:
@jxuanli fine to merge?

jxuanli:
> The fallback label for slabs not found in partial/cpu lists is now [full] instead of the misleading [inactive].

full slabs may be active as well and I think its an important to distinguish a slab  currently linked in kmem cache vs those that are not (maybe `[full and inactive]` would be better describing it but idk).

navytuner:
> > The fallback label for slabs not found in partial/cpu lists is now [full] instead of the misleading [inactive].
> 
> full slabs may be active as well and I think its an important to distinguish a slab currently linked in kmem cache vs those that are not (maybe `[full and inactive]` would be better describing it but idk).

I understood. How about distinguishing them as `[active, cpu 1, full]` and `[inactive, full]`?

jxuanli (approved):

PR #3842, opened 2026-03-29
v2p and p2v gdb func helpers

k4lizen:

image


disconnect3d:
>     A kernel .text pointer has multiple virtual address mappings, the one in physmap is returned.

Btw is there a way we could obtain all addresses? Maybe a command for this if we don't have one? Or 2nd argument to p2v?

k4lizen:
ideally the `p2v` command could print all mappings it finds

PR #3845, opened 2026-03-29
refactor: move wrappers/readelf.py to lib/got.py

### LINKED ISSUE #3844: Move wrappers/readelf.py to got.py
This is just a cleanup task from this PR: https://github.com/pwndbg/pwndbg/pull/3470#issuecomment-4150198381

We need to remove the wrappers/readelf.py file and move its function to got.py: though, the question that pops out is whether commands/got.py is a good place for it - if something is not requiring GDB/LLDB or any debugger API, we should move it to lib/got.py (and create it).

But if it requires that, commands/got.py is probably ok..

tysoncung:
## Problem

As discussed in #3470 (comment), the GOT entry parsing logic in `pwndbg/wrappers/readelf.py` does not depend on any debugger API (GDB/LLDB). It should be moved to `pwndbg/lib/` per the project convention for debugger-independent code.

## Solution

- Moved `pwndbg/wrappers/readelf.py` → `pwndbg/lib/got.py`
- Updated all imports in `pwndbg/commands/got.py` to use the new module path
- Removed the unused `cmd_name` variable (no longer a wrapper)
- Updated a stale comment that referenced "readelf output"

## Files Changed

- `pwndbg/wrappers/readelf.py` → `pwndbg/lib/got.py` (renamed)
- `pwndbg/commands/got.py` (import updates)

Closes #3844

### DIFF
--- a/pwndbg/commands/got.py
+++ b/pwndbg/commands/got.py
@@ -14,10 +14,10 @@ import pwndbg.color.memory as mem_color
 import pwndbg.commands
 import pwndbg.lib.memory
 import pwndbg.wrappers.checksec
-import pwndbg.wrappers.readelf
+import pwndbg.lib.got
 from pwndbg.color import message
 from pwndbg.commands import CommandCategory
-from pwndbg.wrappers.readelf import RelocationType
+from pwndbg.lib.got import RelocationType
 
 parser = argparse.ArgumentParser(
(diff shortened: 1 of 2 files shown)


disconnect3d:
Closing in favor of https://github.com/pwndbg/pwndbg/pull/3850

PR #3847, opened 2026-03-29
fix(color): validate color parameter names

### LINKED ISSUE #2874: add color parameter validation

tysoncung:
## Problem

Setting a color parameter to an invalid value (e.g. `set telescope-register-color meow`) causes an unhandled `KeyError` exception because the color name is looked up directly in `globals()` without validation (#2874). Additionally, using `globals()` allows access to non-color entries like `ansi_escape_8bit`, causing `TypeError`.

## Solution

Added a `VALID_COLOR_NAMES` frozenset containing all valid color function names and validate against it in `generateColorFunction()` before the globals lookup. Invalid names now print a helpful warning listing valid options instead of crashing.

### Before
```
KeyError: 'meow'
```

### After
```
Warning: invalid color 'meow'. Valid colors: black, blue, bold, cyan, ...
```

## Files Changed

- `pwndbg/color/__init__.py` — added `VALID_COLOR_NAMES` set and validation in `generateColorFunction()`

Closes #2874

### DIFF
--- a/pwndbg/color/__init__.py
+++ b/pwndbg/color/__init__.py
@@ -196,6 +196,35 @@ class ColorConfig:
         raise AttributeError(f"ColorConfig object for {self._namespace} has no attribute '{attr}'")
 
 
+VALID_COLOR_NAMES: frozenset[str] = frozenset(
+    {
+        "none",
+        "normal",
+        "black",
+        "red",
+        "green",
+        "yellow",
+        "blue",
+        "purple",
+        "cyan",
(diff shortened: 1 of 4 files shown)


disconnect3d:
Pls fix lint issue and regenerate docs
image

disconnect3d:
@tysoncung please open another PR if u fix the conflicts etc.

Issue #3840, opened 2026-03-28
gdb pwndbg loads file only after reading .gdbinit


k4lizen:
### Description

Oftentimes I write a `.gdbinit` that has various commands / temp breakpoints / whatever, that require the debugger to know about the binary. With the `pwndbg` command, this doesn't work.

### Steps to reproduce

My `.gdbinit` connecting to a kernel (the bug is not related to kernel debugging (I think)):
```
tar rem :1234

tb mm/slub.c:1000
```

Using the "source pwndbg from gdb method":
```
$ gdb vmlinux
Reading symbols from vmlinux...
0xffffffff82bb5d07 in pv_native_safe_halt () at arch/x86/kernel/paravirt.c:63
63	}
Temporary breakpoint 1 at 0xffffffff816b6bc4: mm/slub.c:1000. (4 locations)
```
Using the `pwndbg` command:
```
$ pwndbg vmlinux
~/.gdbinit: Skipped loading Pwndbg from `source path/gdbinit.py` - already loaded.

⚠️ warning: No executable has been specified and target does not support
determining executable automatically.  Try using the "file" command.
0xffffffff82bb5d07 in ?? ()
❌️ No symbol table is loaded.  Use the "file" command.
pwndbg: loaded 212 pwndbg commands. Type pwndbg [filter] for a list.
pwndbg: created 9 GDB functions (can be used with print/break). Type help function to see them.
Reading symbols from vmlinux...
```

You can sort of get around this by doing `file vmlinux` at the top of the `.gdbinit` but:
1. It is a footgun that will be easily encountered
2. I get the `Reading symbols from vmlinux...` message now after the breakpoint is set, which means we are reading the binary twice, which means we are slowing down startup time
3. I've seen someone have an issue with this setup, I don't remember what it was, but it was resolved after they switched to `gdb` (pwndbg being sourced)

### My setup
```
Pwndbg:   2026.02.18 (Linux)
Python:   3.14.3
GDB:      18.0.50.20260221-git
Capstone: 5.0.7
Unicorn:  2.1.4
Pwnlib:   4.15.0
```

k4lizen:
cc: @patryk4815

PR #3849, opened 2026-03-29
Ensure executable is loaded before pwndbg init

SWAPNILVERMA108:
Ensure the executable is loaded before pwndbg initialization to prevent delayed symbol loading and issues with .gdbinit execution.

### DIFF
--- a/pwndbginit/gdbinit.py
+++ b/pwndbginit/gdbinit.py
@@ -58,6 +58,16 @@ def main() -> None:
     except gdb.error as e:
         print(f"Warning: Cannot set gdb charset: '{e}'")
 
+        # Ensure executable is loaded before pwndbg init
+    try:
+        if not gdb.current_progspace().filename:
+            # attempt to load file if provided
+            args = gdb.execute("show args", to_string=True)
+            if args:
+                gdb.execute("file " + args.strip(), to_string=True)
+    except Exception:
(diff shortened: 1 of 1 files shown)


SWAPNILVERMA108:
I added a fix so the executable loads before pwndbg initializes. Let me know if anything needs to be changed or improved.

disconnect3d:
Sorry, but what's the bug that this fix? How do we reproduce? (Is there an issue for it btw?)

SWAPNILVERMA108:
This fixes the issue where pwndbg initializes before the executable is loaded, which causes GDB to show warnings like "No executable has been specified" and delays symbol loading.

To reproduce:
1. Run "pwndbg vmlinux"
2. Observe that symbols are not loaded initially and warnings are shown
3. Symbols are only loaded later, after initialization

This behavior is described in issue #3840.

My fix ensures that the executable is loaded before pwndbg initialization, so symbols are available immediately and the warning is avoided.

disconnect3d:
cc @k4lizen @jxuanli @patryk4815

k4lizen:
image

same with 

`pwndbg /bin/sh`

k4lizen:
i don't see how this fixes the issue, reopen if you can actually repro the issue and the fix, or if i'm missing something

PR #3861, opened 2026-04-01
build base pwndbg image workflow: always inject buildkit cache mounts regardless of cache hit

jackmisbach:

https://github.com/reproducible-containers/buildkit-cache-dance/pull/61
After fixing an upstream issue in buildkit-cache-dance that caused the action to hang during extraction (released in v3.3.2 and adopted here in PR #3806 by dependabot), the workflow now works properly. So, emulated arches can run in minutes rather than hours even when registry cache is invalidated. 

ex: https://github.com/pwndbg/pwndbg/actions/runs/23828715881
<table>
<tr>
<td>chrome_NT3bfO58Dq</td>
<td>chrome_KEKYZN2Jl1</td>
</tr>
</table>

---

There is just one bug that needs to be fixed, though

`if: steps.cache.outputs.cache-hit != 'true'` 
This needs to be removed; otherwise, for example, in this PR from dependabot https://github.com/pwndbg/pwndbg/pull/3858, where `uv.lock` wasn't updated, the cache wasn't injected, and the registry cache only stores image layers, not the mount, which is the actual purpose of buildkit-cache-dance anyways, so it just always needs to be injected instead.

### DIFF
--- a/.github/workflows/base-image.yml
+++ b/.github/workflows/base-image.yml
@@ -115,7 +115,6 @@ jobs:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Inject/extract cache mounts
-        if: steps.cache.outputs.cache-hit != 'true'
         uses: reproducible-containers/buildkit-cache-dance@v3.3.2
         with:
           cache-map: |


disconnect3d:
> emulated arches can run in minutes rather than hours.

This is awesome! Thanks!

PR #3863, opened 2026-04-05
Fix ImportError: handle CS_MODE_RISCV_C renaming in capstone6

Mel-SRK:

### DIFF
--- a/pwndbg/aglib/arch_mod.py
+++ b/pwndbg/aglib/arch_mod.py
@@ -26,7 +26,10 @@ from capstone6pwndbg import CS_MODE_MIPS32
 from capstone6pwndbg import CS_MODE_MIPS64
 from capstone6pwndbg import CS_MODE_RISCV32
 from capstone6pwndbg import CS_MODE_RISCV64
-from capstone6pwndbg import CS_MODE_RISCVC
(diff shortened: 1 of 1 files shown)


Mel-SRK:
Verified that the CI failure on Arch Linux is unrelated to this PR, as other concurrent PRs are experiencing the same issue in the test suite. I have confirmed this fix locally on my Arch Linux machine with Python 3.14.

patryk4815:
Report this to archlinux for using unsupported version

image

Mel-SRK:
> Report this to archlinux for using unsupported version
> 
> 
> 
> image
> 
> 

ok.I submitted this PR to provide a temporary bridge for Arch users until the upstream dependency or packaging issue is resolved. If the team prefers to wait for a fix from Arch/Capstone side, feel free to close or hold this PR.

OBarronCS:
Unfortunately, a fix like this will give a false sense of security. While it will make the symbol names line up, thus avoiding a crash at startup, there will be runtime bugs because the underlying meaning of the symbols that capstone exports has changed between two versions (as mentioned in the screenshot you posted, we are waiting for additional fixes before merging an update of Capstone).

PR #3868, opened 2026-04-11
go: fix type `direct_iface` resolution

RocketMaDev:

This is a fix for go1.26, which breaks direct_iface behavior. The commit
golang/go@f7d167f
move direct_iface from Kind to TFlag, leading to this error.

This PR is intended to fix go_dump errors on archlinux test.

### DIFF
--- a/pwndbg/aglib/godbg.py
+++ b/pwndbg/aglib/godbg.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import dataclasses
 import json
+import re
 import string
 import struct
 import textwrap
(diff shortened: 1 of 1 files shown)


pwndbg/aglib/godbg.py
```diff
@@ -624,7 +624,8 @@ def _inner_decode_runtime_type(
         addr,
         size=size,
         align=align,
-        direct_iface=(kind_raw & (1 << 5)) != 0,
+        # go 1.26 move direct/indirect flag from Kind to TFlag
+        direct_iface=((kind_raw & (1 << 5)) | tflag & (1 << 5)) != 0,
```
disconnect3d, line 628: I assume this works for both 1.26 and the older versions right?

PR #3869, opened 2026-04-14
Fixed 'vbar' reg in AArch64 which should be `vbar_el1`

wired0ut:
I have been debugging an arm64 kernel, when I've noticed I could not get `kbase` and neither could `context` run correctly. 

I have then debugged it (after setting ptrace_scope to 0), and found out its due to the fact we try to find the `kbase` using the `vbar` register, which seems to be deprecated in aarch64 (https://developer.arm.com/documentation/ddi0601/2025-12/AArch64-Registers).

This is a small change, and so I've not opened an issue on the case or anything, but if it's needed let me know.

The environment in which it occurred was using `qemu-system-aarch64` on an aarch64 compiled android kernel (namely 4.14). After changing to `vbar_el1` it could find the `kbase` and I could debug normally.

### DIFF
--- a/pwndbg/aglib/kernel/paging.py
+++ b/pwndbg/aglib/kernel/paging.py
@@ -667,7 +667,11 @@ class Aarch64PagingInfo(ArchPagingInfo):
     @property
     @pwndbg.lib.cache.cache_until("stop")
     def kbase(self) -> int | None:
-        return self._kbase(pwndbg.aglib.regs.read_reg("vbar_el1"))
+        vbar_reg = pwndbg.aglib.regs.read_reg("vbar")
+        if vbar_reg is None:
+            vbar_reg = pwndbg.aglib.regs.read_reg("vbar_el1")
+
+        return self._kbase(vbar_reg)
 
     @property
     @pwndbg.lib.cache.cache_until("stop")
--- a/pwndbg/aglib/kernel/paging.py
+++ b/pwndbg/aglib/kernel/paging.py
@@ -667,7 +667,7 @@ class Aarch64PagingInfo(ArchPagingInfo):
     @property
     @pwndbg.lib.cache.cache_until("stop")
     def kbase(self) -> int | None:
-        return self._kbase(pwndbg.aglib.regs.read_reg("vbar"))
+        return self._kbase(pwndbg.aglib.regs.read_reg("vbar_el1"))
 
     @property
     @pwndbg.lib.cache.cache_until("stop")


patryk4815:
What qemu version?

For me it looks:
a) you use old qemu-system version, and qemu folks change register name
b) you use new qemu-system version, and  qemu folks change register name

Now your fix introduced "regression" in CI tests:
kbase return None, but tests expected valid integer
image

wired0ut:
> What qemu version?
> 
> For me it looks:
> a) you use old qemu-system version, and qemu folks change register name
> b) you use new qemu-system version, and  qemu folks change register name
> 
> Now your fix introduced "regression" in CI tests:
> kbase return None, but tests expected valid integer
> image
> 

I figured this might also be a problem.
Maybe the correct fix (I could change mine) would be to check whether 'vbar' or 'vbar_el1' exists (which one of them isn't None). I should be in a newer qemu version, as it's a fresh install from 'pacman'.

I propose a check like (pseudo-code):
```
if regs.vbar is None:
     return regs.vbar_el1
return regs.vbar
```
Instead of my change


Edit: Pushed the change I  proposed, take a look.

patryk4815:
@jxuanli please look, thanks!

jxuanli:
The changes make sense to me though I think qemu should try to be backwards
compatible with those things. One question would be would it make sense to
add an `*aliases: str` arg to the function `read_register` definition,
maybe @logo could commend on that.

jxuanli:
@OBarronCS

OBarronCS:
@wired0ut

Thanks for reporting this bug and making a fix! Could you let us know what version of qemu you are using?

For the scope of this PR, I think it's probably fine to hardcode the check of which register to use because it's so specific to this one callsite.

A more generic solution could be good for down the road, considering Pwndbg supports multiple backend debuggers (gdb, lldb, windbg support in the works)  and runtime environments (local, remote, emulated), and that register names could vary wildly between them. So far everything happens to have lined up, but inevitably we are going to get more cases like this. GDB has a Python API for reading supported registers, which might come in handy down the road: <https://sourceware.org/gdb/current/onlinedocs/gdb.html/Registers-In-Python.html>

> I have been debugging an arm64 kernel, when I've noticed I could not get kbase and neither could context run correctly.

Could you also please send a screenshot/describe what happens to the context (without this patch?). Ideally, our code can gracefully handle failures like this and the context can still display.

wired0ut:
@OBarronCS 

Sure thing, glad to help out.

The qemu version is :
```
❯ qemu-system-aarch64 --version
QEMU emulator version 10.2.0
Copyright (c) 2003-2025 Fabrice Bellard and the QEMU Project developers
```

And it is run with the following cmdline:
```
qemu-system-aarch64 \
  -M virt \
  -cpu cortex-a72 \
  -m 4G \
  -smp 1 \
  -kernel kernel/common/arch/arm64/boot/Image \
  -initrd initramfs.cpio.gz \
  -append "console=ttyAMA0 nokaslr panic=1" \
  -nographic \
  -gdb tcp::1234
```

I switched my version of `pwndbg` since the version of the `context` not working so I cannot fully reproduce the error that it had before (and I am also not close to the computer to debug right now). It looked like (maybe in an old version of pwndbg I had installed):
```
error: context: unsupported operand types for > int and NoneType
```

These are examples of other things that did not work that I have saved in my log:
```
 pwndbg> slab info kmalloc-512
 Slab Cache @ 0xffff8000da802680
     Name: kmalloc-512
     Flags: SLAB_DEBUG_FREE | SLAB_RED_ZONE | SLAB_POISON | SLAB_STORE_USER
     Offset: 0x208
     Slab size: 0x2000
     Size (including metadata): 0x400
     Align: 0x80
     Object Size: 0x200
     kmem_cache_cpu @ 0xffff8000fffda750 [CPU 0]:
         Freelist: 0x0
         Active Slab: (none)
         Partial Slabs: (none)
     kmem_cache_node @ 0xffff8000da801980 [NUMA node 0, nr_partial/min_partial: 0x1/0x5]:
slab: Prints information about the linux kernel's slab allocator SLUB.
Exception occurred: slab: unsupported operand type(s) for -: 'int' and 'NoneType' (<class 'TypeError'>)
pwndbg> kbase
Unable to locate the kernel base 
```

I will try to reproduce the error when I can, but I might be away from this computer in the next 2-3 days.

wired0ut:
Also, I think a more robust solution would be to use the 'aliases' as was suggested here, alongside the RegisterDescription gdb offers.

Then, inside the register lookup we can have a fallback that looks for any of the aliases. Way cleaner IMO.

disconnect3d:
@wired0ut 
> I will try to reproduce the error when I can, but I might be away from this computer in the next 2-3 days.

Did you maybe try it? :)

disconnect3d:
I'll merge this for now and create a GH issue to fix it more world-wide/globally. Please help us refactor that if you can

wired0ut:
@disconnect3d
Hey! Sorry for the late reply. 
Thanks for merging. I hadn't the time to try and reproduce, but I'll update if I succeed.

I'll try to help in whatever way I can with the issue, of course.

Issue #3875, opened 2026-04-21
Refactor register fetching to support aliases / backend-specific register names


disconnect3d:
We recently merged two AArch64 fixes that work around register-name mismatches in newer environments:

* https://github.com/pwndbg/pwndbg/pull/3869
* https://github.com/pwndbg/pwndbg/pull/3871

These fixes solve the immediate breakages, but they also expose a broader issue: pwndbg currently hardcodes some register names at individual callsites, while the actual names exposed by the debugger/runtime can vary across architectures, debugger backends, and emulation environments.

Some idea from ChatGPT:

<img width="636" height="553" alt="Image" src="https://github.com/user-attachments/assets/fb91aa2e-0bbb-43cf-a482-168ddac9e159" />

Please also see the discussion on https://github.com/pwndbg/pwndbg/pull/3869 on how to approach/fix this -- it seems debuggers may give the list of register names etc.

PR #3877, opened 2026-04-22
Add register-name aliases to read_reg (refactor)

dimssu:
Refs #3875

### What

Extend `RegisterManager.read_reg` (and its `_in_frame` / `_uncached` siblings) with a `*aliases: str` varargs. The first name whose read returns a non-None value wins; any trailing names are probed only on miss:

```python
pwndbg.aglib.regs.read_reg("vbar", "vbar_el1")
pwndbg.aglib.regs.read_reg("SCTLR", "SCTLR_EL1")
```

This is the `*aliases` shape proposed by @jxuanli in https://github.com/pwndbg/pwndbg/pull/3869#issuecomment-4255805541.

### Why

Different debugger backends, QEMU versions, and toolchains expose the same architectural register under different names. Recent fixes for AArch64 (#3869 for `vbar`/`vbar_el1`, #3871 for `SCTLR`/`SCTLR_EL1`) each added an inline `if X is None: try Y` fallback. This PR introduces a generic mechanism for that pattern and folds the two existing callsites onto it.

### Changes

- `pwndbg/aglib/regs_mod.py`: add `*aliases` to `read_reg`, `read_reg_in_frame`, `read_reg_uncached`, and `read_reg_uncached_in_frame`. The per-name cache (`@cache_until("stop")`) is preserved by delegating each name to a private `_read_single_reg_in_frame` helper — single-name calls behave identically to before.
- `pwndbg/aglib/kernel/paging.py`: collapse the `vbar` / `vbar_el1` fallback in `Aarch64PagingInfo.kbase` onto `read_reg("vbar", "vbar_el1")`.
- `pwndbg/aglib/kernel/__init__.py`: same for `SCTLR` / `SCTLR_EL1` in `Aarch64Ops.paging_enabled`.

Purely additive API change — every existing caller that passes a single name continues to behave exactly as before.

### Out of scope

- Not touching other ARM sysreg callsites (`TCR_EL1`, `ID_AA64MMFR2_EL1`, `TTBR{0,1}_EL1`, `sp_el0`, `TPIDR_EL0`) — no reported mismatches there yet.
- Not building a `RegisterSet.aliases` registry — @OBarronCS noted per-callsite handling is fine for now; `*aliases` keeps that option open.
- Not touching `pwndbg/emu/emulator.py`'s unicorn `read_register` (orthogonal).

### Testing

- `./lint.sh --check` equivalent: ✅ ruff format, ruff check, custom-lint, mypy, vermin all clean.
- `tests/unit_tests/`: ✅ 46 passed, 4 skipped.
- **Not manually run against an aarch64 QEMU kernel locally** — relying on the existing `tests/library/qemu_system` suite in CI to cover that path. A maintainer smoke-test of `kbase` / `context` / `vmmap` on an aarch64 kernel session would be appreciated.

### PR checklist

+ [ ] I am providing a screenshot of the (new feature) / (fixed bug). (N/A — internal API refactor with no user-visible UI change; the user-visible outcome was captured in the screenshots on #3869.)

### DIFF
--- a/pwndbg/aglib/regs_mod.py
+++ b/pwndbg/aglib/regs_mod.py
@@ -59,7 +59,18 @@ class RegisterManager:
 
         return None
 
-    def read_reg_uncached_in_frame(self, reg: str, frame: pwndbg.dbg_mod.Frame) -> int | None:
+    def read_reg_uncached_in_frame(
+        self, reg: str, frame: pwndbg.dbg_mod.Frame, *aliases: str
+    ) -> int | None:
+        for name in (reg, *aliases):
+            value = self._read_single_reg_uncached_in_frame(name, frame)
+            if value is not None:
+                return value
+        return None
+
+    def _read_single_reg_uncached_in_frame(
+        self, reg: str, frame: pwndbg.dbg_mod.Frame
+    ) -> int | None:
         reg = reg.lstrip("$")
         try:
             value = self.get_register(reg, frame)
@@ -85,23 +96,36 @@ class RegisterManager:
         except (ValueError, pwndbg.dbg_mod.Error):
             return None
 
-    def read_reg_uncached(self, reg: str) -> int | None:
+    def read_reg_uncached(self, reg: str, *aliases: str) -> int | None:
         frame = pwndbg.dbg.selected_frame()
         if frame is None:
             return None
-        return self.read_reg_uncached_in_frame(reg, frame)
+        return self.read_reg_uncached_in_frame(reg, frame, *aliases)
 
-    @pwndbg.lib.cache.cache_until("stop")
-    def read_reg_in_frame(self, reg: str, frame: pwndbg.dbg_mod.Frame) -> int | None:
+    def read_reg_in_frame(self, reg: str, frame: pwndbg.dbg_mod.Frame, *aliases: str) -> int | None:
         """
         Same as read_reg() except for the provided frame, rather than the currently
         selected frame.
         """
-        return self.read_reg_uncached_in_frame(reg, frame)
+        for name in (reg, *aliases):
+            value = self._read_single_reg_in_frame(name, frame)
+            if value is not None:
+                return value
+        return None
 
-    def read_reg(self, reg: str) -> int | None:
+    @pwndbg.lib.cache.cache_until("stop")
+    def _read_single_reg_in_frame(self, reg: str, frame: pwndbg.dbg_mod.Frame) -> int | None:
+        return self._read_single_reg_uncached_in_frame(reg, frame)
+
+    def read_reg(self, reg: str, *aliases: str) -> int | None:
         """
         Query the underlying debugger for the value of a register.
+
+        If `aliases` are provided, each is tried in turn after `reg` until one
+        returns a non-None value. This is useful for architectural registers
+        whose exposed names differ across debugger backends, runtime
+        environments, or toolchain versions (for example, `vbar` vs
+        `vbar_el1` on AArch64).
 
         Note that in some rare cases, debuggers won't directly expose the values of some special model specific registers.
         Although we can sometimes determine these by other indirect means, this function does not run any extra logic to handle these special cases.
@@ -116,7 +140,7 @@ class RegisterManager:
         frame = pwndbg.dbg.selected_frame()
         if frame is None:
             return None
-        return self.read_reg_in_frame(reg, frame)
+        return self.read_reg_in_frame(reg, frame, *aliases)
 
     def write_reg(self, reg: str, value: int) -> None:
         if not pwndbg.dbg.selected_frame().reg_write(reg, value):
--- a/pwndbg/aglib/regs_mod.py
+++ b/pwndbg/aglib/regs_mod.py
@@ -63,38 +63,33 @@ class RegisterManager:
         self, reg: str, frame: pwndbg.dbg_mod.Frame, *aliases: str
     ) -> int | None:
         for name in (reg, *aliases):
-            value = self._read_single_reg_uncached_in_frame(name, frame)
-            if value is not None:
-                return value
+            name = name.lstrip("$")
+            try:
+                value = self.get_register(name, frame)
+                if value is None and name.lower() == "xpsr":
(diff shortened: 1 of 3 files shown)


pwndbg/aglib/regs_mod.py
```diff
@@ -59,7 +59,18 @@ def get_qemu_register(self, name: str) -> int | None:
...
 
-    def read_reg_uncached_in_frame(self, reg: str, frame: pwndbg.dbg_mod.Frame) -> int | None:
+    def read_reg_uncached_in_frame(
+        self, reg: str, frame: pwndbg.dbg_mod.Frame, *aliases: str
+    ) -> int | None:
+        for name in (reg, *aliases):
+            value = self._read_single_reg_uncached_in_frame(name, frame)
+            if value is not None:
+                return value
+        return None
```
disconnect3d, line 92: What is the error that we are getting out of the debugger if the register name does not match?

Can't we just do something like that? I feel like this would be much simpler. Because now we are introducing some additional APIs :(
```py
try:
    <try logic>
except WhateverErrorIsThereWhenRegIsMissing:
    if aliases[0] is not None:
        return self.read_reg_uncached_in_frame(aliases[0], frame, aliases[1:])
<other exceptions handling>
```

We should also:
1) Document each place that uses multiple register names, why certain names are used, where exactly do they come from - like versions of QEMU.
2) Order those register names by the most probable one that will be used by users to the least probable one. So likely, new QEMU versions first, then older ones etc.

pwndbg/aglib/regs_mod.py
```diff
@@ -85,24 +96,37 @@ def read_reg_uncached_in_frame(self, reg: str, frame: pwndbg.dbg_mod.Frame) -> i
...
+        return self._read_single_reg_uncached_in_frame(reg, frame)
+
+    def read_reg(self, reg: str, *aliases: str) -> int | None:
         """
         Query the underlying debugger for the value of a register.
 
+        If `aliases` are provided, each is tried in turn after `reg` until one
+        returns a non-None value. This is useful for architectural registers
+        whose exposed names differ across debugger backends, runtime
+        environments, or toolchain versions (for example, `vbar` vs
```
disconnect3d: The comment here is meh: the "runtime environments, or toolchain versions" has nothing to do here, does it?

pwndbg/aglib/regs_mod.py
dimssu: Good call — simplified. Pushed 9f91b1c.

To answer the question: the debugger doesn't raise when a register name is missing. `get_register()` calls `regs.by_name(name)` which just returns `None`. So exception-based dispatch isn't needed; `value is None` → `continue` to the next alias is the only path.

That let me drop the private helper entirely: the loop is now inlined in `read_reg_uncached_in_frame` with `continue` on each miss, and `read_reg_in_frame` is back to being a one-liner with `@cache_until("stop")` wrapping the `(reg, frame, *aliases)` tuple as a single cache key.

On the two requests:
1. **Documenting each callsite** — added comments at both callsites explaining what the register is, which QEMU releases use which name, and the relevant PR references (#3869, #3871, #3875).
2. **Ordering newer-first** — flipped both callsites: `read_reg("vbar_el1", "vbar")` and `read_reg("SCTLR_EL1", "SCTLR")`. Newer QEMU (where the reporter hit the bug) is tried first.

pwndbg/aglib/regs_mod.py
dimssu: Fair. Trimmed to just mention backend/QEMU-version differences with `vbar_el1` vs `vbar` as the concrete example. Pushed in 9f91b1c.

pwndbg/aglib/regs_mod.py
wired0ut: @disconnect3d
The logic could also use the GDB Python API to simply find which of the names are available, I think it's a cleaner solution than a try except, no? I planned to implement it, but I see there's an MR now.

disconnect3d: @wired0ut Probably? How would u do this? Map the names or..?

Ideally, we don't want to do this on each register access/cache clear etc.

wired0ut: @disconnect3d 
We might be able to utilize the following API: https://sourceware.org/gdb/current/onlinedocs/gdb.html/Registers-In-Python.html#Registers-In-Python.

I tried to poke around and see if there are any unique identifiers to the registers which we will be able to use, but I did not see anything trivial that we can use here (in `maint print xml-tdesc` there are IDs, but I don't think they'll persist when the registers change etc).

We can save the names one time and use a cache to not re-call the API each time (or something else if you have a suggestion?) and then simply search for the registers instead of a try-except. As an example:
```py
pwndbg> pi
>>> arch = gdb.selected_frame().architecture()
>>> arch.registers()
<gdb.RegisterDescriptorIterator object at 0x7fa7f1687e10>
>>> regs = arch.registers()
>>> regs.find("VBAR")
>>> regs.find("VBAR_EL1")
<gdb.RegisterDescriptor object at 0x7fa7f16a1050>
```


On a side note, if we want to search for a better solution:
You probably know more about the QEMU-gdb integration. Do you know of any IDs we can map them with? The `RegisterDescriptor` only gives a name and nothing else we can use to map unfortunately.

disconnect3d: I guess that would work fine, then we can save that information in <somewhere> (architecture object? somewhere else?).

wired0ut: @disconnect3d
Yes, some sort of an Architecture object, if it's pointed to by the frame it's even better.

k4lizen (approved):
Thanks for the PR!

I think this is good to merge.

k4lizen:
@wired0ut if you'd like to take a stab at implementing a cleaner solution by querying the underlying providers API and seeing which regs are available, i think this would be pretty neat, feel free to do it :>

wired0ut:
@k4lizen I'll try to find time and implement the cleaner solution, hopefully soon.

PS: nice to see another Lain picture 🙃

PR #3893, opened 2026-05-01
procinfo: show peer process for unix-domain sockets

disconnect3d:
For local debug targets, query the kernel via NETLINK_SOCK_DIAG to map each unix socket inode to its peer inode, then resolve that peer back to a (pid, fd, comm) tuple by walking /proc/*/fd. This turns lines like "unix '(anonymous)'" into "unix '(anonymous)' (inode=N, peer pid=M 'comm' fd=K)", which is the missing piece for understanding which processes a debuggee is actually talking to.

SOCK_DIAG is netlink-only (no /proc fallback), so the lookup is gated on aglib.remote.is_remote(): for gdbserver-on-another-machine and QEMU-system targets we silently skip rather than reporting peers from the wrong kernel. Same kernel for kernel-mode targets is a separate problem (walk unix_sock->peer) and is left for later.

### DIFF
--- a/pwndbg/commands/procinfo.py
+++ b/pwndbg/commands/procinfo.py
@@ -6,10 +6,12 @@ import string
 import pwndbg.aglib.file
 import pwndbg.aglib.proc
 import pwndbg.aglib.qemu
+import pwndbg.aglib.remote
 import pwndbg.auxv
 import pwndbg.commands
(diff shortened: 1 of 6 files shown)


disconnect3d:
Before:
image

After:
image

PR #3894, opened 2026-05-01
Fix #3150: procinfo: show pipe peers and read/write end

### LINKED ISSUE #3150: show pipe destination in `procinfo`
Image

Show the pid and name of the process to which the pipe is connected. If there are multiple (https://www.man7.org/linux/man-pages/man3/mkfifo.3.html, in this case the pipe name should be shown as well, if it isn't already) show them all (possibly in a new line if there are a lot of them).

See: https://serverfault.com/questions/48330/how-can-i-get-more-info-on-open-pipes-show-in-proc-in-linux
The `ls -l /proc/*/fd/ | grep $PIPE_ID; ls -l /proc/*/fd/$FD | grep $PIPE_ID` thing might work.

Additional features:
+ Add some coloring.
+ If the process owns both ends of a pipe, make that evident. From the screenshot this is the case here:
```
fd[5]      pipe:[1109875]
fd[6]      pipe:[1109875]
```
+ Show whether the fd is the receiving or transmiting end of a pipe.

disconnect3d:
Walk /proc/*/fd to collect every endpoint of each pipe inode our debuggee holds, read /proc/PID/fdinfo/FD to tag each as the read or the write end (O_RDONLY vs O_WRONLY in the bottom two bits of the open flags), then attach the resulting (pid, fd, comm, mode) list to the corresponding FD in procinfo's output. Same-process peers (e.g. our own pipe(2) pair, or a fork that kept both ends) become obvious because the peer pid matches ours; cross-process peers identify the actual reader or writer.

Like the unix-socket peer feature, this is local-only and gated on aglib.remote.is_remote() so we don't report peers from the wrong kernel when the inferior lives on another machine.

### DIFF
--- a/pwndbg/commands/procinfo.py
+++ b/pwndbg/commands/procinfo.py
@@ -11,6 +11,7 @@ import pwndbg.auxv
 import pwndbg.commands
 import pwndbg.lib.cache
 import pwndbg.lib.net
+import pwndbg.lib.proc_fd
 import pwndbg.lib.sock_diag
 from pwndbg.color import message
 from pwndbg.commands import CommandCategory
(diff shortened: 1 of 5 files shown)


disconnect3d:
That's how it looks now:
image

pwsmaster (approved):

pwsmaster (approved):

pwsmaster (approved):

Issue #3784, opened 2026-02-19
show context disassm above the exact line where a breakpoint was hit


k4lizen:
Here I set a breakpoint right after an important function call, because I want to quickly get the `$rax` return value. 

<img width="1050" height="287" alt="Image" src="https://github.com/user-attachments/assets/2d98b494-b8ca-4ff2-a7ad-a2ed312f0323" />

It is however quite annoying to me that I do not see the disassembly of the call instruction. Especially because this is x86_64 which has variable length instruction encoding so I legitimately don't know how to "go up" in disassembly other than guessing the offset.

OBarronCS:
Currently, this works in x86 under one condition: the disassembly system has already disassembled the call instruction/the stuff before it. In other, words you were already paused before the call, and it disassembled the call and the subsequent instruction you set the breakpoint at. During this process, it caches the instruction flow, and upon hitting the breakpoint it would remember where you came from.

For fixed-width instruction sets, this currently always works since we don't need to have already disassembled the instructions to know where they lie.

For the case of x86, one solution that would work in this scenario would be to identify that we haven't encountered this instruction flow before, and don't know how to disassemble backwards. Then, find the current function we are in based on the symbol (`module_ioctl`), start disassembling from the start until you get to where you are, which is a safe way to determine where the instructions are so we can display the ones behind the PC.

k4lizen:
that's pretty smart!

dunno what bata24/gef is doing, might be worth checking out, since the few instructions above are shown when using it

OBarronCS:
Good call - for variable width instruction sets, `gef` uses a "guessing" mechanism to find the previous instructions (it disassembles 1 byte before, then 2, then 3, up to 15 bytes, until a valid instruction is seen). This can fail if the bytes decode to instructions too early, although this is rarely seen because they search multiple instruction back using the mechanism, and decoding real x86 instruction streams "self-aligns"/resynchronizes in practice, so the real instructions become seen. 

https://github.com/bata24/gef/blob/a69f6d0112d5318cbf81b962f2e346f1210c36c6/gef.py#L6258-L6278

disconnect3d:
That would be neat feature tbh.

OBarronCS:
Implemented in #3898 and #4142

k4lizen:
yay

PR #3898, opened 2026-05-07
Disassemble backwards in variable width instruction sets

OBarronCS:
This implements disassembling backwards in variable width instruction architectures to implement #3784.

Prior to this PR, if you hit a breakpoint or did `nearpc random_address`, and if the disassembly system had not already encountered the surrounding sequence of instructions before, then we were unable to display the "previous" instructions behind the instruction pointer. 

This implements a method to  "guess" the valid sequence of instructions that lead to the current instruction, allowing disassembling backwards. Since ISA's like x86 are not self-synchronizing, we do a best-effort guess at determining the true instruction boundaries.

The method is simple: start a certain amount of bytes in the past, and start disassembling towards the current instruction. We will assume that after a certain number of disassembled instructions, that the sequence will "self-align" to the real sequence (which works in practice).

There is one case where this might not be desirable (i.e. changes longstanding pwndbg behavior), which is stepping into a function call. Previously, when you stepped into a call, the first instruction would be the first one displayed. Now, because we can disassemble backwards, it displays the instructions linearly behind the instruction.

This can get confusing, giving we are now mixing emulation (displaying the true sequence of instruction) with sometimes displaying instructions linearly behind the address.

image

image

### DIFF
--- a/pwndbg/aglib/disasm/disassembly.py
+++ b/pwndbg/aglib/disasm/disassembly.py
@@ -576,6 +576,7 @@ def near(
     show_prev_insns=True,
     use_cache=False,
     linear=False,
+    max_backwards_linear_count: int | None = None,
 ) -> tuple[list[PwndbgInstruction], int, int]:
     """
     Disassembles instructions near given `address`. Passing `emulate` makes use of
@@ -665,6 +666,7 @@ def near(
     # The assumption is that the instruction list will start with the linear instructions, and then transition to the emulated one
     index_of_last_linearly_disassembled_instruction = -1
 
+    count_backwards_linear = 0
     if show_prev_insns:
         saveptr = InstructionSequenceSavePointer(None)
 
@@ -675,6 +677,12 @@ def near(
             insn, was_linear = prev_instruction_fetch
 
             if was_linear:
+                count_backwards_linear += 1
+                if (
+                    max_backwards_linear_count is not None
+                    and count_backwards_linear > max_backwards_linear_count
+                ):
+                    break
                 index_of_last_linearly_disassembled_instruction += 1
 
             if DEBUG_ENHANCEMENT:
@@ -684,7 +692,10 @@ def near(
             insns.append(insn)
 
             prev_instruction_fetch = get_previous_instruction(
-                insn.address, use_cache=use_cache, linear=linear, saveptr=saveptr
+                insn.address,
+                use_cache=use_cache,
+                linear=linear,
+                saveptr=saveptr,
             )
         insns.reverse()
 
--- a/pwndbg/aglib/disasm/disassembly.py
+++ b/pwndbg/aglib/disasm/disassembly.py
@@ -572,10 +572,10 @@ def near(
     backward_count: int = 0,
     total_count: int | None = None,
     end_address: int | None = None,
-    emulate=False,
-    show_prev_insns=True,
-    use_cache=False,
-    linear=False,
+    emulate: bool = False,
+    show_prev_insns: bool = True,
+    use_cache: bool = False,
+    linear: bool = False,
     max_backwards_linear_count: int | None = None,
 ) -> tuple[list[PwndbgInstruction], int, int]:
     """
--- a/pwndbg/aglib/disasm/disassembly.py
+++ b/pwndbg/aglib/disasm/disassembly.py
@@ -55,6 +55,14 @@ Enabling this may make disassembly slower.
 """,
 )
 
+# This global flag is used in tests to disable heuristic-based linear backwards disassembly
+heuristic_backwards_linear_disassembly_enabled = pwndbg.config.add_param(
+    "heuristic-backwards-disasm",
+    True,
+    "toggle backwards linear disassembly during emulation",
+    param_class=pwndbg.lib.config.PARAM_BOOLEAN,
+)
+
 # Caching strategy:
 # To ensure we don't have stale register/memory information in our cached PwndbgInstruction,
 # we clear the cache whenever we DON'T do a `stepi`, `nexti`, `step`, or `next` command.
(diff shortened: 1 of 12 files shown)


OBarronCS:
The test failures are from now disassembling backwards where we previously couldn't. I'm considering adding a setting for the tests like `disable_backwards_heuristic_disassembly" for the tests, rather than changing the 75+ that have changed.

k4lizen:
> The test failures are from now disassembling backwards where we previously couldn't. I'm considering adding a setting for the tests like `disable_backwards_heuristic_disassembly" for the tests, rather than changing the 75+ that have changed.

I think that's fine, though would be nice to then have some tests that cover this behavior specifically.

k4lizen:
I'm thinking if we should only disassemble backwards until we hit a new symbol, or if it makes sense to do it this way where you can see behind the function you call into.

k4lizen:
> This can get confusing, giving we are now mixing emulation (displaying the true sequence of instruction) with sometimes displaying instructions linearly behind the address.

We could color code them (greying them out seems quite intuitive) or set a marker right above the PC that denotes this maybe (something like `┻`? to signal "everything above this is linear").

k4lizen:
btw i didn't check the source yet but do you use function beginnings as a heuristic to help with disassembling backwards

OBarronCS:
It currently doesn't check function boundaries/symbols.

image

OBarronCS:
~~I checked the tests that failed and played around with it more: before merging, I think that in the context disasm view, we should cut off disassembling backwards at function boundaries (dozens of the tests rely on this behavior), and I think it would be most natural (or maybe make this configurable).~~

I think it would be nice to make it configurable if, on for the context disasm view, we want to automatically disassemble backwards beyond the start of the current symbol.

This would rely on the ability to accurately function boundaries even without debugging symbols.

But in the current state, it's fine.

One change that does need to be made is that in the context disasm view, the number of instructions that are disassembled backwards from the heuristics needs to be capped: it's not nice to have half the screen filled with these grayed out instructions when jumping into a function call when `context-disasm-lines` is a large number.

k4lizen:
makes sense

OBarronCS:
I added a cap to the max number of linearly disassembled backwards instructions are displayed in the context view (with a config to change it).

The logic is:
```
 if disasm_backwards_linear_lines == "auto":
      max_backwards_linear_count = min(10, total_lines_on_screen // 3)
  else:
      max_backwards_linear_count = int(disasm_backwards_linear_lines)
```

pwndbg/commands/context.py
```diff
@@ -1335,6 +1335,12 @@ def get_regs(in_regs: list[str | VisitableRegister | None] | None = None) -> lis
     "context-disasm-lines", 10, "number of additional lines to print in the disasm context"
 )
 
+disasm_backwards_linear_lines = pwndbg.config.add_param(
+    "context-disasm-back-linear-lines",
+    "auto",
+    "maximum number of lines to disassemble backwards linearly in memory",
+)
```
OBarronCS, line 1342: I couldn't find any examples of other config variables that are both integers but could also be special word "auto" for a default behavior

OBarronCS:
image

image

OBarronCS:
There are so many different callsite for the disassembly. The backwards disassembly is breaking tests. A global disable during the tests for the feature is going to be the easiest way to not break all existing tests.

OBarronCS:
Added a test for this feature + some edge cases I came across. This is ready to go. 

To summarize, this PR adds two closely related features:

1. In the context disassembly view, we used to ONLY show instructions that were disassembled as part of the emulated disassembly flow. Now, we can disassemble "backwards" from the PC linearly, to show the linearly previous instructions, which are displayed greyed out to indicate this. See screenshots earlier in this PR

3. A method to disassemble instructions "backwards" in memory. So if you do "nearpc *random_address", or `continue` the program until a breakpoint, it uses a "heuristic" method to disassemble instructions backwards in memory. This is referred to as a "heuristic" in the code, since due to variable length instructions sets there is no perfect way to do it.

The two new configs.

```python
# In the context disasm view, the max amount of how many grayed out "linear" lines disassembled backwards that you can display. "auto" is a nice default, with a max of a 3rd of the total instructions or 10 instructions, whichever is smallest
set context-disasm-back-linear-lines <number|auto> # auto by default
# Disable the heuristic all together
set heuristic-backwards-disasm <on/off>                 # on by default
```

k4lizen (commented):
can we get rid of `heuristic-backwards-disasm` and just use `context-disasm-back-linear-lines == 0`?

OBarronCS:
> can we get rid of `heuristic-backwards-disasm` and just use `context-disasm-back-linear-lines == 0`?

The heuristic is for disassembling backwards when we know don't instruction flow ahead of time. This setting enables it in ALL disassembly contexts where we are disassembling around a point.

The `context-disasm-back-linear-lines` setting is specifically is that "for the context disasm view, show this many grayed out linear backwards instructions".

We may want to show 0 lines in the context view, but for normal "nearpc" command invocation, still have it do the backwards disassembly. Having the two settings makes this possible.

pwndbg/aglib/disasm/disassembly.py
```diff
@@ -127,6 +130,14 @@ class InstructionSequenceSavePointer:
     node: InstructionSequenceNode | None
 
 
+# Dict of Address -> previous instruction sequentially in memory
+# Some architectures don't have fixed-sized instructions, so this is used
+# to disassemble backwards linearly in memory for those cases
+linear_backward_address_cache: collections.defaultdict[int, int] = collections.defaultdict(
+    lambda: None
```
disconnect3d, line 137: shouldn't this be: `lambda: 0`? if e map int to int?

disconnect3d, line 137: Or should the type be: `collections.defaultdict[int, int | None]`

OBarronCS: You're right, it should be type `int | None`. Fixed this.

PR #3899, opened 2026-05-07
Add  GDB convenience variable to detect presence of pwndbg in .gdbinit

OBarronCS:
This creates a variable, `$pwndbg_plugin_enabled`, when Pwndbg starts up in GDB.

This allows `.gdbinit` files to detect the presence of pwndbg and conditionally set pwndbg-specific settings

Example `.gdbinit`:
```
source /path/to/pwndbg/gdbinit.py

if ! $_isvoid($pwndbg_plugin_enabled)

    set context-code-lines 40

end
```

This makes it easier to quickly edit the `.gdbinit` and commenting/uncommenting the one `source /path/to/gdbinit.py` file to toggle it, and not have to change the rest of the pwndbg-specific lines in the file.

### DIFF
--- a/pwndbg/dbg_mod/gdb/__init__.py
+++ b/pwndbg/dbg_mod/gdb/__init__.py
@@ -1765,12 +1765,12 @@ class GDB(pwndbg.dbg_mod.Debugger):
 
         from pwndbg.dbg_mod.gdb import debug_sym as debug_sym
 
-        self._load_gdbinit()
-
         # Set a variable, "$pwndbg_plugin_enabled", that a .gdbinit config can use to detect
         # that pwndbg is enabled with `if ! $_isvoid($pwndbg_plugin_enabled)`
         # This allows the .gdbinit to conditionally configure pwndbg settings
         gdb.set_convenience_variable("pwndbg_plugin_enabled", True)
+
+        self._load_gdbinit()
 
         # show_hint must be called after loading ~/.gdbinit, this order allow disabling show_hint
         prompt.show_hint()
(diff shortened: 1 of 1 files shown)


pwndbg/dbg_mod/gdb/__init__.py
```diff
@@ -1767,6 +1767,11 @@ def setup(self):
 
         self._load_gdbinit()
```
patryk4815: `gdb.set_convenience_variable` should be before this line?

patryk4815: for `pwndbg` binary/entrypoint

OBarronCS: I moved it above

k4lizen:
i think this is a small enough change that it doesn't hurt, but i don't really understand the usecase, you can set your script up like this:

```.gdbinit
# pwndbg config
define pwndbg_run
	source ~/code/pwndbg/gdbinit.py
	set max-visualize-chunk-size 0x50
	set exception-verbose on
	set show-flags on
	set auto-explore-pages no
	set disasm-reg-alias on
	set gdb-workaround-stop-event disabled-deadlock
end

# == Run the appropriate gdb script ==
pwndbg_run
```
If you want a quick way to comment out the pwndbg invocation.

Would be nice to mention the var somewhere in the docs as well (and set it in LLDB too) if you do have a use-case that you think could come in handy for other people as well

OBarronCS:
Yeah, my use case could be done with the `define pwndbg_run` you gave. I'll just use that, no need to add this right now.

PR #3901, opened 2026-05-08
Fix messed up stack repeating logic

### LINKED ISSUE #3900: command repeat logic does not account for external invocations
### Description

Reported by a lakectf player (ty `vy`).

```
file /bin/ls
starti
stack
<ENTER>
context
si
```
The stack shown in the `context` command retains the `repeat` flag, even though it shouldn't, and the stack is not shown from `$sp`. The repeat should only happen on explicit invocations of the `tele` or `stack` command by the user, or if something using this API explicitly requests repetion.

k4lizen:
Fixes https://github.com/pwndbg/pwndbg/issues/3900 .


pwndbg/commands/telescope.py
```diff
@@ -312,9 +324,12 @@ def regs_or_frame_offset(addr: int, bp: int | None, regs: dict[int, str], longes
 @pwndbg.commands.OnlyWhenRunning
 def stack(count, offset, frame, inverse) -> None:
     ptrsize = pwndbg.aglib.typeinfo.ptrsize
-    telescope.repeat = stack.repeat
```
k4lizen, line 315: essentially, this code should have been setting `telescope.repeat = False` after the invocation, but I feel this is error prone so I just made it an arg, similar to how it is for nearpc.

PR #3906, opened 2026-05-09
Mark unresolved branches in disassembly view with question mark

OBarronCS:
While disassembling instructions for the disasm context view, emulation is used to determine the predicted instruction flow. However, we do not emulate through through certain instruction, such as syscalls and function calls. Upon hitting these instructions, emulation is disabled.

This PR adds a question mark icon to indicate that a future branch in the disassembly view has yet to have its condition resolved (because we are unable to determine the condition).

image


pwndbg/aglib/disasm/instruction.py
```diff
@@ -140,8 +140,10 @@ class InstructionCondition(Enum):
     TRUE = 1
     # Conditional instruction, but action is not taken
     FALSE = 2
-    # Unconditional instructions (most instructions), or we cannot reason about the instruction
-    UNDETERMINED = 3
+    # Conditional instruction, but we cannot reason about if the condition is true or not
+    UNDETERMINED_CONDITIONAL = 3
+    # Unconditional instructions (most instructions). This is the default
+    UNCONDITIONAL = 4
```
OBarronCS, line 146: This is the core change: being able to delineate between unconditional instructions and conditional instructions that we cannot resolve at the moment. Previously these two groups were grouped together.

tests/library/qemu_user/tests/test_arm.py
```diff
@@ -15,6 +15,15 @@
 mov r0, 0
 mov r7, 0xf8
 swi #0
+nop
```
OBarronCS, line 18: Added these nops because otherwise it was disassembling some random bytes that were there (that happened to be a conditional instruction, which broke the tests).

k4lizen:
tests do be failing

tests/library/qemu_user/tests/test_arm.py
```diff
@@ -538,13 +551,13 @@ def test_arm_exclusive_store(qemu_assembly_run):
     expected = (
         "LEGEND: STACK | HEAP | CODE | DATA | WX | RODATA\n"
         "──────────────────[ DISASM / arm / arm mode / set emulate on ]──────────────────\n"
-        " ► 0x200d4 <_start>       ldr    r0, [pc, #0x34]     R0, [_start+60] => 0x3011c (value1) ◂— 0\n"
-        "   0x200d8 <_start+4>     ldr    r1, [pc, #0x34]     R1, [_start+64] => 0x87654321\n"
-        "   0x200dc <_start+8>     ldr    r2, [pc, #0x34]     R2, [_start+68] => 0x12345678\n"
+        " ► 0x200d4 <_start>       ldr    r0, [pc, #0x58]     R0, [_start+96] => 0x30140 (value1) ◂— 0\n"
```
k4lizen, line 554: hmm, where do these changes come from?

k4lizen (approved):

tests/library/qemu_user/tests/test_arm.py
OBarronCS: Because I added the extra `nops` to one of the snippets, variable locations (which are placed after the instructions) were pushed down a bit.

disconnect3d (approved):

Issue #3341, opened 2025-10-09
lldb stepi is quite slow on macOS


disconnect3d:
I noticed that we are a bit slow on macOS: why is that so? Seems like some regression/bug :/

Is this maybe through some vmmap changes? Or something else? Need to benchmark it somehow.

patryk4815:
@disconnect3d did you confirm this with old release?

disconnect3d:
Yes, the 2025.05 release is also quite slow :(

disconnect3d:
@mbrla0 can you help with this?

patryk4815:
`GetMemoryRegions` from lldb is slow

```
In [5]: %timeit pwndbg.dbg.selected_inferior().process.GetMemoryRegions()
270 ms ± 4.12 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

mbrla0:
I was able to bring it down to 0.28ms from 0.68ms by caching one extra thing on our end, which makes things better, mostly by having us avoid interacting with LLDB. But, as @patryk4815 noted, there is a practical limit on how fast we can go, because of how slow LLDB is to fetch memory maps.

Here's the profile for a single `si` command, after the extra caching decorator was added.
```
         72155 function calls (71972 primitive calls) in 0.472 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    0.264    0.264 /nix/store/63r688f8bqdnxcm4585ccw7sagbak5kv-pwndbg-env/lib/python3.13/site-packages/pwndbg/dbg/lldb/__init__.py:2112(_fire_prompt_hook)
        1    0.000    0.000    0.264    0.264 /nix/store/63r688f8bqdnxcm4585ccw7sagbak5kv-pwndbg-env/lib/python3.13/site-packages/pwndbg/dbg/lldb/hooks.py:84(prompt_hook)
      2/1    0.000    0.000    0.264    0.264 /nix/store/63r688f8bqdnxcm4585ccw7sagbak5kv-pwndbg-env/lib/python3.13/site-packages/pwndbg/commands/__init__.py:368(__call__)
        1    0.000    0.000    0.264    0.264 /nix/store/63r688f8bqdnxcm4585ccw7sagbak5kv-pwndbg-env/lib/python3.13/site-packages/pwndbg/commands/context.py:708(context)
[...]

PR #3913, opened 2026-05-17
Persist vmmap across stops on macOS

disconnect3d:
See https://github.com/pwndbg/pwndbg/issues/3341

This is an experiment; I'm figuring out if we can do better here


pwndbg/aglib/vmmap.py
```diff
@@ -23,6 +26,25 @@
...
+vmmap_cache_param = pwndbg.config.add_param(
+    "vmmap-cache",
+    _vmmap_cache_default,
+    "cache the memory map for the whole run on slow targets (macOS)",
+    help_docstring=(
+        "On macOS, fetching the process memory map via LLDB is slow (every "
+        "region requires a Mach IPC round-trip). When this is on, the memory "
+        "map is fetched once per launch/attach and reused across stops until "
+        "the program exits, you re-launch/attach, or you run `vmmap --refresh`. "
+        "Has no effect when the debuggee target isn't Darwin."
```
k4lizen: docstring should clarify that this is on by default on macos

pwndbg/commands/vmmap.py
```diff
@@ -198,6 +201,14 @@ def print_vmmap_gaps(pages: tuple[Page, ...]) -> None:
     action="store_true",
     help="Display unmapped memory gap information in the memory map.",
 )
+parser.add_argument(
+    "--refresh",
+    action="store_true",
+    help=(
+        "Drop the cached memory map (used on macOS) and re-fetch it. "
```
k4lizen:
```suggestion
        "Drop the cached memory map (used by default on macOS) and re-fetch it. "
```

k4lizen:
lgtm but should be tested by someone running macos ( @mbrla0 )

@disconnect3d i'm guessing you have access to a macos machine since you opened the issue?

can you drop a screenshot of the new statusline?

disconnect3d:
@k4lizen I do have macOS to test it on; I'll update the PR in upcoming days and drop screenshots here

PR #3914, opened 2026-05-18
Fix large frame offset

### LINKED ISSUE #3802: frame offset cut off for large frames
https://github.com/pwndbg/pwndbg/pull/3801#issuecomment-3968866180

Image

doesn't make sense to cut it off here

marywinter3:

Closes #3802

Previously, the offset was only displayed correctly when it fit within the ±0xFFF range. 
With this change, offsets up to ±0xFFFF are shown correctly, and the offset column has a fixed width so all rows stay aligned. 
Is a fixed width of ±0xFFFF sufficient, or would it be better to calculate the column width dynamically based on the actual largest offset in the current output?

### DIFF
--- a/pwndbg/commands/telescope.py
+++ b/pwndbg/commands/telescope.py
@@ -270,12 +270,15 @@ def telescope(
 
 def regs_or_frame_offset(addr: int, bp: int | None, regs: dict[int, str], longest_regs: int) -> str:
     # bp only set if print_framepointer_offset=True
-    if bp is None or regs[addr]:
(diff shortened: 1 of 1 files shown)


disconnect3d (approved):

k4lizen:
can we see a before/after screenshot?

k4lizen:
image

In my issue I thought the solution would be to do it dynamically, but I think this solution is much cleaner.

k4lizen (approved):

k4lizen:
hm I cannot update this branch with dev, i dunno why

disconnect3d:
Thanks!

Issue #3911, opened 2026-05-14
Add test for attaching to non-Linux program run by qemu-system


OBarronCS:
It would be good to have a test where we have the debugger attach to a qemu-system running a custom, non-Linux kernel going through the x86 boot process (starting at 16-bit real mode, GDB architecture i8086).

This would catch cases of event handlers or other code that cause crashes.

Past example of a crash: #3518

Currently, this function: https://github.com/pwndbg/pwndbg/blob/5ccbcb4c35d3685a312e106d2964228b8419b8f9/pwndbg/aglib/kernel/symbol.py#L292-L303

causes a crash when debugging this sort of situation:
```
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /home/user/pwndbg/pwndbg/gdblib/events.py:365 in invoke_event                                    │
│                                                                                                  │
│   362 │   │   # Run all the event handlers for this event from lowest to highest priority.       │
│   363 │   │   for prio in EventHandlerPriority:                                                  │
│   364 │   │   │   for func in handlers.get(prio, []):                                            │
│ ❱ 365 │   │   │   │   func()                                                                     │
│   366 │   except Exception as e:                                                                 │
│   367 │   │   import pwndbg.exception                                                            │
│   368                                                                                            │
│                                                                                                  │
│ /home/user/pwndbg/pwndbg/gdblib/events.py:315 in _dev_wrapper                                    │
│                                                                                                  │
│   312 │   │   │   │   │   f"{event_name} ({priority.name}) {fn.__module__}.{fn.__qualname__}\n   │
│   313 │   │   │   │   )                                                                          │
│   314 │   │   │                                                                                  │
│ ❱ 315 │   │   │   return fn(*a, **kw)                                                            │
│   316 │   │                                                                                      │
│   317 │   │   # Register this event handler with us.                                             │
│   318 │   │   registered[event_registry].setdefault(priority, []).append(_dev_wrapper)           │
│                                                                                                  │
│ /home/newt/pwndbg/pwndbg/aglib/kernel/symbol.py:296 in load_common_structs_on_load_linux         │
│                                                                                                  │
│   293 def load_common_structs_on_load_linux() -> None:                                           │
│   294 │   if pwndbg.aglib.qemu.is_qemu_kernel() and pwndbg.dbg.selected_inferior().is_linux():   │
│   295 │   │   try:                                                                               │
[...]

PR #3918, opened 2026-05-21
Fix crash on load_common_structs_on_load_linux

OBarronCS:
Fixes the crash detailed in #3911 (the event listener would cause pwndbg to crash when debugging a program run by qemu-system that invoked a crash deeper in the call stack)


pwndbg/aglib/kernel/symbol.py
k4lizen:
```suggestion
            # FIXME: NotImplementedError: Currently raised for non-linux OSes by stuff like `_paginginfo`. See #3911 .
```

k4lizen (approved):

pwndbg/aglib/kernel/symbol.py
OBarronCS: Good idea, thanks for adding this

PR #3919, opened 2026-05-21
Rename pwndbg.aglib.disasm.arch to pwndbg.aglib.disasm.assistant

OBarronCS:
I've been wanting to rename this file forever, as arch.py doesn't make much sense for this file.


OBarronCS:
While at it, I also made some minor changes (changing parameter names, adding more correct type annotations) to reduce red underlines in the IDE

k4lizen (approved):
nice

k4lizen:
i wonder if we should change to a more

smoke test -> merge -> full test -> potentially revert

workflow. it's kinda annoying that we now have 20 tests running, that I know we will have to re-run because the branch is not up-to-date. And I can't even update it now since https://github.com/pwndbg/pwndbg/pull/3918 is going to get merged in some minutes

k4lizen:
image

is this a crash in the python interpreter?

disconnect3d (approved):

Issue #2806, opened 2025-03-28
Telescope bug when incorrectly detecting Linux + debugging system without memory mappings


OBarronCS:
I came across a bug related to the interaction of various parts of the codebase that culminate in telescope printing incorrect values when debugging a system that doesn't expose memory mappings, such as `qemu-system`, and when we incorrectly detect the process as a Linux process.

Incorrect telescope:

![Image](https://github.com/user-attachments/assets/1523b652-f25e-433c-b3c9-a42681881533)

With the mapping explicitely explored with `vmmap_explore`

![Image](https://github.com/user-attachments/assets/c49e735e-9597-4f21-8237-f9c43d14a652)

# Reproduce the bug
For anyone wanting to reproduce - you may want to do this in an Ubuntu docker container (that exposes port 1234 so you can connect to the gdb, as a lot of packages are installed
```sh
git clone https://github.com/google/esh.git
cd esh
# this installs a bunch of packages
./manage -s
cd examples/emulation/armv7M-32
# this runs an Arm cortex M program, waits on port 1234 for gdb to connect
make debug

# run telescope / look at stack output, values are incorrect. Read memory directly with `x/wx addr` to see discrepancy
# Run "vmmap_explore 0" to explicitly add the memory mapping
```

# Bugs

1. `is_linux` returns True incorrectly.

I was running https://github.com/google/esh, which provides a quick way to run a simple program in various architectures with emulation. I ran it with Arm Cortex M (`examples/emulation/armv7M-32`).

In this case, `pwndbg.dbg.selected_inferior().is_linux()` returns True, even when this is not the case.

It internally calls `show osabi`, and GDB returns `GNU/Linux`.

2. Cascading errors in the case memory mappings do not exist:

`chain.get()` ignores vmmap in case we detect are in Linux, and dereferences pointers until a memory error is thrown.

https://github.com/pwndbg/pwndbg/blob/a82e152f94ad818b7720ed913e53529d8d6296e6/pwndbg/chain.py#L76-L88

The second-to-last pointer that chain.get returns gets passed to `pwndbg.enhance.enhance()` at the end of chain.format(), 

https://github.com/pwndbg/pwndbg/blob/a82e152f94ad818b7720ed913e53529d8d6296e6/pwndbg/chain.py#L169-L176

In chain.get we already dereferenced and know it's a valid pointer. In this call to enhance, we assume that enhance will dereference it.

[...]

PR #3921, opened 2026-05-24
Fix enhance errors in environment lacking paging information

OBarronCS:
Provides a fix for #2806 - see that issue for details on the bug.

The issue occurs when a memory address is dereferenced, but it does not exist in page in the vmmap list.

This basically implements the idea here: https://github.com/pwndbg/pwndbg/issues/2806#issuecomment-4051963537. While calling `enhance`, we pass it info indicating that we know this pointer can be dereferenced.

This fixes issue of telescoping printing pointers to point to themselves, instead of the memory they point to.

### DIFF
--- a/pwndbg/chain.py
+++ b/pwndbg/chain.py
@@ -184,8 +184,8 @@ def format(
 
         page = pwndbg.aglib.vmmap.find(pointer_to_enhance)
 
-        # Check if we dereferenced the last pointer, but it is not in the vmmap mappings
-        # If so, we have to create the page permissions
+        # If the pointer is not in the vmmap mappings, we create page permissions for it.
+        # We know it's a valid pointer, because we dereferenced it.
         if not page:
             mem_flags = pwndbg.aglib.vmmap_custom.get_memory_flags(pointer_to_enhance)
             if mem_flags is not None:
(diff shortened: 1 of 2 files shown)


OBarronCS:
Old buggy behavior:
image


New behavior:
image

Note that the system in this screenshot doesn't have pagefaults, so 0 is deemed a valid pointer in this context (which is true, real data is mapped there). The contents at `0x3ffff` contains a zero, hence the `◂— 0`. We have logic that already prevents infinite circular memory dereferencing.

OBarronCS:
Note a couple things about this:
- this maintains the previous behavior where we don't add the "newly found" page to the global vmmap list. This might be desirable, but would require some more thinking because we have settings like `auto-explore-pages` that control auto exploration in certain cases. We would also likely need to clear all caches everytime this happens, as the pages change.

- This makes telescope printing "just work" in environments without page faults and without vmmap info. Other cases to consider in future changes include a system having MMIO (which means we likely want to completely disable all pwndbg memory dereferences). A "has-mmio" setting might be appropriate.

- `is_pagefault_supported()` returns "True" basically always in GDB's compiled nowadays. Returning "false" in a system without vmmap`ings would mean that in the cases like in the screenshot above, there would be no recursive dereferencing, which is probably not ideal. A future fix may want to reconsider that logic, and adding a user setting to explicitly control it?

PR #3923, opened 2026-05-27
lint: enable all of flake8 return && ignore a subset of rules (instead of whitelisting some rules) 

c0rydoras:
#[]()1 out of X PRs regarding ruff config :innocent: 

From the [`RET` ruleset](https://docs.astral.sh/ruff/rules/#flake8-return-ret), the following two can't be "trivially"* applied/fixed:
- [`RET503`](https://docs.astral.sh/ruff/rules/implicit-return/)
- [`RET504`](https://docs.astral.sh/ruff/rules/unnecessary-assign/) 

instead of selecting a subset of `RET`, ignore only these 2

_*trivially as in "automated" without potentially degrading code quality/losing information_


k4lizen:
better PR names please, otherwise I have to edit them when merging the commit

PR #3924, opened 2026-05-28
:broom:: enable SIM ruleset

c0rydoras:

Start by enabling the `SIM` ruleset and ignoring individual rules, then (in seperate PRs) gradually enable more rules until stuff is nice

### DIFF
--- a/pwndbg/aglib/godbg.py
+++ b/pwndbg/aglib/godbg.py
@@ -879,10 +879,7 @@ class BasicType(Type):
         elif ty in ("int", "uint", "uintptr", "funcptr"):
             self.sz = word_size()
             self.algn = word_size()
-        elif ty == "string":
-            self.sz = word_size() * 2
-            self.algn = word_size()
-        elif ty in ("any", "interface"):
+        elif ty == "string" or ty in ("any", "interface"):
             self.sz = word_size() * 2
             self.algn = word_size()
         else:
--- a/pwndbg/aglib/godbg.py
+++ b/pwndbg/aglib/godbg.py
@@ -879,7 +879,7 @@ class BasicType(Type):
         elif ty in ("int", "uint", "uintptr", "funcptr"):
             self.sz = word_size()
             self.algn = word_size()
-        elif ty == "string" or ty in ("any", "interface"):
+        elif ty in ("string", "any", "interface"):
             self.sz = word_size() * 2
             self.algn = word_size()
         else:
--- a/pwndbg/aglib/heap/ptmalloc.py
+++ b/pwndbg/aglib/heap/ptmalloc.py
@@ -1887,7 +1887,7 @@ class HeuristicHeap(
 
     def _get_heap_page(self) -> pwndbg.lib.memory.Page | None:
         """Get the [heap] memory page."""
-        return next((p for p in pwndbg.aglib.vmmap.get() if p.objfile == "[heap]"), None)
+        return next((p for p in pwndbg.aglib.vmmap.get() if p.is_heap), None)
 
     def _get_heap_range(self) -> pwndbg.lib.memory.Page | range:
         """Get the heap start & end"""
--- a/pwndbg/aglib/heap/ptmalloc.py
+++ b/pwndbg/aglib/heap/ptmalloc.py
@@ -1105,9 +1105,9 @@ class GlibcMemoryAllocator(pwndbg.aglib.heap.heap.MemoryAllocator, Generic[TheTy
         if not self.has_tcache():
             return None
         mp = self.mp
-        if "tcache_small_bins" in mp.type.keys():
+        if "tcache_small_bins" in mp.type:
             return int(mp["tcache_small_bins"])
-        if "tcache_bins" in mp.type.keys():
+        if "tcache_bins" in mp.type:
             return int(mp["tcache_bins"])
         return None
 
@@ -1199,7 +1199,7 @@ class GlibcMemoryAllocator(pwndbg.aglib.heap.heap.MemoryAllocator, Generic[TheTy
         val = self.malloc_chunk
         if val is None:
             return None
-        chunk_keys = [renames[key] if key in renames else key for key in val.keys()]
+        chunk_keys = [renames.get(key, key) for key in val]
         try:
             return chunk_keys.index(key) * pwndbg.aglib.arch.ptrsize
         except Exception:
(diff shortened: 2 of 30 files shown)


pwndbg/aglib/onegadget.py
```diff
@@ -583,11 +583,13 @@ def check_gadget(
...
-    elif is_valid_gadget == UNSAT and show_unsat:
-        print(output_msg)
-    elif is_valid_gadget == UNKNOWN and not no_unknown:
+    if (
+        is_valid_gadget == SAT
+        or is_valid_gadget == UNSAT
+        and show_unsat
+        or is_valid_gadget == UNKNOWN
+        and not no_unknown
+    ):
```
c0rydoras: 🤔

pwndbg/enhance.py
```diff
@@ -131,11 +131,7 @@ def enhance(
         instr = None
 
     # If it's on the stack, don't display it as code in a chain.
-    if instr and "[stack" in page.objfile:
-        retval = [intval, szval]
-
-    # If it's RWX but a small value, don't display it as code in a chain.
```
k4lizen, line 137: deleting a comment like this is not good

these two if-statements have different semantics even though they reach the same code, so this refactor makes this code harder to read.

also
> instr and "[stack" in page.objfile or instr and rwx and intval0 < 0x1000

you need to be aware of operator precedence to parse this properly, which is extra mental overhead imo

k4lizen: i think its fine to merge the two ifs into one if it can be done in a way which
1. doesn't delete the comments
2. makes it obvious which part of the condition is for which logic

also I would even report this lint to ruff as a bug

k4lizen (changes requested):

pwndbg/enhance.py
k4lizen: (this same issue is present in a couple of places)

c0rydoras: I thought I got rid of everything that would delete comments, looks like I missed some :sweat_smile:

c0rydoras: all cleaned up now :)

pwndbg/aglib/heap/ptmalloc.py
```diff
@@ -1585,7 +1586,8 @@ def main_arena(self) -> Arena | None:
     def has_tcache(self) -> bool:
         # tcache_bins was renamed to tcache_small_bins in GLIBC 2.42
         return self.mp is not None and any(
-            x in self.mp.type.keys() for x in ["tcache_bins", "tcache_small_bins"]
+            x in self.mp.type.keys()  # noqa: SIM118 (mp is not a dict)
```
OBarronCS, line 1589: Is this a Ruff bug? Would the auto-converted code just not work correctly, since `self.mp` is not a dict?

c0rydoras: iirc it's an unsafe fix

for dicts, iterating over it gives you its keys, the same is not true for `CStruct2GDB` (which to be fair, we could add/make behave the same way)

pwndbg/aglib/kernel/vmmap.py
```diff
@@ -208,9 +208,7 @@ def kernel_vmmap_via_page_tables() -> tuple[Page, ...]:
     ptrsize: int = pwndbg.aglib.arch.ptrsize
     if arch == "aarch64":
         arch_backend = PT_Aarch64_Backend(machine_backend)
-    elif arch == "i386":
-        arch_backend = PT_x86_64_Backend(machine_backend)
-    elif arch == "x86-64":
+    elif arch == "i386" or arch == "x86-64":
```
disconnect3d: `arch == "x86-64":` should probably be first checked here (tho it probably doesnt matter)

pwndbg/aglib/heap/ptmalloc.py
```diff
@@ -1585,7 +1586,8 @@ def main_arena(self) -> Arena | None:
...
+            for x in ["tcache_bins", "tcache_small_bins"]
```
disconnect3d:
```suggestion
            for x in ("tcache_bins", "tcache_small_bins")
```

pwndbg/aglib/kernel/vmmap.py
c0rydoras: not rlly, but might as well fix it

pwndbg/aglib/godbg.py
```diff
@@ -879,10 +879,7 @@ def __post_init__(self) -> None:
         elif ty in ("int", "uint", "uintptr", "funcptr"):
             self.sz = word_size()
             self.algn = word_size()
-        elif ty == "string":
-            self.sz = word_size() * 2
-            self.algn = word_size()
-        elif ty in ("any", "interface"):
+        elif ty == "string" or ty in ("any", "interface"):
```
disconnect3d:
```suggestion
        elif ty in ("string", "any", "interface"):
```

Like this?

pwndbg/commands/binder.py
```diff
@@ -114,7 +114,7 @@ def _format_field(
             if t.code == pwndbg.dbg_mod.TypeCode.INT:
                 value = int(value)
             elif t.code == pwndbg.dbg_mod.TypeCode.BOOL:
-                value = True if int(value) else False
+                value = bool(int(value))
```
disconnect3d, line 117: Can this be just `bool(value)`?

pwndbg/commands/context.py
```diff
@@ -1383,7 +1383,7 @@ def context_disasm(
         lambda: pwndbg.aglib.nearpc.nearpc(
             back_lines=additional_disasm_lines // 2,
             total_lines=additional_disasm_lines + 1,
-            emulate=bool(not pwndbg.config.emulate == "off"),
+            emulate=pwndbg.config.emulate != "off",
```
disconnect3d, line 1386: oh wow the prev version was so ugly

pwndbg/commands/binder.py
c0rydoras: assuming `value` is a string `"0"`, would be true, so don't think so

pwndbg/aglib/godbg.py
c0rydoras: there's some rule that does it, but idk which one (will fix em manually rq)

pwndbg/aglib/kernel/vmmap.py
c0rydoras: fixed

pwndbg/aglib/godbg.py
c0rydoras: fixed

k4lizen (approved):

PR #3940, opened 2026-06-03
slab: update field name of struct kmem_cache

navytuner:
Add a new lookup path of `struct kmem_cache` due to the update of field name and its data structure.
Due to this kernel patch[1], slab commands failed like below image.

Reference
[1] https://lore.kernel.org/all/20260311-b4-slab-memoryless-barns-v1-1-70ab850be4ce@kernel.org/

Screenshot 2026-06-03 at 1 05 17 AM

### DIFF
--- a/pwndbg/aglib/kernel/slab.py
+++ b/pwndbg/aglib/kernel/slab.py
@@ -207,7 +207,10 @@ class SlabCache:
     def node_caches(self) -> Generator[NodeCache, None, None]:
         """returns node caches for all NUMA nodes"""
         for node in range(kernel.num_numa_nodes()):
-            yield NodeCache(self._slab_cache["node"][node], self, node)
+            if self._slab_cache.dereference().type.has_field("node"):
+                yield NodeCache(self._slab_cache["node"][node], self, node)
+            else:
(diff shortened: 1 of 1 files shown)


pwndbg/aglib/kernel/slab.py
```diff
@@ -207,7 +207,10 @@ def cpu_caches(self) -> Generator[CpuCache, None, None]:
     def node_caches(self) -> Generator[NodeCache, None, None]:
         """returns node caches for all NUMA nodes"""
         for node in range(kernel.num_numa_nodes()):
-            yield NodeCache(self._slab_cache["node"][node], self, node)
+            if self._slab_cache.dereference().type.has_field("node"):
```
jxuanli: should we check the kernel version instead

jxuanli (approved):

jxuanli:
also this wouldnt work for the cases where it doesnt have typeinfo but i can work on that

pwndbg/aglib/kernel/slab.py
navytuner: That sounds good!

PR #3942, opened 2026-06-03
chore: cleanup pwndbg.aglib.onegadget

c0rydoras:
this can still be improved, a lot, but it's a start

_PR'd on top of #3941_

PR #3943, opened 2026-06-03
chore(aglib): cleanup onegadget & next & elf

c0rydoras:
PR'd on top of #3942

### DIFF
--- a/pwndbg/aglib/next.py
+++ b/pwndbg/aglib/next.py
@@ -125,14 +125,16 @@ async def break_next_branch(
     """
     ins = next_branch(address, including_current=including_current)
 
+    if not ins:
+        return None
+
     inf = pwndbg.dbg.selected_inferior()
-    if ins:
-        # If the branch we found was not at the current program counter, we should step to it.
-        # Otherwise, return the current instruction.
-        if ins.address != pwndbg.aglib.regs.pc:
-            with inf.break_at(BreakpointLocation(ins.address), internal=True) as bp:
-                await ec.cont(bp)
-        return ins
+    # If the branch we found was not at the current program counter, we should step to it.
+    # Otherwise, return the current instruction.
+    if ins.address != pwndbg.aglib.regs.pc:
+        with inf.break_at(BreakpointLocation(ins.address), internal=True) as bp:
+            await ec.cont(bp)
+    return ins
 
 
 async def break_next_interrupt(
@@ -145,13 +147,13 @@ async def break_next_interrupt(
     If no such interrupt exists or a jump is in the way, return None.
     """
     ins = next_int(address, honor_current_branch=honor_current_branch)
-    proc = pwndbg.dbg.selected_inferior()
-    if ins:
-        with proc.break_at(BreakpointLocation(ins.address), internal=True) as bp:
-            await ec.cont(bp)
-        return ins
+    if not ins:
+        return None
 
-    return None
+    inf = pwndbg.dbg.selected_inferior()
+    with inf.break_at(BreakpointLocation(ins.address), internal=True) as bp:
+        await ec.cont(bp)
+    return ins
 
 
 async def break_next_interrupt_filtered(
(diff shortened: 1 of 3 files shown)


pwndbg/aglib/next.py
```diff
@@ -145,13 +147,13 @@ async def break_next_interrupt(
...
-    proc = pwndbg.dbg.selected_inferior()
-    if ins:
-        with proc.break_at(BreakpointLocation(ins.address), internal=True) as bp:
-            await ec.cont(bp)
-        return ins
+    if not ins:
+        return None
 
-    return None
+    proc = pwndbg.dbg.selected_inferior()
```
k4lizen: i'd change the var name to `inf`

pwndbg/aglib/next.py
```diff
@@ -168,17 +170,17 @@ async def break_next_interrupt_filtered(
...
+    with proc.break_at(BreakpointLocation(ins.address), internal=True) as bp:
+        await ec.cont(bp)
+    try:
+        if predicate():
+            return ins
+    except Exception:
```
k4lizen, line 182: this kinda bad but idk how to solve it

k4lizen (approved):

pwndbg/aglib/onegadget.py
```diff
@@ -42,12 +42,12 @@
...
-    "(u32)": lambda x: ctypes.c_uint32(x).value,
-    "(s32)": lambda x: ctypes.c_int32(x).value,
-    "(u64)": lambda x: ctypes.c_uint64(x).value,
-    "(s64)": lambda x: ctypes.c_int64(x).value,
+    "(u16)": ctypes.c_uint16,
+    "(s16)": ctypes.c_int16,
+    "(u32)": ctypes.c_uint32,
+    "(s32)": ctypes.c_int32,
+    "(u64)": ctypes.c_uint64,
+    "(s64)": ctypes.c_int64,
```
disconnect3d, line 50: Is this ok? Since this is not doing the same. The previous one would return an `int` and here this would return a `ctypes` wrapper object.

On the long term this should probably be a function with `if: .. elif ...: ctypes_obj = ctypes.c_*` and returning a `ctypes_obj.value`

pwndbg/aglib/onegadget.py
```diff
@@ -65,16 +65,16 @@ def __str__(self) -> str:
         return self.name
 
     def __and__(self, other: CheckSatResult) -> CheckSatResult:
-        if self == CheckSatResult.UNSAT or other == CheckSatResult.UNSAT:
+        if CheckSatResult.UNSAT in {self, other}:
             return CheckSatResult.UNSAT
-        if self == CheckSatResult.UNKNOWN or other == CheckSatResult.UNKNOWN:
+        if CheckSatResult.UNKNOWN in {self, other}:
```
disconnect3d: This does the check against a set object, so needs to hash etc, whereas the previous would short circuit if first condition is true etc.

I'm okay with both versions though

k4lizen: i guess `in (self, other)` is more canonical?

c0rydoras: It's more readable (imo)

_and prematurely optimizing is kinda pointless, if it even is faster/slower (idk)_

pwndbg/aglib/onegadget.py
c0rydoras: this is okay because in the only place it's used we now do `.value` so it's effectively the same but without lambdas :)

pwndbg/aglib/next.py
c0rydoras: yes

pwndbg/aglib/next.py
c0rydoras: done (in the other places as well :))

pwndbg/aglib/onegadget.py
c0rydoras: yes, changed it to tuples

PR #3944, opened 2026-06-03
chore: enable SIM105

c0rydoras:
+ [x ] I read the contributing documentation.

### DIFF
--- a/pwndbg/commands/ai.py
+++ b/pwndbg/commands/ai.py
@@ -12,6 +12,7 @@ import json
 import os
 import pprint
 import re
+from contextlib import suppress
 
 import gdb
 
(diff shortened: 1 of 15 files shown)


pwndbg/commands/ai.py
```diff
@@ -205,10 +203,8 @@ def build_context_prompt_body():
     local_vars = None
     ## and source information, if available
     source = ""
-    try:
+    with contextlib.suppress(gdb.error):
```
k4lizen, line 206: why not `suppress(gdb.error)`

pwndbg/commands/killthreads.py
```diff
@@ -57,7 +57,7 @@ def killthreads(thread_ids: list[int] | None = None, all: bool = False) -> None:
                     return
         for thread_id in thread_ids:
             gdb.execute(f"thread {thread_id}", to_string=True)
-            try:
+            try:  # noqa: SIM105
                 gdb.execute("call (void) pthread_exit(0)", to_string=True)
             except gdb.error:
                 # gdb will throw an error, because the thread dies during the call, which is expected
```
k4lizen, line 63: does this comment get deleted by this lint?

pwndbg/commands/ai.py
k4lizen: i think either are fine, but better be consistent

k4lizen (approved):

pwndbg/commands/ai.py
c0rydoras: good question actually

pwndbg/commands/killthreads.py
c0rydoras: it doesn't (it's also an explicitly unsafe fix)

but it looks kinda weird otherwise (like if you rewrite the comment that could be changed...)

pwndbg/commands/ai.py
c0rydoras: changed it to just `suppress` in files where it's used more than once

PR #3947, opened 2026-06-04
Reading `/proc/pid/mem` by default when scanning pagetables

jxuanli:
As discussed, reading `/proc/pid/mem` is faster. When benchmarked against `gdb-pt-dump` with optimized (compared to the one provided in the `gdb-pt-dump` repo`) reads, the current implementation is still up to 10x faster.

### DIFF
--- a/pwndbg/aglib/kernel/paging.py
+++ b/pwndbg/aglib/kernel/paging.py
@@ -359,10 +359,10 @@ class ArchPagingInfo:
             for level in levels:
                 if level.phys is None or level.level is None:
                     continue
-                level.virt = level.phys + self.physmap - self.phys_offset
+                level.virt = level.phys + self.physmap - self.ram_phys_start
                 level.name = self.pagetable_level_names[level.level]
             if result.phys is not None:
-                result.virt = result.phys + self.physmap - self.phys_offset
+                result.virt = result.phys + self.physmap - self.ram_phys_start
         except Exception:
             pass
         finally:  # so that the PhyMemMode value is always restored
@@ -390,7 +390,7 @@ class ArchPagingInfo:
         raise NotImplementedError()
 
     @property
-    def phys_offset(self) -> int:
+    def ram_phys_start(self) -> int:
         return 0
 
     @property
@@ -724,7 +724,7 @@ class Aarch64PagingInfo(ArchPagingInfo):
             self.VMEMMAP_START = pwndbg.aglib.arch.unsigned(-0x40000000 - self.VMEMMAP_SIZE)
 
         # obtained through debugging -- kaslr offset of physmap determines the offset of vmemmap
-        vmemmap_kaslr = (self.physmap - self.PAGE_OFFSET - self.phys_offset) >> vmemmap_shift
+        vmemmap_kaslr = (self.physmap - self.PAGE_OFFSET - self.ram_phys_start) >> vmemmap_shift
         return self.VMEMMAP_START + vmemmap_kaslr
 
     @property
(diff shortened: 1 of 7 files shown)


pwndbg/aglib/kernel/paging.py
```diff
@@ -64,23 +65,24 @@ def __init__(self, pi: ArchPagingInfo) -> None:
         # for scanning
         self.pagesz = pi.page_size
         self.ptrsize = pwndbg.aglib.arch.ptrsize
-        self.inf = pwndbg.dbg.selected_inferior()
         self.fmt = "<" + ("Q" if self.ptrsize == 8 else "I") * (self.pagesz // self.ptrsize)
         self.cache: dict[tuple[int, int], list[tuple[int, int, int]]] = {}
         self.entry_cache: dict[int, tuple[int]] = {}
         self.arch = pwndbg.aglib.arch.name
+        # the default is slightly slower but does not require setting yama.ptrace_scope
+        self.read: Callable[[int, int], bytearray] = pwndbg.dbg.selected_inferior().read_memory
```
k4lizen, line 73: this is a pretty clean redesign

k4lizen (changes requested):
can you document the qemu contiguousness assumption?

k4lizen:
image

Here the memory is not contiguous. I am running qemu as

```bash
qemu-system-x86_64 \
    -m 4G \
    -smp 4 \
    -object memory-backend-ram,id=mem0,size=2G \
    -object memory-backend-ram,id=mem1,size=2G \
    -numa node,nodeid=0,memdev=mem0,cpus=0-1 \
    -numa node,nodeid=1,memdev=mem1,cpus=2-3 \
    -kernel ~/opt/linux-stable/arch/x86/boot/bzImage \
    -initrd debugfs.cpio \
    -append "console=ttyS0 nokaslr loglevel=3 oops=panic panic=1 nopti quiet" \
    -cpu qemu64,+smep,+smap \
    -monitor /dev/null \
    -nographic \
    -no-reboot \
    -nic user,model=virtio\
    -gdb tcp::1234
```

(also side note, why is vmmap dying?)

jxuanli:
@k4lizen this now works with multiple ram backends

pwndbg/aglib/qemu.py
```diff
@@ -102,26 +102,100 @@ def exec_file_supported() -> bool:
     return b"qXfer:exec-file:read" in response
 
 
+class QemuPhysAddressNotResolvedError(Exception):
+    """
+    We tried to recover (i.e. look up in the debugger and find through heuristics) the
```
k4lizen: stale docstring xd

k4lizen (approved):
OKE OKE, nowww we're cooking

k4lizen:
just as a sanity check:
+ did you confirm that running mtree does not offset the perf benefit of reading from /proc/pid
+ can the output of mtree change during the runtime of the kernel, and do we properly invalidate it if it does?

patryk4815:
how can we detected when `/proc/pid/mem` is available using `target remote ip:port`?

jxuanli:
@k4lizen I don't think it remaps to a different address after the remote port becomes available (I only empirically verified that). I also benchmarked, its about 2x faster than direct read

pwndbg/aglib/qemu.py
```diff
@@ -85,3 +100,211 @@ def exec_file_supported() -> bool:
...
+        self.file = os.open(f"/proc/{self.pid}/mem", os.O_RDONLY)
+        self.mtree = QemuMtree()
+
+    def __del__(self) -> None:
+        if self.file is not None:
+            with contextlib.suppress(OSError):
+                os.close(self.file)
+
+    @staticmethod
+    def search_pids_for_file(pids: list[str], filename: str) -> str | None:
```
jxuanli, line 199: @patryk4815 this is what it has always been done (since we were using pt-dump)

PR #3948, opened 2026-06-04
Loading kernel common structs only when necessary

### LINKED ISSUE #3820: pwndbg slow to load kernel
When connecting to a kernel with pwndbg, I feel it is perceptably slower then when connecting to the same kernel with gef.

Since this is a very common operation, we should look into optimizing it to be faster.

jxuanli:
Should partially mitigate #3911
Might fix #3820

### DIFF
--- a/pwndbg/aglib/kernel/__init__.py
+++ b/pwndbg/aglib/kernel/__init__.py
@@ -107,11 +107,6 @@ def requires_debug_info(default: D = None) -> Callable[[Callable[P, T]], Callabl
     return decorator
 
 
-# Set by pwndbg.aglib.kernel.symbol.load_common_structs_on_load_linux() when page typeinfo
-# recovery fails.
-page_typeinfo_recovery_failure: None | TypeNotRecoveredError = None
-
-
 def typeinfo_recovery(
     name: str, requires_kversion: bool = False, requires_kbase: bool = False
 ) -> Callable[[Callable[P, str]], Callable[P, None]]:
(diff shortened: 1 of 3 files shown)


pwndbg/aglib/kernel/__init__.py
```diff
@@ -107,11 +107,6 @@ def func(*args: P.args, **kwargs: P.kwargs) -> T | D:
...
 
 
-# Set by pwndbg.aglib.kernel.symbol.load_common_structs_on_load_linux() when page typeinfo
-# recovery fails.
-page_typeinfo_recovery_failure: None | TypeNotRecoveredError = None
-
-
 def typeinfo_recovery(
     name: str, requires_kversion: bool = False, requires_kbase: bool = False
 ) -> Callable[[Callable[P, str]], Callable[P, None]]:
```
k4lizen, line 112:
```suggestion
) -> Callable[[Callable[P, str]], Callable[P, None]]:
    """
    Attached to functions which return the C source-code of the `name` type (usually struct).
    
    Compiles and adds the debug info for the struct into the debugger, from then on accessible via APIs
    like `pwndbg.aglib.memory.get_typed_pointer("struct list_head", db_list)`.
    """
```

k4lizen (approved):
makes sense!

OBarronCS:
Nice. I confirmed locally this avoids the crash described in https://github.com/pwndbg/pwndbg/issues/3911

pwndbg/aglib/kernel/__init__.py
```diff
@@ -126,6 +121,9 @@ def func(*args: P.args, **kwargs: P.kwargs) -> None:
                 return
             if pwndbg.aglib.typeinfo.lookup_types(name) is not None:
                 return
+            recover_common_types_func = pwndbg.aglib.kernel.symbol.recover_page_typeinfo
+            if f.__name__ != recover_common_types_func.__name__:
+                recover_common_types_func()
```
OBarronCS, line 133: I don't fully understand the context of when `typeinfo_recovery` is called. When is this `if` branch taken?

k4lizen: (almost) always, it is just there to prevent infinite recursion, since recover_page_typeinfo is annotated with this

OBarronCS (approved):

PR #3950, opened 2026-06-05
kernel: add function percpu

navytuner:
Add `$percpu(addr, cpu)` function for better kernel debugging.

It performs exactly the same as `pwndbg.aglib.kernel.per_cpu(addr, cpu)`. It allows user to convert `__per_cpu` variables to real virtual address, which makes it possible to dereference the per-cpu pointers.

Screenshot 2026-06-05 at 4 47 06 PM

### DIFF
--- a/pwndbg/gdblib/functions.py
+++ b/pwndbg/gdblib/functions.py
@@ -406,6 +406,7 @@ def p2v(paddr: gdb.Value) -> int:
 
     return vaddr
 
+
 @GdbFunction(only_when_running=True)
 def percpu(addr: gdb.Value, cpu: gdb.Value = gdb.Value(-1)) -> gdb.Value:
     """
--- a/pwndbg/gdblib/functions.py
+++ b/pwndbg/gdblib/functions.py
@@ -410,7 +410,7 @@ def p2v(paddr: gdb.Value) -> int:
 @GdbFunction(only_when_running=True)
 def percpu(addr: gdb.Value, cpu: gdb.Value = gdb.Value(-1)) -> gdb.Value:
     """
-    Resolve a Linux kernel per-cpu pointer to its absolute address.
+    Resolve a Linux kernel per-cpu pointer to its absolute virtual address.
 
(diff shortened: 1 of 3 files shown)


pwndbg/gdblib/functions.py
```diff
@@ -405,3 +405,23 @@ def p2v(paddr: gdb.Value) -> int:
         raise gdb.GdbError("Physical to virtual address failed, invalid physical address?")
 
     return vaddr
+
+
+@GdbFunction(only_when_running=True)
+def percpu(addr: gdb.Value, cpu: gdb.Value = gdb.Value(-1)) -> gdb.Value:
+    """
+    Resolve a per-cpu pointer to its address for the given CPU.
```
disconnect3d: Can we elaborate a bit more?

Should be: `Linux kernel per-cpu pointer` + likely link to kernel sources about it, or, some explanation "what is it".

Also, may want to add a Linux kernel version for which this applies/work with/was tested with -- since this goes to our documentation, or, may be read by users when they do 'help'.

navytuner: Sure!

pwndbg/gdblib/functions.py
```diff
@@ -405,3 +405,44 @@ def p2v(paddr: gdb.Value) -> int:
         raise gdb.GdbError("Physical to virtual address failed, invalid physical address?")
 
     return vaddr
+
+
+@GdbFunction(only_when_running=True)
+def percpu(addr: gdb.Value, cpu: gdb.Value = gdb.Value(-1)) -> gdb.Value:
+    """
+    Resolve a Linux kernel per-cpu pointer to its absolute address.
```
k4lizen:
```suggestion
    Resolve a Linux kernel per-cpu pointer to its absolute virtual address.
```

k4lizen (approved):
LGTM, thanks

Image

You'll need to update this so docs CI doesn't fail

PR #3953, opened 2026-06-05
Handle ljmp instruction target resolution for i8086

OBarronCS:
Handle target resolution of `ljmp` instruction, which uses segment selector to augment the target PC. This is relevant for 16-bit real mode. This will now correctly work for ljmps that are within 16-bit real mode. More complex changes would be necessary to capture an `ljmp` that changes from 16-real mode to 32-bit protected mode that will go in a different PR.

Before:
image

After:
image


k4lizen (approved):
wild, nice!

Issue #2713, opened 2025-01-28
Cache issue with manually changing architecture


OBarronCS:
When manually setting the architecture during debugging, such as with `set architecture i8086` in GDB, sometimes you must step the program up to twice with `si` to see the architecture change take effect in Pwndbg's output. This is caused by the behavior of the `cache_until` decorator.

Here's an example when this is seen:

Assume that GDB detects the wrong architecture with the wrong register set/behavior (like the difference between i8086 and i386). You want to force it to the correct one to get the correct register set. When you call `set architecture i8086`, and do `ctx`, the output does not reflect the new architecture. You must `step` twice to see everything catch up to the new arch.

This is because inside many functions, the `arch.name` is used to fetch values, such as the current register set. These functions are decorated with `cache_until("stop")`, meaning if they are called multiple times while the process is stopped, they will return the cached return value. When you `set architecture`, and then do something like `ctx` to reprint the context, the functions will still return the cached register values, not taking into account the new architecture 

Further compounding this issue is that many key functions are invoked via the debuggers `event handlers`. 

Here is the function that updates the pwndbg arch value.
https://github.com/pwndbg/pwndbg/blob/1ce7ee36e8e3c05aa81308ed357da922c86d915d/pwndbg/aglib/arch.py#L71-L77

And here is the GDB hook that calls this function whenever the program stops:

https://github.com/pwndbg/pwndbg/blob/1ce7ee36e8e3c05aa81308ed357da922c86d915d/pwndbg/gdblib/hooks.py#L40-L44

If the arch update hook is not called first, then other hooks that are called before it and read the pwndbg arch.name will get the old arch name. This is why you would need to step the process one more time for the arch.name to propagate everywhere.

Possible solutions:
- Set a priority on the event handlers to force the arch update handler to run first.
- When the user manually sets architecture (can we detect this?), then force clear function caches. Could we clear all caches? Or just hand-pick the relevant ones?

disconnect3d:
Hah, I've fallen on this exact issue like 2 days ago but didn't have time to report it!

> Set a priority on the event handlers to force the arch update handler to run first.

We should definetely do that. But also do we trigger 'stop' handler on `set arch x`? I think not?
Though: prompt hook would be triggered I believe... so this could maybe work to re-trigger this there?

Though I wonder if that doesn't make things slower unnecessarily...


> When the user manually sets architecture (can we detect this?), then force clear function caches. Could we clear all caches? Or just hand-pick the relevant ones?

[...]

PR #3954, opened 2026-06-05
Listen for 'set architecture' command to reset caches

OBarronCS:
Listens for the "set architecture" command, and resets pwndbg caches and polls for the debugger for the architecture to immediately align with GDB, instead of having to "step" once.

This uses a GDB hook  `define set hookpost-architecture` which runs the command AFTER `set architecture` is run. There is no equivalent in the Python API.

Original bug described here: #2713

This bug came up again due to work in #3911, where user has to manually do `set architecture` to tell GDB it was is i8086, not i386, which causes the cache bug

### DIFF
--- a/pwndbg/dbg_mod/gdb/__init__.py
+++ b/pwndbg/dbg_mod/gdb/__init__.py
@@ -1719,25 +1719,30 @@ class GDB(pwndbg.dbg_mod.Debugger):
 
         prompt.set_prompt()
 
-        pre_commands = """
-        set auto-load safe-path /
-        set confirm off
-        set verbose off
-        set pagination off
-        set history save on
-        set follow-fork-mode child
-        set backtrace past-main on
-        set step-mode on
-        set print pretty on
-        set output-radix 16
-        handle SIGALRM nostop print nopass
-        handle SIGBUS  stop   print nopass
-        handle SIGPIPE nostop print nopass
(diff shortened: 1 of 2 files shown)


pwndbg/dbg_mod/gdb/__init__.py
```diff
@@ -1719,25 +1719,30 @@ def setup(self):
...
+            "set step-mode on",
+            "set print pretty on",
+            "set output-radix 16",
+            "handle SIGALRM nostop print nopass",
+            "handle SIGBUS  stop   print nopass",
+            "handle SIGPIPE nostop print nopass",
+            "handle SIGSEGV stop   print nopass",
+            """\
+define set hookpost-architecture
+python pwndbg.aglib.arch_mod.force_reload_architecture()
```
OBarronCS, line 1739: Just making sure: this is safe, right? I've been using this (in my own .gdbinit) for a while now (using pwndbg from source), and it works. But there are many pwndbg installation methods: is the pwndbg module always in the global scope and accessible via gdb "python" command (no matter the pwndbg installation method?)

OBarronCS:
Closing this for now: this is handling an edge case (pwndbg misdetects architecture, usually due to buggy behavior of underlying debugger, forcing user to override) that should really be handled within pwndbg itself. If this is needed in a specific case, it can be added to .gdbinit.

PR #3956, opened 2026-06-07
Fix --nx still loading ~/.gdbinit when pwndbg is sourced from a .gdbinit

### LINKED ISSUE #3896: The `--nx` flag in pwndbg still loads `~/.gdbinit` when pwndbg is sourced from a different `.gdbinit` file
### Description

This issue is an extension from Issue #3881. So make sure to read that discussion before reading this.

I realized that I didn't properly communicate my issue when I reported it earlier, with my previous issue I wasn't complaining about the double-load error in pwndbg rather I meant that even after using the `--nx` flag, pwndbg still loads the `~/.gdbinit` from the home directory which should not be the case. 

And even though it looks like a GDB issue on surface, it actually originates from the loader in pwndbg.

You can clearly see it here in the screenshot posted by @disconnect3d:

Image

## Proof that this is a pwndbg issue

Create two `.gdbinit` files: `~/.gdbinit` and `~/.gdbinit.local`

`~/.gdbinit`:

```
echo RAN HOME_INIT\n
```

`~/.gdbinit.local`:

```
echo RAN LOCAL_INIT\n
source /tmp/pwndbg/gdbinit.py
```

Now try this command to load gdb: `gdb --nx -x ~/.gdbinit.local --quiet`

## Output

Image

Now remove the path to source pwndbg to load vanilla GDB:

`~/.gdbinit`:

```
echo RAN HOME_INIT\n
```

`~/.gdbinit.local`:

```
echo RAN LOCAL_INIT\n
```

## Output from Vanilla GDB

Image

This clearly shows that Vanilla GDB works as intended while pwndbg loads both `.gdbinit` files somewhere during the loading process.

## Root cause

After manually looking through the loading process I observed that:

- `~/.gdbinit.local` sources `pwndbg/gdbinit.py`
- `pwndbg/gdbinit.py` builds src_root and calls `pwndbginit.gdbinit.main_try()` where pwndbg is imported

```
    check_doubleload()        <------ this is to make sure that pwndbg does not get imported twice, but right now it hasn't even been imported once.
    verify_venv()

    # Force UTF-8 encoding (to_string=True to skip output appearing to the user)
    try:
        gdb.execute("set target-wide-charset UTF-8", to_string=True)
        gdb.execute("set charset UTF-8", to_string=True)
    except gdb.error as e:
        print(f"Warning: Cannot set gdb charset: '{e}'")

    import pwndbg  # noqa: F811        <-------- this is where pwndbg gets imported from sourced path
    import pwndbg.dbg_mod.gdb

    # Mark that pwndbg was loaded from `pwndbg` binary (for double-load detection)        <-------- this assumption is the root cause of our problem in my opinion, more about this later.
    setattr(pwndbg, "_is_loaded_from_pwndbg", True)

    pwndbg.dbg = pwndbg.dbg_mod.gdb.GDB()                    <-------- this path later calls `self._load_gdbinit()` which loads `~/.gdbinit` from the home directory. 
    pwndbg.dbg.setup()
```

But `_load_gdbinit()` has a check in place to make sure that we don't load `.gdbinit` files in case `--nx` or `--nh` flag is passed. 

```
    def _load_gdbinit(self):
        # Emulate how `gdb` loads `.gdbinit` files (home and local)
        disable_any, disable_home = self._disable_gdbinit_loading()
        if disable_any:
            return
``` 

The check in `_disable_gdbinit_loading`:

```
 return disable_any_gdbinit >= 2, disable_home_gdbinit >= 1
```

But this check doesn't work in our case because of one assumption which is present in `pwndbg/pwndbginit/gdbinit.py`:

```
    # Mark that pwndbg was loaded from `pwndbg` binary (for double-load detection)
    setattr(pwndbg, "_is_loaded_from_pwndbg", True)
```

Here it is assumed that the imported pwndbg came from the portable build binary, but in my case pwndbg was sourced from a local path.

This is important because this assumption causes: `disable_any_gdbinit >= 2` instead of `disable_any_gdbinit >= 1`. Now we have to pass two `--nx` flags to stop `~/.gdbinit` from loading, because it is assumed that one of the `--nx` flags comes when we launch pwndbg throught the portable binary as shown below:

`pwndbg/pwndbginit/pwndbg_gdb.py`:
```
def main():
    prepend_venv_bin_to_path()

    gdb_argv = [
        sys.argv[0],
        "-q",
        "-nx",
        "-iex",
        "py import pwndbginit.gdbinit; pwndbginit.gdbinit.main_try()",
        *sys.argv[1:],
    ]
    sys.argv = gdb_argv
```

This is why this issue can be reproduced by using the `--nx` flag but not the `--nh` flag. I know that using `--nh` is a good workaround but I feel that there should be consistency across both the commands.

Also, it's important to note that if you have the portable build of pwndbg, your pwndbg will try to load both the `.gdbinit` files but ignore the `~/.gdbinit` as shown in the very first screenshot. This is because it is caught earlier in `pwndbg/gdbinit.py`:

```
    if "pwndbg" in sys.modules and hasattr(sys.modules["pwndbg"], "_is_loaded_from_pwndbg"):
        print(
            "\033[90m~/.gdbinit: Skipped loading Pwndbg from `source path/gdbinit.py` - already loaded.\033[0m",
            flush=True,
        )
        return
```

## Likely fix

I think there should be a distinction about where pwndbg is loaded from either the portable build or a sourced path from a local checkout. Once we have this we can then patch the `_disable_gdbinit_loading()` function to handle the `--nx` flag count based on this information.

Please do let me know that if there was any mistake in my analysis and feel free to correct me🙏

### My setup

GDB: GNU gdb (GDB) 17.1
OS: Arch Linux (CachyOS)
Python: 3.10.17

clmoon2:
when pwndbg is sourced from a user's own .gdbinit (not the portable binary), a single --nx wasn't enough to skip ~/.gdbinit because the loader always assumed an extra -nx had been injected and so wanted two of them. that injected -nx only actually happens in portable mode, so i made the distinction explicit. the portable launcher now sets PWNDBG_LOADED_FROM_PORTABLE and we only mark _is_loaded_from_pwndbg in that case, then _disable_gdbinit_loading needs two --nx for portable and one when sourced, which matches vanilla gdb. went with the _is_loaded_from_pwndbg approach k4lizen mentioned in the thread.

pulled the flag counting out into pwndbg/lib/gdbinit.py so it's testable without dragging in real gdb. added unit tests for the portable/sourced/--nh/--args cases and they pass.

closes #3896

### DIFF
--- a/pwndbg/dbg_mod/gdb/__init__.py
+++ b/pwndbg/dbg_mod/gdb/__init__.py
@@ -1633,23 +1633,15 @@ class GDB(pwndbg.dbg_mod.Debugger):
 
         import psutil
 
-        disable_home_gdbinit = 0
-        disable_any_gdbinit = 0
+        import pwndbg
+        import pwndbg.lib.gdbinit
+
+        # _is_loaded_from_pwndbg only gets set when we came in through the portable
+        # binary, which is the thing that injects an extra -nx.
+        loaded_from_portable = hasattr(pwndbg, "_is_loaded_from_pwndbg")
+
         proc = psutil.Process(os.getpid())
-        for arg in proc.cmdline():
-            if arg in ("-args", "--args"):
-                break
-            if arg in ("-nh", "--nh"):
-                disable_home_gdbinit += 1
-            elif arg in ("-nx", "--nx", "-n", "--n"):
-                disable_any_gdbinit += 1
-
-        if disable_any_gdbinit == 0:
-            # The `--nx` option is added only in pwndbg-portable mode.
-            # This check allows using OLD syntax, eg: `source /path/to/pwndbg/gdbinit.py`, from ~/.gdbinit
-            return True, True
-
-        return disable_any_gdbinit >= 2, disable_home_gdbinit >= 1
+        return pwndbg.lib.gdbinit.disable_gdbinit_loading(proc.cmdline(), loaded_from_portable)
 
     def _load_gdbinit(self):
         # Emulate how `gdb` loads `.gdbinit` files (home and local)
--- a/pwndbg/lib/gdbinit.py
+++ b/pwndbg/lib/gdbinit.py
@@ -1,0 +1,37 @@
+"""Small helpers for figuring out whether we should skip loading gdbinit files.
+
+Kept dependency-free so it can be unit tested without dragging in the real gdb
+module.
+"""
+
+from __future__ import annotations
+
+
+def disable_gdbinit_loading(cmdline: list[str], loaded_from_portable: bool) -> tuple[bool, bool]:
+    """Mirror gdb's --nx/--nh handling, accounting for the portable launcher.
+
+    Returns (disable_any, disable_home). cmdline is the process argv, and
+    loaded_from_portable says whether we got here through the portable `pwndbg`
+    binary (which injects its own -nx) instead of being sourced from a user's
+    own .gdbinit.
+    """
+    disable_home_gdbinit = 0
+    disable_any_gdbinit = 0
+    for arg in cmdline:
+        if arg in ("-args", "--args"):
+            break
+        if arg in ("-nh", "--nh"):
+            disable_home_gdbinit += 1
+        elif arg in ("-nx", "--nx", "-n", "--n"):
+            disable_any_gdbinit += 1
+
+    if disable_any_gdbinit == 0:
+        # The `--nx` option is added only in pwndbg-portable mode.
+        # This keeps the OLD syntax working, eg: `source /path/to/pwndbg/gdbinit.py` from ~/.gdbinit
+        return True, True
+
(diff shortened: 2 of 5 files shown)


k4lizen:
> when pwndbg is sourced from a user's own .gdbinit (not the portable binary), a single --nx wasn't enough to skip ~/.gdbinit

idkwym, if you run `gdb --nx`, pwndbg won't be sourced. are you doing `pwndbg --nx` and also sourcing pwndbg from your gdbinit? could you provide a screenshot of what happens before and after your PR?

also no debugger-specific code is allowed in /lib/

clmoon2:
yeah youre right on both counts. the /lib/ placement breaks the no-debugger-code rule, and honestly i cant give you a clean before/after repro of the sourced case the way i described it. closing this, sorry for the noise.

Issue #3615, opened 2026-01-11
fix jemalloc failing tests


k4lizen:
### Description

Tests are failing flakily. Something needs to be fixed:
https://github.com/pwndbg/pwndbg/actions/runs/20827026073/job/59830949094?pr=3605

In particular, these three tests (or the jemalloc code that those tests touch) need to be fixed:
```
Failing tests:
- tests/library/dbg/tests/heap/test_heap.py::test_jemalloc_find_extent
- tests/library/dbg/tests/heap/test_heap.py::test_jemalloc_heap
- tests/library/dbg/tests/heap/test_heap.py::test_jemalloc_extent_info
```

PR #3957, opened 2026-06-07
fix jemalloc flaky tests \\ bump jemalloc 5.3.0 > 5.3.1 \\ fix arch build \\ dont try to push if build fails

jackmisbach:

fix jemalloc flaky tests - issue #3615 

---

* root cause: GitHub runners where LG_VADDR = 48 passed, those with 57-bit (5-level paging) failed. The majority of runners are 48-bit, which is why it would pass most of the time before. Here are some runs as well

https://github.com/Monthly-Recurring-Revenue/pwndbg/actions/runs/27079429263 - jemalloc.py before on 5.3.0 - fail vaddr = 57
https://github.com/Monthly-Recurring-Revenue/pwndbg/actions/runs/27078453899 - jemalloc.py after on 5.3.0 - pass vaddr = 57
https://github.com/Monthly-Recurring-Revenue/pwndbg/actions/runs/27080298137 - jemalloc.py before on 5.3.1 - fail vaddr = 57
https://github.com/Monthly-Recurring-Revenue/pwndbg/actions/runs/27080169172 - jemalloc.py after on 5.3.1 - pass vaddr = 57

---

This PR also bumps jemalloc from 5.3.0 to 5.3.1 https://github.com/jemalloc/jemalloc/releases/tag/5.3.1 - 390 commits, first release in 4 years

---

Also disables jemalloc's C++ integration, as it doesn't compile against newer libstdc++ like in Arch's GCC, even on 5.3.1, which gets the Arch build past the jemalloc step. Separately, musl is now detected by test-compiling (not just by checking that musl-gcc exists), so a broken musl-gcc (as in Arch) is skipped rather than failing the build.

---

Also, makes it so that the workflow doesn't try to push the image if the build step fails.
Discord_loWWUKzKoB

Discord_lCQTVIrYOg

related to PR #3884 

I didn't think it would make sense to separate this stuff, so sorry for all the things in one PR

### DIFF
--- a/pwndbg/aglib/heap/jemalloc.py
+++ b/pwndbg/aglib/heap/jemalloc.py
@@ -1,5 +1,8 @@
 from __future__ import annotations
 
+from typing import NamedTuple
+
+import pwndbg
 import pwndbg.aglib.memory
 import pwndbg.aglib.symbol
 import pwndbg.aglib.typeinfo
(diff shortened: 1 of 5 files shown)


setup-dev.sh
```diff
@@ -184,11 +184,11 @@ install_jemalloc() {
     else
         echo "Jemalloc not found in system. Downloading, configuring, building, and installing..."
 
-        # Install jemalloc version 5.3.0
-        JEMALLOC_TAR_URL="https://github.com/jemalloc/jemalloc/releases/download/5.3.0/jemalloc-5.3.0.tar.bz2"
-        JEMALLOC_TAR_SHA256="2db82d1e7119df3e71b7640219b6dfe84789bc0537983c3b7ac4f7189aecfeaa"
-        JEMALLOC_TAR_PATH="/tmp/jemalloc-5.3.0.tar.bz2"
-        JEMALLOC_EXTRACT_PATH="/tmp/jemalloc-5.3.0"
+        # Install jemalloc version 5.3.1
+        JEMALLOC_TAR_URL="https://github.com/jemalloc/jemalloc/releases/download/5.3.1/jemalloc-5.3.1.tar.bz2"
```
OBarronCS, line 188: Did any of the changes in jemalloc.py happen due to the version change, or is it unrelated code cleanup / related to fixing 57 bit builds?

If the tests still succeed, sounds like the internal structures of jemalloc that the tests are targeting didn't change between these two versions?

tests/binaries/host/makefile
```diff
@@ -38,7 +38,10 @@ JEMALLOC_CFLAGS := $(shell $(PKG_CONFIG) --cflags jemalloc)
 JEMALLOC_LIBS   := $(shell $(PKG_CONFIG) --libs jemalloc)
 endif
 
-HAVE_MUSL := $(shell command -v $(firstword $(MUSLCC)) >/dev/null 2>&1 && echo yes || echo no)
+# musl is "available" only if musl-gcc actually compiles+links a trivial program.
+# Some distros (e.g. Arch) ship a musl-gcc whose spec is broken against the current
+# GCC, so an existence check alone is not enough.
```
k4lizen, line 43: hm but this is pretty major no? you're saying musl-gcc just doesn't work on arch? but I see no bug reports on the package: https://gitlab.archlinux.org/archlinux/packaging/packages/musl/-/work_items (although i do have issues locally, but i figured its just my setup thats broken)

pwndbg/aglib/heap/jemalloc.py
```diff
@@ -179,6 +182,58 @@ def lg_ceil(x):
...
+
+def _load_rtree_geom() -> _RtreeGeom:
+    """Read jemalloc's rtree geometry from the target instead of assuming it.
+
+    jemalloc derives the tree's height, per-level bit splits and leaf layout from
+    LG_VADDR at build time -- 48 on 4-level-paging hosts, 57 on 5-level (LA57) -- and
+    the two layouts are incompatible, so assuming 48 misreads a 57-bit build (the
+    #3615 flake: it depended on the CI build host's CPU). Falls back to the 48-bit
+    layout if jemalloc's ``rtree_levels`` table can't be read."""
+    try:
```
k4lizen, line 217: can we get a link to the source where the definition of the structure (dependant on paging) is

pwndbg/aglib/heap/jemalloc.py
```diff
@@ -179,6 +182,58 @@ def lg_ceil(x):
...
+                int(ctx.evaluate_expression(f"rtree_levels[{i}].cumbits")),
+            )
+            for i in range(height)
+        ]
+        nhib = levels[0][1] - levels[0][0]
+        node_sizeof = int(ctx.evaluate_expression("sizeof(struct rtree_node_elm_s)"))
+        leaf_sizeof = int(ctx.evaluate_expression("sizeof(struct rtree_leaf_elm_s)"))
+        # The leaf packs the edata pointer with its metadata into one word only when
+        # enough high address bits are free (LG_VADDR <= 48); otherwise it stores a
+        # plain edata pointer plus a separate metadata word (twice the size).
```
k4lizen: link to source

k4lizen (approved):
thanks!

ngl i don't feel uber confident about the jemalloc changes (nor do i have enough expertise to review them) but i suppose its fine

setup-dev.sh
jackmisbach: Great question, I was considering splitting this up into 2 PRs because of this clarity, but didn't like the ideas I had of doing that.

So the version change did not cause any jemalloc.py changes, to answer your question. The version bump was something I did afterwards.

> sounds like the internal structures of jemalloc that the tests are targeting didn't change between these two versions?

It seems so. The geometry/leaf logic is still the same as 5.3.0.

pwndbg/aglib/heap/jemalloc.py
jackmisbach: https://github.com/jemalloc/jemalloc/blob/81034ce1f1373e37dc865038e1bc8eeecf559ce8/include/jemalloc/internal/rtree.h#L21-L38
Also updated the docstring with it though

tests/binaries/host/makefile
jackmisbach: I believe it's a GCC 16 issue, actually, not musl. but musl-gcc uncovers it

specs show
```
$ gcc -dumpspecs | grep -o -- '-latomic_asneeded'
-latomic_asneeded
```
so musl-gcc will fail to link anything
-fno-link-libatomic works tho

plain gcc works still, though, and tolerates it, as it finds it as /usr/lib is on the default search, but musl-gcc doesn't search /usr/lib so ld can't find -latomic_asneeded and the link will fail. 
https://www.mail-archive.com/gentoo-commits@lists.gentoo.org/msg1122689.html

https://github.com/rui314/mold/issues/1545 - https://gcc.gnu.org/bugzilla/show_bug.cgi?id=123654

But the makefile changes should be the proper way for handling this, I believe

PR #3959, opened 2026-06-07
Support offline uv sync in Docker builds

jackmisbach:

Running the pwndbg container offline fails because uv run re-validates the environment and rebuilds pwndbg's editable install, which, under default build isolation, spins up an isolated build env and fetches the build backend (hatchling) from the network.

This:

adds a build dependency group (hatchling, editables) so the backend is pinned in uv.lock and available from the uv cache,
in setup.sh, pre-installs that group, then runs uv sync --no-build-isolation-package pwndbg, so building pwndbg reuses the already-installed backend instead of fetching it in an isolated env. Scoped to setup.sh so other uv sync calls (docs, lint) keep default build isolation.

Refs: 
[uv build isolation](https://docs.astral.sh/uv/concepts/projects/config/#build-isolation), 
[uv no-build-isolation-package](https://docs.astral.sh/uv/reference/settings/#no-build-isolation-package)


jackmisbach:
Discord_zhjmC8lJoT

PR #3966, opened 2026-06-10
Libc version testing

jackmisbach:
libc version testing - the main goal is to test more libc versions in CI than the current distros we run the Docker tests on provide

glibc 2.35-2.43 (9 versions)
musl 1.1.24 + 1.2.1-1.2.6 (7 versions)
tested on both x86-64 and aarch64

The libc versions are built from source in [`Dockerfile.glibc-test-libs`](https://github.com/pwndbg/pwndbg/pull/3966/changes#diff-8888bcac4d7a22bb96c91f1d8846871c60003a966965099555e9357df06bdc3c) and [`Dockerfile.musl-test-libs`](https://github.com/pwndbg/pwndbg/pull/3966/changes#diff-eacd82c195885b322eac29af03de6cad674b0228c942f08425debeefb386adbd). These are also the place where you add/remove a new version to test.

[`test_heap_glibc_versions.py`](https://github.com/pwndbg/pwndbg/pull/3966/changes#diff-deb4c6135df82df3963f839b6a5ce24e82a88fd2e43869859cdea7426633b38f) and [`test_musl_versions.py`](https://github.com/pwndbg/pwndbg/pull/3966/changes#diff-4419940638b359a8fc23e6b36482cfb308bc240612c8a6393fbd00755f200f6f) are the dedicated per-version tests.

[`heap-libc-tests.yml`](https://github.com/pwndbg/pwndbg/pull/3966/changes#diff-c7c13ccf1fa6ecc4b96b28d734c24267a8ec7f65edbbd4f52ed9376bab68ee64) runs the tests in CI. It pulls (or builds) the libc images, extracts them, and then runs the tests. The initial build is slow, but the images are published (later runs just pull), and their layers are cached (a rebuild is much faster when needed). The tests run on [relevant file changes](https://github.com/pwndbg/pwndbg/pull/3966/changes#diff-c7c13ccf1fa6ecc4b96b28d734c24267a8ec7f65edbbd4f52ed9376bab68ee64R45-R70)  (heap/libc code, tests, or the [Docker/build setup](https://github.com/pwndbg/pwndbg/pull/3966/changes#diff-c7c13ccf1fa6ecc4b96b28d734c24267a8ec7f65edbbd4f52ed9376bab68ee64R71-R76)) on PRs and pushes. 

**NOTICE:** Claude was used in the initial creation of this PR


tests/library/dbg/tests/test_glibc_heap_bins.py
```diff
@@ -1,19 +1,28 @@
 from __future__ import annotations
 
+from pathlib import Path
+
```
OBarronCS, line 4: I think it would be a good idea to rename all the test files that target `glibc` specifically to all include "glibc" in the file name.

OBarronCS: i.e. test_glibc_heap_bins, in this case (or whatever you find most appropriate)

tests/binaries/host/Makefile
```diff
@@ -70,12 +87,125 @@ ifeq ($(HAVE_MUSL),yes)
 ALL_MUSL_TARGETS += heap_musl_dyn.native.out heap_musl_static.native.out heap_musl_static_stripped.native.out
 endif
 
+GLIBC_TEST_VERSIONS := $(shell sed -n 's/^FROM base-builder AS build-\([0-9][0-9.]*\).*/\1/p' $(PWD)/../../../Dockerfile.glibc-test-libs)
+GLIBC_HEAP_BINARIES := heap_malloc_chunk heap_malloc_chunk_dump heap_bins heap_find_fake_fast
```
OBarronCS: A comment about this to explain the file naming convention would be useful:

`X.native.c` I believe is the pattern here? So the developer makes a file named "heap_malloc_chunk.native.c", and this Makefile will automatically find it, and compile it multiple times, linking it against each respective veresion of glibc

docs/contributing/writing-tests.md
```diff
@@ -140,3 +141,22 @@ int main () {
...
+
+- **glibc** (2.35-2.43): version detection; the heap commands (allocator, bins, `malloc-chunk`, `dt`, `find-fake-fast`) via debug symbols; the forced-heuristic path; and a no-symbol heuristic run against a fully stripped libc.
+- **musl** (1.1.24-1.2.6): detection and exact version, static (where the mallocng fingerprint exists) and dynamic, plus the full `test_mallocng.py` suite.
+
+**Adding or removing a glibc or musl version touches two files:** add (or remove) the `FROM base-builder AS build-<version>` stage and its scratch-stage `COPY` line(s) in the relevant `Dockerfile.*-test-libs`, and pin the new release tarball's sha256 in `scripts/build-one-<libc>.sh`.
+
+**Running locally** (needs Docker):
+
+```bash
+./scripts/download-test-libs.sh glibc    # or musl
```
OBarronCS: Could you add a comment (and at the top of the download script) that clarifies that this script downloads/builds into a Docker image, and doesn't put the libc files onto the host.

This is a really cool way of doing it!

tests/binaries/host/Makefile
```diff
@@ -13,6 +13,22 @@ else ifeq ($(HOST_ARCH),i686)
     HOST_ARCH := i386
 endif
 
+ifeq ($(HOST_ARCH),x86-64)
```
OBarronCS, line 16: Small change we can make along the way: capitalize `makefile` to `Makefile`

scripts/download-test-libs.sh
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+LIBC="${1:?Usage: $0 <glibc|musl>}"
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
```
OBarronCS, line 7: This is really cool.

To test this, I ran this download script on my device (building the glibc tests), and my computer froze after about a minute, likely because the Docker image is building the independent layers (of each target glibc version) in parallel. Could there maybe be a flag, like `--serial`, that causes the layers to build one at a time, to avoid using too much ram + cpu while building this?

tests/library/dbg/tests/__init__.py
```diff
@@ -41,6 +42,26 @@ def get_binary(name: str) -> Path:
...
+
+def glibc_test_versions() -> list[str]:
+    return _dockerfile_versions("glibc")
+
+
+def musl_test_versions() -> list[str]:
+    return _dockerfile_versions("musl")
+
+
+def glibc_version_binaries(stem: str) -> list[tuple[str, Path]]:
```
OBarronCS: Could you document, through an example, what the stem here looks like?

docs/contributing/writing-tests.md
```diff
@@ -140,3 +141,22 @@ int main () {
...
+make -C tests/binaries/host -j4 all  # glibc/musl compile per-version binaries
+./tests.sh -d gdb -g dbg test_heap_glibc_versions
```
OBarronCS, line 165: Does this command result in the tests running inside the docker containers that we built, with the target libc? 

Is this command supposed to be inside of the container?

scripts/download-test-libs.sh
OBarronCS: Tested it again: the build takes a lot of RAM since it's building glibc multiple times in parallel, and it will run for a while (40+ minutes).

Discussed with the contributor, let's do a `--build` flag instead that causes local Docker build, otherwise pull the images.

docs/contributing/writing-tests.md
jackmisbach: Just responding here, although we did talk about this on Discord, and you were asking a bit of a different question, at least in comparison to the way I'm going to respond to it on here, just so I can provide some more context.

> Does this command result in the tests running inside the docker containers that we built, with the target libc?
>
> Is this command supposed to be inside of the container?

Not inside the libc images (glibc-test-libs/musl-test-libs).

The workflow `heap-libc-tests.yml` uses the pwndbg ubuntu24.04 image that we push through `docker.yml`  for the Run version tests step, which does the `make` and then the `./tests.sh -d gdb -g dbg "${{ matrix.test_filter }}` 

The Dockerfiles that I created in this PR (glibc-test-libs/musl-test-libs) are only used for the build step, whose contents are then extracted and placed into the ubuntu24.04 container via a bind mount, almost mimicking `./scripts/download-test-libs.sh glibc` which does the extract locally.

docs/contributing/writing-tests.md
jackmisbach: Heh, actually, it does put them onto the host; it extracts them from the image.

scripts/download-test-libs.sh
jackmisbach: I do just want to point out for clarity, right now it does already pull the images by default unless they do not exist, which in this case they don't exist, so it builds.

What I will introduce, however, from our discussion is that it will not fallback to build, but will rather only build if you pass the --build flag. If there is an issue somehow that the image can't be found/pulled, it will simply exit with a different warning and not fall back to building.

tests/binaries/host/Makefile
jackmisbach: I added comments to explain this. They would just make a file named "heap_malloc_chunk.native.c" and add it as heap_malloc_chunk here in the list of `GLIBC_HEAP_BINARIES` and then if they needed to add a version they would add it as a stage to the Dockerfile, as I wanted to minimize the "source of truth" so the Dockerfiles are the source of truth, that and the build scripts, in which they are sha pinned to verify. That way, new versions only have to be updated in the Dockerfile and the build script, rather than many other places.

jackmisbach: Wait, lol, I think I can maybe even make it just the Dockerfile, so it's just one place, let me test if this can work how I'm imagining though

tests/library/dbg/tests/__init__.py
jackmisbach: Adding a docstring

jackmisbach:
Alright, I don't think I missed anything, pls reping me if I did or if you have more questions as you get deeper into the code.

Also want to share this from our conversation to add context.
Discord_TPvItbiWiz
This, however, I can add into a new PR at a later time, just to try and keep the scope down in this PR as it is already so massive.

tests/binaries/host/Makefile
jackmisbach: Alright, looks like this was possible, so versions and shas are all handled within the glibc-test-libs/musl-test-libs Dockerfiles; this makes me happy.

k4lizen:
needs description + rebase I suppose

.github/workflows/heap-libc-tests.yml
```diff
@@ -0,0 +1,251 @@
...
+      matrix:
+        include:
+          - libc: glibc
+            arch: x86-64
+            runs_on: ubuntu-latest
+            image: glibc-test-libs
+            dockerfile: Dockerfile.glibc-test-libs
+            artifact_dir: glibcs
+            sentinel: ld-linux-x86-64.so.2
+            test_filter: "test_glibc|test_heap_glibc_versions|test_command_dt"
```
OBarronCS, line 96: Are these filters a bit brittle? Fine for now, no change needed now. But in the future we may want to consolidate this to a single word. If adding a new test right now, it need to follow one of these naming conventions to apply to this workflow.

.github/workflows/heap-libc-tests.yml
```diff
@@ -0,0 +1,251 @@
+name: Heap Libc Version Tests
```
OBarronCS, line 1: Can you please add a comment to the top of this file indicating what the different jobs do. Right now, this one file appears to both runs tests AND build the images.

docs/contributing/writing-tests.md
```diff
@@ -140,3 +141,22 @@ int main () {
 ## QEMU Tests
 
 To test architecture specific features, like disassembly annotations, we use emulate the appropriate architecure with qemu-user and attach to its debug port. These tests are located in [`tests/library/qemu_user/tests`](https://github.com/pwndbg/pwndbg/tree/dev/tests/library/qemu_user/tests). They are currently `gdb-only` and thus follow the same format as the `gdb/` tests. They require a Python function with a Pytest fixture name as the parameter (it matches based on the name). You call the argument/fixture to start debugging a binary. The `qemu_assembly_run` fixture takes in a Python string of assembly code, compiles it in the appropriate architecture, and runs it - no need to create an external file or edit a Makefile.
+
+## Libc Version Testing
+
+`tests/library/dbg/tests/test_heap_glibc_versions.py` and `test_musl_versions.py` exercise pwndbg against many libc versions.
```
OBarronCS: Please specify in this section a key part of the system: the docker images are built with the specific libc library artifacts, and before running the tests those artifacts are copied to the host. The tests run on the host, not inside the new libc-building container.

OBarronCS: Also please note more details about the building of the libc versions. i.e. both with and without debug symbols, etc.

tests/binaries/host/Makefile
```diff
@@ -70,12 +87,129 @@ ifeq ($(HAVE_MUSL),yes)
...
+
 CUSTOM_TARGETS = heap_find_fake_fast.native.out reference_bin_pie.native.out reference_bin_nopie.native.out reference_bin_nopie.i386.out symbol_1600_and_752.native.out initialized_heap.x86-64.out initialized_heap_big.i386.out linked_lists.native.out onegadget.x86-64.out onegadget.i386.out heap_musl_dyn.native.out heap_musl_static.native.out heap_musl_static_stripped.native.out gosample.i386.out gosample.x86-64.out
 NATIVE_CUSTOM_TARGETS  := $(filter %.native.out %.$(HOST_ARCH).out %.$(HOST_ARCH_32).out,$(CUSTOM_TARGETS))
 
 ALL_TARGETS := $(LINKED_C) $(LINKED_ASM) $(LINKED_GO) $(NATIVE_CUSTOM_TARGETS) \
     $(ALL_JEMALLOC_TARGETS) \
-    $(ALL_MUSL_TARGETS)
+    $(ALL_MUSL_TARGETS) \
+    $(ALL_GLIBC_VERSION_TARGETS) \
+    $(ALL_MUSL_VERSION_TARGETS)
```
OBarronCS, line 212: Building both musl statically linked to binary + dynamically linked. Do tests run both configurations? Why? Please add a comment explaining this in the writing-tests.md

tests/library/dbg/tests/test_glibc_heap.py
```diff
@@ -1,20 +1,42 @@
...
 from . import get_binary
+from . import glibc_version_binaries
 from . import launch_to
 from . import pwndbg_test
 
 HEAP_MALLOC_CHUNK = get_binary("heap_malloc_chunk.native.out")
-HEAP_MALLOC_CHUNK_DUMP = get_binary("heap_malloc_chunk_dump.native.out")
 HEAP_GLIBC2_43 = get_binary("heap_glibc2.43.native.out")
 
+_HEAP_BINARIES = glibc_version_binaries("heap_malloc_chunk")
```
OBarronCS: Can you rename this variable? One first glance, it appears to indicate it refers to all heap related binaries, but it's just a specific tests. Same goes for any other ones like this.

tests/library/dbg/tests/test_heap_glibc_versions.py
```diff
@@ -0,0 +1,295 @@
...
+    await ctrl.cont()
+
+    result = allocator.largebins()
+    assert result is not None
+    assert result.bin_type == BinType.LARGE
+
+    await ctrl.execute("bins")
+
+
+@pytest.mark.parametrize("glibc_version", GLIBC_VERSIONS)
```
OBarronCS: Some test use @glibc_version_binaries() decorator, which tests in this file do not, instead doing this manual string interpolation to find the binary. Should these be changed to the other method?

docs/contributing/writing-tests.md
OBarronCS: It might be nice to point to files like `test_heap_glibc_versions.py` and `test_musl_versions.py` as places to start when looking to understand the setup more deeply.

OBarronCS (changes requested):

OBarronCS:
This looks great. I made comments related to increasing clarity of the code mostly through docs changes. Once those are addressed this is good to go.

.github/workflows/heap-libc-tests.yml
k4lizen: i agree

docs/contributing/writing-tests.md
```diff
@@ -140,3 +141,26 @@ int main () {
...
+
+## Libc Version Testing
+
+`tests/library/dbg/tests/test_heap_glibc_versions.py` and `test_musl_versions.py` exercise pwndbg against many libc versions, and are where you should start to get a better understanding of the libc version testing.
+
+Rather than write new per-version tests, we parametrize the existing tests over the available versions:
+
+- **glibc** (2.35-2.43): version detection; the heap commands (allocator, bins, `malloc-chunk`, `dt`, `find-fake-fast`) via debug symbols; the forced-heuristic path; and a no-symbol heuristic run against a fully stripped libc.
```
k4lizen: im a bit confused, this makes it sound like there are three runs, is this the case?

docs/contributing/writing-tests.md
```diff
@@ -140,3 +141,26 @@ int main () {
...
+
+**Why both musl linkages?** A statically linked binary has the whole libc (including the mallocng allocator and its symbols) compiled into the ELF, while a dynamically linked one loads `libc.so` as a separate mapping. pwndbg locates and parses the allocator differently in each case, so the mallocng suite runs against both to confirm detection and that the `ng` heap commands work regardless of how musl was linked.
+
+**Adding or removing a glibc or musl version is a one-file change** in the relevant `Dockerfile.*-test-libs`: add (or remove) the `FROM base-builder AS build-<version>` stage, its `RUN` line (which passes the release tarball's sha256), and its scratch-stage `COPY` line(s).
```
k4lizen: this is not really true, as all the docs references and comments which mention the version range should be updated as well.

tests/binaries/host/heap_musl.native.c
```diff
@@ -1,6 +1,14 @@
 #include <stdlib.h>
 #include <string.h>
 
+// zig 0.14.1 drops -Wl,--dynamic-linker on -nostdlib links
+// https://github.com/ziglang/zig/issues/23813 - fixed in zig 0.15.0
+// So, instead we define the .interp section here.
+#ifdef MUSL_DYNAMIC_INTERP
+const char __musl_dynamic_interp[] __attribute__((section(".interp"), used)) =
+    MUSL_DYNAMIC_INTERP;
+#endif
```
k4lizen, line 11: can you be more clear on what issue this fixes in the comment?

tests/binaries/host/Makefile
```diff
@@ -70,12 +87,129 @@ ifeq ($(HAVE_MUSL),yes)
 ALL_MUSL_TARGETS += heap_musl_dyn.native.out heap_musl_static.native.out heap_musl_static_stripped.native.out
 endif
 
+# Versions come from the build-<version> stages in Dockerfile.glibc-test-libs (the source of truth).
+GLIBC_TEST_VERSIONS := $(shell sed -n 's/^FROM base-builder AS build-\([0-9][0-9.]*\).*/\1/p' $(PWD)/../../../Dockerfile.glibc-test-libs)
+# Add a binary by creating <stem>.native.c and listing it here; add a version via a Dockerfile stage.
+GLIBC_HEAP_BINARIES := heap_malloc_chunk heap_malloc_chunk_dump heap_bins heap_find_fake_fast
```
k4lizen: as a more general comment, the per-libc version testing should not have anything to do with allocators / heap. 

we can, do, and should, have code which is per-libc but not related to allocators, like e.g. `tests/binaries/host/exithandlers.native.c` which is not included here (but it should be)

k4lizen: this libc / heap intertwining permates this PR in a lot of places, I wouldn't block the merging because of this but it should be fixed in a future PR

tests/binaries/host/Makefile
```diff
@@ -70,12 +87,129 @@ ifeq ($(HAVE_MUSL),yes)
...
+define MUSL_HEAP_STATIC_TARGET
+$(1).musl-$(2)-static.out: $(1).native.c $(PWD)/musls/$(2)/lib/libc.a $(PWD)/musls/$(2)/lib/crt1.o $(PWD)/musls/$(2)/lib/crti.o $(PWD)/musls/$(2)/lib/crtn.o
+	@echo "[+] Building '$$@' against musl $(2) (static)"
+	$${ZIGCC} $${CFLAGS} -static -no-pie -fno-pie -nostdlibinc -nostdlib \
+	-isystem $$(PWD)/musls/$(2)/include \
+	-Wl,-u,__libc_version \
+	-o $$@ \
+	$$(PWD)/musls/$(2)/lib/crt1.o $$(PWD)/musls/$(2)/lib/crti.o \
+	$$< \
+	$$(PWD)/musls/$(2)/lib/libc.a $$(PWD)/musls/$(2)/lib/crtn.o
```
k4lizen, line 171: why are we compiling like this?

does something like this not work?

Image

tests/library/dbg/tests/test_glibc_heap.py
```diff
@@ -583,19 +613,22 @@ async def test_thread_cache_heuristic(ctrl: Controller, is_multi_threaded: bool)
         )
 
 
+@glibc_versions
```
k4lizen: this is awesome

tests/library/dbg/tests/test_heap_glibc_versions.py
```diff
@@ -0,0 +1,295 @@
...
+    assert main_arena is not None, (
+        f"main_arena not found for glibc {glibc_version} (heuristic={use_heuristic})"
+    )
+
+    result = await ctrl.execute_and_capture("heap")
+    assert len(result) > 0, f"'heap' command produced no output for glibc {glibc_version}"
+
+
+_XFAIL_242 = pytest.mark.xfail(
+    reason="stripped glibc 2.42 heuristic can't recover main_arena",
```
k4lizen: thats kinda bad XD

k4lizen: also weird, 2.42 cant but 2.43 can?

tests/library/dbg/tests/test_mallocng.py
```diff
@@ -191,17 +212,15 @@ async def test_mallocng_slot_start(ctrl: Controller, binary: Path):
...
         # Doing `ng slots buffer5` will give you garbage since buffer5 is not
         # a valid slot start.
         assert slotu_buffer5_out != slots_buffer5_out
 
 
 @pwndbg_test
-@pytest.mark.parametrize(
-    "binary", [HEAP_MALLOCNG_DYN, HEAP_MALLOCNG_STATIC], ids=["dynamic", "static"]
-)
+@pytest.mark.parametrize("binary", _MALLOCNG_BINARIES, ids=_MALLOCNG_IDS)
```
k4lizen, line 223: is the order of decorators important?

tests/library/dbg/tests/test_musl_versions.py
```diff
@@ -0,0 +1,52 @@
...
+from ....host import Controller
+from . import get_binary
+from . import launch_to
+from . import musl_test_versions
+from . import pwndbg_test
+
+MUSL_VERSIONS = musl_test_versions()
+assert MUSL_VERSIONS, "no musl versions parsed from Dockerfile.musl-test-libs"
+
+_MALLOCNG_MIN = (1, 2, 1)
```
k4lizen: i dont think mallcng should affect whether musl is being tested dynamically

.gitignore
```diff
@@ -81,6 +81,12 @@ tests/binaries/**/div_zero_binary/core
 tests/binaries/**/div_zero_binary/binary
 !tests/binaries/**/*.go
 
+# libc artifacts from glibc and musl builds
+/tests/binaries/host/glibcs/*
+!/tests/binaries/host/glibcs/2.33/
+/tests/binaries/host/glibcs-nodebug/
+/tests/binaries/host/musls/
```
k4lizen, line 88: this can start with `/` really? wild

k4lizen (approved):
left some comments, but all in all looks pretty awesome

k4lizen:
i'd also like to see docs on how to add support for a new libc (since e.g. support for bionic is coming in parallel to your work)

tests/library/dbg/tests/test_heap_glibc_versions.py
k4lizen: agreed, i would hope we can somehow unify this so each test doesn't need to implement its own decorator

tests/binaries/host/Makefile
jackmisbach: I only cached the compiled libc, and not musl-gcc, which wouldn't be "relocatable" (it hard-codes the install path, so it'd break once cached and pulled back into a container at a different path), so I hand-linked the artifacts with Zig instead.

tests/library/dbg/tests/test_mallocng.py
jackmisbach: nah

tests/binaries/host/Makefile
jackmisbach: I'll just rename everything from heap to version in this PR just committed that now 

and ill add the exithandlers and whatever else there is when I take a look on the follow up pr with some other things I want to add as well

tests/library/dbg/tests/test_musl_versions.py
jackmisbach: That's for static only, they're all tested dynamically

ill rename the var to 
STATIC_DETECTION_MIN

jackmisbach: actually nvm, I can remove this, but anyway, static 1.1.24 still isn't built because there would be nothing additional to test

tests/library/dbg/tests/test_heap_glibc_versions.py
jackmisbach: yeah I just put xfail for now, probably something that needs fixing though at some point

tests/binaries/host/heap_musl.native.c
jackmisbach: i expanded the comment a bit with more info on the issue

docs/contributing/writing-tests.md
jackmisbach: yeah lol idk why the docs themselves needing updating didn't even cross my mind, I was writing from like a 3rd person POV, but yeah I guess I'll put "Adding or removing a glibc or musl version is a one-file change (besides these docs)", or something if that works

docs/contributing/writing-tests.md
jackmisbach: yes (I updated the comment as well)

jackmisbach:
> i'd also like to see docs on how to add support for a new libc (since e.g. support for bionic is coming in parallel to your work)
Yeah, I'll definitely add something like this. Every libc could be quite different, especially bionic, but I'll try to make a good baseline of what to add.

OBarronCS:
This is good to merge once the comment in the "tests/library/dbg/tests/test_heap_glibc_versions.py" file is addressed.

tests/library/dbg/tests/test_heap_glibc_versions.py
jackmisbach: Okay, I just added a shared helper and made all the glibc-version tests use it, so each test doesn't need to build its own parameterization logic

OBarronCS (approved):
Thank you for making the requested changes.

OBarronCS:
One more thing before merging: 

When on this branch, I `./tests.sh -g dbg -d gdb` to run all tests. I have NOT run the download lib script to generate the libc files, so I expect those tests to be skipped/not run.

However, one of the outputs of the tests it this:

image

`glibc_version0-binary0`. Is this intended? Is it a bug in the regex that finds the versions from the docker file?
https://github.com/pwndbg/pwndbg/blob/c18608b83175c0ef9bc8671c0a84b88a6ab6621e/tests/library/dbg/tests/__init__.py#L47-L49

jackmisbach:
> glibc_version0-binary0. Is this intended? Is it a bug in the regex that finds the versions from the docker file?

I guess it wouldn't be considered a regex bug, but rather confusing/ugly.

I suppose there are two options: we can either do something like this, where each version shows skipped

```
test_heap_version_detection[2.35] SKIPPED (glibc 2.35 binary not available)
test_heap_version_detection[2.36] SKIPPED (glibc 2.36 binary not available)
...
test_heap_version_detection[2.43] SKIPPED (glibc 2.43 binary not available)
```

so if only two versions exist, we get

```
[2.35] PASSED
[2.36] PASSED
[2.37] SKIPPED
...
[2.43] SKIPPED
```

OR

we can do 

```
test_heap_version_detection[no-glibc-binaries] SKIPPED (no glibc version test binaries available)
```

but then if only two versions exist

```
[2.35] PASSED
[2.36] PASSED
```
We don't see the skips for others, but in CI this wouldn't happen because it would fail earlier due to missing versions. I mean, because it verifies that they all exist, so I think we can go with this (2nd option) to avoid the redundancy option 1 would cause (every version saying that the binary is missing) while still making it more clear, and do the same for musl

OBarronCS:
One nice part about the showing "skipped" for each one is that it gives the developer the syntax for the name of the test. So they can see  `test_heap_version_detection[2.43] SKIPPED`, and have an idea immediately of all the versions that are available, and how to filter to run only a specific version.

i.e. it's an intuitive way to show "this is how you run only this one test,  since now you know the format of that pytest name"
```sh
./tests.sh -g dbg -d gdb "test_heap_version_detection\[2.43\]"
```

Also, showing "skipped" for all can validate that you added a thing to the Dockerfile correctly, but just haven't pulled it yet (so you understand why it's not running, the libc files just aren't available for it yet)

OBarronCS:
@jackmisbach there seems to be an ongoing issue, #4091, that could have benefited from the work in this PR. Once the contents of the previous comments are implemented, I would like to merge this.

jackmisbach:
Should be good to go

OBarronCS:
Awesome! Thank you! This is really great to have in pwndbg, and it will help use catch bugs in our glibc functionality faster.

PR #3971, opened 2026-06-14
kernel: fix current_task lookup crash

navytuner:
There was a memory access error on vmmap, which turned out that it has calculated the address of `current` by `common_cpu_up`, not by `current_task`. I set `prefer_symbol=True` on `kernel_symbol_func` to find the offset from `current_task` symbol first. 

Screenshot 2026-06-14 at 8 00 00 PM

```
Error: Cannot access memory at address 0xffff8881b799a5f8
> /home/compsec/opt/pwndbg/pwndbg/dbg_mod/gdb/__init__.py(713)read_memory()
-> raise pwndbg.dbg_mod.Error(e)
(Pdb) bt
  /home/compsec/opt/pwndbg/pwndbg/commands/__init__.py(517)__call__()
-> return self.function(*args, **kwargs)
  /home/compsec/opt/pwndbg/pwndbg/commands/vmmap.py(245)vmmap()
-> pwndbg.aglib.kernel.vmmap.annotate(vmmap)
  /home/compsec/opt/pwndbg/pwndbg/lib/cache.py(214)decorator()
-> value = func(*a, **kw)
  /home/compsec/opt/pwndbg/pwndbg/aglib/kernel/vmmap.py(100)annotate()
-> _handle_user_stack_and_filepaths(pages)
  /home/compsec/opt/pwndbg/pwndbg/aglib/kernel/vmmap.py(79)_handle_user_stack_and_filepaths()
-> task = pwndbg.aglib.kernel.current_task()
  /home/compsec/opt/pwndbg/pwndbg/lib/cache.py(214)decorator()
-> value = func(*a, **kw)
  /home/compsec/opt/pwndbg/pwndbg/aglib/kernel/__init__.py(871)current_task()
-> return pwndbg.aglib.memory.read_pointer_width(ptr)
  /home/compsec/opt/pwndbg/pwndbg/aglib/memory.py(201)read_pointer_width()
-> return pwndbg.aglib.arch.unpack(read(addr, pwndbg.aglib.arch.ptrsize))
  /home/compsec/opt/pwndbg/pwndbg/aglib/memory.py(36)read()
-> return pwndbg.dbg.selected_inferior().read_memory(address=addr, size=count, partial=partial)
> /home/compsec/opt/pwndbg/pwndbg/dbg_mod/gdb/__init__.py(713)read_memory()
-> raise pwndbg.dbg_mod.Error(e)
```


jxuanli:
use symbol for this particular one often yields incorrect result so we should use symbol only as a fall back

pwndbg/aglib/kernel/symbol.py
```diff
@@ -419,7 +417,7 @@ def prog_idr(self) -> type[NeedLookup]:
         return NeedLookup
 
     # using symbols usually yield incorrect results
```
jxuanli, line 419: the comment, the `prefer_symbol` was only really added because of `currect_task`

pwndbg/aglib/kernel/symbol.py
```diff
@@ -332,10 +332,8 @@ def func(*args: P.args, **kwargs: P.kwargs) -> int | None:
                             self, f"_{f.__name__}"
                         )
                         result = arch_heuristic_handle()
-            if result is None and not prefer_symbol:
```
jxuanli, line 335: the check of `prefer_symbol` here is to prevent to redo what we already did

pwndbg/aglib/kernel/symbol.py
navytuner: I can't get why heuristic lookup is needed first in `current_task`. Could you please let me know the kernel version and configs where `&current_task` + `gs_base` doesn't work?

navytuner: After the patch(https://lore.kernel.org/all/20250123190747.745588-9-brgerst@gmail.com, included in `v6.15`), it seems that we cannot find the current task pointer by just adding 0x5f8 to `gs_base`

jxuanli: i don't have a kernel build atm to show case but it should have been tested on builds later than the patch

jxuanli: the offset is dynamically determined fwiw - like why is the page pointed to by gs_base not readable?

navytuner: I think that we need to interpret `gs_base` as just the offset(`__per_cpu_offset[i]`), not the base address of real per-cpu region anymore. `gs_base` doesn't point to valid page because it's just the offset(real per-cpu region - template region).

jxuanli: ok `result = syms.current_task()` should have returned a value such that `per_cpu(result, cpu=cpu)` returns a readable address. (`result` is likely negative when interpreted as signed long), i think it was prob a bug on the `current_task` side

jxuanli: could you disasm `common_cpu_up`

navytuner: You're right! The regex pattern has matched incorrectly

navytuner:
```asm
Dump of assembler code for function common_cpu_up:
   0xffffffff81277680 <+0>:     nop    DWORD PTR [rax-0x2a]
   0xffffffff81277684 <+4>:     nop    DWORD PTR [rax+rax*1+0x0]
   0xffffffff81277689 <+9>:     push   rbp
   0xffffffff8127768a <+10>:    push   rbx
   0xffffffff8127768b <+11>:    mov    rbx,rsi
   0xffffffff8127768e <+14>:    mov    ebp,edi
   0xffffffff81277690 <+16>:    call   0xffffffff8124e7c0 <alternatives_enable_smp>
   0xffffffff81277695 <+21>:    mov    eax,ebp
   0xffffffff81277697 <+23>:    mov    rax,QWORD PTR [rax*8-0x7d2434d0]
   0xffffffff8127769f <+31>:    mov    QWORD PTR [rax-0x7bd85ff8],rbx
   0xffffffff812776a6 <+38>:    mov    rax,QWORD PTR [rbx+0x5f8]
   0xffffffff812776ad <+45>:    movsxd rdi,ebp
   0xffffffff812776b0 <+48>:    mov    rcx,QWORD PTR [rdi*8-0x7d2434d0]
   0xffffffff812776b8 <+56>:    mov    QWORD PTR [rcx-0x7bd85fe0],rax
   0xffffffff812776bf <+63>:    pop    rbx
   0xffffffff812776c0 <+64>:    pop    rbp
   0xffffffff812776c1 <+65>:    jmp    0xffffffff81245110 <irq_init_percpu_irqstack>
End of assembler dump.
```

navytuner: It should match the store instruction with - sign
`QWORD PTR [rax-0x7bd85ff8],rbx`

pwndbg/aglib/kernel/symbol.py
```diff
@@ -419,7 +417,7 @@ def prog_idr(self) -> type[NeedLookup]:
         return NeedLookup
 
     # using symbols usually yield incorrect results
-    @kernel_symbol_func(prefer_symbol=False)
+    @kernel_symbol_func(prefer_symbol=True)
```
jxuanli: this need to be restored

pwndbg/aglib/kernel/symbol.py
```diff
@@ -469,18 +467,12 @@ def qword_op_reg_memoff(
             return int(result.group(1), 16)
         return None
 
-    # op [... +/- (0x...)] ...
-    # if negative, the `-0x...`` is a kernel address displayed as a negative number
-    # returns the first 0x... as an int if exists
-    def dword_op_memoff_reg(
```
jxuanli, line 475: don't change the function signature and only the regex please. the idea here is to let the caller specify which sign (`+/-`) / opcode should be matched - don't coalesce the two cases

pwndbg/aglib/kernel/symbol.py
```diff
@@ -469,18 +467,12 @@ def qword_op_reg_memoff(
...
-        self, disass: str, op: str, sign: str = "-", nth: int = 0
-    ) -> int | None:
-        result = self.regex(disass, rf"{op}.*\[.*{re.escape(sign)}\s(0x[0-9a-f]{{1,8}})\]", nth)
-        if result is not None:
-            if sign == "-":
-                return (1 << 64) - int(result.group(1), 16)
-            return int(result.group(1), 16)
-        return None
+    def dword_store_memoff_reg(self, disass: str, nth: int = 0) -> int | None:
+        result = self.regex(disass, r"mov.*\[.*([+-])\s(0x[0-9a-f]{1,8})\]\s*,\s*\w", nth)
```
jxuanli: it seems you are essentially adding `\s*,\s*\w` at the end of the regex string?

navytuner: Thank you for the review. We need to catch the 2nd value from these three cases:
```asm
mov    rax,QWORD PTR [rax*8-0x7d2434d0]
mov    QWORD PTR [rax-0x7bd85ff8],rbx
mov    rax,QWORD PTR [rbx+0x5f8]
```

Calling `dword_op_memoff_reg` with just `sign="-"` doesn't work because it matches the first line out of three. So it seems that we need to focus on store operation, which has the regex `"{op}.*\[.*([+-])\s(0x[0-9a-f]{{1,8}})\]\s*,"`.

navytuner: This is my sample code. If you have a better idea, please let me know. 
```python
def dword_op_memoff_reg(
        self, disass: str, op: str, nth: int = 0
    ) -> int | None:
        result = self.regex(disass, rf"{op}.*\[.*([+-])\s(0x[0-9a-f]{{1,8}})\]\s*,", nth)
        if result is None:
            return None
        value = int(result.group(2), 16)
        return (1 << 64) - value if result.group(1) == "-" else value
```

jxuanli: maybe we can add a `first: bool = False` arg to indicate if the memory offset should be from the first or second oprand

maybe
```py
    def dword_op_memoff_reg(
        self, disass: str, op: str, sign: str = "-", first: bool = True, nth: int = 0
    ) -> int | None:
        r = rf"{op}.*\[.*{re.escape(sign)}\s(0x[0-9a-f]{{1,8}})\]\s*,"
        if not first:
            r = rf"{op}.*,\s*\[.*{re.escape(sign)}\s(0x[0-9a-f]{{1,8}})\]"
        result = self.regex(disass, r, nth)
        if result is not None:
            if sign == "-":
                return (1 << 64) - int(result.group(1), 16)
            return int(result.group(1), 16)
        return None
```

lmk what you think

navytuner: lgtm. For now, let's go with this.

navytuner: We need to match both + and - to cover previous versions as well.

k4lizen:
(if/when this PR is ready for re-review let us know)

jxuanli (approved):
yea, lets merge this

PR #3973, opened 2026-06-15
feat: add exithandlers command to view glibc exit handlers 

h4kk4p4kk4:
+ [X] I am providing a screenshot of the new feature.

Implements a command to get the pointer_guard and list exit handlers from __exit_funcs and tls_dtor_list. Works with or without glibc debug symbols.

image


h4kk4p4kk4:
Failed test appears to be some random gdb-internal integer overflow related to debuginfod, not caused by this PR.

image

pwndbg/commands/exithandlers.py
```diff
@@ -0,0 +1,181 @@
...
+import pwndbg.commands
+import pwndbg.dintegration
+import pwndbg.emu.emulator
+
+
+def rol(val: int, amount: int) -> int:
+    amount %= pwndbg.aglib.arch.ptrbits
+    return (
+        (val << amount) | val >> (pwndbg.aglib.arch.ptrbits - amount)
+    ) & pwndbg.aglib.arch.ptrmask
```
RocketMaDev: pwntools already provide `rol` and `ror`, just use them

pwndbg/commands/exithandlers.py
```diff
@@ -0,0 +1,181 @@
...
+    print(f"__exit_funcs: {pwndbg.color.memory.get(exit_function_list)}")
+    print("Registered handlers:")
+    exit_handlers = []
+    while True:
+        if exit_function_list == 0:
+            break
+        num_handlers = pwndbg.aglib.memory.read_pointer_width(
+            exit_function_list + pwndbg.aglib.arch.ptrsize
+        )
+        for i in reversed(range(num_handlers)):
```
RocketMaDev: why in reverse order?

pwndbg/commands/exithandlers.py
```diff
@@ -0,0 +1,181 @@
...
+    exit_function_list = pwndbg.aglib.memory.read_pointer_width(exit_funcs_ptr)
+    print(f"__exit_funcs: {pwndbg.color.memory.get(exit_function_list)}")
+    print("Registered handlers:")
+    exit_handlers = []
+    while True:
+        if exit_function_list == 0:
+            break
+        num_handlers = pwndbg.aglib.memory.read_pointer_width(
+            exit_function_list + pwndbg.aglib.arch.ptrsize
+        )
```
RocketMaDev: `struct exit_function_list *next` is not considered?

pwndbg/commands/exithandlers.py
h4kk4p4kk4: good catch, wasn't sure how commonly used pwnlib was in the codebase, will switch to this

pwndbg/commands/exithandlers.py
h4kk4p4kk4: they get called in reverse order in __run_exit_handlers (see https://elixir.bootlin.com/glibc/glibc-2.43/source/stdlib/exit.c):

```
while (cur->idx > 0)
	{
	  struct exit_function *const f = &cur->fns[--cur->idx];
```

I though it made more sense to show them in the order they'll be called rather than in memory order

pwndbg/commands/exithandlers.py
```diff
@@ -0,0 +1,181 @@
...
+        emulator.until_jump()  # call __run_exit_handlers
+        esp = emulator.read_register("esp")
+        if esp is None:
+            print(pwndbg.color.message.error("Failed to read ESP register"))
+            return None
+        exit_funcs_ptr_bytes = emulator.read_memory(esp + 8, 4)
+        if exit_funcs_ptr_bytes is None:
+            print(pwndbg.color.message.error("Failed to read &__exit_funcs from stack"))
+            return None
+        exit_funcs_ptr = int.from_bytes(exit_funcs_ptr_bytes, "little")
```
RocketMaDev: use unpack series functions from pwntools instead

pwndbg/commands/exithandlers.py
h4kk4p4kk4: I do consider it, line 175 traverses the linked list.

```
        exit_function_list = pwndbg.aglib.memory.read_pointer_width(exit_function_list)
```

I know the code here is slightly weirdly structured but its the same as the glibc source code (null check at the start of the loop, update at the end)

RocketMaDev: Oh, I see. Adding a comment may be better.

RocketMaDev:
What about tls_dtors? You can refer to pwntools dev branch to see the structure I wrote.

This feature may require debug symbol to perform. `tls_dtor_list` is a thread-local variable located via debug symbol. It will be much harder to resolve without debug symbol.

h4kk4p4kk4:
yeah, I'd be happy to also read tls dtors as part of this, but it is probably slightly harder to heuristically locate tls_dtor_list. could at a stretch try disassembling or emulating __call_tls_dtors, which seems to be exported in stripped libc. will take a look in a few hours.

disconnect3d:
> This feature may require debug symbol to perform. tls_dtor_list is a thread-local variable located via debug symbol. It will be much harder to resolve without debug symbol.

Fwiw we know where tls is; and if its a linked list we can probably figure out via a pattern "ptr -> ptr -> 0"?

Or maybe the TLS has some "standard structure" though I doubt but some heuristic may work?

One thing I am wondering here is naming. Maybe this should be `libc --exitfuncs`, heh...

But I am probably fine with naming it `exitfuncs` or `exithandlers` hm..

cc: @pwndbg/vip-contributors thoughts on naming?

RocketMaDev:
> But I am probably fine with naming it `exitfuncs` or `exithandlers` hm..

Since every libc should implement something like exitfuncs, perhaps we can leave this job to others if anyone would like to implement recognizing exitfuncs in other libc. (Though this command should be limited to glibc in current stage)

h4kk4p4kk4:
I've implemented tls_dtor_list resolution for x86-64 by emulating __call_tls_dtors until we reach an instruction with `fs` as one of its operands. This should be a good enough heuristic to work for pretty much all glibc builds, but I need to implement the same for other architectures. 

On the naming point, I personally prefer exitfuncs but exithandlers is more libc agnostic as the name __exit_funcs is glibc specific, if a future goal is to support other libcs. 

image

pwndbg/commands/exithandlers.py
```diff
@@ -0,0 +1,288 @@
...
+                pwndbg.color.message.error(
+                    f"Failed to read second argument register ({second_arg_reg})"
+                )
+            )
+            return None
+    return exit_funcs_ptr
+
+
+def _get_tls_dtor_list_from_emulator() -> int | None:
+    call_tls_dtors = pwndbg.aglib.symbol.lookup_symbol("__call_tls_dtors")
```
RocketMaDev, line 367: This is a `GLIBC_PRIVATE` function exposed in public APIs. Interesting.

h4kk4p4kk4: Yeah it is a bit weird, the declaration in stdlib.h is marked

```
extern void __call_tls_dtors (void);
libc_hidden_proto (__call_tls_dtors)
```

but the actual definition in cxa_thread_atexit_impl.c doesn't have `attribute_hidden`. I'm not too clear on what the rules are for glibc symbol visibility but I suspect this is why.

RocketMaDev: Sometimes gcc call them when compiling the code. For example, `__stack_chk_fail` is a `GLIBC_PRIVATE` public function.

pwndbg/commands/exithandlers.py
```diff
@@ -0,0 +1,288 @@
...
+            return None
+        pointer_guard_offset = pwndbg.aglib.arch.ptrsize * 6
+        return pwndbg.aglib.memory.read_pointer_width(tls_addr + pointer_guard_offset)
+    if pwndbg.aglib.arch.name in {"aarch64", "arm"}:
+        pointer_chk_guard = pwndbg.aglib.symbol.lookup_symbol(
+            "__pointer_chk_guard"
+        ) or pwndbg.aglib.symbol.lookup_symbol("__pointer_chk_guard_local")
+        if pointer_chk_guard is None:
+            print(
+                pwndbg.color.message.error(
```
k4lizen: nit: `from pwndbg.color import message` is more common in the codebase

pwndbg/commands/exithandlers.py
```diff
@@ -0,0 +1,288 @@
...
+def _get_pointer_guard() -> int | None:
+    if pwndbg.aglib.arch.name in {"x86-64", "i386"}:
+        tls_addr = (
+            pwndbg.aglib.tls.find_address_with_register()
+            or pwndbg.aglib.tls.find_address_with_pthread_self()
+        )
+        if tls_addr is None:
+            print(pwndbg.color.message.error("Failed to get TLS address"))
+            return None
+        pointer_guard_offset = pwndbg.aglib.arch.ptrsize * 6
```
k4lizen: use the `tcbhead_t` type to find the offset if its present (i.e. the libc has debug info), and only do this otherwise. add a comment linking to the type (e.g. https://elixir.bootlin.com/glibc/glibc-2.43.9000/source/sysdeps/x86_64/nptl/tls.h#L52).

the offset is not really 6 pointers, i would use `pwndbg.aglib.typeinfo.sint.sizeof` to get the size of the `int` type to be more portable.

write a comment on what is arch-specific here

pwndbg/commands/exithandlers.py
```diff
@@ -0,0 +1,288 @@
...
+        return string
+
+
+def _ptr_demangle(pointer_guard: int, ptr: int) -> int:
+    if pwndbg.aglib.arch.name in {"x86-64", "i386"}:
+        return (
+            typing.cast(int, ror(ptr, pwndbg.aglib.arch.ptrsize * 2 + 1, pwndbg.aglib.arch.ptrbits))
+            ^ pointer_guard
+        ) & pwndbg.aglib.arch.ptrmask
+    return ptr ^ pointer_guard
```
k4lizen: link to source. is this actually accurate on all other architectures? if not, you can add an assert / error message / exception and catch it.

pwndbg/commands/exithandlers.py
```diff
@@ -0,0 +1,340 @@
...
+    obj: int
+    map: int
+
+    def __str__(self) -> str:
+        decomp_stack_vars = pwndbg.dintegration.manager.get_stack_var_dict_all()
+        string = f"{pwndbg.color.memory.get(self.address)}: "
+        string += pwndbg.color.memory.get_address_and_symbol(self.func, decomp_stack_vars)
+        string += f" [obj = {pwndbg.chain.format(self.obj)}"
+        string += f", map = {pwndbg.color.memory.get(self.map)}]"
+        return string
```
k4lizen, line 139: could we make the output of this command column aligned by using `Table.grid` similarly as in `context_backtrace()`? I think its a small thing that would make things more pretty :)

same comment for ExitFunctionEntry

(but I suppose you can't implement it as a __str__ method then, since you need to gather them all and only then create the string)

pwndbg/commands/exithandlers.py
```diff
@@ -0,0 +1,340 @@
...
+            "__pointer_chk_guard"
+        ) or pwndbg.aglib.symbol.lookup_symbol("__pointer_chk_guard_local")
+        if pointer_chk_guard is None:
+            print(
+                pwndbg.color.message.error(
+                    "Could not find __pointer_chk_guard or __pointer_chk_guard_local symbols"
+                )
+            )
+            return None
+        return pwndbg.aglib.memory.read_pointer_width(int(pointer_chk_guard))
```
k4lizen: ig you could also use `pointer_chk_guard.dereference()`

pwndbg/commands/exithandlers.py
```diff
@@ -0,0 +1,340 @@
...
+    for entry in exit_handlers:
+        print(str(entry))
+
+    tls_dtor_list = (
+        pwndbg.aglib.symbol.lookup_symbol("tls_dtor_list") or _get_tls_dtor_list_from_emulator()
+    )
+    if tls_dtor_list is None:
+        print(pwndbg.color.message.error("Failed to locate tls_dtor_list"))
+        return
+    print(f"\ntls_dtor_list: {pwndbg.color.memory.get(tls_dtor_list)}")
```
k4lizen: your screenshot does not show this output.

also, for both cases, if you can find the exit handlers but none are actually registered, you should print something like "No __exit_funcs handlers registered." instead of an empty "Registered __exit_funcs handlers:"

pwndbg/commands/exithandlers.py
```diff
@@ -0,0 +1,340 @@
...
+
+parser = argparse.ArgumentParser(description="List currently registered glibc exit handlers.")
+
+
+@pwndbg.commands.Command(
+    parser, category=pwndbg.commands.CommandCategory.LINUX, aliases=["exitfuncs"]
+)
+@pwndbg.commands.OnlyWhenRunning
+@pwndbg.aglib.proc.OnlyWithArch(["x86-64", "i386", "aarch64", "arm"])
+def exithandlers() -> None:
```
k4lizen, line 469: exit if libc.which() is not glibc or unknown

pwndbg/commands/exithandlers.py
```diff
@@ -0,0 +1,340 @@
...
+        # update to cur_exit_function_list->next
+        cur_exit_function_list = pwndbg.aglib.memory.read_pointer_width(cur_exit_function_list)
+    return handlers
+
+
+def _list_tls_dtors(pointer_guard: int, tls_dtor_list: int) -> list[_TlsDtorEntry]:
+    cur_tls_dtor = pwndbg.aglib.memory.read_pointer_width(tls_dtor_list)
+    dtors: list[_TlsDtorEntry] = []
+    while cur_tls_dtor != 0:
+        func = _ptr_demangle(pointer_guard, pwndbg.aglib.memory.read_pointer_width(cur_tls_dtor))
```
k4lizen: link to source and write the struct here since its small and important

```c
struct dtor_list
{
  dtor_func func;
  void *obj;
  struct link_map *map;
  struct dtor_list *next;
};
```

pwndbg/commands/exithandlers.py
```diff
@@ -0,0 +1,340 @@
...
+            pwndbg.color.message.error(
+                "Failed to get TLS offset to tls_dtor_list from emulating __call_tls_dtors"
+            )
+        )
+        return None
+    return tls_addr + offset
+
+
+def _list_exit_handlers(pointer_guard: int, exit_funcs: int) -> list[_ExitFunctionEntry]:
+    handlers: list[_ExitFunctionEntry] = []
```
k4lizen, line 402: link to source, put
```
struct exit_function_list
  {
    struct exit_function_list *next;
    size_t idx;
    struct exit_function fns[32];
  };
```
in comment here.

pwndbg/commands/exithandlers.py
```diff
@@ -0,0 +1,340 @@
...
+    cur_exit_function_list = exit_funcs
+    while True:
+        if cur_exit_function_list == 0:
+            break
+        num_handlers = pwndbg.aglib.memory.read_pointer_width(
+            cur_exit_function_list + pwndbg.aglib.arch.ptrsize
+        )
+        for i in reversed(range(num_handlers)):
+            exit_function_struct_base = cur_exit_function_list + pwndbg.aglib.arch.ptrsize * (
+                2 + 4 * i
```
k4lizen: typeinfo.size_t.sizeof (and in general, fix this type of thing everywhere)

k4lizen (approved):
Thanks for the PR! I think this is pretty cool!

Here are some comments, nothing too major though.

This will need a test (https://pwndbg.re/dev/contributing/writing-tests/).

pwndbg/commands/exithandlers.py
k4lizen: good idea! (write this in a comment :p )

pwndbg/commands/exithandlers.py
h4kk4p4kk4: done, have left the offset as 6 * ptrsize for now as i'm otherwise not sure how to represent the struct padding of multiple_threads

pwndbg/commands/exithandlers.py
```diff
@@ -0,0 +1,399 @@
...
+    if pwndbg.aglib.arch.name in {"x86-64", "i386"}:  # x86 stores pointer_guard in TLS
+        tls_addr = (
+            pwndbg.aglib.tls.find_address_with_register()
+            or pwndbg.aglib.tls.find_address_with_pthread_self()
+        )
+        if tls_addr is None:
+            print(message.error("Failed to get TLS address"))
+            return None
+
+        # https://elixir.bootlin.com/glibc/glibc-2.43.9000/source/sysdeps/x86_64/nptl/tls.h#L42
```
RocketMaDev: Please use `https://elixir.bootlin.com/glibc/glibc-2.43/source/sysdeps/x86_64/nptl/tls.h#L42` rather than  `https://elixir.bootlin.com/glibc/glibc-2.43.9000/source/sysdeps/x86_64/nptl/tls.h#L42`. The `.9000` tag is not a stable release tag.

pwndbg/commands/exithandlers.py
h4kk4p4kk4: nice idea, how would it look? something like

```
| struct address | func | obj | map |
```

?

then for __exit_funcs I guess

```
| struct address | flavor | fn | arg | dso_handle |
```

but this feels like it could be a bit weird with different exit_function flavors having some of the fields be irrelevant, or in the case of ef_at, named differently

also I'm not sure how arg being rendered with chain.format would look. could replace it with color.memory.get but I think it's useful having a telescope view shown

h4kk4p4kk4:
I'm going to be offline for a week, so won't be able to work on this until I'm back, but have added tests and addressed most of the review comments aside from https://github.com/pwndbg/pwndbg/pull/3973#discussion_r3415518378.  

Only other major stuff I can think of still todo would be implementing support for musl (https://git.musl-libc.org/cgit/musl/tree/src/exit/atexit.c), which might be a bit of a chore and worth its own PR further down the line, and testing and implementing the more brittle parts (like all the emulated function parsing stuff) for more architectures.

pwndbg/commands/exithandlers.py
k4lizen: I would just leave it the same as you have it, I think it looks good, just having the output aligned.

image

My red markings here should be one below the other.

k4lizen (approved):
Awesome, thanks!

PR #3984, opened 2026-06-26
fix(heap): colorize realloc pointer and fix realloc(ptr, 0) in heap t…

### LINKED ISSUE #3090: heap tracker: color or mark pointers?

hiratinspace:
Completes the heap-tracker pointer coloring (#3090): the realloc() return pointer is now colorized via colorize_ptr() like malloc/free, so an allocation can be visually matched with its free.

Also fixes two bugs on the realloc(ptr, 0) path that the original coloring change left behind:
  - an AttributeError from referencing a non-existent self.freed_pointer, and
  - a missing exit_memory_management() that left the tracker flagged as in-progress, breaking the next allocation.

Adds GDB tests (the tracker is GDB-only) covering the colorized realloc output and the realloc(ptr, 0) path. Verified the tests fail without the fix and pass with it, and `./lint.sh` passes.

Closes #3090

## Screenshot
Screenshot 2026-06-25 at 10 48 14 PM


pwndbg/gdblib/ptmalloc2_tracking.py
```diff
@@ -514,12 +514,13 @@ def stop(self) -> bool:
             # There's no right way to handle realloc(..., 0). C23 says it's
             # undefined behavior, and prior versions say it's implementation-
             # defined. Either way, print a warning and do nothing.
-            ptr_str = self.tracker.colorize_ptr(self.freed_pointer)
+            ptr_str = self.tracker.colorize_ptr(freed_pointer)
             print(
                 message.warn(
                     f"[-] realloc({ptr_str}, {requested_size}) ignored, as realloc(0, ...) is implementation defined"
```
disconnect3d:
```suggestion
                    f"[-] realloc({ptr_str}, {requested_size}) ignored, as realloc(..., 0) is implementation defined"
```

k4lizen:
put the test into `dbg/` and guard it behind a pytest skip with `pwndbg.dbg.name() == DebuggerType.GDB`

disconnect3d:
Closing this for now due to @hiratinspace inactivity/no response to our feedback.

Issue #3092, opened 2025-06-07
heap tracker: display object type or a symbol from backtrace?


disconnect3d:
Currently the `track-heap enable` shows us only that a malloc/calloc/free etc. happened and which size/address it was passed or returned, but it would be awesome to show where in the program the allocation or deallocation happened - or - for what kind of object.

This would need to be smart somehow so it is useful, as we probably don't want to display the whole backtrace for each malloc/calloc/free etc. but just one symbol - like "SomeObject::SomeMethod".

This could also work e.g. only if sources are available or maybe the user should pass some regex for the backtrace and we should only show a symbol that matches?

disconnect3d:
When I think about it more: since we track malloc/free etc and we can figure out their origin, maybe we can use this tracking information to mark those allocation in `vis-heap-chunks`?

If we also find a way to figure out the object type somehow - either by user specifying it somehow once - or - via some LLM+MPC magic - or with some heuristics (e.g. C++ types?) - we could display this as well in vis-heap-chunk!

dbgbgtf1:
I just tried track-heap, but it doesn't work in my case.  
Can you explain what happened here or teach me how to debug this?
I don't know how to debug pwndbg

![Image](https://github.com/user-attachments/assets/eba1e413-c29d-4582-bd83-453dc9ebb92b)

mileaage:
Sounds like a splendid idea. I'd definitely like to see this implemented. However, achieving this cross-platform could impose significant challenges due to fundamental differences in how operating systems and compilers manage memory and debug information.

PR #3986, opened 2026-06-26
feat(heap-tracker): annotate alloc/realloc/free events with calling symbol (--where)

Builder106:
## Summary

Adds a `--where` flag to `track-heap enable` that annotates each `malloc`, `calloc`, `realloc`, and `free` event with the nearest program-code symbol that triggered it, printed on its own aligned line beneath the event:

```
[*] malloc(0x18) -> 0x..., 0x20 bytes real size
    @ do_alloc+16
[*] realloc(0x..., 0x30) -> 0x..., 0x40 bytes real size
    @ do_realloc+28
[*] free(0x...)
    @ main+52
```

Falls back to the raw address when no symbol resolves. Closes `#3092`.

## Implementation

`program_caller_frame()` — new function in `gdblib/ptmalloc2_tracking.py` that walks the GDB frame chain outward from `gdb.newest_frame()`, returning the first frame whose PC falls inside an executable page of the program binary (not libc/ld). This generalizes the existing `in_program_code_stack()` helper, which is now a one-line wrapper around it.

`caller_symbol()` — calls `pwndbg.aglib.symbol.resolve_addr(int(frame.pc()))` on that frame. `resolve_addr` is already `@cache_until("objfile")`-cached and returns `name+offset` strings directly; falls back to the raw address when no symbol resolves.

`Tracker.show_symbols` flag is threaded through `install()` and `Tracker.__init__` to the three print sites (`AllocExitBreakpoint`, `ReallocExitBreakpoint`, `FreeExitBreakpoint`). The symbol is colored with `pwndbg.color.memory.c.code` — the same color pwndbg already uses elsewhere for addresses inside executable memory — rather than an arbitrary color.

## Tests

Two new tests in `tests/library/gdb/tests/heap/test_track_heap_symbols.py`, backed by `tests/binaries/host/heap_tracker_symbols.native.c`:

- `test_track_heap_symbols_annotates_caller` — verifies `@ do_alloc`, `@ do_realloc`, and `@ main` all appear in the output when `--where` is passed (ANSI-stripped before assertion).
- `test_track_heap_without_symbols_is_unchanged` — verifies no `@` annotation appears in default mode.

Both pass on aarch64. The pre-push hook (`ruff` + `mypy` + `vermin` + custom lint) is clean.

Screenshot 2026-07-02 at 11 19 28 PM


patryk4815:
please provide screenshots

Builder106:
sowwy 🥺

Builder106:
@patryk4815 Screenshot added ✅

k4lizen:
i think i would take inspiration from `strace` on how to format this output, i like the idea very much but it looks a bit wonky

some ideas:
+ symbol should be different color from rest of output
+ print addr if symbol is not available
+ use `--where` instead of `--symbols`
+ use `@` instead of `<--`
+ maybe print out two lines instead of one, one for addr+symbol, one for alloc, so there is better alignment

k4lizen:
btw why did you create a test in the gdb/ folder?

Builder106:
> btw why did you create a test in the gdb/ folder?

Yo, @k4lizen. Thanks for the quick response. `track-heap` hooks `gdb.events.inferior_call_post`, which is a GDB-only API, so it can't go in `tests/library/dbg/` with the other heap tests since that tree runs on both GDB and LLDB. I followed the same pattern as `test_inferior_call_cache.py`, which already lives in `tests/library/gdb/tests/heap/` for the exact same reason.

Builder106:
> i think i would take inspiration from `strace` on how to format this output, i like the idea very much but it looks a bit wonky
> 
> some ideas:
> 
> * symbol should be different color from rest of output
> * print addr if symbol is not available
> * use `--where` instead of `--symbols`
> * use `@` instead of `<--`
> * maybe print out two lines instead of one, one for addr+symbol, one for alloc, so there is better alignment

Heyyy @k4lizen. 
Thanks for the feedback. Renamed `--symbols` to `--where`, swapped `<--` for `@`, and put the annotation on its own line under each event so malloc/realloc/free all line up.

Also added the address fallback you mentioned. If `caller_symbol()` can't resolve a name it just prints the raw address instead of nothing.

And swapped `pwndbg.color.bold` for `pwndbg.color.memory.c.code` on the symbol. It's the same color pwndbg already uses for addresses in executable memory, so it felt more consistent than just picking something random.

Updated the screenshot too. Let me know if anything's still off.

k4lizen (changes requested):
oke oke, this looks pretty nice now :>

I would rename `show_symbols` to something like `show_location` now to be inline with the flag changes.

> track-heap hooks gdb.events.inferior_call_post, which is a GDB-only API, so it can't go in tests/library/dbg/ with the other heap tests since that tree runs on both GDB and LLDB. I followed the same pattern as test_inferior_call_cache.py, which already lives in tests/library/gdb/tests/heap/ for the exact same reason.

That is fair rationale. That said, all new tests should go to tests/library/dbg , and if the functionality is GDB-only, you can gate the test with an if-statement and a pytest skip (`pwndbg.dbg.name() == DebuggerType.GDB`). This way if we add this functionality to LLDB later, we don't need to port the test to the new system, but just get rid of the if-statement.

Also the current test is not testing much, ideally it should check that the caught location is the actual location where the malloc is.

k4lizen (approved):
looks good to me, is this all or do you want to add anything else?

PR #3988, opened 2026-06-28
feat: add initial Android libc Bionic support

zeguilherme1:
The main goal of this pull request is to start the improvement of Android debugging in Pwndbg.

Modifications and additions:

bionic.py: Implemented the core logic for detecting Android Bionic libc. It provides essential information including libc version, linking status, relevant documentation URLs, and symbol information.

facade.py: Implemented the logic for path resolution for Android. By using vmmap and filtering based on specific platform paths (such as /apex/ and /system/), it correctly resolves the base load addresses for libc and the loader.

How to test:

Using pwndbg-lldb, run:
`platform select remote-android`
`platform connect connect://localhost:port`
`attach {PID}`

After sucessfull attach, run `libc` command to get all the information

image


pwndbg/libc/facade.py
```diff
@@ -330,7 +331,21 @@ def addr() -> int:
...
+    for page in pwndbg.aglib.vmmap.get():
+        obj = page.objfile
+
+        if obj and obj.endswith(target_name):
+            # Android shared libraries are typically mapped under /apex/ or /system/.
+            # Also accept generic /lib paths for environments with different layouts.
+            if "/apex/" in obj or "/system/" in obj or "/lib" in obj:
+                yes = page.start
+                break
+
```
k4lizen, line 348: why is this change necessary? does `filepath()` not consistently return the correct path to the libc? 

i see two issues with this change:
1. it leaks libc-specific code to facade.py which is bad
2. it invalidates the contract that " If the program is statically linked this will return the address of the main objfile."

pwndbg/libc/facade.py
```diff
@@ -347,7 +362,18 @@ def loader_addr() -> int:
...
+    target_name = os.path.basename(str(loader_filepath()))
+
+    yes = None
+    for page in pwndbg.aglib.vmmap.get():
+        obj = page.objfile
+
+        if obj and obj.endswith(target_name):
+            yes = page.start
+            break
+
```
mbrla0, line 376: Can you add some context for this change? Why would you check only against the basename here?

Additionally, this change breaks in the - admittedly convoluted, but still valid - scenario in which an application maps a file that has the same name as its loader before the loader itself in memory.

pwndbg/libc/bionic.py
```diff
@@ -0,0 +1,103 @@
...
+
+    https://android.googlesource.com/platform/bionic/+/refs/heads/main/libc/bionic/__libc_init_main_thread.cpp
+    https://android.googlesource.com/platform/bionic/+/master/libc/libc.map.txt
+    """
+    return (
+        pwndbg.aglib.symbol.lookup_symbol(
+            "android_reset_stack_guards", objfile_endswith=mapping_name
+        )
+        is not None
+    )
```
mbrla0, line 25: Won't checking only against `android_reset_stack_guards` cause this function to return `false` - and thus make the facade reject the hypothesis that a given libc is Bionic - on Android 11 and earlier?

This is alright for now, but we should probably look into this.

pwndbg/libc/bionic.py
```diff
@@ -0,0 +1,103 @@
...
+        24: "preadv",  # Android 7.0
+        23: "error_at_line",  # Android 6.0
+        21: "epoll_create1",  # Android 5.0
+    }
+    for api_level, api_symbol in api_symbols.items():
+        if (
+            pwndbg.aglib.symbol.lookup_symbol(api_symbol, objfile_endswith=libc_filepath)
+            is not None
+        ):
+            return (api_level,)
```
mbrla0: There's an interesting bit of nuance here. Python's `dics` [have preserved insertion order since Python 3.7](https://mail.python.org/pipermail/python-dev/2017-December/151283.html), so this code is correct. However, I feel the dependency on the order could be made more explicit. Perhaps by doing something like this (Please test it if you do include it!):

```suggestion
    api_symbols = [
        (37, "sched_getattr"),  # Android 17
        (36, "lchmod"),  # Android 16
        (35, "android_crash_detail_register"),  # Android 15
        (34, "close_range"),  # Android 14
        (33, "backtrace"),  # Android 13
        (31, "android_reset_stack_guards"),  # Android 12
        (29, "android_get_device_api_level"),  # Android 10
        (28, "__ppoll64_chk"),  # Android 9
        (26, "catclose"),  # Android 8.0
        (24, "preadv"),  # Android 7.0
        (23, "error_at_line"),  # Android 6.0
        (21, "epoll_create1"),  # Android 5.0
    ]
    for api_level, api_symbol in api_symbols:
        if (
            pwndbg.aglib.symbol.lookup_symbol(api_symbol, objfile_endswith=libc_filepath)
            is not None
        ):
            return (api_level,)
```

Keep in mind this is only my personal preference, not a matter of correctness.

pwndbg/libc/bionic.py
```diff
@@ -0,0 +1,103 @@
...
+
+    return (-1, -1)
+
+
+def has_internal_symbols(libc_filepath: str) -> bool:
+    return (
+        pwndbg.aglib.symbol.lookup_symbol("__libc_globals", objfile_endswith=libc_filepath)
+        is not None
+    )
```
mbrla0, line 63: Since which version has this symbols been present? Would be nice to have a comment saying that. Can be as simple as "Since Android `<version>`"

pwndbg/libc/bionic.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def has_debug_info() -> bool:
+    return pwndbg.aglib.typeinfo.load("scudo::Chunk::UnpackedHeader") is not None
```
mbrla0, line 68: Same thing here as in `has_internal_symbols`. Since when has this type been present in Bionic?

pwndbg/libc/bionic.py
```diff
@@ -0,0 +1,103 @@
...
+        git="https://android.googlesource.com/platform/bionic.git",
+    )
+
+
+def verify_ld_candidate(mapping_name: str) -> bool:
+    return False
+
+
+def libc_same_as_ld() -> bool:
+    return False
```
mbrla0, line 105: I'm not sure on my `libc` subsystem internals so @k4lizen might be better suited to help with this than me. Does Android's Bionic not implement dynamic loading? If so, should `libc_same_as_ld` not return `true` and `verify_ld_candidate` be implemented to some extent?

It's alright if that's intended to be addressed in a future PR, just wanted to know.

mbrla0 (commented):
I've added a few observations. Don't consider any of them blocking, or that they need to addressed right now - it's alright if you want to address them in a later PR! -  but I think they're important to point out.

mbrla0:
Also, it seems this has caused some regressions in testing! Please check those out.

pwndbg/libc/bionic.py
k4lizen: the way it is currently is the intended way.

`libc_same_as_ld()` is just asking whether they get mapped to the same object file, which some libc's do, and doesn't affect whether they do or do not support dynamic loading of shared libraries.

If `verify_libc_candidate()` confirms that the current process is using bionic, then the mapped ld-linux will be found via a regex on the objfile name. This is not as robust as it can be, but implementing an accurate `verify_ld_candidate` is really hard, so it is not required. If we run into issues with our detection being inaccurate in the future, we may look into tightening it up.

mbrla0: Alrighty, makes sense to me. Thanks!

k4lizen:
the binder parcel command should likely be a seperate PR

zeguilherme1:
Hi guys, the filepath() method was not getting successfully the path of Android libc or loader because when doing remote debugging using lldb-server, for example, it returns the local module cache on the host machine but the vmmap fetched from the target contains the remote path. Because the directory prefixes differ, the path match fails and pwndbg cannot find the libc and loader base address

I added a fallback logic aiming to solve this:

First, it tries the exact path match using pwndbg.aglib.vmmap.named_region_start(target_path)

If yes stills None, it will try to find the base address searching in the mapped memory by the basename.

Thanks for the reviews!!

mbrla0:
> Hi guys, the filepath() method was not getting successfully the path of Android libc or loader because when doing remote debugging using lldb-server, for example, it returns the local module cache on the host machine 

Wouldn't using `SBModule::GetPlatformFileSpec` solve that?

zeguilherme1:
> > Hi guys, the filepath() method was not getting successfully the path of Android libc or loader because when doing remote debugging using lldb-server, for example, it returns the local module cache on the host machine
> 
> Wouldn't using `SBModule::GetPlatformFileSpec` solve that?

Tried to follow this way but don't get a successfully result, it still returns the local module cache.

mbrla0:
Hmm, I see. @zeguilherme1 can you add that as a comment? Especially the potential for false positives. @k4lizen I think we can consider moving on with this, as long as we take note of the limitations, and Ze's willing to polish it later. There's a lot of work whose upstreaming is blocked by this PR.

pwndbg/libc/bionic.py
k4lizen: the phone that i use is android 10 :<

k4lizen:
so the problem here is that filepath() which calls __get_libc() which calls module_section_locations() which calls
self.target.GetModuleAtIndex(i) returns objfile paths to some local lldb cache?

but when you actually iterate vmmap() which calls get_known_pages() which calls lldb.SBMemoryRegionInfo().GetMemoryRegionAtIndex(i, region).GetName() you get the "correct" remote paths?

a couple of thoughts come to mind

first of all it seems like we could just call vmmap() instead of module_section_locations() in __get_libc() and this would fix the
issue.

second, i don't understand why they differ, and i feel like this is likely to be a source of bugs in the future,
neither of these functions document this particularity.

i also didn't know there was a local module cache, what do the paths look like for it?

so this is a problem for all libc's with LLDB? doesn't seem to be android-specific? do we not repro this for
glibc remote debugging with lldb?

i feel like:
1. if we move on from this now it seems someone else will have to pick it up (matt <333333)
2. i'm fine with moving on from this if we leave some trace in the code that there are issues

so i would like the comments matt said should be there, and these two snippets should be pulled out to a function since
they are copy-paste.

its subopt that verify_libc_candidate() / this detection won't work on my phone since my phone is too old, if this could be fixed quikcly by choosing a different symbol that would be nice, if not okay we can move on

> it invalidates the contract that " If the program is statically linked this will return the address of the main objfile."

i don't remember why i said this, if you statically compile on android and debug, does it return the address of the main objfile?

mbrla0:
> if we move on from this now it seems someone else will have to pick it up (matt <333333)

@k4lizen I can handle that.

mbrla0:
@zeguilherme1 Can you address the rest of the stuff he's said? If so I feel we could merge this before the deadline on your final report, maybe?

zeguilherme1:
Sure, I will make the necessary changes and upload it for review. Thanks guys

mbrla0 (approved):
Good enough for now. Should probably still have documented the shortcomings of using `basename` as part of the search, but it's alright.

PR #3989, opened 2026-06-29
slab: fix slab info consistency

navytuner:
Small cleanup of the `slab info` command output:

1. `Slab @` now shows the `struct slab` address (with the data/virt address moved to brackets), making it consistent with the other @ lines (`Slab Cache @`, `kmem_cache_cpu @`, `kmem_cache_node @`), which all print the address of their kernel struct.
2. Move `nr_partial/min_partial` from the `kmem_cache_node` header to the Partial Slabs line, where they more naturally describe the partial list and we can avoid duplicate `nr_partial` output.

image
image

No functional changes — output formatting only.

### DIFF
--- a/pwndbg/commands/slab.py
+++ b/pwndbg/commands/slab.py
@@ -225,16 +225,12 @@ def print_node_cache(node_cache: NodeCache, verbose: bool) -> None:
     )
     # https://elixir.bootlin.com/linux/v6.13/source/mm/slub.c#L3140
     indent.print(
-        f"{indent.prefix('kmem_cache_node')} @ {indent.addr_hex(address)} [NUMA node {node}, nr_partial/min_partial: {indent.aux_hex(nr_partial)}/{indent.aux_hex(min_partial)}]:"
+        f"{indent.prefix('kmem_cache_node')} @ {indent.addr_hex(address)} [NUMA node {node}]:"
     )
     with indent:
         partial_slabs = node_cache.partial_slabs
-        if not partial_slabs:
-            indent.print("Partial Slabs: (none)")
-            return
-
         indent.print(
-            f"{indent.prefix('Partial Slabs')} [nr_partial: {indent.aux_hex(len(partial_slabs))}]"
+            f"{indent.prefix('Partial Slabs')} [nr_partial/min_partial: {indent.aux_hex(nr_partial)}/{indent.aux_hex(min_partial)}]"
         )
         with indent:
             for slab in partial_slabs:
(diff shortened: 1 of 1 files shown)


pwndbg/commands/slab.py
```diff
@@ -225,16 +225,12 @@ def print_node_cache(node_cache: NodeCache, verbose: bool) -> None:
...
     # https://elixir.bootlin.com/linux/v6.13/source/mm/slub.c#L3140
     indent.print(
-        f"{indent.prefix('kmem_cache_node')} @ {indent.addr_hex(address)} [NUMA node {node}, nr_partial/min_partial: {indent.aux_hex(nr_partial)}/{indent.aux_hex(min_partial)}]:"
+        f"{indent.prefix('kmem_cache_node')} @ {indent.addr_hex(address)} [NUMA node {node}]:"
     )
     with indent:
         partial_slabs = node_cache.partial_slabs
-        if not partial_slabs:
-            indent.print("Partial Slabs: (none)")
-            return
```
k4lizen, line 234: i think we should keep this

pwndbg/commands/slab.py
```diff
@@ -131,7 +131,7 @@ def freelist_desc(freelist: Freelist) -> str:
 
 def print_slab(slab: Slab, verbose: bool) -> None:
     indent.print(
-        f"- {indent.prefix('Slab')} @ {indent.addr_hex(slab.virt_address)} [{indent.aux_hex(slab.slab_address)}]:"
+        f"- {indent.prefix('Slab')} @ {indent.addr_hex(slab.slab_address)} [{indent.aux_hex(slab.virt_address)}]:"
```
k4lizen, line 134: hmm but for kmem_cache_cpu it's clear that it is an internal struct, while "Slab" is more of a concept. i'm fine having the ordering either way, but it may trip up new people who are used to the old one.

i guess you found the current print counterintuitive? maybe we could clarify it in some other way

navytuner: I found it counterintuitive because all values after @ are addresses of their struct(e.g., `Slab Cache @` and `kmem_cache_node @`)

pwndbg/commands/slab.py
jxuanli: yea i agree, why are we removing this?

jxuanli:
this pr is fine, see comment above

pwndbg/commands/slab.py
navytuner: yes I reverted :)

PR #3993, opened 2026-07-01
Add Unix socket parsing regression test

### LINKED ISSUE #1551: lib/net.py: more functionality and unit tests
* The lib/net.py code could have unit tests implemented, e.g. for https://github.com/pwndbg/pwndbg/pull/1545#discussion_r1090028340
* It could have more features. For example, we could find out the other end of Unix Domain Sockets and show which other processes do use them: this can be super useful for knowing that e.g. a given socket is related to X11 or that a given socket is used for IPC between this and that process[es]

eshun4:
## What does this PR do?

Adds a focused unit test for `pwndbg.lib.net.unix()` covering Unix socket paths that contain a carriage return (`\r`).

## Why was this PR needed?

Issue #1551 mentions adding more unit tests for `lib/net.py`. This test covers the behavior discussed in the earlier abstract Unix Domain Socket work, where the parser intentionally uses `data.split("\n")` instead of `.splitlines()` so socket paths containing `\r` are preserved instead of being split incorrectly.

## What are the relevant issue numbers?

Closes #1551

## Testing

Ran:

```bash
./unit-tests.sh tests/unit_tests/test_net.py -k unix
```

Result:

```text
68 passed, 4 skipped
```

## Checklist

* [x] Tests added for new/changed behavior
* [x] Tests pass locally
* [x] No unrelated changes introduced

### DIFF
--- a/pwndbg/lib/net.py
+++ b/pwndbg/lib/net.py
@@ -218,7 +218,6 @@ def unix(data: str) -> list[UnixSocket]:
 
         u = UnixSocket()
         prev_had_path = len(fields) >= 8
-        prev_had_path = len(fields) >= 8
         if prev_had_path:
             u.path = fields[7]
         u.inode = int(fields[6])
--- a/pwndbg/lib/net.py
+++ b/pwndbg/lib/net.py
@@ -175,6 +175,7 @@ def tcp(data: str, endianness: str) -> list[Connection]:
 def tcp6(data: str, endianness: str) -> list[Connection]:
     return _tcp_parser(data, socket.AF_INET6, endianness)
 
+
 def _is_unix_entry(fields: list[str]) -> bool:
     # A valid /proc/net/unix entry looks like:
     # "0000000000000000: 00000002 00000000 00010000 0001 01 12345 /some/path"
@@ -190,6 +191,7 @@ def _is_unix_entry(fields: list[str]) -> bool:
     except ValueError:
         return False
     return True
+
 
 def unix(data: str) -> list[UnixSocket]:
     if not data:
--- a/pwndbg/lib/net.py
+++ b/pwndbg/lib/net.py
@@ -175,6 +175,21 @@ def tcp(data: str, endianness: str) -> list[Connection]:
 def tcp6(data: str, endianness: str) -> list[Connection]:
     return _tcp_parser(data, socket.AF_INET6, endianness)
 
+def _is_unix_entry(fields: list[str]) -> bool:
+    # A valid /proc/net/unix entry looks like:
+    # "0000000000000000: 00000002 00000000 00010000 0001 01 12345 /some/path"
+    # i.e. a hex slot number ending with ":", five hex fields, a decimal
+    # inode, and an optional path.
+    if len(fields) < 7 or not fields[0].endswith(":"):
+        return False 
+    try:
+        int(fields[0][:-1], 16)
+        for f in fields[1:6]:
+            int(f, 16)
+        int(fields[6])
+    except ValueError:
+        return False 
+    return True
 
 def unix(data: str) -> list[UnixSocket]:
     if not data:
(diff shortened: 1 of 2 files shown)


disconnect3d (changes requested):
The comment in the code - https://github.com/pwndbg/pwndbg/blob/db4502e546d9ea911ebed6eebd96cd2b2d0c8b22/pwndbg/lib/net.py#L184-L189 - says that we want to split with `.split("\n")` instead of `.splitlines()` so that a path containing `\r` is not splitted and it is preserved and reported correctly -- and this is what you test in this PR, it works, and it is fine.

However, if you think about it, there are other cases that should be tested too. And this is when a socket path contains a newline character as well, and this is currently not supported due to us splitting by `\n`.

I would like you to extend the test so that your case contains both \r and \n characters and so that we ensure this is tested properly -- and -- fix the bug so that we do work with this properly.

If you create a binary that opens a socket to `/tmp/sock\nnewline_character` and then cat its `/proc/$tid/net/unix` file, you will see that it contains the newlines as well:

```sh
$ cat /proc/3678745/net/unix
Num       RefCount Protocol Flags    Type St Inode Path
0000000000000000: 00000003 00000000 00000000 0001 03 27785983 /run/systemd/journal/stdout
0000000000000000: 00000003 00000000 00000000 0002 03  9571
0000000000000000: 00000003 00000000 00000000 0001 03 27791062
...
0000000000000000: 00000002 00000000 00010000 0001 01 31956456 /tmp/sock
with_newline
0000000000000000: 00000003 00000000 00000000 0001 03 31568421
...
```

(Note: I truncated the output, the "..." does not exist normally in the output)

And then if you execute `procinfo` on such a program that has this socket opened, its parsing will just crash.

This is a bug we need to fix here and test for it additionally.

disconnect3d:
Useful here - from `man proc_net`:
```
       /proc/net/unix
              Lists the UNIX domain sockets present within the system and their status.  The format is:

              Num RefCount Protocol Flags    Type St Inode Path
               0: 00000002 00000000 00000000 0001 03    42
               1: 00000001 00000000 00010000 0001 01  1948 /dev/printer

              The fields are as follows:

              Num:      the kernel table slot number.

              RefCount: the number of users of the socket.

              Protocol: currently always 0.

              Flags:    the internal kernel flags holding the status of the socket.

              Type:     the  socket  type.   For  SOCK_STREAM  sockets,  this  is 0001; for SOCK_DGRAM sockets, it is 0002; and for
                        SOCK_SEQPACKET sockets, it is 0005.

              St:       the internal state of the socket.

              Inode:    the inode number of the socket.

              Path:     the bound pathname (if any) of the socket.  Sockets in the abstract namespace are included in the list, and
                        are shown with a Path that commences with the character '@'.
```

Heh: https://chatgpt.com/share/6a450db9-b240-83ed-a2d0-3d8fe8f78fd2

disconnect3d:
or: When iterating over the lines if u find unparsable line, add it to the last line's path? I guess that would work and would be very easy to add, huh.

k4lizen:
> Heh: https://chatgpt.com/share/6a450db9-b240-83ed-a2d0-3d8fe8f78fd2

thats pretty funny

eshun4:
Thanks for the detailed review — you were right, my earlier attempt at a newline test just crashed the parser with an IndexError, which confirmed the bug.

I've now implemented the fix using the approach you suggested: when a line doesn't parse as a valid /proc/net/unix entry, it's treated as a continuation of the previous socket's path and glued back with the \n restored. Entry validation checks the expected shape (hex slot number ending with :, five hex fields, decimal inode).
The new test covers a socket path containing both \r and \n, followed by a pathless entry to make sure parsing resumes correctly. Kept the original \r-only regression test as well.
One caveat: since the kernel writes paths verbatim with no escaping, a path fragment could in theory mimic a valid entry line, so this is best-effort. 

Let me know if you's like me to touch on any more updates.

pwndbg/lib/net.py
```diff
@@ -186,15 +203,23 @@ def unix(data: str) -> list[UnixSocket]:
...
 
+        if not _is_unix_entry(fields):
+            if prev_had_path:
+                result[-1].path += "\n" + line
+            continue
+
         u = UnixSocket()
-        if len(fields) >= 8:
+        prev_had_path = len(fields) >= 8
+        prev_had_path = len(fields) >= 8
```
disconnect3d: This is a duplicated line.
```suggestion
```

Issue #3917, opened 2026-05-19
Stack visualisation feature


disconnect3d:
TL;DR: A friend of mine pinged me we should add `vis` for stack: this should go over stack frames and visualise them nicely.

I think this could be `stack --vis` or `stack -v` flag maybe? Or a separate command?

ksaweryr:
(I'm the friend in question) I think `stack --vis` or some argument to make `stackf` recursive could work nicely. Alternatively it could be a command like `vis` but for the stack (`vis-stack-frames`?) where we don't dereference pointers but instead colour different stack frames with different colours.

I'm not sure which of those two approaches would be more useful (pointer dereferencing is nice but `vis`-style printing would produce shorter output, especially if large stack frames are involved).

I would also be down to implement this feature myself (but not for another month or so).

k4lizen:
seperate command makes more sense if the output looks completely different.

`stack-vis` might make sense as it would be inline with the allocator command pattern of `<allocator> vis`

i like the idea

probably part of why it feels needed is because our `hexdump` command is really lacking (no option to hexdump in qwords etc.)

i guess ideally we should have all `vis` commands be backed up by some generic aglib function, and then only recolored according to subsystem, but idk how viable that is

disconnect3d:
Done in #3994

PR #3994, opened 2026-07-01
Add stack-vis command

ksaweryr:
This PR adds the `stack-vis` command, as discussed in #3917 . To implement it, i extracted the printing code from `vis-heap-chunks` into aglib to avoid code duplication. I have tested it with GDB on some simple programs that call a few nested functions and on a program that uses POSIX threads. The command seems to work as expected.

image


disconnect3d:
Extend the `stack-vis --help` so it explains what are the symbols on the right side

Or: the output could inform on that in a header (which would need to be added)

pwndbg/aglib/memory.py
```diff
@@ -469,3 +471,162 @@ def is_pagefault_supported() -> bool:
...
+        # round down to align with 2*cell_size
+        begin_addr = pwndbg.lib.memory.round_down(cursor, cell_size << 1)
+        end_addr = pwndbg.lib.memory.round_down(stop, cell_size << 1)
+
+        # Reset repeat tracking at chunk boundaries (only if skip_repeating is enabled)
+        if skip_repeating:
+            flush_repeats()
+
+        while cursor != stop:
+            # skip the middle part of a huge chunk
```
k4lizen:
```suggestion
            # skip the middle part of a huge block
```

pwndbg/commands/__init__.py
```diff
@@ -1142,6 +1142,7 @@ def load_commands() -> None:
     import pwndbg.commands.sigreturn
     import pwndbg.commands.slab
     import pwndbg.commands.spray
+    import pwndbg.commands.stack
```
k4lizen: need to rename this file to stackvis

pwndbg/commands/stack.py
```diff
@@ -0,0 +1,115 @@
...
+parser = argparse.ArgumentParser(
+    description="""Visualize stack frames of the current thread.
+
+Repeated lines can be collapsed by setting 'vis-skip-repeating-val' config (on by default)."""
+)
+group = parser.add_mutually_exclusive_group()
+group.add_argument(
+    "count",
+    nargs="?",
+    type=lambda n: max(int(n, 0), 1),
```
k4lizen:
```suggestion
    type=lambda n: max(fix_int_reraise_arg(n, 0), 1),
```
from pwndbg.commands

k4lizen (changes requested):
ay pretty nice!

since this is a new command it will require a test as well https://pwndbg.re/dev/contributing/writing-tests/ .

a couple of questions:
+ does the coloring work when a stack frame has an odd number of qwords?
+ what about when you have two stack frames consecutively which are just one qword each? will they be annotated properly on the right?

i think it would be interesting to differently color or bold the frame pointer or return address, but idk if it's easy to do here so its fine if thats out of scope. also might be more clear visually if the function name was written between the stack frame blocks as opposed to the right of them, but yeah same comment stands for that.

ksaweryr:
@k4lizen 

> does the coloring work when a stack frame has an odd number of qwords?

No, it's a bug that's also present in the current implementation of `vis-heap-chunks`.
EDIT: I'm planning to fix it.

> what about when you have two stack frames consecutively which are just one qword each? will they be annotated properly on the right?

Yes, both symbol names will be listed on a single line (comma-separated).

Bolding frame pointers or return addresses would probably be possible with some modifications to the (now generalized) printing code - maybe `pprint_blocks` could take an additional argument with a callable that for given address (and data) somehow signals what additional formatting should be applied? Function name between frames instead of on the side would indeed be more readable but it would also make the output much longer if a lot of nested calls on functions with small frames are performed.

PR #4003, opened 2026-07-06
kernel: Take flag integer properly from struct slab and struct page

k4lizen:
v6.18: https://elixir.bootlin.com/linux/v6.18/source/mm/slab.h#L53
```c
struct slab {
	memdesc_flags_t flags;
// ...
```
v6.17: https://elixir.bootlin.com/linux/v6.17/source/mm/slab.h
```c
struct slab {
	unsigned long flags;
```
Same for `struct page`.

### Before
image

### After
image

### DIFF
--- a/pwndbg/aglib/kernel/macros.py
+++ b/pwndbg/aglib/kernel/macros.py
@@ -44,6 +44,21 @@ def _arr(x: pwndbg.dbg_mod.Value, n: int) -> pwndbg.dbg_mod.Value:
     return (ptr + n).dereference()
 
 
+def memdesc_flag_or_int(val: pwndbg.dbg_mod.Value) -> int:
+    """
+    Return flag integer value from `val` which is either of type `unsigned long`
+    or `memdesc_flags_t`.
+
+    Use this to read `flags` variable of `struct slab` and `struct page`.
+    """
+    # Since kernel v6.18 the flag integer is tucked away in a struct.
+    # (https://elixir.bootlin.com/linux/v6.18/source/include/linux/mm_types.h#L79)
+    # (https://elixir.bootlin.com/linux/v6.18/source/mm/slab.h#L53)
+    if val.type.has_field("f"):
+        return int(val["f"])
+    return int(val)
+
+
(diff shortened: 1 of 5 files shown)


k4lizen:
i think it would've been better if the `int(page["flags"])` exception actually propagated and is shown to the user rather than been swallowed up by the ambiguous error message, we would've caught this way sooner, and this bug has literarly been haunting me.

pwndbg/aglib/kernel/macros.py
```diff
@@ -57,8 +57,17 @@ def compound_head(page: pwndbg.dbg_mod.Value) -> pwndbg.dbg_mod.Value:
...
 
+    # Fetch page flags.
+    maybeflags = page["flags"]
+    # Since kernel v6.18 the flag integer is tucked away in a struct.
+    # (https://elixir.bootlin.com/linux/v6.18/source/include/linux/mm_types.h#L79)
+    if maybeflags.type.has_field("f"):
+        flagint: int = int(maybeflags["f"])
+    else:
+        flagint = int(maybeflags)
+
```
jxuanli: i'm thinking we should create a separate func for this. iirc, the group working on this (memdesc) hasn't fully done yet so it may change in the future

pwndbg/aglib/kernel/symbol.py
```diff
@@ -213,8 +213,12 @@ def recover_page_typeinfo() -> str:
     result += "\n".join(f"#define {s}" for s in defs)
     result += COMMON_TYPES
     result += """
-    struct page { // just a simplied page struct with relavent fields
+    struct page { // just a simplied page struct with relevant fields
+#if KVERSION < KERNEL_VERSION(6, 18, 0)
         unsigned long flags;
+#else
+        struct { unsigned long f; } flags;
+#endif
```
k4lizen, line 224: btw this is not exactly identical to the real definition, because in the real one you might need to "strip typedef" for some stuff while here you don't need to, but I think it shouldn't actually make any issues anywhere

k4lizen:
I'm not really sure which aglib.kernel files are supposed to be dependent on which other aglib.kernel files, so not sure where to put this function without creating a new file / circular dep

jxuanli (approved):

PR #4014, opened 2026-07-13
chore(ruff): enable flake8-bugbear ruleset

c0rydoras:

### DIFF
--- a/pwndbg/aglib/kernel/kallsyms.py
+++ b/pwndbg/aglib/kernel/kallsyms.py
@@ -141,7 +141,7 @@ class Kallsyms:
         current_index = 0x30
 
         position -= 1
-        for _tokens_backwards in range(current_index):
+        for _ in range(current_index):
             for chars_in_token in range(50):
                 position -= 1
                 assert position >= 0
@@ -414,7 +414,7 @@ class Kallsyms:
             tokens = []
             position = self.token_table
 
-            for _num_token in range(256):
+            for _ in range(256):
                 token = ""
                 while self.kernel_ro_mem[position]:
                     token += chr(self.kernel_ro_mem[position])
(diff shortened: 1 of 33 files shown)


pwndbg/lib/cache.py
```diff
@@ -254,4 +254,4 @@ def clear_function_cache(func: Callable[..., T]) -> None:
     clears the cache for all object instances.
     """
     assert hasattr(func, "cache")
-    getattr(func, "cache").clear()
+    func.cache.clear()
```
k4lizen, line 257: this makes sense but it flares up pyright so I'd rather leave it as is

pwndbg/emu/emulator.py
```diff
@@ -397,7 +397,7 @@ def telescope(self, address: int, limit: int, read_size: int = None) -> list[int
         if not read_size <= 8:
             return result
 
-        for i in range(limit):
+        for _i in range(limit):
```
k4lizen: i think these changes hurt legibility. if the var is not used and we're just iterating numbers it should just be `_ in range()` rather than `_i`.

pwndbg/commands/elf.py
```diff
@@ -82,7 +82,7 @@ def elfsections(no_rebase: bool) -> None:
 
         # if the binary is started, use the memory permission for the coloring
         if pwndbg.aglib.proc.alive() and not no_rebase:
-            for start, end, size, name, privilege in sections:
+            for start, end, size, name, _privilege in sections:
```
k4lizen: here i think the change improves legibility

pwndbg/aglib/kernel/kallsyms.py
```diff
@@ -414,7 +414,7 @@ def get_token_table(self):
             tokens = []
             position = self.token_table
 
-            for num_token in range(256):
+            for _num_token in range(256):
```
k4lizen: if its not used why not just `_`?

tests/host/gdb/__init__.py
```diff
@@ -28,10 +28,12 @@ def __init__(
     def _run_gdb(
         self,
         target: str,
-        gdb_args_before: list[str] = [],
+        gdb_args_before: list[str] = None,
```
k4lizen, line 31: this is a very good lint

k4lizen (approved):

pwndbg/lib/cache.py
k4lizen: ah i see it complains either way, then its fine

PR #4015, opened 2026-07-13
feat(lldb): implement support for passing stdin to process launch

Ninja3047:
Screenshot 2026-07-13 at 13 57 34

Implemented `process launch -i` to allow stdin from a file in lldb.

### DIFF
--- a/Dockerfile.arch
+++ b/Dockerfile.arch
@@ -32,8 +32,7 @@ COPY ./pyproject.toml /pwndbg/
 # pyproject.toml requires these files, pip install would fail
 RUN touch README.md && mkdir pwndbg && touch pwndbg/empty.py
 
-RUN sed -i 's/read -p "Do you want to do a full system update?/#read/' ./setup.sh && \
-    ./setup.sh
+RUN printf 'y\n' | ./setup.sh
 
 # Comment these lines if you won't run the tests.
 COPY ./setup-dev.sh /pwndbg/
--- a/pwndbg/dbg_mod/lldb/repl/io.py
+++ b/pwndbg/dbg_mod/lldb/repl/io.py
@@ -178,17 +178,20 @@ class IODriver:
         raise NotImplementedError()
 
 
-def get_io_driver() -> IODriver:
+def get_io_driver(stdin_path: str | None = None) -> IODriver:
     """
     Instances a new IODriver using the best strategy available in the current
     system. Meaning a PTY on Unix and plain text on Windows.
+
+    If stdin_path is set, the launched process reads its standard input from
+    that file instead of from the Pwndbg terminal.
     """
     if PTY_AVAILABLE:
         pty = make_pty()
         if pty is not None:
             worker, manager = pty
-            return IODriverPseudoTerminal(worker=worker, manager=manager)
-    return IODriverPlainText()
+            return IODriverPseudoTerminal(worker=worker, manager=manager, stdin_path=stdin_path)
+    return IODriverPlainText(stdin_path=stdin_path)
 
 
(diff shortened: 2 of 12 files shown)


pwndbg/dbg_mod/lldb/repl/io.py
```diff
@@ -581,7 +598,8 @@ def start(self, process: lldb.Process) -> None:
         assert self.process is None, "Multiple calls to start()"
         self.process = process
         self.stop_requested.clear()
-        os.set_blocking(sys.stdin.fileno(), False)
+        if self.stdin_path is None:
+            os.set_blocking(sys.stdin.fileno(), False)
```
k4lizen: not exactly sure what this does, but why is it fine to not set blocking for file-backed stdin?

k4lizen (commented):
thats pretty awesome, thanks

lets let @mbrla0 take a look

Dockerfile.arch
```diff
@@ -32,8 +32,7 @@ COPY ./pyproject.toml /pwndbg/
 # pyproject.toml requires these files, pip install would fail
 RUN touch README.md && mkdir pwndbg && touch pwndbg/empty.py
 
-RUN sed -i 's/read -p "Do you want to do a full system update?/#read/' ./setup.sh && \
```
mbrla0, line 35: Feels out of scope.

setup-dev.sh
```diff
@@ -136,7 +136,7 @@ EOF
 EOF
     fi
 
-    sudo pacman -Syu --noconfirm || true
```
mbrla0, line 139: Feels out of scope.

setup.sh
```diff
@@ -72,7 +72,7 @@ install_pacman() {
     # user want to perform a full system upgrade
     answer=${answer:-n} # n is default
     if [[ "$answer" == "y" ]]; then
-        sudo pacman -Syu || true
```
mbrla0, line 75: Feels out of scope.

scripts/common.sh
```diff
@@ -33,7 +33,7 @@ else
         echo "Error: 'uv' binary not found." >&2
         UV="${PWNDBG_VENV_PATH}/bin/uv"
     fi
-    UV_RUN="${UV} run"
```
mbrla0, line 36: Feels out of scope.

mbrla0 (commented):
Thanks for taking the time to implement this!

Generally I feel this is good to go, with two caveats. First is that there's changes in this PR that I feel are out of scope - which I've marked in comments -, and that should be split off into their own thing. And second is that I personally would prefer this behavior implemented in a different type that inherits from `IODriver`, rather than be tacked on as a special case to both existing driver types, which now have to pull double duty.

Keep in mind, though, that I don't feel too strongly about either of these, so don't consider these remarks to be blocking.

pwndbg/dbg_mod/lldb/repl/io.py
mbrla0: It is. Setting non-blocking is only important there because we copy to and from our own `stdin` when passing the data through either a PTY or LLDB's process I/O methods. Skipping it in favor of something - as they do here - means there's no need to set or clear non-blocking mode.

mbrla0: Had to comment this twice because GitHub's UI is a mess and I hate it (I deleted my previous comment by accident).

disconnect3d:
Also some conflicts needs to be resolved here

k4lizen:
(ping when ready, ty)

disconnect3d:
@Ninja3047 can u fix conflicts?:(

PR #4016, opened 2026-07-14
feat: validate color theme parameters and prevent crashes on invalid values

### LINKED ISSUE #2874: add color parameter validation

dj-wise-ronin:
### Summary
This PR adds input validation to color theme config parameters and prevents potential runtime `KeyError` or `TypeError` crashes when rendering output.

### Changes
1. **Color Function Validation**:
   - Updated `pwndbg.color.generate_color_function` to actively validate each split element against a pre-defined list of valid style/color names.
   - If an invalid color or style is specified, a `ValueError` listing the correct choices is raised immediately.

2. **Graceful GDB Configuration Rollback**:
   - Modified `Parameter.get_set_string` in `pwndbg/gdblib/config.py` to wrap trigger executions in a `try...except` block.
   - On exception, it rolls back both GDB's parameter value and the library's config value to the previous valid state.
   - Re-runs the triggers to restore the active color functions to their previous functional state.
   - Raises `gdb.GdbError` to display the error cleanly to the user in GDB without a traceback.

3. **Unit Tests**:
   - Added `tests/unit_tests/test_color_validation.py` to verify color configuration validation behavior (valid inputs, invalid inputs, multiple inputs, whitespace tolerance).

Resolves #2874

### DIFF
--- a/pwndbg/color/__init__.py
+++ b/pwndbg/color/__init__.py
@@ -208,11 +208,48 @@ def generate_color_function(
     if disable_colors:
         return function
 
-    for color in config.split(","):
+    config_str = config.value if hasattr(config, "value") else str(config)
+
+    valid_colors = [
+        "normal",
+        "black",
+        "red",
+        "green",
+        "yellow",
+        "blue",
+        "purple",
+        "cyan",
+        "light_gray",
+        "light_grey",
+        "foreground",
+        "gray",
+        "grey",
+        "light_red",
+        "light_green",
+        "light_yellow",
+        "light_blue",
+        "light_purple",
+        "light_cyan",
+        "white",
+        "bold",
+        "underline",
+        "none",
+    ]
+
+    for color in config_str.split(","):
+        color = color.strip()
+        if not color:
+            continue
         func_name = color.lower().replace("-", "_")
+        if func_name not in valid_colors:
+            raise ValueError(
+                f"Invalid color/style '{color}'. Valid choices are: {', '.join(sorted(valid_colors))}"
+            )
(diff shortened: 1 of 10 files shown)


pwndbg/color/__init__.py
```diff
@@ -208,11 +208,48 @@ def generate_color_function(
     if disable_colors:
         return function
 
-    for color in config.split(","):
+    config_str = config.value if hasattr(config, "value") else str(config)
+
+    valid_colors = [
```
k4lizen: this must be tied to the actual color definitions otherwise it may run out of drift

tests/unit_tests/test_color_validation.py
```diff
@@ -0,0 +1,34 @@
+from __future__ import annotations
+
+import pytest
+
+import pwndbg.color
+
```
k4lizen, line 6: the more tricky behaviour is the gdb color setting config shenanigens, not raising an exception, this should be a `tests/library/dbg/` test

k4lizen (changes requested):
i think i would prefer if we explicitly validated the color in `get_set_string`

i also am not a complete fan of us eating the exception traceback in all cases, ideally we would eat it just in these user-error cases like color validation

the two blockers for this PR are tests and `valid_colors = [`

k4lizen:
tests are red :<

k4lizen:
Closed in favour of https://github.com/pwndbg/pwndbg/pull/4016 , I added you as a co-author since I based my PR off of yours, thanks.

PR #4017, opened 2026-07-14
Fix #3897: procinfo: show pipe peers and read/write end on remote targets

### LINKED ISSUE #3897: show pipe destination in procinfo, but on remote
This https://github.com/pwndbg/pwndbg/issues/3150 . Should be possible to make work with remote as well. (that was my original use-case, doing android debugging)

Kaz-scr:
+ [x] Output of the feature is shown below.

Fixes #3897 (follow-up to #3150, as discussed in the issue).

## Approach

The local implementation walks `/proc/*/fd`, which can't be done over the remote protocol: `vFile:*` packets can read files but have no readdir. Instead we ask the stub for its **`files` osdata table** (`qXfer:osdata:read`, the machinery behind `info os files`). gdbserver performs the same `/proc/*/fd` walk on its side and returns pid / command / fd / readlink target for every FD on the system. Each endpoint's read/write mode then comes from reading `/proc/PID/fdinfo/FD` over vFile, which also works for processes the stub isn't attached to.

Degradation: stubs without osdata support (e.g. qemu-user) fall back to r/w modes plus same-process peers derived from the debuggee's own FD table; without vFile the output stays as it is today. The fallback is also applied per-inode locally when `/proc` entries of other processes aren't readable.

New modules: `pwndbg/lib/osdata.py` (pure osdata XML parsing, unit-testable anywhere) and `pwndbg/aglib/osdata.py` (chunked qXfer transfer via `send_remote`).

## Output over `target remote` (gdbserver, x86-64)

```
fd[3]      pipe:[92005] (r, peers: pid=13190 'pipetest' fd=4 w)
fd[4]      pipe:[92005] (w, peers: pid=13190 'pipetest' fd=3 r)
fd[5]      pipe:[92006] (r, peers: pid=13203 'pipetest' fd=6 w)
```

pid 13203 is a forked child the debugger is not attached to; the cross-process case works, including its r/w mode.

## Testcase (as requested in the issue)

`tests/library/gdb/tests/test_remote_pipe_peers.py` plus a new `reference-binary-pipe-fork.native.c`: the binary forks so one pipe's write end lives in a different process, which the same-process fallback cannot discover, so the test only passes if the osdata path genuinely works. It skips when gdbserver isn't installed.

## Tested

- Unit tests for the XML/fdinfo parsers (malformed input, truncation, escapes)
- New gdbserver testcase passes; the existing local `test_command_procinfo_pipe` still passes (no local regression)
- Verified live against gdbserver (x86-64) and qemu-user 8.2 (osdata-less fallback path, no crash, modes still resolved via vFile)
- `./lint.sh` clean including mypy

@k4lizen since Android debugging was your original use case, trying this branch against your remote target would be the perfect real-world validation. Everything degrades gracefully if the stub lacks a capability.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/pwndbg/aglib/osdata.py
+++ b/pwndbg/aglib/osdata.py
@@ -14,9 +14,11 @@ import pwndbg.aglib.file
 import pwndbg.dbg_mod
 import pwndbg.lib.osdata
 
-# Large enough to keep round-trips low, small enough to fit any stub's packet
-# size buffer (gdbserver's default is 16KiB, the protocol minimum is 16KiB too,
-# but qemu and embedded stubs can be much smaller).
+# The length we request per qXfer read. It is only an upper bound: the stub
+# trims each reply to fit its own packet buffer (gdbserver advertises ~128KiB
+# via qSupported PacketSize, QEMU's stub caps at 4KiB, embedded probes can be
+# ~1KiB) and the loop below advances by the bytes actually received, so a
+# stub with a smaller buffer just needs more round-trips.
 _CHUNK = 0xF00
 
 
--- a/pwndbg/aglib/osdata.py
+++ b/pwndbg/aglib/osdata.py
@@ -1,0 +1,74 @@
+"""
+Fetch ``info os`` tables from a remote GDB stub via ``qXfer:osdata:read``.
+
+This is how we enumerate system-wide state (like every process's open file
+descriptors) on a remote target: the vFile packets can read individual files
+but cannot list directories, while osdata asks the stub itself to walk /proc
+on its side. gdbserver implements it on Linux; stubs that don't (qemu-user,
+embedded probes) answer with an empty reply and we report that as None.
+"""
+
+from __future__ import annotations
+
+import pwndbg.aglib.file
+import pwndbg.dbg_mod
+import pwndbg.lib.osdata
+
+# Large enough to keep round-trips low, small enough to fit any stub's packet
+# size buffer (gdbserver's default is 16KiB, the protocol minimum is 16KiB too,
+# but qemu and embedded stubs can be much smaller).
+_CHUNK = 0xF00
+
+
+def read(table: str) -> str | None:
+    """Fetch the raw XML of the given osdata table ("files", "processes", ...).
+
+    Returns None if the stub doesn't support qXfer:osdata:read (empty reply),
+    doesn't know the table (error reply), or the transfer fails part-way.
+    """
+    data = bytearray()
+    offset = 0
+
+    while True:
+        try:
+            response = pwndbg.dbg.selected_inferior().send_remote(
+                f"qXfer:osdata:read:{table}:{offset:x},{_CHUNK:x}"
+            )
+        except pwndbg.dbg_mod.Error:
+            return None
+
+        if not response:
+            # Empty reply is the protocol's "packet not supported".
+            return None
+
+        marker, payload = response[:1], response[1:]
+        # The reply is the binary-escaped chunk prefixed with 'm' (more data
+        # follows) or 'l' (this is the last chunk). Anything else ('E NN', ...)
+        # is an error.
+        if marker not in (b"m", b"l"):
+            return None
+
+        chunk = pwndbg.aglib.file.gdb_memtox_inverse(payload)
+        data += chunk
+
+        if marker == b"l":
+            break
+        if not chunk:
+            # 'm' with no data would loop forever; treat as a broken stub.
+            return None
+        # qXfer offsets count unescaped object bytes.
+        offset += len(chunk)
+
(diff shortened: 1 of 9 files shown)


pwndbg/aglib/osdata.py
```diff
@@ -0,0 +1,74 @@
...
+
+from __future__ import annotations
+
+import pwndbg.aglib.file
+import pwndbg.dbg_mod
+import pwndbg.lib.osdata
+
+# Large enough to keep round-trips low, small enough to fit any stub's packet
+# size buffer (gdbserver's default is 16KiB, the protocol minimum is 16KiB too,
+# but qemu and embedded stubs can be much smaller).
```
disconnect3d: > can be much smaller

Do we know if they are? What happens if they are smaller?

pwndbg/lib/proc_fd.py
```diff
@@ -57,37 +64,74 @@ def _format_peer(peer: tuple[int, int, str, str]) -> str:
...
-    """Return access mode of an FD from /proc/PID/fdinfo/FD as 'r'/'w'/'rw'.
+    """Local-kernel variant of `parse_fdinfo_mode`, reading procfs directly.
 
-    Returns '?' if fdinfo can't be read or doesn't carry the flags line. The
-    fdinfo flags field is octal text and contains the original open() flags;
-    the bottom two bits encode the access mode (O_RDONLY / O_WRONLY / O_RDWR).
+    Returns '?' if fdinfo can't be read.
     """
     try:
         with open(f"/proc/{pid}/fdinfo/{fd}") as f:
```
disconnect3d: Shall we just use the `pwndbg.aglib.file.get` here instead of having multiple functions doing similar stuff but over file _here_ vs _remote_ etc?

pwndbg/aglib/osdata.py
Kaz-scr: Good question. I probed this against a live gdbserver rather than trusting the Claude's comment, and the comment turned out to be stale: gdbserver on Ubuntu 24.04 advertises PacketSize=0x2001f (~128 KiB) in qSupported, QEMU's stub caps packets at 4 KiB, and embedded probes can be around 1 KiB.

As for what happens when the stub's buffer is smaller than the request: the length in qXfer:...:read:offset,length is only an upper bound. The stub trims each reply to fit its own buffer and marks it m (more data follows), and the read loop advances by the number of bytes actually received (offset += len(chunk)), not by the requested length, so a smaller buffer just costs more round-trips. I verified this by requesting 0x40000 per chunk: gdbserver returned 131102-byte chunks (its buffer minus framing) and the 267 KB files table still reassembled into well-formed XML in 3 round-trips. If a stub errors out or replies with something malformed instead, read() returns None and procinfo falls back to same-process peers.

```
stub qSupported PacketSize=2001f
requested 0xf00    -> marker=m got=3840 bytes
requested 0x10000  -> marker=m got=65536 bytes
requested 0x40000  -> marker=m got=131102 bytes
oversized-chunk transfer: 267321 bytes in 3 round-trips, well-formed XML: True
```

pwndbg/lib/proc_fd.py
Kaz-scr: Agreed. The one constraint is that pwndbg/lib/ can't import aglib (the lib_is_pure custom-lint rule), so the walker can't call pwndbg.aglib.file.get itself. I restructured it to get the same effect: _read_fdinfo_mode is gone, the walk now returns bare (pid, fd, comm) endpoints, and the r/w mode of every endpoint is resolved in one place in procinfo via the pwndbg.aglib.file.get based helper, identically for local and remote targets. Side benefit: the debuggee's own fdinfo is no longer read twice on remote targets.

Output unchanged after the restructure, over a real gdbserver session:

image

Done in eb6c0b5.

tests/library/gdb/tests/test_remote_pipe_peers.py
```diff
@@ -0,0 +1,98 @@
...
+        line = process.stderr.readline()
+        if not line:
+            break
+        match = re.search(r"Listening on port (\d+)", line)
+        if match:
+            port = int(match.group(1))
+            break
+    if port is None:
+        process.kill()
+        pytest.fail("gdbserver did not report a listening port")
```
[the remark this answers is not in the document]
Kaz-scr: Both points addressed in eb6c0b5: the fixture now does a non-blocking deadline read of stderr (the readline version could indeed wedge — confirmed while testing) and reaps the process on the failure path.

disconnect3d:
@k4lizen mind trying this on Android? would that even work here? Can gdbserver look into a target process /proc/$pid/* entries?

PR #4027, opened 2026-07-16
fix: don't allow `set`ting to invalid colours

### LINKED ISSUE #2874: add color parameter validation

c0rydoras:

---
- would (somewhat) fix #2874

---

- a lot of this is probably in the wrong place (code wise)
- currently only implemented for gdb


k4lizen:
Closed in favour of https://github.com/pwndbg/pwndbg/pull/4016 , I added you as a co-author since I based my PR off of yours, thanks.

PR #4030, opened 2026-07-17
chore: cleanup `pwndbg.hexdump`

c0rydoras:
`dX_call` is only ever used once, and not coupled to `hexdump` at all really

therefore it should **not** be in the top level `pwndbg.hexdump` but rather, in the one place it's actually used; which is `pwndbg.commands.windbg`

### DIFF
--- a/pwndbg/aglib/ipi_helpers.py
+++ b/pwndbg/aglib/ipi_helpers.py
@@ -33,7 +33,7 @@ def hd(addr: int, count: int = 0x40) -> None:
         hd(0x400000, 0x100)
     """
     data = pwndbg.aglib.memory.read(addr, count)
-    for line in pwndbg.hexdump.hexdump(bytes(data), address=addr, count=count):
+    for line in pwndbg.hexdump.hexdump(bytes(data), address=addr):
         print(line)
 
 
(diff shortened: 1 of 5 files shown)


pwndbg/aglib/ipi_helpers.py
```diff
@@ -33,7 +33,7 @@ def hd(addr: int, count: int = 0x40) -> None:
         hd(0x400000, 0x100)
     """
     data = pwndbg.aglib.memory.read(addr, count)
-    for line in pwndbg.hexdump.hexdump(bytes(data), address=addr, count=count):
+    for line in pwndbg.hexdump.hexdump(bytes(data), address=addr):
```
k4lizen, line 36: does this not change the semantics of the command?

k4lizen (approved):
makes sense, thanks

pwndbg/aglib/ipi_helpers.py
c0rydoras: `count` was/is only ever used in the `dX` branch

k4lizen: i guess whoever wrote that code got confused by that footgun, hopefully now it'll be more clear :>

PR #4032, opened 2026-07-19
aglib: narrow broad exceptions to OSError in file.py

### LINKED ISSUE #4025: Broad `except Exception` in `aglib/file.py` should be narrowed to `OSError`
In `pwndbg/aglib/file.py`, several functions like `get()` and `readlink()` use a very broad `except Exception:` clause when attempting to read files or readlink paths locally:
https://github.com/pwndbg/pwndbg/blob/e2109a93658282d85a9a3631c07785ae93f81ffe/pwndbg/aglib/file.py#L122-L125

Since standard library file operations raise `OSError` subclasses (like `FileNotFoundError`, `PermissionError`), using `except OSError:` would be much safer and more idiomatic, avoiding swallowing unexpected logic errors or other unrelated exceptions.

I can drop a quick PR to narrow these down!

Builder106:
Fixes #4025.

**Working on recording**

This PR addresses the overly broad `except Exception` blocks in `pwndbg.aglib.file` that were silently swallowing unrelated exceptions during file operations.

By narrowing these blocks to catch only `OSError` (and its subclasses like `FileNotFoundError` or `PermissionError`), we ensure that genuine logic bugs or invalid arguments — such as `TypeError` or `AttributeError` — are properly surfaced rather than failing silently and returning fallback values like empty strings or byte arrays.

The accompanying recording demonstrates this behavior using a synthetic test case. Prior to this patch, passing an integer to `pwndbg.aglib.file.readlink()` triggered a `TypeError` that was silently caught, causing the function to return an empty string. With this patch applied, the `TypeError` is correctly bubbled up, making potential bugs in the codebase significantly easier to identify and debug.

*(Note: This PR also bumps `decomp2dbg` to `>=4.0.0` in the dependencies.)*

### DIFF
--- a/pwndbg/aglib/file.py
+++ b/pwndbg/aglib/file.py
@@ -121,7 +121,7 @@ def get(path: str) -> bytes:
     try:
         with open(local_path, "rb") as f:
             return f.read()
-    except Exception:
+    except OSError:
         return b""
 
 
@@ -138,13 +138,13 @@ def readlink(path: str) -> str:
         if can_download_remote_file():
             try:
                 return vfile_readlink(path).decode("utf-8")
-            except Exception:
+            except OSError:
                 return ""
         return ""
 
     try:
         return os.readlink(path)
-    except Exception:
+    except OSError:
         return ""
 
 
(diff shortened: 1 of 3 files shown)


Builder106:
My previous PR (#4031) was from the wrong fork. I've fixed that with this one.

k4lizen:
closed as spam

PR #4033, opened 2026-07-19
prevent invalid colors from being set in GDB

### LINKED ISSUE #2874: add color parameter validation

k4lizen:
Supercedes: https://github.com/pwndbg/pwndbg/pull/4016 and https://github.com/pwndbg/pwndbg/pull/4027 .
Closes: https://github.com/pwndbg/pwndbg/issues/2874 .

image

I'm scared of the re/triggering approach and exception handling, this way the changes are focused on the coloring issue.


c0rydoras (commented):
why not have like a `@color` decorator so we don't have to hardcode this?

k4lizen:
> why not have like a @color decorator so we don't have to hardcode this?

makes sense makes sense, I would've accepted a PR with something like that

(though i don't think it matters that much in this particular case, since we've not added more functions in years)

PR #4040, opened 2026-07-22
Read from ~/.pwndbg_lldbinit to send commands to lldb process

jameskr97:
This PR lets LLDB users define a `~/.pwndbg_lldbinit` file, which functions identically to the `~/.lldbinit` file. I added this specifically so I can use `pwndbg-lldb` on an M-series mac for rust code. The rust toolchains provides `lldb_lookup.py` and `lldb_commands`, and adding those to `{.lldbinit,.pwndbg_lldbinit}` makes printed variables legible in lldb.

- I have linted this code before submitting this PR.

### DIFF
--- a/pwndbginit/pwndbg_lldb.py
+++ b/pwndbginit/pwndbg_lldb.py
@@ -150,11 +150,18 @@ def main() -> None:
 
     # Prepare the startup commands based on those arguments.
     startup = []
+
+    # load user commands from dotfile
+    dotfile = os.path.expanduser("~/.pwndbg_lldbinit")
+    if os.path.exists(dotfile):
+        with open(dotfile) as f:
+            startup += [line.strip("\n") for line in f]
+
     if args.target:
         # DEVIATION: The LLDB CLI silently ignores any target information passed
         # to it when using either '--attach-name' or '--attach-pid', but Pwndbg
         # unconditionally uses it, with a warning.
-        startup = [f"target create '{args.target}'"]
+        startup.append(f"target create '{args.target}'")
 
     if args.attach_name is not None:
         wait = "--waitfor" if args.wait_for else ""


disconnect3d:
@mbrla0 @patryk4815 thoughts? I think we can merge this?

disconnect3d (approved):

patryk4815:
I don't like this idea with `.pwndbg_lldbinit`, better if we stay with standard `. lldbinit` file

pwndbginit/pwndbg_lldb.py
```diff
@@ -150,11 +150,18 @@ def main() -> None:
 
     # Prepare the startup commands based on those arguments.
     startup = []
+
+    # load user commands from dotfile
+    dotfile = os.path.expanduser("~/.pwndbg_lldbinit")
+    if os.path.exists(dotfile):
+        with open(dotfile) as f:
+            startup += [line.strip("\n") for line in f]
```
mbrla0, line 158: Dumb nitpick, but shouldn't we ignore empty lines?

mbrla0:
Looks good to me.

> I don't like this idea with `.pwndbg_lldbinit`, better if we stay with standard `.lldbinit` file

Personally, I think it makes more sense to have it be a separate file. Even though we do our best to stay compatible with vanilla LLDB, we deviate from it in enough places that having it be the same file could force users to pick a favorite between the two and leave the other as a subpar experience, or to be careful that they only choose commands that are exactly the same in both.

disconnect3d:
@OBarronCS @k4lizen thoughts?

I agere with @mbrla0 that it makes sense to have our own file too, but it needs to be documented appropriately.

k4lizen (changes requested):
putting this here deviates from how we do it in `GDB(pwndbg.dbg_mod.Debugger).setup()`.

also, if we're adding this, it should respect the freedesktop dir convention and not just load from home.

I remember we had a discussion about doing a ~/.pwndbginit (for GDB?) and decided against it. @patryk4815 do you remember what the reasons were? iirc you had a PR where you wrote them out.

i don't think we should merge this now since this decision has some history, it should be handled with care (and probably discussed in a dedicated issue)

k4lizen:
@jameskr97 maybe i'm missing something, but why can't you just put your pwndbg settings into ~/.lldbinit ?

patryk4815:
> @jameskr97 maybe i'm missing something, but why can't you just put your pwndbg settings into ~/.lldbinit ?

liblldb python api, don't import `~/.lldbinit`

k4lizen:
> liblldb python api, don't import ~/.lldbinit

so why don't we just read `~/.lldbinit` then?

pwndbginit/pwndbg_lldb.py
jameskr97: From when I was looking into it, lldb already does this. I felt no need to re-ignore, and thought lldb was capable of handling that.

jameskr97:
> > liblldb python api, don't import ~/.lldbinit
> 
> 
> 
> so why don't we just read `~/.lldbinit` then?

I used the `.pwndbg` prefix since I saw it was a separate binary (with how I originally installed it via homebrew) and felt like that saw sufficient enough to warrant a different config file. Incase someone wanted a different config from when using pwndbg vs when using normal lldb.

This was just how I was thinking initially though, feel free to change however you see best fit!

k4lizen:
will close this PR for now, but thanks for the good intentions nonetheless :7

jameskr97:
no problem! out of curiosity, what would you want to see to get this change merged?

k4lizen:
two things, the code to be cleaned up:

> also, if we're adding this, it should respect the freedesktop dir convention and not just load from home.

and me and @patryk4815 to dedicate some time to think about how we want to do this

PR #4049, opened 2026-07-27
Add preliminary glibc 2.44 heap support

k4lizen:
I didn't look at the new malloc source, but the commands work.


k4lizen:
Tested both with debug info and without

RocketMaDev:
> malloc: Remove unused tcache code from unsorted bin scan

`tcache_unsorted_limit` was removed in [ea4c36c3](https://sourceware.org/git/?p=glibc.git;a=commitdiff;h=ea4c36c36bc1fcdc4683127edd2312cbdf540a45)

Issue #4050, opened 2026-07-27
fix archlinux failing tests (~glibc 2.44 support)


k4lizen:
```
Failing tests:
- tests/library/dbg/tests/test_command_canary.py::test_command_canary[x86-64]
- tests/library/dbg/tests/test_context_commands.py::test_source_code_tabstop
- tests/library/dbg/tests/test_context_commands.py::test_context_all_sections_flag
- tests/library/dbg/tests/test_jemalloc.py::test_jemalloc_find_extent
- tests/library/dbg/tests/test_jemalloc.py::test_jemalloc_extent_info
- tests/library/dbg/tests/test_jemalloc.py::test_jemalloc_heap
- tests/library/dbg/tests/test_try_free.py::test_try_free_invalid_misaligned
- tests/library/dbg/tests/test_try_free.py::test_try_free_invalid_size_minsize
- tests/library/dbg/tests/test_try_free.py::test_try_free_invalid_overflow
- tests/library/dbg/tests/test_try_free.py::test_try_free_invalid_size_misaligned
- tests/library/dbg/tests/test_try_free.py::test_try_free_double_free_tcache
- tests/library/dbg/tests/test_try_free.py::test_try_free_invalid_next_size_fast
- tests/library/dbg/tests/test_try_free.py::test_try_free_double_free_or_corruption_top
- tests/library/dbg/tests/test_try_free.py::test_try_free_corrupted_consolidate_backward
- tests/library/dbg/tests/test_try_free.py::test_try_free_double_free_or_corruption_prev
- tests/library/dbg/tests/test_try_free.py::test_try_free_invalid_next_size_normal
- tests/library/dbg/tests/test_try_free.py::test_try_free_double_free
- tests/library/dbg/tests/test_try_free.py::test_try_free_double_free_or_corruption_out
- tests/library/dbg/tests/test_try_free.py::test_try_free_invalid_fastbin_entry
- tests/library/dbg/tests/test_command_canary.py::test_command_canary[i386]
- tests/library/dbg/tests/test_command_exithandlers.py::test_command_exithandlers
```

`test_command_canary` i think someone will need to take a look
`test_command_exithandlers` test issue, code is fine i think
`test_try_free...` probably also complicated

I think for these its some debuginfo detection issue:
```
- tests/library/dbg/tests/test_context_commands.py::test_source_code_tabstop
- tests/library/dbg/tests/test_context_commands.py::test_context_all_sections_flag
[...]

PR #4055, opened 2026-07-28
fix(got): filter non-ELF pages

RocketMaDev:

Use a simple read to probe if the page is actually mapped from an executable rather than a regular file.
https://sourceware.org/git/?p=glibc.git;a=commitdiff;h=b9957a70b82203d91b0659b93cfe8632f954aba3 maps /etc/ld.so.cache into memory, which is not a file, breaking unit test. However, these regular files should not be treated as shared libraries at first.

image

This PR is part of archlinux test fix.

Ref #4050

### DIFF
--- a/pwndbg/commands/got.py
+++ b/pwndbg/commands/got.py
@@ -190,6 +190,9 @@ def iter_objfiles() -> Iterator[pwndbg.lib.memory.Page]:
         if page.objfile in uniq:
             continue
         if pwndbg.aglib.memory.read(page.start, 4, True) != b"\x7fELF":
+            # FIXME: hacky, should be pulled out to a page.is_disk_backed_elf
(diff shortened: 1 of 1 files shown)


k4lizen (approved):
not really ideal to fix it like this but fixing it properly is more involved so lets at least leave a comment about it (i added it)

which test was failing for you though? i didn't see any got-related test failures

RocketMaDev:
> which test was failing for you though? i didn't see any got-related test failures

It's `[GDB] - tests/library/gdb/tests/test_commands_elf.py::test_command_got_for_target_binary_and_loaded_library` which runs `got -p l`, but accidently selects regular-file `/etc/ld.so.cache`, so it runs into error. I fix it by removing non-ELF files from "shared libraries" list.

PR #4059, opened 2026-07-31
Add hexdump --highlight for spotting bad chars

### LINKED ISSUE #1028: Add support for bad char/bad bytes comparisons similar to what mona has
As discussed in the pwndbg Discord server with disconnect3d, I am writing this FR.

It would be nice if there was a bad char comparison function in pwndbg that is equivalent to monas `cmp` command:
```
!mona cmp -f C:\mona\vulnserver\bytearray.bin -a <badchars memory address>
```
In the example above, the bytearray.bin contain bytes from 0x00-0xff / the payload you send. The cmp command will then compare this byte array with a specific memory address (or better yet, search through the memory for a match)

I looked at the encoder https://docs.pwntools.com/en/stable/encoders.html where you can specify bytes you want to avoid, but it would be nice if there exists something to detect them as well. 

The purpose of this would allow you to compare a certain array of bytes with a specific section in memory at a given address. 

Imagine you send a payload of 256 bytes (0x00-0xff), then i.e 0x0b is the only bad byte, the command would compare your byte array/payload to what is found in memory and highlight the mismatch/bad bytes in a (red?) color to make it easy to spot. 

I believe this could be very helpful when debugging payload shellcode that is not working as intended in cases which bad bytes are the reason behind it.

LALITH0110:
Closes #1028.

Adds a `--highlight` / `-H` flag to `hexdump` that colors chosen byte values wherever they appear, in **both** the hex and the ASCII panes. This is aimed at the mona `cmp` workflow described in the issue: dump a buffer and immediately see which bad chars survived into memory or which bytes got mangled.

The pattern syntax follows @disconnect3d's suggestion in the thread — two hex nibbles, where `?` matches any nibble:

```
hexdump $rsp 64 --highlight=00,0a,0d
hexdump $rsp 64 --highlight=4?
```

### Screenshot

hexdump --highlight

> Note on this image: it is real output from `pwndbg.hexdump.hexdump()`, rendered to PNG. I developed this on macOS, where the bundled GDB reports `Don't know how to run`, so I could not capture a live session with an inferior attached. The bytes and colors are exactly what the command emits; only the surrounding gdb prompt is absent. Happy to replace it with a genuine session capture if you'd prefer.

### Notes on the design

- **Why the underline in the default color.** `hexdump-zero-color` already paints `0x00` red, and `0x00` is the most common bad char there is. A color-only highlight was nearly invisible in exactly the case the flag exists for, so `hexdump-highlight-color` defaults to `light-red,underline`. Underlining follows the idiom already used by `hexdump-highlight-group-lsb`.
- **Why exactly two nibbles.** Rejecting a bare `4` avoids the ambiguity of whether it means `0x04` or `0x40`.
- **Performance.** Patterns are expanded eagerly into a `set[int]` at parse time, so the render loop only does one set-membership test per byte. `hexdump` can emit thousands of lines, so this stays off the hot path.
- `parse_byte_patterns()` lives in `pwndbg/lib/common.py` so it carries no debugger dependencies and is unit-testable.

### Testing

- 10 new unit tests in `tests/unit_tests/test_byte_patterns.py` covering wildcards, `0x` prefixes, case-insensitivity, dedup of overlapping patterns, and every error path.
- Verified the render path against real pwndbg output, including that highlighting only ever adds color and never alters the dumped text, and that `highlight=None` / `highlight=set()` are exact no-ops.
- `ruff check`, `ruff format --check`, `scripts/custom-lint.py`, `vermin` (3.10) and `mypy` all clean.

### One thing to flag

The two files under `docs/` are autogenerated, but `scripts/generate-docs.sh` **hangs in its LLDB extraction stage on macOS**, so I could not rebuild the markdown locally. I wrote both by hand to match the generator's exact format. If `verify-docs.sh` disagrees in CI, please tell me and I'll push a fixup — or feel free to regenerate them yourself.


k4lizen:
closed as untested, spam.

Issue #3805, opened 2026-02-28
improve `heap` command output


k4lizen:
<img width="647" height="909" alt="Image" src="https://github.com/user-attachments/assets/41ad8665-af8c-4ebe-a680-3b90bbbaaf41" />

I pretty much never use this command, but the reason is moreso that the formatting is bad than the fact that "tell me a bit of info about every chunk" is a fundamentally bad idea.

Each chunk output should be exactly one line.

disconnect3d:
+1, agree.

Can someone show us how does gef or bata24/gef display those?

cc: @CptGibbon

k4lizen:
<img width="1334" height="1025" alt="Image" src="https://github.com/user-attachments/assets/751d28b6-5be9-4039-bb9e-40ab3610b93e" />

RocketMaDev:
> <img width="1334" height="1025" alt="Image" src="https://github.com/user-attachments/assets/751d28b6-5be9-4039-bb9e-40ab3610b93e" />

`addr - base` is always 0x10? Perhaps we can discard one of these two var.

bata24:
The reason for intentionally leaving both is to make it easier to search for addresses using `less`, and also to make it easier to copy addresses when creating scripts.

PR #4072, opened 2026-08-11
Improved heap command output.

Luisss003:
#3805 

Changes `heap` output so that each chunk is displayed on a single line. Related commands such as `hi` and `malloc-chunk` retain their existing output, since issue only specified `heap`.

1. Adds the chunk data address to compact output (default just outputs base address)
2. Separates chunk type and chunk flags for readability
3. Keeps existing `malloc-chunk` output unchanged.
4. `Flags: ` output is left blank to explicitly show that no flags are set if that's the case.
5. Updates heap address extraction tests for the new format
6. Adds a regression test for single line heap output.

+ [X] I am providing a screenshot of the (new feature) / (fixed bug).

image

### DIFF
--- a/pwndbg/commands/ptmalloc2.py
+++ b/pwndbg/commands/ptmalloc2.py
@@ -248,7 +248,7 @@ def heap(
         while chunk is not None:
             if should_stop(chunk.address):
                 break
-            malloc_chunk(chunk.address, verbose=verbose, simple=simple)
+            malloc_chunk(chunk.address, verbose=verbose, simple=simple, compact=True)
             printed_chunks += 1
             chunk = chunk.next_chunk()
         return
@@ -263,7 +263,7 @@ def heap(
     for chunk in h:
         if should_stop(chunk.address):
             break
-        malloc_chunk(chunk.address, verbose=verbose, simple=simple)
+        malloc_chunk(chunk.address, verbose=verbose, simple=simple, compact=True)
         printed_chunks += 1
 
 
@@ -526,6 +526,7 @@ def malloc_chunk(
     simple: bool = False,
     next: int = 0,
     dump: bool = False,
+    compact: bool = False,
 ) -> None:
     """Print a malloc_chunk struct's contents."""
     allocator = pwndbg.aglib.heap.current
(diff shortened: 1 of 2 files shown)


k4lizen (changes requested):
Thanks! The changes look pretty good, imo the output is much better than before.

I am concerned about the `-v` output though, it is pretty unreadable.

Image

For it to be more readable, the output should be aligned. You could achieve this by adding whitespace for padding, though I think a cleaner approach might be just making all pointers (e.g. `fd: 0x555555688910`) show their full 8 bytes (including the leading zeroes) (so `fd: 0x0000555555688910`) - at least in the verbose output (we shouldn't do this for the non-verbose Addr and Data pointers).

k4lizen:
Also it would be cool if you could change up the order of the fields to make the output aligned even for free chunks, so we don't have stuff like this

image

See how bata24 does it: https://github.com/pwndbg/pwndbg/issues/3805#issuecomment-4366447802 . There are two more points I like in his output:
+ Color-coding the free chunks
+ Showing the tcache size and index (like we do in `vis`)

These changes aren't necessary for your PR to be merged though.

CptGibbon:
> You could achieve this by adding whitespace for padding.

Optionally I think the tabulate library is imported by ptmalloc2.py?

k4lizen:
> Optionally I think the tabulate library is imported by ptmalloc2.py?

or maybe preferably `from rich.table import Table`, but since the output of the command is streaming (theres a lot of results being printed in real time), i'm not sure its possible to use it without waiting for all the results to be resolved

tbh there should probably be a limit on the default amount of printed results

CptGibbon:
> i'm not sure its possible to use it without waiting for all the results to be resolved

Oh I hadn't considered that.

disconnect3d:
@CptGibbon @k4lizen shall we merge this?

k4lizen:
since i requested changes it means i think we shouldn't merge as-is

CptGibbon:
I shall leave this in your capable hands

k4lizen:
beep boop reminding you of this PR, but no pressure

PR #4083, opened 2026-08-19
fix(heap): heap command crash on stripped libc without debug info

### LINKED ISSUE #4076: `heap` command does not work on glibc 2.44 due to bad malloc_initialized detection
Legit don't know how this is possible since our tests pass. I had this issue before, then didn't have it, and now have it again. Does it surface when we don't have debug info for glibc?

```
pwndbg> heap
heap: Iteratively print chunks on a heap.

Default to the current thread's active heap.

Usage:
  heap
  heap --count <N>
  heap <addr_start>
  heap <addr_start> <addr_end> (includes chunks with start address <= addr_end)

Range mode can be combined with --count; walking stops when either limit is hit first.
│ /home/user/code/pwndbg/pwndbg/commands/__init__.py:517 in __call__                               │
│    514 │                                                                                         │
│    515 │   def __call__(self, *args: Any, **kwargs: Any) -> str | None:                          │
│    516 │   │   try:                                                                              │
│ ❱  517 │   │   │   return self.function(*args, **kwargs)                                         │
│    518 │   │   except TypeError:                                                                 │
│    519 │   │   │   print(f"{self.command_name}: {self.description}")                             │
│    520 │   │   │   pwndbg.exception.handle(self.function.__name__)                               │
│ /home/user/code/pwndbg/pwndbg/commands/__init__.py:792 in _OnlyWhenUserspace                     │
│    789 │   @functools.wraps(function)                                                            │
│    790 │   def _OnlyWhenUserspace(*a: P.args, **kw: P.kwargs) -> T | None:                       │
│    791 │   │   if not pwndbg.aglib.qemu.is_qemu_kernel():                                        │
│ ❱  792 │   │   │   return function(*a, **kw)                                                     │
│    793 │   │   log.error(                                                                        │
│    794 │   │   │   f"{func_name(function)}: This command may only be run when not debugging a Q  │
│    795 │   │   )                                                                                 │
│ /home/user/code/pwndbg/pwndbg/commands/__init__.py:883 in _OnlyWhenHeapIsInitialized             │
│    880 def OnlyWhenHeapIsInitialized(function: Callable[P, T]) -> Callable[P, T | None]:         │
│    881 │   @functools.wraps(function)                                                            │
│    882 │   def _OnlyWhenHeapIsInitialized(*a: P.args, **kw: P.kwargs) -> T | None:               │
│ ❱  883 │   │   if pwndbg.aglib.heap.current is not None and pwndbg.aglib.heap.current.is_initia  │
│    884 │   │   │   return function(*a, **kw)                                                     │
│    885 │   │   log.error(f"{func_name(function)}: Heap is not initialized yet.")                 │
│    886 │   │   return None                                                                       │
│ /home/user/code/pwndbg/pwndbg/aglib/heap/ptmalloc.py:1700 in is_initialized                      │
│   1697 │   │   │   addr = pwndbg.aglib.symbol.lookup_symbol_addr("__malloc_initialized")         │
│   1698 │   │   # fallback for GLIBC 2.42 as __malloc_initialized was removed                     │
│   1699 │   │   if addr is None:                                                                  │
│ ❱ 1700 │   │   │   return int(self.mp["sbrk_base"]) != 0                                         │
│   1701 │   │   return pwndbg.aglib.memory.s32(addr) > 0                                          │
│   1702                                                                                           │
│   1703                                                                                           │
TypeError: 'NoneType' object is not subscriptable
PS: Pull requests are welcome
pwndbg> libcinfo
libc: glibc
libc version: 2.44
linked: dynamically
URLs:
    project homepage:       https://sourceware.org/glibc/
    read the source:        https://elixir.bootlin.com/glibc/glibc-2.44/source
    download the archive:   https://ftp.gnu.org/gnu/libc/glibc-2.44.tar.gz
    git clone               https://sourceware.org/git/glibc.git
Mappings:
    libc is at:             0x7ffff7c00000
           /usr/lib/libc.so.6
    ld is at:               0x7ffff7fbe000
           /usr/lib/ld-linux-x86-64.so.2
Symbolication:
    has exported symbols:  yes
    has internal symbols:  no
    has debug info:        no
pwndbg> 
```

M4idCat:
## Summary

Fixes the `heap` command crash on a stripped libc (glibc >= 2.42), reported in #4076.

### Before glibc 2.42

glibc tracked malloc initialization with a `static` flag in `malloc/malloc.c`
(the `__malloc_initialized` / `__libc_malloc_initialized` symbols). pwndbg's
`DebugSymsHeap.is_initialized()` read that symbol.

### Since glibc 2.42

The flag was **removed entirely** and initialization is now tracked by the
internal `sbrk_base` field of the malloc parameters struct:

- `char *sbrk_base;` field — [`malloc/malloc.c#L1625`](https://elixir.bootlin.com/glibc/glibc-2.44/source/malloc/malloc.c#L1625)
- `static struct malloc_par mp_ = { ... };` — [`malloc/malloc.c#L1651`](https://elixir.bootlin.com/glibc/glibc-2.44/source/malloc/malloc.c#L1651)
- init check — [`malloc/malloc.c#L2349`](https://elixir.bootlin.com/glibc/glibc-2.44/source/malloc/malloc.c#L2349):

```c
if (mp_.sbrk_base == NULL)
    mp_.sbrk_base = brk;   /* first sbrk → malloc is initialized */
```

### The pwndbg bug

PR #3464 replaced the old `assert` with a fallback for the removed symbol:

```python
if addr is None:
    return int(self.mp["sbrk_base"]) != 0
```

but `self.mp` is only resolvable when the libc has **debug info**: `mp_` is a
`static` (non-exported) symbol and the `struct malloc_par` type comes from
debug info. On a stripped libc `self.mp` is `None`, so the `heap` command
crashes with:

```
TypeError: 'NoneType' object is not subscriptable
```

## When the bug triggers

| glibc version | libc debug info | before this fix | after this fix |
|---|---|---|---|
| >= 2.42 (2.42/2.43/2.44) | no (stripped) | :boom: TypeError (the reported bug) | :white_check_mark: works |
| >= 2.42 | yes | :white_check_mark: works (via `sbrk_base`) | :white_check_mark: works |
| < 2.42 (e.g. 2.41) | no (stripped) | :boom: `AssertionError` (pre-#3464) / TypeError | :white_check_mark: works |
| < 2.42 | yes | :white_check_mark: works (via `__malloc_initialized`) | :white_check_mark: works |

So the trigger condition is "stripped libc (no debug info)", which exists
across all glibc versions — 2.42 just removed the symbol and made the crash
path the default.

## Fix

Guard the fallback: when `self.mp is None`, fall back to checking whether the
`[heap]` mapping exists — the same signal `HeuristicHeap.is_initialized()` uses:

```python
if addr is None:
    if self.mp is None:
        return any(page.is_heap for page in pwndbg.aglib.vmmap.get())
    return int(self.mp["sbrk_base"]) != 0
```

## Reproduction

Built glibc 2.41 and 2.44 from source, compiled a small test program
(malloc/free of 0x20 chunks), stripped `libc.so.6` to reproduce the
no-debug-info condition, and ran `heap`:

- **Before**: `heap` crashes with `TypeError: 'NoneType' object is not subscriptable` on glibc 2.44 (stripped)
- **After**: `heap` correctly prints the chunks (allocated + free tcache chunks + top chunk)
- Also verified with and without debug info on glibc 2.41 / 2.43 / 2.44 (see matrix above)
```bash
./.venv/bin/pwndbg .repro/test_program_244g
~/.gdbinit: Skipped loading Pwndbg from `source path/gdbinit.py` - already loaded.
pwndbg: loaded 200 pwndbg commands. Type pwndbg [filter] for a list.
pwndbg: created 12 GDB functions (can be used with print/break). Type help function to see them.
Reading symbols from .repro/test_program_244g...
Pwndbg's hexdump colorizes by byte type and lets you re-run with no args to keep dumping forward
pwndbg> set debuginfod enabled off
pwndbg> start
libcinfo
Temporary breakpoint 1 at 0x1277: file test_program.c, line 15.
Using host libthread_db library "/home/xxx/Global/Dev/pwndbg/.repro/glibc-install/lib/libthread_db.so.1".

Temporary breakpoint 1, main () at test_program.c:15
15          setbuf(stdout, NULL);
─────────────────────────────────────────────────────────────────────────────────────────────[ REGISTERS / show-flags off / show-compact-regs off ]──────────────────────────────────────────────────────────────────────────────────────────────
 RAX  0x7ffff7faffa8 (environ) —▸ 0x7fffffffd4e0 —▸ 0x7fffffffd8dd ◂— 'USER=xxx'
 RBX  0
 RCX  0x555555557d78 (__do_global_dtors_aux_fini_array_entry) —▸ 0x5555555551e0 (__do_global_dtors_aux) ◂— endbr64
 RDX  0x7fffffffd4e0 —▸ 0x7fffffffd8dd ◂— 'USER=xxx'
 RDI  2
 RSI  0x7fffffffd4c8 —▸ 0x7fffffffd89f ◂— '/home/xxx/Global/Dev/pwndbg/.repro/test_program_244g'
 R8   0x7ffff7fa8840 —▸ 0x7ffff7faa160 ◂— 0
 R9   0x7ffff7faa160 ◂— 0
 R10  0
 R11  0x7ffff7ffcab0 (__pointer_chk_guard_local) ◂— 0x4265a3da0c9c90a8
 R12  2
 R13  0x7ffff7ffd000 (_rtld_local) —▸ 0x7ffff7ffe3c0 —▸ 0x555555554000 ◂— 0x10102464c457f
 R14  0x7fffffffd4e0 —▸ 0x7fffffffd8dd ◂— 'USER=xxx'
 R15  0x555555557d78 (__do_global_dtors_aux_fini_array_entry) —▸ 0x5555555551e0 (__do_global_dtors_aux) ◂— endbr64
 RBP  0x7fffffffd3a0 —▸ 0x7fffffffd4c8 —▸ 0x7fffffffd89f ◂— '/home/xxx/Global/Dev/pwndbg/.repro/test_program_244g'
 RSP  0x7fffffffd370 {a} ◂— 0
 RIP  0x555555555277 (main+12) ◂— mov rax, qword ptr [rip + 0x2d92]
──────────────────────────────────────────────────────────────────────────────────────────────────────[ DISASM / x86-64 / set emulate on ]───────────────────────────────────────────────────────────────────────────────────────────────────────
 ► 0x555555555277 <main+12>    mov    rax, qword ptr [rip + 0x2d92]     RAX, [stdout@GLIBC_2.2.5] => 0x7ffff7fa9740 (_IO_2_1_stdout_) ◂— 0xfbad2084
   0x55555555527e <main+19>    mov    esi, 0                            ESI => 0
   0x555555555283 <main+24>    mov    rdi, rax                          RDI => 0x7ffff7fa9740 (_IO_2_1_stdout_) ◂— 0xfbad2084
   0x555555555286 <main+27>    call   setbuf@plt                  <setbuf@plt>
 
   0x55555555528b <main+32>    call   getpid@plt                  <getpid@plt>
 
   0x555555555290 <main+37>    mov    edx, eax
   0x555555555292 <main+39>    lea    rax, [rip + 0xda1]     RAX => 0x55555555603a ◂— 'PID: %d\n'
   0x555555555299 <main+46>    mov    esi, edx
   0x55555555529b <main+48>    mov    rdi, rax               RDI => 0x55555555603a ◂— 'PID: %d\n'
   0x55555555529e <main+51>    mov    eax, 0                 EAX => 0
   0x5555555552a3 <main+56>    call   printf@plt                  <printf@plt>
────────────────────────────────────────────────────────────────────────────────────────────────────────────────[ SOURCE (CODE) ]────────────────────────────────────────────────────────────────────────────────────────────────────────────────
In file: /home/xxx/Global/Dev/pwndbg/test_program.c:15
    9     getchar();
   10 }
   11 
   12 int main(void)
   13 {
   14     // Disable stdout buffering so output shows up promptly while debugging
 ► 15     setbuf(stdout, NULL);
   16 
   17     printf("PID: %d\n", getpid());
   18 
   19     pause_here("Program start");
   20 
   21     /*
   22      * malloc(0x20)
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────[ STACK ]────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
00:0000│ rsp 0x7fffffffd370 {a} ◂— 0
... ↓        4 skipped
05:0028│-008 0x7fffffffd398 {big} —▸ 0x7fffffffd4c0 ◂— 2
06:0030│ rbp 0x7fffffffd3a0 —▸ 0x7fffffffd4c8 —▸ 0x7fffffffd89f ◂— '/home/xxx/Global/Dev/pwndbg/.repro/test_program_244g'
07:0038│+008 0x7fffffffd3a8 —▸ 0x7ffff7decfad ◂— mov edi, eax
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────[ BACKTRACE ]──────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 ► 0 0x555555555277 main+12
   1 0x7ffff7decfad
   2 0x7ffff7ded0cb __libc_start_main+139
   3 0x555555555165 _start+37
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
pwndbg> libcinfo
libc: glibc
libc version: 2.44
linked: dynamically
URLs:
    project homepage:       https://sourceware.org/glibc/
    read the source:        https://elixir.bootlin.com/glibc/glibc-2.44/source
    download the archive:   https://ftp.gnu.org/gnu/libc/glibc-2.44.tar.gz
    git clone               https://sourceware.org/git/glibc.git
Mappings:
    libc is at:             0x7ffff7dc6000
           /home/xxx/Global/Dev/pwndbg/.repro/glibc-install/lib/libc.so.6
    ld is at:               0x7ffff7fc2000
           /home/xxx/Global/Dev/pwndbg/.repro/glibc-install/lib/ld-linux-x86-64.so.2
Symbolication:
    has exported symbols:  yes
    has internal symbols:  no
    has debug info:        no
pwndbg> break test_program.c:42
Breakpoint 2 at 0x555555555371: file test_program.c, line 42.
pwndbg> continue
Continuing.
PID: 1145416

========== Program start ==========
Press Enter to continue...

a = 0x555555559420
b = 0x555555559450
c = 0x555555559480

Breakpoint 2, main () at test_program.c:42
42          pause_here("After malloc a/b/c");
─────────────────────────────────────────────────────────────────────────────────────────────[ REGISTERS / show-flags off / show-compact-regs off ]──────────────────────────────────────────────────────────────────────────────────────────────
*RAX  0x555555559480 ◂— 'CCCCCCCC'
 RBX  0
*RCX  0x4343434343434343 ('CCCCCCCC')
*RDX  0x13
*RDI  0x7fffffffd1a0 —▸ 0x7fffffffd1d0 ◂— 'c = 0x555555559480\nUUU'
*RSI  0x7fffffffd1d0 ◂— 'c = 0x555555559480\nUUU'
*R8   0x13
*R9   0
 R10  0
*R11  0x202
 R12  2
 R13  0x7ffff7ffd000 (_rtld_local) —▸ 0x7ffff7ffe3c0 —▸ 0x555555554000 ◂— 0x10102464c457f
 R14  0x7fffffffd4e0 —▸ 0x7fffffffd8dd ◂— 'USER=xxx'
 R15  0x555555557d78 (__do_global_dtors_aux_fini_array_entry) —▸ 0x5555555551e0 (__do_global_dtors_aux) ◂— endbr64
 RBP  0x7fffffffd3a0 —▸ 0x7fffffffd4c8 —▸ 0x7fffffffd89f ◂— '/home/xxx/Global/Dev/pwndbg/.repro/test_program_244g'
 RSP  0x7fffffffd370 {a} —▸ 0x555555559420 ◂— 'AAAAAAAA'
*RIP  0x555555555371 (main+262) ◂— lea rax, [rip + 0xcf1]
──────────────────────────────────────────────────────────────────────────────────────────────────────[ DISASM / x86-64 / set emulate on ]───────────────────────────────────────────────────────────────────────────────────────────────────────
b► 0x555555555371 <main+262>    lea    rax, [rip + 0xcf1]     RAX => 0x555555556069 ◂— 'After malloc a/b/c'
   0x555555555378 <main+269>    mov    rdi, rax               RDI => 0x555555556069 ◂— 'After malloc a/b/c'
   0x55555555537b <main+272>    call   pause_here                  <pause_here>
 
   0x555555555380 <main+277>    mov    rax, qword ptr [rbp - 0x30]
   0x555555555384 <main+281>    mov    rdi, rax
   0x555555555387 <main+284>    call   free@plt                    <free@plt>
 
   0x55555555538c <main+289>    lea    rax, [rip + 0xce9]              RAX => 0x55555555607c ◂— 'After free(a)'
   0x555555555393 <main+296>    mov    rdi, rax                        RDI => 0x55555555607c ◂— 'After free(a)'
   0x555555555396 <main+299>    call   pause_here                  <pause_here>
 
   0x55555555539b <main+304>    mov    rax, qword ptr [rbp - 0x28]
   0x55555555539f <main+308>    mov    rdi, rax
────────────────────────────────────────────────────────────────────────────────────────────────────────────────[ SOURCE (CODE) ]────────────────────────────────────────────────────────────────────────────────────────────────────────────────
In file: /home/xxx/Global/Dev/pwndbg/test_program.c:42
   36     printf("c = %p\n", c);
   37 
   38     strcpy(a, "AAAAAAAA");
   39     strcpy(b, "BBBBBBBB");
   40     strcpy(c, "CCCCCCCC");
   41 
 ► 42     pause_here("After malloc a/b/c");
   43 
   44     /*
   45      * After free(), small chunks usually go into the tcache.
   46      */
   47     free(a);
   48 
   49     pause_here("After free(a)");
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────[ STACK ]────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
00:0000│ rsp 0x7fffffffd370 {a} —▸ 0x555555559420 ◂— 'AAAAAAAA'
01:0008│-028 0x7fffffffd378 {b} —▸ 0x555555559450 ◂— 'BBBBBBBB'
02:0010│-020 0x7fffffffd380 {c} —▸ 0x555555559480 ◂— 'CCCCCCCC'
03:0018│-018 0x7fffffffd388 {d} ◂— 0
04:0020│-010 0x7fffffffd390 {e} ◂— 0
05:0028│-008 0x7fffffffd398 {big} —▸ 0x7fffffffd4c0 ◂— 2
06:0030│ rbp 0x7fffffffd3a0 —▸ 0x7fffffffd4c8 —▸ 0x7fffffffd89f ◂— '/home/xxx/Global/Dev/pwndbg/.repro/test_program_244g'
07:0038│+008 0x7fffffffd3a8 —▸ 0x7ffff7decfad ◂— mov edi, eax
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────[ BACKTRACE ]──────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 ► 0 0x555555555371 main+262
   1 0x7ffff7decfad
   2 0x7ffff7ded0cb __libc_start_main+139
   3 0x555555555165 _start+37
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
pwndbg> heap

Allocated chunk | PREV_INUSE
Addr: 0x555555559000
Size: 0x410 (with flag bits: 0x411)

Allocated chunk | PREV_INUSE
Addr: 0x555555559410
Size: 0x30 (with flag bits: 0x31)

Allocated chunk | PREV_INUSE
Addr: 0x555555559440
Size: 0x30 (with flag bits: 0x31)

Allocated chunk | PREV_INUSE
Addr: 0x555555559470
Size: 0x30 (with flag bits: 0x31)

Top chunk | PREV_INUSE
Addr: 0x5555555594a0
Size: 0x20b60 (with flag bits: 0x20b61)

pwndbg> 
```
## Testing

- Added regression test `test_is_initialized_without_debug_info` in `tests/library/dbg/tests/test_heap.py`
- Heap test suite: 20 passed, 4 skipped, 0 failed

## Screenshots

Fixes #4076

---

- [x] I read the contributing documentation.
- [x] I am providing a screenshot of the (fixed bug).

### DIFF
--- a/pwndbg/aglib/dynamic.py
+++ b/pwndbg/aglib/dynamic.py
@@ -156,7 +156,7 @@ def r_debug_link_map_changed_remove_listener(handler: Callable[..., Any]) -> Non
     R_DEBUG_LINK_MAP_CHANGED_LISTENERS.remove(handler)
 
 
-def link_map_head():
+def link_map_head() -> LinkMapEntry | None:
     """
     Acquires a reference to the head entry of the link map.
     """
@@ -171,6 +171,7 @@ def link_map_head():
 
     if r_map != 0:
         return LinkMapEntry(r_map)
+    return None
 
 
 def link_map():
--- a/pwndbg/aglib/heap/ptmalloc.py
+++ b/pwndbg/aglib/heap/ptmalloc.py
@@ -1513,7 +1513,7 @@ class GlibcMemoryAllocator(pwndbg.aglib.heap.heap.MemoryAllocator, Generic[TheTy
             return self.largebin_index_32_big(sz)
         return self.largebin_index_32(sz)
 
-    def is_initialized(self):
+    def is_initialized(self) -> bool:
         raise NotImplementedError()
 
     def is_statically_linked(self) -> bool:
--- a/pwndbg/commands/__init__.py
+++ b/pwndbg/commands/__init__.py
@@ -18,6 +18,7 @@ from typing_extensions import ParamSpec
 from typing_extensions import override
 
 import pwndbg.aglib
+import pwndbg.aglib.dynamic
 import pwndbg.aglib.heap
 import pwndbg.aglib.kernel
 import pwndbg.aglib.proc
@@ -34,8 +35,14 @@ from pwndbg.aglib.heap.ptmalloc import HeuristicHeap
 from pwndbg.color import message
 from pwndbg.lib import SymbolNotRecoveredError
 from pwndbg.lib import TypeNotRecoveredError
+from pwndbg.libc import LibcNotFound
 
 log = logging.getLogger(__name__)
+
+_HEURISTIC_FALLBACK_WARNING = (
+    "pwndbg will try to resolve the heap symbols via heuristic now since we cannot resolve the heap via the debug symbols.\n"
+    "This might not work in all cases. Use `help set resolve-heap-via-heuristic` for more details.\n"
+)
 
 T = TypeVar("T")
 P = ParamSpec("P")
@@ -882,10 +889,68 @@ def OnlyWithTcache(function: Callable[P, T]) -> Callable[P, T | None]:
 def OnlyWhenHeapIsInitialized(function: Callable[P, T]) -> Callable[P, T | None]:
     @functools.wraps(function)
     def _OnlyWhenHeapIsInitialized(*a: P.args, **kw: P.kwargs) -> T | None:
-        if pwndbg.aglib.heap.current is not None and pwndbg.aglib.heap.current.is_initialized():
-            return function(*a, **kw)
-        log.error(f"{func_name(function)}: Heap is not initialized yet.")
-        return None
+        if pwndbg.aglib.heap.current is None or not pwndbg.aglib.proc.alive():
+            log.error(f"{func_name(function)}: Heap is not initialized yet.")
+            return None
+
+        try:
+            libc_type = pwndbg.libc.which()
+        except LibcNotFound:
+            log.error(f"{func_name(function)}: Heap is not initialized yet.")
+            return None
+
+        # At the dynamic linker's entry point, _r_debug is available but its
+        # link map has not been populated yet. Treat that as "heap not
+        # initialized" instead of misreporting the active libc type.
+        if (
+            libc_type == pwndbg.libc.LibcType.UNKNOWN
+            and pwndbg.dbg.selected_inferior().is_dynamically_linked()
+            and pwndbg.aglib.dynamic.is_dynamic()
+            and pwndbg.aglib.dynamic.link_map_head() is None
+        ):
+            log.error(f"{func_name(function)}: Heap is not initialized yet.")
+            return None
+
+        allocator = _resolve_heap_allocator()
+        if allocator is None:
+            return None
+
+        previous_allocator = pwndbg.aglib.heap.current
+        warn_about_heuristic = isinstance(previous_allocator, DebugSymsHeap) and isinstance(
+            allocator, HeuristicHeap
+        )
+
+        # Some allocator methods read the global `heap.current` (e.g.
+        # Arena.__init__() uses it to fetch malloc_state), so the candidate must
+        # be installed while checking initialization. Restore the previous
+        # allocator if the check fails or raises.
+        is_initialized: bool | None = None
+        if allocator is not previous_allocator:
+            pwndbg.aglib.heap.current = allocator
+        try:
+
+            @functools.wraps(function)
+            def _is_initialized() -> bool:
+                return allocator.is_initialized()
+
+            is_initialized = _try2run_heap_command(_is_initialized)
+        finally:
+            if not is_initialized:
+                pwndbg.aglib.heap.current = previous_allocator
+
+        if is_initialized is None:
+            return None
+        if not is_initialized:
+            log.error(f"{func_name(function)}: Heap is not initialized yet.")
+            return None
+
+        # The candidate is initialized and is already installed as the current
+        # allocator. Print the fallback warning only for the first
+        # DebugSymsHeap -> HeuristicHeap transition.
+        if warn_about_heuristic:
+            log.warning(_HEURISTIC_FALLBACK_WARNING)
+
+        return function(*a, **kw)
 
     return _OnlyWhenHeapIsInitialized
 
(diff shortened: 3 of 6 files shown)


M4idCat:
i can not understand the failure of [Test Docker / docker_x86-64 (fedora42) (pull_request)](https://github.com/pwndbg/pwndbg/actions/runs/32245910209/job/96046291603?pr=4083)Failing after 9m

k4lizen (changes requested):
why can we not assume that we have debug info in a function of the DebugSymsHeap class?

k4lizen:
dw about the test failure above, its flaky

M4idCat:
> why can we not assume that we have debug info in a function of the DebugSymsHeap class?

yes, what you doubt is right. the DebugSymsHeap class should be used when we have debug info.i'll dive into it

M4idCat:
@k4lizen You asked why code inside `DebugSymsHeap` cannot simply assume that debug information is available. After tracing the decorator order, I agree that it should be able to make that assumption.

The problem was not inside `DebugSymsHeap.is_initialized()` itself. The problem was that I called it before checking whether `DebugSymsHeap` was actually resolvable.

The heap decorators run in this order:

```text
OnlyWhenUserspace
    ↓
OnlyWhenHeapIsInitialized
    ↓
OnlyWithResolvedHeapSyms
    ↓
heap command
```

Previously, `OnlyWhenHeapIsInitialized` directly called:

```python
pwndbg.aglib.heap.current.is_initialized()
```

The allocator-resolution logic did not run until the next decorator, `OnlyWithResolvedHeapSyms`.

This meant that, on a stripped libc, the flow was:

```text
current allocator is DebugSymsHeap
    ↓
call DebugSymsHeap.is_initialized()
    ↓
only afterwards check whether DebugSymsHeap can be resolved
```

On glibc 2.42 and newer, `DebugSymsHeap.is_initialized()` may read `mp_.sbrk_base`. Without debug information, `mp_` is unavailable, so this caused the reported crash.

I could not fix this by swapping the decorators. As the existing source comment explains, `OnlyWhenHeapIsInitialized` must remain before `OnlyWithResolvedHeapSyms`; otherwise heap commands may access the inferior while the process is not alive and hit an assertion.

I therefore moved the allocator-resolution check earlier, into `OnlyWhenHeapIsInitialized`, before calling `is_initialized()`.

The new flow is:

```text
check that the process is alive
    ↓
select an allocator that can be resolved
    ↓
check that allocator's initialization state
    ↓
run the remaining decorator and heap command
```

To avoid copying the existing resolver logic, I extracted it from
`OnlyWithResolvedHeapSyms` into:

```python
_resolve_heap_allocator()
```

This helper keeps the existing behavior for:

- checking whether the active libc is glibc;
- preferring `DebugSymsHeap` when debug symbols are available;
- falling back to `HeuristicHeap` in `auto` mode;
- preserving the existing `auto`, `force`, `never`, static, and dynamic
  linking paths.

It now returns a candidate allocator instead of immediately changing `pwndbg.aglib.heap.current` or running the command.

Here, the current and candidate allocators are two Pwndbg analysis backends for the same target heap:

- `DebugSymsHeap` uses libc debug symbols and debugger-provided types;
- `HeuristicHeap` reconstructs the required information when those symbols are
  unavailable.

For example, on a stripped libc:

```text
current allocator   = DebugSymsHeap
candidate allocator = HeuristicHeap
```

Moving resolution earlier introduced another requirement: I could not simply
call:

```python
candidate.is_initialized()
```

while leaving `pwndbg.aglib.heap.current` unchanged.

Some existing heap objects read the global allocator rather than only using the allocator passed through `self`. For example, `Arena.__init__()` reads:

```python
pwndbg.aglib.heap.current.malloc_state
```

If I checked a `HeuristicHeap` candidate while the global current allocator was still `DebugSymsHeap`, `Arena` could use the wrong allocator's types.

I therefore temporarily install the candidate during the initialization probe:

```python
previous_allocator = pwndbg.aglib.heap.current
pwndbg.aglib.heap.current = candidate
```

Because this is only a probe, I also needed rollback behavior:

```text
is_initialized() returns False
    → restore the previous allocator

is_initialized() raises and the error is handled
    → restore the previous allocator

is_initialized() is re-raised in verbose/debugger mode
    → restore the previous allocator in finally

is_initialized() returns True
    → keep the candidate allocator
```

This prevents a failed heap command from leaving Pwndbg in a different global allocator state.

I also delayed the heuristic fallback warning until the initialization probe succeeds. Without that change, moving resolution earlier would make `heap` print the heuristic warning even before malloc had initialized the heap:

```text
pwndbg will try to resolve the heap symbols via heuristic...
heap: Heap is not initialized yet.
```

The updated behavior is:

```text
before malloc:
    report that the heap is not initialized
    restore the previous allocator
    do not print the heuristic warning

after malloc:
    keep HeuristicHeap
    print the fallback warning once
    run the heap command
```

I wrapped the initialization probe with the existing
`_try2run_heap_command()` handler instead of adding a new error-handling path:

```python
@functools.wraps(function)
def _is_initialized() -> bool:
    return allocator.is_initialized()

is_initialized = _try2run_heap_command(_is_initialized)
```

This means initialization failures now use the same diagnostics as other heap command failures. `functools.wraps()` also preserves the original command name, so the error is reported as `heap`, `bins`, `arena`, etc., rather than `is_initialized`.

Moving libc resolution earlier also affected the `starti` state. At the dynamic loader entry, the process is alive but libc has not yet been added to the link map, so libc detection may return `UNKNOWN`.

I could not treat every dynamically linked `UNKNOWN` libc as `starti`, because an already loaded libc with an unusual `LD_PRELOAD` filename may also be reported as `UNKNOWN`.

I therefore distinguish the two cases using the actual dynamic-linker state:

```python
libc_type == LibcType.UNKNOWN
and inferior.is_dynamically_linked()
and pwndbg.aglib.dynamic.is_dynamic()
and pwndbg.aglib.dynamic.link_map_head() is None
```

This produces:

```text
starti, empty link map:
    Heap is not initialized yet.

unknown but already loaded libc:
    The currently active libc isn't glibc. It's unknown.
```

Finally, I removed the vmmap fallback from `DebugSymsHeap`. With the resolver now running first, `DebugSymsHeap.is_initialized()` is only selected through the debug-symbol resolver path, while the no-debug-info case is handled by `HeuristicHeap`.

So the final responsibility split is:

```text
DebugSymsHeap
    → assumes debug symbols are available

HeuristicHeap
    → handles the no-debug-info case

heap command decorators
    → select the correct implementation before checking initialization
```

I added regression tests for:

- `heap` at `starti`;
- GDB and LLDB behavior for the starti diagnostic;
- selecting `HeuristicHeap` before initialization;
- making the candidate visible through global `heap.current` during the
  initialization probe;
- rollback after `False`, handled exceptions, and re-raised exceptions;
- committing the candidate only after successful initialization;
- printing the heuristic warning exactly once;
- preserving the unknown-but-loaded libc diagnostic.

k4lizen:
slop

PR #2839, opened 2025-04-06
### FIX:Refactor: Make callstack a generator to fix #2837

Shyanil:
### **Description**

This pull request refactors the `pwndbg.aglib.stack.callstack` function to return a generator instead of a list, providing better memory efficiency and lazy evaluation. Additionally, a new function `callstack_with_symbols()` is introduced to yield `(address, symbol)` tuples from the call stack.

All test files and command references using `callstack()` have been updated accordingly to maintain compatibility and ensure correctness.

This change **resolves issue #2837**.

---

### **Motivation**

- **Memory efficiency**: Stack frames are now processed one-by-one, not all at once.
- **Lazy evaluation**: Enables better handling for deeper stacks or streaming use cases.
- **Enhanced extensibility**: Makes it easier to compose logic that consumes callstack data.

---

### **Changes Made**

---

#### ✅ `pwndbg/aglib/stack.py` – Refactored core logic

```python
# Before
def callstack() -> List[int]:
    frame = pwndbg.dbg.selected_frame()
    addresses = []
    while frame:
        addr = frame.pc()
        if pwndbg.aglib.memory.is_readable_address(addr):
            addresses.append(addr)
        frame = frame.parent()
    return addresses
```

```python
# After
def callstack() -> Generator[int, None, None]:
    frame = pwndbg.dbg.selected_frame()
    while frame:
        addr = frame.pc()
        if pwndbg.aglib.memory.is_readable_address(addr):
            yield addr
        frame = frame.parent()

def callstack_with_symbols() -> Generator[Tuple[int, Optional[str]], None, None]:
    for addr in callstack():
        symbol = pwndbg.symbol.get(addr) if hasattr(pwndbg, 'symbol') else None
        yield addr, symbol
```

---

#### ✅ `pwndbg/tests/gdb-tests/tests/test_callstack.py` – Updated tests

```python
def test_callstack_readable(start_binary):
    start_binary(REFERENCE_BINARY)
    gdb.execute("b break_here")
    gdb.execute("r")

    addresses = list(pwndbg.aglib.stack.callstack())

    assert len(addresses) > 0
    assert all(pwndbg.aglib.memory.is_readable_address(address) for address in addresses)

def test_callstack_with_symbols(start_binary):
    start_binary(REFERENCE_BINARY)
    gdb.execute("b break_here")
    gdb.execute("r")

    for addr, symbol in pwndbg.aglib.stack.callstack_with_symbols():
        assert pwndbg.aglib.memory.is_readable_address(addr)
[...]

Issue #2837, opened 2025-04-06
pwndbg.aglib.stack.callstack() should yield


disconnect3d:
Some refactoring ideas:

1) pwndbg.aglib.stack.callstack() should yield
2) there should be version that yields addrs+symbol
3) this should be documented?
4) retaddr function should be refactored to use the yielding version probably?
5) context_backtrace should use .callstack() or the version that yields addr+symbol

This would also be a nice/better API

Shyanil:
Hi! @disconnect3d 👋
I'd like to work on this issue.

trukwava:
I'd like to pick this up, if nobody else is on it.

Locally I have `callstack_iter()` plus a `callstack_symbols_iter()` that yields `(pc, symbol)` pairs, with `callstack()` kept as a list wrapper per your comment on #2839. Tests pass under both GDB and LLDB.

Two questions before I open a PR:

1. `retaddr` does `value in addresses` and then `del addresses[:index]`, so it needs the whole remaining list — a lazy iterator doesn't buy it anything there. Leave it on `callstack()`?
2. `context_backtrace` walks children as well as parents, windows output by `backtrace-lines`, and compares frame identity to mark the active frame, so it can't consume a flat list of PCs. Should that stay out of scope, or would you rather have a frame-level iterator for it?

disconnect3d:
not sure if its that important to solve the other parts of this issue; tbh we can revisit if it ever needs to be revisited

PR #4084, opened 2026-08-23
Add callstack_iter() and callstack_symbols_iter() to aglib.stack

trukwava:
Refs #2837 — covers items 1-3. See the open questions below for 4 and 5.

### What changed

- `callstack_iter()` — generator yielding each frame's program counter, innermost first, skipping unreadable ones. Frames are walked as they are consumed.
- `callstack_symbols_iter()` — the same walk, yielding `(pc, symbol)` pairs (item 2).
- `callstack()` keeps its `list[int]` return type and is now a thin wrapper over `callstack_iter()`, per your review comment on #2839.
- Docstrings for all three (item 3). The old `callstack()` docstring said it returned "the address of the return address for the current frame", which described neither the return value nor the fact that unreadable frames are skipped.

Symbols resolve through `pwndbg.aglib.symbol.resolve_addr` so the module stays debugger-agnostic — the earlier attempt in #2839 reached for `pwndbg.gdblib.symbol` and broke under LLDB with `ModuleNotFoundError: No module named 'gdb'`.

### Tests

On #2839 you noted the existing tests "are not great; we probably want to assert specific symbols here". `tests/library/dbg/tests/test_callstack.py` now asserts that `break_here` is the innermost frame and that `main` appears after it, that `callstack_iter()` and `callstack()` agree, and that a partial consumer leaves the generator suspended rather than exhausted.

```
./tests.sh -d gdb  -g dbg                 143 passed, 86 skipped, 0 failed
./tests.sh -d lldb -g dbg test_callstack     4 passed
```

`mypy --strict` error count is unchanged against `dev` (1607 on both).

### Open questions (items 4 and 5)

Neither is in this PR — happy to follow up in either direction.

1. **`retaddr`** does `value in addresses` and then `del addresses[:index]`, so it needs the whole remaining list at every stack slot. A lazy iterator buys it nothing there, so I left it on `callstack()`. Switch it anyway for consistency?
2. **`context_backtrace`** walks children as well as parents, windows output by `backtrace-lines`, and compares frame identity to mark the active frame, so it can't consume a flat list of PCs. Sharing code with it would mean a frame-level iterator rather than the PC-level one this issue describes. Worth a separate change?

### Checklist

+ [ ] I am providing a screenshot of the (new feature) / (fixed bug) — there is no user-visible change here, this is an internal `aglib` API addition, so there is nothing to screenshot.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pwndbg/aglib/stack.py
+++ b/pwndbg/aglib/stack.py
@@ -7,10 +7,13 @@ binaries do things to remap the stack (e.g. pwnies' postit).
 
 from __future__ import annotations
 
+from collections.abc import Iterator
+
 import pwndbg
 import pwndbg.aglib
 import pwndbg.aglib.elf
 import pwndbg.aglib.memory
+import pwndbg.aglib.symbol
 import pwndbg.aglib.vmmap
 import pwndbg.aglib.vmmap_custom
 import pwndbg.lib.cache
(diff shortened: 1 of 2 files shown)


k4lizen:
i'm not gonna review this

PR #4085, opened 2026-08-23
Fixup order of instruction enhancement for instructions backwards in memory

OBarronCS:
This is a minor change in the order we "enhance" instructions will disassembling. Previously, it was done first for the instruction at the PC, then at instructions moving backwards from the program counter, and then from instructions moving forwards from the PC.

Now, we enhance instructions in the original program order. This allows us to take advantage of the feature added in #2963 for these instructions (manually tracking register value propagation), which results in nicer disassembly, mostly visibly seen in syscalls names being annotated.

An example below, is where we do `nearpc` after a syscall instruction.

Before: note that the syscall name is not annotated
image

Note that the syscall name is annotated now, since we enhance in order and can propagate the value of `eax` to the syscall instruction.

image

### DIFF
--- a/pwndbg/aglib/disasm/assistant.py
+++ b/pwndbg/aglib/disasm/assistant.py
@@ -175,6 +175,7 @@ class DisassemblyAssistant:
 
         This is the only public method that should be called on this object externally.
         """
+        instruction.enhanced = True
         # It is assumed that the emulator's pc is at the instruction's address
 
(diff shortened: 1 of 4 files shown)


OBarronCS:
Unrelated flaky fedora test is failing:

image

k4lizen (approved):
awesome, thanks!

PR #4089, opened 2026-08-24
Clean up pwndbginit/ a bit

k4lizen:


gdbinit.py
```diff
@@ -72,7 +72,9 @@ def get_venv_path(src_root: Path):
 
 
 def main() -> None:
-    # Check if pwndbg was already loaded from `pwndbg` binary
+    # Check if pwndbg was already loaded
+    # Can happen if you run `pwndbg /bin/sh` and have `source /path/to/gdbinit.py`
+    # in your `~/.gdbinit`.
     if "pwndbg" in sys.modules and hasattr(sys.modules["pwndbg"], "_is_loaded_from_pwndbg"):
```
patryk4815, line 78: please rename `_is_loaded_from_pwndbg` also? or don't

pwndbginit/common.py
```diff
@@ -170,18 +172,44 @@ def set_debuginfod_timeouts() -> None:
...
+def pre_debugger_init() -> StreamHandler[TextIO]:
     """
     Initialization to run before any debugger-specific stuff gets loaded.
     """
+    import pwndbg
+
+    # Marker used to detect double-loading (checked in ../gdbinit.py).
+    # Can happen if you run `pwndbg /bin/sh` and have `source /path/to/gdbinit.py`
+    # in your `~/.gdbinit`.
+    pwndbg._pwndbg_is_loaded = True
```
patryk4815: I'm a bit against rename, because then anyone who tries to "source" the old version of pwndbg with the new pwndbg will crash again

k4lizen: agh okay you're right, thanks

gdbinit.py
k4lizen: ah IDE rename didn't work because its hasattr, mb

PR #4091, opened 2026-08-25
fix(heap): don't query DebugSymsHeap when libc has no debug info

### LINKED ISSUE #4076: `heap` command does not work on glibc 2.44 due to bad malloc_initialized detection

M4idCat:
i read the heap command related code and found the decorator sequntial is fixed, so add an extra check in function `_OnlyWhenHeapIsInitialized`, when you run heap command at a none debuginfo glibc, it would skip the `is_initialized()` call and just go down to `@OnlyWithResolvedHeapSyms`, then that decorator will pick `HeuristicHeap` for us and everything works again.
@k4lizen you asked "why can we not assume that we have debug info in a function of the DebugSymsHeap class?" — u are right, we CAN. the bug is the caller breaking its precondition, so this time i dont  touch DebugSymsHeap at all.
i add the "-> bool" for `is_initialized` because the mypy check
image

Fixes #4076


CptGibbon:
That em dash tells me folks are prompting their models to write less formally in order to pass off their output as human-written.

M4idCat:
> That em dash tells me folks are prompting their models to write less formally in order to pass off their output as human-written.

sry for that, i'm not native english speaker so worry about the way that im presenting

M4idCat:
> That em dash tells me folks are prompting their models to write less formally in order to pass off their output as human-written.

i thanslate my expression by model, if u feel offended i wouldn't use it

M4idCat:
if anyone can point out the fix direction i would appreciate )

k4lizen (changes requested):
reviewing this PR requires some thinking on what the previous design was and what the implications of it were and are here, if you could write down this stuff it would make me review sooner.

the comment in the code is disguistingly slopped so i don't want to see it merged

why is aarch64 skipped?

M4idCat:
# heap command design

**100% human-written**
---
The `heap` command is decorated by 4 decorator, and 3 of them do lots of heap related checks before actual execution, the sequence of which  is `OnlyWhenUserspace` -> `OnlyWhenHeapIsInitialized` ->`OnlyWithResolvedHeapSyms`
```python
@pwndbg.commands.Command(parser, category=CommandCategory.PTMALLOC2)
@pwndbg.commands.OnlyWhenUserspace
@pwndbg.commands.OnlyWhenHeapIsInitialized
@pwndbg.commands.OnlyWithResolvedHeapSyms
def heap(
    addr_start: int | None = None,
    addr_end: int | None = None,
    count: int | None = None,
    verbose: bool = False,
    simple: bool = False,
) -> None:
```
In OnlyWhenHeapIsInitialized, there is a check about whether the heap is initialized. Diveing into it, the initialized check is under the DebugSymHeap object, which meas the libc has debuginfo and can find some useful symbol in it. From this perspective, this call path is correct.
```python
def _OnlyWhenHeapIsInitialized(*a: P.args, **kw: P.kwargs) -> T | None:
        if pwndbg.aglib.heap.current is not None and pwndbg.aglib.heap.current.is_initialized():

def is_initialized(self) -> bool:
    addr = pwndbg.aglib.symbol.lookup_symbol_addr("__libc_malloc_initialized")
    if addr is None:
        addr = pwndbg.aglib.symbol.lookup_symbol_addr("__malloc_initialized")
    # fallback for GLIBC 2.42 as __malloc_initialized was removed
    if addr is None:
        return int(self.mp["sbrk_base"]) != 0
    return pwndbg.aglib.memory.s32(addr) > 0
```
But the issue #4076 occurred here, when a stripped libc are loaded into pwndbg, this check would be crashed when executing this line `int(self.mp["sbrk_base"]) != 0` because we can't find `mp_` symbol in a striped >=2.42 libc, especially this member variable `sbrk_base`  is static type. So the problem is here, we can not make a stripped libc go throw this call chain. Let's check the `current`, which holds the heap allocator implementation, to inspect where and when the `current` is assigned.

When user typing start command to execute a program firstly, the `current` is assigned according to the `resolve_heap_via_heuristic` mode.
```python
@pwndbg.config.trigger(resolve_heap_via_heuristic)
def resolve_heap(is_first_run: bool = False) -> None:
    global pwndbg
    import pwndbg.aglib.heap.ptmalloc

    global current
    if resolve_heap_via_heuristic == "force":
        current = pwndbg.aglib.heap.ptmalloc.HeuristicHeap()
        if not is_first_run and pwndbg.aglib.proc.alive() and pwndbg.libc.has_debug_info():
            print(
                message.warn(
                    "You are going to resolve the heap via heuristic even though you have libc debug symbols."
                    " This is not recommended!"
                )
            )
    else:
        current = pwndbg.aglib.heap.ptmalloc.DebugSymsHeap()
```
`resolve_heap_via_heuristic` 's default value is `auto`, which means the else branch would be executed, pwndbg treat allocator as DebugSymsHeap.We will see the paradox below.

We use DebugSymsHeap.initialized to check whether a stripped glibc is initialized and all the stripped libc fallback mechanism is in `OnlyWithResolvedHeapSyms`, but we can not go into `OnlyWithResolvedHeapSyms` without the successful return result by `OnlyWhenHeapIsInitialized`. That is root cause.

What i wanted patch is eithor modifying the current assigned way or just skip the initialized check in OnlyWhenHeapIsInitialized when facing a strpped glibc and move it to OnlyWithResolvedHeapSyms.

The former fix is doing by assigning the HeuristicHeap to current directly when loading a stripped libc at auto mode, but i found there is related fallback mechanism in `OnlyWithResolvedHeapSyms`.And this patch will break down the fallback mechanism after that.
```python
def OnlyWithResolvedHeapSyms(function: Callable[P, T]) -> Callable[P, T | None]:
    @functools.wraps(function)
    def _OnlyWithResolvedHeapSyms(*a: P.args, **kw: P.kwargs) -> T | None:
        ...
        static = not pwndbg.dbg.selected_inferior().is_dynamically_linked()
        if (
            isinstance(pwndbg.aglib.heap.current, DebugSymsHeap)
            and pwndbg.config.resolve_heap_via_heuristic == "auto"
        ):
            # In auto mode, if the debug symbols are not enough, we will try to use the heuristic if possible
            heuristic_heap = HeuristicHeap()
```
The later patch is to skip the first initialized check and move the duty to `OnlyWithResolvedHeapSyms`. When `OnlyWithResolvedHeapSyms` find the glibc is stripped and  switch to the HeuristicHeap, the `HeuristicHeap.initialized` would be executed correctly.

Lets explain my patch:
```python
if (
    isinstance(pwndbg.aglib.heap.current, GlibcMemoryAllocator)      
    and pwndbg.aglib.proc.alive()                                   
    and (                                                            
        not pwndbg.aglib.heap.current.can_be_resolved()
        or pwndbg.aglib.heap.current.is_initialized()
    )
):
    return function(*a, **kw)
```
When the current is DebugSymsHeap, we execute DebugSymsHeap.can_be_resolved to check the glibc is whether stripped or not, if it is, skip it and return nomally.If current is HeuristicHeap, go throw the same check by HeuristicHeap.is_initialized.

The only problem is that when we run starti command (not load glibc ), DebugSymsHeap.can_be_resolved will return false.At this case, the first check in OnlyWithResolvedHeapSyms would be triggered and output a  `The currently active libc isn't glibc. It's unknown.` rather than `The heap is not initialized`
```python
if pwndbg.libc.which() != pwndbg.libc.LibcType.GLIBC:
    e(f"The currently active libc isn't glibc. It's {pwndbg.libc.which().value}.")
            return None
```
And because the switch of current, it also output some warinings, but when user execute heap twice, it has gone.
```bash
pwndbg> heap

heap: An unknown error occurred when running this command.
Maybe you can try to determine the libc symbols addresses manually, set them appropriately and re-run this command. For this, see the `heap-config` command output and set the `main_arena`, `mp_`, `global_max_fast`, `tcache` and `thread_arena` addresses.
Set print a full stacktrace for exceptions raised in Pwndbg commands to 'on'.
pwndbg> heap
heap: Heap is not initialized yet.
pwndbg> heap
heap: Heap is not initialized yet.
pwndbg> heap
heap: Heap is not initialized yet.
```
I'm not sure the explaination above could make you clear @k4lizen, if you need more information or something else, i would offer.And after your reivew, i could modify the comment or choose another patch

k4lizen:
but how come this never surfaced before if it has always been like this, shouldn't we just permanently be failing on all heap commands when we get a stripped glibc?

M4idCat:
> but how come this never surfaced before if it has always been like this, shouldn't we just permanently be failing on all heap commands when we get a stripped glibc?

This is the fact, every srtipped glibc would failed now. This is a regression bug after #3929, and It has been only three months so far. I think it is a coincidence that most users use the stable distribution with debuginfo libc, except the rolling release like arch etc. And the most important is that all the test didn't find out it.

M4idCat:
i'm not familiar with the whole test process,  i think it need an core contributor to add testing.But back to the heap command, i need your advice. @k4lizen Would you prefer the order of decorator or add defence check in decorator

k4lizen:
Hi, thank you for your investigation. You were right in the cause of the issue. I wanted to take some time to properly address it, and am doing so in https://github.com/pwndbg/pwndbg/pull/4126 . I added you as a co-author so you will be co-author of that PR.

I am closing this PR in favour of that one.

M4idCat:
OK, i'm looking forward to your fix!

PR #4092, opened 2026-08-25
Disassembly fixes for instructions seen multiple times

OBarronCS:
tl;dr caching is hard, and this will be the first of a couple more fixes to how we track the emulated flow of instructions.

I came across a bug in handling x86 `RET` that resulted in the disassembly incorrectly following the emulated flow of instructions. This occurs when the same `ret` instruction is executed multiple times in the future flow of instructions, which is not too uncommon in long ROP chains and happens in retsleds, where you have back to back ret  gadgets.

This resulted in broken disassembly while the CPU was paused on this instruction.

Basically, we previously had a check for `instruction.address == pwndbg.aglib.regs.pc`, which signaled that we are currently paused on this instruction and thus can read process state during enhancement (because the current register set/memory is the state that the `RET` instruction will access when it runs). This is an incorrect assumption, because during disassembly, we may encounter the same instruction address in the future multiple times due to loops / ROP. This is now handled correctly.

I added tests for this.

Double `ret` instruction before (note that the first `ret` causes program counter to jump back to the same instruction address, to execute another ret):
double_ret_did_not_used_to_work

After:
double_ret_works_now

Other case (we breakpoint at the first ret and continue there. Then we step 1 instruction):

Before, history would be backfilled with a bunch of rets, which was incorrect
xuntil_xret_and_step_before

After:
xuntil_xret_and_step_works

However, while making tests for this, I encountered other quirks which mess with how we track the flow of instructions due a multi-leveled caching system, which will require larger changes.

Future changes include:
1.  Currently, every call to display disassembled code goes through one function, `near()`, which is used for too many purposes and has outgrown it's use cases (at least 4: context disasm with emulation, without emulation, nearpc command, and emulate command). This function computes an in-place mutation of the cache state, which is bad. This messes with the caching (ie when running `nearpc`, this modifies state that affects the `ctx disasm` display!). This will be fixed.
2. Even more tests for the disassembly system, across the multiple arches Pwndbg supports.

### DIFF
--- a/pwndbg/aglib/disasm/instruction.py
+++ b/pwndbg/aglib/disasm/instruction.py
@@ -192,6 +192,17 @@ class DisassemblySource(Enum):
     DEBUGGER = auto()
 
 
+# This is used for debugging issues in the disassembly display when instructions behind
+# the program counter are unexpectedly appearing
+class CacheSource(Enum):
+    LINKED_LIST_DYNAMIC = "L"
+    FALLBACK_DYNAMIC = "F"
+    CACHE_LINEAR = "C"
+    HEURISTIC_LINEAR = "H"
+
+    NOT_FROM_CACHE = ""
+
+
 # Interface for enhanced instructions - there are two implementations defined in this file
 class PwndbgInstruction(Protocol):
     cs_insn: CsInsn
@@ -221,6 +232,8 @@ class PwndbgInstruction(Protocol):
 
     enhanced: bool
     disassembly_source: DisassemblySource
+
+    cache_source: CacheSource
 
     @property
     def call_like(self) -> bool: ...
@@ -443,6 +456,12 @@ class PwndbgInstructionImpl(PwndbgInstruction):
         """
         Mapping of Capstone register id to integer value. During enhancement, we might manually determine
         that an instruction writes some value to a register, and this is stored here.
+        """
+
+        self.cache_source = CacheSource.NOT_FROM_CACHE
+        """
+        This is purely for debugging.
+        This may change during the lifetime of this instruction, as this value is set during the latest time this instruction was displayed.
         """
 
     @property
@@ -792,6 +811,8 @@ class ManualPwndbgInstruction(PwndbgInstruction):
         self.emulated = False
 
         self.register_writes = {}
+
+        self.cache_source = CacheSource.NOT_FROM_CACHE
 
     @property
     def bytes(self) -> bytearray:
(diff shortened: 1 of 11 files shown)


pwndbg/commands/branch.py
```diff
@@ -28,7 +28,7 @@ def should_stop(self):
         # where previously it was not. The enhancement process will figure out if all the conditions
         # this branch requires in order to be taken have been met.
         assistant = pwndbg.aglib.disasm.disassembly.get_disassembly_assistant_for_current_arch()
-        assistant.enhance(self.instruction)
+        assistant.enhance(self.instruction, current_cpu_instruction=True)
```
OBarronCS, line 31: I want to note that there are a couple places in the codebase, here and in the `nextret` family of instruction, that do something like this:

```python
ins = pwndbg.aglib.disasm.disassembly.one(pwndbg.aglib.regs.pc)
# check some value of it
check ins.next
or
check ins.is_conditional_jump_taken
```

It disassembles an instructions and reads some computed property on it. However, they are implicitly relying on the fact that since we are disassembling at the current CPU program counter, the enhancement process can determine a lot more information about it (we are able to determine if indeed the branch is taken, or what the jump target is, for example), which would not be possible otherwise.

Adding `current_cpu_instruction=True` makes this intention more explicit.

OBarronCS: Due to other bugs that need to be addressed anyways in a future PR (see the `test_emulate_nextret_to_double_ret` test), I haven't gone and changed any of the `nextret` functions yet.

pwndbg/aglib/disasm/instruction.py
```diff
@@ -192,6 +192,16 @@ class DisassemblySource(Enum):
...
 
+# This is used for debugging issues in the disassembly display when instructions behind
+# the program counter are unexpectedly appearing
+class CacheSource(Enum):
+    LINKED_LIST_DYNAMIC = "L"
+    FALLBACK_DYNAMIC = "F"
+    CACHE_LINEAR = "C"
+
+    NOT_FROM_CACHE = ""
+
```
OBarronCS, line 203: Added this: it is immensely helpful when debugging. It adds a letter before instructions to indicate how an instruction was retrieved

Image

Enabled with:
```sh
pi pwndbg.color.disasm.DEBUG_CACHE = True
```

k4lizen:
awesome stuff, thanks :D

PR #4096, opened 2026-08-27
Fix nextret when it can encounter a loop

OBarronCS:
#2884 introduced a subtle bug that caused nextret to get stuck on certain types of loops.

If you had a binary like the following, where a sequence the instruction in the loop ends in the branch, that can either fall through or not, and there were no intermediate branches, nextret would get stuck forever.

```
xor ecx, ecx

loop:
    inc ecx
    cmp ecx, 4
    jl loop
ret

```

This is fixed and a test, and I added a test for it

### DIFF
--- a/pwndbg/aglib/next.py
+++ b/pwndbg/aglib/next.py
@@ -57,6 +57,8 @@ def next_branch(address=None, including_current=False) -> PwndbgInstruction | No
     Return the next branch instruction that the process will encounter with repeated usage of the "nexti" command.
 
     If including_current == True, then if the instruction at the address is already a branch, return it.
+    
+    Returns the next branch instruction, or None if cannot disassemble anymore.
     """
     if address is None:
         ins = pwndbg.aglib.disasm.disassembly.one(pwndbg.aglib.regs.pc)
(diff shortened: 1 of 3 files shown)


pwndbg/aglib/next.py
```diff
@@ -131,7 +133,7 @@ async def break_next_branch(
     inf = pwndbg.dbg.selected_inferior()
     # If the branch we found was not at the current program counter, we should step to it.
     # Otherwise, return the current instruction.
-    if ins.address != pwndbg.aglib.regs.pc:
+    if not including_current or ins.address != pwndbg.aglib.regs.pc:
```
OBarronCS, line 136: This function should return with the current instruction only in the case that we passed in `including_current=True`, and we are currently at a branch.

In all other cases, we can assume the `ins` is not the current instruction, and that we should step to it.

OBarronCS:
Unrelated test failure:

image

disconnect3d:
Do I understand correctly that this was a disassembly bug?

EDIT: Ah, no, its a bug in break_* func. I see the comment now. I think it makes sense. Lets merge.

disconnect3d:
Thanks!

PR #4099, opened 2026-08-28
Remove hardcoded tcache dereference limit

Eirikwst:
Small change (glibc 2.43 heap)

PR #650 introduced a separate limit for tcache bins because heap-dereference-limit defaulted to 5 while tcache bins could contain up to 7 entries.

glibc 2.43 increased the default tcache bin length to 16, so tcache rendering should probably follow the configured deref-limit instead of using its own hardcoded limit.

Also add a regression test for tcache bins with more than 8 entries.

image

### DIFF
--- a/tests/library/dbg/tests/test_heap_bins.py
+++ b/tests/library/dbg/tests/test_heap_bins.py
@@ -205,6 +205,38 @@ async def test_heap_bins(ctrl: Controller) -> None:
 
 
 @pwndbg_test
+async def test_tcache_bins_respects_heap_dereference_limit(ctrl: Controller) -> None:
+    """Ensure tcache rendering uses heap-dereference-limit for chains longer than seven."""
+    import pwndbg.aglib.heap
+    from pwndbg.aglib.heap.ptmalloc import GlibcMemoryAllocator
+
+    await ctrl.execute("set context-output /dev/null")
+    await ctrl.launch(BINARY, env={"GLIBC_TUNABLES": "glibc.malloc.tcache_count=16"})
+
+    # Free 11 chunks into one tcache bin
+    break_at_sym("breakpoint")
+    await ctrl.cont()
+    await ctrl.cont()
+    await ctrl.cont()
+
+    assert isinstance(pwndbg.aglib.heap.current, GlibcMemoryAllocator)
+    allocator = pwndbg.aglib.heap.current
+
+    await ctrl.execute("set heap-dereference-limit 12")
+
+    bins = allocator.tcachebins()
+    assert bins is not None
+    tcache_bin = bins.bins[allocator._request2size(0x20)]
+    assert tcache_bin.count == 11
+    assert len(tcache_bin.fd_chain) == 12
+    assert tcache_bin.fd_chain[-1] == 0
+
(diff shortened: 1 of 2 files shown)


disconnect3d:
LGTM; cc @k4lizen @CptGibbon - is that fine?

tests/library/dbg/tests/test_heap_bins.py
```diff
@@ -204,6 +204,38 @@ async def test_heap_bins(ctrl: Controller) -> None:
...
 
 
+@pwndbg_test
+async def test_tcache_bins_respects_heap_dereference_limit(ctrl: Controller) -> None:
+    """Ensure tcache rendering uses heap-dereference-limit for chains longer than seven."""
+    import pwndbg.aglib.heap
+    from pwndbg.aglib.heap.ptmalloc import GlibcMemoryAllocator
+
+    await ctrl.execute("set context-output /dev/null")
+    await ctrl.launch(BINARY, env={"GLIBC_TUNABLES": "glibc.malloc.tcache_count=16"})
```
k4lizen, line 216: why are we setting the tunable if glibc 2.43 increased it by default?

tests/library/dbg/tests/test_heap_bins.py
```diff
@@ -204,6 +204,38 @@ async def test_heap_bins(ctrl: Controller) -> None:
     await ctrl.execute("bins")
 
 
+@pwndbg_test
+async def test_tcache_bins_respects_heap_dereference_limit(ctrl: Controller) -> None:
+    """Ensure tcache rendering uses heap-dereference-limit for chains longer than seven."""
+    import pwndbg.aglib.heap
+    from pwndbg.aglib.heap.ptmalloc import GlibcMemoryAllocator
+
+    await ctrl.execute("set context-output /dev/null")
```
k4lizen, line 213: why :?

k4lizen (approved):
makes sense to me, just curious about the test

tests/library/dbg/tests/test_heap_bins.py
Eirikwst: Two other tests testing heap dereferencing do this, so I kind of just assumed it was standard. My guess was keeping logs cleaner?

tests/library/dbg/tests/test_heap_bins.py
Eirikwst: AFAIK the 2.43 tests only run on if 2.43 is available and this is *techincally* not a 2.43 specific change, which is why I made them for the BINARY and not GLIBC_2_43 test binary.
Also clarifies that this test is for the non-7 size.

tests/library/dbg/tests/test_heap_bins.py
k4lizen: oki

tests/library/dbg/tests/test_heap_bins.py
k4lizen: makes sense, but can you add a comment here mentioning the glibc 2.43 change?

Eirikwst: Done 👍

k4lizen:
Thanks for the PR!

PR #4101, opened 2026-08-29
nextproginstr bug fix

OBarronCS:
I came across a bug where if we are currently on a `ret` instruction, then `nextproginstr` does not work. This was a small logic bug (explained in a comment below), which is now fixed + a test added.

### DIFF
--- a/pwndbg/aglib/next.py
+++ b/pwndbg/aglib/next.py
@@ -24,22 +24,24 @@ from pwndbg.dbg_mod import BreakpointLocation
 interrupts = {CS_GRP_INT}
 
 
-def next_int(address=None, honor_current_branch=False):
+def next_int(address: int | None = None, honor_current_branch: bool = False):
     """
-    If there is a syscall in the current basic black,
-    return the instruction of the one closest to $PC.
+    Start at `address`, disassembling forwards to find the next syscall.
+    If there is a branch before the syscall, return None.
 
     If honor_current_branch is True, then if the address is already a branch, return None.
 
     If no interrupt exists or a jump is in the way, return None.
     """
     if address is None:
-        ins = pwndbg.aglib.disasm.disassembly.one(pwndbg.aglib.regs.pc)
-        if not ins:
-            return None
-        if honor_current_branch and ins.jump_like:
-            return None
-        address = ins.next
+        address = pwndbg.aglib.regs.pc
+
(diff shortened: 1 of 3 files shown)


pwndbg/aglib/next.py
```diff
@@ -300,7 +310,7 @@ async def break_on_program_code(ec: pwndbg.dbg_mod.ExecutionController) -> bool:
         if pwndbg.aglib.proc.stopped_with_signal():
             return False
 
-        await break_next_ret(ec)
+        await break_next_ret(ec, including_current=True)
         await ec.single_step()
```
OBarronCS, line 323: The logic here is the "step to the next ret, because this ends the function, and step one instruction to leave the function". If we are currently at a ret, we should stay there, and then do a single step.

Prior to this PR, if we were at a `ret`, we wouldn't stay there, and instead the `break_next_ret` would disassemble and put a break at an unrelated `ret` instruction.

pwndbg/aglib/next.py
```diff
@@ -24,22 +24,24 @@
...
-            return None
-        address = ins.next
+        address = pwndbg.aglib.regs.pc
+
+    ins = pwndbg.aglib.disasm.disassembly.one(address)
+    if not ins:
+        return None
+    if honor_current_branch and ins.jump_like:
+        return None
+    address = ins.next
```
OBarronCS, line 44: The logic here was bugged (but was never encountered in the current codebase because of how we call `next_int`). Previously, only if we passed in `None` as the address would the `honor_current_branch` condition be checked. This should happen no matter what address we provide.

I fixed this same bug in `next_branch`. No code in pwndbg actually ever calls these functions with an explicit `address` currently, so the bug was never seen at runtime.

pwndbg/commands/next.py
```diff
@@ -78,15 +80,17 @@ async def _stepret(ec: pwndbg.dbg_mod.ExecutionController):
     """
     Execution controller for the `stepret` command.
     """
-    while (
-        pwndbg.aglib.proc.alive()
-        and not (await pwndbg.aglib.next.break_next_ret(ec))
-        and (await pwndbg.aglib.next.break_next_branch(ec))
+    first = True
+    while pwndbg.aglib.proc.alive() and (
+        ins := await pwndbg.aglib.next.break_next_branch(ec, including_current=True)
```
OBarronCS: `stepret` was also not working, so I fixed it along the way.

Issue #3749, opened 2026-02-15
fail assert on subcommands with no help


k4lizen:
Consider something like this:

<img width="759" height="228" alt="Image" src="https://github.com/user-attachments/assets/719db545-5111-44b8-b405-d745afa7b434" />

Each of these subcommands should have a line, just like explain does. This line is generated by the `help` argument of `add_parser`:
```python
explain_parser = subparsers.add_parser(
    "explain",
    description="Gives a quick explanation of musl's mallocng allocator.",
    help="Gives a quick explanation of musl's mallocng allocator.",
)
```
This should be checked and asserted on somewhere in `pwndbg/commands/__init__.py:initialize_parser_recursively()`. See relevant comment:
```python
            if not isinstance(action, argparse._SubParsersAction) and action.help is None:
                # When we do `cmd -h` we want each argument to have a one-line
                # description.
                # Unfortunately, I don't know how to enforce that each subcommand has a help=
                # passed to its add_parser() :(
                print(message.error(f"Error parsing arguments for command: {parser.prog}"))
                print("You must add a `help=` string to your argument.")
                print(f"Erroneous action:\n\t{repr(action)}\n")
                assert False, "You must add a `help=` string to your argument."
```

k4lizen:
Likely the assert should check that the first line of the description is equal to the help.

PR #4103, opened 2026-08-30
Fix help subcommands

marywinter3:
Screenshot from 2026-06-14 21-55-18

The new assertion correctly detects subcommands missing help= . When help= is removed from a subcommand, pwndbg fails to load with a clear error message pointing to the problematic subcommand.
After adding help= back, pwndbg loads without errors.


disconnect3d:
Hi, what's the actual bug here? When does this happen?

Isn't it weird that we do `description=x, help=x` with the same text? Can't we just hijack it under the hood somehow -- if this hack is needed at all?

k4lizen:
It's kinda weird yeah but they don't need to be the same.

image

pwndbg/commands/profiler.py
```diff
@@ -19,9 +19,14 @@
 """
 )
 subparsers = parser.add_subparsers(dest="command")
-subparsers.required = True
```
k4lizen, line 22: why delete this?

k4lizen (changes requested):
i don't really understand this code, could you explain how it works? (and then add comments to it i suppose?)

marywinter3:
This PR resolves issue #3749. The point was that subcommands should have description in the list shown by --help, generated from the help= argument of add_parser(). For regular arguments pwndbg already enforced the presence of help=, but for subcommands there was no such check - there was only a comment saying it wasn't clear how to enforce it. 

Closes #3749

pwndbg/commands/profiler.py
marywinter3: This was accidentally removed line

k4lizen (approved):

k4lizen:
LGTM, thanks!

PR #4111, opened 2026-09-01
Feature Add Instruction-Trace Command

latortuga71:

Hello. First of all thank you for such a great project. I and many others i work with love using this project whenever we can.
I would like to propose a new command, simply put it will create an instruction trace file and write it to disk using single steps to step until a specific address is hit or the program exits. 

I think this feature would be useful in cases when doing recon on libraries for example. Determining what functions inside Macos ImageIO Framework are being called when decoding a specific type of image.

There are downsides it is slow due to the single stepping. But it is up to the user to know what distance they need to trace.

Here is what the command would look like.
`instr-trace /tmp/trace.txt 0x555555555164`

The -v flag can be added to include the path to the file backing the memory region.
`instr-trace /tmp/trace.txt 0x555555555164 -v`

Command Help Output

image

Example of trace output

image
 
 Please let me know if there are any questions.  I have ran the linter and passed all the tests.


k4lizen (commented):
Heya! Thanks for the PR!

If your goal is to trace which functions get executed, wouldn't it be better to just output the unique function enters rather than each instruction?

k4lizen:
I would spruce up the output as well, if we're tracing functions I think it'd also be cool to retreive their arguments and put that into the output too

latortuga71:
Hmm what i had in mind was more general, such as just getting all executions until a particular end address. So you can trace exactly whats happening. Getting unique instruction calls sounds interesting. Is there a way using the abstract debugger API to get when a function is called?  
From my understanding i think we would need to decode each call instruction per cpu platform?

Also to be clear the actual output goes to a file and isn't shown in the debugger that screenshot is a cat of the file that is output.

disconnect3d:
I think that tracing individual instructions make sense and it'd be nice if you could get not just instruction dump but any other data u want like register value or sth? Maybe this should accept some format string to control the printed data?

k4lizen:
> Hmm what i had in mind was more general, such as just getting all executions until a particular end address. So you can trace exactly whats happening.

Right but what is the benefit of this? You don't see which instructions themselves are being executed anyway.

> Getting unique instruction calls sounds interesting. Is there a way using the abstract debugger API to get when a function is called?

Not sure, look into the `nextcall` family of commands.

> Also to be clear the actual output goes to a file and isn't shown in the debugger that screenshot is a cat of the file that is output.

Yeah I understand. Do you have any thoughts on why not output to both the debugger and the file?

latortuga71:
> > Hmm what i had in mind was more general, such as just getting all executions until a particular end address. So you can trace exactly whats happening.
> 
> Right but what is the benefit of this? You don't see which instructions themselves are being executed anyway.
> 
> > Getting unique instruction calls sounds interesting. Is there a way using the abstract debugger API to get when a function is called?
> 
> Not sure, look into the `nextcall` family of commands.
> 
> > Also to be clear the actual output goes to a file and isn't shown in the debugger that screenshot is a cat of the file that is output.
> 
> Yeah I understand. Do you have any thoughts on why not output to both the debugger and the file?

The benefit (at least for me) is i want to know **what the general execution flow of something is quickly and easily.** 


For example i already have the library loaded into binary ninja. But im not sure what functions are being executed. With this i can run it and it will give me a dump of the execution flow. 

And lets say i have a different input i can run it with that new input and check what the difference between the two traces was. 

I don't necessarily need to know exactly what instruction was executed i can just look into bninja using the offsets. But i think that would be a cool add as well. Let me know if this makes sense.

Also i didn't output to the debugger because im not sure how to do that lol. (Plus its usually alot of output) Currently it single steps and prints out like if a human was single stepping. Ideally it wouldn't do that but im not sure how to clear the debugger input and make it show what i want etc.

pwndbg/commands/instr_trace.py
```diff
@@ -0,0 +1,87 @@
...
+    print(message.info(f"instr-trace: Dumped {count} instructions"))
+
+
+@pwndbg.commands.Command(
+    parser,
+    category=pwndbg.commands.CommandCategory.MISC,
+    examples="instr-trace /tmp/trace.txt 0x555555555158 -v",
+)
+@pwndbg.commands.OnlyWhenRunning
+def instr_trace(trace_file: str, end_address: int, verbose: bool) -> None:
```
k4lizen: Hm so we already have some tracing commands, e.g. `track-heap` and `track-got`, but I do think your naming scheme is more memorable so I like the command name `instr-trace`.

pwndbg/commands/instr_trace.py
```diff
@@ -0,0 +1,87 @@
...
+    vmmap_object = pwndbg.aglib.vmmap.get_memory_map()
+
+    async def dump_trace_log(ec: pwndbg.dbg_mod.ExecutionController) -> None:
+        with open(trace_file, "w+") as f:
+            if not verbose:
+                await _step_and_log(ec, file_handle=f, vmmap_object=None, end_address=end_address)
```
k4lizen: Image

wrong type in function

pwndbg/commands/instr_trace.py
```diff
@@ -0,0 +1,87 @@
...
+)
+
+
+async def _step_and_log(
+    ec: pwndbg.dbg_mod.ExecutionController,
+    file_handle: io.TextIOWrapper,
+    vmmap_object: pwndbg.dbg_mod.MemoryMap,
+    end_address: int | None = None,
+) -> None:
+    if end_address == 0:
```
k4lizen: why this special case?

pwndbg/commands/instr_trace.py
```diff
@@ -0,0 +1,87 @@
...
+        return
+    count = 0
+    while pwndbg.aglib.proc.alive():
+        current_pc = pwndbg.aglib.regs_mod.regs.pc
```
k4lizen, line 38: `current_pc` may be None here, use a python type checker (e.g. mypy or pyright) in your editor plz

pwndbg/commands/instr_trace.py
```diff
@@ -0,0 +1,87 @@
...
+            break
+        current_symbol = pwndbg.aglib.symbol.resolve_addr(int(current_pc))
+        await ec.single_step()
+        if vmmap_object is not None:
+            page_ = vmmap_object.lookup_page(current_pc)
+
+            if page_:
+                log_line = f"[{hex(current_pc)}]!{current_symbol} : {page_.objfile:<8}\n"
+            else:
+                log_line = f"[{hex(current_pc)}]!{current_symbol}\n"
```
k4lizen: should enforce it to be front-padded so there are at least 6 bytes shown, so output is aligned if we get shorter PCs.

maybe we'd like the output to look something like
```
[0x7ffff7c3e221] (libc.so.6       ) __sigsetjmp+17
                              ^^padding so we're aligned with longer .so filenames
```
and then a flag like `--full-path` which shows to full path to the mapping instead of just the file name. cuz i think usually you only are about the filename.

k4lizen (commented):
aight fair enough, i see your point. i think we should spend a bit more time on what the output to the file looks like but I'm a bit too sleepy to think about it rn. maybe take some inspiration from strace

pwndbg/commands/instr_trace.py
latortuga71: Sounds good ill work on these fixes today

latortuga71: @k4lizen i have added 3 types of output 

`lighthouse` which is compatible with the [lighthouse coverage tool](https://github.com/gaasedelen/lighthouse) Which is module+offset as shown below

```
libc.so.6+0x131650
libc.so.6+0x131647
libc.so.6+0x131645
libc.so.6+0x13163e
testlib.so+0xe94
testlib.so+0xe93
testlib.so+0xe92
```

Also i have added `full`

which adds symbols as well (Inspired by [symbolizer-rs](https://github.com/0vercl0k/symbolizer-rs))
```
libc.so.6!printf+171
libc.so.6!printf+176
libc.so.6!printf+185
libc.so.6!printf+187
libc.so.6!printf+194
testlib.so!SomeExternalLibrary+99
testlib.so!SomeExternalLibrary+100
testlib.so!SomeExternalLibrary+101
```

And finally `raw` which is just the pc

```
0x7ffff7dfde3e
0x7ffff7dfde40
0x7ffff7dfde42
0x7ffff7dfde44
0x7ffff7df29ab
0x7ffff7df29b0
0x7ffff7df29b9
0x7ffff7df29bb
0x7ffff7df29c2
0x7ffff7fbb16c
0x7ffff7fbb16d
0x7ffff7fbb16e
```
Also i have tested on an arm64 macbook and it's what i was expecting,  outputs such as 

```
ImageIO!PNGReadPlugin::Get_colorspace(ReadPluginData const&, PNGPluginData&, png_struct_def*, png_info_def*, double*)+400
ImageIO!PNGReadPlugin::Get_colorspace(ReadPluginData const&, PNGPluginData&, png_struct_def*, png_info_def*, double*)+408
ImageIO!_cg_png_get_valid
ImageIO!_cg_png_get_valid+4
ImageIO!_cg_png_get_valid+8
ImageIO!_cg_png_get_valid+12
libPng.dylib!_cg_png_get_valid
libPng.dylib!_cg_png_get_valid+4
libPng.dylib!_cg_png_get_valid+8
```
Let me know your thoughts.

k4lizen: I like the lighthouse support, the raw format definitely should not be the default, I think the readability of the `symbolizer-rs` format is poor and that a) you shouldn't use `!` as the delimiter but rather a couple of whitespaces b) you should try to get the function names aligned vertically similarly to what i described above

latortuga71: sounds good, i will work on that

latortuga71: @k4lizen would this formatting work for the full output? also i have set full to be the default trace format.

```
libc.so.6           __printf_buffer+305
libc.so.6           __printf_buffer+311
libc.so.6           __printf_buffer+318
libc.so.6           __printf_buffer+319
libc.so.6           __printf_buffer+320
libc.so.6           __printf_buffer+322
libc.so.6           __printf_buffer+324
libc.so.6           __printf_buffer+326
libc.so.6           __printf_buffer+328
libc.so.6           __vfprintf_internal+545
libc.so.6           __vfprintf_internal+548
libc.so.6           __printf_buffer_to_file_done
libc.so.6           __printf_buffer_to_file_done+3
libc.so.6           __printf_buffer_to_file_done+5
libc.so.6           __printf_buffer_to_file_done+7
libc.so.6           __printf_buffer_to_file_done+8
libc.so.6           __printf_buffer_to_file_done+11
libc.so.6           __printf_buffer_flush_to_file
```

k4lizen: yeah looks good to me

PR #4112, opened 2026-09-01
Fix GDB internally calling malloc while we iterate stack symbols, causing crash

OBarronCS:
While stepping through the early instructions of a binary compiled by Zig, I came across a bug where upon entering a function, GDB would suddenly seemingly jump to a completely unrelated function and segfault.

We are here:
image

Then `si`, and it suddenly crashes here:
image

At the moment of the crash, if we do `bt`, we can see how it got here:
image

Turns out that while we are iterating stack variables in a stack frame for the purposes of displaying variables names (associated with an address), we unknowingly may cause GDB internally to invoke `malloc` within the inferior process. It calls malloc within the process for reasons described below, but it segfaults within malloc almost immediately, since the library hasn't been initialized yet. The malloc implementation assumes that things like thread local memory (FS register) have been setup.

The root cause of this bug starts around here. While displaying integers and pointers to the display, we "enhance" them. One way we do so is to see if the pointer is to a local variable, so we can display the local variable name. Upon entering a new frame, we iterate all symbols in it in this function:

https://github.com/pwndbg/pwndbg/blob/2267f2eb102c53552900129224873481995654da/pwndbg/dbg_mod/gdb/__init__.py#L121-L141

One of the checks in this function is:
```python
if value.address is None:
    continue
```
This `.address` lookup, in certain cases, will cause GDB to invoke malloc in the process. This Python attribute lookup invokes a function in GDB, [valpy_get_address](https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=gdb/python/py-value.c;h=5d8fab90c04c4b5c98d9efc60700ac5f503138d2;hb=refs/heads/gdb-17-branch#l402). 

This function will try to resolve the address of the variable in question, calling [value_coerce_to_target](https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=gdb/valops.c;h=85d69fb11f27e283a8a400ae0205df022c8499d2;hb=refs/heads/gdb-17-branch#l1467) for some variable types.

If the variable type pass [this check](https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=gdb/valops.c;h=85d69fb11f27e283a8a400ae0205df022c8499d2;hb=refs/heads/gdb-17-branch#l1439) (it's an array or string that is an `lvalue`, basically an object that is not backed by memory of any kind currently)

If this is the case, it internally calls [value_allocate_space_in_inferior](https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=gdb/valops.c;h=85d69fb11f27e283a8a400ae0205df022c8499d2;hb=refs/heads/gdb-17-branch#l165), which calls malloc within the inferior process!
```c
struct value *
value_allocate_space_in_inferior (int len)
{
struct objfile *objf;
struct value *val = find_function_in_inferior ("malloc", &objf);
struct gdbarch *gdbarch = objf->arch ();
struct value *blocklen;
...
```
Presumably it is doing this to be able to actually return an address for the variable in question (after placing a value into it). It is basically "constructing" a location that the variable can be referenced at.

But now this calls `malloc` within the inferior, but the libc library has not been setup yet! So the `FS` register as not be set to the thread local memory region yet, and it crashes when it is referenced.

image

This then results in a segfault and stops the program.

This PR fixes this by checking if the variable is optimized out (meaning it definitely has no memory location) before calling `value.address`, which could invoke this long chain that calls `malloc` under the hood.

# Replicate
To replicate the issue (which crashes without this PR), you can:

Running with zig version `0.16.0`
// repro.zig file
```
pub fn main() !void {
}
```
```sh
# Build
zig build-exe repro.zig -lc -target x86_64-linux -O ReleaseSafe

# Debug it
gdb ./repro

# Within gdb, run `starti`, then `si` around 30 times until you get to the memset. Step into the memset and observe the crash.
```

If this is ever encountered in other contexts, you can also fix this bug by setting this gdb setting:
```sh
set may-call-functions off
```
This disables GDB calling functions in the inferior in any case.

### DIFF
--- a/pwndbg/dbg_mod/gdb/__init__.py
+++ b/pwndbg/dbg_mod/gdb/__init__.py
@@ -126,6 +126,9 @@ def _get_frame_stack_variables(frame: gdb.Frame) -> tuple[tuple[int, int, str], 
             try:
                 value = sym.value(frame)
 
+                # If a variable is optimized out, it surely has no memory location
+                # This check also prevents GDB from calling `malloc` within the inferior process
+                # which can cause a crash. See https://github.com/pwndbg/pwndbg/pull/4112
                 if value.is_optimized_out:
                     continue
 
--- a/pwndbg/dbg_mod/gdb/__init__.py
+++ b/pwndbg/dbg_mod/gdb/__init__.py
@@ -125,6 +125,10 @@ def _get_frame_stack_variables(frame: gdb.Frame) -> tuple[tuple[int, int, str], 
 
             try:
                 value = sym.value(frame)
+
+                if value.is_optimized_out:
+                    continue
+
                 # value.address can be None
                 # https://sourceware.org/gdb/current/onlinedocs/gdb.html/Values-From-Inferior.html#Values-From-Inferior:~:text=Variable%3A%20Value%2Eaddress
                 # https://sourceware.org/bugzilla/show_bug.cgi?id=33860


OBarronCS:
I suppose it just so happens that Zig emits variables in `memset` that match the exact criteria to GDB invoking `malloc`, which caused this bug.

k4lizen (approved):
Thanks for the thorough investigation. Maybe we could also check `.is_unavailable`?

k4lizen:
Also I guess there are other places in the code where this should be added...

k4lizen:
Could you report this to upstream? I don't think they are aware their logic can go this badly.

OBarronCS:
> Maybe we could also check .is_unavailable?

Hmmm I think this check may prevent displaying some variable names in some cases. I didn't see this  .is_unavailable attribute at first, but now I see it in the docs. One example they gave is this is used for corefiles: for example, we may know the location in memory of a variable, just not be able to access it (in the corefile, the pointer still exists, but not the underlying value since that specific page was not saved to the coredump). We still get the address, so we can still show the variable name associated with the address, which I think we still want.

OBarronCS:
> Also I guess there are other places in the code where this should be added...
Yes, it may be possible that other places in the code might accidentally trigger this if they enumerate/query a bunch of symbols from GDB.

One idea I considered was a context manager, that was like

```python
with context with "set may-call-functions off":
     code_here

// and when the context manager exits, the `set may-call-functions` is reset to what is was before starting the block
```
And then wrap any code that may invoke this with this context manager.

k4lizen:
> One idea I considered was a context manager, that was like

Makes sense, though it's not entierly clear to my head when we would want to use this.

Like, if I set this, does this mean GDB will not be able to retreive the addresses of some variables of which it otherwise would be able to?

OBarronCS:
The details as to why GDB calls malloc in these scenarios is a little unclear to me still. I understand that GDB would need to call malloc if someone does GDB command `call puts(
"this is a string")`. It needs to be the string in the process memory, and allocate space for it.

In the case of what was encountered in this PR, GDB is seemingly artificially creating a memory location for some variable that we are trying to lookup, for the sake of say "yes, here is where you can read from memory to read the variable you are trying to read" (if this variable otherwise never lived in memory during program execution).

The returned address in these case will always be in the heap.

I can't imagine any scenario where we would want this behavior to occur. If a variable/symbol never actually existed in memory, then we wouldn't want it to artificially create a location for it. But GDB seems to do it if a `gdb.Value`'s address is looked up, because it doesn't know in what context we are looking up the address. It may assume "hey, the user program is asking for an address, surely they want to dereference this address" (which we are NOT doing in this case).

I think for most context, we are looking at gdb.Values that will not trigger this scenario. This check in GDB source code needs to pass: https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=gdb/valops.c;h=85d69fb11f27e283a8a400ae0205df022c8499d2;hb=refs/heads/gdb-17-branch#l1439. We have a variable which is an `lval` and also a string or an array.

k4lizen:
I don't think its about returning a fake address, i think its about GDB expecting to modify the var, see doc
```
/* Return true if VAL does not live in target memory, but should in order
   to operate on it.  Otherwise return false.  */

extern bool value_must_coerce_to_target (struct value *arg1);
```

k4lizen:
But yeah its very weird, plz report :p

PR #4113, opened 2026-09-01
Fix checkmark/x for certain aarch64 conditional instructions

OBarronCS:
The checkmark/x that is marked for branches/conditional instructions was incorrect for certain AArch64 conditional instructions which we don't know the results of yet (because they are after a function call, which we do not emulate through).

These instructions should be marked with a question mark, but instead were defaulting to an `X`.

Before:
image

After:
image


OBarronCS:
Some conditional branches (like b.eq) didn't have a question mark:

Before:
image

After:
image

OBarronCS:
I also found that `ret` was assumed to be conditional:

image

k4lizen (approved):
i love changes like these which stack up and provide a much better disasm experience :>

disconnect3d:
Awesome, thanks!

PR #4122, opened 2026-09-04
Add branch target in disassembly for indirect jmps that dereference writable memory

### LINKED ISSUE #4119: Disassembly jump targets: displaying memory address containing function pointer when content is unknown
In disassembly, we display the target of branches with `<concrete_target>`, such as

```
ret                                <_dl_start+1489>

call   qword ptr [rcx]             <frame_dummy>

call   qword ptr [rip + 0xaa81]    <setlocale>
```

For the first two cases, only at runtime do we know that address, since we either read it from the top of the stack, or from a location in memory defined by a register.

In the third case, the address calculated by `rip + 0xaa81` is constant, and we know in this case that it points to readonly memory. Because of this, we can know statically that the target is `setlocale`, since we can dereference the address to get the function pointer no matter where the current program counter is.

But in some cases, although the address of the function pointer is constant, the contents is in writeable memory, so we cannot read it in a static context, such as when disassembling with `nearpc`.

This pattern is common with the `plt` in cases where the GOT is writeable (see all the jumps with no target indicated):.

Image

This is proposing a way to handle these cases in a nicer way when disassembling in a static context. Currently, because we cannot resolve the target, we show no target when doing `nearpc` at these address. Of course, at runtime with emulation, we are able to resolve the concrete address

When disassembling these types of jumps with nearpc, we can check if:
- The address is constant
- But the address is writeable, so the contents may change so we cannot read it to get a concrete address.

In this scenario, it may be nice to do something like 

```
jmp    qword ptr [rip + 0x2fe4]    <load fnptr from [write@got[plt]>
```

Any ideas for a different syntax to indicate this?

Example `telescope` near the GOT region (thes are the names of the symbols that could be used here)
   
```
00:0000│     0x404000 (_GLOBAL_OFFSET_TABLE_) —▸ 0x403e20 (_DYNAMIC) ◂— 1
01:0008│     0x404008 (_GLOBAL_OFFSET_TABLE_+8) —▸ 0x7ffff7ffe3c0 ◂— 0
02:0010│     0x404010 (_GLOBAL_OFFSET_TABLE_+16) —▸ 0x7ffff7fd3ee0 (_dl_runtime_resolve_xsavec) ◂— endbr64
03:0018│     0x404018 (putchar@got[plt]) —▸ 0x401036 (putchar@plt+6) ◂— push 0 /* 'h' */
04:0020│     0x404020 (write@got[plt]) —▸ 0x401046 (write@plt+6) ◂— push 1
05:0028│     0x404028 (fclose@got[plt]) —▸ 0x401056 (fclose@plt+6) ◂— push 2
06:0030│     0x404030 (read@got[plt]) —▸ 0x401066 (read@plt+6) ◂— push 3
07:0038│     0x404038 (srand@got[plt]) —▸ 0x7ffff7c43e50 (srandom) ◂— endbr64
08:0040│     0x404040 (time@got[plt]) —▸ 0x7ffff7fbccd0 (time) ◂— lea rax, [rip - 0x5cd7]
09:0048│     0x404048 (fopen@got[plt]) —▸ 0x401096 (fopen@plt+6) ◂— push 6
0a:0050│     0x404050 (perror@got[plt]) —▸ 0x4010a6 (perror@plt+6) ◂— push 7
0b:0058│     0x404058 (atoi@got[plt]) —▸ 0x4010b6 (atoi@plt+6) ◂— push 8
0c:0060│     0x404060 (exit@got[plt]) —▸ 0x4010c6 (exit@plt+6) ◂— push 9 /* 'h\t' */
0d:0068│     0x404068 (getc@got[plt]) —▸ 0x4010d6 (getc@plt+6) ◂— push 0xa /* 'h\n' */
0e:0070│     0x404070 (rand@got[plt]) —▸ 0x4010e6 (rand@plt+6) ◂— push 0xb /* 'h\x0b' */
0f:0078│     0x404078 (data_start) ◂— 0
```

OBarronCS:
An implementation for the idea in #4119

Now, when running `nearpc` (and `context disasm` for that matter), when we are enhancing a indirect branch instruction that reads it's target from writable memory, but we are not confident in knowing the concrete contents of said memory at the moment the instruction would execute, then we will include text alongside the branch that indicates the symbol associated with the address (or just the address if the symbol cannot be found), as well as the current contents of that memory location (which is the jump target).

These screenshots show this in action with the `plt`.

Before:
image

After:
image

Note that the "now=target" is the branch target, assuming the contents of memory are not changed by the time we actually get to the branch. The `now` notes to the user that this is current contents of memory, but it may change by the time the user steps to it.

This choice to read from writable memory and to display the dereferenced contents in the disassembly display is constrained to only branch instructions that have a constant memory address (typically `RIP` relative addressing or just a constant address)`


pwndbg/color/disasm.py
```diff
@@ -97,6 +98,22 @@ def instructions_and_padding(instructions: list[PwndbgInstruction], linear: bool
 
             asm = f"{ljust_colored(asm, 36)} <{sym}>"
```
disconnect3d: The above `sym` variable is not needed and this could just be:
```py
            asm = f"{ljust_colored(asm, 36)} <{ins.target_string}>"
```

pwndbg/color/disasm.py
disconnect3d, line 132: Those could just be:

```py
if cur_padding_len is None or cur_padding_len - raw_len < MIN_SPACING:
    cur_padding_len = raw_len + MIN_SPACING
```

disconnect3d (approved):

disconnect3d:
There are some conflicts right now, but see copilot comments too; otherwise I'd merge it

pwndbg/color/disasm.py
OBarronCS: Good point, fixed.

pwndbg/color/disasm.py
OBarronCS: Done

OBarronCS:
Awesome. I made the fixes noted above + added 2 tests for this.


image

OBarronCS:
I changed a test to actually take into account branch visualization with these sorts of branches

k4lizen:
test still failing :<

disconnect3d:
@copilot resolve the merge conflicts in this pull request

OBarronCS:
rebased

PR #4123, opened 2026-09-04
Fix certain arm instructions branch resolution logic being incorrect

OBarronCS:
While working on #4122, I was looking at `plt` tables to ARM, and saw there was a bug in how the jump target is determined for certain `ARM` instructions that explicitly write to the program counter.

In ARM, a bunch of general purpose instructions support writing directly to the `pc`: https://support.arm.com/documentation/ddi0406/cb/Application-Level-Architecture/Application-Level-Programmers--Model/ARM-core-registers/Writing-to-the-PC?lang=en

The code didn't take these instructions into account in the correct way when calculating jump targets, and instead used a default jump handler to resolve their addresses, which was incorrect when disassembling them in a static context.

Before: as visualized with the arrows, the branches appear to be going to nearby addresses, but this is incorrect. For all these branches, we don't know statically what the address is.
image

After (no branch target indicated, since we don't know any of them with statically):
image

### DIFF
--- a/pwndbg/aglib/disasm/assistant.py
+++ b/pwndbg/aglib/disasm/assistant.py
@@ -514,7 +514,7 @@ class DisassemblyAssistant:
         if operand.type == CS_OP_MEM:
             # Assume that we are reading ptrsize - subclasses should override this function
             # to provide a more specific value if needed
-            self._read_memory(value, pwndbg.aglib.arch.ptrsize, instruction, emu)
+            return self._read_memory(value, pwndbg.aglib.arch.ptrsize, instruction, emu)
 
         return None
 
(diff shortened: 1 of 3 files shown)


OBarronCS:
`LDR to pc` handled correctly when we are stepping on it, emulation is disabled.

image

pwndbg/aglib/disasm/assistant.py
```diff
@@ -514,7 +514,7 @@ def _resolve_used_value(
         if operand.type == CS_OP_MEM:
             # Assume that we are reading ptrsize - subclasses should override this function
             # to provide a more specific value if needed
-            self._read_memory(value, pwndbg.aglib.arch.ptrsize, instruction, emu)
+            return self._read_memory(value, pwndbg.aglib.arch.ptrsize, instruction, emu)
```
OBarronCS, line 517: wow.... this return statement was missing.... meaning for non-x86 architectures, when emulation was disabled (or not available), certain instructions that load from memory (like the LDR one fix in the screenshow below) were not annotated. We didn't see this in x86 because it overrides this function, and properly returns all paths.

A Python equivalent of `[[nodiscard]]` would be great, and would have caught this bug.

disconnect3d: Seems there is nothing like this yet in the ecosystem: https://discuss.python.org/t/proposal-typing-no-discard-a-decorator-to-indicate-that-the-return-value-should-not-be-discarded/33138 :(

it would be quite easy to add it as some kind of custom rule, or even as Semgrep rule, though.

PR #4124, opened 2026-09-04
Fix heap tracking when realloc returns NULL

### LINKED ISSUE #3998: track-heap: uncaught exception when realloc() returns NULL

N3thunt3r69:
Fixes #3998

## Problem

`ReallocExitBreakpoint.stop()` tried to read chunk metadata at `0xfffffffffffffff8` when a nonzero-size `realloc()` returned `NULL`. The exception stopped the inferior and left heap-tracker state unfinished.

The bug reproduced on:

- release `2026.02.18`
- commit `91d23aa95`
- commit `8e1a2a5d5`
- current `dev` at `51e94b37a`

## Change

Return immediately when `realloc()` fails, close the active memory-management operation, and leave the original allocation tracked. The regression test performs a failing `realloc(..., SIZE_MAX)`, continues to normal process exit, and verifies that the original pointer is still recognized when freed.

## Verification

- `docker compose run --rm ubuntu24.04-mount ./tests.sh -d gdb -g dbg -v test_track_heap_failed_realloc_preserves_original_allocation`
- `docker compose run --rm ubuntu24.04-mount ./tests.sh -d gdb -g dbg -v test_track_heap_symbols`
- `./lint.sh`
- Manual batch-mode reproduction from #3998 now reaches `free()` and exits normally without a Python exception.

+ [ ] I am providing a screenshot of the fixed bug. The failure is non-visual; the before/after command output and regression test cover it instead.

### DIFF
--- a/pwndbg/gdblib/ptmalloc2_tracking.py
+++ b/pwndbg/gdblib/ptmalloc2_tracking.py
@@ -552,14 +552,14 @@ class ReallocExitBreakpoint(gdb.FinishBreakpoint):
         # Figure out what the reallocated pointer is.
         ret_ptr = int(self.return_value)
         if ret_ptr == 0:
-            # No change.
-            malloc = None
+            # The original allocation remains valid when realloc fails.
+            self.tracker.exit_memory_management()
+            return False
         chunk = get_chunk(ret_ptr, self.requested_size)
-        malloc = lambda: self.tracker.malloc(chunk)
 
         if not self.tracker.free(self.freed_ptr):
             # This is a chunk we'd never seen before.
-            malloc()
+            self.tracker.malloc(chunk)
             self.tracker.exit_memory_management()
 
             msg = f"realloc() to {self.requested_size} bytes with previously unknown pointer {self.freed_str}"
@@ -571,7 +571,7 @@ class ReallocExitBreakpoint(gdb.FinishBreakpoint):
                 last_issue = message.error(msg)
             return stop_on_error
 
-        malloc()
+        self.tracker.malloc(chunk)
         self.tracker.exit_memory_management()
 
         origin = caller_symbol() if self.tracker.show_location else None
--- a/tests/library/dbg/tests/test_track_heap_symbols.py
+++ b/tests/library/dbg/tests/test_track_heap_symbols.py
@@ -58,3 +58,24 @@ async def test_track_heap_without_symbols_is_unchanged(ctrl: Controller) -> None
     output = await ctrl.execute_and_capture("continue")
 
     assert "@" not in output
+
+
+@pwndbg_test
+async def test_track_heap_failed_realloc_preserves_original_allocation(
+    ctrl: Controller,
+) -> None:
+    import pwndbg
+    import pwndbg.aglib.proc
+    from pwndbg.dbg_mod import DebuggerType
+
+    if pwndbg.dbg.name() != DebuggerType.GDB:
+        pytest.skip("track-heap hooks a GDB-only event (inferior_call_post)")
+        return
+
+    await launch_to(ctrl, REFERENCE_BINARY, "main")
+
+    await ctrl.execute("track-heap enable")
+    output = await ctrl.execute_and_capture("continue")
+
+    assert not pwndbg.aglib.proc.alive()
+    assert "[*] free(" in output
--- a/tests/library/dbg/tests/test_track_heap_symbols.py
+++ b/tests/library/dbg/tests/test_track_heap_symbols.py
@@ -41,6 +41,9 @@ async def test_track_heap_symbols_annotates_caller(ctrl: Controller) -> None:
     assert _annotation_after(output, "[*] realloc(").startswith("@ do_realloc")
     assert _annotation_after(output, "[*] free(").startswith("@ main")
 
+    # Expect only one realloc annotation as the impossible size (MAX_SIZE) realloc should fail
+    assert output.count("[*] realloc") == 1
+
 
 @pwndbg_test
 async def test_track_heap_without_symbols_is_unchanged(ctrl: Controller) -> None:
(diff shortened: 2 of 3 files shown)


pwndbg/gdblib/ptmalloc2_tracking.py
```diff
@@ -552,14 +552,14 @@ def stop(self):
         # Figure out what the reallocated pointer is.
         ret_ptr = int(self.return_value)
         if ret_ptr == 0:
-            # No change.
-            malloc = None
```
disconnect3d, line 556: Wth? This code was just bad lol, since this was overwritten anyway

disconnect3d (changes requested):

tests/library/dbg/tests/test_track_heap_symbols.py
```diff
@@ -58,3 +58,24 @@ async def test_track_heap_without_symbols_is_unchanged(ctrl: Controller) -> None
...
+        pytest.skip("track-heap hooks a GDB-only event (inferior_call_post)")
+        return
+
+    await launch_to(ctrl, REFERENCE_BINARY, "main")
+
+    await ctrl.execute("track-heap enable")
+    output = await ctrl.execute_and_capture("continue")
+
+    assert not pwndbg.aglib.proc.alive()
+    assert "[*] free(" in output
```
disconnect3d: Ech I see what this is testing but I don't necessarily like it. I feel like we could just add the free assert to another test and that would be it?

disconnect3d: I mean, the other tests for tracking heap are shit since they do not assert anything that is and really should be printed out, so lets maybe add assertions there and not bloat the test suite with another test?

disconnect3d: Actually, I was wrong.

There is a test that catches this error with realloc anyway after the code is added. It should just also make sure that `realloc` tracking message is not in the output.

This is running the test without the fixed realloc tracker code:
```
$ ./tests.sh -d gdb -g dbg -v test_track_heap_symbols_annotates_caller
...
i          = 21
line       = 'pwndbg.dbg_mod.Error: Cannot access memory at address 0xfffffffffffffff8'
lines      = ['[*] malloc(24) -> 0x10042a0, 0x20 bytes real size',
 '    @ do_alloc+14',
 '[*] realloc(0x10042a0, 48) -> 0x10042a0, 0x40 bytes real size',
 '    @ do_realloc+26',
 'Traceback (most recent call last):',
 '  File "/home/dc/pwndbg/pwndbg/dbg_mod/gdb/__init__.py", line 1562, in '
 '__int__',
 '    return int(self.inner)',
 '           ^^^^^^^^^^^^^^^',
 'gdb.MemoryError: Cannot access memory at address 0xfffffffffffffff8',
 '',
 'During handling of the above exception, another exception occurred:',
 '',
 'Traceback (most recent call last):',
 '  File "/home/dc/pwndbg/pwndbg/gdblib/ptmalloc2_tracking.py", line 557, in '
 'stop',
 '    chunk = get_chunk(ret_ptr, self.requested_size)',
 '            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^',
 '  File "/home/dc/pwndbg/pwndbg/gdblib/ptmalloc2_tracking.py", line 446, in '
 'get_chunk',
 '    size = int(pwndbg.aglib.memory.get_typed_pointer_value(ty, address - '
 'ty.sizeof))',
 '           '
 '^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^',
 '  File "/home/dc/pwndbg/pwndbg/dbg_mod/gdb/__init__.py", line 1564, in '
 '__int__',
 '    raise pwndbg.dbg_mod.Error(e)',
 'pwndbg.dbg_mod.Error: Cannot access memory at address 0xfffffffffffffff8']
marker     = '[*] free('

tests/library/dbg/tests/test_track_heap_symbols.py:22: AssertionError
=========================== short test summary info ============================
FAILED tests/library/dbg/tests/test_track_heap_symbols.py::test_track_heap_symbols_annotates_caller - AssertionError: no line containing '[*] free(' found in output:
============================== 1 failed in 0.52s ===============================
--------------------------------------------------------------------------------
If you want to debug tests locally, run ./tests.sh with the --pdb flag
--------------------------------------------------------------------------------


*********************************
********* TESTS SUMMARY *********
*********************************
Time Spent   : 2.45s (cumulative: 1.35s)
Tests Passed : 0
Tests Skipped: 0
Tests Failed : 1

Failing tests:
- tests/library/dbg/tests/test_track_heap_symbols.py::test_track_heap_symbols_annotates_caller
```

disconnect3d (approved):
I think it is good to go after my change, lets see if CI works

disconnect3d:
Thanks!

PR #4125, opened 2026-09-04
Decouple string length from debugger settings

### LINKED ISSUE #4110: Crash when GDB `set print elements 0' or unlimited
If you are in GDB and run `set print elements 0`, the next call to `pwndbg.enhance.enhance()` will likely crash.

We fetch this GDB config value here, and use the result to set an internal "string limit" (how many bytes to read when enhancing strings) to be zero, causing a zero width memory lookup, which is not allowed.
https://github.com/pwndbg/pwndbg/blob/49de3a4b07b6224d6748de9730321aa0c57629b8/pwndbg/dbg_mod/gdb/__init__.py#L2042-L2049

https://github.com/pwndbg/pwndbg/blob/49de3a4b07b6224d6748de9730321aa0c57629b8/pwndbg/aglib/strings.py#L19-L28

See stack trace below for how this propagates to a `enhance()`

According to GDB docs, if the input is `0`, then this implies unlimited:

https://sourceware.org/gdb/current/onlinedocs/gdb.html/Print-Settings.html#:~:text=Set%20a%20limit%20on%20how%20many%20elements,number%20of%20elements%20to%20print%20is%20unlimited.

> When GDB starts, this limit is set to 200. Setting number-of-elements to unlimited or zero means that the number of elements to print is unlimited.

The error is that both "0" and "unlimited" would return 0 from the function that fetches the config value, which results in a zero-byte memory read.

This could be handed by either:
- Not having this internal string limit be determined by this GDB config (we don't really see this elsewhere in the code.
- If it's "unlimited", use some sane large value.

```
│ /home/user/pwndbg/pwndbg/commands/__init__.py:519 in __call__                                    │
│    516 │                                                                                         │
│    517 │   def __call__(self, *args: Any, **kwargs: Any) -> str | None:                          │
│    518 │   │   try:                                                                              │
│ ❱  519 │   │   │   return self.function(*args, **kwargs)                                         │
│    520 │   │   except TypeError:                                                                 │
│    521 │   │   │   print(f"{self.command_name}: {self.description}")                             │
│    522 │   │   │   pwndbg.exception.handle(self.function.__name__)                               │
│ /home/user/pwndbg/pwndbg/commands/context.py:784 in context                                      │
│    781 │   │   result_settings[target].update(settings)                                          │
│    782 │   │   with target as out:                                                               │
│    783 │   │   │   result[target].extend(                                                        │
│ ❱  784 │   │   │   │   func(                                                                     │
│    785 │   │   │   │   │   target=out,                                                           │
│    786 │   │   │   │   │   width=settings.get("width", None),                                    │
│    787 │   │   │   │   │   height=settings.get("height", None),                                  │
│ /home/user/pwndbg/pwndbg/commands/context.py:418 in _serve_context_history                       │
│    415 │   │   │   ):                                                                            │
│    416 │   │   │   │   current_output = context_history[section_name][-1]                        │
│    417 │   │   │   else:                                                                         │
│ ❱  418 │   │   │   │   current_output = function(*a, **kw)                                       │
│    419 │   │   │   if (                                                                          │
│    420 │   │   │   │   len(context_history[section_name]) == 0                                   │
│    421 │   │   │   │   or context_history[section_name][-1] != current_output                    │
│ /home/user/pwndbg/pwndbg/commands/context.py:1150 in context_regs                                │
│   1147 │   width: int | None = None,                                                             │
│   1148 │   height: int | None = None,                                                            │
│   1149 ) -> list[str]:                                                                           │
│ ❱ 1150 │   regs = get_regs()                                                                     │
│   1151 │   if pwndbg.config.show_compact_regs.value != CompactRegsOptions.NO.value:              │
│   1152 │   │   regs = compact_regs(regs, target=target, width=width)                             │
│   1153                                                                                           │
│ /home/user/pwndbg/pwndbg/commands/context.py:1327 in get_regs                                    │
│   1324 │   │   # Resolve "sp" and "pc" to the real architectural register names                  │
│   1325 │   │   reg = pwndbg.aglib.regs.current.resolve_aliases(reg)                              │
│   1326 │   │                                                                                     │
│ ❱ 1327 │   │   desc = rc.register_context_default(reg)                                           │
│   1328 │   │   if desc is not None:                                                              │
│   1329 │   │   │   result.append(desc)                                                           │
│   1330                                                                                           │
│ /home/user/pwndbg/pwndbg/commands/context.py:1268 in register_context_default                    │
│   1265 │   │   val = self.get_register_value(reg)                                                │
│   1266 │   │   if val is None:                                                                   │
│   1267 │   │   │   return None                                                                   │
│ ❱ 1268 │   │   desc = pwndbg.chain.format(val)                                                   │
│   1269 │   │   prefix = self.get_prefix(reg)                                                     │
│   1270 │   │   return f"{prefix} {desc}"                                                         │
│   1271                                                                                           │
│ /home/user/pwndbg/pwndbg/chain.py:202 in format                                                  │
│   199 │   │   │   │   start, end, mem_flags, pointer_to_enhance - start, pwndbg.aglib.arch.ptr   │
│   200 │   │   │   )                                                                              │
│   201 │   │                                                                                      │
│ ❱ 202 │   │   enhanced = pwndbg.enhance.enhance(                                                 │
│   203 │   │   │   pointer_to_enhance,                                                            │
│   204 │   │   │   code=code,                                                                     │
│   205 │   │   │   safe_linking=safe_linking,                                                     │
│ /home/user/pwndbg/pwndbg/enhance.py:107 in enhance                                               │
│   104 │   │   │   # which are baked in during enhancement. This removes those spaces.            │
│   105 │   │   │   instr = " ".join(pwndbg_instr.asm_string.split())                              │
│   106 │                                                                                          │
│ ❱ 107 │   szval = pwndbg.aglib.strings.get(value, maxlen=enhance_string_len) or None             │
│   108 │   szval0 = szval                                                                         │
│   109 │   if szval:                                                                              │
│   110 │   │   szval = E.string(repr(szval))                                                      │
│ /home/user/pwndbg/pwndbg/aglib/strings.py:45 in get                                              │
│   42 │   │   maxread = length                                                                    │
│   43 │                                                                                           │
│   44 │   try:                                                                                    │
│ ❱ 45 │   │   bytesz = pwndbg.aglib.memory.string(address, maxread)                               │
│   46 │   except pwndbg.dbg_mod.Error:  # should not happen, but sanity check?                    │
│   47 │   │   return None                                                                         │
│   48                                                                                             │
│ /home/user/pwndbg/pwndbg/aglib/memory.py:154 in string                                           │
│   151 │   │   An empty bytearray, or a NULL-terminated bytearray.                                │
│   152 │   """                                                                                    │
│   153 │   if peek(addr):                                                                         │
│ ❱ 154 │   │   data = read(addr, max, partial=True)                                               │
│   155 │   │                                                                                      │
│   156 │   │   try:                                                                               │
│   157 │   │   │   return data[: data.index(b"\x00")]                                             │
│ /home/user/pwndbg/pwndbg/aglib/memory.py:38 in read                                              │
│    35 │   │   `bytearray` The memory at the specified address,                                   │
│    36 │   │   or ``None``.                                                                       │
│    37 │   """                                                                                    │
│ ❱  38 │   return pwndbg.dbg.selected_inferior().read_memory(address=addr, size=count, partial=   │
│    39                                                                                            │
│    40                                                                                            │
│    41 def readtype(type: pwndbg.dbg_mod.Type, addr: int) -> int:                                 │
│ /home/user/pwndbg/pwndbg/dbg_mod/gdb/__init__.py:711 in read_memory                              │
│    708 │   │   addr = address                                                                    │
│    709 │   │                                                                                     │
│    710 │   │   try:                                                                              │
│ ❱  711 │   │   │   result = gdb.selected_inferior().read_memory(addr, count)                     │
│    712 │   │   │   return bytearray(result)                                                      │
│    713 │   │   except gdb.error as e:                                                            │
│    714 │   │   │   if not partial:                                                               │
```

N3thunt3r69:
Fixes #4110

## Problem

GDB treats both `set print elements 0` and `set print elements unlimited` as an unlimited display setting. Pwndbg copied that value into its internal string-read length on every stop. The resulting zero-length memory read raised `ValueError: Argument 'count' should be greater than zero` during string enhancement.

## Change

Use Pwndbg's existing 200-byte limit independently of debugger display settings. Remove the obsolete `Debugger.string_limit()` interface, its GDB and LLDB implementations, and the per-stop refresh hooks.

The regression test covers both accepted GDB spellings, forces a stop so the old refresh path runs, and verifies that `telescope` still enhances the executable path.

## Verification

- `docker compose run --rm ubuntu24.04-mount ./tests.sh -d gdb -g gdb -v test_command_telescope_with_unlimited_print_elements`
- `docker compose run --rm ubuntu24.04-mount ./tests.sh -d gdb -g gdb -v test_command_telescope`
- `./lint.sh`
- Manual batch-mode reproduction against `dev` at `51e94b37a`

+ [ ] I am providing a screenshot of the fixed bug. The failure is non-visual; the traceback and regression test cover it instead.

### DIFF
--- a/pwndbg/aglib/strings.py
+++ b/pwndbg/aglib/strings.py
@@ -14,18 +14,7 @@ import pwndbg.aglib.memory
 import pwndbg.dbg_mod
 from pwndbg.lib.memory import Page
 
-length = 15
-
-
-def update_length() -> None:
-    r"""
-    Unfortunately there's not a better way to get at this info.
-
-    >>> gdb.execute('show print elements', from_tty=False, to_string=True)
-    'Limit on string chars or array elements to print is 21.\n'
-    """
-    global length
-    length = pwndbg.dbg.string_limit()
+length = 200
 
 
 def get(address: int, maxlen: int | None = None, maxread: int | None = None) -> str | None:
--- a/pwndbg/dbg_mod/__init__.py
+++ b/pwndbg/dbg_mod/__init__.py
@@ -1396,12 +1396,6 @@ class Debugger:
         """
         raise NotImplementedError()
 
-    def string_limit(self) -> int:
-        """
-        The maximum size of a string.
-        """
-        raise NotImplementedError()
-
     def addrsz(self, address: Any) -> str:
         """
         Format the given address value.
--- a/pwndbg/dbg_mod/gdb/__init__.py
+++ b/pwndbg/dbg_mod/gdb/__init__.py
@@ -2047,15 +2047,6 @@ class GDB(pwndbg.dbg_mod.Debugger):
         return literal
 
     @override
-    def string_limit(self) -> int:
-        message = gdb.execute("show print elements", from_tty=False, to_string=True)
-        message = message.split("\n")[0].split()[-1]
-        message = message.strip(".")
-        if message == "unlimited":
-            return 0
-        return int(message)
-
-    @override
     def addrsz(self, address: Any) -> str:
         address = int(address) & pwndbg.aglib.arch.ptrmask
         return f"%#{2 * pwndbg.aglib.arch.ptrsize}x" % address
--- a/pwndbg/dbg_mod/gdb/hooks.py
+++ b/pwndbg/dbg_mod/gdb/hooks.py
@@ -8,7 +8,6 @@ import pwndbg
 import pwndbg.aglib.arch_mod
 import pwndbg.aglib.file
 import pwndbg.aglib.memory
-import pwndbg.aglib.strings
 import pwndbg.aglib.typeinfo
 import pwndbg.dbg_mod
 import pwndbg.gdblib.events
@@ -90,7 +89,6 @@ def on_start() -> None:
 
 @pwndbg.dbg.event_handler(EventType.STOP)
 def on_stop() -> None:
-    pwndbg.aglib.strings.update_length()
     pwndbg.dbg_mod.number_of_stops_since_birth += 1
 
 
--- a/pwndbg/dbg_mod/lldb/__init__.py
+++ b/pwndbg/dbg_mod/lldb/__init__.py
@@ -2536,10 +2536,6 @@ class LLDB(pwndbg.dbg_mod.Debugger):
         return literal
 
     @override
-    def string_limit(self) -> int:
-        return 200
-
-    @override
     def get_cmd_window_size(self) -> tuple[int, int]:
         return None, None
 
--- a/pwndbg/dbg_mod/lldb/hooks.py
+++ b/pwndbg/dbg_mod/lldb/hooks.py
@@ -9,7 +9,6 @@ import pwndbg.aglib.arch_mod
 import pwndbg.aglib.file
 import pwndbg.aglib.kernel
 import pwndbg.aglib.memory
-import pwndbg.aglib.strings
 import pwndbg.aglib.typeinfo
 import pwndbg.dbg_mod
 import pwndbg.lib.cache
@@ -47,7 +46,6 @@ def on_start() -> None:
 
 @pwndbg.dbg.event_handler(EventType.STOP)
 def on_stop() -> None:
-    pwndbg.aglib.strings.update_length()
     pwndbg.dbg_mod.number_of_stops_since_birth += 1
 
 
--- a/tests/library/gdb/tests/test_command_telescope.py
+++ b/tests/library/gdb/tests/test_command_telescope.py
@@ -1,8 +1,10 @@
 from __future__ import annotations
 
 import re
+from collections.abc import Callable
 
 import gdb
+import pytest
 
 import pwndbg.aglib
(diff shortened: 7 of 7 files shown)


OBarronCS:
Nice, thanks! Currently, it will be hardcoded to 200 characters max. I think this is good. I doubt anyone has relied on this configuration to change this behavior GDB.

OBarronCS (approved):

OBarronCS:
Please rebase from dev branch, as there are conflicts

k4lizen (commented):
test should be in library/dbg and gated with an if-statement if its GDB-only.

library/gdb is deprecated

OBarronCS:
I think it's okay to drop the test. It tests "does setting this GDB config to a certain value crash us", but now we don't use that setting at all, and I can't see us using it in the future.

PR #4131, opened 2026-09-06
Resolve function boundaries in the absence of debugging symbols

### LINKED ISSUE #3908: function boundary detection does not work on gdb without debug info
### Description

This (`pwndbg/dbg_mod/gdb/__init__.py`):
```python
    @override
    def get_function_boundaries(self, address: int) -> tuple[int, int] | None:
        block = gdb.block_for_pc(address)

        if block is not None:
            # Find the top-level function that this block resides in
            while block.superblock is not None and block.superblock.function is not None:
                block = block.superblock

            return block.start, block.end

        return None
```
i.e. `gdb.block_for_pc()` does not work when there is no debug info. This is used by `nearpc --function` which does not have parity with `disass` because of this.

As far as we can tell, this is a limitation of GDB. See tracking issue: https://sourceware.org/bugzilla/show_bug.cgi?id=34141 .

This could also be used for backwards disassembly of variable length instruction sets.

OBarronCS:
Resolves #3908

This implements a way to get the function boundaries of the function which an address is within in the absence of debugging symbols. This is currently used for `nearpc -f`, making our nearpc command now work the same way as the debugger native `disassemble` commands.

When debug symbols are absent, the underlying debugger needs to do a little more work to determine function boundaries. At a high level, they use the symbol information (if they have symbol_start address, and symbol_end address) to determine the boundaries, and as a fallback, if the symbol size is unknown, they find the next symbol contiguously in memory, and treat it as the end of the current symbol. This is how the builtin `disass` does it.

Turns out LLDB already exposes a way for us to get function boundaries such that we get the same info as the built-in `disass`.

For GDB, no such API is exposed. To implement fetching function boundary, in this PR, we call the GDB command `disassemble`, which internally gets the the function boundaries. It's a bit hacky, but we parse the output of `disassemble` to get the boundaries. This guarantees that whenever someone could use `disass`, they could instead use `nearpc -f`, and take advantage of all of our additional disassembly features.

Other methods of determining function boundaries were considered (all of them were hacky ways involving querying symbols), but for this use case, calling `disass` guarantees we get the function boundary correct to the same extent that GDB already does.

lldb

Note that some functions, such as `_dl_fixup` that I came across while testing this, have more than one contiguous block of addresses that define the function. While querying the function boundary, we get all the ranges, and then return the range within with the input address sits.

lldb_libc_function

### DIFF
--- a/pwndbg/aglib/symbol.py
+++ b/pwndbg/aglib/symbol.py
@@ -9,6 +9,7 @@ import pwndbg.aglib.memory
 import pwndbg.dbg_mod
 import pwndbg.dintegration
 import pwndbg.lib.cache
+from pwndbg.dbg_mod import EventType
 from pwndbg.dbg_mod import SymbolLookupType
 
 
@@ -117,6 +118,14 @@ def resolve_addr(addr: int) -> str | None:
     return pwndbg.dintegration.manager.symbol_at_address(addr)
 
 
+set_existing_ranges: set[tuple[int, int]] = set()
+
+
+@pwndbg.dbg.event_handler(EventType.NEW_MODULE)
+def _clear_set_existing_ranges() -> None:
+    set_existing_ranges.clear()
+
+
 @pwndbg.lib.cache.cache_until("objfile")
 def resolve_function_boundaries(addr: int) -> tuple[int, int] | None:
     """
@@ -128,4 +137,13 @@ def resolve_function_boundaries(addr: int) -> tuple[int, int] | None:
     """
     assert addr >= 0, "address must be positive"
 
-    return pwndbg.dbg.selected_inferior().get_function_boundaries(addr)
+    for cached_range in set_existing_ranges:
+        if cached_range[0] <= addr < cached_range[1]:
+            return cached_range
+
+    fn_range = pwndbg.dbg.selected_inferior().get_function_boundaries(addr)
+
+    if fn_range is not None:
+        set_existing_ranges.add(fn_range)
+
+    return fn_range
--- a/pwndbg/aglib/symbol.py
+++ b/pwndbg/aglib/symbol.py
@@ -141,10 +141,15 @@ def resolve_function_boundaries(addr: int) -> tuple[int, int] | None:
     """
     assert addr >= 0, "address must be positive"
 
+    # Binary search for the range which has the highest start which is <= addr,
+    # and out of those, we pick the one with the highest end.
+    # Does not find a match for some overlapping ranges (e.g. addr=150, set_existing_ranges=[(100,200), (148,149)])
+    # but this should almost never happen, and getting a cache miss is still fine correctness-wise.
     i = bisect_right(set_existing_ranges, addr, key=lambda _range: _range[0]) - 1
-    if i >= 0 and addr <= set_existing_ranges[i][1]:
+    if i >= 0 and addr < set_existing_ranges[i][1]:
         return set_existing_ranges[i]
 
+    # Invokes GDBs `disass` which might be slow.
     fn_range = pwndbg.dbg.selected_inferior().get_function_boundaries(addr)
 
     if fn_range is not None:
--- a/pwndbg/aglib/symbol.py
+++ b/pwndbg/aglib/symbol.py
@@ -5,10 +5,14 @@ vice-versa.
 
 from __future__ import annotations
 
+from bisect import bisect_right
+from bisect import insort
+
 import pwndbg.aglib.memory
 import pwndbg.dbg_mod
 import pwndbg.dintegration
 import pwndbg.lib.cache
+from pwndbg.dbg_mod import EventHandlerPriority
 from pwndbg.dbg_mod import EventType
 from pwndbg.dbg_mod import SymbolLookupType
 
@@ -118,15 +122,15 @@ def resolve_addr(addr: int) -> str | None:
     return pwndbg.dintegration.manager.symbol_at_address(addr)
 
 
-set_existing_ranges: set[tuple[int, int]] = set()
+set_existing_ranges: list[tuple[int, int]] = []
 
 
-@pwndbg.dbg.event_handler(EventType.NEW_MODULE)
+@pwndbg.dbg.event_handler(EventType.START, priority=EventHandlerPriority.CACHE_CLEAR)
+@pwndbg.dbg.event_handler(EventType.EXIT, priority=EventHandlerPriority.CACHE_CLEAR)
 def _clear_set_existing_ranges() -> None:
     set_existing_ranges.clear()
 
 
-@pwndbg.lib.cache.cache_until("objfile")
 def resolve_function_boundaries(addr: int) -> tuple[int, int] | None:
     """
     Return the function start and end address for a function that
@@ -137,13 +141,13 @@ def resolve_function_boundaries(addr: int) -> tuple[int, int] | None:
     """
     assert addr >= 0, "address must be positive"
 
-    for cached_range in set_existing_ranges:
-        if cached_range[0] <= addr < cached_range[1]:
-            return cached_range
+    i = bisect_right(set_existing_ranges, addr, key=lambda _range: _range[0]) - 1
+    if i >= 0 and addr <= set_existing_ranges[i][1]:
+        return set_existing_ranges[i]
 
     fn_range = pwndbg.dbg.selected_inferior().get_function_boundaries(addr)
 
     if fn_range is not None:
-        set_existing_ranges.add(fn_range)
+        insort(set_existing_ranges, fn_range)
 
     return fn_range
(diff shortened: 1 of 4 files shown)


pwndbg/dbg_mod/gdb/__init__.py
```diff
@@ -963,14 +1031,18 @@ def lookup_symbol(
...
     def get_function_boundaries(self, address: int) -> tuple[int, int] | None:
-        block = gdb.block_for_pc(address)
 
-        if block is not None:
-            # Find the top-level function that this block resides in
-            while block.superblock is not None and block.superblock.function is not None:
-                block = block.superblock
+        # While GDB internally has multiple ways of determine function boundaries in the absence
+        # of debugging symbols (using symbol sizes if available, then falling back to using the order symbols in memory to determine boundaries)
+        # These methods are internally used to determine the these methods are not exposed to the Python API
```
k4lizen, line 1057: maybe link to https://github.com/pwndbg/pwndbg/issues/3908

pwndbg/dbg_mod/gdb/__init__.py
```diff
@@ -963,14 +1031,18 @@ def lookup_symbol(
 
     @override
     def get_function_boundaries(self, address: int) -> tuple[int, int] | None:
```
k4lizen, line 1053: You should update the description of `get_function_boundaries` in `pwndbg/dbg_mod/__init__.py`

pwndbg/dbg_mod/lldb/__init__.py
```diff
@@ -1641,10 +1641,33 @@ def get_function_boundaries(self, address: int) -> tuple[int, int] | None:
...
+            ]
+
+            for start_block, end_block in ranges:
+                if start_block <= address < end_block:
+                    return start_block, end_block
+
+        # Fallback to finding the symbol that contains this address, and use it to determine the start/end of this function
+        sym = ctx.GetSymbol()
+        if sym.IsValid():
+            start: int = sym.GetStartAddress().GetLoadAddress(self.target)
```
k4lizen: Image

maybe we should be checking validity on these

pwndbg/dbg_mod/gdb/__init__.py
```diff
@@ -605,6 +605,74 @@ def remove(self) -> None:
...
+    Returns list of tuples representing [start,end) of the addresses that make up this function.
+
+    Note that the `end` address is exclusive of the start of the last instruction. It is guaranteed to be at least `address of last instruction` + 1.
+    It gets fully accurate end address in the that `address` is in a function with multiple address ranges.
+
+    This allows us to disassemble the entire function correctly, as we can start the disassembly of the final instruction (it will still be within the range returned by this function)
+    """
+
+    try:
+        disass_output: str = gdb.execute(f"disassemble {address}", to_string=True)
```
k4lizen: i would consider adding a cache for this since it shouldn't get invalidated realistically, though not sure what datastructure/algo is good for range lookups

k4lizen (changes requested):

pwndbg/dbg_mod/gdb/__init__.py
k4lizen: ah nvm, i misread the code

k4lizen (approved):

pwndbg/dbg_mod/gdb/__init__.py
OBarronCS: Done

pwndbg/dbg_mod/lldb/__init__.py
OBarronCS: Good call. I also added a check for `start != lldb.LLDB_INVALID_ADDRESS`, looks like other places in this file also have this. Seems to be a separate check that needs to be done.

pwndbg/dbg_mod/gdb/__init__.py
OBarronCS: I think the only way really is to add the cache per address, which I added (which clears on objfile loads, as new symbols/things may be loaded so this function needs to recheck).

We could keep a high level cache that stores the address ranges that we binary search over, but this works for now, and this is something is only invoked when user does `nearpc -f` manually.

k4lizen:
> We could keep a high level cache that stores the address ranges that we binary search over, but this works for now, and this is something is only invoked when user does nearpc -f manually.

hmm didn't we also want to use it when we enter functions to not disassemble backwards futher than a function boundary?

pwndbg/dbg_mod/gdb/__init__.py
OBarronCS: nvm, removed the cache due to cache issues with the mutable return type

OBarronCS:
> > We could keep a high level cache that stores the address ranges that we binary search over, but this works for now, and this is something is only invoked when user does nearpc -f manually.
> 
> hmm didn't we also want to use it when we enter functions to not disassemble backwards futher than a function boundary?

Oh yes, good point.... We do want a quick way to lookup the `[start,end)` of a function, and rerunning `disass` everytime in this context is definitely bad.

In that sense, I think the best solution is add a cache like
```
existing_ranges_sorted: list[tuple[int, int]] = []
```

And then before shelling out to `gdb`, binary search through this list to see if in the input address is in this list.

k4lizen:
maybe a set works more naturally instead of a list?

OBarronCS:
Awesome, added a set that caches existing function boundaries. Also fixed the weird thing with the `end` address on gdb. Now it reports the true [start,end) (end exclusive) function boundary (if not, it would have messed with the caching when querying addresses of the last instruction in a function).

pwndbg/aglib/symbol.py
```diff
@@ -117,6 +118,14 @@ def resolve_addr(addr: int) -> str | None:
     return pwndbg.dintegration.manager.symbol_at_address(addr)
 
 
+set_existing_ranges: set[tuple[int, int]] = set()
+
+
+@pwndbg.dbg.event_handler(EventType.NEW_MODULE)
```
k4lizen: make it EventHandlerPriority.CACHE_CLEAR

pwndbg/aglib/symbol.py
```diff
@@ -117,6 +118,14 @@ def resolve_addr(addr: int) -> str | None:
...
+def _clear_set_existing_ranges() -> None:
+    set_existing_ranges.clear()
+
+
 @pwndbg.lib.cache.cache_until("objfile")
```
k4lizen: i would delete this, probably having two caching layers only hurts

pwndbg/aglib/symbol.py
```diff
@@ -117,6 +118,14 @@ def resolve_addr(addr: int) -> str | None:
     return pwndbg.dintegration.manager.symbol_at_address(addr)
 
 
+set_existing_ranges: set[tuple[int, int]] = set()
+
+
+@pwndbg.dbg.event_handler(EventType.NEW_MODULE)
+def _clear_set_existing_ranges() -> None:
+    set_existing_ranges.clear()
```
k4lizen, line 131: why are we clearing when we get a new module though? we don't save "misses" anyway

pwndbg/aglib/symbol.py
```diff
@@ -128,4 +137,13 @@ def resolve_function_boundaries(addr: int) -> tuple[int, int] | None:
     """
     assert addr >= 0, "address must be positive"
 
-    return pwndbg.dbg.selected_inferior().get_function_boundaries(addr)
+    for cached_range in set_existing_ranges:
+        if cached_range[0] <= addr < cached_range[1]:
+            return cached_range
```
k4lizen: ah i thought we would be binary searching them which is why i recommended a set ;;

but yeah it's tricky (considering we might have overlaps) and prolly not worth it perf wise

i guess if we're just iterating a list is fine/better

k4lizen (changes requested):
i have a few bikesheds left 😓

pwndbg/aglib/symbol.py
OBarronCS: Oh nice, thanks.

pwndbg/aglib/symbol.py
OBarronCS: That's a good point. I guess I was thinking about starting another process altogether within the same gdb session, but I suppose other parts of `pwndbg` don't necessarily take this into account when caching anyways?

k4lizen: you can clear it in EventType.START / EventType.EXIT then i suppose (maybe double check), i saw that glibc code does that, but i agree that most places in pwndbg dont

pwndbg/aglib/symbol.py
OBarronCS: Wait how can we binary search over a `set`? With the list we can sort it everytime we insert then do the search, but I don't know how to do this with a set

k4lizen: idk i was thinking something like bisect.bisect_left , but not sure about the details

OBarronCS:
Fixed the caching and made it binary search for the cache.

k4lizen (approved):
LGTM

Issue #3905, opened 2026-05-09
Negative and unexpected tcachebin counts on glibc 2.43


ShlomoJune:
## Environment

- Pwndbg: 2026.02.18 build 6c501f2b5
- GDB: 15.1
- glibc: 2.43
- Target: how2heap malloc_playground

## Case 1: before manual allocation

After starting malloc_playground, before I manually allocate any chunk, the heap layout is:

```
pwndbg> heap
Support for tcache large bins (a GLIBC 2.42 addition) has not been fully implemented. PR contributions are highly appreciated!
Allocated chunk | PREV_INUSE
Addr: 0x6159dfb06000
Size: 0xb0 (with flag bits: 0xb1)

Allocated chunk | PREV_INUSE
Addr: 0x6159dfb060b0
Size: 0x1010 (with flag bits: 0x1011)

Top chunk | PREV_INUSE
Addr: 0x6159dfb070c0
Size: 0x1ff40 (with flag bits: 0x1ff41)

pwndbg> bins
tcachebins
0x20 [ 16]: 0
0x30 [ 16]: 0
0x40 [ 16]: 0x1011
0x50 [ 16]: 0
0x60 [ 16]: 0
0x70 [ 16]: 0
0x80 [ 16]: 0
0x90 [ 16]: 0
0xa0 [ 16]: 0
0xb0 [ 16]: 0
0xc0 [ 16]: 0
0xd0 [ 16]: 0
0xe0 [ 16]: 0
0xf0 [ 16]: 0
```

<img width="1106" height="598" alt="Image" src="https://github.com/user-attachments/assets/b5b01a87-0f83-46b0-b6a3-7526b0fb8e33" />

## Case 2: after malloc(0x10)

After manually running `malloc(0x10)` in malloc_playground, the heap layout is:

```
pwndbg> heap
Support for tcache large bins (a GLIBC 2.42 addition) has not been fully implemented. PR contributions are highly appreciated!
Allocated chunk | PREV_INUSE
Addr: 0x6450e275d000
Size: 0xb0 (with flag bits: 0xb1)

Allocated chunk | PREV_INUSE
Addr: 0x6450e275d0b0
Size: 0x1010 (with flag bits: 0x1011)

Allocated chunk | PREV_INUSE
Addr: 0x6450e275e0c0
Size: 0x20 (with flag bits: 0x21)

Top chunk | PREV_INUSE
Addr: 0x6450e275e0e0
[...]

PR #4133, opened 2026-09-06
Fix incorrectly finding tcache on glibc >= 2.42 (#3905)

kozabrada123:
+ [X] I am providing a screenshot of the (new feature) / (fixed bug).

Before (current dev pwndbg):
bug

If we allocate a large enough chunk as the first chunk on the heap (on glibc >= `2.42`), pwndbg will think it is the `tcache_perthread_struct`, even though it isn't (note that `tcache` is `NULL` here), causing (#3905).

This happens because glibc 2.42 changes when the tcache is allocated: it used to be after _any_ allocation, but now it is only after the first tcache sized one ([see](https://elixir.bootlin.com/glibc/glibc-2.42/source/malloc/malloc.c#L3478)).

This patch:

(See convo below about the modified warning message)


k4lizen:
But then when the tcache does actually get allocated afterwards, does it get correctly found?

Even if not, I do think it's better to say "can't find the tcache" than actively lie to users (which your patch works towards)

kozabrada123:
> But then when the tcache does actually get allocated afterwards, does it get correctly found?

Yes, on both versions, since the tcache variable is set to the correct pointer.

> Even if not, I do think it's better to say "can't find the tcache" than actively lie to users (which your patch works towards)

Sure; would you agree to keep the old message on older glibc versions and update this one to be "the thread hasn't performed any **small enough** allocations?" (or maybe "tcache-sized allocations"?)

Personally, I dislike "can't find the tcache" when we specifically know the tcache is a null pointer. That implies that there is a tcache, we just don't know where, vs the reality of "the tcache is not initialized yet".

k4lizen:
> Yes, on both versions, since the tcache variable is set to the correct pointer.

ahh makes sense

> Sure; would you agree to keep the old message on older glibc versions and update this one to be "the thread hasn't performed any small enough allocations?" (or maybe "tcache-sized allocations"?)

yup that makes a lot of sense, i'd use "tcache-sized"

k4lizen:
there's a heap-related test failing thou

k4lizen:
maybe related to https://github.com/pwndbg/pwndbg/pull/4128 ?

kozabrada123:
I've updated the message, it seems that this does affect the test though, I'll take a look:
> FAILED tests/library/dbg/tests/test_glibc_heap_bins.py::test_tcache_bins_respects_heap_dereference_limit - assert '0x10324f0' in "No tcache bins found for thread 1 (the thread hasn't...

[link](https://github.com/pwndbg/pwndbg/actions/runs/34059381030/job/101557372201#step:7:1182)
(from the previous commit; this happens because fedora43 ships w/ glibc 2.42)

kozabrada123:
hold on, this only happens on aarch64??

kozabrada123:
> [DBG-GDB] pwndbg> p/x tcache
[DBG-GDB] $1 = \<optimized out\>

right.
[logs](https://github.com/kozabrada123/pwndbg/actions/runs/34155967017/job/101847765641?pr=1#step:7:1035)
also reported at https://github.com/pwndbg/pwndbg/issues/3905#issuecomment-4599413365

OBarronCS:
There was recently a large change to libc testing merged, where we can now run tests per-libc version. You'll need to `rebase` from the latest dev branch to get the changes.

If you search everything undo `tests/`, and look for the "2.42" string, it looks like there has previously some tests disabled (`pytest.mark.xfail`) for version 2.42. Are any of these relevant here, and could they be enabled to verify the changes here work?

Here is an example of a test, and how to run with the specific libc locally:
```
# Download the glibc files
./scripts/download-test-libs.sh glibc

# Run a specific test with a specific version
# This specific test skips 2.42: see the decorator. You can enable it changing the decorator arguments
./tests.sh -d gdb -g dbg "test_heap_heuristic_nodebug_glibc_version\[2.42\]"
```

OBarronCS:
And potentially, if a test for what this PR is fixing doesn't exist, could you add one?

kozabrada123:
Thanks! I can also add a test, sure; I'll check, but I don't think any existing tests are relevant here

this test actually found a major issue that very much needs to be fixed, though I am unsure if it's in scope for this pr / if I will manage to fix it

on aarch64, we currently don't have a good way to find the tcache, since the symbol is optimized out, and we (as demonstrated above) can no longer assume it's the first chunk on the heap. I'll take a look at how it's actually read / written in the assembly

kozabrada123:
also, glibc 2.43 further changes how the tcache symbol works..

> static const union
{
  struct [tcache_perthread_struct](https://elixir.bootlin.com/glibc/glibc-2.43/C/ident/tcache_perthread_struct) [inactive](https://elixir.bootlin.com/glibc/glibc-2.43/C/ident/inactive);
  struct
  {
    char [pad](https://elixir.bootlin.com/glibc/glibc-2.43/C/ident/pad);
    struct [tcache_perthread_struct](https://elixir.bootlin.com/glibc/glibc-2.43/C/ident/tcache_perthread_struct) [disabled](https://elixir.bootlin.com/glibc/glibc-2.43/C/ident/disabled);
  };
} [__tcache_dummy](https://elixir.bootlin.com/glibc/glibc-2.43/C/ident/__tcache_dummy);

> /* TCACHE is never NULL; it's either "live" or points to one of the
   above dummy entries.  The dummy entries are all zero so act like an
   empty/unusable tcache.  */
static __thread [tcache_perthread_struct](https://elixir.bootlin.com/glibc/glibc-2.43/C/ident/tcache_perthread_struct) *[tcache](https://elixir.bootlin.com/glibc/glibc-2.43/C/ident/tcache) =
  ([tcache_perthread_struct](https://elixir.bootlin.com/glibc/glibc-2.43/C/ident/tcache_perthread_struct) *) &[__tcache_dummy](https://elixir.bootlin.com/glibc/glibc-2.43/C/ident/__tcache_dummy).[inactive](https://elixir.bootlin.com/glibc/glibc-2.43/C/ident/inactive);

This isn't that critical, since as far as I understand (I'll test though) this should just show an empty tcache before it's allocated

kozabrada123:
I admit I have a combined 5 minutes of aarch64 experience, however:

```asm
0x0000fffff7e9043c <+16>:	bl	0xfffff7e90184 <__libc_malloc2 ; malloc tcache_perthread_struct, result is in x0
0x0000fffff7e90440 <+20>:	adrp	x2, 0xfffff7faf000 <_nl_C_locobj+144> ; find the page offset
0x0000fffff7e90444 <+24>:	ldr	x2, [x2, #3432] ; load some other offset from there
0x0000fffff7e90448 <+28>:	mrs	x1, tpidr_el0 ; load thread ptr into x1
0x0000fffff7e9044c <+32>:	add	x1, x1, x2
0x0000fffff7e90450 <+36>:	str	x0, [x1, #8] ; store malloced ptr from x0 into x1 + 8 (where tcache actually is)
```

```c
tcache = (tcache_perthread_struct *) __libc_malloc2(sizeof (tcache_perthread_struct));
```

[assembly](https://github.com/kozabrada123/pwndbg/actions/runs/34254885725/job/102158097413?pr=1#step:7:1383), [malloc.c](https://elixir.bootlin.com/glibc/glibc-2.42/source/malloc/malloc.c#L3380)

Seems like we can't read from the symbol because it's relative to the thread pointer. It should (?) be possible to have some custom logic to read it, though I think the main issue here would be finding these offsets

k4lizen:
> Seems like we can't read from the symbol because it's relative to the thread pointer. It should (?) be possible to have some custom logic to read it, though I think the main issue here would be finding these offsets

But it's also relative to the Thread Local Storage for x86_64 (i think fs register) and we can read the symbol.

But in any case, pwndbg supports heap debugging without symbol information, e.g. if I do on x86_64 arch linux
```
~❯ gdb /bin/sh
pwndbg> set debuginfod enabled off
pwndbg> run
Ctrl-C
pwndbg> p tcache
❌ No symbol "tcache" in current context.
pwndbg> bins
tcachebins
0x20 [  1]:  0x5555556bd2e0 ◂— 0
.....
```
And we do have these tcache offset TLS heuristics implemented, see `HeuristicHeap.thread_cache()`. Note that your changes are only in `DebugSymsHeap.thread_cache()`. I'm guessing the aarch64 test does not have symbols? So there's something wrong with these heursitics?

Also regardless of the aarch64 stuff you maybe should modify `HeuristicHeap.thread_cache()` for the same reasons you modified `DebugSymsHeap.thread_cache()`

kozabrada123:
Thanks, I'll take a look :D

>  I'm guessing the aarch64 test does not have symbols? So there's something wrong with these heuristics?

I think it does have symbols, the symbol just shows as optimized out, so maybe these heuristics aren't triggered because of that (?)

k4lizen:
> I think it does have symbols, the symbol just shows as optimized out, so maybe these heuristics aren't triggered because of that (?)

this would make sense

k4lizen:
I'm gonna merge this because we want to cut a new release today/tomorrow, and it does help with the DebugSymsHeap, i'll open two issues regarding the behaviour right now w.r.t. aarch64 and the heuristic heap, would be awesome if you could take a stab at those :p

PR #4134, opened 2026-09-06
print error on register read if it is non-standard

k4lizen:

### Before

image

### After

image

### DIFF
--- a/pwndbg/dbg_mod/gdb/__init__.py
+++ b/pwndbg/dbg_mod/gdb/__init__.py
@@ -158,10 +158,17 @@ class GDBRegisters(pwndbg.dbg_mod.Registers):
     def by_name(self, name: str) -> pwndbg.dbg_mod.Value | None:
         try:
             return GDBValue(self.frame.inner.read_register(name))
-        except (gdb.error, ValueError):
-            # GDB throws an exception if the name is unknown, we just return
-            # None when that is the case.
-            pass
+        except ValueError as e:
+            if "Bad register" in str(e):
+                # GDB throws a ValueError exception if the name is unknown, we just return
+                # None when that is the case.
+                return None
+            # Otherwise some weird shenanigents might be going on, so we print the message
+            # as well.
+            err_str: str = str(e)
+        except gdb.error as e:
+            err_str = str(e)
+        print(message.error(f"gdb register read error: {err_str}"))
         return None
 
 


k4lizen:
Reproduces at
```
pwndbg> 
   0 0x5555555551ad thrower()+4
   1 0x5555555551de middle()+9
 ► 2 0x5555555551ef main+14
   3 0x7ffff7827781 __libc_start_call_main+129
   4 0x7ffff78278b9 __libc_start_main+137
   5 0x5555555550d5 _start+37
```

for
```cpp
#include <iostream>
#include <cstdint>

void thrower() {
	throw 1lu;
}

void middle() {
	__asm__(".cfi_escape 0x16, 0x03, 0x03, 0x53, 0x32, 0x24");
	thrower();
}

int main() {
	try {
		middle();
	} catch(uint64_t i) {
		uint64_t rbx;
		__asm__("mov %%rbx, %0" : "=r"(rbx));
		std::cout << "RBX: " << rbx << std::endl;
	}
}
```

k4lizen:
relevant commits: 84f16455c1 , f48f53f7bc . I think @mbrla0 just accidentally wrote the wrong thing? (the alternative explanation being that older gdb versions behaved differently)

k4lizen:
this gdb commit from 2022 changes the behaviour: `git show bdc8cfc1e43e` but i can't build the version on my arch machine so idk what it actually did before

disconnect3d:
Reproduce with:
```
g++ thrower.cpp
gdb ./a.out

b thrower
run
up
up
regs
```
After the 2x `up` u should be in sth like:
```
pwndbg>
   0 0x555555555231 thrower()+8
   1 0x555555555266 middle()+13
 ► 2 0x55555555527b main+18
```

What is further sad is that its just GDB not giving out the register value at all, even with:
```
pwndbg> info registers
rax            0x555555555269      0x555555555269
DWARF-2 expression error: `DW_OP_reg' operations must be used either alone or in conjunction with DW_OP_piece or DW_OP_bit_piece.
pwndbg> info registers rbx
DWARF-2 expression error: `DW_OP_reg' operations must be used either alone or in conjunction with DW_OP_piece or DW_OP_bit_piece.
pwndbg> info registers rcx
rcx            0x555555557d38      0x555555557d38
```

disconnect3d:
Fwiw we can actually get the register value through the low level GDB commands:
```
pwndbg> maintenance print raw-registers
 Name         Nr  Rel Offset    Size  Type            Raw value
 rax           0    0      0       8 int64_t         0x0000555555555269
 rbx           1    1      8       8 int64_t         0x00007fffffffe218
 rcx           2    2     16       8 int64_t         0x0000555555557d38
...
```
or:
```
pwndbg> maintenance print cooked-registers
 Name         Nr  Rel Offset    Size  Type            Cooked value
 rax           0    0      0       8 int64_t         0x0000555555555269
 rbx           1    1      8       8 int64_t         0x00007fffffffe218
 rcx           2    2     16       8 int64_t         0x0000555555557d38
...
```

k4lizen:
In this case GDB is right to error out because the dwarf is in fact malformed.

OBarronCS:
Interesting... so GDB could raise either `gdb.error` or `ValueError`. I think this is definitely good to print more information.

k4lizen:
yup, double checked, it can be both

image

PR #4135, opened 2026-09-07
lldb: Don't crash the REPL on a malformed command line

### LINKED ISSUE #4130: LLDB crashes when `\` character in invoked command
Looks like there is a bug in CLI parsing, which I found when accidentally had a `\` as part of an argument:

Reproduce with
```sh
pwndbg-lldb ls

# backslash is near enter key, accidently pressed it
nearpc -f\
```

Backtrace
```
pwndbg-lldb> nearpc -f\
  File "/home/newt/.local/bin/pwndbg-lldb", line 10, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/newt/pwndbg/pwndbginit/pwndbg_lldb.py", line 206, in main
    launch(drive(startup), debug=debug)
  File "/home/newt/pwndbg/pwndbginit/pwndbg_lldb.py", line 101, in launch
    run_repl(controller, *args, debug=debug)
  File "/home/newt/pwndbg/pwndbg/dbg_mod/lldb/repl/fuzzy.py", line 303, in _wrapped
    return function(*a, **kw)
           ^^^^^^^^^^^^^^^^^^
  File "/home/newt/pwndbg/pwndbg/dbg_mod/lldb/repl/__init__.py", line 404, in run
    if not exec_repl_command(line, sys.stdout, dbg, driver, relay):
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/newt/pwndbg/pwndbg/dbg_mod/lldb/repl/__init__.py", line 468, in exec_repl_command
    return _exec_repl_command(line, output_to.buffer, dbg, driver, relay)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/newt/pwndbg/pwndbg/dbg_mod/lldb/repl/__init__.py", line 744, in _exec_repl_command
    dbg.relay_exceptions()
  File "/home/newt/pwndbg/pwndbg/dbg_mod/lldb/__init__.py", line 2250, in relay_exceptions
    raise pwndbg.dbg_mod.Error(e)
pwndbg.dbg_mod.Error: No escaped character
```

It doesn't trigger the crash if the `\` is part of the first word, like `nearpc\` (no spaces in-between)

patryk4815:
`LLDB.lex_args` used `shlex.split`, which raises `ValueError` on a command line ending in a backslash ("No escaped character") or with an unterminated quote ("No closing quotation"). `Command.invoke` only catches `TypeError` and `DebuggerError`, so the `ValueError` escaped into the LLDB command handler, got stashed in `_exception_relay` and was re-raised by `relay_exceptions` as a `DebuggerError`, tearing down the whole REPL:

    pwndbg-lldb> nearpc -f\
    ...
    pwndbg.dbg_mod.DebuggerError: No escaped character

GDB's `string_to_argv` is lenient here — it yields the incomplete argument instead of failing — so `nearpc -f\` just runs as `nearpc -f` under GDB. Add `pwndbg.lib.strings.lex_args`, which drives `shlex` directly and, on an unterminated escape or quote, keeps the token the lexer had accumulated. This matches GDB for every case checked against it, so a typo no longer kills the session and the two backends agree.

Fixes #4130

### DIFF
--- a/pwndbg/dbg_mod/lldb/__init__.py
+++ b/pwndbg/dbg_mod/lldb/__init__.py
@@ -7,7 +7,6 @@ import functools
 import os
 import random
 import re
-import shlex
 import sys
 from asyncio import CancelledError
 from collections.abc import Awaitable
@@ -32,6 +31,7 @@ import pwndbg.dbg_mod
 import pwndbg.lib.cache
 import pwndbg.lib.memory
 import pwndbg.lib.path
+import pwndbg.lib.strings
 from pwndbg.color import message
 from pwndbg.dbg_mod import EventHandlerPriority
 from pwndbg.dbg_mod import selection
@@ -2293,7 +2293,7 @@ class LLDB(pwndbg.dbg_mod.Debugger):
 
     @override
     def lex_args(self, command_line: str) -> list[str]:
-        return shlex.split(command_line)
+        return pwndbg.lib.strings.lex_args(command_line)
 
     def _any_inferior(self) -> LLDBProcess:
         """
--- a/pwndbg/lib/strings.py
+++ b/pwndbg/lib/strings.py
@@ -1,8 +1,41 @@
 from __future__ import annotations
 
 import re
+import shlex
 
 
 def strip_colors(text):
     """Remove all ANSI color codes from the text"""
     return re.sub(r"\x1b[^m]*m", "", text)
+
+
+def lex_args(command_line: str) -> list[str]:
+    """
+    Split a command line into arguments, the way a POSIX shell would.
+
+    Unlike `shlex.split`, this never raises on a malformed command line. A
+    trailing backslash or an unterminated quote simply yields the incomplete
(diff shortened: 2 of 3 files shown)


k4lizen:
but so, if I write `\n` in a command line will it interpret it like a newline?

patryk4815:
> but so, if I write `\n` in a command line will it interpret it like a newline?

no. `\` is same like in bash/shell, `\` is for new line

OBarronCS:
Nice! Tested it, and this brings the behavior in line with GDB.

OBarronCS (approved):

k4lizen:
needs to be linted :>

disconnect3d:
I'm ok with +2 lint errors from this :x 
(fwiw would be nice if the CI written the added lint errors down in the log)

Issue #4088, opened 2026-08-24
don't do backwards linear disassembly in the context when stepping


k4lizen:
We don't want to show the linear backwards disassembly in the disassembly context when we step into a function (because those are just the instructions from the previous function, and they clutter the output).

We want to show backwards linear disassembly only when we continued the process in some way `e.g. pwndbg> continue` or `pwndbg> nexti`, and hit a stop because of that command (e.g. pressed Ctrl-C, hit a breakpoint, caught a signal).

I don't really see a way in the gdb python docs to make this differentiation though: https://sourceware.org/gdb/current/onlinedocs/gdb.html/Events-In-Python.html#Events-In-Python . GDB seems to simply not expose enough event info.

Maybe we could leverage this: https://sourceware.org/gdb/current/onlinedocs/gdb.html/Hooks.html . But seems flaky.

Side-note: possibly with `nearpc` we want to do backwards disassembly in more cases than in the disassembly context.

k4lizen:
@OBarronCS maybe until we figure out something better, we can make it so that we do backwards disassembly only when:
1. we just stopped
2. the previous instructions aren't cached
3. we are now on a user-set breakpoint OR we just caught a signal

thoughts?

OBarronCS:
I think those are good.

One thing that is a shame is that currently on gdb, if you do `catch syscall`, and continue, it will break at the instruction after the syscall, but without a signal that we can observe. In these cases, it still would be nice to do this backwards disassembly. I wonder if there is a way to detect this event as well.

PR #4142, opened 2026-09-08
In context disasm, do not go backwards linearly above the current function

OBarronCS:
Addresses #4088.

#3898 introduced a way to "disassemble backwards" in memory. This is great for `nearpc`, however in the context disasm display, it can get confusing, such as when stepping into a function (we will display instructions from before the first instruction of the function, which is irrelevant)

This cleans things up by disallowing disassembling backwards beyond the current function when disassembling for the context display.

Before (starting `gdb ls`), first time context is printed:
image

After:
image

It will still go backwards within a function, which is nice, such as when doing `catch syscall`, or breaking at a breakpoint.
image

image

### DIFF
--- a/pwndbg/aglib/disasm/disassembly.py
+++ b/pwndbg/aglib/disasm/disassembly.py
@@ -744,17 +744,17 @@ def near(
     if instruction_flow_cache is not None:
         instruction_flow_cache.copy_current_instruction_flow_to_next()
 
-    function_boundaries: tuple[int, int] | None = None
-    if emulate:
-        # Only bother computing this when emulating
-        function_boundaries = pwndbg.aglib.symbol.resolve_function_boundaries(address)
-
     # Keep track of which of the previous instructions were disassembly linearly so we can display them as gray while emulating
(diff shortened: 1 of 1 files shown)


pwndbg/aglib/disasm/disassembly.py
```diff
@@ -743,6 +744,11 @@ def near(
     if instruction_flow_cache is not None:
         instruction_flow_cache.copy_current_instruction_flow_to_next()
 
+    function_boundaries: tuple[int, int] | None = None
+    if emulate:
+        # Only bother computing this when emulating
+        function_boundaries = pwndbg.aglib.symbol.resolve_function_boundaries(address)
```
disconnect3d: Should we maybe move this added code to the `if show_prev_insns:` branch?

OBarronCS: Good call, I moved it there

k4lizen (approved):

k4lizen:
I guess when there is no symbol it cannot work huh

image

image

maybe we could stop disassembling backwards when we hit `endbr64` :p

PR #4145, opened 2026-09-09
mypy not catching bad types?

### LINKED ISSUE #4115: mypy not catching bad types?
This PR has some erroneous typing but our lint does not catch it for some reason?

https://github.com/pwndbg/pwndbg/actions/runs/33532749547/job/99939678515?pr=4111

victorwon2001:

Use `mypy --strict --strict-optional` for both sides of the PR-vs-dev strict error baseline so new Optional/None incompatibilities are counted without changing the repository-wide mypy configuration. Contributor documentation now describes the same strict baseline mode.

## Verification
`./lint.sh --check` passes, and a focused type-check confirms that passing `None` to a `MemoryMap` parameter is rejected. A full strict-optional run exposes the existing type-error baseline, which remains handled by the unchanged PR-vs-dev count comparison.

Fixes #4115

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -79,35 +79,35 @@ jobs:
           filter_mode: nofilter
           fail_on_error: true
 
-      # The PR must not contain more mypy --strict errors than the dev branch does
-      - name: Count mypy --strict errors in PR and dev branch.
+      # The PR must not contain more mypy --strict --strict-optional errors than the dev branch does
+      - name: Count mypy --strict --strict-optional errors in PR and dev branch.
         id: mypy_strict_count
         run: |
           set +e
           # Taken from ./lint.sh
           LINT_FILES="pwndbg pwndbginit tests *.py scripts"
 
-          # Count `mypy --strict` errors on PR branch.
-          uv run --group dev --group lint --group tests --all-extras mypy --strict $LINT_FILES > mypy_pr.txt 2>&1
+          # Count `mypy --strict --strict-optional` errors on PR branch.
+          uv run --group dev --group lint --group tests --all-extras mypy --strict --strict-optional $LINT_FILES > mypy_pr.txt 2>&1
           PR_ERRORS=$(grep -c "error:" mypy_pr.txt || true)
           echo "pr_count=$PR_ERRORS" >> $GITHUB_OUTPUT
-          echo "PR mypy (--strict) errors: $PR_ERRORS"
+          echo "PR mypy (--strict --strict-optional) errors: $PR_ERRORS"
 
-          # Count `mypy --strict` errors on dev branch.
+          # Count `mypy --strict --strict-optional` errors on dev branch.
           git fetch origin dev
           git checkout origin/dev
-          uv run --group dev --group lint --group tests --all-extras mypy --strict $LINT_FILES > mypy_dev.txt 2>&1
+          uv run --group dev --group lint --group tests --all-extras mypy --strict --strict-optional $LINT_FILES > mypy_dev.txt 2>&1
           DEV_ERRORS=$(grep -c "error:" mypy_dev.txt || true)
           echo "dev_count=$DEV_ERRORS" >> $GITHUB_OUTPUT
-          echo "Dev mypy (--strict) errors: $DEV_ERRORS"
+          echo "Dev mypy (--strict --strict-optional) errors: $DEV_ERRORS"
 
           echo "============================="
-          echo "mypy --strict diff:"
+          echo "mypy --strict --strict-optional diff:"
           echo "============================="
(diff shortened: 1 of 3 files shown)


k4lizen:
show me that this fixes it on the example i provided please

disconnect3d:
> and a focused type-check confirms that passing `None` to a `MemoryMap` parameter is rejected.

So where is this focused type-check confirmation? Why don't you show us the code on which you tested it and a proof that the chnages you made improve this?

victorwon2001:
Here is the focused check behind that statement. I used the repository's actual `MemoryMap` type:

```python
import pwndbg.dbg_mod

def takes_memory_map(value: pwndbg.dbg_mod.MemoryMap) -> None:
    pass

takes_memory_map(None)
```

With the repository configuration (`strict_optional = false`), this passes:

```text
$ uv run --group dev --group lint --group tests --all-extras mypy --strict --follow-imports=silent -c '...'
Success: no issues found in 1 source file
```

Adding the flag used by this PR catches it:

```text
$ uv run --group dev --group lint --group tests --all-extras mypy --strict --strict-optional --follow-imports=silent -c '...'
<string>:6: error: Argument 1 to "takes_memory_map" has incompatible type "None"; expected "MemoryMap"  [arg-type]
Found 1 error in 1 file (checked 1 source file)
```

This also matches the real regression from #4111. In commit `a024bb9dd2ff5d1ef895d7b1285d82987886d9e0`, the function was declared with `vmmap_object: pwndbg.dbg_mod.MemoryMap`, while the non-verbose path called it as:

```python
await _step_and_log(ec, file_handle=f, vmmap_object=None, end_address=end_address)
```

GitHub Actions Lint run `33532749547` used that exact SHA and still concluded `success`; its `Run mypy`, `Count mypy --strict errors in PR and dev branch.`, and `Compare mypy --strict error counts` steps all succeeded. The workflow change here keeps the normal mypy step unchanged and adds explicit `--strict-optional` only to both sides of the existing PR-vs-dev strict error-count baseline, so existing Optional debt remains baselined while a new error like this is counted.

mbrla0:
@k4lizen Kind of a jank test, but applying the commit in this PR to #4111 does seem to cause it to fail, as it should. See [this workflow](https://github.com/pwndbg/pwndbg/actions/runs/34619468799) from my [horrible test PR](https://github.com/pwndbg/pwndbg/pull/4157).

k4lizen:
awesome sauce so mergeable?

mbrla0:
Give me a minute. There's some other stuff I want to check.

mbrla0:
Okay I think it's fine, but it's a bit strange. `--strict-optional` is [not documented](https://web.archive.org/web/20260101191335/https://mypy.readthedocs.io/en/stable/command_line.html#none-and-optional-handling) even in the version we use despite the fact that [it is present in the code](https://github.com/python/mypy/blob/5ff46f8b3706d005fabab2227ec84476d605bfd4/mypy/main.py#L735).

It will very likely increase the number of errors being reported in the `dev` branch, but it's only limited to diffing at PRs, so it's okay on that front.

mbrla0:
All in all, I think it's alright @k4lizen.

mbrla0:
Nope, hold the phone. That workflow failed for some other reason, it didn't even get to diffing. Urgh, hold on, sorry.

mbrla0:
Okay, that sure was a mess. Thanks, GitHub.

Testing _properly_ this time against #4111 both before and after applying the commit in this PR to it, both pass linting. See [here](https://github.com/mbrla0/pwndbg/actions/runs/34632840700/job/103376576387) for the before and [here](https://github.com/mbrla0/pwndbg/actions/runs/34634418086/job/103378879358) for the after. It's clear that in the after there are way more errors in general, but there's no increase in the number of errors in either scenario.

@victorwon2001 @k4lizen I could not reproduce the fix on #4111.

k4lizen:
Right but are we on the same page that the type issues have been fixed by the contributor after my comment on their PR, meaning that to reproduce them, you have to test on not their latest commit, but commit [a024bb9](https://github.com/pwndbg/pwndbg/pull/4111/commits/a024bb9dd2ff5d1ef895d7b1285d82987886d9e0).

mbrla0:
@k4lizen Yeah that was really silly of me, guess that's what I get for doing things in a rush in the middle of work. Against that commit specifically [it does work seem to work fine](https://github.com/mbrla0/pwndbg/pull/4). With a sensible-looking error and everything.

k4lizen:
Okay neat looks good, I'm not an ultra fan of how this PR looks because i'm afraid we'll just continue stacking flags, but alright maybe let's not overcomplicate for now if there is no need.

LGTM!

k4lizen (approved):

PR #4151, opened 2026-09-10
Fix: Show arrow right for contiguous marker in address chain

### LINKED ISSUE #4114: arrow in wrong direction in deref chain
Imo here:

Image

The last arrow should not be `<-` but it should rather be `->`.

Eirikwst:
Small consistency fix
As mentioned in https://github.com/pwndbg/pwndbg/issues/4114 it makes more sense for the contiguous marker to be preceded by a right arrow ->

image

image

Fixes https://github.com/pwndbg/pwndbg/issues/4114

### DIFF
--- a/pwndbg/emu/emulator.py
+++ b/pwndbg/emu/emulator.py
@@ -457,6 +457,9 @@ class Emulator:
         if len(chain) == 1:
             return enhanced
 
+        if len(chain) > limit:
+            return arrow_right.join([*rest, enhanced])
+
         return arrow_right.join(rest) + arrow_left + enhanced
 
     def telescope_enhance(self, value: int, code: bool = True, enhance_string_len: int = None):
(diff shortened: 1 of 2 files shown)


pwndbg/emu/emulator.py
```diff
@@ -457,6 +457,9 @@ def format_telescope_list(
         if len(chain) == 1:
             return enhanced
 
+        if len(chain) > limit:
+            return arrow_right.join([*rest, enhanced])
```
disconnect3d, line 461: Wouldn't it be better to do this instead?:

```py
# show "-> ..." if we are truncating the display
arrow = arrow_left if len(chain) <= limit else arrow_right

return arrrow_right.join(rest) + arrow + enhanced
```

disconnect3d:
Closing in favor of https://github.com/pwndbg/pwndbg/pull/4164

PR #4152, opened 2026-09-10
feat: add scudo chunk debugging command

zeguilherme1:
### feat: add scudo chunk debugging command

Scudo is the memory allocator present in Android since version 11. The main goal of this PR is add a debugging command to check info about a scudo chunk, which is a block of memory that is managed by the allocator.

The Unpacked Header of scudo consists in:

``` 
struct UnpackedHeader {
    uptr ClassId : 8;
    u8 State : 2;
    u8 OriginOrWasZeroed : 2;
    uptr SizeOrUnusedBytes : 20;
    uptr Offset : 16;
    uptr Checksum : 16;
};
```

If we check the scudo implementation, we can see that the header comes before the user_ptr, which is the pointer that we get when memory is allocated.

``` 
inline AtomicPackedHeader *getAtomicHeader(void *Ptr) {
    return reinterpret_cast<AtomicPackedHeader *>(
    reinterpret_cast<uptr>(Ptr) - getHeaderSize());
}
```

### Implementation
This command automates the information extraction of the Scudo header fields. By taking the user pointer (the address returned by the allocator) and applying the header offset.

```
ClassId (8 bits): Identifies the size class of the allocation.

State (2 bits): Indicates whether the chunk is available, allocated, or quarantined.

OriginOrWasZeroed (2 bits): Tracks the allocation origin (e.g., malloc, new) or zeroing state.

SizeOrUnusedBytes (20 bits): Stores the exact size for primary allocations or unused bytes for secondary allocations.

Offset (16 bits): The offset to the beginning of the block, used for alignment.

Checksum (16 bits): A security mechanism to detect header corruption.
```

### Usage

Pass the user pointer of the allocated chunk directly to the command:

```
pwndbg-lldb> scudo-chunk <address>
```

Example Output:

```
[*] Scudo Chunk at 0x778a6ecac000
classId : 14
state   : 1
origin  : 0
size    : 256 (0x100)
offset  : 0
checksum: 54321
```

scudo_chunk


k4lizen:
there should be some classes/funcs written for the scudo allocator aglib (like for jemalloc / glibc / mallocng) and this command should hook into that rather than reading / parsing raw memory in `pwndbg/commands/` like this

k4lizen:
the output should also follow the format of the previous heap inspection commands

PR #4155, opened 2026-09-11
Fix source install

### LINKED ISSUE #4144: Install from source method is broken
This <https://pwndbg.re/stable/setup/#system-gdb-sourced-from-gdbinit> install method causes the following error on `gdb` startup:

```
  File "/home/ubuntu/.local/share/uv/tools/pwndbg/share/pwndbg/gdbinit.py", line 109, in <module>
    main()
    ~~~~^^
  File "/home/ubuntu/.local/share/uv/tools/pwndbg/share/pwndbg/gdbinit.py", line 104, in main
    from pwndbginit.gdbinit import main_try
ModuleNotFoundError: No module named 'pwndbginit'
```

The root cause appears to be that `is_system_installation` is returning `True` incorrectly.

https://github.com/pwndbg/pwndbg/blob/3e08c7b45cce5b4091099db7bad80ad664b404b7/gdbinit.py#L87-L102

Based on the comment, it appears that it thinks it is being installed via a distro package..

Steps to reproduce:
```sh
docker run -it ubuntu:26.04

# Following commands are inside the container
apt update
DEBIAN_FRONTEND=noninteractive apt install -y gdb curl git
curl -LsSf https://astral.sh/uv/install.sh | sh

exec bash

# Clone pwndbg
git clone https://github.com/pwndbg/pwndbg.git

cd pwndbg

# Optionally, checkout a branch
# git fetch origin pull/4102/head:pr-4102
# git checkout pr-4102

# Install
PY_VER=$(gdb -nx --batch -iex 'py import sysconfig; print(sysconfig.get_config_var("VERSION"))')
uv tool install --python=$PY_VER .
echo "source $(uv tool dir)/pwndbg/share/pwndbg/gdbinit.py" >> ~/.gdbinit
```

OBarronCS:
Fixes #4144

I tested it in a docker container and it the install from source method now works.

The code was using a `is_system_installation`, which return a incorrect_value (it checked for a `.pwndbg_root` file next to `gdbinit.py`, doesn't exist in `uv tool install` installation).

Instead, change the function to check if we have a venv we need to fixup paths with.


gdbinit.py
```diff
@@ -84,22 +79,25 @@ def main() -> None:
...
-    # we don't import the `pwndbginit` package before `fixup_paths()` fixes `sys.path`.
-    skip_venv = is_system_installation(src_root)
+    # If Pwndbg is installed by a distro package manager, we don't have a virtualenv that requires path fixups
+    # Note we do NOT import `pwndbginit` package before `fixup_paths()` has corrected `sys.path`. Otherwise,
+    # a system-wide Pwndbg installation (e.g. an Arch `pwndbg` package in the global
+    # site-packages) could shadow the source checkout that is being sourced, leading
+    # to a confusing mix of modules loaded from two different locations. See:
+    # https://github.com/pwndbg/pwndbg/issues/3963
+
+    venv_dir: Path | None = get_venv_path_if_fixup_needed(src_root)
```
patryk4815: redundant type

```suggestion
    venv_dir = get_venv_path_if_fixup_needed(src_root)
```

patryk4815 (approved):
logic looks good

gdbinit.py
```diff
@@ -36,39 +36,34 @@ def fixup_paths(src_root: Path, venv_path: Path):
...
         return src_root.parent.parent
 
-    return src_root / ".venv"
+    # If pwndbg is installed in `/venv/lib/pythonX.Y/site-packages/pwndbg/`,
+    # the `.pwndbg_root` file will not exist because `src_root` will point to the
+    # `/venv/lib/pythonX.Y/site-packages/` directory, not the original source directory.
+    #
+    # However, if pwndbg is installed in editable mode (our recommended way), this file
+    # will exist, and the condition will be False, allowing auto-update.
+    if (src_root / ".pwndbg_root").exists():
```
patryk4815, line 58: Comment above is "improper" 🤔 it should say something like?:
Handle case when you use `source /path/to/pwndbg-git-dir/gdbinit.py` + `.pwndbg_root` + only venv

return None:
Handle case when you use `source /path/to/not-pwndbg-dir/gdbinit.py` + without venv, only system pwndbg
Eg handle archlinux case: `source /usr/share/pwndbg/gdbinit.py`

k4lizen: pls update the comments so they're as accurate as possible, the loading of pwndbg is such a confusing part of the codebase :woozy_face:

OBarronCS: I updated the comments

k4lizen (approved):

k4lizen:
for sure we're done and installation will never break again :partying_face:

k4lizen:
thanks!

PR #4159, opened 2026-09-11
Make branch visualization offsets better

OBarronCS:
A small change to branch visualization. If you have two overlapping jumps in the disassembly output, now the shorter jump (which is possibly contained within the larger jump) has it's lines closer to the right side of the disassembly.

Before:
image

After:
image


k4lizen:
That makes a lot of sense, nice!

k4lizen (approved):

PR #4160, opened 2026-09-11
Reorganize install docs

OBarronCS:
Reorganize the install sections of the docs

image

image

### DIFF
--- a/docs/contributing/index.md
+++ b/docs/contributing/index.md
@@ -98,12 +98,6 @@ If you wish to preview the documentation locally, you may do so by running:
 ```{.bash .copy}
 ./scripts/docs-live.sh
 ```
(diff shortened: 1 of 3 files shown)


k4lizen:
hmm do you disagree with the idea of making it multiple pages? helix does it this way for instance and i think its quite legible https://docs.helix-editor.com/install.html . though i wouldn't make the secondary pages appear in the top bar, just be linked from the main one

k4lizen:
Your changes make it much more managable so I feel less of a need to separate it out. I added uninstall instructions and modified a few things, let me know how you feel about it.

k4lizen:
image

OBarronCS:
Nice! It's looking much better now

Issue #4158, opened 2026-09-11
nearpc - don't show arguments when on a call


OBarronCS:
When calling nearpc, it should not show arguments if you are stepped on top of a call

<img width="1062" height="386" alt="Image" src="https://github.com/user-attachments/assets/266ec348-dafe-4162-990a-019e741b1ba1" />

OBarronCS:
Fixed in #4161

PR #4161, opened 2026-09-12
Do not show args with nearpc command

OBarronCS:
The `nearpc` command is meant for disassembling linear flow of instructions. It is an replacement for gdb `x/20 ADDRESS`. It sometimes showing arguments (if the program PC happens to be sitting on a call) makes for a confusing display sometimes, as it conditionally introduces new rows when disassembling a function, just because the `pc` happens to be there right now.

It still shows arguments on the context disasm and with `emulate` command.

Before:
See screenshot in #4158

After:
image


k4lizen:
test :<

k4lizen (approved):
LGTM

PR #4162, opened 2026-09-12
Context disasm: target now shown for future ret to unmapped address

OBarronCS:
While stepping through a ROP chain, I came across an interesting bug: if you were NOT stepped on a ret (it was in the future), and this `ret` caused the program counter to jump to a non-mapped memory address, then a target would not be displayed:

image

Note that the `ret` has no `<target>` marked.

The emulator crashes in this case, and because `ret` is special (the return address is an implicit operand on the stack), we didn't have anything to handle getting the address that faulted in this case.

This PR fixes this by adding a better system that tracks how the Unicorn engine fails to execute an instruction (typically a memory fault), and uses this information to display the `ret to unmapped page`.

I also found some an interesting edge cases in the emulator (that produces strange behavior when jumping to address 0), and added better type annotations around the code that used an emulator object.

After:
image

### DIFF
--- a/pwndbg/emu/emulator.py
+++ b/pwndbg/emu/emulator.py
@@ -7,7 +7,9 @@ from __future__ import annotations
 import binascii
 import re
 import string
-from typing import NamedTuple
+from dataclasses import dataclass
+from enum import Enum
+from enum import auto
 
 import capstone6pwndbg as C
 import unicorn as U
@@ -183,7 +185,7 @@ enable_virtual_tlb: dict[PWNDBG_SUPPORTED_ARCHITECTURES_TYPE, bool] = {
 ) = (0, 1, 2, 4, 8, 16, 32, 64, 128)
 
 DEBUG = NO_DEBUG
-# DEBUG = -1 # ALL
+# DEBUG = -1  # ALL
 # DEBUG = DEBUG_EXECUTING | DEBUG_MEM_MAP | DEBUG_MEM_READ
 
 if DEBUG != NO_DEBUG:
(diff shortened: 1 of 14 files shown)


OBarronCS:
Added a test for this as well

pwndbg/emu/emulator.py
```diff
@@ -755,6 +836,13 @@ def emulate_with_hook(self, hook, count=512) -> None:
         # which have Thumb mode transitions, Unicorn will internally handle them.
         pc |= self.read_thumb_bit()
 
+        # Tell Unicorn that we want to stop at some explicit exit points, and
+        # then make the list of exit points empty.
+        # This prevents Unicorn from using the "until" argument (second argument below),
+        # which would cause it to always stop before executing at address 0 (which may be mapped!)
+        self.uc.ctl_exits_enabled(True)
+        self.uc.ctl_set_exits([])
+
```
OBarronCS, line 848: This fixes an interesting edge case I came across while debugging this.

To step the emulator, we do `self.emu_start(pc, 0, count=count)`, which sends `0` as the "until" parameter to Unicorn. This means that if the PC is ever set the 0, Unicorn will stop executing. This meant that the address at 0 was never attempted to be executed, and thus would not segfault.

These settings make Unicorn ignore the 'until' setting, making it step no matter what instruction it is on

k4lizen (approved):

pwndbg/emu/emulator.py
k4lizen: haha wild

k4lizen:
need to merge dev after https://github.com/pwndbg/pwndbg/pull/4161 gets merged

PR #4164, opened 2026-09-12
Fix telescope not taking dereference-limit plus showing wrong arrow when deref limit is hit

### LINKED ISSUE #4114: arrow in wrong direction in deref chain

disconnect3d:
Last link of telescope deref chain was always joined with left arrow, even when chain was cut off by dereference limit and the last element is contiguous marker (`...`).

This was misleading, so changing it so left arrow is used only when whole chain was dereferenced. Otherwise, the right arrow is used.

Applying same fix to the emulator code as well.

Fixes #4114

In addition, the `limit = int(LIMIT)` -> `limit = LIMIT` change in `pwndbg.chain.format` fixes the problem that `telescope ptr` did not apply the dereference limit set via `set dereference-limit 1`.

### DIFF
--- a/pwndbg/chain.py
+++ b/pwndbg/chain.py
@@ -141,9 +141,6 @@ def format(
     else:
         chain = get(value, limit, offset, hard_stop, hard_end, safe_linking=safe_linking) or []
 
-    arrow_left = c.arrow(f" {config_arrow_left} ")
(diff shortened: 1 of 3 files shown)


k4lizen:
nice!

Issue #4163, opened 2026-09-12
Display indirect jumps that dereference unmapped memory


OBarronCS:
Currently, if we have an indirect jmp, where the location of the jump target is in an unmapped memory page, there is no indication of this in the context display (executing `si` in these examples would segfault)

<img width="1245" height="98" alt="Image" src="https://github.com/user-attachments/assets/6c76d4f7-6b31-41ce-bfbb-5a1fc9020760" />

<img width="1355" height="108" alt="Image" src="https://github.com/user-attachments/assets/cf25b6ee-3192-41d5-be18-c293b8b1e681" />

OBarronCS:
Will look like this:

<img width="1345" height="101" alt="Image" src="https://github.com/user-attachments/assets/edaf192e-f37d-4251-94e8-b7059990d3bc" />

OBarronCS:
Implemented in #4165

PR #4165, opened 2026-09-13
Context disasm: handle memory indirect jumps better

OBarronCS:
There 3 cases for a memory indirect jump (in context disasm display), such as `call [rax]`

- (1). The memory address is known and readable. The jump target is a valid memory location: display the address. This has worked

- (2). The memory address is known and dereferenced. The target of the jump is in unmapped memory
    - This PR fixes this case by showing the target correctly:
 
    - (https://github.com/pwndbg/pwndbg/pull/4162 fixed it for specifically for rets)

Before:
image

After:
image

(this did not work because of a bug in how/when we queried the emulator for the target memory location. tl;dr: the instruction would cause the emulator to segfault and disable itself, preventing us from querying from its memory to determine what the jump target even was)

- (3). The memory address is known, but cannot be dereferenced
    - Previously, nothing was shown. This PR fixes this as well

    
Before:
image

After:
image

Added tests for both of these cases, which work with emulation for the current + future instructions, and without emulation when you are stepped on the instruction.


OBarronCS:
Resolves #4163

pwndbg/emu/emulator.py
```diff
@@ -318,7 +320,8 @@ def __init__(self) -> None:
         debug(DEBUG_INIT, "uc = U.Uc(%r, %r)", (arch_to_UC[self.arch], self.uc_mode))
         self.uc = U.Uc(arch_to_UC[self.arch], self.uc_mode)
 
-        if enable_virtual_tlb.get(self.arch, False):
+        # TODO: this is temporary for testing. Probably make this always true
+        if True or enable_virtual_tlb.get(self.arch, False):
             debug(DEBUG_INIT, "# Setting TLB mode to virtual")
             self.uc.ctl_set_tlb_mode(U.UC_TLB_VIRTUAL)  # type: ignore[attr-defined]
```
OBarronCS: Virtual TLB means that Unicorn does not do CPU-specific logic for address translation. If an instruction dereference memory at address `x`, the address read from memory is `x` (treating it as a single linear map of memory)

Without this, on x86_64, Unicorn is masking all addresses to 52 bits. This means when we request any address (such as kernel address) with the high bits set, they got masked away, and a different address is requested than intended. This is because the the x86_64 CPU's MMU being used has this behavior, so all memory access get masked.

Enable virtual tlb fixes this issue. We map pages to the emulator at their virtual address, so we want the allow the emulator to just access everything we map directly.

OBarronCS:
Due to enabling the virtual TLB on Unicorn for all architectures, this PR also has the side effect of making emulation work while debugging x86_64 kernels.

pwndbg/emu/emulator.py
```diff
@@ -164,12 +164,6 @@ def create_reg_to_const_map(
     "s390x": create_reg_to_const_map(arch_to_UC_consts["s390x"]),
 }
 
-# Architectures for which we want to enable virtual TLB mode
-enable_virtual_tlb: dict[PWNDBG_SUPPORTED_ARCHITECTURES_TYPE, bool] = {
-    "s390x": True,
-    "powerpc": True,
-    "sparc": True,
-}
```
k4lizen, line 172: huh why did we only do it on these arches b4?

pwndbg/aglib/disasm/instruction.py
```diff
@@ -212,6 +213,27 @@ class CacheSource(Enum):
     NOT_FROM_CACHE = ""
 
 
+@dataclass
+class MemoryDereferenceInfo:
+    """
```
k4lizen, line 218: i like that you make these classes often, i think it helps readability

k4lizen (approved):

pwndbg/emu/emulator.py
OBarronCS: These were enabled because userspace processes in these architectures had similar memory addressing issues that came up when testing emulation for them, where the MMU was changing the address in ways we didn't want to. However, because we just didn't realize the same bug was happening (until this PR) in amd64 kernel addresses, I assumed that it was just quirks of these architectures.

Issue #4170, opened 2026-09-14
sanity check heuristic heap tcache detection


k4lizen:
In the docstring it says:
```python
    @property
    @override
    def thread_cache(self) -> pwndbg.aglib.heap.glibc_structs.TcachePerthreadStruct | None:
        """Locate a thread's tcache struct. We try to find its address in Thread Local Storage (TLS) first,
        and if that fails, we guess it's at the first chunk of the heap.
        """
```
In the latest version of glibc, it may not be the first chunk of the heap. Does this heuristic still make sense, should we really be guessing this?

See related: https://github.com/pwndbg/pwndbg/pull/4133#issuecomment-5634278059 .

Issue #4169, opened 2026-09-14
aarch64 glibc heap debug syms tcache detection not working because variable is optimized out


k4lizen:
See https://github.com/pwndbg/pwndbg/actions/runs/34154457298/job/101843304674?pr=4133 and https://github.com/pwndbg/pwndbg/pull/4133#issuecomment-5634989580 .

I don't know if we have a better option here than downgrading to heuristic heap detection, maybe thats fine, but needs to be actually done.

PR #4172, opened 2026-09-15
glibc heap improvements

kozabrada123:
+ [X] I read the contributing documentation
+ [ ] I include before and after screenshots (or relevant outputs) of the new feature/fixed bug

This has now evolved into a broader refactor of the glibc heap, merging `HeuristicHeap`, `DebugSymsHeap` and `GlibcMemoryAllocator` (the abstract class they both inherited) into just `GlibcHeap`. This was done mostly because a) as seen in #4133 and #3905 especially, symbols aren't just there or not there, and b) some code was duplicated between both classes, and `HeuristicHeap` already checked for symbols in most cases.

`resolve-heap-via-heuristic`, which previously allowed users (and tests) to specify which implementation to use, has now been moved into an enum inside `GlibcHeap` to allow for the same functionality.

This pr (as per its original purpose) also changes discovering the tcache on the heap after glibc 2.42, as it is now no longer necessarily the first chunk on the heap (see #4133). As a last resort, we take the first chunk that is of the appropriate size, which _can_ result in false positives (though is a last resort anyway, as other heuristics or symbols are more reliable and always used whenever possible).

Lastly, another fix (8ca9ad575476bfd48c1f03c43dfadfaed639b7a3) improves how we discover the new tcache dummy objects: we can now leverage debug symbols for that as well, along with accounting for optimizations of the dummy into .text (RX vs just RO).

Original description:

> Should be the follow up to #4133, doing #4169 and #4170 - the heap heuristic is pretty useful actually; I've tried to expand it now to also try to find the tcache in heap chunks, as a last resort. (Note: this is prone to an edgecase if we allocate a chunk that is exactly the same size as the tcache before the tcache itself. -- I think this isn't that big of an issue, since we should find it in the tls anyway - this is a redundancy for a redundancy.) 
> 
> My main criticism of what I've done is the copied code from the heuristic heap, as I am unsure how to cleanly reuse it in this codebase - as far as I understand, it would be virtually impossible to directly call those functions, and I don't know how I would go about abstracting them to be useful in both.


kozabrada123:
Ah, it seems the actions are borked

k4lizen:
Very cool!

I think instead of copying heuristicheap code over to the debugsymsheap, we should detect that the debugsymsheap cannot resolve the heap using the symbols it has on hand, and make it give up and offload its work to the heuristicheap.

One way to do this could be by checking if we can resolve the `tcache` symbol (or whatever we had issues with being optimized out on aarch64) in DebugSymsHeap.can_be_resolved

kozabrada123:
Thanks :p

I'll check, but I think that almost everything else could still use the DebugSymsHeap implementation, so we'd probably only want to partially offload to the heuristic heap, which I couldn't find a clean way of doing.

The new `thread_cache` methods differ _slightly_ due to how the struct is handled in both implementations. Ultimately it would be a pretty big win if we could call `_search_tls`, `_is_tcache_dummy` and `_is_tcache_struct` from the heuristic heap directly in DebugSymsHeap - they use the exact same members from `self` in both classes.

k4lizen:
i understand that it may be functionally a bit better to do it this way, but i think it is bad for the maintainability
of the code to do it this way, and this subsystem is already quite complicated

realistically these two classes should probably be merged into one that just first tries to use debug info / symbols and if they are not present hooks into heuristics (and does so for each piece of symbol / debug info), rather than having two classes, but thats a bigger refactor for another day

kozabrada123:
fair, that's totally understandable

at a quick glances it seems like `HeuristicHeap` already tries to use symbols in many places; I might actually give a quick try to merging them both

k4lizen:
> at a quick glances it seems like HeuristicHeap already tries to use symbols in many places; I might actually give a quick try to merging them both

that'd be awesome!

kozabrada123:
I've certainly managed something :p
Down to just one failing test on aarch64 now, I'll take a look tomorrow

k4lizen:
image

i love to see this

pwndbg/aglib/heap/glibc.py
```diff
@@ -1129,1183 +1141,1061 @@ def arenas(self) -> tuple[Arena, ...]:
...
 
-                sbrk_region = self.get_region(sbrk_base)
-                if sbrk_region is None:
-                    raise ValueError("mp_.sbrk_base is unmapped or points to unmapped memory.")
-                sbrk_region.memsz = sbrk_region.end - sbrk_base
-                sbrk_region.vaddr = sbrk_base
+class HeapDebugMethod(int, Enum):
+    Heuristic = 0  # Force using only heuristics
+    DebugInfo = 1  # Force using only debug info
+    Auto = 2  # Allow both
```
k4lizen: I would move this definition up above

But more than that I would probably just remove it and make `resolve-heap-via-heuristic` a noop.

pwndbg/aglib/heap/glibc.py
```diff
@@ -1129,1183 +1141,1061 @@ def arenas(self) -> tuple[Arena, ...]:
...
+                #   .mutex = _LIBC_LOCK_INITIALIZER,
+                #   .next = &main_arena,
+                #   .attached_threads = 1
+                # };
 
-    @property
-    def heap_info(self) -> TheType | None:
-        raise NotImplementedError()
+                # FIXME: unsure if there's a way to make it work with pwndbg.dbg_mod.Type,
+                # the main issue appears to be creating a proper instance of the type
```
k4lizen, line 1194: i don't understand this comment

pwndbg/aglib/heap/glibc.py
```diff
@@ -1129,1183 +1141,1061 @@ def arenas(self) -> tuple[Arena, ...]:
...
 
-    @property
-    @pwndbg.lib.cache.cache_until("objfile")
-    def size_sz(self) -> int:
-        """Corresponds to SIZE_SZ in glibc malloc.c"""
-        return pwndbg.aglib.arch.ptrsize
+                # If we are still not able to find the main_arena, probably we are debugging a binary with statically linked libc and no PIE enabled
+                if not self._main_arena_addr:
+                    # Try to find the default main_arena struct in the .data section
+                    for i in range(0, size - self.malloc_state.sizeof, pwndbg.aglib.arch.ptrsize):
```
k4lizen, line 1269: fixable type error on self.malloc_state.sizeof

pwndbg/aglib/heap/glibc.py
```diff
@@ -1129,1183 +1141,1061 @@ def arenas(self) -> tuple[Arena, ...]:
...
+                for x in ("tcache_bins", "tcache_small_bins")
+            )
 
-    def _request2size(self, req: int) -> int:
-        """Corresponds to request2size in glibc malloc.c"""
-        if req + self.size_sz + self.malloc_align_mask < self.minsize:
-            return self.minsize
-        return (req + self.size_sz + self.malloc_align_mask) & ~self.malloc_align_mask
+        # There is no debug symbols, we determine the tcache_bins existence by checking glibc version only
+        return self.is_initialized() and pwndbg.libc.version() >= (2, 26)
```
k4lizen: i'm confused by this because for the heruistic heap `self.mp` also resolved to a struct, and even now to me it seems like `SymbolNotRecoveredError` will be raised if `self.mp is None`, so when is this code ever reached?

pwndbg/aglib/heap/glibc.py
```diff
@@ -1129,1183 +1141,1061 @@ def arenas(self) -> tuple[Arena, ...]:
...
-            size & NON_MAIN_ARENA,
-        )
+    def _get_heap_page(self) -> pwndbg.lib.memory.Page | None:
+        """Get the [heap] memory page."""
+        return next((p for p in pwndbg.aglib.vmmap.get() if p.is_heap), None)
 
-    def chunk_key_offset(self, key: str) -> int | None:
-        """Find the index of a field in the malloc_chunk struct.
+    def _get_heap_range(self) -> pwndbg.lib.memory.Page | range:
+        """Get the heap start & end"""
```
k4lizen:
```suggestion
        """Get the start & end of the [heap] mapping"""
```
This should not return `pwndbg.lib.memory.Page | range`, one type should be used.

pwndbg/aglib/heap/glibc.py
```diff
@@ -1129,1183 +1141,1061 @@ def arenas(self) -> tuple[Arena, ...]:
...
+        if not pwndbg.aglib.memory.is_readable_address(addr):
+            return False
 
-    def get_heap(self, addr: int) -> TheValue | None:
-        raise NotImplementedError()
+        cand_arena = Arena(addr)
 
-    def get_tcache(self, tcache_addr: int | None = None) -> TheValue | None:
-        raise NotImplementedError()
+        if not pwndbg.aglib.memory.is_readable_address(cand_arena.next):
```
k4lizen, line 1322: why can `cand_arena.next` be None?

pwndbg/aglib/heap/glibc.py
```diff
@@ -1129,1183 +1141,1061 @@ def arenas(self) -> tuple[Arena, ...]:
...
+        """Check if addr points to a tcache dummy (glibc >= 2.43, read-only, all zeros)."""
+        if self._thread_cache_dummy_addr is not None and addr == self._thread_cache_dummy_addr:
+            return True
 
-        if arena is None:
-            return None
+        # NOTE: ro / rx, can be optimized inside .text as it's just a bunch of zeroes
+        page = pwndbg.aglib.vmmap.find(addr)
+        if page.read and not page.write:
+            return False
```
k4lizen, line 1345: i think this check might be inverted?

pwndbg/aglib/heap/glibc.py
```diff
@@ -1129,1183 +1141,1061 @@ def arenas(self) -> tuple[Arena, ...]:
...
-        # counts was renamed to num_slots in newer version of GLIBC 2.42
-        try:
-            counts = tcache["num_slots"]
-        except Exception:
-            counts = tcache["counts"]
-        entries = tcache["entries"]
+            ptr_size = pwndbg.aglib.arch.ptrsize
+            if pwndbg.libc.version() >= (2, 42):
+                return chunk.real_size - ptr_size == tcache_size
+            return chunk.real_size - ptr_size * 2 == tcache_size
```
k4lizen, line 1371: wild that glibc changed this

pwndbg/aglib/heap/glibc.py
```diff
@@ -1129,1183 +1141,1061 @@ def arenas(self) -> tuple[Arena, ...]:
...
 
-        num_tcachebins = entries.type.sizeof // entries.type.target().sizeof
-        safe_lnk = pwndbg.libc.glibc.check_safe_linking(pwndbg.libc.version())
+        if pwndbg.libc.version() < (2, 43):
+            return False
```
k4lizen: i would pull this into _is_tcache_dummy()

pwndbg/aglib/heap/glibc.py
```diff
@@ -1129,1183 +1141,1061 @@ def arenas(self) -> tuple[Arena, ...]:
...
-                return (TCACHE_LARGE_START_SIZE << (idx - TCACHE_SMALL_BINS)) - self.size_sz
-            return idx * self.malloc_alignment + self.minsize - self.size_sz
+        return self._is_tcache_dummy(addr)
 
-        # TODO: use `__tcache_dummy` symbol when we have debug syms
-        page = pwndbg.aglib.vmmap.find(tcache.address)
-        assert page
-        is_dummy = pwndbg.libc.version() >= (2, 43) and page.ro
+    @property
+    def thread_arena(self) -> Arena | None:
```
k4lizen, line 1376: hm it being `None` i think means the pointer is null which is different from SymbolNotRecovered which happens in the
```
     if not self.method.allow_heuristics:
            return None
```
codepath. Maybe a docstring should be added.

pwndbg/aglib/heap/glibc.py
```diff
@@ -1129,1183 +1141,1061 @@ def arenas(self) -> tuple[Arena, ...]:
...
-        """Returns: tuple(chain, count) or None"""
-        # Delay import so that libc can be loaded
-        from pwndbg.aglib.heap.glibc_structs import DEFAULT_MP_
-        from pwndbg.aglib.heap.glibc_structs import TCACHE_SMALL_BINS
+        self._thread_cache_dummy_addr = addr
+        return True
 
-        TCACHE_LARGE_START_SIZE = 1 << (DEFAULT_MP_.tcache_max_bytes.value.bit_length() - 1)
+    def _is_tcache_struct(self, addr: int) -> bool:
+        """Check if addr points to a valid tcache_perthread_struct."""
```
k4lizen: the docstring should be clarified to mention that this `!=` that the chunk is actually a free tcache chunk, just similar looking

pwndbg/aglib/heap/glibc.py
```diff
@@ -1129,1183 +1141,1061 @@ def arenas(self) -> tuple[Arena, ...]:
...
-                        else 124 + (sz >> 18)
-                        if (sz >> 18) <= 2
-                        else 126
-                    )
-                )
+    @property
+    def global_max_fast(self) -> int:
+        if self.method.allow_debuginfo:
+            global_max_fast_via_symbol = pwndbg.aglib.symbol.lookup_symbol_addr(
+                "global_max_fast", prefer_static=True
```
k4lizen, line 1563: so like lots of places in the code that check symbol checks behind `self.method.allow_debuginfo` whereas this didn't used to be the case. also the naming is a bit confusing since a symbol like this is not necessarily debug info but can be just non-stripped.

pwndbg/aglib/heap/glibc.py
```diff
@@ -1129,1183 +1141,1061 @@ def arenas(self) -> tuple[Arena, ...]:
...
 
-        tcache = self.get_tcache(tcache_addr)
+        if not pwndbg.aglib.memory.is_readable_address(addr - chunk_header_size):
+            return False
+        if not pwndbg.aglib.memory.is_readable_address(addr + tcache_size):
+            return False
 
-        if tcache is None:
-            return None
+        heap_range = self._get_heap_range()
```
k4lizen, line 1364: so `_is_tcache_struct` only works on single threaded / the main heap :? should be documented

pwndbg/aglib/heap/glibc.py
```diff
@@ -1129,1183 +1141,1061 @@ def arenas(self) -> tuple[Arena, ...]:
...
+        if malloc's `mp_` struct can't be resolved.
+        """
+        # Initialize malloc's mp_ struct if necessary.
+        if not self._mp_addr:
+            try:  # noqa: SIM105
+                # the `mp` getter has lots of "side effects"
+                # one of which is setting `self._mp_addr` (if resolving mp succeeds)
+                self.mp  # noqa: B018
             except Exception:
+                # Should only raise SymbolNotRecoveredError, but the heuristic heap implementation is still buggy so catch all exceptions for now.
```
k4lizen: we should only catch `SymbolNotRecoveredError`

pwndbg/aglib/heap/glibc.py
```diff
@@ -1129,1183 +1141,1061 @@ def arenas(self) -> tuple[Arena, ...]:
...
-                    for relocation in relocations:
-                        r_offset = relocation.entry.r_offset
+                return sbrk_region
+            raise ValueError("mp_.sbrk_base is unmapped or points to unmapped memory.")
+        raise SymbolNotRecoveredError("mp_", "Heuristic failed.")
 
-                        # We only care about the relocation in .data section
-                        if r_offset - next_field_offset < data_section_offset:
-                            continue
+    def get_region(self, addr: int | pwndbg.dbg_mod.Value | None) -> pwndbg.lib.memory.Page | None:
```
k4lizen, line 1851: this doesn't use self and should probably not be in this class

pwndbg/aglib/heap/glibc.py
```diff
@@ -1129,1183 +1141,1061 @@ def arenas(self) -> tuple[Arena, ...]:
...
+        return self.largebin_index_32(sz)
 
-    @property
-    @pwndbg.lib.cache.cache_until("objfile")
-    @override
-    def malloc_par(self) -> type[pwndbg.aglib.heap.glibc_structs.MallocPar] | None:
-        if not self.struct_module:
-            return None
-        return self.struct_module.MallocPar
+    def is_initialized(self) -> bool:
```
k4lizen, line 2174: you deleted the docstring

pwndbg/aglib/heap/glibc.py
```diff
@@ -1129,1183 +1141,1061 @@ def arenas(self) -> tuple[Arena, ...]:
...
-        if not self._mp_addr:
-            try:  # noqa: SIM105
-                # the `mp` getter has lots of "side effects"
-                # one of which is setting `self._mp_addr` (if resolving mp succeeds)
-                self.mp  # noqa: B018
-            except Exception:
-                # Should only raise SymbolNotRecoveredError, but the heuristic heap implementation is still buggy so catch all exceptions for now.
-                pass
+    def is_statically_linked(self) -> bool:
+        return not pwndbg.dbg.selected_inferior().is_dynamically_linked()
```
k4lizen, line 2198: should not be in this class

pwndbg/aglib/heap/glibc.py
```diff
@@ -1096,19 +1089,38 @@ def largebin_size_range_from_index(self, index: int):
...
+                    importlib.import_module("pwndbg.aglib.heap.glibc_structs")
+                )
+            except AssertionError:
+                raise
+            except Exception:
+                pass
+        return self._structs_module
+
     def can_be_resolved(self) -> bool:
-        """
```
k4lizen, line 1100: why was the docstring removed, has the contract changed?

k4lizen (changes requested):
Alright so since I had to go through all the code I left comments which are not applicable to your PR but are just to the general code so I/we can reference them in the future.

I think this is pretty great. The only thing I need you to check before merging is the
```python
         if page.read and not page.write:
```
check, everything else is nitpicks.

k4lizen:
Also as a general statement we should make resolve-heap-via-heuristic a no-op and not gate any debug info checks imo.

Non-related to your PR but from looking at the code:

If we fail to resolve a symbol/type once we shouldn't look it up again because it can be an expensive operation and we should just instantly fall back to heuristics.

Also it's ultra unclear to me why some classes return None and why some throw SymbolNotResolved(). Should be more consistent and clarified on the ones that stray from the general behaviour.

kozabrada123:
Hi, thanks for the review :D

I'll take a closer look at each point soon, just a few quick thoughts:

> The only thing I need you to check before merging is the
> `if page.read and not page.write:`
> check, everything else is nitpicks.

Good catch, seems I goofed up there

> Also as a general statement we should make resolve-heap-via-heuristic a no-op and not gate any debug info checks imo.

I thought about this, I think it would make for fairly nicer code. Ultimately I decided to keep it mostly since tests rely on `set resolve-heap-via-heuristic force` to test the heuristics work*, but also to maintain compatibility - if we're sure about the public api change, we can remove it. 

\* though sometimes they do fake the symbol not being available, which is probably a better option

> If we fail to resolve a symbol/type once we shouldn't look it up again because it can be an expensive operation and we should just instantly fall back to heuristics.

Good point; I'm not 100% sure how I'd go about implementing this on a per-symbol basis, maybe just saving a list of symbols we know don't exist?

Since this is already such a major refactor I'm also not against fixing the small stuff along the way before merging :P

k4lizen:
> Good point; I'm not 100% sure how I'd go about implementing this on a per-symbol basis, maybe just saving a list of symbols we know don't exist?

yeah i wouldn't touch this rn it'd probably be a bit involved

> if we're sure about the public api change, we can remove it.

yeah i'm sure. we shouldn't remove the config rn so as not to break lots of people but making it a no-op is fine. we would ideally put a hook in `_load_gdbinit()` or `get_set_string()` so we can print a deprecation warning

> Since this is already such a major refactor I'm also not against fixing the small stuff along the way before merging :P

if you wanna take a look at the smaller stuff see if you can handle some of
```bash
uv run --all-groups --all-extras mypy --strict --strict-optional ./pwndbg/aglib/heap/glibc.py
uv run --all-groups --all-extras mypy --strict --strict-optional ./pwndbg/commands/glibc_malloc.py
uv run --all-groups --all-extras mypy --strict --strict-optional ./tests/library/dbg/tests/test_glibc_heap.py
# etc
```
hopefully if everything is architectured/typed properly stuff should start falling into place

pwndbg/aglib/heap/glibc.py
kozabrada123: To elaborate a bit more, due to the merge of both heaps, `self.malloc_state` can now be `pwndbg.dbg_mod.Type` or `glibc_structs.MallocState` - the heuristic here creates an `expected` object and searches for its literal byte representation, which afaik is only possible with `glibc_structs.MallocState` (forcing us to fallback to the struct module)
```python
                expected = malloc_state._c_struct()
                expected.attached_threads = 1
                [...]
                            expected.next = addend
                [...]
                # If we are still not able to find the main_arena, probably we are debugging a binary with statically linked libc and no PIE enabled
                if not self._main_arena_addr:
                    # Try to find the default main_arena struct in the .data section
                    for i in range(0, size - self.malloc_state.sizeof, pwndbg.aglib.arch.ptrsize):
                        expected.next = data_section_offset + i
                        if bytes(expected) == data_section_data[i : i + malloc_state_size]:
                            # This also might be a false positive, but it is very unlikely too, so should also be fine :)
                            self._main_arena_addr = data_section_address + i
                            break
```

pwndbg/aglib/heap/glibc.py
kozabrada123: You're quite right; it's also weird that the implementation fetched `mp` just to refer to its type (which can be accessed with just `malloc_par)`; I've changed it to:

```diff
     def has_tcache(self) -> bool:
-
         # tcache_bins was renamed to tcache_small_bins in GLIBC 2.42
-        if self.mp is not None and len(self.mp.type.keys()) > 0:
-            return any(
-                x in self.mp.type.keys()  # noqa: SIM118 (mp is not a dict)
-                for x in ("tcache_bins", "tcache_small_bins")
-            )
-
-        # There is no debug symbols, we determine the tcache_bins existence by checking glibc version only
-        return self.is_initialized() and pwndbg.libc.version() >= (2, 26)
+        return any(
+            x in self.malloc_par.keys()  # noqa: SIM118 (mp is not a dict)
+            for x in ("tcache_bins", "tcache_small_bins")
+        )
```

pwndbg/aglib/heap/glibc.py
kozabrada123: Changed to
```python
    def _get_heap_range(self) -> range:
        """Get the start & end of the [heap] mapping"""
        arena = self.thread_arena
        if not arena:
            page = self._get_heap_page()
            assert page is not None
            return range(page.start, page.end)
        return range(arena.active_heap.start, arena.active_heap.end)
```

k4lizen:
(no rush just lmk when it's ready for rereview)

